PR #10721, opened 2021-03-25
Additional swaps network support

danjm:
Swaps now works on another network.


shared/constants/swaps.js
```diff
@@ -10,6 +10,8 @@ export const SWAPS_FETCH_ORDER_CONFLICT = 'swaps-fetch-order-conflict';
 // An address that the metaswap-api recognizes as ETH, in place of the token address that ERC-20 tokens have
 const ETH_SWAPS_TOKEN_ADDRESS = '0x0000000000000000000000000000000000000000';
 
+const BNB_SWAPS_TOKEN_ADDRESS = '0x0000000000000000000000000000000000000000';
```
Gudahtt, line 13: Nit: Maybe instead of adding an additional constant, we could rename the old one to no longer imply that it's network-specific? It doesn't seem network-specific.

Gudahtt (dismissed):
LGTM!

shared/constants/swaps.js
danjm: good suggestion. Done in 2ee45891e

Gudahtt (approved):

PR #10722, opened 2021-03-25
[skip e2e] Update changelog for v9.2.1

danjm:


CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## Current Develop Branch
 
 ## 9.2.1 Thu Mar 25 2021
+- [#10692](https://github.com/MetaMask/metamask-extension/pull/10692): Prevent users from hitting the missing template error
```
Gudahtt, line 6: This might be a bit clearer:
```suggestion
- [#10692](https://github.com/MetaMask/metamask-extension/pull/10692): Prevent UI crash when a 'wallet_requestPermissions" confirmation is queued behind a "wallet_addEthereumChain" confirmation
```

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
...
+- [#10712](https://github.com/MetaMask/metamask-extension/pull/10712): Fix infinite spinner when attempting an approve transaction for some token contracts
```
Gudahtt, line 7: "some token contracts" seems a bit vague. I thought this applied to tokens with empty symbols - is there more to it than that? :thinking: 
```suggestion
- [#10712](https://github.com/MetaMask/metamask-extension/pull/10712): Fix infinite spinner when attempting an approve transaction for token with empty symbol
```

Gudahtt: cc @brad-decker

danjm: I'm going to push a more precise message

Gudahtt: That doesn't sound right :thinking: I thought this error occurred when we did successfully retrieve the token symbol, but it was an empty string. That's what the PR description implies as well.

danjm: This issue occurs when the `fetchSymbolAndDecimals` method returns the `DEFAULT_SYMBOL`. See https://github.com/MetaMask/metamask-extension/blob/develop/ui/app/helpers/utils/token-util.js#L81

This happens when the request for the symbol fails. It could also happen when the request for the symbol successfully returns an empty string, but I am not sure that we have ever seen that in practice.

Gudahtt (approved):
LGTM!

PR #10724, opened 2021-03-25
Workaround to fix lag/freeze on external monitors

### LINKED ISSUE #10173: Mitigate CSS transition lag Chromium bug on macOS external monitors
There is a [Chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=971701) that causes extreme lag in CSS transitions on external monitors on macOS. There don't seem to be any great workarounds, other than manually forcing the page to re-draw on an interval or avoiding CSS transitions altogether.

This issue is meant to consolidate efforts to mitigate or workaround this issue, since it has been brought up many times as different symptoms have been reported. I'll mark it as "blocked" for now, unless someone comes up with a viable mitigation at least.

Related issues:
* #8324
* #10202

igorbarbashin:
Fixes: #10173 

Explanation:  
Add a bit of CSS that renders a visually invisible rotating 1px element that forces the UI to update on each tick

Manual testing steps:  
  - Move Chrome window to an external monitor
  - Open Metamask popup and type in any input (password or a send amount)
  - See the UI updating after each keystroke

Here's a comparison before the fix: (UI only refreshes after a mouse move)
before-fix

After the fix: (UI updates on each keystroke)
after-fix

### DIFF
--- a/ui/app/css/itcss/components/newui-sections.scss
+++ b/ui/app/css/itcss/components/newui-sections.scss
@@ -14,6 +14,47 @@ $sub-mid-size-breakpoint-range: "screen and (min-width: #{$break-large}) and (ma
   align-items: center;
 }
 
+// Fix for UI lag on external monitor: https://github.com/MetaMask/metamask-extension/issues/10173
+.app::after {
+  content: "";
+  position: fixed;
+  top: 0;
+  left: 0;
+  width: 1px;
+  height: 1px;
+  background-color: $Grey-000;
+  -webkit-animation-name: emptySpinningDiv;
+  animation-name: emptySpinningDiv;
+  -webkit-animation-duration: 2s;
+  animation-duration: 2s;
+  -webkit-animation-iteration-count: infinite;
+  animation-iteration-count: infinite;
+  -webkit-animation-timing-function: linear;
+  animation-timing-function: linear;
+}
+
+@-webkit-keyframes emptySpinningDiv {
+  0% {
+    -webkit-transform: rotate(0deg);
+    transform: rotate(0deg);
+  }
+  to {
+    -webkit-transform: rotate(1turn);
+    transform: rotate(1turn);
+  }
+}
+
+@keyframes emptySpinningDiv {
+  0% {
+    -webkit-transform: rotate(0deg);
+    transform: rotate(0deg);
+  }
+  to {
+    -webkit-transform: rotate(1turn);
+    transform: rotate(1turn);
+  }
+}
+
 // Main container
 .main-container {
   z-index: $main-container-z-index;
--- a/ui/app/css/itcss/components/newui-sections.scss
+++ b/ui/app/css/itcss/components/newui-sections.scss
@@ -23,10 +23,7 @@ $sub-mid-size-breakpoint-range: "screen and (min-width: #{$break-large}) and (ma
   width: 1px;
   height: 1px;
   background-color: $Grey-000;
-  animation-name: emptySpinningDiv;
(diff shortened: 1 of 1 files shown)


igorbarbashin:

ui/app/css/itcss/components/newui-sections.scss
```diff
@@ -14,6 +14,47 @@ $sub-mid-size-breakpoint-range: "screen and (min-width: #{$break-large}) and (ma
...
+// Fix for UI lag on external monitor: https://github.com/MetaMask/metamask-extension/issues/10173
+.app::after {
+  content: "";
+  position: fixed;
+  top: 0;
+  left: 0;
+  width: 1px;
+  height: 1px;
+  background-color: $Grey-000;
+  -webkit-animation-name: emptySpinningDiv;
```
Gudahtt, line 26: I suspect that these prefixed entries aren't required, because we use `gulp-autoprefixer` when building our styles.

Gudahtt (commented):
Thanks, this looks great! I had one suggestion on simplifying these style rules, but aside from that they look good.

I'd like to get some sense of the performance impact of this first before merging, particularly on underpowered devices. I can test that sometime tomorrow.

It would be great to only use this on macOS and Chrome specifically, and skip it for other device/browser combinations, so that we can avoid burdening them with this invisible spinner. I'm not sure what the performance impact of this is but surely it's non-zero. That can be done in a follow-up PR though.

igorbarbashin:
Good point @Gudahtt. Indeed looks like it's a mac-only issue and it does make sense to detect a mac before adding this piece of code.
Do you know if mac detection is used anywhere in the code already?

ui/app/css/itcss/components/newui-sections.scss
igorbarbashin: Spot on. I wasn't sure if autoprefixer was implemented. Found other `-webkit-` rules in other css files and decided to keep it.
Gonna update

igorbarbashin:
Though I kinda like that it's a CSS-only solution and doesn't introduce any new points of failure

igorbarbashin:
Updated to make the css footprint a bit smaller

igorbarbashin:
@Gudahtt any luck testing it?
Is there anything I can help with?

igorbarbashin:
@darkwing please let me know what do you think of this PR

Gudahtt:
Sorry for the delay @igorbarbashin. Within a few days I'll put up a PR to handle the browser and OS detection, I had some idea how to do that. I'll find someone else on the team to test this on macOS as well.

igorbarbashin:
@Gudahtt, no worries.

Here's the simplest mac detection I found:

`const isMac = navigator.userAgent.includes('Macintosh')`

I'll see where it makes sense to add it to the codebase. But let me know if know exactly where it makes sense

Gudahtt:
I'm not sure how reliable the user-agent is :thinking: I'd rather that we rely upon the extension API if possible. There is a [`getPlatformInfo` function](https://developer.chrome.com/docs/extensions/reference/runtime/#method-getPlatformInfo) that should reliably tell us which operating system family is being used.

As for the browser, we create separate builds for each browser. We can [modify our build system to inject an environment variable](https://github.com/MetaMask/metamask-extension/blob/develop/development/build/scripts.js#L419) representing the current browser, and use that to ensure this class is only used on Chrome.

darkwing:
Superceded by https://github.com/MetaMask/metamask-extension/pull/10952

Thank you @igorbarbashin

PR #10727, opened 2021-03-25
Fix 10725 - Make 'forget device' work again

### LINKED ISSUE #10725: "Forget Device" link on Connect Hardware Wallet page doesn't work
**Describe the bug**
When connecting a hardware wallet, once connected, the page lists accounts and also has a "Forget Device" link.  This link doesn't appear to be working.  Upon inspection with @Gudahtt , this was an issue that existed before our 9.2.0 update, so it's not a result of the multiple hardware wallet feature.

1. From the main menu, click "Connect Hardware Wallet"
2. Once connected, connect a few accounts
3. Click "Connect Hardware Wallet"
4. Click "Forget device"

**Expected behavior**
I would expect all accounts connected to that device be removed.  The bug is that no accounts are removed.

darkwing:
Fixes: #10725

Explanation:  

The call to `keyring.forgetDevice` wasn't sufficient in remove accounts from the keyringController and other important keyring properties.  Shout out to @brad-decker for the help!

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -1306,6 +1306,10 @@ export default class MetamaskController extends EventEmitter {
    */
   async forgetDevice(deviceName) {
     const keyring = await this.getKeyringForDevice(deviceName);
+    const accounts = await keyring.getAccounts();
+    await Promise.all(
+      accounts.map((account) => this.removeAccount(account.toLowerCase())),
+    );
     keyring.forgetDevice();
     return true;
   }


brad-decker (dismissed):
LGTM!

Gudahtt (commented):
What happens to the identity state when this happens? The `identities` object and the current selected account.

We should ensure we don't have leftover entries in `identities` pointing at these removed accounts, and that the current selected account is updated.

PR #10732, opened 2021-03-26
Lavamoat for webapp - UI

kumavis:
applying lavamoat to webapp ui.js only


kumavis:
will revisit this
pr for background here https://github.com/MetaMask/metamask-extension/pull/12242

PR #10735, opened 2021-03-26
ci - cache deps before patch-package

### LINKED ISSUE #10734: bug: CI - prep deps - we're caching the node_modules after patch
Build - we're caching the `node_modules` after patch, then recovering them, then apply the patch again. this sometimes fails

kumavis:
Fixes: https://github.com/MetaMask/metamask-extension/issues/10734

we are accidentally caching the patched package and then patching them again

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -135,13 +135,13 @@ jobs:
     steps:
       - checkout
       - restore_cache:
-          key: dependency-cache-{{ checksum "yarn.lock" }}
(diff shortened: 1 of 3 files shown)


EtDu (dismissed):
/sparkle

kumavis:
cache is bad for branches that are sensitive to this bug

EtDu (approved):
new cache :)

Gudahtt (approved):
LGTM!

PR #10738, opened 2021-03-26
eth-json-rpc-middleware@8.0.0

rekmarks:
We're bumping from `^6` to `^8`. All imports are now named, and they have been updated. This is a breaking change, in that support for `eth_signTransaction` is added in `^8.0.0`. We do not support this method in our UI, so our middleware stack has been instrumented to reject.

In addition, there are some non-breaking behavioral changes in this version that reviewers should be aware of, see the [7.0.0 release](https://github.com/MetaMask/eth-json-rpc-middleware/releases).


darkwing:
@rekmarks Is this still relevant?  What can we do to get this out of draft?

darkwing (approved):

ryanml (approved):

PR #10743, opened 2021-03-26
Use correct block explorer name and link in swaps when on custom network

danjm:
Fixes an issue found by @tmashuang when QAing v9.3.0

Before:

-

After:

-


ui/app/pages/swaps/awaiting-swap/view-on-ether-scan-link/view-on-ether-scan-link.js
```diff
@@ -18,7 +19,7 @@ export default function ViewOnEtherScanLink({
       onClick={() => global.platform.openTab({ url: blockExplorerUrl })}
     >
       {isCustomBlockExplorerUrl
-        ? t('viewOnCustomBlockExplorer', [blockExplorerUrl])
+        ? t('viewOnCustomBlockExplorer', [blockExplorerBaseUrl])
```
Gudahtt, line 22: Instead of passing in a second "base URL" for custom block explorers, it might be better to update `getBlockExplorerUrlForTx` to accept the custom block explorer URL.

The library used by `getBlockExplorerUrlForTx` was [recently updated](https://github.com/MetaMask/etherscan-link/releases/tag/v2.0.0) with new methods that accept a custom block explorer URL, so this should be easy enough to do. Then we'd get a URL directly to the correct transaction, just as on Etherscan.

ui/app/pages/swaps/build-quote/build-quote.js
```diff
@@ -416,11 +418,13 @@ export default function BuildQuote({
...
                         className="build-quote__token-etherscan-link build-quote__underline"
                         key="build-quote-etherscan-link"
-                        href={`https://etherscan.io/token/${selectedToToken.address}`}
+                        href={`${rpcPrefs.blockExplorerUrl}/token/${selectedToToken.address}`}
                         target="_blank"
                         rel="noopener noreferrer"
                       >
-                        {t('etherscan')}
+                        {rpcPrefs.blockExplorerUrl
+                          ? rpcPrefs.blockExplorerUrl
```
Gudahtt, line 426: Hmm - this seems OK for now, but we should come up with some other text to go here other than the URL.

Maybe for now we could at least limit it to the hostname, rather than the full URL? That might look a bit nicer.

```suggestion
                          ? new URL(rpcPrefs.blockExplorerUrl).hostname
```

ui/app/pages/swaps/build-quote/build-quote.js
```diff
@@ -416,11 +418,13 @@ export default function BuildQuote({
                       <a
                         className="build-quote__token-etherscan-link build-quote__underline"
                         key="build-quote-etherscan-link"
-                        href={`https://etherscan.io/token/${selectedToToken.address}`}
+                        href={`${rpcPrefs.blockExplorerUrl}/token/${selectedToToken.address}`}
```
Gudahtt, line 421: Both this URL and the one on like 457 below will be incorrect for some URLs (e.g. anything with a query string, hash, API key, etc.). We should use our `@metamask/etherscan-link` library to generate these instead.

ui/app/pages/swaps/awaiting-swap/view-on-ether-scan-link/view-on-ether-scan-link.js
danjm: The purpose of this was to show a shorter url, as the link already correctly goes to the correct transaction. Based on your other comment, maybe this would be best here:
`t('viewOnCustomBlockExplorer', [new URL(blockExplorerUrl).hostname])`

ui/app/pages/swaps/build-quote/build-quote.js
danjm: Done in 7a5b0dd92

ui/app/pages/swaps/build-quote/build-quote.js
danjm: Done in 7a5b0dd92

ui/app/pages/swaps/awaiting-swap/view-on-ether-scan-link/view-on-ether-scan-link.js
Gudahtt: Oops, I see, I misread. Yes, perhaps the hostname would be better here.

ui/app/pages/swaps/build-quote/build-quote.js
```diff
@@ -215,6 +218,11 @@ export default function BuildQuote({
     );
   };
 
+  const tokenLinkForChain = createTokenTrackerLinkForChain(
```
Gudahtt, line 221: This will only work for the Infura networks. You'll need to use `createCustomTokenTrackerLink` if on a custom network.

danjm: fixed in 82eb5364b

ui/app/pages/swaps/awaiting-swap/view-on-ether-scan-link/view-on-ether-scan-link.js
danjm: Done in c42554a2d

Gudahtt: This still appears to be here. Did you forgot to push?

danjm: oops, pushed now

ui/app/pages/swaps/build-quote/build-quote.js
Gudahtt: I think you may have misunderstood - I meant that `createCustomTokenTrackerLink` is needed for custom networks, but `createTokenTrackerLinkForChain` is still needed for Infura networks.

Also the wrong parameters are now being passed into `createCustomTokenTrackerLink`

danjm: Ah, sorry for the mix up. Part of my confusion was that we were still on etherscan-link v1.5.0 and I was looking at the readme  in the node modules directory.

To use `createCustomTokenTrackerLink` we need etherscan-link v2.0.0. I created a new PR to add that #10747, and for now am basing this branch off of that.

And I've added a commit which ensures that the correct links are created for the current chain 60f1d347c

video of it working on mainnet, bsc and testnet is here: https://streamable.com/jkn155

ui/app/pages/swaps/build-quote/build-quote.js
```diff
@@ -215,6 +223,13 @@ export default function BuildQuote({
     );
   };
 
+  const tokenLinkForChain = isMainnet
```
Gudahtt, line 226: It would be better to use `rpcPrefs.blockExplorerURl` here, just like for the text portions. This would break on testnets, which don't have a custom block explorer URI.

And actually, I guess this would break on any custom network without a custom block explorer as well :thinking: 

I guess the ideal logic would be:

* If there is a custom block explorer, use that
* Else if the network is among the Infura networks, use Etherscan
* Else hide the explorer link - there is nothing to link to.

ui/app/pages/swaps/build-quote/build-quote.js
Gudahtt: Ah, whoops, didn't realize it wasn't updated yet

ui/app/pages/swaps/build-quote/build-quote.js
danjm: Good points. Also, now I'm realizing we may want to default to https://bscscan.com for the binance chain, if no  `rpcPrefs.blockExplorerURl` is provided. This is the recommended equivalent of etherscan for that chain. It could be added to `shared/constants/swaps.js`. I'll check with product people on this.

danjm: @Gudahtt a731fc97d applies logic very similar to what you suggested above. First it checks if there is a custom block explorer url, then it checks if the current chain is explicitly supported in `shared/constants/swaps.js`, then if it is among the etherscan networks, it uses etherscan. The same logic is also applied in awaiting-swap.js

On the build-quote page I decided to hide the link if there is nothing to link to, as it fits properly with the intention of this fix and the overall feature of supporting a custom networks for swaps. Right now, this situation can only happen on a testnet, but I think it is better that we don't ship broken links.

shared/constants/swaps.js
```diff
@@ -74,3 +76,7 @@ export const SWAPS_CHAINID_DEFAULT_TOKEN_MAP = {
   [SWAPS_TESTNET_CHAIN_ID]: TEST_ETH_SWAPS_TOKEN_OBJECT,
   [BNB_CHAIN_ID]: BNB_SWAPS_TOKEN_OBJECT,
 };
+
+export const SWAPS_CHAINID_DEFAULT_BLOCK_EXPLORER_URL_MAP = {
+  [BNB_CHAIN_ID]: BSC_DEFAULT_BLOCK_EXPLORER_URL,
```
Gudahtt, line 81: This default should probably not be swaps-specific. We could use this as a default custom block explorer everywhere. We can adjust that later though.

package.json
```diff
@@ -87,7 +87,7 @@
     "@metamask/controllers": "^6.2.1",
     "@metamask/eth-ledger-bridge-keyring": "^0.3.0",
     "@metamask/eth-token-tracker": "^3.0.1",
-    "@metamask/etherscan-link": "^1.5.0",
+    "@metamask/etherscan-link": "^2.0.0",
```
Gudahtt, line 90: I guess this PR needs a rebase - this has already been merged separately.

ui/app/pages/swaps/awaiting-swap/awaiting-swap.js
```diff
@@ -105,8 +108,21 @@ export default function AwaitingSwap({
     category: 'swaps',
   });
 
-  const blockExplorerUrl =
-    txHash && getBlockExplorerUrlForTx({ chainId, hash: txHash }, rpcPrefs);
+  let blockExplorerUrl;
+  if (txHash && rpcPrefs.blockExplorerUrl) {
+    blockExplorerUrl = getBlockExplorerUrlForTx({ hash: txHash }, rpcPrefs);
```
Gudahtt, line 113: It looks like `getBlockExplorerUrlForTx` has its own means of generating URLs for custom networks, and it's substantially more error prone than the method used by `@metamask/etherscan-link`. But it should work for these two cases at least. We can improve this separately.

app/_locales/en/messages.json
```diff
@@ -1937,7 +1937,12 @@
     "message": "Using the best quote"
   },
   "swapVerifyTokenExplanation": {
-    "message": "Multiple tokens can use the same name and symbol. Check Etherscan to verify this is the token you're looking for."
+    "message": "Multiple tokens can use the same name and symbol. Check Etherscan to verify this is the token you're looking for.",
+    "description": "This appears in a tooltip next to the verifyThisTokenOn message. It gives the user more information about why they should check the token on a block explorer. This is the message that appears when we want to direct the user to etherscan."
+  },
+  "swapVerifyTokenOnBlockExplorerExplanation": {
+    "message": "Multiple tokens can use the same name and symbol. Check $1 to verify this is the token you're looking for.",
```
Gudahtt, line 1944: Hmm. Could we take this same approach with each other locale too by substituting `Etherscan` for `$1`, which would be `Etherscan` when Etherscan was the block explorer being used? That would ensure we have _something_ understandable for this message in each locale at least, and we know that the sentences are all gramatically correct with proper noun where 'Etherscan' is.

I guess it's possible something about the sentence could change depending on the input, but.... I don't know, that seems difficult to handle perfectly for all cases even if we wanted to.

If we consolidate these two messages, we can also simplify the logic used in `build-quote.js`, because the tooltip text would be the same for each branch.

ui/app/pages/swaps/awaiting-swap/awaiting-swap.js
```diff
@@ -105,8 +108,21 @@ export default function AwaitingSwap({
...
+  } else if (txHash && SWAPS_CHAINID_DEFAULT_BLOCK_EXPLORER_URL_MAP[chainId]) {
+    blockExplorerUrl = createCustomExplorerLink(
+      txHash,
+      SWAPS_CHAINID_DEFAULT_BLOCK_EXPLORER_URL_MAP[chainId],
+    );
+  } else if (txHash) {
```
Gudahtt, line 119: This will use the wrong link for custom networks without a block explorer set.

Not a blocker for this PR, since it's a pre-existing problem, but not ideal. We should at least capture this somewhere so we can fix it later.

shared/constants/swaps.js
Gudahtt: We should really consolidate this link/label generation logic in a utility function or hook, so that it's easier to add more special cases like this everywhere at once.

Unless we wanted to consider making this block explorer the default in the network form itself, and adding it via a migration for anyone that already has this network set. :thinking:

ui/app/pages/swaps/awaiting-swap/awaiting-swap.js
danjm: Ticket created here: https://github.com/MetaMask/metamask-extension/issues/10749

app/_locales/en/messages.json
danjm: Yeah, I think this is a good idea. Done in 1cfbe6c5e

ui/app/pages/swaps/awaiting-swap/awaiting-swap.js
danjm: This is an easy fix and prevents shipping broken links for the test network. So the fix has been added in bcd4b0aee

shared/constants/swaps.js
danjm: I created a ticket for this here https://github.com/MetaMask/metamask-extension/issues/10750

Gudahtt (approved):
LGTM!

PR #10744, opened 2021-03-26
Ensure swaps gas prices are fetched from the correct chain specific endpoint

danjm:
Fixes an issue found by @tmashuang when QAing v9.3.0

The chainId was not being passed to the `fetchSwapsGasPrices` function, resulting in the incorrect gas prices being fetched.

### DIFF
--- a/ui/app/ducks/swaps/swaps.js
+++ b/ui/app/ducks/swaps/swaps.js
@@ -747,6 +747,7 @@ export const signAndSendTransactions = (history, metaMetricsEvent) => {
 export function fetchMetaSwapsGasPriceEstimates() {
   return async (dispatch, getState) => {
     const state = getState();
+    const chainId = getCurrentChainId(state);
     const priceEstimatesLastRetrieved = getSwapsPriceEstimatesLastRetrieved(
       state,
     );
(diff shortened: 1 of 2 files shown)


ui/app/ducks/swaps/swaps.js
```diff
@@ -760,12 +761,13 @@ export function fetchMetaSwapsGasPriceEstimates() {
     let priceEstimates;
     try {
       if (Date.now() - timeLastRetrieved > 30000) {
-        priceEstimates = await fetchSwapsGasPrices();
+        priceEstimates = await fetchSwapsGasPrices(chainId);
       } else {
         const cachedPriceEstimates = await getStorageItem(
           'METASWAP_GAS_PRICE_ESTIMATES',
```
Gudahtt, line 767: This cache isn't chainId-aware either. It needs to be, otherwise cached values from the wrong chain will end up being used.

danjm: I thought it was chainId aware, to the extent that a different url will be used on a different chain. `fetchWithCache` keys the local storage cache by `const cacheKey = `cachedFetch:${url}`;`, so a different url's for different chains would get different values from the cache, or trigger fresh requests if no key matching the url is already in the cache.

Gudahtt: Right, but look at this line though. `getStorageItem` is only being passed `METASWAP_GAS_PRICE_ESTIMATES`. No URL, no chainId.

This wasn't cached by ChainID before because it used to only be used for one.

Gudahtt: Perhaps the best solution here would be to get rid of `METASWAP_GAS_PRICE_ESTIMATES` and `METASWAP_GAS_PRICE_ESTIMATES_LAST_RETRIEVED`, and rely upon `fetchWithCache` for our cache.

That would require passing through the desired cache interval to `fetchSwapsGasPrices, or changing it to be `30000`, if that's what's desired (that's what we're using now).

Gudahtt (approved):
LGTM!

PR #10745, opened 2021-03-27
Remove all uses of createAsyncMiddleware

rekmarks:
After I lost hours to a bug in a `createAsyncMiddleware` middleware — due to an unsafe pattern that it forces the consumer to use — I realized that `createAsyncMiddleware` is superfluous. Free your mind; delete `createAsyncMiddleware`.


Gudahtt:
How are errors handled in middleware? With the async middleware, thrown errors would be captured and handled as errors. I believe switching back to a normal callback would require wrapping the entire middleware body in a `try .. catch` block to preserve the equivalent behaviour.

Maybe we should put this on ice until the changes to "normal" middleware made in this PR have been merged and published: https://github.com/MetaMask/json-rpc-engine/pull/81/files

darkwing:
@rekmarks Is this still relevant?  What can we do to get this out of draft?

PR #10753, opened 2021-03-29
Ensure that the correct default currency symbols are used for fees on the view quote screen

danjm:
This PR fixes an issues found by @tmashuang while reviewing v9.3.0

It ensures that fees on the view quote screen - in both the fees card and in the gas modal - use the correct currency symbol. On mainnet this will be ETH. On custom networks, this will be the currency symbol set by the user, or if no currency symbol has been set, it will be the symbol of the default token for that network.

Demo video is here: https://www.veed.io/download/a2f4f25e-f0b6-4658-8909-4f48730f1890


brad-decker (approved):
LGTM.

PR #10763, opened 2021-03-29
Ensure that the approval fee in the swaps custom gas modal is in network specific currency

danjm:
Addresses an issue found by @tmashuang while QAing v9.3.0

The `extraInfoRow.value` property passed from view-quote.js to the custom gas modal was using the `amountInEth` translation, meaning that the approve fee shown in the modal always had "ETH" as the currency symbol instead of the current networks symbol (e.g. BNB). This PR corrects that.

After:

Screenshot from 2021-03-29 21-25-10

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -143,10 +143,6 @@
   "amount": {
     "message": "Amount"
   },
-  "amountInEth": {
-    "message": "$1 ETH",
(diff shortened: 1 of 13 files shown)


darkwing (approved):
I can confirm that the BNB and ETH labels are displaying properly in both mainnet and BNB swaps!

PR #10767, opened 2021-03-30
Handling gas price fetch failure 

### LINKED ISSUE #10556: TypeError: Failed to fetch (gasPrices api on the confirmation page)
Sentry Issue: [METAMASK-3H10](https://sentry.io/organizations/metamask/issues/749961265/?referrer=github_integration)

```
TypeError: Failed to fetch
```

This is coming from the calls of `fetchBasicGasEstimates` and ultimately `basicGasPriceQuery` which are defined in `gas.duck.js`

The solution will require at least three pieces:
- a fallback when `fetchBasicGasEstimates` fails to `eth_gasPrice`
- ensure the UI appropriately responds to the fallback
- handle the case when both `fetchBasicGasEstimates` fails to `eth_gasPrice` fail, which will require both gas price logic and UI updates

NiranjanaBinoy:
Fixes: #10556 

Explanation:  Currently when the gas estimate service is unavailable, the gas row send and confirm page will go into infinite loading loop.

**Changes introduced:**
- Advanced gas input will be displayed in case the main service fails, eth_gasprice will be fetched and the submission buttons will be active.
- In cases where both main service eth_gasprice and main service fails, Advanced gas input will be displayed, gasPrice displayed will be 0 and the primary button `next` button in the send flow or `confirm` button in the confirmation flow will be disabled.

Manual testing steps:  

**Gas estimation service failing and eth_gasprice fetched:**

You can mimmic the gas price fetch failure by replacing `basicGasPriceQuery` call in `fetchExternalBasicGasEstimates` by `Promise.reject()`

**_Send Page:_**
Screen Shot 2021-04-22 at 4 37 27 PM

**_Confirmation page:_** 

https://user-images.githubusercontent.com/43930900/115785726-d7baec80-a38d-11eb-8d47-66a295280082.mov

**Both the gas estimation services failed:**

You can mimmic the gas price fetch failure by replacing `global.eth.gasPrice()` call in `fetchEthGasPriceEstimates` by `Promise.reject()`;

**_Send Page:_**
Screen Shot 2021-04-22 at 4 19 15 PM

**_Confirmation page:_**

https://user-images.githubusercontent.com/43930900/115785732-d8ec1980-a38d-11eb-9655-a8ff5c194b11.mov

### DIFF
--- a/ui/app/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.component.js
+++ b/ui/app/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.component.js
@@ -123,19 +123,25 @@ export default class GasModalPageContainer extends Component {
       infoRowProps: { newTotalFiat, newTotalEth, sendAmount, transactionFee },
     } = this.props;
 
-    let tabsToRender = [
-      {
-        name: this.context.t('basic'),
-        content: this.renderBasicTabContent(gasPriceButtonGroupProps),
-      },
-      {
-        name: this.context.t('advanced'),
-        content: this.renderAdvancedTabContent(),
-      },
-    ];
-
+    let tabsToRender;
     if (hideBasic) {
-      tabsToRender = tabsToRender.slice(1);
+      tabsToRender = [
+        {
+          name: this.context.t('advanced'),
+          content: this.renderAdvancedTabContent(),
+        },
+      ];
+    } else {
+      tabsToRender = [
+        {
+          name: this.context.t('basic'),
+          content: this.renderBasicTabContent(gasPriceButtonGroupProps),
+        },
+        {
+          name: this.context.t('advanced'),
+          content: this.renderAdvancedTabContent(),
+        },
+      ];
     }
 
     return (
--- a/ui/app/ducks/gas/gas.duck.js
+++ b/ui/app/ducks/gas/gas.duck.js
@@ -8,17 +8,24 @@ import {
 import { getIsMainnet, getCurrentChainId } from '../../selectors';
 import fetchWithCache from '../../helpers/utils/fetch-with-cache';
 
+const BASIC_ESTIMATE_STATES = {
+  LOADING: 'LOADING',
+  FAILED: 'FAILED',
+  READY: 'READY',
+};
+
+const GAS_SOURCE = {
+  METASWAPS: 'MetaSwaps',
+  ETHGASPRICE: 'eth_gasprice',
+};
+
 // Actions
-const BASIC_GAS_ESTIMATE_LOADING_FINISHED =
-  'metamask/gas/BASIC_GAS_ESTIMATE_LOADING_FINISHED';
-const BASIC_GAS_ESTIMATE_LOADING_STARTED =
-  'metamask/gas/BASIC_GAS_ESTIMATE_LOADING_STARTED';
+const BASIC_GAS_ESTIMATE_STATUS = 'metamask/gas/BASIC_GAS_ESTIMATE_STATUS';
 const RESET_CUSTOM_DATA = 'metamask/gas/RESET_CUSTOM_DATA';
 const SET_BASIC_GAS_ESTIMATE_DATA = 'metamask/gas/SET_BASIC_GAS_ESTIMATE_DATA';
 const SET_CUSTOM_GAS_LIMIT = 'metamask/gas/SET_CUSTOM_GAS_LIMIT';
 const SET_CUSTOM_GAS_PRICE = 'metamask/gas/SET_CUSTOM_GAS_PRICE';
-const SET_IS_ETH_GAS_PRICE_FETCHED =
-  'metamask/gas/SET_IS_ETH_GAS_PRICE_FETCHED';
+const SET_ESTIMATE_SOURCE = 'metamask/gas/SET_ESTIMATE_SOURCE';
 
 const initState = {
   customData: {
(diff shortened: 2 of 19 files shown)


ui/app/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.component.js
```diff
@@ -123,19 +123,25 @@ export default class GasModalPageContainer extends Component {
...
-        name: this.context.t('basic'),
-        content: this.renderBasicTabContent(gasPriceButtonGroupProps),
-      },
-      {
-        name: this.context.t('advanced'),
-        content: this.renderAdvancedTabContent(),
-      },
-    ];
-
+    let tabsToRender;
```
brad-decker, line 126: [Q] This change looks like a good one, as it prevents calling the render method of the basic tab content entirely when hideBasic is true. Was that the intent here? does this serve a purpose in the new error message?

ui/app/ducks/gas/gas.duck.js
```diff
@@ -29,6 +31,7 @@ const initState = {
     fast: null,
   },
   basicEstimateIsLoading: true,
+  isEthGasPriceFetched: false,
```
brad-decker, line 34: It seems that this introduces a competing state flag with 'basicEstimateIsLoading'. What I mean to say here is that in order to properly update state we have to check the difference between two keys instead of referencing one key.

What if we refactored the gas slice state to be more like:
```
basicEstimateStatus: 'LOADING' | 'READY' | 'FAILED';
```

In this way we can check one key and branch from the possible states instead of storing the 'status' of the state slice across a few different keys.

brad-decker: As i'm reading more through the logic below I am wondering if this key doesn't actually mean what I think it does. It seems its only set to true if fetchExternalBasicGasEtimates fails and fetchEthGasPriceEstimates succeeds. Could you help me understand a little more?

ui/app/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.component.js
NiranjanaBinoy: Originally, I introduced this change because when the gas fetch fails without any backup estimate, the `basicEstimates` value will be null and calling `renderBasicTabContent` will crash the extension even if hideBasic is true. But later I removed the `Advanced option` from send page and `edit` for gas row from confirmation page in cases of failures `advanced input option` is already present in the pages , currently only the approve-token page will have the `edit ` option, which calls the gas modal. (this realization came to be when I started replying to this message 😓)

Also when the gas fetch fails without backup, we are disabling submission from the corresponding page, in our test dapp the user cannot reach the approve token page without successfully completing create token, if  that is the scenario everywhere we can maybe remove this changes and use the old one.

I think preventing the rendering of basic tab when hideBasic is true might be a better option as it will results in application crashing if in case the basicEstimates is null, unless there is any specific reason to add both the tabs and remove the `basic tab` in case `hideBasic` is true previously.

ui/app/ducks/gas/gas.duck.js
NiranjanaBinoy: In mainnet, `isEthGasPriceFetched` flag is set to `true` only when the original [gas estimation service](https://api.metaswap.codefi.network/gasPrices) fails, but eth_gasprice is available (`fetchEthGasPriceEstimates` is called), and a warning message will be displayed based on this flag. In testnet the flag will be false, as testnets are not affected by the failure of this service.

When both the [gas estimation service](https://api.metaswap.codefi.network/gasPrices) and eth_gasprice fails, the basicEstimates being null is used to identify the condition, and the corresponding changes affects both mainnet and testnets.

brad-decker: Alright so I might suggest the following to avoid using `null` as a logic check:

```
basicEstimateStatus: 'LOADING' | 'READY' | 'FAILED',
estimateSource: 'MetaSwaps' | 'eth_gasprice',
```

Now you can check for failed state by `basicEstimateStatus === 'FAILED'` and the source for your error message by `estimateSource === 'eth_gasprice'` and the code will be easier to maintain going forward. Thoughts on that?

NiranjanaBinoy: I have updated the PR based on these comments.

brad-decker (changes requested):

ui/app/selectors/custom-gas.js
```diff
@@ -355,3 +357,20 @@ export function getRenderableEstimateDataForSmallButtonsFromGWEI(state) {
     },
   ];
 }
+
+export function getIsEthGasPriceFetched(state) {
+  const gasState = state.gas;
+  return Boolean(
+    gasState.estimateSource === 'eth_gasprice' &&
+      gasState.basicEstimateStatus === 'READY' &&
+      getIsMainnet(state),
```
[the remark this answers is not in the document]
NiranjanaBinoy: My intent here was to make sure that the estimates fetched using eth_gasprice in this scenario are from Mainnet and not from the Testnets. Because we are displaying a warning to the user when we fetch gas price from eth_gasprice in Mainnet when Metaswaps service fails, this warning is not needed in the case of Testnets.

ui/app/selectors/custom-gas.js
```diff
@@ -355,3 +357,20 @@ export function getRenderableEstimateDataForSmallButtonsFromGWEI(state) {
...
+  );
+}
+
+export function getNoGasPriceFetched(state) {
+  const gasState = state.gas;
+  return Boolean(
+    gasState.estimateSource === 'eth_gasprice' &&
+      gasState.basicEstimateStatus === 'FAILED',
```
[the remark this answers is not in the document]
NiranjanaBinoy: Yea 😅

ui/app/pages/send/send-footer/send-footer.component.js
```diff
@@ -27,6 +27,7 @@ export default class SendFooter extends Component {
     gasEstimateType: PropTypes.string,
     gasIsLoading: PropTypes.bool,
     mostRecentOverviewPage: PropTypes.string.isRequired,
+    noGasPrice: PropTypes.bool,
```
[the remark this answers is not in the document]
NiranjanaBinoy: 'gasIsLoading' seems to be from `state.appState` and it is based on not just the gas price estimate fetch, whereas `noGasPrice` is based on the status of gas estimation fetch alone. I am confused on why we have 'gasIsLoading' when we also had  `basicGasEstimatesLoadingStarted`  and `basicGasEstimatesLoadingFinished` from `state.gas`!! 🤔

ui/app/selectors/custom-gas.js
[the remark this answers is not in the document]
NiranjanaBinoy: change updated in the PR

ui/app/pages/send/send-footer/send-footer.component.js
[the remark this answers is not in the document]
brad-decker: educating myself and will follow up with you shortly!

noncash15 (approved):

brad-decker (approved):
LGTM! future work that I would suggest here is more of a critique of our infrastructure and not this feature.

PR #10769, opened 2021-03-30
Build - refactor background process to use html

kumavis:
aligns background process entry point to match format of the other entry points (ui/home, ui/popup, ...). this will enable the "flat file" build refactor

additonally, instead of logic for adding/removing scripts from the build, i opted for copying empty files to replace the expected files


development/build/manifest.js
```diff
@@ -37,27 +33,11 @@ function createManifestTasks({ browserPlatforms }) {
 
   // dev: remove bg-libs, add chromereload, add perms
```
Gudahtt, line 34: Nit: Comment could use an update

```suggestion
  // dev: add perms
```

development/build/manifest.js
```diff
@@ -37,27 +33,11 @@ function createManifestTasks({ browserPlatforms }) {
...
-    );
-    scripts.push('chromereload.js');
-    manifest.background = {
-      ...manifest.background,
-      scripts,
-    };
     manifest.permissions = [...manifest.permissions, 'webRequestBlocking'];
   });
 
   // testDev: remove bg-libs, add perms
```
Gudahtt, line 39:
```suggestion
  // testDev: add perms
```

development/build/static.js
```diff
@@ -80,13 +80,34 @@ const copyTargetsDev = [
     pattern: '/chromereload.js',
     dest: ``,
   },
+  // empty files to suppress missing file errors
```
Gudahtt, line 83: Huh, that's an interesting idea!

I had been meaning to address this warning by updating the build script to use these dependency bundles in development too, rather than just in prod. But I never got around to it. This seems like a simpler solution for now.

Gudahtt (dismissed):
LGTM! Made a couple of corrections to comments, but aside from that everything seems good.

ryanml (dismissed):
++ (discussed on Discord, but for now the `empty.js` file workaround is preferable to the alternative, which can be accomplished in a separate task) 

@Gudahtt left some nits, but ran through smoke tests on both `dev` and `prod` builds and all looked good to me!

development/build/static.js
kumavis: yeah this solution is sort of dumb and silly but saves a good chunk of code dedicated to conditionally flipping on and off script imports. simplicity won me over here.

ryanml (approved):

PR #10774, opened 2021-03-30
Use network specific swaps contract address when checking swap contract token approval

danjm:
Found during QA of v9.3.0

- on binance, swap an ERC20 token to any other token, but don't swap away all of your tokens
- this will require the "approval" fee on the first swap
- then try to swap the same token again, every request for quotes will fail

The problem was that we had one hard coded instance of the mainnet swaps contract address. This was being used when checking whether a token had been approved. This PR fixes the issue by making the contract address network specific via `SWAPS_CHAINID_CONTRACT_ADDRESS_MAP[chainId]`


darkwing (approved):
Excellent, my swaps all worked great!

Issue #8252, opened 2020-03-30
Send token screen does not use token icons from custom-added assets.


danfinlay:
As a user who has added a token with a custom icon via EIP-747, I expect to see that custom icon in all places where the token icon is rendered.

Currently the send screen does not use the custom icon:
![Screen Shot 2020-03-30 at 10 37 53 AM](https://user-images.githubusercontent.com/542863/77943813-b8c55600-7272-11ea-8ef0-56c9c25f8e29.png)

danjm:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-747.md

PR #10777, opened 2021-03-30
Ensure correct primary currency image is displayed on home screen and token list

darkwing:
Explanation:

When on Binance Smart Chain, it's a bit jarring to see ETH as the default token image in the home header and token listing. This PR aims to fix that.

https://user-images.githubusercontent.com/46655/113065431-8056a180-917e-11eb-876e-32f75648e2e5.mp4


shared/constants/swaps.js
```diff
@@ -46,7 +46,7 @@ const BSC_CONTRACT_ADDRESS = '0x1a1ec25DC08e98e5E93F1104B5e5cdD298707d31';
 const METASWAP_ETH_API_HOST = 'https://api.metaswap.codefi.network';
 
 const METASWAP_BNB_API_HOST = 'https://bsc-api.metaswap.codefi.network';
-const BNB_CHAIN_ID = '0x38';
+export const BNB_CHAIN_ID = '0x38';
```
Gudahtt, line 49: This seems better suited for the `network` constants module - that's where we have all of our other chain IDs. Especially since it's now being used for something unrelated to swaps.

ui/app/selectors/selectors.js
```diff
@@ -475,3 +481,11 @@ export function getIsSwapsChain(state) {
   const chainId = getCurrentChainId(state);
   return ALLOWED_SWAPS_CHAIN_IDS[chainId];
 }
+
+export function getPrimaryTokenImage(state) {
+  const chainId = getCurrentChainId(state);
+
+  return chainId === BNB_CHAIN_ID
```
darkwing, line 488: I may want to add another check here for Mainnet ETH so we show the traditional light image as we did before.

shared/constants/swaps.js
Gudahtt: Also, why is this called `BNB_CHAIN_ID`? Isn't the chain called "BSC"? cc @danjm

ui/app/selectors/selectors.js
```diff
@@ -475,3 +481,11 @@ export function getIsSwapsChain(state) {
   const chainId = getCurrentChainId(state);
   return ALLOWED_SWAPS_CHAIN_IDS[chainId];
 }
+
+export function getPrimaryTokenImage(state) {
+  const chainId = getCurrentChainId(state);
```
Gudahtt, line 486: Perhaps it would be better to use `nativeCurrency` here instead of the `chainId`? That's what this seems to be for - a representation of the native currency.

This would let us fix this for _all other_ custom networks as well, by displaying _nothing_ if we don't recognize the native currency. We'd ideally want to show the ETH symbol only of the native currency is ETH, rather than as a fallback all of the time.

ui/app/selectors/selectors.js
```diff
@@ -475,3 +481,11 @@ export function getIsSwapsChain(state) {
   const chainId = getCurrentChainId(state);
   return ALLOWED_SWAPS_CHAIN_IDS[chainId];
 }
+
+export function getPrimaryTokenImage(state) {
```
Gudahtt, line 485: Nit: The thing we're getting an image for here is referred to as the "native currency" throughout the extension. It is certainly not a token.

```suggestion
export function getNativeCurrencyImage(state) {
```

shared/constants/swaps.js
danjm: ack, BNB is the symbol of the default token on the chain, so this is a mistake

ui/app/selectors/selectors.js
darkwing: Good idea!

ui/app/pages/send/send-content/send-asset-row/send-asset-row.component.js
```diff
@@ -121,7 +127,8 @@ export default class SendAssetRow extends Component {
         <div className="send-v2__asset-dropdown__asset-icon">
           <Identicon
             diameter={36}
-            address={nativeCurrency === ETH ? undefined : nativeCurrency}
+            image={nativeCurrencyImage}
+            address={nativeCurrency}
```
Gudahtt, line 131: Passing in the native currency as the "address" seems wrong :thinking: I see that we were already doing this for non-ETH native currencies though, for some reason :confused:

shared/constants/swaps.js
danjm: You could add the export to the network constants module, as well as a `CHAIN_ID_TOKEN_IMAGE_MAP` export, and then leave the `constants/swaps` file unchanged. We can then address redundancies in a later PR.

ui/app/selectors/selectors.js
```diff
@@ -475,3 +481,15 @@ export function getIsSwapsChain(state) {
   const chainId = getCurrentChainId(state);
   return ALLOWED_SWAPS_CHAIN_IDS[chainId];
 }
+
+export function getNativeCurrencyImage(state) {
+  if (getIsMainnet(state)) {
```
danjm, line 486: See my previous comment. If we have a `CHAIN_ID_TOKEN_IMAGE_MAP`, which includes the test networks, then we can simplify this logic to:

```
function getNativeCurrencyImage(state) {
    return CHAIN_ID_TOKEN_IMAGE_MAP[chainId];
}
```

This will then allow us to add future symbols without adding more `else if` checks here.

shared/constants/swaps.js
Gudahtt: I brought up in a [different comment](https://github.com/MetaMask/metamask-extension/pull/10777#discussion_r604868465) that this image corresponds with the native currency, not the chain ID. Different chains with the same currency will want the same image here (e.g. Mainnet and the Ethereum testnets)

ui/app/pages/send/send-content/send-asset-row/send-asset-row.component.js
danjm: The effect of this, and perhaps this was the intent, is to have a jazzicon display when there is a custom symbol for the custom network. Like this:
Screenshot from 2021-03-31 20-26-23

If no symbol is provided, it looks like this.

Screenshot from 2021-03-31 20-27-00

I agree that we should not do this, and instead should just show the new default of Identicon: `<div style={getStyles(diameter)}></div>;`

Now, the only problem with changing this is if some users have become acustomed to seeing the jazzicon for their particular custom network... so I think it'd be best not to break that in this PR.

shared/constants/swaps.js
danjm: Ah, sorry, yes. So a `NATIVE_CURRENCY_TOKEN_IMAGE_MAP` is what we want here.

ui/app/pages/send/send-content/send-asset-row/send-asset-row.component.js
Gudahtt: Yeesh. So we're treating this as an address without validating that it's even remotely similar to an address. That's unfortunate.

Reminds me of this bug: https://github.com/MetaMask/metamask-extension/issues/8252.

Leaving this alone for this PR seems sensible to me. I'm not sure we should be concerned about users being accustomed to this icon though. It seems very broken and confusing to me. Particularly because we use jazzicons for people, and we don't want tokens mistaken for people. Maybe we can discuss it in a future standup.

shared/constants/swaps.js
darkwing: Implemented `NATIVE_CURRENCY_TOKEN_IMAGE_MAP`.

ui/app/selectors/selectors.js
darkwing: Done.

shared/constants/network.js
```diff
@@ -81,3 +88,13 @@ export const CHAIN_ID_TO_NETWORK_ID_MAP = Object.values(
   chainIdToNetworkIdMap[chainId] = networkId;
   return chainIdToNetworkIdMap;
 }, {});
+
+export const NATIVE_CURRENCY_TOKEN_IMAGE_MAP = {
```
danjm, line 92: So this needs to be a map from currency symbols ("ETH', 'TESTETH', 'BNB') to urls, as per gudhatt's comment

> I brought up in a different comment that this image corresponds with the native currency, not the chain ID. Different chains with the same currency will want the same image here (e.g. Mainnet and the Ethereum testnets)

danjm: oh, but I guess this implmentation has the same effect...

danjm: and this ensures that if someone adds the BSC network but doesn't specify a symbol, the image will still be correctly selected

Gudahtt: I don't think this is equivalent. Now the native currency and the image displayed for this currency can get out-of-sync.

For example:
>this ensures that if someone adds the BSC network but doesn't specify a symbol, the image will still be correctly selected

In this situation, the balance would be shown in ETH (the default, if no symbol is entered), but the picture would be of the Binance logo.

danjm: > In this situation, the balance would be shown in ETH (the default, if no symbol is entered), but the picture would be of the Binance logo.

ah, that's a good point

darkwing: Looking at `ui/app/selectors/send.js`'s `getNativeCurrency` function, `state.metamask.nativeCurrency` is `ETH` on test networks, yet `TEST_ETH_SWAPS_TOKEN_OBJECT.symbol` is `TESTETH`, so that seems weird.  I can update the map to be keyed by symbol easily enough, but the `TESTETH`/`ETH` mismatch will be problematic.

danjm: so what I think is needed is:
```
export const NATIVE_CURRENCY_TOKEN_IMAGE_MAP = {
  [ETH_SWAPS_TOKEN_OBJECT.symbol]: ETH_SWAPS_TOKEN_OBJECT.iconUrl,
  [TEST_ETH_SWAPS_TOKEN_OBJECT.symbol]: TEST_ETH_SWAPS_TOKEN_OBJECT.iconUrl,
  [BNB_SWAPS_TOKEN_OBJECT.symbol]: TEST_ETH_SWAPS_TOKEN_OBJECT.iconUrl,
};
```
Does that make it less problematic @darkwing?

danjm: later we can clean it up so that we are not importing these values from swaps, I think the logic here is correct

danjm (approved):
LGTM

PR #10778, opened 2021-03-31
Refactoring ethereum-on.spec.js to use fixtures

### LINKED ISSUE #10771: Refactor ethereum-on.spec.js to follow pattern in metamask-extension/test/e2e/tests
This is part of https://github.com/MetaMask/metamask-extension/issues/10754, and is the to-do item for the ethereum-on.spec.js

`metamask-extension/test/e2e/tests/personal-sign.spec.js` can be used an example

 Only the `has the ganache network id within the dapp` and below tests need to be included.

NiranjanaBinoy:
Fixes: #10771 

Explanation:  
1. Moved the file to `metamask-extension/test/e2e/tests`
2. Updated `ethereum-on.spec.js`  to use fixtures.
3. Renamed `ethereum-on.spec.js` to `provider-events.spec.js`
4. `provider-events.spec.js` is using the same fixture as the `personal-sign`
5. Renamed `personal-sign` folder for fixture to `connected-state`


test/e2e/tests/provider-events.spec.js
```diff
@@ -0,0 +1,79 @@
...
+        const addressInput = await driver.findElement(
+          By.css('.readonly-input__input'),
+        );
+        const publicAddress = await addressInput.getAttribute('value');
+
+        const accountModal = await driver.findElement(By.css('span .modal'));
+        await driver.clickElement(By.css('.account-modal__close'));
+
+        await driver.wait(until.stalenessOf(accountModal));
+        await driver.delay(regularDelayMs);
```
Gudahtt, line 43: It seems as though the only purpose of this section is to obtain the public address. We don't need to get this interactively upon every test though - we already know it. It's in the fixture: `0x5cfe73b6021e818b776b421b1c4db2474086a7e1`

We can hard-code this instead, to save all of this time.

test/e2e/tests/provider-events.spec.js
```diff
@@ -0,0 +1,79 @@
...
+
+        await driver.openNewPage('http://127.0.0.1:8080/');
+
+        const windowHandles = await driver.getAllWindowHandles();
+        const extension = windowHandles[0];
+        const dapp = await driver.switchToWindowWithTitle(
+          'E2E Test Dapp',
+          windowHandles,
+        );
```
Gudahtt, line 52: We don't need to switch to the E2E test dapp window here - it's already the current window. This should be avoided if possible, because it cycles through every open window to find the one being requested, even though it's already open.

We do need the reference to the `dapp` window handle, but not until later. We can move this `switchToWindowWithTitle` call to the first instance where we need the `dapp` handle (the ` await driver.switchToWindow(dapp);` step below on line 68).

Gudahtt (commented):
Looks good! I left a few comments about steps that can be removed entirely, but aside from that this seems functionally correct.

test/e2e/tests/provider-events.spec.js
```diff
@@ -0,0 +1,79 @@
...
+        await driver.delay(largeDelayMs);
+
+        await driver.switchToWindow(dapp);
+        const switchedNetworkDiv = await driver.findElement(By.css('#network'));
+        const chainIdDiv = await driver.findElement(By.css('#chainId'));
+        const accountsDiv = await driver.findElement(By.css('#accounts'));
+
+        assert.equal(await switchedNetworkDiv.getText(), '3');
+        assert.equal(await chainIdDiv.getText(), '0x3');
+        assert.equal(await accountsDiv.getText(), publicAddress.toLowerCase());
```
Gudahtt, line 75: I've just noticed that this last assertion doesn't really work. It's meant to test the `accountsChanged` event, but, that event is never emitted here and this still works :thinking: 

Not a blocker for this PR since it's an existing problem in the test, but we should move this out into its own `it` block at some point, where we compare this text value before and after switching accounts.

Gudahtt (commented):
I suspect we don't need any of these `delay` calls. Those were added a long time ago when we were directly using the selenium webdriver. Since then we've switched to our own driver abstraction, which is somewhat more robust.

NiranjanaBinoy:
> I suspect we don't need any of these `delay` calls. Those were added a long time ago when we were directly using the selenium webdriver. Since then we've switched to our own driver abstraction, which is somewhat more robust.

test/e2e/tests/provider-events.spec.js
NiranjanaBinoy: I have not made any change to the last assertion as of yet, will handle that as a separate PR once after the current one is merged.

Gudahtt (approved):
LGTM!

diff
:heart_eyes:

PR #10780, opened 2021-03-31
Build - fix bug in empty file copying

kumavis:
whoops, so sorry

the source/pattern/dest copy does not quite work the way i expected :grimacing:

### DIFF
--- a/development/build/static.js
+++ b/development/build/static.js
@@ -82,13 +82,11 @@ const copyTargetsDev = [
   },
   // empty files to suppress missing file errors
   {
-    src: './development/',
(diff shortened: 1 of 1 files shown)


kumavis:
develop resulted in a file called `chromereload.jsempty.js` in the prod build
this gave me an idea for a snapshot of the build output, as a sanity check
may implement in another PR

EtDu (approved):
ggwp

ryanml (approved):
++

kumavis:
fyi @Gudahtt

Issue #10752, opened 2021-03-29
Improved release automation


Gudahtt:
This is a tracking issue for the various release automation improvements were are planning to make.

- [ ] Tag & publish tip of release branch, rather than after release branch merge.
  * This will let us drop the `master` branch, and target the main branch instead (`develop`)
  * Any conflicts between the release branch and the main branch should be resolved *after* publishing, but *before* merging the release branch. This means that the trigger for starting the release process cannot be the merging of the release branch. It'll have to be triggered by something else (e.g. a GitHub comment).
- [ ] Improve automatic changelog generation
  * It should generate a changelog in the ["keep a changelog" format](https://keepachangelog.com/en/1.0.0/), which each commit initially uncategorized.
  * it should exclude any entries already listed in the changelog
  * It should be run automatically when a new release candidate is being prepared
  * It should group new entries under the header of the release currently being prepared
- [ ] Write instructions for new release process
  * These should be included in the repository
  * These instructions should assume as little as possible. Include examples of each command that should be run.
  * Include instructions for both normal and hotfix releases
  * Include instructions on how to monitor for regressions after the release has been published

PR #10784, opened 2021-03-31
Add changelog entries under release candidate header

Gudahtt:
Instead of always placing new changelog entries under the "Current Develop Branch" header, the changelog script now places them under the header for the current release if that release has not yet been tagged.

This eliminates one manual step from the release process.

Relates to #10752

Manual testing steps :

For development changelog updates:
- Run `yarn update-changelog`
- See that it places changelog entries under 'Current Develop Branch'
 
For RC changelog updates:
- Bump the `version` in `app/manifest/_base.json`, to simulate an RC-like environment
- Add a new release header to the changelog for the new version
- Run `yarn update-changelog`
- See that it places changelog entries under the new release header
 
For updates without a release header:
- Bump the `version` in `app/manifest/_base.json`, to simulate an RC-like environment
- Run `yarn update-changelog`
- See that it throws an error because it cannot find the release header
 
For testing that it doesn't confuse v10.0.0 and v10.0.0-prefix:
- Bump the `version` in `app/manifest/_base.json` to v10.0.0, to simulate an RC-like environment
- Add a header to the changelog for v10.0.0-beta.0
- Run `yarn update-changelog`
- See that it throws an error because it cannot find the release header, showing that it doesn't confuse v10.0.0 for v10.0.0-beta.0

PR #10786, opened 2021-03-31
Prevent duplicate changelog entries

Gudahtt:
The changelog update script now prevents duplicate entries from being added. Specifically, it will ensure that if a PR has been referenced already in an entry, it will not add it again.

This should prevent it from adding duplicate entries for changes that were cherry-picked into hotfix releases.

Note that this duplication prevention only works for entries containing a PR number. We don't have any way to prevent duplicate entries yet in cases where we don't know the associated PR. We will be preventing this possibility entirely pretty soon in some upcoming release automation changes though, so I'm not concerned about this omission.

Relates to #10752

Manual testing steps:  
  - Run `yarn update-changelog`
  - Look at the changelog
  - Run `yarn update-changelog` again
  - See that no entries including a PR number have been added a second time

### DIFF
--- a/development/auto-changelog.js
+++ b/development/auto-changelog.js
@@ -27,7 +27,7 @@ async function main() {
     `${mostRecentTag}..HEAD`,
   ]);
 
-  const changelogEntries = [];
+  const commitEntries = [];
   for (const commit of commitsSinceLastRelease) {
     const [subject] = await runCommand('git', [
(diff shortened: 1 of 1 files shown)


Gudahtt:
This depends upon #10784

brad-decker (approved):
LGTM

PR #10788, opened 2021-03-31
only applies rules to the appropriate files

brad-decker:
Previously the mocha rules were disabled for anything without the .test.js suffix, which unfortunately included the .spec.js files in E2E folder. This update aims to simplify overrides for tests by only extending the mocha configuration for files that need it. 

Also, this change now only applies react rules to the UI folder.


Gudahtt (approved):
LGTM!

PR #10789, opened 2021-03-31
upgrade eslint deps

brad-decker:
Upgrade eslint dependencies to improve lint performance and accuracy. Fixes one new warning from react hooks rules.


Gudahtt (approved):
LGTM!

PR #10794, opened 2021-04-01
Add release header when updating changelog

Gudahtt:
The changelog update script now adds a release header if it doesn't find one already that matches the current release candidate version.

Relates to #10752

Manual testing steps:  

For development changelog updates:
- Run `yarn update-changelog`
- See that it places changelog entries under 'Current Develop Branch'
 
For RC changelog updates with existing release header:
- Bump the `version` in `app/manifest/_base.json`, to simulate an RC-like environment
- Add a new release header to the changelog for the new version
- Run `yarn update-changelog`
- See that it places changelog entries under the new release header
 
For updates without a release header:
- Bump the `version` in `app/manifest/_base.json`, to simulate an RC-like environment
- Run `yarn update-changelog`
- See that it places changelog entries under the newly added release header

PR #10795, opened 2021-04-01
Update the changelog when creating an RC

Gudahtt:
The changelog will now be automatically updated when a release branch is created. A new release header along with changelog entries for any new commits will be added.

Note that this changelog will still need to be manually cleaned up, but it's one less manual step at least.

The old Bash script for adding a new release header to the changelog has been removed, as that functionality is now built into the changelog update script.

A new script has been added to commit any changes made to the manifest and changelog. This step used to happen at the end of the bump manifest version script, but now the changelog update relies upon the manifest version bump happening first, so it needed to be re-ordered. The changes should only be committed on the first run of the branch, as it's contingent upon the manifest changing (due to the version bump). Further changelog updates won't trigger new automatic commits.

Relates to #10752

Manual testing steps:  
  - Create a new release branch
  - See that it updates the changelog automatically


Gudahtt:
This depends upon #10794

.circleci/scripts/release-commit-verison-bump.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
```
rekmarks, line 1: Typo in file name: `verison` -> `version`.

It appears to be correctly spelled in the CircleCI config, but we should review all references.

rekmarks (changes requested):
Looks good except for a typo.

.circleci/scripts/release-commit-verison-bump.sh
Gudahtt: Oops :facepalm: Good catch! Fixed.

rekmarks (dismissed):
LGTM!

Gudahtt:
I've just created the branch `Version-v9.3.1` to test this, just to double check that it all works :eyes:

rekmarks:
ffffuuuuuuu
```
@lavamoat/allow-scripts automatically updating configuration

removing unneeded configuration for packages:
- fsevents
Done in 2.65s.
allow-scripts configuration requires updates
```

.circleci/scripts/release-commit-version-bump.sh
```diff
@@ -18,7 +18,7 @@ fi
 
 printf '%s\n' 'Commit the manifest version and changelog if the manifest has changed'
 
-if ! git diff --quiet CHANGELOG.md app/manifest/_base.json;
+if git diff --quiet app/manifest/_base.json;
```
rekmarks, line 21: 😓

Gudahtt:
It worked! https://github.com/MetaMask/metamask-extension/commit/d6b5abdc68eece5ed281600900698bc694633767

... after fixing a couple more mistakes :sweat_smile: . I put each bug fix in a separate commit - check out the most recent two since your approval to see what they were.

Gudahtt:
That status check shouldn't affect this PR thankfully, because this isn't a release branch. It was only triggered because this commit was used in a release branch. GitHub applies these status checks by commit it seems, not by branch.

rekmarks (approved):
Alright, now it looks good!

PR #10208, opened 2021-01-18
use dart sass, and update related modules

brad-decker:
### Rationale
At some point in the future, I would like to refactor the way we use our scss files to make use of the `@use` rule. switching from `node-sass` to `sass` is a requirement of doing that, because `node-sass` binds to `libsass` which hasn't received new language features since November 2018. The official implementation of sass is `dart-sass` which has js bindings through the `sass` dependency.

More short-termed: I have need of the `math.pow` function that was added officially to the `math` module but never landed in `libsass` 

This PR
1. removes top-level `node-sass` dependency (gulp-sass still depends on it and it must be compiled in prep-deps on circleci). 
2. adds `sass` (js bindings to `dart-sass`)
3. ~adds fibers, which is recommended by both `gulp-sass` and `sass-loader` for improving performance during compilation.~ despite this recommendation, the `fibers` package is marked as deprecated and heavily discouraged from use. Furthermore, it didn't appear to do anything in terms of performance on the build. I removed this complexity from the PR.
5. updates `sass-loader`
6. passes `sass` as implementation to `sass-loader`
7. updates `design-system/index.scss` for `@forward` its contents, ~which should have zero impact on the final result~(1). However, if `node-sass` was still being used it would fail to compile.

1) It *did* end up having an impact, apparently `@forward` doesn't forward maps? I'm not sure what happened there but in a future PR where I used `$color-map` from `desing-system/colors` it didn't work until I `@use`'d colors directly. This isn't blocking this PR because `$color-map` is not used yet.

PR #10797, opened 2021-04-01
remove node-sass dependency

brad-decker:
Back in #10208 I changed us over to using dart-sass instead of node-sass, but gulp-sass had a dependency on node-sass that still required it in our build. This change uses `gulp-dart-sass` to remove that element.


Gudahtt:
@brad-decker Instructions on how to manage lavamoat policies and allow-scripts configuration are here: https://github.com/MetaMask/metamask-extension#changing-dependencies

And yes, deleting the patch would be the correct approach. That reminds me - we don't document how to deal with patches yet :thinking: I should address that.

brad-decker:
Erg... the policy file changes made me look at bundles

https://bundlephobia.com/result?p=gulp-dart-sass@1.0.2 vs https://bundlephobia.com/result?p=gulp-sass@4.1.0

2.2mb minified vs 141kb minified... I'm thinking we just keep node-sass around for a while.

brad-decker:
Ew, the diff on p

Gudahtt:
Check out the install scripts for node sass though. Are they downloading a binary? The true size can be hidden a lot of the time.

Plus... having both around doesn't do us much good if we're only using one.

Gudahtt:
That is misleading because of node-sass, which downloads a binary as part of the install script I think? Thus hiding the true size.

Plus the bulk of the size of the dart package is due to the sass package, which we already have in our dependency tree. I do think this would be a strict reduction in our dependencies.

kumavis (dismissed):
looks good. huge yarn-lock reduction :eyes: :+1:

Gudahtt:
I have just rebased this to fix the `yarn.lock` conflict

Gudahtt (approved):
LGTM!

PR #10800, opened 2021-04-01
Set the BSC_CONTRACT_ADDRESS to lowercase

danjm:
Set the BSC_CONTRACT_ADDRESS to lowercase. This makes it consistent with other address constants and the way all addresses are handled in our front-end. This fixes a bug found while I was QAing where swaps transactions don't appear in token specific views while on binance.


Gudahtt:
>swaps transactions don't appear in token specific views while on binance.

How did the case of the address cause this to happen? :thinking:

tmashuang:
Is this the case where `txParams.to` is not checksummed?

danjm:
> How did the case of the address cause this to happen? thinking
@Gudahtt See Thomas' comment above

danjm:
But maybe then there is a question of why the `to` params are not checksummed... perhaps because we get it from the api?

danjm:
oh, we normalize it to lowercase in `addUnapprovedTransaction`

tmashuang (approved):
LGTM

PR #10802, opened 2021-04-01
Add support for locators into driver abstraction

brad-decker:
supersedes #10801 

Adds support for inline locators in driver element matcher methods, instead of using Selenium's `By` abstraction. In the interim, they ALSO accept the latter. Additional PRs will be set up to convert usages to remove the usage of `By`


Gudahtt (approved):
LGTM!

PR #10804, opened 2021-04-02
Refactoring address-book.spec.js to use fixtures

### LINKED ISSUE #10798: Refactor address-book.spec.js to follow pattern in metamask-extension/test/e2e/tests
This is part of #10754, and is the to-do item for the address-book.spec.js

NiranjanaBinoy:
Fixes: #10798 

Explanation:  

- Moved the file to test/e2e/tests
- The address-book.spec.js has 2 `it` parts first part adds a new address and second uses an existing address entry, first one is re-using `imported-account` and  a new state.json is added to `test/e2e/fixtures/address-entry` which is used in the second scenario.
- Updated run-all.sh to use the new test file.


Gudahtt (approved):
LGTM! I like where you chose to split the tests up, that worked out well

PR #10805, opened 2021-04-02
Update changelog headers and fix dates

Gudahtt:
The changelog release header format has been updated to match the ["keep a changelog" ](https://keepachangelog.com/en/1.0.0/) format. Each header is now the bracketed version number followed by a dash, then the release date in ISO-8601 format.

The release dates in each header were also updated to match the date of the corresponding [GitHub Release](https://github.com/MetaMask/metamask-extension/releases). Many of these dates were incorrect because they were set on the day we created the release candidate, rather than on the day of release.

Any changelog release entries without a corresponding GitHub release was left with the date already specified.

The three oldest release headers were missing dates. For the first two, I used the date of the version bump commit. For the third, I removed it since no changes were listed anyway, and it represented a range of releases rather than a single one.

The `auto-changelog.js` script has been updated to account for this new format as well.

Manual testing steps: 

For RC changelog updates with existing release header:
- Bump the `version` in `app/manifest/_base.json`, to simulate an RC-like environment
- Add a new release header to the changelog for the new version
- Run `yarn update-changelog`
- See that it places changelog entries under the new release header
 
For updates without a release header:
- Bump the `version` in `app/manifest/_base.json`, to simulate an RC-like environment
- Run `yarn update-changelog`
- See that it places changelog entries under the newly added release header

PR #10808, opened 2021-04-02
Add links to release headers

Gudahtt:
Each release header now includes a link to the range of commits included with that release. These links are at the end of the document, in accordance with the ["keep a changelog"](https://keepachangelog.com/en/1.0.0/) format.

For the purpose of this changelog, the "previous release" is the most recent release mentioned in the changelogs. The diffs ignore any releases that were omitted from the changelog. This is mainly an issue with older releases, so it seemed acceptable. All releases have been documented for a couple of years now, and will be going forward as well.

The name of the "Current Develop Branch" section was changed to "Unreleased" to confirm with "keep a changelog".

The `auto-changelog.js` script has been updated to update/add these links whenever adding a new release header as well.

Relates to #10752

Manual testing steps:  

For development changelog updates:
- Run `yarn update-changelog`
- See that it places changelog entries under 'Unreleased`

For RC changelog updates with existing release header:
- Bump the `version` in `app/manifest/_base.json`, to simulate an RC-like environment
- Add a new release header to the changelog for the new version
- Run `yarn update-changelog`
- See that it places changelog entries under the new release header along with a new release header link, and it updates the `Unreleased` link.
 
For updates without a release header:
- Bump the `version` in `app/manifest/_base.json`, to simulate an RC-like environment
- Run `yarn update-changelog`
- See that it places changelog entries under the newly added release header


Gudahtt:
This depends upon #10795 and #10805

development/auto-changelog.js
```diff
@@ -96,8 +96,22 @@ async function main() {
     const firstReleaseHeaderIndex = changelogLines.findIndex((line) =>
       line.match(/## \[\d+\.\d+\.\d+\]/u),
     );
+    const [, previousVersion] = changelogLines[firstReleaseHeaderIndex].match(
+      /## \[(\d+\.\d+\.\d+)\]/u,
+    );
     changelogLines.splice(firstReleaseHeaderIndex, 0, versionHeader, '');
     releaseHeaderIndex = firstReleaseHeaderIndex;
+
+    // Update release links
```
rekmarks, line 105: nit: "links" is a confusing reference. It would be better if the comment stated that it's rewriting the `Unreleased` link and inserting a link for new version.

development/auto-changelog.js
```diff
@@ -75,12 +75,12 @@ async function main() {
   const isReleaseCandidate = mostRecentVersion !== version;
   const versionHeader = `## [${version}]`;
   const escapedVersionHeader = escapeRegExp(versionHeader);
-  const currentDevelopBranchHeader = '## Current Develop Branch';
+  const currentDevelopBranchHeader = '## [Unreleased]';
```
rekmarks, line 78: This may be a preexisting issue, but I'm confused by the error thrown in side the `if` statement on L88 below, which will always have the message `Failed to find release header '## [Unreleased]'`, which should always (in terms of its natural language meaning) be false.

rekmarks (commented):
This will be a joy to use.

development/auto-changelog.js
Gudahtt: That error is to protect against the changelog getting malformed. If it can't find the "Unreleased" header, it can't know where to place the changelog entries so it blows up.

rekmarks: Some defensive programming. I'm all for it!

development/auto-changelog.js
Gudahtt: To be more specific, it's a link reference definition: https://github.github.com/gfm/#link-reference-definitions
I'll update the comment to better explain this.

Gudahtt: Addressed here: bad85c4fd6c825faffca9b2f89f2e9d928bfe6af

brad-decker (approved):
LGTM! nice work

rekmarks (approved):
LGTM!

PR #10819, opened 2021-04-04
Fix _getPermittedAccounts type safety

rekmarks:
`PermissionsController._getPermittedAccounts` could return an array or `null`, but it was used in `removeAllAccountPermissions` as though it could only return an array.

I don't believe it _would_ ever return anything other than an array as things stand, but it's not type safe, and we should fix it. Things with `getAccounts` in the name tend to only return arrays in our codebase, so I made `_getPermittedAccounts` only return arrays, and updated calls expecting `null` to expect an empty array instead.


darkwing (approved):
I very, very much like this change!  LGTM!

PR #10820, opened 2021-04-04
Refactoring signature-request.spec.js to use fixtures

### LINKED ISSUE #10806: Refactor signature-request.spec.js to follow pattern in metamask-extension/test/e2e/tests
This is part of #10754, and is the to-do item for the signature-request.spec.js

NiranjanaBinoy:
Fixes: #10806

Explanation:

- Moved the file to `test/e2e/tests`
- Updated `signature-request.spec.js` to use fixtures.
- Updated `run-all.sh` to use the new test file.
- Deleted the old `signature-request.spec.js` file


test/e2e/tests/signature-request.spec.js
```diff
@@ -0,0 +1,93 @@
...
+        assert.equal(await name.getText(), 'Ether Mail');
+        assert.equal(await origin.getText(), 'http://127.0.0.1:8080');
+        assert.equal(
+          await address.getText(),
+          `${publicAddress.slice(0, 8)}...${publicAddress.slice(
+            publicAddress.length - 8,
+          )}`,
+        );
+
+        // signs the transaction
```
Gudahtt, line 63: Nit: This is a mistake in the original test, but, the test description "signs the transaction" doesn't seem right :thinking: Since this isn't a transaction being signed here, it's typed data.

```suggestion
        // Approve signing typed data
```

test/e2e/tests/signature-request.spec.js
```diff
@@ -0,0 +1,93 @@
...
+        const addressInput = await driver.findElement(
+          By.css('.readonly-input__input'),
+        );
+        const newPublicAddress = await addressInput.getAttribute('value');
+        const accountModal = await driver.findElement(By.css('span .modal'));
+
+        await driver.clickElement(By.css('.account-modal__close'));
+
+        await driver.wait(until.stalenessOf(accountModal));
+        assert.equal(newPublicAddress.toLowerCase(), publicAddress);
```
Gudahtt, line 89: I'm not really sure why this last part was included with the test :thinking: It doesn't seem especially relevant.

```suggestion
```

Gudahtt: We don't ever confirm that the signature worked though. That seems like a problem.

We could confirm it worked by switching to the test dapp and checking for a result (it should say "Result: [hash]" if it worked. Then there is a "Verify" button to verify the signature, though that verification step is unnecessary if we know the exact hash to expect.

I won't block on adding a last step to confirm that the signature worked though, since it's a pre-existing problem. That can happen in a separate PR if you'd prefer.

Gudahtt (commented):
Looks good! I think we can remove more from the end, but I can approve after that.

I think the last section was intended to verify the signature, but it doesn't actually do that unfortunately.

Gudahtt (approved):
LGTM!

PR #10826, opened 2021-04-05
Update dependabot to standard settings

darkwing:
Puts our dependabot in line with new template.  The only consideration is whether we want to remove `"@metamask/*"` -- we may get an avalanche of non-MM repo dependency updates. (i.e. react widgets, etc.)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,9 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+      time: "06:00"
     allow:
       - dependency-name: "@metamask/*"
-    versioning-strategy: "lockfile-only"
+    target-branch: "develop"
+    versioning-strategy: "increase-if-necessary"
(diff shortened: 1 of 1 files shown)


Gudahtt:
The 'standard settings' in the template repo are more appropriate for libraries. That's where lockfile-only changes aren't useful.

Since this is the end product, lockfile changes do impact the final build. If we don't ask Dependabot to update `@metamask/*` lockfile dependencies, we're back to manually doing it ourselves.

Maybe we could use the standard "increase-if-necessary" style configuration for _all_ dependencies, but keep the "lockfile-only" configuration for internal dependencies? That would alert us whenever substantial dependency changes were made that might deserve our attention, but wouldn't bother us about every little patch that we might not care about.

Gudahtt:
Nevermind, `update-if-necessary` still does lockfile-only updates. I misunderstood.

PR #10831, opened 2021-04-06
Removing unnecessary params from withFixtures function call.

NiranjanaBinoy:
Removing the `dapp=true` parameter from `withFixtures` function call on the tests that are not using the Dapp.


Gudahtt (approved):
LGTM, good catch!

PR #10833, opened 2021-04-06
use locator abstraction in tests folder

brad-decker:
Implements the abstraction for the 'by' method from selenium implemented in #10802

### DIFF
--- a/test/e2e/metamask-ui.spec.js
+++ b/test/e2e/metamask-ui.spec.js
@@ -1,7 +1,6 @@
 const assert = require('assert');
-const webdriver = require('selenium-webdriver');
+const { Key, until } = require('selenium-webdriver');
 
-const { By, Key, until } = webdriver;
 const enLocaleMessages = require('../../app/_locales/en/messages.json');
 const { tinyDelayMs, regularDelayMs, largeDelayMs } = require('./helpers');
 const { buildWebDriver } = require('./webdriver');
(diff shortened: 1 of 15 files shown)


Gudahtt:
This has conflicts, and will conflict will https://github.com/MetaMask/metamask-extension/pull/10820 as well.

cc @NiranjanaBinoy, maybe we can coordinate this to avoid further conflicts? e.g. merge #10820, then this, then the rest of the fixture migrations after that (or something along those lines).

brad-decker:
I can also break this out into smaller pieces, prioritizing those files that are either a few steps away from being refactored or have already been refactored.

Gudahtt:
That could work! Though I expect that I'd be able to review this fairly quickly, since the changes are all pretty easy to follow. If we're quick to review and merge this, the all-at-once approach might result in the least conflicts.

Gudahtt:
#10820 has been merged now as well, and I've just heard from @NiranjanaBinoy that there aren't any further big changes in draft yet, so now would be a good time to get this updated and reviewed/merged

brad-decker:
@Gudahtt updated the branch with latest from develop

test/e2e/metamask-ui.spec.js
```diff
@@ -325,145 +299,127 @@ describe('MetaMask', function () {
...
       await inputAmount.sendKeys('1');
 
       const inputValue = await inputAmount.getAttribute('value');
       assert.equal(inputValue, '1');
 
       // Set the gas price
-      await driver.clickElement(
-        By.xpath(`//button/div/div[contains(text(), "Fast")]`),
-      );
+      await driver.clickElement({ text: 'Fast', tag: 'button/div/div' });
```
Gudahtt, line 344: I guess this means the `tag` property is only compatible with xpath :thinking: 

It might be worth using the `xpath` locator here explicitly if we want to make the `text` and `tag`-based locators implementable without using xpath.

test/e2e/metamask-ui.spec.js
```diff
@@ -476,41 +432,34 @@ describe('MetaMask', function () {
...
+      await driver.clickElement('.account-menu__icon');
       await driver.delay(regularDelayMs);
 
-      await driver.clickElement(
-        By.xpath(`//div[contains(text(), 'Settings')]`),
-      );
+      await driver.clickElement({ text: 'Settings', tag: 'div' });
 
-      // await driver.findElement(By.css('.tab-bar'))
+      // await driver.findElement('.tab-bar')
```
Gudahtt, line 440: :joy: Whoops! I wonder how that got there.

test/e2e/metamask-ui.spec.js
brad-decker: How would you make text locators implementable without xpath with Selenium? For what its worth this can be expressed very easily with Playwright `button > div > div:has-text("fast")`

Gudahtt: Oh I meant in Playwright - I think we need to use xpath for this in Selenium.

if this slash syntax is easily parsed/handled by Playwright that works too! Just wasn't sure it was portable. That approach you just gave an example of seems fine to me.

Gudahtt (approved):
LGTM!

The only remaining uses of the selenium `By` API are in the drivers themselves. All of the changes here seem functionally equivalent to me.

PR #10836, opened 2021-04-06
Add validation for the `fee` property from the `/trades` API response

### LINKED ISSUE #9681: Add fee property to validation of swaps `/trades` response
Comment from @Gudahtt 

> I just noticed that the fee is missing from the quote validator. We should add the fee to that at some point.

> We might also want to consider filtering the API response to just the validated parameters, so that we're forced to consider this in the future as the API changes.

https://github.com/MetaMask/metamask-extension/pull/9677#pullrequestreview-514843497

dan437:
Fixes: #9681

## Explanation
Now we are validating that the `fee` property from the `/trades` API response is the `number` type. If the API would return a different type, we would show an error page with: "No quotes available".

## Manual testing steps
  - Click on the `Swap` button inside the MetaMask extension
  - Choose `Swap from` and `Swap to` tokens
  - Click on `Review Swap`, at which time the `/trades` API is called and response validation happens

## Screenshots
Swap page:
image

`fee` validation passed and a user can see the Swap Review page:
image

`fee` property in the `/trades` API response:
image

"No quotes available" page would be displayed if the type validation fails (e.g. when we would expect an object instead of number):
image

All unit are tests green:
image

### DIFF
--- a/ui/app/pages/swaps/swaps-util-test-constants.js
+++ b/ui/app/pages/swaps/swaps-util-test-constants.js
@@ -101,6 +101,7 @@ export const MOCK_TRADE_RESPONSE_1 = [
     sourceAmount: '10000000000000000',
     destinationAmount: '2248687',
     error: null,
+    fee: 0.875,
     sourceToken: TOKENS[0].address,
     destinationToken: TOKENS[1].address,
     fetchTime: 553,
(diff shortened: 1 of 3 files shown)


dan437:

danjm (approved):
LGTM!

PR #10842, opened 2021-04-07
Increase default slippage from 2% to 3%, show Advanced Options by default

dan437:
## Explanation: 
- Increase default slippage from 2% to 3%
- Increase slippage from 1% to 2%
- Always show Advanced Options

## Manual testing steps:  
  - Click on the `Swap` button in the MetaMask extension
  - See the `Swap` page with Advanced Options visible and 3% slippage tolerance selected by default

## Screenshots:
Swap page: Advanced Options visible and 3% slippage tolerance selected by default
image

Verifying that the 3% slippage is send to the server:
image

Pre-popullating previously set custom value when going back to the Swap page:
image

All unit tests are green:
image

### DIFF
--- a/ui/app/pages/swaps/slippage-buttons/slippage-buttons.js
+++ b/ui/app/pages/swaps/slippage-buttons/slippage-buttons.js
@@ -12,7 +12,12 @@ export default function SlippageButtons({
   currentSlippage,
 }) {
   const t = useContext(I18nContext);
-  const [customValue, setCustomValue] = useState('');
+  const [customValue, setCustomValue] = useState(() => {
+    if (currentSlippage && currentSlippage !== 2 && currentSlippage !== 3) {
+      return currentSlippage;
+    }
+    return '';
+  });
   const [enteringCustomValue, setEnteringCustomValue] = useState(false);
   const [activeButtonIndex, setActiveButtonIndex] = useState(() => {
     if (currentSlippage === 3) {
@@ -20,7 +25,6 @@ export default function SlippageButtons({
     } else if (currentSlippage === 2) {
       return 0;
     } else if (currentSlippage) {
-      setCustomValue(currentSlippage);
       return 2;
     }
     return 1; // Choose activeButtonIndex = 1 for 3% slippage by default.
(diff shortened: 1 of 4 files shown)


danjm:
This change looks good, but actually reveals a pre-existing bug that I think should be addressed within this PR.

To see this bug on this branch:

- set the slippage to 2
- click "Review Swap"
- once on the "View Quote" screen click "Back"
- the selection will be defaulted back to 3, even though the `fetchParams.slippage` still equals 2

I've attached a file that shows a change that resolves that problem:
[slippage-patch.txt](https://github.com/MetaMask/metamask-extension/files/6274151/slippage-patch.txt)

However, there will still be a similar problem if a custom amount is entered. That should be resolved too, I think.

ui/app/pages/swaps/slippage-buttons/slippage-buttons.js
```diff
@@ -6,12 +6,25 @@ import ButtonGroup from '../../../components/ui/button-group';
...
-  const [open, setOpen] = useState(false);
   const [customValue, setCustomValue] = useState('');
   const [enteringCustomValue, setEnteringCustomValue] = useState(false);
-  const [activeButtonIndex, setActiveButtonIndex] = useState(1);
+  const [activeButtonIndex, setActiveButtonIndex] = useState(() => {
+    if (currentSlippage === 3) {
+      return 1;
+    } else if (currentSlippage === 2) {
+      return 0;
+    } else if (currentSlippage) {
```
danjm, line 22: I think it might be best not to call `setCustomValue` in the callback to `useState`, as this is incidental to the purpose of this function. I think the intent of the code would be clearer if we added a callback to the `useState` call that sets up the `customValue` which would set the initial custom value to `currentSlippage` if `currentSlippage` is truthy but not equal to 2 or 3.

danjm (changes requested):
One small issue to address, otherwise this code looks good

ui/app/pages/swaps/slippage-buttons/slippage-buttons.js
dan437: I like that!

dan437: Done.

danjm (approved):
LGTM!

PR #10844, opened 2021-04-07
add abstraction for waitForSelector

brad-decker:
Adds an abstraction for waitForSelector, a method that exists in the playwright API. This abstraction aims to eventually become a paper-thin abstraction over the playwright version. In the meantime `until` is used behind the scenes to mimick the behavior.

Included is a refactor of `metamask-ui.spec.js` which served as a robustness check for the utility.

### DIFF
--- a/test/e2e/metamask-ui.spec.js
+++ b/test/e2e/metamask-ui.spec.js
@@ -1,5 +1,5 @@
 const assert = require('assert');
-const { Key, until } = require('selenium-webdriver');
+const { Key } = require('selenium-webdriver');
 
 const enLocaleMessages = require('../../app/_locales/en/messages.json');
 const { tinyDelayMs, regularDelayMs, largeDelayMs } = require('./helpers');
@@ -143,10 +143,12 @@ describe('MetaMask', function () {
       await driver.findVisibleElement('.qr-code__wrapper');
       await driver.delay(regularDelayMs);
 
-      const accountModal = await driver.findElement('span .modal');
+      // wait for permission modal to be visible.
+      await driver.waitForSelector('span .modal');
       await driver.clickElement('.account-modal__close');
 
-      await driver.wait(until.stalenessOf(accountModal));
+      // wait for account modal to be removed from DOM.
+      await driver.waitForSelector('span .modal', { state: 'detached' });
       await driver.delay(regularDelayMs);
     });
   });
@@ -242,10 +244,10 @@ describe('MetaMask', function () {
     });
 
     it('balance renders', async function () {
-      const balance = await driver.findElement(
-        '[data-testid="wallet-balance"] .list-item__heading',
-      );
-      await driver.wait(until.elementTextMatches(balance, /100\s*ETH/u));
+      await driver.waitForSelector({
+        css: '[data-testid="wallet-balance"] .list-item__heading',
+        text: '100 ETH',
+      });
       await driver.delay(regularDelayMs);
     });
   });
@@ -317,10 +319,10 @@ describe('MetaMask', function () {
         return confirmedTxes.length === 1;
       }, 10000);
 
-      const txValues = await driver.findElement(
-        '.transaction-list-item__primary-currency',
-      );
-      await driver.wait(until.elementTextMatches(txValues, /-1\s*ETH/u), 10000);
+      await driver.waitForSelector({
+        css: '.transaction-list-item__primary-currency',
+        text: '-1 ETH',
+      });
     });
   });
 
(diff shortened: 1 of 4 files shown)


test/e2e/webdriver/driver.js
```diff
@@ -28,8 +29,31 @@ class Driver {
       // xpath locator.
       return By.xpath(locator.xpath);
     } else if (locator.text) {
-      // Providing a text prop, and optionally a tag, will use xpath to look
-      // for an element with the tag that has matching text.
+      // Providing a text prop, and optionally a tag or css prop, will use
+      // xpath to look for an element with the tag that has matching text.
+      if (locator.css) {
+        // When providing css prop we use cssToXPath to build a xpath string
```
brad-decker, line 35: A majority of our usage of `until.elementMatchesText` was using a css selector then waiting until the returned element had matching text. Building an xpath combining CSS selection and text selection more closely mimics Playwright's API (page.waitForSelector('.hello >> text="hi"'). XPath IS slower than CSS selectors, BUT in the case of searching for text, it is not possible without XPath. Combining them into XPath is likely faster than issuing two statements looking for both a CSS selector AND then waiting for text to match.

In the end, though, this entire method (buildLocator) will go away and we will use the much more powerful and expressive playwright selector API.

test/e2e/metamask-ui.spec.js
```diff
@@ -1435,19 +1445,21 @@ describe('MetaMask', function () {
...
+      });
 
-      const txValues = await driver.findElements(
-        '.transaction-list-item__primary-currency',
+      await driver.waitForSelector(
+        {
+          css: '.transaction-list-item__primary-currency',
+          text: '-1.5 TST',
+        },
+        { timeout: 10000 },
```
Gudahtt, line 1461: Why so long of a wait here? Is this the same transaction that's being waited for on the line above this? If so I'd expect it to be fairly quick to locate.

test/e2e/metamask-ui.spec.js
```diff
@@ -1510,17 +1516,11 @@ describe('MetaMask', function () {
     });
 
     it('finds the transaction in the transactions list', async function () {
-      await driver.wait(async () => {
-        const confirmedTxes = await driver.findElements(
-          '.transaction-list__completed-transactions .transaction-list-item',
-        );
-        return confirmedTxes.length === 5;
```
Gudahtt, line 1517: Getting rid of these checks might end up bypassing the intent here, if the previous transaction happens to have the same text as the one this was supposed to wait for. Then the test would immediately find the wrong element and continue on.

It might be safer to restore these.

test/e2e/metamask-ui.spec.js
```diff
@@ -1053,19 +1065,19 @@ describe('MetaMask', function () {
       await driver.clickElement('.advanced-gas-options-btn');
       await driver.delay(regularDelayMs);
 
-      gasModal = await driver.findElement('span .modal');
+      // Wait for gas modal to be visible
+      await driver.waitForSelector('span .modal');
       await driver.delay(regularDelayMs);
     });
 
     it('opens customize gas modal', async function () {
```
Gudahtt, line 1073: Nit: Pre-existing issue, but... this description is inaccurate :thinking:

Gudahtt (commented):
This looks fantastic! This seems like a very good way of abstracting this functionality. I just had a couple of questions about the `metamask-ui.spec.js` changes.

test/e2e/metamask-ui.spec.js
brad-decker: I struggled with this one, I left it in a few cases where it was very hard to tell if the amount of ETH being sent was different. These tests are already not idempotent, relying on the order of operations to be successful. Skipping a test for expediency results in systematic failure of the entire spec after the last test. I'll restore these but I strongly feel like this whole file needs to be broken out into chunks.

brad-decker: Maybe idempotency isn't a necessity in E2E tests, 🤔 it just feels like there is more than one thing being tested out here

Gudahtt: Oh for sure, these tests are a mess. They'll be broken out soon though hopefully!

test/e2e/metamask-ui.spec.js
brad-decker: i didn't know what to do here so I moved the steps from the prior block that actually opened the modal into this step and reworded it slightly.

test/e2e/metamask-ui.spec.js
brad-decker: Done in latest

Gudahtt (approved):
LGTM!

PR #10845, opened 2021-04-07
Adds jest dependency

tmashuang:
Adds jest dependency as per the discussion with the codefi team to write the swaps test using jest + react-testing-library.


Gudahtt (dismissed):
LGTM!

These lavamoat policy changes aren't necessary here (it's just `fsevent` and it's dependencies being removed), but we've been allowing these changes for convenience sake until lavamoat better supports optional dependencies.

Gudahtt (approved):

PR #10846, opened 2021-04-07
Add `--help` flag to changelog script

Gudahtt:
The changelog script now accepts a `--help` flag, which prints a help text explaining how the script works and what each flag does.


ryanml (approved):
++

PR #10847, opened 2021-04-08
Refactor changelog parsing and generation

Gudahtt:
The `auto-changelog.js` script has been refactoring into various different modules. This was done in preparation for migrating this to a separate repository, where it can be used in our libraries as well.

Functionally this should act _mostly_ the same way, but there have been some changes. It was difficult to make this a pure refactor because of the strategy used to validate the changelog and ensure each addition remained valid. Instead of being updated in-place, the changelog is now parsed upfront and stored as a "Changelog" instance, which is a new class that was written to allow only valid changes. The new changelog is then stringified and completely overwrites the old one.

The parsing had to be much more strict, as any unanticipated content would otherwise be erased unintentionally. This script now also normalizes the formatting of the changelog (though the individual change descriptions are still unformatted).

The changelog stringification now accommodates non-linear releases as well. For example, you can now release v1.0.1 *after* v2.0.0, and it will be listed in chronological order while also correctly constructing the `compare` URLs for each release.

Relates to #10752

Manual testing steps:  

To update the changelog with unreleased changes: 
  - Run `yarn update-changelog`
  - See that it puts changes under the "Unreleased" header
 
To update the changelog for a release candidate:
- Run `yarn update-changelog --rc`
- See that it puts changes under the header for the current release candidate

### DIFF
--- a/development/auto-changelog.js
+++ b/development/auto-changelog.js
@@ -1,25 +1,27 @@
 #!/usr/bin/env node
 const fs = require('fs').promises;
-const assert = require('assert').strict;
-const path = require('path');
-const { escapeRegExp } = require('lodash');
-const { version } = require('../app/manifest/_base.json');
-const runCommand = require('./lib/runCommand');
 
-const URL = 'https://github.com/MetaMask/metamask-extension';
+const path = require('path');
+const { version } = require('../app/manifest/_base.json');
+const { updateChangelog } = require('./lib/changelog/updateChangelog');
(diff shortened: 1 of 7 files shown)


brad-decker (approved):
Output looks good for both baseline and --rc flag, code looks much cleaner and broken into fairly easy-to-digest pieces. LGTM 👍🏻 Especially like the amount of comments / docblocks to explain what each piece does.

PR #10849, opened 2021-04-08
Refactoring threebox.spec.js to use fixtures

### LINKED ISSUE #10848: Refactor threebox.spec.js to follow pattern in metamask-extension/test/e2e/tests
This is part of #10754, and is the to-do item for the threebox.spec.js

NiranjanaBinoy:
Fixes: #10848

Explanation:  

- Moved the file to test/e2e/tests
- Updated threebox.spec.js to use fixtures.
- Updated run-all.sh to use the new test file.
- Updated the test/e2e/mock-3box/server.js to be called from the test instead of running it as a script.


test/e2e/mock-3box/threebox-mock-server.js
```diff
@@ -0,0 +1,57 @@
+const http = require('http');
```
Gudahtt, line 1: We should remove the script that this is replacing as well

Gudahtt (approved):
LGTM!

Issue #10755, opened 2021-03-29
Complete the abstraction of, and migration to, our own wrapper/api around selenium


danjm:
Currently, we have our own `Driver` class that wraps many selenium-webdriver methods that we use in our e2e tests. `metamask-extension/test/e2e/webdriver/driver.js`

This instances of this class form all the `driver`s that are used in all of our e2e tests. So anywhere that there is a `driver.method()` call in our e2e tests, we are calling methods defined in the `Driver` class or available on that classes objects by virtue of being defined on the WebDriver instance passed to that class. However, we still do import some methods and objects from the `selenium-webdriver` module.

We need to abstract all methods that we use from selenium-webdriver in an API which ultimately can be decoupled from the driver / e2e test library that backs it.

The goal here is to make it easy to replace seleniuem-webdriver with something else, e.g. playwright.

To complete this task we need to:

- [x] ensure that our driver class has equivalents for all methods from selenium-webdriver (note that, to start, all of these methods will just be wrappers around webdriver)
- [x] double check that these methods can be properly supported by Playwright, which is the top candidate for what we will replace selenium-webdriver with. In particular, the `By` and `until` methods and patterns may require more sophisticated abstractions to create an api that can support selenium-webdriver and its playwright equivalent.
- [x] refactor all our current tests to fully use the methods defined in this api

danfinlay:
Consider publishing docs for external consumption?

Gudahtt:
This is not meant for external consumption; this is purely an internal construct used for our e2e tests.

brad-decker:
Related PRS:
#10802 
#10833
#10844 
#10854 
#10852

PR #10852, opened 2021-04-08
use waitForSelector instead of until

brad-decker:
This PR completes the abstraction of the until utility from Selenium.

When refactoring to use the work completed in #10844 I ran into a few cases where I believe a race condition made tests flakey at best. The waitForSelector method when passed the `{ state: 'detached' }` options bag would try to find an element with the provided selector and wait for it to be removed from the DOM. However, in every case we used the detached flag we are triggering the close of the modal/element before we call `waitForSelector`. If the modal is removed from the DOM **before** the waitForSelector is called then it would timeout. 

To fix this I have added another item into the Element API wrapper that matches the API from playwright. The new utility method is `waitForElementState`. It works in much the same way as the page level `waitForSelector` but the element handle is stable and the race condition is abated. I have left the detached support in waitForSelector because it exists in the playwright APi that it is modeled after.

### DIFF
--- a/test/e2e/incremental-security.spec.js
+++ b/test/e2e/incremental-security.spec.js
@@ -146,10 +146,13 @@ describe('MetaMask', function () {
     });
 
     it('should have the correct amount of eth', async function () {
-      await driver.waitForSelector({
+      const currencyDisplay = await driver.waitForSelector({
         css: '.currency-display-component__text',
         text: '1',
       });
+      const balance = await currencyDisplay.getText();
+
+      assert.strictEqual(balance, '1');
     });
   });
 
@@ -213,10 +216,13 @@ describe('MetaMask', function () {
     });
 
     it('should have the correct amount of eth', async function () {
-      await driver.waitForSelector({
+      const currencyDisplay = await driver.waitForSelector({
         css: '.currency-display-component__text',
         text: '1',
       });
+      const balance = await currencyDisplay.getText();
+
+      assert.strictEqual(balance, '1');
     });
 
     it('should not show a backup reminder', async function () {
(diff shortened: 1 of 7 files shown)


test/e2e/webdriver/driver.js
```diff
@@ -9,13 +9,23 @@ const cssToXPath = require('css-to-xpath');
...
     await element.sendKeys(input);
   };
+  element.waitForElementState = async (state, timeout) => {
+    switch (state) {
+      case 'hidden':
+        return await driver.wait(until.stalenessOf(element), timeout);
+      case 'visible':
+        return await driver.wait(until.elementIsVisible(element), timeout);
+      default:
+        return null;
```
Gudahtt, line 26: Should this throw? I'm having trouble imagining how this could be useful.

test/e2e/incremental-security.spec.js
```diff
@@ -134,22 +135,21 @@ describe('MetaMask', function () {
...
     });
 
     it('should have the correct amount of eth', async function () {
-      const balances = await driver.findElements(
-        '.currency-display-component__text',
-      );
-      await driver.wait(until.elementTextMatches(balances[0], /1/u), 15000);
-      const balance = await balances[0].getText();
-
-      assert.equal(balance, '1');
```
Gudahtt, line 152: I believe the intent of this assertion was to ensure this test doesn't pass if the balance is a number _containing_ the digit 1 (e.g. `1.51`, `100`, etc.). It might be worth preserving this assertion, both here and below where a similar assertion was removed.

Gudahtt:
I think this is the last step in completing #10755

test/e2e/incremental-security.spec.js
brad-decker: will do! Thanks

test/e2e/webdriver/driver.js
brad-decker: Probably, yeah.

Gudahtt (approved):
LGTM!

PR #10854, opened 2021-04-08
add key literals to driver

brad-decker:
This PR aims to improve the abstraction around sending keystrokes/values to input fields. The API aims to match playwright's implementation to allow a more straightforward migration in the near future. There are a few pieces:

1. Add the 'Key' constants to the driver itself, hardcoded instead of using selenium's. These can be later replaced with 'Enter' and 'Backspace' to match the playwright API for `.press`
2. Add `driver.fill` API to the driver that matches playwright's `fill` method. Fill differs from `sendKeys` in that it *replaces* the current value instead of adding to it.
3. Add `driver.press` API to the driver that matches playwright's `press` method. Playwright's implementation differs in that it expects key names like 'Enter' while selenium's expects unicode encoded key values.
4. Wraps all elements returned from driver with `fill` and `press` methods. 

This PR includes the refactor to use these.

### DIFF
--- a/test/e2e/benchmark.js
+++ b/test/e2e/benchmark.js
@@ -3,7 +3,6 @@
 const path = require('path');
 const { promises: fs, constants: fsConstants } = require('fs');
 const ttest = require('ttest');
-const { Key } = require('selenium-webdriver');
 const { withFixtures } = require('./helpers');
 const { PAGES } = require('./webdriver/driver');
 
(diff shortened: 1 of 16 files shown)


Gudahtt (commented):
LGTM! Apart from the conflicts. I can re-review after those are resolved.

Gudahtt (approved):
LGTM!

PR #10855, opened 2021-04-08
Jest config

tmashuang:
Configures jest and lint rules for the `/ui/app/pages/swaps/*.test.js`. 
<strike>Omit the swaps folder from unit tests for now.</strike>
Adds `test:unit:jest` script to run the jest tests isolated to the swap ui folder.
Rewrite `ui/app/pages/swaps/swaps.util.test.js` to jest. Replaces `proxyquire` with `jest.mock` and use `nock` for api mocking.


package.json
```diff
@@ -20,10 +20,11 @@
...
     "dapp-forwarder": "concurrently -k -n forwarder,dapp -p '[{time}][{name}]' 'yarn forwarder' 'yarn dapp'",
     "sendwithprivatedapp": "node development/static-server.js test/e2e/send-eth-with-private-key-test --port 8080",
-    "test:unit": "mocha --exit --require test/env.js --require test/setup.js --recursive './{ui,app,shared}/**/*.test.js'",
+    "test:unit": "mocha --exit --require test/env.js --require test/setup.js --recursive './{ui,app,shared}/**/!(swaps)/*.test.js'",
     "test:unit:global": "mocha --exit --require test/env.js --require test/setup.js --recursive test/unit-global/*.test.js",
-    "test:unit:lax": "mocha --exit --require test/env.js --require test/setup.js --ignore './app/scripts/controllers/permissions/*.test.js' --recursive './{ui,app,shared}/**/*.test.js'",
+    "test:unit:lax": "mocha --exit --require test/env.js --require test/setup.js --ignore './app/scripts/controllers/permissions/*.test.js' --recursive './{ui,app,shared}/**/!(swaps)/*.test.js'",
     "test:unit:strict": "mocha --exit --require test/env.js --require test/setup.js --recursive './app/scripts/controllers/permissions/*.test.js'",
+    "test:unit:jest": "jest",
```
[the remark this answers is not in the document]
tmashuang: I was going to hold off on this after the `/ui/app/pages/swaps/*.test.js` were going to be written. The `test:lax` is what we use for the CI and my understanding is that there needs to be a different config for jest within the CI to run properly as well as a different coverage script. I was going to set up the CI commands for jest in a different task in another PR, if that is ok?
My understanding is that `test:unit` is meant for being locally run, it isn't being used in CI as far as I can tell, and theirs no need to include it right now from my POV. Later `test:unit` will get changed to be `test:unit:jest`.

jest.config.js
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable import/unambiguous */
```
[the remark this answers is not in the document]
tmashuang: Resolved in 0f3ee4d.

jest.config.js
```diff
@@ -0,0 +1,6 @@
...
+module.exports = {
+  clearMocks: true,
```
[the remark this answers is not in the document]
tmashuang: Resolved in 2eee19b.

package.json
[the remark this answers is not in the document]
Gudahtt: Hah, oh, I hadn't realized there are no tests yet! Yeah that all makes sense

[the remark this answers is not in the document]
tmashuang: After thinking about it I think I'll just add the commits onto this PR. It seems within the scope to create the new jest ci scripts into this PR.

adamlaska (commented):
Helo world

package.json
```diff
@@ -14,22 +14,25 @@
...
     "build:test:metrics": "SEGMENT_HOST='http://localhost:9090' SEGMENT_WRITE_KEY='FAKE' SEGMENT_LEGACY_WRITE_KEY='FAKE' yarn build test",
-    "test": "yarn test:unit && yarn lint",
+    "test": "yarn lint && yarn test:unit && yarn test:unit:jest",
     "dapp": "node development/static-server.js node_modules/@metamask/test-dapp/dist --port 8080",
     "dapp-chain": "GANACHE_ARGS='-b 2' concurrently -k -n ganache,dapp -p '[{time}][{name}]' 'yarn ganache:start' 'sleep 5 && yarn dapp'",
     "forwarder": "node ./development/static-server.js ./node_modules/@metamask/forwarder/dist/ --port 9010",
     "dapp-forwarder": "concurrently -k -n forwarder,dapp -p '[{time}][{name}]' 'yarn forwarder' 'yarn dapp'",
     "sendwithprivatedapp": "node development/static-server.js test/e2e/send-eth-with-private-key-test --port 8080",
-    "test:unit": "mocha --exit --require test/env.js --require test/setup.js --recursive './{ui,app,shared}/**/*.test.js'",
+    "test:unit": "mocha --exit --require test/env.js --require test/setup.js --recursive './{ui,app,shared}/**/!(swaps)/*.test.js'",
```
Gudahtt, line 23: Nit: This `!(swaps)` syntax will exclude _all_ directories named `swaps`, which there may be multiple of. It might be better to use the `--ignore` flag here instead. The syntax is simpler to understand too.

ui/app/pages/swaps/swaps.util.test.js
```diff
@@ -1,54 +1,31 @@
...
+jest.mock('../../../lib/storage-helpers.js', () => ({
+  getStorageItem: jest.fn(),
+  setStorageItem: jest.fn(),
+}));
 
-describe('Swaps Util', function () {
-  describe('fetchTradesInfo', function () {
+describe('Swaps Util', () => {
+  afterAll(() => {
+    nock.cleanAll();
```
Gudahtt, line 25: It would be better to call this after each test, rather than after all of them, to ensure they don't incidentally rely upon an interceptor setup for a different test. Similarly, the interceptors being setup in `beforeAll` hooks in various places can be moved to `beforeEach`.

jest.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  restoreMocks: true,
+  setupFiles: ['./test/setup.js', './test/env.js'],
+  testMatch: ['**/ui/app/pages/swaps/**/?(*.)+(test).js'],
+};
```
dan437, line 5: Can we please add coverageThreshold for Jest? Something like we have in the Controllers repo: https://github.com/MetaMask/controllers/blob/develop/package.json#L108
We can start with whatever numbers we get now from the coverage report and slowly but surely keep increasing them with new code.

Once we have this minimum threshold set up, I can send a PR for setting up enforcement of that threshold on every Git push. It could be done e.g. via the `husky` node module: https://github.com/typicode/husky
image

Then whoever is contributing to the Swaps feature (and later on to any feature in this repo) will have to add enough unit tests, otherwise they won't be able to push their code.

Let me know what you think.

.eslintrc.js
```diff
@@ -107,11 +107,16 @@ module.exports = {
     },
     {
       files: ['**/*.test.js'],
+      excludedFiles: ['**/swaps/**/*.test.js'],
```
Gudahtt, line 110: Nit: we should probably specify `pages/swaps` here for consistency with the other config block.

Or even better, we could be more explicit in both places to ensure it only applies to this one directory:
```suggestion
      excludedFiles: ['ui/pages/swaps/**/*.test.js'],
```

.circleci/config.yml
```diff
@@ -501,6 +501,9 @@ jobs:
       - run:
           name: test:coverage
           command: yarn test:coverage
+      - run:
+          name: test:coverage:jest
+          command: yarn test:coverage:jest
```
Gudahtt, line 506: It would be nice to store these coverage reports somewhere :thinking: But `coverage` is taken. Maybe we can set the `coverageDir` to `jest-coverage` or something? Then we could persist it a couple lines down from here.

jest.config.js
Gudahtt: Setting a minimum coverage goal and ratcheting it up is a good idea - that's basically what we've done with the `test:unit:strict` and `test:unit:lax` scripts (the "strict" ones have 100% coverage).

Not sure it needs to go in at this point though? The coverage is incredibly low at the moment. Seems like it could wait for another PR.

The Husky suggestion is bound to be controversial :sweat_smile: Some people find husky very irritating. Certainly a discussion to have elsewhere. I've been considering removing the husky setup from the controllers repo and replacing it with something where the git hook is still version-controlled but is opt-in rather than automatic, so that people who prefer not using it or have their IDE running the linter/tests on each change already don't have to add `--no-verify` to every command.

Gudahtt (approved):
LGTM!

PR #10857, opened 2021-04-08
Adjust renderWithProvider to accommodate redux-less components

tmashuang:
This adds a conditional to the `renderWithProvider()` render components that aren't connected to the `store` to render without the `Provider` and `Router`.


Gudahtt (approved):
LGTM!

PR #10867, opened 2021-04-09
Quote globs in prettier scripts

rekmarks:
The globs in our Prettier scripts should be quoted. This PR ensures that they are.

> Don’t forget the quotes around the globs! The quotes make sure that Prettier CLI expands the globs rather than your shell, which is important for cross-platform usage.

https://prettier.io/docs/en/cli.html


Gudahtt (approved):
LGTM!

PR #10868, opened 2021-04-09
Add jest coverage

tmashuang:
This will add coverage for any tests ran in jest under the `test:coverage:jest` command, which is currently being used in CI. I set the values to the current test coverage in `ui/app/pages/swaps`.


Gudahtt (approved):
LGTM!

PR #10878, opened 2021-04-13
Build - flat build system

kumavis:
### Proposal:
- no longer build bundles that need to be size-capped, factored/de-duped, and remapped back to source.
- "build" many individual module definitions closely matching their source and import the ones needed for the context (background, ui, ...)

here "flat" means many separate files, sometimes in a deeply nested path, instead of in a series of bundles

layout can either match source, as seen in the graphic below (`node_modules/abc/node_modules/xyz/index.js`)
or can match package label (`xyz@1.2.3/index.js`)
or some other format.

this should **not** impact:
- our ability to use sourcemaps in sentry

this should improve:
- our debugging experience, thanks to smaller files that are comprehensible even without sourcemaps (optional)
- improve our final file size, due to optimal deduping

### Ask
:thinking: currently the flat file approach seems to impact startup performance 
please evaluate the experience on your machine and report
:pray: :pray: :pray: 

### Output format
[detailed dist file structure](https://gist.github.com/kumavis/3545f6878d942a4dd52b9b70afae2b82)
image

image


kumavis:
even with a manifest of everything  that needed to be loaded, performance of loading in the js was too bad and killed my utopia
sad sad

PR #10882, opened 2021-04-13
Swaps: Show a network name dynamically in a tooltip

dan437:
## Explanation
Show a selected network for Swaps dynamically in a tooltip. Only these 3 are supported at the moment:
- Ethereum
- BSC
- Testnet

## Manual testing steps
  - Open the MetaMask extension
  - Enter your password
  - Click on the "Swap" button
  - Select "From" and "To" tokens
  - Click on the "Review Swap" button and wait
  - On the page with a quote you can hover your mouse on the tooltip next to "Estimated network fee" and after it opens it shows the network dynamically: "The network fee covers the cost of processing your swap and storing it on the Ethereum network." It can either be "Testnet", "Ethereum" or "BSC" at the moment.

## Screenshots
Showing "Testnet network" when using LOCALHOST_CHAIN_ID:
image

Showing "Ethereum network" by default if there is no localized network name:
image


app/_locales/en/messages.json
```diff
@@ -1781,7 +1781,7 @@
     "description": "$1 is the number of quotes that the user can select from when opening the list of quotes on the 'view quote' screen"
   },
   "swapNetworkFeeSummary": {
-    "message": "The network fee covers the cost of processing your swap and storing it on the Ethereum network. MetaMask does not profit from this fee."
+    "message": "The network fee covers the cost of processing your swap and storing it on the $1 network. MetaMask does not profit from this fee."
```
Gudahtt, line 1784: Note that this will break for any translations that don't include a `$1`. I'd suggest going through each translation to either replace `Ethereum` with `$1`, or if that seems difficult then deleting this string so that we can ask for a new translation.

shared/constants/swaps.js
```diff
@@ -42,6 +43,12 @@ export const TEST_ETH_SWAPS_TOKEN_OBJECT = {
   iconUrl: TEST_ETH_TOKEN_IMAGE_URL,
 };
 
+export const SWAPS_CHAIN_ID_TO_NETWORK_NAME_MAP = {
+  [MAINNET_CHAIN_ID]: 'Ethereum',
```
Gudahtt, line 47: These strings should be localized. "Test" probably isn't a very good name in all locales for example.

dan437: I believe that network names should be the same across all languages. I like to think of it as names of companies.. we also don't translate them. 'Test' is a special one just for testing purposes and I'm open to changing it if there is something better. I thought that devs are used to hear Test Net(work), so that's why I chose "Test".

app/_locales/en/messages.json
dan437: I've talked with @danjm and he mentioned I should only care about `en` translations. 

I've actually just tested what you mentioned and it should work just fine for other languages without `$1`. I just removed the `$1` from the English translation and it shows whatever text is there right now (see below on the screenshot). That means that other languages will see `Ethereum` until their strings contain `$1`, which should be done by our localization team.
image

Gudahtt: That will flood sentry with errors. Our localization system expects each message to have the same number of substitutions across all locales

dan437: Good to know. I've tried to replace all `Ethereum` occurrences with `$1`, but I don't see `Ethereum` in the ko/messages.json file: 
image

shared/constants/swaps.js
dan437: I've changed it to "Testnet", because I believe that some translations don't use the word "network" after our dynamic value, so it would just say something like: "The network fee covers the cost of processing your swap and storing it on the Test".

app/_locales/en/messages.json
danjm: @dan437 Sorry for misleading you. When adding new translations, we only need to worry about `en` translations.When updating existing translations, if there is an obvious change to make to non-english translations, we can do that manually. If not, it is usually best to add a translation as new.

danjm: According to google translate, 이더 리움 is korean for ethereum

danjm: which appears to be the fifth set of characters from the left

Gudahtt: This does hint that localizing the actual network names is something we should do as well.

dan437: That's interesting, thanks, Dan and Mark! I've replaced `이더리움` with `$1` and will make sure to localize network names as well.

ui/app/pages/swaps/fee-card/__snapshots__/fee-card.test.js.snap
```diff
@@ -0,0 +1,175 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
dan437, line 1: From the official Jest site: "The snapshot artifact should be committed alongside code changes, and reviewed as part of your code review process."

ui/app/pages/swaps/fee-card/fee-card.test.js
```diff
@@ -0,0 +1,31 @@
...
+      isBestQuote: true,
+      numberOfQuotes: 6,
+      onQuotesClick: jest.fn(),
+      tokenConversionRate: 0.015,
+      chainId: MAINNET_CHAIN_ID,
+      ...customProps,
+    };
+  };
+
+  it('renders the component with initial props', () => {
```
dan437, line 27: I will be adding many more unit tests for the Swaps feature in upcoming PRs.

danjm (dismissed):
This looks good to me!

ui/app/pages/swaps/fee-card/fee-card.js
```diff
@@ -26,6 +28,12 @@ export default function FeeCard({
     bestQuoteText = t('swapBetterQuoteAvailable');
   }
 
+  const getTranslatedNetworkName = () => {
+    const networkNameKey =
+      SWAPS_CHAIN_ID_TO_NETWORK_NAME_KEY_MAP[chainId] || 'networkNameEthereum';
+    return t(networkNameKey) || 'Ethereum'; // If there are no translated network names, it defaults to 'Ethereum'
```
Gudahtt, line 34: The translation system already defaults to using the `en` translation - this fallback is redundant.

app/_locales/ko/messages.json
```diff
@@ -1654,7 +1654,7 @@
     "description": "Provides information about the fee that metamask takes for swaps. $1 is a decimal number."
   },
   "swapNetworkFeeSummary": {
-    "message": "네트워크 요금에는 스왑을 처리하고 이더리움 네트워크에 보관하는 비용이 적용됩니다. MetaMask는 이 요금을 이용해 이득을 얻지 않습니다."
+    "message": "네트워크 요금에는 스왑을 처리하고 $1 네트워크에 보관하는 비용이 적용됩니다. MetaMask는 이 요금을 이용해 이득을 얻지 않습니다."
```
Gudahtt, line 1657: It might be worth adding `이더리움` as a translation for `networkNameEthereum` so we don't have a regression here.

ui/app/pages/swaps/fee-card/fee-card.js
dan437: Awesome, will remove it.

Gudahtt, line 34: Nit: In general we should avoid passing variables to `t`, because our localized string linting script then has a more difficult time discovering whether the string is used or not.  There is a 'strict mode' that we've been slowly introducing one file at a time that assumes `t` is passed a string literal, so any use of a variable will make that task more difficult.

The alternative I'd suggest is deleting `SWAPS_CHAIN_ID_TO_NETWORK_NAME_KEY_MAP`, and replacing it with a function that accepts `t` and the chain ID, and returns the correct network name.

ui/app/pages/swaps/fee-card/fee-card.js
```diff
@@ -26,6 +28,12 @@ export default function FeeCard({
     bestQuoteText = t('swapBetterQuoteAvailable');
   }
 
+  const getTranslatedNetworkName = () => {
+    const networkNameKey =
+      SWAPS_CHAIN_ID_TO_NETWORK_NAME_KEY_MAP[chainId] || 'networkNameEthereum';
```
Gudahtt, line 33: Maybe we should throw here instead of defaulting to `'networkNameEthereum'`? :thinking: This should be unreachable right now, and in the future could only be hit if we add support for a fourth network in swaps but forget to add a network name. That scenario would be worth throwing an error for.

Unless there's some case I'm missing.

app/_locales/ko/messages.json
dan437: Done.

ui/app/pages/swaps/fee-card/fee-card.js
dan437: Done.

danjm: > In general we should avoid passing variables to t, because our localized string linting script then has a more difficult time discovering whether the string is used or not
This is a good point, I always forget this.

> The alternative I'd suggest is deleting SWAPS_CHAIN_ID_TO_NETWORK_NAME_KEY_MAP, and replacing it with a function that accepts t and the chain ID, and returns the correct network name.
Wouldn't that still result in a variable passed to `t`?

danjm: > Wouldn't that still result in a variable passed to t?

Ah, no, I see. The function would map chainId to calls to `t` using string literals

dan437: So, there is https://github.com/i18next/react-i18next , which is one of the best internationalization frameworks in the world for React / React Native. Interpolation (passing dynamic values into translations) is one of the most used functionalities there: https://www.i18next.com/translation-function/interpolation

I believe we should support passing dynamic values into translations, since it's a very common thing to do. With i18next it's done like this:
```
{
    "key": "{{what}} is {{how}}"
}
i18next.t('key', { what: 'i18next', how: 'great' });
```

Gudahtt: We're not talking about interpolation here. Our translation helper does support interpolation.

ui/app/pages/swaps/fee-card/fee-card.js
dan437: It sounds fine to me. What about something like this? 
```
if (!networkNameKey) {
  throw new Error('This network is not supported for token swaps');
}
```

Gudahtt: That works!

ui/app/pages/swaps/fee-card/fee-card.js
dan437: We've just clarified it over Slack - the issue is not about passing variables in the second param to `t`, but in the first one. Will create a function as suggested.

danjm (dismissed):

danjm (approved):

PR #10883, opened 2021-04-13
Handling infura blockage

ryanml:
Fixes: https://github.com/MetaMask/MetaMask-planning/issues/42

Screen Shot 2021-04-14 at 12 07 00 PM

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -879,7 +879,7 @@
     "message": "Info & Help"
   },
   "infuraBlockedNotification": {
-    "message": "MetaMask is unable to reach our hosted blockchain connection. Review possible reasons $1.",
+    "message": "MetaMask is unable to connect to the blockchain host. Review possible reasons $1.",
     "description": "$1 is a clickable link."
   },
   "initialTransactionConfirmed": {
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -880,7 +880,7 @@
   },
   "infuraBlockedNotification": {
     "message": "MetaMask is unable to connect to the blockchain host. Review possible reasons $1.",
-    "description": "$1 is a clickable link."
+    "description": "$1 is a clickable link with with text defined by the 'here' key"
   },
   "initialTransactionConfirmed": {
     "message": "Your initial transaction was confirmed by the network. Click OK to go back."
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -878,6 +878,10 @@
   "infoHelp": {
     "message": "Info & Help"
   },
+  "infuraBlockedNotification": {
+    "message": "MetaMask is unable to reach our hosted blockchain connection. Review possible reasons $1.",
+    "description": "$1 is a clickable link."
+  },
   "initialTransactionConfirmed": {
     "message": "Your initial transaction was confirmed by the network. Click OK to go back."
   },
--- a/app/scripts/controllers/network/network.js
+++ b/app/scripts/controllers/network/network.js
@@ -17,17 +17,19 @@ import {
   NETWORK_TYPE_TO_ID_MAP,
   MAINNET_CHAIN_ID,
   RINKEBY_CHAIN_ID,
-  INFURA_BLOCKED_MESSAGE,
+  INFURA_BLOCKED_KEY,
 } from '../../../../shared/constants/network';
 import {
   isPrefixedFormattedHexString,
   isSafeChainId,
 } from '../../../../shared/modules/network.utils';
+import getFetchWithTimeout from '../../../../shared/modules/fetch-with-timeout';
 import createMetamaskMiddleware from './createMetamaskMiddleware';
 import createInfuraClient from './createInfuraClient';
 import createJsonRpcClient from './createJsonRpcClient';
 
 const env = process.env.METAMASK_ENV;
+const fetchWithTimeout = getFetchWithTimeout(30000);
 
 let defaultProviderConfigOpts;
 if (process.env.IN_TEST === 'true') {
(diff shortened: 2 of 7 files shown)


app/scripts/controllers/preferences.js
```diff
@@ -679,6 +681,25 @@ export default class PreferencesController {
...
+    });
+  }
+
+  /**
+   * A setter for the `infuraBlocked` property
+   * @param {boolean} isBlocked - Bool indicating whether Infura is blocked
+   * @returns {Promise<boolean>} A promise of the updated blockage status
+   */
+  _setInfuraBlocked(isBlocked) {
+    this.store.updateState({ infuraBlocked: isBlocked });
```
Gudahtt, line 699: Nit: we should avoid a state update here if `infuraBlocked` is already set to the same value as `isBlocked`, to avoid refreshing the UI when there is no change

app/scripts/controllers/network/network.js
```diff
@@ -52,6 +53,10 @@ export const NETWORK_EVENTS = {
   NETWORK_DID_CHANGE: 'networkDidChange',
   // Fired when the actively selected network *will* change
   NETWORK_WILL_CHANGE: 'networkWillChange',
+  // Fired when Infura returns an error indicating no support
+  INFURA_IS_BLOCKED: 'infuraIsBlocked',
+  // Fired when Infura returns no error, indicating support
```
Gudahtt, line 58: Nit:
```suggestion
  // Fired when not using an Infura network or when Infura returns no error, indicating support
```

app/_locales/en/messages.json
```diff
@@ -878,6 +878,10 @@
   "infoHelp": {
     "message": "Info & Help"
   },
+  "infuraBlockedNotification": {
+    "message": "MetaMask is unable to connect to the blockchain host. Review possible reasons $1.",
+    "description": "$1 is a clickable link."
```
Gudahtt, line 883: The most important thing about `$1` for translators to understand is that `$1` will have the text "here" as a link.
```suggestion
    "description": "$1 is a clickable link with with text defined by the 'here' key"
```

Gudahtt (commented):
Looks good aside from the redundant state update part! Though I didn't test this locally.

app/scripts/controllers/network/network.js
ryanml: suggested change implemented

app/_locales/en/messages.json
ryanml: suggested change implemented

app/scripts/controllers/preferences.js
ryanml: done, will not be unnecessarily set now

app/scripts/controllers/preferences.js
```diff
@@ -679,6 +681,31 @@ export default class PreferencesController {
...
+  /**
+   * A setter for the `infuraBlocked` property
+   * @param {boolean} isBlocked - Bool indicating whether Infura is blocked
+   * @returns {Promise<boolean>} A promise of the updated blockage status
+   */
+  _setInfuraBlocked(isBlocked) {
+    const { infuraBlocked } = this.store.getState();
+
+    if (infuraBlocked === isBlocked) {
+      return Promise.resolve(isBlocked);
```
Gudahtt, line 702: Why does this return a Promise if it's synchronous :thinking:

shared/constants/network.js
```diff
@@ -96,3 +96,6 @@ export const NATIVE_CURRENCY_TOKEN_IMAGE_MAP = {
   [TEST_ETH_SYMBOL]: TEST_ETH_TOKEN_IMAGE_URL,
   [BNB_SYMBOL]: BNB_TOKEN_IMAGE_URL,
 };
+
+export const INFURA_BLOCKED_MESSAGE =
+  'EthQuery - RPC Error - This service is not available in your country';
```
Gudahtt, line 101: Curious about this error message format - I don't think I've seen this type before. I guess you saw this in testing?

ryanml: As this is an EthQuery call, the error will be prefixed with `EthQuery - RPC Error - ` per: https://github.com/ethereumjs/eth-query/blob/master/index.js#L67

Then per https://consensys.slack.com/archives/GV90KPED9/p1618353684021400?thread_ts=1618348658.016900&cid=GV90KPED9, the full message will be `EthQuery - RPC Error - This service is not available in your country`

app/scripts/controllers/preferences.js
ryanml: Good catch! I had this in line with other setter methods, but given it's usage only in private context, no need to return Promises here

shared/constants/network.js
Gudahtt: Thanks, I was just able to trace it down. It all makes sense now. I was expecting the status code to be included with the error somewhere but apparently we don't do that yet, so I've opened an issue for that: https://github.com/MetaMask/eth-json-rpc-infura/issues/45

ryanml: ++

ui/app/pages/home/home.component.js
```diff
@@ -74,10 +76,12 @@ export default class Home extends PureComponent {
     originOfCurrentTab: PropTypes.string,
     disableWeb3ShimUsageAlert: PropTypes.func.isRequired,
     pendingConfirmations: PropTypes.arrayOf(PropTypes.object).isRequired,
+    infuraBlocked: PropTypes.bool.isRequired,
   };
 
   state = {
     mounted: false,
+    canShowBlockageNotification: true,
```
Gudahtt, line 84: This might be pretty annoying for users, as it'll be restored every time they navigate back to the home page. e.g. "Home => Settings => Home" will force them to dismiss it again.

But nonetheless, I like this solution for now. It's simple and pragmatic. We can improve the dismissal logic later if we feel the need. Given that there's not a lot users can do in this state, it might not be super important to improve this.

ryanml: My exact thoughts here as well, I think when we approach improving network connectivity UX as a whole, this should definitely be a point for refactoring

Gudahtt (commented):
I've discovered a problem: the `net_version` request never goes to Infura. We intercept it and return the constant we have for the network version.

The two options I can see are checking the block tracker requests for fetch failures, or making a separate fetch request each time the network changes. The former is much nicer idea because it saves network traffic but unfortunately the fetch error details aren't returned by the block tracker at the moment - that would require an update to that library. Which we should absolutely do, but we don't have time for that now. So, direct `fetch` it is I suppose.

I'd suggest making it a non-blocking fetch as well, so that we don't slow down the network switch process.

I created a mock Infura server to test this with, which is available here: https://github.com/MetaMask/metamask-extension/compare/mock-infura . It also includes a patch to update `eth-json-rpc-infura` to send request to localhost instead of to Infura. That patch might not be necessary if you're calling `fetch` directly but it was important for testing anything that relies upon our provider.

app/scripts/controllers/network/network.js
```diff
@@ -235,6 +251,31 @@ export default class NetworkController extends EventEmitter {
...
+      const response = await fetchWithTimeout(rpcUrl, {
+        method: 'POST',
+        body: JSON.stringify({
+          jsonrpc: '2.0',
+          method: 'eth_blockNumber',
+          params: [],
+          id: 1,
+        }),
+      });
+      if (response.ok) {
```
Gudahtt, line 266: We might also want to listen in this function for network changes, and bail here without emitting any event if the network has switched since the function began. Otherwise this might resolve after a subsequent switch, setting this flag on the wrong network.

app/scripts/controllers/network/network.js
```diff
@@ -235,6 +251,45 @@ export default class NetworkController extends EventEmitter {
   // Private
   //
 
+  async _checkInfuraAvailability(network) {
+    const rpcUrl = `https://${network}.infura.io/v3/${this._infuraProjectId}`;
+
+    let networkChanged = false;
+    this.on(NETWORK_EVENTS.NETWORK_DID_CHANGE, () => {
```
Gudahtt, line 258: Nit: This should probably use `once` instead of `on`, since we only care about the first time we see this

```suggestion
    this.once(NETWORK_EVENTS.NETWORK_DID_CHANGE, () => {
```

ryanml: Makes sense, fixed

Gudahtt (approved):
LGTM!

PR #10887, opened 2021-04-15
Build: is bg-libs being used?

kumavis:
will CI fail?

also see https://github.com/MetaMask/metamask-extension/pull/10888 https://github.com/MetaMask/metamask-extension/pull/10889


kumavis:
answer is that in bg-libs in tests we stub the only thing thats in there
in ui-libs we only test homt.html

PR #10895, opened 2021-04-16
Fix 10458 - Understand where to get support

### LINKED ISSUE #10458: Understand where to get support
As a MetaMask Extension user, I'd like to better understand what the official support channels for MetaMask are and understand that fake support channels contain scams, so that my funds will be more secure.

**description**

Telegram scams seem to be growing as one of the primary scam vectors with the rise of new more sophisticated and official-looking MetaMask channels.

One user in particular shared concern that he didn't know where to find support, so he went looking on Telegram. One way we could reduce scams is by making the official support in the extension more prominent and accessible.

We could also add language around only seek support from official support channels to avoid scams to the onboarding journey.

Needs designs.

darkwing:
Fixes: #10458

Explanation:  

Provides multiple new ways to get to official support.

## Unlock screen
LockScreenHelp

## Home Screen
TokenHelp

## Account Menu Dropdown
SupportMenu


ui/app/pages/home/index.scss
```diff
@@ -124,4 +124,13 @@
     height: 100%;
     width: 100%;
   }
+
+  &__support {
+    padding: 10px 0 20px 0;
+    text-align: center;
+
+    a {
+      color: $Blue-500;
```
brad-decker, line 133: [REQ] Instead of $Blue-500 could we use $primary-1 -- its the same value but matches the primary theme variables. Sorry its confusing :(

brad-decker (changes requested):
one minor nit request, this looks good to me!

darkwing:
Sorry for using the wrong var!  Updated!

brad-decker (dismissed):
LGTM

rachelcope (dismissed):
This looks great! Two very small things:
1) update form line to blue, instead of orange
2) update link copy to "or import using seed phrase"
image

darkwing:
Updated @rachelcope !

BlueUnderlineNewText

rachelcope:
Great! If it's not too late, can you make the same blue as the button? Thanks!

ui/app/components/ui/text-field/text-field.component.js
```diff
@@ -24,7 +24,7 @@ const styles = {
   materialFocused: {},
   materialUnderline: {
     '&:after': {
-      borderBottom: '2px solid #f7861c',
+      borderBottom: `2px solid ${'$primary-1'}`,
```
brad-decker, line 27: wait what? does this work?

darkwing: Nope, simply didn't push my hard-coded update.  Was left over from testing.

brad-decker: aw man it doesn't lol. 😢

brad-decker (approved):

darkwing:
YayYay

rachelcope:
Looks great! 🚢

PR #10900, opened 2021-04-19
Increase Jest unit test coverage for the Swaps feature to ~25%

dan437:
## Explanation
I've added the most popular way to test React components - snapshot testing, starting with happy paths. I've combined it with targeted assertions for specific things as well.

The goal is to increase coverage at least to 90%, so next PRs will cover:
- Testing with different sets of props
- Covering more files
- Testing events (click, typing, submit, client-side validations, etc.)

I've created a new folder `test/jest`, where we have the `setup.js` file for setting up React Testing Library with Jest assertions and then the `test/jest` folder contains files with functions that are reusable across Jest tests (rendering, returning a mock Redux store). I believe that all things for Jest except for the tests themselves should be in the `test/jest` folder, so it's all nicely organized in one place.

## Manual testing steps
- To run Jest tests in the silent watch mode you can run this: `yarn test:unit:jest:watch:silent` (it will not show a plenty of React warnings, since it's the silent mode). You can also run `yarn test:unit:jest:watch`
- For Jest coverage that is only set up for the Swaps feature now you can run: `npm run test:coverage:jest`
- If you only want to run tests for the Swaps feature, you can open `jest.config.js` and use this line: `testMatch: ['<rootDir>/ui/app/**/swaps/**/*.test.js'],`

## Screenshots
Increased test coverage for the Swaps feature:
image

All Swaps tests are green:
image

All Jest tests are green with these changes:
image

CircleCI is red if snapshots are longer than 50 lines:
image


test/jest/rendering.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
```
brad-decker, line 1: We have a similar helper in `test/lib/render-helpers` which we should probably use instead of adding a new file?

jest.config.js
```diff
@@ -3,12 +3,13 @@ module.exports = {
...
-      lines: 11.76,
-      statements: 11.78,
+      branches: 21.24,
+      functions: 23.01,
+      lines: 27.19,
+      statements: 27.07,
     },
   },
   setupFiles: ['./test/setup.js', './test/env.js'],
+  setupFilesAfterEnv: ['./test/jest/setup.js'],
```
brad-decker, line 13: How does this differ from setupFiles? Can we not set up jest-dom in one of our existing files?

test/jest/mock-store.js
```diff
@@ -0,0 +1,69 @@
+import { MAINNET_CHAIN_ID } from '../../shared/constants/network';
+
+export const createSwapsMockStore = () => {
```
brad-decker, line 3: I like having a mock store for tests! Do we think there is value in creating specific store mocks (swaps, vs general state?) I think instead of separating by feature it might make sense to separate by a scenario where it makes sense. Like "mainnetAcctWithMultipleTokens" or "customNetworkPowerUser" -- this is just brainstorming atm! 

also related but unrelated: #10895 introduces a localization "mock" to the provider

ui/app/__mocks__/react-router-dom.js
```diff
@@ -0,0 +1,6 @@
+const originalModule = jest.requireActual('react-router-dom');
```
brad-decker, line 1: is this an automock? Does jest look for these in every sub-level folder? I thought it only looked for node_modules mocks at the app root.

ui/app/pages/swaps/actionable-message/actionable-message.test.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+
+import { renderWithProvider } from '../../../../../test/jest';
+import ActionableMessage from '.';
+
+describe('ActionableMessage', () => {
+  const createProps = (customProps = {}) => {
```
brad-decker, line 7: Is there any reason that this method is created inside the describe block?

ui/app/pages/swaps/awaiting-swap/__snapshots__/awaiting-swap.test.js.snap
```diff
@@ -0,0 +1,958 @@
...
+  <div
+    class="awaiting-swap"
+  >
+    <div
+      class="awaiting-swap__content"
+    >
+      <div
+        style="z-index: 0;"
+      >
+        <svg
```
brad-decker, line 14: this is one where we should break apart this snapshot, hopefully in such a way that we don't end up covering the SVG / animation with a snapshot. Polygons. Polygons everywhere.

ui/app/pages/swaps/awaiting-swap/awaiting-swap.test.js
```diff
@@ -0,0 +1,32 @@
...
+    };
+  };
+  const store = configureMockStore()(createSwapsMockStore());
+
+  it('renders the component with initial props', () => {
+    const { container, getByText } = renderWithProvider(
+      <AwaitingSwap {...createProps()} />,
+      store,
+    );
+    expect(getByText('[swapProcessing]')).toBeInTheDocument();
```
brad-decker, line 29: This [swapProcessing] string should be the actual text value we'd expect to see, imo. The change mentioned in the localization mock will make that a possibility.

brad-decker (changes requested):
Took a first pass at this to give some early feedback and ask a few questions

test/jest/rendering.js
dan437: Hi, I actually deliberately took it from there, because I believe it's better to have all things for Jest in the `test/jest` folder, except the Jest tests themselves. I don't think we should mix shared functions in the same files for Mocha and Jest tests. 

Also, our test folder looks like this:
image

There are folders like `data`, `helpers`, `lib`, `mocks`, `stub` and it's not really immediately clear if they have code for Mocha unit tests, Jest unit tests, e2e tests or upcoming Playwright e2e tests or some combination of these. With the `test/jest` folder we can have all shared functions for Jest nicely organized in one place, so it will be easier to do Jest testing.

I will happily search + replace all paths with `test/lib/render-helpers` to `test/jest` for the `renderWithProvider` function and move all Jest-specific shared functions to `test/jest`.

jest.config.js
dan437: I've tried to load the file in `setupFiles`, but I'm getting this error for many tests:
image

With `setupFilesAfterEnv` it works correctly. `setupFilesAfterEnv` is also mentioned on the official react testing library site: https://github.com/testing-library/jest-dom#usage

test/jest/mock-store.js
dan437: Nice, will check the localization "mock", so we can get translations into our unit tests instead of just keys!

About creating a mock store for a feature vs one general state.. I think a general state could be pretty big if it would contain data for all features. I would do it per feature and if there is some shared state across features, it can be abstracted in the mock-store.js file into its own function and used in feature specific functions like `createSwapsMockStore`. For example:
```
export const createSwapsMockStore = () => {
  ...createGeneralMockStore(),
  swapsKey1: 'swapsValue1',
  swapsKey2: 'swapsValue2',
}
```

For specific scenarios within a feature there is always an option to call e.g. `createSwapsMockStore` to get default feature state and then modify that state in the code. For example:
```
const swapsMockStore = createSwapsMockStore();
swapsMockStore.metamask.provider.chainId = BSC_CHAIN_ID;
swapsMockStore.metamask.selectedAddress: 'differentAddress0x0dcd5d886577d5081b0c52e242ef29e70be3e7bc'
```

ui/app/__mocks__/react-router-dom.js
dan437: This behaves like if you would have `"automock": true` in jest config, except it only mocks this node module and only if that module is imported somewhere in the `ui/app` folder. More about mocking node modules in Jest here: https://jestjs.io/docs/manual-mocks#mocking-node-modules

Without this we would need to do jest.mock('react-router-dom') in many React files. With this we just do it once here and then this mock version will be used every time we import that module in our React files:
```import { useHistory } from 'react-router-dom';```

File: ui/app/pages/swaps/actionable-message/actionable-message.test.js (hunk 5 above)
Comment (dan437) [reply ui/app/pages/swaps/actionable-message/actionable-message.test.js:7]: No reason, I will move it outside.

File: ui/app/pages/swaps/awaiting-swap/__snapshots__/awaiting-swap.test.js.snap (hunk 6 above)
Comment (dan437) [reply ui/app/pages/swaps/awaiting-swap/__snapshots__/awaiting-swap.test.js.snap:14]: Haha yes! This is a great example when taking a snapshot of the whole component is not easily readable.

File: ui/app/pages/swaps/awaiting-swap/awaiting-swap.test.js (hunk 7 above)
Comment (dan437) [reply ui/app/pages/swaps/awaiting-swap/awaiting-swap.test.js:29]: Absolutely agree. I have this on my todo list to make sure we use real translations instead of keys in our unit tests. I will try to do that as part of this PR.

File: ui/app/__mocks__/react-router-dom.js (hunk 4 above)
Comment (brad-decker) [reply ui/app/__mocks__/react-router-dom.js:1]: cool, i didn't realize it would mock node_modules via relative path like that. neat.

File: test/jest/index.js [hunk 8]
```diff
@@ -0,0 +1,4 @@
+import { createSwapsMockStore } from './mock-store';
+import { renderWithProvider } from './rendering';
+
+export { createSwapsMockStore, renderWithProvider };
```
dan437, line 4: FYI, the reason for this file is to export all shared functions from `test/jest` and then when we need to import some of them in our tests, we only need to import this file instead of multiple files.

ui/app/pages/swaps/actionable-message/actionable-message.test.js
dan437: Done in all tests.

ui/app/pages/swaps/awaiting-swap/__snapshots__/awaiting-swap.test.js.snap
dan437: Done.

ui/app/pages/swaps/searchable-item-list/searchable-item-list.test.js
```diff
@@ -0,0 +1,46 @@
...
+    itemsToSearch: [
+      {
+        iconUrl: 'iconUrl',
+        selected: true,
+        primaryLabel: 'primaryLabel',
+        secondaryLabel: 'secondaryLabel',
+        rightPrimaryLabel: 'rightPrimaryLabel',
+        rightSecondaryLabel: 'rightSecondaryLabel',
+      },
+    ],
```
tmashuang, line 19: The `fuseSearchKeys` prop is require by ListItemSearch, https://github.com/MetaMask/metamask-extension/blob/1e44c34e1e0f7d70ef5841d04eb8da6e9e23b94f/ui/app/pages/swaps/searchable-item-list/list-item-search/list-item-search.component.js#L83

https://github.com/MetaMask/metamask-extension/blob/76a2a9bb8b6ea04025328d36404ac3b59121dfc8/ui/app/pages/swaps/searchable-item-list/searchable-item-list.js#L65-L70

An example is 
```
  "fuseSearchKeys": [
    {
      "name": "name",
      "weight": 0.499
    },
    {
      "name": "symbol",
      "weight": 0.499
    },
    {
      "name": "address",
      "weight": 0.002
    }
  ],
```

ui/app/pages/swaps/fee-card/fee-card.test.js
```diff
@@ -1,31 +1,43 @@
...
-      isBestQuote: true,
-      numberOfQuotes: 6,
-      onQuotesClick: jest.fn(),
-      tokenConversionRate: 0.015,
-      chainId: MAINNET_CHAIN_ID,
-      ...customProps,
-    };
+const createProps = (customProps = {}) => {
+  return {
+    primaryFee: '1 ETH',
```
tmashuang, line 9: The `primaryFee` prop is an object. https://github.com/MetaMask/metamask-extension/blob/16f533e6d9edf0c29b5fb103ab26bfd24d70494c/ui/app/pages/swaps/fee-card/fee-card.js#L174-L177

ui/app/pages/swaps/fee-card/fee-card.test.js
```diff
@@ -1,31 +1,43 @@
...
+    secondaryFee: '2500 USD',
```
tmashuang, line 10: The `secondaryFee` prop is an object. https://github.com/MetaMask/metamask-extension/blob/16f533e6d9edf0c29b5fb103ab26bfd24d70494c/ui/app/pages/swaps/fee-card/fee-card.js#L178-L181

tmashuang: An example could be 
```
  "secondaryFee": {
    "fee": "$101.98",
    "maxFee": "$112.17"
  },
```

ui/app/pages/swaps/fee-card/fee-card.test.js
tmashuang: An example could be 
```
  "primaryFee": {
    "fee": "0.0441 ETH",
    "maxFee": "0.04851 ETH"
  },
```

tmashuang:
The only other react warnings after the aforementioned ones are 

```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `FeeCard`. See https://fb.me/react-warning-keys for more information.
    in Fragment
    in FeeCard
    in LegacyI18nProvider (created by WithoutStore)
    in I18nProvider (created by WithoutStore)
    in Router (created by MemoryRouter)
    in MemoryRouter (created by WithoutStore)
    in WithoutStore (created by Wrapper)
    in Wrapper
```
and 
```
Warning: Failed prop type: ActionableMessage: prop type `withRightButton` is invalid; it must be a function, usually from the `prop-types` package, but received `undefined`.
      in ActionableMessage
      in ViewQuotePriceDifference
      in Context.Provider
```

Which are issues with the implementation of those components, not necessarily the tests, and can be viewed as an extraneous fix from the scope of this PR.

ui/app/pages/swaps/searchable-item-list/searchable-item-list.test.js
dan437: Thanks! I didn't see `.isRequired` for that specific prop inside `searchable-item-list.js`, but will add the array you mentioned into default props. With future PRs I will be adding testing of different sets of props, not just happy paths.

ui/app/pages/swaps/fee-card/fee-card.test.js
dan437: Nice catch! I see it's there from a previous PR, will update it.

dan437:
Agree, I can solve these pre-existing React warnings as part of this PR.

tmashuang (approved):

brad-decker:
One more thing I'm concerned about with the introduction of a new render helper -- we now have two with differing implementation. I'd like to solve this here by either 1) removing the new one, or 2) moving the newly added code from the old one to the new one in jest/ folder and upgrading all usages to use that one. 


Thoughts?

dan437:
Can I create a user story to migrate all shared Jest functions into `test/jest` folder and update all references to it in our Jest tests?

brad-decker (approved):
LGTM!

PR #10838, opened 2021-04-07
LavaMoat - protect app process background 

EtDu:
🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 
the next step in incrementally locking down metamask:
protecting the **background** process with **lavamoat**
ui to follow in a later PR
🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 

### tasks
- [x] e2e metrics tests passing
- [ ] sourcemaps
- [ ] mozilla lint
  - [ ] bundle splitting
- [ ] minimize diff
  - [ ] most patches should already be on `develop`
  - [ ] maybe move dep upgrades to `develop`
  
blocked on https://github.com/MetaMask/metamask-extension/pull/11080

PR #10902, opened 2021-04-20
Dep Upgrades for Lavamoat Patches

EtDu:
Upgrades:
`fast-json-patch`
`luxon`
+ adds patches to minimize diff size of #10838


kumavis (approved):

ryanml (approved):
++

PR #10905, opened 2021-04-20
rough in switchEthereumChain

brad-decker:
Implements basic swithEthereumChain feature for suggesting a user to switch to a specified chain that **they already have added to MetaMask**. This method will work with default networks like mainnet.

### DIFF
--- a/app/scripts/lib/rpc-method-middleware/handlers/switch-ethereum-chain.js
+++ b/app/scripts/lib/rpc-method-middleware/handlers/switch-ethereum-chain.js
@@ -1,0 +1,92 @@
+import { ethErrors } from 'eth-rpc-errors';
+import { omit } from 'lodash';
+import { MESSAGE_TYPE } from '../../../../../shared/constants/app';
+import {
+  isPrefixedFormattedHexString,
+  isSafeChainId,
+} from '../../../../../shared/modules/network.utils';
+
+const switchEthereumChain = {
+  methodNames: [MESSAGE_TYPE.SWITCH_ETHEREUM_CHAIN],
+  implementation: switchEthereumChainHandler,
+};
+export default switchEthereumChain;
+
+async function switchEthereumChainHandler(
+  req,
+  res,
+  _next,
+  end,
+  { getCurrentChainId, findCustomRpcBy, updateRpcTarget, requestUserApproval },
+) {
+  if (!req.params?.[0] || typeof req.params[0] !== 'object') {
+    return end(
+      ethErrors.rpc.invalidParams({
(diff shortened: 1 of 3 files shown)


app/scripts/lib/rpc-method-middleware/handlers/switch-ethereum-chain.js
```diff
@@ -0,0 +1,92 @@
...
+    }
+    try {
+      await updateRpcTarget(
+        await requestUserApproval({
+          origin,
+          type: MESSAGE_TYPE.SWITCH_ETHEREUM_CHAIN,
+          requestData: {
+            rpcUrl: existingNetwork.rpcUrl,
+            chainId: existingNetwork.chainId,
+            nickname: existingNetwork.nickname,
```
shanejonas, line 80: [EIP-3085](https://eips.ethereum.org/EIPS/eip-3085) has `chainName`

brad-decker: true, but internally we call it nickname. existingNetwork comes from our state. EIP-3085 is implemented in the add-ethereum-chain file that gets the chainName from the dapp.

app/scripts/lib/rpc-method-middleware/handlers/switch-ethereum-chain.js
```diff
@@ -0,0 +1,92 @@
...
+            ticker: existingNetwork.ticker,
```
shanejonas, line 81: EIP-3085 uses:

```
  nativeCurrency?: {
    name: string;
    symbol: string;
    decimals: number;
  };
  ```
  
  not sure if this is the same, or maybe just `symbol` and it has to match

Comment (shanejonas) [reply app/scripts/lib/rpc-method-middleware/handlers/switch-ethereum-chain.js:81]: ah, only `chainId` is from the dapp

File: app/scripts/lib/rpc-method-middleware/handlers/switch-ethereum-chain.js [hunk 3]
```diff
@@ -0,0 +1,92 @@
...
+    );
+  }
+
+  if (!isSafeChainId(parseInt(_chainId, 16))) {
+    return end(
+      ethErrors.rpc.invalidParams({
+        message: `Invalid chain ID "${_chainId}": numerical value greater than max safe value. Received:\n${chainId}`,
+      }),
+    );
+  }
```
rekmarks, line 62: I don't think this check is necessary for the purposes of this method, but it's nice from a "fail early and often" perspective.

rekmarks (approved):
LGTM!

Issue #7452, opened 2019-11-17
Transaction Record Missing


safepayment:
**Describe the bug**
I sent a erc20 transaction on TrustWallet, but it's not shown on MetaMask.
Same private key and address on both TrustWallet and MetaMask.

**To Reproduce**
1. Sent a erc20 tx on Trust Wallet. It's success on etherscan.
2. It's not shown on MetaMask.

**Expected behavior**
The tx should be shown on MetaMask.

**Screenshots**
It's easy to reproduce.

**Browser details (please complete the following information):**
 - OS: Address X
 - Browser chrome
 - MetaMask Version 7.5.3

**Additional context**
No

Gudahtt:
Thanks for the report! This is not a bug, but it is a very important feature that we are lacking. Currently we only show transactions sent from the current MetaMask instance or incoming transactions; we don't show transactions _from_ the current account that were sent elsewhere.

This relates to #505, which is about tracking outgoing transactions from other MetaMask instances. This request seems a bit more broad; to track outgoing transactions from non-MetaMask wallets as well.

PR #10908, opened 2021-04-21
Replace incoming with external transactions

### LINKED ISSUE #505: Load transaction history from other clients
Right now we only log outgoing transactions from the same instance of MetaMask.

Long term we should load all transactions from that account, so even if someone has multiple metamask instances, they both appear "up to date".

Related to #133 but kindof is the inverse of it.

PatrykLucka:
Fixes: #505 #7452

Explanation:  

1.  `IncomingTransactionsController` has been replaced with `ExternalTransactionsController`
2.  `showIncomingTransactions` feature flag has been changed to `showExternalTransactions`
3. `ExternalTransactionsController` now does not filter out outgoing transactions
4.  New transactions type `SENT` has been added, wich is described as "External Send" transaction
5.  `_normalizeTxFromEtherscan` sets the new transaction type `SENT` when address of the sender is the same as the one passed as the parameter

Screenshot 2021-04-20 at 12 51 58

Manual testing steps:  
  - Import account using private key
  - Use the same private key to import it to other instance of MM or any other wallet
  - Send transaction using 2nd wallet
  - Wait for the block request in 1st MM instance
  - Check if "External Send" transaction will show up

### DIFF
--- a/app/scripts/controllers/external-transactions.js
+++ b/app/scripts/controllers/external-transactions.js
@@ -44,11 +44,11 @@ const fetchWithTimeout = getFetchWithTimeout(30000);
  */
 
 /**
- * This controller is responsible for retrieving incoming transactions. Etherscan is polled once every block to check
- * for new incoming transactions for the current selected account on the current network
+ * This controller is responsible for retrieving external transactions. Etherscan is polled once every block to check
+ * for new external transactions for the current selected account on the current network
  *
  * Note that only the built-in Infura networks are supported (i.e. anything in `INFURA_PROVIDER_TYPES`). We will not
- * attempt to retrieve incoming transactions on any custom RPC endpoints.
+ * attempt to retrieve external transactions on any custom RPC endpoints.
  */
 const etherscanSupportedNetworks = [
   GOERLI_CHAIN_ID,
@@ -58,7 +58,7 @@ const etherscanSupportedNetworks = [
   ROPSTEN_CHAIN_ID,
 ];
 
-export default class IncomingTransactionsController {
+export default class ExternalTransactionsController {
   constructor(opts = {}) {
     const {
       blockTracker,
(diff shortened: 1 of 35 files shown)


app/scripts/controllers/external-transactions.js
```diff
@@ -282,7 +281,10 @@ export default class IncomingTransactionsController {
         value: bnToHex(new BN(etherscanTransaction.value)),
       },
       hash: etherscanTransaction.hash,
-      type: TRANSACTION_TYPES.INCOMING,
+      type:
```
brad-decker, line 284: The type parameter here will not be this simple, this could be any currently supported type of transaction not just sent or incoming. In the main transaction controller we use a method that determines what the transaction's type is.. we could potentially reuse some of that logic.

brad-decker (commented):
There's some more nuances here that we have to figure out before this can merge in. First off we currently trim transaction history of the metamask originated transactions to 50 transactions while there is no such cap for this controller. Removing the filter on which transactions it considers could result in new behavior never before seen in the extension. For example, given that the UI groups transactions by nonce, and relies on the history object of our transaction state for displaying the history, throwing in new transactions that do not have that key into the mix could be potentially confusing. Further more transactions sent from outside of MetaMask could potentially confuse the UI. Consider a transaction that is initiated and fails before being submitted, it'll have a nonce associated with it. It becomes the "primary" transaction for informing the UI what occurred. We then bring in a transaction remotely, issued from outside of MetaMask, that shares the same nonce. The status of the transaction will change in the UI but the title and the details of the transaction do not. We have this problem today, adding in outside sources will further complicate the issue. 

I have some inflight work preparing for EIP-1559 and others that will change the way transactions are grouped and I think that is the appropriate time to consider this and incorporate it into that work.

Gudahtt:
Moving this into draft until we complete the aforementioned work

brad-decker:
Hey @PatrykLucka, i'm going to close this PR. We definitely want this work, but as stated before a large amount of effort is needed to modify how we group transactions so that we can effectively handle transactions from clients outside of MetaMask. I have added a comment to #505 which is the issue originally describing this problem.

PR #10583, opened 2021-03-04
Whats new popup

danjm:
Fixes  https://github.com/MetaMask/MetaMask-planning/issues/23

This PR adds the "See what's new" feature to the extension. This PR requires #329 of https://github.com/MetaMask/controllers, and [#375](https://github.com/MetaMask/controllers/pull/375) which aims to update #329 

<details>
<summary>Videos of the behaviour are below:</summary>
<br>

https://user-images.githubusercontent.com/7499938/109920653-2ead1b00-7c95-11eb-8011-d4ed59d0e63f.mp4

https://user-images.githubusercontent.com/7499938/109920654-2f45b180-7c95-11eb-97d4-69af1be3ae25.mp4

https://user-images.githubusercontent.com/7499938/109920656-2fde4800-7c95-11eb-9cd8-d2e1007f54f8.mp4

</details>

PR #10911, opened 2021-04-22
remove the ui/app and ui/lib folders

brad-decker:
### Problem statement
Giving instructions to new engineers (or even those who have been around the block) to go to the `ui/app/components/app` folder is somewhat confusing. This would be fine to just overcome, but it also provides no real value. In fact it disjoints the file structure by implying significance of `ui/libs` over things like `ui/app/helpers` and to further complicate the matter the `ui/libs` folder had test helpers in it (some were unused) 

### Solution
Relocate all files in `ui/lib` to the most sensible destination, for most this was `ui/app/helpers/utils`. Once done, flatten the contents of `ui/app` into `ui/` making `ui/` the top level folder of the entire frontend portion of the extension.

Also, 1316 files later, I'm sorry.


darkwing:
Is there a way we can do this without losing all history 😬😬😬 ?  Also, it might be good to separate the actual file moves into one commit and the config changes in another, so that rebases etc. aren't a horrible experience.

brad-decker:
@darkwing I don't believe preserving history at the file level to be possible in all cases, but we will have the ability to see the file changes and connect the history manually. As an example the first entry for any "new" file in this PR will be the commit in this PR, which if clicked will inform the viewer of the old location. Browsing the repo at that point in history will give the rich file history while it was located in the previous place. 

Not ideal, but unfortunately obscuring file history in this way isn't avoidable in many cases. 

I did split out the configuration changes into a separate commit

darkwing (dismissed):

brad-decker:
Waiting for the [What's new feature](https://github.com/MetaMask/metamask-extension/pull/10583) to land before merging

kumavis:
@brad-decker why do you think these are not being recognized as file renames?

kumavis:
reading suggestions to prefer `git mv` when possible

kumavis:
ohhh

> git commit never detects renames. It just writes content to the repository. Renames (and copies as well) are only detected after the fact, i.e. when running git diff, git merge and friends. That's because git does not store any rename/copy information.

https://stackoverflow.com/questions/14832963/how-can-i-control-the-rename-threshold-when-committing-files-under-git

brad-decker:
@kumavis Thanks for the lesson! I didn't know the entire interface is just being deciphered! I won't complain about github loading slowly on my huge PRs anymore lol.

ryanml (approved):
great work!

PR #10912, opened 2021-04-22
Add contract address validation for token swaps

dan437:
## Explanation
Add contract address validation for token swaps before sending it to blockchain. That is done in 2 ways:
- If it's ETH -> WETH or WETH -> ETH, we validate it against a specific contract address
- For everything else, we check if a contract address matches an expected one for current chainId

## Manual testing steps
Validation is triggered on the page with a quote:
image

It would be great to manually test scenarios that are covered in unit tests.

## Screenshots and videos
Validation passed for TESTETH -> WETH on Testnet:
image

Contract address validation failed:
image


ui/app/ducks/swaps/swaps.js
```diff
@@ -578,6 +587,33 @@ export const fetchQuotesAndSetQuoteState = (
   };
 };
 
+const isContractAddressValid = (
```
brad-decker, line 590: [REQ] Could we add a jsdoc block to this function to describe its purpose briefly and expected parameters?

brad-decker: Also because this doesn't interact with state or dispatch directly I think its better suited to be in the swaps utils file in the page folder. It can be directly exported from there instead of the testtables object

brad-decker (changes requested):

ui/app/ducks/swaps/swaps.js
dan437: Done.

brad-decker (approved):
LGTM, thanks!

PR #10915, opened 2021-04-23
use one segment instance

brad-decker:
Fixes: no longer splitting analytics into two mixpanels to separate old analytics schema from new, work in progress one. Instead, we will segment in mixpanel to keep visualizations separate.


darkwing (approved):
Awesome!  Works for me!

PR #10932, opened 2021-04-26
add view account-details menu item to token-options menu

### LINKED ISSUE #10148: Add "account details" to hamburger menu in asset view
Given I am on the assets page/view
When I click on the hamburger menu
Then I want to see the 'Account detail" link as one of the options
So that I can send funds to that account

NOTE: on ETH assets view add hamburger menu, include "View on Etherscan" and "Account details" links

adonesky1:
(partially) Fixes: #10148

Explanation:  Adds the Account Details menu item/link to the hamburger menu on token pages

Manual testing steps:  
  - Navigate to a token page (other than Eth)
  - Click the hamburger/meatball menu in the top right hand corner
  - Confirm that there is an Account Details menu item available and that when clicked it pops the correct modal.

### DIFF
--- a/ui/app/pages/asset/components/token-asset.js
+++ b/ui/app/pages/asset/components/token-asset.js
@@ -43,6 +43,9 @@ export default function TokenAsset({ token }) {
               );
               global.platform.openTab({ url });
             }}
+            onViewAccountDetails={() => {
+              dispatch(showModal({ name: 'ACCOUNT_DETAILS' }));
+            }}
             tokenSymbol={token.symbol}
           />
         }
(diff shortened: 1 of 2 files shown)


adonesky1:
@danjm Just want to highlight here there are segment events that are raised by the click handlers of the base account level menu items. Not sure if there is an existing issue to port those over to these asset-level menus, but if not I'm happy to write one up?

brad-decker (dismissed):
LGTM. We could probably just invoke `dispatch(showModal` directly from the token-options component but given these two files are already linked tightly it doesn't matter and your contribution matches what is already happening so I think we should land it as is.

brad-decker:
Oops missed the proptypes. Just need to update those and I think this can land

brad-decker (approved):

PR #10933, opened 2021-04-26
upgrade eth-keyring-controller

brad-decker:
Supports newer versions of @ethereumjs/tx in HD and simple keyring types, while maintaining backwards compatibility. 

Required two changes to tests:
1. _privKey and _pubKey are replaced by privateKey and publicKey
2. The error message for messages of incorrect length has been updated (I can't find this change in the dependencies so it must be a transient dependency update)


NiranjanaBinoy (approved):
LGTM

PR #10934, opened 2021-04-26
Increase Jest unit test coverage for the Swaps feature to ~43%

dan437:
## Explanation
This PR increases Swaps coverage from 25% to about 43%, which is getting us one step closer to 90% coverage.

## Manual testing steps
`npm run test:coverage:jest`

## Screenshots
Swaps Jest unit test coverage is getting greener (43.12% for Lines coverage):
image


ui/app/pages/swaps/index.test.js
```diff
@@ -0,0 +1,63 @@
...
+  });
+
+  afterAll(() => {
+    nock.cleanAll();
+  });
+
+  it('renders the component with initial props', async () => {
+    const store = configureMockStore(middleware)(createSwapsMockStore());
+    const { container, getByText } = renderWithProvider(<Swap />, store);
+    await waitFor(() => expect(tokensNock.isDone()).toBe(true));
```
dan437, line 58: This is a great way to wait until a mocked API call is done.

ui/app/__mocks__/react-router-dom.js
```diff
@@ -3,4 +3,9 @@ const originalModule = jest.requireActual('react-router-dom');
 module.exports = {
   ...originalModule,
   useHistory: jest.fn(),
+  useLocation: jest.fn(() => {
```
brad-decker, line 6: presumably when we expand test coverage in our other areas of the repo we'll need to mock this out in those files to override the pathname?

ui/app/pages/swaps/awaiting-swap/view-on-ether-scan-link/view-on-ether-scan-link.test.js
```diff
@@ -0,0 +1,37 @@
...
+  it('renders the component with a custom block explorer link', () => {
+    const { container, getByText } = renderWithProvider(
+      <ViewOnEtherScanLink
+        {...createProps({
+          blockExplorerUrl: 'https://custom-blockchain.explorer',
+          isCustomBlockExplorerUrl: true,
+        })}
+      />,
+    );
+    expect(getByText('View at custom-blockchain.explorer')).toBeInTheDocument();
```
brad-decker, line 34: Is it possible for us to test what the URL will go to here instead of just checking the link text? It'd help to ensure that changes made to the URL but not the text don't pass through our test cases.

ui/app/pages/swaps/loading-swaps-quotes/__snapshots__/loading-swaps-quotes-stories-metadata.test.js.snap
```diff
@@ -0,0 +1,30 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`storiesMetadata matches expected values for storiesMetadata 1`] = `
+Object {
```
brad-decker, line 4: I am assuming this snapshot exists just to confirm that the loading-swaps-quotes-stories-metadata file isn't touched, or if it is that it is intentional?

ui/app/pages/swaps/index.test.js
brad-decker: Nifty!

.eslintrc.js
```diff
@@ -130,6 +130,7 @@ module.exports = {
       rules: {
         'jest/no-restricted-matchers': 'off',
         'import/unambiguous': 'off',
+        'import/named': 'off',
```
brad-decker, line 133: Why did this get turned off?

package.json
```diff
@@ -33,7 +33,7 @@
     "test:e2e:firefox": "SELENIUM_BROWSER=firefox test/e2e/run-all.sh",
     "test:e2e:firefox:metrics": "SELENIUM_BROWSER=firefox mocha test/e2e/metrics.spec.js",
     "test:coverage": "nyc --silent --check-coverage yarn test:unit:strict && nyc --silent --no-clean yarn test:unit:lax && nyc report --reporter=text --reporter=html",
-    "test:coverage:jest": "jest --coverage --maxWorkers=2 --collectCoverageFrom=**/ui/app/**/swaps/**",
+    "test:coverage:jest": "jest --coverage --maxWorkers=2",
```
brad-decker, line 36: why the removal of the collectCoverageFrom flag

ui/app/__mocks__/react-router-dom.js
dan437: That's right. My assumption is that most of the tests will be fine with this path, but if they need to use a different one, they can just do this:
```
jest.mock('react-router-dom', () => {
  const originalModule = jest.requireActual('react-router-dom');
  return {
    ...originalModule,
    useHistory: jest.fn(),
    useLocation: jest.fn(() => {
      return {
        pathname: '/feature-z/page-1',
      };
    }),
  }
});
```

ui/app/pages/swaps/awaiting-swap/view-on-ether-scan-link/view-on-ether-scan-link.test.js
dan437: Great point, will add an assertion for a URL.

ui/app/pages/swaps/loading-swaps-quotes/__snapshots__/loading-swaps-quotes-stories-metadata.test.js.snap
dan437: Exactly. If there would be any change in `loading-swaps-quotes-stories-metadata.js`, a test would fail with an outdated snapshot. It's just a double-check that someone didn't accidentally modify it + it was showing 0% test coverage, which was lowering our overall coverage.

.eslintrc.js
dan437: Great question! That's because I used `module.exports` in the link below: https://github.com/MetaMask/metamask-extension/pull/10934/files#diff-8fd96322fc664541ebc0326e3c5963e55db825809abb977c6ffad8be7555cbd1R7
After which I started getting these errors:
image

I will give it little more time to use `export` instead of `module.exports` when re-exporting imported things. Then we hopefully wouldn't need to turn `import/named` off for tests.

package.json
dan437: I've moved it to `jest.config.js`: https://github.com/MetaMask/metamask-extension/pull/10934/files/581c2fb2962fbb64d0772f57c451c89d3e1dd12c#diff-1e058ca1442e46581b13571fb8d261f9e1f5657e26c96634d4c1072f0f0347f1R4

ui/app/pages/swaps/awaiting-swap/view-on-ether-scan-link/view-on-ether-scan-link.test.js
dan437: I've checked the code and we don't render a link, but open a new tab via an `onClick` event on a `div` element:
```onClick={() => global.platform.openTab({ url: blockExplorerUrl })}```

I will be testing events in following PRs, so I would prefer to do it there.

File: .eslintrc.js (hunk 5 above)
Comment (dan437) [reply .eslintrc.js:133]: So, I've changed it to `export` only:
```
export { createSwapsMockStore } from './mock-store';
export { renderWithProvider } from './rendering';
export { setBackgroundConnection } from './background';
export * as MOCKS from './mocks';
export * as CONSTANTS from './constants';
```

But even while tests are green, I'm getting the same ESLint issue when re-exporting:
image

That's why I would prefer to keep `'import/named': 'off',`, just for tests.

brad-decker: looks legit. https://github.com/benmosher/eslint-plugin-import/issues/1998

brad-decker (approved):

.eslintrc.js
dan437: Nice find!

PR #10938, opened 2021-04-27
add hamburger menu to eth page

### LINKED ISSUE #10148: Add "account details" to hamburger menu in asset view

adonesky1:
(Completes) Fixes: #10148

Explanation:  Adds the hamburger menu that currently exists for non-native asset token pages to the native asset (eth) page fork navigating to account details and Etherscan.

Manual testing steps:  
  - Navigate to native asset (eth) page
  - Confirm that there is a hamburger menu available in top right hand corner
  - Confirm that links in menu work as expected.

### DIFF
--- a/ui/app/pages/asset/components/asset-options.js
+++ b/ui/app/pages/asset/components/asset-options.js
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types';
 import { I18nContext } from '../../../contexts/i18n';
 import { Menu, MenuItem } from '../../../components/ui/menu';
 
-const TokenOptions = ({
+const AssetOptions = ({
   onRemove,
   onViewEtherscan,
   onViewAccountDetails,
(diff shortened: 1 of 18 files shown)


ui/app/pages/asset/components/native-asset.js
```diff
@@ -1,19 +1,30 @@
...
 
 export default function NativeAsset({ nativeCurrency }) {
   const selectedAccountName = useSelector(
     (state) => getSelectedIdentity(state).name,
   );
+  const dispatch = useDispatch();
+
+  const chainId = useSelector(getCurrentChainId);
+  const rpcPrefs = useSelector(getRpcPrefsForCurrentProvider);
+  const selectedIdentity = useSelector(getSelectedIdentity);
```
brad-decker, line 26: There is also `getSelectedAddress` selector which just returns the address. The getSelectedIdentity selector first calls getSelectedAddress then gets the full identity object by address. Might be worthwhile to use `getSelectedAddress` here if we don't expect to use any of the identity fields.

brad-decker (changes requested):
One minor suggestion! Great work

ui/app/pages/asset/components/native-asset.js
```diff
@@ -1,19 +1,30 @@
...
+  getSelectedIdentity,
+  getCurrentChainId,
+  getRpcPrefsForCurrentProvider,
+} from '../../../selectors/selectors';
+import { showModal } from '../../../store/actions';
+import getAccountLink from '../../../../lib/account-link';
 import { DEFAULT_ROUTE } from '../../../helpers/constants/routes';
-
 import AssetNavigation from './asset-navigation';
+import TokenOptions from './token-options';
```
Gudahtt, line 16: This might be a good time to rename this to something other than "token options", since it's now being used for native assets as well.

ui/app/pages/asset/components/native-asset.js
adonesky1: Thanks! Will do!

ui/app/pages/asset/components/native-asset.js
adonesky1: That makes sense to me. I'll make this change!

adonesky1: @Gudahtt so I may have gone beyond what you were thinking with renaming classnames and such. There is a issue with the verify-locales process no longer seeing tokenOptions present in the UI, but I'm not sure how or if the translations should change in the locales files if we do wish to change these classnames to match the new more generic verbiage?

adonesky1: Looks like E2E tests are failing from these classname changes, thought I'm not totally clear why yet

adonesky1: It looks like the E2E errors are the same as the locales issues. @brad-decker @Gudahtt Should I fully revert the name changes because of the translation blocker for now? Or just revert the classnames?

ui/app/pages/asset/components/asset-options.js
```diff
@@ -0,0 +1,80 @@
...
+  const [assetOptionsOpen, setAssetOptionsOpen] = useState(false);
+
+  return (
+    <>
+      <button
+        className="fas fa-ellipsis-v asset-options__button"
+        data-testid="asset-options__button"
+        onClick={() => setAssetOptionsOpen(true)}
+        ref={setAssetOptionsButtonElement}
+        title={t('AssetOptions')}
```
brad-decker, line 27: You changed this translation key which is trying to find a key in app/_locales/en/messages.json  named 'AssetOptions' which doesn't exist. If you do want to change this you'll need to change the key to 'assetOptions' and update 'tokenOptions' to match, as long as that's wanted.

adonesky1: @brad-decker ah yes that makes more sense. I had originally changed the key in the locale files but wasn't sure that was the right thing to do given that the translations won't really match. Not sure what the right move is here, but feel like it could make sense to revert this key you've highlighted but leave the other name changes as is?

brad-decker: In these cases we'd change en/messages.json `tokenOptions: "Token options"` -> `assetOptions: "Asset options"`, then we'd make sure the code is `t('assetOptions')` and finally run yarn lint:fix which will remove the unused translations in the other files. At some point in the future the translation service will add the new 'assetOptions' translations to other files.

brad-decker: The lack of translation here is unfortunate but not the end of the world and aligns more with our other terminology anyways. The tab is the 'assets' tab which includes the tokens that would have previously had the 'Token Options'

brad-decker (approved):
LGTM

PR #10939, opened 2021-04-27
use token custom icons where possible

### LINKED ISSUE #8252: Send token screen does not use token icons from custom-added assets.

adonesky1:
Fixes: #8252

Explanation:  Adds custom icon for a token to the send token flow in the asset-dropdown

Manual testing steps:  
  - Add a custom token via the methods outlined in [EIP-747](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-747.md)
  - Attempt to send some of this token
  - Confirm that the custom icon that was supplied when first importing the token is visible beside token name(rather than generic identicon).
  
  Here is a method that can be called in the browser console to initialize the addition of a token with a custom icon into MetaMask:
  
  `ethereum.request({
  method: 'wallet_watchAsset',
  params: {
    type: 'ERC20',
    options: {
      address: '0xb60e8dd61c5d32be8058bb8eb970870f07233155',
      symbol: 'FOO',
      decimals: 18,
      image: 'https://w7.pngwing.com/pngs/1018/815/png-transparent-luminiferous-aether-symbol-diethyl-ether-symbol-miscellaneous-purple-chemical-element.png',
    },
  },
})
  .then((success) => {
    if (success) {
      console.log('FOO successfully added to wallet!')
    } else {
      throw new Error('Something went wrong.')
    }
  })
  .catch(console.error)`

### DIFF
--- a/ui/app/pages/send/send-content/send-asset-row/send-asset-row.component.js
+++ b/ui/app/pages/send/send-content/send-asset-row/send-asset-row.component.js
@@ -16,6 +16,7 @@ export default class SendAssetRow extends Component {
       }),
     ).isRequired,
     accounts: PropTypes.object.isRequired,
+    assetImages: PropTypes.object,
     selectedAddress: PropTypes.string.isRequired,
     sendTokenAddress: PropTypes.string,
     setSendToken: PropTypes.func.isRequired,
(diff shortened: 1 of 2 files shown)


brad-decker (approved):
LGTM

PR #10954, opened 2021-04-29
fix transaction sync logic

brad-decker:
Fixes: Issue with RC and syncing transactions to mobile. 

cc @tmashuang


tmashuang (approved):
LGTM

PR #10956, opened 2021-04-30
Handling custom token decimal fetch failure due to network error

### LINKED ISSUE #9777: Warn user when custom token addition has defaulted to 0
Currently, when `attemptToAutoFillTokenParams` in `add-token.component.js` is called, if a network issue causes the decimals to default to `0`, the user is not notified. This can lead to tokens being added with an erroneous decimals value of `0`. Notifying the user that no decimal info was retrievable could prevent this. Perhaps even better we should require the user to directly input decimals in such cases.

These changes might have prevented the issue detailed in https://github.com/MetaMask/metamask-extension/issues/9771

### LINKED ISSUE #10534: Intermittent token display rounding error
I have not seen this issue myself, but it is shared here and then several people in the replies claim they’ve seen it before themselves.

I checked for duplicate issues, I can find no former reports with us, did not check in ZenDesk though, there might be some repro state logs in there.

https://twitter.com/missbitcoin_mai/status/1365668356079017984?s=21

### LINKED ISSUE #11081: TypeError: e.trim is not a function
Sentry Issue: [METAMASK-GQTZ](https://sentry.io/organizations/metamask/issues/2212388792/?referrer=github_integration)

```
TypeError: e.trim is not a function
  at handleCustomDecimalsChange (ui/app/pages/add-token/add-token.component.js:216:34)
  at y.attemptToAutoFillTokenParams (ui/app/pages/add-token/add-token.component.js:158:10)
```

NiranjanaBinoy:
Fixes: #9777 #10534 #11081

Explanation:  Currently while fetching the custom token if the decimal value fetch fails, it is defaulting to 0 without warning the user.

Fix:- If decimal value fetch fails the field will become editable and shows warning messages for network failure.

Manual testing steps:

- Create a custom token from test-dapp, and copy the token address.
- From the extension homepage, click the `Add Token` button, in the `Custom Token` tab and enter the token address.
  - The default behavior should be that the symbol and decimal will get populated automatically and the `Next` button will be active.
  - Based on the current changes when the decimal fetch fails, the decimal tab will be empty and editable, a warning message will be shown below the text field with a link to verify the token decimal and the `Next` button will be disabled.

https://user-images.githubusercontent.com/43930900/116836899-5ede4080-ab96-11eb-9e08-095147eae58f.mov


ui/pages/add-token/add-token.component.js
```diff
@@ -211,16 +220,20 @@ class AddToken extends Component {
...
 
   handleCustomDecimalsChange(value) {
-    const customDecimals = value.trim();
-    const validDecimals =
-      customDecimals !== null &&
-      customDecimals !== '' &&
-      customDecimals >= MIN_DECIMAL_VALUE &&
-      customDecimals <= MAX_DECIMAL_VALUE;
-    let customDecimalsError = null;
+    const inValidDecimals = !(
```
ryanml, line 223: Could this be `const inValidDecimals = value < MIN_DECIMAL_VALUE || value > MAX_DECIMAL_VALUE`?

ui/pages/add-token/add-token.component.js
```diff
@@ -211,16 +220,20 @@ class AddToken extends Component {
...
+      value >= MIN_DECIMAL_VALUE && value <= MAX_DECIMAL_VALUE
+    );
 
-    if (!validDecimals) {
-      customDecimalsError = this.context.t('decimalsMustZerotoTen');
+    let customDecimals;
+    let customDecimalsError = null;
+    if (value) {
+      customDecimals = Number(value.trim());
```
ryanml, line 230: nit: could we maybe define this with numerical coercion at the top of the function? Then we can use strict comparison operators in the decimal validation check and fail early if needed if `value` is falsey/not a number

ryanml, line 230: I think we should define `customDecimals` at the beginning of this function so we can fail early if needed and use it in the validation checks

NiranjanaBinoy: I have moved the validity check to inside the if condition, so now the function processing starts by checking for the existence of value. Our lint rules do not allow the negated check so we can't check for `if(!value)`

brad-decker (dismissed):
LGTM!

app/_locales/en/messages.json
```diff
@@ -2158,6 +2158,9 @@
   "tokenContractAddress": {
     "message": "Token Contract Address"
   },
+  "tokenDecimalFetchFailed": {
+    "message": "Token decimals not found."
```
adonesky1, line 2162: @NiranjanaBinoy I think we may need to explain this a bit more clearly to lay users?

adonesky1 (changes requested):
Just a check on warning verbiage. @rachelcope could you take a look and sanity check my comment?

ui/pages/add-token/add-token.component.js
```diff
@@ -236,10 +247,16 @@ class AddToken extends Component {
...
       customSymbolError,
       customDecimalsError,
-      autoFilled,
       forceEditSymbol,
+      symbolAutoFilled,
+      decimalAutoFilled,
     } = this.state;
 
+    const { chainId } = this.props;
+    const blockExplorerTokenLink = createTokenTrackerLinkForChain(
```
adonesky1, line 256: I have an open PR to bump our etherscan-link version and make use of a new method `getTokenTrackerLink`, in which you can pass the RPCPrefs to get custom network block explorer link generation. This will probably land first though so we can change this to use the new method later on I suppose.

adonesky1 (commented):
LGTM, only thing is that it doesn't seem the blockExplorer link will work for tokens on custom networks. You could bump the etherscan-link version number and make use of the new `getTokenTrackerLink` method to get this functionality working.

ui/pages/add-token/add-token.component.js
```diff
@@ -253,10 +253,8 @@ class AddToken extends Component {
     } = this.state;
 
     const { chainId } = this.props;
-    const blockExplorerTokenLink = createTokenTrackerLinkForChain(
-      customAddress,
-      chainId,
-    );
+    const blockExplorerTokenLink = getTokenTrackerLink(customAddress, chainId);
```
adonesky1, line 256: Here in order for the custom block explorers to work you'll need to get RPCPrefs from state and pass them in as the 5th argument here... I made a stupid function signature (probably need to change it). See line 51 in the token-asset.js file [here](https://github.com/MetaMask/metamask-extension/pull/11017/files) for reference.

adonesky1: @NiranjanaBinoy ^

adonesky1 (approved):
LGTM!

PR #10958, opened 2021-04-30
Migrate check for GITHUB_TOKEN

Gudahtt:
The check for the GITHUB_TOKEN environment variable was being done in the wrong release script. It has been migrated to the relevant script.

A second check for the username has also been added, as it is also required.


brad-decker (approved):
LGTM

PR #10680, opened 2021-03-21
[Fix] 10365 My Accounts Removal

BboyAkers:
Fixes: #10365

Explanation: This PR removes
My Accounts string and flow,
`CONTACT_MY_ACCOUNTS_ROUTE`
`CONTACT_MY_ACCOUNTS_VIEW_ROUTE`
`CONTACT_MY_ACCOUNTS_EDIT_ROUTE`

Manual testing steps:  
  - All "My Contact Account" routes should not be usable.

PR #10963, opened 2021-04-30
Ensure method of adding contact when contacts exist

darkwing:
While doing QA on 9.5 today, I noticed an issue that we missed during https://github.com/MetaMask/metamask-extension/pull/10680 review -- when there are existing contacts, there is no "Add Contact" button, since that button got moved to the main content area when contacts don't exist.  This PR aims to rectify that problem.

Add

AddC


brad-decker (approved):
LGTM

PR #10967, opened 2021-04-30
add setting to dismiss seed phrase backup reminder

### LINKED ISSUE #10660: Allow dismissing seed backup reminder
Screen Shot 2021-03-16 at 12 44 25 PM

Our current seed backup reminder is useful for some users, but others want to just be left alone, they know what they're doing (maybe they're using a hardware wallet as their primary account).

We might want to provide a dismiss button, (maybe with some extra friction, really making sure the user knows what they're doing), to allow skipping this for longer periods of time, or MAYBE even indefinitely.

adonesky1:
Fixes: #10660

Explanation:  Provides a setting option under advanced to allow user to dismiss the seed-phrase back up reminder message

Manual testing steps:  
  - Open MetaMask with an account that contains mainnet Eth.
  - While in onboarding flow select "remind me later" when prompted to back up seed-phrase
  - Confirm that this message is present: image
  - Navigate to settings -> advanced and at the bottom toggle on the setting "Dismiss seed-phrase backup reminder"
  - Navigate back to the home page and confirm that the reminder message is no longer there.

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -556,6 +556,12 @@
   "dismiss": {
     "message": "Dismiss"
   },
+  "dismissReminderDescriptionField": {
+    "message": "Turn this on to dismiss the recovery phrase backup reminder message. We highly recommend that you back up your seed phrase to avoid loss of funds"
+  },
+  "dismissReminderField": {
+    "message": "Dismiss recovery phrase backup reminder"
+  },
   "done": {
     "message": "Done"
   },
--- a/app/scripts/controllers/preferences.js
+++ b/app/scripts/controllers/preferences.js
@@ -43,6 +43,7 @@ export default class PreferencesController {
       useBlockie: false,
       useNonceField: false,
       usePhishDetect: true,
+      dismissSeedBackUpReminder: false,
 
       // WARNING: Do not use feature flags for security-sensitive things.
       // Feature flag toggling is available in the global namespace
@@ -669,7 +670,7 @@ export default class PreferencesController {
 
   /**
    * A setter for the `useLedgerLive` property
-   * @param {bool} domain - Value for ledger live support
+   * @param {bool} useLedgerLive - Value for ledger live support
    * @returns {Promise<string>} A promise of the update to useLedgerLive
    */
   async setLedgerLivePreference(useLedgerLive) {
@@ -683,6 +684,17 @@ export default class PreferencesController {
    */
   getLedgerLivePreference() {
     return this.store.getState().useLedgerLive;
+  }
+
+  /**
+   * A setter for the user preference to dismiss the seed phrase backup reminder
+   * @param {bool} dismissBackupReminder- User preference for dismissing the back up reminder
+   * @returns {void}
+   */
+  async setDismissSeedBackUpReminder(dismissSeedBackUpReminder) {
+    await this.store.updateState({
+      dismissSeedBackUpReminder,
+    });
   }
 
   //
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -733,6 +733,10 @@ export default class MetamaskController extends EventEmitter {
         preferencesController.addKnownMethodData,
         preferencesController,
       ),
+      setDismissSeedBackUpReminder: nodeify(
+        this.preferencesController.setDismissSeedBackUpReminder,
+        this.preferencesController,
+      ),
 
       // AddressController
       setAddressBook: nodeify(
--- a/ui/pages/settings/advanced-tab/advanced-tab.component.js
+++ b/ui/pages/settings/advanced-tab/advanced-tab.component.js
@@ -35,6 +35,8 @@ export default class AdvancedTab extends PureComponent {
     ipfsGateway: PropTypes.string.isRequired,
     useLedgerLive: PropTypes.bool.isRequired,
     setLedgerLivePreference: PropTypes.func.isRequired,
+    setDismissSeedBackUpReminder: PropTypes.func.isRequired,
+    dismissSeedBackUpReminder: PropTypes.bool.isRequired,
   };
 
   state = {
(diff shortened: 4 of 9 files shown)


app/_locales/en/messages.json
```diff
@@ -556,6 +556,12 @@
   "dismiss": {
     "message": "Dismiss"
   },
+  "dismissReminderDescriptionField": {
+    "message": "Turn this on to dismiss the seed-phrase backup reminder message. We highly recommend that you back up your seed phrase to avoid loss of funds"
```
adonesky1, line 560: This is placeholder text. @rachelcope do you want to rethink this verbiage?

ui/pages/settings/advanced-tab/advanced-tab.component.js
```diff
@@ -490,6 +492,40 @@ export default class AdvancedTab extends PureComponent {
...
+
+    console.log('dismissSeedBackUpReminder', dismissSeedBackUpReminder);
+
+    return (
+      <div
+        className="settings-page__content-row"
+        data-testid="advanced-setting-dismiss-reminder"
+      >
+        <div className="settings-page__content-item">
+          <span>{this.context.t('dismissReminderField')}</span>
```
ryanml, line 510: nit, this can just be `t('dismissReminderField')` :)

app/scripts/controllers/app-state.js
```diff
@@ -23,6 +23,7 @@ export default class AppStateController extends EventEmitter {
       timeoutMinutes: 0,
       connectedStatusPopoverHasBeenShown: true,
       defaultHomeActiveTabName: null,
+      dismissSeedBackUpReminder: false,
```
brad-decker, line 26: This should probably be in the preferences controller rather than appState. I believe the distinction barrier between these two should be that if the user toggles/sets a setting that it goes in preferences controller. If the state is set just due to some interaction with the app then it goes into the App state.

app/scripts/metamask-controller.js
```diff
@@ -757,6 +757,10 @@ export default class MetamaskController extends EventEmitter {
         this.appStateController.setConnectedStatusPopoverHasBeenShown,
         this.appStateController,
       ),
+      setDismissSeedBackUpReminder: nodeify(
```
brad-decker, line 760: great work adding a function to the API 💯

ui/pages/settings/advanced-tab/advanced-tab.component.js
```diff
@@ -490,6 +492,40 @@ export default class AdvancedTab extends PureComponent {
...
   }
 
+  renderDismissSeedBackupReminderControl() {
+    const { t } = this.context;
+    const {
+      dismissSeedBackUpReminder,
+      setDismissSeedBackUpReminder,
+    } = this.props;
+
+    console.log('dismissSeedBackUpReminder', dismissSeedBackUpReminder);
```
brad-decker, line 502: left over console.log

brad-decker (changes requested):
LGTM, just need to relocate the state but the glue looks good and the frontend code looks good.

app/scripts/controllers/app-state.js
adonesky1: @brad-decker I was wondering the same thing. Unless @danjm feels strongly that this should live in appState (https://github.com/MetaMask/metamask-extension/issues/10660#issuecomment-827835784)?

app/_locales/en/messages.json
rachelcope: This looks good. We are updating "seed phrase" to "recovery phrase" so we might want to update that here as well

app/scripts/controllers/preferences.js
```diff
@@ -669,7 +669,7 @@ export default class PreferencesController {
 
   /**
    * A setter for the `useLedgerLive` property
-   * @param {bool} domain - Value for ledger live support
+   * @param {bool} useLedgerLive - Value for ledger live support
```
adonesky1, line 672: trivial little param name tweak fixing copy paste error

app/scripts/controllers/preferences.js
```diff
@@ -685,6 +686,17 @@ export default class PreferencesController {
     return this.store.getState().useLedgerLive;
   }
 
+  /**
+   * A setter for the user preference to dismiss the seed phrase backup reminder
+   * @param {bool} dismissBackupReminder- User preference for dismissing the back up reminder
+   * @returns {void}
+   */
+   async setDismissSeedBackUpReminder(dismissSeedBackUpReminder) {
```
ryanml, line 694: per the linter this is just one space over-indented: https://app.circleci.com/pipelines/github/MetaMask/metamask-extension/9157/workflows/0ea649f7-a905-47d5-9553-3cd5a39bd86c/jobs/272007

app/scripts/controllers/app-state.js
danjm: @adonesky1 Yes the preferences controller is correct in this case. My suggestion to put a flag in the appstate controller was made when we were thinking that there would be a dismiss button on the notification itself. Since making that comment, we decided to put the toggle in settings, and so I should have added an updated comment to suggest use of the preferences controller.

adonesky1: @brad-decker I've moved this method to the preferences controller

ui/pages/settings/advanced-tab/advanced-tab.component.js
```diff
@@ -490,6 +492,38 @@ export default class AdvancedTab extends PureComponent {
...
 
+  renderDismissSeedBackupReminderControl() {
+    const { t } = this.context;
+    const {
+      dismissSeedBackUpReminder,
+      setDismissSeedBackUpReminder,
+    } = this.props;
+
+    return (
+      <div
```
brad-decker, line 503: just as an aside to any other engineers who review this and go "dang why did Brad not recommend Typography/Box components here" -- these styles and HTML are all pre-existing or copied from other toggles. At some point in the future we should simplify the settings pages but at present this is a-ok 🆗

brad-decker (approved):

PR #10970, opened 2021-05-03
Remove context shims from unit tests

darkwing:
In https://github.com/MetaMask/metamask-extension/pull/10895 we started using our real localization messages, so we should no longer need to shim in our own contexts for localization.


darkwing:
I'll revisit this in the future.

PR #10978, opened 2021-05-03
Change selector of the selected account name to check for visibility.

tmashuang:
After creating a new account, their seems to be a loading screen after confirming new account creation, momentarily/initially setting the account name to `Account {index}`. This behavior is affecting the automated tests that's assuming that the account name is immediately populated with no wait time. The fix is to wait for the element to be visible, which I assume take into account the change in opacity from the loading screen. An alternative would be to set a delay before the account name assertion.


brad-decker (dismissed):
LGTM

PR #10979, opened 2021-05-03
Fixing alignment issue with bottom notification

ryanml:
`align-items: left;` is invalid, and `align-items: center;` on the notification container was producing an undesired result, the notifications will align appropriately to the left without it. 

**Before**
Screen Shot 2021-05-03 at 2 03 19 PM

**After**
Screen Shot 2021-05-03 at 2 04 55 PM


tmashuang (approved):
LGTM

danjm (approved):

PR #10985, opened 2021-05-04
Create MetaMask Beta build

### LINKED ISSUE #10977: Create MetaMask Beta Build
The following should be considered first steps for creating a MMB build:

- [ ] Create a new `yarn` command to signal that the build should be beta
- [ ] Create a basic framework where the BETA label is known everywhere
- [ ] Create another screen in the onboarding flow to let the user know what's different about BETA

darkwing:
Fixes: #10977

Explanation:  

Running `yarn beta` on this WIP PR creates a zip file properly labeled `BETA`, with the manifest containing changes from a separate JSON file.  Additionally, running `yarn beta 3` changes the preminor to `3`, ex: `9.x.x.3`, in the case we have multiple betas.


development/build/utils.js
```diff
@@ -0,0 +1,21 @@
+function getNextBetaVersionMap(currentVersion, platforms) {
```
darkwing, line 1: This function is fairly naive in that it presumes this is `premajor 0`, so a second beta of the same version would have its premajor set to 0 as well; we'll need to be clever here.  We could use a `semver` package but that might just be another dependency.
2.

development/build/utils.js
```diff
@@ -0,0 +1,21 @@
...
+  // eslint-disable-next-line prefer-const
+  let [major, minor, patch, premajor] = currentVersion.split('.');
+  minor = Number(minor) + 1;
+  patch = 0;
+
+  return platforms.reduce((platformMap, platform) => {
+    if (platform === 'firefox') {
+      premajor = 'beta0';
+    } else {
+      premajor = '0';
```
darkwing, line 11: For this we should create a command line param or environment variable where we allow the beta premajor to be passed in, in the rare event we have multiple beta versions.

shanejonas:
+1 for `METAMASK_BUILD_TYPE`, this is awesome @darkwing, great work

ui/pages/home/home.component.js
```diff
@@ -285,6 +287,43 @@ export default class Home extends PureComponent {
     );
   }
 
+  getFooterContent() {
+    const { t } = this.context;
+
+    if (isBeta()) {
```
brad-decker, line 293: You're probably going to hate me but in a few places in this PR we add new class methods for easily switching between two variations of content (beta vs not beta) imo this is extending a pattern that we have agreed to eliminate (class components -> functional components).

Although I'm not proposing that you refactor these components to functional ones, I think adding new methods like this increases the difficulty to refactor them later. In addition this is likely setting the stage for what is a pattern for beta build in the future. 

I'd prefer it if we either:
1) inline the content using ternaries where the content doesn't become obtuse. 
2) create new components where content is incredibly different.

home-footer-beta.js and home-footer.js. I don't know if we support dynamic imports and code splitting yet (@kumavis ?) but we could utilize this pattern in the future to completely extract beta content from production builds and vice versa.  For now, I think its fine to import both normally and render one or the other.

ui/components/app/app-header/index.scss
```diff
@@ -95,6 +95,10 @@
     flex: 1 0 auto;
     width: 0;
     justify-content: flex-end;
+
+    .beta & span {
+      color: #bbc0c5;
```
brad-decker, line 100: please use $ui-3 variable

ui/components/app/app-header/index.scss
```diff
@@ -103,4 +107,8 @@
     background-size: contain;
     background-position: center;
   }
+
+  .beta & {
+    background: #24292e;
```
brad-decker, line 112: $ui-black

ui/pages/first-time-flow/welcome/index.scss
```diff
@@ -27,8 +27,15 @@
...
-    div {
+    p {
       @include Paragraph;
+
+      text-align: start;
+      margin-bottom: 22px;
+    }
+
+    a {
+      color: $Blue-500;
```
brad-decker, line 38: $primary-1 😬 $Blue-500 is deprecated and i'm a slacker who hasn't furthered that plan still

brad-decker (changes requested):
This looks ssiiiiiicccckkk!

all of my change requests are just about setting the right patterns to be extended by future beta-only features (if they become a thing)

darkwing:
I did a major refactor of this branch, making it more extensible and also allowing the extension size to stay smaller.

The only issue is that, for some reason, the mascot is always the standard MM, and not the beta 3D animation.  Will look at it again tomrrow.

brad-decker (dismissed):

danjm:
Verify design requests from support are complete:

- [ ] Update links from zendesk
- [ ] Add new link

cc @darkwing @kevinghim

rekmarks (changes requested):
Looking good! Just a handful of requests.

development/build/static.js
```diff
@@ -13,15 +13,26 @@ module.exports = function createStaticAssetTasks({
...
     shouldIncludeLockdown,
   );
 
+  const copyTargetsBeta = [
+    ...copyTargetsProd,
+    {
+      src: './app/build-types/beta/',
+      dest: `images`,
+    },
+  ];
```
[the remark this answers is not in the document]
darkwing: We don't want to do that, because there are loads of images we still need there; this last object simply overrides a selection of them.

rekmarks (dismissed):
LGTM!

ui/pages/swaps/select-quote-popover/quote-details/quote-details.js
```diff
@@ -80,7 +80,7 @@ const QuoteDetails = ({
       <div className="quote-details__row quote-details__row--high">
         <div className="quote-details__detail-header">
           <img
-            src="./images/logo/metamask-fox.svg"
+            src="/images/logo/metamask-fox.svg"
```
kumavis, line 83: this may break storybook. why was it changed?

kumavis (dismissed):
looks like a good start

ui/pages/swaps/select-quote-popover/quote-details/quote-details.js
darkwing: Don't recall why I made this change -- I've reverted this change.

rekmarks (approved):

kumavis (approved):

PR #10988, opened 2021-05-04
Allowing for scroll overflow when swaps content is too tall

### LINKED ISSUE #10957: Button hidden for custom token and slippage high
To reproduce bug, login to app, add custom token, enable advanced slippage settings on in settings, create quote, and enter manual slippage 6 percent or higher. You see this error "Slippage amount is very high. Make sure you know what you are doing!" and the plugin hides most of the primary submit button as seen in screen shot here.
image

ryanml:
Fixes MetaMask/metamask-extension#10957

Reproduction steps defined in issue


darkwing:
This worked for me!  Did we get the nod on removing the second sentence in the localization though?

darkwing (approved):

PR #10990, opened 2021-05-05
Fix dependency vulnerability by upgrading xmlhttprequest-ssl via yarn.lock

danjm:
This fixes these code vulnerabilities:

Screenshot from 2021-05-05 11-18-20

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "**/redux/symbol-observable": "^2.0.3",
     "**/redux-devtools-instrument/symbol-observable": "^2.0.3",
     "**/rxjs/symbol-observable": "^2.0.3",
+    "**/xmlhttprequest-ssl": "^1.6.2",
     "3box/ipfs/ipld-zcash/zcash-bitcore-lib/lodash": "^4.17.19",
     "3box/ipfs/ipld-zcash/zcash-bitcore-lib/elliptic": "^6.5.4",
     "3box/**/libp2p-crypto/node-forge": "^0.10.0",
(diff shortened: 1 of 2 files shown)


danjm:
Further comments on this fix:

- Currently xmlhttprequest-ssl is at version 1.5.5 in our codebase
- These are the commits between that and the version with the security fix (1.6.2):

Screenshot from 2021-05-04 23-13-44

Of those, only 4 appear to be potentially functional:
https://github.com/mjwwit/node-XMLHttpRequest/pull/6
https://github.com/mjwwit/node-XMLHttpRequest/commit/bf53329b61ca6afc5d28f6b8d2dc2e3ca740a9b2
https://github.com/mjwwit/node-XMLHttpRequest/commit/6a0a91d832419430adf13d64c4b70690228634d7
https://github.com/mjwwit/node-XMLHttpRequest/commit/ee1e81fc67729c7c0eba5537ed7fe1e30a6b3291

I think all of these will be non-breaking for engine-io.client:

- https://github.com/mjwwit/node-XMLHttpRequest/commit/ee1e81fc67729c7c0eba5537ed7fe1e30a6b3291 is the security fix, which should be non-functional for any use case that matters
- https://github.com/mjwwit/node-XMLHttpRequest/commit/6a0a91d832419430adf13d64c4b70690228634d7 will be non-functional for uses cases possibly relevant to use
- https://github.com/mjwwit/node-XMLHttpRequest/pull/6 would only be breaking if the user of the library was (as explained in a comment by the maintainer) "relying on the specific error type or 503 status code being returned", but engine-io.client is not
- https://github.com/mjwwit/node-XMLHttpRequest/commit/bf53329b61ca6afc5d28f6b8d2dc2e3ca740a9b2 won't effect engine-io.client as that library is just passing options from its caller on through to XMLHttpRequest, and in any effect, it just defaults to stronger

Even if my read of all of those is incorrect, I am 99% sure that 3box's use of ipfs does not rely on libp2p-webrtc-star. 3box creates an ipfs instance, which in turn instantiates libp2p-webrtc-star. However, 3box only uses the following properties/methods on these ipfs instances:
ipfs.swarm
ipfs.id
ipfs.dag
ipfs.stop
ipfs.pubsub
ipfs.on
ipfs.add

And none of these rely on libp2p-webrtc-star

It also seems safe for eth-trezor-keyring, which only relies on the `build-tx` module of hd-wallet, which does not rely on socket.io-client (but that gets imported because the whole of hd-wallet is imported)

Gudahtt (approved):
LGTM!

darkwing (approved):

PR #10993, opened 2021-05-05
Replace `auto-changelog` script

Gudahtt:
The `auto-changelog` script has been replaced with the package `@metamask/auto-changelog`. This package includes a script that has an `update` command that is roughly equivalent to the old `auto-changelog.js` script, except better. The script also has a `validate` command.

The `repository` field was added to `package.json` because it's utilized by the `auto-changelog` script, and this was easier than specifying the repository URL with a CLI argument.


package.json
```diff
@@ -2,6 +2,10 @@
   "name": "metamask-crx",
   "version": "0.0.0",
   "private": true,
+  "repository": {
+    "type" : "git",
+    "url" : "https://github.com/MetaMask/metamask-extension"
```
ryanml, line 7: linter note: spaces before: `:`

ryanml:
Besides the lint issue looks good to me!

package.json
Gudahtt: Oops! Fixed.

brad-decker (approved):

PR #11000, opened 2021-05-06
Update `@metamask/controllers` to v8

Gudahtt:
Update `@metamask/controllers` from v6 to v8. The breaking changes were all in controllers that aren't used by the extension, so no changes were required.


ryanml (approved):
++

PR #11003, opened 2021-05-06
Remove `SET_CURRENT_FIAT` action

Gudahtt:
The `SET_CURRENT_FIAT` action has been removed. It has been replaced by a call to `forceUpdateMetamaskState`. The only purpose of this action was to eagerly update the current fiat currency settings before the next state update. Forcing a state update instead is simpler and safer.

The `setCurrentCurrency` function in the background has been updated to no longer return the state, now that it's no longer needed.


brad-decker (approved):
LGTM

PR #11005, opened 2021-05-06
Migrate to new CurrencyRateController

Gudahtt:
The CurrencyRateController has been migrated to the BaseControllerV2 API, which includes various API changes. These changes include:
* The constructor now expects to be passed a `RestrictedControllerMessenger`.
* State changes are subscribed to via the `ControllerMessenger` now, rather than via a `subscribe` function.
* The state and configration are passed in as one "options" object, rather than as two separate parameters
* The polling needs to be started explicitly by calling `start`. It can be stopped and started on-demand now as well.
* Changing the current currency or native currency will now throw an error if we fail to update the conversion rate.

Manual testing instructions:
* Check that fiat is shown correctly when the extension starts
* Check that fiat values are shown correctly after the currency is changed in settings.
* Check that fiat values are shown correctly after switching networks

### DIFF
--- a/app/scripts/lib/ComposableObservableStore.js
+++ b/app/scripts/lib/ComposableObservableStore.js
@@ -1,18 +1,36 @@
 import { ObservableStore } from '@metamask/obs-store';
 
 /**
+ * @typedef {import('@metamask/controllers').ControllerMessenger} ControllerMessenger
+ */
+
+/**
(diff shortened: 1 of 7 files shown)


Gudahtt:
Depends upon #11009

shanejonas (approved):

PR #11007, opened 2021-05-07
Hide basic tab in advanced gas modal when on testnets

### LINKED ISSUE #10980: Hide basic tab in advanced gas modal when on testnets
On testnets, we don't use the gas prices api to get slow, medium and fast estimates. Instead we just get a single estimate via `eth_estimateGas`. In such cases, when the gas edit modal is opened, we should not show the basic tab in the advanced gas modal.

Screenshot from 2021-05-04 10-53-26

NiranjanaBinoy:
Fixes: #10980 

The only location the custom gas modal is accessed in test-net is in `confirm-approve`. This PR checks if the network is not Mainnet or test environment (for e2e test).


ryanml (approved):
lgtm!

PR #11009, opened 2021-05-07
Replace uses of `hasOwnProperty.call` with simpler alternatives

Gudahtt:
Three uses of `hasOwnProperty.call` have been replaced with simpler alternatives. I did this because I found this convention confusing.

The `ComposableObservableStore` changes required adding a check for whether the `config` is defined, because the old syntax would allow iterating over `undefined` as if it was an empty object. Functionally it should be identical. A new test was added to highlight this functionality.


brad-decker (approved):
LGTM

PR #11012, opened 2021-05-07
Use strict assertion mode everywhere

Gudahtt:
The `assert` module has two modes: "Legacy" and "strict". When using strict mode, the "strict" version of each assertion method is implied. Whereas in legacy mode, by default it will use the deprecated, "loose" version of each assertion.

We now use strict mode everywhere. A few tests required updates where they were asserting the wrong thing, and it was passing beforehand due to the loose matching.


rekmarks (approved):
LGTM!

PR #11013, opened 2021-05-07
Prevent signature request component from canceling hardware wallet signing

darkwing:
Explanation:  

While working on Ledger EIP-712 signing, I was noticing that the Ledger was being asked to confirm signature but MetaMask was immediately showing a cancel/rejection.  This appears to be happening because we're manually calling `cancel` from the UI based on a `beforeunload` event.  In the case that we know the user has pressed the cancel or submit button, we should remove that onUnload event.


Gudahtt (approved):
LGTM!

PR #11015, opened 2021-05-07
Prevent an undefined gas price from breaking the transaction list

darkwing:
An undefined gas price causes errors when signing, so setting that value to `0x00` is appropriate.


brad-decker (approved):
LGTM

PR #11016, opened 2021-05-07
Group signature transactions

darkwing:
We need to group the explicitly known signing types with `null` and undefined so they are treated properly by the transaction list.

### DIFF
--- a/shared/constants/transaction.js
+++ b/shared/constants/transaction.js
@@ -1,3 +1,5 @@
+import { MESSAGE_TYPE } from './app';
+
 /**
  * Transaction Type is a MetaMask construct used internally
  * @typedef {Object} TransactionTypes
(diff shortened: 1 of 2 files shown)


shared/constants/transaction.js
```diff
@@ -51,6 +53,9 @@ export const TRANSACTION_TYPES = {
   DEPLOY_CONTRACT: 'contractDeployment',
   SWAP: 'swap',
   SWAP_APPROVAL: 'swapApproval',
+  SIGN: MESSAGE_TYPE.ETH_SIGN,
+  SIGN_TYPED_DATA: MESSAGE_TYPE.ETH_SIGN_TYPED_DATA,
+  PERSONAL_SIGN: MESSAGE_TYPE.PERSONAL_SIGN,
```
brad-decker, line 58: Should this list include 'DECRYPT' or does that not trigger this flow?

darkwing: Good shout!  Updated!

ghost (approved):
Dope

ghost (approved):

ghost (approved):

brad-decker (approved):

PR #11021, opened 2021-05-10
Adding periodic reminder modal for backing up recovery phrase

ryanml:
The goal is to display this reminder two days after wallet creation/import (or in the case of an update, two days from that point). After the user dismisses the reminder, it will subsequently appear every 90 days.

## Manual Test Plan
1. In `app/scripts/controllers/app-state.js`, set `REMINDER_CHECK_INTERVAL` to equal `time.SECOND * 30` (30 seconds), `INITIAL_REMINDER_FREQUENCY` to equal `time.MINUTE * 2` (2 minutes), and `FOLLOWUP_REMINDER_FREQUENCY` to equal `time.MINUTE * 4` (4 minutes)
2. Run a dev build from this branch
3. Create/Import a wallet
4. View Notifications and dismiss "what's new popup" (this notification should not appear when the "what's new popup" is open)
5. Confirm after 2 minutes that the reminder appears.
6. Dismiss the reminder
7. Confirm that after 4 minutes, the reminder appears agin.
8. Dismiss the reminder, re-confirm step 7

### (User has not backed up their seed phrase)
Screen Shot 2021-05-09 at 8 02 20 PM (1)

### (User has backed up their seed phrase)
Screen Shot 2021-05-09 at 8 04 13 PM (1)

### DIFF
--- a/app/scripts/controllers/app-state.js
+++ b/app/scripts/controllers/app-state.js
@@ -1,6 +1,14 @@
 import EventEmitter from 'events';
 import { ObservableStore } from '@metamask/obs-store';
 import { METAMASK_CONTROLLER_EVENTS } from '../metamask-controller';
+import * as time from '../../../shared/constants/time';
+
+// 1 hour
+const REMINDER_CHECK_INTERVAL = time.HOUR;
+// 2 days
+const INITIAL_REMINDER_FREQUENCY = time.DAY * 2;
+// 90 days
+const FOLLOWUP_REMINDER_FREQUENCY = time.DAY * 90;
 
 export default class AppStateController extends EventEmitter {
   /**
@@ -24,9 +32,13 @@ export default class AppStateController extends EventEmitter {
       connectedStatusPopoverHasBeenShown: true,
       defaultHomeActiveTabName: null,
       browserEnvironment: {},
+      recoveryPhraseReminderHasBeenShown: false,
+      recoveryPhraseReminderLastShown: 0,
+      shouldShowRecoveryPhraseReminder: false,
       ...initState,
     });
     this.timer = null;
+    this.interval = REMINDER_CHECK_INTERVAL;
 
     this.isUnlocked = isUnlocked;
     this.waitingForUnlock = [];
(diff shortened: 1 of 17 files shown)


ui/components/ui/popover/popover.component.js
```diff
@@ -17,6 +17,7 @@ const Popover = ({
   showArrow,
```
ryanml, line 17: Per the designs, there is no 'X' button (closure happens upon clicking "Got it!" at the bottom of the popup), and the title is centered.

shared/constants/time.js
```diff
@@ -0,0 +1,5 @@
+export const MILLISECOND = 1;
```
ryanml, line 1: After merge I'll put up a change to use these across the codebase where applicable

app/scripts/controllers/app-state.js
```diff
@@ -134,6 +169,69 @@ export default class AppStateController extends EventEmitter {
...
+    // For the first interval run, set recoveryPhraseReminderLastShown to the current time.
+    if (recoveryPhraseReminderLastShown === 0) {
+      this.store.updateState({
+        recoveryPhraseReminderLastShown: new Date().getTime(),
+      });
+      return;
+    }
+
+    // Capture the current timestamp
+    const currentTime = new Date().getTime();
```
darkwing, line 187: Nit: we could move this var declaration above `if (recoveryPhraseReminderLastShown === 0) {` and then use it on line 181.

ryanml: good catch, i can update this

ryanml: Updated

danjm:
There might be a way that this can be simplified so that fewer background code additions are needed and no migration is required.

What I'm thinking is:
- `recoveryPhraseReminderLastShown` can be exposed via state to the front end
- Instead of `shouldShowRecoveryPhraseReminder` as a property, it could be a redux selector. This selector could comparse the current time to the `recoveryPhraseReminderLastShown` time. If `recoveryPhraseReminderLastShown` is not yet defined, assume that the user has never been shown the reminder, and return true if `(currentTime - 0) > 2 days`. If `recoveryPhraseReminderLastShown` is defined, assume the user has seen the reminder before, and return true if `(currentTime - 0) > 30 days`
- then, instead of calling `setRecoveryPhraseReminderHasBeenShown` from the front end when the reminder is closed, just call `setRecoveryPhraseReminderLastShown` with the current date/time

With this, I think we can greatly reduce the amount of background code added with this PR, and remove the need for a migration.

Let me know what you think.

ryanml:
@danjm - I have refactored this PR per our discussion. cc: @darkwing

danjm (approved):
My major suggestion involved the refactor to simplify controller logic. Those changes look good and this PR looks good overall to me. Nice work!

PR #11029, opened 2021-05-10
Migrate version from `_base` manifest to `package.json`

Gudahtt:
The version field is now stored in the main `package.json` file rather than in the base manifest. It is built into the final manifest during the build script.

This makes it easier to communicate what the current version should be to our `auto-changelog` script. It's also generally a more conventional place to keep track of the version, even considering that we're not publishing to npm.


ryanml (approved):
lgtm!

PR #11030, opened 2021-05-10
Update switchEthereumChain unrecognized chain error

rekmarks:
Return a specific error (code `4902`) for `switchEthereumChain` requests for chains that aren't already in the user's wallet.


brad-decker (approved):
LGTM

PR #11031, opened 2021-05-10
Fix error behavior of addEthereumChain

rekmarks:
Previously, we were returning a user rejection error for `wallet_addEthereumChain` requests if the user declined to switch the network, even if the network was successfully added. This violates the specification of this RPC method. This PR ensures that we do not return errors if the user declines to switch the network after the network is successfully added (or if it was already added).


brad-decker (approved):
LGTM

PR #11043, opened 2021-05-11
updaing copy to "Send ether"

### LINKED ISSUE #10974: Copy update:
past tense "sent ether" on the transaction type at confirmation screen. Update to present tense "Send Ether"
 
 
Screenshot from 2021-05-03 12-52-00

jamiehewitt15:
Closes: https://github.com/MetaMask/metamask-extension/issues/10974

Explanation:  
- Changing the past tense "sent ether" on the transaction type at confirmation screen. Updating it to present tense "Send Ether"

### DIFF
--- a/app/_locales/am/messages.json
+++ b/app/_locales/am/messages.json
@@ -940,7 +940,7 @@
   "sendTokens": {
     "message": "ተለዋጭ ስሞችን ላክ"
   },
-  "sentEther": {
+  "ethSendAction": {
     "message": "የተላከ ether"
   },
   "separateEachWord": {
--- a/app/_locales/ar/messages.json
+++ b/app/_locales/ar/messages.json
@@ -936,7 +936,7 @@
   "sendTokens": {
     "message": "إرسال عملات رمزية"
   },
-  "sentEther": {
+  "ethSendAction": {
     "message": "أرسل عملة إيثير"
   },
   "separateEachWord": {
--- a/app/_locales/bg/messages.json
+++ b/app/_locales/bg/messages.json
@@ -939,7 +939,7 @@
   "sendTokens": {
     "message": "Изпращане на жетони"
   },
-  "sentEther": {
+  "ethSendAction": {
     "message": "изпратен етер"
   },
   "separateEachWord": {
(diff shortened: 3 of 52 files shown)


jamiehewitt15:

darkwing:
Thank you for your contribution, @jamiehewitt15, and I apologize it took us so long to have a look!  The change looks good but when we change the value of a localization we need to remove that localization key from all of the other localization files.  Could you remove usages of `sentEther` elsewhere?

darkwing:
Also, I think we should consider updating that key name; maybe `ethSendAction`?

brad-decker:
> Thank you for your contribution, @jamiehewitt15, and I apologize it took us so long to have a look! The change looks good but when we change the value of a localization we need to remove that localization key from all of the other localization files. Could you remove usages of `sentEther` elsewhere?

This will result in fundamental changes to how transactions are created at the controller level. Not that we shouldn't do it but 'ethSendAction' as a transaction type might be a bit confusing. Let me look into the implications.

brad-decker:
I think we should likely divide the relationship between transaction type and translation keys, but that will require some additional effort and might be best to be done in a PR ahead of this one @jamiehewitt15. A good example of what we could do is here:

```js
  const getTranslatedNetworkName = () => {
    switch (chainId) {
      case MAINNET_CHAIN_ID:
        return t('networkNameEthereum');
      case BSC_CHAIN_ID:
        return t('networkNameBSC');
      case LOCALHOST_CHAIN_ID:
        return t('networkNameTestnet');
      default:
        throw new Error('This network is not supported for token swaps');
    }
  };
```

This design follows a few of our current best practices which is
1. don't use translationKeys as application logic
2. don't use variables to store translationKeys, call them directly in the `t` method invocation. 

https://github.com/MetaMask/metamask-extension/blob/09d81ac5f2c94d7730d2e981b4f05734858030bf/ui/pages/swaps/fee-card/fee-card.js

brad-decker:
(sorry just reading the linked issue)
One other thing to consider is the tense is correct in some cases, for example on the transaction activity list. Where as in confirmations it should be present tense because it hasn't yet happened.

PR #10695, opened 2021-03-22
Excluding sourcemaps comment in production builds

ryanml:
Fixes MetaMask/metamask-extension#7077

Manual testing steps:  
  - Build production MetaMask with `yarn dist`
  - Load the extension in to the browser
  - Navigate to a site such as `https://www.cryptokitties.co/`
  - Open Dev Tools > Console
  - Ensure sourcemap console warnings are not present
  
**Before:**
Screen Shot 2021-03-22 at 2 34 31 PM

**After:**
Screen Shot 2021-03-22 at 2 38 46 PM

PR #11051, opened 2021-05-11
Fix Sentry sourcemaps

Gudahtt:
The method used for uploading release artifacts to Sentry has been updated to allow `sentry-cli` to associate our minified bundles with the corresponding source map file. This should help Sentry display rich stack traces.

Previously Sentry had used the `sourceMappingURL` to associate source maps with bundles, but we recently removed this in #10695. The hope is that this change to the upload process will ensure the mapping works correctly without the `sourceMappingURL` comment.

The `upload_bundles` function was removed because the later `upload_sourcemaps` function actually uploaded both the bundles and source maps.

The `--rewrite` flag was added to enable a newer ["rewrite" feature of the Sentry CLI that they recommend using](https://docs.sentry.io/product/cli/releases/#sentry-cli-sourcemaps). This rewrite is where they associate source maps with bundles.

The `url-prefix` has been updated to be `metamask` rather than `sourcemaps`. I don't think `sourcemaps` was ever the correct prefix. We normalize our errors to have the path `metamask/` before sending any reports to Sentry.

### DIFF
--- a/development/sentry-upload-artifacts.sh
+++ b/development/sentry-upload-artifacts.sh
@@ -27,33 +27,10 @@ Options:
 EOF
 }
 
-function upload_bundles {
-  local release="${1}"; shift
-
-  for filepath in ./dist/chrome/*.js
-  do
-    if [[ -f $filepath ]]
-    then
-      upload_bundle "${release}" "${filepath}"
-    fi
-  done
-}
-
-function upload_bundle {
-  local release="${1}"; shift
-  local filepath="${1}"; shift
-  local filename
-
-  filename="$( basename "${filepath}" )"
-
-  printf 'Uploading %s\n' "${filename}"
-  sentry-cli releases --org 'metamask' --project 'metamask' files "${release}" upload "${filepath}" "metamask/${filename}"
-}
-
 function upload_sourcemaps {
   local release="${1}"; shift
 
-  sentry-cli releases --org 'metamask' --project 'metamask' files "${release}" upload-sourcemaps ./dist/sourcemaps/ --url-prefix 'sourcemaps'
+  sentry-cli releases --org 'metamask' --project 'metamask' files "${release}" upload-sourcemaps ./dist/chrome/ ./dist/sourcemaps/ --rewrite --url-prefix 'metamask'
 }
 
 function main {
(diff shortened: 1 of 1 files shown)


Gudahtt:
I think the best way to test this would be to create a test release, then run `yarn sentry:publish` in this test release environment. It's a bit tricky to do in practice though - we'll have to be careful to label the test version correctly, avoid overwriting real production Sentry releases, and we'll need various credentials that are typically only in CI. We can coordinate this test via Slack.

Gudahtt:
Here are the results after the latest commit:

```
+---------------------------------+--------------+------------------------+----------+                                    
| Name                            | Distribution | Source Map             | Size     |                                    
+---------------------------------+--------------+------------------------+----------+                                    
| metamask/background.js          |              | background.js.map      | 3.90MB   |                                    
| metamask/background.js.map      |              |                        | 13.19MB  |                                    
| metamask/bg-libs.js             |              | bg-libs.js.map         | 2.39MB   |                                    
| metamask/bg-libs.js.map         |              |                        | 8.47MB   |                                    
| metamask/chromereload.js        |              |                        | 42B      |                                    
| metamask/contentscript.js       |              | contentscript.js.map   | 255.10KB |                                    | metamask/contentscript.js.map   |              |                        | 463.82KB |
| metamask/disable-console.js     |              | disable-console.js.map | 2.26KB   |
| metamask/disable-console.js.map |              |                        | 9.81KB   |
| metamask/globalthis.js          |              |                        | 8.81KB   |
| metamask/initSentry.js          |              | initSentry.js.map      | 144.57KB |
| metamask/initSentry.js.map      |              |                        | 607.11KB |
| metamask/inpage.js              |              | inpage.js.map          | 137.14KB |
| metamask/inpage.js.map          |              |                        | 549.55KB |      
| metamask/lockdown.js            |              |                        | 167.76KB |
| metamask/phishing-detect.js     |              | phishing-detect.js.map | 409.80KB |
| metamask/phishing-detect.js.map |              |                        | 1.81MB   |
| metamask/runLockdown.js         |              |                        | 684B     |
| metamask/ui.js                  |              | ui.js.map              | 3.82MB   |
| metamask/ui.js.map              |              |                        | 11.55MB  |
| metamask/ui-libs.js             |              | ui-libs.js.map         | 1.08MB   |
| metamask/ui-libs.js.map         |              |                        | 4.85MB   |                                    +---------------------------------+--------------+------------------------+----------+

```

ryanml (approved):
this lgtm!

PR #11063, opened 2021-05-12
Swaps: Add anonymized tracking for HW wallet and HW wallet type

dan437:
## Explanation
We want to see how many users use a HW wallet and which type during swaps, so we can make better data-driven decisions about upcoming features / priorities. Tracking of these 2 params is anonymized.

## Manual testing steps
- Click on "Swap" on the main page
- Select "Token From" and "Token To"
- When submitted, we log the first event called "Quotes Requested"
- After receiving quotes, we track another event called "Quotes Received"

### DIFF
--- a/test/data/mock-state.json
+++ b/test/data/mock-state.json
@@ -15,6 +15,12 @@
     "provider": {
       "chainId": "0x4"
     },
+    "keyrings": [
+      {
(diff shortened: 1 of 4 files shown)


darkwing (approved):
Awesome!

brad-decker (approved):

PR #11065, opened 2021-05-12
return early from metaRPCHandler if stream has ended

shanejonas:
still cant reproduce locally. but i believe this could solve `write after end`.

### DIFF
--- a/app/scripts/lib/createMetaRPCHandler.js
+++ b/app/scripts/lib/createMetaRPCHandler.js
@@ -2,6 +2,9 @@ import { ethErrors, serializeError } from 'eth-rpc-errors';
 
 const createMetaRPCHandler = (api, outStream) => {
   return (data) => {
+    if (outStream._writableState.ended) {
+      return;
+    }
     if (!api[data.method]) {
       outStream.write({
         jsonrpc: '2.0',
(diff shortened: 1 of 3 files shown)


Gudahtt (dismissed):
LGTM! I suspect that this will address https://sentry.io/organizations/metamask/issues/2394407935/.

It would be great to have this covered by a unit test too.

Gudahtt:
I think there is still one outstanding `write after error` case that will require a similar change to be made in `metamask-controller`: https://sentry.io/organizations/metamask/issues/2394666692/

rekmarks (commented):
Nice!!

rekmarks (approved):
LGTM!

I'd recommend that we use [`duplex.writableEnded`](https://nodejs.org/api/stream.html#stream_writable_writableended), but the stream implementation we use emulates a Node version well before that property was introduced. (Specifically, our version of `readable-stream` emulates Node 8, while the `writableEnded` property was introduced in Node 12.)

rekmarks (changes requested):
Although, per @Gudahtt's suggestion, should we just include an early return here as well? https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/metamask-controller.js/#L2049-L2057

rekmarks (approved):
LGTM!

PR #11066, opened 2021-05-12
Fix conversion rate issues

### LINKED ISSUE #9492: Custom network incorrect account balance when symbol is left blank.
**Describe the bug**
When a custom network symbol that we optionally take is left blank we assume it is an ETH network. So then we apply the ETH conversion rate to getBalance. This is easily result in users thinking they have more funds they they actually do when viewing their balance. The max amount in the send field will show the correct amount. 

**To Reproduce (REQUIRED)**

1. Add the xDai network, steps are [xDai MetaMask setup](https://www.xdaichain.com/for-users/wallets/metamask/metamask-setup)
2. Omit the asset symbol which we say is optional
3. The current conversion rate of ETH should times the balance we get from the network and apply the ETH suffix to the value.

**Expected behavior**
Either require the network symbol field to prevent this behavior, or when on a network other than the defaults do not apply the conversionRate to it.

**Screenshots**
image

adonesky1:
Fixes: #9492

Explanation:  In cases where a user has not entered a symbol ticker for a custom RPC network (or our conversion rate API doesn't recognize the asset), don't currently update our conversionRate in state and use the conversion rate from the previously used network with a valild conversion rate. Moreover, in useCurrencyDisplay (where we apply the conversion for most UI elements where the value of a secondary currency is required) we haven't had a clear conditional separation of values that require conversion (Secondary currencies / Fiat) and those that do not (Primary Currency / native assets).

 This PR, in tandem with this [one](https://github.com/MetaMask/controllers/pull/465) allows conversionRates to be null when we cannot calculate them. In such cases we do not show a converted secondary currency (even if the user has selected to show conversions in testnets). This prevents us from showing inaccurate conversions. It also adds logic to avoid applying conversion rates to primary currencies which results in some of the more alarming bugs (show in video in linked PR above).

tldr: when reviewing start at useCurrencyDisplay.js
 
 Additional work here:
- Removes user-preferenced-token-input & user-preferenced-currency-input components because they are just wrapping token-input and currency-input and passing a single additional prop (useNativeCurrencyAsPrimaryCurrency) that can be added by the component's containers. IMO this pattern makes it harder to tell where props are coming from and how to reason about what the components are trying to achieves. Furthermore there is currently no where that the token-input or currency-input is used where it wasn't wrapped by user-preference-token-input /user-preference-currency-input so the pattern isn't creating any meaningful distinction between the wrapped and unwrapped versions.
- add functionality such that when a user has selected fiat as their primary currency token asset list items also show fiat as primary.

Work to follow:
- Proposal to replace useCurrencyDisplay with a hook that is more agnostic and versatile to consolidate the many patterns we have for calculating and displaying `currentCurrency` and `nativeCurrency` together and clean up the impossible to reason about logic that lives in components like currency-input.component and token-input.component.
- A PR to change the names of `currentCurrency` and `nativeCurrency` to `preferredCurrency` and `networkCurrency` respectively per this [thread](https://consensys.slack.com/archives/G1L7H42BT/p1620939857402600). If you have thoughts on this please include comments here?
 
Manual testing steps:  
  - In advanced settings, turn `Show Conversion on Testnets` on.
  - Go between mainnet, a custom rpc network where you have added a ticker symbol and a custom rpc network where you haven't added a ticker symbol.
  - Check that conversions are correct for all assets across these different networks (for the network with no ticker symbol there should be no conversions).


darkwing:
Added DONOTMERGE based on Alex's PR description, which could be lost in the rest of the text

ui/components/app/asset-list-item/asset-list-item.js
```diff
@@ -98,11 +98,10 @@ const AssetListItem = ({
         <button
           className="asset-list-item__token-button"
           onClick={onClick}
-          title={`${primary} ${tokenSymbol}`}
+          title={primary}
         >
           <h2>
             <span className="asset-list-item__token-value">{primary}</span>
-            <span className="asset-list-item__token-symbol">{tokenSymbol}</span>
```
Gudahtt, line 105: In some other places, we intentionally put the symbol and the value in separate elements so that we can use CSS to ellipse long values without obscuring the symbol. That seems like it would be useful here too :thinking: But that said I can't see any evidence of such a strategy being used here.

adonesky1: Added back!

ui/components/app/asset-list-item/asset-list-item.js
```diff
@@ -71,17 +73,21 @@ const AssetListItem = ({
             updateSendToken({
               address: tokenAddress,
               decimals: tokenDecimals,
-              symbol: tokenSymbol,
+              symbol: primaryIsFiat ? secondarySymbol : primarySymbol,
```
ryanml, line 76: Nit, since we are using `primaryIsFiat ? secondarySymbol : primarySymbol` twice, maybe we can cache it in a var called `symbol`?

adonesky1: added :)

ui/components/app/wallet-overview/eth-overview.js
```diff
@@ -72,6 +76,8 @@ const EthOverview = ({ className }) => {
   });
   const defaultSwapsToken = useSelector(getSwapsDefaultToken);
 
+  const swapPrimary = !conversionRate && !useNativeCurrencyAsPrimaryCurrency;
```
brad-decker, line 79: [req] i assumed this was related to swaps, lol. maybe `treatSecondaryCurrencyAsPrimary` ? 😬 naming things is hard

ui/components/ui/currency-display/currency-display.component.js
```diff
@@ -27,25 +28,30 @@ export default function CurrencyDisplay({
...
-        {parts.prefix}
-        {parts.value}
-      </span>
-      {parts.suffix && (
-        <span className="currency-display-component__suffix">
-          {parts.suffix}
-        </span>
-      )}
-    </div>
+    <>
```
brad-decker, line 34: is the fragment here required?

ui/components/ui/currency-display/currency-display.component.js
```diff
@@ -27,25 +28,30 @@ export default function CurrencyDisplay({
...
+      {parts ? (
```
brad-decker, line 35: when would parts not be defined?

brad-decker (changes requested):
haven't gotten very far but some initial thoughts. I'll keep going on this asap.

brad-decker (changes requested):
image
Is this file supposed to be here? Seems off

ui/components/app/wallet-overview/eth-overview.js
adonesky1: > image
> Is this file supposed to be here? Seems off

definitely not 😁

adonesky1:
> image
> Is this file supposed to be here? Seems off

definitely not 😁 ... been having weird stuff happening with my icloud sync 😢

ui/components/ui/currency-display/currency-display.component.js
adonesky1: So part of what this PR does is in cases where we don't have a valid conversion rate we no longer show the fiat/user-preferred currency. the useCurrencyDisplay hook can return [null, null] if we can't provide an accurate value for a currency that requires conversion. So for instance in the eth-overview.js component if we don't have a conversionRate for a secondary currency it will return null and we won't show a secondary currency display. This pattern still leaves a lot to be desired but it was an intermediate step for a larger refactor that I still hope to complete soonish.

ui/components/ui/currency-display/currency-display.component.js
adonesky1: In order to enable the entire return body of this component to be conditionally rendered ( per the nullable parts variable discussed below )it appears to be necessary to wrap in a fragment.

ui/components/app/wallet-overview/eth-overview.js
```diff
@@ -88,15 +95,15 @@ const EthOverview = ({ className }) => {
...
+                type={treatSecondaryCurrencyAsPrimary ? SECONDARY : PRIMARY}
                 ethNumberOfDecimals={4}
                 hideTitle
               />
               {balanceIsCached ? (
                 <span className="eth-overview__cached-star">*</span>
               ) : null}
             </div>
-            {showFiat && (
+            {showFiat && conversionRate && (
```
brad-decker, line 106: are there any cases where we should showFiat if conversionRate is null? If not could we move the check for conversionRate into the getShouldShowFiat selector?

ui/components/app/wallet-overview/eth-overview.js
```diff
@@ -72,6 +76,9 @@ const EthOverview = ({ className }) => {
   });
   const defaultSwapsToken = useSelector(getSwapsDefaultToken);
 
+  const treatSecondaryCurrencyAsPrimary =
```
brad-decker, line 79: is this logic duplicated anywhere? Are there any cases where we wouldn't swap these types if there isn't a conversion rate and the primary currency is not the native currency?

ui/components/app/wallet-overview/eth-overview.js
adonesky1: Good question. Let me poke around!

adonesky1: Moved this into a selector and consolidated in a few places as a result. There isn't anywhere I can think/find that we should showFiat if we don't have a valid conversion rate!

ui/components/app/wallet-overview/eth-overview.js
adonesky1: This precise logic doesn't appear to be duplicated anywhere. And as far as I know we would never want to show the "fiat" (or userPreferred currency) when we don't have an accurate conversionRate.

adonesky1: @brad-decker the pattern isn't as clear as I would like it to be and, as I said in a separate thread, I have a dormant larger refactor. But basically we're not showing the user preferred currency if we don't have a valid conversionRate and this placeholder piece of logic allows us to determine whether or not to sub in the native currency in this scenario.

ui/hooks/useTokenFiatAmount.js
```diff
@@ -32,7 +32,7 @@ export function useTokenFiatAmount(
   const conversionRate = useSelector(getConversionRate);
   const currentCurrency = useSelector(getCurrentCurrency);
   const userPrefersShownFiat = useSelector(getShouldShowFiat);
-  const showFiat = overrides.showFiat ?? userPrefersShownFiat;
+  const showFiat = overrides?.showFiat ?? userPrefersShownFiat;
```
brad-decker, line 35: This file throws an error when getConversionRate is hardwired to return undefined, because it doesn't check if this is available before trying to compute values.

ui/components/app/asset-list-item/asset-list-item.js
```diff
@@ -20,13 +20,15 @@ const AssetListItem = ({
   iconClassName,
   onClick,
   tokenAddress,
-  tokenSymbol,
+  primarySymbol,
```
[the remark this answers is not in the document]
brad-decker: Also related, when conversionRate is undefined (hardcoded in getConversionRate):

image

ui/selectors/confirm-transaction.js
```diff
@@ -230,13 +230,16 @@ export const transactionFeeSelector = function (state, txData) {
...
   }
 
-  const fiatTransactionAmount = getValueFromWeiHex({
-    value,
-    fromCurrency: nativeCurrency,
-    toCurrency: currentCurrency,
-    conversionRate,
-    numberOfDecimals: 2,
-  });
+  const fiatTransactionAmount = conversionRate
```
brad-decker, line 233: image
left: this PR, "Fiat" selected for primary, getConversionRate set to return undefined.

My assumption would be that, when the conversionRate is undefined, that the original UI (right) is shown but without the $0.00?

ui/hooks/useEthFiatAmount.js
```diff
@@ -23,10 +23,13 @@ export function useEthFiatAmount(
   const conversionRate = useSelector(getConversionRate);
   const currentCurrency = useSelector(getCurrentCurrency);
   const userPrefersShownFiat = useSelector(getShouldShowFiat);
-  const showFiat = overrides.showFiat ?? userPrefersShownFiat;
+  const showFiat = Boolean(overrides.showFiat ?? userPrefersShownFiat);
```
brad-decker, line 26: this boolean needs to be updated to take into account the possibility of conversionRate not being present.

ui/pages/swaps/swaps.util.js
```diff
@@ -451,6 +451,8 @@ export function getRenderableNetworkFeesForQuote({
   chainId,
   nativeCurrencySymbol,
 }) {
+  const showInFiat = Boolean(nativeCurrencySymbol && conversionRate);
```
brad-decker, line 454: Something in the swaps changes has caused the middle row to never fill in fee estimate regardless of conversionRate being defined or not: 

image

brad-decker (changes requested):
There's a number of issues with this PR due to conversionRate possibly being undefined where it was never assumed to be undefined before. Might want to hardcode the selector to return undefined to figure out what should be displayed in these cases. I've left a number of feedback items here too. I also started a thread in the design team channel to discuss how we should handle the changes to make assets display fiat as the "primary" display value when the appropriate setting is set.

ui/components/app/asset-list-item/asset-list-item.js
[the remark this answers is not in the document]
adonesky1: Re the last comment, the conversionRate should be null when not available? and I'm not seeing that behavior where some tokens are USD and others are with their symbols...?

Regarding the shape you suggest for primary/secondary -> asset/conversionAsset, I agree it is an improvement. I'll move forward with that for now.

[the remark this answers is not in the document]
brad-decker: ah sorry, you have to set "primary" currency as "Fiat" in settings to see this. I did update the getConversionRate selector to return null instead of undefined and same result.

ui/selectors/custom-gas.js
```diff
@@ -294,10 +294,10 @@ export function getRenderableEstimateDataForSmallButtonsFromGWEI(state) {
 
   const { showFiatInTestnets } = getPreferences(state);
   const isMainnet = getIsMainnet(state);
-  const showFiat = isMainnet || Boolean(showFiatInTestnets);
+  const { conversionRate } = state.metamask;
+  const showFiat = Boolean((isMainnet || showFiatInTestnets) && conversionRate);
```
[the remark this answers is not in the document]
adonesky1: yessir!

adonesky1:
The core of this PR was extracted and [merged](https://github.com/MetaMask/metamask-extension/pull/11335). The remaining pieces will be added in subsequent PRs.

PR #11072, opened 2021-05-12
gitignore ui/app/css/output css output files

tmashuang:
Adds `ui/app/css/output/`, css output, to the to the gitignore list.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ build-artifacts
 
 #ignore css output and sourcemaps
 ui/css/output/
+ui/app/css/output/
 
 notes.txt
 


.gitignore
```diff
@@ -40,6 +40,7 @@ build-artifacts
 
 #ignore css output and sourcemaps
 ui/css/output/
+ui/app/css/output/
```
brad-decker, line 43: these can be safely deleted and should not reappear because ui/app no longer exists

brad-decker (changes requested):

.gitignore
tmashuang: Hmm seems the files only shows up for me locally? If I just delete them it should be fine I suppose?

brad-decker: yeah should be fine, happened to me a lot during the refactor to remove the folder

PR #11085, opened 2021-05-13
Migrate Sentry settings to environment variables

Gudahtt:
Sentry is now configured with environment variables, rather than with hard-coded values. This makes it easier to test Sentry functionality using a different Sentry account, as we did recently during QA of v9.5.1.

The only change for the normal build process is the introduction of the `SENTRY_DSN_DEV` variable, which can be set via `.metamaskrc` or via an environment variable. This determines where error reports are sent. It still defaults to our team Sentry account's `metamask-testing` project.

The `sentry:publish` script now requires SENTRY_ORG and SENTRY_PROJECT to be set in order to publish release artifacts. The CircleCI configuration has been updated with these values, so it should act the same as it did before. Previously we had used a CLI flag to specify the organization and project, but [Sentry already natively supports these environment variables](https://docs.sentry.io/product/cli/configuration/#configuration-values).

Manual testing steps for `SENTRY_DSN_DEV`
  - Create a free Sentry account and log in, and create a project. Take note of the DSN.
  - Build the extension using `yarn build:test` or `yarn dist` (only test/prod builds send Sentry errors, they are disabled for dev)
  - Install the extension
  - Do something that causes an error (it may be easiest to introduce a false error for testing)
  - Observe the network tab of whichever process the error is thrown in, and see that it's sending to the correct DSN. You should also see the error on your Sentry dashboard.
  
  Manual testing steps for release artifact publishing:
  - Create a free Sentry account and log in, and create a project.
  - Navigate to "Settings => Developer Settings => "New Internal Integration"
  - Assign a name to this integration, and grant it a "Release" permission of "Admin".
  - After saving this new integration, take note of the token that should be listed on the integration page.
  - Run the command `yarn sentry:publish` with the environment variable `SENTRY_AUTH_TOKEN` set to the token from that integration, and with `SENTRY_ORG` set to your account name and `SENTRY_PROJECT` set to the project name you created.
  - It should complete successfully. You should see the release under "Releases" with the correct project selected. It should have 22 artifacts.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -505,7 +505,7 @@ jobs:
           at: .
       - run:
           name: sentry sourcemaps upload
-          command: yarn sentry:publish
+          command: SENTRY_ORG=metamask SENTRY_PROJECT=metamask yarn sentry:publish
       - run:
           name: Create GitHub release
           command: |
--- a/app/scripts/lib/setupSentry.js
+++ b/app/scripts/lib/setupSentry.js
@@ -7,9 +7,8 @@ import extractEthjsErrorMessage from './extractEthjsErrorMessage';
 // Destructuring breaks the inlining of the environment variables
 const METAMASK_DEBUG = process.env.METAMASK_DEBUG;
 const METAMASK_ENVIRONMENT = process.env.METAMASK_ENVIRONMENT;
+const SENTRY_DSN_DEV = process.env.SENTRY_DSN_DEV;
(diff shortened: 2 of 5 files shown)


Gudahtt:
I have followed both manual testing instructions, and confirmed that this works with all three of these new environment variables set to point at a different Sentry account.

brad-decker (approved):
followed testing instructions for both reporting and artifact uploading and everything works well.

PR #11087, opened 2021-05-13
Swaps: Add more anonymized tracking for HW wallet and HW wallet type

dan437:
## Explanation
It adds more anonymized tracking of HW wallets. More details in the previous PR: https://github.com/MetaMask/metamask-extension/pull/11063


brad-decker (approved):
LGTM

PR #11088, opened 2021-05-13
Swaps: Add specific error content if Contract data are not enabled on Ledger

dan437:
## Explanation
It seems that by default Ledger has Contract data disabled inside its Ethereum app. When a user tries to use a Ledger device with its default config, they will see a Swaps Failed page when they try to swap tokens and it's not really clear what the user should do.

This PR shows a new error page with specific error and description about how to resolve it, so a user can enable Contract data on their Ledger device and do Swaps successfully.

## Manual testing steps
- Try to use Ledger with default config for swaps
- After agreeing with a quote a new error page is displayed
- Enable Contract data on a Ledger device
- Click on "Try again"
- Swapping works now

## Screenshots
image

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -1963,10 +1963,10 @@
     "message": "No quotes available"
   },
   "swapContractDataDisabledErrorDescription": {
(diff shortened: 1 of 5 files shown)


brad-decker (approved):
Super cool, tested with my ledger and works for me.

PR #11089, opened 2021-05-13
create safer isValidAddress method

brad-decker:
Instead of using the `isValidAddress` method from `ethereumjs-util` that now throws when a non hex prefixed string is provided, or the overloaded `isValidAddress` helper we maintain that checks a number of things, we should use a more effective shared method. 

This PR creates two new shared methods:
1. isBurnAddress - is the address exactly 0x0000000000000000000000000000000000000000 which is an address that is used sometimes for burning tokens and one we want to guard our users from using. 
2. isValidHexAddress - checks if the address is possibly a checksum address and uses the checksum address validation method or the address validation method appropriately. 

The tests from our 'isValidAddress' implementation have been copied over and only ONE has been altered fundamentally. This new method does not return true for forty-character non hex-prefixed strings. If user input would be valid without the hex prefix, then it should be hex prefixed before calling this method.

### DIFF
--- a/app/scripts/controllers/preferences.js
+++ b/app/scripts/controllers/preferences.js
@@ -867,7 +867,7 @@ export default class PreferencesController {
         `Invalid decimals "${decimals}": must be 0 <= 36.`,
       );
     }
-    if (!isValidHexAddress(address, false)) {
+    if (!isValidHexAddress(address, { allowNonPrefixed: false })) {
       throw ethErrors.rpc.invalidParams(`Invalid address "${address}".`);
     }
   }
--- a/app/scripts/controllers/preferences.js
+++ b/app/scripts/controllers/preferences.js
@@ -2,12 +2,12 @@ import { strict as assert } from 'assert';
 import { ObservableStore } from '@metamask/obs-store';
 import { ethErrors } from 'eth-rpc-errors';
 import { normalize as normalizeAddress } from 'eth-sig-util';
-import { isValidAddress } from 'ethereumjs-util';
 import ethers from 'ethers';
 import log from 'loglevel';
 import { LISTED_CONTRACT_ADDRESSES } from '../../../shared/constants/tokens';
 import { NETWORK_TYPE_TO_ID_MAP } from '../../../shared/constants/network';
 import { isPrefixedFormattedHexString } from '../../../shared/modules/network.utils';
+import { isValidHexAddress } from '../../../shared/modules/hexstring-utils';
 import { NETWORK_EVENTS } from './network';
 
 export default class PreferencesController {
@@ -867,7 +867,7 @@ export default class PreferencesController {
         `Invalid decimals "${decimals}": must be 0 <= 36.`,
       );
     }
-    if (!isValidAddress(address)) {
+    if (!isValidHexAddress(address, false)) {
       throw ethErrors.rpc.invalidParams(`Invalid address "${address}".`);
     }
   }
--- a/app/scripts/lib/typed-message-manager.js
+++ b/app/scripts/lib/typed-message-manager.js
@@ -160,7 +160,8 @@ export default class TypedMessageManager extends EventEmitter {
     assert.ok('data' in params, 'Params must include a "data" field.');
     assert.ok('from' in params, 'Params must include a "from" field.');
     assert.ok(
-      typeof params.from === 'string' && isValidHexAddress(params.from, false),
+      typeof params.from === 'string' &&
+        isValidHexAddress(params.from, { allowNonPrefixed: false }),
       '"from" field must be a valid, lowercase, hexadecimal Ethereum address string.',
     );
 
--- a/app/scripts/lib/typed-message-manager.js
+++ b/app/scripts/lib/typed-message-manager.js
@@ -3,12 +3,12 @@ import { strict as assert } from 'assert';
 import { ObservableStore } from '@metamask/obs-store';
 import { ethErrors } from 'eth-rpc-errors';
 import { typedSignatureHash, TYPED_MESSAGE_SCHEMA } from 'eth-sig-util';
-import { isValidAddress } from 'ethereumjs-util';
 import log from 'loglevel';
 import jsonschema from 'jsonschema';
 import { MESSAGE_TYPE } from '../../../shared/constants/app';
 import { METAMASK_CONTROLLER_EVENTS } from '../metamask-controller';
 import createId from '../../../shared/modules/random-id';
+import { isValidHexAddress } from '../../../shared/modules/hexstring-utils';
 
 /**
  * Represents, and contains data about, an 'eth_signTypedData' type signature request. These are created when a
@@ -160,7 +160,7 @@ export default class TypedMessageManager extends EventEmitter {
     assert.ok('data' in params, 'Params must include a "data" field.');
     assert.ok('from' in params, 'Params must include a "from" field.');
     assert.ok(
-      typeof params.from === 'string' && isValidAddress(params.from),
+      typeof params.from === 'string' && isValidHexAddress(params.from, false),
       '"from" field must be a valid, lowercase, hexadecimal Ethereum address string.',
     );
 
--- a/shared/modules/hexstring-utils.js
+++ b/shared/modules/hexstring-utils.js
@@ -17,17 +17,20 @@ export function isBurnAddress(address) {
  * does not throw an error when provided values that are not hex strings. In
  * addition, and by default, this method will return true for hex strings that
  * meet the length requirement of a hex address, but are not prefixed with `0x`
- * Finally, if a mixed case string is provided this method assumes the address
- * is a checksum address and will validate it as such.
+ * Finally, if the mixedCaseUseChecksum flag is true and a mixed case string is
+ * provided this method will validate it has the proper checksum formatting.
  * @param {string} possibleAddress - Input parameter to check against
  * @param {Object} [options] - options bag
- * @param {boolean} [options.allowNonPrefixed] - If true will first ensure '0x' is
- *  prepended to the string
+ * @param {boolean} [options.allowNonPrefixed] - If true will first ensure '0x'
+ *  is prepended to the string
+ * @param {boolean} [options.mixedCaseUseChecksum] - If true will treat mixed
+ *  case addresses as checksum addresses and validate that proper checksum
+ *  format is used
  * @returns {boolean} whether or not the input is a valid hex address
  */
 export function isValidHexAddress(
   possibleAddress,
-  { allowNonPrefixed = true } = {},
+  { allowNonPrefixed = true, mixedCaseUseChecksum = false } = {},
 ) {
   const addressToCheck = allowNonPrefixed
     ? addHexPrefix(possibleAddress)
@@ -36,12 +39,14 @@ export function isValidHexAddress(
     return false;
   }
 
-  const prefixRemoved = addressToCheck.slice(2);
-  const lower = prefixRemoved.toLowerCase();
-  const upper = prefixRemoved.toUpperCase();
-  const allOneCase = prefixRemoved === lower || prefixRemoved === upper;
-  if (!allOneCase) {
-    return isValidChecksumAddress(addressToCheck);
+  if (mixedCaseUseChecksum) {
+    const prefixRemoved = addressToCheck.slice(2);
+    const lower = prefixRemoved.toLowerCase();
+    const upper = prefixRemoved.toUpperCase();
+    const allOneCase = prefixRemoved === lower || prefixRemoved === upper;
+    if (!allOneCase) {
+      return isValidChecksumAddress(addressToCheck);
+    }
   }
 
   return isValidAddress(addressToCheck);
--- a/shared/modules/hexstring-utils.js
+++ b/shared/modules/hexstring-utils.js
@@ -11,7 +11,24 @@ export function isBurnAddress(address) {
   return address === BURN_ADDRESS;
 }
 
-export function isValidHexAddress(possibleAddress, allowNonPrefixed = true) {
+/**
+ * Validates that the input is a hex address. This utility method is a thin
+ * wrapper around ethereumjs-util.isValidAddress, with the exception that it
+ * does not throw an error when provided values that are not hex strings. In
+ * addition, and by default, this method will return true for hex strings that
+ * meet the length requirement of a hex address, but are not prefixed with `0x`
+ * Finally, if a mixed case string is provided this method assumes the address
+ * is a checksum address and will validate it as such.
+ * @param {string} possibleAddress - Input parameter to check against
+ * @param {Object} [options] - options bag
+ * @param {boolean} [options.allowNonPrefixed] - If true will first ensure '0x' is
+ *  prepended to the string
+ * @returns {boolean} whether or not the input is a valid hex address
+ */
+export function isValidHexAddress(
+  possibleAddress,
+  { allowNonPrefixed = true } = {},
+) {
   const addressToCheck = allowNonPrefixed
     ? addHexPrefix(possibleAddress)
     : possibleAddress;
--- a/shared/modules/hexstring-utils.js
+++ b/shared/modules/hexstring-utils.js
@@ -1,0 +1,31 @@
+import {
+  isHexString,
+  isValidAddress,
+  isValidChecksumAddress,
+  addHexPrefix,
+} from 'ethereumjs-util';
+
+export const BURN_ADDRESS = '0x0000000000000000000000000000000000000000';
+
+export function isBurnAddress(address) {
+  return address === BURN_ADDRESS;
+}
+
+export function isValidHexAddress(possibleAddress, allowNonPrefixed = true) {
+  const addressToCheck = allowNonPrefixed
+    ? addHexPrefix(possibleAddress)
+    : possibleAddress;
+  if (!isHexString(addressToCheck)) {
+    return false;
+  }
+
+  const prefixRemoved = addressToCheck.slice(2);
+  const lower = prefixRemoved.toLowerCase();
+  const upper = prefixRemoved.toUpperCase();
+  const allOneCase = prefixRemoved === lower || prefixRemoved === upper;
+  if (!allOneCase) {
+    return isValidChecksumAddress(addressToCheck);
+  }
+
+  return isValidAddress(addressToCheck);
+}
--- a/ui/components/ui/identicon/identicon.component.js
+++ b/ui/components/ui/identicon/identicon.component.js
@@ -2,8 +2,8 @@ import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import contractMap from '@metamask/contract-metadata';
+import { isHexString, addHexPrefix } from 'ethereumjs-util';
 
-import { isValidHexAddress } from '../../../../shared/modules/hexstring-utils';
 import { checksumAddress } from '../../../helpers/utils/util';
 import Jazzicon from '../jazzicon';
 import BlockieIdenticon from './blockieIdenticon';
@@ -85,8 +85,8 @@ export default class Identicon extends PureComponent {
       return this.renderImage();
     }
 
-    if (isValidHexAddress(address)) {
-      const checksummedAddress = checksumAddress(address);
+    if (isHexString(addHexPrefix(address))) {
+      const checksummedAddress = checksumAddress(addHexPrefix(address));
 
       if (contractMap[checksummedAddress]?.logo) {
         return this.renderJazzicon();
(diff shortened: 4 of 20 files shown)


jsconfig.json
```diff
@@ -1,14 +1,3 @@
...
-    "builds",
-    "coverage",
-    "dist",
-    "docs",
-    "lavamoat",
-    "node:console",
-    "node_modules",
-    "patches",
-    "test-artifacts"
-  ]
```
ryanml, line 13: were these exclusions no needed here or?

brad-decker: oh, not part of this i'll remove. With them in autocomplete just randomly starts failing for me.

ryanml (dismissed):
lgtm!

brad-decker:
I discovered that our internal 'isHex' tool also doesn't validate that the hex string starts with '0x' so the previous fix by @tmashuang was not sufficient to prevent errors for real addresses that aren't hex prefixed, so this PR replaces the usages of that with isValidHexAddress (in no case where it was used did we expect anything other than an address. Note that Thomaas' change does prevent issues when non hex strings (like domain names) were sent to the icons. 

Also reverted the change that made non 0x prefixed  hex strings invalid for the purposes of this method but we should explore changing this behavior in the future.

brad-decker:
One more tweak, i went through all usages of isValidAddress (both our util and the ethereumjs-util function) and restored it's rightful behavior by adding a new prop to the new helper called 'allowNonPrefixed' ...why?

ethereumjs-util.isValidAddress prior to my update would return false for non hex prefixed strings, whereas our method would return true provided that it was a 40 character hex string. 

This new prop allows us to have both behaviors in one method, and a path for gradual migration of using non hex prefixed values and allowing them from users.

shared/modules/hexstring-utils.js
```diff
@@ -0,0 +1,31 @@
...
+  addHexPrefix,
+} from 'ethereumjs-util';
+
+export const BURN_ADDRESS = '0x0000000000000000000000000000000000000000';
+
+export function isBurnAddress(address) {
+  return address === BURN_ADDRESS;
+}
+
+export function isValidHexAddress(possibleAddress, allowNonPrefixed = true) {
```
Gudahtt, line 14: Nit: This might be easier to understand if the second parameter was an options bag.

e.g. `isValidHexAddress(address, { allowNonPrefixed: false })` v.s. `isValidHexAddress(address, false)`

I was a bit confused about what the `false` meant at first, reading through this PR.

ui/components/ui/identicon/identicon.component.js
```diff
@@ -84,13 +85,11 @@ export default class Identicon extends PureComponent {
       return this.renderImage();
     }
 
-    if (address) {
-      if (isHex(address)) {
-        const checksummedAddress = checksumAddress(address);
+    if (isValidHexAddress(address)) {
```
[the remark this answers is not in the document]
Gudahtt: Also, this might prevent the display of identicons for any invalid addresses, instead showing a plain circle. We use identicons in _a lot_ of places, so I would expect that some of them will still have "invalid" addresses.

ui/helpers/utils/icon-factory.js
```diff
@@ -18,7 +19,7 @@ function IconFactory(jazzicon) {
 IconFactory.prototype.iconForAddress = function (address, diameter) {
   let addr = address;
 
-  if (isHex(address)) {
+  if (isValidHexAddress(address, false)) {
```
Gudahtt: This seems dangerous. Previously `checksumAddress` would add the prefix, and checksum the address regardless of whether it's mixed-case or not. Now it requires the address is already hex-prefixed and either checksummed already or all one case. This means for any other case, we're skipping this normalization step, and we'll be generating a different icon than before.

shared/modules/hexstring-utils.js
brad-decker: I did that and added docs 🤦🏻 was maybe rushing too much last week.

ui/pages/add-token/add-token.component.js
```diff
@@ -167,7 +165,7 @@ class AddToken extends Component {
       autoFilled: false,
     });
 
-    const addressIsValid = isValidAddress(customAddress);
+    const addressIsValid = isValidHexAddress(customAddress, false);
```
Gudahtt: This behavior is pre-existing, but it does seem strange to me that we'd show an "invalid address" error when the address is mixed-case but doesn't match our checksum. Particularly when we're not checksumming the addresses properly according to EIP-1191, as we're omitting the chainID. As a result, we'd be showing "invalid address" when an address was properly checksummed in accordance with EIP-1191.

ui/pages/send/send-content/add-recipient/add-recipient.component.js
```diff
@@ -101,7 +104,7 @@ export default class AddRecipient extends Component {
 
     let content;
 
-    if (isValidAddress(query)) {
+    if (!isBurnAddress(query) && isValidHexAddress(query)) {
```
Gudahtt, line 107: It seems overly harsh to disallow differently-checksummed addresses here.

The same goes for the other uses of this function in the Send flow and the contacts flow.

ui/pages/settings/settings.container.js
```diff
@@ -64,7 +64,7 @@ const mapStateToProps = (state, ownProps) => {
 
   const addressName = getAddressBookEntryName(
     state,
-    isHex(pathNameTail) && isValidAddress(pathNameTail) ? pathNameTail : '',
+    isValidHexAddress(pathNameTail) ? pathNameTail : '',
```
Gudahtt, line 67: I wonder if we have any existing entries that this would be `false` for? :thinking:

ui/pages/swaps/swaps.util.js
```diff
@@ -74,8 +74,8 @@ const QUOTE_VALIDATORS = [
     validator: (trade) =>
       trade &&
       validHex(trade.data) &&
-      isValidAddress(trade.to) &&
-      isValidAddress(trade.from) &&
+      isValidHexAddress(trade.to) &&
```
Gudahtt, line 77: It looks like these APIs give all lower-cased addresses (at least at first glance), so they might not be affected by this checksumming issue.

Gudahtt (changes requested):
I like the new function, and the consolidation of all of those utility functions. But it seems that there are a lot of potentially dapp-breaking changes in here still, and a few undesirable changes to the UI.

It makes perfect sense to me that we'd ensure our dapp interactions use properly-formatted addresses, but when making _any_ non-backwards-compatible changes to our provider (however reasonable) we should consider how best to communicate these changes before making them. Giving plenty of advanced notice, and maybe shipping a deprecation notice ahead of time, would be much friendlier to devs. Especially in cases like this where there are few downsides for us to preserving this behaviour a little longer.

And regarding checksumming, we should consider adopting support for EIP-1191, and considering addresses checksummed by EIP-1191 or by EIP-55 as valid. Or alternatively, we could skip validating mixed-case addresses as checksummed for now (or not extending our validation further at least). I'd hate to see us accidentally stand in the way of a new, helpful EIP becoming more widespread.

brad-decker:
@Gudahtt my goal was to preserve the behavior pre 9.5.0 and the `ethereumjs-util` and I *thought* I was being diligent in doing so. 😓 I'm going to add a new parameter (validateChecksumAddress) that will default to false and be opt-in. I hope you won't mind being a second pair of eyes again while I do another pass :(

app/scripts/controllers/preferences.js
```diff
@@ -867,7 +867,7 @@ export default class PreferencesController {
         `Invalid decimals "${decimals}": must be 0 <= 36.`,
       );
     }
-    if (!isValidAddress(address)) {
+    if (!isValidHexAddress(address, false)) {
```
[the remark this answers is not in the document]
brad-decker: to triple check this I installed `ethereumjs-util` 5.1.0 in a new project and tested in node.

```
> var x = require('ethereumjs-util')
undefined
> x.isValidAddress('0x0000000000000000000000000000000000000000')
true
> x.isValidAddress('0000000000000000000000000000000000000000')
false
```

the behavior change is that now it *throws* an error versus returning false, so requiring hex prefix enforces the old behavior, but checksum checking is still erroneous here so the new parameter being added will prevent it from validating checksum addresses.

app/scripts/lib/typed-message-manager.js
```diff
@@ -160,7 +160,7 @@ export default class TypedMessageManager extends EventEmitter {
     assert.ok('data' in params, 'Params must include a "data" field.');
     assert.ok('from' in params, 'Params must include a "from" field.');
     assert.ok(
-      typeof params.from === 'string' && isValidAddress(params.from),
+      typeof params.from === 'string' && isValidHexAddress(params.from, false),
```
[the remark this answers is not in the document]
brad-decker: it was previously required, but would just return false for non-prefixed, rather than throwing an error.

shared/modules/hexstring-utils.js
```diff
@@ -0,0 +1,31 @@
...
+  if (!isHexString(addressToCheck)) {
+    return false;
+  }
+
+  const prefixRemoved = addressToCheck.slice(2);
+  const lower = prefixRemoved.toLowerCase();
+  const upper = prefixRemoved.toUpperCase();
+  const allOneCase = prefixRemoved === lower || prefixRemoved === upper;
+  if (!allOneCase) {
+    return isValidChecksumAddress(addressToCheck);
```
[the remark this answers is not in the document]
brad-decker: Very good point, i'm makin this behavior opt-in and will turn it on in the places where the old 'isValidAddress' custom helper was used.

app/scripts/controllers/preferences.js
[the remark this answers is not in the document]
Gudahtt: Ah, good catch! I hadn't realized that it returned `false`  in v5.1.0.

ui/pages/send/send-content/add-recipient/add-recipient.component.js
brad-decker: While that is true, it is the previously implemented behavior. Should I change that as part of this PR? These places used the previous 'isValidAddress' custom function which did check for mixed case and do validation using isValidChecksum.

Gudahtt:
Sure, of course! That solution sounds perfect to me.

ui/pages/send/send-content/add-recipient/add-recipient.component.js
Gudahtt: It would be perfectly reasonable to stick with targeting pre-v9.5.0 behavior IMO. Particularly since this is destined for a hotfix.

ui/components/ui/identicon/identicon.component.js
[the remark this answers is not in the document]
brad-decker: wait, so what are the cases we want to not show a jazzicon right now? that might be easier to reason about? IF jazzicons allow non hex strings should we just move the part where we render the blockie or jazzicon (if there isn't a matching entry in the contractMap) out one level and remove the fallback render?

ui/helpers/utils/icon-factory.js
brad-decker: under the hood, the utility method 'checksumAddress' calls ethereumjs-util's toChecksumAddress which in 5.1 adds the '0x' prefix, and in the new version, it throws an error. It seems like we should probably add a new utility method that mimics this behavior because as is even when allowing this to validate true for a non hex prefixed string, the string is not modified meaning it would immediately throw on the next line.

I'm going to add this utility method now, and replace usages of the old helper method.

brad-decker: actually I should do that in a new PR instead to prevent convoluting this with new logic

brad-decker:
There are no futher usages of either:
1. `ethereumjs-util.isValidAddress`
2. `isValidAddress` from our helpers

in all cases to restore pre 9.5 behavior the new isValidHexAddress should be called in the following ways:
a. For 1 above should call `isValidHexAddress(input, { allowNonPrefixed: false })`
b. For 2 above should call `isValidHexAddress(input, { mixedCaseUseChecksum: true })` and separately check if the burn address is used.

I did an audit of all places where isValidHexAddress is invoked (29 invocations)
1. app/scripts/controllers/preferences.js - 1 invocation - previously used 1, correctly applied a.
2. app/scripts/controllers/transactions/lib/util.js - 2 invocations - previously used 1, correctly applied a.
3. app/scripts/lib/typed-message-manager.js - 1 invocation - previously used 1, correctly applied a.
4. shared/modules/hexstring-utils.test.js - 8 invocations - new test file, irrelevant
5. ui/pages/add-token/add-token.component.js - 1 invocation - previously used 1, correctly applied a.
6. ui/pages/send/send.component.js - 1 invocation - previously used 1, correctly applied a.
7. ui/pages/send/send-content/add-recipient/add-recipient.component.js - 1 invocation - previously used 2, correctly applied b.
8. ui/pages/send/send-content/add-recipient/add-recipient.js - 1 invocation - previously used 2, correctly applied b.
9. ui/pages/send/send-content/add-recipient/ens-input.component.js - 3 invocations - previously used 2, correctly applied b.
10. ui/pages/settings/settings.container.js - 1 invocation - previously used 2, correctly applied b.
11. ui/pages/settings/contact-list-tab/add-contact/add-contact.component.js - 1 invocation - previously used 2, correctly applied b.
12. ui/pages/settings/contact-list-tab/edit-contact/edit-contact.component.js - 1 invocation - previously used 2, correctly applied b.
13. ui/pages/swaps/swaps.util.js - 4 invocations - previously used 1, correctly applied a.

There are two files that deserve special mention:
1. ui/helpers/utils/icon-factory.js - 2 invocations - previously only had 1 usage using 1, and so in that instance applied a.
2. ui/components/ui/identicon/identicon.component.js - 1 invocation - previously did not have an invocation

the new usages of this utility method in the above two instances will be *removed* in a new PR adding the toChecksumHexAddress.

ui/pages/settings/settings.container.js
brad-decker: not sure I understand the question here, this code previously used 'isValidAddress' from ethereumjs-util and as part of the original ethereumjs-util upgrade I added 'isHex' ahead of this because non-address strings in the URL were being checked to see if they were a valid address and failing. Can't remember the exact path that was being checked when I noticed the failure case :/

ui/helpers/utils/icon-factory.js
```diff
@@ -52,7 +53,9 @@ IconFactory.prototype.generateNewIdenticon = function (address, diameter) {
 
 function iconExistsFor(address) {
   return (
-    contractMap[address] && isValidAddress(address) && contractMap[address].logo
+    contractMap[address] &&
+    isValidHexAddress(address) &&
```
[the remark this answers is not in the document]
brad-decker: yep, and it was previously using the ethereumjs-util method so it would have returned false for any non prefixed ones.

ha842:
Pull

ui/pages/add-token/add-token.component.js
```diff
@@ -167,7 +165,9 @@ class AddToken extends Component {
       autoFilled: false,
     });
 
-    const addressIsValid = isValidAddress(customAddress);
+    const addressIsValid = isValidHexAddress(customAddress, {
```
Gudahtt, line 168: This was using the `isValidAddress` function from our util file, so we'd want to take the 'B' strategy that you listed (check burn address and set `mixedCaseUseChecksum` to `true`).

ui/pages/send/send.component.js
```diff
@@ -171,7 +171,7 @@ export default class SendTransactionScreen extends Component {
     if (qrCodeData) {
       if (qrCodeData.type === 'address') {
         scannedAddress = qrCodeData.values.address.toLowerCase();
-        if (isValidAddress(scannedAddress)) {
+        if (isValidHexAddress(scannedAddress)) {
```
Gudahtt, line 174: This should be following the B strategy too I think?

ui/pages/swaps/swaps.util.js
```diff
@@ -101,12 +101,12 @@ const QUOTE_VALIDATORS = [
   {
     property: 'sourceToken',
     type: 'string',
-    validator: isValidAddress,
+    validator: isValidHexAddress,
```
Gudahtt, line 104: This should be following the A strategy, but I guess that would require creating a local helper function to set the option, since it's passed in as a value :thinking: 

Seems fine to leave it as-is if we're confident the data here is always prefixed though, since its our API.

ui/pages/settings/settings.container.js
Gudahtt: I was thinking of the difference in checksumming, wondering if this change would prevent this property from being set when viewing a contact with a mixed-case, non-checksummed address.

But with the most recent changes, that's no longer a concern.

ui/pages/add-token/add-token.component.js
brad-decker: https://github.com/MetaMask/metamask-extension/blob/v9.4.0/ui/app/pages/add-token/add-token.component.js#L164

ui/pages/send/send.component.js
brad-decker: https://github.com/MetaMask/metamask-extension/blob/v9.4.0/ui/app/pages/send/send.component.js#L174

ui/pages/swaps/swaps.util.js
brad-decker: just to be super safe, I could do `validator: (input) => isValidHexAddress(input, { allowNonPrefixed: false })` ?

Gudahtt: That works!

ui/pages/add-token/add-token.component.js
Gudahtt: Ah, I didn't realize we switched between those two methods during the `ethereumjs-util` update. Good catch!

ui/pages/send/send.component.js
Gudahtt: Whoops! So should we use strategy A here instead?

brad-decker: Huh, thought I did 🤦🏻  done now!

ui/pages/swaps/swaps.util.js
brad-decker: Done in latest

Gudahtt (commented):
Everything looks good aside from these two identicon-related changes. I suggested that we revert changes for now so that we're not making things any worse, until we're prepared to fix those properly. But let me know if you think there's a good reason to keep these changes.

ui/helpers/utils/icon-factory.js
```diff
@@ -18,7 +19,7 @@ function IconFactory(jazzicon) {
 IconFactory.prototype.iconForAddress = function (address, diameter) {
   let addr = address;
 
-  if (isHex(address)) {
+  if (isValidHexAddress(address)) {
```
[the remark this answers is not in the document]
brad-decker: rather than reverting the old custom 'isHex' function which did not validate the existence of prefix '0x', I did the following:
```if (isHexString(addHexPrefix(address))) {
      const checksummedAddress = checksumAddress(addHexPrefix(address));``` which will eventually be encapsulated in the helper method mentioned abovee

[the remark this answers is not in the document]
Comment (Gudahtt) [reply ui/helpers/utils/icon-factory.js]: Makes sense! This looks good

File: ui/components/ui/identicon/identicon.component.js [hunk 17]
```diff
@@ -84,13 +85,11 @@ export default class Identicon extends PureComponent {
       return this.renderImage();
     }
 
-    if (address) {
-      if (isHex(address)) {
-        const checksummedAddress = checksumAddress(address);
+    if (isHexString(addHexPrefix(address))) {
```
Gudahtt, line 88: Ok, down to one last issue! I think this is an excellent improvement for the old `if (isHex(address)) { ... }` block, but the `if (address)` part should be restored. Otherwise we're preventing jazzicons from being rendered for token/currency symbols (which, confusingly, are supplied using the `address` prop).

brad-decker:
```
    if (address) {
      const checksummedAddress =
        isHexString(addHexPrefix(address)) &&
        checksumAddress(addHexPrefix(address));

      if (checksummedAddress && contractMap[checksummedAddress]?.logo) {
        return this.renderJazzicon();
      }

      return (
        <div
          className={classnames({ 'identicon__address-wrapper': addBorder })}
        >
          {useBlockie ? this.renderBlockie() : this.renderJazzicon()}
        </div>
      );
    }
```
Think that does it!

Gudahtt (approved):
LGTM!

brad-decker:
thank you thank you thank you, @Gudahtt. 🙏🏻 🎉 💯 🥇 🚀

PR #11101, opened 2021-05-14
Improve type checking and handle incorrect category in transaction-icon component

danjm:
This PR makes the transaction-icon component resilient against erroneous props. Better proptype checking is added, and changes are made to ensure that if flawed data is passed it doesn't break the whole UX.

In cases where a non-required transaction category is passed, we default to a grey circle.

Screenshot from 2021-05-14 17-34-26


ui/components/app/transaction-icon/index.scss
```diff
@@ -0,0 +1,8 @@
+.transaction-icon {
```
brad-decker, line 1: [REQ] lets use the name of the component as the filename instead of index, so that we can more easily search/and find these and understand where they're coming from in source maps

brad-decker (changes requested):

ui/components/app/transaction-icon/index.scss
danjm: done in the latest commit

Gudahtt (approved):
LGTM!

brad-decker (approved):

PR #11102, opened 2021-05-14
Fixing unlocalized strings

### LINKED ISSUE #11094: Localize 'View More' button on transaction list
The "View More" button on the transaction list should be localized.

It is here: https://github.com/MetaMask/metamask-extension/blob/85eb62e8cf887e3de6179342fcd15dcfd909148f/ui/components/app/transaction-list/transaction-list.component.js#L162

ryanml:
Fixes MetaMask/metamask-extension#11094

### DIFF
--- a/ui/components/app/modals/metametrics-opt-in-modal/metametrics-opt-in-modal.component.js
+++ b/ui/components/app/modals/metametrics-opt-in-modal/metametrics-opt-in-modal.component.js
@@ -50,12 +50,6 @@ export default class MetaMetricsOptInModal extends Component {
                     {t('metametricsCommitmentsSendAnonymizedEvents')}
                   </div>
                 </div>
-                <div className="metametrics-opt-in__row">
-                  <i className="fa fa-check" />
-                  <div className="metametrics-opt-in__row-description">
-                    {t('metametricsCommitmentsMaintainAggregate')}
-                  </div>
-                </div>
                 <div className="metametrics-opt-in__row metametrics-opt-in__break-row">
                   <i className="fa fa-times" />
                   <div className="metametrics-opt-in__row-description">
--- a/ui/components/app/modals/metametrics-opt-in-modal/metametrics-opt-in-modal.component.js
+++ b/ui/components/app/modals/metametrics-opt-in-modal/metametrics-opt-in-modal.component.js
@@ -69,7 +69,7 @@ export default class MetaMetricsOptInModal extends Component {
                     {t('metametricsCommitmentsNeverCollectIP', [
                       <span
                         className="metametrics-opt-in__bold"
-                        key="neverCollectKeys"
+                        key="neverCollectIP"
                       >
                         {t('metametricsCommitmentsBoldNever')}
                       </span>,
@@ -82,7 +82,7 @@ export default class MetaMetricsOptInModal extends Component {
                     {t('metametricsCommitmentsNeverSellDataForProfit', [
                       <span
                         className="metametrics-opt-in__bold"
-                        key="neverCollectKeys"
+                        key="neverSellData"
                       >
                         {t('metametricsCommitmentsBoldNever')}
                       </span>,
--- a/ui/components/app/modals/metametrics-opt-in-modal/metametrics-opt-in-modal.component.js
+++ b/ui/components/app/modals/metametrics-opt-in-modal/metametrics-opt-in-modal.component.js
@@ -27,76 +27,87 @@ export default class MetaMetricsOptInModal extends Component {
               <img src="images/metrics-chart.svg" alt="" />
             </div>
             <div className="metametrics-opt-in__title">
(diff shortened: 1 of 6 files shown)


ui/components/app/modals/metametrics-opt-in-modal/metametrics-opt-in-modal.component.js
```diff
@@ -27,76 +27,87 @@ export default class MetaMetricsOptInModal extends Component {
...
                   <i className="fa fa-check" />
                   <div className="metametrics-opt-in__row-description">
-                    Send anonymized click & pageview events
+                    {t('metametricsCommitmentsSendAnonymizedEvents')}
                   </div>
                 </div>
                 <div className="metametrics-opt-in__row">
                   <i className="fa fa-check" />
                   <div className="metametrics-opt-in__row-description">
-                    Maintain a public aggregate dashboard to educate the
```
Gudahtt, line 59: Ah, I was wondering what was going on here - I thought we localized this already! We did, but only during onboarding. This second copy of the metametrics agreement was forgotten.

This is intended to match the onboarding metrics agreement exactly, which has had this bullet removed (since we didn't actually deliver on this commitment, and had no immediate plans to). This should be compared with the onboarding page to ensure no other differences exist either.

Gudahtt: I'm not even entirely sure we need this modal still :thinking: It was used to catch an edge case where a user might bypass the metrics screen. I don't recall how that might occur.

ryanml: I can remove this bullet to match the full page metrics component.

I was able to get the modal to appear like so:

1. Setup MetaMask, select import
2. On the metrics opt-in tab, navigate to `initialize/create-password/import-with-seed-phrase`
3. Enter seed phrase + new password
4. Logout once your on the home screen
5. Log back in
6. See the modal

Gudahtt: Ah cool, nice job! Probably best to keep it anyway as a safeguard.

ryanml: This bullet has been removed from the modal for consistency

ui/components/app/modals/metametrics-opt-in-modal/metametrics-opt-in-modal.component.js
```diff
@@ -27,76 +27,81 @@ export default class MetaMetricsOptInModal extends Component {
...
                 </div>
                 <div className="metametrics-opt-in__row">
                   <i className="fa fa-times" />
                   <div className="metametrics-opt-in__row-description">
-                    <span className="metametrics-opt-in__bold">Never</span>{' '}
-                    collect your full IP address
+                    {t('metametricsCommitmentsNeverCollectIP', [
+                      <span
+                        className="metametrics-opt-in__bold"
+                        key="neverCollectKeys"
```
Gudahtt, line 72: Nit: probably better to use a different `key` name here, just so it doesn't confuse readers

```suggestion
                        key="neverCollectIP"
```

ui/components/app/modals/metametrics-opt-in-modal/metametrics-opt-in-modal.component.js
```diff
@@ -27,76 +27,81 @@ export default class MetaMetricsOptInModal extends Component {
...
                 </div>
                 <div className="metametrics-opt-in__row">
                   <i className="fa fa-times" />
                   <div className="metametrics-opt-in__row-description">
-                    <span className="metametrics-opt-in__bold">Never</span> sell
-                    data for profit. Ever!
+                    {t('metametricsCommitmentsNeverSellDataForProfit', [
+                      <span
+                        className="metametrics-opt-in__bold"
+                        key="neverCollectKeys"
```
Gudahtt, line 85: Nit: probably better to use a different `key` name here, just so it doesn't confuse readers

```suggestion
                        key="neverSellData"
```

Gudahtt (dismissed):
LGTM!

ui/components/app/modals/metametrics-opt-in-modal/metametrics-opt-in-modal.component.js
ryanml: Updated both here and in the full page component

ui/components/app/modals/metametrics-opt-in-modal/metametrics-opt-in-modal.component.js
ryanml: Updated both here and in the full page component

Gudahtt (approved):
LGTM!

PR #11103, opened 2021-05-14
fix splitting of transaction nonce groups in state

### LINKED ISSUE #11069: MetaMask encountered an error
Message: Minified React error #130; visit https://reactjs.org/docs/error-decoder.html?invariant=130&args[]=undefined&args[]= for the full message or use the non-minified dev environment for full errors and additional helpful warnings.

Code: Error

Stack:

Error: Minified React error #130; visit https://reactjs.org/docs/error-decoder.html?invariant=130&args[]=undefined&args[]= for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
  at Mu (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/ui-libs.js:75:109949)
  at chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/ui-libs.js:75:57194
  at za (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/ui-libs.js:75:66624)
  at jl (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/ui-libs.js:75:104838)
  at bu (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/ui-libs.js:75:90053)
  at pu (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/ui-libs.js:75:89978)
  at iu (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/ui-libs.js:75:87326)
  at eu (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/ui-libs.js:75:84094)
  at zu (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/ui-libs.js:75:112010)
  at chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/ui-libs.js:75:113243
  at lu (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/ui-libs.js:75:87674)
  at Ku (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/ui-libs.js:75:113228)
  at render (chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/ui-libs.js:75:115151)
  at chrome-extension://nkbihfbeogaeaoehlefnkodbefgpgknn/ui.js:152:1356161

brad-decker:
Fixes: #11069 

Explanation:  
The tx-state-manager has always deleted the oldest "finalized" transaction from the state after the limit (default is 50), this is erroneous and can lead to UI issues when you split a cancel transaction away from the transaction that it was canceling and other various edge cases. 

This change makes sure we remove entire groups of transactions after we hit a limit. This will result in a number greater than 50 transactions being possible, if large groups of transactions occur. Grouping is done by nonce by network, but only the `${limit}` most recent unique nonces are considered.

### DIFF
--- a/app/scripts/controllers/transactions/tx-state-manager.js
+++ b/app/scripts/controllers/transactions/tx-state-manager.js
@@ -187,28 +187,43 @@ export default class TransactionStateManager extends EventEmitter {
     const transactions = this.getTransactions({
       filterToCurrentNetwork: false,
     });
-    const txCount = transactions.length;
     const { txHistoryLimit } = this;
 
     // checks if the length of the tx history is longer then desired persistence
     // limit and then if it is removes the oldest confirmed or rejected tx.
     // Pending or unapproved transactions will not be removed by this
-    // operation.
+    // operation. For safety of presenting a fully functional transaction UI
+    // representation, this function will not break apart transactions with the
+    // same nonce, per network. Not accounting for transactions of the same
+    // nonce and network combo can result in confusing or broken experiences
+    // in the UI.
     //
     // TODO: we are already limiting what we send to the UI, and in the future
     // we will send UI only collected groups of transactions *per page* so at
     // some point in the future, this persistence limit can be adjusted. When
     // we do that I think we should figure out a better storage solution for
     // transaction history entries.
-    if (txCount > txHistoryLimit - 1) {
-      const index = transactions.findIndex((metaTx) => {
-        return getFinalStates().includes(metaTx.status);
-      });
-      if (index !== -1) {
-        this._deleteTransaction(transactions[index].id);
-      }
-    }
+    const nonceNetworkSet = new Set();
+    const txsToDelete = transactions
+      .reverse()
+      .filter((tx) => {
+        const { nonce } = tx.txParams;
+        const { chainId, metamaskNetworkId, status } = tx;
+        const key = `${nonce}-${chainId ?? metamaskNetworkId}`;
+        if (nonceNetworkSet.has(key)) {
+          return false;
+        } else if (
+          nonceNetworkSet.size < txHistoryLimit - 1 ||
+          getFinalStates().includes(status) === false
+        ) {
+          nonceNetworkSet.add(key);
+          return false;
+        }
+        return true;
+      })
+      .map((tx) => tx.id);
 
+    this._deleteTransactions(txsToDelete);
     this._addTransactionsToState([txMeta]);
     return txMeta;
   }
@@ -608,6 +623,22 @@ export default class TransactionStateManager extends EventEmitter {
   _deleteTransaction(targetTransactionId) {
     const { transactions } = this.store.getState();
     delete transactions[targetTransactionId];
+    this.store.updateState({
+      transactions,
+    });
+  }
+
+  /**
+   * removes multiple transaction from state. This is not intended for external use.
+   *
+   * @private
+   * @param {number[]} targetTransactionIds - the transactions to delete
+   */
+  _deleteTransactions(targetTransactionIds) {
+    const { transactions } = this.store.getState();
+    targetTransactionIds.forEach((transactionId) => {
+      delete transactions[transactionId];
+    });
     this.store.updateState({
       transactions,
     });
--- a/app/scripts/controllers/transactions/tx-state-manager.test.js
+++ b/app/scripts/controllers/transactions/tx-state-manager.test.js
@@ -6,6 +6,8 @@ import {
 } from '../../../../shared/constants/transaction';
 import {
   KOVAN_CHAIN_ID,
+  MAINNET_CHAIN_ID,
+  RINKEBY_CHAIN_ID,
   KOVAN_NETWORK_ID,
 } from '../../../../shared/constants/network';
 import TxStateManager from './tx-state-manager';
@@ -636,7 +638,10 @@ describe('TransactionStateManager', function () {
       const limit = txStateManager.txHistoryLimit;
       // In this test case the earliest two transactions are a dropped attempted ether send and a
       // following cancel transaction with the same nonce. these two transactions should be dropped
-      // together as soon as the 11th unique nonce is attempted to be added, mocked here with limit + 1
+      // together as soon as the 11th unique nonce is attempted to be added. We use limit + 2 to
+      // first get into the state where we are over the "limit" of transactions because of a set
+      // of transactions with a unique nonce/network combo, then add an additional new transaction
+      // to trigger the removal of one group of nonces.
       const txs = generateTransactions(limit + 2, {
         chainId: currentChainId,
         to: VALID_ADDRESS,
@@ -653,15 +658,14 @@ describe('TransactionStateManager', function () {
       const result = txStateManager.getTransactions();
       assert.equal(result.length, limit, `limit of ${limit} is enforced`);
       assert.notEqual(result[0].id, 0, 'first tx should be removed');
-      assert.notEqual(
-        result[0].status,
-        TRANSACTION_STATUSES.DROPPED,
-        'first tx should not be dropped',
-      );
-      assert.notEqual(
-        result[1].type,
-        TRANSACTION_TYPES.CANCEL,
-        'Cancel tx should be dropped',
+      assert.equal(
+        result.some(
+          (tx) =>
+            tx.status === TRANSACTION_STATUSES.DROPPED ||
+            tx.status === TRANSACTION_TYPES.CANCEL,
+        ),
+        false,
+        'the cancel and dropped transactions should not be present in the result',
       );
     });
 
@@ -673,14 +677,14 @@ describe('TransactionStateManager', function () {
       // single transaction but the other shouldn't be dropped until adding the fifth additional
       // transaction
       const txs = generateTransactions(limit + 5, {
-        chainId: (i) => ([0, 1, 4, 5].includes(i) ? currentChainId : '0x1'),
+        chainId: (i) => {
+          if (i === 0 || i === 1) return MAINNET_CHAIN_ID;
+          else if (i === 4 || i === 5) return RINKEBY_CHAIN_ID;
+          return currentChainId;
+        },
         to: VALID_ADDRESS,
         from: VALID_ADDRESS_TWO,
-        nonce: (i) => {
-          if (i === 1) return '0';
-          else if (i === 5) return '4';
-          return `${i}`;
-        },
+        nonce: (i) => ([0, 1, 4, 5].includes(i) ? '0' : `${i}`),
         status: (i) =>
           i === 0 || i === 4
             ? TRANSACTION_STATUSES.DROPPED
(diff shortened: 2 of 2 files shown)


app/scripts/controllers/transactions/tx-state-manager.js
```diff
@@ -187,28 +187,43 @@ export default class TransactionStateManager extends EventEmitter {
...
     // transaction history entries.
-    if (txCount > txHistoryLimit - 1) {
-      const index = transactions.findIndex((metaTx) => {
-        return getFinalStates().includes(metaTx.status);
-      });
-      if (index !== -1) {
-        this._deleteTransaction(transactions[index].id);
-      }
-    }
+    const nonceNetworkSet = new Set();
```
danjm, line 206: I want to make sure I understand this.

So the goal is to ensure that if there are two or more transactions with the same nonce in the transaction list, that they don't get deleted.

So we iterate through the transactions. If a transaction is either counted as below `txHistoryLimit - 1` or is not in a final state, then it is NOT deleted. If a transaction is counted higher than `txHistoryLimit - 1` or is in a final state BUT has the same nonce as a transaction already flagged as one that shouldn't be deleted, then it is NOT deleted. If a transaction is above the txHistoryLimit count or is in a final state and does not match the nonce of a transaction already identified as one that shouldn't be deleted, then it is deleted...

So something about this isn't adding up for me... probably me misunderstanding, but I want to check.

Suppose that there are only 10 transactions and they are all "final states". Wouldn't they all be deleted according to this logic?

Should the `if (txCount > txHistoryLimit - 1) {` that was removed above wrap this tx deletion code?

And this is in a function that only adds a single tx. Before it only removed a single tx. Why now attempt to remove multiple within this function?

danjm (changes requested):

app/scripts/controllers/transactions/tx-state-manager.js
danjm: maybe I forget how `filter` works...

danjm: maybe not

app/scripts/controllers/transactions/tx-state-manager.test.js
```diff
@@ -614,6 +631,76 @@ describe('TransactionStateManager', function () {
...
       assert.equal(result[1].id, 2, 'early txs truncated');
     });
+
+    it('cuts off entire groups of transactions by nonce when adding new transaction', function () {
+      const limit = txStateManager.txHistoryLimit;
+      // In this test case the earliest two transactions are a dropped attempted ether send and a
+      // following cancel transaction with the same nonce. these two transactions should be dropped
+      // together as soon as the 11th unique nonce is attempted to be added, mocked here with limit + 1
+      const txs = generateTransactions(limit + 1, {
+        chainId: currentChainId,
```
danjm, line 641: should the `nonce` method param be defined here, to ensure more than once transaction of the same nonce is in the test?

app/scripts/controllers/transactions/tx-state-manager.js
danjm: Based on my reading of the tests I think I did misunderstand the goal though

danjm: Can the purpose of this code change be stated as:

If the addition of a transaction brings the transaction count above the txHistoryLimit then all transactions that share a nonce with the oldest "finalized" transaction should be deleted.

?

Gudahtt: >So the goal is to ensure that if there are two or more transactions with the same nonce in the transaction list, that they don't get deleted.

It's to ensure that one transaction in a group isn't deleted without deleting the entire group.

We can't let _all_ groups of more than 1 transaction persist indefinitely - then we'd have no limit on storage. But we can't break groups apart either without breaking the UI.

So the solution we went with here was to interpret the transaction history limit of 50 as meaning 50 _groups of transactions_. If there are over 50 groups, we delete any groups beyond that 50 (which should always be a single group if this is the only means of adding transactions, but might be more than 1 group if transactions are added elsewhere).

brad-decker: when we return false from this method we are *Not* including it, when we return true we *are* including it. So when return true we are saying "this is an item we want to delete". if there are only ten transactions non will be deleted because they will never hit the limit and thus will all be added to the nonceSet

brad-decker: basically yes as stated that will work

app/scripts/controllers/transactions/tx-state-manager.js
```diff
@@ -187,28 +187,43 @@ export default class TransactionStateManager extends EventEmitter {
...
+      .reverse()
+      .filter((tx) => {
+        const { nonce } = tx.txParams;
+        const { chainId, metamaskNetworkId, status } = tx;
+        const key = `${nonce}-${chainId ?? metamaskNetworkId}`;
+        if (nonceNetworkSet.has(key)) {
+          return false;
+        } else if (
+          nonceNetworkSet.size < txHistoryLimit - 1 ||
+          getFinalStates().includes(status) === false
```
Gudahtt, line 217: I'm not sure about this :thinking: This dictates that a transaction in its final state won't count towards the limit. But it should count - I don't see why it wouldn't.

app/scripts/controllers/transactions/tx-state-manager.test.js
brad-decker: Yes! must have dropped during code cleanup

app/scripts/controllers/transactions/tx-state-manager.js
brad-decker: its transactions NOT in their final state, AKA unapproved/pending txs

Gudahtt: Yes, but that's still wrong isn't it?

e.g. if we have 500 transactions with unique nonces that are all completed, this would skip over all of them, removing nothing.

Gudahtt: Oh nevermind, my bad. it's OR not and. I get it now.

app/scripts/controllers/transactions/tx-state-manager.test.js
brad-decker: fixed

danjm:
This looks good to me. One other non-blocking improvement I could suggest: add another test case for when there are transactions with the same nonce, but the number of generated transactions is less than the limit, and verify that entire groups of transactions are not cut off.

danjm (dismissed):

brad-decker:
@danjm added and also updated the test above it to be more specific because the last checks could have applied to either the first group of transactions or the second and the test was specifically checking that the first group was cut off and the second was not.

app/scripts/controllers/transactions/tx-state-manager.test.js
```diff
@@ -540,72 +573,56 @@ describe('TransactionStateManager', function () {
...
-            from: VALID_ADDRESS,
-          },
-        };
-        txStateManager.addTransaction(tx);
-      }
+      const txs = generateTransactions(limit + 1, {
+        chainId: currentChainId,
+        to: VALID_ADDRESS,
+        from: VALID_ADDRESS_TWO,
+        status: TRANSACTION_STATUSES.CONFIRMED,
```
Gudahtt, line 594: Given the description of this test and that the previous one is identical aside from the status, I think the intention was to ensure that it worked the same with the status was `REJECTED` as it did when it was `CONFIRMED`

```suggestion
        status: TRANSACTION_STATUSES.REJECTED,
```

app/scripts/controllers/transactions/tx-state-manager.test.js
```diff
@@ -614,6 +631,136 @@ describe('TransactionStateManager', function () {
       );
       assert.equal(result[1].id, 2, 'early txs truncated');
     });
+
+    it('cuts off entire groups of transactions by nonce when adding new transaction', function () {
+      const limit = txStateManager.txHistoryLimit;
+      // In this test case the earliest two transactions are a dropped attempted ether send and a
+      // following cancel transaction with the same nonce. these two transactions should be dropped
+      // together as soon as the 11th unique nonce is attempted to be added, mocked here with limit + 1
+      const txs = generateTransactions(limit + 2, {
```
Gudahtt, line 640: Shouldn't this be `limit + 1`? :thinking: That's what the description implies. And if I'm understanding this correctly, `limit + 1` is when the behavior we're testing is executed.

app/scripts/controllers/transactions/tx-state-manager.test.js
```diff
@@ -614,6 +631,136 @@ describe('TransactionStateManager', function () {
...
+      // different network. The first two transactions should be dropped after adding even another
+      // single transaction but the other shouldn't be dropped until adding the fifth additional
+      // transaction
+      const txs = generateTransactions(limit + 5, {
+        chainId: (i) => ([0, 1, 4, 5].includes(i) ? currentChainId : '0x1'),
+        to: VALID_ADDRESS,
+        from: VALID_ADDRESS_TWO,
+        nonce: (i) => {
+          if (i === 1) return '0';
+          else if (i === 5) return '4';
```
Gudahtt, line 681: In order to test that we're doing the network grouping correctly, wouldn't we want the two groups on different networks to share the same nonce? That way we can ensure we aren't considering them as one group.

As currently implemented, I think this test would pass if we were grouping by nonce alone.

app/scripts/controllers/transactions/tx-state-manager.test.js
brad-decker: the text is confusing and probably needs updated, but limit applies to unique nonces now, and so one of the + 1s is to get us into the state where we have a group of transactions at the beginning of the array that are together, the second + 1 is to push us over the limit of unique nonces.

brad-decker: at the very least the limit + 1 should be removed :P

app/scripts/controllers/transactions/tx-state-manager.test.js
brad-decker: yeah that's a better test. so I should flip the logic so that the [0, 1, 4, 5].includes(i) check returns the same nonce, and this if else returns different networks. I'll do that patch soon

brad-decker: that actually matches what I wrote in the description I think? 😕

app/scripts/controllers/transactions/tx-state-manager.test.js
Gudahtt: Ah right, gotcha. Makes sense.

app/scripts/controllers/transactions/tx-state-manager.test.js
```diff
@@ -614,6 +631,136 @@ describe('TransactionStateManager', function () {
...
+      assert.notEqual(result[0].id, 0, 'first tx should be removed');
+      assert.notEqual(
+        result[0].status,
+        TRANSACTION_STATUSES.DROPPED,
+        'first tx should not be dropped',
+      );
+      assert.notEqual(
+        result[1].type,
+        TRANSACTION_TYPES.CANCEL,
+        'Cancel tx should be dropped',
```
Gudahtt, line 664: It might be better to assert that the dropped and cancel transactions are _nowhere_ in the list of transactions. Just to ensure they're really gone.

If we had accidentally removed just the first transaction, this assertion would still pass because the cancel transaction isn't at index 1 anymore.

app/scripts/controllers/transactions/tx-state-manager.test.js
Gudahtt: Yeah I think it does!

app/scripts/controllers/transactions/tx-state-manager.test.js
```diff
@@ -614,6 +631,136 @@ describe('TransactionStateManager', function () {
...
+        '0x4',
+        'the second tx should be from the second group',
+      );
+    });
+
+    it('does not cut off entire groups of transactions when adding new transaction when under limit', function () {
+      // In this test case the earliest two transactions are a dropped attempted ether send and a
+      // following cancel transaction with the same nonce. Then, a bit later the same scenario on a
+      // different network. None of these should be dropped because we haven't yet reached the limit
+      const txs = generateTransactions(9, {
```
Gudahtt, line 729: Nit: it would probably be better to express this in terms of the limit. (e.g. `limit - 1`), so that it's more clear why this number was chosen, and so that it continues to work if the limit changes.

brad-decker: done in latest

app/scripts/controllers/transactions/tx-state-manager.test.js
brad-decker: done here and in the nonce+network test

app/scripts/controllers/transactions/tx-state-manager.test.js
brad-decker: done

app/scripts/controllers/transactions/tx-state-manager.test.js
brad-decker: modifed the explainer text

Gudahtt (approved):
LGTM!

danjm (approved):

PR #11107, opened 2021-05-14
Migration to remove erroneous tx state

brad-decker:
requires: #11103

### DIFF
--- a/app/scripts/migrations/059.js
+++ b/app/scripts/migrations/059.js
@@ -31,7 +31,7 @@ function transformState(state) {
     const nonceNetworkGroupedObject = groupBy(
       Object.values(transactions),
       (tx) => {
-        return `${tx.txParams.nonce}-${tx.chainId ?? tx.metamaskNetworkId}`;
+        return `${tx.txParams?.nonce}-${tx.chainId ?? tx.metamaskNetworkId}`;
       },
     );
 
--- a/app/scripts/migrations/059.js
+++ b/app/scripts/migrations/059.js
@@ -1,0 +1,52 @@
+import {
+  cloneDeep,
+  concat,
+  groupBy,
+  keyBy,
+  pickBy,
+  isPlainObject,
+} from 'lodash';
+import { TRANSACTION_TYPES } from '../../../shared/constants/transaction';
+
+const version = 59;
+
+/**
+ * Removes orphaned cancel and retry transactions that no longer have the
+ * original transaction in state, which results in bugs.
+ */
+export default {
+  version,
+  async migrate(originalVersionedData) {
+    const versionedData = cloneDeep(originalVersionedData);
+    versionedData.meta.version = version;
+    const state = versionedData.data;
+    versionedData.data = transformState(state);
+    return versionedData;
+  },
+};
+
+function transformState(state) {
+  const transactions = state?.TransactionController?.transactions;
+  if (isPlainObject(transactions)) {
+    const nonceNetworkGroupedObject = groupBy(
+      Object.values(transactions),
+      (tx) => {
+        return `${tx.txParams.nonce}-${tx.chainId ?? tx.metamaskNetworkId}`;
+      },
+    );
+
+    const withoutOrphans = pickBy(nonceNetworkGroupedObject, (group) => {
+      return group.some(
+        (tx) =>
+          tx.type !== TRANSACTION_TYPES.CANCEL &&
+          tx.type !== TRANSACTION_TYPES.RETRY,
+      );
+    });
+    state.TransactionController.transactions = keyBy(
+      concat(...Object.values(withoutOrphans)),
+      (tx) => tx.id,
+    );
+  }
+
+  return state;
+}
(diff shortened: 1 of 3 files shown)


moymoy08:
https://www.contributor-covenant.org/version/2/0/code_of_conduct.html

moymoy08:
https://www.contributor-covenant.org/version/2/0/code_of_conduct.html

moymoy08:
https://www.contributor-covenant.org/version/2/0/code_of_conduct.html

app/scripts/migrations/059.js
```diff
@@ -0,0 +1,50 @@
...
+    versionedData.data = transformState(state);
+    return versionedData;
+  },
+};
+
+function transformState(state) {
+  const transactions = state?.TransactionController?.transactions;
+  if (
+    typeof transactions === 'object' &&
+    Array.isArray(transactions) === false &&
```
danjm, line 25: Why the need to verify that it is not an array? And is any update to state needed when it is an array?

danjm: Also, lodash can help simplify this:

Screenshot from 2021-05-17 11-44-09

brad-decker: If transactions are an array here it'll mean that the state is corrupt because in a previous migration we migrated the transaction state store to an id indexed object. So if its an array this point the user has bigger problems. I added these checks to make the code as verbose as possible. There are also corresponding tests to see what would happen in the case where it isn't in the proper state.

app/scripts/migrations/059.test.js
```diff
@@ -0,0 +1,384 @@
...
+import {
+  KOVAN_CHAIN_ID,
+  MAINNET_CHAIN_ID,
+  RINKEBY_CHAIN_ID,
+  GOERLI_CHAIN_ID,
+} from '../../../shared/constants/network';
+import { TRANSACTION_TYPES } from '../../../shared/constants/transaction';
+import migration59 from './059';
+
+const ERRONEOUS_TRANSACTION_STATE = {
```
danjm, line 12: Can we add a case to these tests that where there is a transaction with a 'cancel' type, but it shares a nonce with a non-cancel transaction, to verify that it won't be removed?

brad-decker: there is one. line 192

```js
 it('should not drop cancel transactions with matching non cancel or retry in same network and nonce', async function () {
    const oldStorage = {
      meta: {},
      data: {
        TransactionController: {
          transactions: {
            ...ERRONEOUS_TRANSACTION_STATE,
            11: {
              ...ERRONEOUS_TRANSACTION_STATE['0'],
              id: 11,
              type: TRANSACTION_TYPES.SENT_ETHER,
            },
          },
        },
        foo: 'bar',
      },
    };

    const newStorage = await migration59.migrate(oldStorage);
    assert.deepEqual(newStorage.data, {
      TransactionController: {
        transactions: oldStorage.data.TransactionController.transactions,
      },
      foo: 'bar',
    });
  });
```

app/scripts/migrations/059.js
brad-decker: added isPlainObject in latest

Gudahtt (commented):
This doesn't seem to account for pending transactions :thinking: 

I'm not entirely sure what we should do for those though. I was thinking that we could exclude them from the migration altogether, but what about the case where there's a pending "CANCEL" with no base transaction from which to infer the original intent?

brad-decker:
> I'm not entirely sure what we should do for those though. I was thinking that we could exclude them from the migration altogether, but what about the case where there's a pending "CANCEL" with no base transaction from which to infer the original intent?

if they have no base transaction to infer from they need to be removed because they are part of the erroneous state... the only alternative would be to change them to ether transaction which would potentially confuse the user. This does seem to be an edge case though because when creating the cancel transaction the original should be in pending status (and immune from excision), and the status of the original should only change if it is A) confirmed (in which case the cancel will be dropped) of B) dropped (when the cancel in confirmed)...

Gudahtt:
Ah, right good point. OK, in that case it seems fair to treat them the same as finished transactions.

I just thought of another potential complication: Unapproved transactions. They typically don't have a nonce yet, so they'll be in a nonce group with `undefined` in it. Which.... is probably fine? They can't possibly be cancel/retry so they won't get filtered. But it might be worth adding a test at least to confirm they don't blow anything up.

brad-decker:
I changed one transaction in the ERRONEOUS_TRANSACTION_STATE to be an unapproved, nonce-less tx. I actually just deleted the txParams object entirely which did result in an error so I added optional chaining on the txParams.nonce in the migration.

brad-decker:
any hints on the benchmark fail?

Gudahtt:
That's an intermittent failure specific to chromedriver. It happens sporadically on the chrome e2e tests as well, except we have a retry for those, so we don't see it bubble up as often.

Gudahtt (approved):
LGTM!

PR #11111, opened 2021-05-16
Avoid showing  "Gas price extremely low" warning in advanced tab for testnets

### LINKED ISSUE #11076: "Gas price extremely low" warning showing on speed ups on test nets
- Add a custom network (e.g. https://consensys.slack.com/archives/C01CZRZ9H0T/p1618512321118000)
- send a transaciton on that network
- try to speed up the transaction
- notice that the "Gas price extremely low" warning shows no matter what price is entered

Instead, the "Gas price extremely low" warning should only show if the gas price is below the recommended price for that network

https://user-images.githubusercontent.com/7499938/118056350-5b9c3f00-b364-11eb-82d4-ed48f6e24206.mp4

NiranjanaBinoy:
Fixes: #11076 

Explanation:  
Problem: While speeding up or canceling a transaction on testnets, the gasprice on the advanced tab is showing "Gas price extremely low" warning no matter what price is entered.

Fix: On testnets, we don't use the gas prices API to get slow, medium, and fast estimates. Instead, we just get a single estimate via eth_gasprice. So the isCustomPriceSafe() will be called only for Mainnet and e2e test and a default `true` value will be sent for testnets.

Manual testing steps:  

- Add a custom network (e.g. https://consensys.slack.com/archives/C01CZRZ9H0T/p1618512321118000)
- Send a transaction on that network
- Try to speed up the transaction, check the advanced input tab and try adjusting the gasPrice.

### DIFF
--- a/ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.container.js
+++ b/ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.container.js
@@ -142,7 +142,8 @@ const mapStateToProps = (state, ownProps) => {
     customGasLimit: calcCustomGasLimit(customModalGasLimitInHex),
     customGasTotal,
     newTotalFiat,
-    customPriceIsSafe: isCustomPriceSafe(state),
+    customPriceIsSafe:
+      isMainnet || process.env.IN_TEST ? isCustomPriceSafe(state) : true,
     customPriceIsExcessive: isCustomPriceExcessive(state),
     maxModeOn,
     gasPriceButtonGroupProps: {
--- a/ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.container.js
+++ b/ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.container.js
@@ -38,6 +38,7 @@ import {
   getSendMaxModeState,
   getAveragePriceEstimateInHexWEI,
   isCustomPriceExcessive,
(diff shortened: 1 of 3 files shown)


Gudahtt (dismissed):
LGTM! This is at least a solid improvement.

I think there remains an issue when the gas estimation API is offline. If we don't have any value cached for the safe low price, this problem will continue to affect users on mainnet I think?

I haven't thought this through, so I don't know how we might fix that issue. If a solution isn't clear, perhaps we can open a separate issue for it, and merge this for now.

NiranjanaBinoy:
> LGTM! This is at least a solid improvement.
> 
> I think there remains an issue when the gas estimation API is offline. If we don't have any value cached for the safe low price, this problem will continue to affect users on mainnet I think?
> 
> I haven't thought this through, so I don't know how we might fix that issue. If a solution isn't clear, perhaps we can open a separate issue for it, and merge this for now.

Gudahtt:
Ah, great idea! I hadn't seen this `basicEstimateStatus`.

Though, I think we'll need to use a different selector. This one will return `false` if the gas estimates are still loading for example, wrongly displaying this warning until they finish. It'll also show the warning if the gas estimates failed to load completely.

Maybe we should only show the warning if `basicEstimateStatus` is `READY`, and the `estimateSource` is `MetaSwaps`? (or _not_ `eth_gasprice`, alternatively)

Gudahtt (approved):
LGTM!

PR #11115, opened 2021-05-17
Hide basic tab in advanced gas modal for speedup and cancel when on testnets

### LINKED ISSUE #11114: Hide basic tab in advanced gas modal when on testnets for speedup and cancel
On testnets, we don't use the gas prices API to get slow, medium, and fast estimates. Instead, we just get a single estimate via eth_gasprice. In such cases, when the gas edit modal is opened, we should not show the basic tab in the advanced gas modal.

NiranjanaBinoy:
Fixes: #11114 

Since on testnets, we don't use the gas prices API to get slow, medium, and fast estimates. Instead, we just get a single estimate via eth_gasprice. 

This PR will hide the basic modal of the advanced gas modal when opened as `sidebar` during speedup and cancel of a transaction.

### DIFF
--- a/ui/components/app/sidebars/sidebar.component.js
+++ b/ui/components/app/sidebars/sidebar.component.js
@@ -30,12 +30,16 @@ export default class Sidebar extends Component {
 
   renderSidebarContent() {
     const { type, sidebarProps = {} } = this.props;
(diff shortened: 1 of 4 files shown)


brad-decker (approved):
Tested locally, works well. LGTM

PR #11116, opened 2021-05-17
fix metaRPCClientFactory id handling

shanejonas:
Explanation:  
these changes affect `null` and `undefined` and `0` handling for `metaRPC` to reflect the [JSON-RPC Spec](https://www.jsonrpc.org/specification)

Rel: https://github.com/MetaMask/json-rpc-middleware-stream/pull/18

### DIFF
--- a/app/scripts/lib/metaRPCClientFactory.js
+++ b/app/scripts/lib/metaRPCClientFactory.js
@@ -6,6 +6,7 @@ class MetaRPCClient {
   constructor(connectionStream) {
     this.connectionStream = connectionStream;
     this.notificationChannel = new SafeEventEmitter();
+    this.uncaughtErrorChannel = new SafeEventEmitter();
     this.requests = new Map();
     this.connectionStream.on('data', this.handleResponse.bind(this));
     this.connectionStream.on('end', this.close.bind(this));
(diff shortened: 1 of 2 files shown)


app/scripts/lib/metaRPCClientFactory.js
```diff
@@ -17,34 +18,46 @@ class MetaRPCClient {
...
   }
 
+  onUncaughtError(handler) {
+    this.uncaughtErrorChannel.addListener('error', (error) => {
+      handler(error);
+    });
+  }
+
   close() {
     this.notificationChannel.removeAllListeners();
```
shanejonas, line 28:
```suggestion
    this.notificationChannel.removeAllListeners();
    this.uncaughtErrorChannel.removeAllListeners();
```

rekmarks: I endorse this change, lol

app/scripts/lib/metaRPCClientFactory.js
```diff
@@ -17,34 +18,46 @@ class MetaRPCClient {
     });
   }
 
+  onUncaughtError(handler) {
+    this.uncaughtErrorChannel.addListener('error', (error) => {
+      handler(error);
+    });
+  }
```
rekmarks, line 25: We should use this new method, probably by adding a listener next to [this](https://github.com/MetaMask/metamask-extension/blob/09d81ac/ui/index.js/#L131-L141) in `ui/index.js`.

That should be done in a follow-up, though.

rekmarks (approved):
LGTM!

PR #11117, opened 2021-05-17
use safer toChecksum utility

brad-decker:
Fixes: the previous version of ethereumjs-util that we relied upon would handle values other than prefixed hex strings without failing. The new version throws an error. This new utility replaces our downstream direct usage of the toChecksumAddress method in favor of the more safe wrapped method that handles non prefixed hex strings. This also replaces one thin wrapper around the method that was used throughout the frontend code.

### DIFF
--- a/ui/helpers/utils/icon-factory.js
+++ b/ui/helpers/utils/icon-factory.js
@@ -1,7 +1,8 @@
 import contractMap from '@metamask/contract-metadata';
-import { isHexString, addHexPrefix } from 'ethereumjs-util';
-import { isValidHexAddress } from '../../../shared/modules/hexstring-utils';
-import { checksumAddress } from './util';
+import {
+  isValidHexAddress,
+  toChecksumHexAddress,
+} from '../../../shared/modules/hexstring-utils';
 
 let iconFactory;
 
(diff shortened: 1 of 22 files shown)


jsconfig.json
```diff
@@ -1,14 +1,3 @@
 {
-  "exclude": [
```
Gudahtt, line 2: What prompted this? :eyes:

ui/helpers/utils/icon-factory.js
```diff
@@ -18,11 +19,7 @@ function IconFactory(jazzicon) {
 }
 
 IconFactory.prototype.iconForAddress = function (address, diameter) {
-  let addr = address;
-
-  if (isHexString(addHexPrefix(address))) {
-    addr = checksumAddress(addHexPrefix(address));
-  }
+  const addr = toChecksumHexAddress(address);
```
Gudahtt, line 22: I was about to comment that this was a change in behaviour, because the hex prefix always gets added. But apparently we were doing this in v9.4 as well! So this is good as-is.

jsconfig.json
brad-decker: i keep removing all of this stuff cause its killing autocomplete. I'll remove it for this pr and figure out a proper fix.

brad-decker: removed

Gudahtt (approved):
LGTM!

PR #11118, opened 2021-05-17
Removing support survey notification from What's New

ryanml:
## Test Plan
1. Ensure from a fresh install that the support notification is not shown
2. Ensure the support notification is removed after an upgrade where it would otherwise be shown

### DIFF
--- a/app/scripts/migrations/060.js
+++ b/app/scripts/migrations/060.js
@@ -1,4 +1,4 @@
-import { cloneDeep } from 'lodash';
+import { cloneDeep, isPlainObject } from 'lodash';
 
 const version = 60;
 const SUPPORT_NOTIFICATION_KEY = 2;
@@ -20,11 +20,16 @@ export default {
 };
 
 function transformState(state) {
-  if (
-    state?.NotificationController?.notifications[SUPPORT_NOTIFICATION_KEY]
-      .date === SUPPORT_NOTIFICATION_DATE
-  ) {
-    delete state.NotificationController.notifications[SUPPORT_NOTIFICATION_KEY];
+  const notifications = state?.NotificationController?.notifications;
+  if (isPlainObject(notifications)) {
+    if (
+      notifications[SUPPORT_NOTIFICATION_KEY]?.date ===
+      SUPPORT_NOTIFICATION_DATE
+    ) {
+      delete state.NotificationController.notifications[
+        SUPPORT_NOTIFICATION_KEY
+      ];
+    }
   }
   return state;
 }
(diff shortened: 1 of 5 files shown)


ryanml (commented):
I didn't re-index the translation keys as I didn't want to cause an issue with translations already in progress

Gudahtt (commented):
If we're going to leave the translation keys as-is, perhaps it would be simpler to leave the notification IDs as-is, and remove #2? Rather than shift them all by one. There is no requirement for them to be consecutive.

Also, we should remove the translations that we're no longer using.

ryanml:
No longer re-indexing notification ids/keys. Good thought to keep it simple. Also, removed the now unused translations

app/scripts/migrations/060.test.js
```diff
@@ -0,0 +1,75 @@
...
+
+    migration60
+      .migrate(oldStorage)
+      .then((newStorage) => {
+        assert.deepEqual(newStorage.meta, {
+          version: 60,
+        });
+        done();
+      })
+      .catch(done);
```
Gudahtt, line 21: Nit: any reason not to use `async/await` here? We have been for the migration tests for a long while now

app/scripts/migrations/060.js
```diff
@@ -0,0 +1,30 @@
...
+    const state = versionedData.data;
+    const newState = transformState(state);
+    versionedData.data = newState;
+    return versionedData;
+  },
+};
+
+function transformState(state) {
+  if (
+    state?.NotificationController?.notifications[SUPPORT_NOTIFICATION_KEY]
```
Gudahtt, line 24: Nit: it would be great to have tests for all branches, such as the case where `NotificationController` is nullish.

Probably best to add a few more `?.` operators here as well, in case `notifications` is nullish, and in case `notifications[SUPPORT_NOTIFICATION_KEY]` is nullish. So that this won't crash if the state is invalid.

Gudahtt (dismissed):
LGTM! Left a few suggestions on making tests more robust.

danjm (dismissed):

app/scripts/migrations/060.js
ryanml: `transformState` updated to accommodate instances where `NotificationController` or `notifications` are non-existent. Test cases added to cover additional branches

app/scripts/migrations/060.test.js
ryanml: No reason in particular, this has been update to use `async/await`

danjm (approved):

PR #11124, opened 2021-05-18
Swaps: Create a new swap

dan437:
## Explanation
As a user, I want to submit multiple swaps in a row without clicking on multiple buttons to reset the flow.

We've also changed the button to `Close` after swapping is completed, which redirects a user to the main page.

## Manual testing steps

Make another swap:
- Make a swap
- On the success page click on `Create a new swap`
- The `Build Quote` page is displayed again

Close button:
- Make a swap
- Click on the `Close` button
- The main page is displayed

## Screenshots
Success page with the `Create a new swap` link and `Close` button:
image


ui/pages/swaps/awaiting-swap/awaiting-swap.js
```diff
@@ -262,12 +285,8 @@ export default function AwaitingSwap({
             );
           } else if (errorKey) {
             await dispatch(navigateBackToBuildQuote(history));
-          } else if (
-            isSwapsDefaultTokenSymbol(destinationTokenInfo?.symbol, chainId)
-          ) {
-            history.push(DEFAULT_ROUTE);
           } else {
-            history.push(`${ASSET_ROUTE}/${destinationTokenInfo?.address}`);
```
danjm, line 270: Why is this line deleted?

ui/pages/swaps/awaiting-swap/awaiting-swap.js
```diff
@@ -230,6 +236,22 @@ export default function AwaitingSwap({
     );
   }
 
+  const MakeAnotherSwap = () => {
+    return (
+      <Box marginBottom={3}>
+        <a
```
darkwing, line 242: I think it would be best to use a `<button>` here

danjm (changes requested):
Code looks good, but I left a question about a functionality change.

ui/pages/swaps/awaiting-swap/awaiting-swap.js
dan437: Thanks for the suggestion! We had a quick chat about when to use a button or a link and since we do page redirection on click here, a link seems fine.

ui/pages/swaps/awaiting-swap/awaiting-swap.js
dan437: Great question! We synced up yesterday and here is a summary: the requirement is to show the `Close` button after a swap is done and that will redirect to the main page. However, before the swap is done, a user can click on `View in activity`: 
image

Which redirects to the Activity tab, where a user can see progress of their transaction.

dan437: An update.. I was redirecting to the Activity tab on the main page. I changed it back to the Activity tab on an asset's page, so it looks like this now: 
image

danjm (approved):
LGTM!

PR #11134, opened 2021-05-19
bump allow scripts

EtDu:
- Fixes https://app.circleci.com/pipelines/github/MetaMask/metamask-extension/9600/workflows/0dbdedd0-7bb8-45fa-9a29-7357d5fd8907/jobs/280180

### DIFF
--- a/package.json
+++ b/package.json
@@ -209,7 +209,7 @@
     "@babel/preset-env": "^7.5.5",
     "@babel/preset-react": "^7.0.0",
     "@babel/register": "^7.5.5",
-    "@lavamoat/allow-scripts": "^1.0.5",
+    "@lavamoat/allow-scripts": "^1.0.6",
     "@metamask/auto-changelog": "^1.0.0",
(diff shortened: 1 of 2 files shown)


Gudahtt:
So the only change is that there is now a trailing newline - is that correct?

kumavis (approved):

kumavis:
@Gudahtt  correct - it now appends a newline when generating the policy https://npmfs.com/compare/@lavamoat/allow-scripts/1.0.5/1.0.6/

EtDu:
For the sake of our linter. Maybe we broke someone else's... 🤷

Gudahtt:
Doubtful. It's a requirement of the POSIX standard, for one thing. The npm CLI will always include a trailing newline in newly generated `package.json` files as well, and they typically dictate the format used by any tooling.

Though this might still be broken on Windows, for any projects that are setup to use `\r\n`. Ideally `allow-scripts` would preserve the pre-existing formatting.

Gudahtt (approved):

PR #11140, opened 2021-05-20
@metamask/auto-changelog@2.0.1

rekmarks:
Adds the new `@metamask/auto-changelog`. The big difference is that PR links (and numbers) are returned to the end of the line for all change entries. Since that's what we're now doing, I moved the PR links to the end of each line for all entries. There were some completely malformed lines in there that are also fixed.

Oh, and it's also more robust and less buggy!


ryanml (approved):
++

kumavis (approved):
i dig it

PR #11145, opened 2021-05-20
Improve UI + content for price difference notifications

dan437:
## Explanation
The goal of this PR is to improve UI + content for price difference notifications, so users will understand it better.

## Manual testing steps
TBD (I've tested it via mocking specific code, which showed notifications)

## Screenshots / videos
`Medium` price diff notification:
image

`High` price diff notification:
image

Unknown price notification:
image

Tooltip shared for medium and high price diff notifications and unknown price notifications:
image


app/_locales/en/messages.json
```diff
@@ -873,6 +873,9 @@
   "history": {
     "message": "History"
   },
+  "iUnderstand": {
```
darkwing, line 876: Rather than have the key mimic the text, we should key off of what they represent instead.

dan437: I like that!

darkwing:
This is awesome @dan437 !  Well done!

darkwing (approved):

PR #11153, opened 2021-05-21
Capture exception instead of throw error in useTransactionDisplayData

danjm:
Ensures the user experience doesn't break if a user has an unrecognized type in their transaction history.

### DIFF
--- a/ui/hooks/useTransactionDisplayData.js
+++ b/ui/hooks/useTransactionDisplayData.js
@@ -1,4 +1,5 @@
 import { useSelector } from 'react-redux';
+import { captureException } from '@sentry/browser';
 import { getKnownMethodData } from '../selectors/selectors';
 import {
   getStatusKey,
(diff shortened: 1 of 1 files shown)


ryanml (commented):
Seems reasonable to me! Curious, was there any reason it was important that an error was being explicitly thrown per: https://github.com/MetaMask/metamask-extension/pull/11104?

danjm:
> Seems reasonable to me! Curious, was there any reason it was important that an error was being explicitly thrown per: #11104?
I don't think so. Before this issue was silently failing within this function, which resulted in data being passed to `transaction-icon.js` that caused metamask to break. We explicitly added code to that component to capture when such an error occurs, but not to throw an error, so that the experience doesn't break. With 11104, we threw an error when really we only needed to capture  one.

ryanml (approved):

brad-decker:
What is displayed to the user when this happens? The values that the hook returns are not set when in this else block? Better than breaking the extension but just want to understand the outcome

danjm:
>What is displayed to the user when this happens? The values that the hook returns are not set when in this else block? Better than breaking the extension but just want to understand the outcome

I think the icon will be a grey circle, and the title on the transaction will be blank.

brad-decker (approved):

PR #11159, opened 2021-05-21
implement safer to buffer

brad-decker:
Fixes another sentry issue that has surfaced related to ethereumjs-util upgrade. This time the toBuffer method from ethereumjs-util is the issue. I have made a new buffer-utils file in shared/modules that is safe to use even with non hex prefixed strings, but I don't know that we always want to allow for that kind of input. There are two cases where I haven't replaced using toBuffer with this method, which are in the transaction controller and account import code. I feel like we'd want an error thrown in these cases if ever a non prefixed hex string is returned from either signing a transaction or importing an account.

### DIFF
--- a/shared/modules/buffer-utils.js
+++ b/shared/modules/buffer-utils.js
@@ -1,0 +1,16 @@
+import { toBuffer as ethUtilToBuffer, isHexString } from 'ethereumjs-util';
+
+/**
+ * Returns a buffer from the provided input, via ethereumjs-util.toBuffer but
+ * additionally handling non hex strings. This is a failsafe as in most cases
+ * we should be primarily dealing with hex prefixed strings with this utility
+ * but we do not want to break the extension for users.
+ * @param {import('ethereumjs-util').ToBufferInputTypes | string} input
+ * @returns {Buffer}
+ */
+export function toBuffer(input) {
+  if (typeof input === 'string' && isHexString(input) === false) {
+    return Buffer.from(input);
+  }
+  return ethUtilToBuffer(input);
+}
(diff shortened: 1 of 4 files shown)


jsconfig.json
```diff
@@ -7,7 +7,6 @@
     "docs",
     "lavamoat",
     "node:console",
-    "node_modules",
```
Gudahtt, line 10: :eyes:

brad-decker: everytime.

shared/modules/buffer-utils.js
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Returns a buffer from the provided input, via ethereumjs-util.toBuffer but
+ * additionally handling non hex strings. This is a failsafe as in most cases
+ * we should be primarily dealing with hex prefixed strings with this utility
+ * but we do not want to break the extension for users.
+ * @param {import('ethereumjs-util').ToBufferInputTypes | string} input
+ * @returns {Buffer}
+ */
+export function toBuffer(input) {
+  if (typeof input === 'string' && isHexString(input) === false) {
```
Gudahtt, line 12: What is the `typeof` check meant to catch?

brad-decker: this method will work with any input type that `ethereumjs.util` supports including BN, buffer, arrays of numbers, etc. But it only works with 0x prefixed hex strings, not plain strings. So this check is checking that if this is a string and not a hexstring then we just user Buffer.from

Gudahtt:
>There are two cases where I haven't replaced using toBuffer with this method, which are in the transaction controller and account import code. I feel like we'd want an error thrown in these cases if ever a non prefixed hex string is returned from either signing a transaction or importing an account.

Agreed - seems safer to leave those as-is for now.

shared/modules/buffer-utils.js
Gudahtt: Ah I see, makes sense. I didn't realize so many other types were supported.

Gudahtt (approved):
LGTM!

Issue #9243, opened 2020-08-15
Increase Token Symbol Length 


adrianmcli:
## What problem are you trying to solve?

**In brief:** The token symbol length requirement of 12 characters is too short, we should raise it to 20 or more if possible.

As expiring tokens become more popular, it is often desired to have the expiration date within the token symbol itself. For example, if I were to create a synthetic yielding token pegged to the USD while backed by ETH and expiring in September of 2020, I might want to use a token symbol of `xyUSDETH-SEP20` (14 characers).

However, some users in our community at UMA have voiced a desire for lexicographically sortable symbol names. And the best way of doing this is to use the ISO standard of [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) calendar date-inspired format of YYYY-MM-DD. A possible desired symbol might be `xyUSDETH-2020.09.01` (19 characters).

## Describe the solution you'd like

Increase the allowed token symbol length to 20 (or more). The specific length can be further discussed but I think 20 is a reasonable starting point.

## Additional context

This previously merged PR increased token symbol length from 10 to 12: https://github.com/MetaMask/metamask-extension/pull/5816

Implementation should be quite simple if we follow that PR. In that PR, the problem being addressed was a little different: https://github.com/MetaMask/metamask-extension/issues/5815

mc25:
Unless there is a technical reason for the 12 char limit. It makes absolute sense to open up the design space for more readable and usefull tickers for crypto teams.

chrismaree:
This would be pretty great! There are many projects that need longer symbol length and removing this barrier would be helpful in improving their UX.

Also, [EIP747](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-747.md) which lets you click to add tokens to Metamask is even more restrictive! This is limited to token symbol length of 6 chars. If metamask was updated to support longer symbol length then this should properly also be updated to take full advantage of these features.

rekmarks:
I'll note that the [ERC20](https://eips.ethereum.org/EIPS/eip-20) standard does not specify any limit on the symbol length.

There is of course some kind of practical limit to the symbol length, because it needs to be able to function as a ticker in the UI.

That said, we should be able to do something with this, and are addressing it internally.

claytonroche:
I would also be keen to see this happen.  I know most tickers should be short for ease of use, but given the variety of very specific-use tokens, I think having a little more information in the titles would be hugely useful.

danfinlay:
[...]

PR #11162, opened 2021-05-22
change erc20 token symbol length limit from 11 to 20

nymmrx:
**Fixes:** #9243

**Explanation:**  
The token symbol length requirement of 11 characters is too short, we should raise it to 20. 

Newer DeFi protocols, which often do _something_ with existing tokens, need effective ways to communicate that in the symbol. At Yearn, we preface our vault tokens with `yv-`, and add additional considerations for LP vault tokens. For example, our yVault that accepts tokens from Curve's Compound pool is `yvCurve-Compound` (16 chars), and tokens for a YFI-ETH Sushi LP vault would be `yvSushi-YFI-ETH` (15 chars). Many other protocols have compound token nomenclature such as this, so it makes sense to expand the acceptable limit.

As an example example Uniswaps's [tokenlist standard](https://github.com/Uniswap/token-lists/blob/b5865fbded944da8715e65fa2f3286efc6c18f87/src/tokenlist.schema.json#L180) defines an hard limit of 20 characters for a token symbol. 

This PR replaces the 11 characters limit with a 20 character limits, updates relevant i18n & test files and adds small tweaks for the UI to handle edge cases.

**Screenshots for the relevant UI tweaks:**  

<div>
<kbd >

</kbd>
<kbd>

</kbd>
<kbd>

</kbd>
</div>

### DIFF
--- a/app/_locales/am/messages.json
+++ b/app/_locales/am/messages.json
@@ -985,8 +985,8 @@
   "symbol": {
     "message": "ምልክት"
   },
-  "symbolBetweenZeroTwelve": {
-    "message": "ምልክቱ 11 ቁምፊዎች ወይም ከዚያ ያነሰ መሆን አለበት።"
+  "symbolBetweenZeroTwentyOne": {
+    "message": "ምልክቱ 20 ቁምፊዎች ወይም ከዚያ ያነሰ መሆን አለበት።"
   },
   "syncWithMobile": {
     "message": "ከሞባይል ጋር አሳምር"
--- a/app/_locales/ar/messages.json
+++ b/app/_locales/ar/messages.json
@@ -981,8 +981,8 @@
   "symbol": {
     "message": "رمز"
   },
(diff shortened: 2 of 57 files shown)


mcmire:
Hi, sorry for not getting back to you about this. We have brought this up internally a few times, and while we are generally for this change, we need to double-check each screen and make sure this doesn't break any of the UI. Also, there is a concern that if we truncate tokens we could make phishing campaigns easier. We also need to think about how this dovetails with token autodetection. We will post an update when we've made a decision.

brad-decker:
@nymmrx thank you for doing this work. I am closing this PR due to the staleness, but the branch is referenced from the ticket related to this work. If you want to take on the work you can rebase and resolve conflicts and reopen this PR and we will consider it and help push it across the finish line. I would check first if the issue, #15819 has assignees or even ask in that thread if work has begun before beginning the rebase work. Thank you again!

PR #11165, opened 2021-05-24
Add event property to event sent on tx:status-update in metamask-controller

danjm:
The metrics event sent when an onchain failure occurs, in the handler for the `tx:status-update` event in `metamask-controller.js`, was updated in https://github.com/MetaMask/metamask-extension/commit/3d4dfc74a8. However, no `event` property was added

No event name was added. This causes a `Must specify event and category.` event to be registered in sentry.

This PR corrects that.


dan437 (approved):
Looks good!

Gudahtt (approved):
LGTM!

PR #11166, opened 2021-05-24
Add stringified payload to metametrics controller trackEvent error message

danjm:
This adds the stringified payload to the error message thrown when their is no event or category passed to the `trackEvent` method  of the metametrics controller. This will make it easier to identify the source of the error.

### DIFF
--- a/app/scripts/controllers/metametrics.js
+++ b/app/scripts/controllers/metametrics.js
@@ -331,9 +331,15 @@ export default class MetaMetricsController {
     // event and category are required fields for all payloads
     if (!payload.event || !payload.category) {
       throw new Error(
-        `Must specify event and category. Passed payload was: ${JSON.stringify(
+        `Must specify event and category. Event was: ${
+          payload.event
+        }. Category was: ${payload.category}. Payload keys were: ${Object.keys(
           payload,
-        )}`,
+        )}. ${
+          typeof payload.properties === 'object'
+            ? `Payload property keys were: ${Object.keys(payload.properties)}`
+            : ''
+        }`,
       );
     }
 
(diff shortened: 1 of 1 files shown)


Gudahtt (dismissed):
LGTM!

app/scripts/controllers/metametrics.js
```diff
@@ -330,7 +330,11 @@ export default class MetaMetricsController {
   async trackEvent(payload, options) {
     // event and category are required fields for all payloads
     if (!payload.event || !payload.category) {
-      throw new Error('Must specify event and category.');
+      throw new Error(
+        `Must specify event and category. Passed payload was: ${JSON.stringify(
+          payload,
```
brad-decker, line 335: Do we track the user's MetaMetricsId in sentry context @Gudahtt? If so tracking the entire payload is problematic here for anonymity reasons. It's likely safe here to just include the event OR category and possibly an array of keys supplied in the payload versus the values.

brad-decker (changes requested):

app/scripts/controllers/metametrics.js
Gudahtt: I had the same thought, but it looks like the payload doesn't include the MetaMetricsId. That's in the options.

Gudahtt: Oh but the `properties` and `sensitiveProperties` are _a lot_. Especially the `sensitiveProperties` - those explicitly should not be included for privacy reasons. Good point - I agree with your recommendation.

brad-decker: its also in the *sentry* appState context. 
Screen Shot 2021-05-24 at 12 29 34 PM

danjm:
@brad-decker @Gudahtt Addressed above comments in d6f355f21

brad-decker (approved):

PR #11169, opened 2021-05-24
Disable Ledger Live toggle for Firefox users

darkwing:
During 9.6.0 QA @tmashuang had mentioned Firefox users could still toggle this setting and prevent Ledger U2F from working, despite there being a warning.  Instead it would be good to disable this toggle for Firefox users so that U2F is always used.

### DIFF
--- a/ui/components/ui/toggle-button/toggle-button.component.js
+++ b/ui/components/ui/toggle-button/toggle-button.component.js
@@ -47,15 +47,19 @@ const colors = {
 };
 
 const ToggleButton = (props) => {
-  const { value, onToggle, offLabel, onLabel } = props;
+  const { value, onToggle, offLabel, onLabel, disabled } = props;
 
   const modifier = value ? 'on' : 'off';
 
   return (
-    <div className={classnames('toggle-button', `toggle-button--${modifier}`)}>
+    <div
+      className={classnames('toggle-button', `toggle-button--${modifier}`, {
+        'toggle-button--disabled': disabled,
+      })}
+    >
       <ReactToggleButton
         value={value}
-        onToggle={onToggle}
+        onToggle={disabled ? undefined : onToggle}
         activeLabel=""
         inactiveLabel=""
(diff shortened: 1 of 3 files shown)


brad-decker (dismissed):
LGTM

Gudahtt:
Do we know why the Ledger Live connection doesn't work on Firefox? I didn't think it was doing anything Chrome-specific.

darkwing:
Firefox ignores the `window.open` call because we're calling it without a true click heuristic.  I tried a variety of methods to make this happen, including using `img src` and `iframe src`, as well as a `window.location` experiment, but none worked.  With WebUSB being the other option, and Firefox not supporting WebUSB, Firefox has boxed us in a bit.

Gudahtt:
Ahhhh, interesting, that makes sense. Thanks, now that you mention it I do recall this coming up once before.

The action _is_ triggered by a real user click in this case, so there might be some way for us to make this work :thinking: But that would be a pretty large development effort. I think we'd have to move the `window.open` call into the UI. Not something we can rush.

ui/components/ui/toggle-button/toggle-button.component.js
```diff
@@ -47,15 +47,22 @@ const colors = {
...
-    <div className={classnames('toggle-button', `toggle-button--${modifier}`)}>
+    <div
+      className={classnames('toggle-button', `toggle-button--${modifier}`, {
+        'toggle-button--disabled': disabled,
+      })}
+    >
       <ReactToggleButton
         value={value}
-        onToggle={onToggle}
+        onToggle={disabled ? noop : onToggle}
```
Gudahtt, line 65: Nit: we can pass in `undefined` here instead - the ReactToggleButton prop doesn't require us passing in a function.

Gudahtt (dismissed):
LGTM! Had one minor suggestion, but it doesn't really matter.

Gudahtt (approved):

PR #11170, opened 2021-05-24
Update `@metamask/auto-changelog` from v9.0.1 to v9.1.0

Gudahtt:
This update includes a bug fix that made v9.0.1 incompatible with valid entries for the `package.json` "repository" field. Specifically, that field required that the repository be the GitHub repo URL, but the field is meant to point at the _git_ repo URL (the difference between the two on GitHub is the `.git` suffix).

Now that that bug as been fixed, we can update the `repository` field to point at `https://github.com/MetaMask/metamask-extension.git`, which is what it should be.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "repository": {
     "type": "git",
-    "url": "https://github.com/MetaMask/metamask-extension"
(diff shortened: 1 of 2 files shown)


brad-decker (approved):
LGTM

PR #11179, opened 2021-05-25
Warn users when they attempt to add a network that is already configured

### LINKED ISSUE #10632: Warn users on custom network form when chain_id matches known one
As a user who wants to not get phished,

If I try to enter a custom network with a chainID that is equal to a known one (or one I already have entered), I should get a warning, something like:

- This chain ID is currently used by the _____ network. Transactions sent on this provider will be valid transactions on that network also. If you do not intend to be adding a new connection for that same network, do not add this provider.

adonesky1:
Fixes: #10632

Explanation:  Adds a warning when a user attempts to add a custom network with a chainId that matches a network they already have configured. Modifies the existing warning for a matching RPC Url to include the network that matches

This does not include warnings if a custom network is added via API.

Manual testing steps:  
  - Attempt to add a custom network with a chainID and/or RPC URL that matches that of a network that is already configured
  - Confirm that an warning message pops beneath the input fields for each of the matching entries

### DIFF
--- a/test/e2e/helpers.js
+++ b/test/e2e/helpers.js
@@ -27,6 +27,7 @@ async function withFixtures(options, testSuite) {
   } = options;
   const fixtureServer = new FixtureServer();
   const ganacheServer = new Ganache();
+  let secondaryGanacheServer;
   let dappServer;
   let segmentServer;
   let segmentStub;
@@ -34,6 +35,16 @@ async function withFixtures(options, testSuite) {
   let webDriver;
   try {
     await ganacheServer.start(ganacheOptions);
+    if (ganacheOptions?.concurrent) {
+      const { port, chainId } = ganacheOptions.concurrent;
+      secondaryGanacheServer = new Ganache();
+      await secondaryGanacheServer.start({
+        blockTime: 2,
+        _chainIdRpc: chainId,
+        port,
+        vmErrorsOnRPCResponse: false,
+      });
+    }
     await fixtureServer.start();
     await fixtureServer.loadState(path.join(__dirname, 'fixtures', fixtures));
     if (dapp) {
@@ -103,6 +114,9 @@ async function withFixtures(options, testSuite) {
   } finally {
     await fixtureServer.stop();
     await ganacheServer.quit();
+    if (ganacheOptions?.concurrent) {
+      secondaryGanacheServer.quit();
+    }
     if (webDriver) {
       await webDriver.quit();
     }
--- a/test/e2e/helpers.js
+++ b/test/e2e/helpers.js
@@ -115,7 +115,7 @@ async function withFixtures(options, testSuite) {
     await fixtureServer.stop();
     await ganacheServer.quit();
     if (ganacheOptions?.concurrent) {
-      secondaryGanacheServer.quit();
+      await secondaryGanacheServer.quit();
     }
     if (webDriver) {
       await webDriver.quit();
--- a/test/e2e/tests/custom-rpc-history.spec.js
+++ b/test/e2e/tests/custom-rpc-history.spec.js
@@ -12,10 +12,12 @@ describe('Stores custom RPC history', function () {
     ],
   };
   it(`creates first custom RPC entry`, async function () {
+    const port = 8546;
+    const chainId = 1338;
     await withFixtures(
       {
         fixtures: 'imported-account',
-        ganacheOptions,
+        ganacheOptions: { ...ganacheOptions, concurrent: { port, chainId } },
         title: this.test.title,
       },
       async ({ driver }) => {
@@ -23,9 +25,8 @@ describe('Stores custom RPC history', function () {
         await driver.fill('#password', 'correct horse battery staple');
         await driver.press('#password', driver.Key.ENTER);
 
-        const rpcUrl = 'https://rpc-mumbai.matic.today';
-        const chainId = '80001';
-        const networkName = 'Matic';
+        const rpcUrl = `http://127.0.0.1:${port}`;
+        const networkName = 'Secondary Ganache Testnet';
 
         await driver.clickElement('.network-display');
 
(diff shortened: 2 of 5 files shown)


test/e2e/tests/custom-rpc-history.spec.js
```diff
@@ -33,17 +34,65 @@ describe('Stores custom RPC history', function () {
...
         await chainIdInput.sendKeys(chainId);
 
         await driver.clickElement('.network-form__footer .btn-secondary');
-        await driver.findElement({ text: rpcUrl, tag: 'div' });
+        await driver.findElement({ text: networkName, tag: 'div' });
+      },
+    );
+  });
+
+  it('warns user when they enter url or chainId for an already configured network', async function () {
```
brad-decker, line 56: 💯 very nice.

ui/pages/settings/networks-tab/network-form/network-form.component.js
```diff
@@ -332,11 +333,24 @@ export default class NetworkForm extends PureComponent {
   };
 
   validateChainIdOnChange = (chainIdArg = '') => {
+    const { networksToRender } = this.props;
     const chainId = chainIdArg.trim();
     let errorMessage = '';
     let radix = 10;
+    const hexChainId =
+      chainId === '0x' ? chainId : `0x${decimalToHex(chainId)}`;
```
brad-decker, line 341: should this be chainId.startsWith('0x') ?

ui/pages/settings/networks-tab/network-form/network-form.component.js
```diff
@@ -433,23 +447,29 @@ export default class NetworkForm extends PureComponent {
...
-    const isValidUrl = validUrl.isWebUri(url) && url !== '';
+    const isValidUrl = validUrl.isWebUri(url);
     const chainIdFetchFailed = this.hasError(
       'chainId',
       t('failedToFetchChainId'),
     );
+    const [matchingRPCUrl] = networksToRender.filter((e) => e.rpcUrl === url);
 
-    if (!isValidUrl) {
+    if (!isValidUrl && url !== '') {
```
brad-decker, line 459: What does this change do for the error state of the field?

brad-decker (dismissed):
Looks pretty good. Great improvement. Mostly just questions, nothing blocking.

ui/pages/settings/networks-tab/network-form/network-form.component.js
adonesky1: so when url is an empty string, `validUrl.isWebUri(url)` is `undefined`. And so `const isValidUrl = validUrl.isWebUri(url) && url !== ''` would equal `undefined` when url is an empty string and `!isValidUrl` would equal `true`, causing an incorrect validation error "Invalid RPC URL" to show when the string is empty.

test/e2e/tests/custom-rpc-history.spec.js
```diff
@@ -23,8 +23,9 @@ describe('Stores custom RPC history', function () {
         await driver.fill('#password', 'correct horse battery staple');
         await driver.press('#password', driver.Key.ENTER);
 
-        const rpcUrl = 'http://127.0.0.1:8545/1';
-        const chainId = '0x539'; // Ganache default, decimal 1337
+        const rpcUrl = 'https://rpc-mumbai.matic.today';
```
danjm, line 26: Why was this url change needed? This would make our e2e tests dependent on network conditions, which we try to avoid. There might be a way to mock the behaviour this change is aiming to take advantage of.

danjm (changes requested):
Looks good except for the one issue with e2e tests that I commented on

test/e2e/tests/custom-rpc-history.spec.js
adonesky1: Yeah I figured that could be an issue. The reason I changed it was because in this test flow the ganache network is already a configured network and with the changes to validation in this PR it won't allow you to add the same network twice so this test wasn't able to validate the adding of network. I suppose we could delete and re-add the ganache network instead.

adonesky1: @PeterYinusa and @danjm the prepopulated ganache network here doesn't appear to be tied to the state of the network controller in the fixtures json... is there a way to not have the ganache server present in the configured RPCs in the test env?

adonesky1: I suppose a larger question that this raises is should this validation be (as the title of the PR currently suggests) just a warning and not an error that disables the ability to save duplicate networks (as the PR's actually currently functioning)?

adonesky1: @danjm ^

danjm: > I suppose a larger question that this raises is should this validation be (as the title of the PR currently suggests) just a warning and not an error that disables the ability to save duplicate networks (as the PR's actually currently functioning)?

I don't know about this. We should ask for opinions on this. Although, I think the way this PR is currently implemented is probably fine.

danjm: Regarding how test having a different network.

ganache-cli allows this flag:

`--chainId: Specify the Chain ID ganache-cli will use for eth_chainId RPC and the CHAINID opcode. For legacy reasons, the default is currently 1337 for eth_chainId RPC and 1 for the CHAINID opcode. Setting this flag will align the chainId values. This will be fixed in the next major version of ganache-cli and ganache-core!`

We could maybe add the option to our test helpers to run two ganache servers at the same time, with different chain ids.

Gudahtt: There might be a simpler option: update the text fixture to remove ganache. It's no longer a "special" baked-in network - it's just present by default, and can be removed (IIRC).

test/e2e/helpers.js
```diff
@@ -103,6 +114,9 @@ async function withFixtures(options, testSuite) {
   } finally {
     await fixtureServer.stop();
     await ganacheServer.quit();
+    if (ganacheOptions?.concurrent) {
+      secondaryGanacheServer.quit();
```
Gudahtt, line 118: There is an `await` missing here

```suggestion
      await secondaryGanacheServer.quit();
```

SirJosh1987: Good?

danjm (approved):
LGTM

test/e2e/tests/custom-rpc-history.spec.js
Gudahtt: We later realized that my suggestion here doesn't make sense, because we want the initial network to be a local one to avoid the test depending on the network.

PR #11182, opened 2021-05-25
Fixing ever-present scrollbar issue on Swaps view

ryanml:
Related: 

https://github.com/MetaMask/metamask-extension/pull/10988

**With "Always Show Scroll Bars" set, no overflow content**
Screen Shot 2021-05-25 at 12 22 47 PM

**With "Always Show Scroll Bars" set, with overflow content**
Screen Shot 2021-05-25 at 12 26 29 PM

### DIFF
--- a/ui/pages/swaps/index.scss
+++ b/ui/pages/swaps/index.scss
@@ -31,11 +31,7 @@
     height: 100%;
     width: 100%;
     overflow-x: hidden;
-    overflow-y: scroll;
-
-    &--scrollable {
-      overflow: auto;
-    }
+    overflow-y: auto;
 
     @media screen and (min-width: 576px) {
       width: 460px;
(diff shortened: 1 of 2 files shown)


ui/pages/swaps/index.scss
```diff
@@ -31,11 +31,7 @@
     height: 100%;
     width: 100%;
     overflow-x: hidden;
-    overflow-y: scroll;
-
-    &--scrollable {
```
ryanml, line 36: This was never being used

dan437 (dismissed):
Looks good!

dan437:
The main problem was resolved, which is great! One little thing, which we discussed.. if we can change `max-height` from 428px to 420px to show 1 scrollbar instead of 2 on the View Quote page, that will be awesome!

```
@media screen and (max-width: 576px)
.view-quote__content {
    overflow-y: auto;
    max-height: 420px;
}
```

ryanml:
@dan437 I've pushed that change, can you re-run through your visual tests once more? Thanks!

dan437 (approved):
I've retested it and it looks great! A scrollbar shows in Swaps only when it's needed and there are no 2 scrollbars on the same page.

danjm (approved):

Issue #11136, opened 2021-05-19
Refactor metamask-ui.spec.js to follow new pattern


danjm:
This is a ticket to cover the remaining work of https://github.com/MetaMask/metamask-extension/issues/10754

As that ticket describes we want to migrate our e2e tests so that they use fixtures to set up test state, and test specific pieces of functionality in isolation.

We can see an example of one such migration in this PR https://github.com/MetaMask/metamask-extension/pull/10907/files

There is one test left to migrate: metamask-ui.spec.js

We will need to migrate it in pieces. We started with https://github.com/MetaMask/metamask-extension/issues/11058

Now we need to migrate each remaining set of tests in the file piece by piece. Let's track our progress on that here.

david0xd:
As discussed with @danjm and @PeterYinusa I've split this into three tickets:

`Refactor custom token tests (add/approve | metamask-ui.spec.js)` https://github.com/MetaMask/metamask-extension/issues/13674
`Refactor custom token tests (send | metamask-ui.spec.js)` https://github.com/MetaMask/metamask-extension/issues/13676
`Refactor onboarding tests (metamask-ui.spec.js)` https://github.com/MetaMask/metamask-extension/issues/13677

ElvirCe:
We are blocked on [13677  ](https://github.com/MetaMask/metamask-extension/issues/13677) until we get more clarified situation with Onboarding V2.

PR #11183, opened 2021-05-25
refactor add/hide token tests to use fixture

PeterYinusa:
Related: #11136

Explanation:  

  - moved hide token and add existing token tests
  - updated tests to use fixtures


NiranjanaBinoy (approved):
LGTM

PR #11184, opened 2021-05-26
Supporting existing default networks in switchEthereumChain

ryanml:
Supports existing default networks:
Screen Shot 2021-05-25 at 6 54 42 PM

Continues to work with added custom networks
Screen Shot 2021-05-25 at 6 55 11 PM


app/scripts/lib/rpc-method-middleware/handlers/switch-ethereum-chain.js
```diff
@@ -75,7 +91,6 @@ async function switchEthereumChainHandler(
           origin,
           type: MESSAGE_TYPE.SWITCH_ETHEREUM_CHAIN,
           requestData: {
-            rpcUrl: existingNetwork.rpcUrl,
```
ryanml, line 78: Per discussion on Slack, we don't need this property

brad-decker (approved):
LGTM

PR #11194, opened 2021-05-27
[POC] - Testing @etherumjs/tx update with various keyrings. 

brad-decker:
This simply constructs a new @ethereumjs/common object set to the Rinkeby network and instantiates the transaction using new version of @ethereumjs/tx. This should be enough to test transactions on Rinkeby with the various keyring types to ensure it works. This may need rebased etc to get all the updates that have landed. :D

### DIFF
--- a/app/scripts/controllers/transactions/index.js
+++ b/app/scripts/controllers/transactions/index.js
@@ -1,7 +1,8 @@
 import EventEmitter from 'safe-event-emitter';
 import { ObservableStore } from '@metamask/obs-store';
 import { bufferToHex, keccak, toBuffer } from 'ethereumjs-util';
-import Transaction from 'ethereumjs-tx';
+import Common from '@ethereumjs/common';
+import { TransactionFactory } from '@ethereumjs/tx';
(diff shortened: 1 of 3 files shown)


app/scripts/controllers/transactions/index.js
```diff
@@ -558,20 +559,34 @@ export default class TransactionController extends EventEmitter {
...
+      chainId,
+    };
+    console.log('txParams', txParams);
     // sign tx
     const fromAddress = txParams.from;
-    const ethTx = new Transaction(txParams);
-    await this.signEthTx(ethTx, fromAddress);
+    const unsignedEthTx = TransactionFactory.fromTxData(txParams, {
+      common,
+      freeze: false,
```
brad-decker, line 577: This shouldn't be strictly necessary, please test with and without to see if it works. By default transaction objects are frozen using Object.freeze making them immutable @rickycodes @darkwing

brad-decker:
Oh so this is old and can be closed, #11288 was merged instead

PR #11200, opened 2021-05-27
Show custom tokens in Swaps, add a custom token in Swaps

dan437:
## Explanation
This PR solves a few things:
- Shows custom tokens in Swaps if a user added them on the main page in the `Assets` tab
- Adds a possibility to paste a contract address in Swaps and if a token is found, a user can import it
- Fixes a bug where we were showing a scrollbar in a dropdown for tokens, even if there was only 1 token
- Fixes a bug where we were pushing away a token icon in the `Swap From` dropdown if a token name was too long
- Removes a cursor from a heading

TODOs:
- Test a successful swap with a custom token added on the main page
- Test a successful swap with a custom token added in Swaps

## Manual testing steps
Show a custom token in Swaps:
- On the main page go into the `Assets` tab and click on the `Add Token` button
- Click on `Custom Token`
- Paste your contract address and click on `Next`
- Click on `Swap` on the main page
- Search for your custom token on the `Build Quote` page, you should see your custom token there

Add a custom token in Swaps:
- Click on `Swap` on the main page
- Click on `Select a token` for the `Select to` dropdown
- Paste a contract address of your custom token
- If a token is found, you will see one item in the list and it will have an `Import` button
- Click on the item in the list or its `Import` button
- A new modal window for importing a token is displayed
- Click on `Import`
- Your custom token is selected in the `Select to` dropdown
 
## Screenshots
Show a notification with a list of tokens about adding a token via a contract address:
image

Show a notification if no token is found about adding a token via a contract address:
image

Token found by a contract address:
image

Import token popup:
image

Token imported:
image

List of custom tokens in the `Assets` tab on the main page:
image

List of custom tokens when selecting tokens in Swaps:
image

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -52,6 +52,9 @@
   "addContact": {
     "message": "Add contact"
   },
+  "addCustomTokenByContractAddress": {
+    "message": "Manually add tokens not listed by copy and pasting a token contract address in the search above."
+  },
   "addEthereumChainConfirmationDescription": {
     "message": "This will allow this network to be used within MetaMask."
   },
@@ -409,6 +412,9 @@
   },
   "continueToWyre": {
     "message": "Continue to Wyre"
+  },
+  "contract": {
+    "message": "Contract"
   },
   "contractAddressError": {
     "message": "You are sending tokens to the token's contract address. This may result in the loss of these tokens."
@@ -892,6 +898,12 @@
   "importAccountText": {
     "message": "or $1",
     "description": "$1 represents the text from `importAccountLinkText` as a link"
+  },
+  "importTokenQuestion": {
+    "message": "Import token?"
+  },
+  "importTokenWarning": {
+    "message": "Add at your own risk. Anyone can create a token, including creating fake versions of existing tokens that claim to represent projects."
   },
   "importWallet": {
     "message": "Import wallet"
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -53,7 +53,8 @@
     "message": "Add contact"
   },
   "addCustomTokenByContractAddress": {
-    "message": "Manually add tokens not listed by copy and pasting a token contract address in the search above."
+    "message": "Can’t find a token? You can manually add any token by pasting its address. Token contract addresses can be found on $1.",
+    "description": "$1 is a blockchain explorer for a specific network, e.g. Etherscan for Ethereum"
   },
   "addEthereumChainConfirmationDescription": {
     "message": "This will allow this network to be used within MetaMask."
@@ -903,7 +904,7 @@
     "message": "Import token?"
   },
   "importTokenWarning": {
-    "message": "Add at your own risk. Anyone can create a token, including creating fake versions of existing tokens that claim to represent projects."
+    "message": "Anyone can create a token with any name, including fake versions of existing tokens. Add and trade at your own risk!"
   },
   "importWallet": {
     "message": "Import wallet"
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -2079,12 +2079,12 @@
     "message": "Swap $1 to $2",
     "description": "Used in the transaction display list to describe a swap. $1 and $2 are the symbols of tokens in involved in a swap."
   },
+  "swapTokenVerificationAddedManually": {
+    "message": "This token has been added manually."
+  },
   "swapTokenVerificationMessage": {
     "message": "Always confirm the token address on $1.",
     "description": "Points the user to Etherscan as a place they can verify information about a token. $1 is replaced with the translation for \"Etherscan\" followed by an info icon that shows more info on hover."
-  },
-  "swapTokenVerificationNoSource": {
-    "message": "This token has not been verified."
   },
   "swapTokenVerificationOnlyOneSource": {
     "message": "Only verified on 1 source."
@@ -2369,6 +2369,10 @@
   },
   "verifyThisTokenOn": {
     "message": "Verify this token on $1",
+    "description": "Points the user to etherscan as a place they can verify information about a token. $1 is replaced with the translation for \"etherscan\""
+  },
+  "verifyThisUnconfirmedTokenOn": {
+    "message": "Verify this token on $1 and make sure this is the token you want to trade.",
     "description": "Points the user to etherscan as a place they can verify information about a token. $1 is replaced with the translation for \"etherscan\""
   },
   "viewAccount": {
--- a/ui/hooks/useTokensToSearch.js
+++ b/ui/hooks/useTokensToSearch.js
@@ -2,7 +2,7 @@ import { useMemo } from 'react';
 import { useSelector } from 'react-redux';
 import contractMap from '@metamask/contract-metadata';
 import BigNumber from 'bignumber.js';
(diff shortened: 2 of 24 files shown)


ui/pages/swaps/dropdown-search-list/dropdown-search-list.js
```diff
@@ -81,13 +113,74 @@ export default function DropdownSearchList({
...
+
   return (
     <div
       className={classnames('dropdown-search-list', className)}
       onClick={onClickSelector}
       onKeyUp={onKeyUp}
       tabIndex="0"
     >
+      {tokenForImport && isImportTokenModalOpen && (
+        <Popover
```
dan437, line 145: I will abstract this modal window for importing a token to its own component and use that one here.

ui/hooks/useTokensToSearch.js
```diff
@@ -119,7 +119,7 @@ export function useTokensToSearch({ usersTokens = [], topTokens = {} }) {
       others: [],
     };
 
-    memoizedTokensToSearch.forEach((token) => {
+    [...memoizedTokensToSearch, ...memoizedUsersToken].forEach((token) => {
```
dan437, line 122: If you know about more elegant way of doing this, let me please know!

ui/hooks/useTokensToSearch.js
```diff
@@ -129,10 +129,14 @@ export function useTokensToSearch({ usersTokens = [], topTokens = {} }) {
       );
       if (
         isSwapsDefaultTokenSymbol(renderableDataToken.symbol, chainId) ||
-        (usersTokensAddressMap[token.address] &&
-          Number(renderableDataToken.balance ?? 0) !== 0)
```
dan437, line 133: I deleted this line, because I want to include tokens with 0 balance (e.g. a newly imported custom token).

ui/hooks/useTokensToSearch.js
```diff
@@ -129,10 +129,14 @@ export function useTokensToSearch({ usersTokens = [], topTokens = {} }) {
...
+        usersTokensAddressMap[token.address]
       ) {
-        tokensToSearchBuckets.owned.push(renderableDataToken);
+        const foundItem = tokensToSearchBuckets.owned.find((item) => {
+          return item.symbol === token.symbol;
+        });
+        if (!foundItem) {
+          tokensToSearchBuckets.owned.push(renderableDataToken);
```
dan437, line 138: This way I won't add a custom token that is already in the `tokensToSearchBuckets.owned` array

ui/pages/swaps/dropdown-search-list/dropdown-search-list.js
dan437: Done.

ui/hooks/useTokensToSearch.js
darkwing: This makes sense but from a product perspective, I'd be interested in knowing if Gal would prefer that the native currency always display first.  Just a thought worth exploring.

darkwing (commented):
Wow, this is really slick work, I love it!  A few comments included in the PR.  Also, I saw this:

PaddingHere

We probably want to add some padding before the button.  Additionally, if I click to Import that item, confirm, and then open the dropdown again, it shows me the import button again.  Maybe the information is cached?

app/_locales/en/messages.json
```diff
@@ -893,6 +899,12 @@
     "message": "or $1",
     "description": "$1 represents the text from `importAccountLinkText` as a link"
   },
+  "importTokenQuestion": {
+    "message": "Import token?"
+  },
+  "importTokenWarning": {
+    "message": "Add at your own risk. Anyone can create a token, including creating fake versions of existing tokens that claim to represent projects."
```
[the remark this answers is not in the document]
dan437: Good point! I've just added new content from a PO. We have tracking for how many people import a custom token, so if we see low numbers, we can make changes.

dan437:
@darkwing Thank you so much for a review! I've disabled adding a custom token in the `Swap from` field, so now it's only available in the `Swap to` field, where it makes the most sense. When it's imported in Swaps, we will only add it to a list of tokens in the `Assets` tab on the main page if swapping was successful.

ui/hooks/useTokensToSearch.js
danjm: However, there is no way to distinguish between a token that had balance at one point and has recently gone to 0, and a token that was just added using the "Add Token" flow. We could add something like a `hasThisTokenEverHadBalance` property to all custom tokens, but that might be over-optimizing at this point.

So perhaps this should be left as it is in this PR for now, but we need to make sure to flag the UX implications to Gal, and this may require improvement in the future.

ui/hooks/useTokensToSearch.js
danjm: Nit: It would be nice if there were no duplicates in the array being loop over. Instead of doing this check within the `forEach` loop, we could filter or do a lodash `uniqBy` or something like that on the array before `forEach` is called.

danjm: Ultimately, I am guessing that we can improve performance (and maybe clean up the code) by using objects instead of arrays to store this data, but that can be a future improvement after this PR.

ui/pages/swaps/searchable-item-list/list-item-search/list-item-search.component.js
```diff
@@ -18,17 +20,47 @@ export default function ListItemSearch({
   fuseSearchKeys,
   searchPlaceholderText,
   defaultToAll,
+  listContainerClassName,
 }) {
   const fuseRef = useRef();
   const [searchQuery, setSearchQuery] = useState('');
+  const isTokenToDropdown =
+    listContainerClassName === 'build-quote__open-to-dropdown';
```
danjm, line 28: It looks like `listContainerClassName` is only used as a check for whether the search should include a search for imported tokens. Instead of basing this logic on the classname of the parent, could we pass a boolean whose name makes the use of the property clear (`shouldSearchForImports` or something like that)

danjm (changes requested):
Looks good overall. Nice work! I have left a set of small-ish suggestions.

ui/hooks/useTokensToSearch.js
dan437: Thanks for all the details, I will check it with product and likely create an improvement user story for this.

ui/hooks/useTokensToSearch.js
dan437: I like the idea to filter it in advance via Lodash, will do that!

ui/hooks/useTokensToSearch.js
```diff
@@ -119,7 +119,12 @@ export function useTokensToSearch({ usersTokens = [], topTokens = {} }) {
       others: [],
     };
 
-    memoizedTokensToSearch.forEach((token) => {
+    const memoizedSwapsAndUserTokensWithoutDuplicities = uniqBy(
+      [...memoizedTokensToSearch, ...memoizedUsersToken],
+      'symbol',
```
danjm, line 124: I think we should use the `address` as the property to determine uniqueness, because different tokens can share symbols.

danjm (changes requested):
Only one more suggestion, and then I think this is good.

danjm (approved):

PR #11203, opened 2021-05-27
fix stream.push after EOF

shanejonas:
Fixes sentry issue: https://sentry.io/organizations/metamask/issues/2394407935/?environment=production&project=273505&query=is%3Aunresolved+release%3A9.5.4&sort=freq&statsPeriod=14d

Explanation: 
this is related to [return early from metaRPCHandler if stream has ended #11065
](https://github.com/MetaMask/metamask-extension/pull/11065). Its a duplex stream so we should check that the readable and writable streams have ended.


Gudahtt (commented):
I was about to approve this, but after testing it I was still able to reproduce the error. It happens fairly often - maybe 1/3 of the time - when I _remove_ the extension fully from Chrome and then add the unpacked extension back, open the background console, wait for it to finish loading, then click the browser action button. That first display of the UI triggers the bug.

Here's the error I saw (same message as the Sentry error):
stackt

The stack trace ends in the `extension-port-stream` library. So it's a _different stream_ that's causing this error - the one between the background and the contentscript. Not this stream between the UI and the background.

Gudahtt:
Welp. So we've discovered that the error I posted a screenshot of above was distinct, and unrelated to the one this was meant to fix.

darkwing (dismissed):
Thank you Shane!

Gudahtt:
Oddly I'm still seeing the exact error this was meant to prevent :thinking: 
metaRPCstack

Gudahtt:
I think all of the changes made so far are good, and have improved the situation, but we need one more check to ensure the parent stream isn't ended by the time we're writing to the substream. That's the last missing piece here I think. The check when creating the substream isn't sufficient because the connection is broken after that, but before the write causing the above error.

Gudahtt (changes requested):
I just realized something: `write` takes a callback. We're not passing in the callback. That's why we're not seeing the error!

ghost (dismissed):

darkwing:
@shanejonas Is this PR still relevant?

darkwing:
@shanejonas Is this PR still relevant?

shanejonas:
PR and marks comments are still relevant but ran into issues reproducing on my own machine

Gudahtt:
It looks like I was wrong about the fix involving just a write callback. That didn't help.

The streams we're writing to here are the _substreams_. They just pass the message along to the parent - they don't write anything internally, and they don't directly throw when something goes wrong. The parent is what emits an error. Except in this case the parent isn't emitting an error either because the `extension-port-stream` package doesn't listen for and propagate disconnect errors.

In other words, this is blocked by https://github.com/MetaMask/extension-port-stream/issues/22

brad-decker:
@shanejonas @Gudahtt It seems the next step here is to fix the issue in extension-port-stream package, which may be a long running task? Considering this PR is now over a year old should we track this as an issue and link to this PR for a future time when we have bandwidth?

PR #11205, opened 2021-05-28
Feature/use zeroAddress

rickycodes:
Noticed this the other day... We can actually get the burn address from `ethereumjs-util@7.0.10`. This PR updates `ethereumjs-util` to that version and uses `zeroAddress()`


darkwing (approved):
I compared the output of both of `zeroAddress` and can confirm it's the same!  Thank you!

PR #11207, opened 2021-05-28
Fix notification locale format

danjm:
Fixes an error hit by users of two variants of portugese, two variants of chinese, and one variant of spanish.


Gudahtt (approved):
LGTM!

PR #11210, opened 2021-05-28
add erc-721 token detection and flag to disable sending

### LINKED ISSUE #10625: Disable the ability to try to send NFTs on Extension
**Describe the bug**
Currently a user can add NFTs to the extension but cannot specify the token ID.  So when sending in-extension, the send fails, and will likely spend a % of the gas.

**Expected behavior**
Disable the send button, similar to swaps on multi-chain, or add full NFT functionality.

**Screenshots**
image

adonesky1:
Fixes: #10625

Explanation:  Adds detection of ERC-721 tokens and a flag to the token object in preference controller state indicating whether or not this token supports the ERC-721 interface. The UI then disables sending for tokens for which `isERC721 === true`.

Manual testing steps:  
  - [Add an ERC-721 token](https://etherscan.io/token/0x629a673a8242c2ac4b7b8c5d8735fbeac21a6205)
  - Check that the send button is disabled.
  - Add an ERC-20 token 
  - Make sure that the send button is enabled

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -2387,6 +2387,10 @@
     "message": "verify the network details",
     "description": "Serves as link text for the 'unrecognizedChain' key. This text will be embedded inside the translation for that key."
   },
+  "unsendableAsset": {
+    "message": "Sending collectible (ERC-721) tokens is not currently supported",
+    "description": "This is an error message we show the user if they attempt to send a collectible asset type, for which currently don't support sending"
+  },
   "updatedWithDate": {
     "message": "Updated $1"
   },
--- a/app/scripts/controllers/detect-tokens.test.js
+++ b/app/scripts/controllers/detect-tokens.test.js
@@ -11,7 +11,7 @@ import PreferencesController from './preferences';
 
 describe('DetectTokensController', function () {
   const sandbox = sinon.createSandbox();
-  let keyringMemStore, network, preferences;
+  let keyringMemStore, network, preferences, provider;
 
   const noop = () => undefined;
 
@@ -23,12 +23,16 @@ describe('DetectTokensController', function () {
     keyringMemStore = new ObservableStore({ isUnlocked: false });
     network = new NetworkController();
     network.setInfuraProjectId('foo');
-    preferences = new PreferencesController({ network });
+    network.initializeProvider(networkControllerProviderConfig);
+    provider = network.getProviderAndBlockTracker().provider;
+    preferences = new PreferencesController({ network, provider });
     preferences.setAddresses([
       '0x7e57e2',
       '0xbc86727e770de68b1060c91f6bb6945c73e10388',
     ]);
-    network.initializeProvider(networkControllerProviderConfig);
+    sandbox
+      .stub(preferences, '_detectIsERC721')
+      .returns(Promise.resolve(false));
   });
 
   after(function () {
@@ -125,6 +129,7 @@ describe('DetectTokensController', function () {
         address: existingTokenAddress.toLowerCase(),
         decimals: existingToken.decimals,
         symbol: existingToken.symbol,
+        isERC721: false,
       },
     ]);
   });
@@ -177,11 +182,13 @@ describe('DetectTokensController', function () {
         address: existingTokenAddress.toLowerCase(),
         decimals: existingToken.decimals,
         symbol: existingToken.symbol,
+        isERC721: false,
       },
       {
         address: tokenAddressToAdd.toLowerCase(),
         decimals: tokenToAdd.decimals,
         symbol: tokenToAdd.symbol,
+        isERC721: false,
       },
     ]);
   });
@@ -234,11 +241,13 @@ describe('DetectTokensController', function () {
         address: existingTokenAddress.toLowerCase(),
         decimals: existingToken.decimals,
         symbol: existingToken.symbol,
+        isERC721: false,
       },
       {
         address: tokenAddressToAdd.toLowerCase(),
         decimals: tokenToAdd.decimals,
         symbol: tokenToAdd.symbol,
+        isERC721: false,
       },
     ]);
   });
--- a/app/scripts/controllers/preferences.js
+++ b/app/scripts/controllers/preferences.js
@@ -506,11 +506,8 @@ export default class PreferencesController {
         let addressBookKey = rpcDetail.chainId;
         if (!addressBookKey) {
           // We need to find the networkId to determine what these addresses were keyed by
-          const provider = new ethers.providers.JsonRpcProvider(
-            rpcDetail.rpcUrl,
-          );
           try {
-            addressBookKey = await provider.send('net_version');
+            addressBookKey = await this.ethersProvider.send('net_version');
             assert(typeof addressBookKey === 'string');
           } catch (error) {
             log.debug(error);
--- a/app/scripts/controllers/preferences.js
+++ b/app/scripts/controllers/preferences.js
@@ -796,17 +796,28 @@ export default class PreferencesController {
     if (contractsMap[checksumAddress]?.erc721 === true) {
       return Promise.resolve(true);
     }
-    const tokenContract = await new ethers.Contract(
+    const tokenContract = await this._createEthersContract(
       tokenAddress,
       abiERC721,
       this.ethersProvider,
     );
+
     return await tokenContract
       .supportsInterface(ERC721METADATA_INTERFACE_ID)
       .catch((error) => {
+        console.log('error', error);
         log.debug(error);
         return false;
       });
+  }
+
+  async _createEthersContract(tokenAddress, abi, ethersProvider) {
+    const tokenContract = await new ethers.Contract(
+      tokenAddress,
+      abi,
+      ethersProvider,
+    );
+    return tokenContract;
   }
 
   /**
--- a/app/scripts/controllers/preferences.js
+++ b/app/scripts/controllers/preferences.js
@@ -421,12 +421,12 @@ export default class PreferencesController {
 
   async updateTokenType(tokenAddress) {
     const { tokens } = this.store.getState();
-    const previousIndex = tokens.findIndex((token) => {
+    const tokenIndex = tokens.findIndex((token) => {
       return token.address === tokenAddress;
     });
-    tokens[previousIndex].isERC721 = await this._detectIsERC721(tokenAddress);
+    tokens[tokenIndex].isERC721 = await this._detectIsERC721(tokenAddress);
     this.store.updateState({ tokens });
-    return Promise.resolve(tokens[previousIndex]);
+    return Promise.resolve(tokens[tokenIndex]);
   }
 
   /**
@@ -786,25 +786,22 @@ export default class PreferencesController {
    */
   async _detectIsERC721(tokenAddress) {
     const checksumAddress = toChecksumHexAddress(tokenAddress);
-    let isERC721;
     // if this token is already in our contract metadata map we don't need
     // to check against the contract
     if (contractsMap[checksumAddress]?.erc721 === true) {
-      isERC721 = Promise.resolve(true);
-    } else {
-      const tokenContract = await new ethers.Contract(
-        tokenAddress,
-        abiERC721,
-        this.ethersProvider,
-      );
-      isERC721 = await tokenContract
-        .supportsInterface(ERC721METADATA_INTERFACE_ID)
-        .catch((error) => {
-          log.debug(error);
-          return false;
-        });
+      return Promise.resolve(true);
     }
-    return isERC721;
+    const tokenContract = await new ethers.Contract(
+      tokenAddress,
+      abiERC721,
+      this.ethersProvider,
+    );
+    return await tokenContract
+      .supportsInterface(ERC721METADATA_INTERFACE_ID)
+      .catch((error) => {
+        log.debug(error);
+        return false;
+      });
   }
 
   /**
--- a/app/scripts/controllers/preferences.js
+++ b/app/scripts/controllers/preferences.js
@@ -2,13 +2,20 @@ import { strict as assert } from 'assert';
 import { ObservableStore } from '@metamask/obs-store';
 import { ethErrors } from 'eth-rpc-errors';
 import { normalize as normalizeAddress } from 'eth-sig-util';
-import ethers from 'ethers';
+import { ethers } from 'ethers';
 import log from 'loglevel';
+import abiERC721 from 'human-standard-collectible-abi';
+import contractsMap from '@metamask/contract-metadata';
 import { LISTED_CONTRACT_ADDRESSES } from '../../../shared/constants/tokens';
 import { NETWORK_TYPE_TO_ID_MAP } from '../../../shared/constants/network';
 import { isPrefixedFormattedHexString } from '../../../shared/modules/network.utils';
-import { isValidHexAddress } from '../../../shared/modules/hexstring-utils';
+import {
+  isValidHexAddress,
+  toChecksumHexAddress,
+} from '../../../shared/modules/hexstring-utils';
 import { NETWORK_EVENTS } from './network';
+
+const ERC721METADATA_INTERFACE_ID = '0x5b5e139f';
 
 export default class PreferencesController {
   /**
@@ -73,11 +80,18 @@ export default class PreferencesController {
     };
 
     this.network = opts.network;
+    this.ethersProvider = new ethers.providers.Web3Provider(opts.provider);
     this.store = new ObservableStore(initState);
     this.store.setMaxListeners(12);
     this.openPopup = opts.openPopup;
     this.migrateAddressBookState = opts.migrateAddressBookState;
-    this._subscribeToNetworkDidChange();
+
+    this.network.on(NETWORK_EVENTS.NETWORK_DID_CHANGE, () => {
+      const { tokens, hiddenTokens } = this._getTokenRelatedStates();
+      this.ethersProvider = new ethers.providers.Web3Provider(opts.provider);
+      this._updateAccountTokens(tokens, this.getAssetImages(), hiddenTokens);
+    });
+
     this._subscribeToInfuraAvailability();
 
     global.setPreference = (key, value) => {
@@ -393,6 +407,8 @@ export default class PreferencesController {
     });
     const previousIndex = tokens.indexOf(previousEntry);
 
+    newEntry.isERC721 = await this._detectIsERC721(newEntry.address);
+
     if (previousEntry) {
       tokens[previousIndex] = newEntry;
     } else {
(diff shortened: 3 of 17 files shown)


app/_locales/en/messages.json
```diff
@@ -2329,6 +2329,9 @@
     "message": "verify the network details",
     "description": "Serves as link text for the 'unrecognizedChain' key. This text will be embedded inside the translation for that key."
   },
+  "unsendableAsset": {
+    "message": "Sending collectible (erc-721) tokens is not currently supported"
```
Gudahtt, line 2333: Nit: "ERC" is usually capitalized, and this message could use a description to help translators understand what it's for.

app/scripts/controllers/preferences.js
```diff
@@ -708,6 +729,8 @@ export default class PreferencesController {
   _subscribeToNetworkDidChange() {
     this.network.on(NETWORK_EVENTS.NETWORK_DID_CHANGE, () => {
       const { tokens, hiddenTokens } = this._getTokenRelatedStates();
+      const { type, rpcUrl } = this.network.getProviderConfig();
+      this.provider = ethers.getDefaultProvider(rpcUrl || type);
```
Gudahtt, line 733: I'd recommend using the `provider` we use everywhere else in the background, just to ensure this behaves as expected. I don't know if there are any differences in behavior, but just something to consider. You can see an example of how that can be passed into `ethers` in the `swaps` controller.

ui/store/actions.js
```diff
@@ -1398,6 +1398,23 @@ export function addToken(
   };
 }
 
+export function updateTokenType(tokenAddress) {
+  return (dispatch) => {
+    dispatch(showLoadingIndication());
+    return new Promise((resolve, reject) => {
+      background.updateTokenType(tokenAddress, (err, tokens) => {
```
Gudahtt, line 1405: It would be better to use the `promisifiedBackground` for all new actions - that's the one we're switching everything to use.

app/scripts/controllers/preferences.js
adonesky1: I thought I'd tried this before but apparently not correctly. I am making this change now.

app/scripts/controllers/preferences.js
```diff
@@ -763,6 +778,38 @@ export default class PreferencesController {
...
+   *
+   */
+  async _detectIsERC721(tokenAddress) {
+    const checksumAddress = toChecksumHexAddress(tokenAddress);
+    let isERC721;
+    // if this token is already in our contract metadata map we don't need
+    // to check against the contract
+    if (
+      contractsMap[checksumAddress] &&
+      contractsMap[checksumAddress].erc721 === true
```
darkwing, line 794: I think this can be `if (contractsMap[checksumAddress]?.erc721 === true) {`

ui/pages/send/send-content/send-asset-row/send-asset-row.component.js
```diff
@@ -31,13 +34,37 @@ export default class SendAssetRow extends Component {
...
+  clearUnsendableAssetError = () => {
+    this.props.setUnsendableAssetError(false);
+    this.props.updateSendErrors({ unsendableAssetError: null });
+  };
+
+  selectToken = async (token) => {
+    if (token && token.isERC721 === undefined) {
+      const updatedToken = await this.props.updateTokenType(token.address);
+      if (updatedToken.isERC721) {
+        const error = `Sending of collectible token ${token.symbol} is not currently supported`;
```
darkwing, line 58: We should localize this :)

ui/pages/send/send-content/send-content.component.js
```diff
@@ -50,10 +59,13 @@ export default class SendContent extends Component {
         <div className="send-v2__form">
           {gasError && this.renderError(gasError)}
           {isEthGasPrice && this.renderWarning(ETH_GAS_PRICE_FETCH_WARNING_KEY)}
-          {error && this.renderError()}
+          {unsendableAssetError && this.renderError(UNSENDABLE_ASSET_ERROR_KEY)}
```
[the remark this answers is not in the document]
adonesky1: @darkwing this error will only occur if a user enters the send flow with an NFT token added before these changes were introduced, because the token will not show up as available to send if it has been added with these changes in place (whether it is an ERC721 token is detected upon add here). If you want to test this functionality and don't have any nft tokens previously added, you can comment out line 410 in Preference.js - `newEntry.isERC721 = await this._detectIsERC721(newEntry.address)` in addToken function. Then add an NFT token and then enter the send flow and attempt to select that token to send. Let me know if this makes sense and/or if you want to walk through it together?

ui/pages/send/send-content/send-asset-row/send-asset-row.component.js
adonesky1: actually this error specificity isn't really needed. Its just a backend error that is never shown to the user. I'm going to remove and add generic error string.

ui/pages/send/send-content/send-asset-row/send-asset-row.component.js
```diff
@@ -31,13 +34,41 @@ export default class SendAssetRow extends Component {
...
+    this.props.setUnsendableAssetError(false);
+    this.props.updateSendErrors({
+      unsendableAssetError: null,
+      gasLoadingError: null,
+    });
+  };
+
+  selectToken = async (token) => {
+    if (token && token.isERC721 === undefined) {
+      const updatedToken = await this.props.updateTokenType(token.address);
```
darkwing, line 59: I'm hoping for some feedback from others on this, but should we be doing this type of property/token update in a send component?  I feel like we may want to do this at a higher level, but I'd really appreciate feedback from others.

app/scripts/controllers/preferences.js
```diff
@@ -403,6 +419,16 @@ export default class PreferencesController {
     return Promise.resolve(tokens);
   }
 
+  async updateTokenType(tokenAddress) {
+    const { tokens } = this.store.getState();
+    const previousIndex = tokens.findIndex((token) => {
```
darkwing, line 424: Minor, but thoughts on using `tokenIndex` instead of `previousIndex`?  `previous` makes me feel like it's going to or has changed.

ui/pages/send/send-content/send-asset-row/send-asset-row.component.js
adonesky1: So per this [thread](https://consensys.slack.com/archives/G1L7H42BT/p1622590868094100) (albeit with little feedback) this was the simplest solution to allow for multiple attempts to update the tokenType in the case of network issues. Its not perfect but I'm not sure if there is a better place for it. Open to suggestions though!

app/scripts/controllers/preferences.js
```diff
@@ -763,6 +778,35 @@ export default class PreferencesController {
...
+   * @param {string} tokensAddress - the token contract address.
+   *
+   */
+  async _detectIsERC721(tokenAddress) {
+    const checksumAddress = toChecksumHexAddress(tokenAddress);
+    let isERC721;
+    // if this token is already in our contract metadata map we don't need
+    // to check against the contract
+    if (contractsMap[checksumAddress]?.erc721 === true) {
+      isERC721 = Promise.resolve(true);
```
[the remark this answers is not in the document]
adonesky1: Yup that makes sense :)

app/scripts/controllers/preferences.js
adonesky1: Also makes sense. I'll make this change.

ui/pages/send/send-content/send-asset-row/send-asset-row.component.js
adonesky1: @darkwing ^

app/_locales/en/messages.json
```diff
@@ -2384,6 +2384,10 @@
     "message": "verify the network details",
```
danjm, line 2384: :+1:

ui/helpers/constants/error-keys.js
```diff
@@ -5,3 +5,4 @@ export const TRANSACTION_NO_CONTRACT_ERROR_KEY = 'transactionErrorNoContract';
 export const ETH_GAS_PRICE_FETCH_WARNING_KEY = 'ethGasPriceFetchWarning';
```
danjm, line 5: :+1:

app/scripts/controllers/preferences.js
```diff
@@ -2,14 +2,21 @@ import { strict as assert } from 'assert';
 import { ObservableStore } from '@metamask/obs-store';
 import { ethErrors } from 'eth-rpc-errors';
 import { normalize as normalizeAddress } from 'eth-sig-util';
-import ethers from 'ethers';
+import { ethers } from 'ethers';
```
danjm, line 5: Why was this change in the ethers import needed? How does it work?

adonesky1: This appears to have been the only place where ethers wasn't imported with destructuring (before this PR).  I can't find anything in the docs where it is imported as a default. It was required to access the ethers.Contract() method on line 791. I've tested the existing use case of ethers in this component and it doesn't change functionality. Though I'm actually going to push a commit to remove the redunant instantiation of the provider in the updateRPC method. Tested this as well and no functionality change.

app/scripts/controllers/preferences.js
```diff
@@ -403,6 +419,16 @@ export default class PreferencesController {
     return Promise.resolve(tokens);
   }
 
+  async updateTokenType(tokenAddress) {
```
danjm, line 422: We should add some basic documentation for this method, as is done with other token related methods in this file.

app/scripts/controllers/preferences.js
```diff
@@ -480,11 +506,8 @@ export default class PreferencesController {
         let addressBookKey = rpcDetail.chainId;
         if (!addressBookKey) {
           // We need to find the networkId to determine what these addresses were keyed by
-          const provider = new ethers.providers.JsonRpcProvider(
```
danjm, line 483: This change is okay because networkcontroller.setrpc is called before `updateRpc`

app/scripts/controllers/preferences.js
danjm: And maybe a unit test for this method.

app/scripts/metamask-controller.js
```diff
@@ -132,11 +132,17 @@ export default class MetamaskController extends EventEmitter {
     this.networkController = new NetworkController(initState.NetworkController);
```
danjm, line 132: :+1:

app/scripts/controllers/detect-tokens.test.js
```diff
@@ -11,7 +11,7 @@ import PreferencesController from './preferences';
```
danjm, line 11: :+1:

package.json
```diff
@@ -153,6 +153,7 @@
     "fast-safe-stringify": "^2.0.7",
```
danjm, line 153: :+1:

ui/components/app/asset-list-item/asset-list-item.js
```diff
@@ -27,6 +27,7 @@ const AssetListItem = ({
   primary,
```
danjm, line 27: :+1:

ui/components/app/token-cell/token-cell.js
```diff
@@ -15,6 +15,7 @@ export default function TokenCell({
   string,
```
danjm, line 15: :+1:

danjm:
Hitting this when switching accounts:

Screenshot from 2021-06-18 17-04-25

danjm:
@adonesky1 Regarding the error posted above, it seems that there is a race condition whereby, when switching accounts, `tokens` and `tokensWithBalances` can reflect data in different accounts. To recreate, have two accounts with two different lists of tokens and switch between.

ui/components/app/token-list/token-list.js
```diff
@@ -46,8 +46,18 @@ export default function TokenList({ onTokenClick }) {
   return (
     <div>
       {tokensWithBalances.map((tokenData, index) => {
+        const { isERC721 } = tokens.find(
```
danjm, line 49: Maybe we should update `useTokenTracker` so that `isERC721` is a property on the items in the `tokensWithBalances` array? Then we don't need to search tokens for a matching address. This might make the error I found go away too.

ui/pages/send/send-content/send-asset-row/send-asset-row.component.js
```diff
@@ -22,6 +22,9 @@ export default class SendAssetRow extends Component {
     setSendToken: PropTypes.func.isRequired,
```
danjm, line 22: :+1:

ui/pages/send/send-content/send-asset-row/send-asset-row.container.js
```diff
@@ -6,7 +6,11 @@ import {
   getSendTokenAddress,
```
danjm, line 6: :+1:

ui/pages/send/send-content/send-content.component.js
```diff
@@ -17,6 +18,10 @@ export default class SendContent extends Component {
     t: PropTypes.func,
   };
 
+  state = {
```
danjm, line 21: Any reason to maintain this state here instead of in redux?

ui/pages/send/send-content/send-content.component.js
```diff
@@ -97,12 +109,11 @@ export default class SendContent extends Component {
     );
   }
 
-  renderError(gasError = '') {
+  renderError(error) {
     const { t } = this.context;
```
danjm, line 113: This makes sense

ui/store/actions.js
```diff
@@ -1234,6 +1234,21 @@ export function addToken(
   };
```
danjm, line 1234: :+1:

ui/components/app/token-list/token-list.js
adonesky1: So it seems that error can be resolved by switching away from destructuring to optional chaining to protect against an undefined value. But I'll look into using the useTokenTracker as a more elegant solution!

ui/pages/send/send-content/send-content.component.js
adonesky1: I guess only that this state is used only within this component and the setter can be passed just one component down so felt unnecessary to kick up to redux but if its the preferred pattern happy to move it over!

ui/components/app/token-list/token-list.js
adonesky1: @danjm looks like you read [this conversation](https://github.com/MetaMask/metamask-extension/pull/11210#discussion_r653040540). Do you think that we should move from updating token types when users attempt to send to a more pro-active pattern of updating when all tokens in the token list that don't already have the isERC721 flag already set?

adonesky1: > Then we don't need to search tokens for a matching address. 

I recognize that for a user with a lot of tokens this adds some overhead for just rendering their token list...

adonesky1: Oh wait I see. You're not recommending that we do the update here just add the isERC721 flag where available in the useTokenTracker hook. Disregard my previous comments.

adonesky1: @danjm I've opened a PR to add the isERC721 field to the eth-token-tracker class: https://github.com/MetaMask/eth-token-tracker/pull/62. Though I can't seem to get the unit tests to pass

adonesky1: @danjm just wanted to check if you've gotten a chance to read through this yet

danjm: So I agree that this is cleaner from the perspective of extension code. The only challenge with this is that `eth-token-tracker` wasn't really meant to support non-ERC20 tokens. It creates a contract object using https://www.npmjs.com/package/human-standard-token-abi, which is meant for ERC-20. The fact that it works with ERC-721 is sort of an accident, and part of the reason we have this incomplete "support" of ERC-721 in the extension.

For now, I'd rather not add details about ERC-721 at the level of this library. We will need to deal with this soon, when we add actual NFT support. But it might be best to first just band-aid something in the hook, and then we can deal with a broader refactor of eth-token-tracker at a time when we are ready to add proper ERC-721 support.

Sorry! I realize this means redoing your work. I'm open to alternative opinions on this.

adonesky1: @danjm no worries! I'll add the band-aid loop in the hook for now!

danjm:
Apart from the one outstanding thread/issue, all code changes look good!

app/scripts/controllers/preferences.test.js
```diff
@@ -100,6 +101,45 @@ describe('preferences controller', function () {
...
+            decimals: '0',
+          },
+        ],
+      });
+      const sandbox = sinon.createSandbox();
+      sandbox
+        .stub(preferencesController, '_detectIsERC721')
+        .callsFake(() => true);
+      // sandbox.stub(ethers, 'Contract');
+      // sinon.stub(ethers.Contract.prototype, 'supportsInterface').returns(true);
```
adonesky1, line 134: @danjm @brad-decker do either of you happen to know how to stub this method correctly? I've poured through all related content on stackoverflow but can't seem to figure it out.

adonesky1:
> Apart from the one outstanding thread/issue, all code changes look good!

addressed that issue/thread

app/scripts/controllers/preferences.test.js
danjm: so this is a method on the object/class returend by the call to

```
new ethers.Contract(
      tokenAddress,
      abiERC721,
      this.ethersProvider,
    )
```

danjm: So first I would just validate that `ethers.Contract` is called with the expected parameters.

danjm: but then I think you would need to stub `ethers.Contract`, have it return an object with a `supportsInterface` method, verify that method gets called with the expected param, and stub `supportsInterface` to return true

danjm: Also, you may need to stub the constructor. So maybe something like:
```
const supportsInterfaceStub = sandbox.stub().returns(true);
sandbox.stub(ethers.Contract, 'constructor')
    .callsFake(() => ({ supportsInterface: supportsInterfaceStub }))
```

Gudahtt: Please do not introduce any more usages of `sinon.sandbox()` :pray: I've been slowly trying to get rid of them everywhere. They just make sinon more difficult to use. They're only meant for obscure cases where you need to manage multiple pools of distinct stubs simultaneously. The default sandbox works for everything we do.

ui/hooks/useTokenTracker.js
```diff
@@ -23,11 +23,19 @@ export function useTokenTracker(
...
+          (t) => t.address === token.address,
+        );
+        return { ...token, isERC721: additionalTokenData?.isERC721 };
+      });
+      setTokensWithBalances(matchingTokensWithIsERC721Flag);
       setLoading(false);
       setError(null);
     },
-    [hideZeroBalanceTokens],
+    [hideZeroBalanceTokens, tokens],
```
danjm, line 38:
```suggestion
    [hideZeroBalanceTokens, memoizedTokens],
```

danjm: This way, the `updateBalances` callback will only be updated when `tokens` changes in the sense of deep equality.

danjm (dismissed):
One last suggested change, and then this is good.

ui/hooks/useTokenTracker.js
adonesky1: Makes sense! Adding now!

adonesky1: Pushed

danjm (approved):
LGTM

brad-decker (approved):
LGTM! *prepares rebase skills*

PR #11215, opened 2021-05-31
Sync `master` with `develop`

danjm:
This brings develop up-to-date with `v9.5.7`


app/manifest/_base.json
```diff
@@ -71,5 +71,6 @@
     "notifications"
   ],
   "short_name": "__MSG_appName__",
+  "version": "9.5.7",
```
Gudahtt, line 74: Note that we started tracking the version in `package.json` rather than the manifest, so this is in the wrong place.

This move in where the version is tracked is a part of v9.6.0, so it's not on `master` yet

danjm: I have pushed a commit to fix this.

Gudahtt (approved):
LGTM!

PR #11216, opened 2021-05-31
Version v9.5.8

danjm:
This version readds the changes that were originally brought in with v9.5.6 but reverted with v9.5.7

It includes the fix to the breaking change of v9.5.6 that was made in https://github.com/MetaMask/metamask-extension/pull/11207


Gudahtt (approved):
LGTM!

I didn't QA this, but the commits and the diff all look correct to me.

tmashuang (approved):

PR #11221, opened 2021-06-02
Sync master with develop

danjm:
This brings develop up-to-date with v9.5.8


Gudahtt (changes requested):
The changelog is not well formatted. The categories are in the wrong order (`Added` comes before `Fixed`), and the reference link definition at the bottom of the changelog is missing.

danjm:
> The changelog is not well formatted. The categories are in the wrong order (Added comes before Fixed), and the reference link definition at the bottom of the changelog is missing.
Just pushed a commit to fix this

Gudahtt (approved):
LGTM!

PR #11227, opened 2021-06-03
Add confirm-encryption-public-key component to Storybook

EtDu:
- Adds `confirm-encryption-public-key` component to Storybook. 
- Modifies the actual component use pull the account from props instead of state because it otherwise doesn't update on account change.


Alanling888 (dismissed):

SirJosh1987 (approved):

SirJosh1987 (dismissed):

kumavis (changes requested):
@EtDu looks like it needs a rebase to deconflict and get a dep update

PR #11229, opened 2021-06-03
Resolve ws dependency to version ^7.4.6 to fix security vulnerability

danjm:
This fixes the `ws` security vulnerabilty warning which is currently breaking our builds.

I did a full audit of each of the cases where `ws` is used in our dependency tree. It turns out that it does not affect our code at all, and is incidentally imported by our dependencies. A case by case break down of my analysis is here: https://gist.github.com/danjm/e03db8e5aae65797532c87f6ec3d230b

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,8 @@
     "analytics-node/axios": "^0.21.1",
     "ganache-core/lodash": "^4.17.21",
     "netmask": "^2.0.1",
-    "pull-ws": "^3.3.2"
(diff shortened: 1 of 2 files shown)


Gudahtt (approved):
LGTM! Great writeup

We should update `ethers` to v5.3.0 soon as well, which includes an update to the latest version of `ws`, to reduce our dependence upon this resolution. But that can be in a separate PR. For the other cases I couldn't see any straightforward upgrade path.

PR #11239, opened 2021-06-04
Patch gas price issue - add fallback gas price

### LINKED ISSUE #11226: Gas price issues: use basic estimate average as fallback for gas price
**Describe the bug**
On 06/02 users reported transactions with proposed gas prices of 0. We subsequently learned this was due to an [issue with Infura's gas estimation enpoint](https://consensys.slack.com/archives/G8RSKCNCD/p1622651069157600). 

This issue is only intermittently reproducible. But to try:
Attempt to make a swap on uniswap and check the proposed gas price in the confirmation screen.

**Expected behavior**
Gas prices should never be 0.

 - Seen on Chrome and Brave

adonesky1:
Fixes: #11226

Explanation:  Adds a fallback (metaswap average gas price estimation) gas price if the infura gas price estimation endpoint fails.

Manual testing steps:  
  - Requires either modifying data in dev tools because we can't reliably reproduce the issue where infura's api fails

### DIFF
--- a/ui/selectors/confirm-transaction.js
+++ b/ui/selectors/confirm-transaction.js
@@ -12,6 +12,7 @@ import {
 import { sumHexes } from '../helpers/utils/transactions.util';
 import { transactionMatchesNetwork } from '../../shared/modules/transaction.utils';
 import { getNativeCurrency } from '../ducks/metamask/metamask';
+import { getAveragePriceEstimateInHexWEI } from './custom-gas';
 import { getCurrentChainId, deprecatedGetCurrentNetworkId } from './selectors';
 
 const unapprovedTxsSelector = (state) => state.metamask.unapprovedTxs;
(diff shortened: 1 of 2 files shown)


ui/selectors/confirm-transaction.js
```diff
@@ -217,10 +218,18 @@ export const transactionFeeSelector = function (state, txData) {
...
   const nativeCurrency = getNativeCurrency(state);
+  const metaSwapAverageGasEstimate = getAveragePriceEstimateInHexWEI(state);
 
-  const {
-    txParams: { value = '0x0', gas: gasLimit = '0x0', gasPrice = '0x0' } = {},
+  const { txParams: { value = '0x0', gas: gasLimit = '0x0' } = {} } = txData;
+
+  // if the gas price from our infura endpoint is null or undefined
+  // use the metaswap average price estimation as a fallback
+  let {
```
darkwing, line 227: This change makes sense at face value but have we confirmed that there are no other places in the extension that Infura being down would effect, that don't use this selector?  I don't think I have the answer to this, but just wanted to make sure it was considered.

adonesky1: Good question. I'll take dig into this.

adonesky1: I'm not finding other places we can patch up right now and in any case I can't think of how it would cause any issues elsewhere to patch this by itself?

ui/selectors/confirm-transaction.js
```diff
@@ -217,10 +218,18 @@ export const transactionFeeSelector = function (state, txData) {
...
+    txParams: { gasPrice },
   } = txData;
+  if (!gasPrice) {
+    gasPrice = metaSwapAverageGasEstimate || '0x0';
+  }
```
NiranjanaBinoy, line 232: These changes looks good. 

Q: Please check the function `updateTxDataAndCalculate` in`confirm-transaction.duck` as it seems to be getting the `gasPrice` from infura as well. (This function is called from confirm-transaction.component>>componentDidUpdate>>setTransactionToConfirm).

adonesky1: Good catch! Thanks Niranjana. I'll add the fallback here!

NiranjanaBinoy (dismissed):
LGTM

NiranjanaBinoy (approved):
LGTM

PR #10965, opened 2021-04-30
Refactor send page state management

brad-decker:
### Refactor Send Page State Management

*Problem Statements*
1. The "state" of the send page is not centralized making it near impossible to understand the logic because each component owns it's own impact on the state. The result was the same overall effect being applied in multiple places (ex: updating the max amount when the max button is clicked, AND anytime the gas field is modified AND anytime the asset was changed, each implementation in it's respective component's componentDidUpdate/lifecycle methods)
2. The Components that make up the UI are too complex due to being the sole managers of state, making it near impossible to refactor or update. 
3. Changes required for future logic updates would need to be carried through many different places

*Design goals of refactor*
1. No functional change to the UX. 
2. Centralize all state update logic as it pertains to the send page into a send slice using redux-toolkit.
3. keep metamask/background state and send page state clearly separated.
4. capture the *intent* of the send page, which is to create a draft transaction.
5. eliminate any dead code found or created by way of this refactor
6. adhere to the redux-toolkit design guide as much as possible

PR #11248, opened 2021-06-07
Add gas constants

brad-decker:
First PR breaking up from #10965 - This one just adds gas constants for 21000 and 100000 that are shared throughout the app.

### DIFF
--- a/test/stub/tx-meta-stub.js
+++ b/test/stub/tx-meta-stub.js
@@ -1,3 +1,4 @@
+import { GAS_LIMITS } from '../../shared/constants/gas';
 import {
   TRANSACTION_STATUSES,
   TRANSACTION_TYPES,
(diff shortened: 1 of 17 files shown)


jsconfig.json
```diff
@@ -1,14 +1,3 @@
...
-    "coverage",
-    "dist",
-    "docs",
-    "lavamoat",
-    "node:console",
-    "node_modules",
-    "patches",
-    "test-artifacts"
-  ]
+  "exclude": ["node:console"]
```
darkwing, line 2: Is this supposed to be in this PR?  Seems out of place.

test/stub/tx-meta-stub.js
```diff
@@ -16,7 +17,7 @@ export const txMetaStub = {
       type: TRANSACTION_TYPES.SENT_ETHER,
       txParams: {
         from: '0xf231d46dd78806e1dd93442cf33c7671f8538748',
-        gas: '0x5208',
+        gas: GAS_LIMITS,
```
darkwing, line 20: I think this needs to be `GAS_LIMITS.SIMPLE`

darkwing (changes requested):
I think I spotted an incorrect variable and an extra modified file, so marking as Request Changes.

brad-decker:
thanks @darkwing, ready for another review.

darkwing (approved):

PR #11249, opened 2021-06-07
move metamask state selectors out of send

brad-decker:
Moves selectors from the 'send' file in the selectors folder to the metamask duck, which more closely aligns with where the values are derived from. Previously they were in the send selectors because they are predominately used in that flow but that isn't the best organization model. This is pulled out of #10965

### DIFF
--- a/ui/components/app/asset-list/asset-list.js
+++ b/ui/components/app/asset-list/asset-list.js
@@ -11,10 +11,10 @@ import { useMetricEvent } from '../../../hooks/useMetricEvent';
 import { useUserPreferencedCurrency } from '../../../hooks/useUserPreferencedCurrency';
 import {
   getCurrentAccountWithSendEtherInfo,
-  getNativeCurrency,
   getShouldShowFiat,
   getNativeCurrencyImage,
 } from '../../../selectors';
+import { getNativeCurrency } from '../../../ducks/metamask/metamask';
 import { useCurrencyDisplay } from '../../../hooks/useCurrencyDisplay';
 
 const AssetList = ({ onClickAsset }) => {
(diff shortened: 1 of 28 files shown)


ui/components/app/asset-list/asset-list.js
```diff
@@ -11,10 +11,10 @@ import { useMetricEvent } from '../../../hooks/useMetricEvent';
 import { useUserPreferencedCurrency } from '../../../hooks/useUserPreferencedCurrency';
 import {
   getCurrentAccountWithSendEtherInfo,
-  getNativeCurrency,
   getShouldShowFiat,
   getNativeCurrencyImage,
```
darkwing, line 15: Would we also want to pull this one out as well?

ui/selectors/send.test.js
```diff
@@ -279,78 +248,12 @@ describe('send selectors', () => {
...
-    });
-  });
-
   describe('getTokenBalance()', () => {
     it('should', () => {
       expect(getTokenBalance(mockState)).toStrictEqual(3434);
     });
   });
 
-  describe('getUnapprovedTxs()', () => {
```
darkwing, line 328: Were all of these tests unhelpful, or could/should they be moved elsewhere?

ui/components/app/asset-list/asset-list.js
brad-decker: This one is defined in the base selectors file, which isn't really any better but at least its not in `selectors/send` -- I think long term all selectors should move into appropriate slices to more clearly indicate which part of state they are "selecting" from. For now though we can leave it where it's at.

ui/selectors/send.test.js
brad-decker: Good call! Moved them to the metamask duck test file

brad-decker:
ready for another peek @darkwing

darkwing:
Testing now!

darkwing (approved):

PR #11252, opened 2021-06-07
Relocate send state into send duck from metamask duck

brad-decker:
Moves the send state out of the metamask duck, which confuses the distinction between frontend and background state. This is predicate work towards #10965 and should make the review of said work slightly easier.

~There is one strange quirk from this work that required changes outside of just relocating values:~

1. ~on develop, the send state exists inside the metamask slice which is not correct.~
2. ~this pr moves that state and related selectors/action creators into the send duck.~
3. ~when you click ‘next’ at the bottom of the send page, a transaction is created in the background. This will become an unapproved tx that you have to confirm.~
4. ~the next button will also forward you to the confirmation page.~
5. ~there doesn’t seem to be any guarantee of order of operations between when the unapproved tx is added and when the route changes. We do a .then() on the sign() call in send-footer.component but that method doesn’t appear to guarantee to return a promise.~
6. ~without modifying the .onSubmit functionality in send-footer, the relocation of the send state outside of metamask slice seems to make this race condition evident.~
7. ~result: you click next and land on the homepage and you (moments / fraction of a second) later see an unapproved tx appear in your activity list.~

~to fix this I had to add a manual delay of 500ms in the signTx and signTokenTx methods to make them wait long enough to be effective.~

EDIT:

thanks @Gudahtt for finding the issue!


ui/ducks/app/app.test.js
```diff
@@ -352,22 +352,4 @@ describe('App State', () => {
 
     expect(state.isMouseUser).toStrictEqual(true);
   });
-
-  it('sets gas loading', () => {
```
brad-decker, line 356: These tests are removed versus being moved because they will eventually be obsolete in #10965

ui/ducks/metamask/metamask.test.js
```diff
@@ -480,32 +307,6 @@ describe('MetaMask Reducers', () => {
...
-        type: actionConstants.UPDATE_SEND_ENS_RESOLUTION,
-        payload: '0x1337',
-      },
-    );
-
-    expect(state.send.ensResolution).toStrictEqual('0x1337');
-    expect(state.send.ensResolutionError).toStrictEqual('');
-  });
-
-  it('update ensResolutionError', () => {
```
brad-decker, line 496: These tests were removed instead of relocated because they will eventually be obsolete in #10965

ui/store/actions.test.js
```diff
@@ -853,67 +853,6 @@ describe('Actions', () => {
     });
   });
 
-  describe('#updatedGasData', () => {
```
brad-decker, line 856: Tests deleted instead of moved because will eventually be obsolete in #10965

test/e2e/tests/simple-send.spec.js
```diff
@@ -23,8 +23,9 @@ describe('Simple send', function () {
           '0x985c30949c92df7a0bd42e0f3e3d539ece98db24',
         );
         await driver.fill('.unit-input__input', '1');
-        await driver.clickElement('[data-testid="page-container-footer-next"]');
-        await driver.clickElement('[data-testid="page-container-footer-next"]');
+        await driver.clickElement({ text: 'Next', tag: 'button' });
```
brad-decker, line 26: Due to having to add a small delay after submitting a tx from send flow, the testid selector is *too fast* and results in trying to select the element that is hidden behind the loading indicator. This matches the other tests now and is *slow enough* to not break. Also all of this will be unnecessary once #10965 lands

ui/store/actions.js
```diff
@@ -624,196 +625,36 @@ export function signTypedMsg(msgData) {
...
 export function signTokenTx(tokenAddress, toAddress, amount, txData) {
   return async (dispatch) => {
     dispatch(showLoadingIndication());
 
     try {
       const token = global.eth.contract(abi).at(tokenAddress);
-      const txPromise = token.transfer(toAddress, addHexPrefix(amount), txData);
-      dispatch(showConfTxPage());
+      token.transfer(toAddress, addHexPrefix(amount), txData);
+      await wait(500);
```
brad-decker, line 655: This wait is necessary because there is no guarantee that after the `token.transfer` call is made that the transaction has fully made its way into the unapproved transaction state. I am not sure why moving the send state out of the metamask state makes this issue happen but the way it sufficient to rectify it. Tried with 200,300 and 400.500 is successful 100% of the time while 400 had intermittent test failures.

ui/store/actions.js
```diff
@@ -624,196 +625,36 @@ export function signTypedMsg(msgData) {
...
 export function signTx(txData) {
-  return (dispatch) => {
+  return async (dispatch) => {
+    dispatch(showLoadingIndication());
     global.ethQuery.sendTransaction(txData, (err) => {
       if (err) {
         dispatch(displayWarning(err.message));
       }
     });
+    await wait(500);
```
brad-decker, line 635: This wait is necessary because there is no guarantee that after the token.transfer call is made that the transaction has fully made its way into the unapproved transaction state. I am not sure why moving the send state out of the metamask state makes this issue happen but the way it sufficient to rectify it. Tried with 200,300 and 400.500 is successful 100% of the time while 400 had intermittent test failures.

ui/pages/send/send-footer/send-footer.container.js
```diff
@@ -88,7 +88,7 @@ function mapDispatchToProps(dispatch) {
         to,
       });
 
-      sendToken
+      return sendToken
```
brad-decker, line 91: in the send-footer component we do a .then chain off of the sign method but we don't return anything from this function so its unclear to me how the .then() ever functioned before this.

ryanml (dismissed):
This looks good to me! Did a smoke test of things as well

ui/store/actions.js
```diff
@@ -624,196 +625,36 @@ export function signTypedMsg(msgData) {
...
   return async (dispatch) => {
     dispatch(showLoadingIndication());
 
     try {
       const token = global.eth.contract(abi).at(tokenAddress);
-      const txPromise = token.transfer(toAddress, addHexPrefix(amount), txData);
-      dispatch(showConfTxPage());
+      token.transfer(toAddress, addHexPrefix(amount), txData);
+      await wait(500);
       dispatch(hideLoadingIndication());
```
Gudahtt, line 656: Why remove the loading indicator before showing the confirmation page? I had thought that was intended to give the confirmation page time to finish any on-mount stuff before it was shown. I'm not really sure though.

ui/store/actions.js
Gudahtt: Possibly related, though I'm not sure how yet, but.... it looks like the call to `showConfTxPage` does nothing but hide the loading indicator and clear any warnings that are set (if any). It doesn't do anything. It does not show the confirm transaction page :confused: 

So that has nothing to do with the change in behaviour here. It must be in the UI component.

ui/store/actions.js
brad-decker: showConfTxPage is pretty confusing. It doesn't actually show the confTxPage it sets the txId in the appState though (but we don't know the txId here so it gets set to nothing)... also txId is only referenced (to my knowledge) in conf-tx which ... only works for signature requests? 

```
        <Route
          exact
          path={`${CONFIRM_TRANSACTION_ROUTE}/:id?${SIGNATURE_REQUEST_PATH}`}
          component={ConfTx}
        />
```

All of this to say this wasn't an intentional change I just didn't understand what it was doing I'm going to restore the previous order.

ui/store/actions.js
brad-decker: in send-footer.component, onSubmit:
```js
    Promise.resolve(promise).then(() => {
      metricsEvent({
        eventOpts: {
          category: 'Transactions',
          action: 'Edit Screen',
          name: 'Complete',
        },
        customVariables: {
          gasChanged: gasEstimateType,
        },
      });
      history.push(CONFIRM_TRANSACTION_ROUTE);
    });
```

brad-decker: what's weird to me is before I started messing with this, the 'sign' method didn't return a promise... just the 'update' method. At least that's how I interpreted it.

brad-decker: oh haha, Promise.resolve(promise) is just going to return a resolved promise that resolves whatever is passed in 🤦🏻

Gudahtt (dismissed):
I think I found the problem - the `send` prop in `confirm-transaction.container.js` hasn't been updated, so it's failing to find the send flow transaction id and redirecting back to home

brad-decker:
You did indeed find the problem. Working on removing the stupid bits of code now

ryanml (approved):

PR #11011, opened 2021-05-07
Update jsconfig excludes array

rekmarks:
I and other have experienced degraded performance in VS Code recently. In particular, I've constantly received notifications that I need to exclude files for the benefit of VS Code's native JavaScript and TypeScript extension. 

VS Code's [documentation](https://code.visualstudio.com/Docs/languages/javascript) has this to say on the matter:

> If not all JavaScript files in your workspace should be considered part of a single JavaScript project. `jsconfig.json` files let you exclude some files from showing up in IntelliSense.

This PR adds a bunch of directories and one glob to the `excludes` array of our `jsconfig.json`, in the hope of improving or restoring VS Code / IntelliSense performance for the extension repository.

PR #11253, opened 2021-06-08
fix jsconfig woes

brad-decker:
in #11011 @rekmarks correctly identified an issue that was slowing down our vscode experience, for some reason the dist folder is like a black hole sucking all autocomplete into it. If you ignore it, initializing language features takes *forever* and crashes frequently. If you don't ignore it autocomplete is fast but you get warned about slow features.

The solution seems to *only* include our three primary js directories.

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -1,14 +1,7 @@
 {
-  "exclude": [
-    "*.log",
-    "builds",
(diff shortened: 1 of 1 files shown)


NiranjanaBinoy (approved):
LGTM

PR #11254, opened 2021-06-08
fix issue with importing ui module space inside shared

brad-decker:
I accidently created an issue where we are importing the UI module space inside the shared module space which would inadvertently cause leaking of imports across UI into the app/ folder when importing the gas file. Because the numbers involved are lower than the max safe integer value it's safe to just use toString(16)


ryanml (approved):
lgtm!

PR #11259, opened 2021-06-09
Bringing recovery phrase text notification to the top

ryanml:
Screen Shot 2021-06-09 at 3 56 27 AM

### DIFF
--- a/shared/notifications/index.js
+++ b/shared/notifications/index.js
@@ -24,7 +24,7 @@ export const UI_NOTIFICATIONS = {
   },
   5: {
     id: 5,
-    date: '2021-05-18',
+    date: '2021-06-09',
   },
   6: {
     id: 6,


shared/notifications/index.js
```diff
@@ -24,7 +24,7 @@ export const UI_NOTIFICATIONS = {
   },
   5: {
     id: 5,
-    date: '2021-05-18',
+    date: '2021-06-09',
```
ryanml, line 27: Since notification order is sorted by date by the `getSortedNotificationsToShow` selector, it made sense to just to update the date of the Recovery phrase entry rather than re-order any other static structures

danjm (approved):

ryanml:
Merging as the audit failure is unrelated to this change

PR #11264, opened 2021-06-09
Add warning when domain name doesn't match origin for typed messages requesting signature

### LINKED ISSUE #10576: Ensure signTypedData restricted to a domain
EIP712 has a `domain` field that is meant to be used as a security measure, to restrict the range of possible signatures.

We do not currently enforce `domain` in any way that could prove to a contract that a signature was suggested to the wallet from a given web app.

This has been requested by an L2 team using `signTypedData`.

adonesky1:
Fixes: #10576

Explanation:  Add warning when domain name doesn't match origin for typed messages requesting signature

Manual testing steps:  
  - run the e2e test dapp
  - click the Sign Typed Data V3 and/or Sign Typed Data V4
  - Check that there is a warning at the top of the MetaMask pop up: `Domain passed to message - Ether Mail - does not match the origin of the signature request: localhost`
  - 
Screen Shot 2021-06-14 at 12 23 15 PM

### DIFF
--- a/app/scripts/lib/typed-message-manager.js
+++ b/app/scripts/lib/typed-message-manager.js
@@ -123,8 +123,16 @@ export default class TypedMessageManager extends EventEmitter {
     if (req) {
       msgParams.origin = req.origin;
     }
-    this.validateParams(msgParams);
-
+    let warning;
+    try {
+      this.validateParams(msgParams);
+    } catch (error) {
+      if (error.message === 'Domain name does not match origin') {
+        warning = `Domain passed to message - ${error.actual} - does not match the origin of the signature request: ${error.expected}`;
+      } else {
+        throw error;
+      }
+    }
     log.debug(
       `TypedMessageManager addUnapprovedMessage: ${JSON.stringify(msgParams)}`,
     );
@@ -138,6 +146,7 @@ export default class TypedMessageManager extends EventEmitter {
       time,
       status: 'unapproved',
       type: MESSAGE_TYPE.ETH_SIGN_TYPED_DATA,
+      warning,
     };
     this.addMsg(msgData);
 
@@ -209,6 +218,13 @@ export default class TypedMessageManager extends EventEmitter {
             `Provided chainId "${chainId}" must match the active chainId "${activeChainId}"`,
           );
         }
+        const { domain } = data;
+        assert.equal(
+          domain?.name,
+          new URL(params?.origin).hostname,
+          'Domain name does not match origin',
+        );
+
         break;
       }
       default:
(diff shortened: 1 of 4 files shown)


adonesky1:
@rachelcope and @danfinlay I think this probably could use a design upgrade?

darkwing:
@adonesky1 Is this blocked on design?

adonesky1:
> @adonesky1 Is this blocked on design?

Yes exactly

app/scripts/lib/typed-message-manager.js
```diff
@@ -209,6 +218,13 @@ export default class TypedMessageManager extends EventEmitter {
             `Provided chainId "${chainId}" must match the active chainId "${activeChainId}"`,
           );
         }
+        const { domain } = data;
+        assert.equal(
+          domain?.name,
+          new URL(params?.origin).hostname,
```
darkwing, line 224: `new URL` can throw if it's passed bad information, so I'm thinking we need a `try/catch` here or simply do `params.origin`.  Do we know if this info could be absent?

darkwing (changes requested):
@adonesky1 Could you ping design on this so we can push it forward?

adonesky1:
> @adonesky1 Could you ping design on this so we can push it forward?

Sorry @darkwing somehow just seeing your followups here! Yes will ping design about this.

app/scripts/lib/typed-message-manager.js
Gudahtt: At the moment at least it's guaranteed to be present. It's set by the browser when we receive a message.

We might have to rethink this if/when we start handling messages from snaps though. Unless we give them valid origins too.

Gudahtt:
Marked this as draft since it appears to be blocked, so this doesn't show up in anyone's PR review queue.

app/scripts/lib/typed-message-manager.js
```diff
@@ -123,8 +123,16 @@ export default class TypedMessageManager extends EventEmitter {
     if (req) {
       msgParams.origin = req.origin;
     }
-    this.validateParams(msgParams);
-
+    let warning;
+    try {
+      this.validateParams(msgParams);
```
Gudahtt, line 128: Rather than include this validation here then explicitly suppress it, maybe it would be simpler to perform the validation in the UI? At least until we graduate this from a warning to a hard error.

Gudahtt:
It looks like this doesn't include any console warning or metrics for this case. We ought to at least make some effort to tell dapp devs not to do this if we want to discourage it, and Finlay had mentioned in the issue wanting to collect metrics on how often this happens.

adonesky1:
This has been open forever. Going to close and revisit when we have designs

app/scripts/lib/typed-message-manager.js
adonesky1: 👀

PR #11265, opened 2021-06-09
Use verifyPassword instead of submitPassword in exportAccounts in actions.js

danjm:
Errors during execution of `submitPassword` while logged in could result in an accidentally deleted vault. We don't need to call submit password in `exportAccounts`, we can call `verifyPassword` instead.


Gudahtt (approved):
LGTM!

ryanml (approved):

PR #11268, opened 2021-06-09
Fix built-in networks switch-ethereum-chain, Including RPC url in switchEthereumChain requestData

ryanml:
Related: https://github.com/MetaMask/metamask-extension/pull/11184 - test plan should be similar


brad-decker (approved):
LGTM!

PR #11271, opened 2021-06-10
Migrate linting and unit tests to GitHub Actions

Gudahtt:
Unit tests and linting has been migrated from CircleCI to GitHub Actions. It should behave roughly the same as before.

The `metamaskbot` comment has been updated to remove the coverage report, because now this report will be stored on GitHub rather than on CircleCI with the other data we link in that comment. It will be restored when the `metamaskbot` comment is moved to GitHub Actions.

The GitHub Actions job "All jobs passed" will need to be added to the set of required status checks after this is merged.

Coverage data is uploaded as an artifact called `coverage`. The HAR file from the Yarn install is saved as an artifact called `install-artifacts`. The install HAR files are included with the dependency cache so that we can save them again as artifacts each time we get a cache hit, so they're easy to locate. Otherwise we might have a difficult time tracking down which previous build was the first to build the cache we're using.


brad-decker:
😱 😄 🚀

Gudahtt:
Oops, closed this by accident.

This seems to work! I think this PR is ready but I'm going to leave it in draft until I have a more complete picture of where the CircleCI => GitHub Actions migration is headed, in case we need to revise this at all.

Gudahtt:
I'm going to close this for now, as I think merging this before we figure out E2E tests would be a mistake. Ideally everything linked in the metamaskbot comment would be kept together. I'll leave the omnibus PR open for now to track progress towards this though: #11279

PR #11279, opened 2021-06-11
Migrate from CircleCI to GitHub Actions

Gudahtt:
This PR is based upon #11271. This represents the work remaining to fully migrate from CircleCI to GitHub Actions. This PR is primarily for testing purposes - this will be split up into smaller PRs for review later.

This PR is still incomplete. I will be adding to it incrementally so that I can test each change separately.


Gudahtt:
Unfortunately we don't have a great story here for retrying failed jobs. GitHub actions doesn't appear to support re-running individual jobs. We could split up the jobs into a separate workflow each, but that makes it difficult to tie them all together while still reporting status checks back to the relevant PR. It's definitely possible, but it seems pretty convoluted to me. It also makes handling artifacts more complicated.

PR #11295, opened 2021-06-15
Enable Chrome E2E logs

Gudahtt:
Chrome logs are now enabled for E2E tests when the 'ENABLE_CHROME_LOGS' environment variable is set to anything other than `false`.

This was helpful to me in debugging Chrome crashes on CI, the ones with the error "unknown error: DevToolsActivePort file doesn't exist". This was the only way to discover the cause of the error. It's also useful for discovering console errors from the background process or from the UI.

It's disabled by default because it makes the test output quite noisy and difficult to read.

This required a patch to `selenium-webdriver`, which has also been submitted upstream here: https://github.com/SeleniumHQ/selenium/pull/9575


brad-decker (approved):

Gudahtt:
Just tested this on this commit: https://github.com/MetaMask/metamask-extension/pull/11279/commits/bc2d4d7425b8a331a115bcd09436b7d225ccbbb8
it seems to work!

PR #11299, opened 2021-06-15
Start dapp directly from `metamask-ui.spec.js`

Gudahtt:
The dapp is now started directly from the `metamask-ui.spec.js` test module. This makes it easier to run independently, and brings it in-line with our other E2E tests. Mainly this was done to facilitate further refactors which will come in later PRs.


brad-decker (approved):
Awesome! LGTM

PR #11302, opened 2021-06-15
Update Chrome version used for E2E testing to v91

Gudahtt:
The Chrome version used for E2E testing has been updated from v79 to v91. This was delayed so long because all attempts to do this in the past had failed. We have since discovered that the failure was caused by a lack of a display to render to. This was resolved by using `xvfb` to create a virtual display when running tests.

Fixes: #

Explanation:  

Manual testing steps:  
  - Try running `yarn test:e2e:chrome` locally with Chrome v91 installed. It should just work!
  - All e2e tests (chrome and firefox) and the benchmark script should work correctly on CI
  - All e2e tests should behave as they did before locally

### DIFF
--- a/test/e2e/run-e2e-test.js
+++ b/test/e2e/run-e2e-test.js
@@ -63,8 +63,17 @@ async function main() {
     throw error;
   }
 
+  let command = 'yarn';
+  const args = ['mocha', '--no-timeouts', e2eTestPath];
+
+  // Run test with virtual display on CI
+  if (process.env.CI === 'true') {
+    command = 'xvfb-run';
+    args.unshift(...['-e', '/dev/stderr', '-a', 'yarn']);
+  }
+
   await retry(retries, async () => {
-    await runInShell('yarn', ['mocha', '--no-timeouts', e2eTestPath]);
+    await runInShell(command, args);
   });
 }
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -448,7 +448,7 @@ jobs:
           command: mv ./builds-test ./builds
       - run:
           name: Run page load benchmark
-          command: yarn benchmark:chrome --out test-artifacts/chrome/benchmark/pageload.json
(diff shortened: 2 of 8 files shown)


Gudahtt:
This depends upon #11301

Gudahtt:
Well, I thought that `xvfb-run` was the key to getting this to work. But something still isn't right here. It seems to work far more often than before, but I'm still regularly seeing failures with the error message "Unable to open X display."

I'm also running into what appear to be legitimate failures. Maybe because of differences in timing with these updates, or maybe due to browser changes.

test/e2e/run-e2e-test.js
```diff
@@ -63,8 +63,17 @@ async function main() {
     throw error;
   }
 
+  let command = 'yarn';
+  const args = ['mocha', '--no-timeouts', e2eTestPath];
+
+  // Run test with virtual display on CI
+  if (process.env.CI === 'true') {
+    command = 'xvfb-run';
```
Gudahtt, line 71: Seemingly the Firefox E2E tests don't like this :thinking: 

I'm not sure why, given that Xvid is already running at this point. And this shouldn't conflict - the `-a` flag ensures it finds a free port to use. We can consider using this just for Chrome, but.... it's unsettling not understanding what's going on here.

In the GitHub Actions migration PR I placed this in the CI conflig directly, rather than here. It was needed for Firefox there, since the GItHub Actions base image doesn't run `Xvfb` automatically like the CircleCI one does.

I have no idea what impact running it per-test-suite has v.s. once for all test suites. If it's not shutting down properly or starting up fast enough, that could explain some crashes.

Gudahtt:
To give a bit more context: This isn't a problem for most projects because most projects use Chrome in headless mode when running E2E tests. Headless mode doesn't support extensions (and it likely never will according to a few discussions in bug reports I saw), so we can't do that.

Issue #11296, opened 2021-06-15
Provided chainId "x" must match the active chainId "x"


joehawary:
**Describe the bug**
When performing a Signed transaction using signTypedData an error is thrown.

This includes when passing ChainID as an INT, a HEX, a BigNumber.

This issue has been introduced in 9.6.0 and was not present in 9.5.9.

**Steps to reproduce (REQUIRED)**
1. perform _signTypedData with domain including ChainId
2. Returned error Provided chainId "x" must match the active chainId "x"

**Expected behavior**
No error

**Screenshots**
![Screenshot 2021-06-15 at 14 15 22](https://user-images.githubusercontent.com/32411671/122059307-65421800-cde4-11eb-80c5-a3d99b801629.png)

![Screenshot 2021-06-15 at 10 14 36](https://user-images.githubusercontent.com/32411671/122026821-80039500-cdc2-11eb-97be-e048aad6a502.png)

![Screenshot 2021-06-15 at 10 18 23](https://user-images.githubusercontent.com/32411671/122027473-0d46e980-cdc3-11eb-8985-dd066f40455a.png)

**Browser details (please complete the following information):**
 - OS: Mac
 - Hardware Wallet: None
 - Browser: Edge Chromium & Chrome
 - MetaMask Version: 9.6.0

joehawary:
I've also tried as a string "4"

Even passing the hex as a string "0x4" (for chainID 4)

joehawary:
Also getting the ChainId from the signer and passing it back is the same.

joehawary:
Trials of data being passed:


BigNumber/Hex
![Screenshot 2021-06-15 at 10 43 08](https://user-images.githubusercontent.com/32411671/122031328-81cf5780-cdc6-11eb-8463-164e5579a9bc.png)

INT
![Screenshot 2021-06-15 at 10 42 41](https://user-images.githubusercontent.com/32411671/122031336-83991b00-cdc6-11eb-9bd3-533b0ca7ae90.png)

String
![Screenshot 2021-06-15 at 10 43 59](https://user-images.githubusercontent.com/32411671/122031422-9ad80880-cdc6-11eb-9dac-a03482a977ff.png)

joehawary:
Okay, so after further investigation, it only works if you remove the ChainId entirely from the signing request.

I believe this will break compatibility to previous versions.

joehawary:
[...]

PR #11309, opened 2021-06-15
patch breaking change in signTypeData param validation

### LINKED ISSUE #11308: Breaking `eth_signTypedData` change in `9.6.0`.
**Describe the bug**
9.6.0 introduced a breaking change with https://github.com/MetaMask/metamask-extension/pull/11012 where string `chainId`s are no longer supported (but were supported as prior to this release).

Specifically the issue stems from the fact that:

```
> assert.equal(1, "1")
undefined
> assert.strict.equal(1, "1")
Uncaught AssertionError
```

**Steps to reproduce**
Use `eth_signTypedData` to sign a message where the chain ID is specified as a string instead of a number. The popular Ethers.js does this when generating a payload for `eth_signTypedData`:

- **With Ethers.js**
  ```js
  const { chainId } = await provider.getNetwork();
  const signature = await signer._signTypedData(
    { chainId },
    { Foo: [{ name: "foo", type: "string" }] },
    { foo: "bar" },
  );
  ```
- **With the injected provider directly**
  ```js
  await ethereum.request({
    method: "eth_signTypedData_v4",
    params: [
      accountAddress,
      JSON.stringify({
        types: {
          Foo: [{ name: "foo", type: "string" }],
          EIP712Domain: [{ name: "chainId", type: "uint256" }],
        },
        domain: { chainId: chainId.toString() },
        primaryType: "Foo",
        message: { foo: "bar" },
      }),
    ],
  });
  ```

The above code worked in versions prior to `9.6.0`.

**Expected behavior**

I expect the `chainId` to support values as strings as it did in previous versions. This would be required in order to properly support `uint256` chain IDs which cannot correctly be represented with JavaScript numbers:
```
> JSON.parse('{"chainId":9007199254740993}').chainId
9007199254740992
```

Furthermore, I think there is a larger problem with the EIP itself. Specifically, it doesn’t go into much detail about the JSON representation of the `message` or `domain` object and the supported Solidity primitive values.

Specifically for the `chainId` in the [EIP JSON RPC sample request](https://eips.ethereum.org/EIPS/eip-712#specification-of-the-eth_signtypeddata-json-rpc) it uses `"chainId": 1` as a number (which is what MetaMask was changed to enforce). However, as previously mentioned, this has issues WRT correctly representing `uint256` chain IDs since JavaScript numbers just can't represent all 256-bit integers.

This also raises the larger question of what values should be accepted for Solidity primitives? For example, what is an acceptable encoding for a `uint32` message field? Are all `100`, `1e2`, `"100"`, `"0x64"` acceptable? Or should they be treated like [`QUANTITIES`](https://eth.wiki/json-rpc/API#hex-value-encoding) so only "0x64" is valid (what I would lean towards for consistency with other JSON RPC methods)? Currently, MetaMask for example, seems to accept both hexadecimal and decimal strings for integers. This also makes it quite weird that message fields accept multiple formats (numbers, decimal and hexidecimal strings) while `chainId` strictly requires a JavaScript number.

 - OS: Fedora 34
 - Browser: Firefox 89
 - MetaMask Version: 9.6.0

**Additional context**

For additional discussion about this issue see comments in the Ledger sign typed data issue #10240:
- https://github.com/MetaMask/metamask-extension/issues/10240#issuecomment-856808423
- https://github.com/MetaMask/metamask-extension/issues/10240#issuecomment-861568240

adonesky1:
(Hot) Fixes: #11308 & #11296

Explanation:  In 9.6 we added a strictEqual check when validating the specified chainId against the current chainId for `signTypedData` calls, this runs afoul of how the domain data is often formatted (as a string or hex string), causing issues for some developers.

### DIFF
--- a/app/scripts/lib/typed-message-manager.js
+++ b/app/scripts/lib/typed-message-manager.js
@@ -1,5 +1,5 @@
 import EventEmitter from 'events';
-import { strict as assert } from 'assert';
+import assert from 'assert';
 import { ObservableStore } from '@metamask/obs-store';
 import { ethErrors } from 'eth-rpc-errors';
 import { typedSignatureHash, TYPED_MESSAGE_SCHEMA } from 'eth-sig-util';
@@ -177,7 +177,7 @@ export default class TypedMessageManager extends EventEmitter {
         break;
       case 'V3':
       case 'V4': {
-        assert.equal(
+        assert.strictEqual(
           typeof params.data,
           'string',
           '"params.data" must be a string.',
@@ -191,7 +191,7 @@ export default class TypedMessageManager extends EventEmitter {
           data.primaryType in data.types,
           `Primary type of "${data.primaryType}" has no type definition.`,
         );
-        assert.equal(
+        assert.strictEqual(
           validation.errors.length,
           0,
           'Signing data must conform to EIP-712 schema. See https://git.io/fNtcx.',


app/scripts/lib/typed-message-manager.js
```diff
@@ -177,7 +177,7 @@ export default class TypedMessageManager extends EventEmitter {
         break;
       case 'V3':
       case 'V4': {
-        assert.equal(
+        assert.strictEqual(
```
adonesky1, line 180: preserving strict equality check where there is no reported issue.

ryanml (approved):

brad-decker (approved):
LGTM, but I do think going back to legacy assertion is a regression in itself ... We will need to return to this after the patch to figure out how to handle this long term. This is already discussed in internal (slack) communication but wanted to raise it here as well.

adonesky1:
> LGTM, but I do think going back to legacy assertion is a regression in itself ... We will need to return to this after the patch to figure out how to handle this long term. This is already discussed in internal (slack) communication but wanted to raise it here as well.

agreed. In this particular instance could just involve parsing all possible matching chainIds (string, hex, bigNumber) and making sure that the active is in that set?

otherwise @ryanml do we feel good about merging and cutting a patch?

ryanml:
@adonesky1 - no objections from me, but we should file an issue for followup reasonably soon per @brad-decker

adonesky1:
> @adonesky1 - no objections from me, but we should file an issue for followup reasonably soon per @brad-decker

Just wrote one up: https://github.com/MetaMask/metamask-extension/issues/11310

PR #11316, opened 2021-06-15
Prevent propType errors in for TransactionIcon statuses

darkwing:
While looking at the send flow for EIP1559, I spotted a console error about 'confirmed' and 'submitted' not being valid status.  Fixing this.


ryanml (dismissed):
++

ui/components/app/transaction-icon/transaction-icon.js
```diff
@@ -63,6 +63,8 @@ TransactionIcon.propTypes = {
     TRANSACTION_STATUSES.REJECTED,
     TRANSACTION_GROUP_STATUSES.CANCELLED,
     TRANSACTION_STATUSES.DROPPED,
+    TRANSACTION_STATUSES.CONFIRMED,
```
brad-decker, line 66: these statuses are not dealt with and default to the OK_COLOR which is fine for confirmed, but needs to be PENDING_COLOR for submitted. I'm also unsure why we have a mix of individual statuses and group statuses in this map but that's a story for a different day. If we just add SUBMITTED: PENDING_COLOR to the status color map that'll be sufficient for now.

brad-decker (changes requested):
One small fix to make sure we are handling the submitted status appropriately.

brad-decker (approved):

adonesky1 (approved):

PR #11318, opened 2021-06-15
Update `benchmark.js` script to use `yargs`

Gudahtt:
Our benchmark script now uses `yargs`. Functionally it should be nearly the same as before, except that now it has more documentation and validation. The one functional difference aside from that is that the `--pages` flag now takes space-separated arguments rather than comma-separated.

Manual testing steps:  
  - Try out the `yarn benchmark:chrome` and `yarn benchmark:firefox` npm scripts! I'd suggest passing in the `--samples` option though - the default is 20, which takes quite a while. Be sure to use at least 2 samples though - it throws an error with 1.

### DIFF
--- a/test/e2e/benchmark.js
+++ b/test/e2e/benchmark.js
@@ -2,6 +2,8 @@
 
 const path = require('path');
 const { promises: fs, constants: fsConstants } = require('fs');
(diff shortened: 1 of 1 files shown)


brad-decker (approved):
LGTM

PR #11319, opened 2021-06-15
Add retries to the benchmark script

Gudahtt:
The benchmark script can now be set to retry upon failure, like the E2E tests do. The default is zero, just as with the E2E tests. A retry of 2 has been set in CI to match the E2E tests as well.

The `retry` module had to be adjusted to throw an error in the case of failure. Previously it just set the exit code, but that only worked because it was the last thing called before the process ended. That is no longer the case.

Manual testing steps:  
  - Try out the `yarn benchmark:chrome` and `yarn benchmark:firefox` npm scripts with the `--retries` option! I'd suggest passing in the `--samples` option though - the default is 20, which takes quite a while. Be sure to use at least 2 samples though - it throws an error with 1.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -448,7 +448,7 @@ jobs:
           command: mv ./builds-test ./builds
       - run:
           name: Run page load benchmark
-          command: yarn benchmark:chrome --out test-artifacts/chrome/benchmark/pageload.json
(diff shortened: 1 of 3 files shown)


Gudahtt:
This depends upon #11318

brad-decker (dismissed):

brad-decker (approved):

PR #11320, opened 2021-06-15
Fix benchmark script with 1 sample

Gudahtt:
Previously the benchmark script would throw an error if asked to take just 1 sample. Now it works, though the stats returned are of dubious use.

The problem was that it was impossible to calculate the standard deviation or margin of error of a set of 1. Instead it now returns zero for both of those values in the single-sample case, which is what it would return for two identical samples.

Manual testing steps:  
  - Try out the `yarn benchmark:chrome` and `yarn benchmark:firefox` npm scripts with `--samples` set to 1.


brad-decker (approved):
Tested locally 👍🏻

PR #11321, opened 2021-06-15
Add `--leave-running` flag to E2E test script

Gudahtt:
The `--leave-running` flag has been added to the E2E test runner. This ensures the browser, ganache, and everything else stays running upon test failure. This is useful for local debugging, for investigating what state the extension was in when it failed.

Manual testing steps:  
  - Try running `yarn test:e2e:single --browser firefox --leave-running` to see that it works correctly for a working test (it should exit normally).
  - Try running `yarn test:e2e:single --browser firefox --leave-running` on a test that will fail (e.g. one that you introduced an error into). Everything should stay running. It should all exit cleanly after sending an interrupt (CTRL+C).

### DIFF
--- a/test/e2e/helpers.js
+++ b/test/e2e/helpers.js
@@ -33,6 +33,7 @@ async function withFixtures(options, testSuite) {
   let segmentStub;
 
   let webDriver;
+  let failed = false;
   try {
     await ganacheServer.start(ganacheOptions);
     if (ganacheOptions?.concurrent) {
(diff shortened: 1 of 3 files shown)


brad-decker (dismissed):
I assume this is intended not to work yet with the metamask-ui file knowing that it'll eventually be replaced with using the fixtures?

Gudahtt:
Oh! No, that was intended, I just forgot. I'll update this to do that too. That might take a while to get rid of after all.

Gudahtt:
Ok, this should work with `metamask-ui.spec.js` now as well.

Spent some time trying to find an equivalent for `this.currentTest.state` for the `after` hook, but it doesn't appear to exist :thinking: :shrug:

brad-decker:
that seems counterintuitive at the least. working well locally though!

brad-decker (approved):

PR #11322, opened 2021-06-16
Make MAX button accessible

darkwing:
The MAX button is not accessible via keyboard, which is very problematic.  Let's fix that.


ryanml (approved):
love it!

ryanml:
@darkwing just for documentation's sake, would you mind posting a screenshot of the max button with this diff? :) I know it should look the same.

darkwing:
https://user-images.githubusercontent.com/46655/122145041-229b3280-ce1a-11eb-910b-35956b1c265f.mp4

PR #11323, opened 2021-06-16
fix xpath to get firefox uuid

hsy822:
When running the test code in Firefox, if the Firefox language setting is not English, the 'Internal UUID' property cannot be fetched and the test can not proceed. So I just changed 'Internal UUID' to 'UUID'

### Before (Cannot access element 'Internal UUID')
before1

### After (It works!)
after


Gudahtt (approved):
LGTM, thanks for the contribution!

PR #11326, opened 2021-06-16
Fix case sensitive check of chainID

### LINKED ISSUE #11325: Connecting to a new network throws "endpoint returned a different chain ID" even when entering the same value
**Describe the bug**
When attempting to connect to an Infura endpoint for a new network, and entering the chainId, metamask throws an error saying the chainId is different even though the value of the chainId entered is the same.

image

1. Go to 'Networks -> New Network'
2. Set the RPC Url to https://palm-mainnet.infura.io/v3/6071ae03c33343bba487b780a9851329
3. Set the chainId to 11297108109
4. Attempt to save the network 
5. Error is thrown 

**Expected behavior**
The chainId entered is the same and should work. The error thrown isn't clear as I'm entering the same value it requires but it says "The endpoint returned a different chain ID"

**Screenshots**

 - OS: [Linux - Ubuntu 20.04]
 - Hardware Wallet: [n/a]
 - Browser: [e.g. Chrome Version Version 90.0.4430.212 (Official Build) (64-bit)]
 - MetaMask Version: [e.g. 9.6.0, 9.6.1]

adonesky1:
Fix issue where case sensitive check of chainID while adding network causes incorrect validation error

Fixes: #11325

Explanation:  
In the add custom network flow we are incorrectly doing a case sensitive comparison of the user input chainID and the chainId returned when checking against the RPC endpoint.

### DIFF
--- a/ui/pages/settings/networks-tab/network-form/network-form.component.js
+++ b/ui/pages/settings/networks-tab/network-form/network-form.component.js
@@ -176,7 +176,7 @@ export default class NetworkForm extends PureComponent {
         blockExplorerUrl,
       } = this.state;
 
-      const formChainId = stateChainId.trim().toLowerCase();
+      const formChainId = stateChainId.trim();
       const chainId = this.prefixChainId(formChainId);
 
       if (!(await this.validateChainIdOnSubmit(formChainId, chainId, rpcUrl))) {
@@ -394,7 +394,7 @@ export default class NetworkForm extends PureComponent {
 
     if (providerError || typeof endpointChainId !== 'string') {
       errorMessage = t('failedToFetchChainId');
-    } else if (parsedChainId !== endpointChainId) {
+    } else if (parsedChainId.toUpperCase() !== endpointChainId.toUpperCase()) {
       // Here, we are in an error state. The endpoint should always return a
       // hexadecimal string. If the user entered a decimal string, we attempt
       // to convert the endpoint's return value to decimal before rendering it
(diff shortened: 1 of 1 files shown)


brad-decker:
The underlying issue this PR was created for has been resolved on the client to return lowercase chainId. We may want to push for standardization but in my opinion that standard should be lowercase as this is what is returned from calling `.toString(16)`

JTraversa (approved):
Any link to the resolution in the client? 

Currently methods based on chainid are still blocked in chrome on our end.

brad-decker:
@JTraversa -- Can you elaborate on your experience around this issue. Either here or in #11325

brad-decker:
Is there a specific network you are experiencing this on?

JTraversa:
Sure. As with many projects ours relies on ethereum.signtypeddata which includes a chainId param.

When placing a limit order users sign our order object with this method.

Currently those using metamask through chrome are prevented from signing, finding the RPC rejects the method because the chainid doesnt match. We're experiencing this on Kovan, so the chainid is 42 (0x2a).

brad-decker:
@adonesky1 can correct me if i'm wrong but I think you're looking more for https://github.com/MetaMask/metamask-extension/issues/11296 this issue which was fixed by this #11309 and including in 9.6.1 which will be rolling out to all users soon

JTraversa:
You are correct, apologies, I tracked down the wrong PR after finding that issue.

brad-decker:
no worries at all!

adonesky1:
@danjm @brad-decker @darkwing Do we have any new thoughts/feelings about whether to add case insensitivity here? Or do we want to punish clients that don't follow standard (lowercase)?

adonesky1:
Closing for now. Will reopen if this issue is re-raised by users.

PR #11327, opened 2021-06-16
Adding edit navigation button flow back to send token flow

### LINKED ISSUE #11317: Missing edit button for token send flow
**Describe the bug**
While reviewing @brad-decker send slice refactor PR I found [some invalid logic](https://github.com/MetaMask/metamask-extension/pull/10965#discussion_r652108114) around editing a transaction in the send flow where the recipient address and amount were switched. Upon further digging I realized this wasn't causing a visible issue because the (back to) edit option that is available for ETH transactions is not available for token transactions (where this logic would be used). I don't believe there is a good reason that we should allow users to return to edit a transaction in the ETH send flow but not in a token send flow? 

1. Start to send a token (on any network, and token that is not its native asset)
2. Once at the confirm screen look for the edit button in the top left corner and see that it isn't present

**Expected behavior**
Users should be able to navigate back to edit a transaction while sending a token.

**Screenshots**

 - OS: OSX
 - Browser:  Chrome
 - MetaMask Version: 9.6.0

adonesky1:
Fixes: #11317

Explanation:  While reviewing @brad-decker send slice refactor PR I found [invalid logic](https://github.com/MetaMask/metamask-extension/compare/add-edit-back-send-tokens?expand=1#diff-25bd56a27f74d50f95a8063ba5e8186a8ef9e608cf2ceff651d964d11c2c77eb) around editing a transaction in the send flow where the recipient address and amount were switched. Upon further digging I realized this wasn't causing a visible issue because the (back to) edit option that is available for ETH transactions is not available for token transactions (where this logic would be used). This PR adds back the edit navigation button for the send-token flow and fixes the faulty logic.

Manual testing steps:  
- Start to send a token (on any network, and token that is not its native asset)
- Once at the confirm screen look for the edit button in the top left corner and click it
- Edit the transaction and click forward again to confirm.
- Send the transaction and verify that the transaction sent the correct amount to the correct address

### DIFF
--- a/ui/ducks/confirm-transaction/confirm-transaction.duck.js
+++ b/ui/ducks/confirm-transaction/confirm-transaction.duck.js
@@ -3,7 +3,7 @@ import {
   currentCurrencySelector,
   unconfirmedTransactionsHashSelector,
 } from '../../selectors';
-import { getNativeCurrency } from '../metamask/metamask';
+import { getNativeCurrency, getTokens } from '../metamask/metamask';
 
 import {
   getValueFromWeiHex,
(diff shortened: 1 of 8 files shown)


brad-decker:
@adonesky1 this is going to have some conflicts with #10965 and #11328-- my recommendation would be to apply your changes to these branches. I moved the logic of the confirmation pages 'editTransaction' function call into the send slice in #11328.

adonesky1:
> @adonesky1 this is going to have some conflicts with #10965 and #11328-- my recommendation would be to apply your changes to these branches. I moved the logic of the confirmation pages 'editTransaction' function call into the send slice in #11328.

I should PR against both branches?

adonesky1:
@brad-decker Oh I see  #11328  is PRed against #10965

brad-decker:
Yeah just #11328.

adonesky1:
Closing this PR. I've opened a new one against #11328.

PR #11328, opened 2021-06-16
fix issues found during QA of send slice

brad-decker:
@PeterYinusa found the following issues in QA (so far):

1. When clicking the blue dialog for "new address detected click here to add to address book" and then supplying a name in the modal for saving the contact, the new name isn't reflected in the address bar.
2. When clicking 'edit' from a transaction being sent to an account in the address book, the name wasn't being reflected in the address bar. 
3. @adonesky1 brought up that the edit token flow had inversed parameters in the send token confirmation.
4. @PeterYinusa found an issue where once an error is displayed for input it isn't removed when the error goes away. Example: start typing a domain name slowly so that denounced validation occurs. You'll get an invalid recipient error. Then complete the ENS domain and the error doesn't go away even though a resolution is found. 

This PR, so far, 
1. Adds a new action for when the address book changes, which we now listen for in the send slice and update the state.recipient.nickname when the state.recipient.address appears in the new address book.
2. Moves to using an action creator for editTransaction, which pulls the transaction out of unapprovedTxs by id. Now all setup for editTransaction occurs in the send slice with just a few details needed to be provided by the various confirmation page. This fixes that issue @adonesky1 but doesn't implement edit flow for tokens (we can enable in a new PR). Adds tests for this case as well.
3. Clears out errors/warnings when the error cases don't evaluate to true


ui/ducks/send/send.js
```diff
@@ -260,7 +267,7 @@ async function estimateGasLimitForSend({
   }
 }
 
-async function getERC20Balance(token, accountAddress) {
+export async function getERC20Balance(token, accountAddress) {
```
adonesky1, line 270: I'm not seeing this method used outside of this file?

ui/ducks/send/send.js
```diff
@@ -892,6 +903,16 @@ const slice = createSlice({
           slice.caseReducers.validateSendState(state);
         }
       })
+      .addCase(ADDRESS_BOOK_UPDATED, (state, action) => {
+        const { addressBook } = action.payload;
+        console.log(state.recipient.address, addressBook);
```
adonesky1, line 908: stray console.log 🧑‍🔧

ui/ducks/send/send.js
brad-decker: its being exported so it can be mocked in tests

adonesky1: Ah gotcha. I didn't realize that was necessary. Cool

ui/ducks/send/send.js
```diff
@@ -1240,6 +1261,76 @@ export function signTransaction() {
...
+      await dispatch(
+        actions.editTransaction({
+          id: transactionId,
+          gasLimit,
+          gasPrice,
+          from,
+          amount,
+          address,
+          nickname,
+        }),
```
adonesky1, line 1293: I might be missing something but is there a reason we don't `updateSendAsset({type: ASSET_TYPES.NATIVE ...)}` here?

adonesky1: Or actually I think it might be that the updateSendAsset call in the else block is redundant? From playing around with it a bit I can't think of or find a scenario where the assetType / details aren't already updated and aligned when you reach the updateSendAsset call on line 1313, could very well be missing something though!

brad-decker: You have to set the asset on the send flow that you are going to be editing. The default send asset is native so that's why we don't call it in the first block.

Scenario 1: clicking edit from the send-eth-confirmation page
1. State is initialized with asset type = NATIVE
2. we call editTransaction which changes stage to EDIT
2. initialization of the send slice happens and sets the asset.balance to equal account.balance

Scenario 2: clicking edit from the send-token-confirmation page
1. State is initialized with asset type = TOKEN
2. We first update asset/balance which sets the appropriate values
3. We then call editTransaction which changes stage to EDIT
4. initilization occurs and uses the balance and asset we just set in state.

adonesky1: Yup that checks out!

PR #11330, opened 2021-06-16
WIP: don't leak UI node_modules into the background bundle

brad-decker:
Explanation:  
We added a shared folder to hold things that are required in both `ui/` and `app/` to avoid accidentally bringing unrelated dependencies along with code when importing pieces across the module barrier. We didn't ever go back to untangle these bundles though. This PR is a POC for what it would take to eliminate importing UI from the app folder.


brad-decker:
@tmashuang @dan437 @PeterYinusa  -- I think we'll want to use jest in the /shared folder for tests instead of mocha to ease this transition. Would any of you be interested in setting that up. @tmashuang i know you were already working on the mocha -> jest conversion for the extension?

brad-decker:
@dan437 -- This is going to require moving a lot of swaps functionality into the shared swaps.util file -- could you see how you'd like to do this best? This PR just cut and rearranged a lot of stuff, including moving a lot of extension utils into shared folder.

brad-decker:
Although I think keeping the boundaries between applications strict is still important, the work by @kumavis and others on updating the build system so that it chunks code into smaller bundles makes this work less impactful.

PR #11332, opened 2021-06-17
move @ethereumjs/common to bg-libs

brad-decker:
Fixes: 

Not much, buys us a little bit of package room to prevent leaky deps #11330 from increasing bundle size past the 4mb mark.

This impacted me in #10965


brad-decker:
This isn't, itself, enought to alleviate this concern. This PR has been included in the send slice refactor where the issue occurred.

PR #11334, opened 2021-06-17
use jest for testing shared

brad-decker:
as discovered as part of exploration in #11330, the pieces of code we are importing from the UI folder inside of the app folder/background process are bringing unused, unnecessary dependencies with it. This PR simply makes jest the test runner for the shared folder. The tests in this folder will run correctly but will produce warnings in eslint for not having used jest's expect. This will be remedied by future PRs!


dan437 (approved):
Looks good to me!

darkwing (approved):

ryanml (approved):
++ lgtm

PR #11335, opened 2021-06-17
Adding edit navigation button flow back to send token flow

### LINKED ISSUE #11317: Missing edit button for token send flow

adonesky1:
Fixes: #11317

Explanation: While reviewing @brad-decker send slice refactor PR I found invalid logic around editing a transaction in the send flow where the recipient address and amount were switched. Upon further digging I realized this wasn't causing a visible issue because the (back to) edit option that is available for ETH transactions is not available for token transactions (where this logic would be used). This PR adds back the edit navigation button for the send-token flow and fixes the faulty logic.

Manual testing steps:
- Start to send a token (on any network, and token that is not its native asset)
- Once at the confirm screen look for the edit button in the top left corner and click it
- Edit the transaction and click forward again to confirm.
- Send the transaction and verify that the transaction sent the correct amount to the correct address

### DIFF
--- a/ui/ducks/confirm-transaction/confirm-transaction.duck.js
+++ b/ui/ducks/confirm-transaction/confirm-transaction.duck.js
@@ -3,7 +3,7 @@ import {
   currentCurrencySelector,
   unconfirmedTransactionsHashSelector,
 } from '../../selectors';
-import { getNativeCurrency } from '../metamask/metamask';
+import { getNativeCurrency, getTokens } from '../metamask/metamask';
 
 import {
   getValueFromWeiHex,
@@ -25,6 +25,7 @@ const createActionType = (action) => `metamask/confirm-transaction/${action}`;
 
 const UPDATE_TX_DATA = createActionType('UPDATE_TX_DATA');
 const UPDATE_TOKEN_DATA = createActionType('UPDATE_TOKEN_DATA');
+const UPDATE_TOKEN_PROPS = createActionType('UPDATE_TOKEN_PROPS');
 const CLEAR_CONFIRM_TRANSACTION = createActionType('CLEAR_CONFIRM_TRANSACTION');
 const UPDATE_TRANSACTION_AMOUNTS = createActionType(
   'UPDATE_TRANSACTION_AMOUNTS',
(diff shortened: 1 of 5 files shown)


ui/ducks/send/send.js
```diff
@@ -570,7 +570,7 @@ const slice = createSlice({
       state.recipient.nickname = action.payload.nickname;
       state.draftTransaction.id = action.payload.id;
       state.draftTransaction.txParams.from = action.payload.from;
-      slice.caseReducers.updateDraftTransaction(state);
+      slice.caseReducers.validateSendState(state);
```
adonesky1, line 573: @brad-decker I banged my head against the wall here for a while and found that this is required for updating the validating and updating the state.status after modifying the state.stage here when navigating back to edit in a send-token flow. Let me know what you think?

ui/ducks/send/send.js
```diff
@@ -571,7 +571,11 @@ const slice = createSlice({
       state.recipient.nickname = action.payload.nickname;
       state.draftTransaction.id = action.payload.id;
       state.draftTransaction.txParams.from = action.payload.from;
-      slice.caseReducers.updateDraftTransaction(state);
+      if(state.asset.type === ASSET_TYPES.NATIVE){
```
adonesky1, line 574: @brad-decker can we take a look at this PR together sometime today? I've spent too much time trying to figure out whats going wrong. The navigate back to edit flow is behaving inconsistently for me

adonesky1:
@brad-decker with the fix you added for the gasLoading issue this PR appears to be working as I expect it to. Take a look when you have chance?

brad-decker (approved):
:100: works like a charm!

danjm (approved):

PR #11341, opened 2021-06-18
Add 'Transaction Added' metric event to TransactionController

ryanml:
Example Metrics payload for Eth transaction on Rinkeby:
Screen Shot 2021-06-17 at 9 42 22 PM

### DIFF
--- a/app/scripts/controllers/transactions/index.js
+++ b/app/scripts/controllers/transactions/index.js
@@ -202,8 +202,32 @@ export default class TransactionController extends EventEmitter {
   @emits ${txMeta.id}:unapproved
   */
   addTransaction(txMeta) {
+    const {
+      type,
+      status,
+      chainId,
+      origin: referrer,
+      txParams: { gasPrice },
+      metamaskNetworkId: network,
+    } = txMeta;
+    const source = referrer === 'metamask' ? 'user' : 'dapp';
+
     this.txStateManager.addTransaction(txMeta);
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/transactions/index.js
```diff
@@ -202,8 +202,36 @@ export default class TransactionController extends EventEmitter {
...
+      status,
+      chainId: id,
+      origin: referrer,
+      metamaskNetworkId,
+      txParams: { gasPrice },
+    } = txMeta;
+
+    const networkId = this.networkStore.getState();
+    const network = metamaskNetworkId || networkId;
+    const chainId = id || this._getCurrentChainId();
```
ryanml, line 216: @brad-decker let me know if these are the incorrect fallbacks for both network and chainId

brad-decker: If you just leave these undefined they'll be filled by the metametrics controller with the current values so you shouldn't need to get the current network/chainId here. :D

brad-decker: LGTM though other than that.

ryanml: 👍 will update per our call

ryanml: fixed

brad-decker (approved):
LGTM!

darkwing (approved):

PR #11343, opened 2021-06-18
Increase Jest unit test coverage for the Swaps feature to ~60%

dan437:
## Explanation
This PR increases Swaps to about 60%, which is getting us another step closer to at least 90% coverage.

## Manual testing steps
```npm run test:coverage:jest```

## Screenshots
image

### CODE REVIEW

File: ui/pages/swaps/swaps-gas-customization-modal/swaps-gas-customization-modal.container.test.js [hunk 1]
```diff
@@ -0,0 +1,34 @@
...
+      store,
+    );
+    expect(document.querySelector('.page-container__header')).toMatchSnapshot();
+    expect(getByText('Basic')).toBeInTheDocument();
+    expect(getByText('Advanced')).toBeInTheDocument();
+    expect(getByText('Estimated Processing Times')).toBeInTheDocument();
+    expect(getByText('Send Amount')).toBeInTheDocument();
+    expect(getByText('Transaction Fee')).toBeInTheDocument();
+    expect(
+      document.querySelector('.gas-modal-content__info-row__send-info'),
```
brad-decker, line 24: Are these meant to put *sub sections* of the component into snapshots? Also I'm not a fan of using querySelector to overcome the lack of class selectors in react testing library. They aren't included as a way of gracefully nudging us towards using more tests as the user would experience the app versus how we know it to be structured.

brad-decker (changes requested):
I have a question around the usage of querySelector and the snapshots, but otherwise looks good!

darkwing (dismissed):

ui/pages/swaps/swaps-gas-customization-modal/swaps-gas-customization-modal.container.test.js
dan437: Great comment! That's correct that I use it to get sub sections of the component into snapshots. I only use `querySelector` for snapshots, never for other assertions. Will check what react testing library recommends when it comes to snapshot testing.

brad-decker: >So, avoid huge snapshots and take smaller, more focused ones. While you're at it, see if you can actually change it from a snapshot to a more explicit assertion (because you probably can 😉).

Source: https://kentcdodds.com/blog/effective-snapshot-testing

```js
import { render } from '@testing-library/react'
import '@testing-library/jest-dom/extend-expect'

test('renders a message', () => {
  const { container, getByText } = render(<Greeting />)
  expect(getByText('Hello, world!')).toBeInTheDocument()
  expect(container.firstChild).toMatchInlineSnapshot(`
    <h1>Hello, World!</h1>
  `)
})'
```

source: https://testing-library.com/docs/react-testing-library/api/

(second source originally sent to me by @dan437 in slack) 

--- Follow up thoughts ---
I mostly rescind my feedback. I still do not like the usage of querySelector to bypass the lack of support for class name lookup... the example stated doesn't do that, either, but I get this bit of feedback (from the slack thread) from @dan437 as well:

> Basically, they are using container.firstChild to check a snapshot. In our case we want to check multiple things in a component and doing something like container.firstChild.firstChild etc. wouldn’t be so readable. Another option is to add the data-testid attribute to elements, but that sounded like a lot of additional work, which is not really needed if we can use classes for snapshots.

but by using the querySelector API we are sending a signal to all engineers that this is an accepted pattern and, although some might only use it for snapshot testing, it becomes the burden of the reviewer to catch these things when they are used inappropriately. 

#### Additional feedback 

```snapshot
exports[`GasCustomizationModalContainer renders the component with initial props 5`] = `
<div
  class="page-container__footer"
>
  <footer>
    <button
      class="button btn-secondary page-container__footer-button"
      data-testid="page-container-footer-next"
      disabled=""
      role="button"
      tabindex="0"
    >
      Save
    </button>
  </footer>
</div>
`;
```

what is the role of this snapshot? what is the bit that matters to this test? that it has a save button? If so then this belongs as a snapshot of the footer component, not the parent page? I'm maybe missing something with those these snapshots were chosen. 

Lets try to see if we can whittle down these cases where we need to cover a large component in sub snapshots so that we are only snapshotting the components themselves instead of creating small snapshots of this component many times in different places? The footer snapshot above shows up in a few places, each with different classes and disabled showing or not showing. If the goal here is to test that the button behaves as expected then we should use functional testing, not snapshot testing, to test that this one component is behaving as appropriate in the component.

dan437: > by using the querySelector API we are sending a signal to all engineers that this is an accepted pattern and, although some might only use it for snapshot testing, it becomes the burden of the reviewer to catch these things when they are used inappropriately

That's a good point that devs would need to know to only use it for snapshot testing and it can be tricky to enforce.

dan437: I've done a bit more research and the official documentation has a section about manual queries: https://testing-library.com/docs/queries/about/#manual-queries
> On top of the queries provided by the testing library, you can use the regular querySelector DOM API to query elements. Note that using this as an escape hatch to query by class or id is not recommended because they are invisible to the user. Use a testid if you have to, to make your intention to fall back to non-semantic queries clear and establish a stable API contract in the HTML.

I still feel that snapshot testing is beneficial, because otherwise we would need to write many more regular assertions to test a React component (e.g. to check that elements are nested correctly with the right classes / attributes). Even the creator of the react-testing-library sees value in snapshot testing: https://kentcdodds.com/blog/effective-snapshot-testing
> I'd suggest that there is value in them if you use them effectively
> Edit: Since I wrote this blog post, a new lint rule was added to eslint-plugin-jest called "no-large-snapshots" that could be pretty helpful in discouraging large snapshots

We already use the `no-large-snapshots` ESLint rule limited to 50 lines

So, I would say we can do this:
- If we are doing a snapshot for a tiny part of code, we can replace that snapshot with a regular assertion from the react-testing-library, e.g. by using `getByText`
- When our snapshot is bigger (but less than the current maximum of 50 lines per snapshot), we can add a `data-testid` attribute and then use the `getByTestId` selector instead of `document.querySelector`

What do you think, @brad-decker?

brad-decker (approved):

brad-decker:
Looks good to me, thank you, @dan437!

darkwing (approved):

PR #11345, opened 2021-06-18
Set account nickname to ENS when is available

### LINKED ISSUE #11273: Show ENS reverse-resolved name on account detail view
If a user has not changed their account nickname (it is not `\Account $[0-9]+\`), and an ENS reverse-resolved name is available for that account, then show the reverse-resolved name for that account.

hsy822:
Fixes: #11273 

Explanation:  If the user has not changed the account nickname and ENS is available, ENS reverse-resolved name is displayed on Account detail view instead of the nickname such as 'Account 1'. 

Manual testing steps:  
  - Import the account registered in ENS to Metamask
  - Set the alias of the account to a default value such as 'Account 1'
  - Go to  the Account detail view and check if the ENS value is loaded.


brad-decker:
hey @hsy822 rebasing on master should fix up your tests!

PR #11352, opened 2021-06-19
Updating client side metrics for EIP-1559 UI

ryanml:
- Adding 'Changed Gas Button' metrics event
- Adding gasEstimateType to 'Changed Gas Button' metrics event

Example metrics payload for 'Changed Gas Button':
Screen Shot 2021-06-21 at 11 38 03 AM


darkwing (dismissed):
LGTM!

brad-decker (changes requested):
I hate to delay us, but I think we should reason about what we would like the schema to be fore this event in the non-legacy tracking plan and update to use the new method/analytics setup.

brad-decker (approved):
b-e-a utiful.

darkwing (approved):

Issue #9419, opened 2020-09-16
Display on-chain failure reason on tx confirmation


danfinlay:
We currently show a warning when a transaction is likely to run out of gas, but we could also show the error text related to any transaction failure.

This probably wouldn’t require any additional design, and could simply be added into the current warning field.

https://twitter.com/smpalladino/status/1305991209257775105?s=12

danfinlay:
Come to think of it, there are two places we could use this:
- [ ] Warn users of possible failure on tx confirmation (and show reason as warning).
- [ ] Upon actual tx failure, display the on-chain reported reason in the tx history.

danfinlay:
Related to #10338

PR #11358, opened 2021-06-22
Adding tx metric events for Approved, Rejected, Submitted, and Finalized

ryanml:


app/scripts/controllers/transactions/index.js
```diff
@@ -1090,4 +1072,43 @@ export default class TransactionController extends EventEmitter {
       }
     }
   }
+
+  _trackTxEvent(txMeta, event, extraOpts = {}) {
```
brad-decker, line 1076: let's add some documentation to this method, and my request would be not to use the tx shorthand and not to hide the metrics portion of the 'event' considering events mean a few things in the controllers portion of our codebase ( something like trackTransactionMetricsEvent is just verbose enough to describe what we're doing to a newcomer. )

app/scripts/controllers/transactions/index.js
```diff
@@ -607,12 +584,13 @@ export default class TransactionController extends EventEmitter {
       // sign transaction
       const rawTx = await this.signTransaction(txId);
       await this.publishTransaction(txId, rawTx);
+      this._trackTxEvent(txMeta, 'Transaction Approved');
```
brad-decker, line 587: Let's lift the strings for these events into constants at the top of this file as its unlikely they'll need to be used/referenced outside of this.

brad-decker: so const TRANSACTION_APPROVED = 'Transaction Approved'

app/scripts/controllers/transactions/index.js
```diff
@@ -1090,4 +1072,43 @@ export default class TransactionController extends EventEmitter {
...
+    } = txMeta;
+    const source = referrer === 'metamask' ? 'user' : 'dapp';
+
+    this._trackMetaMetricsEvent({
+      event,
+      category: 'Transactions',
+      sensitiveProperties: {
+        type,
+        status,
+        gasPrice,
```
brad-decker, line 1093: gas_price, properties should be lower snake cased for events

app/scripts/controllers/transactions/index.js
```diff
@@ -1090,4 +1072,43 @@ export default class TransactionController extends EventEmitter {
...
+        referrer,
+        source,
+        network,
+        chain_id: chainId,
```
brad-decker, line 1097: let's add 'first_seen:' and use txMeta.time (i think) which is only set when the tx is first seen

app/scripts/controllers/transactions/index.js
```diff
@@ -695,6 +673,9 @@ export default class TransactionController extends EventEmitter {
     this.setTxHash(txId, txHash);
 
     this.txStateManager.setTxStatusSubmitted(txId);
+
+    const { gas } = txMeta.txParams;
+    this._trackTxEvent(txMeta, 'Transaction Submitted', { gas });
```
brad-decker, line 678: lets track this value as `gas_limit: gas` as its confusing by itself and we'll be using gas_limit exclusively sometime in the future

app/scripts/controllers/transactions/index.js
```diff
@@ -1090,4 +1072,43 @@ export default class TransactionController extends EventEmitter {
...
+        chain_id: chainId,
+        ...extraOpts,
+      },
+    });
+  }
+
+  _failTransaction(txId, error) {
+    this.txStateManager.setTxStatusFailed(txId, error);
+    const txMeta = this.txStateManager.getTransaction(txId);
+    this._trackTxEvent(txMeta, 'Transaction Finalized');
```
brad-decker, line 1106: lets go ahead and include the error message in the event extra opts here because this could be an on_chain error and I think that'll be really valuable in assessing how we might better serve our customers.

app/scripts/controllers/transactions/index.js
```diff
@@ -726,6 +707,7 @@ export default class TransactionController extends EventEmitter {
       };
       this.txStateManager.setTxStatusConfirmed(txId);
       this._markNonceDuplicatesDropped(txId);
+      this._trackTxEvent(txMeta, 'Transaction Finalized');
```
brad-decker, line 710: Let's pull some values out of the txReceipt such as gasUsed, and explore in that object because I'm pretty sure we can get the time it was included in the network. We don't want to include that exact time, but instead txMeta.submittedTime - txReceipt contains blockNumber and then you can use getBlockByNumber to get the timestamp of when the block was collated.

brad-decker: without going the extra mile of solving #9419, we should at least catch cases wher the txReceipt.status is '0x0' which is onChain, and track that case as a failure for metrics instead of as a success. We also track a metrics event for on chain failures in metamask-controller (check for looking at txReceipt.status ,and if you search for txReceipt?.status you'll find a few more cases of checking that in the UI). 

I think if we can just check the status and log the error as an event here we have fulfilled our duty to this story and we can revisit #9419 and the differences between on-chain and in-metamask failures later ;)

brad-decker (changes requested):
Solid work, hopefully these comments help! lemme know if you wanna jump on a call tomorrow

app/scripts/controllers/transactions/index.js
ryanml: function name updated and documentation added

app/scripts/controllers/transactions/index.js
ryanml: add a structure the top containing string values 👍

app/scripts/controllers/transactions/index.js
ryanml: done

app/scripts/controllers/transactions/index.js
ryanml: done

app/scripts/controllers/transactions/index.js
ryanml: sgtm, done

app/scripts/controllers/transactions/index.js
ryanml: good though! error message added

app/scripts/controllers/transactions/index.js
ryanml: `completion_time` is now factored in and included in the metrics payload here

app/scripts/controllers/transactions/index.js
```diff
@@ -727,6 +718,24 @@ export default class TransactionController extends EventEmitter {
...
+      const metricsParams = { gas_used: gasUsed };
+      const completionTime = await this._getTransactionCompletionTime(
+        blockNumber,
+        submittedTime,
+      );
+
+      if (completionTime) {
+        metricsParams.completion_time = completionTime;
+      }
+
```
brad-decker, line 732: Lets do one more clause here
```suggestion
if (txReceipt.status === '0x0') {
  metricsParams.status = 'failed on-chain'; ((overriding the status we set in that event))
  // metricsParams.error = TODO: figure out a way to get the on-chain failure reason
}
```

brad-decker (changes requested):
One little thing then we can land this bad boy

brad-decker (approved):

app/scripts/controllers/transactions/index.js
ryanml: done

app/scripts/controllers/transactions/index.js
```diff
@@ -727,6 +718,29 @@ export default class TransactionController extends EventEmitter {
...
+      if (txReceipt.status === '0x0') {
+        metricsParams.status = 'failed on-chain';
+        // metricsParams.error = TODO: figure out a way to get the on-chain failure reason
```
darkwing, line 735: Do we intend to have this kept in?

darkwing (approved):

app/scripts/controllers/transactions/index.js
ryanml: Yes this will be a followup after some exploration

brad-decker: Yeah its TODO to add later

PR #11361, opened 2021-06-22
Update controllers with conversionRate change with minimal required changes in extension

### LINKED ISSUE #9492: Custom network incorrect account balance when symbol is left blank.

adonesky1:
Fixes: #9492

Explanation:  My original [PR to fix this issue](https://github.com/MetaMask/metamask-extension/pull/11066) got, as @brad-decker gently pointed out to me, a bit out of hand. This PR aims to add the core functionality needed to land [the changes in the controller's repo](https://github.com/MetaMask/controllers/pull/465) for the extension - which set conversionRate to null when we don't have a valid conversion rate from the current network's native currency to the user's preferred currency (usually fiat).
 
Manual testing steps:
- In advanced settings, turn `Show Conversion on Testnet`s on.
- Go between mainnet, a custom rpc network where you have added a ticker symbol and a custom rpc network where you haven't added a ticker symbol.
- Check that conversions are correct for all assets across these different networks (for the network with no ticker symbol there should be no conversions).


ui/components/app/asset-list-item/asset-list-item.js
```diff
@@ -92,7 +92,7 @@ const AssetListItem = ({
 
   return (
     <ListItem
-      className={classnames('asset-list-item', className)}
+      className={classnames('asset-list-item', className, !secondary ? 'no-subtitle' : '')}
```
brad-decker: could we follow BEM syntax here so asset-list-item--no-subtitle (as a BEM modifier)

ui/components/ui/list-item/index.scss
```diff
@@ -110,3 +110,18 @@
       '.    actions  actions  actions  actions  mid mid mid mid right right right';
   }
 }
+
+.no-subtitle {
+  grid-template-areas:
+    'icon head     head     head     head     head     head     head     right right right right'
+    'icon head     head     head     head     head     head     head     right right right right'
+    'icon head     head     head     head     head     head     head     right right right right';
```
brad-decker, line 118: this solution is great, but we can't use the lack of secondary prop to nix the 'actions' row (this is reserved for speed up / cancel buttons on transaction/activity items that are pending).

So instead of no-subtitle, lets try something like 'single-content-row' and reduce the grid template areas to a single row. So if !children && !secondary instead of just !secondary

ui/components/app/asset-list-item/asset-list-item.js
```diff
@@ -92,7 +92,11 @@ const AssetListItem = ({
 
   return (
     <ListItem
-      className={classnames('asset-list-item', className)}
+      className={classnames(
+        'asset-list-item',
+        className,
+        secondary ? '' : 'no-subtitle',
```
brad-decker, line 98: Can we use BEM style here so 'asset-list-item--no-subtitle'

brad-decker (changes requested):
Feedback of this round is specific to the very great idea of changing the grid template under certain criteria.

ui/components/ui/list-item/index.scss
adonesky1: Yes this class is only being added to asset-list-item so I don't think it could cause the issue as implemented (no actions but I see your point and will make this change in case someone tries to use it in a different context!

brad-decker (dismissed):
Looks fantastic!

darkwing (dismissed):
Needs a rebase but this worked great!

brad-decker (approved):

adonesky1:
@darkwing  just rebased with Brad's changes onboard now

darkwing (approved):

PR #11363, opened 2021-06-23
Allowing custom rpc form submission when chainId is a duplicate

ryanml:
Fixes the following cases:
Screen Shot 2021-06-22 at 7 42 03 PM

### DIFF
--- a/ui/pages/settings/networks-tab/network-form/network-form.component.js
+++ b/ui/pages/settings/networks-tab/network-form/network-form.component.js
@@ -280,6 +280,7 @@ export default class NetworkForm extends PureComponent {
   }) {
     const { errors } = this.state;
     const { viewOnly } = this.props;
+    const errorMessage = errors[fieldKey]?.msg || '';
 
     return (
       <div className="networks-tab__network-form-row">
@@ -305,7 +306,7 @@ export default class NetworkForm extends PureComponent {
           margin="dense"
           value={value}
           disabled={viewOnly}
-          error={errors[fieldKey]}
+          error={errorMessage}
           autoFocus={autoFocus}
         />
       </div>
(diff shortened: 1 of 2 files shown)


ui/pages/settings/networks-tab/network-form/network-form.component.js
```diff
@@ -451,28 +499,26 @@ export default class NetworkForm extends PureComponent {
...
-        this.context.t(
-          this.isValidWhenAppended(url) ? 'urlErrorMsg' : 'invalidRPC',
-        ),
-      );
+      const errorKey = this.isValidWhenAppended(url)
+        ? 'urlErrorMsg'
+        : 'invalidRPC';
+      this.setErrorTo(stateKey, {
+        key: errorKey,
+        msg: t(errorKey),
```
ryanml, line 511: Note for tomorrow: Pass the key directly per our "pass string literals directly to `t(..)`" rule

danjm (dismissed):
LGTM

adonesky1:
@ryanml So I'm seeing some slightly strange UX here:

So I go to edit a network, and I'm allowed to change the chainID and click save:
Screen Shot 2021-06-23 at 9 34 00 AM

After hitting save, save button is disabled and you see warning message:
Screen Shot 2021-06-23 at 9 34 06 AM

Then you go back to edit it to the original chainID and you get both the error message and save is disabled.
Screen Shot 2021-06-23 at 9 34 12 AM

Really we never saved the new (incorrect) chainId so from a functional stand point it doesn't matter, but it could still be confusing UX.

brad-decker:
The behavior that @adonesky1 outlined is live on 9.6.1 so not introduced in this PR, so maybe address in another PR?

adonesky1:
> The behavior that @adonesky1 outlined is live on 9.6.1 so not introduced in this PR, so maybe address in another PR?

I thought this PR was partially intended to address this kindof behavior?

ryanml:
@adonesky1 - I specifically aimed to address the two items in the PR description regarding the duplicate chain id error. I'm happy to address the other issues in another PR following this. For the sake of addressing this particular regression and moving 9.7 forward, I think this should be merged as is (unless there is other feedback on behavior/code as it pertains to the duplicate chain id error)

adonesky1:
Ok, I think I was confused because I thought I was describing the second issue in the 9.7 regressions doc but I guess the issue I'm describing is different in that I save the incorrect chainId before attempting to revert.

ui/pages/settings/networks-tab/network-form/network-form.component.js
ryanml: fixed

adonesky1 (approved):
LGTM!

danjm (approved):

PR #11367, opened 2021-06-23
add maxPriorityFeePerGas and maxFeePerGas to incoming controller

brad-decker:
Explaination:

Adds new EIP-1559 fields while retaining gasPrice fields.

### DIFF
--- a/app/scripts/controllers/incoming-transactions.js
+++ b/app/scripts/controllers/incoming-transactions.js
@@ -34,8 +34,10 @@ const fetchWithTimeout = getFetchWithTimeout(SECOND * 30);
  * @typedef {Object} EtherscanTransaction
  * @property {string} blockNumber - The number of the block this transaction was found in, in decimal
  * @property {string} from - The hex-prefixed address of the sender
- * @property {string} gas - The gas limit, in decimal WEI
- * @property {string} gasPrice - The gas price, in decimal WEI
+ * @property {string} gas - The gas limit, in decimal GWEI
+ * @property {string} [gasPrice] - The gas price, in decimal GWEI
+ * @property {string} [maxFeePerGas] - The maximum fee per gas, inclusive of tip, in decimal GWEI
+ * @property {string} [maxPriorityFeePerGas] - The maximum tip per gas in decimal GWEI
  * @property {string} hash - The hex-prefixed transaction hash
  * @property {string} isError - Whether the transaction was confirmed or failed (0 for confirmed, 1 for failed)
  * @property {string} nonce - The transaction nonce, in decimal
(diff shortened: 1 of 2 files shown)


brad-decker:
adding tests to this

darkwing (dismissed):

app/scripts/controllers/incoming-transactions.js
```diff
@@ -267,21 +269,33 @@ export default class IncomingTransactionsController {
...
+      from: etherscanTransaction.from,
+      gas: bnToHex(new BN(etherscanTransaction.gas)),
+      nonce: bnToHex(new BN(etherscanTransaction.nonce)),
+      to: etherscanTransaction.to,
+      value: bnToHex(new BN(etherscanTransaction.value)),
+    };
+
+    if (etherscanTransaction.gasPrice) {
+      txParams.gasPrice = bnToHex(new BN(etherscanTransaction.gasPrice));
+    } else if (etherscanTransaction.maxFeePerGas) {
```
danjm, line 282: Do we at all need to consider the case where one of `maxFeePerGas` or `maxPriorityFeePerGas` are truthy and the other is not?

danjm: Also, `etherscanTransaction.maxFeePerGas` will never be of type number and equal to 0 right?

app/scripts/controllers/incoming-transactions.js
```diff
@@ -267,21 +269,33 @@ export default class IncomingTransactionsController {
...
+      txParams.maxFeePerGas = bnToHex(
+        new BN(etherscanTransaction.maxFeePerGas),
```
danjm, line 284: Should we specify the base in these `BN` calls?

... Although, that wasn't done before, so I guess this is fine.

app/scripts/controllers/incoming-transactions.js
brad-decker: A transaction isn't valid if it has one or the other, per the JSON-rpc spec it would be immediately rejected. 


https://github.com/ethereum/eth1.0-specs/blob/master/json-rpc/spec.json#L1709-L1730

So like, the new fields appear to be missing a 'type' parameter, also I just noticed it *does* specify WEI not GWEI i assume that we are then converting the gasPrice (also in wei) elsewhere in our code for display to the user?

app/scripts/controllers/incoming-transactions.js
brad-decker: I'm curious about this with the comment above regarding wei vs gwei.

app/scripts/controllers/incoming-transactions.js
danjm: > i assume that we are then converting the gasPrice (also in wei) elsewhere in our code for display to the user?
yes, we are

danjm: for example, in `metamask-extension/ui/components/app/transaction-breakdown/transaction-breakdown.component.js`

danjm: see the use of the `CurrencyDisplay` component

danjm (dismissed):

danjm (approved):
LGTM

darkwing (approved):

PR #11368, opened 2021-06-23
remove EIP-1559 fields from txParams before calling estimateGas

brad-decker:
Depends on: #11367 
Explanation:  
Also removes maxFeePerGas and maxPriorityFeePerGas from txParams before estimating gas. Though the JSON RPC spec for EIP-1559 says 'oneOf' the two variations of gas fields should be included, but 'gasPrice' is not required on legacy. I'm getting some validation from geth devs that this will still work. 

Micah essentially confirmed this is correct, supporting screenshots below

<details>
<summary>Question asked</summary>

image

</details>

<details>
<summary>Response

</summary>

Screen Shot 2021-06-24 at 9 33 18 AM

Redacted portions are from an unrelated response. 

</details>


onie92 (dismissed):

NiranjanaBinoy (approved):
LGTM

adonesky1 (approved):
LGTM

PR #11369, opened 2021-06-24
add method to detect EIP 1559 support

brad-decker:
Adds getEIP1559Compatibility method to network controller that detects EIP-1559 support by checking the baseFeePerGas property on blockHeader. Fully wired up and ready for consumption (Properly detects that Ropsten has support and providers a selector for UI work)

### DIFF
--- a/app/scripts/controllers/network/network.js
+++ b/app/scripts/controllers/network/network.js
@@ -375,7 +375,7 @@ export default class NetworkController extends EventEmitter {
   }
 
   _switchNetwork(opts) {
-    // Indicate to subsribers that network is about to change
+    // Indicate to subscribers that network is about to change
     this.emit(NETWORK_EVENTS.NETWORK_WILL_CHANGE);
     // Set loading state
     this.setNetworkState('loading');
--- a/app/scripts/controllers/network/network.js
+++ b/app/scripts/controllers/network/network.js
@@ -83,9 +83,11 @@ export default class NetworkController extends EventEmitter {
     // will require a decent sized refactor of how we're accessing network
     // state. Currently this is only used for detecting EIP 1559 support but
     // can be extended to track other network details.
-    this.networkDetails = new ObservableStore({
-      ...defaultNetworkDetailsState,
-    });
+    this.networkDetails = new ObservableStore(
+      opts.networkDetails || {
+        ...defaultNetworkDetailsState,
+      },
+    );
     this.store = new ComposedStore({
       provider: this.providerStore,
       previousProviderStore: this.previousProviderStore,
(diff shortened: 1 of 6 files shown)


app/scripts/controllers/network/network.js
```diff
@@ -120,6 +135,42 @@ export default class NetworkController extends EventEmitter {
...
+          return resolve(block);
+        },
+      );
+    });
+  }
+
+  /**
+   * Method to check if the block header contains fields that indicate EIP 1559
+   * support (baseFeePerGas).
+   * @returns {boolean} true if current network supports EIP 1559
```
adonesky1, line 161: NIT: should be returns {Promise<boolean>}

app/scripts/controllers/network/network.js
```diff
@@ -120,6 +135,42 @@ export default class NetworkController extends EventEmitter {
...
+   */
+  async getEIP1559Compatibility() {
```
adonesky1, line 163: Wondering if, since you've made the networkDetails store quite general to all EIPs, whether we should have a method that is abstracted away from the particular EIP we're checking against. Obviously this is the important one right now, so not a big deal, just a thought given the open-endedness of the networkDetails store shape.

app/scripts/controllers/network/network.js
```diff
@@ -298,9 +375,15 @@ export default class NetworkController extends EventEmitter {
   }
 
   _switchNetwork(opts) {
+    // Indicate to subsribers that network is about to change
```
darkwing, line 378: Typo -- subscribers

darkwing (dismissed):
I really like the way you've done this.  Really tiny nit in text.

One thing to consider, if we need to do a lot of EIP testing, is creating a generic object with an EIP number and a function to test support.  Since we only have one now, it's probably not necessary.

app/scripts/controllers/network/network.js
brad-decker: I think we'd need specific implementations for each one, this function is checking for a specific field on a block header...other EIPs will have other detection methods, there is a generic method for *saving* support for an EIP and one for clearing all support.

brad-decker:
Yeah good call, we can expand this implementation in the near future

app/scripts/controllers/network/network.js
adonesky1: yeah makes sense. It wouldn't make any sense until there are more EIPs to check against, but I was thinking something more like a reducer with cases for different EIPs

adonesky1: But definitely not a REQ here.

adonesky1 (dismissed):
One tiny nit outstanding but otherwise LGTM

app/scripts/controllers/network/network.js
brad-decker: Yeah we'd need to think that one through a bit, but not opposed to it long term.

brad-decker:
@adonesky1 fixed that type issue, and the typo from @darkwing

adonesky1 (approved):

darkwing (approved):

PR #11374, opened 2021-06-24
account tests using fixtures

PeterYinusa:
Related: #11136

Explanation:  

- moved add & lock and unlock account test to separate spec
- updated add account test to use fixture
- updated lock and unlock test to use fixture


Gudahtt (approved):
LGTM!

adonesky1 (approved):
LGTM 👍

PR #11380, opened 2021-06-24
add account eip1559 detection support

brad-decker:
check to determine if the keyring for the currently selected account is a trezor or ledger device and returns false if so, until such a time as those devices are updated to support EIP-1559

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -73,6 +73,16 @@ export const METAMASK_CONTROLLER_EVENTS = {
   UPDATE_BADGE: 'updateBadge',
 };
 
+/**
+ * Accounts can be instantiated from simple, HD or the two hardware wallet
+ * keyring types. Both simple and HD are treated as default but we do special
+ * case accounts managed by a hardware wallet.
+ */
+const KEYRING_TYPES = {
+  LEDGER: 'Ledger Hardware',
+  TREZOR: 'Trezor Hardware',
+};
+
 export default class MetamaskController extends EventEmitter {
   /**
    * @constructor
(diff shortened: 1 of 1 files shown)


darkwing (dismissed):

NiranjanaBinoy (dismissed):
LGTM

app/scripts/metamask-controller.js
```diff
@@ -1934,6 +1934,21 @@ export default class MetamaskController extends EventEmitter {
...
+   * TODO: remove this method when Ledger and Trezor release their supported
+   * client utilities for EIP-1559
+   * @returns {boolean} true if the keyring type supports EIP-1559
+   */
+  getCurrentAccountEIP1559Compatibility() {
+    const selectedAddress = this.preferencesController.getSelectedAddress();
+    const keyring = this.keyringController.getKeyringForAccount(
+      selectedAddress,
+    );
+    return keyring.type !== 'ledger' && keyring.type !== 'trezor';
```
adonesky1, line 1949: Didn't test this locally so not sure, but should these strings be 'Ledger Hardware' and 'Trezor Hardware'?

brad-decker: oh snap, yeah you're right! I was conflating deviceName with keyring type.

darkwing: Good catch @adonesky1 !

brad-decker:
added a commit to move these to a constant and used constants for the instances of Ledger Hardware and Trezor Hardware

adonesky1 (approved):
LGTM!

darkwing (approved):

PR #11381, opened 2021-06-24
remove unused sufficient balance method

brad-decker:
this method relied on gasPrice being present and is safe to remove


ryanml (approved):
++

darkwing (approved):
A project search concludes we indeed do not use this anywhere. ++

PR #11382, opened 2021-06-24
Add methods to easily detect transaction type based on gas fields

brad-decker:
Helper methods that will be used in future PRs to determine how to handle a transaction.

### DIFF
--- a/shared/modules/transaction.utils.js
+++ b/shared/modules/transaction.utils.js
@@ -1,6 +1,40 @@
+import { isHexString } from 'ethereumjs-util';
+
 export function transactionMatchesNetwork(transaction, chainId, networkId) {
   if (typeof transaction.chainId !== 'undefined') {
     return transaction.chainId === chainId;
   }
   return transaction.metamaskNetworkId === networkId;
+}
+
+/**
+ * Determines if the maxFeePerGas and maxPriorityFeePerGas fields are supplied
+ * and valid inputs. This will return false for non hex string inputs.
+ * @param {import("../constants/transaction").TransactionMeta} transaction -
+ *  the transaction to check
+ * @returns {boolean} true if transaction uses valid EIP1559 fields
+ */
+export function isEIP1559Transaction(transaction) {
+  return (
+    isHexString(transaction.txParams.maxFeePerGas) &&
+    isHexString(transaction.txParams.maxPriorityFeePerGas)
+  );
+}
+
+/**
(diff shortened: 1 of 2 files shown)


darkwing (dismissed):
We might want to provide a reworded commit message, but the code looks good.

brad-decker:
@darkwing -- I'm open to rewording. What do you see as the issue with current one and what would you suggest?

darkwing:
Just added some capitalization to the beginning 🆙

shared/modules/transaction.utils.js
```diff
@@ -1,6 +1,40 @@
...
+ * @param {import("../constants/transaction").TransactionMeta} transaction -
+ *  the transaction to check
+ * @returns {boolean} true if transaction uses valid Legacy fields OR lacks
+ *  EIP1559 fields
+ */
+export function isLegacyTransaction(transaction) {
+  return (
+    typeof transaction.txParams.maxFeePerGas === 'undefined' &&
+    typeof transaction.txParams.maxPriorityFeePerGas === 'undefined' &&
+    (typeof transaction.txParams.gasPrice === 'undefined' ||
```
adonesky1, line 37: Forgive my ignorance but when would gasPrice be 'undefined' in a legacy transaction?

shared/modules/transaction.utils.test.js
```diff
@@ -0,0 +1,83 @@
...
+      expect(
+        isEIP1559Transaction({
+          txParams: { maxFeePerGas: '0x1' },
+        }),
+      ).toBe(false);
+    });
+  });
+
+  describe('isLegacyTransaction', function () {
+    it('should return true if both maxFeePerGas and maxPriorityFeePerGas are not supplied', () => {
```
adonesky1, line 41: maybe 'should return true if maxFeePerGas, maxPriorityFeePerGas and gasPrice are not supplied'?

shared/modules/transaction.utils.js
brad-decker: when calling eth_estimateGas you can choose to remove gasPrice so that it doesn't throw insufficient balance errors.

shared/modules/transaction.utils.test.js
brad-decker: reworded to 'return true if no gas related fields are supplied'

adonesky1 (approved):
LGTM

NiranjanaBinoy (approved):
LGTM

PR #11387, opened 2021-06-25
Call Wyre’s API via our backend to generate Wyre’s Checkout URL

dan437:
## Explanation
Call Wyre’s API via our backend to generate Wyre’s Checkout URL. It's a strongly recommended way how to create their Checkout URL, because it has some additional query params that we don't currently have.

## Manual testing steps
  - Open the MetaMask browser extension
  - Click on `Buy` on the main page
  - Click on `Continue to Wyre` on the `Deposit Ether` Page
  - Check that a new tab is opened with Wyre's Checkout URL, has prefilled ETH address and there is a query param called `reservation`. If it's not there, it means a fallback URL was used instead. 

## Screenshots
`Deposit Ether` Page:
image


app/scripts/lib/buy-eth-url.test.js
```diff
@@ -12,7 +13,7 @@ describe('buy-eth-url', function () {
   const mainnet = {
     chainId: MAINNET_CHAIN_ID,
     amount: 5,
-    address: '0x0dcd5d886577d5081b0c52e242ef29e70be3e7bc',
+    address: '0x0dcd5d886577d5581b0c524242ef2ee70be3e7bc',
```
darkwing, line 16: Tiny suggestion but this address appears all over this test file -- might want to make it a constant and use that throughout.

darkwing (dismissed):

adonesky1 (dismissed):
LGTM!

brad-decker (approved):

darkwing (approved):

PR #11392, opened 2021-06-25
Create ChipWithInput Component

adonesky1:
Explanation:  Creates `ChipWithInput` wrapper of `Chip` component. We would like to reuse the `Chip` component for Backup Confirmation of Recovery Phrase stage of the new onboarding flow:
Screen Shot 2021-06-25 at 3 25 21 PM

Storybook:
Screen Shot 2021-06-25 at 3 28 44 PM

Screen Shot 2021-06-25 at 3 29 50 PM

### DIFF
--- a/ui/components/ui/chip/chip-with-input.js
+++ b/ui/components/ui/chip/chip-with-input.js
@@ -1,0 +1,37 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import { COLORS } from '../../../helpers/constants/design-system';
+import Chip from '.';
+
+export function ChipWithInput({
+  className,
+  borderColor = COLORS.UI1,
+  inputValue,
+  setInputValue,
+}) {
+  return (
+    <Chip
+      className={classnames(className, 'chip--with-input')}
+      borderColor={borderColor}
+    >
+      {setInputValue && (
+        <input
+          type="text"
+          className="chip__input"
+          onChange={(e) => {
+            setInputValue(e.target.value);
+          }}
+          value={inputValue}
+        />
+      )}
+    </Chip>
+  );
+}
+
+ChipWithInput.propTypes = {
+  borderColor: PropTypes.oneOf(Object.values(COLORS)),
+  className: PropTypes.string,
+  inputValue: PropTypes.string,
+  setInputValue: PropTypes.func,
+};
--- a/ui/components/ui/chip/chip.js
+++ b/ui/components/ui/chip/chip.js
@@ -63,4 +63,6 @@ Chip.propTypes = {
   rightIcon: PropTypes.node,
   className: PropTypes.string,
   onClick: PropTypes.func,
+  inputValue: PropTypes.string,
+  setInputValue: PropTypes.func,
 };
--- a/ui/components/ui/chip/chip.scss
+++ b/ui/components/ui/chip/chip.scss
@@ -37,7 +37,20 @@
     }
   }
 
+  &--with-input &__input {
+    direction: ltr;
+    border: none;
+    background: transparent;
+    text-align: center;
 
+    &:focus {
+      text-align: left;
+    }
+
+    &:focus-visible {
+      outline: none;
+    }
+  }
 
   &--with-right-icon {
     padding-right: 4px;
--- a/ui/components/ui/chip/chip.stories.js
+++ b/ui/components/ui/chip/chip.stories.js
@@ -1,10 +1,11 @@
 /* eslint-disable react/prop-types */
 
-import React from 'react';
+import React, { useState } from 'react';
 import { select, text } from '@storybook/addon-knobs';
 import { COLORS, TYPOGRAPHY } from '../../../helpers/constants/design-system';
 import ApproveIcon from '../icon/approve-icon.component';
 import Identicon from '../identicon/identicon.component';
+import { ChipWithInput } from './chip-with-input';
 import Chip from '.';
 
 export default {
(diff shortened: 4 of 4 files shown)


ui/components/ui/chip/chip.js
```diff
@@ -14,7 +14,10 @@ export default function Chip({
   leftIcon,
   rightIcon,
   onClick,
+  isInput,
```
brad-decker, line 17: I think you'll likely want the value to be accessible outside of this chip, so might I suggest:

```suggestion
inputValue,
setInputValue,
```

ui/components/ui/chip/chip.js
```diff
@@ -14,7 +14,10 @@ export default function Chip({
...
 }) {
+  const [inputValue, setInputValue] = useState('');
```
brad-decker, line 19:
```suggestion

```

ui/components/ui/chip/chip.js
```diff
@@ -48,6 +51,16 @@ export default function Chip({
         </Typography>
       )}
       {rightIcon && <div className="chip__right-icon">{rightIcon}</div>}
+      {isInput && (
```
brad-decker, line 54:
```suggestion
      {inputValue && (
```

ui/components/ui/chip/chip.js
```diff
@@ -48,6 +51,16 @@ export default function Chip({
...
+        <input
+          type="text"
+          className="chip__input"
+          onChange={(e) => {
+            setInputValue(e.target.value);
```
brad-decker, line 59:
```suggestion
            setInputValue && setInputValue(e.target.value);
```

ui/components/ui/chip/chip.js
```diff
@@ -63,4 +76,5 @@ Chip.propTypes = {
   rightIcon: PropTypes.node,
   className: PropTypes.string,
   onClick: PropTypes.func,
+  isInput: PropTypes.bool,
```
brad-decker, line 79:
```suggestion
  inputValue: PropTypes.string,
  setInputValue: PropTypes.function,
```

ui/components/ui/chip/chip.stories.js
```diff
@@ -80,3 +80,7 @@ export const WithBothIcons = () => (
     }
   />
 );
+
+export const AsInput = () => (
+  <Chip isInput borderColor={select('borderColor', COLORS, COLORS.UI3)} />
```
brad-decker, line 85:
```suggestion
  <Chip inputValue={inputValue} setInputvalue={setInputValue} borderColor={select('borderColor', COLORS, COLORS.UI3)} />
```

ui/components/ui/chip/chip.stories.js
```diff
@@ -80,3 +80,7 @@ export const WithBothIcons = () => (
     }
   />
 );
+
```
brad-decker, line 83:
```suggestion
  const [inputValue, setInputValue] = useState('');
```

brad-decker (changes requested):
I probably overdid it with the suggestions, and it might fail lint if you accept them all, so feel free to accept these or make the changes locally yourself. This is awesome but we'll want to track the value of the input outside of the Chip so we need to be able to pass the value in and get a update when it changes.

adonesky1:
> I probably overdid it with the suggestions, and it might fail lint if you accept them all, so feel free to accept these or make the changes locally yourself. This is awesome but we'll want to track the value of the input outside of the Chip so we need to be able to pass the value in and get a update when it changes.

Yeah I originally had it like this and then made it stateful with the intention to pass in the setters/value when it was used. But that was dumb...

brad-decker:
Actually now that I'm thinking about it, we already have a children prop on this component, considering that the styles are not at all required to be on the chip itself, why not do something like the following:

```
const ChipWithInput = () => {
  return (
    <Chip>
      {... your input related code}
    </Chip>
   );
```

That way we don't overcomplicate the Chip Component? Just a suggestion

ui/components/ui/chip/chip-with-input.js
```diff
@@ -0,0 +1,33 @@
...
+import { COLORS } from '../../../helpers/constants/design-system';
+
+export function ChipWithInput({
+  className,
+  borderColor = COLORS.UI1,
+  inputValue,
+  setInputValue,
+}) {
+  return (
+    <Chip className={className} borderColor={borderColor}>
```
brad-decker, line 13: [req] lets add 'chip--with-input' to className, just in case we need to modify any styles in the future from the main component.

ui/components/ui/chip/index.js
```diff
@@ -1 +1 @@
-export { default } from './chip';
+export { default } from './chip';
```
brad-decker, line 1: unnecessary change?

brad-decker (dismissed):
LGTM, just a nit

ui/components/ui/chip/index.js
adonesky1: fixed 💯

ui/components/ui/chip/chip.scss
```diff
@@ -37,7 +37,20 @@
     }
   }
 
-
+  &__input {
```
brad-decker, line 40: one more minor change req:

make this &--with-input &__input {

just so that the styles are specific to the ChipWithInput component

brad-decker (changes requested):
sorry one minor suggestion on making style slightly more specific.

brad-decker (approved):

ryanml (approved):
lgtm!

Issue #560, opened 2016-08-19
Create unit tests for balances.


Zanibas:

flyswatter:
Aw, I didn't think it was *that* big a deal, it's a nice thing to have going forward.  But, if you want to, there should be some really quick ones to write.

PR #11398, opened 2021-06-27
Integrating the TokenListController to Extension

NiranjanaBinoy:
Adding the TokeListcontroller to Extension code


app/scripts/controllers/detect-tokens.js
```diff
@@ -41,7 +44,7 @@ export default class DetectTokensController {
     if (this._network.store.getState().provider.chainId !== MAINNET_CHAIN_ID) {
       return;
     }
-
+    console.log(this._tokenList)
```
jasonzhouu, line 47: maybe `import log from 'loglevel'; log.info(this._tokenList);` will be better.

app/scripts/controllers/detect-tokens.js
```diff
@@ -175,7 +223,16 @@ export default class DetectTokensController {
       }
     });
   }
-
+  /**
+   * @type {Object}
```
jasonzhouu, line 227: `@type {Array}`

ui/helpers/utils/icon-factory.js
```diff
@@ -49,18 +52,19 @@ IconFactory.prototype.generateNewIdenticon = function (address, diameter) {
...
-    contractMap[address].logo
+    tokenList[address].iconUrl
   );
 }
 
-function imageElFor(address) {
-  const contract = contractMap[address];
-  const fileName = contract.logo;
-  const path = `images/contract/${fileName}`;
+function imageElFor(address, useStaticTokenList, tokenList) {
```
danjm, line 63: Question: if `useStaticTokenList` is false, should we continue to use `contractMap`?

danjm: Oops, I mean, if it is true?

danjm: Oh, wait, I guess if `useStaticTokenList` is `true`, then `tokenList` will be the list from `eth-contract-metadata`?

danjm: Because `tokenList` is managed by the controller, and we import `eth-contract-metadata` there, and make a decision about whether to use it in the controller.

app/scripts/controllers/detect-tokens.js
```diff
@@ -91,6 +91,54 @@ export default class DetectTokensController {
...
   }
 
+  /**
+   * For each token in tokenlist ptovided by the TokenListController, find check selectedAddress balance.
+   */
+  async detectNewTokens() {
+    if (!this.isActive) {
+      return;
+    }
+    if (
```
danjm, line 101: Are binance and mainnet the only two chains that the new token detection api supports? And is there a way to determine supported chains dynamically?

app/scripts/controllers/detect-tokens.js
```diff
@@ -24,24 +23,25 @@ export default class DetectTokensController {
...
     this.network = network;
     this.keyringMemStore = keyringMemStore;
+    this.tokenList = tokenList;
   }
 
   /**
    * For each token in @metamask/contract-metadata, find check selectedAddress balance.
-   */
+   *
   async detectNewTokens() {
```
danjm, line 38: I guess this old `detectNewTokens` can be deleted now?

ui/helpers/utils/icon-factory.js
danjm: > Oh, wait, I guess if useStaticTokenList is true, then tokenList will be the list from eth-contract-metadata?

hmm... `useStaticTokenList` is set in the preferences controller, and in the token list controller... I need to think about this :thinking:

NiranjanaBinoy: `useStaticTokenList` is set in the preferences controller, and it is passed on to the TokenListController during its initiation along with an `onPreferencesStateChange` method.

ui/helpers/utils/token-util.js
```diff
@@ -48,15 +40,22 @@ async function getDecimalsFromContract(tokenAddress) {
   }
 }
 
-function getContractMetadata(tokenAddress) {
-  return tokenAddress && casedContractMap[tokenAddress.toLowerCase()];
+function getTokenMetadata(tokenAddress, tokenList) {
+  const casedTokenList = Object.keys(tokenList).reduce((acc, base) => {
+    console.log(base);
```
danjm, line 45: Remove this console.log

ui/pages/add-token/token-search/token-search.component.js
```diff
@@ -1,27 +1,9 @@
 import React, { Component } from 'react';
```
danjm, line 1: :+1:

ui/pages/confirm-transaction-base/confirm-transaction-base.container.js
```diff
@@ -102,9 +96,19 @@ const mapStateToProps = (state, ownProps) => {
   const { name: fromName } = identities[fromAddress];
   const toAddress = propsToAddress || txParamsToAddress;
 
+  const tokenList = getTokenList(state);
+  const useStaticTokenList = getUseStaticTokenList(state);
```
danjm, line 100: Is the casing of the static and dynamic lists returned from the controller different?

NiranjanaBinoy: The addresses in `contract-metadata` are checksum addresses and the addresses we get from the API are all lower case.

app/scripts/controllers/detect-tokens.js
```diff
@@ -24,36 +23,55 @@ export default class DetectTokensController {
...
+    tokenList,
   } = {}) {
     this.preferences = preferences;
     this.interval = interval;
     this.network = network;
     this.keyringMemStore = keyringMemStore;
+    this.tokenList = tokenList;
+  }
+
+  async _getTokenBalances(tokens) {
```
danjm, line 35: Note to self: this is the same as before, just moved within the file

app/scripts/controllers/detect-tokens.js
```diff
@@ -24,36 +23,55 @@ export default class DetectTokensController {
...
   /**
-   * For each token in @metamask/contract-metadata, find check selectedAddress balance.
+   * For each token in tokenlist ptovided by the TokenListController, find check selectedAddress balance.
    */
   async detectNewTokens() {
     if (!this.isActive) {
       return;
     }
-    if (this._network.store.getState().provider.chainId !== MAINNET_CHAIN_ID) {
+    console.log(this._tokenList.state.tokenList);
```
danjm, line 56: Remove this console.log

app/scripts/controllers/detect-tokens.js
```diff
@@ -24,36 +23,55 @@ export default class DetectTokensController {
...
+    if (
+      this._network.store.getState().provider.chainId !== MAINNET_CHAIN_ID &&
+      this._network.store.getState().provider.chainId !== '0x38'
```
danjm, line 59: I might have asked this question elsewhere: are we only supporting token detection on mainnet and binance to start?

danjm: Also, probably helpful to replace `'0x38'` with a descriptive constant like `MAINNET_CHAIN_ID`

app/scripts/metamask-controller.js
```diff
@@ -16,7 +16,6 @@ import TrezorKeyring from 'eth-trezor-keyring';
 import LedgerBridgeKeyring from '@metamask/eth-ledger-bridge-keyring';
```
danjm, line 16: :+1:

ui/helpers/utils/token-util.js
```diff
@@ -80,23 +78,12 @@ async function getDecimals(tokenAddress) {
   return decimals;
 }
 
-export async function getSymbolAndDecimals(tokenAddress, existingTokens = []) {
```
danjm, line 83: Was `existingTokens` not used anywhere?

ui/hooks/useTokensToSearch.js
```diff
@@ -9,6 +9,8 @@ import {
   getCurrentCurrency,
```
danjm, line 9: I remember you told me you were encountering some problem with this file. What was it again?

danjm:
When is `setUseStaticTokenList` going to be called? Is that going to be added in another PR?

app/scripts/controllers/detect-tokens.js
NiranjanaBinoy: I am unsure of that so I changed that to check if the `tokenList` is empty.

app/scripts/controllers/detect-tokens.js
NiranjanaBinoy: I am unsure of that so I changed that to check if the `tokenList` is empty.

ui/helpers/utils/token-util.js
NiranjanaBinoy: `getSymbolAndDecimals` is used only in two locations in `tokenInfoGetter` in `tokens-util` and `getTokenParams` in `actions` and both of them seems to be returning the token info if it is an existingToken.

app/scripts/controllers/detect-tokens.js
```diff
@@ -24,36 +22,51 @@ export default class DetectTokensController {
...
     }
-    if (this._network.store.getState().provider.chainId !== MAINNET_CHAIN_ID) {
+    if (Object.keys(this._tokenList.state.tokenList).length <= 0) {
       return;
     }
 
     const tokensToDetect = [];
     this.web3.setProvider(this._network._provider);
-    for (const contractAddress in contracts) {
+    const apiTokens = this._tokenList.state.tokenList;
```
adonesky1, line 61: I feel like this variable should just be `tokenList`. I think you're trying to call out that these tokens are coming from the tokenList API as opposed to our static list but I think this is more confusing?

app/scripts/controllers/detect-tokens.js
```diff
@@ -138,9 +142,13 @@ export default class DetectTokensController {
       });
       this.hiddenTokens = hiddenTokens;
     });
-    preferences.store.subscribe(({ selectedAddress }) => {
-      if (this.selectedAddress !== selectedAddress) {
+    preferences.store.subscribe(({ selectedAddress, useStaticTokenList }) => {
+      if (
```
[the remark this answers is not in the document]
adonesky1: No I think this is right. If the selectedAddress changes the tokens we should detect will change too because you may have balances for different tokens across different addresses

app/scripts/metamask-controller.js
```diff
@@ -1303,8 +1306,9 @@ export default class MetamaskController extends EventEmitter {
             ? accountTokens[address][chainId].filter(
                 ({ address: tokenAddress }) => {
                   const checksumAddress = toChecksumHexAddress(tokenAddress);
-                  return contractMap[checksumAddress]
-                    ? contractMap[checksumAddress].erc20
+                  return this.preferencesController.store.getState()
+                    .useStaticTokenList && tokenList[checksumAddress]
+                    ? tokenList[checksumAddress].erc20
```
adonesky1, line 1311: I'm not understanding why we check the useStaticTokenList preference here?

NiranjanaBinoy: At the moment, I am also wondering if we need filtering for erc20 at this level with the current tokenList changes because we are filtering out the tokens that are not erc20 in TokenListController itself. 
@danjm Can I get your input on if we can skip the filtering and directly use `accountTokens[address][chainId]`??!!

ui/hooks/useTokensToSearch.js
NiranjanaBinoy: The application was crashing when I add shuffling to the tokenList Data, I will leave a comment at the line of code.

ui/hooks/useTokensToSearch.js
```diff
@@ -83,7 +90,12 @@ export function useTokensToSearch({ usersTokens = [], topTokens = {} }) {
   const conversionRate = useSelector(getConversionRate);
   const currentCurrency = useSelector(getCurrentCurrency);
   const defaultSwapsToken = useSelector(getSwapsDefaultToken);
-
+  const tokenList = useSelector(getTokenList);
+  const useStatictokenList = useSelector(getUseStaticTokenList);
+  // TODO: replace contract-metadata with list from state and introduce shuffle to tokenlist
+  const shuffledTokenList = useStatictokenList
+    ? shuffledContractMap
+    : Object.values(tokenList);
```
NiranjanaBinoy, line 98: When I added shuffling to `Object.values(tokenList)` here, the application started crashing.

app/scripts/controllers/detect-tokens.js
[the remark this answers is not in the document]
danjm: @adonesky1 fyi: this originally was `&&`, and was changed in this commit https://github.com/MetaMask/metamask-extension/commit/bbbb2b056e1ea2c5dc12cad83e0551e3ef0cd0ed

My original question was asked when it was `&&`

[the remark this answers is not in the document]
danjm: So I agree with your assessment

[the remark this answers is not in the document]
adonesky1: Ah ok! 👍

test/e2e/webdriver/index.js
```diff
@@ -67,6 +73,7 @@ async function setupFetchMocking(driver) {
       return { cancel: true };
     }
     function clone(obj) {
+      console.log(`clone`);
```
adonesky1, line 76: remove?

danjm:
I added a commit which will fix e2e tests once we upgrade to controllers v14.2.0

NiranjanaBinoy:
> When is `setUseStaticTokenList` going to be called? Is that going to be added in another PR?

The changes in the preferences controller are already added, I am planning to add the remaining changes to call the `setUseStaticTokenList with the PR for settings change.

app/scripts/metamask-controller.js
NiranjanaBinoy: Removed the filtering as the` tokenList` from the controller holds only erc20 tokens.

app/scripts/controllers/detect-tokens.js
```diff
@@ -24,71 +22,80 @@ export default class DetectTokensController {
...
   /**
-   * For each token in @metamask/contract-metadata, find check selectedAddress balance.
+   * For each token in tokenlist ptovided by the TokenListController, find check selectedAddress balance.
    */
   async detectNewTokens() {
     if (!this.isActive) {
       return;
     }
-    if (this._network.store.getState().provider.chainId !== MAINNET_CHAIN_ID) {
+    if (Object.keys(this._tokenList.state.tokenList).length <= 0) {
```
adonesky1, line 55: [nit] could this ever be less than 0?

app/scripts/controllers/detect-tokens.js
```diff
@@ -24,71 +22,80 @@ export default class DetectTokensController {
...
     this.web3.setProvider(this._network._provider);
-    for (const contractAddress in contracts) {
+    const { tokenList } = this._tokenList.state;
+    for (const tokenAddress in tokenList) {
       if (
-        contracts[contractAddress].erc20 &&
-        !this.tokenAddresses.includes(contractAddress.toLowerCase()) &&
-        !this.hiddenTokens.includes(contractAddress.toLowerCase())
+        !this.tokenAddresses.includes(tokenAddress.toLowerCase()) &&
+        !this.hiddenTokens.includes(tokenAddress.toLowerCase())
```
adonesky1, line 65: I can do this when I rebase with my PR but I think these should be case insensitive checks?

app/scripts/controllers/preferences.js
```diff
@@ -51,7 +51,11 @@ export default class PreferencesController {
       useNonceField: false,
       usePhishDetect: true,
       dismissSeedBackUpReminder: false,
-      useStaticTokenList: false,
+
+      // contradictory to the name the useStaticTokenList
+      // set to true means the dynamic list from the API is being used
+      // set to false will be using the static list from contract-metadata
+      useStaticTokenList: true,
```
adonesky1, line 58: I've mentioned this in a previous comment but I really think we should switch this around so that useStaticTokenList actually corresponds to using the static token list rather than inverting it...

app/scripts/metamask-controller.js
```diff
@@ -238,7 +237,7 @@ export default class MetamaskController extends EventEmitter {
     });
     this.tokenListController = new TokenListController({
       chainId: hexToDecimal(this.networkController.getCurrentChainId()),
-      useStaticTokenList: this.preferencesController.store.getState()
+      useStaticTokenList: !this.preferencesController.store.getState()
```
adonesky1, line 240: And then just not invert this value? Why would we do it this way?

test/e2e/webdriver/index.js
adonesky1: still remove?

adonesky1, line 76: Still need to remove this?

app/scripts/controllers/preferences.js
NiranjanaBinoy: The flag useStaticTokenList was added first in the controller repo and the default value was supposed to be `false`. But while adding the extension changes, I realized we need the toggle to be ON by default, hence the flag has to be `true`, after a discussion with @danjm we decided to not change the controller repo instead inverted the value we are sending from the extension.

I kept the same variable name to keep it consistent with the controller, now I am wondering if I should change that to `useTokenDetection`?

app/scripts/metamask-controller.js
NiranjanaBinoy: The `Use Token Detection` toggle in extension in ON by default to use the token service api, so we need the value of `useStaticTokenList` to be true. while in the controller, the `true` value for `useStaticTokenList` means we are using the static token list.

test/e2e/webdriver/index.js
NiranjanaBinoy: sorry, I thought I committed this change but I lost it somewhere along the branch switching 😟

app/scripts/controllers/preferences.js
adonesky1: Thanks for the background context @NiranjanaBinoy. I would vote yes to changing it to `useTokenDetection`!

ui/ducks/send/send.js
```diff
@@ -1006,13 +1014,19 @@ const slice = createSlice({
...
-          isValidHexAddress(recipient.userInput) &&
-          (toChecksumAddress(recipient.userInput) in contractMap ||
-            checkExistingAddresses(recipient.userInput, tokens))
-        ) {
-          recipient.warning = KNOWN_RECIPIENT_ADDRESS_WARNING;
+        if (isSendingToken && isValidHexAddress(recipient.userInput)) {
+          // token from dynamic api list is fetched when useStaticTokenList is true
+          const userInput = useStaticTokenList
+            ? recipient.userInput
+            : toChecksumAddress(recipient.userInput);
```
adonesky1, line 1021: I don't follow why we Checksum the input if useStaticTokenList is false?

NiranjanaBinoy: When the flag is false the tokenList contains tokens from `contract-metadata`, and the address of tokens in `contract-metadata` are checksum addresses.

app/scripts/controllers/detect-tokens.test.js
```diff
@@ -117,7 +117,7 @@ describe('DetectTokensController', function () {
     });
     tokenListController = new TokenListController({
       chainId: '1',
-      useStaticTokenList: false,
+      useTokenDetection: true,
```
adonesky1, line 120: Should the value change here?

app/scripts/controllers/detect-tokens.js
```diff
@@ -52,7 +52,7 @@ export default class DetectTokensController {
     if (!this.isActive) {
       return;
     }
-    if (Object.keys(this._tokenList.state.tokenList).length <= 0) {
+    if (Object.keys(this._tokenList.state.tokenList).length === 0) {
```
adonesky1, line 55: 👍

test/e2e/webdriver/index.js
```diff
@@ -73,7 +73,6 @@ async function setupFetchMocking(driver) {
       return { cancel: true };
     }
     function clone(obj) {
-      console.log(`clone`);
```
adonesky1: 👍

ui/ducks/send/send.js
adonesky1: But then you shouldn't need to cast them to checkSum again? So If anything it should be the other way around right? And it doesn't hurt to checkSum an address that is already formatted as such. I think we should remove the ternary? @ryanml any thoughts on this?

ui/helpers/utils/icon-factory.js
```diff
@@ -20,13 +20,13 @@ function IconFactory(jazzicon) {
...
   address,
   diameter,
-  useStaticTokenList,
+  useTokenDetection,
   tokenList,
 ) {
-  // token from dynamic api list is fetched when useStaticTokenList is true
-  const addr = useStaticTokenList ? address : toChecksumHexAddress(address);
+  // token from dynamic api list is fetched when useTokenDetection is true
+  const addr = useTokenDetection ? address : toChecksumHexAddress(address);
```
adonesky1, line 27: Same thing here. Maybe remove ternary and just toCheckSumHexAddress the address either way.

ui/ducks/send/send.js
adonesky1: ping here @ryanml @NiranjanaBinoy @Gudahtt

app/scripts/controllers/preferences.test.js
```diff
@@ -869,21 +869,21 @@ describe('preferences controller', function () {
       );
     });
   });
-  describe('setUseStaticTokenList', function () {
+  describe('setUseTokenDetection', function () {
     it('should default to false', function () {
       const state = preferencesController.store.getState();
-      assert.equal(state.useStaticTokenList, false);
+      assert.equal(state.useTokenDetection, true);
```
adonesky1, line 875: This test should either be renamed... or, and I think its this one, the value should still be false?

app/scripts/controllers/preferences.test.js
```diff
@@ -869,21 +869,21 @@ describe('preferences controller', function () {
...
+        preferencesController.store.getState().useTokenDetection,
+        true,
       );
-      preferencesController.setUseStaticTokenList(true);
+      preferencesController.setUseTokenDetection(false);
       assert.equal(
-        preferencesController.store.getState().useStaticTokenList,
-        true,
+        preferencesController.store.getState().useTokenDetection,
+        false,
```
adonesky1, line 886: same with all the values in this test.

app/scripts/controllers/preferences.test.js
NiranjanaBinoy: I renamed the test, the default value should be true.

app/scripts/controllers/preferences.test.js
NiranjanaBinoy: The default value is `true`, so the test is checking if it is getting set to `false` now but using the function,

ui/helpers/utils/icon-factory.js
NiranjanaBinoy: When `useTokenDetection` flag is true, the `tokenList` will be fetched from the token list service API and the address of the token provided by the API is not a checksum address. Whereas when the flag is false the token list will be fetched from `contract-metadata` and then the address is a checksum address. So tokenList will hold checksum address and non-checksum address at different points based on the list from which it is fetched.

adonesky1: Ok I still think we might as well checksum either way. It won't hurt if it's already checksummed? And if not doing that, could you add a comment explaining why we only checkSum the staticList addresses?

app/scripts/controllers/detect-tokens.test.js
```diff
@@ -18,7 +22,9 @@ describe('DetectTokensController', function () {
   const networkControllerProviderConfig = {
     getAccounts: noop,
   };
-
+  before(function () {
+    globalThis.AbortController = window.AbortController;
```
adonesky1, line 26: This should be remove since #11988 has been merged

app/scripts/controllers/detect-tokens.test.js
```diff
@@ -36,10 +42,92 @@ describe('DetectTokensController', function () {
...
+      onNetworkStateChange: sinon.spy(),
+      onPreferencesStateChange: sinon.spy(),
+      messenger: tokenListMessenger,
+    });
+    await tokenListController.start();
   });
 
   after(function () {
     sandbox.restore();
+    delete globalThis.AbortController;
```
adonesky1, line 130: This too (#11988)

app/scripts/controllers/detect-tokens.test.js
```diff
@@ -144,27 +244,26 @@ describe('DetectTokensController', function () {
...
       existingToken.symbol,
       existingToken.decimals,
     );
 
     const tokenAddressToAdd = erc20ContractAddresses[1];
-    const tokenToAdd = contracts[tokenAddressToAdd];
+    const tokenToAdd = tokenList[tokenAddressToAdd];
 
-    const contractAddresssesToDetect = contractAddresses.filter(
+    const contractAddresssesToDetect = erc20ContractAddresses.filter(
```
adonesky1, line 266: [nit]
```suggestion
    const contractAddressesToDetect = erc20ContractAddresses.filter(
```

app/scripts/controllers/detect-tokens.test.js
```diff
@@ -203,27 +302,26 @@ describe('DetectTokensController', function () {
...
```
adonesky1, line 324: [nit] 
```suggestion
    const contractAddressesToDetect = erc20ContractAddresses.filter(
```

app/scripts/controllers/detect-tokens.test.js
```diff
@@ -203,27 +302,26 @@ describe('DetectTokensController', function () {
...
       (address) => address !== existingTokenAddress,
     );
     const indexOfTokenToAdd = contractAddresssesToDetect.indexOf(
```
adonesky1, line 327:
```suggestion
    const indexOfTokenToAdd = contractAddressesToDetect.indexOf(
```

app/scripts/metamask-controller.js
```diff
@@ -242,11 +241,17 @@ export default class MetamaskController extends EventEmitter {
...
           return cb(modifiedNetworkState);
         }),
-      onPreferencesStateChange: this.preferencesController.store.subscribe.bind(
-        this.preferencesController.store,
-      ),
+      onPreferencesStateChange: (cb) =>
+        this.preferencesController.store.subscribe((preferencesState) => {
+          const modifiedPreferencesState = {
+            ...preferencesState,
+            useStaticTokenList: !this.preferencesController.store.getState()
```
adonesky1, line 248: I won't block on this at some point I think we should just switch the TokenListController to expect the `useTokenDection` value instead.

ui/ducks/send/send.js
adonesky1: Ah sorry @NiranjanaBinoy I understand this now. Could you add a comment explaining why the ternary is necessary?

ui/helpers/utils/icon-factory.js
adonesky1: Also here. I get why there's the ternary now. Please add a comment explaining?

ui/hooks/useTokensToSearch.js
```diff
@@ -17,7 +19,8 @@ import { isSwapsDefaultTokenSymbol } from '../../shared/modules/swaps.utils';
 import { toChecksumHexAddress } from '../../shared/modules/hexstring-utils';
 import { useEqualityCheck } from './useEqualityCheck';
 
-const tokenList = shuffle(
+// TODO: replace contract-metadata with list from state
+const shuffledContractMap = shuffle(
```
adonesky1, line 23: Should this const live in the hook?

app/scripts/metamask-controller.js
NiranjanaBinoy: The changes for the switch were dropped as it will be a breaking change in controllers repo for both mobile and extension at that time. [#560](https://github.com/MetaMask/controllers/pull/560) tracks the related changes if we decided to move forward with that.

brad-decker (dismissed):
LGTM

ui/hooks/useTokensToSearch.js
NiranjanaBinoy: The `const` is pre-existing, in this PR it is just renamed and updated the usage to be conditional based on the useTokenDetection flag. This `const` will not be used when the flag is false, in that case, won't it be better for the `const` to be outside the hook?!!

app/scripts/controllers/detect-tokens.test.js
adonesky1: no?

app/scripts/controllers/detect-tokens.test.js
adonesky1: shouldn't this be removed?

brad-decker (dismissed):

app/scripts/controllers/detect-tokens.test.js
```diff
@@ -18,7 +22,9 @@ describe('DetectTokensController', function () {
   const networkControllerProviderConfig = {
     getAccounts: noop,
   };
-
+  before(function () {
+    globalThis.AbortController = window.AbortController;
+  });
```
adonesky1, line 27: @NiranjanaBinoy should this be removed?

NiranjanaBinoy: Done 😊 I ended up reverting the changes after resolving it here for some reason I cannot recollect now and missed removing it afterward.😅

brad-decker (dismissed):

app/scripts/controllers/detect-tokens.js
```diff
@@ -24,71 +23,84 @@ export default class DetectTokensController {
...
     }
-    if (this._network.store.getState().provider.chainId !== MAINNET_CHAIN_ID) {
+    if (Object.keys(this._tokenList.state.tokenList).length === 0) {
       return;
     }
 
     const tokensToDetect = [];
     this.web3.setProvider(this._network._provider);
-    for (const contractAddress in contracts) {
+    const { tokenList } = this._tokenList.state;
```
darkwing, line 62: Nit: We're referencing this property on line 56 -- maybe move this up, and use `tokenList` in both places?

app/scripts/controllers/detect-tokens.js
```diff
@@ -24,71 +23,84 @@ export default class DetectTokensController {
...
+    const sliceOfTokensToDetect = [];
+    sliceOfTokensToDetect[0] = tokensToDetect.slice(0, 1000);
+    sliceOfTokensToDetect[1] = tokensToDetect.slice(
+      1000,
+      tokensToDetect.length - 1,
+    );
+    for (const tokensSlice of sliceOfTokensToDetect) {
+      let result;
+      try {
+        result = await this._getTokenBalances(tokensSlice);
```
darkwing, line 84: This could potentially be a huge list, correct?  Is there a reason we can't do `Promise.all` here?

app/scripts/controllers/detect-tokens.js
```diff
@@ -24,71 +23,84 @@ export default class DetectTokensController {
...
-
-    tokensToDetect.forEach((tokenAddress, index) => {
-      const balance = result[index];
-      if (balance && !balance.isZero()) {
-        this._preferences.addToken(
-          tokenAddress,
-          contracts[tokenAddress].symbol,
-          contracts[tokenAddress].decimals,
+    const sliceOfTokensToDetect = [];
+    sliceOfTokensToDetect[0] = tokensToDetect.slice(0, 1000);
```
darkwing, line 76: Nit: we could use `push` here.

test/e2e/tests/permissions.spec.js
```diff
@@ -62,7 +62,7 @@ describe('Permissions', function () {
           text: 'Connected sites',
           tag: 'h2',
         });
-
+        await driver.delay(xLargeDelayMs);
```
darkwing, line 65: Is this related in some way?

test/e2e/webdriver/index.js
```diff
@@ -52,6 +52,12 @@ async function setupFetchMocking(driver) {
         if (url.match(/featureFlags$/u)) {
           return { json: async () => clone(mockResponses.swaps.featureFlags) };
         }
+      } else if (
+        url.match(/^https:\/\/(token-api\.airswap-prod\.codefi\.network)/u)
```
darkwing, line 56: Can this be `else if(url === ` instead of the regular expression?

ui/ducks/send/send.js
```diff
@@ -1006,13 +1014,21 @@ const slice = createSlice({
...
+          const userInput = useTokenDetection
+            ? recipient.userInput
+            : toChecksumAddress(recipient.userInput);
+          if (
+            tokenAddressList.includes(userInput) ||
+            checkExistingAddresses(userInput, tokens)
+          ) {
+            recipient.warning = KNOWN_RECIPIENT_ADDRESS_WARNING;
+          } else {
+            recipient.warning = null;
```
darkwing, line 1030: Can we initialize `recipient.warning` to `null`?  If so, we can remove this `else` clause and the one underneath it.

ui/helpers/utils/util.js
```diff
@@ -56,6 +56,12 @@ export function isDefaultMetaMaskChain(chainId) {
   return false;
 }
 
+// Both inputs should be strings. This method is currently used to compare tokenAddress hex strings.
+export function isEqualCaseInsensitive(value1, value2) {
```
darkwing, line 60: Yay, I like this!

ui/pages/add-token/add-token.component.js
```diff
@@ -358,15 +366,16 @@ class AddToken extends Component {
   }
 
   renderSearchToken() {
+    const { tokenList = {} } = this.props;
```
darkwing, line 369: It might be cleaner if we use `AddToken .defaultProps` to set this `tokenList` value to empty object.

ui/pages/add-token/token-search/token-search.component.js
```diff
@@ -34,17 +16,40 @@ export default class TokenSearch extends Component {
...
     this.setState({ searchQuery });
-    const fuseSearchResult = fuse.search(searchQuery);
-    const addressSearchResult = contractList.filter((token) => {
-      return token.address.toLowerCase() === searchQuery.toLowerCase();
+    const fuseSearchResult = this.tokenSearchFuse.search(searchQuery);
+    const addressSearchResult = this.tokenList.filter((token) => {
+      return (
+        token.address &&
+        searchQuery &&
+        token.address.toLowerCase() === searchQuery.toLowerCase()
```
darkwing, line 51: Can we use `isEqualCaseInsensitive` here?

ui/ducks/send/send.js
brad-decker: This is likely fine, double checked to make sure it wouldn't interfere with other validation.

test/e2e/tests/permissions.spec.js
NiranjanaBinoy: The test started failing without that delay, it was not related to token detection changes.

test/e2e/webdriver/index.js
NiranjanaBinoy: yes, `else if(url === ` does work, but the above URLs are all using regular expressions for matching. And won't regular expression make it easier to add more params to the URL if needed in the future?

brad-decker (dismissed):
LGTM

app/scripts/controllers/detect-tokens.js
```diff
@@ -24,71 +23,85 @@ export default class DetectTokensController {
...
+          return reject(error);
+        }
+        return resolve(result);
+      });
+    });
   }
 
   /**
-   * For each token in @metamask/contract-metadata, find check selectedAddress balance.
+   * For each token in tokenlist ptovided by the TokenListController, find check selectedAddress balance.
```
adonesky1, line 50: [nit]
```suggestion
   * For each token in the tokenlist provided by the TokenListController, check selectedAddress balance.
```

app/scripts/controllers/detect-tokens.js
```diff
@@ -24,71 +23,85 @@ export default class DetectTokensController {
...
-        !this.hiddenTokens.includes(contractAddress.toLowerCase())
+        !this.tokenAddresses.find((address) =>
+          isEqualCaseInsensitive(address, tokenAddress),
+        ) &&
+        !this.hiddenTokens.find((address) =>
+          isEqualCaseInsensitive(address, tokenAddress),
+        )
       ) {
-        tokensToDetect.push(contractAddress);
+        tokensToDetect.push(tokenAddress.toLowerCase());
```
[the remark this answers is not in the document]
NiranjanaBinoy: The address in tokenList can be a checksum address or non-checksum address based on whether the user is using a dynamic or static list, I think my instinct was to keep them uniform

ui/hooks/useTokensToSearch.js
adonesky1: 👍

ui/ducks/send/send.js
```diff
@@ -1005,16 +1013,20 @@ const slice = createSlice({
...
+        recipient.warning = null;
+        if (isSendingToken && isValidHexAddress(recipient.userInput)) {
+          // When useTokenDetection flag is true the tokenList contains tokens with non-checksum address from the dynamic token service api,
+          // When useTokenDetection flag is false the tokenList contains tokens with checksum addresses from contract-metadata.
+          // So the flag indicates whether the address of tokens currently on the tokenList is checksum or not.
+          const userInput = useTokenDetection
+            ? recipient.userInput
+            : toChecksumAddress(recipient.userInput);
+          if (
+            tokenAddressList.includes(userInput) ||
```
adonesky1, line 1025: Ok I think I know why this has been bugging me. I think rather than doing the checksum if useTokenDetection is true, we should use the isEqualCaseSensitive check here. The checkExistingAddresses already does a case insensitive check.

brad-decker: [Q] doestokenAddressList contain checksum addresses? The code this is replacing used the checksumaddress to look up in contractMap which was keyed by checksum address.

NiranjanaBinoy: tokenAddressList will have checksumaddress when the user opt-out of the new token service in that scenario the tokenList will be holding the same data from contract-metadata. I am working on Alex's suggestion that way I can stick to the old code with minimal changes.

app/scripts/controllers/detect-tokens.js
```diff
@@ -24,71 +23,85 @@ export default class DetectTokensController {
...
-    tokensToDetect.forEach((tokenAddress, index) => {
-      const balance = result[index];
-      if (balance && !balance.isZero()) {
-        this._preferences.addToken(
-          tokenAddress,
-          contracts[tokenAddress].symbol,
-          contracts[tokenAddress].decimals,
+    const sliceOfTokensToDetect = [];
+    sliceOfTokensToDetect.push(tokensToDetect.slice(0, 1000));
+    sliceOfTokensToDetect.push(
```
darkwing, line 78: `push`, cool!  You actually don't need the two `push` directives, you can add multiple items in one go:

```
[].push("hello", "world");
```

We could also initialize the array with those items:

```
const sliceOfTokensToDetect = [
  tokensToDetect.slice(0, 1000),
  tokensToDetect.slice(1000, tokensToDetect.length - 1)
];
```

What is the significance of doing the slicing though?  It's not immediately apparent to me.

NiranjanaBinoy: The size of the data (around 5k) returned from the token service API slows down the token detection, we are splitting the list based on the number of occurrences (assuming the probability of a user having a popular token is higher) so that they can be detected faster followed by the remaining list, rather than waiting on the whole list.

ui/components/ui/identicon/identicon.component.js
```diff
@@ -78,16 +88,25 @@ export default class Identicon extends PureComponent {
...
 
     if (address) {
-      const checksummedAddress = toChecksumHexAddress(address);
-
-      if (checksummedAddress && contractMap[checksummedAddress]?.logo) {
+      // token from dynamic api list is fetched when useTokenDetection is true
+      const tokenAddress = useTokenDetection
+        ? address
+        : toChecksumHexAddress(address);
+      if (tokenAddress && tokenList[tokenAddress]?.iconUrl) {
```
adonesky1, line 109: Could do a similar thing here: 
```
const tokenAddressKey = Object.keys(tokenList).find((tokenAddress) => isEqualCaseInsensitive(address, tokenAddress))

if (tokenAddress && tokenList[tokenAddressKey]?.iconUrl) {
...
```

ui/helpers/utils/icon-factory.js
adonesky1: And here!

ui/hooks/useTokensToSearch.js
```diff
@@ -32,9 +34,14 @@ export function getRenderableTokenData(
...
   chainId,
+  tokenList,
+  useTokenDetection,
 ) {
   const { symbol, name, address, iconUrl, string, balance, decimals } = token;
-
+  // token from dynamic api list is fetched when useTokenDetection is true
+  const tokenAddress = useTokenDetection
+    ? address
+    : toChecksumHexAddress(address);
```
adonesky1, line 44: and here

ui/components/ui/identicon/identicon.component.js
NiranjanaBinoy: Won't be better to do a single check and conversion if needed, rather than looping through the whole list of tokens and converting them to lowercase every time the component is used??

adonesky1: yeah you're right! 🙏

ui/hooks/useTokensToSearch.js
```diff
@@ -59,31 +66,40 @@ export function getRenderableTokenData(
     ) || '';
   const usedIconUrl =
     iconUrl ||
-    (contractMap[toChecksumHexAddress(address)] &&
-      `images/contract/${contractMap[toChecksumHexAddress(address)].logo}`);
+    (tokenList[tokenAddress] &&
+      `images/contract/${tokenList[tokenAddress].iconUrl}`);
```
adonesky1, line 70: If tokenList is static won't this still be `logo`?

NiranjanaBinoy: No, in TokenListController we are reassigning `logo` to `iconUrl`.
https://github.com/MetaMask/controllers/blob/c7d7480339b9dc62bdd2a2b8cada49a96d62b01e/src/assets/TokenListController.ts#L249

ui/pages/add-token/token-list/token-list.component.js
```diff
@@ -35,13 +37,17 @@ export default class TokenList extends Component {
           {Array(6)
             .fill(undefined)
             .map((_, i) => {
-              const { logo, symbol, name, address } = results[i] || {};
+              const { iconUrl, symbol, name, address } = results[i] || {};
+              // token from dynamic api list is fetched when useTokenDetection is true
+              const iconPath = useTokenDetection
+                ? iconUrl
+                : `images/contract/${iconUrl}`;
```
adonesky1, line 44: sorry to keep asking for comments but can you explain why in the static list case we need this string interpolation?

NiranjanaBinoy: In the static list, the iconUrl will be holding only a filename for the image, we have a pre-existing setup that will make the images available in the `images/contract/` location (I think created when the contract-metadata package is added to the repo) so that it can be accessed using the filename in iconUrl.

adonesky1: Would you mind adding this as a comment in the code?

NiranjanaBinoy: sure, will add.

ui/store/actions.js
```diff
@@ -2131,7 +2135,7 @@ export function setPendingTokens(pendingTokens) {
       : selectedTokens;
 
   Object.keys(tokens).forEach((tokenAddress) => {
-    tokens[tokenAddress].unlisted = !LISTED_CONTRACT_ADDRESSES.includes(
+    tokens[tokenAddress].unlisted = !tokenAddressList.includes(
```
adonesky1, line 2138: another place we could use `isEqualCaseInsensitive`?

adonesky1 (approved):
LGTM! Man these conflicts are going to be awful. Wish me luck!

darkwing:
I think we did something here that broke our icons.  When I went to add a token, I'm seeing the following:

MissingIcons

MissingIcons2

darkwing (approved):

ryanml (approved):
i've looked at this previously, thanks for all of the changes and the hard work! :D

brad-decker (approved):

PR #11404, opened 2021-06-28
EIP-1559 - Implement TransactionDetail for confirmations

darkwing:
UI:

gas

There's extra padding at the bottom of the first item because we're missing the "Very likely" text on the left, and the gas range on the right.

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -2367,6 +2367,9 @@
   "transactionCreated": {
     "message": "Transaction created with a value of $1 at $2."
   },
+  "transactionDetailGasHeading": {
+    "message": "Estimated gas fee"
+  },
   "transactionDropped": {
     "message": "Transaction dropped at $2."
   },
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -2370,6 +2370,9 @@
   "transactionDetailGasHeading": {
     "message": "Estimated gas fee"
   },
+  "transactionDetailGasTotalSubtitle": {
+    "message": "Amount + gas fee"
+  },
   "transactionDropped": {
     "message": "Transaction dropped at $2."
   },
--- a/ui/components/app/transaction-detail-item/transaction-detail-item.component.js
+++ b/ui/components/app/transaction-detail-item/transaction-detail-item.component.js
@@ -27,13 +27,17 @@ export default function TransactionDetailItem({
           {detailTitle}
         </Typography>
         {detailText && (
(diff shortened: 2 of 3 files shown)


brad-decker (approved):
💯

ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
```diff
@@ -31,6 +31,10 @@ import { getTransactionTypeTitle } from '../../helpers/utils/transactions.util';
 import ErrorMessage from '../../components/ui/error-message';
 import { toBuffer } from '../../../shared/modules/buffer-utils';
 
+import TransactionDetail from '../../components/app/transaction-detail/transaction-detail.component';
```
dan437, line 34: Why do we use the `.component` postfix for file names, when it's already in the `/components` folder?

dan437 (approved):
It looks great, I'm going to use the component soon!

app/_locales/en/messages.json
```diff
@@ -2367,6 +2367,12 @@
   "transactionCreated": {
```
danjm, line 2367: :+1:

ui/components/app/transaction-detail-item/transaction-detail-item.component.js
```diff
@@ -27,13 +27,17 @@ export default function TransactionDetailItem({
           {detailTitle}
```
danjm, line 27: :+1:

ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
danjm: It is a practice we are moving away from. The reason it exists is to distinguish between "container" components, when we used that pattern for redux. So we have `confirm-transaction-base.component.js` and `confirm-transaction-base.container.js`. This is discouraged for all new components though

ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
```diff
@@ -302,6 +306,91 @@ export default class ConfirmTransactionBase extends Component {
       />
     );
 
+    const nonceField = useNonceField ? (
```
danjm, line 309: Lines 309-337 are not change from before, the code is just moved :+1:

danjm (approved):

ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
dan437: Thanks for the explanation, great to know!

PR #11406, opened 2021-06-28
update txParams normalization and validation. 

brad-decker:
Adds new TRANSACTION_ENVELOPE_TYPES constant to add types for EIP-2718 'type' field on txParams.

Also updates normalize / validate txParams methods to take into consideration these fields such that a dapp will get error messages useful in debugging when submitting transactions of various types. Note, we don't currently support EIP-2930 access lists and I am not working on that thread until after EIP-1559

  -

### DIFF
--- a/app/scripts/controllers/transactions/lib/util.js
+++ b/app/scripts/controllers/transactions/lib/util.js
@@ -7,18 +7,17 @@ import {
 import { isValidHexAddress } from '../../../../../shared/modules/hexstring-utils';
 
 const normalizers = {
-  from: (from) => addHexPrefix(from),
+  from: addHexPrefix,
   to: (to, lowerCase) =>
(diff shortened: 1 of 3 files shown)


darkwing (commented):
These functions look very similar; could we generisize the validation checking and use that generic function within each `validate...` function?

brad-decker:
> These functions look very similar; could we generisize the validation checking and use that generic function within each validate... function?

Good call, I have changed this or significantly:
1. I realized that the normalize methods were each creating a useless function except in one case, so these are just now passed 'addHexPrefix'
2. validation of fields now uses reusable methods
3. instead of using the 'validateMaxFeePerGas' type functions that no longer exist, I'm calling validateTxParams in tests and ensuring that an appropriate error message is thrown.

darkwing (approved):

ryanml (approved):

PR #11407, opened 2021-06-28
update speedup and cancel to make room for EIP-1559

brad-decker:
Makes the positional arguments of createCancelTransaction and createSpeedUpTransaction an option bag so that other fields (maxFeePerGas, etc) can be included in future PRs, this also updates configuring new gas properties so that when EIP-1559 transactions roll in they will be handled in the background. Right now in this PR if a user were to create a EIP-1559 transaction and then attempt to speed it up with a custom gas value this would not be used. Instead a 1.10 multiplier would be applied to both gas fields. 

### Manual Testing Steps
1. Make sure SHOW_EIP_1559_UI is false/0
2. Submit a transaction with low gasPrice (< 0.5) on any testnet
3. Attempt to cancel the transaction, and see that the suggested gasPrice is 10% higher than the original.
4. Attempt to speed up the cancellation and see that the gasPrice is 10% higher than the cancel.
5. Check transaction details for the proper history of resubmitting

Repeat the steps with a plain speed up and no cancel.

### DIFF
--- a/ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.container.js
+++ b/ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.container.js
@@ -264,7 +264,6 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => {
       if (ownProps.onSubmit) {
         dispatchHideSidebar();
         dispatchCancelAndClose();
-        console.log('here');
         ownProps.onSubmit({ gasLimit, gasPrice });
         return;
       }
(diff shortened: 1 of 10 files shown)


ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.container.js
```diff
@@ -264,7 +264,8 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => {
       if (ownProps.onSubmit) {
         dispatchHideSidebar();
         dispatchCancelAndClose();
-        ownProps.onSubmit(gasLimit, gasPrice);
+        console.log('here');
```
darkwing, line 267: Needs removal.

darkwing (changes requested):
Outside of the console.log, this works as expected, and the code makes sense.  Would appreciate more eyes though.

ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.container.js
brad-decker:
```suggestion
```

darkwing (approved):

NiranjanaBinoy (approved):
LGTM

PR #11408, opened 2021-06-28
add eip-1559 fields to event schema

brad-decker:
Adds EIP-1559 fields to the schema for tracking events.

### DIFF
--- a/app/scripts/controllers/transactions/index.js
+++ b/app/scripts/controllers/transactions/index.js
@@ -29,6 +29,7 @@ import {
   MAINNET,
   NETWORK_TYPE_RPC,
 } from '../../../../shared/constants/network';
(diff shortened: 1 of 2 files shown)


ryanml (commented):
Per discussion, let's add `gas_limit` to the default metrics params in `_trackTransactionMetricsEvent`. We are passing it in as an extra prop here: https://github.com/MetaMask/metamask-extension/blob/45836f8d02b49592d63bb1b92f102364b00467fa/app/scripts/controllers/transactions/index.js#L696

darkwing (dismissed):

ryanml (approved):
++ thanks for the changes!

darkwing (approved):

PR #5282, opened 2018-09-18
Add Cancel Transaction feature. Refactor modals. Add Transaction Details modal.

alextsg:

PR #4691, opened 2018-06-29
Refactor and redesign confirm transaction views

alextsg:

PR #11411, opened 2021-06-28
remove ui lookup of lastGasPrice field

brad-decker:
In the transaction controller when creating a cancel or speed up transaction, a parameter called 'lastGasPrice' is added to the transaction object. It serves a purpose of being used to reference whether or not to use a generated nonce (which can probably also be removed) but I found this code hard to understand how it was used.

Assumptions I need help with:
lastGasPrice will only be available on speed ups and cancels
speed ups and cancels are automatically approved and therefore a confirmation window should never be shown for them.

So in https://github.com/MetaMask/metamask-extension/pull/5282 we updated logic to the controller for this and in https://github.com/MetaMask/metamask-extension/pull/4691 we created the confirm-transaction duck that includes this logic, but I just don't understand the connection and why this was ever required.

### DIFF
--- a/ui/ducks/confirm-transaction/confirm-transaction.duck.js
+++ b/ui/ducks/confirm-transaction/confirm-transaction.duck.js
@@ -11,8 +11,6 @@ import {
   getHexGasTotal,
   addFiat,
   addEth,
-  increaseLastGasPrice,
-  hexGreaterThan,
 } from '../../helpers/utils/confirm-tx.util';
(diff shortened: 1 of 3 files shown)


danjm (approved):
LGTM, as per discussion in this thread: https://consensys.slack.com/archives/GTQAGKY5V/p1624920487198800

darkwing (approved):

PR #11412, opened 2021-06-29
Fixing invalid minimumGasLimit prop (AdvancedGasInputs)

ryanml:
Screen Shot 2021-06-28 at 7 16 11 PM

Manual testing steps:  
  - Enter Send Flow on a testnet
  - Enter a recipient address
  - Confirm when the amount/gas inputs view loads, that the above error does not exist (Extra check, ensure that the gas limit is appropriately 21000 or otherwise (of type `number`)

### DIFF
--- a/ui/components/app/gas-customization/advanced-gas-inputs/advanced-gas-inputs.container.js
+++ b/ui/components/app/gas-customization/advanced-gas-inputs/advanced-gas-inputs.container.js
@@ -4,6 +4,7 @@ import {
   decimalToHex,
   hexWEIToDecGWEI,
 } from '../../../../helpers/utils/conversions.util';
+import { MIN_GAS_LIMIT_DEC } from '../../../../pages/send/send.constants';
 import AdvancedGasInputs from './advanced-gas-inputs.component';
 
 function convertGasPriceForInputs(gasPriceInHexWEI) {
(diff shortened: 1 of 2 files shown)


danjm (dismissed):
LGTM

ui/components/app/gas-customization/advanced-gas-inputs/advanced-gas-inputs.container.js
```diff
@@ -14,19 +15,25 @@ function convertGasLimitForInputs(gasLimitInHexWEI) {
   return parseInt(gasLimitInHexWEI, 16) || 0;
 }
 
+function convertMinimumGasLimitForInputs(minimumGasLimit) {
+  return parseInt(minimumGasLimit, 10) || Number(MIN_GAS_LIMIT_DEC);
```
danjm, line 19: Very minor non-blocking nit: if `minimumGasLimit` is undefined, then `parseInt(minimumGasLimit, 10)` will return `NaN`. This is okay because it is falsy and so this function will return `Number(MIN_GAS_LIMIT_DEC)`

This function might be conceptually simpler if we default an undefined `minimumGasLimit` param to `MIN_GAS_LIMIT_DEC` and remove the need to default to a call to `Number`

This way no piece of this ever produces `NaN`. Also, if for some reason we ever do want to pass `0` as the `minimumGasLimit`, then that would be supported as well (current logic would result in `MIN_GAS_LIMIT_DEC` being return if the function is passed `0`)

ryanml: Good suggestion! fixed

brad-decker (approved):

danjm (approved):

Issue #11394, opened 2021-06-25
Ensure that users can read the error message RPC error messages that result upon attempting to submit a transaction


kevinghim:
JSON-RPC error issue raised by Optimism.

- Metamask chrome extension does not surface RPC methods to users, it just reverts.
- When we reject transactions for paying either too much or not high enough of a fee, we reject it at the RPC level and return an informative error message. Whenever this happens in the MM chrome extension right now, the error message is displayed for a fraction of a second and then transaction just appears as failed. It would be ideal if we could surface this error message to users.

https://user-images.githubusercontent.com/6951929/123484628-ab625d00-d5d6-11eb-9e1b-7de9dc102157.mov

PR #11416, opened 2021-06-29
Allow users to see confirm screen errors.

danjm:
This PR makes it possible to actually view transaction errors that occur upon submission. Currently, when on the confirm screen and attempting to submit a transaction, an error is rendered for a half second and then the user is taken to the home screen. This leaves the user with no information about why the transaction submission attempt failed, making it harder for them to correct their own actions, and adding friction to the support process.

This PR corrects this keeping a record of failed transactions that we wish to show, and then utilizing the confirm screen to show them. When in this state, transaction data cannot be edited. For now, this is just utilized to show failed transactions immediately after they fail. In the future we could add the ability to inspect transactions that have failed in the past via the same UI, but we don't need to.

This was motivated by developers and users of custom networks that have varying gas limits. It was suggested that allowing users to see this error would help them adapt to the custom network more easily.

This is meant to be an iterative but clear improvement over the current state, which simply allows a user to actually read an error that we render but navigate away from too quickly.

### Before

https://user-images.githubusercontent.com/7499938/123792053-c1b03780-d8ba-11eb-831c-5b2e4a0cf298.mp4

### After:

https://user-images.githubusercontent.com/44811/137756296-fba57419-b4fc-4e6f-aa42-f61475a765f1.mov


danjm:
closing for now, will re-open once there are proper designs

kevinghim:
Reopening with Yanrong's new designs: https://www.figma.com/file/vAvkQ7sbGlTCFqffCHfIfY/confirm-screen-errors?node-id=0%3A1

danjm:
Once resolved, this PR will also address: https://github.com/MetaMask/metamask-extension/issues/11394

danjm:
It might be best to close this draft PR and just reference it in the ticket.

segun:
I'm looking to fix this, but I can't seem to find where to pause the process from going back to the home page when there's an error. It sort of auto switched to the home page once the button is clicked.

@danjm @brad-decker

darkwing:
Something weird happened here with the number of files changed...

segun:
> Something weird happened here with the number of files changed...

ooops, it was 21. I did a rebase. I think that funked things up.

Gudahtt (changes requested):

jpuri (commented):
Nice work 👍 , I added some small feedbacks.

Gudahtt (changes requested):

ui/store/actions.js
```diff
@@ -1103,6 +1113,32 @@ export function updateMetamaskState(newState) {
       type: actionConstants.UPDATE_METAMASK_STATE,
       value: newState,
     });
+
+    // check that tx status has not changed and remove it from failed transactions if it has
+    const foundTx = currentNetworkTxList && currentNetworkTxList.find((currentTx, index) => {
```
Gudahtt, line 1118: The `currentNetworkTxList` should always be set. If it's empty, it'll be an empty array.

```suggestion
    const foundTx = currentNetworkTxList.find((currentTx, index) => {
```

Gudahtt: Also... this only finds the first transaction in `currentNetworkTxList` that should be removed from `failedTransactionsToDisplay`. If multiple transactions were to be removed in one update, it'd miss the second one.

I think it would be better to loop through `failedTransactionsToDisplay` and check each one against `currentNetworkTxList` for state changes. That way we ensure we remove _all_ that should be removed. 

e.g. `const transactionIdsToRemove = failedTransactionsToDisplay.filter(...)`

Gudahtt (commented):
The name `failedTxesToDisplay` seems to wrongly apply that all of the transactions in the list have failed already. But actually this is a list of transactions that _might_ encounter a network failure, which we want to display in the event of an error. Maybe we can come up with a more accurate name.

And on a less important note, I think it would be better for readability to spell out works completely rather than use shortened forms like "txs".

`transactionsToDisplayUponFailure` maybe? I'm having trouble thinking of alternate names.

ui/store/actions.js
```diff
@@ -1103,6 +1113,32 @@ export function updateMetamaskState(newState) {
...
+      if (
+        failedTransactionsToDisplay &&
+        failedTransactionsToDisplay[currentTx.id]
+      ) {
+        const newTxStatus = getStatusKey(newTx);
+        if (
+          isEqual(newTx.status, currentTx.status) === false &&
+          newTxStatus !== TRANSACTION_STATUSES.FAILED &&
+          newTxStatus !== TRANSACTION_STATUSES.SIGNED &&
+          newTxStatus === TRANSACTION_STATUSES.APPROVED
```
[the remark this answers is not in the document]
Gudahtt: It looks like this hasn't been addressed yet

ui/store/actions.js
```diff
@@ -1116,6 +1126,37 @@ export function updateMetamaskState(newState) {
       type: actionConstants.UPDATE_METAMASK_STATE,
       value: newState,
     });
+
+    // Check that the transaction was not submitted successfully, and remove it from failed transactions if it was.
+    const transactionIdsToRemove = Object.values(
```
Gudahtt, line 1131: Was this meant to be `Object.keys`? The values are all booleans, not ids.

ui/components/ui/error-message/index.scss
```diff
@@ -19,4 +19,18 @@
   &__text {
     overflow: auto;
   }
+
+  &__link {
+    text-decoration: none;
+    cursor: pointer;
```
Gudahtt, line 25: Now that the link is an `<a>` element, it already has a pointer cursor. This is not needed.

```suggestion
```

ui/components/ui/error-message/error-message.component.js
```diff
@@ -12,14 +12,25 @@ const ErrorMessage = (props, context) => {
         alt=""
         className="error-message__icon"
       />
-      <div className="error-message__text">{error}</div>
+      <div className="error-message__text">
+        {error}
+        {linkText ? (
+          <a className="error-message__link" onClick={onErrorMessageClick}>
```
Gudahtt, line 18: IIRC this needs an `href` to be keyboard focusable. Elsehwere we've been settting `href` to `"#"` if we don't want to navigate anywhere, just to ensure it's focusable.

```suggestion
          <a className="error-message__link" href="#" onClick={onErrorMessageClick}>
```

ui/store/actions.js
segun: yes object.keys.

ui/store/actions.js
```diff
@@ -1116,6 +1126,37 @@ export function updateMetamaskState(newState) {
       type: actionConstants.UPDATE_METAMASK_STATE,
       value: newState,
     });
+
+    // Check that the transaction was not submitted successfully, and remove it from failed transactions if it was.
+    const transactionIdsToRemove = Object.keys(
+      transactionsToDisplayOnFailure,
+    ).filter((id) => {
+      const foundTx = currentNetworkTxList.find((currentTx, index) => {
```
Gudahtt, line 1134: I don't think this is doing what you meant it to do. It looks like this is finding the first transaction in `currentNetworkTxList` that has changed to a relevant status, irrespective of the id. Then it's checking the ID last. This will find at-most a single transaction.

ui/store/actions.js
```diff
@@ -1116,6 +1126,39 @@ export function updateMetamaskState(newState) {
...
+        .map((tx, index) => {
+          return { id: tx.id, status: tx.status, index };
+        })
+        .find((currentTx) => {
+          return id === currentTx.id;
+        });
+
+      // check if the status of foundTx has changed
+      if (foundTx) {
+        const newTx = newNetworkTxList[foundTx.index];
```
Gudahtt, line 1144: There is no guarantee that the transaction at this index is the correct transaction. We need to find this transaction by id.

ui/store/actions.js
```diff
@@ -1116,6 +1126,39 @@ export function updateMetamaskState(newState) {
...
+          newTx.status !== foundTx.status &&
+          newTxStatus !== TRANSACTION_STATUSES.FAILED &&
+          newTxStatus !== TRANSACTION_STATUSES.SIGNED &&
+          newTxStatus !== TRANSACTION_STATUSES.APPROVED
+        ) {
+          return true;
+        }
+      }
+
+      return false;
```
Gudahtt, line 1156: This can be done more concisely, e.g.

```suggestion
      const currentTx = currentNetworkTxList.find((tx) => tx.id === id);
      const newTx = newNetworkTxList.find((tx) => tx.id === id);

      return (
        newTx.status !== currentTx.status &&
        newTx.status !== TRANSACTION_STATUSES.FAILED &&
        newTx.status !== TRANSACTION_STATUSES.SIGNED &&
        newTx.status !== TRANSACTION_STATUSES.APPROVED
      );
```

This also solves the issue I mentioned in my other comment about how to find the transaction in the `newNetworkTxList`.

ui/selectors/selectors.js
```diff
@@ -403,6 +403,28 @@ export function getUnapprovedConfirmations(state) {
...
 
+export function getFailedTransactionsToDisplay(state) {
+  const { currentNetworkTxList = [] } = state.metamask;
+  const { transactionsToDisplayOnFailure = {} } = state.appState;
+  const failedTransactions = currentNetworkTxList.reduce(
+    (_failedTransactions, tx) => {
+      if (
+        transactionsToDisplayOnFailure[tx.id] &&
+        (tx.err || tx.status === 'approved' || tx.status === 'signed')
+      ) {
```
Gudahtt, line 414: I don't think we need to filter transactions by state or by the presence of an error here. If a transaction is in the `transactionsToDisplayOnFailure` map, we should include it here. The logic for filtering out completed transactions is already present in the background state update handler - we don't need to duplicate it.

This is dangerous in that it could hide an error if we did include a transaction by mistake that doesn't belong here. Also it would result in a mismatch between this list and the "display count" derived below.

```suggestion
      if (transactionsToDisplayOnFailure[tx.id]) {
```

ui/pages/confirm-transaction-base/confirm-transaction-base.container.js
```diff
@@ -84,10 +89,13 @@ const mapStateToProps = (state, ownProps) => {
...
     provider: { chainId },
   } = metamask;
+
+  const transactionsToDisplayOnFailure = getFailedTransactionsToDisplay(state);
+
   const { tokenData, txData, tokenProps, nonce } = confirmTransaction;
   const { txParams = {}, id: transactionId, type } = txData;
   const transaction =
-    Object.values(unapprovedTxs).find(
+    Object.values({ ...unapprovedTxs, ...transactionsToDisplayOnFailure }).find(
```
Gudahtt, line 98: In this case we want to ensure that the unapproved transaction is always chosen first, if it's present in both collections. It probably works correctly here because `Object.values` returns entries in insert order in most cases. But I do not think the iteration order of `Object.values` is guaranteed to be insert-order in all cases. I don't know if this works consistently in all of our supported browsers.

To make certain that this check is done in the correct order, we should look in each of these collections separately:

e.g. something like:

```
const transaction =
  Object.values(unapprovedTxs).find(({ id }) => id === (transactionId || Number(paramsTransactionId))) ||
  Object.values(transactionsToDisplayOnFailure).find(({ id }) => id === (transactionId || Number(paramsTransactionId))) ||
  {};
```

ui/pages/confirm-transaction-base/confirm-transaction-base.container.js
```diff
@@ -180,6 +188,10 @@ const mapStateToProps = (state, ownProps) => {
     fromAddress,
   );
 
+  const isFailedTransaction = Boolean(
+    transactionsToDisplayOnFailure[fullTxData.id],
```
Gudahtt, line 192: The presence of the transaction in this collection does not indicate that it is a failed transaction. We should be checking here that the transaction _isn't_ in `unapprovedTxs` - not that it _is_ in `transactionsToDisplayOnFailure`.

ui/components/app/confirm-page-container/confirm-page-container-container.test.js
```diff
@@ -61,6 +61,7 @@ describe('Confirm Page Container Container Test', () => {
     showAddToAddressBookModal: sinon.spy(),
     contact: undefined,
     isOwnedAccount: false,
+    hideTitle: false,
```
Gudahtt, line 64: Why was this change needed? It appears to be unrelated to the other changes.

ui/components/app/confirm-page-container/confirm-page-container-content/confirm-page-container-content.component.js
```diff
@@ -117,15 +124,32 @@ export default class ConfirmPageContainerContent extends Component {
         {this.renderContent()}
         {(errorKey || errorMessage) && (
           <div className="confirm-page-container-content__error-container">
-            <ErrorMessage errorMessage={errorMessage} errorKey={errorKey} />
+            <ErrorMessage
+              errorMessage={this.context.t('somethingWentWrong')}
+              errorKey={errorKey}
```
Gudahtt, line 129: In the case where `errorKey` is set, this component is broken.

`errorKey`, when set, is used to obtain the error message. This is a deprecated pattern that we're trying to get rid of because it makes these keys impossible to find with strict-mode of the localized message linter, but nonetheless it's something we need to continue to support until we remove all uses.

Right now if it's set, the full error message will be shown (instead of "something went wrong") along with the "more details" button. When the user clicks "more details", they'll be shown an empty modal.

We need to decide - do we want to treat keyed errors the same way we do text errors? If so, we need to stop passing it into `ErrorMessage` here, and use the key to get the full error message below when passing it into `TransactionErrorDetailsModal`.

If we want to treat keyed errors differently than raw string errors (because they have translations), we can show them directly here and omit the "More details" link altogether.

ui/components/app/gas-customization/advanced-gas-inputs/advanced-gas-inputs.component.js
```diff
@@ -21,6 +21,7 @@ export default class AdvancedGasInputs extends Component {
     minimumGasLimit: PropTypes.number,
     customPriceIsExcessive: PropTypes.bool,
     networkSupportsSettingGasPrice: PropTypes.bool,
+    disabled: PropTypes.bool,
```
Gudahtt, line 24: Why were the changes in this component made? It looks like this prop is never set.

segun: These crept in after the last rebase....I am going to compare with previous commits and fix.

ui/components/app/confirm-page-container/confirm-page-container-container.test.js
segun: This snuck in from the last rebase.

ui/components/app/confirm-page-container/confirm-page-container-content/confirm-page-container-content.component.js
segun: @Gudahtt What do you about about splitting this component, so we have one with the ability to accept links and the other one without?

Gudahtt: That could work! They seem different enough - one with plain text (the error key case) and one with a generic error with a more details button. Good idea.

segun: cool

ui/pages/confirm-transaction-base/confirm-transaction-base.container.js
segun: @Gudahtt this doesn't work as expected. At this point in the execution, the transaction is in unapprovedTxs even though it's a failed tx.

Gudahtt: Hmmm. That should not be the case. The list of `unapprovedTxs` necessarily have a status of "unapproved".

ui/components/ui/detailed-error-message/detailed-error-message.component.js
```diff
@@ -0,0 +1,40 @@
...
+          ''
+        )}
+      </div>
+    </div>
+  );
+};
+
+DetailedErrorMessage.propTypes = {
+  errorMessage: PropTypes.string,
+  errorKey: PropTypes.string,
```
Gudahtt, line 31: I think we should omit this prop, since we want to discourage the use of "error keys" (because they are incompatible with the strict mode of our verify-locales linter) and none are currently used with this component.

ui/pages/confirm-transaction-base/confirm-transaction-base.container.js
segun: When I use the above condition (isFailedTransaction = isn't in unapprovedTxs), I can't go past the confirm page because the confirm button is not shown (i.e isFailedTransaction is true).

Pretty late for me. I will take a closer look in the morning

ui/components/ui/detailed-error-message/detailed-error-message.component.js
```diff
@@ -0,0 +1,40 @@
...
+  linkText: PropTypes.string,
+  onErrorMessageClick: PropTypes.func,
```
Gudahtt, line 33: Could we make these mandatory? I can't imagine we'd have a "detailed error message" that had no way to click to see the details.
```suggestion
  linkText: PropTypes.string.isRequired,
  onErrorMessageClick: PropTypes.func.isRequired,
```

This would let us simplify the render function as well, by removing the conditions that check if these props are present or not. And it would help validate that `linkText` is always set with `onErrorMessageClick`.

ui/pages/confirm-transaction-base/confirm-transaction-base.container.js
Gudahtt: Interesting! Sounds good, I can take a look tomorrow as well

ui/components/app/confirm-page-container/confirm-page-container-content/confirm-page-container-content.component.js
Gudahtt: The second new component looks great!

ui/pages/confirm-transaction-base/confirm-transaction-base.container.js
segun: Ok, so I fixed this. The real issue was fullTxData.id is number and unapproveTxs ids are string.

ui/components/app/confirm-page-container/confirm-page-container-content/confirm-page-container-content.component.js
```diff
@@ -117,15 +125,38 @@ export default class ConfirmPageContainerContent extends Component {
         {this.renderContent()}
         {(errorKey || errorMessage) && (
           <div className="confirm-page-container-content__error-container">
-            <ErrorMessage errorMessage={errorMessage} errorKey={errorKey} />
+            {errorKey ? (
+              <ErrorMessage
+                errorMessage={this.context.t('somethingWentWrong')}
```
Gudahtt, line 130: This prop is not used if `errorKey` is passed in

```suggestion
```

ui/pages/confirm-transaction-base/confirm-transaction-base.container.js
Gudahtt: This still looks like the opposite of what it should be. Did you forget a `!`?

Gudahtt: Did you discover what was going on here after?

danjm:
@segun can you address this previous comment:

> The name failedTxesToDisplay seems to wrongly apply that all of the transactions in the list have failed already. But actually this is a list of transactions that might encounter a network failure, which we want to display in the event of an error. Maybe we can come up with a more accurate name.

> transactionsToDisplayUponFailure maybe? I'm having trouble thinking of alternate names.

danjm:
Oh, I see that was done at some point: https://github.com/MetaMask/metamask-extension/commit/ac896b3f871fe530ea46745486dba0d78fbb2cec

danjm:
Oh, it is just some methods that have names similar to the old `failedTxesToDisplay`

ui/pages/confirm-transaction/confirm-transaction.container.js
```diff
@@ -28,10 +36,24 @@ const mapStateToProps = (state, ownProps) => {
...
-    ? unapprovedTxs[id] || unconfirmedTransactions[0]
-    : {};
+
+  if (failedTransactionsToDisplay && failedTransactionsToDisplay[id]) {
+    const status = getStatusKey(failedTransactionsToDisplay[id]);
+    if (
+      status !== TRANSACTION_STATUSES.FAILED &&
+      status !== TRANSACTION_STATUSES.SIGNED
+    ) {
+      delete failedTransactionsToDisplay[id];
```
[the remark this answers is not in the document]
danjm: This was addressed in https://github.com/MetaMask/metamask-extension/commit/f2345d2a62e779732c1047a20977ec82ec0eabeb

ui/store/actions.js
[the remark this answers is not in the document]
danjm: addressed in https://github.com/MetaMask/metamask-extension/commit/9f1be122cb88e682da264b837b14f6b57717a6a3#diff-3c58b916cefff81e4927a9afd392acadbf4515f99f41783964a9851b7774791b

ui/store/actions.js
danjm: this was done in https://github.com/MetaMask/metamask-extension/commit/ac896b3f871fe530ea46745486dba0d78fbb2cec#diff-3c58b916cefff81e4927a9afd392acadbf4515f99f41783964a9851b7774791b

danjm:
Looks like this comment needs resolution https://github.com/MetaMask/metamask-extension/pull/11416#discussion_r756946798

Gudahtt:
The e2e test error does seem like a legitimate issue. I tried locally and it also fails. Infinite loading screen, no console errors in UI or background.

danjm (commented):
LGTM

danjm:
Explanation of last commit: `Object.keys` always returns an array of strings, even if the keys are numbers.

danjm:
Rebased again to fix the latest merge conflict

ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
```diff
@@ -1088,7 +1106,12 @@ export default class ConfirmTransactionBase extends Component {
           onEdit={() => this.handleEdit()}
           onCancelAll={() => this.handleCancelAll()}
           onCancel={() => this.handleCancel()}
-          onSubmit={() => this.handleSubmit()}
+          onSubmit={() =>
+            isFailedTransaction
+              ? this.handleFailedTxClose()
+              : this.handleSubmit()
+          }
+          onConfirmAnyways={() => this.handleConfirmAnyways()}
```
Gudahtt, line 1114: What is this for? It doesn't appear to be used.

```suggestion
```

ui/store/actions.js
```diff
@@ -1105,6 +1112,28 @@ export function updateMetamaskState(newState) {
       type: actionConstants.UPDATE_METAMASK_STATE,
       value: newState,
     });
+
+    // Check that the transaction was not submitted successfully, and remove it from failed transactions if it was.
+
```
Gudahtt, line 1117: Nit: This wording is a bit awkward (my bad, I think this is from an earlier suggestion of mine), and this newline isn't needed. Here is a suggested improvement.

 ```suggestion
    // Remove successful transactions from `transactionsToDisplayOnFailure`.
```

File: ui/ducks/app/app.js [hunk 21]
```diff
@@ -388,6 +405,19 @@ export function hideWhatsNewPopup() {
 export function toggleGasLoadingAnimation(value) {
   return { type: actionConstants.TOGGLE_GAS_LOADING_ANIMATION, value };
 }
+export function addTxToFailedTxesToDisplay(txId) {
```
Gudahtt, line 408: Could we rename these two functions and the actions as well? This isn't "failed transactions to display" anymore, it's "transactions to display on failure".

So this could be `ADD_TRANSACTION_TO_DISPLAY_ON_FAILURE` and `addTransactionToDisplayOnFailure` instead for example.

ui/ducks/app/app.js
```diff
@@ -388,6 +405,19 @@ export function hideWhatsNewPopup() {
...
   return { type: actionConstants.TOGGLE_GAS_LOADING_ANIMATION, value };
 }
+export function addTransactionToDisplayOnFailure(txId) {
+  return {
+    type: actionConstants.ADD_TRANSACTION_TO_DISPLAY_ON_FAILURE,
+    value: txId,
+  };
+}
+
+export function removeTxFromFailedTxesToDisplay(txId) {
```
Gudahtt, line 415: Sorry I had meant that both functions should be renamed. This function should be renamed too, for the same reason as `addTransactionToDisplayOnFailure`.

segun: I'm very bad with names...so this is what I come up with

`removeTransactionsDisplayedOnFailure` and `REMOVE_TRANSACTIONS_DISPLAYED_ON_FAILURE`

Gudahtt: Hmm.... to be the obvious choice would be to match the parallel "add" function, but replace "add" with "remove". e.g.

`REMOVE_TRANSACTION_TO_DISPLAY_ON_FAILURE` and `removeTransactionToDisplayOnFailure`

ui/components/app/confirm-page-container/confirm-page-container.component.js
```diff
@@ -85,6 +85,7 @@ export default class ConfirmPageContainer extends Component {
     contact: PropTypes.object,
     isOwnedAccount: PropTypes.bool,
     supportsEIP1559V2: PropTypes.bool,
+    isFailedTransaction: PropTypes.bool,
```
Gudahtt, line 88: The prop types for this component are organized into categories. This prop should be in the "Footer" category, but it's currently listed under "Gas Popover".

ui/components/app/confirm-page-container/confirm-page-container.component.js
```diff
@@ -240,8 +247,14 @@ export default class ConfirmPageContainer extends Component {
...
-              submitText={this.context.t('confirm')}
+              submitText={
+                isFailedTransaction
+                  ? this.context.t('close')
+                  : this.context.t('confirm')
+              }
+              submitButtonType={isFailedTransaction ? 'default' : 'confirm'}
               disabled={disabled}
             >
               {unapprovedTxCount > 1 && (
```
Gudahtt, line 260: Should we hide the "Reject X" button as well when showing a failed transaction? It would be a bit confusing if that button didn't reject what is currently shown.

Gudahtt (commented):
I noticed that this breaks our display of multiple transactions. The code for showing the "X of Y" header when displaying multiple transactions doesn't count failed transactions, and neither does the "Reject X" button at the bottom.

I'm not entire sure what to do here design wise. We can update the counter to include a failed transaction that we want to display. But should we still show the "Reject X" button when displaying a failed transaction? Should the "X" include the transaction being shown or not (since we can't really "reject" it because it has already failed).

ui/components/app/confirm-page-container/confirm-page-container.component.js
segun: The "Reject X" button is not shown.

Gudahtt: It is shown of there is more than one unapproved transaction. It is possible for that to be the case while viewing a failed transaction.

danjm (commented):
This looks good to me. But I can't approve because I am the author of the PR

ui/components/app/modals/transaction-error-details-modal/index.scss
```diff
@@ -0,0 +1,43 @@
...
+      font-weight: bold;
+
+      h2 {
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+      }
+
+      button {
+        margin-right: 24px;
```
darkwing, line 35: This might be best served as `margin-inline-end` to better support RTL.

ui/components/ui/error-message/index.scss
```diff
@@ -19,4 +19,17 @@
...
+    color: $Blue-500;
+
+    &:hover {
+      color: $Blue-400;
+    }
+
+    &:active {
+      color: $Blue-600;
+    }
+  }
```
darkwing, line 34: Since we moved to CSS variables, we need to change `$Blue-500` to `var(--Blue-500)`, as with the other two.

darkwing (changes requested):
ErrorDetails

The spacing of the icon and the text seems slim, and we should probably make the text size a bit smaller.  The other thing I'm seeing is that "More Details" text shows the text cursor upon hover, where it should show a pointer cursor.

ui/components/ui/detailed-error-message/detailed-error-message.component.js
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <div className="error-message">
+      <img
+        src="./images/alert-red.svg"
+        alt=""
+        className="error-message__icon"
+      />
+      <div className="error-message__text">
+        {errorMessage}
+        <a className="error-message__link" onClick={onErrorMessageClick}>
```
darkwing, line 16: We should make this a `<Button type="link">` component for the sake of accessibility.

darkwing:
Functionally this works great!  Just a few nits for CSS and accessibility improvements.

brad-decker:
@danjm is this something we still want? Possible to get this rebased so we can get it landed?

brad-decker:
@danjm closing this PR feel free to reopen it if you think it should remain open. in the meantime the branch will stay put if we want to return to this.

PR #11417, opened 2021-06-29
Build System: Don't run subprocesses in Lavamoat if the main process is not run in Lavamoat

EtDu:
As the title says


danjm (approved):
LGTM!

darkwing (approved):

PR #11418, opened 2021-06-29
Ensure that eth_estimateGas is called to estimate gas limit for simple sends on custom networks

danjm:
This PR updates our gas estimation logic so that if we are doing a simple send on a network that is not one of the default/standard networks in metamask (e.g. eth mainnet, or the test networks supported by infura), then we set the gas estimate to the result of calling `eth_gasEstimate`

This will enable custom networks to set gas prices for simple sends that are different than `21000`

### DIFF
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -251,6 +251,17 @@ async function estimateGasLimitForSend({
     );
   }
 
+  // The buffer multipler reduces transaction failures by ensuring that the
+  // estimated gas is always sufficient. Without the multiplier, estimates
+  // for contract interactions can become inaccurate over time. This is because
+  // gas estimation is non-deterministic. The gas required for the exact same
+  // transaction call can change based on state of a contract or changes in the
+  // contracts environment (blockchain data or contracts it interacts with).
+  // Applying the 1.5 buffer has proven to be a useful guard against this non-
+  // deterministic behaviour.
+  //
+  // Gas estimation of simple sends should, however, be deterministic. As such
+  // no buffer is needed in those cases.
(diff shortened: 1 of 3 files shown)


ui/selectors/selectors.js
```diff
@@ -376,6 +376,10 @@ export function getIsTestnet(state) {
   return TEST_CHAINS.includes(chainId);
 }
 
+export function getIsNonStandardEthChain(state) {
+  return !(getIsMainnet(state) || getIsTestnet(state));
```
brad-decker, line 380: you should probably return false if IN_TEST to match some of our other logic to treat ganache as mainnet in tests.

brad-decker (dismissed):
looks good to me, minor fix that should fix e2e

brad-decker (dismissed):

ui/ducks/send/send.js
```diff
@@ -231,25 +237,30 @@ async function estimateGasLimitForSend({
...
+      multiplyCurrencies(blockGasLimit, 0.95, {
+        multiplicandBase: 16,
+        multiplierBase: 10,
+        roundDown: '0',
+        toNumericBase: 'hex',
+      }),
+    );
+  }
+
+  const bufferMultiplier = isSimpleSendOnNonStandardNetwork ? 1 : 1.5;
```
adonesky1, line 254: Would it be possible to add a comment explaining why we do this?

danjm: @adonesky1 added in the latest commit

adonesky1 (approved):
LGTM! Code looks great and functions as described!

brad-decker (approved):

PR #11419, opened 2021-06-29
EIP 1559 - Provide a GWEI label for advanced gas control fields

darkwing:
Since @jakehaugen 's original design for `GWEI` text positioning won't work, we're going to place that label next to the field name:

GWEI


brad-decker (dismissed):

brad-decker (approved):

ryanml (approved):
lgtm!

PR #11421, opened 2021-06-29
wire up gasFeeController

brad-decker:
Initial pass at wiring up gasFeeController. This PR gets values flowing to the UI but does not yet use them for anything. To validate that this PR achieves its goal check out redux devtools and check that gasFeeEstimates is included in metamask state tree when entering the send flow. If you switch to ropsten you should see an update where 'gasPrice' is replaced with a slew of new estimates for EIP-1559. 

### This is a feature branch, the following PRs have merged into this and were reviewed separately
* [Wire up send flow to gasFeeController](https://github.com/MetaMask/metamask-extension/pull/11421/commits/b27f024ce0333547bcaae1845322f8278591a4e0) 
* [Pull logic out of gas modal into shared method](https://github.com/MetaMask/metamask-extension/pull/11432)


app/scripts/metamask-controller.js
```diff
@@ -174,6 +175,23 @@ export default class MetamaskController extends EventEmitter {
       initState: initState.MetaMetricsController,
     });
 
+    this.gasFeeController = new GasFeeController({
```
danjm, line 178: This looks correct

ui/ducks/send/send.js
```diff
@@ -356,6 +357,8 @@ export const initializeSendState = createAsyncThunk(
     // Initiate gas slices work to fetch gasPrice estimates. We need to get the
     // new state after this is set to determine if initialization can proceed.
     await thunkApi.dispatch(fetchBasicGasEstimates());
+
```
danjm, line 360: This looks correct

danjm (dismissed):
LGTM

app/scripts/metamask-controller.js
```diff
@@ -174,6 +175,23 @@ export default class MetamaskController extends EventEmitter {
...
+      interval: 10000,
+      messenger: controllerMessenger,
```
Gudahtt, line 180: It's rather important to pass in a restricted messenger here, not the global one. See the `currencyRateMessenger` for an example of this. This would fail to compile if this were in TypeScript.

package.json
```diff
@@ -100,7 +100,7 @@
     "@lavamoat/preinstall-always-fail": "^1.0.0",
     "@material-ui/core": "^4.11.0",
     "@metamask/contract-metadata": "^1.26.0",
-    "@metamask/controllers": "^10.0.0",
+    "@metamask/controllers": "https://github.com/MetaMask/controllers#2ad3f2d1e5a235e5d6bdc47851f151476f2efa80",
```
Gudahtt, line 103: This PR should be marked as DO-NOT-MERGE or in draft until this is released. We don't want to merge this.

Gudahtt (dismissed):

package.json
brad-decker: This will require this PR to become a long running feature branch for our work to continue, I strongly suggest we merge this so that the code can be used for future iteration. Is there anyway around this @Gudahtt ?

Gudahtt: Yes - merge and release the gas fee controller.

If we merge this as-is, then `develop` ceases to be releasable. Keeping `develop` releasable at all times is critical to if we want to keep with the regular release cadence we've been aiming for.

app/scripts/metamask-controller.js
brad-decker: Resolved :D

brad-decker:
This PR is gonna start collecting other work from PRs that have gone through the review process. I'll compile the list of merged PRs into this one in the description

darkwing (dismissed):
Everything makes sense and this worked with EIP1559 off and on.  I'd love eyes with someone more familiar with our legacy gas controller.

danjm (approved):

darkwing (approved):

PeterYinusa:
LGTM!

PR #11426, opened 2021-06-30
Fix gas-modal-page-container.container check for custom gas price safety

danjm:
There is a bug in v9.7.0 that can be recreated like this:

1. Create a transaction with a gas price that is too low, so will be pending a long time
2. While the transaction is still pending, close metamask
3. Open the browser, open metamask
4. Try to "speed up" or cancel the pending transaction (before doing anything else)

Users will hit an error page with a "bignumber" error.

This is because we were calling `isCustomPriceSafeForCustomNetwork` when on mainnet but `isGasEstimate` is false. `isGasEstimate` is set to true after fetching the price when hitting the send screen. If there is no cached price, then this error will be hit.

The fix is to ensure that we only call `isCustomPriceSafeForCustomNetwork` when we are on a custom network.

### DIFF
--- a/ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.container.js
+++ b/ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.container.js
@@ -138,12 +138,10 @@ const mapStateToProps = (state, ownProps) => {
       });
   const isGasEstimate = getIsGasEstimatesFetched(state);
 
-  let customPriceIsSafe;
+  let customPriceIsSafe = true;
   if ((isMainnet || process.env.IN_TEST) && isGasEstimate) {
     customPriceIsSafe = isCustomPriceSafe(state);
-  } else if (isTestnet) {
-    customPriceIsSafe = true;
-  } else {
+  } else if (!(isMainnet || process.env.IN_TEST || isTestnet)) {
     customPriceIsSafe = isCustomPriceSafeForCustomNetwork(state);
   }
 
--- a/ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.container.js
+++ b/ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.container.js
@@ -41,6 +41,7 @@ import {
   isCustomPriceExcessive,
   getIsGasEstimatesFetched,
   getShouldShowFiat,
+  getIsCustomNetworkGasPriceFetched,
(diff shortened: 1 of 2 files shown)


ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.container.js
```diff
@@ -138,12 +138,10 @@ const mapStateToProps = (state, ownProps) => {
       });
   const isGasEstimate = getIsGasEstimatesFetched(state);
 
-  let customPriceIsSafe;
+  let customPriceIsSafe = true;
```
danjm, line 141: Default to true if on a standard testnet or in the fallback scenario where the call for gas prices has failed (i.e. `isGasEstimate` is false). This was the intent of https://github.com/MetaMask/metamask-extension/pull/11111

ui/selectors/custom-gas.js
```diff
@@ -105,6 +105,10 @@ export function isCustomPriceSafeForCustomNetwork(state) {
     return true;
   }
 
+  if (!estimatedPrice) {
```
danjm, line 108: While not strictly required to fix the bug this PR aims to correct, this ensures that no errors are hit if on a testnet and `state.gas.basicEstimates.average` somehow got set to a falsy value. So this will prevent possible big number errors in the future.

brad-decker (dismissed):

ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.container.js
```diff
@@ -138,12 +138,10 @@ const mapStateToProps = (state, ownProps) => {
...
   if ((isMainnet || process.env.IN_TEST) && isGasEstimate) {
     customPriceIsSafe = isCustomPriceSafe(state);
-  } else if (isTestnet) {
-    customPriceIsSafe = true;
-  } else {
+  } else if (!(isMainnet || process.env.IN_TEST || isTestnet)) {
```
adonesky1, line 144: Don't we need to check if gasEstimates have been fetched yet in this block as well?

danjm: Thanks. I added this check in the most recent commit.

brad-decker (approved):

adonesky1 (approved):
LGTM!

PR #11429, opened 2021-06-30
Wire up send flow to gasFeeController

brad-decker:
starts wiring up the send flow with new controller. This PR specifically works only when the SHOW_EIP_1559_UI flag is set to 1, and it only operates on the gasPrice field (actual EIP 1559 support in send flow will be in a later PR)


ui/store/actions.js
```diff
@@ -2712,10 +2727,21 @@ export async function updateTokenType(tokenAddress) {
...
+ *
+ * @returns {string} a unique identify of the polling request that can be used
+ *  to remove that request from consideration of whether polling needs to
+ *  continue.
+ */
 export function getGasFeeEstimatesAndStartPolling() {
   return promisifiedBackground.getGasFeeEstimatesAndStartPolling();
 }
 
+export function disconnectGasFeeEstimatePoller(pollToken) {
```
brad-decker, line 2741: @darkwing this will be helpful for you, this is what should be called in the return method of that useEffect hook so that it stops polling.

ui/pages/send/send-content/send-content.component.js
```diff
@@ -57,7 +57,7 @@ export default class SendContent extends Component {
           {this.maybeRenderAddContact()}
           <SendAssetRow />
           <SendAmountRow />
-          <SendGasRow />
+          {process.env.SHOW_EIP_1559_UI ? null : <SendGasRow />}
```
brad-decker, line 60: In our new flow the gas is not editable on the send screen, edits will only happen on confirmations.

ui/pages/send/send-content/send-amount-row/amount-max-button/amount-max-button.js
```diff
@@ -29,17 +29,20 @@ export default function AmountMaxButton() {
     dispatch(toggleSendMaxMode());
   };
 
+  const disabled = process.env.SHOW_EIP_1559_UI
+    ? isDraftTransactionInvalid
+    : buttonDataLoading || isDraftTransactionInvalid;
```
brad-decker, line 34: we only need to consider the button loading state in the old UI flow

ui/ducks/send/send.js
```diff
@@ -995,9 +1021,11 @@ const slice = createSlice({
         // the gasPrice in our slice. We call into the caseReducer
         // updateGasPrice to also tap into the appropriate follow up checks
         // and gasTotal calculation.
-        slice.caseReducers.updateGasPrice(state, {
-          payload: getGasPriceInHexWei(action.value.average),
-        });
+        if (Boolean(process.env.SHOW_EIP_1559_UI) === false) {
+          slice.caseReducers.updateGasPrice(state, {
+            payload: getGasPriceInHexWei(action.value.average),
```
brad-decker, line 1026: we won't have this value if the flag is set so don't attempt to set it, further more our estimate comes from the background instead of an API call/redux slice.

ui/ducks/send/send.js
```diff
@@ -354,19 +355,37 @@ export const initializeSendState = createAsyncThunk(
...
     // Default gasPrice to 1 gwei if all estimation fails
-    const gasPrice =
-      basicEstimateStatus === BASIC_ESTIMATE_STATES.READY
-        ? getGasPriceInHexWei(basicEstimates.average)
-        : '0x1';
+    let gasPrice = '0x1';
+    let basicEstimateStatus = BASIC_ESTIMATE_STATES.LOADING;
+    let gasEstimatePollToken = null;
+
+    if (Boolean(process.env.SHOW_EIP_1559_UI) === false) {
```
brad-decker, line 364: use the old way if the flag is falsy.

ui/ducks/send/send.js
```diff
@@ -354,19 +355,37 @@ export const initializeSendState = createAsyncThunk(
...
+        gasPrice = getGasPriceInHexWei(basicEstimates.average);
+      }
+    } else {
+      // Instruct the background process that polling for gas prices should begin
+      gasEstimatePollToken = await getGasFeeEstimatesAndStartPolling();
+      const {
+        metamask: { gasFeeEstimates },
+      } = thunkApi.getState();
+
+      basicEstimateStatus = BASIC_ESTIMATE_STATES.READY;
```
brad-decker, line 384: kind of hacky using a constnat from another slice but we need to know this for other parts of this method.

darkwing (approved):
LGTM; I tested manually and worked well; code makes sense!

ryanml (approved):
lgtm!!

PR #11435, opened 2021-07-01
Add list of custom networks with special gas limit buffers

danjm:
This PR adds special gas limit buffer multipliers for specific custom networks. In the future we will look for a way to standardize this so that specific networks do not need to be hardcoded; this is a temporary solution.


brad-decker (approved):
Looks solid to me!

ryanml (approved):
++

PR #11443, opened 2021-07-02
Use ESM imports in migrations index.

tmpfs:
So that spack does not complain about mixing CommonJs and ESM.

Fixes: N/A

Explanation:  

I am working with @kumavis on LavaMoat and researching creation of a bundler that could be deeply integrated with LavaMoat/SES. Whilst doing this research I ran into an issue using [the SWC library](https://github.com/swc-project/swc/) with the migrations code which is using a mix of ESM -> CommonJs -> ESM that caused bundling to fail using `spack`.

Looking at [migrations/index.js](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/migrations/index.js) I can't think of any reason why `require()` should be used instead of static ESM imports.

This PR replaces the use of `require()` with `import` and allows us to bundle using `spack`.

Manual testing steps:  

I don't know enough about the repository to test this thoroughly so would appreciate some help verifying this works with the existing build process. In theory this should just work.

### DIFF
--- a/app/scripts/migrations/index.js
+++ b/app/scripts/migrations/index.js
@@ -5,68 +5,130 @@
 // The `migrate` function receives the previous
 // config data format, and returns the new one.
 
+import m002 from './002';
+import m003 from './003';
+import m004 from './004';
+import m005 from './005';
+import m006 from './006';
+import m007 from './007';
+import m008 from './008';
+import m009 from './009';
+import m010 from './010';
+import m011 from './011';
+import m012 from './012';
+import m013 from './013';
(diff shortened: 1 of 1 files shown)


tmpfs:

Gudahtt (approved):
LGTM! I had meant to do this ages ago. I also can't think of a reason not to use `import` here.

These aren't tested comprehensively by our test suite right now, but some of them are run during e2e tests. So the fact that they passed is a good sign.

brad-decker (approved):

PR #11447, opened 2021-07-02
Add some metrics events to the 3box controller

danjm:
Adds some metrics events to the 3box controller. This will inform some upcoming migrations and deprecations we have to do in relation to this feature.


ryanml (approved):
lgtm!

brad-decker (approved):

PR #11450, opened 2021-07-02
Add form-field component and new account view

adonesky1:
- add form-field component which takes the detail additions to an input from the advanced-gas-control-row and makes it more reusable in other contexts.
- Swaps out old advanced-gas-control-row in advancedGasControls for this new component, with slightly tweaked props
-  Use this new form-field component with a new password creation view as part of the redesigned onboarding flow: 
Screen Shot 2021-07-02 at 4 06 31 PM


ui/pages/onboarding-flow/new-account/new-account.js
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    setConfirmPassword(confirmPasswordInput);
+    setConfirmPasswordError(error);
+  };
+
+  const handleCreate = async (event) => {
+    event.preventDefault();
+
+    if (!isValid) {
```
jasonzhouu, line 76: Just a little suggest. When it's not valid, the button is disabled, so maybe there's no need to check it here. 🙂

brad-decker (dismissed):
🤩

ui/pages/onboarding-flow/new-account/new-account.js
adonesky1: thanks @jasonzhouu, I think this can serve as a backup to prevent invalid account credentials so I will leave it for now.

ui/pages/onboarding-flow/new-account/new-account.js
```diff
@@ -0,0 +1,184 @@
...
+            titleDetail={
+              <button
+                className="new-account__form--password-button"
+                type="button"
+                onClick={(e) => {
+                  e.preventDefault();
+                  setShowPassword(!showPassword);
+                }}
+              >
+                {showPassword ? 'Hide' : 'Show'}
```
ryanml, line 124: can we translate these? thanks!

ryanml (approved):

brad-decker (approved):

PR #11451, opened 2021-07-02
move conversion utils to shared folder

brad-decker:
Going to need to have some shared methods for calculations, as we do a number of gas calculations in a few places. This is prep work for EIP-1559 related changes.


NiranjanaBinoy (approved):
LGTM

darkwing (approved):

PR #11433, opened 2021-06-30
Provide Estimate Data to EditGasDisplay

darkwing:
Manual testing steps:  
  - Create a transaction on Ropsten with a very very low gas fee
  - Click "Speed up"
  - See new form, able to play around with the inputs and radio buttons.

## To Do:
- [x] There's some weirdness initially where the numbers don't appear to match what they should -- something went a bit wrong with elevating state up.
- [x] I still need to add the dollar estimates:  
MONEY

PR #11462, opened 2021-07-06
Provide data wiring for Gas Timing component

darkwing:
Depends on: https://github.com/MetaMask/metamask-extension/pull/11433

Instead of requiring a ton of information be provided to this component, and since it's featured in multiple places in the UI, it makes sense to make this component as standalone as possible.

## ToDo:
- [x]  Get sign off that this logic is correct
- [x]  Create jest test
- [x] Add this component to all UIs that should display it


dan437 (dismissed):
Looks good to me! Can we do additional changes in another PR if needed, so this can be merged and used right away?

ui/components/app/gas-timing/gas-timing.component.js
```diff
@@ -1,12 +1,112 @@
...
 
-export default function GasTiming({ text, tooltipText, attitude }) {
+function toHumanReadableTime(milliseconds) {
+  const rtf = new Intl.RelativeTimeFormat(undefined, {
+    localeMatcher: 'best fit',
+    numeric: 'always',
+    style: 'long',
+  });
+
+  return rtf.format(milliseconds / 1000, 'second');
```
darkwing, line 19: I'm not convinced this API will work -- it prefixes `in` on its time estimates, which is problematic.

brad-decker:
```
rtf.formatToParts(100, "day");
// > [{ type: "literal", value: "in " },
// >  { type: "integer", value: "100", unit: "day" },
// >  { type: "literal", value: " days" }]
```
you could do something like this and use the pieces, just insert the `<` symbol but that won't be internationalized (might be in the wrong place?)

brad-decker (dismissed):
🎉

ui/components/app/gas-timing/gas-timing.component.js
```diff
@@ -1,12 +1,66 @@
...
+  }
+
+  const { low, medium, high } = gasFeeEstimates;
+
+  let text = '';
+  let attitude = '';
+
+  // Anything medium or faster is positive
+  if (
+    Number(maxPriorityFeePerGas) >= Number(medium.suggestedMaxPriorityFeePerGas)
```
darkwing, line 41: I'm seeing these properties as strings in a few places -- can we convert to numbers in `useGasFeeEstimates` and `useGasFeeInputs`, or is there reason not to?

darkwing: @brad-decker ^

ui/components/app/gas-timing/gas-timing.component.js
```diff
@@ -1,12 +1,66 @@
...
+  const toHumanReadableTime = (milliseconds = 1) => {
+    const seconds = Math.ceil(milliseconds / 1000);
+    if (seconds <= SECOND_CUTOFF) {
+      return t('gasTimingSeconds', [seconds]);
+    }
+    return t('gasTimingMinutes', [Math.ceil(seconds / 60)]);
+  };
+
+  // Don't show anything if we don't have enough information
+  if (isGasEstimatesLoading || maxPriorityFeePerGas < 1) {
```
danjm, line 30: What if `maxPriorityFeePerGas` is 0.5? Or any other number greater than 0 but less than 1? Then we should still show something right? Or is 1 a protocol enforced minimum to maxPriorityFeePerGas?

ui/components/app/gas-timing/gas-timing.component.js
```diff
@@ -1,12 +1,66 @@
...
+    return null;
+  }
+
+  const { low, medium, high } = gasFeeEstimates;
+
+  let text = '';
+  let attitude = '';
+
+  // Anything medium or faster is positive
```
danjm, line 39: Is the logic here consistent with mobile?

danjm (changes requested):
Left some questions/comments. It is possible none of these require action, depending on the answers

ui/components/app/gas-timing/gas-timing.component.js
darkwing: This was confirmed to me yesterday from @jakehaugen

ui/components/app/gas-timing/gas-timing.component.js
darkwing: Good point, removed this check!

ui/components/app/gas-timing/gas-timing.component.test.js
```diff
@@ -0,0 +1,105 @@
...
+    sinon.stub(useGasFeeEstimatesExport, 'useGasFeeEstimates').returns({
+      isGasEstimatesLoading: false,
+      gasFeeEstimates: MOCK_FEE_ESTIMATE,
+    });
+
+    sinon.stub(useGasFeeInputsExport, 'useGasFeeInputs').returns({
+      maxPriorityFeePerGas: 10,
+    });
+
+    const wrapper = shallow(<GasTiming />);
```
dan437, line 74: It can probably be done in some future PR, but it would be awesome if we would use `react-testing-library` instead, since we are migrating away from Enzyme.

dan437 (dismissed):
It looks great. I was wondering how to get timing info in Swaps for this component, but you have resolved it in this PR, so I can use this component without any props!

ui/components/app/gas-timing/gas-timing.component.js
```diff
@@ -1,12 +1,66 @@
...
 import { TYPOGRAPHY } from '../../../helpers/constants/design-system';
-import InfoTooltip from '../../ui/info-tooltip';
 
-export default function GasTiming({ text, tooltipText, attitude }) {
+// Once we reach this second threshold, we switch to minutes as a unit
+const SECOND_CUTOFF = 90;
+
+export default function GasTiming() {
+  const { gasFeeEstimates, isGasEstimatesLoading } = useGasFeeEstimates();
+  const { maxPriorityFeePerGas } = useGasFeeInputs();
```
darkwing, line 16: This isn't going to work -- we'll need to pass this in as props.

ui/components/app/gas-timing/gas-timing.component.js
```diff
@@ -1,12 +1,65 @@
...
+
+  const t = useContext(I18nContext);
+
+  // Shows "seconds" as unit of time if under SECOND_CUTOFF, otherwise "minutes"
+  const toHumanReadableTime = (milliseconds = 1) => {
+    const seconds = Math.ceil(milliseconds / 1000);
+    if (seconds <= SECOND_CUTOFF) {
+      return t('gasTimingSeconds', [seconds]);
+    }
+    return t('gasTimingMinutes', [Math.ceil(seconds / 60)]);
```
danjm, line 25: This would show both 61 seconds and 119 seconds as "2 minutes". Is that okay?

danjm (dismissed):
I left one other question, but I won't block on it.

ui/components/app/gas-timing/gas-timing.component.js
darkwing: The 61 second case would not happen, as `SECOND_CUTOFF` is 90 -- it would display as 61 seconds.

dan437 (dismissed):

ui/components/app/gas-timing/gas-timing.component.js
danjm: oh, yeah, oops

darkwing:
Rebased!

danjm (dismissed):

ui/components/app/gas-timing/gas-timing.component.js
```diff
@@ -15,18 +68,10 @@ export default function GasTiming({ text, tooltipText, attitude }) {
...
 GasTiming.propTypes = {
-  text: PropTypes.string.isRequired,
-  tooltipText: PropTypes.string,
-  attitude: PropTypes.oneOf(['positive', 'negative', 'warning', '']),
-};
-
-GasTiming.defaultProps = {
-  tooltipText: '',
-  attitude: '',
+  maxPriorityFeePerGas: PropTypes.object,
```
dan437, line 76: Shouldn't this be: `PropTypes.string.isRequired` for now or `PropTypes.number.isRequired` once we only pass numbers in it?

ui/components/app/transaction-detail/transaction-detail.stories.js
```diff
@@ -21,7 +21,7 @@ const rows = [
     }
     detailText="0.00896 ETH"
     detailTotal="$15.73"
-    subTitle={<GasTiming text="Very likely in < 15 seconds" />}
+    subTitle={<GasTiming />}
```
dan437, line 24: I guess we will need to update Storybook with the right props in some future PRs? Assuming that `maxPriorityFeePerGas` is mandatory.

ui/components/app/gas-timing/gas-timing.component.js
dan437: In some places `maxPriorityFeePerGas` is in decimal GWEI and in other places it's in hex WEI. Do we want to use some postfix to make it clearer, e.g. `maxPriorityFeePerGasHexWEI` or `maxPriorityFeePerGasDecGWEI`?

dan437 (approved):
Looks good!

adonesky1 (approved):
LGTM

PR #11467, opened 2021-07-06
update implementation to match latest

brad-decker:
#### Metaswaps API support
The gas fee controller has been updated to use the legacy metaswaps API endpoint, and this PR updates the configuration such that it will work under those conditions.

#### Specific Types
the gas fee controller now exports GAS_ESTIMATE_TYPES which has 'NONE' 'LEGACY' 'FEE_MARKET' and 'ETH_GASPRICE' keys. These each correspond to a specific type of estimate state:

`NONE` refers to the state where an estimate has not yet been fetched
`LEGACY` refers to legacy metaswaps api estimates, (slow/avg/fast), but responds in the new low/medium/high format
`ETH_GASPRICE` refers to estimates gained from eth_gasPrice
`FEE_MARKET` refers to new estimates in the low/medium/high and estimatedBaseFee format

### DIFF
--- a/ui/ducks/metamask/metamask.js
+++ b/ui/ducks/metamask/metamask.js
@@ -288,6 +288,10 @@ export function isEIP1559Network(state) {
   return state.metamask.networkDetails.EIPS[1559] === true;
 }
 
+export function getGasEstimateType(state) {
+  return state.metamask.gasEstimateType;
+}
+
 export function getGasFeeEstimates(state) {
   return state.metamask.gasFeeEstimates;
 }
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -10,7 +10,7 @@ import {
   multiplyCurrencies,
   subtractCurrencies,
 } from '../../../shared/modules/conversion.utils';
-import { GAS_LIMITS } from '../../../shared/constants/gas';
+import { GAS_ESTIMATE_TYPES, GAS_LIMITS } from '../../../shared/constants/gas';
 import {
   CONTRACT_ADDRESS_ERROR,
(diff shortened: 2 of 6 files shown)


ui/ducks/metamask/metamask.js
```diff
@@ -288,6 +288,10 @@ export function isEIP1559Network(state) {
   return state.metamask.networkDetails.EIPS[1559] === true;
 }
 
+export function getGasEstimateType(state) {
+  return state.metamask.gasEstimateType;
```
brad-decker, line 292: New selector for getting the gasEstimateType

ui/ducks/send/send.js
```diff
@@ -414,13 +414,20 @@ export const initializeSendState = createAsyncThunk(
...
       gasEstimatePollToken = await getGasFeeEstimatesAndStartPolling();
       const {
-        metamask: { gasFeeEstimates },
+        metamask: { gasFeeEstimates, gasEstimateType },
       } = thunkApi.getState();
 
-      basicEstimateStatus = BASIC_ESTIMATE_STATES.READY;
-      if (gasFeeEstimates.gasPrice) {
-        gasPrice = gasFeeEstimates.gasPrice;
+      if (gasEstimateType === GAS_ESTIMATE_TYPES.LEGACY) {
```
brad-decker, line 420: Here's an example of using the gasEstimateType to determine which fees to grab

darkwing:
This looks good to me but we shouldn't merge due to package.json change, correct?  We need a release?

brad-decker:
@darkwing this is merging into the gas-fee-controller branch, rather than develop. The package json change updates the commit hash to the latest

PR #11470, opened 2021-07-07
Swaps: Add conditional routing to new APIs based on a feature flag

dan437:
## Explanation
This PR adds conditional routing to either v1 or v2 APIs for Swaps, based on feature flags from our backend. The goal is to migrate to v2 APIs only, but we want to have a fallback solution to v1, just to be on the safe side.

## Manual testing steps
v1 API calls:
- Open the `swaps.util.js` file and return the following from the `fetchSwapsFeatureFlags` function:
```
return {
    bsc: {
      mobile_active: false,
      extension_active: false,
      fallback_to_v1: true,
    },
    ethereum: {
      mobile_active: false,
      extension_active: false,
      fallback_to_v1: true,
    },
    polygon: {
      mobile_active: false,
      extension_active: false,
      fallback_to_v1: false,
    },
  };
```
- Run `yarn start` with a branch that includes this PR
- Open the extension
- On the main page click on Swaps
- Check that network calls are going to our old APIs

v2 API calls:
- Open the `swaps.util.js` file and return the following from the `fetchSwapsFeatureFlags` function:
```
return {
    bsc: {
      mobile_active: false,
      extension_active: true,
      fallback_to_v1: true,
    },
    ethereum: {
      mobile_active: false,
      extension_active: true,
      fallback_to_v1: true,
    },
    polygon: {
      mobile_active: false,
      extension_active: true,
      fallback_to_v1: false,
    },
  };
```
- Run `yarn start` with a branch that includes this PR
- Open the extension
- On the main page click on Swaps
- Check that network calls are going to our new APIs (`https://api2.metaswap.codefi.network/`)

## Screenshots
v2 /featureFlags API:
image

v1 API calls:
image

v1 successful swapping:
image

v2 API calls:
image

### DIFF
--- a/app/scripts/controllers/swaps.js
+++ b/app/scripts/controllers/swaps.js
@@ -19,7 +19,6 @@ import { isSwapsDefaultTokenAddress } from '../../../shared/modules/swaps.utils'
 
 import {
   fetchTradesInfo as defaultFetchTradesInfo,
-  fetchSwapsFeatureLiveness as defaultFetchSwapsFeatureLiveness,
   fetchSwapsQuoteRefreshTime as defaultFetchSwapsQuoteRefreshTime,
 } from '../../../ui/pages/swaps/swaps.util';
 import { MINUTE, SECOND } from '../../../shared/constants/time';
@@ -86,7 +85,6 @@ export default class SwapsController {
     getProviderConfig,
     tokenRatesStore,
     fetchTradesInfo = defaultFetchTradesInfo,
-    fetchSwapsFeatureLiveness = defaultFetchSwapsFeatureLiveness,
     fetchSwapsQuoteRefreshTime = defaultFetchSwapsQuoteRefreshTime,
     getCurrentChainId,
   }) {
@@ -95,7 +93,6 @@ export default class SwapsController {
     });
 
     this._fetchTradesInfo = fetchTradesInfo;
-    this._fetchSwapsFeatureLiveness = fetchSwapsFeatureLiveness;
     this._fetchSwapsQuoteRefreshTime = fetchSwapsQuoteRefreshTime;
     this._getCurrentChainId = getCurrentChainId;
 
--- a/app/scripts/controllers/swaps.js
+++ b/app/scripts/controllers/swaps.js
@@ -462,8 +462,9 @@ export default class SwapsController {
 
   setSwapsLiveness(swapsLiveness) {
     const { swapsState } = this.store.getState();
+    const { swapsFeatureIsLive, useNewSwapsApi } = swapsLiveness;
     this.store.updateState({
-      swapsState: { ...swapsState, ...swapsLiveness },
+      swapsState: { ...swapsState, swapsFeatureIsLive, useNewSwapsApi },
     });
   }
 
--- a/app/scripts/controllers/swaps.js
+++ b/app/scripts/controllers/swaps.js
@@ -73,6 +73,7 @@ const initialState = {
     topAggId: null,
     routeState: '',
     swapsFeatureIsLive: true,
+    useNewSwapsApi: false,
     swapsQuoteRefreshTime: FALLBACK_QUOTE_REFRESH_TIME,
   },
 };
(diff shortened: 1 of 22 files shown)


ui/pages/swaps/build-quote/build-quote.js
```diff
@@ -141,7 +141,9 @@ export default function BuildQuote({
   const toTokenIsNotDefault =
     selectedToToken?.address &&
     !isSwapsDefaultTokenAddress(selectedToToken?.address, chainId);
-  const occurances = Number(selectedToToken?.occurances || 0);
+  const occurrences = Number(
+    selectedToToken?.occurances || selectedToToken?.occurrences || 0,
```
danjm, line 145: Is this change because the `occurances` property will be spelled differently on api v2?

dan437: Yup, v2 uses correct naming, but here I support both v1 and v2 versions.

shared/constants/network.js
```diff
@@ -4,6 +4,9 @@ export const KOVAN = 'kovan';
 export const MAINNET = 'mainnet';
 export const GOERLI = 'goerli';
 export const NETWORK_TYPE_RPC = 'rpc';
+export const ETHEREUM = 'ethereum';
```
danjm, line 7: could these three constants be moved to `shared/constants/swaps`?

I believe they will only be used in swaps, because we already use `'mainnet'` as the name for `'ethereum'`, and for all custom networks they are identified by chainId.

dan437: Sure, will move them there.

app/scripts/controllers/swaps.js
```diff
@@ -449,22 +460,22 @@ export default class SwapsController {
     this.store.updateState({ swapsState: { ...swapsState, routeState } });
   }
 
-  setSwapsLiveness(swapsFeatureIsLive) {
+  setSwapsLiveness(swapsLiveness) {
```
danjm, line 463: The removal of the `swapsFeatureIsLive` param, along with spreading the new param onto state, was a bit confusing when I first read through this. It might be good to make it obvious which properties are getting set on state by the `swapsLiveness` parameter. Also, with this change, any property in the object passed to `setSwapsLiveness` will be written to state.

It might be good to destructure `swapsLiveness` and explictly write the `swapsFeatureIsLive` and `useNewSwapsApi` properties in the `updateState` call below.

dan437: That's great feedback! I will explicitly use these 2 props when updating state to make it clearer and safer.

ui/ducks/swaps/swaps.js
```diff
@@ -395,15 +404,22 @@ export const fetchQuotesAndSetQuoteState = (
...
+    let swapsLivenessForNetwork = {
+      swapsFeatureIsLive: false,
+      useNewSwapsApi: false,
+    };
     try {
-      swapsFeatureIsLive = await fetchSwapsFeatureLiveness(chainId);
+      const swapsFeatureFlags = await fetchSwapsFeatureFlags();
+      swapsLivenessForNetwork = getSwapsLivenessForNetwork(
+        swapsFeatureFlags,
+        getCurrentChainId(getState()),
```
danjm, line 415: Nit: `getState()` is called above when the `state` const is set. The `state` const could be used here instead of called `getState()` again.

danjm: Actually, the same is true for `chainId` and `getCurrentChainId`. So instead of `getCurrentChainId(getState())` here, it can just be `chainId`

ui/ducks/swaps/swaps.js
```diff
@@ -600,15 +616,22 @@ export const signAndSendTransactions = (history, metaMetricsEvent) => {
...
```
danjm, line 627: `chainId` is set on line 617, so `getCurrentChainId(getState())` can be replaced with `chainId`

danjm:
This function is in `ui/ducks/gas/gas.duck.js`:

```
async function basicGasPriceQuery() {
  const url = `https://api.metaswap.codefi.network/gasPrices`;
  return await fetchWithCache(
    url,
    {
      referrer: url,
      referrerPolicy: 'no-referrer-when-downgrade',
      method: 'GET',
      mode: 'cors',
    },
    { cacheRefreshTime: 75000 },
  );
}
```

This is used for fetching gas prices for transactions outside of swaps (the standard send/confirm flows in metamask).

Does this need to be updated as well?

danjm (changes requested):
Left a few comments. Most are relatively minor, the issue with the `basicGasPriceQuery` is probably most significant.

dan437:
Hi, thank you a lot for the review! I've resolved all your comments except of the `basicGasPriceQuery` one. I will check with the team, but I believe we can keep using the v1 API there, because v1 will be still running for a while. Once we push these changes to production and verify that everything is fine with v2 after running it for a while, we can point `basicGasPriceQuery` to the v2 API.

app/scripts/controllers/swaps.js
```diff
@@ -73,6 +72,7 @@ const initialState = {
     topAggId: null,
     routeState: '',
     swapsFeatureIsLive: true,
+    useNewSwapsApi: false,
```
darkwing, line 75: Instead of `useNewSwapsApi`, should we use a more meaningful setting name and value, like API version number?  `new` is a relative term and isn't very meaningful after a certain amount of time.

danjm: I think this will only be in place for a few weeks. Once the switch to v2 is made, this property will be removed.

danjm (dismissed):
LGTM

app/scripts/controllers/swaps.js
dan437: I agree with you, @darkwing , if it would be a long term solution. As Dan mentioned, once we switch to v2 and we feel comfortable with it, this and calls to v1 will be removed.

danjm (approved):

darkwing (approved):
Worked great for me!

PR #11477, opened 2021-07-08
fix up ens/hex address validation error handling

adonesky1:
Fixes: The following issue raised in the [v9.8.0 bug/regression report](https://docs.google.com/document/d/1-RpPcuYaPRmubH7fCBNN4eyEwfh_QyzgOY0VDSLtcKk/edit#) : Upon landing on the send/add recipient page, while on a network that doesn’t support ENS, the error “Network does not support ENS” is shown immediately. Additionally, entering an incorrect address should update the error to “Recipient address is invalid”, screenshot is an address without the last character. Before, user input would trigger the appropriate error messages and their priority/case.

### DIFF
--- a/ui/ducks/ens.js
+++ b/ui/ducks/ens.js
@@ -92,18 +92,20 @@ const slice = createSlice({
     },
     disableEnsLookup: (state) => {
       state.stage = 'NO_NETWORK_SUPPORT';
-      state.error = ENS_NOT_SUPPORTED_ON_NETWORK;
+      state.error = null;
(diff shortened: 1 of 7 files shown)


brad-decker (dismissed):
Changes seem reasonable to me

adonesky1:
@brad-decker just pushed lint/test fixes

danjm (commented):
LGTM

danjm (approved):

brad-decker (approved):

PR #11480, opened 2021-07-08
add useGasInputs hook

brad-decker:
Uses the new useGasEstimates hook and state hooks to return values and properly calculates display related fiat conversions.

### DIFF
--- a/ui/hooks/useGasFeeInputs.js
+++ b/ui/hooks/useGasFeeInputs.js
@@ -1,0 +1,311 @@
+import { addHexPrefix } from 'ethereumjs-util';
+import { useCallback, useState } from 'react';
+import { useSelector } from 'react-redux';
+import { GAS_ESTIMATE_TYPES } from '../../shared/constants/gas';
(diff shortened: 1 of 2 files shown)


brad-decker:
@darkwing the last commit has a breaking change from how it was hooked up in #11433 -- it adds two new fields: 'estimatedMaximumFiat' and 'estimatedMaximumNative'. The former is a direct copy of maxFeePerGasFiat because it uses the same value as its basis but that value will be valid for gasPrice not just maxFeePerGas. 

estimateMaximumNative was renamed from 'maxFeePerGasPrimary' -- note the names might change again as I iterate on this hook to work when users have selected various setting options.

I have added *VERY* limited test coverage and will be doing TDD from this point forward to get stuff to the UI and make sure all use cases are cared for.

dan437 (approved):
Looks great! As we discussed on Slack, if any updates are needed, it can be done in future PRs.

darkwing (approved):

adonesky1 (approved):

PR #11485, opened 2021-07-09
Ensure simple send gas estimation happens after the recipient is iden…

danjm:
Issue found in QA:

>Custom Network Gas Limits have increased not only for Optimism, but for xDai and Binance. Their eth_estimateGas returns 53000 on those networks for just “simple sends” when 21000 is default/sufficient for “simple sends”. I have a feeling we are incorrectly passing the incorrect params to eth_estimateGas for xDai/Binance. It seems that params for eth_estimateGas just need to be params: [{“to”: “ADDRESS”}] for simple sends, and params: [{“to”: “ADDRESS”, data: “DATA”}] for txs with data. Looks like we are passing “from” as the param, rather than “to”. Possible further discussion on here or Slack.

This was happening because without the 'to' address in the parameters passed to `estimateGas`, estimates for "simple sends" were actually returning estimates for contract deployments.

This PR fixes this by ensuring that estimateGas calls only happen once the recipient is set.

### DIFF
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -437,7 +437,8 @@ export const initializeSendState = createAsyncThunk(
         : GAS_LIMITS.SIMPLE;
     if (
       basicEstimateStatus === BASIC_ESTIMATE_STATES.READY &&
-      stage !== SEND_STAGES.EDIT
+      stage !== SEND_STAGES.EDIT &&
+      recipient.address
     ) {
       // Run our estimateGasLimit logic to get a more accurate estimation of
       // required gas. If this value isn't nullish, set it as the new gasLimit
@@ -1267,12 +1268,9 @@ export function useMyAccountsForRecipientSearch() {
  * @returns {void}
  */
 export function updateRecipient({ address, nickname }) {
-  return async (dispatch, getState) => {
+  return async (dispatch) => {
     await dispatch(actions.updateRecipient({ address, nickname }));
-    const state = getState();
-    if (state.send.asset.type === ASSET_TYPES.TOKEN) {
-      await dispatch(computeEstimatedGasLimit());
-    }
+    await dispatch(computeEstimatedGasLimit());
   };
 }
 
--- a/ui/ducks/send/send.test.js
+++ b/ui/ducks/send/send.test.js
@@ -1400,7 +1400,7 @@ describe('Send Slice', () => {
         );
       });
 
-      it('should create actions to update recipient and recalculate gas limit if the asset is a token', async () => {
+      it('should create actions to reset recipient input and ens, calculate gas and then validate input', async () => {
         const tokenState = {
           metamask: {
             blockGasLimit: '',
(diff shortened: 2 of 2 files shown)


ui/ducks/send/send.test.js
```diff
@@ -1349,11 +1349,35 @@ describe('Send Slice', () => {
       };
 
       it('should create an action to update recipient', async () => {
```
adonesky1, line 1351:
```suggestion
      it('should create actions to update recipient and recalculate gas limit if the asset type is not set ', async () => {
```

ui/ducks/send/send.test.js
```diff
@@ -1451,14 +1484,23 @@ describe('Send Slice', () => {
...
+        expect(actionResult[2].type).toStrictEqual(
+          'send/computeEstimatedGasLimit/pending',
+        );
         expect(actionResult[3].type).toStrictEqual(
+          'metamask/gas/SET_CUSTOM_GAS_LIMIT',
+        );
+        expect(actionResult[4].type).toStrictEqual(
+          'send/computeEstimatedGasLimit/fulfilled',
+        );
+        expect(actionResult[5].type).toStrictEqual('ENS/resetResolution');
```
adonesky1, line 1502: I think if tests are re-run with #11477 merged this will fail 
```suggestion
        expect(actionResult[5].type).toStrictEqual('ENS/resetEnsResolution');
```

ui/ducks/send/send.test.js
```diff
@@ -1443,6 +1469,13 @@ describe('Send Slice', () => {
               address: 'Address',
```
adonesky1, line 1469: ^^^ This test should probably be renamed: 'should create actions to reset recipient input and ens, calculate gas and then validate input'

danjm: Done in 2c6dd9b8d

ui/ducks/send/send.js
```diff
@@ -1267,12 +1268,9 @@ export function useMyAccountsForRecipientSearch() {
...
  */
 export function updateRecipient({ address, nickname }) {
-  return async (dispatch, getState) => {
+  return async (dispatch) => {
     await dispatch(actions.updateRecipient({ address, nickname }));
-    const state = getState();
-    if (state.send.asset.type === ASSET_TYPES.TOKEN) {
-      await dispatch(computeEstimatedGasLimit());
-    }
+    await dispatch(computeEstimatedGasLimit());
```
brad-decker, line 1273: This will not result in over calling eth_gasEstimate because we will return GAS_LIMITS.SIMPLE early if its a eth tx on a L1 network?

brad-decker (dismissed):
one non blocking question.

ui/ducks/send/send.js
danjm: correct

adonesky1 (dismissed):
Looks like you've got a lint issue but otherwise LGTM

adonesky1 (approved):

brad-decker (approved):

PR #11496, opened 2021-07-09
provide skeleton for submitting gas form

brad-decker:
This PR continues the process of peeling out parts of the old UI. For example, instead of relying on the sidebar component for our SHOW_EIP_1559_UI logic branch, this PR now embeds the popovers directly in the Transaction list item. 

**What to expect if NOT behind the SHOW_EIP_1559_UI flag**
1. clicking speed up or cancel behaves exactly like it used to.

**What to expect if behind the SHOW_EIP_1559_UI flag**
1. Clicking speed up or cancel opens the edit gas popover directly
2. ~Clicking submit on these forms will successfully close the popover, but speed ups and cancels will fail because the 110% increase of price is not set.~ the background controller will actually bump the prices to safe minimums if no value supplied. So they *may* succeed but they won't reflect the values in the form

### DIFF
--- a/ui/components/app/edit-gas-popover/edit-gas-popover.component.js
+++ b/ui/components/app/edit-gas-popover/edit-gas-popover.component.js
@@ -8,12 +8,26 @@ import EditGasDisplay from '../edit-gas-display';
 import EditGasDisplayEducation from '../edit-gas-display-education';
 
 import { I18nContext } from '../../../contexts/i18n';
-import { hideModal, hideSidebar } from '../../../store/actions';
+import {
+  createCancelTransaction,
+  createSpeedUpTransaction,
+  hideModal,
+  hideSidebar,
+  updateTransaction,
+} from '../../../store/actions';
+
+export const EDIT_GAS_MODE = {
+  SPEED_UP: 'speed-up',
+  CANCEL: 'cancel',
+  MODIFY_IN_PLACE: 'modify-in-place',
+};
 
 export default function EditGasPopover({
   popoverTitle,
   confirmButtonText,
   editGasDisplayProps,
+  transaction,
+  mode,
   onClose,
 }) {
   const t = useContext(I18nContext);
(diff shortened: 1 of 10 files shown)


ui/hooks/useCancelTransaction.js
```diff
@@ -34,47 +34,58 @@ export function useCancelTransaction(transactionGroup) {
...
+          // gasPrice/gasLimit to start at.
+          dispatch(setCustomGasPriceForRetry(customGasSettings.gasPrice));
+          dispatch(setCustomGasLimit(GAS_LIMITS.SIMPLE));
+        }
+        const tx = {
+          ...primaryTransaction,
+          txParams: {
+            ...primaryTransaction.txParams,
+            gas: GAS_LIMITS.SIMPLE,
+            value: '0x0',
```
brad-decker, line 62: We'll set these values in the EditGasPopover, or in the hook, because I'm positive they're used for gasLimit estimation at some point.

ui/hooks/useCancelTransaction.js
```diff
@@ -34,47 +34,58 @@ export function useCancelTransaction(transactionGroup) {
...
-                  newGasFee: userCustomizedGasTotal,
-                  transactionId: primaryTransaction.id,
-                  customGasSettings: newGasSettings,
-                }),
-              );
-            },
+      if (process.env.SHOW_EIP_1559_UI) {
+        setShowCancelEditGasPopover(true);
+      } else {
+        if (isLegacyTransaction(primaryTransaction)) {
```
brad-decker, line 50: this whole branch will go away. We'll do any necessary logic inside the EditGasPopover (setting initial gasPrice/maxFeePerGas/maxPriorityFeePerGas to be 10% higher)

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
```diff
@@ -37,6 +51,38 @@ export default function EditGasPopover({
     }
   }, [showSidebar, onClose, dispatch]);
 
+  const onSubmit = useCallback(() => {
+    if (transaction && mode) {
```
dan437, line 55: To avoid one level of nesting, we can do an early return:
```
if (!transaction || !mode) {
  return;
}
```

ui/components/app/transaction-list-item/transaction-list-item.component.js
```diff
@@ -203,6 +210,22 @@ export default function TransactionListItem({
           cancelDisabled={!cancelEnabled}
         />
       )}
+      {process.env.SHOW_EIP_1559_UI && showRetryEditGasPopover && (
+        <EditGasPopover
+          popoverTitle={t('cancelPopoverTitle')}
+          onClose={closeRetryEditGasPopover}
+          mode={EDIT_GAS_MODE.SPEED_UP}
+          transaction={transactionGroup.primaryTransaction}
```
dan437, line 218: I really like that we can pass the `transaction` prop now!

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
brad-decker: Done in latest

darkwing (approved):

adonesky1 (approved):

PR #11497, opened 2021-07-10
Setting tx param type before signing transactions

### LINKED ISSUE #11491: Set txParams.type before signing transactions in transaction controller
### Background
We imlemented `TransactionFactory.fromTxData` as part of #11288, but my understanding was the appropriate `type` parameter was set in this method based on the presence of fields. I realize now that determining the transaction type programmatically is error prone at scale (AccessLists can have gasPrice OR the EIP 1559 fields)

For MVP we need to add the type before signing either 0 for legacy, or 2 for EIP 1559. This can be done where we build the txParams with extra properties before signing:
https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/controllers/transactions/index.js#L729-L733

and [isEIP1559Transaction](https://github.com/MetaMask/metamask-extension/blob/develop/shared/modules/transaction.utils.js#L10-L22) can be used to determine if it should be type 2.

[The code for fromTxData uses BigNumber](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/tx/src/transactionFactory.ts#L29-L40) so we can use our hex encoded constants for [transaction envelope type](https://github.com/MetaMask/metamask-extension/blob/develop/shared/constants/transaction.js#L84-L88)

ryanml:
Fixes MetaMask/metamask-extension#11491

### DIFF
--- a/app/scripts/controllers/transactions/index.js
+++ b/app/scripts/controllers/transactions/index.js
@@ -23,6 +23,7 @@ import { getSwapsTokensReceivedFromTxMeta } from '../../../../ui/pages/swaps/swa
 import {
   TRANSACTION_STATUSES,
   TRANSACTION_TYPES,
+  TRANSACTION_ENVELOPE_TYPES,
 } from '../../../../shared/constants/transaction';
 import { METAMASK_CONTROLLER_EVENTS } from '../../metamask-controller';
 import { GAS_LIMITS } from '../../../../shared/constants/gas';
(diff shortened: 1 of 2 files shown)


SirJosh1987 (approved):

app/scripts/controllers/transactions/index.js
```diff
@@ -726,7 +727,11 @@ export default class TransactionController extends EventEmitter {
     const txMeta = this.txStateManager.getTransaction(txId);
     // add network/chain id
     const chainId = this.getChainId();
+    const type = isEIP1559Transaction(txMeta)
```
brad-decker, line 730: One final thing I forgot is that we should honor the type provided by dapps if they set type themselves.

brad-decker (changes requested):
One minor req, thanks for taking this one on!

app/scripts/controllers/transactions/index.js
brad-decker: oh i see that the type parameter is set *before* the spread of txParams so this is already cared for!

brad-decker (approved):

adonesky1 (approved):

PR #11501, opened 2021-07-11
Removing dead post-send tx code from ConfirmTransactionBase

ryanml:
Screen Shot 2021-07-11 at 2 15 17 AM

This code wasn't being reached as the component is unmounted at that point, the same logic is already covered in the `updateAndApproveTx` action (https://github.com/MetaMask/metamask-extension/blob/develop/ui/store/actions.js#L711)

**Test Plan:**
1. Create a send transaction
2. Confirm the transaction
3. Confirm that the error is no longer present once redirection the home page happens

### DIFF
--- a/ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
+++ b/ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
@@ -693,8 +693,6 @@ export default class ConfirmTransactionBase extends Component {
       sendTransaction,
       clearConfirmTransaction,
       txData,
-      history,
-      mostRecentOverviewPage,
       updateCustomNonce,
       maxFeePerGas,
       maxPriorityFeePerGas,
@@ -733,18 +731,7 @@ export default class ConfirmTransactionBase extends Component {
         this._removeBeforeUnload();
 
         sendTransaction(txData)
-          .then(() => {
-            clearConfirmTransaction();
-            this.setState(
-              {
-                submitting: false,
-              },
-              () => {
-                history.push(mostRecentOverviewPage);
-                updateCustomNonce('');
-              },
-            );
-          })
+          .then(() => clearConfirmTransaction())
           .catch((error) => {
             this.setState({
               submitting: false,


ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
```diff
@@ -590,18 +588,7 @@ export default class ConfirmTransactionBase extends Component {
...
 
         sendTransaction(txData)
-          .then(() => {
-            clearConfirmTransaction();
-            this.setState(
-              {
-                submitting: false,
-              },
-              () => {
-                history.push(mostRecentOverviewPage);
```
brad-decker, line 600: Without this does the user still return to their mostRecentOverviewPage when done confirming tx?

adonesky1: Also do we not need to set submitting to false if there is no error?

ryanml: @brad-decker the user was still being returned yes, I can double check though

@adonesky1 issue here was at this point in the flow, the component is unmounted so state changes become moot. (If there's no error redirection happens) I'll confirm though that there isn't a latent race condition issue that I wasn't seeing

darkwing:
@ryanml Is this still relevant?

ryanml:
@darkwing - yes actually, thanks for bumping this up, i've rebased it. this particular component side code is still unnecessary, at least from my confirmations. @brad-decker @adonesky1 when you get a chance I'd appreciate a test from either of you to make sure I'm not missing something xD

ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
```diff
@@ -733,18 +731,7 @@ export default class ConfirmTransactionBase extends Component {
...
-                updateCustomNonce('');
```
NiranjanaBinoy, line 744: It seems like we are resetting the `custom nonce` over here I can see that it is done for the error condition as well, I wonder this is handled with the current changes as well.

ryanml:
Closing as stale

PR #11506, opened 2021-07-12
moving some properties to identifiable payload

brad-decker:
Moves the following fields to the identifiable payload:
1 - chain_id, this was previously on the `properties` payload, and isn't enough information to fully identify the transaction.
2 - type, this is a new thing, but doesn't provide enough details even in combination with other fields to identify the transaction.
3 - network (same as chainId)
4 - referrer - this was always on the properties object and doesn't relate to on-chain data directly. in any event it doesn't identify the tx.
5 - source - derived from the referrer.

### DIFF
--- a/app/scripts/controllers/transactions/index.js
+++ b/app/scripts/controllers/transactions/index.js
@@ -1238,13 +1238,15 @@ export default class TransactionController extends EventEmitter {
     this._trackMetaMetricsEvent({
       event,
       category: 'Transactions',
-      sensitiveProperties: {
-        type,
-        status,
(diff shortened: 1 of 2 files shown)


ryanml (dismissed):
++

darkwing (dismissed):

brad-decker:
@darkwing and @ryanml rebased

NiranjanaBinoy (approved):
LGTM

danjm (approved):

ryanml (approved):

PR #11510, opened 2021-07-12
EIP-1559 - Elevate gas properties to the Popover, disable submission if errors

darkwing:
Since we need to be able to validate data before submission, we'll elevate the state of gas fro the Display to the Popover.

### DIFF
--- a/shared/constants/gas.js
+++ b/shared/constants/gas.js
@@ -29,3 +29,12 @@ export const GAS_RECOMMENDATIONS = {
   MEDIUM: 'medium',
   HIGH: 'high',
 };
+
+/**
+ * These represent the different edit modes presented in the UI
+ */
+export const EDIT_GAS_MODES = {
+  SPEED_UP: 'speed-up',
+  CANCEL: 'cancel',
+  MODIFY_IN_PLACE: 'modify-in-place',
+};
--- a/ui/components/app/edit-gas-display/edit-gas-display.component.js
+++ b/ui/components/app/edit-gas-display/edit-gas-display.component.js
@@ -85,7 +85,7 @@ export default function EditGasDisplay({
             />
           </div>
         )}
-        {type === 'speed-up' && (
+        {mode === EDIT_GAS_MODES.SPEED_UP && (
           <div className="edit-gas-display__top-tooltip">
             <Typography
               color={COLORS.BLACK}
--- a/ui/components/app/edit-gas-display/edit-gas-display.component.js
+++ b/ui/components/app/edit-gas-display/edit-gas-display.component.js
@@ -1,7 +1,10 @@
 import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
 
-import { GAS_RECOMMENDATIONS } from '../../../../shared/constants/gas';
+import {
+  GAS_RECOMMENDATIONS,
+  EDIT_GAS_MODES,
+} from '../../../../shared/constants/gas';
 
 import Button from '../../ui/button';
 import Typography from '../../ui/typography/typography';
@@ -54,14 +57,13 @@ export default function EditGasDisplay({
   showAdvancedForm,
   setShowAdvancedForm,
   warning,
+  mode,
 }) {
   const t = useContext(I18nContext);
 
   const requireDappAcknowledgement = Boolean(
     dappSuggestedGasFee && !dappSuggestedGasFeeAcknowledged,
   );
-
-  console.log('defaultEstimateToUse: ', defaultEstimateToUse);
 
   return (
     <div className="edit-gas-display">
@@ -136,31 +138,33 @@ export default function EditGasDisplay({
             </Typography>
           </div>
         )}
-        {!requireDappAcknowledgement && (
-          <RadioGroup
-            name="gas-recommendation"
-            options={[
-              {
-                value: GAS_RECOMMENDATIONS.LOW,
-                label: t('editGasLow'),
-                recommended: defaultEstimateToUse === GAS_RECOMMENDATIONS.LOW,
-              },
-              {
-                value: GAS_RECOMMENDATIONS.MEDIUM,
-                label: t('editGasMedium'),
-                recommended:
-                  defaultEstimateToUse === GAS_RECOMMENDATIONS.MEDIUM,
-              },
-              {
-                value: GAS_RECOMMENDATIONS.HIGH,
-                label: t('editGasHigh'),
-                recommended: defaultEstimateToUse === GAS_RECOMMENDATIONS.HIGH,
-              },
-            ]}
-            selectedValue={estimateToUse}
-            onChange={setEstimateToUse}
-          />
-        )}
+        {!requireDappAcknowledgement &&
+          ![EDIT_GAS_MODES.SPEED_UP, EDIT_GAS_MODES.CANCEL].includes(mode) && (
+            <RadioGroup
+              name="gas-recommendation"
+              options={[
+                {
+                  value: GAS_RECOMMENDATIONS.LOW,
+                  label: t('editGasLow'),
+                  recommended: defaultEstimateToUse === GAS_RECOMMENDATIONS.LOW,
+                },
+                {
+                  value: GAS_RECOMMENDATIONS.MEDIUM,
+                  label: t('editGasMedium'),
+                  recommended:
+                    defaultEstimateToUse === GAS_RECOMMENDATIONS.MEDIUM,
+                },
+                {
+                  value: GAS_RECOMMENDATIONS.HIGH,
+                  label: t('editGasHigh'),
+                  recommended:
+                    defaultEstimateToUse === GAS_RECOMMENDATIONS.HIGH,
+                },
+              ]}
+              selectedValue={estimateToUse}
+              onChange={setEstimateToUse}
+            />
+          )}
         {!alwaysShowForm && (
           <button
             className="edit-gas-display__advanced-button"
@@ -212,6 +216,7 @@ export default function EditGasDisplay({
 EditGasDisplay.propTypes = {
   alwaysShowForm: PropTypes.bool,
   type: PropTypes.oneOf(['customize-gas', 'speed-up']),
+  mode: PropTypes.string,
   showEducationButton: PropTypes.bool,
   onEducationClick: PropTypes.func,
   dappSuggestedGasFee: PropTypes.number,
--- a/ui/components/app/edit-gas-display/edit-gas-display.component.js
+++ b/ui/components/app/edit-gas-display/edit-gas-display.component.js
@@ -1,4 +1,4 @@
-import React, { useState, useContext } from 'react';
+import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
 
 import { GAS_RECOMMENDATIONS } from '../../../../shared/constants/gas';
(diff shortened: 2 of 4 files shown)


ui/components/app/edit-gas-display/edit-gas-display.component.js
```diff
@@ -19,54 +19,49 @@ import AdvancedGasControls from '../advanced-gas-controls/advanced-gas-controls.
...
-    setGasLimit,
-    estimateToUse,
-    setEstimateToUse,
-    estimatedMinimumFiat,
-    estimatedMaximumFiat,
-    isMaxFeeError,
-    isMaxPriorityFeeError,
-    isGasTooLow,
-  } = useGasFeeInputs(defaultEstimateToUse);
+  console.log('defaultEstimateToUse: ', defaultEstimateToUse);
```
brad-decker, line 64: console.log here

brad-decker (dismissed):
forgot a console.log but looks good to me

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
```diff
@@ -55,26 +94,42 @@ export default function EditGasPopover({
     if (!transaction || !mode) {
       closePopover();
     }
+
+    const cancelSpeedUpGas = process.env.SHOW_EIP_1559_UI
+      ? {
+          gasLimit: decimalToHex(gasLimit),
```
darkwing, line 100: I need to figure out the root cause of `Cannot convert string to buffer. toBuffer only supports 0x-prefixed hex strings and this string was given: 0x5.100000001` when using this `decimalToHex` function.

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
```diff
@@ -55,26 +94,42 @@ export default function EditGasPopover({
     if (!transaction || !mode) {
       closePopover();
     }
+
+    const cancelSpeedUpGas = process.env.SHOW_EIP_1559_UI
```
brad-decker, line 98: i'd switch on gasEstimateType === GAS_ESTIMATE_TYPES.FEE_MARKET

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
brad-decker: Ah, you need to convert these using `decGWEIToHexWEI` from `shared/modules/conversion.util.js` it does two conversions, first from gwei to wei then dec to hex. Not gasLimit though, gasLimit is just dec to hex string.

brad-decker (changes requested):

shared/constants/gas.js
```diff
@@ -29,3 +29,12 @@ export const GAS_RECOMMENDATIONS = {
   MEDIUM: 'medium',
   HIGH: 'high',
 };
+
+/**
+ * These represent the different edit modes presented in the UI
+ */
+export const EDIT_GAS_MODES = {
```
brad-decker, line 36: [REQ] These are UI specific, they don't need to be in the shared folder. I'd recommend leaving these declared in the EditPopover as they primarily modify the behavior of this component and should likely only be imported when also importing EditPopover

ui/components/app/edit-gas-display/edit-gas-display.component.js
```diff
@@ -19,55 +22,49 @@ import AdvancedGasControls from '../advanced-gas-controls/advanced-gas-controls.
...
 
 import { I18nContext } from '../../../contexts/i18n';
-import { useGasFeeInputs } from '../../../hooks/useGasFeeInputs';
 
 export default function EditGasDisplay({
-  alwaysShowForm,
-  type,
-  showEducationButton,
+  alwaysShowForm = false,
+  type = 'customize-gas',
```
brad-decker, line 28: I think the 'mode' prop removes our need to have the 'type' prop?

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
```diff
@@ -55,26 +93,43 @@ export default function EditGasPopover({
     if (!transaction || !mode) {
       closePopover();
     }
+
+    const cancelSpeedUpGas =
+      gasEstimateType === GAS_ESTIMATE_TYPES.FEE_MARKET
+        ? {
+            gasLimit: decGWEIToHexWEI(gasLimit),
```
brad-decker, line 100: gasLimit is not in gwei, it needs to be just a straight conversion to hex. decimalToHex.

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
```diff
@@ -55,26 +93,43 @@ export default function EditGasPopover({
     if (!transaction || !mode) {
       closePopover();
     }
+
+    const cancelSpeedUpGas =
```
brad-decker, line 97: rename to newGasSettings and use it for all three 'modes' below

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
```diff
@@ -55,26 +93,43 @@ export default function EditGasPopover({
...
         break;
-      case EDIT_GAS_MODE.MODIFY_IN_PLACE:
+      case EDIT_GAS_MODES.MODIFY_IN_PLACE:
         dispatch(
           updateTransaction({
             ...transaction,
-            txParams: { ...transaction.txParams /** ...newGasSettings */ },
+            txParams: {
+              ...transaction.txParams,
+              ...(process.env.SHOW_EIP_1559_UI
```
brad-decker, line 122: this is essentially the same as the cancelSpeedUpGas, so you can use it here too, just add the gas key manually. it won't hurt anything to add both gasLimit and gas (only one will be used)

brad-decker (changes requested):
Looks good just a few little things.

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
darkwing: Good to know!  Updated!

shared/constants/gas.js
darkwing: Pushing back on this slightly; `GAS_RECOMMENDATIONS` is also a shared constant which is only used in the UI.  I think having it here will be really helpful here.

brad-decker: I didn't comment on the addition of `GAS_RECOMMENDATIONS` because we'll need that in the background now, too. We'll be using it when we set default gas estimations on incoming transactions from dapps. 

I understand the desire to place this here, but it is important that we keep the shared folder strict to its purpose of storing code shared by both `app/` and `ui/` folders to keep the module boundary line clearly defined.

brad-decker (dismissed):
LGTM with the exception of the location of the constant -- but that can either be moved here or we can wait until we are sure its not used in the background before we relocate it at some point in the future.

brad-decker (dismissed):

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
```diff
@@ -55,26 +96,36 @@ export default function EditGasPopover({
     if (!transaction || !mode) {
       closePopover();
     }
+
+    const newGasSettings =
+      gasEstimateType === GAS_ESTIMATE_TYPES.FEE_MARKET
+        ? {
+            gas: decimalToHex(gasLimit),
+            gasLimit: decimalToHex(gasLimit),
```
Gudahtt, line 104: This seems like a mistake :thinking: 
```suggestion
```

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
```diff
@@ -16,25 +28,54 @@ import {
...
+    gasLimit,
+    setGasLimit,
+    estimateToUse,
+    setEstimateToUse,
+    estimatedMinimumFiat,
+    estimatedMaximumFiat,
+    isMaxFeeError,
+    isMaxPriorityFeeError,
+    isGasTooLow,
+  } = useGasFeeInputs(defaultEstimateToUse);
```
Gudahtt, line 77: I don't have the proper context here, but passing this many props around seems like a red flag to me. This is quite difficult to understand. It was overly complex before too when it was isolated to the "edit-gas-display" component, and now it's worse because the complexity has found its way into a component API now as well. I don't see how we could make a component with over a dozen props easily understood or used.

A better way to handle UI complexity would be to track UI state in a Redux store, and grab it directly from Redux wherever a value is needed. Then derived state can be calculated using selectors. Then hopefully the UI logic would be just that - UI logic, not cluttered with the rules of how editing gas works.

Anyway - that's not something that was introduced here, nor can it be fixed here in this PR. Just commenting on it now that I've noticed it.

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
darkwing: Good shout, removed!

ui/components/app/edit-gas-display/edit-gas-display.component.js
```diff
@@ -141,31 +137,33 @@ export default function EditGasDisplay({
...
-                label: t('editGasHigh'),
-                recommended: defaultEstimateToUse === GAS_RECOMMENDATIONS.HIGH,
-              },
-            ]}
-            selectedValue={estimateToUse}
-            onChange={setEstimateToUse}
-          />
-        )}
+        {!requireDappAcknowledgement &&
+          ![EDIT_GAS_MODES.SPEED_UP, EDIT_GAS_MODES.CANCEL].includes(mode) && (
```
Gudahtt, line 141: No recommendations for speed up or cancellation?

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
```diff
@@ -83,11 +133,29 @@ export default function EditGasPopover({
...
+    gasLimit,
+    gasPrice,
+    maxFeePerGas,
+    maxPriorityFeePerGas,
+    gasEstimateType,
+  ]);
+
+  let title = t('editGasTitle');
+  if (popoverTitle) {
+    title = popoverTitle;
```
Gudahtt, line 150: Just noting that the title priorities have changed here as well. Previously the education title would be shown over a custom title, but no longer.

This seems more appropriate though I think :thinking:

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
brad-decker: Our goal is to get to a working solution in as iterative a process as possible. We've discussed moving much of this logic into redux but that will happen at a later point when we have a clearer picture of where we are heading.

Most values come directly from the background state, and are computed using conversion methods and other hooks (useCurrencyDisplay)

Other values come from the transaction being modified which we used to set in state and access in redux, or passed in through modal or popover.

So I'd like to take more time to think about how we integrate this into redux

Gudahtt (dismissed):
LGTM! Seems equivalent except up one layer, except that now the new gas settings are included with the speedup/cancellation call, and the form is disabled when there is an error. And there were a few more functional changes that seemed probably intentional, but I left comments so that someone can double-check.

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
brad-decker: Need this, its part of the new options bag for speed up and cancel methods, and transactionFactory from ethereumjs/tx looks for gasLimit too. We need this line for now we can figure out if its redundant later.

brad-decker: https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/controllers/transactions/index.js#L557-L625

ui/components/app/edit-gas-display/edit-gas-display.component.js
brad-decker: yeah i believe we should be showing recommendations for these. Previously we would just default the fields to 110% of the previous value but there was no warning for going under that. Now we're going to use present estimations but warn if the value is less than 110% of previous value. That work doesn't have to be done here but I'm also curious about this logic change.

darkwing:
Restored `gasLimit`

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
Gudahtt: It's far easier to iterate with smaller and simpler APIs than it is with enormously complex ones. I don't think an iterative approach is in conflict with good API design.

When I suggested Redux I hadn't looked closely enough to be honest. I see now that some of this is input field state, and that definitely makes sense to be at the component level. So maybe some of this is best here, and maybe some is best as utility functions or selectors or as hooks broken out into more logical chunks. Or maybe some should be in Redux, I don't know. But I'm sure we can come up with a more readable abstraction than this enormous list of props.

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
Gudahtt: That's disappointing. The type definition even states that there is no gas limit: https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/controllers/transactions/index.js#L56

ui/components/app/edit-gas-display/edit-gas-display.component.js
darkwing: I can restore it, but they aren't present in the Figma file:  https://www.figma.com/file/NXvUKcZ0Ocu1GPcT7VxRMx/1559---Gas-Controls?node-id=643%3A5

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
Gudahtt: What is the intended difference between `gas` and `gasLimit`? I'm finding this very hard to follow. `generateNewGasParams` for example uses the truthiness of `gasLimit` to decide whether to ensure `gas` was set, but it doesn't use `gasLimit` directly. It doesn't seem to be explained in any comments either :disappointed:

brad-decker: A big reason for the introduction of `gasLimit` is because `@ethereumjs/tx` has no mention of a `gas` field in its [`txData` type](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/tx/src/types.ts#L112-L163), but does specify `gasLimit`.

`generateNewGasParams` has some flaws in its types defined and I did not do a good enough job writing down my internal thoughts/strategy. I had assumed that the usage of `gasLimit` by `@ethereumjs/tx` was a strong signal that we should also be using it instead of `gas` but in the midst of that discovered that the JSON RPC spec does not use `gasLimit`. It doesn't matter because we only use that package to generate a serialized tx that abstracts away field names. 

My thought was to consolidate to using one field name, personally gasLimit makes the most sense to me and is more descriptive than gas, but I wanted to take the time to understand the impact that would have. You'll see in the signTransaction method I added gasLimit to the txParams before signing to match the type. 

All of this to say, I made some choices during individual units of work that led us here. I'll make it a priority to update the documentation of the new method and the create speed up/cancel methods. I don't think this should be addressed here. Happy to have further conversation about this.

ui/components/app/edit-gas-display/edit-gas-display.component.js
brad-decker: talked on the sync up this morning and they do not get recommendations right now we may restore at a later time.

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
Gudahtt: Hmm. So just to be clear - both properties have the same intended purpose? They're just duplicated for the aforementioned reasons.

I'm asking because your earlier comment made me think that both were needed for speed up and cancellation specifically, but not for other transactions.

brad-decker (approved):

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
brad-decker: yes, they are redundant fields, they ultimately set the `gas` value on txParams.

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
brad-decker: We know we can come up with something, as well. I don't disagree with your sentiment but we're doing the best we can with our current knowledge of the problem, in the time constraints we are under without breaking anything that is currently in place for transactions today (before the fork). The consolidation of logic into a hook was a purposeful step meant to get us into a place where abstracting that into a redux store (or wherever) would be easier in the future.

danjm (approved):

PR #11511, opened 2021-07-12
rely upon gas fee controller for gas price estimates

brad-decker:
Changes
* returns `false` from getNetworkEIP1559Compatibility if the `SHOW_EIP_1559_UI` environment variable isn't true. This allows testnets such as Ropsten, Goerli, Rinkeby, etc that have launched EIP-1559 to still get good `gasPrice` estimates until we release our UI elements.
* Plugs in values from the `GasFeeController` state into the custom-gas selectors. These selectors will be removed in favor of directly accessing the underlying state values (or creating new, more apt named selectors) in a future PR.
* updates tests to have mock data appropriate for these changes. I attempted to only modify test mock data or remove tests that were no longer relevant (covering removed code)
* Adds calls to start polling on the gasFeeController from a number of places that were previously calling the fetch method from gas.duck.js.
* Removes all fetch related code from gas.duck.js, it now ONLY exists to store custom data which will also go away as we get closer to merging our EIP-1559 work. 

Items remaining
- [x] The gasPrice on send page is being constantly overwritten with values. I need to tap into the customGasPrice state to see if it has been set and then avoid changing gasPrice once that happens. We'll need to figure this out for EIP-1559 too because the maxFeePerGas/gasPrice will update periodically but **isn't visible to the user**, resulting in possibly randomly displaying insufficient funds warnings? 
- [x] Failing test that I'm not sure why is happening. 

### QA / Testing Steps
- Use the send flow on various networks and ensure a reasonable gasPrice is set. this is 1 on rinkeby, 5 on Goerli, 10 on Kovan, 2 on Ropsten (as of this writing 7/14/2021 1:20 CST).
- On the send flow, on custom networks, switch networks and see that the gasPrice updates within a few seconds (no more than 10)
- On the send flow, if you change your gasPrice, your price does not change every 10 seconds.
- Confirmation edit gas price logic works on mainnet (I *believe* that custom networks do not use estimates they set the gasPrice to the transaction gasPrice, the only place estimates are displayed is in the buttons)
- speed up/cancel works

### Known Issues
<details>
<summary>Gas customization modal on mainnet does not update gasPrice values when estimates change</summary>

https://user-images.githubusercontent.com/4448075/125673410-f903853e-933d-4f5b-afb9-5cb81fe72ae7.mov

<br/>
<ul>
<li>I don't know how much time we should spend resolving this as it doesn't actually change the *result* of the behavior from mainnet. If the user had opened this window the gasPrices wouldn't have updated but it doesn't mean the expectations of the network hadn't changed.</li>
<li>Fixing this will require introducing a whole layer of complexity to the existing gas modal that we intend on deprecating soon</li>
</ul>

</details>

### DIFF
--- a/app/scripts/controllers/network/network.js
+++ b/app/scripts/controllers/network/network.js
@@ -162,6 +162,9 @@ export default class NetworkController extends EventEmitter {
    */
   async getEIP1559Compatibility() {
     const { EIPS } = this.networkDetails.getState();
+    if (process.env.SHOW_EIP_1559_UI === false) {
+      return false;
+    }
     if (EIPS[1559] !== undefined) {
       return EIPS[1559];
     }
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -195,8 +195,16 @@ export default class MetamaskController extends EventEmitter {
         this,
       ),
       legacyAPIEndpoint: `https://gas-api.metaswap.codefi.network/networks/<chain_id>/gasPrices`,
-      getCurrentNetworkLegacyGasAPICompatibility: () => true,
-      getChainId: () => this.networkController.getCurrentChainId(),
+      EIP1559APIEndpoint: `https://gas-api.metaswap.codefi.network/networks/<chain_id>/suggestedGasFees`,
+      getCurrentNetworkLegacyGasAPICompatibility: () => {
+        const chainId = this.networkController.getCurrentChainId();
+        return process.env.IN_TEST || chainId === MAINNET_CHAIN_ID;
+      },
+      getChainId: () => {
+        return process.env.IN_TEST
+          ? MAINNET_CHAIN_ID
+          : this.networkController.getCurrentChainId();
+      },
     });
 
     this.appStateController = new AppStateController({
@@ -2050,6 +2058,7 @@ export default class MetamaskController extends EventEmitter {
   }
 
   estimateGas(estimateGasParams) {
+    console.log(estimateGasParams);
     return new Promise((resolve, reject) => {
       return this.txController.txGasUtil.query.estimateGas(
         estimateGasParams,
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -194,11 +194,9 @@ export default class MetamaskController extends EventEmitter {
       getCurrentAccountEIP1559Compatibility: this.getCurrentAccountEIP1559Compatibility.bind(
         this,
       ),
-      getCurrentNetworkLegacyGasAPICompatibility: () =>
-        this.networkController.getCurrentChainId() === MAINNET_CHAIN_ID,
-      getChainId: this.networkController.getCurrentChainId.bind(
-        this.networkController,
-      ),
+      legacyAPIEndpoint: `https://gas-api.metaswap.codefi.network/networks/<chain_id>/gasPrices`,
+      getCurrentNetworkLegacyGasAPICompatibility: () => true,
+      getChainId: () => this.networkController.getCurrentChainId(),
     });
 
     this.appStateController = new AppStateController({
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -2058,7 +2058,6 @@ export default class MetamaskController extends EventEmitter {
   }
 
   estimateGas(estimateGasParams) {
-    console.log(estimateGasParams);
     return new Promise((resolve, reject) => {
       return this.txController.txGasUtil.query.estimateGas(
         estimateGasParams,
--- a/ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.component.js
+++ b/ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.component.js
@@ -2,6 +2,10 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import PageContainer from '../../../ui/page-container';
 import { Tabs, Tab } from '../../../ui/tabs';
+import {
+  disconnectGasFeeEstimatePoller,
+  getGasFeeEstimatesAndStartPolling,
+} from '../../../../store/actions';
 import AdvancedTabContent from './advanced-tab-content';
 import BasicTabContent from './basic-tab-content';
 
@@ -17,7 +21,6 @@ export default class GasModalPageContainer extends Component {
     updateCustomGasPrice: PropTypes.func,
     updateCustomGasLimit: PropTypes.func,
     insufficientBalance: PropTypes.bool,
-    fetchBasicGasEstimates: PropTypes.func,
     gasPriceButtonGroupProps: PropTypes.object,
     infoRowProps: PropTypes.shape({
       originalTotalFiat: PropTypes.string,
@@ -39,7 +42,15 @@ export default class GasModalPageContainer extends Component {
   };
 
   componentDidMount() {
-    this.props.fetchBasicGasEstimates();
+    getGasFeeEstimatesAndStartPolling().then((pollingToken) => {
+      this.setState({ pollingToken });
+    });
+  }
+
+  componentWillUnmount() {
+    if (this.state.pollingToken) {
+      disconnectGasFeeEstimatePoller(this.state.pollingToken);
+    }
   }
 
   renderBasicTabContent(gasPriceButtonGroupProps) {
--- a/ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.component.js
+++ b/ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.component.js
@@ -41,6 +41,13 @@ export default class GasModalPageContainer extends Component {
     customPriceIsExcessive: PropTypes.bool.isRequired,
   };
 
+  constructor(props) {
+    super(props);
+    this.state = {
+      pollingToken: undefined,
+    };
+  }
+
   componentDidMount() {
     getGasFeeEstimatesAndStartPolling().then((pollingToken) => {
       this.setState({ pollingToken });
--- a/ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.component.js
+++ b/ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.component.js
@@ -49,12 +49,18 @@ export default class GasModalPageContainer extends Component {
   }
 
   componentDidMount() {
+    this._isMounted = true;
     getGasFeeEstimatesAndStartPolling().then((pollingToken) => {
-      this.setState({ pollingToken });
+      if (this._isMounted) {
+        this.setState({ pollingToken });
+      } else {
+        disconnectGasFeeEstimatePoller(pollingToken);
+      }
     });
   }
 
   componentWillUnmount() {
+    this._isMounted = false;
     if (this.state.pollingToken) {
       disconnectGasFeeEstimatePoller(this.state.pollingToken);
     }
--- a/ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.container.js
+++ b/ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.container.js
@@ -10,7 +10,6 @@ import {
   setCustomGasPrice,
   setCustomGasLimit,
   resetCustomData,
-  fetchBasicGasEstimates,
 } from '../../../../ducks/gas/gas.duck';
 import {
   getSendMaxModeState,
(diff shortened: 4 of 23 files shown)


app/scripts/metamask-controller.js
```diff
@@ -2052,6 +2058,7 @@ export default class MetamaskController extends EventEmitter {
   }
 
   estimateGas(estimateGasParams) {
+    console.log(estimateGasParams);
```
danjm, line 2061: console.log

app/scripts/controllers/network/network.js
```diff
@@ -162,6 +162,9 @@ export default class NetworkController extends EventEmitter {
    */
```
danjm, line 162: :+1:

ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.container.js
```diff
@@ -10,7 +10,6 @@ import {
   setCustomGasPrice,
```
danjm, line 10: :+1:

ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.component.js
```diff
@@ -38,8 +41,23 @@ export default class GasModalPageContainer extends Component {
...
+    };
+  }
+
   componentDidMount() {
-    this.props.fetchBasicGasEstimates();
+    getGasFeeEstimatesAndStartPolling().then((pollingToken) => {
+      this.setState({ pollingToken });
+    });
+  }
+
```
danjm, line 56: Not sure if there is anything we can do about this:

Is it possible for the component to unmount before `getGasFeeEstimatesAndStartPolling()` resolves, thereby polling having been initiated in the controller but disconnect won't be called here, leaving the controller polling indefinitely?

ui/helpers/utils/conversions.util.js
```diff
@@ -80,6 +80,7 @@ export function getWeiHexFromDecimalValue({
   conversionRate,
   fromDenomination,
   invertConversionRate,
+  numberOfDecimals,
```
danjm, line 83: It seems that this new `numberOfDecimals` parameter is not used anywhere

brad-decker: Yes, needs to be removed didn't end up needing it!

ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.component.js
brad-decker: Hmmm.... good point. I think that affects our `useGasEstimates` hook, too. At least for this component, it won't matter long term because I plan on ceremoniously deleting it once @darkwing's efforts are fully realized. :D 

For the useGasEstimates hook we can do something like this 

```
useEffect(() => {
  let pollingToken;
  let isMounted = true;
  getGasFeeEstimatesAndStartPolling().then((token) => {
    if (!isMounted) {
      return disconnectMethodNameThatIForgot(token);
    }
    pollingToken = token;
  });
  return () => {
    isMounted = false;
    if (pollingToken) {
     //disconnect
    }
  }
}, []);
```

ui/pages/confirm-transaction/confirm-transaction.container.js
```diff
@@ -6,7 +6,6 @@ import {
   clearConfirmTransaction,
```
danjm, line 6: :+1:

ui/pages/confirm-send-ether/confirm-send-ether.container.js
```diff
@@ -17,9 +17,9 @@ const mapStateToProps = (state) => {
```
danjm, line 17: :+1:

ui/pages/confirm-send-ether/confirm-send-ether.component.js
```diff
@@ -16,8 +16,9 @@ export default class ConfirmSendEther extends Component {
```
danjm, line 16: :+1:

ui/hooks/useRetryTransaction.js
```diff
@@ -29,6 +26,10 @@ import { useIncrementedGasFees } from './useIncrementedGasFees';
 export function useRetryTransaction(transactionGroup) {
   const { primaryTransaction } = transactionGroup;
   const isMainnet = useSelector(getIsMainnet);
+
+  // Starts polling for gas prices that will be used for retry gas pricing
+  useGasFeeEstimates();
```
danjm, line 31: This might be a bit nitpicky: it feels like it would be nice to separate the "get gas fee estimates" concern from the "start polling" concern. Calling `useGasFeeEstimates()` in some cases for the returned data, and in other cases for the side effect of starting polling, seems like it could be a bit confusing to future readers/maintainers

danjm: But also, why is `useGasFeeEstimates` called here and not in a component?

danjm: And is this also needed for cancels?

brad-decker: agreed, but the good news is this is a temporary measure until users are using the new UI.

brad-decker: I can just copy/paste the polling code from that hook here instead

ui/ducks/gas/gas.duck.js
```diff
@@ -78,138 +36,11 @@ export default function reducer(state = initState, action) {
...
-      timeRetrieved,
-    ),
-  ]);
-
-  return basicEstimates;
-}
-
-export function setCustomGasPriceForRetry(newPrice) {
-  return async (dispatch) => {
-    if (newPrice === '0x0') {
```
danjm, line 197: This had a guard against trying to retry transaction with a `0x0` price. I guess that is unnecessary now because the price will always come from the controller?

app/scripts/metamask-controller.js
```diff
@@ -194,11 +194,17 @@ export default class MetamaskController extends EventEmitter {
       getCurrentAccountEIP1559Compatibility: this.getCurrentAccountEIP1559Compatibility.bind(
```
danjm, line 194: :+1:

ui/pages/confirm-transaction/confirm-transaction.component.js
```diff
@@ -49,27 +52,34 @@ export default class ConfirmTransaction extends Component {
...
       transaction: { txParams: { data, to } = {} } = {},
-      fetchBasicGasEstimates,
       getContractMethodData,
       transactionId,
       paramsTransactionId,
       getTokenParams,
       isTokenMethodAction,
     } = this.props;
 
+    getGasFeeEstimatesAndStartPolling().then((pollingToken) => {
```
danjm, line 74: The same question I raised elsewhere about cancelling polling can be raised here.

ui/hooks/useRetryTransaction.js
brad-decker: > But also, why is useGasFeeEstimates called here and not in a component?

I'm having a hard time remembering why I did this, though it is fine to call a hook from within a hook. Thats part of their power in composability. I do think here it'd be more appropriate to just poll -- however, like I said I'm having hard time remembering why this was necessary.

I did remove the setCustomGasPriceForRetry method that calls the fetch code, but I think this is covered by having a polling creator in the gas modal so this likely can just be removed.

ui/ducks/gas/gas.duck.js
brad-decker: yeah, that's correct.

ui/ducks/send/send.js
```diff
@@ -186,8 +183,16 @@ async function estimateGasLimitForSend({
...
   // blockGasLimit may be a falsy, but defined, value when we receive it from
-  // state, so we use logical or to fall back to MIN_GAS_LIMIT_HEX.
-  const blockGasLimit = options.blockGasLimit || MIN_GAS_LIMIT_HEX;
+  // state, so we use logical or to fall back to MIN_GAS_LIMIT_HEX. Some
+  // network implementations check the gas parameter supplied to
+  // eth_estimateGas for validity. For this reason, we set token sends
+  // blockGasLimit default to a higher number. Note that the current gasLimit
+  // on a BLOCK is 15,000,000 and will be 30,000,000 on mainnet after London.
+  // Meanwhile, MIN_GAS_LIMIT_HEX is 0x5208.
+  const blockGasLimit =
```
danjm, line 192: Before this would be `options.blockGasLimit` if it was defined. Now it will be `GAS_LIMITS.BASE_TOKEN_ESTIMATE` if `options.blockGasLimit` is defined. To ensure backwards compatibility while addressing the issue this is meant to address, I might suggest:
```
let blockGasLimit = MIN_GAS_LIMIT_HEX;
if (options.blockGasLimit) {
blockGasLimit = options.blockGasLimit;
} else if (sendToken) {
blockGasLimit = GAS_LIMITS.BASE_TOKEN_ESTIMATE;
}
```

ui/ducks/send/send.js
```diff
@@ -342,7 +347,7 @@ export const computeEstimatedGasLimit = createAsyncThunk(
     if (send.stage !== SEND_STAGES.EDIT) {
       const gasLimit = await estimateGasLimitForSend({
         gasPrice: send.gas.gasPrice,
-        blockGasLimit: metamask.blockGasLimit,
```
danjm, line 345: We should investigate the difference between `blockGasLimit` and `currentBlockGasLimit` and why we were using something that was seemingly not defined. We don't necessarily have to do this before this PR is merged, but I would like to create an issue, and block release on answering the question, to make sure there are no other related bugs, or no important differences between what we expected `blockGasLimit` to be and what `currentBlockGasLimit` is

ui/ducks/send/send.js
```diff
@@ -360,6 +365,29 @@ export const computeEstimatedGasLimit = createAsyncThunk(
...
+ * This method is used to keep the original logic from the gas.duck.js file
+ * after receiving a gasPrice from eth_gasPrice. First, the returned gasPrice
+ * was converted to GWEI, then it was converted to a Number, then in the send
+ * duck (here) we would use getGasPriceInHexWei to get back to hexWei. Now that
+ * we receive a GWEI estimate from the controller, we still need to do this
+ * weird conversion to get the proper rounding.
+ * @param {T} gasPriceEstimate
+ * @returns
+ */
+function getRoundedGasPrice(gasPriceEstimate) {
```
danjm, line 378: Nit: feels like this should be be in some `util` type file

danjm (changes requested):
Completed review. Overall everything makes sense and looks good. Once my comments and questions are addressed we are good.

ui/ducks/send/send.js
brad-decker: I disagree, as it will only ever be used here I hope, and maybe will eventually be removed entirely. Moving things to Util files is a vote of confidence for their usage and it is not exported on purpose.

ui/ducks/send/send.js
brad-decker: It most likely happened as a result of the send flow refactor where we were taking currentBlockGasLimit from state and passing it to utility methods as a 'blockGasLimit' prop and then I improperly selected it from state.

brad-decker: currentBlockGasLimit is the legitimate value. A prop named blockGasLimit is used in the tx-gas-utils and transaction controller, but its set in state in the `account-tracker` as `currentBlockGasLimit`

```
  async _updateForBlock(blockNumber) {
    this._currentBlockNumber = blockNumber;

    // block gasLimit polling shouldn't be in account-tracker shouldn't be here...
    const currentBlock = await this._query.getBlockByNumber(blockNumber, false);
    if (!currentBlock) {
      return;
    }
    const currentBlockGasLimit = currentBlock.gasLimit;
    this.store.updateState({ currentBlockGasLimit });

    try {
      await this._updateAccounts();
    } catch (err) {
      log.error(err);
    }
  }
```

ui/ducks/send/send.js
brad-decker: if options.blockGasLimit is defined it'll be set to that due to the logical `||` operator. the asset type is only checking if the blockGasLimit is not provided

brad-decker: I can make this change for legibility though.

brad-decker: Done in latest

ui/pages/confirm-transaction/confirm-transaction.component.js
brad-decker: I added a instance variable _isMounted where this occurs, and if _isMounted if alse after the promise resolves we disconnect instead of setting the pollingToken in state.

ui/hooks/useRetryTransaction.js
brad-decker: removed in latest

brad-decker:
I think I resolved all of the feedback @danjm

danjm (approved):
LGTM

PeterYinusa:
LGTM!

darkwing (approved):

PR #11515, opened 2021-07-13
Adding TokenListController to Metamask-Controller

NiranjanaBinoy:
This PR includes the changes needed to add TokenListcontroller to metamask-controller.js. This will add the list of Tokens from the Tokenlistcontrollers to state under the key `tokenList`. The remaining changes to use the TokenList in DetectTokensController and at other usages of `contract-metadata` will be a part of #11398

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -228,7 +228,17 @@ export default class MetamaskController extends EventEmitter {
       chainId: hexToDecimal(this.networkController.getCurrentChainId()),
       useStaticTokenList: this.preferencesController.store.getState()
         .useStaticTokenList,
-      onNetworkStateChange: this._onModifiedNetworkStateChange.bind(this),
+      onNetworkStateChange: (cb) =>
+        this.networkController.store.subscribe(async (networkState) => {
+          const modifiedNetworkState = {
+            ...networkState,
+            provider: {
+              ...networkState.provider,
+              chainId: hexToDecimal(networkState.provider.chainId),
+            },
+          };
+          return await cb(modifiedNetworkState);
+        }),
       onPreferencesStateChange: this.preferencesController.store.subscribe.bind(
         this.preferencesController.store,
       ),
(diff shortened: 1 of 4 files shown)


app/scripts/metamask-controller.js
```diff
@@ -219,6 +221,21 @@ export default class MetamaskController extends EventEmitter {
...
     });
 
+    const tokenListMessenger = controllerMessenger.getRestricted({
+      name: 'TokenListController',
+    });
+    this.tokenListController = new TokenListController({
+      chainId: hexToDecimal(this.networkController.getCurrentChainId()),
+      useStaticTokenList: this.preferencesController.store.getState()
+        .useStaticTokenList,
+      onNetworkStateChange: this._onModifiedNetworkStateChange.bind(this),
```
brad-decker, line 231: Could we just inline this method rather than creating a new method on the controller? I think it would be easier to understand its one-off usage if it was just all here. Also, unless the TokenListController needs the entire state, I'd suggest just passing the fields required back here.

app/scripts/metamask-controller.js
```diff
@@ -723,6 +762,7 @@ export default class MetamaskController extends EventEmitter {
       setUseBlockie: this.setUseBlockie.bind(this),
       setUseNonceField: this.setUseNonceField.bind(this),
       setUsePhishDetect: this.setUsePhishDetect.bind(this),
+      setUseStaticTokenList: this.setUseStaticTokenList.bind(this),
```
brad-decker, line 765: I *think* you can get the same effect by doing

```suggestion
      setUseStaticTokenList: nodeify(
        this.preferencesController.setUseStaticTokenList,
        this.preferencesController
      ),
```

brad-decker: If this suggestion is taken and works then the other method added to the metamask-controller can be removed.

brad-decker (changes requested):
Some very minor things but looks good!

app/scripts/metamask-controller.js
```diff
@@ -219,6 +221,31 @@ export default class MetamaskController extends EventEmitter {
...
+      onNetworkStateChange: (cb) =>
+        this.networkController.store.subscribe(async (networkState) => {
+          const modifiedNetworkState = {
+            ...networkState,
+            provider: {
+              ...networkState.provider,
+              chainId: hexToDecimal(networkState.provider.chainId),
+            },
+          };
+          return await cb(modifiedNetworkState);
```
brad-decker, line 240: [REQ] if cb is a promise returning function then I think the async/await here isn't necessary right? If you return `cb(modifiedNetworkState)` it'll return the promise from the `cb` method which can be awaited in the controller. I could be missing something though!

app/scripts/metamask-controller.js
```diff
@@ -2797,6 +2832,23 @@ export default class MetamaskController extends EventEmitter {
     }
   }
 
+  /**
+   * Sets whether or not to use phishing detection.
+   * @param {boolean} val
+   * @param {Function} cb
+   */
+  setUseStaticTokenList(val, cb) {
```
brad-decker, line 2840: This method can be removed now if the nodeify entry above worked out :)

brad-decker (changes requested):

brad-decker (approved):

danjm (approved):

PR #11518, opened 2021-07-13
CI - enforce dep usage with depcheck

kumavis:
caught some unused and missing deps


package.json
```diff
@@ -216,6 +214,7 @@
     "@babel/preset-react": "^7.0.0",
     "@babel/register": "^7.5.5",
     "@lavamoat/allow-scripts": "^1.0.6",
+    "@lavamoat/preinstall-always-fail": "^1.0.0",
```
Gudahtt, line 217: Could this be omitted entirely? Since it's a dependency of `@lavamoat/allow-scripts` already. It still serves the same purpose when it's a transitive dependency.

.depcheckrc.yml
```diff
@@ -0,0 +1,44 @@
+# things that *are* used, that depcheck is wrong about
+ignores:
+  # 
+  # webapp deps
+  # 
+
+  - "@babel/runtime"
+  - "@fortawesome/fontawesome-free"
+  - "punycode"
```
Gudahtt, line 9: Curious why depcheck complains about this one :thinking: It seems to be a "regular" dependency as far as I can tell.

.circleci/config.yml
```diff
@@ -25,7 +25,10 @@ workflows:
               only:
                 - /^Version-v(\d+)[.](\d+)[.](\d+)/
       - prep-deps
-      - test-deps
+      - test-deps-audit
+      - test-deps-depcheck
```
Gudahtt, line 29: This is the syntax error I believe:
```suggestion
      - test-deps-depcheck:
```

.circleci/config.yml
```diff
@@ -322,6 +325,19 @@ jobs:
       - run:
           name: yarn audit
           command: .circleci/scripts/yarn-audit.sh
+      - run:
+          name: depcheck
+          command: yarn depcheck
```
Gudahtt, line 330: Since the separate job was added:
```suggestion
```

.depcheckrc.yml
kumavis: yeah something about esm + importing an alternate entry point?

package.json
ryanml: Removed

.depcheckrc.yml
```diff
@@ -0,0 +1,46 @@
...
+  - "sass-loader"
+  - "resolve-url-loader"
+
+# files depcheck should not parse 
+ignorePatterns:
+  # seems to incorrectly parse scss @include pragmas?
+  - "**/*.scss"
+  # self-contained bundle used for testing
+  - "**/send-eth-with-private-key-test/web3js.js"
+  - "**/send-eth-with-private-key-test/ethereumjs-tx.js"
```
Gudahtt, line 46: Nit: Missing trailing empty line

Gudahtt (dismissed):
LGTM!

.depcheckrc.yml
kumavis, line 46:
```suggestion
  - "**/send-eth-with-private-key-test/ethereumjs-tx.js"
  
```

Gudahtt (approved):

ryanml (approved):

Issue #11474, opened 2021-07-08
EIP-1159 UI - Implement and Wire Up Insufficent Funds Error


darkwing:
https://www.figma.com/file/NXvUKcZ0Ocu1GPcT7VxRMx/1559---Gas-Controls?node-id=1618%3A23677

<img width="924" alt="InsufficentFunds" src="https://user-images.githubusercontent.com/46655/124975485-b0220a80-dff3-11eb-914f-6469bdb88047.png">

This will need to happen in the `EditGasDisplay` component, where an `error` prop exists.

darkwing:
I believe the best place to do this will be in `ui/hooks/useGasFeeInputs.js`; since all gas estimates are routed back to this function and its state, this is the best spot to (1) ensure the user has a sufficient balance and (2) to have central place where we know that they have an insufficient balance.  Since `EditGasPopover` already uses this hook, we simply need to pass that info down to the `EditGasDisplay`.

PR #11522, opened 2021-07-14
Move custom gas fee properties from useGasFeeInputs state to redux

danjm:
Necessary for https://github.com/MetaMask/metamask-extension/issues/11474

Necessary because the `edit-gas-popover` and `edit-gas-display` both need to be able to access the custom values set in the advanced form. This will allow us to control the button, and also properly display the error message, based on the state of those values.


danjm:
This appears to be redundant given https://github.com/MetaMask/metamask-extension/pull/11510, will reopen later if necessary

danjm:
or maybe this is needed...

PR #11526, opened 2021-07-14
Create different label for Gas Price with GWEI included

darkwing:
The current gas price text is `Gas price (GWEI)` but we need just `Gas price`


ryanml (approved):
++

danjm (approved):

brad-decker (approved):

PR #11536, opened 2021-07-15
Updating formatTxMetaForRpcResult to match latest spec

### LINKED ISSUE #11492: Update `formatTxMetaForRpcResult`
When dapps ask for transactions from the provider we format those transactions so that they match the spec, but the spec has evolved and we need to allow for the presence of EIP-1559, EIP-2930 and EIP-2718 (fee market/fee burning, access lists, and typed transaction envelopes): 

https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/controllers/network/util.js#L4-L24

This must be updated to match [the spec](https://github.com/ethereum/eth1.0-specs/blob/master/json-rpc/spec.json#L1624-L1746)

Namely, we need to pass through gasPrice, maxFeePerGas and maxPriorityFeePerGas and ensure only ONE method of gas estimation is returned (if gasPrice is present, don't return the other two fields, and vice versa). We also need to return the type parameter '0x2' in the case of EIP-1559.

ryanml:
Fixes: https://github.com/MetaMask/metamask-extension/issues/11492

### DIFF
--- a/app/scripts/controllers/network/util.js
+++ b/app/scripts/controllers/network/util.js
@@ -1,24 +1,48 @@
 import { NETWORK_TO_NAME_MAP } from '../../../../shared/constants/network';
+import { TRANSACTION_ENVELOPE_TYPES } from '../../../../shared/constants/transaction';
 
 export const getNetworkDisplayName = (key) => NETWORK_TO_NAME_MAP[key];
 
 export function formatTxMetaForRpcResult(txMeta) {
-  return {
-    blockHash: txMeta.txReceipt ? txMeta.txReceipt.blockHash : null,
-    blockNumber: txMeta.txReceipt ? txMeta.txReceipt.blockNumber : null,
-    from: txMeta.txParams.from,
-    gas: txMeta.txParams.gas,
-    gasPrice: txMeta.txParams.gasPrice,
-    hash: txMeta.hash,
-    input: txMeta.txParams.data || '0x',
-    nonce: txMeta.txParams.nonce,
-    to: txMeta.txParams.to,
-    transactionIndex: txMeta.txReceipt
-      ? txMeta.txReceipt.transactionIndex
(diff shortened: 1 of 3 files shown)


app/scripts/controllers/network/util.js
```diff
@@ -1,24 +1,47 @@
...
+    r,
+    s,
+    to,
+    gas,
+    from,
+    hash,
+    nonce,
+    input: data || '0x',
+    value: value || '0x0',
+    accessList: txReceipt?.accessList || {},
```
ryanml, line 31: Per the spec (https://github.com/ethereum/eth1.0-specs/blob/master/json-rpc/spec.json#L1619), `accessList` would be an object, but not sure if it's fine to fallback to `null` if it isn't defined. Also, I'm unsure where this would exist with in `txMeta`, I just have it as a member of `txReceipt` for now

app/scripts/controllers/network/util.js
```diff
@@ -1,24 +1,47 @@
...
+    s,
+    to,
+    gas,
+    from,
+    hash,
+    nonce,
+    input: data || '0x',
+    value: value || '0x0',
+    blockHash: txReceipt?.blockHash || null,
+    accessList: txReceipt?.accessList || null,
```
ryanml, line 32: This is tentative, as I'm not certain where in `txMeta` `accessList` lives

brad-decker: txParams

ryanml: Thanks! fixed

brad-decker (dismissed):

ryanml:
@brad-decker @darkwing this needs a re-review, did a push, thanks!

danjm (approved):
Looks correct to me

brad-decker (approved):

PR #11537, opened 2021-07-15
Handling non-numeric input for chain id in network form

ryanml:
Catches error that throws when non-numeric input is supplied to `decimalToHex`, and renders validation error to user

New Error Behavior:
Screen Shot 2021-07-15 at 3 12 06 AM

### DIFF
--- a/ui/pages/settings/networks-tab/network-form/network-form.component.js
+++ b/ui/pages/settings/networks-tab/network-form/network-form.component.js
@@ -358,14 +358,27 @@ export default class NetworkForm extends PureComponent {
   };
 
   validateChainIdOnChange = (selfRpcUrl, chainIdArg = '') => {
+    const { t } = this.context;
     const { networksToRender } = this.props;
     const chainId = chainIdArg.trim();
+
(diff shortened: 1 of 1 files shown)


ui/pages/settings/networks-tab/network-form/network-form.component.js
```diff
@@ -358,14 +358,27 @@ export default class NetworkForm extends PureComponent {
...
-      ? chainId
-      : `0x${decimalToHex(chainId)}`;
+    let hexChainId = chainId;
+
+    if (!hexChainId.startsWith('0x')) {
+      try {
+        hexChainId = `0x${decimalToHex(hexChainId)}`;
+      } catch (err) {
+        this.setErrorTo('chainId', {
+          key: 'invalidHexNumber',
```
ryanml, line 375: The message `"Invalid Input"` may serve better for this case, but I wanted to avoid adding a new translation here, and I think serves the purpose well.

danjm (approved):
LGTM

brad-decker (approved):

PR #11540, opened 2021-07-15
EIP-1559 -  Implement form validation for EIP-1559

### LINKED ISSUE #11306: EIP-1559 UI - Wire Up Advanced Gas Error UI
https://www.figma.com/file/NXvUKcZ0Ocu1GPcT7VxRMx/1559---Gas-Controls?node-id=643%3A5

We need to wire real gas data into the `EditGasDisplay` component, as well as show the correct errors when data fields contain invalid information, or there are network or balance errors.

ErrorMessages

UnknownProcessing

darkwing:
Fixes: https://github.com/MetaMask/metamask-extension/issues/11306

Implements error handling and validation for the advanced gas form.

### DIFF
--- a/ui/helpers/constants/gas.js
+++ b/ui/helpers/constants/gas.js
@@ -1,17 +1,20 @@
 export const GAS_FORM_ERRORS = {
-  GAS_LIMIT_TOO_LOW: 'gasLimitTooLow',
+  GAS_LIMIT_OUT_OF_BOUNDS: 'editGasLimitOutOfBounds',
   MAX_PRIORITY_FEE_TOO_LOW: 'editGasMaxPriorityFeeLow',
   MAX_FEE_TOO_LOW: 'editGasMaxFeeLow',
+  MAX_PRIORITY_FEE_ZERO: 'editGasMaxPriorityFeeZeroError',
 };
 
 export function getGasFormErrorText(type, t) {
   switch (type) {
-    case GAS_FORM_ERRORS.GAS_LIMIT_TOO_LOW:
-      return t('gasLimitTooLow');
+    case GAS_FORM_ERRORS.GAS_LIMIT_OUT_OF_BOUNDS:
+      return t('editGasLimitOutOfBounds');
     case GAS_FORM_ERRORS.MAX_PRIORITY_FEE_TOO_LOW:
       return t('editGasMaxPriorityFeeLow');
     case GAS_FORM_ERRORS.MAX_FEE_TOO_LOW:
       return t('editGasMaxFeeLow');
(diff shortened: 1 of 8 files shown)


darkwing:
We'll need to make another pass for implementing all validation -- I'd moreso like validation of the structure here.

brad-decker:
The one flaw with this is that we want to actively move away from using variables to contain translation strings because it makes the verify locales script less strict and it makes tracking down where and how certain translation strings are used difficult.

Instead we could use error constants that we switch on that renders the appropriate translation string. This could even be done in a helper method to keep the code cleaner.

So same concept but just one extra step to ensure we are following best practices

danjm (changes requested):
Code looks good, but I agree with Brad's suggestion

darkwing:
Moved to using a constant

brad-decker:
Still need to move to doing something more like this 
```
switch (error.key) {
  case OUR_CONSTANT_ONE:
    return t('literalKeyValue')// <-- we can't use variables to hold these keys so we do something like this
```

real world example: 
https://github.com/MetaMask/metamask-extension/blob/develop/ui/helpers/utils/transactions.util.js#L201-L231
(consumed in useTransactionDisplayData)

darkwing:
I think the only validation left is the "too high" stuff:

ErrorMessages

brad-decker (dismissed):

brad-decker (dismissed):

danjm (approved):
LGTM

ui/helpers/constants/gas.js
```diff
@@ -0,0 +1,27 @@
+export const GAS_FORM_ERRORS = {
+  GAS_LIMIT_OUT_OF_BOUNDS: 'editGasLimitOutOfBounds',
+  MAX_PRIORITY_FEE_TOO_LOW: 'editGasMaxPriorityFeeLow',
+  MAX_FEE_TOO_LOW: 'editGasMaxFeeLow',
+  MAX_PRIORITY_FEE_ZERO: 'editGasMaxPriorityFeeZeroError',
+  MAX_PRIORITY_FEE_HIGH_WARNING: 'editGasMaxPriorityFeeHigh',
+  MAX_FEE_HIGH_WARNING: 'editGasMaxFeeHigh',
+};
+
+export function getGasFormErrorText(type, t) {
```
dan437, line 10: nit: Do we want to have functions in files with constants? I would probably move the function to `shared/modules/gas.utils.js`

darkwing: We use this pattern elsewhere; on another PR, Brad mentioned that putting things in shared bloats bundle sizes.  These constants aren't actually used in the background, thus this is fine.

dan437 (approved):
Looks great!

PR #11541, opened 2021-07-16
EIP-1559 - Don't show submit button  when in education mode, don't duplicate title

darkwing:
I noticed that the title of the education content displayed twice, both as the popover title and the heading of the content, so I removed that problem.  Additionally, we should hide the submission button when the education content is displaying.


danjm (approved):
LGTM

dan437 (approved):

PR #11543, opened 2021-07-16
Improve handling for block explorer url host extraction (AccountOptionsMenu)

ryanml:
(Sentry error fix) In the event a user has an invalid block explorer url for the current network, (follow up: investigate how this may be given we validate urls in the network form) the existing regex method for host extraction will fail. At any rate, this lets `URL` do the heavy lifting.


danjm (approved):

danjm:
Addresses this error: https://sentry.io/organizations/metamask/issues/2071695817/?environment=production&project=273505&query=is%3Aunresolved+release%3A9.8.0&sort=freq&statsPeriod=14d

NiranjanaBinoy (approved):
LGTM

brad-decker (approved):

PR #11545, opened 2021-07-16
Add support for Polygon

dan437:
## Explanation
Support for Polygon in Swaps

## Manual testing steps
- Add Polygon Mainnet in the extension (https://docs.matic.network/docs/develop/network-details/network/)
- Click on `Swap` on the main page (make sure that you have enough funds to do swapping)
- Select `Swap from` and `Swap to` values and do swapping
- See the success page


shared/constants/swaps.js
```diff
@@ -64,13 +78,16 @@ const SWAPS_TESTNET_HOST = 'https://metaswap-api.airswap-dev.codefi.network';
 
 const BSC_DEFAULT_BLOCK_EXPLORER_URL = 'https://bscscan.com/';
 const MAINNET_DEFAULT_BLOCK_EXPLORER_URL = 'https://etherscan.io/';
+const POLYGON_DEFAULT_BLOCK_EXPLORER_URL = 'https://api.polygonscan.com/api/';
```
matmertz25, line 81: We probably want to change this value to `https://polygonscan.com` rather than the API endpoint

app/scripts/controllers/network/network.js
```diff
@@ -166,7 +166,8 @@ export default class NetworkController extends EventEmitter {
       return EIPS[1559];
     }
     const latestBlock = await this.getLatestBlock();
-    const supportsEIP1559 = latestBlock.baseFeePerGas !== undefined;
+    const supportsEIP1559 =
```
danjm, line 169: Was this change because of an error that was occurin when on polygon?

dan437: That is correct.

danjm (approved):
LGTM

darkwing (approved):

PR #11547, opened 2021-07-16
Fix actionable message icon size

darkwing:
A few PRs collided and the actionable message icon selector no longer matches and was showing at a huge size -- this fixes the selector.


brad-decker (approved):

ryanml (approved):
++

PR #11549, opened 2021-07-16
update send slice to work with EIP 1559 networks

brad-decker:
Opening for QA

### What does this PR do? 
1. updates the send slice state to work with either gasPrice or maxFeePerGas/maxPriorityFeePerGas, including generating a fully formed type 2 (0x2) EIP 1559 transaction for submission.
2. Hooks up the minimum gas paid estimation on confirmations to work with EIP 1559 fields.

### Steps to QA:
With SHOW_EIP_1559_UI set to `0` or `false`:
1. Try to send transactions on the major test nets and make sure that the confirmation page quotes align with expectations.
2. Do this on mainnet, but on testnets also submit transactions and confirm the fields in the block explorer.
3. Modify gas prices using the edit modal and make sure the changes are reflected in the confirmation screen AND on the block explorer on test nets.

With SHOW_EIP_1559_UI set to `1`:
1. Same as above, but note you will not see the gas fields on send screen so it becomes more difficult to validate. You can check redux but I have had some trouble with it since the build system was updated.
2. You'll want to test sending transactions on the various testnets and mainnet. Note, mainnet will be a gasPrice network. You can also run ganache locally and connect to local host for a test network to test sending gasPrice transactions on.
3. In confirmation screens the gas estimate will be `estimatedBaseFee + maxPriorityFeePerGas` or maxFeePerGas, whichever is lowest.

### DIFF
--- a/app/scripts/controllers/transactions/index.js
+++ b/app/scripts/controllers/transactions/index.js
@@ -27,7 +27,11 @@ import {
   TRANSACTION_ENVELOPE_TYPES,
 } from '../../../../shared/constants/transaction';
 import { METAMASK_CONTROLLER_EVENTS } from '../../metamask-controller';
-import { GAS_LIMITS } from '../../../../shared/constants/gas';
+import {
+  GAS_LIMITS,
+  GAS_ESTIMATE_TYPES,
+} from '../../../../shared/constants/gas';
+import { decGWEIToHexWEI } from '../../../../shared/modules/conversion.utils';
 import {
   HARDFORKS,
   MAINNET,
@@ -107,6 +111,7 @@ export default class TransactionController extends EventEmitter {
     this.inProcessOfSigning = new Set();
     this._trackMetaMetricsEvent = opts.trackMetaMetricsEvent;
     this._getParticipateInMetrics = opts.getParticipateInMetrics;
+    this._getEIP1559GasFeeEstimates = opts.getEIP1559GasFeeEstimates;
 
     this.memStore = new ObservableStore({});
     this.query = new EthQuery(this.provider);
@@ -399,7 +404,13 @@ export default class TransactionController extends EventEmitter {
    * @returns {Promise<object>} resolves with txMeta
    */
   async addTxGasDefaults(txMeta, getCodeResponse) {
-    const defaultGasPrice = await this._getDefaultGasPrice(txMeta);
+    const eip1559Compatibility = await this.getEIP1559Compatibility();
+
+    const {
+      gasPrice: defaultGasPrice,
+      maxFeePerGas: defaultMaxFeePerGas,
+      maxPriorityFeePerGas: defaultMaxPriorityFeePerGas,
+    } = await this._getDefaultGasFees(txMeta, eip1559Compatibility);
     const {
       gasLimit: defaultGasLimit,
       simulationFails,
@@ -410,6 +421,67 @@ export default class TransactionController extends EventEmitter {
     if (simulationFails) {
       txMeta.simulationFails = simulationFails;
     }
+
+    if (eip1559Compatibility) {
+      // If the dapp has suggested a gas price, but no maxFeePerGas or maxPriorityFeePerGas
+      //  then we set maxFeePerGas and maxPriorityFeePerGas to the suggested gasPrice.
+      if (
+        txMeta.txParams.gasPrice &&
+        !txMeta.txParams.maxFeePerGas &&
+        !txMeta.txParams.maxPriorityFeePerGas
+      ) {
+        txMeta.txParams.maxFeePerGas = txMeta.txParams.gasPrice;
+        txMeta.txParams.maxPriorityFeePerGas = txMeta.txParams.gasPrice;
+      } else {
+        if (defaultMaxFeePerGas && !txMeta.txParams.maxFeePerGas) {
+          // If the dapp has not set the gasPrice or the maxFeePerGas, then we set maxFeePerGas
+          // with the one returned by the gasFeeController, if that is available.
+          txMeta.txParams.maxFeePerGas = defaultMaxFeePerGas;
+        }
+
+        if (
+          defaultMaxPriorityFeePerGas &&
+          !txMeta.txParams.maxPriorityFeePerGas
+        ) {
+          // If the dapp has not set the gasPrice or the maxPriorityFeePerGas, then we set maxPriorityFeePerGas
+          // with the one returned by the gasFeeController, if that is available.
+          txMeta.txParams.maxPriorityFeePerGas = defaultMaxPriorityFeePerGas;
+        }
+
+        if (defaultGasPrice && !txMeta.txParams.maxFeePerGas) {
+          // If the dapp has not set the gasPrice or the maxFeePerGas, and no maxFeePerGas is available
+          // from the gasFeeController, then we set maxFeePerGas to the defaultGasPrice, assuming it is
+          // available.
+          txMeta.txParams.maxFeePerGas = defaultGasPrice;
+        }
+
+        if (
+          txMeta.txParams.maxFeePerGas &&
+          !txMeta.txParams.maxPriorityFeePerGas
+        ) {
+          // If the dapp has not set the gasPrice or the maxPriorityFeePerGas, and no maxPriorityFeePerGas is
+          // available from the gasFeeController, then we set maxPriorityFeePerGas to
+          // txMeta.txParams.maxFeePerGas, which will either be the gasPrice from the controller, the maxFeePerGas
+          // set by the dapp, or the maxFeePerGas from the controller.
+          txMeta.txParams.maxPriorityFeePerGas = txMeta.txParams.maxFeePerGas;
+        }
+      }
+
+      // We remove the gasPrice param entirely when on an eip1559 compatible network
+
+      delete txMeta.txParams.gasPrice;
+    } else {
+      // We ensure that maxFeePerGas and maxPriorityFeePerGas are not in the transaction params
+      // when not on a EIP1559 compatible network
+
+      delete txMeta.txParams.maxPriorityFeePerGas;
+      delete txMeta.txParams.maxFeePerGas;
+    }
+
+    // If we have gotten to this point, and none of gasPrice, maxPriorityFeePerGas or maxFeePerGas are
+    // set on txParams, it means that either we are on a non-EIP1559 network and the dapp didn't suggest
+    // a gas price, or we are on an EIP1559 network, and none of gasPrice, maxPriorityFeePerGas or maxFeePerGas
+    // were available from either the dapp or the network.
     if (
       defaultGasPrice &&
       !txMeta.txParams.gasPrice &&
@@ -418,6 +490,7 @@ export default class TransactionController extends EventEmitter {
     ) {
       txMeta.txParams.gasPrice = defaultGasPrice;
     }
+
     if (defaultGasLimit && !txMeta.txParams.gas) {
       txMeta.txParams.gas = defaultGasLimit;
     }
@@ -425,20 +498,59 @@ export default class TransactionController extends EventEmitter {
   }
 
   /**
-   * Gets default gas price, or returns `undefined` if gas price is already set
+   * Gets default gas fees, or returns `undefined` if gas fees are already set
    * @param {Object} txMeta - The txMeta object
    * @returns {Promise<string|undefined>} The default gas price
    */
-  async _getDefaultGasPrice(txMeta) {
+  async _getDefaultGasFees(txMeta, eip1559Compatibility) {
     if (
-      txMeta.txParams.gasPrice ||
+      (!eip1559Compatibility && txMeta.txParams.gasPrice) ||
       (txMeta.txParams.maxFeePerGas && txMeta.txParams.maxPriorityFeePerGas)
     ) {
-      return undefined;
+      return {};
     }
+
+    try {
+      const {
+        gasFeeEstimates,
+        gasEstimateType,
+      } = await this._getEIP1559GasFeeEstimates();
+      if (
+        eip1559Compatibility &&
+        gasEstimateType === GAS_ESTIMATE_TYPES.FEE_MARKET
+      ) {
+        const {
+          medium: { suggestedMaxPriorityFeePerGas, suggestedMaxFeePerGas } = {},
+        } = gasFeeEstimates;
+
+        if (suggestedMaxPriorityFeePerGas && suggestedMaxFeePerGas) {
+          return {
+            maxFeePerGas: decGWEIToHexWEI(suggestedMaxFeePerGas),
+            maxPriorityFeePerGas: decGWEIToHexWEI(
+              suggestedMaxPriorityFeePerGas,
+            ),
+          };
+        }
+      } else if (gasEstimateType === GAS_ESTIMATE_TYPES.LEGACY) {
+        // The LEGACY type includes low, medium and high estimates of
+        // gas price values.
+        return {
+          gasPrice: decGWEIToHexWEI(gasFeeEstimates.medium),
+        };
+      } else if (gasEstimateType === GAS_ESTIMATE_TYPES.ETH_GASPRICE) {
+        // The ETH_GASPRICE type just includes a single gas price property,
+        // which we can assume was retrieved from eth_gasPrice
+        return {
+          gasPrice: decGWEIToHexWEI(gasFeeEstimates.gasPrice),
+        };
+      }
+    } catch (e) {
+      console.error(e);
+    }
+
     const gasPrice = await this.query.gasPrice();
 
-    return addHexPrefix(gasPrice.toString(16));
+    return { gasPrice: gasPrice && addHexPrefix(gasPrice.toString(16)) };
   }
 
   /**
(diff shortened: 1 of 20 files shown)


app/scripts/controllers/transactions/index.js
```diff
@@ -309,6 +309,7 @@ export default class TransactionController extends EventEmitter {
    * @returns {txMeta}
    */
   async addUnapprovedTransaction(txParams, origin) {
+    console.log(txParams, origin);
```
brad-decker, line 312: console.log.

app/scripts/controllers/transactions/index.js
```diff
@@ -671,6 +672,7 @@ export default class TransactionController extends EventEmitter {
       this.txStateManager.setTxStatusApproved(txId);
       // get next nonce
       const txMeta = this.txStateManager.getTransaction(txId);
+      console.log(txMeta);
```
brad-decker, line 675: console.log.

ui/hooks/useSafeGasEstimatePolling.js
```diff
@@ -0,0 +1,25 @@
+import { useEffect } from 'react';
+import {
+  disconnectGasFeeEstimatePoller,
+  getGasFeeEstimatesAndStartPolling,
+} from '../store/actions';
+
+export function useSafeGasEstimatePolling() {
```
brad-decker, line 7: add some documentation

ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
```diff
@@ -744,9 +755,21 @@ export default class ConfirmTransactionBase extends Component {
     if (toAddress) {
       tryReverseResolveAddress(toAddress);
     }
+
+    getGasFeeEstimatesAndStartPolling().then((pollingToken) => {
```
brad-decker, line 759: add some docs.

ui/selectors/confirm-transaction.js
```diff
@@ -218,17 +228,46 @@ export const transactionFeeSelector = function (state, txData) {
...
-  // if the gas price from our infura endpoint is null or undefined
-  // use the metaswap average price estimation as a fallback
-  let { txParams: { gasPrice } = {} } = txData;
+  const gasEstimationObject = {
+    gasLimit: txData.txParams?.gas ?? '0x0',
+  };
 
-  if (!gasPrice) {
-    gasPrice = getAveragePriceEstimateInHexWEI(state) || '0x0';
+  switch (gasEstimateType) {
```
brad-decker, line 238: this block is the best stab that shouldn't break anything for current txs. we need to think about the following variables:
TransactionType = 'legacy' | 'fee market'
gasEstiamteType = 'legacy' | 'eth_gasPrice' | 'none' | 'fee market'
doesGasPriceExistInTx = true | false
doesEIPGasFieldsExistOntx = true | false

Now create a table with what to do in each combinations of scenarios :P 

If the transaction type is FEE MARKET but our gas estimate type is ETH_gasPrice (because of a network failure or a dapp submits a type 0x2 tx on a non-supported network (we should fail at the provider level here but that's a todo item)). what do we do? 

when our gas estimates are FEE_MARKET and the dapp suggests gasPrice we just use the gasPrice and fall back to '0x0' for now.

😓

ui/selectors/confirm-transaction.js
```diff
@@ -244,36 +283,52 @@ export const transactionFeeSelector = function (state, txData) {
...
+  const hexMinimumTransactionFee = getMinimumGasTotalInHexWei(
+    gasEstimationObject,
+  );
+  const hexMaximumTransactionFee = getMaximumGasTotalInHexWei(
+    gasEstimationObject,
+  );
 
-  const fiatTransactionFee = getTransactionFee({
-    value: hexTransactionFee,
+  const fiatMinimumTransactionFee = getTransactionFee({
```
brad-decker, line 293: the old value, fiatTransactionFee appears to be unused everywhere. Probably for the best to rely upon our components for that. Will remove in a later PR and I wouldn't tap into these.

brad-decker:
@PeterYinusa got this one queued up for next week.

brad-decker:
@dan437 I added the send slice to the coverage path and ensured it met the threshold. Hopefully, that doesn't mess with your objectives of covering swaps code.

ui/ducks/send/send.js
```diff
@@ -570,6 +584,7 @@ export const initialState = {
       value: '0x0',
       gas: '0x0',
       gasPrice: '0x0',
+      type: TRANSACTION_ENVELOPE_TYPES.LEGACY,
```
danjm, line 587: Note to self: Although the intiialState has the params for a legacy transaction, the `updateDraftTransaction` reducer appropriately handles fee market type transactions

ui/ducks/send/send.js
```diff
@@ -683,9 +698,17 @@ const slice = createSlice({
      * field and send state, then updates the draft transaction.
      */
     calculateGasTotal: (state) => {
-      state.gas.gasTotal = addHexPrefix(
-        calcGasTotal(state.gas.gasLimit, state.gas.gasPrice),
-      );
+      // use gasPrice as the multiplier if working with a LEGACY transaction
+      // otherwise use maxFeePerGas.
```
danjm, line 702: Nit: The below logic is slightly out of sync with the comment here. I think the below logic would be more accurately reflected by the following comment:

```
// use maxFeePerGas as the multiplier if working with a FEE_MARKET transaction
// otherwise us gasPrice
```
(i.e. first describe what is being done in the specific case checked for the in the `if` block, and the describe everything the `else` block handles with `otherwise`)

jest.config.js
```diff
@@ -1,7 +1,10 @@
 module.exports = {
   restoreMocks: true,
   coverageDirectory: 'jest-coverage/',
-  collectCoverageFrom: ['<rootDir>/ui/**/swaps/**'],
+  collectCoverageFrom: [
+    '<rootDir>/ui/**/swaps/**',
+    '<rootDir>/ui/ducks/send/**',
```
dan437, line 6: Thank you for adding more code for the coverage and meeting the minimum coverage threshold! This is getting us closer to having coverage for almost the whole repo!

ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
```diff
@@ -359,84 +364,89 @@ export default class ConfirmTransactionBase extends Component {
 
     if (process.env.SHOW_EIP_1559_UI) {
       return (
```
danjm, line 366: This return block is the same as before the PR, with the exception of the inclusion of the `nonceField` (which conditionally renders if `useNonceField`, which was defined before this PR, is true)

ui/ducks/send/send.js
```diff
@@ -705,12 +728,85 @@ const slice = createSlice({
...
-     * sets the provided gasPrice in state and then recomputes the gasTotal
+     * Sets the appropriate gas fees in state and determines and sets the
+     * appropriate transactionType based on gas fee fields received.
      */
-    updateGasPrice: (state, action) => {
-      state.gas.gasPrice = addHexPrefix(action.payload);
+    updateGasFees: (state, action) => {
+      if (
+        action.payload.transactionType === TRANSACTION_ENVELOPE_TYPES.FEE_MARKET
+      ) {
```
danjm, line 737: Question for me to figure out: under what conditions can we hit this case? (It is a question because if `action.payload.transactionType === TRANSACTION_ENVELOPE_TYPES.FEE_MARKET`, then I assume we are are on EIP1559 compatible network, in which case I thought that we could _not_ update gas feels while on the send screen)... maybe (probably?) this is just for updates that happen in the background, as the user sets token and recipient?

ui/ducks/send/send.js
```diff
@@ -705,12 +728,85 @@ const slice = createSlice({
...
+        state.gas.maxFeePerGas = addHexPrefix(action.payload.maxFeePerGas);
+        state.gas.maxPriorityFeePerGas = addHexPrefix(
+          action.payload.maxPriorityFeePerGas,
+        );
+        state.transactionType = TRANSACTION_ENVELOPE_TYPES.FEE_MARKET;
+      } else {
```
danjm, line 743: I don't understand why this says "Until we remove the old UI we **don't** want to update gasPrice if the user has changed the field value." It seems inconsistent with the logic below, which updates gasPrice... I should come back to this

- [ ] Understand this

ui/ducks/send/send.js
danjm: - [ ] Understand this

ui/ducks/send/send.js
```diff
@@ -705,12 +728,85 @@ const slice = createSlice({
...
+              type: TRANSACTION_ENVELOPE_TYPES.LEGACY,
+              isAutomaticUpdate: true,
+            },
+          });
+          break;
+        case GAS_ESTIMATE_TYPES.ETH_GASPRICE:
+          gasPriceEstimate = getRoundedGasPrice(gasFeeEstimates.gasPrice);
+          slice.caseReducers.updateGasFees(state, {
+            payload: {
+              gasPrice: getRoundedGasPrice(gasFeeEstimates.gasPrice),
```
danjm, line 797: Looks like this can be `gasPrice: gasPriceEstimate`

ui/ducks/send/send.js
```diff
@@ -705,12 +728,85 @@ const slice = createSlice({
...
+        // Until we remove the old UI we don't want to update gasPrice if the
+        // user has changed the field value. When receiving a new estimate the
+        // isAutomaticUpdate property will be on the payload (and set to true).
+        // If isAutomaticUpdate is true, then we check if the previous estimate
+        // was '0x0' or if the previous gasPrice equals the previous
+        // gasEstimate. if either of those cases are true then we update the
+        // gasPrice otherwise we skip it because it indicates the user has
+        // ejected from the estimates by modifying the field.
+        if (
+          action.payload.isAutomaticUpdate !== true ||
```
danjm, line 753: Note that the one case were we can be in this else block and `isAutomaticUpdate` is true is when `updateGasPrice` is called, which is apparently a temporary placeholder to support the old UI

ui/ducks/send/send.js
```diff
@@ -1042,13 +1166,18 @@ const slice = createSlice({
...
+            gasEstimateType: action.payload.gasEstimateType,
+          },
+        });
         state.gas.gasTotal = action.payload.gasTotal;
         state.gas.gasEstimatePollToken = action.payload.gasEstimatePollToken;
         if (action.payload.gasEstimatePollToken) {
           state.gas.isGasEstimateLoading = false;
         }
-        if (state.stage !== SEND_STAGES.UNINITIALIZED) {
+        if (state.stage !== SEND_STAGES.INACTIVE) {
```
danjm, line 1180: TODO: understand the conditions under which stage in `INACTIVE` vs not inactive

danjm: - [ ] Understand this

ui/hooks/useSafeGasEstimatePolling.js
danjm: - [ ] Add some documentaiton

This is handling the case describe in this comment on a separate PR: https://github.com/MetaMask/metamask-extension/pull/11511/files#r670686360

ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
danjm: - [ ] Add documentation

This is handling the case describe in this comment on a separate PR: https://github.com/MetaMask/metamask-extension/pull/11511/files#r670686360

danjm:
rebased and fixed merge conflict

danjm:
I'm looking into some issues with editing dapp created transactions. To determine:

- [ ] Should editing only happen for legacy transactions?

If no, we need to:

- [ ] support edit mode fore EIP1559 transactions

darkwing:
Is the edit mode issue fixed by https://github.com/MetaMask/metamask-extension/pull/11589?

ui/ducks/send/send.js
danjm: The hardcoded initial state has the stage set to `INACTIVE`. It gets changed from this when:
- the `initializeSendState` think is fulfilled
- when the recipient is updated (including when it is first set)
- when beginning to edit a transcaction

I don't think there is a signficant difference between the `UNITIALIZED` and `INACTIVE` names

ui/ducks/send/send.js
```diff
@@ -1090,33 +1219,9 @@ const slice = createSlice({
       .addCase(GAS_FEE_ESTIMATES_UPDATED, (state, action) => {
         // When the gasFeeController updates its gas fee estimates we need to
         // update and validate state based on those new values
-        const { gasFeeEstimates, gasEstimateType } = action.payload;
```
danjm, line 1093: Code that serves the purpose of this set of removed lines is added to `updateGasFeeEstimates` (with appropriate modifications for getting the send slice to work with EIP1559 networks)

ui/selectors/confirm-transaction.js
danjm: After careful review and testing, I believe this code here is correct.

danjm:
> Is the edit mode issue fixed by #11589?

No, I am referring to clicking the `< Edit` button in the top left of the confirm screen, which takes the user back to the send screen.

ui/ducks/send/send.js
danjm: > maybe (probably?) this is just for updates that happen in the background, as the user sets token and recipient?

This is correct. Also, I believe it best that we do not update gas fee properties while in edit mode/stage. I have a change for that locally and will add it to this PR

ui/ducks/send/send.js
danjm: Okay, what is meant here is: "Until we remove the old UI we don't want to _automatically_ update gasPrice if the user has already manually changed the field value"

app/scripts/controllers/transactions/index.js
darkwing: Still needs removal

app/scripts/controllers/transactions/index.js
darkwing: Still needs removal

ui/ducks/send/send.js
darkwing: Is this still relevant?

darkwing (dismissed):

ui/ducks/send/send.js
adonesky1: @danjm Seems to still be a nit fix here. Looks like just above we're setting gasPriceEstimate to the same value we want to set here so we can just use that variable instead.

ui/ducks/send/send.js
```diff
@@ -705,12 +733,86 @@ const slice = createSlice({
...
+              isAutomaticUpdate: true,
+            },
+          });
+          break;
+        case GAS_ESTIMATE_TYPES.ETH_GASPRICE:
+          gasPriceEstimate = getRoundedGasPrice(gasFeeEstimates.gasPrice);
+          slice.caseReducers.updateGasFees(state, {
+            payload: {
+              gasPrice: getRoundedGasPrice(gasFeeEstimates.gasPrice),
+              type: TRANSACTION_ENVELOPE_TYPES.LEGACY,
```
adonesky1, line 804: @danjm shouldn't this be `TRANSACTION_ENVELOPE_TYPES.ETH_GASPRICE`?

ui/ducks/send/send.js
```diff
@@ -434,6 +440,9 @@ export const initializeSendState = createAsyncThunk(
       metamask: { gasFeeEstimates, gasEstimateType },
     } = thunkApi.getState();
 
+    // Because we are only interested in getting a gasLimit estimation we only
+    // need to worry about gasPrice. So we use maxFeePerGas as gasPrice if we
+    // have a fee market estimation.
     if (gasEstimateType === GAS_ESTIMATE_TYPES.LEGACY) {
       gasPrice = getGasPriceInHexWei(gasFeeEstimates.medium);
```
adonesky1, line 447: @danjm I know this wasn't part of the PR but do we need to protect against a possibly undefined value here?

ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
```diff
@@ -370,123 +377,133 @@ export default class ConfirmTransactionBase extends Component {
...
-                }
-                subTitle={t('transactionDetailGasTotalSubtitle')}
-                subText={t('editGasSubTextAmount', [
-                  <UserPreferencedCurrencyDisplay
-                    key="gas-total-subtext"
-                    type={SECONDARY}
-                    value={addHexes(
-                      txData.txParams.value,
-                      getHexGasTotal({
+        <>
```
adonesky1, line 380: @danjm Did you happen to figure out what's going on with this big diff section?

adonesky1 (dismissed):
LGTM. Just a few small questions.

ui/ducks/send/send.js
```diff
@@ -442,6 +451,10 @@ export const initializeSendState = createAsyncThunk(
       gasPrice = getGasPriceInHexWei(
         gasFeeEstimates.medium.suggestedMaxFeePerGas,
       );
+    } else {
```
danjm, line 454: @PeterYinusa @adonesky1 This is the one change I made since last review

darkwing (dismissed):

ui/ducks/send/send.js
adonesky1: 👍

adonesky1 (approved):

darkwing (approved):

PR #11552, opened 2021-07-16
Integrate TokensController

adonesky1:
Required changes to land the controllers repo [TokensController](https://github.com/MetaMask/controllers/pull/518). This is pre-requisite work to next land the TokenRatesController in the extension and have access to token exchange rates for other custom networks.

Manual testing steps:  
  - Broad regression testing for all token related actions.
  - Testing migration scenario from old Preference Controller based token state to the new TokensController based token state.


app/scripts/controllers/detect-tokens.js
```diff
@@ -48,7 +51,9 @@ export default class DetectTokensController {
       if (
         contracts[contractAddress].erc20 &&
         !this.tokenAddresses.includes(contractAddress.toLowerCase()) &&
-        !this.hiddenTokens.includes(contractAddress.toLowerCase())
+        !this.hiddenTokens.find((token) =>
+          isEqualCaseInsensitive(token.address, contractAddress),
```
danjm, line 55: @NiranjanaBinoy see this `isEqualCaseInsensitive` method, as an alternative to how you handle case differences in your PR. I think this is okay for comparison of token addresses. You will have to change to this within this file, once you rebase.

ui/helpers/utils/util.js
```diff
@@ -50,6 +50,10 @@ export function isDefaultMetaMaskChain(chainId) {
   return false;
 }
 
+export function isEqualCaseInsensitive(value1, value2) {
```
danjm, line 53: If both values are are undefined or of a type that does not have a `toLowerCase` method, should we return `true`?

app/scripts/controllers/preferences.js
```diff
@@ -767,189 +610,4 @@ export default class PreferencesController {
...
-    });
-  }
-
-  /**
-   * Detects whether or not a token is ERC-721 compatible.
-   *
-   * @param {string} tokensAddress - the token contract address.
-   *
-   */
-  async _detectIsERC721(tokenAddress) {
```
danjm, line 803: So this detection is now handled in the assets controller?

ui/helpers/utils/util.js
adonesky1: Good point. We should probably address that. Ideally this would be typed for only string inputs. For its current use case - finding matching token addresses - I don't think want to return true for either case (if both undefined or both are not strings). So I suppose we should rename this function to more specifically call out its use case and avoid confusion or use where those scenarios should return true?

app/scripts/controllers/preferences.js
adonesky1: in the TokensController yes: https://github.com/MetaMask/controllers/blob/176418fc9411fba27edcdb1b92f8a857a8e8700c/src/assets/TokensController.ts#L326

app/scripts/controllers/detect-tokens.js
```diff
@@ -64,17 +69,18 @@ export default class DetectTokensController {
...
-      const balance = result[index];
-      if (balance && !balance.isZero()) {
-        this._preferences.addToken(
-          tokenAddress,
-          contracts[tokenAddress].symbol,
-          contracts[tokenAddress].decimals,
-        );
-      }
-    });
+    await Promise.all(
```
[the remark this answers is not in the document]
adonesky1: 🤔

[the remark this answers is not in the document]
adonesky1: I have no recollection of why I thought this was necessary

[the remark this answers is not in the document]
adonesky1: removed all the unnecessary awaiting!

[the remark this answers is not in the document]
adonesky1: Hmmm moving too fast today I guess. It was made for testing. Is the move here to adapt the tests?

ui/helpers/utils/util.js
```diff
@@ -50,6 +50,12 @@ export function isDefaultMetaMaskChain(chainId) {
   return false;
 }
 
+// Both inputs should be strings. This method is currently used to compare tokenAddress hex strings.
+export function isEqualCaseInsensitive(value1, value2) {
+  if (typeof value1 !== 'string' || typeof value2 !== 'string') return false;
+  return value1?.toLowerCase() === value2?.toLowerCase();
```
ryanml, line 56: Nit: are the `?`'s necessary? given both `value1` and `value2` will be of type `string` at this comparison

adonesky1: no they are not required anymore. Will remove!

ui/hooks/useTokenTracker.js
```diff
@@ -98,7 +103,7 @@ export function useTokenTracker(
       return;
     }
 
-    if (memoizedTokens.length === 0) {
+    if (memoizedTokens?.length === 0) {
```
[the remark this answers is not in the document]
adonesky1: Trying to remember 🤔

ui/store/actions.js
```diff
@@ -1121,18 +1122,17 @@ export function lockMetamask() {
   };
 }
 
-async function _setSelectedAddress(dispatch, address) {
+async function _setSelectedAddress(address) {
   log.debug(`background.setSelectedAddress`);
-  const tokens = await promisifiedBackground.setSelectedAddress(address);
-  dispatch(updateTokens(tokens));
```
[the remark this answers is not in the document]
adonesky1: Tokens state is now managed exclusively in the TokensController. So the tokens are updated when the selectedAddress is [updated with a listener in the TokensController](https://github.com/MetaMask/controllers/blob/5ab9cd605e965340582e71009ab6a06d4cf96826/src/assets/TokensController.ts#L181).

ui/hooks/useTokenTracker.js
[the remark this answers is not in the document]
adonesky1: removed.

app/scripts/controllers/preferences.js
```diff
@@ -4,29 +4,18 @@ import { ethErrors } from 'eth-rpc-errors';
 import { normalize as normalizeAddress } from 'eth-sig-util';
 import { ethers } from 'ethers';
 import log from 'loglevel';
-import abiERC721 from 'human-standard-collectible-abi';
-import contractsMap from '@metamask/contract-metadata';
-import { LISTED_CONTRACT_ADDRESSES } from '../../../shared/constants/tokens';
+import { Promise } from 'globalthis/implementation';
```
danjm, line 7: Probably not needed

app/scripts/metamask-controller.js
```diff
@@ -1284,14 +1307,16 @@ export default class MetamaskController extends EventEmitter {
...
-      tokens,
     } = this.preferencesController.store.getState();
 
+    const { tokens, allTokens } = this.tokensController.state;
+
+    const accountTokens = allTokens[selectedAddress];
+
     // Filter ERC20 tokens
     const filteredAccountTokens = {};
     Object.keys(accountTokens).forEach((address) => {
```
danjm, line 1322: Confirm whether or not we need this

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.component.js
```diff
@@ -90,31 +96,25 @@ export default class ConfirmAddSuggestedToken extends Component {
               </div>
             </div>
             <div className="confirm-add-token__token-list">
-              {Object.entries(pendingTokens).map(([address, token]) => {
```
danjm, line 93: Looks like we need to continue to do a mapping like this

danjm:
Migrations are needed

app/scripts/metamask-controller.js
adonesky1: So I'm fairly certain we still want this. We need to filter out the non-ERC20 tokens because they are handled differently by mobile. We don't have a great way to keep pass say ERC721 tokens through the sync and I think we might as well wait until we have better handling of them in the extension to include them as part of the sync. 

I've altered the logic here to do a more thorough sweep to filter out ERC721 tokens. Please take another look!

app/scripts/migrations/063.js
```diff
@@ -0,0 +1,55 @@
...
+    return versionedData;
+  },
+};
+
+function transformState(state) {
+  const accountTokens = state?.PreferencesController?.accountTokens;
+  const accountHiddenTokens = state?.PreferencesController?.accountHiddenTokens;
+
+  if (accountTokens || accountHiddenTokens) {
+    const flattenedIgnoredTokens = [];
```
adonesky1, line 25: So the ignoredTokens data in the TokensController is currently flat. This makes ignored tokens ignored across all chains and accounts which is probably not ideal... I had made it this way in the TokensController because this was the shape in the AssetsController. Perhaps we should make the TokensController version conform to the current shape of accountHiddenTokens and change mobile to use that pattern instead of what I'm doing here? @danjm

danjm: Yes, I think you are right.

danjm: Might be worth asking mobile if there is any reason we don't do that already. But the new token detection will detect tokens by network, so it makes sense to ignore by network.

adonesky1: @danjm Here is a PR to update the shape in the TokensController/AssetsDetectionController so this can move forward: https://github.com/MetaMask/controllers/pull/570

app/scripts/controllers/detect-tokens.js
```diff
@@ -48,7 +69,9 @@ export default class DetectTokensController {
       if (
         contracts[contractAddress].erc20 &&
         !this.tokenAddresses.includes(contractAddress.toLowerCase()) &&
```
NiranjanaBinoy, line 71: [Q]: Is there any reason we are not using the isEqualCaseInsensitive here as well?

ui/store/actions.js
```diff
@@ -1297,40 +1291,49 @@ export function addTokens(tokens) {
   };
 }
 
-export function removeSuggestedTokens() {
-  return (dispatch) => {
+export function rejectWatchAsset(suggestedAssetID) {
```
brad-decker, line 1294: [Q] is this name change strictly to align with the underlying provider method?

ui/store/actions.js
```diff
@@ -1297,40 +1291,49 @@ export function addTokens(tokens) {
...
+      dispatch(displayWarning(error.message));
+      return;
+    } finally {
+      dispatch(hideLoadingIndication());
+    }
+
+    dispatch(closeCurrentNotificationWindow());
+    if (getEnvironmentType() === ENVIRONMENT_TYPE_NOTIFICATION) {
+      global.platform.closeCurrentWindow();
+    }
```
brad-decker, line 1310: Given that closeCurrentNotificationWindow is defined as such:

```js
export function closeCurrentNotificationWindow() {
  return (_, getState) => {
    if (
      getEnvironmentType() === ENVIRONMENT_TYPE_NOTIFICATION &&
      !hasUnconfirmedTransactions(getState())
    ) {
      global.platform.closeCurrentWindow();
    }
  };
}
```

It appears that line 1307 will make the following 3 lines unnecessary. Also, 1308-1310 do not check for existing notifications before closing.

test/e2e/tests/contract-interactions.spec.js
```diff
@@ -81,10 +81,11 @@ describe('Deploy contract and call contract methods', function () {
         await driver.clickElement('#depositButton');
         await driver.waitUntilXWindowHandles(3);
         windowHandles = await driver.getAllWindowHandles();
-        await driver.switchToWindowWithTitle(
-          'MetaMask Notification',
-          windowHandles,
+        popup = windowHandles.find(
+          (handle) => handle !== extension && handle !== dapp,
         );
+        await driver.switchToWindow(popup);
```
brad-decker, line 87: [Q] say what?

brad-decker (changes requested):
Minor thing i think might be an issue, but other than that LGTM

ui/store/actions.js
adonesky1: Yes. And it also now rejects one pending/suggested asset. The change to handling a single asset at a time is not yet leveraged but I think it should be soon!

ui/store/actions.js
adonesky1: Well sleuthed. Removing 1308-1310

test/e2e/tests/contract-interactions.spec.js
adonesky1: I honestly don't know... For some reason the driver was no longer finding the window titled Metamask Notification, even though I could see that is still in fact the title of the window.

adonesky1: Its possible firefox changed how our (and other extension) notifications are titled: Screen Shot 2021-09-07 at 2 09 59 PM

I think this change is more robust to these kinds of changes?

adonesky1: But why would it only show up on this PR...

brad-decker: i wonder if related to that double notification window close?

adonesky1: oh yeah possibly. Unfortunately I can't confirm that atm because this test is dependent on gas estimation api availability  (which is [currently intermittent](https://consensys.slack.com/archives/GTQAGKY5V/p1631043080226300))

adonesky1: That does appear to have been the issue 🤦 🤛  Pushed a commit to revert this change!

brad-decker (dismissed):
LGTM

app/scripts/controllers/detect-tokens.js
adonesky1: You're probably right that this should be a case insensitive check. Will do now.

adonesky1:
@brad-decker made a small update since your review 😃

brad-decker (dismissed):

ui/components/ui/list-item/list-item.component.js
```diff
@@ -19,7 +19,6 @@ export default function ListItem({
     className,
     subtitle || children ? '' : 'list-item--single-content-row',
   );
-
```
ryanml, line 22: nit: any reason for this?

adonesky1: I didn't like it there 😄 no not really haha

app/scripts/migrations/063.js
```diff
@@ -0,0 +1,78 @@
+import { cloneDeep } from 'lodash';
```
danjm, line 1: This migration looks correct

app/scripts/metamask-controller.js
```diff
@@ -1289,43 +1312,47 @@ export default class MetamaskController extends EventEmitter {
...
+      allERC20Tokens[chainId] = {};
+      Object.keys(allTokens[chainId]).forEach((accountAddress) => {
+        const checksummedAccountAddress = toChecksumHexAddress(accountAddress);
+        allERC20Tokens[chainId][checksummedAccountAddress] = allTokens[chainId][
+          checksummedAccountAddress
+        ].filter((asset) => {
+          if (asset.isERC721 === undefined) {
+            const checksumAddress = toChecksumHexAddress(asset.address);
+            if (contractMap[checksumAddress] !== undefined) {
+              if (contractMap[checksumAddress].erc20) {
```
darkwing, line 1344: Can these two `if`'s be combined into one with `&&`

ui/components/app/modals/hide-token-confirmation-modal/hide-token-confirmation-modal.js
```diff
@@ -74,7 +72,9 @@ class HideTokenConfirmationModal extends Component {
               type="secondary"
               className="hide-token-confirmation__button"
               data-testid="hide-token-confirmation__hide"
-              onClick={() => hideToken(address)}
+              onClick={() => {
+                hideToken(address);
```
darkwing, line 76: Unnecessary change but no big deal

app/scripts/metamask-controller.js
adonesky1: yes they could 😺

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.component.js
```diff
@@ -124,10 +128,11 @@ export default class ConfirmAddSuggestedToken extends Component {
               type="default"
               large
               className="page-container__footer-button"
-              onClick={() => {
-                removeSuggestedTokens().then(() =>
-                  history.push(mostRecentOverviewPage),
-                );
+              onClick={async () => {
+                await suggestedAssets.forEach(async ({ id }) => {
```
darkwing, line 132: I wasn't aware you could do this on a `forEach` ; maybe a `Promise.all`+`suggestedAssets.map` would be more readable?

adonesky1: This does work but happy to make that change if you'd prefer!

adonesky1:
@brad-decker just address minor nits

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.component.js
adonesky1: ping here @darkwing 😅

brad-decker: the await isn't doing anything here. The only reason it works is that the iterator is an async and all the logic is done there. Map + promise.all is a cleaner way of doing this.

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.component.js
```diff
@@ -136,24 +141,23 @@ export default class ConfirmAddSuggestedToken extends Component {
...
-                        token_decimal_precision: pendingToken.decimals,
-                        unlisted: pendingToken.unlisted,
-                        source: 'dapp',
-                      },
-                    });
-                  })
-                  .then(() => history.push(mostRecentOverviewPage));
+              disabled={suggestedAssets.length === 0}
+              onClick={async () => {
+                await suggestedAssets.forEach(async ({ asset, id }) => {
```
brad-decker, line 146: same here as @darkwing 's initial comment.

adonesky1:
Resolved conflicts with @NiranjanaBinoy's PR. Obviously needed to rebase so sorry for breaking review tracking flows. Tests failing because gas API is down. Please review (or re-review) as soon as you can! cc @danjm, @darkwing, @brad-decker, @ryanml. 🙏 🙏 🙏

ui/store/actions.js
```diff
@@ -1298,40 +1291,45 @@ export function addTokens(tokens) {
...
```
brad-decker, line 1326:
```suggestion
    dispatch(closeCurrentNotificationWindow());
```

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.component.js
```diff
@@ -124,10 +128,13 @@ export default class ConfirmAddSuggestedToken extends Component {
...
               large
               className="page-container__footer-button"
-              onClick={() => {
-                removeSuggestedTokens().then(() =>
-                  history.push(mostRecentOverviewPage),
+              onClick={async () => {
+                await Promise.all(
+                  suggestedAssets.map(async ({ id }) => {
+                    await rejectWatchAsset(id);
+                  }),
```
brad-decker, line 135:
```suggestion
                await Promise.all(
                  suggestedAssets.map(({ id }) => rejectWatchAsset(id)
                ),
```

you don't need to await inside here, the await of Promise.all waits for all returned promises in the array to be resolved.

brad-decker (changes requested):
some code style/improvement things, but all good after that.

app/scripts/controllers/detect-tokens.js
```diff
@@ -91,8 +115,14 @@ export default class DetectTokensController {
       await Promise.all(
         tokensSlice.map(async (tokenAddress, index) => {
           const balance = result[index];
-          if (balance && !balance.isZero()) {
-            await this._preferences.addToken(
+          let ignored;
+          if (this.hiddenTokens.length) {
```
darkwing, line 119: I don't think we need this `this.hiddenTokens.length` check and `ignored` can be a `const` with the `find` result, then below, do an `ignored.length` check.  If `this.hiddenTokens` is an empty array, the `find` will never be executed, so there's no advantage to the length check here.

darkwing: Happy to hear what @brad-decker or anyone else thinks though.

ui/store/actions.js
adonesky1: Oh whoops this must have gotten reverted in the rebase somehow... that's a little scary...

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.component.js
adonesky1: Indeed ser!

app/scripts/controllers/detect-tokens.js
adonesky1: I think you're right, though find just returns one element so no need to check its length, just whether or not its defined.

adonesky1: Wondering if this is a more legible approach: 
```
const isIgnored = Boolean(this.hiddenTokens.find((ignoredTokenAddress) =>
            isEqualCaseInsensitive(ignoredTokenAddress, tokenAddress),
          ));
          if (balance && !balance.isZero() && !isIgnored) {
            await this.tokensController.addToken(
              tokenAddress,
              tokenList[tokenAddress].symbol,
              tokenList[tokenAddress].decimals,
            );
          }
```

brad-decker: You can't return true here, it would end up returning a boolean value that can't be passed to Promise.all. A safer approach would be to first create a new list that has only NON ignored tokens in it and then do the promise.all on that map.

The reason for that is two fold. First as mentioned above you can't use .map for branching logic here. You need the return values to be all of the same type, and they must be promises. Finally, iterating on a list inside a .map has potential performance issues. You'll be iterating up to n times per item in tokensSlice. Better to reduce the list of what you are iterating on ahead of time. You could use `Set` to quickly get to a list of tokens that are unique and not ignored.

ui/components/app/wallet-overview/token-overview.js
```diff
@@ -136,11 +133,7 @@ const TokenOverview = ({ className, token }) => {
       }
       className={className}
       icon={
-        <Identicon
-          diameter={32}
-          address={token.address}
-          image={assetImages[token.address]}
-        />
+        <Identicon diameter={32} address={token.address} image={token.image} />
```
darkwing, line 136: Line 109 has `token?.image`, but not here; should it be the same in both places?

app/scripts/controllers/detect-tokens.js
adonesky1: I just realized that the checking for ignored was redundant... that check is already done on line 93. Here is what I have now: 
```
const tokensWithBalance = tokensSlice.filter((_, index) => {
        const balance = result[index];
        return balance && !balance.isZero();
      });

      await Promise.all(
        tokensWithBalance.map((tokenAddress) => {
          return this.tokensController.addToken(
            tokenAddress,
            tokenList[tokenAddress].symbol,
            tokenList[tokenAddress].decimals,
          );
        }),
      );
  ```

File: ui/components/app/wallet-overview/token-overview.js (hunk 25 above)
Comment (adonesky1) [reply ui/components/app/wallet-overview/token-overview.js:136]: good catch. I'll remove the optional chaining, since we infer and rely on the fact that token object is a defined object throughout this component.

File: ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.component.js [hunk 26]
```diff
@@ -136,24 +143,25 @@ export default class ConfirmAddSuggestedToken extends Component {
...
               large
               className="page-container__footer-button"
-              disabled={pendingTokens.length === 0}
-              onClick={() => {
-                addToken(pendingToken)
-                  .then(() => removeSuggestedTokens())
-                  .then(() => {
+              disabled={suggestedAssets.length === 0}
+              onClick={async () => {
+                await Promise.All(
```
darkwing, line 146: This will fail -- should be `Promise.all`

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.component.js
```diff
@@ -175,20 +185,18 @@ export default class ConfirmAddSuggestedToken extends Component {
...
-      .filter((token) => {
-        const dupes = tokens
-          .filter((old) => old.symbol === token.symbol)
-          .filter((old) => old.address !== token.address);
-        return dupes.length > 0;
-      });
+  checkNameReuse(suggestedAssets, tokens) {
+    const duplicates = suggestedAssets.filter(({ asset }) => {
+      const dupes = tokens
+        .filter((old) => old.symbol === asset.symbol)
```
darkwing, line 194: Doing two consecutive filters can be expensive -- best to combine them into one.

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.component.js
adonesky1: whoa... thanks

app/scripts/controllers/detect-tokens.js
brad-decker: that looks good!

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.component.js
adonesky1: great catch hawk eye

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.component.js
adonesky1: ✅

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.component.js
adonesky1: ✅

ui/components/app/wallet-overview/token-overview.js
adonesky1: ✅

app/scripts/controllers/detect-tokens.js
adonesky1: ✅

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.component.js
adonesky1: ✅

ui/store/actions.js
adonesky1: ✅

brad-decker (approved):

darkwing (approved):

danjm (approved):

PR #11555, opened 2021-07-17
build - log the bundle completion

kumavis:
now logs the name of the bundle at start and end
ideally this would happen via each task ending,
but currently tasks dont end when they are watching the filesystem


ryanml (approved):

darkwing:
Works for me!  Would be cool in the future to use Chalk (https://davidwalsh.name/chalk) to use green and yellow for end and start, but obviously not important right now.

darkwing (approved):

Issue #11558, opened 2021-07-19
Can't transfer USDC on xDai


KimiWu123:
**Describe the bug**
can not transfer ERC20 token on xDai

**Steps to reproduce (REQUIRED)**
1. Switch network to xDai
2. click `transfer` and  select USDC (or other non-native tokens) as asset
3. when changed the number of the `Amount`, the `Next` button is disabled

**Expected behavior**
should be able to transfer to another account.

**Screenshots**
![Screen_Shot_2021-07-19_at_11 11 19_AM](https://user-images.githubusercontent.com/19582168/126097620-313c27e6-7a62-454f-86f2-fdd371aceab0.png)

**Browser details (please complete the following information):**
 - OS: Mac OSX
 - Hardware Wallet: na
 - Browser: Chrome 91.0.4472.114
 - MetaMask Version: 9.8.0

**Additional context (Error Messages, etc.)**
none

amitOodles:
Facing a similar issue on my Metamask as well on Firefox, once i have entered the amount the Next button is getting disabled.

amitOodles:
@danjm @darkwing @KimiWu123 
Regarding my issue, i am talking about normal account to account transactions not being sent on custom networks.

TtheBC01:
Also having this problem just trying to transfer between accounts on Hardhat.
![metamask_bug](https://user-images.githubusercontent.com/6946868/126570212-1ae3b97c-b50e-4487-b7a0-3781166ccf69.gif)

OS: Windows 10
Browser: Chrome
Metamask version: 9.8.1

PR #11565, opened 2021-07-20
Ensure gas limit estimation for token sends has correct default/base gas

danjm:
Fixes: https://github.com/MetaMask/metamask-extension/issues/11558

TODO: Provide full explanation as to why the bug occured and why this works.

Note: This change is on develop, brought in by https://github.com/MetaMask/metamask-extension/pull/11511, this PR picks out that part of that change, so that it can be shipped independently. @brad-decker discovered this bug during work on EIP-1559, and believed it to only affect local ganache on develop, but it is affecting some (but not all) custom networks as well.

### DIFF
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -182,6 +182,7 @@ async function estimateGasLimitForSend({
 }) {
   let isSimpleSendOnNonStandardNetwork = false;
 
+  // blockGasLimit may be a falsy, but defined, value when we receive it from
   // state, so we use logical or to fall back to MIN_GAS_LIMIT_HEX. Some
   // network implementations check the gas parameter supplied to
   // eth_estimateGas for validity. For this reason, we set token sends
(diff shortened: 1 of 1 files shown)


ui/ducks/send/send.js
```diff
@@ -182,9 +182,19 @@ async function estimateGasLimitForSend({
 }) {
   let isSimpleSendOnNonStandardNetwork = false;
 
-  // blockGasLimit may be a falsy, but defined, value when we receive it from
-  // state, so we use logical or to fall back to MIN_GAS_LIMIT_HEX.
-  const blockGasLimit = options.blockGasLimit || MIN_GAS_LIMIT_HEX;
+  // state, so we use logical or to fall back to MIN_GAS_LIMIT_HEX. Some
```
darkwing, line 185: I think we want to keep the comment starting on the original line 185 so that the comment block doesn't start with `state,`

danjm: @darkwing applied in latest commit

ryanml (approved):

darkwing (approved):

PR #11571, opened 2021-07-20
fix: lock and unlock events

rynst:
[Investigating Issue #10368](https://github.com/MetaMask/metamask-extension/issues/10368)


rsteubs:

niconiahi:
what's the status on this? would be neat to have it already

rsteubs:
Still working on it @niconiahi

rsteubs:
After digging into this, it looks like lock events just aren't added:
On `line 321` of https://github.com/MetaMask/providers/blob/main/src/BaseProvider.ts, `There are no lock/unlock events.`

I have a commit for this, though there might be other issues to getting it working.  Closing since this is a feature, rather than a bug.
https://github.com/rsteubs/providers/commit/0951d26fefc06bd646df4decf9cbb132fd027918

PR #11576, opened 2021-07-21
EIP-1559 - Show fiat values where necessary

### LINKED ISSUE #11573: EIP-1559 - Bug - Fiat Value Not Displaying Properly in Confirmation screen
The confirmation screen is displaying the ETH value twice, instead of the fiat value:

NumbersBug

darkwing:
Fixes: #11573, 11574

Explanation:  The reason we aren't seeing fiat values is because we can only use 1559 on test networks, and we have an advanced setting which defaults to hiding fiat values.  

My opinion is we should show fiat values for 1559 v1, and if necessary, hide them in v2.  The design looks bare without them.

### DIFF
--- a/ui/hooks/useGasFeeInputs.js
+++ b/ui/hooks/useGasFeeInputs.js
@@ -1,6 +1,5 @@
 import { addHexPrefix } from 'ethereumjs-util';
 import { useCallback, useState } from 'react';
-import { useSelector } from 'react-redux';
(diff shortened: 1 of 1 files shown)


danjm:
Won't the fiat values display on test networks if we turn on the option in settings? (They will display on mainnet by default)

darkwing:
Yeah, good shout, we shouldn't change our present strategy for test nets.

PR #11577, opened 2021-07-21
EIP-1559 - Maintain changed form values and unselect radio buttons when user customizes form value

darkwing:
If a user manually changes a form value, we need to clear the radio buttons while maintaining the form values.


darkwing:
This will require a bit more thought; if the user clicks the "x" to close, we won't want to change the stored values.

danjm (approved):

danjm:
> This will require a bit more thought; if the user clicks the "x" to close, we won't want to change the stored values.

If the user clicks the "x" to close, will the stored values be changed? I thought not: the values stored in the transaction will only be changed when the user clicks "Save". The values stored in the hook will be changed, but when the model is closed, that instance of the hook will be destroyed.

Or am I misunderstanding something?

darkwing:
You're right, please disregard comment!

NiranjanaBinoy (approved):
LGTM

PR #11581, opened 2021-07-21
Ensure send slice gas loading state is set to false upon gas estimation error

### LINKED ISSUE #11579: Cannot submit transactions that send ETH to some contracts on 9.8.0
**Describe the bug**
The "Next" button is disabled when trying to send transactions with value to some contracts. This does not happen when sending to every contract, and does not happen when sending to external accounts (non-contract).

1. Unlock your metamask wallet
2. Go to 'Send'
2. Set this contract 0x30b84DC1A46C58E0BcfE6AA9f74042dFf159277a as target (as example)
3. Set a non-zero amount (eg. 0.001 ETH)
4. The "Next" button is disabled

**Expected behavior**
The "Next" button should not be disabled

**Screenshots**
image

 - OS: Linux & Mac (haven't tested on Windows)
 - Hardware Wallet: none
 - Browser: Chrome & Opera & Firefox 
 - MetaMask Version: 9.8.0

If you enable advanced gas controls in settings and click on advanced settings and save the same gasLimit and gasPrice, then, the Next button is enabled.

danjm:
Fixes: https://github.com/MetaMask/metamask-extension/issues/11579

As #11579 describes, sending to some contracts cannot be done on v9.8.0. The user is prevented from doing so because the "Next" button is disabled.

The reason for this is that the "Next" button will be disabled while our eth_estimateGas call is in flight. This is represented with the `isGasEstimateLoading` property in the send slice state. A bug with this in v9.8.0 is that `isGasEstimateLoading` does not get set back to false when the eth_estimateGas call results in an error.

This PR corrects this by updating the `isGasEstimateLoading` state in the case of as gas estimation error.

This means that we are not doing anything in response to that error - and just defaulting to a 21000 gas limit - but that was the case before v9.8.0 going back as far as MetaMask v5.0.0, so I believe it is correct to stay consistent with that

### DIFF
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -1082,6 +1082,11 @@ const slice = createSlice({
           });
         }
       })
+      .addCase(computeEstimatedGasLimit.rejected, (state) => {
+        // If gas estimation fails, we should set the loading state to false,
+        // because it is no longer loading
+        state.gas.isGasEstimateLoading = false;
+      })
       .addCase(GAS_FEE_ESTIMATES_UPDATED, (state, action) => {
         // When the gasFeeController updates its gas fee estimates we need to
         // update and validate state based on those new values


adonesky1 (approved):
Makes sense to me.

NiranjanaBinoy (approved):
LGTM

PR #11585, opened 2021-07-22
No longer deconstructing account menu ref if it isn't rendered

ryanml:
Related Sentry error:
https://sentry.io/organizations/metamask/issues/2211680764/?project=273505&query=is%3Aunresolved

`setShouldShowScrollButton` is called in a few places throughout the component. There's a race condition scenario if it's called in the instant before the account menu renders. We can safely return inside `setShouldShowScrollButton` if the ref isn't set.

### DIFF
--- a/ui/components/app/account-menu/account-menu.component.js
+++ b/ui/components/app/account-menu/account-menu.component.js
@@ -253,12 +253,13 @@ export default class AccountMenu extends Component {
   }
 
   setShouldShowScrollButton = () => {
+    if (!this.accountsRef) {
+      return;
+    }
+
     const { scrollTop, offsetHeight, scrollHeight } = this.accountsRef;
-
     const canScroll = scrollHeight > offsetHeight;
-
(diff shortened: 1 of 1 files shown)


ui/components/app/account-menu/account-menu.component.js
```diff
@@ -253,12 +253,13 @@ export default class AccountMenu extends Component {
...
 
   setShouldShowScrollButton = () => {
-    const { scrollTop, offsetHeight, scrollHeight } = this.accountsRef;
+    if (!this.accountsRef) {
+      return;
+    }
 
+    const { scrollTop, offsetHeight, scrollHeight } = this.accountsRef;
     const canScroll = scrollHeight > offsetHeight;
-
```
ryanml, line 259: Spacing preferencee B)

danjm (approved):

Gudahtt (approved):
LGTM!

PR #11587, opened 2021-07-22
Add token decimals validation in Swaps

dan437:
## Explanation
We need to make sure a user cannot send a number with more decimals than a token supports.

## Manual testing steps
- Open the extension
- Click on `Swap` on the main page
- Select `USDC` as the `Swap from` token
- Type a number `0.0000001`
- Inline error message is displayed and the `Review Swap` button is disabled

## Screenshots
Inline error message and disabled `Review Swap` button:
image

Inline warning message and enabled `Review Swap` button:
image

### DIFF
--- a/ui/pages/swaps/build-quote/build-quote.js
+++ b/ui/pages/swaps/build-quote/build-quote.js
@@ -74,6 +74,7 @@ export default function BuildQuote({
   maxSlippage,
   selectedAccountAddress,
   isFeatureFlagLoaded,
+  tokenFromError,
 }) {
   const t = useContext(I18nContext);
   const dispatch = useDispatch();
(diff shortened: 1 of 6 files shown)


danjm (approved):

ui/pages/swaps/build-quote/build-quote.js
```diff
@@ -367,6 +368,11 @@ export default function BuildQuote({
     }
   }
 
+  const swapYourTokenBalance = t('swapYourTokenBalance', [
```
darkwing, line 371: Are these arguments backward?  `$1 is a token symbol and $2 is the max. number of decimals allowed for the token`

dan437: This is an existing translation, which I just abstracted here since it was used multiple times.

```
"swapYourTokenBalance": {
    "message": "$1 $2 available to swap",
    "description": "Tells the user how much of a token they have in their balance. $1 is a decimal number amount of tokens, and $2 is a token symbol"
  },
```

darkwing (approved):

PR #11589, opened 2021-07-22
EIP-1559 - Restore custom values in Edit Gas Popover

darkwing:
Here we grab values from an existing transaction so we can restore values chose by the user.  

## ToDo
- [x] Why are these prices being set ahead of time?  We should be defaulting to medium values (are we?)
- [x] Compare those custom values with the gasEstimates to see if we should select any of the radio buttons

### DIFF
--- a/ui/components/app/edit-gas-popover/edit-gas-popover.component.js
+++ b/ui/components/app/edit-gas-popover/edit-gas-popover.component.js
@@ -75,7 +75,7 @@ export default function EditGasPopover({
     hasGasErrors,
     gasErrors,
     onManualChange,
-  } = useGasFeeInputs(defaultEstimateToUse);
+  } = useGasFeeInputs(defaultEstimateToUse, transaction);
 
   /**
    * Temporary placeholder, this should be managed by the parent component but
@@ -120,6 +120,12 @@ export default function EditGasPopover({
         dispatch(createSpeedUpTransaction(transaction.id, newGasSettings));
         break;
       case EDIT_GAS_MODES.MODIFY_IN_PLACE:
+        // TODO:  This shouldn't be required by prevents the following error:
+        // Error: Invalid transaction params: specified gasPrice but also included maxFeePerGas, these cannot be mixed
+        if (process.env.SHOW_EIP_1559_UI) {
+          delete transaction.txParams.gasPrice;
+        }
+
         dispatch(
           updateTransaction({
             ...transaction,
--- a/ui/components/app/edit-gas-popover/edit-gas-popover.component.js
+++ b/ui/components/app/edit-gas-popover/edit-gas-popover.component.js
@@ -120,12 +120,6 @@ export default function EditGasPopover({
         dispatch(createSpeedUpTransaction(transaction.id, newGasSettings));
         break;
       case EDIT_GAS_MODES.MODIFY_IN_PLACE:
-        // TODO:  This shouldn't be required by prevents the following error:
-        // Error: Invalid transaction params: specified gasPrice but also included maxFeePerGas, these cannot be mixed
-        if (process.env.SHOW_EIP_1559_UI) {
-          delete transaction.txParams.gasPrice;
-        }
-
         dispatch(
           updateTransaction({
             ...transaction,
--- a/ui/hooks/useGasFeeInputs.js
+++ b/ui/hooks/useGasFeeInputs.js
@@ -80,6 +80,33 @@ function getGasFeeEstimate(
 }
 
 /**
+ * This method tries to determine if any estimate level matches the
+ * current maxFeePerGas and maxPriorityFeePerGas values. If we find
+ * a match, we can pre-select a radio button in the RadioGroup
+ */
+function getMatchingEstimateFromGasFees(
+  gasFeeEstimates,
+  maxFeePerGas,
+  maxPriorityFeePerGas,
+) {
+  if (process.env.SHOW_EIP_1559_UI) {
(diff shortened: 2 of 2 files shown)


ui/components/app/edit-gas-popover/edit-gas-popover.component.js
```diff
@@ -120,6 +120,12 @@ export default function EditGasPopover({
         dispatch(createSpeedUpTransaction(transaction.id, newGasSettings));
         break;
       case EDIT_GAS_MODES.MODIFY_IN_PLACE:
+        // TODO:  This shouldn't be required by prevents the following error:
```
danjm, line 123: I think this will likely be handled by this change in the send slice update PR: https://github.com/MetaMask/metamask-extension/pull/11549/files#diff-bad76d7279418f3520a0ca93cfaeeb6fcbb1f0d4782f6ff942d422632a1c2780R917

ui/hooks/useGasFeeInputs.js
```diff
@@ -78,6 +81,29 @@ function getGasFeeEstimate(
...
+  if (process.env.SHOW_EIP_1559_UI) {
+    return (
+      findKey(
+        gasFeeEstimates,
+        (estimate) =>
+          estimate?.suggestedMaxPriorityFeePerGas === maxPriorityFeePerGas &&
+          estimate?.suggestedMaxFeePerGas === maxFeePerGas,
+      ) || null
+    );
+  }
```
danjm, line 103: @darkwing Can you add the case for legacy low, medium and high estimates? This would require passing in `gasPrice` and checking when its matches the low, medium or high gasPrice in an else condition. (Even after the mainnet hardfork, this might be valuable on some custom networks, if we do provide a gas price estimate api for those.)

ui/hooks/useGasFeeInputs.js
```diff
@@ -164,6 +179,39 @@ export function useGasFeeInputs(defaultEstimateToUse = 'medium') {
     estimatedGasFeeTimeBounds,
   } = useGasFeeEstimates();
 
+  // This hook keeps track of a few pieces of transitional state. It is
+  // transitional because it is only used to modify a transaction in the
+  // metamask (background) state tree.
+  const [maxFeePerGas, setMaxFeePerGas] = useState(
+    transaction?.txParams?.maxFeePerGas
```
dan437, line 186: So, in Swaps when we want to edit gas on the `View Quote` page, we don't have a transaction yet.

However, we store custom gasLimit, maxFeePerGas and maxPriorityFeePerGas in swapsState, which we could pass to the `EditGasPopover` component, e.g. in the `customGas` object instead of the `transaction` object. 

What do you think?

ui/hooks/useGasFeeInputs.js
darkwing: Done!

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
darkwing: @danjm Is this confirmed?  Should I remove from here?

danjm (approved):

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
danjm: Confirmed that between that change and https://github.com/MetaMask/metamask-extension/pull/11615, this bug goes away and the below if block is not needed

ui/hooks/useGasFeeInputs.js
```diff
@@ -164,6 +180,40 @@ export function useGasFeeInputs(defaultEstimateToUse = 'medium') {
...
+  const [maxPriorityFeePerGas, setMaxPriorityFeePerGas] = useState(
+    transaction?.txParams?.maxPriorityFeePerGas
+      ? Number(hexWEIToDecGWEI(transaction.txParams.maxPriorityFeePerGas))
+      : null,
+  );
+  const [gasPrice, setGasPrice] = useState(
+    transaction?.txParams?.gasPrice
+      ? Number(hexWEIToDecGWEI(transaction.txParams.gasPrice))
+      : null,
+  );
```
danjm, line 200: This change is needed below:
Screenshot from 2021-07-27 04-59-04

danjm: Because the property on `txParams` is called `gas` and no `gasLimit`, and because gas/gasLimit is not in WEI or GWEI (it does not have a unit)

danjm: > the property on txParams is called `gas` and not `gasLimit`

We do need to better improve this naming situation in the codebase. Ideally, it would be the same everywhere... but that is for a different PR

danjm (changes requested):
see latest two comments

darkwing:
Updated!

danjm (approved):

dan437 (approved):

ui/hooks/useGasFeeInputs.js
adonesky1: @darkwing was this addressed ?

dan437: @adonesky1 Yup, but in another place. @danjm has a commit that handles prefilling of the gas editing modal in Swaps for EIP-1559. We just need to get this PR in and then I can include Dan's commit in my PR that I will open shortly.

darkwing: Swaps can be addressed separately from this PR, IMO

PR #11590, opened 2021-07-22
EIP-1559 - Show advanced form controls by default when user has selected custom values or there are gas warnings/errors

darkwing:
This is a big UX improvement I came across when writing https://github.com/MetaMask/metamask-extension/pull/11589 .  If the user has selected custom gas values, or if there were warnings when they chose their values, the next time they click "Edit" to edit gas we should simply show the advanced gas controls by default, since none of the radio buttons will be selected.  Otherwise it's an extra step for the user to see their custom values.


darkwing:
Happy to hear what @jakehaugen thinks.

danjm (approved):

jakehaugen:
Yes, this is great.

ryanml (approved):

PR #11592, opened 2021-07-23
Build with Spack

kumavis:
experiments happening here 🚧


kumavis:
current issue: spack does not provide a way to map "node core modules" (`events`, `assert`) to npm modules

PR #11594, opened 2021-07-23
Fix sending maximum of an ERC20 token

### LINKED ISSUE #11593: Can't sed "Max" of an ERC20 token
> I’m trying to send a token from MM to another wallet on BSC. I have enough BNB for gas (although I’m not making any gas price edits, just going with the default as suggested by MM). The tx keeps failing and I see this error. 

> Ok the problem was I was sending the Max amount, and for some reason the tx failed. I reduced the max amount to remove all the decimal numbers and used a whole number instead and it worked

danjm:
Fixes: https://github.com/MetaMask/metamask-extension/issues/11593

With the send slice refactor, we stopped correctly re-estimating gas limit when setting "Max" in the send flow. This PR ensures that is done correctly.

### DIFF
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -1164,9 +1164,7 @@ export function updateSendAmount(amount) {
     if (state.send.amount.mode === AMOUNT_MODES.MAX) {
       await dispatch(actions.updateAmountMode(AMOUNT_MODES.INPUT));
     }
-    if (state.send.asset.type === ASSET_TYPES.TOKEN) {
-      await dispatch(computeEstimatedGasLimit());
-    }
+    await dispatch(computeEstimatedGasLimit());
   };
 }
 
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -1326,6 +1326,7 @@ export function toggleSendMaxMode() {
       await dispatch(actions.updateAmountMode(AMOUNT_MODES.MAX));
       await dispatch(actions.updateAmountToMax());
     }
+    await dispatch(computeEstimatedGasLimit());
   };
 }
 
--- a/ui/ducks/send/send.test.js
+++ b/ui/ducks/send/send.test.js
@@ -1045,7 +1045,34 @@ describe('Send Slice', () => {
       };
 
       it('should create an action to update send amount', async () => {
-        const store = mockStore(defaultSendAmountState);
+        const sendState = {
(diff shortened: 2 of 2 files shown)


ryanml (dismissed):

LeHoang88 (dismissed):

ui/ducks/send/send.js
```diff
@@ -1326,6 +1326,7 @@ export function toggleSendMaxMode() {
       await dispatch(actions.updateAmountMode(AMOUNT_MODES.MAX));
       await dispatch(actions.updateAmountToMax());
     }
+    await dispatch(computeEstimatedGasLimit());
```
adonesky1, line 1329: @danjm Do we want to do this for non-token sends?

danjm: Yes, in case custom networks like optimism need it

danjm: We could not do it for non-token sends on mainnet

adonesky1: hmmm yeah maybe... Looking a bit more through the send slice, it looks like in the updateSendAmount action creator we are not recomputing the gas limit for non-token assets when the amount is updated, which as you point out for custom networks (like optimism) may cause an issue?

adonesky1: ^^^ https://github.com/MetaMask/metamask-extension/blob/4e87069af3619db2307e08e1d577621339eab3cc/ui/ducks/send/send.js#L1167

adonesky1: @danjm this isn't a blocker. Just raising it here because you've pointed out that we shouldn't use the asset_type to decide whether or not to recompute gas

adonesky1 (dismissed):
A nonblocking/comment question is still open. Otherwise LGTM

ui/ducks/send/send.js
danjm: right, and that could also matter if the user was sending to a contract address that is something other than a token

adonesky1: Should I write up a ticket for this?

danjm:
@adonesky1 Just pushed a change to address our comments above.

adonesky1 (approved):
LGTM!

Russell4115 (approved):

ryanml (approved):

Russell4115 (approved):

PR #11596, opened 2021-07-23
Remove the SHOW_EIP_1559_UI environment variable, replace with network detection where appropriate

### LINKED ISSUE #11490: EIP-1559 - Audit Usage of SHOW_EIP_1559_UI
At present, we use the `SHOW_EIP_1559_UI` environment variable for two purposes:

1.  Whether we want the EIP-1559 components to be used in the extension (so that we could merge work into the extension without the users seeing it when we have releases before 1559)
2. Whether we want to show pieces of UI or old UI

When we have components wired up, we need to remove `SHOW_EIP_1559_UI` and instead replace that functionality with logic to check if a given transaction or flow is an EIP-1559 flow.

darkwing:
Fixes: https://github.com/MetaMask/metamask-extension/issues/11490

I did the best I could to keep this small, minimal.  We'll need follow up work to remove extraneous props, remove unused components.  The difficult one was `ui/pages/confirm-transaction-base/confirm-transaction-base.component.js`, so have a good look at that one.

### DIFF
--- a/ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
+++ b/ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
@@ -79,6 +79,11 @@ export default class ConfirmTransactionBase extends Component {
     actionKey: PropTypes.string,
     contentComponent: PropTypes.node,
     dataComponent: PropTypes.node,
+    primaryTotalTextOverride: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.node,
+    ]),
+    secondaryTotalTextOverride: PropTypes.string,
     hideData: PropTypes.bool,
     hideSubtitle: PropTypes.bool,
     identiconAddress: PropTypes.string,
@@ -257,6 +262,8 @@ export default class ConfirmTransactionBase extends Component {
 
   renderDetails() {
     const {
+      primaryTotalTextOverride,
+      secondaryTotalTextOverride,
       hexTransactionFee,
       hexTransactionTotal,
       useNonceField,
@@ -515,7 +522,7 @@ export default class ConfirmTransactionBase extends Component {
             />,
             <TransactionDetailItem
               key="total-item"
-              detailTitle={t('total')}
+              detailTitle={primaryTotalTextOverride || t('total')}
               detailText={
                 <UserPreferencedCurrencyDisplay
                   type={PRIMARY}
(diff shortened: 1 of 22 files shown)


darkwing:
Test updates look good , but since I did the logic part, I think others should review.

danjm (approved):
This looks good to me. Let's make sure to get some QA on it before merge.

danjm (changes requested):
hold on, need to re-check something

ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
```diff
@@ -360,179 +305,118 @@ export default class ConfirmTransactionBase extends Component {
       </div>
     ) : null;
```
danjm, line 307: I'm confused about something. A lot of code deleted in this file seems to be necessary for EIP1559. And some that is deleted seems necessary non-EIP1559 support. For example, on develop `inlineGasControls` (either `AdvancedGasControls` or `AdvancedGasControls`) is shown on both EIP1559 networks and non-EIP1559 networks, but here it is removed completely.

Why not just replace the `process.env.SHOW_EIP_1559_UI` checks with a check for whether the current network supports eip1559?

danjm: I think I missed this when I first reviewed because the diff of this file was hidden

darkwing: I was under the impression that all gas editing should go through the popover.  Also, we had spoken about removing the inline controls for now, since we need to restructure the data handling.

danjm: Okay, actually, my assessment above is not right. There are some things removed: the `inlineGasControls` functionality is removed with this PR, and so is the `primaryTotalTextOverride` and `secondaryTotalTextOverride`. We still should support these.

danjm: I can see the difficulty in having inline gas controls here, but I think we need to continue to support their use? There removal would be a degradation of UX on non-eip1559 networks.

adonesky1: @danjm is it a degradation if the functionality is still available but just in the popover? Plus on non-eip1559 networks the gas controls are available on the send stage? So for a non-1559 network the user could either use the gas pop over to edit gas on the confirmation screen or navigate back to the send screen and edit gas from there?

ui/pages/send/send-content/send-gas-row/send-gas-row.component.js
```diff
@@ -40,7 +46,11 @@ export default class SendGasRow extends Component {
             category: 'Transactions',
             event: 'Clicked "Advanced Options"',
           });
-          showCustomizeGasModal();
+          if (isEIP1559Network) {
```
danjm, line 49: I don't think we should show the customization modal in send when on an EIP-1559 network

danjm: This code can't be hit, but it should be removed

darkwing:
Closing in favor of Alex's PR!

PR #11602, opened 2021-07-24
EIP-1559 - Allow wrapping for dapp suggested origin

darkwing:
Dapp origins that are long push the fiat value off screen, so we need to break on those origins.


ryanml (approved):
++

danjm (approved):

PR #11604, opened 2021-07-24
Fixing type issue with auto-filled custom token decimals value

ryanml:
Related Sentry Error: https://sentry.io/organizations/metamask/issues/2455739694/?project=273505&query=is%3Aunresolved

An attempt to auto-fill custom token information occurs when a user pastes in a token address in the custom form.  (https://github.com/MetaMask/metamask-extension/blob/develop/ui/pages/add-token/add-token.component.js#L165)

The errors occurs when we need to grab the info from the contract-metadata, where decimals values are stored as type `number` (https://github.com/MetaMask/metamask-extension/blob/develop/ui/helpers/utils/token-util.js#L72) - t

### DIFF
--- a/ui/helpers/utils/token-util.js
+++ b/ui/helpers/utils/token-util.js
@@ -73,7 +73,7 @@ async function getDecimals(tokenAddress) {
     const contractMetadataInfo = getContractMetadata(tokenAddress);
 
     if (contractMetadataInfo) {
-      decimals = contractMetadataInfo.decimals;
+      decimals = contractMetadataInfo.decimals?.toString();
     }
   }
 


ui/helpers/utils/token-util.js
```diff
@@ -73,7 +73,7 @@ async function getDecimals(tokenAddress) {
     const contractMetadataInfo = getContractMetadata(tokenAddress);
 
     if (contractMetadataInfo) {
-      decimals = contractMetadataInfo.decimals;
+      decimals = contractMetadataInfo.decimals?.toString();
```
ryanml, line 76: The `?` isn't completely necessary, as all of the token objects in contract-metadata have the `decimals` attribute with a numerical value, however this is just an extra precaution should anything land there that isn't correctly formatted. We already handle `undefined` cases in `handleCustomDecimalsChange`

danjm (approved):

NiranjanaBinoy (approved):
LGTM

PR #11608, opened 2021-07-26
Adding hex address guard to input validation for token sends

ryanml:
Related Sentry Error: https://sentry.io/organizations/metamask/issues/2508809557/

After validating the user-entered address for a token send, further checks occur to determine if the recipient is known. `toChecksumAddress` should be guarded with `isValidHexAddress`, as the input may be invalid at that point. This fix correctly results with the "Recipient address is invalid" message in the UI, with no JS errors.

## Test Steps
1. Initiate a token send from the asset list (EG, `Send LINK`)
2. Enter an invalid hex address, such as `300`
3. Confirm the correct error is shown, with no errors in the console

### DIFF
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -858,6 +858,7 @@ const slice = createSlice({
 
         if (
           isSendingToken &&
+          isValidHexAddress(recipient.userInput) &&
           (toChecksumAddress(recipient.userInput) in contractMap ||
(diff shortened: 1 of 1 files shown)


ryanml:
Alternate thought:  We could `return;` after the prior validation finds an invalid address error

danjm (approved):

NiranjanaBinoy (approved):
LGTM

PR #11610, opened 2021-07-26
Fix setting of nickname upon pasting a new address into the send flow

danjm:
In prod, when pasting an address into the send flow recipient field, we don't immediately set the nickname, even if the address has a nickname in the address book state. This incorrect UX is shown here:

https://user-images.githubusercontent.com/7499938/126953994-2df91a6e-f08e-425d-b0b4-d84f89c8c769.mp4

This PR fixes this by ensuring that when the recipient is updated in the send flow, the displayed nickname in the send flow is set to the one already saved in state. The working UX after this fix is applied is shown here:

https://user-images.githubusercontent.com/7499938/126954252-1102f400-754c-424a-849c-8a9583dc6bcd.mp4

### DIFF
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -1261,6 +1261,10 @@ export function useMyAccountsForRecipientSearch() {
  * address results in hex data changing because the recipient address is
  * encoded in the data instead of being in the 'to' field. The to field in a
  * token send will always be the token contract address.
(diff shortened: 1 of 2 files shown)


ryanml (approved):
++

NiranjanaBinoy (approved):
LGTM

PR #11615, opened 2021-07-26
Update addTxGasDefaults and _getDefaultGasFees to work with new estimate types, and ensure we correctly handle gas price estimates when on EIP1559 networks

danjm:
This PR ensures that transactions created by dapps will work correctly, on both EIP1559 compatible networks and on networks without compatibility, and both for txes with dapp suggested fees and txes where the dapp does not suggest the fee. It also ensures that if we are on an EIP1559 network, and our request for gas fee estimates fails and so we have to fall back to a eth_gasPrice call, that those estimates are handled correctly as well.

eth_gasPrice estimates on an EIP-1559 network are to be set to the `maxFeePerGas` and the `maxPriorityFeePerGas`

For testing, it is recommended to use this branch: https://github.com/MetaMask/test-dapp/pull/119

(I will add some demo videos here a bit later)

To do:

- [ ] Remove controllers version bump, only included here for QA purposes
- [ ] Needs a rebase

### DIFF
--- a/ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
+++ b/ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
@@ -116,6 +116,7 @@ export default class ConfirmTransactionBase extends Component {
     isEthGasPrice: PropTypes.bool,
     noGasPrice: PropTypes.bool,
     setDefaultHomeActiveTabName: PropTypes.func,
+    supportsEIP1599: PropTypes.bool,
   };
 
   state = {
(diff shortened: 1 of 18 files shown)


darkwing (approved):
This is excellent Dan!  Thank you so much!

ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
```diff
@@ -449,11 +451,13 @@ export default class ConfirmTransactionBase extends Component {
                     />,
                   ])}
                   subTitle={
-                    <GasTiming
```
danjm, line 452: This actually doesn't need to happen in this PR and will be removed when I correctly rebase. But It is okay to have it here for now, so that the error it throws in not a distraction for QA. It has been fixed properly on another PR that will soon be merged

- [ ] Remove change after rebasing onto develop

PeterYinusa:
LGTM!

PR #11616, opened 2021-07-26
Make EIP1559 compatibility in tx controller also require account type support

danjm:
This PR ensures that the checks for EIP1559 in the transaction controller account for whether the account sending the transaction is EIP1559 compatible.


darkwing (approved):

app/scripts/controllers/transactions/index.js
```diff
@@ -739,7 +750,7 @@ export default class TransactionController extends EventEmitter {
     };
     // sign tx
     const fromAddress = txParams.from;
-    const common = await this.getCommonConfiguration();
+    const common = await this.getCommonConfiguration(txParams.from);
```
ryanml, line 753: nit: use `fromAddress` here?

darkwing: `txParams.from` is used everywhere I believe

ryanml (approved):

PR #11620, opened 2021-07-26
Ensure that gas fee/prices params are not updated while in send edit stage/mode

danjm:
This PR fixes the send edit mode/stage. By "send edit mode" I mean: from the confirm screen, click the "< Edit" link in the top left of  the confirm screen, to go to the send screen.

This PR ensures that submission of the send screen form while editing does not update gasPrice/maxPriorityFee/maxFee. These values can change during polling while in the send flow. We may want to make changes to ensure that polling doesn't happen in the future, but for now at least this change is definitely needed, as editing should not change any values that cannot be seen+editing while on the send screen in edit mode.

### DIFF
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -1516,10 +1516,18 @@ export function signTransaction() {
       // merge in the modified txParams. Once the transaction has been modified
       // we can send that to the background to update the transaction in state.
       const unapprovedTxs = getUnapprovedTxs(state);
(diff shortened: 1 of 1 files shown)


darkwing:
This makes good sense, just need to address test failure

dan437 (dismissed):
Looks good!

darkwing (dismissed):

dan437 (approved):

adonesky1 (approved):
LGTM!

PR #11621, opened 2021-07-26
Only show dapp suggested gas fee warning if user has not edited the fee

danjm:
This PR proposes a UX modification to the dapp-ackknowledgement-warning in `edit-gas-display.component.js`

Currently, if a dapp creates a transaction with proposed gas prices included, and the user goes to edit the gas fees, a warning is shown which says "This gas fee has been suggested by {dapp origin}. Overriding this may cause a problem with your transaction. Please reach out to $1 if you have questions." The user has to click a confirmation before they can edit.

I think that if they already have edited the fees provided by the dapp, then we should not show the warning. This PR add this update.


darkwing (approved):
I noticed this as well!  Good catch, nice solution.

dan437 (approved):

PR #11634, opened 2021-07-27
Fixing confirmation layout when dapp has suggested a price (EIP-1559)

### LINKED ISSUE #11612: EIP-1559: spacing/layout issues when we have a dapp suggested gas price
To recreate:

(1) Run metamask with the EIP1559 flag turned on
(2) Go to https://metamask.github.io/test-dapp/ and connect
(3) Click the "Send" button

You will see this:

https://user-images.githubusercontent.com/7499938/126984588-b1f14568-f2a9-4495-870b-d89a3f11449b.mp4

ryanml:
Fixes: https://github.com/MetaMask/metamask-extension/issues/11612

- Only displaying dapp hostname as opposed to protocol + hostname
- Adjusting spacing in detail rows

Screen Shot 2021-07-27 at 7 31 45 AM
Screen Shot 2021-07-27 at 7 30 31 AM

### DIFF
--- a/ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
+++ b/ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
@@ -301,6 +301,14 @@ export default class ConfirmTransactionBase extends Component {
     } = this.props;
     const { t } = this.context;
 
+    const getRequestingOrigin = () => {
+      try {
+        return new URL(txData.origin)?.hostname;
+      } catch (err) {
+        return '';
+      }
+    };
+
     const notMainnetOrTest = !(isMainnet || process.env.IN_TEST);
     const gasPriceFetchFailure = isEthGasPrice || noGasPrice;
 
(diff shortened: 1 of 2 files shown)


adonesky1 (dismissed):
LGTM

NiranjanaBinoy (dismissed):
LGTM

ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
```diff
@@ -303,6 +303,7 @@ export default class ConfirmTransactionBase extends Component {
 
     const notMainnetOrTest = !(isMainnet || process.env.IN_TEST);
     const gasPriceFetchFailure = isEthGasPrice || noGasPrice;
+    const requestingOrigin = new URL(txData.origin)?.hostname;
```
danjm, line 306: `txData` won't always have an origin. I think this is why e2e tests are failing

ryanml: Good point, I've made this more fault tolerant 👍

kumavis (approved):
approving on the condition that `txData.origin` should never be empty while `dappSuggestedGasFees` is present

the empty string is fine as a safer approach to the mostly impossible case, but its like to generate a bad localization string in some languages

darkwing (approved):

PR #11635, opened 2021-07-27
EIP-1559 and Rinkeby Testnet support in Swaps

dan437:
## Explanation
This PR:
- Adds support for EIP-1559 in Swaps
- Uses new components for editing gas in Swaps
- Adds support for a new network in Swaps (Rinkeby), because it already has EIP-1559 support

## Manual testing steps
EIP-1559 happy path:
- Switch to the `Rinkeby` Test Network
- Get test ETH
- Click on "Swap" on the main page
- Make a swap
- Open the transaction in a block explorer (e.g. https://rinkeby.etherscan.io/) and check that Txn Type is: `2 (EIP-1559)` 

Non EIP-1559 happy path:
- Use e.g. the Ethereum Mainnet Network
- Click on "Swap" on the main page
- Make a swap
- Open the transaction in a block explorer (e.g. https://etherscan.io/) and check that Txn Type is not: `2 (EIP-1559)` 

## Screenshots
`View Quote` page on Rinkeby, which has EIP-1559 support:
image

Success page for an EIP-1559 transaction:
image

`Customize Gas` modal window for EIP-1559, Basic Options:
image

`Customize Gas` modal window for EIP-1559, Advanced Options:
image

EIP-1559 Txn Type on rinkeby.etherscan.io:
image

EIP-1559 transaction in the `Activity` tab on the main page:
image

`View Quote` on Ethereum testnet, without EIP-1559 support:
image

Success page for a non-EIP 1559 transaction:
image

Non-EIP-1559 transaction in the `Activity` tab on the main page:
image


ui/store/actions.js
```diff
@@ -2199,6 +2199,22 @@ export function setSwapsTxGasLimit(gasLimit) {
...
 }
 
+export function updateCustomSwapsEIP1559GasParams({
+  gasLimit,
+  maxFeePerGas,
+  maxPriorityFeePerGas,
+}) {
+  return async (dispatch) => {
+    // TODO:Why do we pass true as the 2nd param? It seems the fn only supports gasLimit.
+    await promisifiedBackground.setSwapsTxGasLimit(gasLimit, true);
```
dan437, line 2209: I can change this to: `setSwapsTxGasLimit(gasLimit)` everywhere

app/scripts/controllers/swaps.js
```diff
@@ -514,7 +535,38 @@ export default class SwapsController {
...
+      } = gasFeeEstimates;
+
+      usedGasPrice = addCurrencies(
+        customMaxPriorityFeePerGas || // Is already in hex WEI.
+          decGWEIToHexWEI(suggestedMaxPriorityFeePerGas),
+        decGWEIToHexWEI(estimatedBaseFee),
+        {
+          aBase: 16,
+          bBase: 16,
+          toNumericBase: 'hex',
```
dan437, line 558: Just to double check - are these params correct?

ui/components/app/transaction-detail-item/transaction-detail-item.component.js
```diff
@@ -22,14 +22,14 @@ export default function TransactionDetailItem({
         <Typography
           color={detailTitleColor}
           fontWeight={FONT_WEIGHT.BOLD}
-          variant={TYPOGRAPHY.H6}
+          variant={TYPOGRAPHY.H7}
```
dan437, line 25: I've changed this because according to our design we should use H7. Benefit is that now text fits on one line: 
image

ui/pages/swaps/index.js
```diff
@@ -111,8 +112,11 @@ export default function Swap() {
   const chainId = useSelector(getCurrentChainId);
   const isSwapsChain = useSelector(getIsSwapsChain);
   const useNewSwapsApi = useSelector(getUseNewSwapsApi);
+  const EIP1559NetworkEnabled = useSelector(isEIP1559Network);
   const fromToken = useSelector(getFromToken);
 
+  useGasFeeEstimates(); // This will pre-load gas fees before going to the View Quote page.
```
dan437, line 118: Double check this.

ui/pages/swaps/swaps.util.js
```diff
@@ -93,6 +96,9 @@ const getBaseApi = function (
       return `${baseUrl}/aggregatorMetadata`;
     case 'gasPrices':
       return `${baseUrl}/gasPrices`;
+    case 'suggestedGasFees':
+      // We will probably use the hook "useGasFeeEstimates" instead of this.
+      return `${GAS_API_BASE_URL}/networks/${chainIdDecimal}/suggestedGasFees`;
```
dan437, line 101: I can probably remove this since we use `useGasFeeEstimates` in Swaps, which calls this endpoint.

ui/pages/swaps/view-quote/view-quote.js
```diff
@@ -134,6 +150,8 @@ export default function ViewQuote() {
   const chainId = useSelector(getCurrentChainId);
   const nativeCurrencySymbol = useSelector(getNativeCurrency);
 
+  const gasFeeInputs = useGasFeeInputs('high');
```
dan437, line 153: For Swaps we want to get 'high' estimations by default.

app/scripts/controllers/swaps.js
```diff
@@ -514,7 +535,38 @@ export default class SwapsController {
...
+          numberOfDecimals: 6,
+        },
+      );
+    } else if (gasEstimateType === GAS_ESTIMATE_TYPES.LEGACY) {
+      usedGasPrice = customGasPrice || decGWEIToHexWEI(gasFeeEstimates.high);
+    } else if (gasEstimateType === GAS_ESTIMATE_TYPES.ETH_GASPRICE) {
+      usedGasPrice =
+        customGasPrice || decGWEIToHexWEI(gasFeeEstimates.gasPrice);
+    } else {
+      usedGasPrice = '0x0';
```
darkwing, line 568: If you initialize `usedGasPrice` to `0x0`, you can avoid the `else` block; `0x0` would act as the default value.

ui/pages/swaps/view-quote/view-quote.js
darkwing: Yep!

darkwing:
One thing that I'm seeing that looks weird is that the timing function shows "60" seconds for high, and 45 for medium and low oO.  Weird.

https://user-images.githubusercontent.com/46655/127207237-aca79b1c-474d-4fae-95d5-2a644cf84536.mp4

hendis85 (approved):

dan437:
> One thing that I'm seeing that looks weird is that the timing function shows "60" seconds for high, and 45 for medium and low oO. Weird.
> WeirdTiming.mp4

It seems that for Low and Medium it says: `Likely in < 45 seconds` and for High: `Very likely in < 60 seconds`. Agree that it's a bit confusing for users and so we might want to update the `<GasTiming>` component in another PR.

ui/store/actions.js
```diff
@@ -2199,6 +2199,21 @@ export function setSwapsTxGasLimit(gasLimit) {
...
+  gasLimit,
+  maxFeePerGas,
+  maxPriorityFeePerGas,
+}) {
+  return async (dispatch) => {
+    await promisifiedBackground.setSwapsTxGasLimit(gasLimit);
+    await promisifiedBackground.setSwapsTxMaxFeePerGas(maxFeePerGas);
+    await promisifiedBackground.setSwapsTxMaxFeePriorityPerGas(
+      maxPriorityFeePerGas,
+    );
```
darkwing, line 2212: Doesn't need to block merge but can this be a `Promise.all`, just so we aren't successively awaiting?

darkwing:
This is awesome!  
Successful

ui/store/actions.js
dan437: That sounds promising!

ui/components/app/transaction-detail-item/transaction-detail-item.component.js
```diff
@@ -39,20 +39,24 @@ export default function TransactionDetailItem({
...
       </div>
       <div className="transaction-detail-item__row">
-        <Typography
-          variant={TYPOGRAPHY.H7}
-          className="transaction-detail-item__subtitle"
-          color={COLORS.UI4}
-        >
-          {subTitle}
-        </Typography>
+        {React.isValidElement(subTitle) ? (
```
darkwing, line 49: I'm very much looking forward to this change! :)

app/scripts/controllers/transactions/index.js
```diff
@@ -1200,10 +1200,12 @@ export default class TransactionController extends EventEmitter {
           txMeta.chainId,
         );
 
-        const quoteVsExecutionRatio = `${new BigNumber(tokensReceived, 10)
-          .div(txMeta.swapMetaData.token_to_amount, 10)
-          .times(100)
-          .round(2)}%`;
```
danjm, line 1206: Why is this change needed?

shared/constants/swaps.js
```diff
@@ -54,6 +55,14 @@ export const TEST_ETH_SWAPS_TOKEN_OBJECT = {
   iconUrl: TEST_ETH_TOKEN_IMAGE_URL,
 };
 
+export const RINKEBY_SWAPS_TOKEN_OBJECT = {
```
danjm, line 58: I guess we should include this for QA purposes, but we should remove it before shipping to prod, because swaps does not actually work on rinkeby

ui/ducks/swaps/swaps.js
```diff
@@ -682,7 +715,10 @@ export const signAndSendTransactions = (history, metaMetricsEvent) => {
       .plus(usedQuote.approvalNeeded?.gas || '0x0', 16)
       .toString(16);
     const gasEstimateTotalInUSD = getValueFromWeiHex({
-      value: calcGasTotal(totalGasLimitEstimate, usedGasPrice),
+      value: calcGasTotal(
+        totalGasLimitEstimate,
+        EIP1559Network ? maxFeePerGas : usedGasPrice,
```
danjm, line 720: I think this should be `baseFeePerGas + maxPriorityFeePerGas`

danjm: `usedGasPrice` is use to calculate the "estimate", which is what `baseFeePerGas + maxPriorityFeePerGas` would be used for

ui/pages/swaps/view-quote/view-quote.js
```diff
@@ -607,6 +665,24 @@ export default function ViewQuote() {
...
           />
         )}
+
+        {showEditGasPopover && EIP1559Network && (
+          <EditGasPopover
+            transaction={{
+              txParams: {
+                maxFeePerGas,
+                maxPriorityFeePerGas,
+                gasLimit: maxGasLimit,
```
danjm, line 675: Let's change this to `gas: maxGasLimit`

ui/hooks/useGasFeeInputs.js
```diff
@@ -200,8 +212,8 @@ export function useGasFeeInputs(defaultEstimateToUse = 'medium', transaction) {
       : null,
   );
   const [gasLimit, setGasLimit] = useState(
-    transaction?.txParams?.gas
-      ? Number(hexToDecimal(transaction.txParams.gas))
```
danjm, line 204: Hmm... this won't work for standard transactions in metamask. `gasLimit` is normally called `gas`

Instead could you call the param passed in from "View-quote" `gas`? (Then I think this change won't be necessary

danjm: Also, we need to keep that `Number` conversion present to handle transactions in the standard metamask send flow

Could you convert the value to hex in `View-quote`? You can use a function called `hexToDecimal` for that

danjm: I realize I might have been the original author of these changes :sweat_smile:

ui/pages/swaps/index.js
```diff
@@ -111,8 +112,15 @@ export default function Swap() {
   const chainId = useSelector(getCurrentChainId);
   const isSwapsChain = useSelector(getIsSwapsChain);
   const useNewSwapsApi = useSelector(getUseNewSwapsApi);
+  const EIP1559Network = useSelector(isEIP1559Network);
   const fromToken = useSelector(getFromToken);
 
+  if (EIP1559Network) {
```
danjm, line 118: Do we have to have this `if` block and eslint disable? Will everything still work if it is removed?

danjm (changes requested):
Identified a few issues that need to be addressed

shared/constants/swaps.js
dan437: What if we just disable swapping for Rinkeby after we are done with QA, at least until our EIP 1559 changes are in production and working? It could be done by removing this line: 
https://github.com/MetaMask/metamask-extension/pull/11635/files#diff-9f9eff0a280f38efd9f071f24d503f51a1f7617a30f67e7e654e157dfc060902R98

Once we verify that EIP 1559 changes work well in production, I can send a quick PR for removing Rinkeby from Swaps.

app/scripts/controllers/transactions/index.js
dan437: Great question! For some reason I was getting empty `txMeta.txReceipt.logs` on Rinkeby: 
image

When that happens, the `getSwapsTokensReceivedFromTxMeta` fn call returns null (https://github.com/MetaMask/metamask-extension/blob/develop/ui/pages/swaps/swaps.util.js#L707). Any idea why there are on logs on Rinkeby?

ui/hooks/useGasFeeInputs.js
dan437: Will do the suggested changes.

ui/pages/swaps/index.js
dan437: Originally I didn't have this `if` block, but then I realized we don't need to make unnecessary API calls for non EIP-1559 networks in Swaps via this hook. For non EIP-1559 networks we get gas prices like this: https://github.com/MetaMask/metamask-extension/pull/11635/files#diff-f89a0460c9cb04858f4de957f65d6aab869c1d766f11781b1fe8a61f3321620eR199

ui/hooks/useGasFeeInputs.js
dan437: Done. Btw, on the View Quote page `maxGasLimit` is already in hex, so I can just do this:
image

Edit Gas modal after these changes:
image

shared/constants/swaps.js
danjm: yeah, that makes sense

danjm (approved):

darkwing (approved):
This looks excellent!  The *only* thing that I noticed is that when I click "Edit", the "high" radio isn't selected.  I don't want to block on that though -- functionally working great!

dan437:
> This looks excellent! The _only_ thing that I noticed is that when I click "Edit", the "high" radio isn't selected. I don't want to block on that though -- functionally working great!

Thank you! It seems to be working now, I can see the "High" radio button selected after clicking on "Edit":
image

PeterYinusa:
LGTM!

PR #11652, opened 2021-07-28
Allow higher precision gas prices in the send flow

danjm:
On the palm network, gas prices can be lower than 0.0001, so we need to allow the gas price estimates to have a higher precision.


ryanml (dismissed):

ryanml:
note: check unit test failures

darkwing:
Will approve once test is good to go

ryanml (approved):

darkwing (approved):

Issue #11647, opened 2021-07-28
Metamask Extension: Incorrect gas limit estimates for ETH sends to Contract addresses


yashutanna:
**Describe the bug**
When initiating a transfer transaction to a contract address, the mobile app estimates a gas limit of 64963 while the extension estimates a limit of 34261. This require gas for this type of transaction is 40828

**Steps to reproduce (REQUIRED)**

1. Go to Metamask mobile app (v2.5.0 (750))
2. send ETH to 0xcBC243f991dFA5025511dFdB40225E1D9C4B6F9e and click next
3. set any amount of ETH (we are not going to actually send) and click next
4. click on the gas fee and then navigate to the advanced tab
5. see gas limit is set to 64963
2. follow the same flow on the meta mask extension (v.9.8.2) and see the gas limit is 34261

**Expected behavior**
gas limit should be estimated at more than what would be minimally required

**Screenshots**
![image](https://user-images.githubusercontent.com/6696871/127330335-29ffce64-3d74-4ce1-a33a-71318bdd707f.png)
![image](https://user-images.githubusercontent.com/6696871/127330358-1756d5c5-7198-4268-9971-c5d4e980f96c.png)

**Browser details (please complete the following information):**
 - OS: OSX
 - Browser: Brave Chrome Version 91.0.4472.164 (64-bit)
 - MetaMask Extension Version: 9.8.2
 - MetaMask mobile app (IOS)(v2.5.0 (750))

yashutanna:
This seems to be fixed with 9.8.4 👍

PR #11658, opened 2021-07-28
Use current block gas limit as the limit passed eth_estimateGas

danjm:
Fixes: https://github.com/MetaMask/metamask-extension/issues/11647

We stopped using `state.metamask.currentBlockGasLimit` with this change: https://github.com/MetaMask/metamask-extension/pull/10965/files#diff-4090152fe396753053c3582bfd4e1e51e7aea1c1e1833e4e9993cd0b7709f7c4L54 and instead started using `state.metamask.blockGasLimit`, which can be undefined.

This PR returns us to using the property from state that was used before that change.


ryanml (approved):
lgtm!

darkwing (approved):

PR #11659, opened 2021-07-28
EIP-1559 - Only show radio group and gas timing when network supports 1559

### LINKED ISSUE #11649: EIP-1559 - Edit gas popover estimate and priority on custom networks

1. Login to the extension
2. Click the Network drop down, Select `Binance Smart Chain Testnet`
3. Click the Account identicon, then select Settings and then Advanced
4. Toggle `Show Conversion on Testnets` to ON, then close the menu
2. Click the Send button on the Home page
3. Select a recipient on the Add Recipient page
4. Click the Max amount button on the Send page, then click Next
5. Click the Edit link in the Estimate gas fee section

**Expected behavior**
- The priority radio options are not displayed
- An estimate of the native currency is shown
Edit gas 2 - BSC

**Screenshots**
Current behavior
Edit gas - BSC

darkwing:
Fixes: #11649

The GasTiming fix is required or the component will error out on networks without high/medium/low

### DIFF
--- a/ui/components/app/gas-timing/gas-timing.component.js
+++ b/ui/components/app/gas-timing/gas-timing.component.js
@@ -32,7 +32,6 @@ export default function GasTiming({ maxPriorityFeePerGas }) {
   };
 
   // Don't show anything if we don't have enough information
-  // or if the network doesn't support 1559
   if (
     isGasEstimatesLoading ||
     gasEstimateType !== GAS_ESTIMATE_TYPES.FEE_MARKET
--- a/ui/components/app/gas-timing/gas-timing.component.js
+++ b/ui/components/app/gas-timing/gas-timing.component.js
@@ -1,7 +1,9 @@
 import React, { useContext } from 'react';
+import { useSelector } from 'react-redux';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
+import { isEIP1559Network } from '../../../ducks/metamask/metamask';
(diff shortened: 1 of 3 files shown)


darkwing:
I added another small fix to prevent merge conflicts -- hides the education button if the network doesn't support 1559.

danjm (dismissed):

ui/components/app/gas-timing/gas-timing.component.js
```diff
@@ -26,7 +30,8 @@ export default function GasTiming({ maxPriorityFeePerGas }) {
   };
 
   // Don't show anything if we don't have enough information
-  if (isGasEstimatesLoading) {
+  // or if the network doesn't support 1559
```
danjm, line 33: Just realized that there is a better / more robust check that we can do here. Instead of checking whether the network supports EIP-1559, we should check if `gasEstimateType === GAS_ESTIMATE_TYPES.FEE_MARKET`

`gasEstimateType` is available from `useGasFeeEstimates()`, which is already called above, and `GAS_ESTIMATE_TYPES` will have to be imported from the shared constants.

The reason we should do this check, instead of just the network check, is that if for some reason the request for fee market type estimates failled, and the gas controller fell back to just requesting the gas price, then there will be no `low, medium and high` available on `gasFeeEstimates` and the below code will break

danjm (approved):
Looks good!

adonesky1 (approved):
LGTM!

PR #11669, opened 2021-07-29
Fix to having radio buttons in advanced gas inputs pre-selected

danjm:
The radio buttons in advanced gas inputs, when on an eip-1559 network, where not being automatically selected because we were matching the set gas fees against the estimated gas fees, but these were of different types. This is because of a `Number()` type cast done elsewhere in `useGasInputs`. This PR ensures that the estimated and set values are of the same type when compared


darkwing (approved):

NiranjanaBinoy (approved):
LGTM

PR #11698, opened 2021-07-30
reduce cryptography dependencies bundled in ui

kumavis:
an attempt to reduce the amount of crypto dependencies in ui for faster boot times
secp256k1 was particularly slow during init
still requires keccak256 for checksum addresses
this code was copy pasted from ethereumjs-util and slightly modified to appease the linter
idk if this is a good idea, we should likely instead migrate to utils that dont mix crypto and hex utils

also, i noticed we use toChecksumAddress in the render function in a few places. this should be memoized or moved out. shouldnt be hashing in the render functions!


kumavis:
going to abandon this in the face of manifest v3 roadmap

PR #11710, opened 2021-07-31
Fix 11705 - Reset gas limit when radio button clicked

darkwing:
* Resets the gas limit from custom when the user clicks a radio button
* Adds gas limit as a comparison for when finding a matching estimate

### DIFF
--- a/ui/hooks/useGasFeeInputs.js
+++ b/ui/hooks/useGasFeeInputs.js
@@ -429,13 +429,19 @@ export function useGasFeeInputs(
     (estimateLevel) => {
       setInternalEstimateToUse(estimateLevel);
       if (gasErrors.gasLimit === GAS_FORM_ERRORS.GAS_LIMIT_OUT_OF_BOUNDS) {
-        setGasLimit(hexToDecimal(transaction?.txParams?.gas));
+        const transactionGasLimit = hexToDecimal(transaction?.txParams?.gas);
+        const minimumGasLimitDec = hexToDecimal(minimumGasLimit);
+        setGasLimit(
+          transactionGasLimit > minimumGasLimitDec
+            ? transactionGasLimit
+            : minimumGasLimitDec,
+        );
       }
       setMaxFeePerGas(null);
       setMaxPriorityFeePerGas(null);
       setGasPrice(null);
     },
-    [gasErrors.gasLimit, transaction],
+    [minimumGasLimit, gasErrors.gasLimit, transaction],
   );
 
   return {
--- a/ui/hooks/useGasFeeInputs.js
+++ b/ui/hooks/useGasFeeInputs.js
@@ -103,6 +103,8 @@ function getMatchingEstimateFromGasFees(
   maxFeePerGas,
   maxPriorityFeePerGas,
   gasPrice,
+  gasLimit,
+  minimumGasLimit,
   supportsEIP1559,
 ) {
   return (
@@ -111,7 +113,8 @@ function getMatchingEstimateFromGasFees(
         return (
           Number(estimate?.suggestedMaxPriorityFeePerGas) ===
             Number(maxPriorityFeePerGas) &&
-          Number(estimate?.suggestedMaxFeePerGas) === Number(maxFeePerGas)
+          Number(estimate?.suggestedMaxFeePerGas) === Number(maxFeePerGas) && 
+          Number(gasLimit) === Number(minimumGasLimit)
         );
       }
       return estimate?.gasPrice === gasPrice;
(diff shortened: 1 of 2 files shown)


ui/hooks/useGasFeeInputs.js
```diff
@@ -221,17 +224,22 @@ export function useGasFeeInputs(
       : null,
   );
   const [gasLimit, setGasLimit] = useState(
-    transaction?.txParams?.gas
-      ? Number(hexToDecimal(transaction.txParams.gas))
-      : 21000,
```
danjm, line 226: This change is good.

ui/hooks/useGasFeeInputs.js
```diff
@@ -221,17 +224,22 @@ export function useGasFeeInputs(
...
+    ),
   );
+
   const [estimateToUse, setInternalEstimateToUse] = useState(
     transaction
       ? getMatchingEstimateFromGasFees(
           gasFeeEstimates,
           maxFeePerGas,
           maxPriorityFeePerGas,
           gasPrice,
```
danjm, line 240: I think we should only base the match on the maxFeePerGas, maxPriorityFeePerGas and gasPrice. gasLimit is independent of the estimates, and we should show a match with low, medium or high even if the user has changes the gas limit from what it was originally.

Also, keep in mind, when the user clicks the different low, medium and high settings, the gas limit will not change. Gas limit will be the same across all possible gas pricess / fee per gas.

danjm:
I think that we should only change the gas limit on selecting the radio buttons if the gas limit is in error (i.e. below the minimum).

I don't think the radio buttons should affect the gas limit otherwise. This is because gas limit is not part of our recommendations we only recommend gas price / max fee and max priority fee. Gas limit is independent of all these values and its estimate depends on the computation cost of the transaction, and not the ethereum network conditions.

So if the gas limit is initially set to 29050, and then the user edits it to 32000, and then clicks "High", I believe the gas limit should stay at 32000.

darkwing:
Should we close Peter's issue then?  He was expecting gas price to get set back to 21000 when a radio button is clicked.  We can then also just close this.

ui/hooks/useGasFeeInputs.js
```diff
@@ -240,12 +248,16 @@ export function useGasFeeInputs(
...
   // values at the level specified.
-  const setEstimateToUse = useCallback((estimateLevel) => {
-    setInternalEstimateToUse(estimateLevel);
-    setMaxFeePerGas(null);
-    setMaxPriorityFeePerGas(null);
-    setGasPrice(null);
-  }, []);
+  const setEstimateToUse = useCallback(
+    (estimateLevel) => {
+      setInternalEstimateToUse(estimateLevel);
```
danjm, line 253: As per my other comments, we probably should not `setMaxFeePerGas(null);` in this function, unless there is a "gasLimit" too low error.

In that case, I think we should set the gasLimit to what is in the `txParams` instead of the minimum. `txParams.gas` should be considered the estimate of the gas. If there is no gas limit in the txParams, it can be set to the minimum.

danjm:
> Should we close Peter's issue then? He was expecting gas price to get set back to 21000 when a radio button is clicked. We can then also just close this.

Well, I do think we should reset the gas limit in case where there is an error because the gas limit is too low (as is the case in the video in Peter's ticket). So I think that we just need to modify this a bit.

The UX we should have is:
- the radio buttons being auto selected should only be based on comparisons to gas price / max fee / max priority fee
- if the gas limit is too low, then when one of the radio buttons are clicked, we should reset the gas limit to `txParams.gas`
- if the gas limit is not too low, then it should be unchanged when the radio buttons are clicked.

ui/hooks/useGasFeeInputs.js
darkwing: I'm ... having a hard time understanding what needs to be done here oO

danjm: something like
```
if (gasErrors.gasLimit === GAS_FORM_ERRORS.GAS_LIMIT_OUT_OF_BOUNDS) {
    setGasLimit(hexToDecimal(tx.txParams.gas));
}
```

adonesky1: Ok @danjm I think the confusion might have stemmed from you first mentioning `setMaxFeePerGas(null)` when I think you meant to refer to `setGasLimit(hexToDecimal(minimumGasLimit));`? So this change makes sense to me, but wondering if we want to fall back to the minimum here if the gasLimit on the transaction is itself too low?

danjm: @adonesky1 That sounds like a good idea. With that change, I think this PR is good to go.

danjm (approved):

adonesky1 (approved):

PR #11711, opened 2021-07-31
Fix #11703 - Provide wiring for approval modal

### LINKED ISSUE #11703: EIP-1559 Custom edits made when on token "Approve" confirmation don't have any effect
Repro steps:
(1) Be on rinkeby
(2) go to https://metamask.github.io/test-dapp/
(3) Create a custom token
(4) Approve the token
(5) On the approve confirmation screen, edit the max fee per gas and max priority fee per gas
(6) Submit the transaction
(7) A max priority fee per gas that is different from the one you edited will appear in the transaction receipt / etherscan

Somehow our custom edit while in the approve flow does not actually update the underlying txParam data.

darkwing:
Fixes: #11703

Removes the old `modal` call with a hook modeled after cancel and retry.  With this PR, approval gas modification functions properly.

### DIFF
--- a/ui/pages/confirm-approve/confirm-approve.js
+++ b/ui/pages/confirm-approve/confirm-approve.js
@@ -2,6 +2,7 @@ import React, { useEffect, useRef, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { useParams } from 'react-router-dom';
 import ConfirmTransactionBase from '../confirm-transaction-base';
+import { EDIT_GAS_MODES } from '../../../shared/constants/gas';
 import {
   showModal,
   updateCustomNonce,
@@ -14,11 +15,7 @@ import {
   getTokenValueParam,
 } from '../../helpers/utils/token-util';
 import { useTokenTracker } from '../../hooks/useTokenTracker';
-import {
-  getTokens,
-  getNativeCurrency,
-  isEIP1559Network,
-} from '../../ducks/metamask/metamask';
+import { getTokens, getNativeCurrency } from '../../ducks/metamask/metamask';
 import {
   transactionFeeSelector,
   txDataSelector,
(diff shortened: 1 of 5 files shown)


darkwing:
I think this may need an `insufficients funds` check and notification somewhere...

ui/pages/confirm-approve/confirm-approve.js
```diff
@@ -83,6 +79,12 @@ export default function ConfirmApprove() {
 
   const previousTokenAmount = useRef(tokenAmount);
 
+  const {
+    approveTransaction,
+    showCustomizeGasPopover,
+    closeCustomizeGasPopover,
+  } = useApproveTransaction(transaction);
```
danjm, line 86: It looks like the `transaction` param passed to `useApproveTransaction` is not used (within `useApproveTransaction`)

danjm:
> I think this may need an insufficients funds check and notification somewhere...

Good news: `insufficient funds` warning is handled by the insufficient funds logic within `useGasFeeInputs`, so no further changes are needed in that regards. I took the below screen shot when creating an approval transaction on this branch

Screenshot from 2021-07-31 14-30-25

danjm (approved):

dan437:
After editing max priority fee when approving a token I can see correct numbers: 
image

I've noticed that by default max priority fee and max fee is set to 20 GWEI, but when I edit it and click on "High", it sets to 3 GWEI. It seems that the default 20 GWEI is coming from us (metamask.github.io). 

Maybe in the future we can have some warning that the suggested fees are too high and users should edit it, so they wouldn't need to pay more than they have to. The point is: we should make sure that whichever 3rd party is suggesting fees is not setting up some crazy high prices which are not necessary.
image
image

dan437 (approved):

PR #11713, opened 2021-07-31
EIP-1559 - Only show advanced form toggle when radio buttons are present

darkwing:
I noticed this on mainnet today:

AdvancedOptionsIssuepng

There's no need to show the advanced toggle if the radio group isn't displaying -- doing so allows the user to toggle to essentially seeing only the total, which is pointless.


danjm (approved):
LGTM

ryanml (approved):

PR #11718, opened 2021-08-02
Fall back to aggregator name when icon is unavailable (Swaps Load View)

ryanml:
Related Sentry Error: https://sentry.io/organizations/metamask/issues/2527650528

It is possible for an aggregator's color/icon to be unavailable in the aggregator metadata (https://github.com/MetaMask/metamask-extension/blob/develop/ui/pages/swaps/loading-swaps-quotes/loading-swaps-quotes.js#L224) - in that case, this fix will display the aggregator's name in lieu of the logo. If the name not available either, render nothing. 

Screen Shot 2021-08-01 at 10 58 23 PM

### DIFF
--- a/ui/pages/swaps/loading-swaps-quotes/aggregator-logo.js
+++ b/ui/pages/swaps/loading-swaps-quotes/aggregator-logo.js
@@ -10,22 +10,31 @@ function hexToRGB(hex, alpha) {
   return `rgba(${r}, ${g}, ${b}, ${alpha})`;
 }
 
-export default function AggregatorLogo({ icon, color }) {
+export default function AggregatorLogo({ name, icon, color }) {
   return (
     <div className="loading-swaps-quotes__logo">
-      <div
-        style={{
-          background: color,
-          boxShadow: `0px 4px 20px ${hexToRGB(color, 0.25)}`,
-        }}
-      >
-        <img src={icon} alt="" />
-      </div>
+      {icon && color ? (
+        <div
+          style={{
+            background: color,
+            boxShadow: `0px 4px 20px ${hexToRGB(color, 0.25)}`,
+          }}
+        >
+          <img src={icon} alt="" />
+        </div>
+      ) : (
+        name && (
+          <div>
+            <span>{name}</span>
+          </div>
+        )
+      )}
     </div>
   );
 }
 
 AggregatorLogo.propTypes = {
-  icon: PropTypes.string.isRequired,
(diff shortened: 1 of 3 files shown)


ui/pages/swaps/loading-swaps-quotes/loading-swaps-quotes.js
```diff
@@ -220,7 +220,7 @@ export default function LoadingSwapsQuotes({
                 key={`aggregator-logo-${aggName}`}
               >
                 <AggregatorLogo
-                  aggregatorName={aggName}
```
ryanml, line 223: This prop was not being made use of

ui/pages/swaps/loading-swaps-quotes/aggregator-logo.js
```diff
@@ -10,22 +10,31 @@ function hexToRGB(hex, alpha) {
...
+          </div>
+        )
+      )}
     </div>
   );
 }
 
 AggregatorLogo.propTypes = {
+  name: PropTypes.string.isRequired,
   icon: PropTypes.string.isRequired,
```
Gudahtt, line 38: Perhaps we should mark these props as not being required? Since we seem to expect them to be missing in production some of the time at least.

ui/pages/swaps/loading-swaps-quotes/aggregator-logo.js
```diff
@@ -10,22 +10,31 @@ function hexToRGB(hex, alpha) {
...
+        <div
+          style={{
+            background: color,
+            boxShadow: `0px 4px 20px ${hexToRGB(color, 0.25)}`,
+          }}
+        >
+          <img src={icon} alt="" />
+        </div>
+      ) : (
+        name && (
```
Gudahtt, line 26: Is it possible for the name to not be specified either? The prop type implies otherwise. We should make the prop optional if it's sometimes missing.

ui/pages/swaps/loading-swaps-quotes/aggregator-logo.js
ryanml: sgtm

ui/pages/swaps/loading-swaps-quotes/aggregator-logo.js
ryanml: made this optional 👍

Gudahtt (approved):
LGTM!

darkwing (approved):

PR #11722, opened 2021-08-02
Rejecting EIP-1559 transactions on unsupported networks

### LINKED ISSUE #11550: Throw an error if transaction submitted with EIP-1559 fields/type 0x2 but on a network that doesn't support EIP-1559
We have a state field in the network controller that detects EIP-1559 support. IF a transaction is submitted that otherwise passes validation (meaning it has type 0x2 without a gasPrice field submitted) but the network doesn't support EIP-1559 we need to reject that request with an appropriate error.

ryanml:
Fixes: https://github.com/MetaMask/metamask-extension/issues/11550

The thought here was it was better to reject in this case inside the validation function _prior_ to the actual submission of the transaction


brad-decker (approved):

adonesky1 (approved):
LGTM!

PR #11728, opened 2021-08-02
remove estimate detail in advanced gas controls

adonesky1:
Explanation:  Removes estimate detail above inputs in advanced controls:
Screen Shot 2021-08-02 at 1 56 39 PM


danjm (approved):

dan437 (approved):
Sweet, it will be more straightforward for users with this change.

PR #11735, opened 2021-08-03
EIP-1559 - Ensure transaction detail font-size and icon colors are consistent with Figma design

darkwing:
I noticed that somewhere we got away from the original font-sizes, so I brought up Figma and restored the font-sizes they should be at:

TD

H7s for all of the text was simply too difficult to see.


danjm (approved):

ryanml (approved):
nice!

PR #11736, opened 2021-08-03
EIP-1559 - Provide better validation for gas price and gas limit

darkwing:
While testing Ledger wallets, I noticed that a user could add a gas price and limit of ''.  We should ensure we get valid values there.

### DIFF
--- a/ui/hooks/useGasFeeInputs.js
+++ b/ui/hooks/useGasFeeInputs.js
@@ -366,7 +366,7 @@ export function useGasFeeInputs(
     { value: minimumGasLimit || GAS_LIMITS.SIMPLE, fromNumericBase: 'hex' },
   );
 
-  if (gasLimitTooLow) {
+  if (gasLimitTooLow || gasLimit === '') {
     gasErrors.gasLimit = GAS_FORM_ERRORS.GAS_LIMIT_OUT_OF_BOUNDS;
   }
 
@@ -411,6 +411,9 @@ export function useGasFeeInputs(
     case GAS_ESTIMATE_TYPES.NONE:
       if (networkAndAccountSupports1559) {
         estimatesUnavailableWarning = true;
+      }
+      if (gasPriceToUse === '' || gasPriceToUse < 1) {
+        gasErrors.gasPrice = GAS_FORM_ERRORS.GAS_PRICE_TOO_LOW;
       }
       break;
     default:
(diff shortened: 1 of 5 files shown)


danjm (approved):

ui/hooks/useGasFeeInputs.js
```diff
@@ -412,6 +412,9 @@ export function useGasFeeInputs(
       if (networkAndAccountSupports1559) {
         estimatesUnavailableWarning = true;
       }
+      if (gasPriceToUse === '' || gasPriceToUse < 1) {
```
danjm, line 415: Gas price can be lower than 1 on custom networks

danjm (changes requested):

danjm (dismissed):

ui/hooks/useGasFeeInputs.js
dan437: If gas price can be lower than 1 on custom networks (e.g. 0), how does the code below work for them? 
```
      if (gasPriceToUse === '' || gasPriceToUse === 0) {
        gasErrors.gasPrice = GAS_FORM_ERRORS.GAS_PRICE_TOO_LOW;
      }
```

NiranjanaBinoy (dismissed):
LGTM

brad-decker (dismissed):

darkwing:
I cleaned up validation here -- empty string is no longer possible because we default to 0, and I also made the gas price validation more resilient.

ui/hooks/useGasFeeInputs.js
```diff
@@ -412,6 +412,9 @@ export function useGasFeeInputs(
       if (networkAndAccountSupports1559) {
         estimatesUnavailableWarning = true;
       }
+      if (gasPriceToUse <= 0) {
```
dan437, line 415: Hi, I'm trying to understand @danjm comment: `Gas price can be lower than 1 on custom networks`. If that's correct, that means a custom network can set gas price e.g. to 0, which should be allowed, but it seems this validation doesn't allow it.

dan437 (commented):
Looks good!

ui/hooks/useGasFeeInputs.js
danjm: @dan437 My concern was for `0 < gasPrice < 1`

I don't think there is a case we need to support where `gasPrice === 0`

danjm: For example, on optimism, gasPrice is (currently) always esimated to 0.015

ui/hooks/useGasFeeInputs.js
adonesky1: @dan437 I believe he means it can be a decimal? (e.g. 0.1)

dan437 (approved):

adonesky1 (approved):
LGTM

PR #11746, opened 2021-08-03
Stop GasFeeController polling when pop closes

adonesky1:
Fixes: (#)1 on [RC v10.0 feedback doc](https://docs.google.com/document/d/1jRyoxmVWXGv-jPCFCpdVBvMwAt7ERkGtfRnPAjEp2y0/edit#heading=h.rtflwh1n2ugq)

Explanation: This stops polling in the gasFeeController when a users closes confirmation screen/window with a pending transaction unfinished.

### DIFF
--- a/app/scripts/background.js
+++ b/app/scripts/background.js
@@ -18,6 +18,7 @@ import {
   ENVIRONMENT_TYPE_POPUP,
   ENVIRONMENT_TYPE_NOTIFICATION,
   ENVIRONMENT_TYPE_FULLSCREEN,
+  POLLING_TOKEN_ENVIRONMENT_TYPES,
 } from '../../shared/constants/app';
 import { SECOND } from '../../shared/constants/time';
 import migrations from './migrations';
@@ -302,11 +303,31 @@ function setupController(initState, initLangCode) {
     );
   };
 
-  const checkAndSetClientOpenStatus = () => {
+  const setClientOpenStatusAndDisconnectGasPolling = (environmentType) => {
     const isClientOpen = isClientOpenStatus();
     controller.isClientOpen = isClientOpen;
-    if (!isClientOpen) {
+    // if all instances of metamask are closed we call a method on the controller to stop gasFeeController polling
+    if (isClientOpen === false) {
       controller.onClientClosed();
+      // otherwise we want to only remove the polling tokens for the environment type that has closed
+    } else {
+      // in the case of fullscreen environment a user might have multiple tabs open so we don't want to disconnect all of
+      // its corresponding polling tokens unless all tabs are closed.
+      if (
+        environmentType === ENVIRONMENT_TYPE_FULLSCREEN &&
+        Boolean(Object.keys(openMetamaskTabsIDs).length)
+      ) {
+        return;
+      }
+      const tokensToDisconnect = controller.appStateController.store.getState()[
+        POLLING_TOKEN_ENVIRONMENT_TYPES[environmentType]
+      ];
+      tokensToDisconnect.forEach((token) => {
+        controller.disconnectAndRemovePollingToken(
+          token,
+          POLLING_TOKEN_ENVIRONMENT_TYPES[environmentType],
+        );
+      });
     }
   };
 
@@ -340,7 +361,7 @@ function setupController(initState, initLangCode) {
         popupIsOpen = true;
         endOfStream(portStream, () => {
           popupIsOpen = false;
-          checkAndSetClientOpenStatus();
+          setClientOpenStatusAndDisconnectGasPolling(ENVIRONMENT_TYPE_POPUP);
         });
       }
 
@@ -349,7 +370,9 @@ function setupController(initState, initLangCode) {
 
         endOfStream(portStream, () => {
           notificationIsOpen = false;
-          checkAndSetClientOpenStatus();
+          setClientOpenStatusAndDisconnectGasPolling(
+            ENVIRONMENT_TYPE_NOTIFICATION,
+          );
         });
       }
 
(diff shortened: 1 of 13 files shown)


danjm (approved):

darkwing (dismissed):

darkwing:
This all looks reasonable and works as expected.  I'd still love @Gudahtt's eyes on this, as a senior engineer outside of the 1559 group.

app/scripts/metamask-controller.js
```diff
@@ -2969,15 +2969,29 @@ export default class MetamaskController extends EventEmitter {
   /* eslint-disable accessor-pairs */
   /**
    * A method for recording whether the MetaMask user interface is open or not.
-   * @private
    * @param {boolean} open
    */
   set isClientOpen(open) {
+    if (!open) {
+      this.onPopUpClose();
```
danjm, line 2976: Just thought of something: if we are calling this method from here, then `onPopupClose` is not quite the right name. Because this can be called on close of other contexts as well, which is what we want.

app/scripts/background.js
```diff
@@ -330,9 +330,9 @@ function setupController(initState, initLangCode) {
 
       if (processName === ENVIRONMENT_TYPE_POPUP) {
         popupIsOpen = true;
-
         endOfStream(portStream, () => {
           popupIsOpen = false;
+          controller.onPopUpClose();
```
danjm, line 335: I just a realized a scenario where this doesn't work:

- have a transaction open in a full tab and also have the popup open. Gas will be polling
- close the popup, but leave the full tab with the transaction open. Gas _should_ continue polling, but won't, because we just closed the popup

danjm: :thinking:

danjm (changes requested):
see latest comment

app/scripts/background.js
danjm: First, as an improvement, instead of calling `onPopupClose` in the setter, lets call it in each of the `endOfStream` callbacks in this file. (Also, see my above comment on the rename of `onPopupClose`). We can do this after `controller.isClientOpen` is set, and only do it if it is false.

I briefly commented on this when we are talking yesterday, but with a bit more sleep in me, I think I have come down more firmly on the side of not wanting to have unexpected sideeffects in a setter. If we are overwriting a setter method, I don't think it should do much more than setting. That could be unexpected for a reader of `background.js`

danjm: So that will safely handle the case where all instances of client connections are closed.

danjm: For the edge in my first comment in this thread, or the other edge cases we discussed yesterday when you have multiple tabs or notification windows open and close some but not all, we need something more.

I think what we have to do is:
- maintain a record of poll tokens per environment (e.g. `popupGasPollTokens`, `notificationGasPollTokens`, `fullScreenGasPollTokens`)
- in the `endOfStream` callback for each environment, if it is not the case where `controller.isClientOpen === false`, then just disconnect polling for each token from that environment

We could track those records in the `appState` controller. We can add / remove them in any place in the frontend where we get a gas poll token upon starting gas polling.

danjm: Then I think we will need separate `onPopupEnvClose()`, `onNotificationEnvClose()` and `onFullScreenEnvClose()` methods for each of the three environments. The method we call when all are closed could then call each of these methods.

ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container.component.js
```diff
@@ -57,13 +57,18 @@ export default class GasModalPageContainer extends Component {
         disconnectGasFeeEstimatePoller(pollingToken);
       }
     });
+    window.addEventListener('beforeunload', this._beforeUnload);
   }
 
-  componentWillUnmount() {
+  _beforeUnload = () => {
     this._isMounted = false;
```
Gudahtt, line 64: It seems like this should still be set upon unmount.

danjm: yes, good catch

app/scripts/background.js
adonesky1: > First, as an improvement, instead of calling `onPopupClose` in the setter, lets call it in each of the `endOfStream` callbacks in this file. 

@danjm Would this include `processName === ENVIRONMENT_TYPE_NOTIFICATION`? No right?

app/scripts/controllers/app-state.js
```diff
@@ -191,4 +194,27 @@ export default class AppStateController extends EventEmitter {
...
+      [pollingTokenType]: [
+        ...this.store.getState()[pollingTokenType],
+        pollingToken,
+      ],
+    });
+  }
+
+  removePollingToken(pollingToken, pollingTokenType) {
+    this.store.updateState({
+      [pollingTokenType]: [
```
danjm, line 213: Nit: Could be a little easier to read if we put `this.store.getState()[pollingTokenType]` in a variable at the start of the function, and then reference the variable here

ui/ducks/send/send.js
```diff
@@ -1298,6 +1302,7 @@ export function resetSendState() {
       await disconnectGasFeeEstimatePoller(
         state[name].gas.gasEstimatePollToken,
       );
+      removePollingTokenFromAppState(state[name].gas.gasEstimatePollToken);
```
danjm, line 1305: Let's make sure that `resetSendState` is dispatched on beforeunload in the send component

app/scripts/background.js
```diff
@@ -302,6 +303,34 @@ function setupController(initState, initLangCode) {
...
+      // otherwise we want to only remove the polling tokens for the environment type that has closed
+    } else {
+      // in the case of fullscreen environment a user might have multiple tabs open so we don't want to disconnect all of
+      // its corresponding polling tokens unless all tabs are closed.
+      if (
+        environmentType === ENVIRONMENT_TYPE_FULLSCREEN &&
+        Boolean(Object.keys(openMetamaskTabsIDs).length)
+      ) {
+        return;
+      }
```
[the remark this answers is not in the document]
adonesky1: done!

danjm (approved):

darkwing (approved):

PR #11759, opened 2021-08-04
EIP-1559 - Prevent uncaught exception when passing fees to getGasFeeTimeEstimate

darkwing:
Peter was seeing an uncaught exception in our `GasTiming` component because very small decimal values were being converted to scientific notation, and `gasFeeController.getTimeEstimate` requires two strings -- we were passing numbers.


adonesky1 (approved):
LGTM

danjm (approved):

PR #11760, opened 2021-08-04
Ensure gas fees update in popover on poll for new values

danjm:
Fixes an issue where the gas price estimates in the edit gas popover were not automatically updating after first opening the popover. This PR ensures that the `maxFeePerGas` and `maxPriorityFeePerGas` get set to null on initiatilization of `useGasFeeInputs` if the initial values (those that come from the transaction) are the same as one of the gas fee estimates

### DIFF
--- a/ui/hooks/useGasFeeInputs.js
+++ b/ui/hooks/useGasFeeInputs.js
@@ -206,41 +206,51 @@ export function useGasFeeInputs(
     estimatedGasFeeTimeBounds,
   } = useGasFeeEstimates();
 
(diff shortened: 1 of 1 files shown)


darkwing (approved):
This worked great for updating!  I tested by:

1.  Wait for it to update, updates value successfully
2.  Add custom values, wait for update, custom value doesn't change
3.  Choose a radio button, wait for update, values *do* update

Excellent!

adonesky1 (approved):
LGTM!

PR #11766, opened 2021-08-04
Fix legacy unapproved tx handling

danjm:
This PR handles the following case:

- user is on v9.8.4
- user has an unapproved transaction
- user updates to v10.0.0
- user opens the unapproved transaction again

Currently this will result in the user seeing `0` in the max fee and max priority fee field.

To fix this, we default max fee and max priority fee to gas price if on an eip 1559 network but they do not exist.


adonesky1 (approved):
LGTM

darkwing (approved):

PR #11767, opened 2021-08-04
Ensure that gas fee minimum errors show when api is down

danjm:
Ensures that minimum errors are correctly shown for max fee and max priority fee, even when gas fee api is down

Screenshot from 2021-08-04 19-42-20
Screenshot from 2021-08-04 19-42-10


adonesky1 (approved):
LGTM

darkwing (approved):

Issue #11761, opened 2021-08-04
EIP-1559 - Small UI Tweaks


darkwing:
As provided by @jakehaugen:

- [ ] https://www.notion.so/Missing-hover-state-for-high-med-low-options-77dd4f8c811548ca9701f5b5fd09b67f
- [ ] https://www.notion.so/While-hovering-over-options-the-pointer-link-cursor-should-ap[…]urrently-just-shows-the-norma-fc70efd8913f4ba9959e400f1cb10a2a
- [ ] https://www.notion.so/Radio-buttons-are-small-this-may-be-a-global-change-we-need-t[…]e-everywhere-Should-be-18px-v-1a3740d4230f4b90adcef91f2cf7a7d1
- [ ] https://www.notion.so/Max-fee-should-be-bolded-just-the-label-not-the-amount-1ea80a19e9b54d2e8eef2aa5cb962781
- [ ] https://www.notion.so/Spacing-between-type-and-radio-buttons-should-be-24px-e1e2feb357ab403a8434bfe57c9c6650

jakehaugen:
Adding a few more, happy to discuss priority if/when it makes sense. 

- [ ] https://www.notion.so/Radio-button-options-are-not-centered-in-dialog-Looks-like-maybe-it-s-centered-including-the-labels-66e2d479abdf4d539d38d3e4bf6d1a08

- [ ] https://www.notion.so/Up-to-amount-should-be-Max-fee-amount-aeafa8065ae0410db371f50c75206663

- [ ] https://www.notion.so/Vertical-spacing-between-the-estimated-amount-max-amount-and-time-should-all-be-16px-equal-258da809a8cd45c0ae52d328a62beb2b

- [ ] https://www.notion.so/Selected-radio-button-label-should-be-Black-8929aeb4001f4a17837376396c6c7d8a

PR #11768, opened 2021-08-04
Update all radio button sizes to 18px

darkwing:
Fixes an item in https://github.com/MetaMask/metamask-extension/issues/11761

@jakehaugen  wants all radio buttons to be `18px x 18px` instead of default sizes.  Form controls are notoriously different and strict on different operating systems but I tested in Mac+Chrome, Mac+Firefox, and Windows+Chrome, and this looked good.

### DIFF
--- a/ui/pages/settings/settings-tab/settings-tab.component.js
+++ b/ui/pages/settings/settings-tab/settings-tab.component.js
@@ -4,6 +4,7 @@ import availableCurrencies from '../../../helpers/constants/available-conversion
 import Dropdown from '../../../components/ui/dropdown';
 import ToggleButton from '../../../components/ui/toggle-button';
 import locales from '../../../../app/_locales/index.json';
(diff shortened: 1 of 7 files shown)


ui/pages/settings/settings-tab/settings-tab.component.js
```diff
@@ -185,12 +187,13 @@ export default class SettingsTab extends PureComponent {
                 </label>
               </div>
               <div className="settings-tab__radio-button">
-                <input
-                  type="radio"
+                <RadioButton
+                  name="preferred-currency"
```
ryanml, line 191: should this have a different `name` than the button for the native primary currency?

darkwing:
I think we should punt on this for when the design system comes in, tbh; I think we can close.  Changing radio button size also isn't a standard thing.

PR #11769, opened 2021-08-04
Correct the vertical margins on the RadioGroup component

darkwing:
Fixes an issue here:  https://github.com/MetaMask/metamask-extension/issues/11761

The extra vertical space is for the recommendation system which we've not used yet.


ryanml (approved):

adonesky1 (approved):
LGTM

PR #11771, opened 2021-08-05
Keep radio group contained and equal width

darkwing:
The current radiogroup stretches slightly out of the current margins, as well as gives more space to the "low" panel than the other two options.  This keeps the radio group in bounds, and gives equal widths.

When we come back to reinstating the "Recommended" stuff, we'll need to look at these widths again.


adonesky1 (dismissed):
LGTM

adonesky1 (approved):

ryanml (approved):

PR #11775, opened 2021-08-05
Ensure that gas fee inputs fallback to tx params values if api is down

danjm:
If the user opens the edit gas popover, and has not yet manually set any values, and the api is down, the value we show the user should fallback to the value from the transaction parameters, instead of '0'.


dan437 (dismissed):

PeterYinusa (approved):
LGTM!

dan437 (approved):

PR #11776, opened 2021-08-05
Use bignumber for number comparisons in useGasFeeInput

danjm:
This PR ensures that the number comparisons we do to set errors in useGasFeeInputs will be valid for all inputs. It can prevent errors when comparing stringified numbers, or highly precise numbers.

### DIFF
--- a/ui/helpers/utils/util.js
+++ b/ui/helpers/utils/util.js
@@ -345,3 +345,31 @@ export function constructTxParams({
   }
   return addHexPrefixToObjectValues(txParams);
 }
+
+export function bnGreaterThan(a, b) {
+  if (a === null || a === undefined || b === null || b === undefined) {
+    return null;
+  }
+  return new BigNumber(a, 10).gt(b, 10);
+}
+
+export function bnLessThan(a, b) {
+  if (a === null || a === undefined || b === null || b === undefined) {
+    return null;
+  }
+  return new BigNumber(a, 10).lt(b);
+}
+
+export function bnGreaterThanEqualTo(a, b) {
+  if (a === null || a === undefined || b === null || b === undefined) {
+    return null;
+  }
+  return new BigNumber(a, 10).gte(b);
+}
+
+export function bnLessThanEqualTo(a, b) {
+  if (a === null || a === undefined || b === null || b === undefined) {
+    return null;
+  }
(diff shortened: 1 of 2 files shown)


dan437 (dismissed):
Looks good. If we can add UTs in some future PR for these util functions, that would be great.

ui/helpers/utils/util.js
```diff
@@ -345,3 +345,31 @@ export function constructTxParams({
...
+    return null;
+  }
+  return new BigNumber(a, 10).gt(b, 10);
+}
+
+export function bnLessThan(a, b) {
+  if (a === null || a === undefined || b === null || b === undefined) {
+    return null;
+  }
+  return new BigNumber(a, 10).lt(b);
```
darkwing, line 360: Does `b` need `,10`?

ui/helpers/utils/util.js
```diff
@@ -345,3 +345,31 @@ export function constructTxParams({
...
+}
+
+export function bnGreaterThanEqualTo(a, b) {
+  if (a === null || a === undefined || b === null || b === undefined) {
+    return null;
+  }
+  return new BigNumber(a, 10).gte(b);
```
darkwing, line 367: Does `b` need `,10`?

ui/helpers/utils/util.js
```diff
@@ -345,3 +345,31 @@ export function constructTxParams({
...
+}
+
+export function bnLessThanEqualTo(a, b) {
+  if (a === null || a === undefined || b === null || b === undefined) {
+    return null;
+  }
+  return new BigNumber(a, 10).lte(b);
```
darkwing, line 374: Does `b` need `,10`?

darkwing:
Do we want to return `false` instead of null?  I guess I expect a boolean value from these functions.  We could also trigger an error to sentry or something?

darkwing:
Do we want to return `false` instead of null?  I guess I expect a boolean value from these functions.  We could also trigger an error to sentry or something?

QA'd and it works but love to know if we should be adding the second param for `b`.

darkwing (dismissed):

dan437 (dismissed):

darkwing (approved):

dan437 (approved):

PR #11778, opened 2021-08-05
Allow max fee to be equal to max priority fee

danjm:
We should allow max fee to be equal to max priority fee, it just can't be less than it.


darkwing (dismissed):

NiranjanaBinoy (dismissed):
LGTM

darkwing (approved):

NiranjanaBinoy (approved):
LGTM

PR #11786, opened 2021-08-05
EIP-1559 - Provide support for Ledger

darkwing:
Fixes: Provides Ledger support for EIP-1559

## Depends on:
- [ ] https://github.com/MetaMask/eth-ledger-bridge-keyring/pull/95 being merged
- [ ] https://github.com/MetaMask/eth-ledger-bridge-keyring/pull/93 being merged, and a new release cut and then included in this PR

## How to Test
* Check out this PR locally
* In the extension's `node_modules/@metamask/eth-ledger-bridge-keyring/index.js` file, change `BRIDGE_URL` to `'https://darkwing.github.io/eth-ledger-bridge-keyring'`,  comment out the lines changed in this PR (https://github.com/MetaMask/eth-ledger-bridge-keyring/pull/93), and comment out the `origin` check in the `_sendMessage` method.

### DIFF
--- a/ui/selectors/selectors.js
+++ b/ui/selectors/selectors.js
@@ -82,8 +82,9 @@ export function getCurrentKeyring(state) {
 }
 
 export function isEIP1559Account(state) {
-  // Neither hardware wallet supports 1559 at this time
-  return !isHardwareWallet(state);
+  // Only Ledger supports 1559 at this times
+  const currentKeyring = getCurrentKeyring(state);
+  return currentKeyring && currentKeyring.type !== 'Trezor Hardware';
 }
 
 export function checkNetworkAndAccountSupports1559(state) {
--- a/ui/selectors/selectors.js
+++ b/ui/selectors/selectors.js
@@ -82,7 +82,7 @@ export function getCurrentKeyring(state) {
 }
 
 export function isEIP1559Account(state) {
-  // Only Ledger supports 1559 at this times
+  // Trezor does not support 1559 at this time
   const currentKeyring = getCurrentKeyring(state);
   return currentKeyring && currentKeyring.type !== 'Trezor Hardware';
 }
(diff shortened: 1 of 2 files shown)


darkwing:
Just tested this briefly on the latest Ledger Live software update.  I see `Max Fee` on the Ledger confirmation, but the value is always `0`, so there's work to be done here.  Transaction insta-fails.

bayesianmind:
FWIW signing 1559 transactions works fine with ledger when used with mycrypto.com's ledger support. The ledger confirmation screen shows the max fee as nonzero, and the transaction goes through as expected.

I tried this patch and got the same results as you - Ledger shows 0 max gas, and the txn fails in metamask immediately.

ui/selectors/selectors.js
```diff
@@ -82,8 +82,9 @@ export function getCurrentKeyring(state) {
 }
 
 export function isEIP1559Account(state) {
-  // Neither hardware wallet supports 1559 at this time
-  return !isHardwareWallet(state);
+  // Only Ledger supports 1559 at this times
```
Gudahtt, line 85: Nit: I assume this meant "Among our supported hardware wallets, only Ledger supports 1559". This code is run for local accounts too after all. Maybe this wording would be clearer:

```suggestion
  // Trezor does not support 1559 at this time
```

darkwing:
I've updated the testing steps; to pin to the exact version of Ledger we need for Palm and EIP-1559 support, since updating to a newer version is a breaking change.  Only this PR and the `gh-pages` of the keyring need to be updated.

ui/selectors/selectors.js
```diff
@@ -82,8 +82,9 @@ export function getCurrentKeyring(state) {
...
+  const currentKeyring = getCurrentKeyring(state);
+  return currentKeyring && currentKeyring.type !== 'Trezor Hardware';
```
ryanml, line 87: Nit: this literal `'Trezor Hardware'` is used in a few places throughout the codebase (defined as `KEYRING_TYPES.TREZOR` in `MetamaskController`) this would be a good candidate for a shared constant I think

adonesky1:
This PR was superceded by #11951

PR #11787, opened 2021-08-05
EIP-1559 - Ensure form always displays when there are errors

darkwing:
This is a small but useful UX fix -- we default `showAdvancedForm` to `true` if we start with errors, but if I start with `Medium` and no errors, then pick `High` and there are errors, we need to open the form so they know why the submit button is disabled.


danjm (approved):

NiranjanaBinoy (approved):
LGTM

PR #11802, opened 2021-08-10
Replace hardcoded sent ether label on confirm screen

### LINKED ISSUE #10974: Copy update:

adonesky1:
Fixes #10974
Fixes issue where `sent ether` label appears on confirm screen for sends of non-ether native currencies of custom networks:
Screen Shot 2021-08-10 at 10 42 47 AM

Manual testing steps:  
- Enter send flow on a custom network with native currency other than ETH
- Check that label in top left accurately describes the currency being sent.

Now:
Screen Shot 2021-08-10 at 12 05 15 PM


rachelcope:
This looks like a big improvement 👍

ryanml (dismissed):
tested this out, looks great!

ryanml (dismissed):

shared/constants/transaction.js
```diff
@@ -11,7 +11,8 @@ import { MESSAGE_TYPE } from './app';
  * @property {'approve'} TOKEN_METHOD_APPROVE - A token transaction requesting an
  *  allowance of the token to spend on behalf of the user
  * @property {'incoming'} INCOMING - An incoming (deposit) transaction
- * @property {'sentEther'} SENT_ETHER - A transaction sending ether to a recipient
+ * @property {'sentEther'} SENT_ETHER - A transaction sending ether to a recipient (LEGACY - replaced by sendingNativeAsset)
```
brad-decker, line 14: [Q] Does the sentEther type still exist?

shared/constants/transaction.js
```diff
@@ -11,7 +11,8 @@ import { MESSAGE_TYPE } from './app';
...
+ * @property {'sendingNativeAsset'} SENDING_NATIVE_ASSET - A transaction sending a network's native asset to a recipient
```
brad-decker, line 15: I do not like conflating transaction types with translation strings. I would prefer us to use something like 'simpleSend', and then use the sendingNativeAsset key when we encounter that type of transaction. I know we already conflate the others, but this is an opportunity to move in the right direction.

brad-decker (changes requested):

shared/constants/transaction.js
adonesky1: It would be a legacy transaction type that we could safely remove later on, in case say a user has a pending simple send transaction when the version update takes place. I kept it as a legacy type instead of creating a migration for this edge case.

shared/constants/transaction.js
adonesky1: Ok yeah that makes sense. I'll make this change.

adonesky1:
@brad-decker ready for another look when you have time. @ryanml a few small changes since your approval.

shared/constants/transaction.js
```diff
@@ -49,6 +50,7 @@ export const TRANSACTION_TYPES = {
   TOKEN_METHOD_APPROVE: 'approve',
   INCOMING: 'incoming',
   SENT_ETHER: 'sentEther',
+  SENDING_NATIVE_ASSET: 'simpleSend',
```
brad-decker, line 53: Let's update the variable name too. So that it is descriptive of the actual transaction type.

brad-decker (changes requested):

darkwing (dismissed):

shared/constants/transaction.js
adonesky1: ✅

shared/constants/transaction.js
brad-decker: if a user has a sentEther transaction pending, it would continue to be a sentEther indefinitely right? What mechanism updates a transaction's type?

adonesky1:
@darkwing just changed variable names per @brad-decker's comment

shared/constants/transaction.js
brad-decker: I think a migration is required here to treat the sentEther type as legacy, or to ever remove it.

brad-decker (changes requested):

shared/constants/transaction.js
adonesky1: Sorry I was mistaken. We would need a migration to update transaction types. I've implicitly made it legacy by swapping out `sentEther` as a type we set as a transaction type for `simpleSend`, but kept the logic to support transactions still typed as `sentEther` which would either need to be left there (because it is used for ui related logic with activity log) or migrate and swap all transaction type instances of `sentEther` for `simpleSend`. I was being lazy but a migration is probably called for here.

adonesky1: Added a migration. Should I remove the 'sentEther' type from the `TRANSACTION_TYPES` and replace the reference to this (enum) object in old migrations as well? cc @brad-decker @danjm

app/scripts/migrations/053.test.js
```diff
@@ -64,7 +64,10 @@ describe('migration #53', function () {
       TransactionController: {
         transactions: [
           { type: TRANSACTION_TYPES.CANCEL, txParams: { foo: 'bar' } },
-          { type: TRANSACTION_TYPES.SENT_ETHER, txParams: { foo: 'bar' } },
+          {
+            type: TRANSACTION_TYPES.SENT_ETHER,
```
brad-decker, line 68: Was this a lint related change? If SENT_ETHER doesn't exist anymore this might be a lint issue

shared/constants/transaction.js
brad-decker: Kind of a conundrum. Replacing the value in old migrations obfuscates what the migration was actually doing. I might, instead, create a local variable for SENT_ETHER in the migration and note what it is and why it exists.

shared/constants/transaction.js
```diff
@@ -49,6 +50,7 @@ export const TRANSACTION_TYPES = {
   TOKEN_METHOD_APPROVE: 'approve',
   INCOMING: 'incoming',
   SENT_ETHER: 'sentEther',
```
brad-decker, line 52: Should def remove the legacy type

brad-decker (changes requested):

ryanml (dismissed):
lgtm aside from brad's comments

shared/constants/transaction.js
adonesky1: ✅

app/scripts/migrations/053.test.js
adonesky1: replaced with local const in the test file.

brad-decker (approved):

ryanml (approved):

PR #11080, opened 2021-05-13
Build - refactor for bundle factoring and swappable runtime

kumavis:
this refactor unblocks incremental lavamoat protections for the webapp
this is **does not add lavamoat** yet

some additional benefits:
- automatically factors bundles into entry (phishing/ui/background) and "common" module groups
- module groups are then reduced to subgroups based on a size limit, and then compose into bundles 
- bundles call into a swappable runtime api that allows the same module definition to be used in lavamoat or as in here a normal commonjs runtime. eg background could use lavamoat protections while ui does not, even as both consume the same "common" bundles

additional `watchify` and `patch-package` note:
in order to remove an incompatible watchify feature (binary output statistics), we used `patch-package`. it was determined to be the easiest and least hacky way to disable the non-optional feature.

PR #11813, opened 2021-08-11
Fix Sentry release configuration

Gudahtt:
The Sentry `release` was not being configured correctly. It was being left blank. This is because the location of the extension version was moved in #11029. The build script was correctly updated in that PR, but that work was accidentally undone in a merge error that was included in #11080.


danjm (approved):
LGTM

adonesky1 (approved):

kumavis:
my bad, thanks/

PR #11818, opened 2021-08-11
Add low/medium/high gas recommendations to the speedup/cancel gas popover

adonesky1:
Explanation:  For the speedup and cancel edit gas popovers we currently only show the advanced gas controls pre-populated with values 10% higher than the original values submitted. We should give the users the option to select from updated recommendations of low/medium/high.

If the originally submitted values were very low the 10% bump that we use is still too low:
Screen Shot 2021-08-11 at 3 40 18 PM

User can now opt back into a recommended price that reflects current fee market conditions.
Screen Shot 2021-08-11 at 3 40 21 PM


darkwing (approved):

PeterYinusa:
LGTM!

ryanml (approved):

PR #11823, opened 2021-08-12
Do not attempt to track tx metrics event if txMeta is not available

ryanml:
Related sentry error: https://sentry.io/organizations/metamask/issues/2502760408/?project=273505&query=is%3Aunresolved&sort=freq&statsPeriod=14d

It is possible for the transaction meta to be undefined (https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/controllers/transactions/tx-state-manager.js#L608) for a transaction when setting the status - if that is the case, we should not attempt to de-structure `txMeta` and send a metrics event.

Alternative: Perhaps send the metrics event anyway in this case with message denoting the absence of transaction meta?


darkwing (approved):

NiranjanaBinoy (approved):
LGTM

PR #11829, opened 2021-08-13
Improving handling for non-existent props in state migrations

ryanml:
Related sentry errors:
https://sentry.io/organizations/metamask/issues/2402390721/
https://sentry.io/organizations/metamask/issues/2411887599

We have to be extra cautious when accessing state attributes like controllers, etc during migrations


NiranjanaBinoy (approved):
LGTM

rekmarks (approved):
LGTM!

PR #11845, opened 2021-08-16
Add support for swaps via a direct contract for wrapping and unwrapping

dan437:
## Explanation
When wrapping / unwrapping tokens on different networks, we will only show 1 quote that will do a swap via a direct contract integration.

## Manual testing steps
  - Try to do a swap from ETH -> WETH. You will notice there are no Advanced Options on the Build Quote page and no link to select from different quotes on the View Quote page, since there will only be 1 quote available for wrapping / unwrapping

## Screenshots
Advanced Options are hidden for wrapping / unwrapping across all supported networks:
image

If there is only 1 quote available (which is the case for wrapping / unwrapping), don't show a link to choose from quotes:
image


darkwing (dismissed):
Looks good and works well!

darkwing (dismissed):

shared/constants/swaps.js
```diff
@@ -114,6 +118,37 @@ export const SWAPS_CHAINID_CONTRACT_ADDRESS_MAP = {
   [RINKEBY_CHAIN_ID]: TESTNET_CONTRACT_ADDRESS,
 };
 
+export const SWAPS_WRAPPED_TOKENS_ADDRESSES = {
+  [MAINNET_CHAIN_ID]: WETH_CONTRACT_ADDRESS,
+  [SWAPS_TESTNET_CHAIN_ID]: WETH_CONTRACT_ADDRESS,
+  [BSC_CHAIN_ID]: WBNB_CONTRACT_ADDRESS,
+  [POLYGON_CHAIN_ID]: WMATIC_CONTRACT_ADDRESS,
+  [RINKEBY_CHAIN_ID]: WETH_CONTRACT_ADDRESS,
```
danjm, line 126: Will the `WETH_CONTRACT_ADDRESS` work on rinkeby? WETH probably has a different address on rinkeby...

danjm: yeah, this might be the token on rinkeby: https://rinkeby.etherscan.io/token/0xc778417e063141139fce010982780140aa0cd5ab

danjm (changes requested):
Only one concern that I mention in my comment. The rest of the PR looks good.

danjm (dismissed):

danjm (dismissed):

danjm (approved):

darkwing (approved):

PR #11862, opened 2021-08-17
Increase saliency of mobile sync warning

### LINKED ISSUE #11861: Increase warning saliency on wallet sync screen
Phishers are now asking users to share their extension sync QR code, since it provides a less severe warning than our "secret recovery phrase reveal" screen.

Story (threading broken):
- https://twitter.com/Ape_NFTs/status/1427704710396620803
- https://twitter.com/Ape_NFTs/status/1427705408550998024?s=20
- https://twitter.com/Ape_NFTs/status/1427705527669346316?s=20
- https://twitter.com/Ape_NFTs/status/1427705883933425664?s=20
- https://twitter.com/Ape_NFTs/status/1427706077332873220?s=20
- https://twitter.com/Ape_NFTs/status/1427706478421544969?s=20
- https://twitter.com/Ape_NFTs/status/1427706755270815755?s=20

This is a good sign! It means phishers are finding it more difficult to get secret recovery phrases!

This also means we have inadequate text warnings on the QR sync screen.

E9A-hL_VgAEqsKS
E9A-iSSVEAApt4y

Some improvement ideas:
- On the "Sync with mobile" screen, the red text could say "Showing this code will allow migrating all of your accounts. Support staff will never ask for this, but phishers might."
- Saying "make sure nobody else is looking" on the screen that is sensitive is _too late_. Any warning like that should also be on the screen before.

danfinlay:
Improves on #11861

Explanation:  
Phishers have begun requesting users display the "mobile sync" QR code to steal accounts. This user path does not have adequately severe warnings, so we need to increase them. This PR includes some proposed improved copy.

New translations should follow this PR as soon as possible.

Manual testing steps:  
  - This is only a copy change, and should not affect behavior at all.


app/_locales/en/messages.json
```diff
@@ -1130,8 +1130,8 @@
     "message": "verify the network details",
     "description": "Serves as link text for the 'mismatchedChain' key. This text will be embedded inside the translation for that key."
   },
-  "mobileSyncText": {
-    "message": "Please enter your password to confirm it's you!"
+  "mobileSyncWarning": {
+    "message": "Proceeding will show a code that allows exporting all of your accounts. Support staff will never request this, but phishers would."
```
ryanml, line 1134: Perhaps it's already well implied enough, but would `MetaMask support staff` be better here?

jspence425:
@danfinlay  Might be too lengthy for that panel, but this is a bit more explicit: 

> ⚠️ Proceeding will display a QR code that reveals the Secret Recovery Phrase for all your accounts, allowing for them to be imported into MetaMask Mobile.
> 
> This is for your eyes only. MetaMask will never ask you for it, but a phisher would.

app/_locales/en/messages.json
danfinlay: The current phishing attacks in the linked tweets were conducted by impersonating OpenSea support staff, so I thought it was better to leave this open.

danfinlay:
>  Might be too lengthy for that panel, but this is a bit more explicit:
This isn't exactly how it works. The QR code embeds a channel name and a shared session key, which is good for establishing a one-time connection over which the seed phrase is then encrypted with that symmetrical encryption key.

Rather than trying to explain how it works, I think emphasizing what is possible is just a bit easier while preserving the important content.

jspence425:
So in this instance, the phisher used their own MetaMask Mobile to sync to the victim's QR code? That makes sense.

Copy looks good to me in that sense. Also agree with saying "phishers" over being specific about "metamask support staff"

danjm (dismissed):

app/_locales/en/messages.json
```diff
@@ -1297,8 +1297,8 @@
     "message": "verify the network details",
     "description": "Serves as link text for the 'mismatchedChain' key. This text will be embedded inside the translation for that key."
   },
-  "mobileSyncText": {
-    "message": "Please enter your password to confirm it's you!"
+  "mobileSyncWarning": {
+    "message": "⚠️ Proceeding will show a code that allows exporting all of your accounts. Support staff will never request this. DO NOT SHARE!"
```
rekmarks, line 1301: I don't want to quibble over this urgent fix, but on the other hand my bank doesn't yell at me in all caps and neither should we:

```suggestion
    "message": "⚠️ Proceeding will display a secret QR code that allows access to your accounts. Do not share it with anyone. Support staff will never ask your for it."
```

danfinlay: I agree. I think I took the all caps from another suggestion, even though I agree it rubs against our established text language (have we written that anywhere? I feel like Rachel helped set that standard)

rekmarks (approved):
LGTM!

app/_locales/en/messages.json
```diff
@@ -1297,8 +1297,8 @@
     "message": "verify the network details",
     "description": "Serves as link text for the 'mismatchedChain' key. This text will be embedded inside the translation for that key."
   },
-  "mobileSyncText": {
-    "message": "Please enter your password to confirm it's you!"
+  "mobileSyncWarning": {
+    "message": "⚠️ Proceeding will display a secret QR code that allows access to your accounts. Do not share it with anyone. Support staff will never ask your for it."
```
rekmarks, line 1301: For reviewers: This is what you want to review.

danjm (approved):

PR #11865, opened 2021-08-17
Added iframe execution environment

shanejonas:
This adds the [Iframe Execution Environment](https://github.com/MetaMask/iframe-execution-environment/) as the default service used for snaps.

image

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -32,6 +32,7 @@ import {
   ExternalResourceController,
   WebWorkerExecutionEnvironmentService,
 } from '@mm-snap/controllers';
+import { IframeExecutionEnvironmentService } from '@mm-snap/iframe-execution-environment-service';
 import { TRANSACTION_STATUSES } from '../../shared/constants/transaction';
 import { MAINNET_CHAIN_ID } from '../../shared/constants/network';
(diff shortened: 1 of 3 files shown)


app/scripts/metamask-controller.js
```diff
@@ -382,9 +382,11 @@ export default class MetamaskController extends EventEmitter {
       initState.PermissionsMetadata,
     );
 
-    this.workerController = new WebWorkerExecutionEnvironmentService({
+    this.workerController = new IframeExecutionEnvironmentService({
       setupPluginProvider: this.setupPluginProvider.bind(this),
-      workerUrl: WORKER_BLOB_URL,
+      iframeUrl: new URL(
+        'https://metamask.github.io/iframe-execution-environment/',
```
rekmarks, line 388: We're probably going to need to figure out a different place to host this, but it's fine to in-line it for now.

rekmarks (dismissed):
Yissssssss

rekmarks (approved):

app/scripts/metamask-controller.js
shanejonas: also, if this url was on IPFS I think it would be better when we get more production ready. Content addressable gives us nice benefits.

PR #11871, opened 2021-08-18
Fix content cutoff issue (Mobile Sync View)

ryanml:
There was an unneeded `margin-top` value being applied from the `new-account-import-form` rules, which are used elsewhere

Popup view:
Screen Shot 2021-08-18 at 12 22 19 AM

### DIFF
--- a/ui/pages/mobile-sync/mobile-sync.component.js
+++ b/ui/pages/mobile-sync/mobile-sync.component.js
@@ -408,7 +408,10 @@ export default class MobileSyncPage extends Component {
     const { password } = this.state;
 
     return (
-      <div className="new-account-import-form__buttons" style={{ padding: 30 }}>
+      <div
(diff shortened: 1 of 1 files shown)


ui/pages/mobile-sync/mobile-sync.component.js
```diff
@@ -408,7 +408,10 @@ export default class MobileSyncPage extends Component {
     const { password } = this.state;
 
     return (
-      <div className="new-account-import-form__buttons" style={{ padding: 30 }}>
+      <div
+        className="new-account-import-form__buttons"
```
ryanml, line 412: Ideally the mobile sync component would have its own set of css rules, but they are borrowed elswhere, hence the pre-existing `padding` overridee

danjm (approved):

tmashuang (approved):
LGTM

PR #11872, opened 2021-08-18
Dont default gas limit in gas popover when non is available on transaction params

danjm:
This PR removes the default value for gas limit in our gas editing popover. This case will only occur if no gas limit was suggested by a dapp, and calls to `eth_estimateGas` have failed not due to a simulation error, but a network failure. This would happen if the user is disconnected from the internet, or there is a failure connecting to the provider. In such a case, defaulting to 0 is safe because it requires the user to be intentional about the gas limit.


ryanml (approved):

tmashuang (approved):
LGTM as it stands. There are cases where the gas limit will be provided by a dapp, and if the gas api/eth_gasPrice is sluggish or down altogether, there will not be an error on the confirm tx screen which users might question as to why they won't be able to confirm the proposed tx.

PR #11720, opened 2021-08-02
_locales/en/messages.json - The default network ... is Mainnet.

paulvi:
not "Main Net", see https://eips.ethereum.org/EIPS/eip-2228

Explanation:  The name of the network is correct, but comment was incorrect

Manual testing steps:  
  - this is only resource string(s)

PR #11880, opened 2021-08-18
Update verbiage of Main Net to Mainnet

darkwing:
Supercedes https://github.com/MetaMask/metamask-extension/pull/11720 ; thank you @paulvi!


ryanml:
Thanks for this! While we're in here, do you think we should additionally capitalize the use of `Mainnet` everywhere in our translations? Seems consistently capitalized, here's an instance where it isn't: https://github.com/MetaMask/metamask-extension/blob/develop/app/_locales/en/messages.json#L1523

NiranjanaBinoy (approved):
lgtm

ryanml (approved):

PR #11887, opened 2021-08-19
Fix 'yarn setup' on M1 Macs

mcmire:
There are a few issues encountered when running `yarn setup` on new Apple
Silicon (aka M1, aka arm64) Macs:

* The script halts when attempting to run the install step for
  the `chromedriver` package with the message "Only Mac 64 bits
  supported". This is somewhat misleading as it seems to indicate that
  chromedriver can only be installed on a 64-bit Mac. However, what I
  think is happening is that the installation script for `chromedriver`
  is not able to detect that an arm64 CPU *is* a 64-bit CPU. After
  looking through the `chromedriver` repo, it appears that 87.0.1 is the
  first version that adds a proper check ([1]).

  Note that upgrading chromedriver caused the Chrome-specific tests
  to fail intermittently on CI. I was not able to 100% work out the reason
  for this, but ensuring that X (which provides a way for Chrome to run
  in a GUI setting from the command line) is available seems to fix these
  issues.
* The script also halts when attempting to run the install step for
  the `electron` package. This happens because for the version of
  `electron` we are using (9.4.2), there is no available binary for
  arm64. It appears that Electron 11.x was the first version to support
  arm64 Macs ([2]). This is a bit tricker to resolve because we don't
  explicitly rely on `electron` — that's brought in by `react-devtools`.
  The first version of `react-devtools` that relies on `electron` 11.x
  is 4.11.0 ([3]).

[1]: https://github.com/giggio/node-chromedriver/commit/469dd0a6ee23540bfa832b5f09b4cbde3e152010
[2]: https://www.electronjs.org/blog/apple-silicon
[3]: https://github.com/facebook/react/blob/main/packages/react-devtools/CHANGELOG.md#4110-april-9-2021


mcmire:
Changing this to a draft while I work out the CI issues on this.

test/e2e/webdriver/chrome.js
```diff
@@ -6,7 +6,7 @@ const chrome = require('selenium-webdriver/chrome');
  */
 class ChromeDriver {
   static async build({ responsive, port }) {
-    const args = [`load-extension=dist/chrome`];
+    const args = ['load-extension=dist/chrome', '--no-sandbox'];
```
mcmire, line 9: I found I had to do this to get CI to pass, as I was getting the following error: "DevToolsActivePort file doesn't exist". Here is a link to a failing build: https://app.circleci.com/pipelines/github/MetaMask/metamask-extension/12078/workflows/e69a57ce-0950-4da6-9a5b-aee777a5f4cc/jobs/327566. I googled for "DevToolsActivePort file doesn't exist" and found [this SO answer](https://stackoverflow.com/questions/50642308/webdriverexception-unknown-error-devtoolsactiveport-file-doesnt-exist-while-t) which seemed to indicate that `--no-sandbox` makes the error go away. That said, I am not sure whether this is the best fix or whether there is a better one?

tmashuang (dismissed):
lgtm

package.json
```diff
@@ -249,7 +249,7 @@
     "brfs": "^2.0.2",
     "browserify": "^16.5.1",
     "chalk": "^3.0.0",
```
danjm, line 251: @Gudahtt Any concern with this version change for chromedriver?

Gudahtt: The last time I tried, CI consistently failed. That's the only reason it has stayed at v79.

.circleci/scripts/chrome-install.sh
```diff
@@ -4,11 +4,11 @@ set -e
 set -u
 set -o pipefail
 
-CHROME_VERSION='79.0.3945.117-1'
+CHROME_VERSION='87.0.4280.88-1'
 CHROME_BINARY="google-chrome-stable_${CHROME_VERSION}_amd64.deb"
 CHROME_BINARY_URL="http://mirror.cs.uchicago.edu/google-chrome/pool/main/g/google-chrome-stable/${CHROME_BINARY}"
```
Gudahtt, line 9: We only switched to this mirror because google took v79 down. We can switch back to google if we're updating it again.

The link is
```suggestion
CHROME_BINARY_URL="https://dl.google.com/linux/chrome/deb/pool/main/g/google-chrome-stable/${CHROME_BINARY}"
```

If we're sticking with the mirror, we should at least verify that the hash matches the one from Google. But I expect it'll be faster to download this from Google than from uchicago.edu anyway.

test/e2e/webdriver/chrome.js
Gudahtt: We should have already been setting `--no-sandbox` :thinking: Were we setting it incorrectly I wonder? Here:
https://github.com/MetaMask/metamask-extension/blob/develop/.circleci/scripts/chrome-install.sh#L27

Gudahtt: But if it works, it's definitely better than what we had before! This error is what stopped us from updating Chrome for so long. We've long wanted to update.

Gudahtt:
Here is my previous attempt to update Chrome by the way: https://github.com/MetaMask/metamask-extension/pull/11302/files
.I left notes on my progress, and a few additional ideas that I hadn't pursued yet.

danjm (commented):
LGTM

tmashuang (dismissed):

danjm (dismissed):

mcmire:
I forgot that force-pushing resets the approvals 😣 Can I get another approval pls?

mcmire (commented):
@Gudahtt I think I managed to make this work! Do you want to take a look at this again? I'm happy to chat about this on Monday and walk back through my investigation if you want as well.

development/lib/retry.js
```diff
@@ -5,9 +5,28 @@
  * @param {number} retries - The number of retries upon failure to attempt.
  * @param {function} functionToRetry - The function that will be retried upon failure.
  */
-async function retry(retries, functionToRetry) {
+async function retry(...args) {
```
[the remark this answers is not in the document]
Gudahtt: It looks like the JSDoc comment still needs to be updated here.

Though, I'm not sure how best to document this sort of pattern, where the arguments are positional but the meaning is different depending on how many there are :thinking:. Have you considered switching this to use an options bag instead? e.g. passing in the parameters as an object. That would be easier to document at least.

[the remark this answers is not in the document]
mcmire: Yeah, that would definitely be a simpler API. Let me look into how many places are using this. What's our preferred command for generating docs?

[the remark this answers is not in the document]
Gudahtt: The JSDoc in this repo is just internal at the moment. We used to generate docs from it, but haven't in a long time.

development/lib/run-command.js
```diff
@@ -15,15 +15,18 @@ const spawn = require('cross-spawn');
...
+async function runCommand(command, args, options = {}) {
   const output = [];
   let mostRecentError;
   let errorSignal;
   let errorCode;
   const internalError = new Error('Internal');
   try {
     await new Promise((resolve, reject) => {
-      const childProcess = spawn(command, args, { encoding: 'utf8' });
+      const childProcess = spawn(command, args, {
```
Gudahtt, line 26: What were these changes used for? :thinking:

.circleci/scripts/chrome-install.sh
```diff
@@ -24,6 +24,6 @@ fi
 
 rm -rf "${CHROME_BINARY}"
 
-sudo sed -i 's|HERE/chrome"|HERE/chrome" --disable-setuid-sandbox --no-sandbox|g' "/opt/google/chrome/google-chrome"
```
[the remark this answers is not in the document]
Gudahtt: Huh. That is fascinating. It's great that we'll be using a sandbox! But I am surprised that this works.

Easy enough to restore if it causes problems I suppose!

Which warnings did you see after removing this?

test/e2e/helpers.js
```diff
@@ -83,6 +84,9 @@ async function withFixtures(options, testSuite) {
       });
       await segmentServer.start(9090);
     }
+    if (process.env.SELENIUM_BROWSER === 'chrome') {
+      await ensureXServerIsRunning();
```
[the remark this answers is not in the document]
Gudahtt: Amazing discovery! That explains a lot.

development/lib/run-command.js
mcmire: `runCommand` and `runInShell` now take an options object which allows you to configure `spawn`. In this case these options are specified in `ensureXServerIsRunning` to prevent `xset q` from outputting anything to stdout, which was happening here https://app.circleci.com/pipelines/github/MetaMask/metamask-extension/12187/workflows/777baae3-4b51-42c8-8ff8-c4c7ede52861/jobs/329760 and due to this commit https://github.com/MetaMask/metamask-extension/commit/ab16ddf957433ac049d8dd2a3c93adaee9d8151d. I guess changing `runInShell` isn't technically needed here but I updated both for consistency.

mcmire: Ah, I see. I originally used `runInShell` then changed it to `runCommand`. By itself that might prevent output to stdout. Let me try removing these changes and seeing if that still works.

.circleci/scripts/chrome-install.sh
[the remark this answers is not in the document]
mcmire: Upon re-running this with logging, it doesn't seem that many warnings/messages that get produced are related to the sandbox. The only one I see is this one:

```
[2789:2789:0830/210054.551055:ERROR:sandbox_linux.cc(374)] InitializeSandbox() called with multiple threads in process gpu-process.
```

but I think that's just telling us that the sandbox is being initialized and doesn't indicate a problem.

development/lib/run-command.js
mcmire: `runCommand` does seem to work the way I want it — I'm not getting seeing any output from `xset` in the CircleCI logs. :)

Gudahtt: Nice!

.circleci/scripts/chrome-install.sh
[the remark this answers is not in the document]
Gudahtt: Cool! Good to know. I agree - that does not seem concerning.

I had been under the impression that when you run Chrome in a container it typically runs as root, which requires additional sandbox configuration to work correctly. But maybe this is taken care of for us already in the base image :thinking: . Or maybe I was mistaken, or this changed in a Chrome update.

Gudahtt (approved):
LGTM, great work! :tada:

test/e2e/x-server.js
```diff
@@ -0,0 +1,17 @@
+const { runCommand } = require('../../development/lib/run-command');
+const { retry } = require('../../development/lib/retry');
+
+function ensureXServerIsRunning() {
+  return retry(
+    {
+      retries: 3,
+      delay: 2000,
+      rejectionMessage: 'X server does not seem to be running?!',
```
adonesky1, line 9: 🤣

package.json
```diff
@@ -249,7 +249,7 @@
...
-    "chromedriver": "^79.0.0",
+    "chromedriver": "^87.0.1",
```
adonesky1, line 252: [Q] What's the basis for upgrading to this particular version? It looks like latest is `92.0.2`, matching current chrome version?

adonesky1 (approved):
LGTM! Just one non-blocking question

package.json
Gudahtt: The reasoning was included in the PR description - this was the minimum version that included a fix for the bug that was encountered. I agree we should probably update this further though. Maybe in a future PR?

mcmire: Yeah, once this is through I will submit another PR to bring it up to the latest version so that we can run Chrome tests locally.

Issue #11894, opened 2021-08-19
SignTypedData error -- incorrect chainid


JTraversa:
**Describe the bug**
When attempting to use signTypedData_v3 or v4, I am returned the error `MetaMask - RPC Error: Provided chainId "66" must match the active chainId "42"` 

(signed on Kovan(42), but reading chainId as OkEx Chain(66))

**Steps to reproduce**
Just attempt to sign a typed data object using signtypeddata
(https://swivel.exchange)

**Expected behavior**
The modal with the typed data pops up allowing me to sign the data.

**Browser details (please complete the following information):**
 - OS: Windows
 - Browser: Chrome 92.0.4515.131
 - MetaMask Version: 10.0.2
 (did not happen pre 10.0.2)

**Other comments**
I've tested other deployments of ours which dont appear to have this issue. Perhaps it is an issue with ethers versioning (5.0.31)?

JTraversa:
I've tried replicating outside of this app and haven't been able to.

We're rebuilding our app and releasing a new version next week, so this isn't necessarily a blocker for us as it isn't replicated in the new app. 

It is just definitely strange behavior, and after checking our signing methods, ethers, etc, its definitely a bug on the side of the metamask RPC.

jknight:
I can reproduce this consistently on FF and Chrome. We are using ethers 5.4.4

jknight:
Is this a duplicate of #11830 ?

Pol-Lanski:
Same issue as https://github.com/MetaMask/metamask-extension/issues/11847

jmagder:
I haven't been able to reproduce this but I have two team members who consistently reproduce this.

jmagder:
I am able to reproduce this consistently.  The chainId is being treated as a hex number.  Even if I provide 01337. it gets interpreted as 0x1337.

jmagder:
https://github.com/MetaMask/metamask-extension/pull/11900

Gudahtt:
This should be fixed as of #11900, which will be included in the next release.

Issue #11892, opened 2021-08-19
Receiving transaction payload with chainId set to 0


mortimr:
**Describe the bug**
Since MM 10.0.2, the payload returned from Metamask is setting chainId to 0. In my setup (local mainnet fork with chainId set to 1), everything was working properly yesterday, but this change that probably occured in the last published version is triggering an error in `useDapp`, when the tool checks for the chainId.

**Steps to reproduce (REQUIRED)**
I'm not sure if the issue is happening on all network or what might be the reason behind it so I cannot provide a clear reproduction example.

**Expected behavior**
Chain ID is set to actual chain id.

**Screenshots**
![Screenshot 2021-08-19 at 10 42 43](https://user-images.githubusercontent.com/17824082/130037722-6862b329-98f1-4e1a-99a5-b88603b61ea0.png)

**Browser details (please complete the following information):**
 - OS: macOs
 - Hardware Wallet: none
 - Browser: Version 92.0.4515.159 (Official Build) (arm64)
 - MetaMask Version: 10.0.2

mortimr:
Just tested versions `10.0.1` and `10.0.0` and the issue is there too. Version `9.8.4` works properly (sends back the proper chainId).

vujevits:
getting the same error when using `rinkeby` in `useDapp`

quantizations:
we are getting this same issue

kennym:
Same issue after updating to v10.x.x - also tested in 10.0.1 and still breaks.

mortimr:
In the meantime, here you can find a workaround ! https://github.com/EthWorks/useDApp/issues/289#issuecomment-901944570

PR #11900, opened 2021-08-20
Fix issue with chainId comparison in sign typed message param validation

### LINKED ISSUE #11847: Upgrading to 10.0 for EIP1559 support breaks xDai Network Signing
**Describe the bug**

I'm trying to stake xNODE on the xDai network.

The website is 
https://app.dappnode.io/

Prior to the Version 10 upgrade I would be given a signing request, which I would confirm. Then I would be given a transaction to confirm to stake the xNode tokens.

After upgrading to Metamask 10.0.1 the signature request does not work. Metamask starts to show the pop-up for the signature, then almost immediately closes the pop-up.

1. Go to https://app.dappnode.io/  -- you must already have an account there with some xNODE tokens
2. Click on "Stake xNODE Tokens"
3. Type in the number of tokens you want to stake, then click "Deposit xNODE Tokens"
4. The Metamask pop-up appears for a moment, but then disappears again without showing anything.

**Expected behavior**
I was supposed to get a signature request, followed by a transaction to approve

**Screenshots**

 - OS: Windows 10
 - Hardware Wallet: None
 - Browser: Firefox Browser 91.0 (64-bit)
 - MetaMask Version: 10.0.1

Other users have confirmed the same problem

Rolling back to an earlier version of Metamask fixes the problem.

adonesky1:
Fixes: #11847

Explanation:  Some dapps pass a non-hex string for chainId in signTypedData calls. We should support this.

Manual testing steps:  
  - Attempt to sign a typed message using the test dapp on a network that is not mainnet, one of the ethereum testnets, or a ganache local network. This should now work as expected.

### DIFF
--- a/app/scripts/lib/typed-message-manager.js
+++ b/app/scripts/lib/typed-message-manager.js
@@ -204,7 +204,7 @@ export default class TypedMessageManager extends EventEmitter {
             `Cannot sign messages for chainId "${chainId}", because MetaMask is switching networks.`,
           );
           if (typeof chainId === 'string') {
-            chainId = parseInt(chainId, 16);
+            chainId = parseInt(chainId, chainId.startsWith('0x') ? 16 : 10);
           }
           assert.equal(
             chainId,


danjm:
Would this handle the case of someone submitting a strong like '1a'?

adonesky1:
@danjm well I'm not sure what the expected behavior for that case would be but as it stands '1a' would parse to 1 and so if the user was on mainnet and the `1a` was passed as the chainId in the signTypedData call it would work.

jmagder:
I've been experiencing the same issue the past couple of days.  I looked at the git blame of line 207 and it looks like that has been there for 2 months, yet this issue only started a couple days ago.  Furthermore, I'm not passing in the chainId as a string, but as a number.  Something else must have changed between 10.0.1 (or 10.0.0) and 10.0.2.  As such I'm not sure this is fixing the regression.

kennym:
Does this address #11892?

jmagder:
I just tested with an earlier version, 9.8.4 from https://addons.mozilla.org/en-US/firefox/addon/ether-metamask/versions/, and the wrong chain ID message went away.

danjm:
Sorry @adonesky1 I should have been more clear with my last comment. What if the network is Optimism on Goerli, which has a chainId of 420, which is `1a4` in base 16.

You are correct that `parseInt('1a4', 10) === 1`. Meanwhile, `parseInt('1af', 16) === 420`

So if the user is on the Optimism Goerli network, they will see `Provided chainId "1" must match the active chainId "420"`

I was in a rush so wrote a pithy, cryptic and incomplete question in my previous message. I can now write more fully: what if the chainId provided by the dapp is a hex number, but not prefixed with `0x`

Prior to this PR, I think it would work. With this PR, I think it starts to break. I think the problem here is a lack of specifications in the EIP around acceptable types for this parameter. This was described in the issue description here https://github.com/MetaMask/metamask-extension/issues/11308#issue-921569010

I'm wondering if the correct course of action here is:
- go back to the state we were in after https://github.com/MetaMask/metamask-extension/pull/11309/files, and ship that
- then have the larger conversation that Dan Finlay alludes to here: https://github.com/MetaMask/metamask-extension/issues/11308#issuecomment-861635511
- then once we decide on a more specific set of requirements for the chainId parameter, publish warnings of a breaking change to dapp developers. and after a period of properly warning developers, start enforcing the stricter type checking

danjm:
Ah, and now I am seeing the above messages from jmagder and kennym. So it seems there might be something else happening as well...

@jmagder So this issue didn't happen for you on v10.0.0? And could you share an example of message data where you are passing the chainId as a number but still seeing this error?

danjm:
Also, @jmagder could you clarify: are you passing the number as a String type `typeof chainId === 'string'` or Number type `typeof chainId === 'number'`?

I see your example here: https://github.com/MetaMask/metamask-extension/issues/11894#issuecomment-903043778

Are you passing `1337` or `'01337'`?

kennym:
> Also, @jmagder could you clarify: are you passing the number as a String type `typeof chainId === 'string'` or Number type `typeof chainId === 'number'`?
> 
> I see your example here: [#11894 (comment)](https://github.com/MetaMask/metamask-extension/issues/11894#issuecomment-903043778)
> 
> Are you passing `1337` or `'01337'`?

in our case we're passing a number - `80001` for Mumbai.

jmagder:
> Ah, and now I am seeing the above messages from jmagder and kennym. So it seems there might be something else happening as well...
> 
> @jmagder So this issue didn't happen for you on v10.0.0? And could you share an example of message data where you are passing the chainId as a number but still seeing this error?
> 
> Thank-you

I might have been on the 9.x release, I'm not 100% certain I was on 10.0.  The problem does not exist in 9.x, so my team has been unblocked running the earlier version.  Previously I was passing 1337 as a **number**, not as a string.  I glanced at the changes between 10.0.1 and 10.0.2 and didn't see anything that might have caused the regression.  Though I didn't go through with a fine tooth comb and it was my first time looking at your code base. :-)  The line that was changed in this PR was last modified 2 months ago.  But the code from 2 months ago is what was in 9.x.  I think the regression is likely around some assumptions of a string input.

Let me know if I can be of further help!

adonesky1:
@jmagder could you pull down this branch: https://github.com/MetaMask/metamask-extension/tree/revert-signtypeddata-validation and test if the issue is still happening for you?

jmagder:
I'll give it a shot, just got back in front of the computer.  Will check it out after my grooming meeting.  Cheers!

jmagder:
I'm pleased to report that this branch worked! :tada:  I confirmed it was a number being passed in, as retrieved from  `provider.getSigner().getChainId()` using ethers.js 5.4.0.  The provider was created using `new ethers.providers.Web3Provider(window.ethereum)`

jmagder:
> @jmagder could you pull down this branch: https://github.com/MetaMask/metamask-extension/tree/revert-signtypeddata-validation and test if the issue is still happening for you?

It worked! I confirmed it was a number being passed in, as retrieved from  `provider.getSigner().getChainId()` using ethers.js 5.4.0.  The provider was created using `new ethers.providers.Web3Provider(window.ethereum)`.  Looks like you've got a beautiful lint error: :smile: 

```
/home/circleci/project/app/scripts/lib/typed-message-manager.js
  199:15  error  'chainId' is never reassigned. Use 'const' instead  prefer-const
```

From https://app.circleci.com/pipelines/github/MetaMask/metamask-extension/12105/workflows/1b161007-26d0-4134-82ac-93321c557e70/jobs/328126

Thanks so much for the fix, let me know if you need me to test again.  Cheers!  :partying_face:

app/scripts/lib/typed-message-manager.js
```diff
@@ -204,7 +204,7 @@ export default class TypedMessageManager extends EventEmitter {
             `Cannot sign messages for chainId "${chainId}", because MetaMask is switching networks.`,
           );
           if (typeof chainId === 'string') {
-            chainId = parseInt(chainId, 16);
+            chainId = parseInt(chainId, chainId.startsWith('0x') ? 16 : 10);
```
jmagder, line 207: I've tested the fix at https://github.com/MetaMask/metamask-extension/tree/revert-signtypeddata-validation which fixes this issue without the parsing issues mentioned in this thread. :)

kennym:
Wohoo! What are chances of getting this being merged soon?

danjm:
@kennym There are unresolved concerns with this PR. Notably see my comment above. While this may fix the issue for dapps that are passing the chainId as a decimal string number, it may cause a new regression for dapps that are passing the chainId as a hex string without a `0x` prefix. Once we can either prove that concern invalid, or further modify this to handle that concern, then we can merge this. We will be working on this today.

Gudahtt:
We used to rely on the non-strict `assert.equal` to ensure the `chainId` passed in matches the current chain id. `assert.equal` will infer that `0x`-prefixed strings represent hex numbers, but it will assume non-`0x`-prefixed strings are decimal numbers.

So for the scenario discussed above:
```
const assert = require('assert');

const currentChainId = 420;
const messageChainId = '1af';

assert(messageChainId, currentChainId); // This throws with the error "Uncaught AssertionError [ERR_ASSERTION]: '1af' == 420"
```

So, because of that loose assertion, it's not possible anyone was passing in non-prefixed hex strings before v10 and getting the behaviour they expected. I believe assuming strings are decimal unless hex prefixed will result in the exact same behaviour we saw before v10.

jmagder:
> @kennym There are unresolved concerns with this PR. Notably see my comment above. While this may fix the issue for dapps that are passing the chainId as a decimal string number, it may cause a new regression for dapps that are passing the chainId as a hex string without a `0x` prefix. Once we can either prove that concern invalid, or further modify this to handle that concern, then we can merge this. We will be working on this today.

What about your proposed fix at https://github.com/MetaMask/metamask-extension/tree/revert-signtypeddata-validation

adonesky1:
@jmagder thanks for your patience here. We're just trying to get a few things sorted out and I think we will be able to merge one fix or another soon. Would you mind just confirming whether the changes in this PR (not the other branch I asked you to test) are causing the issue for you?

jmagder:
> @jmagder thanks for your patience here. We're just trying to get a few things sorted out and I think we will be able to merge one fix or another soon. Would you mind just confirming whether the changes in this PR (not the other branch I asked you to test) are causing the issue for you?

Sure thing, I'll give it another run using this branch.

jmagder:
> > @jmagder thanks for your patience here. We're just trying to get a few things sorted out and I think we will be able to merge one fix or another soon. Would you mind just confirming whether the changes in this PR (not the other branch I asked you to test) are causing the issue for you?
> 
> Sure thing, I'll give it another run using this branch.

I rebuilt using this branch and it also seems to fix the issue.  So I guess with @Gudahtt 's comment you don't need to worry about hex handling regressions with this fix.  Awesome!

danjm:
@Gudahtt's comments and explanation above have resolved my concerns.

danjm (approved):

Gudahtt (approved):
LGTM!

PR #11901, opened 2021-08-20
Notify user of snap initial execution errors

shanejonas:
This adds a browser notification to alert the user of an initial snap execution error.

We need to make the error message itself a bit more user friendly when we define the error codes for the execution environment.

This is just a stop-gap updating the filsnap calls to use the error returned from `executePlugin` until we have the full snaps API implemented again.


rekmarks (dismissed):
LGTM!

rekmarks (approved):

PR #11909, opened 2021-08-22
Fixing error handling for `publish()` (MobileSync)

ryanml:
Related Sentry Error: https://sentry.io/organizations/metamask/issues/2213968901/?project=273505&query=is%3Aunresolved+toString&statsPeriod=14d

When `publish` resolves with errors, we shouldn't `reject` with the `response` (which is undefined in error cases), rather `status.errorData`, which contains the `Error` object we want to throw. (https://www.pubnub.com/docs/sdks/javascript/api-reference/publish-and-subscribe)

Also localized the sync failure message

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -2415,6 +2415,9 @@
   "symbolBetweenZeroTwelve": {
     "message": "Symbol must be 11 characters or fewer."
   },
+  "syncFailed": {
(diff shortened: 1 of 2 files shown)


NiranjanaBinoy (approved):
lgtm

adonesky1 (approved):
LGTM!

PR #11915, opened 2021-08-24
Quotes prefetching in Swaps

dan437:
## Explanation
As a user I want to see quotes in Swaps ideally asap after clicking on the "Review Swap" button, instead of waiting for about 3 - 5 seconds on the Quotes Loading page.

We do that via silent quotes prefetching on the Build Quote page, once we have form values and the "Review Swap" button is enabled. If a user clicks on the "Review Swap" button while quotes prefetching is in progress, we still redirect to the Quotes Loading page, but it will be typically finished faster than before, since the API call started before a user clicked on the "Review Swap" button.

## Manual testing steps
- Click on the "Swap" button on the main page
- Select token from, token to and add amount of tokens to be sent
- Once the "Review Swap" button is enabled, we start quotes prefetching based on form values. After clicking on the "Review Swap" button a few seconds later, a user is taken immediately to the View Quote page.

## Screenshots
Build Quote page with form fields filled in:
image

View Quote page displayed asap after a user clicks on the "Review Swap" button on the Build Quote page:
image

Old Quotes Loading page:
image

New Quotes Loading page (displayed when quotes prefetching is in progress):
image

Quotes Timeout page, which is displayed on the View Quote page after we do polling for quotes 3 times:
image


ui/pages/swaps/swaps.util.js
```diff
@@ -418,21 +416,27 @@ export async function fetchSwapsFeatureFlags() {
   return response;
 }
 
-export async function fetchSwapsQuoteRefreshTime(chainId, useNewSwapsApi) {
+export async function fetchSwapsRefreshRates(chainId, useNewSwapsApi) {
```
brad-decker, line 419: My main concern here is that this method is declared here but only ever consumed in the controller. Could we lift this into the controller itself and import the pieces it requires from this file? It isn't a complete fix but it's better than so vastly separating implementation with the intended call site.

ui/pages/swaps/loading-swaps-quotes/loading-swaps-quotes.js
```diff
@@ -195,37 +136,9 @@ export default function LoadingSwapsQuotes({
...
-                style={{
-                  opacity: aggName === aggregatorNames[quoteCount] ? 1 : 0,
-                  top:
-                    aggregatorLocationMap[aggName]?.y + midPointTarget?.y ?? 0,
-                  left:
-                    aggregatorLocationMap[aggName]?.x + midPointTarget?.x ?? 0,
-                }}
-                key={`aggregator-logo-${aggName}`}
-              >
-                <AggregatorLogo
```
brad-decker, line 222: If we are removing AggregatorLogo from being called we should also remove the files.

brad-decker (changes requested):
couple changes I see immediately, sorry my brain is hurting trying to get my context back on this one. I'll do more of a review soon!

ui/pages/swaps/swaps.util.js
dan437: That's a great point! Basically, we don't need it in utils if it's only used once and that code can be in the Swaps controller instead. I don't see the `getBaseApi` nor `fetchWithCache` function calls in the Swaps controller, but we can start doing it now for this one.

ui/pages/swaps/build-quote/build-quote.js
```diff
@@ -392,6 +410,51 @@ export default function BuildQuote({
...
+          maxSlippage,
+          metaMetricsEvent,
+          pageRedirectionDisabled,
+        ),
+      );
+    };
+    // Delay fetching quotes until a user is done typing an input value. If they type a new char in less than a second,
+    // we will cancel previous setTimeout call and start running a new one.
+    timeoutIdForQuotesPrefetching = setTimeout(() => {
+      timeoutIdForQuotesPrefetching = null;
```
danjm, line 441: Should we clear the previous timeout here, in addition to setting the variable to null?

ui/pages/swaps/build-quote/build-quote.js
```diff
@@ -348,6 +365,7 @@ export default function BuildQuote({
 
   useEffect(() => {
     dispatch(resetSwapsPostFetchState());
+    dispatch(setReviewSwapClickedTimestamp());
```
danjm, line 368: Is this dispatch just to set the reviewSwapClickedTimestamp to`undefined`?

danjm (commented):
Okay, I think this looks good. I left a couple of questions and can probably approve once they are answered.

app/scripts/controllers/swaps.js
```diff
@@ -124,50 +121,84 @@ export default class SwapsController {
...
+      typeof refreshRates.quotes !== 'number' ||
+      typeof refreshRates.quotesPrefetching !== 'number'
+    ) {
+      throw new Error(
+        `MetaMask - invalid response for refreshRates: ${response}`,
+      );
+    }
+    // We presently use milliseconds in the UI.
+    return {
+      quotes: refreshRates.quotes * 1000,
```
darkwing, line 142: In theory this could end up being `0`, since `0` is a `number`

ui/pages/swaps/build-quote/build-quote.js
```diff
@@ -392,6 +410,51 @@ export default function BuildQuote({
...
     selectedToToken.address,
   );
+  const isReviewSwapButtonDisabled =
+    tokenFromError ||
+    !isFeatureFlagLoaded ||
+    !Number(inputValue) ||
+    !selectedToToken?.address ||
+    Number(maxSlippage) < 0 ||
+    Number(maxSlippage) > MAX_ALLOWED_SLIPPAGE ||
+    (toTokenIsNotDefault && occurrences < 2 && !verificationClicked);
```
darkwing, line 420: Not your fault, but this conditional is massive and maybe a tiny bit confusing.  I don't have a solution, but maybe we should think on this in the future.

brad-decker (dismissed):
LGTM thanks for making those changes.

app/scripts/controllers/swaps.js
dan437: That's correct, but in that case we would use fallback values instead of 0: https://github.com/MetaMask/metamask-extension/pull/11915/files/690200973b39b440018539daf2580a48359f98b2#diff-71af45d691cfae9944ce2e51b42d9c9830714eaba88b4207404d4ff332317dd1R164

ui/pages/swaps/build-quote/build-quote.js
dan437: Exactly that. When the Build Quote page is loaded, I want to clean it up and set a fresh timestamp on submit.

ui/pages/swaps/build-quote/build-quote.js
dan437: When this `useEffect` call for checking form changes is triggered again, it's automatically cleared by:
```
return () => clearTimeout(timeoutIdForQuotesPrefetching);
```

ui/pages/swaps/build-quote/build-quote.js
dan437: Agree with you! I've just moved all these conditions into a constant, but they were there previously. I will keep an eye on this and will refactor it / add explanatory comments if it would be getting any more complex.

brad-decker (approved):

danjm (approved):

PR #11922, opened 2021-08-25
Set up minimum Jest threshold for "ui" and "shared" folders

dan437:
## Explanation
Up until now we have been using minimum Jest threshold only for the Swaps feature and a little bit of the Send feature.

This PR enforces minimum Jest test coverage for all JavaScript files in the "ui" and "shared" folders, which means that most PRs will need to include unit tests in order to be green. Right now the minimum threshold is pretty low (around 40%), so it should be pretty simple to add enough unit tests to reach it.

This PR is a simple change, which will have high impact on quality, because the more unit tests we have, the less bugs we will have in production, refactoring will be less stressful and can happen more often because of that and we will have more confidence in our production pushes.

I understand that people will need to get used to it, but it's necessary in order to increase our standards.

## Manual testing steps
  - Send a PR with new business logic without any unit tests
  - PR will not be green until you meet the minimum Jest threshold

## Screenshots
Not enough coverage (`yarn test:coverage:jest`):
image

Enough coverage (`yarn test:coverage:jest`):
image

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,17 +1,14 @@
 module.exports = {
   restoreMocks: true,
   coverageDirectory: 'jest-coverage/',
-  collectCoverageFrom: [
-    '<rootDir>/ui/**/swaps/**',
-    '<rootDir>/ui/ducks/send/**',
-  ],
+  collectCoverageFrom: ['<rootDir>/ui/**/*.js', '<rootDir>/shared/**/*.js'],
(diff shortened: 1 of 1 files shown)


brad-decker (approved):
also strongly in favor.

darkwing (approved):

brad-decker:
@dan437 we spoke about this in the extension standup, and are all in agreement to merge. Letting you do the honors ❤️

PR #11936, opened 2021-08-25
Use higher gasfees when attempting to speedup or cancel a transaction

### LINKED ISSUE #11928: Not broadcasting replacement tx unless priority fee is identical
During speed up or cancel, metamask will not broadcast the TX (either increase in gas, cancellation, or replacement) if the Max Priority Fee is not exactly the same as the original TX

Should be a quick fix, but prevents sending txs, so is urgent.

adonesky1:
Possibly Fixes: #11928

Explanation:  When a user wants to speed up or cancel a transaction the fees we should pre-populate the gas popover form with should be the higher of the following two options: the original fee + 10% or the current medium suggested fees (for current market conditions) from our gas estimation api.

Manual testing steps:  
  - Initialize a send transaction with very low maxFee and maxPriorityFee 
  - Attempt to speed up or cancel the transaction
  - The newly suggested fees should be greater than a 10% increase on the originals (unless gas fee prices have precipitously dropped since you sent the transaction)

### DIFF
--- a/ui/hooks/useCancelTransaction.test.js
+++ b/ui/hooks/useCancelTransaction.test.js
@@ -6,6 +6,14 @@ import { getConversionRate, getSelectedAccount } from '../selectors';
 import { increaseLastGasPrice } from '../helpers/utils/confirm-tx.util';
 import { useCancelTransaction } from './useCancelTransaction';
 
+jest.mock('../store/actions', () => ({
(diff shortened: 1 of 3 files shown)


danfinlay (approved):
Looks right to me.

danjm (approved):
Nice work!

PR #11949, opened 2021-08-27
Allow run-ganache script to have additional command line options

bentobox19:
So you can do for example

```
./run-ganache.sh --db /tmp/lololol
```

And be able to persist your development blockchain across runs.


bentobox19:
Disregard. In the end one has to get used to the `GANACHE_ARGS` env var.

PR #11972, opened 2021-08-30
Using a normal bundle instead of a factored bundle for phishing-detect (build-sys)

### LINKED ISSUE #11971: "Continue at your own risk" button does not work
Dear manager,

Metamask protection appears when opening some sites. I was taking the risk. I can't access the site even though I took this risk for a week. What should I do. I am waiting for your help, please

Metamask

ryanml:
Fixes: https://github.com/MetaMask/metamask-extension/issues/11971 - since this was previously a factored build, `phishing.html` was not referencing the correct bundled js file: https://github.com/MetaMask/metamask-extension/blob/develop/app/phishing.html#L8 - this should be fine as a normal bundle

### DIFF
--- a/development/sourcemap-validator.js
+++ b/development/sourcemap-validator.js
@@ -24,7 +24,7 @@ async function start() {
     `common-0.js`,
     `background-0.js`,
     `ui-0.js`,
-    'phishing-detect-0.js',
+    'phishing-detect.js',
     // `contentscript.js`, skipped because the validator is erroneously sampling the inlined `inpage.js` script
     `inpage.js`,
   ];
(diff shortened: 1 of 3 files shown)


kumavis (dismissed):

kumavis:
> Error: SourcemapValidator - failed to load source file for "phishing-detect-0.js"
sourcemap validator needs to be updated

ryanml:
@kumavis - validator updated 👍

rekmarks (dismissed):
LGTM!

development/sourcemap-validator.js
```diff
@@ -24,7 +24,6 @@ async function start() {
     `common-0.js`,
     `background-0.js`,
     `ui-0.js`,
-    'phishing-detect-0.js',
```
rekmarks, line 27: I approved, but we don't want this in here?

rekmarks: By our build system lingo, it's no longer a "factored build", so no, we don't.

kumavis:
we should still validate it

kumavis (approved):

PR #11975, opened 2021-08-31
@metamask/controllers@15.0.0

rekmarks:
Adds the latest version of `@metamask/controllers`, and updates our usage of the `ApprovalController`, which has been migrated to `BaseControllerV2`. Of [the new `controllers` release](https://github.com/MetaMask/controllers/releases/tag/v15.0.0), only the `ApprovalController` migration should be breaking.

This is the first time we use events on the `ControllerMessenger` to update the badge, so I turned the messenger into a property on the main `MetaMaskController` in order to subscribe to events on it in `background.js`. I confirmed that the badge does indeed update during local QA.

As it turns out, [MetaMask/controllers#571](https://github.com/MetaMask/controllers/pull/571) was breaking for a single unit test case, which is now handled during setup and teardown for the related test suite (`metamask-controller.test.js`).

### DIFF
--- a/app/scripts/background.js
+++ b/app/scripts/background.js
@@ -435,9 +435,13 @@ function setupController(initState, initLangCode) {
     METAMASK_CONTROLLER_EVENTS.UPDATE_BADGE,
     updateBadge,
   );
-  controller.approvalController.subscribe(updateBadge);
   controller.appStateController.on(
     METAMASK_CONTROLLER_EVENTS.UPDATE_BADGE,
+    updateBadge,
+  );
+
+  controller.controllerMessenger.subscribe(
+    METAMASK_CONTROLLER_EVENTS.APPROVAL_STATE_CHANGE,
     updateBadge,
   );
 
--- a/app/scripts/metamask-controller.test.js
+++ b/app/scripts/metamask-controller.test.js
@@ -109,6 +109,7 @@ describe('MetaMaskController', function () {
   const noop = () => undefined;
 
   before(async function () {
+    globalThis.AbortController = window.AbortController;
     await ganacheServer.start();
   });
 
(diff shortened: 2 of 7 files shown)


app/scripts/metamask-controller.test.js
```diff
@@ -109,6 +109,7 @@ describe('MetaMaskController', function () {
   const noop = () => undefined;
 
   before(async function () {
+    globalThis.AbortController = window.AbortController;
```
rekmarks, line 112: As an aside, that I had to do this has me wondering when we'll fully migrate to `jest`.

app/scripts/background.js
```diff
@@ -435,12 +435,16 @@ function setupController(initState, initLangCode) {
...
-  controller.approvalController.subscribe(updateBadge);
   controller.appStateController.on(
     METAMASK_CONTROLLER_EVENTS.UPDATE_BADGE,
     updateBadge,
   );
 
+  controller.controllerMessenger.subscribe(
+    METAMASK_CONTROLLER_EVENTS.APPROVAL_STATE_CHANGE,
+    updateBadge,
+  );
```
shanejonas, line 446: somewhat related nit: we never clean up any of these listeners in this file. 

this main controller only gets used once in the lifecycle so we dont need to, but just wanted to point it out

rekmarks: I'm pretty confident that we do have event-related leaks, so it's great to keep an eye out for it!

adonesky1 (approved):
LGTM!

Gudahtt (approved):
LGTM! Most of the breaking changes in this release don't have any impact it seems. Just the ApprovalController really.

shanejonas (approved):

PR #11982, opened 2021-08-31
Sort contacts alphabetically

### LINKED ISSUE #10318: Contacts are not sorted properly.
**Describe the bug**
Contacts are sorted in partial alphabetical order.

1. Add contacts with names starting with various letters, and some starting with the same letter.
2. Notice that each first letter is put into its own section.
3. Notice that when two names start with the same first letter, they are not sorted by the second letter.

**Expected behavior**
Names are sorted consistently.

**Screenshots**
image

 - OS: Windows
 - Hardware Wallet: N/A
 - Browser: Firefox 84.0.2
 - MetaMask Version: 8.1.11

mcmire:
Contacts are grouped together by letter, and the groups are listed
alphabetically, but the contacts in each group are not sorted
alphabetically themselves.

Fixes #10318.

---

### Before

Screen Shot 2021-09-01 at 9 40 01 AM

### After

Screen Shot 2021-08-31 at 5 00 49 PM


ui/components/app/contact-list/contact-list.component.js
```diff
@@ -50,34 +51,36 @@ export default class ContactList extends PureComponent {
...
-          items={groupItems}
-          onSelect={this.props.selectRecipient}
-          selectedAddress={this.props.selectedAddress}
-        />
-      ));
+    const sortedContactGroups = letters.map((letter) => {
+      return [
+        letter,
+        sortBy(unsortedContactsByLetter[letter], (contact) => {
+          return contact.name.toLowerCase();
```
mcmire, line 70: Note that each contact is sorted case-insensitively. Personally I find this less confusing as when I look through I list that I know is sorted, I don't think about case. But I'm happy to change this if we think otherwise.

mcmire (commented):
I tried writing an E2E test for this, but for some reason the app was misbehaving when the add contact form was opened, filled in, and then saved (the new contact was not appearing). If we feel like this is something we want to continue on to ensure proper test coverage, I may need to grab someone to help me with this.

ui/components/app/contact-list/contact-list.test.js
```diff
@@ -0,0 +1,60 @@
...
+
+      const wrapper = shallowWithContext(
+        <ContactList
+          searchForContacts={searchForContacts}
+          selectRecipient={selectRecipient}
+          selectedAddress={selectedAddress}
+        />,
+      );
+
+      expect(wrapper).toMatchElement(
```
mcmire, line 32: I opted to use `jest-enzyme`, which provides this matcher, but happy to use something else if we prefer another way.

mcmire: Okay, I think I know why we don't use this library. It seems to have a dependency on an old version of `jest-environment-jsdom` which is majorly screwing up tests. So, I think this isn't a great idea.

test/jest/matchers.js
```diff
@@ -0,0 +1,38 @@
+import { shallow, mount } from 'enzyme';
```
mcmire, line 1: The code in this file is copied from `enzyme-matchers` (https://github.com/enzymejs/enzyme-matchers/blob/master/packages/enzyme-matchers/src/assertions/toMatchElement.js and https://github.com/enzymejs/enzyme-matchers/blob/master/packages/enzyme-matchers/src/utils/isShallowWrapper.js). The reason I copied it is because `jest-enyzme` [relies on an old version of jsdom](https://github.com/enzymejs/enzyme-matchers/issues/351) and including it would break a bunch of tests. Based on the [inactivity of this repo](https://github.com/enzymejs/enzyme-matchers/issues/353) it doesn't seem like this issue will be addressed any time soon. 

In terms of why I chose to add this matcher, I looked through some existing test files and it seems like in most cases we merely check that a component contains other components. This did not work for my case as I needed to check the contents of the ContactList component (albeit shallowly) and thus confirm the sort order of children elements. Happy to change the test below to accommodate another style, however.

ui/components/app/contact-list/contact-list.test.js
mcmire: Okay, I opted to copy this matcher from Enzyme (see comment above for reasoning).

.eslintrc.js
```diff
@@ -129,7 +129,12 @@ module.exports = {
       },
     },
     {
-      files: ['ui/**/*.test.js', 'ui/__mocks__/*.js', 'shared/**/*.test.js'],
+      files: [
+        'ui/**/*.test.js',
+        'ui/__mocks__/*.js',
+        'shared/**/*.test.js',
+        'test/jest/*.js',
```
adonesky1, line 136: ❤️

ui/components/app/contact-list/contact-list.test.js
adonesky1: Great work here! I'm generally in favor of using something like react-test-library more for UI testing but I'm honestly not sure if it can do the check for element ordering in the DOM that we're looking for here... 🤔

.eslintrc.js
adonesky1: Though, per this [conversation](https://github.com/MetaMask/controllers/pull/579#discussion_r700590783) I had yesterday with @rekmarks this probably merits some conversations in the [eslint config repo](https://github.com/MetaMask/eslint-config)

ui/components/app/contact-list/contact-list.component.js
```diff
@@ -1,5 +1,6 @@
 import React, { PureComponent } from 'react';
```
adonesky1, line 1: 👍

ui/components/app/contact-list/contact-list.test.js
mcmire: Yeah! It seems that the JS community has largely embraced react-testing-library over Enyzme, and it seems that react-testing-library has some more TLC around it than Enzyme, so I am not sure whether we are wanting to move in that direction or not? Checking the entire content of a component rather than its pieces doesn't seem to be something that's common for react-testing-library, but perhaps we could make use of [`toContainHTML`](https://github.com/testing-library/jest-dom#tocontainhtml)?

.eslintrc.js
mcmire: Ah! I didn't know that existed. Should I open a PR there? OTOH the `test/jest` directory seems pretty specific to this repo.

adonesky1: Maybe raise the conversation there? I think that's probably the protocol here.

ui/components/app/contact-list/contact-list.test.js
adonesky1: I feel like in this particular case react-testing-library won't work. I wouldn't block this PR on this. Though there might be folks on the team who don't want to introduce new uses of enzyme. Not sure what teamwide feelings are on this. cc @danjm @darkwing @Gudahtt @NiranjanaBinoy @ryanml

Gudahtt: I'm not sure I follow why `react-testing-library` would be ill-suited for this purpose :thinking: . It has an extensive querying API that should allow grabbing the list contents and checking the order.

Gudahtt: >Checking the entire content of a component rather than its pieces doesn't seem to be something that's common for react-testing-library, but perhaps we could make use of toContainHTML?

"Checking the entire contents of a component" sounds like a snapshot to me, which is supported by `react-testing-library`. I don't think it has anything like `toMatchElement` though, as you're doing here. I don't think snapshots are appropriate to use here either, because a snapshot would be more difficult to read.

As far as I understand it, `react-testing-library` doesn't support this style of testing _on purpose_. They want you to test for the things users see, e.g. ensure this text is present, in this order. The focus is on the expected output, not in the implementation details. Whereas this test, as written, looks for a `RecipientComponent` with a specific set of inputs. If the inputs were renamed or the contents of the component restructured, this test might break even under conditions where the list is still sorted alphanumerically, making it fragile.

mcmire: Cool, thanks for those pointers. For some reason I didn't notice we were already using react-testing-library. I reworked the test to use it instead of a custom matcher. Let me know if that's better!

.eslintrc.js
mcmire: Ended up removing the matcher, so this change isn't needed.

ui/components/app/contact-list/contact-list.test.js
adonesky1: Nicely done! It looks like there are a bunch of methods that I either just didn't know about or have been added to `react-testing-library` since I last wrote a true UI test 🤦 I think what you have now looks alot better!

ryanml (approved):
looks great! thanks for the followup changes

adonesky1 (approved):
Awesome work! Thanks @mcmire

PR #11984, opened 2021-09-01
fix state problems with Storybook

EtDu:
Fixes some of the Approval screens missing some state


kumavis (approved):
i dont entirely understand why this is needed but the changes are confined to storybook so im fine

PR #11988, opened 2021-09-01
bump @metamask/controllers to v15.0.2 and remove AbortController workaround in e2e tests

adonesky1:
Explanation:  @metamask/controllers v15.0.0 introduced the use of the global AbortController. The global AbortController was first introduced in Node 15, so this unintentionally broke support for the extension which only requires Node ^14.15.1. @metmask/controllers v15.0.1 introduces a polyfill for AbortController to restore full compatibility with the extension.

This PR also removes [the workaround](https://github.com/MetaMask/metamask-extension/blob/8a8ce3a0c0095fe99764186b15a0d1e70a03e074/app/scripts/metamask-controller.test.js#L112) we introduced in tests to get around the broken contract.

### DIFF
--- a/app/scripts/background.js
+++ b/app/scripts/background.js
@@ -2,9 +2,6 @@
  * @file The entry point for the web extension singleton process.
  */
 
-// polyfills
-import 'abortcontroller-polyfill/dist/polyfill-patch-fetch';
-
 import endOfStream from 'end-of-stream';
 import pump from 'pump';
 import debounce from 'debounce-stream';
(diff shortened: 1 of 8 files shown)


adonesky1:
Still AbortController bugs 😢 `TypeError: c.AbortController is not a constructor`

NiranjanaBinoy (approved):
lgtm

Gudahtt (approved):
LGTM!

PR #11995, opened 2021-09-01
Bump minimum Chrome version to 66

rekmarks:
This PR updates our minimum supported Chrome version from 63 to 66, so that we may use the `AbortController` browser API without polyfilling it.

Our minimum Firefox version supports the `AbortController`, but our current minimum Chrome version (63, released in December 2017) does not. Chrome shipped the `AbortController` in version 66, in April 2018. We have determined that an extremely small number of users are on Chrome 63 < 66, and that this change is therefore acceptable.

### DIFF
--- a/app/manifest/chrome.json
+++ b/app/manifest/chrome.json
@@ -3,5 +3,5 @@
     "matches": ["https://metamask.io/*"],
     "ids": ["*"]
   },
-  "minimum_chrome_version": "63"
+  "minimum_chrome_version": "66"
 }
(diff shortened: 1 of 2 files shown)


Gudahtt (changes requested):
The minimum browser versions are also in the manifest - it should be updated there as well.

Gudahtt (approved):
LGTM!

adonesky1 (approved):
LGTM!

PR #11996, opened 2021-09-01
Fix for #11503: when you send a transaction with value  as 0x, you get a Bignumber error

### LINKED ISSUE #11503: Bignumber not a base 16. Recent new bug with 9.7.1
image

segun:
Fixes: #11503 

Explanation:  
When you send a transaction with 0x as value, you get a Bignumber error

Manual testing steps:  
To recreate the issue, you can go to https://metamask.github.io/test-dapp/ and enter the below code in the developer console:
```
var requestParams = {
    "method": "eth_sendTransaction",
    "params": [
        {
            "data": "0x6abb9a460000000000000000000000000000000000000000000000000000000000000000",
            "value": "0x",
            "gasPrice": "0x746a528800",
            "from": YOUR_ADDRESS_HERE,
            "to": "0xf7a2571ea4d6f478448551f7423085df61bd7cf4",
            "gas": "0xfb7e"
        }
    ],
}
ethereum.request(requestParams);
```
  - 
  - 
  -

### DIFF
--- a/app/scripts/controllers/transactions/index.js
+++ b/app/scripts/controllers/transactions/index.js
@@ -369,10 +369,12 @@ export default class TransactionController extends EventEmitter {
     txMeta.type = type;
 
     // ensure value
-    txMeta.txParams.value = txMeta.txParams.value
-      ? addHexPrefix(txMeta.txParams.value)
-      : '0x0';
-
+    if (txMeta.txParams.value && txMeta.txParams.value !== '0x') {
+      txMeta.txParams.value = addHexPrefix(txMeta.txParams.value);
+    } else {
+      txMeta.txParams.value = '0x0';
+    }
+    
     this.addTransaction(txMeta);
     this.emit('newUnapprovedTx', txMeta);
 
--- a/app/scripts/controllers/transactions/index.js
+++ b/app/scripts/controllers/transactions/index.js
@@ -369,11 +369,7 @@ export default class TransactionController extends EventEmitter {
     txMeta.type = type;
 
     // ensure value
-    if (txMeta.txParams.value && txMeta.txParams.value !== '0x') {
-      txMeta.txParams.value = addHexPrefix(txMeta.txParams.value);
-    } else {
-      txMeta.txParams.value = '0x0';
-    }
+    txMeta.txParams.value = txMeta.txParams.value ? addHexPrefix(txMeta.txParams.value) : '0x0';
     
     this.addTransaction(txMeta);
     this.emit('newUnapprovedTx', txMeta);
--- a/app/scripts/controllers/transactions/index.js
+++ b/app/scripts/controllers/transactions/index.js
@@ -369,8 +369,10 @@ export default class TransactionController extends EventEmitter {
     txMeta.type = type;
 
     // ensure value
-    txMeta.txParams.value = txMeta.txParams.value ? addHexPrefix(txMeta.txParams.value) : '0x0';
-    
+    txMeta.txParams.value = txMeta.txParams.value
+      ? addHexPrefix(txMeta.txParams.value)
+      : '0x0';
+
     this.addTransaction(txMeta);
     this.emit('newUnapprovedTx', txMeta);
 
--- a/app/scripts/controllers/transactions/lib/util.js
+++ b/app/scripts/controllers/transactions/lib/util.js
@@ -1,5 +1,5 @@
 import { ethErrors } from 'eth-rpc-errors';
-import { addHexPrefix, fixHexValue } from '../../../lib/util';
+import { addHexPrefix } from '../../../lib/util';
 import {
   TRANSACTION_ENVELOPE_TYPES,
   TRANSACTION_STATUSES,
@@ -12,7 +12,7 @@ const normalizers = {
   to: (to, lowerCase) =>
     lowerCase ? addHexPrefix(to).toLowerCase() : addHexPrefix(to),
   nonce: addHexPrefix,
-  value: fixHexValue,
+  value: addHexPrefix,
   data: addHexPrefix,
   gas: addHexPrefix,
   gasPrice: addHexPrefix,
@@ -195,6 +195,12 @@ export function validateTxParams(txParams, eip1559Compatibility = true) {
         if (value.toString().includes('.')) {
           throw ethErrors.rpc.invalidParams(
             `Invalid transaction value of "${value}": number must be in wei.`,
+          );
+        }
+
+        if (!value.match(/^0x[a-fA-F0-9]+$/u)) {
+          throw ethErrors.rpc.invalidParams(
+            `Invalid transaction value of "${value}: not an hex string.`,
           );
         }
         break;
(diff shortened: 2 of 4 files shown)


app/scripts/controllers/transactions/index.js
```diff
@@ -369,9 +369,11 @@ export default class TransactionController extends EventEmitter {
     txMeta.type = type;
 
     // ensure value
-    txMeta.txParams.value = txMeta.txParams.value
-      ? addHexPrefix(txMeta.txParams.value)
-      : '0x0';
+    if (txMeta.txParams.value && txMeta.txParams.value !== '0x') {
```
adonesky1, line 372: I think we may want a more robust check here? You could still have a random string that could cause the same issue (I tried it with 
```"params": [
        {
            "data": "0x6abb9a460000000000000000000000000000000000000000000000000000000000000000",
            "value": **"JIEA"**,
            "gasPrice": "0x746a528800",
            "from": '0x2f318C334780961FB129D2a6c30D0763d9a5C970',
            "to": "0xf7a2571ea4d6f478448551f7423085df61bd7cf4",
            "gas": "0xfb7e"
        }
    ],```
    and got the same error)
   
I'm assuming we want to just default to `0x0` for any invalid hex here? @danjm @darkwing @ryanml

Comment (darkwing) [reply app/scripts/controllers/transactions/index.js:372]: Good call @adonesky1 ; I think we can use `isHexString` here.

Comment (brad-decker) [reply app/scripts/controllers/transactions/index.js:372]: there may be a better place to do this logic, namely in the normalizeAndValidateTxParams method and its constituents in app/scripts/controllers/transactions/lib/utils.js. This method is called anytime a transaction is added or updated in tx-state-manager.

**Review by brad-decker — CHANGES_REQUESTED** [review]

File: app/scripts/lib/util.js [hunk 2]
```diff
@@ -139,6 +139,17 @@ const addHexPrefix = (str) => {
...
 
+const fixHexValue = (str) => {
+  const prefixed = addHexPrefix(str);
+  // if it's a valid hex return it immediately
+  if (prefixed.match(/^-?0x[a-fA-F0-9]+$/u)) {
+    return prefixed;
+  }
+
+  // Else return 0x0 for all non valid hex values
+  return '0x0';
```
brad-decker, line 150: I'm okay with this as it, but just throwing it out there that maybe we shouldn't evaluate all improper formatted values to '0x0'. We may want to throw an error and literally only fix 0x to 0x0. @Gudahtt thoughts?

Gudahtt: Why are we even fixing this case? According to the Ethereum JSON-RPC docs, "0x" is not a valid value for the transaction value. See the definition of "QUANTITY".
https://eth.wiki/json-rpc/API

We should return an error to the dapp if they give us invalid parameters.

brad-decker: We can throw the error in the same file, lower in the validation pieces. It won't make it to the point where the big number error happens if we do this resolving the original bug and improving error messaging for dapps. I think that's a win-win

Gudahtt:
This branch includes 147 commits :confused: My guess is that this was messed up during a rebase.

app/scripts/lib/util.js
```diff
@@ -139,6 +139,17 @@ const addHexPrefix = (str) => {
   return `0x${str}`;
 };
 
+const fixHexValue = (str) => {
```
Gudahtt, line 142: Nit: Since we're now throwing an error rather than fixing the value, we should probably rename this. Maybe `validateHexValue` would be better?

Gudahtt: Also, a JSDoc comment block would be great. They will be required soon, when we update the ESLint config to v9

app/scripts/controllers/transactions/lib/util.js
```diff
@@ -12,7 +12,7 @@ const normalizers = {
   to: (to, lowerCase) =>
     lowerCase ? addHexPrefix(to).toLowerCase() : addHexPrefix(to),
   nonce: addHexPrefix,
-  value: addHexPrefix,
+  value: fixHexValue,
```
Gudahtt, line 15: Nit: could we move this to the `validateTxParams` function instead? It would make more sense semantically if it was there, because we aren't normalizing the value anymore.

app/scripts/lib/util.js
```diff
@@ -139,6 +139,17 @@ const addHexPrefix = (str) => {
...
+  const prefixed = addHexPrefix(str);
+  // if it's a valid hex return it immediately
+  if (prefixed.match(/^-?0x[a-fA-F0-9]+$/u)) {
+    return prefixed;
+  }
+
+  // if we get here, throw invalid hex value detected error
+  throw new Error(`Invalid hex value ${str} detected.`);
```
Gudahtt, line 150: We should return one of the standardized Ethereum JSON-RPC errors here instead, like we do for the other error cases in `validateTxParams`. It should be ` ethErrors.rpc.invalidParams` in this case I think.

Gudahtt (commented):
The code looks good to me! But I'm having trouble testing this :thinking: 

I can reproduce the crash, and can verify that it no longer happens now. But I was testing which value formats we accepted before, and it looks like no matter what value the dapp proposes, the transaction confirmation shows a value of "0" (apart from the crash case).

Gudahtt (dismissed):
LGTM! I was able to test this after all. I was confused by the aggressive rounding our confirmation page uses - the value showed up once I made it large enough.

mcmire (dismissed):
Nice and simple! I like it.

app/scripts/controllers/transactions/lib/util.js
```diff
@@ -197,6 +197,12 @@ export function validateTxParams(txParams, eip1559Compatibility = true) {
             `Invalid transaction value of "${value}": number must be in wei.`,
           );
         }
+
+        if (!value.match(/^0x[a-fA-F0-9]+$/u)) {
+          throw ethErrors.rpc.invalidParams(
+            `Invalid transaction value of "${value}: not an hex string.`,
```
adonesky1, line 203:
```suggestion
            `Invalid transaction value of "${value}: not a valid hex string.`,
```

Gudahtt (approved):

app/scripts/controllers/transactions/lib/util.js
```diff
@@ -197,6 +197,12 @@ export function validateTxParams(txParams, eip1559Compatibility = true) {
             `Invalid transaction value of "${value}": number must be in wei.`,
           );
         }
+
+        if (!value.match(/^0x[a-fA-F0-9]+$/u)) {
+          throw ethErrors.rpc.invalidParams(
+            `Invalid transaction value of "${value}: not a valid hex string.`,
```
Gudahtt, line 203: The closing double-quote is missing:
```suggestion
            `Invalid transaction value of "${value}": not a valid hex string.`,
```

Gudahtt (changes requested):

mcmire (approved):

Gudahtt (approved):

brad-decker (approved):

PR #12003, opened 2021-09-02
Cap resubmit transaction retries to within 50 blocks of first submission

adonesky1:
Explanation:  Currently we do not limit the distance between resubmission attempts when a submitted transaction is not confirmed by the network. This can lead to retries many hours apart when a retry might succeed sooner.

### DIFF
--- a/app/scripts/controllers/transactions/pending-tx-tracker.js
+++ b/app/scripts/controllers/transactions/pending-tx-tracker.js
@@ -136,8 +136,8 @@ export default class PendingTransactionTracker extends EventEmitter {
 
     const retryCount = txMeta.retryCount || 0;
 
-    // Exponential backoff to limit retries at publishing
-    if (txBlockDistance <= Math.pow(2, retryCount) - 1) {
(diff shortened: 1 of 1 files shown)


app/scripts/controllers/transactions/pending-tx-tracker.js
```diff
@@ -136,8 +136,8 @@ export default class PendingTransactionTracker extends EventEmitter {
 
     const retryCount = txMeta.retryCount || 0;
 
-    // Exponential backoff to limit retries at publishing
-    if (txBlockDistance <= Math.pow(2, retryCount) - 1) {
+    // Exponential backoff to limit retries at publishing (capped at ~15 minutes between retries)
```
Gudahtt, line 139: 50 blocks seems more like ~11 minutes based on average block time on Mainnet lately :thinking: Ah well, close enough.

Gudahtt (approved):
LGTM!

ryanml (approved):

PR #12004, opened 2021-09-02
tx-list-item - show primaryWarning

kumavis:
Unknown resubmission errors are captured as `txMeta.warning`
this is a sketch of how we could render that warning in the txMeta


brad-decker:
Closing this in favor of documentation via issue #15813 -- but we'd like to keep the branch open for consideration by team to take this work to fruition.

PR #12007, opened 2021-09-02
Added `getPersistentState` support for V2 controllers `persist` metadata

shanejonas:
This PR adds `getPersistentState` support for BaseController V2 to provide persist metadata to localstorage of the extension

### DIFF
--- a/app/scripts/lib/ComposableObservableStore.js
+++ b/app/scripts/lib/ComposableObservableStore.js
@@ -1,4 +1,5 @@
 import { ObservableStore } from '@metamask/obs-store';
+import { getPersistentState } from '@metamask/controllers';
 
 /**
  * @typedef {import('@metamask/controllers').ControllerMessenger} ControllerMessenger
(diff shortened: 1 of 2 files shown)


Gudahtt:
I've scanned through all of our BaseControllerV2-based controllers to get a sense of what the impact of this will be.

* ApprovalController: Not persisted, no impact
* CurrencyRateController: `pendingCurrentCurrency` and `pendingNativeCurrency` will now no longer be stored. Huh. The fact that they were persisted in the first place sounds like a bug :thinking:. In the right conditions, this could show the wrong conversion rate temporarily upon startup. But it's a self-correcting bug, so not a huge deal. And this PR fixes it anyway.
* TokenListController: It's all persisted now, and `persist` is true for all state properties, so no impact.
* GasFeeController: It's all persisted now, and `persist` is true for all state properties, so no impact.

Gudahtt (approved):
LGTM!

rekmarks (approved):
Nice!

PR #12019, opened 2021-09-05
Fix for: Confirmation screen should show local nicknames even when invoking a contract method

### LINKED ISSUE #11148: Confirmation screen should show local nicknames even when invoking a contract method
When sending to a contract, even if the user adds the address to their address book (adding a nickname) in the send view, the confirmation screen does not show the nickname.

Ideally, we always show a local nickname instead of an address if one is available.

segun:
Fixes: #11148 

Explanation:  

1. Display the new address detected dialog if the contract is not in the address book
2. Use the address book if exists, else use the default.

Manual testing steps:  

1. Go to https://metamask.github.io/test-dapp/
2. Click Deploy Contract - You should see recipient as New Contract
3. After deploying contract successfully, click Deposit - You should see contract address as recipient and a dialog that says add to address book
4. Click the dialog, and  provide a nickname
5. The recipient should now show the Nickname
6. Cancel (or continue the transaction, doesn't matter)
7. Click Deposit Again, the Nickname should now show as sender and the dialog should not show.
  - 
  - 
  -

### DIFF
--- a/ui/components/app/confirm-page-container/confirm-page-container.component.js
+++ b/ui/components/app/confirm-page-container/confirm-page-container.component.js
@@ -1,16 +1,20 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
 import SenderToRecipient from '../../ui/sender-to-recipient';
 import { PageContainerFooter } from '../../ui/page-container';
 import EditGasPopover from '../edit-gas-popover';
 import { EDIT_GAS_MODES } from '../../../../shared/constants/gas';
+import { getAddressBookEntry } from '../../../selectors';
+import * as actions from '../../../store/actions';
+import Dialog from '../../ui/dialog';
 import {
   ConfirmPageContainerHeader,
   ConfirmPageContainerContent,
   ConfirmPageContainerNavigation,
 } from '.';
 
-export default class ConfirmPageContainer extends Component {
+class ConfirmPageContainer extends Component {
   static contextTypes = {
     t: PropTypes.func,
   };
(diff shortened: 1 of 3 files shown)


ui/components/app/confirm-page-container/confirm-page-container.component.js
```diff
@@ -203,3 +229,42 @@ export default class ConfirmPageContainer extends Component {
     );
   }
 }
+
+function mapStateToProps(state, ownProps) {
```
adonesky1, line 233: As far as I know we tend to put the "Container" in a separate file and connect/wrap the component there. You can see this pattern throughout the code base, but for just one example look at `add-token.component.js` and `add-token.container.js`. Alternatively I think we generally want, when we have the bandwidth, to migrate the components to all functional, and use the useDispatch and useSelector hooks for interacting with redux.

brad-decker: This component in particular would be an easy win to convert to a functional component. Right now the only place this component is called is confirm-transaction-base.component which is passing in all of its many props.

brad-decker: i'm fine with the separation into .container as well :D

ui/components/app/confirm-page-container/confirm-page-container.component.js
```diff
@@ -149,6 +171,7 @@ export default class ConfirmPageContainer extends Component {
             />
           )}
         </ConfirmPageContainerHeader>
+        <div>{this.maybeRenderAddContact()}</div>
```
brad-decker, line 174: my preference would be to determine whether to show the dialog and assign to variable. then render inline based on that variable. Pseudocode below.

```js
render() {
  const showAddToAddressDialog = contact.name === undefined && toAddress !== undefined;
  return (
   <>{showAddToAddressDialog && (
      <Dialog
        type="message"
        className="send__dialog"
        onClick={showAddToAddressBookModal}
      >
        {t('newAccountDetectedDialogMessage')}
      </Dialog>
   )}
   </>
  );
```

brad-decker (changes requested):
minor feedback, looks good otherwise.

ui/components/app/confirm-page-container/confirm-page-container.component.js
segun: @adonesky1 I've separated the container and the component in a latest commit. Thanks.

adonesky1 (approved):
LGTM! Could be good to add an e2e test for this but not a blocker for me.

brad-decker (approved):
LGTM!

PR #12020, opened 2021-09-05
Adding usage instructions for Ledger Live users to confirmation screen

### LINKED ISSUE #12014: Alert users of necessary steps required prior to confirming tx with Ledger Live bridge
**Describe the bug**
The user of a ledger device on chrome must have everything open and ready to go prior to clicking confirm on MM otherwise it is prone to fail to send the approval request to the hardware device.

**Improvement request**
Add information on the confirmation screen until further improvements or options are made (webhid, etc.)

**Screenshots**
image

ryanml:
Fixes: https://github.com/MetaMask/metamask-extension/issues/12014

Screen Shot 2021-09-07 at 9 55 38 AM

### DIFF
--- a/ui/components/app/confirm-page-container/confirm-detail-row/index.scss
+++ b/ui/components/app/confirm-page-container/confirm-detail-row/index.scss
@@ -58,4 +58,12 @@
       width: 90px;
     }
   }
+
+  .ledger-live-dialog {
+    font-weight: bold;
+
+    &--step {
+      display: block;
+    }
+  }
 }
--- a/ui/components/app/confirm-page-container/confirm-detail-row/index.scss
+++ b/ui/components/app/confirm-page-container/confirm-detail-row/index.scss
@@ -58,12 +58,4 @@
       width: 90px;
     }
   }
-
-  .ledger-live-dialog {
-    font-weight: bold;
-
-    &--step {
-      display: block;
-    }
-  }
 }
--- a/ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
+++ b/ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
@@ -35,6 +35,7 @@ import TransactionDetailItem from '../../components/app/transaction-detail-item/
 import InfoTooltip from '../../components/ui/info-tooltip/info-tooltip';
 import LoadingHeartBeat from '../../components/ui/loading-heartbeat';
 import GasTiming from '../../components/app/gas-timing/gas-timing.component';
+import Dialog from '../../components/ui/dialog';
 
 import {
   COLORS,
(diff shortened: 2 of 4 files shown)


ui/components/app/confirm-page-container/confirm-detail-row/index.scss
```diff
@@ -58,4 +58,12 @@
       width: 90px;
     }
   }
+
+  .ledger-live-dialog {
```
ryanml, line 62: ToDo: Upon design confirmation use `<Typography>` components

mcmire: Nit, but should this be `.confirm-detail-row__ledger-live-dialog` and `.confirm-detail-row__ledger-live-dialog-step` instead, since they're child elements of `.confirm-detail-row` and not components on their own?

mcmire (dismissed):
One question but looks good to me otherwise.

darkwing (dismissed):
The first step is only applicable if the user is *not* using Firefox; we do have an `isFirefox` function we use elsewhere.

Otherwise, I believe this is helpful!

ui/components/app/confirm-page-container/confirm-detail-row/index.scss
ryanml: @mcmire - I've updated the PR to use our design system typography components, so this css has been removed, thanks!

ryanml:
@darkwing - thanks for catching this! I've made a change to hide this first step for FF users

adonesky1 (approved):
Code looks good!

ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
```diff
@@ -396,6 +402,51 @@ export default class ConfirmTransactionBase extends Component {
       </div>
     ) : null;
 
+    const ledgerInstructionField = isLedgerAccount ? (
+      <div>
+        <div className="confirm-detail-row">
+          <Dialog type="message">
+            <div className="ledger-live-dialog">
+              <Typography
```
mcmire, line 410: Wondering if it's worth refactoring this slightly so you don't have to repeat this?

mcmire (approved):
Another nit, but looks good overall.

ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
ryanml: I think there is, though I think that would look like defining an array containing the locale keys for the header and steps, and spitting out the `Typography` components inside a `map` run. There would have to be two conditionals inside, one to prepend `-` for only the steps, then checking if the current key is step one and if so, are we not on firefox. I think given this, though this current way is a bit more verbose, it's also a bit more explicit. 

If you had another idea or would prefer to see that, let me know and I can make the change :)

mcmire: Yeah, that's the same route as I was thinking. No worries, just wondering!

dpazdan (approved):
looks good!!

Issue #12035, opened 2021-09-08
Add padding to footer of custom network form when opened in popup


danjm:
If I open MetaMask in the popup view, and open the custom network form, there is no padding at the bottom. Padding should be added.

![Screenshot from 2021-09-08 07-05-08](https://user-images.githubusercontent.com/7499938/132485972-6d188446-3868-4807-b076-d34f36446e79.png)

To see this view:

https://user-images.githubusercontent.com/7499938/132486128-ecb04b53-1fb9-4f80-bb65-03320cfb02f4.mp4

jpuri:
The issue has been fixed by PR https://github.com/MetaMask/metamask-extension/pull/12039

PR #12039, opened 2021-09-08
Fix padding at bottom of custom network form in popup view

jpuri:
Fixes: #
https://github.com/MetaMask/metamask-extension/issues/12035

Explanation:  
Custom network form when opened in popup view had no padding at bottom. The PR fixes this issue.  This is how it looks after the fix:

Screenshot 2021-09-08 at 9 07 26 PM

Manual testing steps:  
  - Open metamask in popup view
  - Go to Custom RPC
  - Scroll down to the bottom - there should be a clear padding at bottom

### DIFF
--- a/ui/pages/settings/networks-tab/index.scss
+++ b/ui/pages/settings/networks-tab/index.scss
@@ -226,7 +226,7 @@
   &__footer {
     display: flex;
     flex-flow: row nowrap;
-    margin: 0.75rem 0;
+    padding: 0.75rem 0;
 
     @media screen and (max-width: 575px) {
       width: 93%;


adonesky1:
@jpuri it looks like you just need to comment to pass the CLA requirement: 

> I have read the CLA Document and I hereby sign the CLA

ui/pages/settings/networks-tab/index.scss
```diff
@@ -226,7 +226,7 @@
   &__footer {
     display: flex;
     flex-flow: row nowrap;
-    margin: 0.75rem 0;
+    padding: 0.75rem 0;
```
jpuri, line 229: max-height of this section is set to `465px` I found that in case of Custom RPC there is an overflow. margin does not add to overall height of element but padding does and thus padding-bottom is visible in overflow (some context here https://stackoverflow.com/a/21756251). Thus I replaced margin by padding to fix the issue.

Increasing max-height is also an option to fix it, but that might break the view for other networks. Thus I would avoid it.

darkwing (approved):

danjm (approved):
LGTM!

jpuri:

brad-decker (approved):

PR #12042, opened 2021-09-08
Add Max Fee Per Gas to transaction breakdown

### LINKED ISSUE #11879: Show "max fee per gas" in transaction history
Especially for submitted transactions. Could be useful. Should be easy to add just that one field, it's already on every 1559 transaction.

Will be even more useful when we display the current/recent base fee somewhere.

mcmire:
This comes through in EIP-1559 transactions and so it is helpful to show
this to users.

---

Fixes #11879.

Screen Shot 2021-09-08 at 10 50 43 AM


app/_locales/en/messages.json
```diff
@@ -2564,10 +2564,13 @@
     "message": "Transaction Fee"
   },
   "transactionHistoryBaseFee": {
-    "message": "Base fee (GWEI)"
+    "message": "Base Fee (GWEI)"
```
mcmire, line 2567: Opted to titlecase all of the properties in the transaction breakdown to keep things consistent.

ui/components/ui/currency-display/currency-display.component.js
```diff
@@ -53,7 +53,7 @@ CurrencyDisplay.propTypes = {
   'className': PropTypes.string,
   'currency': PropTypes.string,
   'data-testid': PropTypes.string,
-  'denomination': PropTypes.oneOf([GWEI]),
+  'denomination': PropTypes.oneOf([GWEI, ETH]),
```
mcmire, line 56: For some reason I was finding after adding max fee per gas that I was getting an error about ETH not being a valid denomination. Adding ETH to the list fixed this issue, but I am curious whether there is a better way to solve this problem as I do not totally understand how this works.

brad-decker (dismissed):
LGTM

ui/components/ui/currency-display/currency-display.component.js
adonesky1: Yeah that's very odd given that there are other UserPreferencedCurrencyDisplay instances getting passed `ETH` as denomination in the same component...

adonesky1: That being said this should be included here. So I guess propTypes is as unreliable as we've always assumed

adonesky1 (approved):
LGTM!

brad-decker (approved):

ui/components/ui/currency-display/currency-display.component.js
mcmire: Indeed :)

PR #12043, opened 2021-09-08
Fix e2e to not check for X when running locally

mcmire:
The X server check was only added to make Chrome tests run successfully
on CI.


mcmire:
This is a dev fix, so if you can run some or all of Chrome e2e tests locally then this should be good to go.

PeterYinusa (approved):

brad-decker (approved):
LGTM

PR #12047, opened 2021-09-08
Rationalize build system arguments

rekmarks:
This rationalizes how arguments are passed to and parsed by the build system. To accomplish this, everything that isn't an environment variable from `.metamaskrc` or our CI environment is now passed as an argument on the command line.

Of such arguments, the `entryTask` is still expected as a positional argument in the first position (i.e. `process.argv[2]`), but everything else must be passed as a named argument. We use `minimist` to parse the arguments, and set defaults to preserve existing behavior.

Arguments are parsed in a new function, `parseArgv`, in `development/build/index.js`. They are returned from `parseArgv` to be passed to other functions invoked in the same file.

This change is motivated by our previous inconsistent handling of arguments to the build system, which will grow increasingly problematic as the build system grows in complexity. (Which it will very shortly, as we introduce Flask builds.)

Miscellaneous changes:
- Adds a build system readme at a`development/build/README.md`
- Removes the `beta` package script. Now, we can instead call: `yarn dist --build-type beta`
- Fixes the casing of some log messages in the build system


kumavis (dismissed):

development/build/README.md
```diff
@@ -0,0 +1,37 @@
...
+Here follows basic usage information for the build system.
+
+```text
+Usage: yarn build <entry-task> [options]
+
+Commands:
+  yarn build prod       Create a production-ready build
+  yarn build dev        Create a build for local development
+  yarn build test       Create a build for running e2e tests
+  yarn build testDev    ?
```
rekmarks, line 23: @kumavis @EtDu What is `testDev` for?

kumavis: live reloading (e2e?) test development

development/build/index.js
```diff
@@ -117,4 +133,77 @@ function defineAllTasks() {
...
+
+function parseArgv() {
+  const NamedArgs = {
+    BetaVersion: 'beta-version',
+    BuildType: 'build-type',
+    OmitLockdown: 'omit-lockdown',
+    SkipStats: 'skip-stats',
+  };
+
+  const entryTask = process.argv[2];
```
kumavis, line 149: does minimist provide positional args parsing we can use instead

development/build/index.js
```diff
@@ -117,4 +133,77 @@ function defineAllTasks() {
...
+  if (!entryTask) {
+    throw new Error('MetaMask build: No entry task specified.');
+  }
+
+  // The entry task should never be prefixed with '-' or '--', so that is easily
+  // distinguishable from other arguments.
+  if (entryTask.startsWith('-')) {
+    throw new Error(`MetaMask build: invalid entry task: ${entryTask}`);
```
kumavis, line 157:
```suggestion
    throw new Error(`MetaMask build: invalid entry task: "${entryTask}"`);
```

development/build/index.js
rekmarks: Not as far as I can tell: https://www.npmjs.com/package/minimist

development/build/README.md
rekmarks: Ah, I see it now. Thanks!

development/build/index.js
```diff
@@ -117,4 +133,77 @@ function defineAllTasks() {
...
+  });
+
+  const betaVersion = argv[NamedArgs.BetaVersion];
+  if (!Number.isInteger(betaVersion) || betaVersion < 0) {
+    throw new Error(`MetaMask build: Invalid beta version: ${betaVersion}`);
+  }
+
+  // Arguments that are used throughout or deeply within the build system are
+  // set as environment variables.
+  setImmutableProperties(process.env, [
```
kumavis, line 178: id rather not if we can just put it in a configuration blob and pass that around

development/build/index.js
rekmarks: My practice is to double-quote substitution values in template strings, _unless_ they're the last token of the string and 
follow a colon. Is it time for me to change my ways? 🤔

development/build/index.js
kumavis: its the underbar, eh `{ _: [ 'foo', 'bar', 'baz' ], `

development/build/index.js
rekmarks: image

rekmarks: That's too much refactoring / tracing through the build system than I'd like to do right now, if you're OK with that. In addition, these environment variables are just replacing other environment variables we were already using. Now we just define them in a single spot.

development/build/index.js
rekmarks: Yeah! Which seems, roughly equivalent to what I'm doing?

Hm, actually. I can use the underbar to make this validation stricter. Hold on.

rekmarks: 2ff827c

development/build/index.js
```diff
@@ -4,12 +4,13 @@
 // run any task with "yarn build ${taskName}"
 //
 const livereload = require('gulp-livereload');
+const minimist = require('minimist');
```
mcmire, line 7: It looks like we already use yargs; is there a reason for using minimist over yargs? Should we be switching to minimist across the board?

development/build/index.js
rekmarks: In other words, it's an improvement! 😄

development/build/index.js
kumavis: i like wrapping things so its clear if theres a trailing space or something

development/build/index.js
rekmarks: Ah, I didn't realize we were using `yargs` in the e2e tests. Interesting.

I chose `minimist` for two reasons:
1. It's dead simple, and I didn't want to go through the trouble of giving the build system the complete `yargs` treatment right now.
2. It was already in our dependency graph

Consider this a step in the right direction!

kumavis: i use yargs in lavamoat, with mixed feelings

mcmire: Cool, no worries!

development/build/index.js
rekmarks: I'm changing my ways. I'm never going back.

a04f202

development/build/index.js
rekmarks: I use `yargs` in `snaps-cli`, too! I think it's pretty great, but I have bigger fish to fry right now. I'll mark this resolved.

development/build/index.js
rekmarks: Alright! Fine! I did it! 604e82e

README.md
```diff
@@ -28,6 +28,8 @@ To learn how to contribute to the MetaMask project itself, visit our [Internal D
 
 Uncompressed builds can be found in `/dist`, compressed builds can be found in `/builds` once they're built.
 
+See the [build system readme](./development/build/README.md) for build system usage information.
```
kumavis, line 31: maybe add a tldr saying "yarn start for dev yarn dist for prod"

kumavis (dismissed):
seems rational

README.md
rekmarks: 231e093

darkwing (dismissed):

development/build/index.js
```diff
@@ -177,12 +177,6 @@ function parseArgv() {
     throw new Error('MetaMask build: No entry task specified.');
   }
 
-  // The entry task should never be prefixed with '-' or '--', so that is easily
-  // distinguishable from other arguments.
-  if (entryTask.startsWith('-')) {
-    throw new Error(`MetaMask build: invalid entry task: "${entryTask}"`);
-  }
-
```
rekmarks: Redundant since we started grabbing positionals from the `minimist` `argv` underbar property. It doesn't put `-`-prefixed arguments in that array.

kumavis (approved):
i like it

darkwing (approved):

PR #12048, opened 2021-09-08
remove unused sidebar component

brad-decker:
The Sidebar component and app state are no longer used, the only place it was previously used was as part of the old cancel/speedup flow. This has since been replaced by the EIP-1559 experience.


NiranjanaBinoy (dismissed):
LGTM

ui/store/actions.js
```diff
@@ -1636,6 +1636,7 @@ export function showModal(payload) {
     payload,
   };
 }
+// comment
```
Gudahtt, line 1639: I can confirm that this is a comment

adonesky1 (dismissed):
LGTM! Just looks like @Gudahtt has nit

ui/store/actions.js
brad-decker: hahahaha

darkwing (dismissed):

ui/store/actions.js
brad-decker: removed in latest.

darkwing (approved):

adonesky1 (approved):

PR #12053, opened 2021-09-09
Support for GridPlus Lattice1 hardware wallet

alex-miller-0:
## Overview

Hello 👋  

This PR adds support for the GridPlus [Lattice1 hardware wallet](https://gridplus.io/lattice). I tried to be as surgical as possible and to keep existing styles in mind. The code is also linted. 

GridPlus has been managing a [fork](https://github.com/GridPlus/metamask-extension) of the MetaMask extension and has been keeping it up to speed with official MetaMask updates since December 2020. All (or nearly all) Lattice users use our MetaMask fork. It is unlisted but can be found in our [setup page](https://gridplus.io/setup), which is also a good resource for understanding how the Lattice works.

This PR includes code to utilize the [`eth-lattice-keyring`](https://github.com/GridPlus/eth-lattice-keyring), which adheres to the Keyring API and uses [`gridplus-sdk`](https://github.com/GridPlus/gridplus-sdk) under the hood. If you're interested in how MetaMask communicates with the Lattice at a protocol level, you should check out the SDK and how it is used with the keyring.

I'll now go through a quick tutorial of how a new Lattice user would interact with MetaMask.

### 1. Pairing with a Lattice

The first step is to establish a persistent, secure connection between the requesting app (MetaMask) and a user's Lattice device. This is called "pairing" because it uses out-of-band data entry. The Lattice is connected to wifi  and there is no physical link to the user's computer, so this pairing process happens wirelessly. Once pairing finishes, all subsequent messages sent to the Lattice are end-to-end encrypted.

First, the user chooses the Lattice as a hardware wallet:

<p align="center">
  
</p>

and the keyring module uses `window.open` to [create a window](https://github.com/GridPlus/eth-lattice-keyring/blob/master/index.js#L412) pointing to a `gridplus.io` page:

<p align="center">
  
</p>

The Lattice has a unique "Device ID" which the user can find on the device's main menu. It is a random 6-character code and is used to build a message routing endpoint (if you're interested in how the routing works, you should read [this article](https://blog.gridplus.io/connecting-to-your-lattice1-in-private-41587c43f635)). It is not meant to be secure because all messages are end-to-end encrypted. 

On the new page the user enters that device ID plus a "password", which is really a salt:

<p align="center">
  
</p>

***Note:** the "password" isn't a really critical piece of data. If the user loses it, they can delete the connection on the device and refresh MetaMask -- it will go through this pairing process again.*

Pressing `Connect` will contact the Lattice, which will generate a random, 8 character code and display that on its screen. This is generated using a TRNG on-device and must be entered by the user out of band on the login page within 60 seconds of it being created (if it times out there is a retry button):

<p align="center">
  
</p>

<p align="center">
  
</p>

At this point, the `gridplus.io` window will close and the user will go back to MetaMask, which now has a secure connection to the user's Lattice device, and is therefore "paired".

***Note:** this connection is to the **device**, not the **wallet**. The Lattice device allows users to switch between wallets using [SafeCards](https://docs.gridplus.io/introduction/introduction-to-safecards). Basically, a user has a default wallet on the Lattice itself, but if they insert and unlock a SafeCard with a PIN, the Lattice device will switch from its on-board wallet to the SafeCard wallet such that all requests from securely connected external apps will now be made against the SafeCard wallet. Once the card is removed, it goes back to the on-board Lattice wallet.*

### 2. Selecting Accounts

Once connected, a Lattice user can select one of several derivation paths:

<p align="center">
  
</p>

They can also import as many accounts they want:

<p align="center" float="left">
  
  
</p>

***Note:** This set of accounts is associated with whatever wallet was active on my Lattice when I loaded the address set. If this was an on-board Lattice wallet and I suddenly switched to a SafeCard, this is equivalent to using a separate Ledger device. The user in this case would need to remove the original accounts and refresh the connection to the device. Importantly, though, they would **not** need to go through the pairing process because the secure connection is persistent.*

### 3. Requesting Signatures

Once setup, the user can start using MetaMask like they normally would. The Lattice supports all Ethereum transactions as well as EIP712 and `personal_sign` messages. It also supports all EVM-based chains/execution environments that MetaMask already supports.

The MetaMask window is unchanged relative to the official extension:

<p align="centeR" float="left">
  
  
</p>

The Lattice displays EIP712 messages as JSON objects:

<p align="center">
  
</p>

By default it displays hex code for Ethereum calldata but it is capable of decoding that using an ABI (more in [this article](https://blog.gridplus.io/readable-ethereum-transactions-a-new-standard-945c5e9ef2c7)):

<p align="center" float="left">
  
  
</p>

## Conclusion

I hope this helps show how the Lattice device interacts with MetaMask. Our users have generally given positive UX feedback and our unofficial integration has been in production for over 9 months so we have quite a bit of confidence behind it at this point.

Please let me know if there's anything I can clarity here and of course I look forward to seeing feedback on the code itself. Cheers! 🍻


alex-miller-0:

ui/pages/create-account/connect-hardware/select-hardware.js
```diff
@@ -129,6 +152,7 @@ export default class SelectHardware extends Component {
         return this.renderLedgerTutorialSteps();
       case 'trezor':
         return this.renderTrezorTutorialSteps();
+      case 'lattice': // TODO: Add Lattice tutorial
```
danjm, line 155: Do you intent on resolving this TODO before this PR is merged? Or is this a reminder for the future?

alex-miller-0: This is up to your design team. We didn't want to use the existing Ledger/Trezor asset because it showed a USB device being inserted into a laptop.

We're more than happy to leave the tutorial section empty and have this note as a future reminder but we can also coordinate on an asset if that's important on your side.

rachelcope:
Looks good! A couple questions/comments on the design:

- On the "Select an Account" view, when user clicks the “Select the HD Path” dropdown, does the “Switch Ledger Path” modal open? Is there a reason for this? It seems like the dropdown should just drop down under the dropdown field instead of opening a modal. 

- What happens if a user approves a confirmation in MetaMask, but fails to approve on Lattice? Is the transaction “pending” in the Activity? Should there be a message indicating that a Lattice approval is required to complete the transaction?

ui/pages/create-account/connect-hardware/select-hardware.js
rachelcope: Where in the design is the Lattice tutorial? I don't see it in the screenshots

alex-miller-0: There is no Lattice tutorial at the moment. This line of code doesn't do anything and was just meant to be a placeholder for discussion.

From the GridPlus perspective we don't **need** a tutorial, but if you all want us to have one we're happy to work with you on that. I do think we would need a new asset to match the style of the Ledger/Trezor one -- we don't have anything internally that would work.

alex-miller-0:
> * On the "Select an Account" view, when user clicks the “Select the HD Path” dropdown, does the “Switch Ledger Path” modal open? Is there a reason for this? It seems like the dropdown should just drop down under the dropdown field instead of opening a modal.

The modal is coming from the `click here` link. We found that the main dropdown selector didn't update state in MetaMask and we didn't understand why, but the modal works so we left it alone. Maybe @danjm can shed more light on the origin of this `click here` link.

> * What happens if a user approves a confirmation in MetaMask, but fails to approve on Lattice? Is the transaction “pending” in the Activity?

If the user rejects the transaction on the Lattice or if it gets rejected for some reason, the error will bubble up to MetaMask, so it gets thrown like a MetaMask error:

image

> Should there be a message indicating that a Lattice approval is required to complete the transaction?

This isn't something we'd considered. Does it happen for Ledger or Trezor? I tried to replicate the flow of Ledger/Trezor -- did I miss a message?

alex-miller-0:
We updated the keyring module today and I included it in this PR. I can squash these commits before we merge but for now I'll leave it as a new commit.

The update was just to shim in the old `signMessage` pattern (before `signTypedData` existed) because we found an old app (perp.exchange) that was still using it: https://github.com/GridPlus/eth-lattice-keyring/releases/tag/v0.3.2

hrharder:
is this stalled? would love to have official support for the Lattice in Metamask!

alex-miller-0:
We've updated the `gridplus-sdk` again and I added a new commit to edit `yarn.lock`. Again, this should be squashed. Just waiting on feedback before I start squashing commits, but happy to do that whenever you all feel it is appropriate.

danjm (commented):
Code looks good. I will wait for QA approval before final approval and merge

alex-miller-0:
We realized we never added Firefox support, so we went ahead and did that in the commit above ^

alex-miller-0:
Added a missing error for Firefox. I can squash those two commits as well but just want everyone to see them first.

danjm (approved):

PeterYinusa (approved):

PR #12058, opened 2021-09-09
Truncate long network names

### LINKED ISSUE #10544: Spanish version of the extension overlaps network name over the "fox icon" home button
metamask_issue_thumb

The new update of the Spanish name of the main network causes problems because the text string is very long, and invades the space of the upper left button (the Logo button), overlaying it and preventing its selection.

mcmire:
In some languages, network names are longer than the visible width of
the network dropdown. In these cases, the currently selected network will
overflow the bounds of the network dropdown container.

This commit addresses this issue so that the dropdown
container will grow and shrink appropriately, truncating the currently
selected network if needed.

Fixes: #10544

## Before

Screen Shot 2021-09-09 at 4 16 59 PM

## After

https://user-images.githubusercontent.com/7371/132901182-7b00994e-b888-41df-bdb5-6833206a729e.mov

### DIFF
--- a/ui/components/app/network-display/index.scss
+++ b/ui/components/app/network-display/index.scss
@@ -35,6 +35,11 @@
     background-color: lighten($dodger-blue, 35%);
   }
 
+  &.chip {
+    margin: 0;
+    max-width: 100%;
+  }
+
   & .chip__label {
     overflow: hidden;
     text-overflow: ellipsis;
--- a/ui/components/app/app-header/index.scss
+++ b/ui/components/app/app-header/index.scss
@@ -72,6 +72,7 @@
     flex-direction: row;
     align-items: center;
     flex: 0 0 auto;
+    margin-right: 1rem;
 
     &--clickable {
       cursor: pointer;
--- a/ui/css/itcss/components/network.scss
+++ b/ui/css/itcss/components/network.scss
@@ -73,9 +73,6 @@
 .network-name-item {
   flex: 1;
   color: $dusty-gray;
-  text-overflow: ellipsis;
-  overflow: hidden;
-  white-space: nowrap;
 }
 
 .network-check,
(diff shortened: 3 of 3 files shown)


ui/components/app/network-display/index.scss
```diff
@@ -35,6 +35,11 @@
     background-color: lighten($dodger-blue, 35%);
   }
 
+  &.chip {
```
mcmire, line 38: Ideally it shouldn't be necessary to qualify these styles with `.chip`. As I understand it, Chip is a more abstract component than NetworkDisplay. But `ui` styles get imported after `app` styles, so because Chip defines a `margin` property, that will override anything we put here. We need `.chip` so that these styles will win.

development/lib/run-command.js
```diff
@@ -1,4 +1,5 @@
 const spawn = require('cross-spawn');
+const shellwords = require('shellwords');
```
brad-decker, line 2: I think this is an unrelated change?

ui/components/app/network-display/index.scss
brad-decker: Yeah, I opted for alphabetical ordering of imports before fully understanding the impact it would have. I do prefer us to have *some* scheme for ordering imports, though. Increasing specificity is fine.

brad-decker (changes requested):

brad-decker:
One other thing... I think removing the truncation in the drop down is a mistake. Instead, we should use the title property to provide the full length name on hover. I could be wrong tho. cc @MetaMask/design

development/lib/run-command.js
mcmire: Yeah, I just was trying to debug CI failures better, but will split this off into its own PR.

mcmire:
@brad-decker Okay. I can remove that change until we have a better approach on that, since it's technically out of scope of the original issue.

brad-decker (approved):
LGTM :shipit:

NiranjanaBinoy (approved):

darkwing (approved):

PR #12060, opened 2021-09-10
Add build-time code exclusion using code fencing

rekmarks:
This PR adds build-time code exclusion by means of code fencing. For details, please see the [README](https://github.com/MetaMask/metamask-extension/blob/code-fencing/development/build/transforms/README.md). Note that linting of transformed files as a form of validation is added in a follow-up, #12075.

Hopefully exhaustive tests are added to ensure that the transform works according to its specification. Since these tests are Node-only, they required their own Jest config. The recommended way to work with multiple Jest configs is using the `projects` field in the Jest config, however [that feature breaks coverage collection](https://github.com/facebook/jest/issues/9628). That being the case, I had to set up two separate Jest configs. In order to get both test suites to run in parallel, Jest is now invoked via a script, `./test/run-jest.sh`.

By way of example, this build system feature allows us to add fences like this:

```javascript
this.store.updateStructure({
  ...,
  GasFeeController: this.gasFeeController,
  TokenListController: this.tokenListController,
  ///: BEGIN:ONLY_INCLUDE_IN(beta)
  PluginController: this.pluginController,
  ///: END:ONLY_INCLUDE_IN
});
```

Which at build time are transformed to the following if the build type is not `beta`:

```javascript
this.store.updateStructure({
  ...,
  GasFeeController: this.gasFeeController,
  TokenListController: this.tokenListController,
});
```


development/build/transforms/code-fencing.js
```diff
@@ -0,0 +1,288 @@
...
+  // "next" is called at the end of the transform operation to inform browserify
+  // that it can continue to the next transform.
+  return function (fileName) {
+    if (!['.js', '.cjs', '.mjs'].includes(path.extname(fileName))) {
+      return new PassThrough();
+    }
+
+    return through(function (fileBuffer, _encoding, next) {
+      this.push(
+        removeFencedCode(fileName, buildType, fileBuffer.toString('utf8')),
```
kumavis, line 41: this makes assumptions on chunk contents. i reccomend operating on the whole file instead of chunks. as seen in the example i gave @rekmarks

kumavis: eg chunking barrier could happen in the middle of a fence identifier

kumavis (changes requested):
use the example luke 🧙‍♂️

kumavis:
generally good

darkwing (dismissed):
This looks really good and my testing was successful.  Well done, and very quickly at that!

development/build/transforms/remove-fenced-code.js
```diff
@@ -0,0 +1,416 @@
...
+ *   - `.mjs`
+ *
+ * For details on how the transform mutates source files, see
+ * {@link removeFencedCode} and the documentation.
+ *
+ * @param {string} buildType - The type of the current build.
+ * @returns {(filePath: string) => Transform} The transform function.
+ */
+function createRemoveFencedCodeTransform(buildType) {
+  if (!(buildType in BuildTypes)) {
```
kumavis: in keyword on objects includes the prototype chain eg `'toString' in {} // true`

jest.config.js
```diff
@@ -11,10 +11,11 @@ module.exports = {
...
     },
   },
-  setupFiles: ['./test/setup.js', './test/env.js'],
-  setupFilesAfterEnv: ['./test/jest/setup.js'],
-  testMatch: [
-    '<rootDir>/ui/**/?(*.)+(test).js',
-    '<rootDir>/shared/**/?(*.)+(test).js',
-  ],
+  // TODO: enable resetMocks
+  // resetMocks: true,
```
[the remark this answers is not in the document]
rekmarks: `resetMocks` was already not enabled, and enabling it caused errors. I'm just highlighting that we should enable it here!

kumavis (dismissed):
lgtm 🔥

jest.config.js
```diff
@@ -11,10 +11,11 @@ module.exports = {
...
+  restoreMocks: true,
+  setupFiles: ['<rootDir>/test/setup.js', '<rootDir>/test/env.js'],
+  setupFilesAfterEnv: ['<rootDir>/test/jest/setup.js'],
+  testMatch: ['<rootDir>/ui/**/*.test.js', '<rootDir>/shared/**/*.test.js'],
+  testTimeout: 2500,
```
Gudahtt, line 20: This looks new - is this the default setting?

test/run-jest.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# For storing PIDs of subshells
+PIDS=(0 0)
+
+yarn 'jest' '--config=./jest.config.js' "$@" & PIDS[0]=$!
+yarn 'jest' '--config=./development/jest.config.js' "$@" & PIDS[1]=$!
+
+# This ensures that this shell exists with a non-zero code if any of the
+# subshells we waited for did
+for pid in ${PIDS[*]}; do
```
Gudahtt, line 16: This is giving ShellCheck some difficulty due to word splitting. Here you're relying on word-splitting for the array elements to be correctly distinguished.  `PIDS[*]` will spit out the array as one space-delimited string, and then in context it'll be interpreted as an array due to word splitting. But if one of the elements has a space, it'll be wrongly interpreted as two elements.


Here is the recommended way to loop through an array in Bash:
```suggestion
for pid in "${PIDS[@]}"; do
```
The quotes protect against word splitting, and the `@` will give the entire array rather than the stringified representation of it.

test/run-jest.sh
```diff
@@ -0,0 +1,18 @@
...
+
+set -x
+set -e
+set -u
+set -o pipefail
+
+# For storing PIDs of subshells
+PIDS=(0 0)
+
+yarn 'jest' '--config=./jest.config.js' "$@" & PIDS[0]=$!
```
Gudahtt, line 11: This pattern seems questionable :thinking: This means if one process fails, the other will keep going and there will be no way to stop it. Sounds like a pain.

We have a package that is supposed to do this sort of thing for us: `concurrently`. I don't think it works correctly in all cases, but it might work more often than this.

I think we could scrap this whole script and call `concurrently` directly from the npm script.

jest.config.js
rekmarks: We bumped the timeout at one point in the module template: https://github.com/MetaMask/metamask-module-template/blob/main/jest.config.js#L26

test/run-jest.sh
rekmarks: > This means if one process fails, the other will keep going and there will be no way to stop it. Sounds like a pain.

This is an intentional feature, not a bug! The idea being that, in CI, you want to know of all unit test failures, not just the ones for a particular config. This is an attempt to replicate what the Jest `projects` feature does, which is to run all tests for every project and then print the complete results.

If developing locally, the tests for a particular config can be run by doing `yarn jest --config=PATH_TO_CONFIG`.

On the topic of using `concurrently`, that seems like a good idea, since it appears to be a well-maintained tool that does exactly what I'm trying to do. We may still have to keep this script in order to do something like `yarn test:unit:jest --watch` and pass the `--watch` argument to both invocations of `jest`. That's what the `"$@"` is about in the Bash script.

test/run-jest.sh
rekmarks: Hm, my Shellcheck linting plugin appears to not tell me about warnings? Good to know.

Fixed: 073f7cf

development/build/transforms/remove-fenced-code.js
```diff
@@ -0,0 +1,425 @@
...
+            `"${param}" is not a valid build type.`,
+          ),
+        );
+      }
+    });
+  },
+};
+
+// Matches lines starting with "///:", optionally preceded by whitespace
+const linesWithFenceRegex = /^.*\/\/\/:.*$/gmu;
```
Gudahtt, line 124: This regex doesn't appear to do what is described. First off, it matches on _any_ preceding characters, not just whitespace. But more importantly, this doesn't match lines. The `^` and `$` assertions match the beginning and end of the _input_, not the _line_. And since this is a multi-line regex, this will return multi-line results.

test/run-jest.sh
Gudahtt: Just to be clear, when I say "there is no way to stop it", what I mean is that the interrupt signal no longer halts the script. I certainly hope disabling CTRL+C wasn't intentional! It's incredibly frustrating when scripts can't be halted.

development/build/transforms/remove-fenced-code.js
Gudahtt: I'd recommending using something more like this:

```suggestion
// Matches lines starting with "///:", optionally preceded by whitespace
const lineWithFenceRegex = /^\s*\/\/\/:.*$/u;
```

and doing the line-splitting separately, before using this regex.

rekmarks: Great point on the white space. I'm on mobile so I can't check right now, but I can't see how my tests would patch if it matched the beginning and end of the entire input.

rekmarks: The `m` flag should make the caret and dollar sign match the beginning and end of lines: https://javascript.info/regexp-multiline-mode

development/build/transforms/remove-fenced-code.js
```diff
@@ -0,0 +1,425 @@
...
+ * For details on how the transform mutates source files, see
+ * {@link removeFencedCode} and the documentation.
+ *
+ * @param {string} buildType - The type of the current build.
+ * @returns {(filePath: string) => Transform} The transform function.
+ */
+function createRemoveFencedCodeTransform(buildType) {
+  if (!hasOwnProperty(BuildTypes, buildType)) {
+    throw new Error(
+      `Metamask build: Code fencing transform received unrecognized build type "${buildType}".`,
```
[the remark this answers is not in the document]
rekmarks: Oh, my bad! Hangover from writing other build system errors (the properly cased prefix is ubiquitous).

development/build/transforms/remove-fenced-code.js
Gudahtt: Oh, oops, yes you're right. I actually tested this and got the correct results but misread them them. :facepalm:

test/run-jest.sh
Gudahtt: ... I tested this, and it works :thinking: I guess it's because these tests don't spawn child processes? This was a struggle to get working with the e2e tests.. Our attempts to combine tests just with bash resulted in them continuing on after interrupting the main process.

development/build/transforms/remove-fenced-code.js
```diff
@@ -0,0 +1,425 @@
...
+  // Get the first part to be included
+  // The substring() call returns an empty string if splicingIndices[0] is 0,
+  // which is exactly what we want in that case.
+  retainedSubstrings.push(toSplice.substring(0, splicingIndices[0]));
+
+  // This loop gets us all parts of the string that should be retained, except
+  // the first and the last.
+  // It iterates over all "end" indices of the array except the last one, and
+  // pushes the substring between each "end" index and the next "begin" index
+  // the the array of retained substrings.
```
Gudahtt, line 373: Typo:
```suggestion
  // t the array of retained substrings.
```

development/build/transforms/remove-fenced-code.js
rekmarks: Where is my "😅" react?!

development/build/transforms/remove-fenced-code.js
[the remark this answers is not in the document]
rekmarks:
```suggestion
      `Code fencing transform received unrecognized build type "${buildType}".`,
```

development/build/transforms/remove-fenced-code.js
rekmarks:
```suggestion
// Matches lines starting with "///:", optionally preceded by whitespace
const linesWithFenceRegex = /^\s*\/\/\/:.*$/gmu;
```

development/build/transforms/remove-fenced-code.test.js
```diff
@@ -0,0 +1,564 @@
...
+      // beginning with "///: TERMINUS"
+      const fenceSentinelAndTerminusRegex = /^\/\/\/: \w+/mu;
+
+      const replacements = [
+        '///:BEGIN',
+        '///:XBEGIN',
+        '///:_BEGIN',
+        '///:B',
+        '///:_',
+        '///: ',
```
Gudahtt, line 209: Nit: It would be nice to include a single-character alphabetical test case here. The description isn't met without it.
```suggestion
        '///: ',
        '///: B',
```

development/build/transforms/remove-fenced-code.js
Gudahtt: Or `onto`? :thinking: 
```suggestion
  // onto the array of retained substrings.
```

development/build/transforms/remove-fenced-code.test.js
rekmarks: 474ec02

test/run-jest.sh
rekmarks: After some deliberation, I decided to go with `concurrently` but keep the bash script. I went for `concurrently` because a tool we're already using is probably better than a weird Bash loop-and-`wait` implementation. However, I invoke `concurrently` in the `.sh` file to preserve the ability to pass arguments to both `jest` processes by doing e.g. `yarn test:unit:jest --silent`.

662f5ae

rekmarks:
892a0a1 fixes a RegEx-related bug introduced in ce15bf2, and a completely different bug discovered while trying to fix the RegEx bug.

The RegEx bug was due to (what seems to me) somewhat strange behavior by the JavaScript RegEx engine (or just the spec, who knows) in that `/s*\w+/gmu` will match `\nstuff\n` as the single line `\nstuff`. The RegEx had to be modified to avoid this and still produce the desired effect.

While writing tests to ensure that the parser indeed ignores sentinels preceded by non-whitespace characters (i.e. what ce15bf2 was supposed to accomplish), I discovered that the `multiSplice` function didn't handle the case where `splicingIndices.length === 2`. Specifically, it would insert the last retained part of the original string twice. This has now been fixed.

Gudahtt (approved):
LGTM!

brad-decker (approved):
This is great. Good work. Tested locally and it checks out.

development/build/transforms/README.md
```diff
@@ -0,0 +1,123 @@
+# Local Browserify Transforms
```
Gudahtt, line 1: This is some _incredible_ documentation by the way. Not just this but the inline docs too.

I was expecting this transform to be difficult to understand, but it was really straightforward with all of these explanations at each step.

brad-decker: +1000

PR #12061, opened 2021-09-10
Ensuring only valid numerical inputs for gasPrice and gasLimit

### LINKED ISSUE #11940: BigNumber Error: new BigNumber() not a base 16 number
**Steps to reproduce**

1. Login to the extension
2. Switch to a custom network e.g. `Binance Smart Chain`
3. Click the Send button on the Home page
4. Select a recipient on the Add Recipient page
5. Prepend a `-` to the Gas Price on the Send page

**Expected behavior**

- Error class on the `Gas Price (GWEI)` field
- Next button is disabled

**Error Messages**

- BigNumber Error: new BigNumber() not a base 16 number: 0x-6fc23ac00

```
BigNumber Error: new BigNumber() not a base 16 number: 0x-6fc23ac00
  at ee(metamask/common-0.js:18:72957)
  at S(metamask/common-0.js:18:78437)
  at new z(metamask/common-0.js:18:70451)
  at T(metamask/common-1.js:1:190480)
  at P(metamask/common-1.js:1:191128)
  at n.calcGasTotal(metamask/common-1.js:1:288315)
  at Object.calculateGasTotal(metamask/common-3.js:12:22758)
  at updateGasFees(metamask/common-3.js:12:23626)
  at ? (metamask/common-4.js:10:302433)
  at produce(metamask/common-3.js:19:122502)
  at ? (metamask/common-4.js:10:302410)
  at Array.reduce(<anonymous>)
  at ? (metamask/common-4.js:10:302279)
  at ? (metamask/common-5.js:27:14440)
  at y(metamask/common-5.js:27:15359)
  at ? (metamask/common-5.js:27:16149)
  at dispatch(metamask/common-5.js:27:13586)
  at ? (metamask/common-3.js:12:12533)
  at ? (metamask/common-5.js:27:16140)
  at updateGasPrice(metamask/ui-5.js:10:153760)
  at Object.updateCustomGasPrice(metamask/ui-5.js:10:30783)
  at d.<anonymous>(metamask/ui-5.js:10:328735)
  at m(metamask/common-0.js:1:72601)
  at v(metamask/common-0.js:1:72852)
  at _(metamask/common-0.js:1:72754)
  at a(metamask/sentry-install.js:1:10465)
```

ryanml:
Fixes: https://github.com/MetaMask/metamask-extension/issues/11940

### DIFF
--- a/ui/components/app/gas-customization/advanced-gas-inputs/advanced-gas-inputs.component.js
+++ b/ui/components/app/gas-customization/advanced-gas-inputs/advanced-gas-inputs.component.js
@@ -52,22 +52,32 @@ export default class AdvancedGasInputs extends Component {
     }
   }
 
-  onChangeGasLimit = (e) => {
-    this.setState({ gasLimit: e.target.value });
-    this.changeGasLimit({ target: { value: e.target.value } });
+  onChangeGasLimit = ({ target: { value } }) => {
+    const newGasLimit = Number(value);
+    if (isNaN(newGasLimit) || newGasLimit < 0) {
+      this.setState({ gasLimit: this.state.gasLimit });
+      return;
+    }
+    this.setState({ gasLimit: newGasLimit });
+    this.changeGasLimit(newGasLimit);
   };
 
-  changeGasLimit = (e) => {
-    this.props.updateCustomGasLimit(Number(e.target.value));
+  changeGasLimit = (newGasLimit) => {
+    this.props.updateCustomGasLimit(newGasLimit);
   };
 
-  onChangeGasPrice = (e) => {
-    this.setState({ gasPrice: e.target.value });
-    this.changeGasPrice({ target: { value: e.target.value } });
+  onChangeGasPrice = ({ target: { value } }) => {
+    const newGasPrice = Number(value);
+    if (isNaN(newGasPrice) || newGasPrice < 0) {
+      this.setState({ gasPrice: this.state.gasPrice });
+      return;
+    }
+    this.setState({ gasPrice: newGasPrice });
+    this.changeGasPrice(newGasPrice);
   };
 
-  changeGasPrice = (e) => {
-    this.props.updateCustomGasPrice(Number(e.target.value));
+  changeGasPrice = (newGasPrice) => {
+    this.props.updateCustomGasPrice(newGasPrice);
   };
 
   gasPriceError({


ui/components/app/gas-customization/advanced-gas-inputs/advanced-gas-inputs.component.js
```diff
@@ -52,22 +52,32 @@ export default class AdvancedGasInputs extends Component {
...
+    this.props.updateCustomGasLimit(newGasLimit);
   };
 
-  onChangeGasPrice = (e) => {
-    this.setState({ gasPrice: e.target.value });
-    this.changeGasPrice({ target: { value: e.target.value } });
+  onChangeGasPrice = ({ target: { value } }) => {
+    const newGasPrice = Number(value);
+    if (isNaN(newGasPrice) || newGasPrice < 0) {
+      this.setState({ gasPrice: this.state.gasPrice });
```
jpuri, line 72: Not sure why is this setState needed `this.setState({ gasPrice: this.state.gasPrice });` ?

ryanml: ideally this would be an `event.preventDefault()` call, but this will not work for `onChange` events, this is to restore the input to the last valid value. See chat for more details on this particular PR :)

Gudahtt: So just to be clear - this is being done to overwrite the value in the actual text input HTML element? The state isn't changing here, but setting state here has the side-effect of triggering another render cycle that writes this back to the element.

ui/components/app/gas-customization/advanced-gas-inputs/advanced-gas-inputs.component.js
```diff
@@ -52,22 +52,32 @@ export default class AdvancedGasInputs extends Component {
     }
   }
 
-  onChangeGasLimit = (e) => {
-    this.setState({ gasLimit: e.target.value });
-    this.changeGasLimit({ target: { value: e.target.value } });
+  onChangeGasLimit = ({ target: { value } }) => {
+    const newGasLimit = Number(value);
+    if (isNaN(newGasLimit) || newGasLimit < 0) {
```
Gudahtt, line 57: The `isNaN` check is redundant I think, since the fact that this is a `type=number` component guarantees the value gets coerced to a valid floating point number. I guess it doesn't hurt to check anyway though in case the type of the input is changed.

Gudahtt, line 57: It might be worth preventing the user from setting this to non-integer values as well. I think only integers are valid here.

ui/components/app/gas-customization/advanced-gas-inputs/advanced-gas-inputs.component.js
ryanml: correct yes

ui/components/app/gas-customization/advanced-gas-inputs/advanced-gas-inputs.component.js
ryanml: sorry, just want to make sure I understand, are you suggesting something additional or are you following up on https://github.com/MetaMask/metamask-extension/pull/12061#discussion_r706216696

Gudahtt: This would be something additional. I don't mean to block the PR - what you have here is still an improvement - but I think "Ensuring only valid numerical inputs" should include disallowing non-integer numbers. e.g. we can't allow the gas limit to be "0.5" or "3.1415".

Gudahtt: The same applies to the gas price, except that for the price we can allow a certain number of decimal places. I can't recall how many :thinking: Basically we want the value in WEI to be an integer. But IIRC the input is in GWEI

ryanml: Ah I understand! I'd be happy to update this PR to include that check, may as well do it now :)

darkwing:
Before merging this, I think it's a good time to introduce tests with the different values that require this change to be made.

mcmire:
This is an improvement over what exists, but I'm wondering if we can provide a better user experience. I don't like tend to like experiences where as soon as I type something bad the whole interface changes on me or corrects what I wrote, it ends up being temporarily confusing I feel like. I'm wondering if it would be smoother to validate the input (showing an error message if invalid) onBlur rather than onChange and have the user correct their input instead of changing things on them. This would require changing the input to a text field so I realize it can't be made in this PR, but that's food for thought.

jpuri (approved):
Nice change, but yep would be nice to add tests for the inputs.

mcmire (changes requested):
I'm going to mark this as "needs changes" since I too would like to see tests for this.

brad-decker:
@ryanml -- care to update this work to get it landed? otherwise we should close and add this PR to the issue for a future engineer to lean on your previous work.

brad-decker:
Converting to draft for now.

Issue #12067, opened 2021-09-10
Gas Estimate API (way) overcalled - from activity log


adonesky1:
**Describe the bug**
I introduced a nasty bug in [11936](https://github.com/MetaMask/metamask-extension/pull/11936), where the gas estimation API is called for transaction-list-item in a users activity log, ultimately way overcalling this API and resulting in the service hitting its rate limits.
 - Need to refactor useCancelTransaction and useRetryTransaction to still get the benefits of improved gasEstimation for users attempting to speedup or cancel transactions, without calling the API when the users has only checked the activity log.

danjm:
Looks like there already was a ticket for this https://github.com/MetaMask/metamask-extension/issues/12002

adonesky1:
Closed with merge of https://github.com/MetaMask/metamask-extension/pull/12069

PR #12069, opened 2021-09-10
Fix gas api overcalling

### LINKED ISSUE #12002: Issuing too many gasApi requests
Just sitting on the home page, I can see many requests in the background going to the gasApi
image

This looks like it may be one request per transaction sitting in my current account's queue, even though they've already been mined.

Looks like this is coming from the [GasFeeController](https://github.com/MetaMask/controllers/blob/b8c83704622a5d56ccf513f9ea1440d686c897a5/src/gas/GasFeeController.ts), which is shared with mobile, so this might affect both products, and could impact the load that API endpoint was under (recently needed to scale up)

adonesky1:
**Fixes: #12002 & #12067**

**Explanation**:  I introduced a bug in this [PR](https://github.com/MetaMask/metamask-extension/pull/11936) in which we (way) over call to our gasprice estimate API. This fixes this issue by moving the gas API calls into more appropriate places.

(edit) **More context**:
The PR that introduced the bug was meant to improve our speedup/cancel transaction success rate by introducing gasEstimates as part of the calculation for what speedup/cancel gas fee's should be set to. This was poorly implemented (by me) and introduced a call to the gas Estimates API (via useGasFeeEstimates) in a hook (useIncrementedGasFees) that was called directly on each TransactionListItem in the activity log.

This PR relocates and compartmentalizes the calls to the API ( via useGasFeeEstimates) in such a way that no functionality is lost but only required API calls are made.

**Manual testing steps:**  
  - Open the background console and navigate tot he network tab
  - Send a simple eth transaction on Rinkeby (or Mainnet)
  - Make sure there are not a ton of queries to `SuggestedGasFee` (request URL: https://gas-api.metaswap.codefi.network/networks/4/suggestedGasFees)
  - There should be 1 or 2 every ~10 seconds and no more than that.


ui/components/app/cancel-button/cancel-button.component.js
```diff
@@ -0,0 +1,64 @@
+import { Tooltip } from '@material-ui/core';
```
Gudahtt, line 1: Nit: the `.component.js` naming convention is from when we used to use the "container" / "component" pattern, with the Redux hooks in the "container" and everything else in the "component". Now that we're using React Hooks instead, I don't think we need a `.component.js` suffix anymore.

adonesky1: ✅

darkwing (dismissed):
Really excellent work!

ui/hooks/useIncrementedGasFees.js
```diff
@@ -56,10 +56,9 @@ function getHighestIncrementedFee(originalFee, currentEstimate) {
  * ).CustomGasSettings} - Gas settings for cancellations/speed ups
  */
 export function useIncrementedGasFees(transactionGroup) {
-  const { primaryTransaction } = transactionGroup;
+  const { primaryTransaction = {} } = transactionGroup;
```
Gudahtt, line 59: It should not be possible for the `primaryTransaction` to be falsy :thinking: We assume it exists all over the place.

ryanml, line 59: did this fix any issue in particular? Or is this just a general cleanup item

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
```diff
@@ -63,6 +64,20 @@ export default function EditGasPopover({
 
   const minimumGasLimitDec = hexToDecimal(minimumGasLimit);
 
+  const updatedCustomGasSettings = useIncrementedGasFees(transaction);
+
+  let transactionToUse = transaction;
```
Gudahtt, line 69: Nit: This name is a bit confusing :thinking: Maybe something like `customizedTransaction` would be better? Or `updatedTransaction`? :shrug:

Gudahtt (commented):
Can you confirm whether I understand this correctly? It looks like the key here to preventing the gas API from being called excessively is in moving the `useIncrementedGasFees` call outside of the main render function of the transaction list item, so that it's not called when they get rendered.

If that's the case, this PR does look like an improvement but it's still calling the gas API for each approved pending transaction because of the cancel button render function.

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
adonesky1: I'll roll with updateTransaction.

ui/hooks/useIncrementedGasFees.js
adonesky1: removed 😄

adonesky1:
@Gudahtt yes so we don't avoid all calls to the API in the transaction-list, but now we only call it for the one actually pending transaction (not all historical transaction items) and in the previous implementation some re-render cycling issue was causing the hook to re-rerun the query constantly. Now it hits only when the transactionGroup is updated

adonesky1:
@darkwing just updated some variable names and other small cleanup tasks

test/e2e/fixtures/imported-account/state.json
```diff
@@ -489,6 +492,6 @@
     }
   },
   "meta": {
-    "version": 40
+    "version": 63
```
Gudahtt, line 495: This change in version broke the e2e tests. The `participateInMetaMetrics` state moved in migration 49, so now that modal is showing at the start of each test using this.

For now you should be able to leave this as it was before I think? But in general we really ought to setup some script to run migrations on fixture files to make updating their versions easier.

adonesky1: So I'm trying to figure out the best way to resolve an issue caused by `primaryTransaction` on the transactionGroup being passed to useIncrementedGasFees being `undefined` sometimes in a couple of e2e tests... I'm not entirely sure why. I originally solved this by add a default to the primaryTransaction being destructured in useIncrementedGasFees:  `const { primaryTransaction = {} } = transactionGroup;` but I should probably dig in a bit more into the cause.

Gudahtt: That's curious. This state log has _no_ transactions, so there should be no transaction items shown. Which test did you see this failure in?

adonesky1: Sorry yes I wasn't clear, so I dug in a bit deeper and primaryTransaction (and initial transaction for that matter) aren't on the transaction object (which becomes the transactionGroup object) until in the transaction-list, so there won't be a primaryTransaction when, for instance the user clicks on the edit gas modal in the send flow, as the code stands. So in [my latest commit](https://github.com/MetaMask/metamask-extension/pull/12069/commits/50481bd0139958710c0bfd7d43ac42168c33df69) I check if there is a primaryTransaction defined on the transactionsGroup object and, if not, put the root level txParams in as the primaryTransaction. This may not be exactly the way we want to do it, but its the right direction.

adonesky1: @Gudahtt

Gudahtt: That doesn't make much sense to me :thinking: If it's a transaction group, it has to have a `primaryTransaction`. If it's not, then the parameters are mismatched here.

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
```diff
@@ -62,6 +63,19 @@ export default function EditGasPopover({
   ] = useState(false);
 
   const minimumGasLimitDec = hexToDecimal(minimumGasLimit);
+  const updatedCustomGasSettings = useIncrementedGasFees(transaction);
```
Gudahtt, line 66: I think this is the root of the problem you're seeing. `useIncrementedGasFees` takes a transaction group, but you're passing in a single transaction here instead.

Gudahtt: Perhaps `useIncrementedGasFees` could be updated to take a single transaction? It doesn't seem to make any use of the transaction group - just the primary transaction.

adonesky1: yes that's a far more sensible solution here... 🚀

test/e2e/fixtures/imported-account/state.json
```diff
@@ -443,6 +443,9 @@
         }
       }
     },
+    "TransactionController": {
```
Gudahtt, line 446: Nit: could this change be removed? I don't think it's needed anymore

Gudahtt (dismissed):
LGTM!

test/e2e/fixtures/imported-account/state.json
adonesky1: quite right.

adonesky1: ✅

Gudahtt (dismissed):

darkwing:
I'm trying to figure out why, but on this branch, when I click the `cancel` button, the `Gas Limit` comes up as `0` with an error style.  This isn't happening on `develop` branch though.

adonesky1:
> I'm trying to figure out why, but on this branch, when I click the `cancel` button, the `Gas Limit` comes up as `0` with an error style. This isn't happening on `develop` branch though.

good catch @darkwing! So glad you tested! Fixed in [2f71d5](https://github.com/MetaMask/metamask-extension/pull/12069/commits/2f71d54eef950758dceb4075b58589975b24d286)

danjm (dismissed):
LGTM

ui/hooks/useIncrementedGasFees.js
```diff
@@ -68,19 +66,19 @@ export function useIncrementedGasFees(transactionGroup) {
     // do not have txParams. This is why we use optional chaining on the
     // txParams object in this hook.
     const temporaryGasSettings = {
-      gasLimit: primaryTransaction.txParams?.gas,
-      gas: primaryTransaction.txParams?.gas,
+      gasLimit: transaction?.txParams?.gas,
```
Gudahtt, line 69: Is it possible for the `transaction` to be falsy at this point? I think all of these `?`'s can be dropped. This hook doesn't make a lot of sense if there's no transaction.

adonesky1: Indeed! Removed in latest commit: [902279a](https://github.com/MetaMask/metamask-extension/pull/12069/commits/902279ae2d23f645eb764704bec5b9f6512a6231)

Gudahtt (approved):
LGTM!

darkwing (approved):

PR #12090, opened 2021-09-13
Revisit #12019 : Show save dialog when calling a smart contract method and use the saved contact is already saved

segun:
Fixes: a bug introduced by this PR #12019 

Explanation:  
#12019 is a fix for the below:

Change the [domain] suggested gas fee (i) to Estimated gas fee (i)
Change that line's color from orange to black
Under that line, add Site suggested in italics in gray.

But it introduced the bug discussed [here](https://consensys.slack.com/archives/GTQAGKY5V/p1631220853285800)

`I'm seeing what I believe is a bug on v10.1.0. Creating transaction from within metamask shows the recipient as "New Contract", when it is one of the users own current accounts:
`

This is a fix for the bug.

Should not show add new address dialog under the below conditions
    Address is own adddress
    send to recipient header is not shown

Manual testing steps:  
  - 
  - 
  -

### DIFF
--- a/package.json
+++ b/package.json
@@ -306,6 +306,7 @@
     "randomcolor": "^0.5.4",
     "rc": "^1.2.8",
     "react-devtools": "^4.11.0",
+    "react-test-renderer": "^17.0.2",
     "read-installed": "^4.0.3",
     "redux-mock-store": "^1.5.4",
     "remote-redux-devtools": "^0.5.16",
--- a/package.json
+++ b/package.json
@@ -306,7 +306,6 @@
     "randomcolor": "^0.5.4",
     "rc": "^1.2.8",
     "react-devtools": "^4.11.0",
-    "react-test-renderer": "^17.0.2",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -306,6 +306,7 @@
     "randomcolor": "^0.5.4",
     "rc": "^1.2.8",
     "react-devtools": "^4.11.0",
+    "react-test-renderer": "^17.0.2",
```
ryanml, line 309: is this actually being used? are we not able to accomplish something without this?

segun: No this is not needed. I was experimenting with it, but decided against it.

danjm (changes requested):
This looks good! Nice work.

I think you just need to run `yarn setup` again, and then commit the lockfile changes, so that no lockfile changes are included from when you were experimenting with that other testing module.

danjm (approved):

brad-decker (approved):
LGTM

PR #12091, opened 2021-09-13
lavamoat - add policy override disabling write for eslint in build sys

kumavis:


kumavis:
we cant do this because we need policy expressiveness to be able to specify a noop fn

PR #12100, opened 2021-09-14
Add target object word to "View on Etherscan" links

### LINKED ISSUE #9476: Add target object word to "View on Etherscan" links.
We have "View on Etherscan" in a few places, and it means different things (accounts or tokens). We know this can be confusing to users. One simple way to mitigate this confusion is to update the wording in each case to clarify _what_ is being viewed.
- View Account on Etherscan
- View Token Contract on Etherscan

Bonus: We could use the term "Block explorer" instead of etherscan, to pave the way for user-configured block explorers.

segun:
Fixes: #9476 

Explanation:  
We have "View on Etherscan" in a few places, and it means different things (accounts or tokens). We know this can be confusing to users. One simple way to mitigate this confusion is to update the wording in each case to clarify what is being viewed.

Manual testing steps:  
  - 
  - 
  -

### DIFF
--- a/app/_locales/am/messages.json
+++ b/app/_locales/am/messages.json
@@ -1124,15 +1124,6 @@
   "viewContact": {
     "message": "ዕውቂያን ይመልከቱ"
   },
-  "viewOnCustomBlockExplorer": {
-    "message": "በ $1ይመልከቱ"
-  },
-  "viewOnEtherscan": {
-    "message": "በ Etherscan ላይ ይመልከቱ"
-  },
-  "viewinExplorer": {
-    "message": "በኤክስፕሎረር ተመልከት"
-  },
   "visitWebSite": {
     "message": "ድረ ገጻችንን ይጎብኙ"
   },
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -262,7 +262,7 @@
   "blockExplorerAccountAction": {
     "message": "Account"
   },
-  "blockExplorerAseetAction": {
+  "blockExplorerAssetAction": {
     "message": "Asset"
   },
   "blockExplorerSwapAction": {
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -2781,13 +2781,16 @@
     "message": "View More"
   },
   "viewOnCustomBlockExplorer": {
-    "message": "View $1 at $2"
+    "message": "View $1 at $2",
+    "description": "$1 is the action type. e.g (Account, Transaction, Swap) and $2 is the Custom Block Exporer URL"
   },
   "viewOnEtherscan": {
-    "message": "View $1 on Etherscan"
+    "message": "View $1 on Etherscan",
+    "description": "$1 is the action type. e.g (Account, Transaction, Swap)"
   },
   "viewinExplorer": {
-    "message": "View $1 in Explorer"
+    "message": "View $1 in Explorer",
+    "description": "$1 is the action type. e.g (Account, Transaction, Swap)"
   },
   "visitWebSite": {
     "message": "Visit our web site"
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -260,16 +260,20 @@
     "message": "Welcome to MetaMask Beta"
   },
   "blockExplorerAccountAction": {
-    "message": "Account"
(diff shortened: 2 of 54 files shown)


app/_locales/en/messages.json
```diff
@@ -2754,13 +2754,13 @@
     "message": "View More"
   },
   "viewOnCustomBlockExplorer": {
-    "message": "View at $1"
+    "message": "View $1 at $2"
```
brad-decker, line 2757: [REQ] When we add variables to messages our current best practices are to include a description field that explains what those variables will be. There are examples in the messages.json file. I also realize that the original didn't have this but its a good time to add it now that we are extending it

ui/pages/swaps/awaiting-swap/view-on-ether-scan-link/view-on-ether-scan-link.js
```diff
@@ -12,6 +12,8 @@ export default function ViewOnEtherScanLink({
 }) {
   const t = useContext(I18nContext);
 
+  const blockExplorerViewAction = 'Swap';
```
brad-decker, line 15: [NIT/Weakly held opinion/change my mind] while extracting constants like this is generally good, the likelihood of the object changing in the context of its location is slim to none without major refactors. I have an ever-so-slight preference of seeing the literal string value in the `t()` callsite because it is less contextual overhead to read and understand the output.

brad-decker (changes requested):
This will make a lot of people very happy. 🎉 only minor feedback.

brad-decker:
Looks like you'll need to update some unit tests that look for the `view on etherscan` string

ui/pages/swaps/awaiting-swap/view-on-ether-scan-link/view-on-ether-scan-link.js
darkwing: Could we should localize all of the `blockExplorerViewAction` values?  If they're front-facing to the user, I could see that being beneficial

darkwing (changes requested):
A few nits from Brad and myself.

brad-decker (dismissed):
LGTM great work

darkwing (dismissed):

adonesky1:
Looks like this needs a rebase @segun!

Gudahtt:
It looks like this branch was messed up during a rebase; this includes _way_ more commits than it should.

mcmire (dismissed):
Makes sense to me.

Gudahtt:
>Looks like descriptions need to be added to all of the other languages :(

There is a script to help with this: `yarn verify-locales:fix`

Gudahtt (changes requested):
Because you're adding a new field to the localized message, this will now break for non-English locales that don't accept the second parameter.  The equivalent messages for other locales will need to be removed and replaced with new translations.

app/_locales/en/messages.json
```diff
@@ -259,6 +259,18 @@
   "betaWelcome": {
     "message": "Welcome to MetaMask Beta"
   },
+  "blockExplorerAccountAction": {
+    "message": "Account"
```
Gudahtt, line 263: Ideally these would have descriptions that link them back up to the `viewOnCustomBlockExplorer` and `viewOnEtherscan` messages, so that translators know which context these words will be used in.

app/_locales/am/messages.json
```diff
@@ -1125,13 +1125,16 @@
     "message": "ዕውቂያን ይመልከቱ"
   },
   "viewOnCustomBlockExplorer": {
-    "message": "በ $1ይመልከቱ"
+    "message": "በ $1ይመልከቱ",
```
Gudahtt, line 1128: To clarify, what I meant is that these three messages should be removed from non-English locales completely. Right now the wallet will throw an error when trying to display these messages because they're expected to have a `$1` in them, but there is no `$1`.

The easiest way to delete them all is to rename the English messages, run `yarn verify-locales:fix` (which will delete them in each other locale), then rename the English ones back to the original names.

segun: I see, I ran :fix without renaming first.

Gudahtt (approved):
LGTM!

mcmire (approved):
Nice, looks good!

PR #12101, opened 2021-09-14
E2e fixtures navigate transactions

PeterYinusa:
Related: #11136

Explanation:  
  - create navigate transactions state with 4 unapproved transactions
  - moved navigate transactions test
  - updated navigate transactions test to use fixtures


brad-decker (approved):
😍 LGTM 🎉

test/e2e/tests/navigate-transactions.spec.js
```diff
@@ -0,0 +1,251 @@
...
+    await withFixtures(
+      {
+        dapp: true,
+        fixtures: 'navigate-transactions',
+        ganacheOptions,
+        title: this.test.title,
+      },
+      async ({ driver }) => {
+        await driver.navigate();
+        await driver.fill('#password', 'correct horse battery staple');
```
darkwing, line 24: Nit: I've noticed in our E2E's that we reuse a values like this password, the secret key, etc.  Should we, maybe in a different PR, create a `constants` file and use vars from there?

darkwing (approved):

test/e2e/tests/navigate-transactions.spec.js
brad-decker: that's a fantastic idea.

PeterYinusa: 💯 - i'll revisit that as a separate piece of work

PR #12110, opened 2021-09-15
Adding search tab in import tokens for custom networks

### LINKED ISSUE #12109: Adding Search Tab to custom network with token list
The token service API provides tokenList for custom networks like Binance and Polygon, so we can add a search tab for those networks with the token lists available.

NiranjanaBinoy:
Fixes: #12109 

The token service API provides tokenList for custom networks like Binance and Polygon, so we can add a search tab for the networks for which the token lists are available. But we are hiding the search tab when token detection is OFF as the static list does not support custom network.


danjm (approved):
Nice work!

darkwing (approved):
Works as described; btw, what is the `network.js` change?  Is that required?

NiranjanaBinoy:
> Works as described; btw, what is the `network.js` change? Is that required?

Not totally related, I was testing with polygon network and even though the image was added the mapping was not done as a result the icon was not getting displayed when we switch the network.

PR #12113, opened 2021-09-15
Add more global snap hooks

rekmarks:
This updates / adds more snap functions to the global object. They're now namespaced under `globalThis.snaps` instead of directly on the global.


Gudahtt:
I have just rebased this against the `snaps` branch.

Gudahtt (approved):
LGTM!

PR #12115, opened 2021-09-15
Adding "Enable smart contract data" step to Ledger Instructions (Confirmation Screen)

ryanml:
Follows up: https://github.com/MetaMask/metamask-extension/pull/12020
(Note: This step only shows for transactions containing method data)

Screen Shot 2021-09-15 at 6 04 55 AM
Screen Shot 2021-09-15 at 6 02 47 AM
Screen Shot 2021-09-15 at 6 05 50 AM

### DIFF
--- a/ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
+++ b/ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
@@ -401,45 +401,37 @@ export default class ConfirmTransactionBase extends Component {
       </div>
     ) : null;
 
+    const renderLedgerLiveStep = (text, show = true) => {
+      return (
+        show && (
+          <Typography
+            boxProps={{ margin: 0 }}
+            color={COLORS.PRIMARY3}
+            fontWeight={FONT_WEIGHT.BOLD}
+            variant={TYPOGRAPHY.H7}
+          >
+            {text}
+          </Typography>
+        )
+      );
+    };
+
     const ledgerInstructionField = isLedgerAccount ? (
       <div>
         <div className="confirm-detail-row">
           <Dialog type="message">
             <div className="ledger-live-dialog">
-              <Typography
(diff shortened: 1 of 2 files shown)


darkwing (dismissed):

ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
```diff
@@ -401,45 +402,37 @@ export default class ConfirmTransactionBase extends Component {
...
-                  {`- ${t('ledgerLiveDialogStepOne')}`}
-                </Typography>
+              {renderLedgerLiveStep(t('ledgerLiveDialogHeader'))}
+              {renderLedgerLiveStep(
+                `- ${t('ledgerLiveDialogStepOne')}`,
+                !isFirefox,
+              )}
+              {renderLedgerLiveStep(`- ${t('ledgerLiveDialogStepTwo')}`)}
+              {renderLedgerLiveStep(`- ${t('ledgerLiveDialogStepThree')}`)}
+              {renderLedgerLiveStep(
```
danjm, line 432: Could you make it so that this is conditionally shown based on whether the current transactions has contract data or not?

danjm: ah okay, all that needs to happen is we need to use `txData.txParams.data` instead of `methodData`. `methodData` is the results of calling `getKnownMethodData`, which returns a property from `state.metamask.knownMethodData`, which is not guaranteed to be truthy, even when sending to a contract. But `txData.txParams.data` will always be truthy when sending to a contract

ryanml:
@danjm i've updated this to use `txData.txParams.data`

kumavis:
checkout the storybook highlights 😁🤙

danjm (commented):
LGTM

darkwing (approved):

PR #12120, opened 2021-09-15
Fixing case of password variable and adding test site to readme

jpuri:
Explanation:  
The PR has 2 chores that I felt will be useful from my experience of doing project setup:

1. Fix case of password variable in `.metamaskrc`.
2. Adding test site to readme https://metamask.github.io/test-dapp/


brad-decker:
Has the format of our .metmaskrc file changed? Previously it was just a type of `.env` file that would load environment variables during build process.

jpuri:
> Has the format of our .metmaskrc file changed? Previously it was just a type of `.env` file that would load environment variables during build process.

Hey @brad-decker, we are using this library to parse the file https://www.npmjs.com/package/rc. It should actually support both `json` and `ini` format. I am facing issues with `ini` format and variable `PASSWORD` was not working for me while it did worked for @danjm in `JSON` format. Our initial though was that it is format issue and I created this PR.

But as I dig more into it today I found that issue probably lies in how `password` variable is being used probably [here](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/metamask-controller.js#L626). I will look more into it to not change the format of `.metamaskrc` but rather fix use of `password` variable.

Thanks for the feedback.

jpuri:
Turned out that issue was with case of `password` variable. While other env variables are uppercase `password` is lowercase. An option is to add it to `.metamaskrc.dist` file like [here](https://github.com/MetaMask/metamask-extension/pull/12120/files#diff-acd4fa2c53711dfe68770fac434b05961542d621547211604bffdeab13ab958aR2), though I am not sure if that is ok from security point of view.

brad-decker:
@jpuri it would be fine, IMO, to add the password= to the `.metamaskrc.dist` file without a value. I believe this will hit the block of code that looks for it the exact same as if it hadn't been specified. This would allow for new contributors to more readily copy it into their `.metamaskrc` -- the only concern I would have is someone *accidentally* adding their password to the `.dist` file and committing it. A safer alternative could be uppercasing the password variable to match others and then documenting the casing of these files such that we don't get into a similar situation in the future.

darkwing (dismissed):

jpuri:
> @jpuri it would be fine, IMO, to add the password= to the `.metamaskrc.dist` file without a value. I believe this will hit the block of code that looks for it the exact same as if it hadn't been specified. This would allow for new contributors to more readily copy it into their `.metamaskrc` -- the only concern I would have is someone _accidentally_ adding their password to the `.dist` file and committing it. A safer alternative could be uppercasing the password variable to match others and then documenting the casing of these files such that we don't get into a similar situation in the future.

Hey @brad-decker, I would prefer to fix the case but my little concern with changing the case is that it would need change in local config of all devs. I will merge this PR and we can discuss and create a separate one for case fix.

brad-decker:
@jpuri I think that's an acceptable short-term issue and gets us to a place where the env variables specified in dist are all in the right case. We can first document the requirement then later write a lint script for it to make sure it is enforced. I'm open to other's thoughts on this though. I'm also wondering if this has any unexpected behavior for those who have yet to make their `metamaskrc` file where the provided password is incorrect or if it just silently fails to unlock the wallet and lands the user at the unlock screen as expected.

jpuri:
Hey @brad-decker, I updated the PR with fix for `password` variable case. I will message to all devs to change their local variable case as I merge this.

darkwing (approved):

brad-decker (approved):
LGTM

PR #12122, opened 2021-09-15
fix stuck speedup problem

### LINKED ISSUE #12051: Transaction with `?` gasprice after failed bump, cannot modify any more
After submitting a transaction with a certain gas price, Metamask offers the user the possibility to speed it up via increasing the gas price. The new gas price represents not just the 10% bump needed, but also tries to hone in on current network conditions via the basefee. This can however be overridden by the user. 

Yesterday while the basefee was swinging wildly I was trying to speed up a transaction by just a bit (10%, not the new basefee) so I manually edited the gas price of the proposed speedup tx. I "failed" at that, because I set it to 9.x%, getting the transaction rejected with "replacement underpriced". That would be fine, however now the pending transaction in metamask appears with a  base fee and priority fee of `?`, and there's no more speedup button, so I can't fix my error.

I'm unsure if this is some deliberate protection mechanism, but there's no way to resubmit the transaction even manually now. Cancellation still works, but I don't want to cancel, just resubmit with manually entered as prices.

brad-decker:
Fixes: #12051 

Explanation: Speed up transactions that are failed should not prevent additional retries. Only speed up transactions that are successfully submitted to the network should prevent the transaction from being retried. In addition this clears up some additional user confusion by simply not rendering the baseFee or priorityFee fields when we can't know them, instead of the `?` mark which doesn't do anything to explain why its there. 

Manual testing steps:  
  - Switch to rinkeby
  - Submit a transaction with max and priority fee of 0.1
  - speed up the transaction, with max and priority fee of 0.109 (109%)
  - On develop you'll see that the speed up button is removed, and the speed up is failed (you'll get a notification) but the original transaction is pending.
  - On this branch you'll see the speed up button remains.

### DIFF
--- a/ui/components/app/transaction-breakdown/transaction-breakdown.component.js
+++ b/ui/components/app/transaction-breakdown/transaction-breakdown.component.js
@@ -100,40 +100,32 @@ export default class TransactionBreakdown extends PureComponent {
             />
           </TransactionBreakdownRow>
         )}
-        {isEIP1559Transaction && (
+        {isEIP1559Transaction && typeof baseFee !== 'undefined' ? (
           <TransactionBreakdownRow title={t('transactionHistoryBaseFee')}>
-            {typeof baseFee === 'undefined' ? (
-              '?'
-            ) : (
-              <CurrencyDisplay
-                className="transaction-breakdown__value"
-                data-testid="transaction-breakdown__base-fee"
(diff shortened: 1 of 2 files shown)


darkwing (approved):

adonesky1 (approved):
LGTM!

PR #12127, opened 2021-09-16
Remove unwanted single returns from arrow functions

darkwing:
This is a code styling proposal to remove unnecessary `return` statements inside arrow functions when we could use the implied return.  For example:

```
// from ...
const j = () => {
  return 1;
};

// to....
const je = () => 1;
```

The real work is `arrow-body-style: 1` in our ESLint config -- the rest of these changes are done via `yarn lint --fix`.

Thoughts?


app/scripts/controllers/permissions/permissions-middleware.test.js
```diff
@@ -19,19 +19,17 @@ const { CAVEATS, ERRORS, PERMS, RPC_REQUESTS } = getters;
 
 const { ACCOUNTS, DOMAINS, PERM_NAMES } = constants;
 
-const initPermController = () => {
-  return new PermissionsController({
+const initPermController = () =>
```
mcmire, line 22: I feel like in cases like this, the braces can help with readability.

app/scripts/controllers/permissions/restricted-methods.test.js
```diff
@@ -69,12 +65,11 @@ describe('restricted methods', function () {
     it('should return accounts in keyring order when none are selected', async function () {
       const keyringAccounts = ['0x7e57e2', '0x7e57e3', '0x7e57e4', '0x7e57e5'];
       const restrictedMethods = getRestrictedMethods({
-        getIdentities: () => {
-          return keyringAccounts.reduce((identities, address) => {
+        getIdentities: () =>
+          keyringAccounts.reduce((identities, address) => {
```
mcmire, line 69: Same thing with this. It's not impossible to know where the function begins and ends, of course, because of the indentation, but the braces help in being able to spot it more easily. Also with braces it's possible to jump from the start to the end of a block.

mcmire (commented):
In general I like this change but there are cases where this can reduce readability IMO, so I'm slightly 👎🏻 . Not a big deal though if others feel differently.

.eslintrc.js
```diff
@@ -52,6 +52,7 @@ module.exports = {
     'default-param-last': 'off',
     'prefer-object-spread': 'error',
     'require-atomic-updates': 'off',
+    'arrow-body-style': 1,
```
rekmarks, line 55: Can we use strings instead of numbers here? We've been trying to avoid using numbers so we don't have to remember what they mean.

rekmarks (changes requested):

.eslintrc.js
darkwing: I tried `'error'` but for some reason the linting bombed, requesting a number.  I'll look into it, it was weird.

brad-decker:
I'm in favor of this change. I do see @mcmire 's point on readability in some cases, but in my opinion, it can be generally more readable in more circumstances than it is less. The lack of bracket is a clear indication that the function simply returns whatever follows, and doesn't contain any logic of its own. This is my experience, others may feel differently and should voice their opinions.

Also, at what point does it make sense to elevate this to the eslint config repo?

jpuri:
Agree with @darkwing , @brad-decker : to me also this is more readable and kind of simplified. I always prefer not to have brackets around arrow functions whenever possible.

darkwing:
Closing for upstream https://github.com/MetaMask/eslint-config/pull/205

.eslintrc.js
rekmarks: Update: We can't have nice things: https://github.com/MetaMask/eslint-config/pull/205#issuecomment-921169164

Issue #12128, opened 2021-09-16
Token Detection MVP changes


NiranjanaBinoy:
As the current contract to check for the token balance is not supporting the custom network we have to make the following updations to the token detection MVP for the upcoming release:

- [x] The token detection feature will be turned OFF for all users by default.
- [x] Move the `Use token detection` toggle to the `Advanced` in the settings.
- [x] In Search, use ‘info’ pill -> Link/refer to the toggle on tokens page to make it more discoverable

Related Figma file: https://www.figma.com/file/HgRPmjjq8Y6cmtewnsuFDF/Extension-Token-Detection?node-id=617%3A3477

PR #12129, opened 2021-09-16
Changing Use Token detection to  OFF by default

NiranjanaBinoy:
This is part of #12128 

We are keeping the token detection OFF by default so that the users can opt-in based on their network and helps us to get early feedback reducing the risk of shipping any regression to our whole user base.


Gudahtt (approved):
LGTM!

darkwing (approved):

PR #12136, opened 2021-09-16
Moving UseTokenDetection toggle to new Experimental tab

NiranjanaBinoy:
This PR is a part of #12128 
This PR has to be approved and merged before #12152 

Since Token detection is now introduced as an experimental feature for users to opt-in we are adding a new Experimental tab which is currently positioned just above the `About` tab in the settings(following the pattern in mobile).
Screen Shot 2021-09-17 at 2 18 54 PM

Screen Shot 2021-09-17 at 2 19 09 PM

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -865,6 +865,12 @@
   "expandView": {
     "message": "Expand view"
   },
+  "experimental": {
+    "message": "Experimental"
+  },
+  "experimentalSettingsDescription": {
+    "message": "Token detection & more"
+  },
   "exportPrivateKey": {
     "message": "Export Private Key"
   },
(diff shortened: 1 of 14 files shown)


darkwing (dismissed):

brad-decker (dismissed):

NiranjanaBinoy:
@yanrong-chen and @rachelcope  please confirm if the description of the new tab or the position of the Experimental tab need to be updated

darkwing (approved):
Looks good.

I do worry about having an "Experimental" tab though.  When the feature is no longer experimental, we presumably move it, and then the user has to find it again.

brad-decker (approved):
LGTM I do think @darkwing has expressed an important concern in the relocation of experimental features. Maybe we can leave links from old experimental features to their new locations for a time after they move.

PR #12140, opened 2021-09-17
Adding base to BigNumber constructor to round off value

### LINKED ISSUE #11826: BigNumber Error: new BigNumber() number type has more than 15 significant digits
**Steps to reproduce**

1. Login to the extension
2. Click the Send button on the Home page
3. Select a recipient on the Add Recipient page
4. Click the Max amount button on the Send page, then click Next
5. Click the Edit link in the Estimate gas fee section
6. Enter `5.1111111111111111` in the Max priority fee
7. Enter `6` in the Max fee input field

**Expected behavior**
- Error class on the Max priority fee field
- Save button is disabled

**Error Message**
- new BigNumber() number type has more than 15 significant digits: 5.1111111111111111

jpuri:
Fixes: #11826

Explanation:  Advance gas editing screen throws error when number of significant decimal digits is > 15. I found that at most places in the application base value is passed to `BigNumber` constructor, but it was missing in GasTiming component which was causing this error `new BigNumber() number type has more than 15 significant digits: 5.1111111111111111`. Passing base round-offs the number and avoid this error.

An additional change in this PR is to stop user from entering > 14 decimal places in numeric inputs. I found an issue with numeric inputs as soon as decimal places in numeric input > 15 value is no longer valid number and the input becomes un-controlled. Thus it gives impression to user that he has added any number of decimal places but in reality it can not exceed 15 decimal places as parseFloat does not return > 15 decimal places.

Manual testing steps:  
  - In advance gas editing screen try to enter > 14 decimal places in numeric inputs.
  - User is not able to enter > 14 decimal places.


ui/components/app/gas-timing/gas-timing.component.js
```diff
@@ -71,8 +71,8 @@ export default function GasTiming({
     ) {
       // getGasFeeTimeEstimate requires parameters in string format
       getGasFeeTimeEstimate(
-        new BigNumber(priority).toString(10),
-        new BigNumber(fee).toString(10),
+        new BigNumber(priority, 10).toString(10),
```
danjm, line 74: Do you know if priority and fee are in base 10 when passed to BigNumber? Or are they base 16?

jpuri: It appears to be base 10, the value is coming from this hook https://github.com/MetaMask/metamask-extension/blob/develop/ui/hooks/useGasFeeInputs.js#L220 and is user input on advance gas form.

jpuri:
I found that on send page we actually do not allow user to type > 6 significant decimal digits. I think it will be good to have same behaviour here.

ui/components/app/gas-timing/gas-timing.component.js
jpuri: And default is base 10, thus even previously it was base 10.

ui/components/ui/numeric-input/numeric-input.component.js
```diff
@@ -26,7 +28,10 @@ export default function NumericInput({
           }
         }}
         onChange={(e) => {
-          onChange?.(parseFloat(e.target.value || 0, 10));
+          const newValue = e.target.value;
+          const match = DECIMAL_REGEX.exec(newValue);
+          if (match[1]?.length >= 15) return;
+          onChange?.(parseFloat(newValue || 0, 10));
```
jpuri, line 34: I added above code to ensure that user is not able to enter an invalid numeric value into number input, invalid in sense that it has > 15 decimal places.

---------------------------------------------------------------------------------------------------------------
Number inputs are quite complicated in metamask's scenario. As decimal places in number value exceed 15 the inputs is no longer controlled input in react, it starts behaving in-controlled. Ref: https://github.com/facebook/react/issues/1549

Additional complexity is that `parseFloat` ignores digits after 15 significant digits

parseFloat(1.11111111111111)                          -> 1.111111111111111
parseFloat(1.111111111111111111111111111)    -> 1.111111111111111

Thus user may see that they have added any number of decimal digits but in code it can not exceed 15 - it only looks like value has > 15 decimal places as input becomes un-controlled.

---------------------------------------------------------------------------------------------------------------
Approach I took here is as soon as user enters 15th decimal digit - we reject the change and do not let user enter any more decimal digits. This is to ensure that input does not becomes un-controlled. I expect 14 decimal places to suffice for all use cases.

Adding form error for > 15 decimal places to inputs are also complicated as code outside `onChange` method will see this value `parseFloat(newValue || 0, 10)` which will never exceed 15 decimal places.

---------------------------------------------------------------------------------------------------------------
To allow users to enter more decimal places we need to change the inputs to text inputs.

danjm:
Code changes look good, but e2e test failures should be addressed

danjm (approved):
LGTM

ui/components/ui/numeric-input/numeric-input.component.js
mcmire: I read through the GH issue you linked to. I am curious whether the input becomes uncontrolled still as was described there as a fix for this landed in React a while back. So, I wonder if the weird behavior we are seeing right now is a combination of a floating point issue (floating numbers only support a certain number of significant digits) and the fact that we are using `parseFloat` here but BigNumber later. 

Regardless, it seems that this is a good change. I am wondering whether we could go farther and have the regex be this:

``` javascript
const DECIMAL_REGEX = /^\d+|\d*\.\d{15,}/u;
```

The reason is that even with this change, since we are using `parseFloat` to normalize the value, if you enter something like an `=`, then the input resets to 0. This regex would prevent the user from entering something non-numeric entirely. Then we wouldn't have to run the value through `parseFloat`, as we already know it's in a numeric format. I made a [CodePen](https://codepen.io/mcmire/pen/vYZagRR) that demonstrates this behavior. What do you think?

mcmire (approved):
Had a couple of comments below but this seems like a good improvement.

ui/components/ui/numeric-input/numeric-input.component.js
jpuri: Input does become uncontrolled - it is indeed weird behaviour - at least on chrome number inputs behave weird if value is not valid number. I spent time debugging it and found that input can show any number of decimal places parseFloat returned only 15 and thus app state had 15 at max.

mcmire: Gotcha, interesting. Good to know, thanks for spending time on this.

ui/components/app/gas-timing/gas-timing.component.js
jpuri: I added some details here https://github.com/MetaMask/metamask-extension/pull/12140#issue-998970996, passing a base round-ups the value and saves from exception being thrown.

PR #12148, opened 2021-09-17
Fix issue where duplicate tokens without icons appear in swap-to token list dropdown

### LINKED ISSUE #12103: Using Swaps Adds Custom Token to Token List
After getting a swap quote, a new "custom" token is added to the tokens dropdown:

https://user-images.githubusercontent.com/46655/133307250-17abc43c-4b37-44d7-8b76-0807f3464f12.mp4

This is problematic, as it erodes trust in what the user is swapping to.

When selecting that icon, the user must then acknowledge that they know the token is custom.

adonesky1:
Fixes: #12103 
co-authored with @mcmire 

Explanation:  User tokens weren't being matched correctly against swaps tokensToSearch list (casing issue), resulting in duplicates icon and aggregator data.

**Outstanding tech debt to consider**: 
- [ ] We should agree on and apply normalized casing for token addresses (and all addresses for that matter) from coingecko, contract-metadata and throughout wherever else addresses are stored or received.
- [ ] We noticed a strange bug with the swaps duck (the duck state intermittently doesn't reset correctly):
https://user-images.githubusercontent.com/34557516/133833994-243774ca-3510-4c6d-9799-5d69031ec7c5.mov
https://user-images.githubusercontent.com/34557516/133834055-98714376-e427-414b-9458-cc12a366423d.mov

Fixed behavior in PR:
https://user-images.githubusercontent.com/34557516/133832688-44ac6f66-6798-4252-ae31-05d7a03b29f5.mov

https://user-images.githubusercontent.com/34557516/133832883-13fa77a0-bbd1-40ef-96e5-690038c74bf5.mov

Manual testing steps:  
1)
  - Import token(s) from our enumerated token list
  - Initialize a swap
  - Make sure that there isn't a duplicate of added token(s) in the `Swap to` dropdown
 2)
 - Initialize a swap
 - Select a new token (not currently in your added token list) in the `Swap to` dropdown
 - Click back into the `Swap to` drop down and ensure the token you just selected isn't duplicated in the dropdown list.

### DIFF
--- a/ui/hooks/useTokensToSearch.js
+++ b/ui/hooks/useTokensToSearch.js
@@ -66,7 +66,8 @@ export function getRenderableTokenData(
   const usedIconUrl =
     iconUrl ||
     (tokenList[tokenAddress] &&
-      `images/contract/${tokenList[tokenAddress].iconUrl}`);
+      `images/contract/${tokenList[tokenAddress].iconUrl}`) ||
+    token?.image;
   return {
     ...token,
     primaryLabel: symbol,
@@ -127,7 +128,7 @@ export function useTokensToSearch({
   const memoizedTokensToSearch = useEqualityCheck(tokensToSearch);
   return useMemo(() => {
     const usersTokensAddressMap = memoizedUsersToken.reduce(
-      (acc, token) => ({ ...acc, [token.address]: token }),
+      (acc, token) => ({ ...acc, [token.address.toLowerCase()]: token }),
       {},
     );
 
@@ -139,12 +140,12 @@ export function useTokensToSearch({
 
     const memoizedSwapsAndUserTokensWithoutDuplicities = uniqBy(
       [...memoizedTokensToSearch, ...memoizedUsersToken],
-      'address',
+      (token) => token.address.toLowerCase(),
     );
 
     memoizedSwapsAndUserTokensWithoutDuplicities.forEach((token) => {
       const renderableDataToken = getRenderableTokenData(
-        { ...usersTokensAddressMap[token.address], ...token },
+        { ...usersTokensAddressMap[token.address.toLowerCase()], ...token },
         tokenConversionRates,
         conversionRate,
         currentCurrency,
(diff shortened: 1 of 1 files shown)


mcmire:
Also worth noting that [this line in `fetchQuotesAndSetQuoteState`](https://github.com/MetaMask/metamask-extension/blob/develop/ui/ducks/swaps/swaps.js#L491) is always true because the set of token addresses in `contractExchangeRates` is a different case than the token address that the user selects. We could normalize this address before we check for its existence in `contractExchangeRates`, however, there seems to be a timing issue in that `fetchQuotesAndSetQuoteState` seems to be getting called before `contractExchangeRates` is updated with the rate for the token the user has selected. So in the brief moment in time where the token is not present in `contractExchangeRates`, `addToken` will get called anyway. We are not sure if this is a bug or not, but I think it's worth investigating at some point.

ui/hooks/useTokensToSearch.js
```diff
@@ -66,7 +66,8 @@ export function getRenderableTokenData(
   const usedIconUrl =
     iconUrl ||
     (tokenList[tokenAddress] &&
-      `images/contract/${tokenList[tokenAddress].iconUrl}`);
+      `images/contract/${tokenList[tokenAddress].iconUrl}`) ||
+    token?.image;
```
Gudahtt, line 70: The `token` is destructured earlier in this function, so `token` can't be falsy here.
```suggestion
    token.image;
```

ui/hooks/useTokensToSearch.js
```diff
@@ -66,7 +66,8 @@ export function getRenderableTokenData(
   const usedIconUrl =
     iconUrl ||
     (tokenList[tokenAddress] &&
-      `images/contract/${tokenList[tokenAddress].iconUrl}`);
+      `images/contract/${tokenList[tokenAddress].iconUrl}`) ||
```
Gudahtt, line 69: When is `iconUrl` expected to be missing? I thought it was present for all tokens, dynamic and static.

Confusingly, it seems to be missing from the static token type in the controller, but I can see it being assigned. But for dynamic tokens, it's declared in the type.

mcmire:
@adonesky1 I know I said I'd work on some e2e tests for this, but don't let me hold you up — I ran into an issue that prevents writing e2e tests for any feature that involves Infura, so I'm going to have to figure that out.

ui/hooks/useTokensToSearch.js
adonesky1: So the token could also be [one that the user has imported manually](https://github.com/MetaMask/metamask-extension/blob/a8a1f7121f5cf35b6700e100cf7ae62cbe8016bc/ui/hooks/useTokensToSearch.js#L142) which is neither in our static json list nor in any of the aggregators. In this case it seems unlikely we would have an image for it but for tokens coming from this source it would be of [this type ](https://github.com/MetaMask/controllers/blob/26be9d21ebcc183eb418d4da48b8a685b43aa19f/src/assets/TokenRatesController.ts#L42) so an image property and no iconUrl property.

@Gudahtt

Gudahtt: Makes sense!

brad-decker (approved):

ryanml (approved):
lgtm!

PR #12150, opened 2021-09-17
Remove whitespace from inputs

exbucks:
Fixes:
Add trim functionality for inputs.

Explanation:  
There are some unexpected use cases when adding custom network. When User copy and paste, it might have unnecessary spaces at the end of string, that causes errors, even if User put correct ones.
Here is a screenshot.
https://share.getcloudapp.com/6quYGAYX

Manual testing steps:  
  - 
  - 
  -

### DIFF
--- a/ui/pages/settings/networks-tab/network-form/network-form.component.js
+++ b/ui/pages/settings/networks-tab/network-form/network-form.component.js
@@ -316,7 +316,7 @@ export default class NetworkForm extends PureComponent {
   setStateWithValue = (stateKey, validator) => {
     return (e) => {
       validator?.(e.target.value, stateKey);
-      this.setState({ [stateKey]: e.target.value });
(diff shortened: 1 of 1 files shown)


ui/pages/settings/networks-tab/network-form/network-form.component.js
```diff
@@ -316,7 +316,7 @@ export default class NetworkForm extends PureComponent {
   setStateWithValue = (stateKey, validator) => {
     return (e) => {
       validator?.(e.target.value, stateKey);
-      this.setState({ [stateKey]: e.target.value });
+      this.setState({ [stateKey]: e.target.value.trim() });
```
jpuri, line 319: I thin it is ok for it to fail if users enters wrong input.

darkwing:
This component not longer exists so I will be closing this PR.  Thank you!

PR #12158, opened 2021-09-20
Hardening property check for `stopPropagation` in `ConfirmTxScreen`

ryanml:
Fixes this fairly noisy sentry error: https://sentry.io/organizations/metamask/issues/2255929540/

We check for the existence of the `stopPropagation` on `event` before calling it, but we should also consider if the `event` is there at all

### DIFF
--- a/ui/pages/confirm-transaction/conf-tx.js
+++ b/ui/pages/confirm-transaction/conf-tx.js
@@ -135,7 +135,7 @@ class ConfirmTxScreen extends Component {
   }
 
   stopPropagation(event) {
-    if (event.stopPropagation) {
+    if (event?.stopPropagation) {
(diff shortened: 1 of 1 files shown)


jpuri (approved):

ui/pages/confirm-transaction/conf-tx.js
```diff
@@ -135,7 +135,7 @@ class ConfirmTxScreen extends Component {
   }
 
   stopPropagation(event) {
-    if (event.stopPropagation) {
+    if (event?.stopPropagation) {
```
darkwing, line 138: Could we remove the `if` and just do `event?.stopPropagation?.()` or is that too gross?

darkwing (approved):

PR #12163, opened 2021-09-20
Fixing ledger steps display (Confirm TX)

ryanml:
Should be checking the current keyring type instead of the account type. Also hides step two for FF users


PeterYinusa:
lgtm

tmashuang (dismissed):

danjm (approved):

tmashuang (approved):

PR #12170, opened 2021-09-21
fix pubnub dep vulnerabiltiy

brad-decker:
Fixes: vulnerability reported in yarn audit 

Issue is in a dependency of pubnub. I opted to add a resolution for superagent-proxy. The rest of the dependencies in that chain are all maintained by the same individual all the way down to the problem dependency. Pubnub has not updated its dependency on superagent-proxy yet. 

Note: The sync with extension feature is currently disabled


adonesky1 (approved):
LGTM!

danjm (approved):

PR #12176, opened 2021-09-22
Fix 12145 - Ledger errors being set as `Error: [object, object]`

### LINKED ISSUE #12145: Broken support of Ledger wallets ~12h ago
Hello guys!

**Describe the bug**
You're breaking ledger nano x wallets today.
24 hours ago all was great but in one moment metamask + ledger get RPC error `{code: -32603, message: 'Error: [object Object]'}` on all operations on all sites.

For example
1. Go to https://mint.sadgirlsbar.io/
2. Click on 'Connect walled button' (with connected hardware ledger account)
3. Push 'mint' button. And try to push 'confirm' button. You've got following error:

`inpage.js:1 MetaMask - RPC Error: Error: [object Object] 
{code: -32603, message: 'Error: [object Object]'}
code: -32603
message: "Error: [object Object]"
[[Prototype]]: Object`

**Expected behavior**
Metamask pass transaction to ledger as it does a day ago.

**Screenshots**
image

 - OS: All desktop OSes, all tested metamask versions 10.1.0, 10.0.3, 10.0.2
 - Hardware Wallet: LEDGER NANO X with latest firmware and latest version Ledger Live 
 - Browser: Latest Chrome
 - MetaMask Version: 10.1.0, 10.0.3, 10.0.2

tmashuang:
Fixes: #12145

Recently change the way we pass error messages from the ledger bridge to the extension, https://github.com/MetaMask/eth-ledger-bridge-keyring/pull/104. When transactions fail from the ledger side for any reason, the error object would get stringified, and present as `Error: [object, object]` in state and the UI as a result. The PR handles the case where if the error is an error object we preserve the message as is, and fall back to a stringified error for other cases. Also, preserves `err.stack` when it is present, else sets it as `err.message`.

### DIFF
--- a/app/scripts/controllers/transactions/tx-state-manager.js
+++ b/app/scripts/controllers/transactions/tx-state-manager.js
@@ -521,7 +521,7 @@ export default class TransactionStateManager extends EventEmitter {
 
     const txMeta = this.getTransaction(txId);
     txMeta.err = {
-      message: typeof error === 'object' ? error.message : error.toString(),
+      message: error.message?.toString() || error.toString(),
       rpc: error.value,
       stack: error.stack,
     };
--- a/app/scripts/controllers/transactions/tx-state-manager.js
+++ b/app/scripts/controllers/transactions/tx-state-manager.js
@@ -521,7 +521,7 @@ export default class TransactionStateManager extends EventEmitter {
 
     const txMeta = this.getTransaction(txId);
     txMeta.err = {
-      message: error.toString(),
+      message: typeof error === 'object' ? error.message : error.toString(),
       rpc: error.value,
       stack: error.stack,
     };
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/transactions/tx-state-manager.js
```diff
@@ -521,7 +521,7 @@ export default class TransactionStateManager extends EventEmitter {
 
     const txMeta = this.getTransaction(txId);
     txMeta.err = {
-      message: error.toString(),
+      message: typeof error === 'object' ? error.message : error.toString(),
```
jpuri, line 524: Small may be `error.message?.toString() || error.toString()`

tmashuang: I haven't seen cases where we need to stringify the error messages from the ledger library having tested most of the common errors. 

The idea to convert to string I believe is to account for any, and all, error formats from different providers as there is not a common error format across the board.

tmashuang: Resolved in https://github.com/MetaMask/metamask-extension/pull/12176/commits/aa13c0f823fa1115d2cee6baa89c39251900665c.

brad-decker (approved):

ryanml (approved):

PR #12200, opened 2021-09-23
Prevent fox mesh JSON warning

darkwing:
Explanation:  The fox mesh was displaying a console warning because its value was correctly `undefined` but the presence check wasn't properly coded to allow an undefined value.

### DIFF
--- a/ui/helpers/utils/build-types.js
+++ b/ui/helpers/utils/build-types.js
@@ -20,7 +20,10 @@ export function isBeta() {
 // the current metamask version (i.e. main, beta, etc.)
 export function getBuildSpecificAsset(assetName) {
   const buildType = process.env.METAMASK_BUILD_TYPE;
-  if (!assetList[buildType] || !Object.keys(assetList[buildType]).includes(assetName)) {
+  if (
+    !assetList[buildType] ||
+    !Object.keys(assetList[buildType]).includes(assetName)
+  ) {
     console.warn(
       `Cannot find asset for build ${buildType}: ${assetName}, returning main build asset`,
     );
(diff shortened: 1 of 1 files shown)


ryanml (dismissed):

ui/helpers/utils/build-types.js
```diff
@@ -20,7 +20,7 @@ export function isBeta() {
 // the current metamask version (i.e. main, beta, etc.)
 export function getBuildSpecificAsset(assetName) {
   const buildType = process.env.METAMASK_BUILD_TYPE;
-  if (!assetList[buildType]?.[assetName]) {
+  if (!assetList[buildType] || !(assetName in assetList[buildType])) {
```
Gudahtt, line 23: Nit: There's a [proposed lint change](https://github.com/MetaMask/eslint-config/pull/207) that would prevent using the `in` operator because it looks up the entire prototype chain, which can be unexpected and confusing.

These two alternatives are functionally equivalent and would pass that rule:
```suggestion
  if (!assetList[buildType] || !Object.keys(assetList[buildType]).includes(assetName)) {
```
or
```suggestion
  if (!assetList[buildType] || Reflect.hasOwnProperty.call(assetList[buildType], assetName)) {
```

Gudahtt (dismissed):
LGTM!

Gudahtt (dismissed):

Gudahtt (approved):

kumavis (approved):

kumavis:
e2e failure looks unrelated

kumavis:
merged via admin access

PR #12202, opened 2021-09-24
Fixing alignment of settings close button

ryanml:
Popup:
Screen Shot 2021-09-23 at 9 03 02 PM

Responsive layout:
https://user-images.githubusercontent.com/8732757/134616537-273af6df-4405-44de-8a34-3b930078a53c.mov

### DIFF
--- a/ui/pages/settings/index.scss
+++ b/ui/pages/settings/index.scss
@@ -22,7 +22,7 @@
 
       flex: 1 0 auto;
 
-      @media screen and (max-width: 575px) {
+      @media screen and (max-width: $break-small) {
         text-overflow: ellipsis;
         overflow: hidden;
         white-space: nowrap;
@@ -106,6 +106,10 @@
         transform: rotate(180deg);
       }
     }
+  }
+
+  &__close-button {
+    margin-left: auto;
   }
 
   &__close-button::after {


ui/pages/settings/index.scss
```diff
@@ -26,7 +26,6 @@
         text-overflow: ellipsis;
         overflow: hidden;
         white-space: nowrap;
-        max-width: 250px;
```
ryanml, line 29: @adonesky1 was there a case this was needed for in particular?

ryanml: Ah, now I see it served a purpose for very long contact names, so this needs to not be removed then but rather re-worked for normal use cases

adonesky1: yes I didn't do this elegantly before. But now removing it is causing thi
Screen Shot 2021-09-24 at 9 58 30 AM
Screen Shot 2021-09-24 at 9 58 24 AM
s:

ryanml: @adonesky1 I pushed changes that should cover all cases:
Screen Shot 2021-09-24 at 8 23 28 AM
Screen Shot 2021-09-24 at 8 23 03 AM
Screen Shot 2021-09-24 at 8 22 53 AM

adonesky1 (approved):
LGTM!

darkwing (approved):

PR #12209, opened 2021-09-24
adding checksum conversion

NiranjanaBinoy:
Fixes no: 1, 6 and  8, in [10.2.0 RC issues doc](https://docs.google.com/document/d/1BiTKWFKhYHRx697RrEs8Xz7R2qj0aZSN58h0JSsoyU0/edit#heading=h.lwrpcld9iufi)

When the token detection is OFF, the address will be converted to a checksum address since the data available in the tokenList will be the static token list from contract-metadata.


tmashuang (approved):
lgtm, restored previous token images with the token detection feature is off.

jpuri (approved):
👍 LGTM

PR #12210, opened 2021-09-24
Hiding refresh list on non-Mainnet networks

NiranjanaBinoy:
Fixes no 7, in [10.2.0 RC issues doc](https://docs.google.com/document/d/1BiTKWFKhYHRx697RrEs8Xz7R2qj0aZSN58h0JSsoyU0/edit#heading=h.lwrpcld9iufi)

The token detection is currently available only on Mainnet, so the `refresh list` link has to be displayed on Mainnet only.

### DIFF
--- a/ui/components/app/import-token-link/import-token-link.component.js
+++ b/ui/components/app/import-token-link/import-token-link.component.js
@@ -23,15 +23,17 @@ export default function ImportTokenLink({ isMainnet }) {
   return (
     <Box className="import-token-link" textAlign={TEXT_ALIGN.CENTER}>
       {isMainnet && (
-        <Button
-          className="import-token-link__link"
-          type="link"
-          onClick={() => detectNewTokens()}
-        >
-          {t('refreshList')}
-        </Button>
+        <>
+          <Button
+            className="import-token-link__link"
+            type="link"
+            onClick={() => detectNewTokens()}
+          >
+            {t('refreshList')}
+          </Button>
+          {t('or')}
+        </>
       )}
-      {isMainnet && ' or '}
       <Button
         className="import-token-link__link"
         type="link"
--- a/ui/components/app/import-token-link/import-token-link.component.js
+++ b/ui/components/app/import-token-link/import-token-link.component.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import { useHistory } from 'react-router-dom';
 import { useMetricEvent } from '../../../hooks/useMetricEvent';
 import { useI18nContext } from '../../../hooks/useI18nContext';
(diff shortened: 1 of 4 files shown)


ui/components/app/import-token-link/import-token-link.component.js
```diff
@@ -21,14 +22,16 @@ export default function ImportTokenLink() {
...
+      {isMainnet && (
+        <Button
+          className="import-token-link__link"
+          type="link"
+          onClick={() => detectNewTokens()}
+        >
+          {t('refreshList')}
+        </Button>
+      )}
+      {isMainnet && ' or '}
```
ryanml, line 34: Looks like the `or` should be localized?

jpuri: Also repeating `isMainnet && ` can be avoided.

ui/components/app/import-token-link/import-token-link.component.js
```diff
@@ -37,8 +40,15 @@ export default function ImportTokenLink() {
...
+          : t('importTokens').charAt(0).toUpperCase() +
+            t('importTokens').slice(1)}
       </Button>
     </Box>
   );
 }
+
+ImportTokenLink.propTypes = {
+  isMainnet: PropTypes.bool,
+};
```
jpuri, line 54: `isMainnet` can be given a default value also.

ui/components/app/import-token-link/import-token-link.component.js
```diff
@@ -37,8 +42,15 @@ export default function ImportTokenLink() {
           addTokenEvent();
         }}
       >
-        {t('importTokens')}
+        {isMainnet
+          ? t('importTokens')
```
ryanml, line 46: Any reason we can't just use `t('importTokensCamelCase')` here for non-mainnet networks??

jpuri (approved):

ui/components/app/import-token-link/import-token-link.component.js
NiranjanaBinoy: The message in `t('importTokensCamelCase')` is `Import Token` but we need `Import token` here

ryanml: or if need be can we add a new translation called `'importTokensLabel'` that equals `"Import tokens"`? I now see it's for consistency with `refreshList`

NiranjanaBinoy: Yes, but I was not sure if it is worth adding multiple entries for the same message just for case change, if I add this one that will be 3rd entry for the same message.

ryanml (approved):

PR #12221, opened 2021-09-27
Add client id to GasFeeController

meppsilon:
Explanation:  
- Adds clientId to the GasFeeController in @metamask/controllers

Manual testing steps:  
  1. On mainnet, enter Swaps view by clicking "Swaps"
  2. Open background.html and see the request to Gas API (/suggestedGasFees)
  3. Notice "x-client-id": "extension" in the Request Headers

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -32,6 +32,7 @@ import { TRANSACTION_STATUSES } from '../../shared/constants/transaction';
 import {
   GAS_API_BASE_URL,
   GAS_DEV_API_BASE_URL,
+  SWAPS_CLIENT_ID,
 } from '../../shared/constants/swaps';
 import { MAINNET_CHAIN_ID } from '../../shared/constants/network';
 import { KEYRING_TYPES } from '../../shared/constants/hardware-wallets';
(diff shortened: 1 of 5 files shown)


app/scripts/metamask-controller.js
```diff
@@ -76,6 +76,8 @@ import MetaMetricsController from './controllers/metametrics';
 import { segment } from './lib/segment';
 import createMetaRPCHandler from './lib/createMetaRPCHandler';
 
+const EXTENSION_CLIENT_ID = 'extension'
```
ryanml, line 79: I see also use this constant here: https://github.com/MetaMask/metamask-extension/blob/develop/ui/pages/swaps/swaps.util.js#L56

Could we make `EXTENSION_CLIENT_ID` a shared constant? Perhaps place it in `shared/constants/network.js`?

ryanml: Also, this line needs a `;` at the end

meppsilon: I'm not sure if `shared/constants/network.js` makes sense because it's not network-specific. How about `shared/constants/app.js`? Or if that is too general, how about `shared/constants/swaps.js`?

ryanml: I think you're right that `shared/constants/swaps.js` makes the most sense!

meppsilon: Awesome, pushed those changes. Let me know what you think.

ryanml (approved):
lgtm! thanks for the changes

danjm (approved):

PR #12223, opened 2021-09-27
Ci test config improvement

kumavis:
- Enables chrome e2e logging by default
- increases firefox ci resource class to same as chrome


Gudahtt (dismissed):
LGTM!

Gudahtt (approved):

ryanml (approved):

PR #12230, opened 2021-09-28
Fix gas control flicker on send screen when switching between EIP-1559 networks

### LINKED ISSUE #12188: Fix gas control flicker on send screen when switching between EIP-1559 networks
If you are on the send screen and switch between mainnet and ropsten (or any two EIP-1559 networks) the gas inputs can sometimes be shown for a few seconds. They should only show up on networks that do no support EIP-1559

https://user-images.githubusercontent.com/7499938/134512069-9b946018-9b2a-4cb6-b665-d55eb1e9dcae.mov

jpuri:
Fixes: #12188

Explanation:  While switching between EIP-1559 networks send screen had gas inputs flickering. Reason is that condition `state.metamask.networkDetails?.EIPS[1559] === true` in method `isEIP1559Network` returns true while data is being fetched for the newly selected network.

I added method `isNotEIP1559Network` which returns true only if data is fetched and network is found to not support EIP-1559. An option here was to change existing method `isEIP1559Network` to return true only after data is fetched. But I avoided that as the old method is widely used and changing it may cause regression somewhere.

Manual testing steps:  
  - On home page select option to send and select an address
  - On send screen and toggle between different networks
  - Gas inputs should not flicker on send screen

### DIFF
--- a/test/data/mock-state.json
+++ b/test/data/mock-state.json
@@ -31,6 +31,9 @@
         "name": "Test Account 2"
       }
     },
+    "networkDetails": {
+      "EIPS": { "1559": true }
+    },
     "cachedBalances": {},
     "incomingTransactions": {},
     "unapprovedTxs": {
--- a/test/data/mock-state.json
+++ b/test/data/mock-state.json
@@ -32,7 +32,9 @@
       }
     },
     "networkDetails": {
-      "EIPS": { "1559": true }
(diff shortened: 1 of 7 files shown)


danjm (approved):
LGTM!

brad-decker (approved):

darkwing (approved):

PR #12232, opened 2021-09-28
Turning OFF token detection for test nets

NiranjanaBinoy:
Fixes no 11, in [10.2.0 RC issues doc](https://docs.google.com/document/d/1BiTKWFKhYHRx697RrEs8Xz7R2qj0aZSN58h0JSsoyU0/edit#)

Since the token detection works only for Mainnet at the moment due to the balance check contract we can add the chainId not equal to Mainnet chainId check to block out other networks and testnets.

### DIFF
--- a/app/scripts/controllers/detect-tokens.js
+++ b/app/scripts/controllers/detect-tokens.js
@@ -33,7 +33,6 @@ export default class DetectTokensController {
     this.network = network;
     this.keyringMemStore = keyringMemStore;
     this.tokenList = tokenList;
-    this.useTokenDetection = this.preferences?.store.getState().useTokenDetection;
     this.selectedAddress = this.preferences?.store.getState().selectedAddress;
     this.tokenAddresses = this.tokensController?.state.tokens.map((token) => {
       return token.address;
--- a/app/scripts/controllers/detect-tokens.js
+++ b/app/scripts/controllers/detect-tokens.js
@@ -32,6 +32,7 @@ export default class DetectTokensController {
     this.network = network;
     this.keyringMemStore = keyringMemStore;
     this.tokenList = tokenList;
+    this.useTokenDetection = this.preferences?.store.getState().useTokenDetection;
     this.selectedAddress = this.preferences?.store.getState().selectedAddress;
     this.tokenAddresses = this.tokensController?.state.tokens.map((token) => {
       return token.address;
@@ -79,7 +80,9 @@ export default class DetectTokensController {
     }
 
     const { tokenList } = this._tokenList.state;
-    if (Object.keys(tokenList).length === 0) {
+    console.log(!this.useTokenDetection || Object.keys(tokenList).length === 0);
+    if (!this.useTokenDetection || Object.keys(tokenList).length === 0) {
+      console.log(`this.useTokenDetection: ${this.useTokenDetection}`);
       return;
     }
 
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/detect-tokens.js
```diff
@@ -32,6 +33,7 @@ export default class DetectTokensController {
     this.network = network;
     this.keyringMemStore = keyringMemStore;
     this.tokenList = tokenList;
+    this.useTokenDetection = this.preferences?.store.getState().useTokenDetection;
```
adonesky1, line 36: @NiranjanaBinoy I'm not clear on why this change was necessary?

NiranjanaBinoy: it is not, I was trying another fix at first then switched to this one, this entry was made for the other attempt. I will remove it now.

PeterYinusa (dismissed):

app/scripts/controllers/detect-tokens.js
```diff
@@ -79,7 +80,10 @@ export default class DetectTokensController {
     }
 
     const { tokenList } = this._tokenList.state;
-    if (Object.keys(tokenList).length === 0) {
```
adonesky1, line 82: Do we want to get rid of this check?

NiranjanaBinoy: Even though the tokenList will be empty for testnets when the usetokenDetection flag is ON when the flag is OFF the token List will be holding the static tokenList which mimics the use of contract-metadata before the introduction of token detection. So the above test was failing and the token detection was getting triggered while on testnets. 

The token detection is currently available only on Mainnet, so the new check will make sure the detection will be triggered only for Mainnet. Once we have the token detection enabled in other networks we will need to update the current check I have added to make sure token detection will work on all those networks.

We can also do something like this, but I am wondering if this is worth it considering the current scenario.

    if (this._network.store.getState().provider.chainId !== MAINNET_CHAIN_ID && (!this.useTokenDetection || Object.keys(tokenList).length === 0)) {
      return;
    }

adonesky1: Can we be quite confident that tokenList will never have a length of 0 for Mainnet (or any other network for which we might enable token detection down the line)?

NiranjanaBinoy: If tokenList length is 0, the token detection will not work as we are using the token address from tokenList for balance check. When token detection is ON, as of now the testnets will have an empty token list. When token detection is OFF the tokenList will have the data from contract-metadata. If tokenList is empty for any network the token detection will be disabled but  we can always add any token deployed in a network via `custom token`

danjm: I think the alternative suggested in this comment is good: https://github.com/MetaMask/metamask-extension/pull/12232/files#r717893459

app/scripts/controllers/detect-tokens.js
```diff
@@ -79,7 +80,13 @@ export default class DetectTokensController {
...
+    // since the token detection is currently enabled only on Mainnet
+    // we can use the chainId check to ensure token detection is not triggered for any other network
+    // but once the balance check contract for other networks are deploayed and ready to use, we need to update this check.
+    if (
+      this._network.store.getState().provider.chainId !== MAINNET_CHAIN_ID ||
```
danjm, line 87: looks correct (note to other readers that this is different from the above suggestion)

danjm (approved):
LGTM

app/scripts/controllers/detect-tokens.js
adonesky1: this is what I had in mind!

adonesky1 (approved):
LGTM!

PR #12233, opened 2021-09-28
Fix assets for beta dev build

Gudahtt:
The MetaMask logo used for beta development builds was wrong. The lock screen (and any other place using the `@metamask/logo` logo) showed the correct logo, but all of our static assets used the "regular" logo.

Now the beta logo should be used everywhere for beta development builds.

Manual testing steps:  
  - Run `yarn start --build-type beta
  - See that the logos are correct

### DIFF
--- a/development/build/static.js
+++ b/development/build/static.js
@@ -20,21 +20,24 @@ module.exports = function createStaticAssetTasks({
     shouldIncludeLockdown,
   );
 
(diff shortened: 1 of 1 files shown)


rekmarks (approved):
LGTM!

darkwing (approved):

PR #12238, opened 2021-09-29
Changed snap imports to use `@metamask/` org

shanejonas:
We are now using the `@metamask` org for these packages, this changes the imports and package.json to use those new packages.


package.json
```diff
@@ -107,13 +107,14 @@
...
     "@metamask/eth-token-tracker": "^3.0.1",
     "@metamask/etherscan-link": "^2.1.0",
+    "@metamask/iframe-execution-environment-service": "^0.1.2",
     "@metamask/jazzicon": "^2.0.0",
     "@metamask/logo": "^3.0.1",
     "@metamask/obs-store": "^5.0.0",
     "@metamask/post-message-stream": "^4.0.0",
     "@metamask/providers": "^8.1.1",
+    "@metamask/snap-controllers": "^0.1.2",
     "@mm-snap/controllers": "^0.0.7",
```
rekmarks, line 117: This should get deleted!

package.json
```diff
@@ -107,13 +107,14 @@
...
-    "@mm-snap/iframe-execution-environment-service": "^0.0.8",
     "@mm-snap/rpc-methods": "^0.0.6",
```
rekmarks, line 118: This should also get deleted!

rekmarks (changes requested):
Maybe do a find-all of `@mm-snap` to make sure we get rid of all references to the old packages.

Anon79 (dismissed):

rekmarks (approved):
LGTM!

PR #12242, opened 2021-09-29
Lavamoat build system integration for WebApp

kumavis:
🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 
the next step in incrementally locking down metamask:
providing the build process with a switch for adding **lavamoat** to each page
currently **does not** apply lavamoat to any page
🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 🦊 🌋 

### summary of code changes
- in `app/*.html` include a `if`-block to optionally apply lavamoat by loading the commonjs or lavamoat runtime. commonjs still needs an external lockdown script
- in `build/scripts.js` change packer to `lavapack`, append policy loader, and render html with the `useLavamoat` switch
-  in `lavamoat/browserify`, add combined background and ui policies (ui policy is not enforced yet)
- in `lavamoat/node` update policies to include running lavamoat and other added deps
- in `package.json` add new scripts for generating lavamoat policies and add deps
- in `test/e2e/tests/contract-interactions.spec.js` add delays to fix race condition
- in `yarn.lock`lavamoat dep additions

### summary of work flow changes
- after modifying dependencies or patches used in the background or ui, run `yarn lavamoat:auto` to update the policies. these policies are not applied yet, but makes PRs consistent with the future requirement where we will soon apply these policies when lavamoat is enabled.

### things to be aware of
- there is a change to workflow, see above
- this pr was designed to be minimal and easy to revert
- this version of LavaMoat was prepared to not require eval and requires no change to the manifest.
- while this applies no protection to background or ui, it does changes the runtime of the ui and background (update to lavapack). All non-lavamoat pages now runs in a ses-shim like structure where a `with` statement is wrapping the `globalThis`. This is done to have a matching module initialization structure to the lavamoat protected components.
- the application of this structure may have performance impacts, especially on low resource devices. This is already noticed in CI as additional timeouts were necessary. However, in QA on my machines performance has felt normal. Take this with a grain of salt.
- this change is in preparation to applying lavamoat protections and bringing significant supplychain security enhancements to metamask
- I am committed to polishing the developer experience in response to feedback

### TODO:
- [x] add debug info for policy validation failure
https://github.com/MetaMask/metamask-extension/pull/12259
- [x] include `lockdown-run.js` additions beyond lockdown
https://github.com/MetaMask/metamask-extension/pull/12258
- [x] QA popup init time, esp firefox


development/build/scripts.js
```diff
@@ -306,36 +329,58 @@ function createFactoredBuild({
...
-            renderHtmlFile(
-              'notification',
+            renderHtmlFile({
+              htmlName: 'popup',
               groupSet,
               commonSet,
               browserPlatforms,
-            );
-            renderHtmlFile('home', groupSet, commonSet, browserPlatforms);
+              useLavamoat: false,
```
Gudahtt, line 342: It would be nice to pass in `useLavamoat` as a CLI flag

kumavis:
TODO: review `lockdown-run.js`

Gudahtt:
Lavamoat policy validation is here: https://github.com/MetaMask/metamask-extension/blob/develop/.circleci/config.yml#L171
Which calls this script: https://github.com/MetaMask/metamask-extension/blob/develop/.circleci/scripts/validate-lavamoat-policy.sh

So because the `lavamoat:auto` script now automatically updates both, I think this PR works with that CI step already.

kumavis:
request for QA on firefox esp wrt to popup init perf

development/build/scripts.js
```diff
@@ -259,7 +261,22 @@ function createFactoredBuild({
...
+    // setup lavamoat
+    // lavamoat will add lavapack but it will be removed by bify-module-groups
+    // we will re-add it later by installing a lavapack runtime
+    const lavamoatOpts = {
+      policy: path.resolve(__dirname, '../../lavamoat/browserify/policy.json'),
+      policyOverride: path.resolve(
+        __dirname,
+        '../../lavamoat/browserify/policy-override.json',
+      ),
+      writeAutoPolicy: process.env.WRITE_AUTO_POLICY,
```
Gudahtt, line 273: What does this environment variable do?

kumavis:
add debugging information if policy fails https://github.com/MetaMask/metamask-extension/blob/develop/.circleci/scripts/validate-lavamoat-policy.sh#L9

development/build/scripts.js
```diff
@@ -259,7 +261,22 @@ function createFactoredBuild({
...
+    };
+    Object.assign(bundlerOpts, lavamoatBrowserify.args);
```
Gudahtt, line 275: I'm curious to see what the impact of these two lines is on the final bundle :thinking: @kumavis suspects we will see more duplicated modules because deduping is now disabled.

kumavis: looks like its already disabled by `bify-module-groups` https://npmfs.com/package/bify-module-groups/1.0.0/src/plugin.js#L6

development/build/scripts.js
kumavis: im not sure the usecase, can add later. would need to specify it for every entry/webpage

Gudahtt: I mean so the Lavamoat build can be tested without modifying this file. You'd have to go through and swap every `useLavamoat: false,` to `useLavamoat: true,` right?

brad-decker:
Other than the popup init time issue I am 👍🏻 on this

development/build/scripts.js
kumavis: makes sense

kumavis:
@brad-decker did you see a difference in pop perf time with lavamoat disabled as it is now?

kumavis:
removing e2e test delays that are likely only required for actually turning lavamoat on

kumavis:
ive gotten a few QA reports that firefox popup startup perf looks fine

brad-decker (dismissed):
With QA and my own testing not finding popup delays I'm good with this 💯

kumavis:
removing delays failed :crying_cat_face:

brad-decker (approved):
LGTM

ryanml (approved):
lgtm after testing!

PR #7004, opened 2019-08-13
Connect distinct accounts per site

danfinlay:
This absolutely tremendous PR is finally ready for review.

For the complete picture, be sure to refer to the [inpage provider](https://github.com/MetaMask/metamask-inpage-provider).

Thanks to @Gudahtt for his extensive review work so far!

### Code Owners
- [rpc-cap](https://github.com/MetaMask/rpc-cap): @danfinlay and @rekmarks 
- UI: @danjm 
- Background and [inpage provider](https://github.com/MetaMask/metamask-inpage-provider): @rekmarks 

### Changes (Non-Exhaustive)
- Accounts are now connected individually to different sites, regardless of the globally selected account in the UI
  - Only one account can be connected to any one site at a time
- `inpage.js` has been simplified, and much of its functionality has moved to the [inpage provider module](https://github.com/MetaMask/metamask-inpage-provider)
- The `PermissionsController` has replaced the `ProviderApprovalController`
  - `PermissionsController` wraps [rpc-cap](https://github.com/MetaMask/rpc-cap), and should be the single source of truth for the permissions of all external domains (be they dapps, other extensions, or in the future, plugins)

PR #12243, opened 2021-09-29
Permission System 2.0

rekmarks:
# Permission System 2.0

## Background

This PR migrates the extension permission system to [the new `PermissionController`](https://github.com/MetaMask/snaps-skunkworks/tree/main/packages/controllers/src/permissions).
The original permission system, based on [`rpc-cap`](https://github.com/MetaMask/rpc-cap), introduced [`ZCAP-LD`](https://w3c-ccg.github.io/zcap-ld/)-like permissions to our JSON-RPC stack.
We used it to [implement](https://github.com/MetaMask/metamask-extension/pull/7004) what we called "LoginPerSite" in [version 7.7.0](https://github.com/MetaMask/metamask-extension/releases/tag/v7.7.0) of the extension, which enabled the user to choose which accounts, if any, should be exposed to each dapp.
While that was a worthwhile feature in and of itself, we wanted a permission _system_ in order to enable everything we are going to with Snaps.
Unfortunately, the original permission system was difficult to use, and necessitated the creation of the original `PermissionsController` (note the "s"), which was more or less a wrapper for `rpc-cap`.

With this PR, we shake off the yoke of the original permission system, in favor of the modular, self-contained, ergonomic, and more mature permission system 2.0.

> **ATTN**
> Before continuing, reviewers should read [the `PermissionController` readme](https://github.com/MetaMask/snaps-skunkworks/tree/main/packages/controllers/src/permissions/README.md).
> Note that the `PermissionController` and `SubjectMetadataController` are currently shipped via `@metamask/snap-controllers`. This is a temporary state of affairs, and we'll move them to `@metamask/controllers` once they've landed in prod. Snaps / Core is also adding non-breaking features to them at the moment.

## Changes in Detail

First, the changes in this PR are not as big as they seem. Roughly half of the additions in this PR are fixtures in the test for the new migration (number 68), and a significant portion of the remaining ~2500 lines are due to find-and-replace changes in other test fixtures and UI files.

- The extension `PermissionsController` has been deleted, and completely replaced with the new `PermissionController` from [`@metamask/snap-controllers`](https://www.npmjs.com/package/@metamask/snap-controllers).
- The original `PermissionsController` "domain metadata" functionality is now managed by the new `SubjectMetadataController`, also from [`@metamask/snap-controllers`](https://www.npmjs.com/package/@metamask/snap-controllers).
- The permission activity and history log controller has been renamed `PermissionLogController` and has its own top-level state key, but is otherwise functionally equivalent to the existing implementation.
- Migration number 68 has been added to account for the new state changes.
- The tests in `app/scripts/controllers/permissions` have been migrated from `mocha` to `jest`.

Reviewers should focus their attention on the following files:

- `app/scripts/`
  - `metamask-controller.js`
    - This is where most of the integration work for the new `PermissionController` occurs.
      Some functions that were internal to the original controller were moved here.
  - `controllers/permissions/`
    - `selectors.js`
      - These selectors are for `ControllerMessenger` selector subscriptions. The actual subscriptions occur in `metamask-controller.js`. See the `ControllerMessenger` implementation for details.
    - `specifications.js`
      - The caveat and permission specifications are required by the new `PermissionController`, and are used to specify the `eth_accounts` permission and its JSON-RPC method implementation.
        See the `PermissionController` readme for details.
  - `migrations/068.js`
    - The new state should be cross-referenced with the controllers that manage it.
      The accompanying tests should also be thoroughly reviewed.

Some files may appear new but have just moved and/or been renamed:

- `app/scripts/lib/rpc-method-middleware/handlers/request-accounts.js`
  - This was previously implemented in `controllers/permissions/permissionsMethodMiddleware.js`.
- `test/mocks/permissions.js`
  - A truncated version of `test/mocks/permission-controller.js`.


app/scripts/metamask-controller.js
```diff
@@ -1090,36 +1225,89 @@ export default class MetamaskController extends EventEmitter {
       initializeThreeBox: nodeify(this.initializeThreeBox, this),
 
       // permissions
-      approvePermissionsRequest: nodeify(
-        permissionsController.approvePermissionsRequest,
-        permissionsController,
+      clearPermissions: permissionController.clearState.bind(
```
Gudahtt, line 1228: This API method is apparently unused! I have removed it in https://github.com/MetaMask/metamask-extension/pull/12731

app/scripts/metamask-controller.js
```diff
@@ -2527,40 +2737,51 @@ export default class MetamaskController extends EventEmitter {
         registerOnboarding: this.onboardingController.registerOnboarding,
       }),
     );
+    engine.push(this.permissionLogController.createMiddleware());
```
Gudahtt, line 2740: I noticed that previously this middleware was only added for sites, not for UI connections. But I guess we'd want our log to include permission changes made within our UI as well :thinking:

rekmarks: Hm, yeah, I can't recall why we didn't do it in the first place, but I don't see why not! (It's also possible we were doing it in some convoluted way in the original `PermissionsController`, but w/e.)

app/scripts/lib/rpc-method-middleware/createMethodMiddleware.js
```diff
@@ -35,7 +32,18 @@ export default function createMethodMiddleware(opts) {
...
+        return await implementation(
+          req,
+          res,
+          next,
+          end,
+          selectHooks(hooks, hookNames),
+        );
+      } catch (error) {
+        return end(error);
+      }
```
rekmarks, line 46: The addition of the `try`/`catch` block fixes a pre-existing issue where errors thrown by async middleware functions would be completely uncaught.

test/e2e/fixtures/connected-state/state.json
```diff
@@ -96,90 +101,83 @@
...
             }
-          ]
+          }
         }
       }
     }
   },
   "meta": {
-    "version": 47
+    "version": 68
```
rekmarks, line 181: As it turns out, migrations run for our fixture state, and migration 68 clobbered the new permissions state I defined in this fixture. Therefore, I had to bump this and update the fixture to match version 68 state well enough for tests to pass.

ui/selectors/permissions.js
```diff
@@ -264,22 +260,10 @@ export function getLastConnectedInfo(state) {
...
-      if (counts[host]) {
-        counts[host] += 1;
-      } else {
-        counts[host] = 1;
-      }
-    }
-    return counts;
-  }, {});
-}
-
```
rekmarks, line 280: We're no longer storing the `host` in the subject metadata, and this selector was already quite pointless anyway.

ui/pages/permissions-connect/redirect/permissions-redirect.component.js
```diff
@@ -47,9 +47,9 @@ export default function PermissionsRedirect({ domainMetadata }) {
 }
 
 PermissionsRedirect.propTypes = {
-  domainMetadata: PropTypes.shape({
+  subjectMetadata: PropTypes.shape({
     extensionId: PropTypes.string,
-    icon: PropTypes.string,
+    iconUrl: PropTypes.string,
     host: PropTypes.string.isRequired,
```
Gudahtt, line 53: This `host` isn't here anymore

```suggestion
```

ui/pages/permissions-connect/permissions-connect.stories.js
```diff
@@ -21,10 +21,9 @@ export const ChooseAccountComponent = () => {
           '0xbe0eb53f46cd790cd13851d5eff43d12404d33e8',
         ])
       }
-      targetDomainMetadata={{
+      targetSubjectMetadata={{
         host: 'gnosis-safe.io',
```
Gudahtt, line 25: Another extra `host`
```suggestion
```

ui/pages/permissions-connect/permissions-connect.stories.js
```diff
@@ -56,11 +55,10 @@ export const PermissionPageContainerComponent = () => {
   return (
     <div className="page-container permission-approval-container">
       <PermissionPageContainerContent
-        domainMetadata={{
+        subjectMetadata={{
           extensionId: '1',
           host: 'gnosis-safe.io',
```
Gudahtt, line 60: Another extra `host`
```suggestion
```

ui/pages/permissions-connect/permissions-connect.component.js
```diff
@@ -31,9 +31,9 @@ export default class PermissionConnect extends Component {
     connectPath: PropTypes.string.isRequired,
     confirmPermissionPath: PropTypes.string.isRequired,
     page: PropTypes.string.isRequired,
-    targetDomainMetadata: PropTypes.shape({
+    targetSubjectMetadata: PropTypes.shape({
       extensionId: PropTypes.string,
-      icon: PropTypes.string,
+      iconUrl: PropTypes.string,
       host: PropTypes.string.isRequired,
```
Gudahtt, line 37: Another extra `host`
```suggestion
```

ui/pages/permissions-connect/choose-account/choose-account.component.js
```diff
@@ -31,9 +31,9 @@ export default class ChooseAccount extends Component {
     cancelPermissionsRequest: PropTypes.func.isRequired,
     permissionsRequestId: PropTypes.string.isRequired,
     selectedAccountAddresses: PropTypes.object.isRequired,
-    targetDomainMetadata: PropTypes.shape({
+    targetSubjectMetadata: PropTypes.shape({
       extensionId: PropTypes.string,
-      icon: PropTypes.string,
+      iconUrl: PropTypes.string,
       host: PropTypes.string.isRequired,
```
Gudahtt, line 37:
```suggestion
```

ui/components/app/connected-sites-list/connected-sites-list.component.js
```diff
@@ -9,56 +9,53 @@ export default class ConnectedSitesList extends Component {
...
-    if (domain.extensionId) {
+  getSubjectDisplayName(subject) {
+    if (subject.extensionId) {
       return this.context.t('externalExtension');
     }
 
-    return this.props.domainHostCount[domain.host] > 1
-      ? domain.origin
-      : stripHttpSchemes(domain.origin);
+    return stripHttpSchemes(subject.origin);
```
Gudahtt, line 59: The host count was added so that we couldn't have multiple entries that appear different but are the same. For example if we had the same origin listed with different protocols (`http` and `https`) they'd look identical here.

Instead we could only strip `https://`, leaving `http://` as-is. That might be better anyway, so the user can see clearly that their connection isn't encrypted.

As a further improvement, we should probably also leave `https//` there if there is a non-standard port set. Otherwise it could be misleading (e.g. `foo.com:80` might look innocuous, but `https://foo.com:80` is more obviously non-stardard).

ui/components/app/connected-sites-list/connected-sites-list.component.js
```diff
@@ -9,56 +9,53 @@ export default class ConnectedSitesList extends Component {
...
 
   static propTypes = {
-    connectedDomains: PropTypes.arrayOf(
+    connectedSubjects: PropTypes.arrayOf(
       PropTypes.shape({
         name: PropTypes.string,
-        icon: PropTypes.string,
+        iconUrl: PropTypes.string,
         origin: PropTypes.string,
         host: PropTypes.string,
```
Gudahtt, line 17:
```suggestion
```

ui/hooks/useOriginMetadata.js
```diff
@@ -1,14 +1,12 @@
 import { useSelector } from 'react-redux';
-import { getDomainMetadata } from '../selectors';
+import { getSubjectMetadata } from '../selectors';
 
 /**
  * @typedef {Object} OriginMetadata
  * @property {string} host - The host of the origin
```
Gudahtt, line 6:
```suggestion
```

Gudahtt: Actually this is still being set below, and is derived from the origin. But I don't think we need it. It's never used.

ui/components/app/permission-page-container/permission-page-container.component.js
```diff
@@ -13,7 +13,7 @@ export default class PermissionPageContainer extends Component {
     allIdentitiesSelected: PropTypes.bool,
     request: PropTypes.object,
     requestMetadata: PropTypes.object,
-    targetDomainMetadata: PropTypes.shape({
+    targetSubjectMetadata: PropTypes.shape({
       extensionId: PropTypes.string,
       icon: PropTypes.string,
       host: PropTypes.string.isRequired,
```
Gudahtt, line 19:
```suggestion
```

test/mocks/permissions.js
```diff
@@ -0,0 +1,369 @@
...
+  '0x0dcd5d886577d5081b0c52e242ef29e70be3e7bc',
+  '0xc42edfcc21ed14dda456aa0756c153f7985d8813',
+  '0x7ae1cdd37bcbdb0e1f491974da8022bfdbf9c2bf',
+  '0xcc74c7a59194e5d9268476955650d1e285be703c',
+]);
+
+const SUBJECTS = {
+  a: { origin: 'https://foo.xyz', host: 'foo.xyz' },
+  b: { origin: 'https://bar.abc', host: 'bar.abc' },
+  c: { origin: 'https://baz.def', host: 'baz.def' },
```
Gudahtt, line 20: We don't use the `host` anywhere
```suggestion
  a: { origin: 'https://foo.xyz' },
  b: { origin: 'https://bar.abc' },
  c: { origin: 'https://baz.def' },
```

ui/components/app/alerts/unconnected-account-alert/unconnected-account-alert.js
```diff
@@ -97,7 +97,7 @@ const UnconnectedAccountAlert = () => {
       footer={footer}
     >
       <ConnectedAccountsList
-        accountToConnect={selectedIdentity}
+        accountToConnect={accountToConnect}
```
Gudahtt, line 100: Why change this? I thought this alert was only shown if the current selected account was the same as the "account to connect".

yarn.lock
```diff
@@ -2759,6 +2768,15 @@
     readable-stream "^2.2.2"
     through2 "^2.0.3"
 
+"@metamask/obs-store@^6.0.2":
```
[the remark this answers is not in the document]
rekmarks: Indeed. There's no reason for it! https://github.com/MetaMask/snaps-skunkworks/pull/144

test/e2e/fixtures/connected-state/state.json
Gudahtt: Why update the fixtures at all? The e2e test fixtures are generally left as-is, because they help validate that the migrations are working properly.

jest.strict.config.js
```diff
@@ -0,0 +1,21 @@
+module.exports = {
+  displayName: 'strict',
+  collectCoverageFrom: ['<rootDir>/app/scripts/controllers/permissions/*.js'],
+  coverageDirectory: './jest-coverage/strict/',
+  coverageReporters: ['json', 'lcov', 'text', 'clover'],
+  coverageThreshold: {
+    './app/scripts/controllers/permissions/*.js': {
```
Gudahtt, line 7: Does this have to be a separate config? We could merge it with the base config, since the only difference is the coverage threshold.

app/scripts/background.js
```diff
@@ -533,13 +533,8 @@ function setupController(initState, initLangCode) {
...
 
-    // We're specifcally avoid using approvalController directly for better
-    // Error support during rejection
-    Object.keys(
-      controller.permissionsController.approvals.state.pendingApprovals,
-    ).forEach((approvalId) =>
-      controller.permissionsController.rejectPermissionsRequest(approvalId),
-    );
+    // Finally, reject all approvals managed by the ApprovalController
+    controller.approvalController.clear();
```
Gudahtt, line 537: I had recommended rejecting these via the `permissionsController` in this comment: https://github.com/MetaMask/metamask-extension/pull/12643#discussion_r747590236

I hadn't realized there was a `clear` method that did reject them all properly. Thought the error used by `clear` might not be especially appropriate (`ethErrors.rpc.resourceUnavailable('The request was rejected; please try again.',);`). That makes it sound like it was rejected by the system for because we had insufficient resources to handle it, rather than because the user explicitly rejected it.

app/scripts/controllers/permissions/enums.js
```diff
@@ -1,28 +1,15 @@
...
-};
-
 export const NOTIFICATION_NAMES = {
   accountsChanged: 'metamask_accountsChanged',
   unlockStateChanged: 'metamask_unlockStateChanged',
   chainChanged: 'metamask_chainChanged',
 };
 
 export const LOG_IGNORE_METHODS = [
+  'wallet_addEthereumChain',
```
Gudahtt, line 10: Why are these now ignored?

rekmarks: IMO they should always have been ignored, since they're not strictly related to permissions. I'm open to keeping them, though.

app/scripts/migrations/068.js
```diff
@@ -0,0 +1,159 @@
...
+  const { PermissionsController = {}, PermissionsMetadata = {} } = state;
+
+  const {
+    domainMetadata = {},
+    permissionsHistory = {},
+    permissionsLog = [],
+  } = PermissionsMetadata;
+
+  return {
+    ...state,
```
Gudahtt, line 31: Aren't `PermissionsController` and `PermissionsMetadata` still in this state? This is returning the old state alongside the new state, rather than replacing the old with the new.

app/scripts/controllers/permissions/enums.js
Gudahtt: Hmm. I wasn't aware that this log was meant just for permission-related calls. In any case, it's probably better to at least leave this as-is for this PR, to minimize the changes.

test/e2e/fixtures/connected-state/state.json
rekmarks: Yeah, I restored all fixtures.

app/scripts/background.js
rekmarks: That makes sense, although with the new `PermissionController`, pending approvals are only stored in the `ApprovalController`. That said, we can improve upon this very soon: https://github.com/MetaMask/controllers/pull/656

app/scripts/controllers/permissions/enums.js
rekmarks: Fair enough! b65e88909

app/scripts/migrations/068.js
rekmarks: This was fixed as of e48e20a

jest.strict.config.js
rekmarks: No, they don't have to be separate: 5bfbe8e15

app/scripts/migrations/068.js
```diff
@@ -0,0 +1,163 @@
...
+    {},
+  );
+
+  return {
+    subjects,
+  };
+}
+
+function getPermissionLogControllerState(permissionsHistory, permissionsLog) {
+  return {
```
darkwing, line 120: Nit but could we rename the arguments so we can just return `{ permissionHistory, permissionActivityLog }`?  No big deal, really.

package.json
```diff
@@ -29,16 +29,13 @@
     "test:unit:global": "mocha test/unit-global/*.test.js",
     "test:unit:mocha": "mocha './app/**/*.test.js'",
     "test:unit:lax": "mocha --config '.mocharc.lax.js' './app/**/*.test.js'",
```
Gudahtt, line 31: If we don't have the strict tests anymore, can we get rid fo this too?

app/scripts/migrations/068.js
rekmarks: Ah, I kept the different parameter names to highlight that we are changing the names of the keys as part of the migration. I just got rid of the function here instead: 5774f1b0c

package.json
rekmarks: Yes we can! 074a885c8

app/scripts/migrations/068.js
```diff
@@ -0,0 +1,156 @@
...
+    (transformed, [origin, metadata]) => {
+      const { name, icon = null, extensionId = null, ...other } = metadata;
+
+      // We're getting rid of these.
+      delete other.lastUpdated;
+      delete other.host;
+
+      if (origin) {
+        transformed[origin] = {
+          name: name || origin,
```
Gudahtt, line 141: Why include a fallback name? I don't think we did this before, unless I'm missing something.

app/scripts/migrations/068.test.js
```diff
@@ -0,0 +1,2491 @@
...
+    },
+  };
+}
+
+function getOldState() {
+  return {
+    FooController: { a: 'b' }, // just to ensure it's not touched
+    PermissionsController: {
+      domains: {
+        'https://faucet.metamask.io': {
```
Gudahtt, line 362: Is it necessary to have such a large fixture? This makes it difficult to tell what has changed.

app/scripts/migrations/068.test.js
```diff
@@ -0,0 +1,2491 @@
...
+      data: {
+        PermissionsController: getOldState().PermissionsController,
+      },
+    };
+
+    const newStorage = await migration68.migrate(oldStorage);
+    const { PermissionController } = newStorage.data;
+
+    expect(PermissionController).toStrictEqual(
+      getExpectedPermissionControllerState(),
```
Gudahtt, line 57: This test is very difficult to read. The size of the fixtures plus the fact that they're both declared elsewhere is leaving me flipping back and forth between them to see what this is asserting.

app/scripts/background.js
Gudahtt: What do you mean by "pending approvals are only stored in the `ApprovalController`"? That was the case with the previous permission controller as well.

rekmarks: Ah, I think pending permissions requests were/are also stored in `rpc-cap` under a state key in there, but I see now that we were just getting the IDs from the approval controller regardless. So, never mind that comment!

For present purposes, although I believe that the approval controller is only used for permission requests at the moment, I think it's bad to maintain code based on that assumption, and we should just use `clear` and wait for the improved error message to land.

app/scripts/controllers/permissions/specifications.js
```diff
@@ -0,0 +1,261 @@
...
+            } error: Received unexpected caveats:\n${JSON.stringify(
+              permissionOptions.caveats,
+              null,
+              2,
+            )}`,
+          );
+        }
+
+        // This value will be further validated as part of the caveat.
+        if (!requestData.approvedAccounts) {
```
Gudahtt, line 101: Should we also validate that this is an array of strings? Or is that validation performed elsewhere already?

app/scripts/controllers/permissions/specifications.js
```diff
@@ -0,0 +1,261 @@
...
+        const { caveats } = permission;
+        if (
+          !caveats ||
+          caveats.length !== 1 ||
+          caveats[0].type !== CaveatTypes.restrictReturnedAccounts
+        ) {
+          throw new Error(
+            `${
+              PermissionKeys.eth_accounts
+            } error: Invalid caveats:\n${JSON.stringify(caveats, null, 2)}`,
```
Gudahtt, line 156: This error could also be long and could contain ethereum addresses

darkwing (dismissed):
Mark's review comments need addressing.  Code looks good and testing worked great.  Well done!

app/scripts/controllers/permissions/specifications.js
Gudahtt: Oh, I guess this is taken care of by the caveat validator. Nevermind then.

app/scripts/metamask-controller.js
```diff
@@ -738,10 +774,76 @@ export default class MetamaskController extends EventEmitter {
...
+   * event subscriptions. See the relevant @metamask/controllers documentation
+   * for more information.
+   *
+   * Note that account-related notifications emitted when the extension
+   * becomes unlocked are handled in MetaMaskController._onUnlock.
+   */
+  setupControllerEventSubscriptions() {
+    const getCurrentAccounts = async (origin, newAccounts) => {
+      return newAccounts.length < 2
+        ? newAccounts
```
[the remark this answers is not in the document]
Gudahtt: Wait a minute... why doesn't `getPermittedAccountsByOrigin` return accounts that are sorted correctly, if this is the explanation?

app/scripts/metamask-controller.js
```diff
@@ -738,10 +774,76 @@ export default class MetamaskController extends EventEmitter {
...
+        }
+      }
+    });
+
+    // This handles account changes every time relevant permission state
+    // changes, for any reason.
+    this.controllerMessenger.subscribe(
+      `${this.permissionController.name}:stateChange`,
+      async (currentValue, previousValue) => {
+        if (this.isUnlocked()) {
```
Gudahtt, line 832: Presumably we still want the `this.permissionLogController.updateAccountsHistory(origin, newAccounts);` call to happen even if the wallet is locked. That call is part of `handleAccountsChange`.

package.json
```diff
@@ -28,17 +28,13 @@
...
     "test:unit:mocha": "mocha './app/**/*.test.js'",
-    "test:unit:lax": "mocha --config '.mocharc.lax.js' './app/**/*.test.js'",
-    "test:unit:strict": "mocha './app/scripts/controllers/permissions/*.test.js'",
     "test:e2e:chrome": "SELENIUM_BROWSER=chrome node test/e2e/run-all.js",
     "test:e2e:chrome:metrics": "SELENIUM_BROWSER=chrome node test/e2e/run-e2e-test.js test/e2e/metrics.spec.js",
     "test:e2e:firefox": "SELENIUM_BROWSER=firefox node test/e2e/run-all.js",
     "test:e2e:firefox:metrics": "SELENIUM_BROWSER=firefox node test/e2e/run-e2e-test.js test/e2e/metrics.spec.js",
     "test:e2e:single": "node test/e2e/run-e2e-test.js",
-    "test:coverage": "nyc --silent --check-coverage yarn test:unit:strict && nyc --silent --no-clean yarn test:unit:lax && nyc report --reporter=text --reporter=html",
+    "test:coverage:mocha": "nyc --reporter=text --reporter=html yarn test:unit:mocha",
```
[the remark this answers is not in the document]
rekmarks: I think we should omit it. I'm not sure why we used `--silent` for `test:unit:strict`, but for the `mocha` tests I believe it was to support combining coverage from multiple test scripts as described [here](https://github.com/istanbuljs/nyc#combining-reports-from-multiple-runs). Now, there is only one script, and I think we want to output the reports.

Here's what `nyc --help` has to say about `--silent`:

```text
  --silent, -s                don't output a report after tests finish
                              running           [boolean] [default: false]
```

test/e2e/fixtures/connected-state/state.json
```diff
@@ -1,7 +1,8 @@
 {
   "data": {
     "AppStateController": {
-      "connectedStatusPopoverHasBeenShown": false
+      "connectedStatusPopoverHasBeenShown": false,
+      "recoveryPhraseReminderHasBeenShown": true
```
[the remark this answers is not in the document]
rekmarks: No, this was accidentally left over. Reverted here: bcc123089

app/scripts/migrations/068.js
rekmarks: The name is not optional per the `SubjectMetadataController`'s state type. I'm not sure if we did it before, but I think it's fine?

In retrospect, the `name` should perhaps be optional in the controller, but we can always go back and fix that.

app/scripts/migrations/068.test.js
rekmarks: Hm, probably not. I cut down on it significantly here: cd4d3a5d2

app/scripts/migrations/068.test.js
rekmarks: Yeah, fair enough. I inlined the expected state getters here: f8a006ec8

Combined with pruned fixture from cd4d3a5, is that good enough?

app/scripts/controllers/permissions/specifications.js
```diff
@@ -0,0 +1,261 @@
...
+      targetKey: PermissionKeys.eth_accounts,
+      allowedCaveats: [CaveatTypes.restrictReturnedAccounts],
+
+      factory: (permissionOptions, requestData) => {
+        if (Array.isArray(permissionOptions.caveats)) {
+          throw new Error(
+            `${
+              PermissionKeys.eth_accounts
+            } error: Received unexpected caveats:\n${JSON.stringify(
+              permissionOptions.caveats,
```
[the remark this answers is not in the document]
rekmarks: Yeah good point. Fixed here: 48a3e60bb

app/scripts/controllers/permissions/specifications.js
```diff
@@ -0,0 +1,261 @@
...
+};
+
+/**
+ * Gets the specifications for all permissions that will be recognized by the
+ * PermissionController.
+ *
+ * @param {{
+ *   getKeyringAccounts: () => Promise<string[]>,
+ *   getIdentities: () => Record<string, Identity>,
+ * }} options - Options bag.
```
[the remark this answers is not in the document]
rekmarks: Fixed in: 7607fda82

app/scripts/controllers/permissions/specifications.js
```diff
@@ -0,0 +1,261 @@
...
+ */
+export const getPermissionSpecifications = ({
+  getKeyringAccounts,
```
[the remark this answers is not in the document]
rekmarks: Yeah, I think this name was just a thoughtless carryover. Renamed to `getAllAccounts` here: 7607fda82

app/scripts/controllers/permissions/specifications.js
rekmarks: Fixed here: d939d9c05

app/scripts/metamask-controller.js
rekmarks: Fixed here: 83502cbf8

app/scripts/controllers/permissions/specifications.test.js
```diff
@@ -0,0 +1,340 @@
...
+
+          expect(await methodImplementation()).toStrictEqual([
+            '0x3',
+            '0x4',
+            '0x1',
+            '0x2',
+          ]);
+        });
+
+        it('throws if a keyring account is missing an address (case 1)', async () => {
```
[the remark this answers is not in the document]
rekmarks: Hm, which cases in particular are you referring to? As 83502cbf8, there are two accounts with the same `lastSelected` value in this and the previous test case.

app/scripts/metamask-controller.js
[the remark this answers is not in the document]
rekmarks: I added comments that should clarify the situation here: ef8895171

package.json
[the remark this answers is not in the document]
Gudahtt: Cool, that makes sense to me

app/scripts/migrations/068.js
Gudahtt: An optional name would be better. It should only be shown alongside the origin, so repeating the origin twice would look silly. Better to omit it in that case.

app/scripts/controllers/permissions/specifications.js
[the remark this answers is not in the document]
Gudahtt: Ah, I didn't notice we were using this name before. This new one is better

app/scripts/metamask-controller.js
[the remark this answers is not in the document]
Gudahtt: That does clarify everything but that last question; why doesn't `getPermittedAccountsByOrigin` sort the accounts by last selected?

app/scripts/lib/rpc-method-middleware/handlers/request-accounts.js
```diff
@@ -0,0 +1,103 @@
...
+ * the extension unlocks.
+ * @property {Function} hasPermission - Returns whether the requesting origin
+ * has the specified permission.
+ * @property {Function} requestAccountsPermission - Requests the `eth_accounts`
+ * permission for the requesting origin.
+ */
+
+/**
+ *
+ * @param {import('json-rpc-engine').JsonRpcRequest<unknown>} req - The JSON-RPC request object.
```
Gudahtt, line 43: The parameter has a leading underscore
```suggestion
 * @param {import('json-rpc-engine').JsonRpcRequest<unknown>} _req - The JSON-RPC request object.
```

app/scripts/lib/rpc-method-middleware/handlers/request-accounts.js
```diff
@@ -0,0 +1,103 @@
...
+  // If no accounts, request the accounts permission
+  try {
+    await requestAccountsPermission();
+  } catch (err) {
+    res.error = err;
+    return end();
+  }
+
+  // Get the approved accounts
+  const accounts = await getAccounts();
```
Gudahtt, line 90: We could eliminate this step and use the accounts returned by `requestAccountsPermission` instead. That function returns the permissions now. That eliminates an async call, it eliminates a race condition (where the accounts are granted, then revoked before this function returns), and it eliminates the `istanbul ignore else` comment for that difficult-to-reproduce race condition.

app/scripts/lib/rpc-method-middleware/handlers/request-accounts.js
```diff
@@ -0,0 +1,103 @@
...
+
+  if (hasPermission(MESSAGE_TYPE.ETH_ACCOUNTS)) {
+    // We wait for the extension to unlock in this case only, because permission
+    // requests are handled when the extension is unlocked, regardless of the
+    // lock state when they were received.
+    locks.add(origin);
+    await getUnlockPromise();
+    locks.delete(origin);
+
+    res.result = await getAccounts();
```
Gudahtt, line 77: This is pre-existing, but... should we release the lock before getting the accounts? Technically we could see requests stack up as things currently are.

Not a blocking issue either way, since this is how it acted beforehand.

app/scripts/controllers/permissions/specifications.test.js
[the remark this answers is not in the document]
Gudahtt: I must have missed it I guess! I see it now.

app/scripts/lib/rpc-method-middleware/createMethodMiddleware.js
Gudahtt: Yikes! I wonder how many Sentry errors were caused by this.

app/scripts/metamask-controller.js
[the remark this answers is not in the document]
rekmarks: They're not sorted because that function is just pulling the accounts directly from the `PermissionController` state, where they are unsorted. Are you asking me to add the sorting logic from the `eth_accounts` implementation to `getPermittedAccountsByOrigin`?

app/scripts/controllers/permissions/selectors.test.js
```diff
@@ -0,0 +1,97 @@
...
+      const expected = new Map([
+        ['foo.bar', ['0x1']],
+        ['bar.baz', ['0x2']],
+        ['baz.fizz', ['0x1', '0x2']],
+      ]);
+
+      const received = getPermittedAccountsByOrigin(state);
+
+      expect(received).toStrictEqual(expected);
+      // The selector should return the memoized value if state.subjects is
```
Gudahtt, line 92: It might be worth checking that the memoization works by changing the state, and seeing that the answer changes.

app/scripts/controllers/permissions/permission-log.test.js
```diff
@@ -667,14 +578,90 @@ describe('permissions log', function () {
...
+              accounts: {
+                '0x1': 1,
+              },
+              lastApproved: 1,
+            },
+          },
+        },
+      });
+
+      clock.tick(1);
```
Gudahtt, line 615: What is this for? I didn't see any timer related code that would need this.

app/scripts/metamask-controller.js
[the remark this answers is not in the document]
Gudahtt: I was just wondering if it was considered. I see now that `getPermittedAccountsByOrigin` is a selector, so it wouldn't make a lot of sense for that to be async. This approach seems fine.

Gudahtt:
Just finished my first pass!

app/scripts/migrations/068.js
rekmarks: We are strongly against silliness! cd9cbe4e6

app/scripts/controllers/permissions/permission-log.test.js
rekmarks: It determines the number value of the account properties in the expected state object below.

Gudahtt: Ahh I see

app/scripts/metamask-controller.js
```diff
@@ -1208,29 +1311,19 @@ export default class MetamaskController extends EventEmitter {
       initializeThreeBox: nodeify(this.initializeThreeBox, this),
 
       // permissions
+      getApprovedAccounts: nodeify(this.getPermittedAccounts, this),
```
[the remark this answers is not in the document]
rekmarks: Fixed in: d49f868ae

app/scripts/metamask-controller.js
```diff
@@ -2689,40 +2818,51 @@ export default class MetamaskController extends EventEmitter {
...
-          this.networkController,
+        requestAccountsPermission: this.permissionController.requestPermissions.bind(
+          this.permissionController,
+          { origin },
+          { eth_accounts: {} },
         ),
+        requestPermissionsForOrigin: this.permissionController.requestPermissions.bind(
+          this.permissionController,
+          { origin },
+        ),
```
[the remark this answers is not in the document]
rekmarks: Fixed in: d49f868ae

app/scripts/lib/rpc-method-middleware/handlers/eth-accounts.js
```diff
@@ -0,0 +1,33 @@
+import { MESSAGE_TYPE } from '../../../../../shared/constants/app';
+
+/**
+ * A wrapper for eth_accounts that never errors.
```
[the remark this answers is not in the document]
rekmarks: Fixed in: d49f868ae

app/scripts/lib/rpc-method-middleware/handlers/request-accounts.js
rekmarks: Fixed in: d49f868ae

app/scripts/controllers/permissions/background-api.test.js
```diff
@@ -0,0 +1,187 @@
...
+      expect(permissionController.updateCaveat).not.toHaveBeenCalled();
+    });
+  });
+
+  describe('requestAccountsPermissionWithId', () => {
+    it('request an accounts permission and returns the request id', async () => {
+      const permissionController = {
+        requestPermissions: jest.fn().mockImplementationOnce(() => {
+          return Promise.resolve([null, { id: 'arbitraryId' }]);
+        }),
```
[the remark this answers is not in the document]
rekmarks: Fixed in: d49f868ae

app/scripts/controllers/permissions/selectors.js
```diff
@@ -0,0 +1,81 @@
...
+  if (newAccountsMap === previousAccountsMap) {
+    return changedAccounts;
+  }
+
+  const newOrigins = new Set([...newAccountsMap.keys()]);
+
+  for (const origin of previousAccountsMap.keys()) {
+    const newAccounts = newAccountsMap.get(origin) ?? [];
+
+    // The values of these maps are references to immutable PermissionController
```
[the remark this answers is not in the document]
rekmarks: Yeah, I think this is sound. Expanded the comment here: 26d3949e2

app/scripts/controllers/permissions/selectors.test.js
rekmarks: Agreed! 26d3949e2

app/scripts/lib/rpc-method-middleware/handlers/send-metadata.js
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * @param {import('json-rpc-engine').JsonRpcRequest<unknown>} req - The JSON-RPC request object.
+ * @param {import('json-rpc-engine').JsonRpcResponse<true>} res - The JSON-RPC response object.
+ * @param {Function} _next - The json-rpc-engine 'next' callback.
+ * @param {Function} end - The json-rpc-engine 'end' callback.
+ * @param {SendMetadataOptions} options
+ */
+function sendMetadataHandler(req, res, _next, end, { addSubjectMetadata }) {
+  if (typeof req.params?.name === 'string') {
+    addSubjectMetadata(req.params);
```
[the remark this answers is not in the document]
rekmarks: Great catch! ad5ff590d

app/scripts/lib/rpc-method-middleware/handlers/send-metadata.js
```diff
@@ -0,0 +1,42 @@
...
+ * @param {Function} _next - The json-rpc-engine 'next' callback.
+ * @param {Function} end - The json-rpc-engine 'end' callback.
+ * @param {SendMetadataOptions} options
+ */
+function sendMetadataHandler(req, res, _next, end, { addSubjectMetadata }) {
+  const { params } = req;
+  if (params && typeof params === 'object' && !Array.isArray(params)) {
+    const { icon, ...remainingParams } = params;
+
+    if (typeof remainingParams?.name === 'string') {
```
Gudahtt, line 35: This seems like an odd choice :thinking:

app/scripts/lib/rpc-method-middleware/handlers/send-metadata.js
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * @param {import('json-rpc-engine').JsonRpcRequest<unknown>} req - The JSON-RPC request object.
+ * @param {import('json-rpc-engine').JsonRpcResponse<true>} res - The JSON-RPC response object.
+ * @param {Function} _next - The json-rpc-engine 'next' callback.
+ * @param {Function} end - The json-rpc-engine 'end' callback.
+ * @param {SendMetadataOptions} options
+ */
+function sendMetadataHandler(req, res, _next, end, { addSubjectMetadata }) {
+  const { params } = req;
+  if (params && typeof params === 'object' && !Array.isArray(params)) {
```
Gudahtt, line 32: Should we throw if `params` is invalid?

app/scripts/lib/rpc-method-middleware/handlers/request-accounts.js
rekmarks: Fixed here: a2d793b99

app/scripts/lib/rpc-method-middleware/handlers/send-metadata.js
rekmarks: I agree completely, but that's what the existing implementation is doing: https://github.com/MetaMask/metamask-extension/blob/76a2a9b/app/scripts/controllers/permissions/permissionsMethodMiddleware.js#L74-L82

I suppose the intention is to say that "if there's no name, then this is probably not valid metadata", but there could still be an icon... Maybe we should update the validation, return an error if it fails, and just forward metadata to the controller, name or no?

app/scripts/lib/rpc-method-middleware/handlers/request-accounts.js
rekmarks: We could do this, but I'd have to pull out the sorting logic from the `eth_accounts` method implementation. Since this is pre-existing, can we do it in a follow-up?

app/scripts/lib/rpc-method-middleware/handlers/send-metadata.js
Gudahtt: Interesting!

>Maybe we should update the validation, return an error if it fails, and just forward metadata to the controller, name or no?

That sounds better to me

app/scripts/lib/rpc-method-middleware/handlers/request-accounts.js
Gudahtt: Oh right... I forgot about the sorting. Nevermind then! I thought this was an easy win. Improving this later sounds good to me.

app/scripts/background.js
Gudahtt: Sounds good!

app/scripts/lib/rpc-method-middleware/handlers/send-metadata.js
rekmarks: Done here: 6804a8fd2

app/scripts/lib/rpc-method-middleware/handlers/send-metadata.js
rekmarks: See: https://github.com/MetaMask/metamask-extension/pull/12243#discussion_r763387274

ui/components/app/connected-sites-list/connected-sites-list.component.js
rekmarks: Took a stab at it here: 73cdc0be8

ui/components/app/signature-request-original/signature-request-original.component.js
```diff
@@ -121,22 +121,22 @@ export default class SignatureRequestOriginal extends Component {
...
         <div className="request-signature__origin-label">
           {`${t('origin')}:`}
         </div>
-        {originMetadata?.icon ? (
+        {targetSubjectMetadata?.iconUrl ? (
           <SiteIcon
-            icon={originMetadata.icon}
-            name={originMetadata.hostname}
+            icon={targetSubjectMetadata.iconUrl}
+            name={targetSubjectMetadata.name}
```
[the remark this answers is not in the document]
rekmarks: Changed to use the hostname from the `origin` here: ba7bec986

ui/components/app/alerts/unconnected-account-alert/unconnected-account-alert.js
rekmarks: Ah, you are correct, I just realized that we could avoid finding the selected identity in this component since the `getAccountToConnectToActiveTab` selector returns `identities[selectedAddress]` ([ref](https://github.com/MetaMask/metamask-extension/blob/develop/ui/selectors/permissions.js/#L179-L201)). If `accountToConnect` is `undefined`, that's handled by the `ConnectedAccountsList` component [here](https://github.com/MetaMask/metamask-extension/blob/develop/ui/components/app/connected-accounts-list/connected-accounts-list.component.js/#L74-L76).

ui/components/app/connected-sites-list/connected-sites-list.component.js
Gudahtt: Looks good!

rekmarks: Updated: a736af520

ui/components/app/alerts/unconnected-account-alert/unconnected-account-alert.js
Gudahtt: Can we revert this change then? Or is there some reason to keep it? I don't really follow why we'd want to avoid using the `getSelectedIdentity` selector here.

Gudahtt (commented):
Looks good! I can approve after this is rebased.

ui/components/app/alerts/unconnected-account-alert/unconnected-account-alert.js
rekmarks: This was resolved. We reverted the change.

rekmarks:
For any new reviewers: @Gudahtt conducted a thorough review and all of his feedback has been addressed as of 0c8eadf, which I squashed in order to make rebasing onto latest `develop` easier.

darkwing (approved):

Gudahtt (approved):
LGTM!

PR #12246, opened 2021-09-29
Migrate beta version to the main version field

Gudahtt:
The main `version` field in `package.json` will now include the beta version (if present) rather than it being passed in via the CLI when building. The `version` field is now a fully SemVer-compatible version, with the added restriction that any prerelease portion of the version must match the format `<build type>.<build version>`.

This brings the build in-line with the future release process we will be using for the beta version. The plan is for each future release to enter a "beta phase" where the version would get updated to reflect that it's a beta, and we would increment this beta version over time as we update the beta. The manifest gives us a place to store this beta version. It was also important to replace the automatic minor bump logic that was being used previously, because the version in beta might not be a minor bump.

Additionally, the filename logic used for beta builds was updated to be generic across all build types rather than beta-specific. This will be useful for Flask builds in the future.

Manual testing steps:  
  - Build "main" and "beta" builds (`yarn dist` and `yarn dist --build-type beta`) and ensure both versions work, and that the correct assets are used throughout the app.
  - Try adding a `-beta.<number>` suffix to the `version` field in `package.json` and ensure both beta and non-beta builds work.
  - Try adding invalid suffixes to `version` in `package.json` and ensure the build fails (e.g. anything other than `-beta.<positive integer>`.

### DIFF
--- a/development/build/utils.js
+++ b/development/build/utils.js
@@ -1,24 +1,5 @@
-/**
- * @returns {Object} An object with browser as key and next version of beta
- * as the value.  E.g. { firefox: '9.6.0.beta0', chrome: '9.6.0.1' }
- */
-function getNextBetaVersionMap(currentVersion, betaVersion, platforms) {
-  const [major, minor] = currentVersion.split('.');
-
-  return platforms.reduce((platformMap, platform) => {
-    platformMap[platform] = [
-      // Keeps the current major
-      major,
-      // Bump the minor version
-      Number(minor) + 1,
-      // This isn't typically used
-      0,
-      // The beta number
-      `${platform === 'firefox' ? 'beta' : ''}${betaVersion}`,
-    ].join('.');
-    return platformMap;
-  }, {});
-}
+const semver = require('semver');
+const { version } = require('../../package.json');
 
 const BuildTypes = {
   beta: 'beta',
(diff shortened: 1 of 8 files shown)


development/build/utils.js
```diff
@@ -1,32 +1,64 @@
...
+  const prerelease = semver.prerelease(version);
+
+  let buildType;
+  let buildVersion;
+  if (prerelease) {
+    if (prerelease.length !== 2) {
+      throw new Error(`Invalid prerelease version: '${prerelease.join('.')}'`);
+    }
+    buildType = prerelease[0];
+    buildVersion = prerelease[1];
```
darkwing, line 36: Can this be `[buildType, buildVersion] = prerelease`

darkwing (dismissed):

development/build/utils.js
```diff
@@ -1,32 +1,64 @@
...
+ * as the value.  For example, the version `9.6.0-beta.1` would return the object
+ * `{ firefox: { version: '9.6.0.beta1' }, chrome: { version: '9.6.0.1', version_name: 'beta' } }`.
  */
-function getNextBetaVersionMap(currentVersion, betaVersion, platforms) {
-  const [major, minor] = currentVersion.split('.');
+function getBrowserVersionMap(platforms) {
+  const major = semver.major(version);
+  const minor = semver.minor(version);
+  const patch = semver.patch(version);
+  const prerelease = semver.prerelease(version);
```
rekmarks, line 27: Why not use `semver.parse(version)` here?

Gudahtt: I wasn't sure what that would do for us :thinking: I see that it returns a SemVer object, but that class didn't seem to have any documented `major`, `minor`, `patch`, or `prerelease` methods or properties.

Since you mentioned it, I did dig in and I found that it has these properties. But I'm not sure whether they're public API or not.

brad-decker (dismissed):
LGTM

development/build/utils.js
rekmarks:
```
const semverParse = require('semver/functions/parse')
```

That's listed as a usage example in [their readme](https://github.com/npm/node-semver#usage), so I can't imagine it isn't public. However, I'm fine with it either way!

rekmarks (dismissed):

Gudahtt:
Just rebased this on develop to get the CLA bot to run (it stalled out during the GitHub Actions outage)

rekmarks (approved):

darkwing (approved):

PR #9825, opened 2020-11-07
Revert "Remove unnecessary lock page (#9793)"

Gudahtt:
This reverts commit f5265c24ab2359f7d51b5f527dde23a7132c4bd1.

Apparently it wasn't unnecessary after all. The Lock page served a few different purposes. First, it was used to safeguard the seed phrase, in case the user was interrupted after setting a password. Otherwise anyone could open MetaMask and see the seed phrase without verifying the password. Second, the submit function for the initialization unlock screen also returned the seed phrase, so that it could be set in React state for the confirmation step. Third, the submit function was also responsible for navigating back to the seed phrase reveal page.

Removing the lock page had the effect of causing an infinite render loop if onboarding was interrupted in the "Create" flow after setting a password but before seed phrase confirmation. That redirect loop has now been fixed.

Manual testing steps:  
  - Install the extension
  - Proceed through the 'Create' flow
  - Submit a password
  - Exit the tab/window
  - Open a new tab/window and click the MetaMask browser action button

Before this PR, this would result in an infinite loop. After this PR, it should navigate to the unlock screen, after which onboarding picks up where it left off at the seed phrase reveal page.

PR #12247, opened 2021-09-29
Establish onboarding-flow wrapper/router base and feature flag env variable

### LINKED ISSUE #12198: Routing Wrapper Component
Routing Wrapper component for new onboarding flow

adonesky1:
Fixes: #12198

Explanation:  
 - Adds new routing/wrapper components for the new onboarding flow as well as feature flags to hide new onboarding designs behind env variable.

### DIFF
--- a/development/build/scripts.js
+++ b/development/build/scripts.js
@@ -613,8 +613,7 @@ function getEnvironmentVariables({ buildType, devMode, testing }) {
         ? process.env.SEGMENT_PROD_LEGACY_WRITE_KEY
         : metamaskrc.SEGMENT_LEGACY_WRITE_KEY,
     SWAPS_USE_DEV_APIS: process.env.SWAPS_USE_DEV_APIS === '1',
-    ONBOARDING_V2:
-      process.env.ONBOARDING_V2 === '1' || metamaskrc.ONBOARDING_V2 === '1',
+    ONBOARDING_V2: metamaskrc.ONBOARDING_V2 === '1',
   };
 }
 
--- a/development/build/scripts.js
+++ b/development/build/scripts.js
@@ -27,6 +27,7 @@ const bifyModuleGroups = require('bify-module-groups');
 
 const metamaskrc = require('rc')('metamask', {
   INFURA_PROJECT_ID: process.env.INFURA_PROJECT_ID,
+  ONBOARDING_V2: process.env.ONBOARDING_V2,
   SEGMENT_HOST: process.env.SEGMENT_HOST,
   SEGMENT_WRITE_KEY: process.env.SEGMENT_WRITE_KEY,
   SEGMENT_LEGACY_WRITE_KEY: process.env.SEGMENT_LEGACY_WRITE_KEY,
@@ -612,6 +613,8 @@ function getEnvironmentVariables({ buildType, devMode, testing }) {
         ? process.env.SEGMENT_PROD_LEGACY_WRITE_KEY
         : metamaskrc.SEGMENT_LEGACY_WRITE_KEY,
     SWAPS_USE_DEV_APIS: process.env.SWAPS_USE_DEV_APIS === '1',
+    ONBOARDING_V2:
+      process.env.ONBOARDING_V2 === '1' || metamaskrc.ONBOARDING_V2 === '1',
   };
 }
 
--- a/ui/helpers/higher-order-components/authenticated/authenticated.component.js
+++ b/ui/helpers/higher-order-components/authenticated/authenticated.component.js
@@ -10,12 +10,23 @@ import {
 export default function Authenticated(props) {
   const { isUnlocked, completedOnboarding } = props;
   switch (true) {
+    // For ONBOARDING_V2 dev purposes,
+    // Remove when ONBOARDING_V2 dev complete
     case process.env.ONBOARDING_V2 === true:
       return <Redirect to={{ pathname: ONBOARDING_ROUTE }} />;
+
     case isUnlocked && completedOnboarding:
       return <Route {...props} />;
     case !completedOnboarding:
-      return <Redirect to={{ pathname: INITIALIZE_ROUTE }} />;
+      return (
+        <Redirect
+          to={{
+            pathname: process.env.ONBOARDING_V2
+              ? ONBOARDING_ROUTE
+              : INITIALIZE_ROUTE,
+          }}
+        />
+      );
     default:
       return <Redirect to={{ pathname: UNLOCK_ROUTE }} />;
   }
--- a/ui/helpers/higher-order-components/authenticated/authenticated.component.js
+++ b/ui/helpers/higher-order-components/authenticated/authenticated.component.js
@@ -1,12 +1,17 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Redirect, Route } from 'react-router-dom';
-import { UNLOCK_ROUTE, INITIALIZE_ROUTE } from '../../constants/routes';
+import {
+  UNLOCK_ROUTE,
+  INITIALIZE_ROUTE,
+  ONBOARDING_ROUTE,
+} from '../../constants/routes';
 
 export default function Authenticated(props) {
   const { isUnlocked, completedOnboarding } = props;
-
   switch (true) {
+    case process.env.ONBOARDING_V2 === true:
+      return <Redirect to={{ pathname: ONBOARDING_ROUTE }} />;
     case isUnlocked && completedOnboarding:
       return <Route {...props} />;
     case !completedOnboarding:
--- a/ui/pages/onboarding-flow/onboarding-flow-switch/onboarding-flow-switch.js
+++ b/ui/pages/onboarding-flow/onboarding-flow-switch/onboarding-flow-switch.js
@@ -3,23 +3,23 @@ import { useSelector } from 'react-redux';
 import { Redirect } from 'react-router-dom';
 import {
   DEFAULT_ROUTE,
-  LOCK_ROUTE,
   ONBOARDING_COMPLETION_ROUTE,
   ONBOARDING_GET_STARTED_ROUTE,
   ONBOARDING_UNLOCK_ROUTE,
+  LOCK_ROUTE,
 } from '../../../helpers/constants/routes';
 import {
   getCompletedOnboarding,
   getIsInitialized,
   getIsUnlocked,
   getSeedPhraseBackedUp,
-} from '../../../selectors';
+} from '../../../ducks/metamask/metamask';
 
 export default function OnboardingFlowSwitch() {
   const completedOnboarding = useSelector(getCompletedOnboarding);
   const isInitialized = useSelector(getIsInitialized);
-  const isUnlocked = useSelector(getIsUnlocked);
   const seedPhraseBackedUp = useSelector(getSeedPhraseBackedUp);
+  const isUnlocked = useSelector(getIsUnlocked);
 
   if (completedOnboarding) {
     return <Redirect to={{ pathname: DEFAULT_ROUTE }} />;
--- a/ui/pages/onboarding-flow/onboarding-flow-switch/onboarding-flow-switch.js
+++ b/ui/pages/onboarding-flow/onboarding-flow-switch/onboarding-flow-switch.js
@@ -1,0 +1,41 @@
+import React from 'react';
+import { useSelector } from 'react-redux';
+import { Redirect } from 'react-router-dom';
+import {
+  DEFAULT_ROUTE,
+  LOCK_ROUTE,
+  ONBOARDING_COMPLETION_ROUTE,
+  ONBOARDING_GET_STARTED_ROUTE,
+  ONBOARDING_UNLOCK_ROUTE,
+} from '../../../helpers/constants/routes';
+import {
+  getCompletedOnboarding,
+  getIsInitialized,
+  getIsUnlocked,
+  getSeedPhraseBackedUp,
+} from '../../../selectors';
+
+export default function OnboardingFlowSwitch() {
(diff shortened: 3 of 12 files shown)


darkwing (dismissed):
This is awesome!

darkwing (dismissed):

development/build/scripts.js
```diff
@@ -612,6 +613,8 @@ function getEnvironmentVariables({ buildType, devMode, testing }) {
         ? process.env.SEGMENT_PROD_LEGACY_WRITE_KEY
         : metamaskrc.SEGMENT_LEGACY_WRITE_KEY,
     SWAPS_USE_DEV_APIS: process.env.SWAPS_USE_DEV_APIS === '1',
+    ONBOARDING_V2:
+      process.env.ONBOARDING_V2 === '1' || metamaskrc.ONBOARDING_V2 === '1',
```
Gudahtt, line 617: This should be equivalent:
```suggestion
    ONBOARDING_V2: metamaskrc.ONBOARDING_V2 === '1',
```

Since `process.env.ONBOARDING_V2` is set as the default on line 30.

ui/helpers/higher-order-components/authenticated/authenticated.component.js
```diff
@@ -1,12 +1,17 @@
...
+  UNLOCK_ROUTE,
+  INITIALIZE_ROUTE,
+  ONBOARDING_ROUTE,
+} from '../../constants/routes';
 
 export default function Authenticated(props) {
   const { isUnlocked, completedOnboarding } = props;
-
   switch (true) {
+    case process.env.ONBOARDING_V2 === true:
```
Gudahtt, line 13: This seems like it should be below, alongside the `completedOnboarding` case :thinking: 
Why would we redirect users here before they've completed onboarding?

ui/pages/onboarding-flow/onboarding-flow-switch/onboarding-flow-switch.js
```diff
@@ -0,0 +1,41 @@
...
+  if (completedOnboarding) {
+    return <Redirect to={{ pathname: DEFAULT_ROUTE }} />;
+  }
+
+  if (seedPhraseBackedUp !== null) {
+    return <Redirect to={{ pathname: ONBOARDING_COMPLETION_ROUTE }} />;
+  }
+
+  if (isUnlocked) {
+    return <Redirect to={{ pathname: LOCK_ROUTE }} />;
```
Gudahtt, line 33: This seems like a confusing decision :thinking: Why redirect here if the wallet is unlocked?

ui/pages/onboarding-flow/onboarding-flow.js
```diff
@@ -0,0 +1,115 @@
...
+  const dispatch = useDispatch();
+  const history = useHistory();
+  const isInitialized = useSelector(getIsInitialized);
+  const isUnlocked = useSelector(getIsUnlocked);
+  const completedOnboarding = useSelector(getCompletedOnboarding);
+  const seedPhraseBackedUp = useSelector(getSeedPhraseBackedUp);
+  const nextRoute = useSelector(getFirstTimeFlowTypeRoute);
+
+  useEffect(() => {
+    if (process.env.ONBOARDING_V2) {
```
Gudahtt, line 39: At the moment this would always resolve to `true`, unless the user typed in this route manually. Was this component intended to be used with `ONBOARDING_V2` disabled?

ui/selectors/selectors.js
```diff
@@ -171,6 +171,21 @@ export const getMetaMaskAccounts = createSelector(
     ),
 );
 
+export function getCompletedOnboarding(state) {
+  return state.metamask.completedOnboarding;
```
Gudahtt, line 175: In keeping with the "ducks" pattern, we've started placing selectors alongside the reducers they correspond to. So these would belong at the end of `ui/ducks/metamask/metamask.js`.

ui/pages/onboarding-flow/onboarding-flow.js
```diff
@@ -0,0 +1,115 @@
...
+
+  const handleUnlock = async (password) => {
+    try {
+      const retreivedSeedPhrase = await dispatch(
+        unlockAndGetSeedPhrase(password),
+      );
+      setSeedPhrase(retreivedSeedPhrase);
+      history.push(nextRoute);
+    } catch (error) {
+      throw new Error(error.message);
```
Gudahtt, line 79: What is the purpose of this catch block, and the one above on line 67? These would have the effect of stripping the error of any non-message properties, like `code` and `stack`.

ui/helpers/higher-order-components/authenticated/authenticated.component.js
adonesky1: @Gudahtt honestly this was added in for development convenience purposes. But I can make the change locally while in dev without committing it!

ui/pages/onboarding-flow/onboarding-flow.js
adonesky1: Also in here for dev convenience purposes but will remove for the actual commit!

ui/helpers/higher-order-components/authenticated/authenticated.component.js
Gudahtt: Ah OK. I'm not opposed to adding temporary stuff like this since it's behind the flag, but a comment to help identify these cases would be great.

ui/pages/onboarding-flow/onboarding-flow.js
adonesky1: Honestly this is a copy from current [first-time-flow.component.js file](https://github.com/MetaMask/metamask-extension/blob/6187ab9b01ea757366a42d22b7292413c0f2b119/ui/pages/first-time-flow/first-time-flow.component.js#L97). I created a new file to convert to a functional component and keep the use of env variables down. But you're totally right this is silly. I will remove the try/catch here.

ui/pages/onboarding-flow/onboarding-flow-switch/onboarding-flow-switch.js
adonesky1: Same thing here. A copy from the original file [here](https://github.com/MetaMask/metamask-extension/blob/6187ab9b01ea757366a42d22b7292413c0f2b119/ui/pages/first-time-flow/first-time-flow-switch/first-time-flow-switch.component.js#L37). Good question 🤔

ui/helpers/higher-order-components/authenticated/authenticated.component.js
adonesky1: Gotcha! Yeah I really ought to have called that out! Will add a comment!

ui/pages/onboarding-flow/onboarding-flow.js
adonesky1: Added a comment calling out that this is strictly for dev purposes and to be removed when ONBOARDING_V2 dev is complete

ui/pages/onboarding-flow/onboarding-flow-switch/onboarding-flow-switch.js
adonesky1: I'm feeling like this if block should be excluded? The default on line 40 is the `onboarding_unlock_route`. Not sure why we would use the normal lock route in the onboarding flow at all since we have a distinct onboarding unlock component. And as you called out the logic seems wrong. Probably supposed to be `if(!isUnlocked)`. But weird that it seems to have been like this in prod for a while... @Gudahtt

Gudahtt: I swear I've gone down this rabbit hole before, tried to remove this component completely, then found a case where it was required :thinking: I can't recall why though...

Gudahtt: Oh, more details here: https://github.com/MetaMask/metamask-extension/pull/9825

Gudahtt: Essentially the lock screen ensures that the onboarding flow works if it gets interrupted. It prevents the seed phrase from being seen without entering the password (if the "create" flow is interrupted), and forcing the unlock has the side-effect of allowing us to bring the seed phrase into the UI for the verification page.

At least for the Create flow it's definitely required! We don't need the _page_, but we need to force the unlock somehow in this circumstance. I'm not totally sure if we need it for the import flow or not :thinking:  But it doesn't seem to hurt at least.

Gudahtt: Oh, though we'd also need to ensure the Lock screen redirects to the ONBOARDING_V2 unlock page in this case, not the old one.

adonesky1: Don't we want a lock screen apart from the onboarding flow? For instance if a user clicks the lock button
Screen Shot 2021-09-30 at 1 00 16 PM
or if they use the auto-lock timer?

Gudahtt: Not exactly - we don't show a lock screen when that button is used, or when the auto-lock timer is called. We just lock the wallet, and that redirects them to the unlock page.

This "Lock page" isn't really a page, it's a component with an on-mount hook to lock and redirect.

adonesky1: And I was wrong there is no new ONBOARDING_V2 unlock page so we should continue to use the existing one. 

I'm still not understanding why [this](https://github.com/MetaMask/metamask-extension/blob/6187ab9b01ea757366a42d22b7292413c0f2b119/ui/pages/first-time-flow/first-time-flow-switch/first-time-flow-switch.component.js#L36) is `if(isUnlock)` and not the inverse?

adonesky1: @Gudahtt ^

adonesky1: Ok just writing out the onboarding interruption flow so I can think it through.  So the Lock component associated with the `LOCK_ROUTE` actually just sets isUnlocked to false and then history.pushes to DEFAULT_ROUTE. So this component (onboarding-flow-switch, formerly first-time-flow-switch) catches any `onboarding/` routes we don't catch in the routes established in onboarding-flow.js and routes accordingly. If we hit this if block (`if(isUnlocked)`) it means we `completedOnboarding` is false and `seedPhraseBackup` is null and so at this point we want to lock the extension which will push us back to DEFAULT_ROUTE, which because completedOnboarding is false will route us back to ONBOARDING_ROUTE which will then push us to the UNLOCK_ROUTE.

Gudahtt: Basically yes. But the reason we want to lock the wallet in that case is because we don't want the seed phrase to be shown on the screen without the user verifying that they know the password. If a user got partway through the onboarding flow, closed the page and left the computer, and then a second person came by and opened MetaMask, they'd be shown the seed phrase.

Showing the seed phrase (and similar cases like showing private keys) is the one use case where we try to be careful about shoulder surfers, because of the severe consequences. Hence the "confirm password" step for showing the seed phrase or exporting a private key.

The second reason we wanted to lock the account with the _old_ flow is that the "initializing unlock" screen was a special unlock screen that had a side-effect when it was unlocked: it retrieved the seed phrase itself from the background, and stored it in memory in UI for the seed phrase reveal page and the backup exercise. This is why removing the Lock component caused an infinite loop when the onboarding was interrupted.

I'm not sure how this will work with your new onboarding flow - maybe it makes sense to retrieve it directly when it's needed, rather than store it in state as a side-effect of unlocking on a special onboarding unlock page. So this second reason might not apply anymore, in which case using the normal unlock page might be perfectly fine.

Gudahtt (approved):
LGTM!

ui/pages/onboarding-flow/onboarding-flow-switch/onboarding-flow-switch.js
adonesky1: Just to clarify in my head, it seems like it's the same `UnlockPage` across the board, but in the initialize/onboarding flow we pass an onSubmit method in that retrieves the seedPhrase and then stores it in state. This still could make sense as the flow. The alternative I'm hearing you describe is to retrieve the seedphrase directly from background on the views where it is needed rather than keeping it in state in the onboarding-flow component and passing it around. The only downside to this second approach, I think, is more calls to the background? Does that sound right to you @Gudahtt?

Gudahtt: Yes that sounds correct to me. I agree that either seems fine.

> The only downside to this second approach, I think, is more calls to the background

Possibly, though depending on how we do it, we could still guarantee that only one call to the background is made.

Another downside is that having it as a side-effect of unlock _guarantees_ that the user has just entered their password. Whereas if we rely on routing for that, maybe that's easier to mess up.

adonesky1: Ok thanks so much as always for your in-depth input @Gudahtt. With regard to this particular issue, I'm thinking if it ain't broke. I believe I've added back what's necessary to preserve this behavior in the new onboarding flow, but perhaps I'm still missing something...?

adonesky1: Sorry just noticed that you approved!

darkwing (approved):

PR #12249, opened 2021-09-29
Suppress prop types warning in RadioGroupComponent

mcmire:
When editing the gas fee for a transaction, the following warning is
being output to the console:

    Warning: Failed prop type: Connector: prop type `isFirst` is invalid; it must be a function, usually from the `prop-types` package, but received `undefined`.

This commit fixes this issue.


darkwing (approved):

adonesky1 (approved):
👍

PR #12251, opened 2021-09-29
Implement event fragments

brad-decker:
This is a draft implementation of the spec detailed at a high level in this [proposal](https://docs.google.com/document/d/1_is-9L9tauqxtwB9gsEUQuCW4mMnDnqsoi3Tudb29Z0/edit?usp=sharing). This implementation is meant to be accompanied with the proposal and serve as a technical reference. It has not yet been fully fleshed out or tested but the theory is solid.

#### Checklist
- [x] Test coverage for metametrics controller
- [x] Clean up code and add more documentation
- [x] Lint, Tests, etc.

### DIFF
--- a/app/scripts/controllers/metametrics.js
+++ b/app/scripts/controllers/metametrics.js
@@ -1,11 +1,13 @@
-import { merge, omit } from 'lodash';
+import { merge, omit, omitBy } from 'lodash';
 import { ObservableStore } from '@metamask/obs-store';
 import { bufferToHex, keccak } from 'ethereumjs-util';
+import { generateUUID } from 'pubnub';
 import { ENVIRONMENT_TYPE_BACKGROUND } from '../../../shared/constants/app';
 import {
   METAMETRICS_ANONYMOUS_ID,
   METAMETRICS_BACKGROUND_PAGE_OBJECT,
 } from '../../../shared/constants/metametrics';
+import { SECOND } from '../../../shared/constants/time';
 
 const defaultCaptureException = (err) => {
   // throw error on clean stack so its captured by platform integrations (eg sentry)
@@ -27,15 +29,18 @@ const exceptionsToFilter = {
  * @typedef {import('../../../shared/constants/metametrics').SegmentInterface} SegmentInterface
  * @typedef {import('../../../shared/constants/metametrics').MetaMetricsPagePayload} MetaMetricsPagePayload
  * @typedef {import('../../../shared/constants/metametrics').MetaMetricsPageOptions} MetaMetricsPageOptions
+ * @typedef {import('../../../shared/constants/metametrics').MetaMetricsEventFragment} MetaMetricsEventFragment
  */
 
 /**
  * @typedef {Object} MetaMetricsControllerState
- * @property {?string} metaMetricsId - The user's metaMetricsId that will be
+ * @property {string} [metaMetricsId] - The user's metaMetricsId that will be
  *  attached to all non-anonymized event payloads
- * @property {?boolean} participateInMetaMetrics - The user's preference for
+ * @property {boolean} [participateInMetaMetrics] - The user's preference for
  *  participating in the MetaMetrics analytics program. This setting controls
  *  whether or not events are tracked
+ * @property {{[string]: MetaMetricsEventFragment}} [fragments] - Object keyed
+ *  by UUID with stored fragments as values.
  */
 
 export default class MetaMetricsController {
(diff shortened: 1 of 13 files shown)


Gudahtt (commented):
Great work! This looks like a simple and effective strategy.

One question I did have is whether there's some API-level support for funnels already in Segment or Mixpanel. e.g. is it possible for us to instruct it in how to string a series of events together? It seems that this PR is taking a more DIY approach of tracking each step of the funnel internally and submitting the entire thing as a single event. Which might be the best approach, but if the platforms support funnels, maybe we'd be missing out on Mixpanel/Segment funnel-related features, like visualizations of each step of the funnel?

It seems that Mixpanel has some support for funnels but I'm not yet sure how they work, or how this approach ties in.

Gudahtt:
"multi-step events" and "event fragments" both sound good to me!

app/scripts/controllers/metametrics.js
```diff
@@ -87,6 +118,95 @@ export default class MetaMetricsController {
...
+  }
+
+  /**
+   * Updates a funnel in state
+   * @param {string} id - The funnel id to update
+   * @param {MetaMetricsFunnel} options - Funnel settings and properties
+   *  to initiate the funnel with.
+   * @returns {MetaMetricsFunnel}
+   */
+  updateFunnel(id, payload) {
```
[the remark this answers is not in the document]
brad-decker: I do like the idea of having each fragment be a state machine, but I think that complexity is best tackled in a future PR that starts to make use of this underlying implementation. I agree right now the updateFragment method is a little too open-ended but I think hardening it will be best suited for when I start implementing events using it.

app/scripts/controllers/metametrics.js
```diff
@@ -94,6 +104,28 @@ export default class MetaMetricsController {
...
+    setInterval(() => {
+      this.store.getState().fragments.forEach((fragment) => {
+        if (
+          fragment.timeout &&
+          Date.now() - fragment.lastUpdated / 1000 > fragment.timeout
+        ) {
+          this.finalizeEventFragment(fragment.id, { abandoned: true });
+        }
+      });
+    }, SECOND * 30);
```
jpuri, line 128: Not sure why above code requires `setInterval` ?

app/scripts/controllers/metametrics.js
```diff
@@ -107,6 +139,108 @@ export default class MetaMetricsController {
     );
   }
```
jpuri, line 141: code below throws error in various scenario, I am wondering if that can cause issues for user

ui/contexts/metametrics.js
```diff
@@ -125,10 +124,6 @@ export function MetaMetricsProvider({ children }) {
 
 MetaMetricsProvider.propTypes = { children: PropTypes.node };
 
-export function useMetaMetricsContext() {
-  return useContext(MetaMetricsContext);
-}
-
```
jpuri, line 131: `useMetaMetricsContext` might be useful for hooks

jpuri: I would prefer this over `useContext(MetaMetricsContext)`

jpuri:
LGTM, just some small questions / feedbacks added.

ui/contexts/metametrics.js
brad-decker: The problem is that we don't want the context to be consumed directly because the hooks for tracking events do additional work. I removed so as to make it more difficult to consume the context directly.

app/scripts/controllers/metametrics.js
brad-decker: @Gudahtt @kumavis what is the appropriate way of throwing errors for developers implementing events? @jpuri raises a good point and I'm just curious if anything has changed in my absence for best practices around throwing useful errors in background code that **should** be non user facing.

darkwing (approved):
Code looks good and I like the strategy!

app/scripts/controllers/metametrics.js
brad-decker: Oops. sorry forgot to mention i added more docs.

ui/contexts/metametrics.js
jpuri: I think we should either support it get rid of it altogether. Making it more difficult may not be so useful.

jpuri (approved):

app/scripts/controllers/metametrics.js
Gudahtt: This PR should have all of the relevant context: https://github.com/MetaMask/metamask-extension/pull/12573

Basically our strategy is the same as it always was - we should throw anything unexpected and ensure it's logged to Sentry (either explicitly or by not catching it). If it's expected but still shouldn't be user-facing, we can use console logs (and maybe a metric event, if it's something there is value in tracking).

The recent change in this PR was to make the `trackEvent` function non-blocking, for performance reasons. Typically there is no need to wait on submitting a metric event before proceeding to the next operation. The same approach might be useful here too?

There are actually now two functions for submitting events, one non-blocking (`trackEvent`), and one blocking (`submitEvent`), so that there would be an option in the rare case where we did need to block on an event getting submitted. For example, if we start using Mixpanel funnels, we might want to block on the first event getting submitted to ensure that it always gets sent first before the other funnel events (otherwise they might not be recognized by Mixpanel as being part of the same funnel). But as far as I know we don't use `submitEvent` directly anywhere yet.

PR #12254, opened 2021-09-30
Normalize extension verison to SemVer

Gudahtt:
The extension version used throughout the wallet is now normalized to a SemVer-compliant version that matches the version used in `package.json`. We use this version for display on the "About" page, and we attach it to all error reports and metric events, so it's important that we format it consistently so that we can correlate events on the same version across different browsers.

This normalization step is necessary because Firefox and Chrome both have different requirements for the extension version, and neither is SemVer-compliant.

Manual testing steps:  
  - Try adding a `-beta.0` suffix to the version, and see that it is displayed consistently on both Chrome and Firefox
  - Try leaving the version as-is, and ensure it is displayed correctly.

### DIFF
--- a/app/scripts/platforms/extension.js
+++ b/app/scripts/platforms/extension.js
@@ -79,7 +79,36 @@ export default class ExtensionPlatform {
   }
 
   getVersion() {
-    return extension.runtime.getManifest().version;
+    const {
+      version,
+      version_name: versionName,
+    } = extension.runtime.getManifest();
+
+    const versionParts = version.split('.');
+    if (versionName) {
+      // On Chrome, the build type is stored as `version_name` in the manifest, and the fourth part
+      // of the version is the build version.
+      const buildType = versionName;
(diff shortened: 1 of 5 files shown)


Gudahtt:
Depends upon #12246

darkwing (dismissed):

adonesky1 (dismissed):
LGTM!

app/scripts/platforms/extension.js
```diff
@@ -79,7 +79,36 @@ export default class ExtensionPlatform {
   }
 
   getVersion() {
-    return extension.runtime.getManifest().version;
+    const {
+      version,
+      version_name: versionName,
+    } = extension.runtime.getManifest();
+
```
jpuri, line 86: Will be useful to test this conditions.

darkwing (dismissed):

jpuri (approved):

darkwing (approved):

PR #12258, opened 2021-09-30
lockdown - breakout making globalThis properties non-writable

kumavis:
this is in preperation for using LavaMoat which does `lockdown-run.js` but not `lockdown-more.js`


app/scripts/lockdown-more.js
```diff
@@ -0,0 +1,91 @@
...
+     * @param {Object} descriptor - The propertyName descriptor to check.
+     * @returns {boolean} Whether the propertyName descriptor has any accessors.
+     */
+    function hasAccessor(descriptor) {
+      return 'set' in descriptor || 'get' in descriptor;
+    }
+  })();
+} catch (error) {
+  console.error('Protecting intrinsics failed:', error);
+  if (globalThis.sentry && globalThis.sentry.captureException) {
```
darkwing, line 86: Can we do `if (globalThis.sentry?.captureException) {`?

darkwing:
Naming is difficult but could we make `-more` more specific?  I guess I don't have a better word off the top of my head :/

app/scripts/lockdown-more.js
```diff
@@ -0,0 +1,91 @@
+// Make all "object" and "function" own properties of globalThis
+// non-configurable and non-writable, when possible.
+// We call the a property that is non-configurable and non-writable,
```
rekmarks, line 3:
```suggestion
// We call a property that is non-configurable and non-writable,
```

app/scripts/lockdown-more.js
rekmarks:
```suggestion
  if (globalThis?.sentry.captureException) {
```

rekmarks (changes requested):

rekmarks:
@darkwing @kumavis how about `lockdown-intrinsics`?

kumavis:
i definitely struggled to find an appropriate name

kumavis:
we also considered expanding this beyond intrinsics

rekmarks (approved):
LGTM!

ryanml (approved):

Gudahtt (approved):
LGTM!

Issue #11307, opened 2021-06-15
EIP-1559 UI - Implement Busy Network Warning


darkwing:
https://www.figma.com/file/NXvUKcZ0Ocu1GPcT7VxRMx/1559---Gas-Controls?node-id=643%3A5

<img width="729" alt="BusyNetworkWarning" src="https://user-images.githubusercontent.com/46655/122086910-cad4db00-cdc9-11eb-9b2e-91a9759e2094.png">

darkwing:
When real data exists, we'll need to pipe that information into the `EditGasDisplay` component as a `warning`.

darkwing:
@danjm @brad-decker Do we presently have a method for detecting this case?

danjm:
Blocked by `eth_feeHistory` being available from infura

danjm:
There will be a few different pieces of this:

- add a method to the gas fee controller that simply requests `eth_feeHistory` and returns the result
- add a helper function to the gas fee controller helpers/utils file that uses the result of `eth_feeHistory` to calculate whether the network is busy or not
- (possibly... this might need more thought) include `eth_feeHistory` in the gas fee controller state, and have it be requested and updated during the polling in that controller
- dependent on whether do the immediately above bullet, add an indication of whether the network is busy to the gas fee controller state
- once the above changes are made, update the controllers repo
- update the version of the controllers repo in the extension
- use the new data available to show the warning in the above designs

danjm:
> add a helper function to the gas fee controller helpers/utils file that uses the result of eth_feeHistory to calculate whether the network is busy or not

This might require some discussion

danjm:
https://github.com/MetaMask/controllers/blob/main/src/gas/GasFeeController.ts

danjm:
https://gitlab.com/ConsenSys/codefi/products/metaswap/gas-api/-/tree/master

PR #12268, opened 2021-10-01
Alert users when the network is busy 

mcmire:
When a lot of transactions are occurring on the network, such as during
an NFT drop, it drives gas fees up. When this happens, we want to not
only inform the user about this, but also dissuade them from using a
higher gas fee (as we have proved in testing that high gas fees can
cause bidding wars and exacerbate the situation).

The method for determining whether the network is "busy" is already
handled by GasFeeController, which exposes a `networkCongestion`
property within the gas fee estimate data. If this number exceeds 0.66 —
meaning that the current base fee is above the 66th percentile among the
base fees over the last several days — then we determine that the
network is "busy".

## References

**Fixes:** #11307 

## Manual testing steps

There are two versions of this message: one for the "EIP-1559 V1" designs and one for the "EIP-1559 V2" designs. These designs impact the appearance of the confirm transaction screen and modal that appears when you edit the gas fee. You can switch from V1 to V2 locally by going into `.metamaskrc` and adding the following line:

```
EIP_1559_V2=1
```

Then re-running `yarn start` or `yarn start:dev`.

Once you've done this, attempt to send a transaction. Under the V1 design, you should see a message if the network is busy when you go to edit the gas fee. Under the V2 design, you should see a message if the network is busy right away on the confirm transaction screen.

What does it mean for the network to be busy? There isn't really a great way to see this because it depends on the base fee of the latest block (and the base fee of the last 20,000 blocks). That said, I made a [dashboard](https://observablehq.com/@mcmire/realtime-ethereum-gas-fees) that you can use to monitor current gas fees in realtime. It re-uses the same logic that the extension is using to determine network busyness, which is defined as when the current base fee ranks in the top third of all base fees over the last 20,000 blocks. I've provided this threshold on the dashboard. So, to test, you would wait until the current base fee is greater than this threshold. Once this happens and the base fee turns orange, try sending a transaction through the extension and you should see the "network busy" message. (Note: Based on recent base fees, this is likely to happen more often than not.)

## Screencaps

https://user-images.githubusercontent.com/7371/148132616-680cf30f-749d-49d6-9677-046f46cf43e2.mov

https://user-images.githubusercontent.com/7371/148132643-a779f37c-f43c-497c-b39d-492de21a7c1e.mov

### DIFF
--- a/ui/components/app/edit-gas-display/edit-gas-display.component.js
+++ b/ui/components/app/edit-gas-display/edit-gas-display.component.js
@@ -63,7 +63,6 @@ export default function EditGasDisplay({
   estimatedMaximumFiat,
   dappSuggestedGasFeeAcknowledged,
   setDappSuggestedGasFeeAcknowledged,
-  warning,
   gasErrors,
   gasWarnings,
   onManualChange,
@@ -72,6 +71,7 @@ export default function EditGasDisplay({
   estimatesUnavailableWarning,
   hasGasErrors,
   txParamsHaveBeenCustomized,
+  isNetworkBusy,
 }) {
   const t = useContext(I18nContext);
   const scrollRef = useRef(null);
@@ -93,7 +93,7 @@ export default function EditGasDisplay({
 
   useLayoutEffect(() => {
     if (showAdvancedForm && scrollRef.current) {
-      scrollRef.current.scrollIntoView();
+      scrollRef.current.scrollIntoView?.();
     }
   }, [showAdvancedForm]);
 
@@ -133,14 +133,6 @@ export default function EditGasDisplay({
   return (
     <div className="edit-gas-display">
       <div className="edit-gas-display__content">
-        {warning && !isGasEstimatesLoading && (
-          <div className="edit-gas-display__warning">
-            <ActionableMessage
-              className="actionable-message--warning"
-              message={warning}
-            />
-          </div>
-        )}
         {showTopError && (
           <div className="edit-gas-display__warning">
             <ErrorMessage errorKey={errorKey} />
@@ -156,6 +148,16 @@ export default function EditGasDisplay({
             />
           </div>
         )}
+        {isNetworkBusy ? (
+          <div className="edit-gas-display__warning">
+            <ActionableMessage
+              className="actionable-message--warning"
+              message={t('networkIsBusy')}
+              iconFillColor="#f8c000"
+              useIcon
+            />
+          </div>
+        ) : null}
         {mode === EDIT_GAS_MODES.SPEED_UP && (
           <div className="edit-gas-display__top-tooltip">
             <Typography
@@ -336,7 +338,6 @@ EditGasDisplay.propTypes = {
   estimatedMaximumFiat: PropTypes.string,
   dappSuggestedGasFeeAcknowledged: PropTypes.bool,
   setDappSuggestedGasFeeAcknowledged: PropTypes.func,
-  warning: PropTypes.string,
   transaction: PropTypes.object,
   gasErrors: PropTypes.object,
   gasWarnings: PropTypes.object,
@@ -346,4 +347,5 @@ EditGasDisplay.propTypes = {
   estimatesUnavailableWarning: PropTypes.bool,
   hasGasErrors: PropTypes.bool,
   txParamsHaveBeenCustomized: PropTypes.bool,
+  isNetworkBusy: PropTypes.bool,
 };
--- a/ui/components/app/edit-gas-display/edit-gas-display.test.js
+++ b/ui/components/app/edit-gas-display/edit-gas-display.test.js
@@ -1,0 +1,38 @@
+import React from 'react';
+import { screen } from '@testing-library/react';
+import { renderWithProvider } from '../../../../test/jest';
+import configureStore from '../../../store/store';
+import EditGasDisplay from '.';
+
+jest.mock('../../../selectors');
+jest.mock('../../../helpers/utils/confirm-tx.util');
+jest.mock('../../../helpers/utils/transactions.util');
+
+function render({ componentProps = {} } = {}) {
+  const store = configureStore({});
+  return renderWithProvider(<EditGasDisplay {...componentProps} />, store);
+}
+
+describe('EditGasDisplay', () => {
+  describe('if getIsNetworkBusy returns a truthy value', () => {
+    it('informs the user', () => {
+      render({ componentProps: { isNetworkBusy: true } });
+      expect(
+        screen.getByText(
+          'Network is busy. Gas prices are high and estimates are less accurate.',
+        ),
+      ).toBeInTheDocument();
+    });
+  });
+
+  describe('if getIsNetworkBusy does not return a truthy value', () => {
+    it('does not inform the user', () => {
+      render({ componentProps: { isNetworkBusy: false } });
+      expect(
+        screen.queryByText(
+          'Network is busy. Gas prices are high and estimates are less accurate.',
+        ),
+      ).not.toBeInTheDocument();
+    });
+  });
+});
--- a/ui/components/app/edit-gas-fee-popover/network-statistics/status-slider/status-slider.js
+++ b/ui/components/app/edit-gas-fee-popover/network-statistics/status-slider/status-slider.js
@@ -1,5 +1,6 @@
 import React from 'react';
 
+import { NETWORK_CONGESTION_THRESHOLDS } from '../../../../../../shared/constants/gas';
 import { useGasFeeContext } from '../../../../../contexts/gasFee';
 import I18nValue from '../../../../ui/i18n-value';
 import { NetworkStabilityTooltip } from '../tooltips';
@@ -24,24 +25,24 @@ const determineStatusInfo = (givenNetworkCongestion) => {
   const color = GRADIENT_COLORS[colorIndex];
   const sliderTickValue = colorIndex * 10;
 
-  if (networkCongestion <= 0.33) {
-    return {
-      statusLabel: 'notBusy',
-      tooltipLabel: 'lowLowercase',
-      color,
-      sliderTickValue,
-    };
-  } else if (networkCongestion > 0.66) {
+  if (networkCongestion >= NETWORK_CONGESTION_THRESHOLDS.BUSY) {
     return {
       statusLabel: 'busy',
       tooltipLabel: 'highLowercase',
       color,
       sliderTickValue,
     };
+  } else if (networkCongestion >= NETWORK_CONGESTION_THRESHOLDS.STABLE) {
+    return {
+      statusLabel: 'stable',
+      tooltipLabel: 'stableLowercase',
+      color,
+      sliderTickValue,
+    };
   }
   return {
-    statusLabel: 'stable',
-    tooltipLabel: 'stableLowercase',
+    statusLabel: 'notBusy',
+    tooltipLabel: 'lowLowercase',
     color,
     sliderTickValue,
   };
--- a/ui/components/app/edit-gas-fee-popover/network-statistics/status-slider/status-slider.test.js
+++ b/ui/components/app/edit-gas-fee-popover/network-statistics/status-slider/status-slider.test.js
@@ -20,31 +20,31 @@ const renderComponent = ({ networkCongestion }) => {
 
 describe('StatusSlider', () => {
   it('should show "Not busy" when networkCongestion is less than 0.33', () => {
-    const { queryByText } = renderComponent({ networkCongestion: 0.32 });
-    expect(queryByText('Not busy')).toBeInTheDocument();
+    const { getByText } = renderComponent({ networkCongestion: 0.32 });
+    expect(getByText('Not busy')).toBeInTheDocument();
   });
 
-  it('should show "Not busy" when networkCongestion is 0.33', () => {
-    const { queryByText } = renderComponent({ networkCongestion: 0.33 });
-    expect(queryByText('Not busy')).toBeInTheDocument();
+  it('should show "Stable" when networkCongestion is 0.33', () => {
+    const { getByText } = renderComponent({ networkCongestion: 0.33 });
+    expect(getByText('Stable')).toBeInTheDocument();
   });
 
   it('should show "Stable" when networkCongestion is between 0.33 and 0.66', () => {
-    const { queryByText } = renderComponent({ networkCongestion: 0.5 });
-    expect(queryByText('Stable')).toBeInTheDocument();
+    const { getByText } = renderComponent({ networkCongestion: 0.5 });
+    expect(getByText('Stable')).toBeInTheDocument();
   });
 
-  it('should show "Stable" when networkCongestion is 0.66', () => {
-    const { queryByText } = renderComponent({ networkCongestion: 0.66 });
-    expect(queryByText('Stable')).toBeInTheDocument();
+  it('should show "Busy" when networkCongestion is 0.66', () => {
+    const { getByText } = renderComponent({ networkCongestion: 0.66 });
+    expect(getByText('Busy')).toBeInTheDocument();
   });
 
   it('should show "Busy" when networkCongestion is greater than 0.66', () => {
-    const { queryByText } = renderComponent({ networkCongestion: 0.67 });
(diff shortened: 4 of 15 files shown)


ui/pages/confirm-transaction/confirm-transaction.component.js
```diff
@@ -54,21 +50,7 @@ export default class ConfirmTransaction extends Component {
     setDefaultHomeActiveTabName: PropTypes.func,
   };
 
-  constructor(props) {
```
mcmire, line 57: Another case in which we can use the new hooks.

ui/store/actions.js
```diff
@@ -2481,6 +2481,32 @@ export function setRecoveryPhraseReminderLastShown(lastShown) {
   };
 }
 
+/**
```
mcmire, line 2484: Just moved up from below, as these are AppTokenController methods not GasFeeController methods.

ui/store/actions.js
```diff
@@ -2786,40 +2812,25 @@ export async function updateTokenType(tokenAddress) {
...
  *
- * @returns {string} a unique identify of the polling request that can be used
- *  to remove that request from consideration of whether polling needs to
- *  continue.
+ * @param {string} item - The type of request to add to the polling queue (one
+ * of "gasFeeEstimates" or "isNetworkCongested").
  */
-export function getGasFeeEstimatesAndStartPolling() {
-  return promisifiedBackground.getGasFeeEstimatesAndStartPolling();
+export function updateWithAndStartPollingFor(item) {
```
mcmire, line 2821: These changes come from the corresponding commit in the `controllers` repo: https://github.com/MetaMask/controllers/pull/609

ui/pages/swaps/index.js
```diff
@@ -127,11 +128,10 @@ export default function Swap() {
...
 
-  if (networkAndAccountSupports1559) {
-    // This will pre-load gas fees before going to the View Quote page.
-    // eslint-disable-next-line react-hooks/rules-of-hooks
-    useGasFeeEstimates();
-  }
+  // This will preload gas fees before going to the View Quote page.
+  useSafeGasEstimatePolling({ isEnabled: networkAndAccountSupports1559 });
+  // This will pre-gauge the network before going to the View Quote page.
+  useSafeNetworkCongestionPolling({ isEnabled: networkAndAccountSupports1559 });
```
mcmire, line 134: I wanted to write an e2e test to cover this behavior but Ganache doesn't support EIP-1559 transactions yet :(

Any idea on how we want to test this? The test file that matches this file is pretty sparse.

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
```diff
@@ -48,14 +49,15 @@ export default function EditGasPopover({
...
   );
+  const isNetworkCongested = useSelector(getIsNetworkCongested);
 
   const showEducationButton =
     (mode === EDIT_GAS_MODES.MODIFY_IN_PLACE ||
       mode === EDIT_GAS_MODES.SWAPS) &&
     networkAndAccountSupport1559;
   const [showEducationContent, setShowEducationContent] = useState(false);
 
-  const [warning] = useState(null);
```
mcmire, line 58: I didn't see a need to use `useState` here but happy to do something else if there is a reason for it!

ui/ducks/metamask/metamask.js
```diff
@@ -340,3 +341,7 @@ export function getIsUnlocked(state) {
 export function getSeedPhraseBackedUp(state) {
   return state.metamask.seedPhraseBackedUp;
 }
+
+export function getIsNetworkCongested(state) {
+  return state.metamask.isNetworkCongested;
```
mcmire, line 346: This piece of data is populated in GasFeeController. You can see it in action in this PR: https://github.com/MetaMask/controllers/pull/609

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
jpuri: Agree it is not useful, these variable `warning` even if not in state does not seems very useful.

ui/components/app/edit-gas-popover/edit-gas-popover.test.js
```diff
@@ -0,0 +1,147 @@
...
+          cachedBalances: {},
+          provider: {},
+        },
+      },
+      overrides,
+    ),
+  );
+}
+
+setBackgroundConnection({
```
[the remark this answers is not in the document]
jpuri: Nice work on this test case 👍

ui/components/app/edit-gas-popover/edit-gas-popover.test.js
```diff
@@ -0,0 +1,147 @@
...
+              userFeeLevel: 'medium',
+            }}
+          />,
+          store,
+        );
+        expect(
+          getByText(
+            'Network is busy. Gas prices are high and estimates are less accurate.',
+          ),
+        ).toBeInTheDocument();
```
jpuri, line 143: Here it will be useful to test the absence of above message also if gas fee is not high.

ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container-component.test.js
```diff
@@ -1,19 +1,24 @@
...
 jest.mock('../../../../store/actions', () => ({
-  disconnectGasFeeEstimatePoller: jest.fn(),
-  getGasFeeEstimatesAndStartPolling: jest
+  stopPollingFor: jest.fn(),
+  updateWithAndStartPollingFor: jest
     .fn()
     .mockImplementation(() => Promise.resolve()),
   addPollingTokenToAppState: jest.fn(),
+  removePollingTokenFromAppState: jest.fn(),
 }));
```
jpuri, line 23: `'SLOW'` I think is not values value for `gasEstimateType`, but yep ok as a dummy value.

ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container-component.test.js
```diff
@@ -87,12 +95,76 @@ describe('GasModalPageContainer Component', () => {
...
+        );
+        expect(updateWithAndStartPollingFor).toHaveBeenCalledWith(
+          'gasFeeEstimates',
+        );
+      });
+    });
+
+    describe('when the network supports EIP 1559', () => {
+      it('should call updateWithAndStartPollingFor twice, once to get gas fee estimates and another time to gauge network congestion', () => {
+        expect(updateWithAndStartPollingFor).not.toHaveBeenCalled();
```
jpuri, line 133: I have a doubt here, since it keeps polling already why is a second call needed ?

jpuri:
Great work @mcmire , esp with test coverage.

ui/components/app/gas-customization/gas-modal-page-container/gas-modal-page-container-component.test.js
mcmire: This was the way that `getGasFeeEstimatesAndStartPolling` worked before, so I copied the behavior over, but my guess as to why it works this way is that the timer is set to 15 seconds by default, so if we simply started polling then when the user opened the transaction confirmation screen, they wouldn't see any estimates until 15 seconds had passed. So we make the request once at the beginning manually to populate an initial set of estimates, so that the confirmation screen shows correctly, then start the timer.

package.json
```diff
@@ -104,7 +104,7 @@
     "@fortawesome/fontawesome-free": "^5.13.0",
     "@material-ui/core": "^4.11.0",
     "@metamask/contract-metadata": "^1.28.0",
-    "@metamask/controllers": "^17.0.0",
+    "@metamask/controllers": "MetaMask/controllers#5e18a39556008dd866bf1dece7861e7f382cb30a",
```
mcmire, line 107: This is temporary and will be removed when the controllers PR is merged.

app/scripts/metamask-controller.js
```diff
@@ -217,6 +217,9 @@ export default class MetamaskController extends EventEmitter {
       ),
       legacyAPIEndpoint: `${gasApiBaseUrl}/networks/<chain_id>/gasPrices`,
       EIP1559APIEndpoint: `${gasApiBaseUrl}/networks/<chain_id>/suggestedGasFees`,
+      determineNetworkStatusInfoUrlTemplate: ({ chainId }) => {
```
mcmire, line 220: The changes introduced in https://github.com/MetaMask/controllers/pull/609 require a new argument to be passed to GasFeeController to specify the new endpoint we hit to determine whether the network is busy.

ui/ducks/metamask/metamask.js
```diff
@@ -399,3 +400,7 @@ export function doesUserHaveALedgerAccount(state) {
     return kr.type === KEYRING_TYPES.LEDGER;
   });
 }
+
+export function getIsNetworkBusy(state) {
+  return state.metamask.isNetworkBusy;
```
mcmire, line 405: This will get updated automatically by GasFeeController.

ui/components/app/edit-gas-display/edit-gas-display.component.js
```diff
@@ -91,7 +91,7 @@ export default function EditGasDisplay({
 
   useLayoutEffect(() => {
     if (showAdvancedForm && scrollRef.current) {
-      scrollRef.current.scrollIntoView();
+      scrollRef.current.scrollIntoView?.();
```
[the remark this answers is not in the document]
mcmire: This is still applicable, only I ended up writing a test for EditGasDisplay instead.

ui/components/app/edit-gas-display/edit-gas-display.component.js
```diff
@@ -133,14 +133,6 @@ export default function EditGasDisplay({
   return (
     <div className="edit-gas-display">
       <div className="edit-gas-display__content">
-        {warning && !isGasEstimatesLoading && (
```
mcmire, line 136: I opted to take off the `warning` prop because I didn't see it getting used. We can very easily just add a warning ourselves to this component.

ui/components/app/edit-gas-fee-popover/network-statistics/status-slider/status-slider.js
```diff
@@ -24,24 +25,24 @@ const determineStatusInfo = (givenNetworkCongestion) => {
   const color = GRADIENT_COLORS[colorIndex];
   const sliderTickValue = colorIndex * 10;
 
-  if (networkCongestion <= 0.33) {
```
mcmire, line 27: I changed this slightly so that instead of:

* not busy: <= 0.33
* stable: > 0.33, <= 0.66
* busy: > 0.66

It is now:

* not busy: 0 up to 0.33
* stable: 0.33 up to 0.66
* busy: 0.66-

This is not strictly necessary to change, but given that I need to _just_ test that the network is busy elsewhere, it made more sense in my head to say that the status changes to busy as soon as `networkCongestion` crosses 0.66 instead of as soon as it exceeds it. Happy to change this back though.

ui/components/app/edit-gas-fee-popover/network-statistics/status-slider/status-slider.test.js
```diff
@@ -20,31 +20,31 @@ const renderComponent = ({ networkCongestion }) => {
 
 describe('StatusSlider', () => {
   it('should show "Not busy" when networkCongestion is less than 0.33', () => {
-    const { queryByText } = renderComponent({ networkCongestion: 0.32 });
```
mcmire, line 23: Apparently I forgot the difference between `query*` and `get*` 😆  But at the same time all of these tests change because the definition changed.

ui/ducks/metamask/metamask.js
```diff
@@ -357,6 +361,13 @@ export function getIsGasEstimatesLoading(state) {
   return isGasEstimatesLoading;
 }
 
+export function getIsNetworkBusy(state) {
+  const gasFeeEstimates = getGasFeeEstimates(state);
+  return (
+    gasFeeEstimates?.networkCongestion >= NETWORK_CONGESTION_THRESHOLDS.BUSY
```
mcmire, line 367: Here's the corresponding busy check as referenced above.

ui/pages/confirm-transaction-base/transaction-alerts/transaction-alerts.js
```diff
@@ -105,6 +106,13 @@ const TransactionAlerts = ({
           type="warning"
         />
       )}
+      {isNetworkBusy ? (
```
mcmire, line 109: This warning applies to the EIP-1559 V2 UI and will appear on the confirm transaction screen. It will NOT appear within the edit gas modal because we already have the status slider which indicates the status of the network.

ui/pages/confirm-transaction-base/transaction-alerts/transaction-alerts.test.js
```diff
@@ -1,172 +1,307 @@
 import React from 'react';
```
mcmire, line 1: I opted to refactor these tests so that they're simpler and don't require us to recreate the Redux state so much. I can put these changes in another PR if necessary, but I will point the additional check for `isNetworkBusy` regardless.

ui/pages/confirm-transaction-base/transaction-alerts/transaction-alerts.test.js
```diff
@@ -1,172 +1,307 @@
...
+        });
+        expect(
+          queryByText(
+            'Future transactions will queue after this one. This price was last seen was some time ago.',
+          ),
+        ).not.toBeInTheDocument();
+      });
+    });
+
+    describe('if isNetworkBusy from useGasFeeContext is truthy', () => {
```
mcmire, line 200: Here are the new tests for the isNetworkBusy warning.

ui/selectors/transactions.js
```diff
@@ -10,8 +10,11 @@ import {
   TRANSACTION_TYPES,
 } from '../../shared/constants/transaction';
 import { transactionMatchesNetwork } from '../../shared/modules/transaction.utils';
-import { getCurrentChainId, deprecatedGetCurrentNetworkId } from './selectors';
```
mcmire, line 13: The main change here is to have `getSelectedAddress` come from `ui/selectors/selectors` and not from `ui/selectors`. The previous change created a circular dependency between `ui/selectors` and `ui/selectors/transactions`, which prevents us from `jest.mock`'ing `ui/selectors` correctly in component tests.

ui/components/app/edit-gas-display/edit-gas-display.test.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import { screen } from '@testing-library/react';
+import { renderWithProvider } from '../../../../test/jest';
+import configureStore from '../../../store/store';
+import EditGasDisplay from '.';
+
+jest.mock('../../../selectors');
```
mcmire, line 7: This might have been the simplest component test file I've ever written 😆

ui/components/app/edit-gas-display/edit-gas-display.component.js
```diff
@@ -156,6 +148,16 @@ export default function EditGasDisplay({
             />
           </div>
         )}
+        {isNetworkBusy ? (
```
mcmire, line 151: This warning applies to the EIP-1559 V1 UI. Note that this warning appears within the edit gas modal, NOT the confirm transaction screen, unlike EIP-1559 V2. I could have added it to the confirm transaction screen, but it would be more work. Given that we are soon transitioning to the V2 UI I feel like this is an okay compromise for now, but happy to change if not.

jpuri: Looks good, in V@ as I understand we are showing network busy only on edit gas fee modal.

jpuri (approved):
LGTM

ui/components/app/edit-gas-display/edit-gas-display.component.js
mcmire: In V2 it's slightly different — we are showing the network busy message on the confirm transaction screen, but not in the edit gas fee modal.

NiranjanaBinoy (approved):

PR #12270, opened 2021-10-02
build system task output improvement

kumavis:
concurrently does have a way of adding jobs after instantiation, and so we run them one by one
it really isnt providing very much here, just prefixing the output.
it also only wraps the output for tasks run in child processes.

image


kumavis:
test failures bc i didnt update lavamoat policy

brad-decker:
@kumavis should this be fixed up and made reviewable or closed?

kumavis:
image
pretty! but stats look wrong. also the process isnt closing

brad-decker (approved):
Looks good to me! 💯

PR #12278, opened 2021-10-04
Fix unit test package scripts

rekmarks:
This PR fixes our local unit test package scripts. When the state migration unit tests were migrated to Jest in #12106, it left the `test:unit` script in a broken state, because it didn't tell `mocha` to ignore the state migration tests.

Arguably, that script was already broken, since the most reasonably expectation from its name is that it runs _all_ unit tests. The PR makes it so that it does just that, by means of `concurrently`.

Unfortunately, `concurrently` only outputs errors from child processes once (at the time when they exit, https://github.com/open-cli-tools/concurrently/issues/134). This means that we have to search/navigate the output for this combined script to identify the failure. That said, it's better than the status quo.


kumavis (approved):

jpuri (approved):
Nice change 👍

ryanml (approved):

PR #12291, opened 2021-10-07
Adds sign typed data type to metrics payload for sign typed data events

### LINKED ISSUE #10250: Add distinct metrics events for various signature methods
> we have metrics on signature confirmation but they don't distinguish between eth_sign, personal_sign, eth_signTypedData, eth_signTypedData, eth_signTypedData_v1, and eth_signTypedData_v2. They're all lumped into the same request.
> And we don't have metrics for eth_signTypedData_v3 or eth_signTypedData_v4 at all

We should remedy this, this would be good data for what methods we need to support best.

ryanml:
Fixes: https://github.com/MetaMask/metamask-extension/issues/10250

Additionally adds tracking events for `eth_signTypedData_v3` and `eth_signTypedData_v4`

Example payloads: 
Screen Shot 2021-10-08 at 1 51 17 PM

Screen Shot 2021-10-08 at 1 51 03 PM

Screen Shot 2021-10-08 at 1 51 41 PM

Screen Shot 2021-10-08 at 1 51 29 PM

### DIFF
--- a/ui/components/app/signature-request-original/signature-request-original.component.js
+++ b/ui/components/app/signature-request-original/signature-request-original.component.js
@@ -285,7 +285,9 @@ export default class SignatureRequestOriginal extends Component {
       history,
       mostRecentOverviewPage,
       sign,
+      txData: { type },
     } = this.props;
+    const { metricsEvent, t } = this.context;
 
     return (
       <div className="request-signature__footer">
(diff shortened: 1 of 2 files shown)


mcmire (approved):
Makes sense to me 👍🏻

brad-decker (approved):

PR #12303, opened 2021-10-07
Update `immer` from v8 to v9

Gudahtt:
`immer` has been updated to v9. This didn't require any changes on our part; [the only breaking changes are to the TypeScript types](https://github.com/immerjs/immer/releases/tag/v9.0.0).

The `@reduxjs/toolkit` library has been updated as well, to ensure that it's using the updated version of Immer internally as well. This update makes our patch of that package obsolete, as the problematic pattern that were were patching out is no longer present.


danjm (approved):
LGTM

jpuri (approved):

jpuri (commented):
test-deps-audit is still breaking in this PR

Gudahtt:
Yes, this only addresses the immer-related advisories. Two others remain.

PR #12309, opened 2021-10-08
Removing the icronUrl from fixture state for chrome e2e test failure

NiranjanaBinoy:
Fixing e2e test failure in chrome because CoinGecko iconUrls are not loading.


adonesky1 (approved):
LGTM

Gudahtt (approved):
Not ideal to have no icons in our e2e tests :thinking: But oh well, we can figure that out later. This looks OK for now :+1:

PR #12313, opened 2021-10-08
Simplify Mocha npm scripts

Gudahtt:
The npm scripts used to run Mocha scripts have been greatly simplified. As we transition more tests from Mocha to Jest it was becoming increasingly difficult to update the CLI arguments to keep all of these scripts working correctly. This reorganization should make that process much simpler.

The base Mocha options are in `.mocharc.js` - all except for the target tests to run. Those are still given via the CLI. There is a second config file specifically for the `test:unit:lax` tests (i.e. the Mocha tests that have no coverage requirements) because it requires a change to the `ignored` configuration property. We can create an additional configuration file for each test script we add that needs further configuration changes.

The `test:unit:path` script used to be used to run Mocha tests at a given path. Now that can be done using `yarn mocha` instead, so this script has been removed.

The `yarn watch` command has been broken for some time now, so it has been removed as well. Mocha tests can still be run with a file watcher using `yarn mocha --watch <path>` or `yarn test:unit:mocha --watch`.

The README has been updated to remove references about the `watch` command that was removed. I considered explaining the other test scripts there as well, but they were difficult to explain I will attempt to update the README after making further simplifications instead.

Manual testing steps:  
  - Try running each of the updated scripts, and ensure they run the same tests as they did before.
  - Try running the suggested equivalent commands for the removed scripts.


darkwing (approved):

darkwing (approved):

jpuri (approved):
Nice 👍

PR #12315, opened 2021-10-08
Do case-insensitive comparisons of token addresses in view-quote and build-quote

danjm:
Fixes #12306

We recently change the way that token addresses are saved in state, so that they are now checksummed. This breaks attempts to compare token addresses in state with token addresses from the swaps apis. This PR corrects those comparisons.


ryanml (approved):
lgtm!

adonesky1 (approved):
LGTM

PR #12322, opened 2021-10-09
feat: allow adding rpc endpoints with same chain id

gakonst:
As title, the motivation is to allow adding new RPCs with existing chain ids. 

An example of that is the Flashbots RPC which provides users with frontrunning/backrunning protection and non-reverting transactions.

### DIFF
--- a/app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
+++ b/app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
@@ -113,14 +113,6 @@ async function addEthereumChainHandler(
     );
   }
 
-  if (CHAIN_ID_TO_NETWORK_ID_MAP[_chainId]) {
-    return end(
-      ethErrors.rpc.invalidParams({
-        message: `May not specify default MetaMask chain.`,
-      }),
-    );
-  }
-
   const existingNetwork = findCustomRpcBy({ chainId: _chainId });
 
   if (existingNetwork) {


gakonst:

CodeForcer:
+1. This PR is really important and fixes a major and longstanding UX bug in MetaMask wallet

danfinlay:
I agree there is probably some way this could work (same changes would be needed on mobile).

I think the changes would need to be contingent that the confirmation view made clear the risk of trusting a bad replacement (incorrect state on that network and potentially censored or manipulated transaction submission, to try naming it generally).

I also would prefer that the proposed name be shown alongside the former name, and make it clear that the renaming will not affect the sensitivity of using that network: Transactions to this new provider will still be valid on the previous network. If this network is presenting itself as a test network but replacing a real network, it is still sending transactions for that real network.

ernestognw:
> I agree there is probably some way this could work (same changes would be needed on mobile).
> 
> I think the changes would need to be contingent that the confirmation view made clear the risk of trusting a bad replacement (incorrect state on that network and potentially censored or manipulated transaction submission, to try naming it generally).
> 
> I also would prefer that the proposed name be shown alongside the former name, and make it clear that the renaming will not affect the sensitivity of using that network: Transactions to this new provider will still be valid on the previous network. If this network is presenting itself as a test network but replacing a real network, it is still sending transactions for that real network.

I think it's almost the same risk for non-default networks like BSC. Since they're not already added, if you go to somewhere like https://chainlist.org you're still at risk of having a fake RPC endpoint added, so I think that this change is enough for now.

Any extra UI changes to be made should be in another PR to avoid blocking things like https://flashbotsrpc.com.

I can help with the UI changes if needed.

danfinlay:
> I think it's almost the same risk for non-default networks like BSC.

I don't think that's true at all. If you add a new network and it's your first time using it, you are not going to put any of your previous assets at risk. If you suddenly add a new network for a chain that you have existing assets on, you are now prone to confusion & censorship of those previous assets, particularly if you thought you were adding a new network (not just updating an existing one).

> Any extra UI changes to be made should be in another PR to avoid blocking

I don't think security considerations can be considered "non blocking". This policy _may_ be overly strict, but being overly strict is preferable to being insecure. We'll consider this carefully and proceed from there.

gakonst:
Hey Dan, just wanted to bump this issue. 

> A new RPC is trusted to represent your balance on the given network, and can be used to deceive you into thinking you have received assets that you do not have.
> A new RPC is trusted to broadcast your transactions, and can potentially either censor or delay them.
It's not "A New RPC is trusted", it's "Any RPC is trusted". 

Regarding your two above, they also apply when users use Metamask to connect to the 'official' APIs (e.g. Infura), i.e. it's not "A new RPC", it's "Any RPC".

Given that, I do not really think that warning a user about adding new RPCs is a blocker for this PR (rather, it's part of the general "any RPC is ultimately trusted"), and I'd track it as a separate item.

cc @danfinlay

xinnanyemm:
Please note that `label:team-dapp-api` has been replaced by `label:team-devex` . Thanks.

PR #12331, opened 2021-10-11
Fix toggle button 'off' color setting

jayeclark:
Fixes: #12296 

Explanation:  Updated color setting in the toggle button component. 

toggle buttons with correct color scheme


danjm (approved):
Thanks for the contribution @jayeclark!

This looks good to me

darkwing (approved):

PR #12354, opened 2021-10-13
Refactor checking if address is contract into a new module.

### LINKED ISSUE #12350: Create helper method for determining whether an address is a contract
We currently have two locations where we check whether a given address is a contract:

https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/controllers/transactions/index.js#L1238-L1250

https://github.com/MetaMask/metamask-extension/blob/develop/ui/ducks/send/send.js#L229-L241

We should abstract the common logic into a helper method. This will then be reusable in other locations in the future.

The helper method should probably go somewhere in the `shared/` directory

Let's add appropriate unit tests too.

segun:
Fixes: #12350

Explanation:  

We currently have two locations where we check whether a given address is a contract:
We should abstract the common logic into a helper method. This will then be reusable in other locations in the future.

  - 
  - 
  -

### DIFF
--- a/app/scripts/controllers/transactions/index.js
+++ b/app/scripts/controllers/transactions/index.js
@@ -39,6 +39,7 @@ import {
   CHAIN_ID_TO_GAS_LIMIT_BUFFER_MAP,
 } from '../../../../shared/constants/network';
 import { isEIP1559Transaction } from '../../../../shared/modules/transaction.utils';
+import { addressIsContract } from '../../../../shared/modules/contract-utils';
 import TransactionStateManager from './tx-state-manager';
 import TxGasUtil from './tx-gas-utils';
 import PendingTransactionTracker from './pending-tx-tracker';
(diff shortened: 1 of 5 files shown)


segun:
> @segun Was looking through some stuff and it looks like we already have this function:
> 
> https://github.com/MetaMask/metamask-extension/blob/864800b035d722db97641c01cff9b46a3c260ccf/ui/helpers/utils/transactions.util.js#L146-L151
> 
> Is it possible to reuse this somehow? Or, maybe that function should be changed to use your function?

Yeah, I think the idea is this was done is a few places and we should make sure it's only being done in a single place. I somehow missed this one. I will update to use the new code.

mcmire (dismissed):
Looks good!

jpuri (dismissed):
Looks good, I added small feedback.

danjm (approved):
LGTM!

mcmire (approved):

darkwing (approved):

Issue #10345, opened 2021-02-02
Approve Confirmation design improvements


rachelcope:
#Overview
Improve the Approve Confirmation designs

* ~~**Copy "To" contract address** As a user, I want to be able to copy / expand the contract address in ‘To:’ on Approve screen~~ (This will be handled as part of https://github.com/MetaMask/metamask-extension/issues/12351)
* **Remove "Https://" from origin** Remove "Https://" from the header and subheader (this is handles as a part of https://github.com/MetaMask/metamask-extension/pull/12360)
* **Update designs/components** Update header components to the updated transaction confirmation components

JSRossiter:
I think it would be a big improvement if the Approve Confirmation screen respected "Advanced gas controls" and "Customize transaction nonce" settings. Many users (myself included) set the gas manually for every transaction and this currently takes several clicks for approves. There is no current way to customize the nonce for an approval transaction.

kevinghim:
Related  [Remove "Https://" from origin Remove "Https://" from the header and subheader](https://www.notion.so/Remove-Https-from-origin-Remove-Https-from-the-header-and-subheader-818810f0780c4fd0b35e4c62690381e1)

PR #12360, opened 2021-10-14
Approve screen header and sub-header improvement

### LINKED ISSUE #7574: When submitting an ERC20 token approval this is not for a specific site but for a contract
This message is extremely inaccurate as it implies the user is allowing a site to manage the token but in reality it is a contract that could be used by many different sites. I don't think simplifying this to the user by singling out the domain this is used on is in any way helpful. 

If I authorize a contract, I can't control the domains it's used on.

token_allowance

NiranjanaBinoy:
This is a part of #10345.
Fixes: #7574 

Problem: it is not "the site" that has access to the token. It is "the contract"

Change:

- Update header copy: "Give permission to access your [token]?”
- Update subheader copy: "By granting permission, you are allowing the following [contract/account] to access your funds”
- Include contract address and its jazzicon as additional information; allow users to copy the address or view it in the block explorer
- Use the most recent header icon style to show the site that is interacting with MM

Screen Shot 2021-10-28 at 2 42 26 PM

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -142,9 +142,9 @@
   "allowExternalExtensionTo": {
     "message": "Allow this external extension to:"
   },
-  "allowOriginSpendToken": {
-    "message": "Allow $1 to spend your $2?",
-    "description": "$1 is the url of the site and $2 is the symbol of the token they are requesting to spend"
+  "allowSpendToken": {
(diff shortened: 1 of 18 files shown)


jpuri (dismissed):
Looks good, small comment added.

ryanml (approved):

jpuri (approved):

Issue #12295, opened 2021-10-07
Design updates in the transaction details view


danjm:
![Screenshot from 2021-10-07 07-20-49](https://user-images.githubusercontent.com/7499938/136361957-33ceee7c-4dba-49a4-abbe-c188cd2f147f.png)

2 in the above list can be ignored as address truncation has been handled in a separate PR

To see this component, create and send a transaction. Then click on that transaction in the transaction activity list.

Gudahtt:
@igorms-cons Can this issue be closed?

PR #12364, opened 2021-10-15
[DEL] Fix/ux papercuts transaction details issue #12295

igorms-cons:
updating ui on transaction modal view


igorms-cons:
image

Also on the issue https://github.com/MetaMask/metamask-extension/issues/12295 it was mentioned to pay attention to the format dat (24:40 instead of 00:40). Of course i did not tried to send a transaction at midnight but when i test with a milli timestamp between 0:00am and 1:am, the format was good. 

what is happening for the CLA? the 2nd is my good username, the one inside metamask team :)

PR #12366, opened 2021-10-15
Use consistent naming for welcome route of onboarding

darkwing:
This was the only route that didn't end in `ROUTE`, so I've fixed that.


adonesky1 (approved):
Good catch

ryanml (approved):

PR #12370, opened 2021-10-17
Preventing swaps data re-fetch when navigating back to home screen

ryanml:
Resolves many of: https://sentry.io/organizations/metamask/issues/?project=273505&query=is%3Aunresolved+Swaps+API+calls+are+disabled+for+chainId&sort=freq&statsPeriod=14d

Steps to reproduce/test:

1. Add the Polygon (Matic network)
2. Switch to the network
3. Click "Swap"
4. Click "Cancel"
5. Observe the errors per Sentry in the console

Screen Shot 2021-10-16 at 5 26 24 PM

This occurs because the swaps state is reset before pushing the default route to history, that sets `useNewSwapsApi` from `true` -> `false`, thereby re-triggering [this](https://github.com/MetaMask/metamask-extension/blob/develop/ui/pages/swaps/index.js#L195) block to run again with the new parameter. The fix includes caching the existing `useNewSwapsApi`, and preventing a refetch if it changes. (This fetch needs to only happen once)

A general smoke/regression test for Swaps would be a good thing to do when testing this fix out.

### DIFF
--- a/ui/pages/swaps/index.js
+++ b/ui/pages/swaps/index.js
@@ -118,6 +118,7 @@ export default function Swap() {
   const chainId = useSelector(getCurrentChainId);
   const isSwapsChain = useSelector(getIsSwapsChain);
   const useNewSwapsApi = useSelector(getUseNewSwapsApi);
+  const prevUseNewSwapsApi = useRef(useNewSwapsApi);
   const networkAndAccountSupports1559 = useSelector(
     checkNetworkAndAccountSupports1559,
   );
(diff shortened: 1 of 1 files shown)


dan437 (approved):
@ryanml Hi Ryan, thanks a lot for finding the root issue and resolving this. I've reproduced the issue locally and then gotten your branch and it seems good - no unnecessary re-fetch. I've done some testing and swapping seems to be working properly. It would be great if our QA folks can have a look as well.

danjm (approved):

PR #12403, opened 2021-10-19
fix issue where regional locales are not matched against our locales  list

### LINKED ISSUE #12353: Incorrect default locale
Screen Shot 2021-10-13 at 6 08 42 PM

adonesky1:
Fixes: #12353

Explanation:  Some users experience issues in onboarding if their browsers' language settings are set such that their primary language is coded regionally, and another secondary language - for which we have translations - is not. Before this PR we were not normalizing regionally coded language settings to their base (e.g. 'en-us' => 'en') if we didn't have that regional coding in our translations. This PR adds this normalization.

Manual testing steps:  
  - Change the language primary language in your browser's preferences to one with regional coding (e.g. French (Switzerland), code 'fr-CH'), for which we have translations for the non-regionally coded (in this case just French (code 'fr')
  - Reload the app
  - Check that the language is set to the language you just changed it to (French if you followed step 1)

### DIFF
--- a/app/scripts/controllers/preferences.js
+++ b/app/scripts/controllers/preferences.js
@@ -68,7 +68,7 @@ export default class PreferencesController {
     this.store.setMaxListeners(12);
     this.openPopup = opts.openPopup;
     this.migrateAddressBookState = opts.migrateAddressBookState;
-
+    this.setCurrentLocale(opts.initLangCode);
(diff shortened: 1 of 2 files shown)


Gudahtt (approved):
LGTM!

jpuri (approved):

PR #12405, opened 2021-10-19
Fix weth->eth direct wrapping allowance issue

meppsilon:
Fixes: Issue with trading WETH->ETH throwing an error due to allowance.

Explanation: I left URL logic in this PR to allow for testing but the URL logic does not need to go into production.

Manual testing steps:  
  - Start with a new wallet
  - Trade Eth for Weth
  - Trade back Weth to Eth
  - See that the trade succeeds


dan437:
Great job! It seems the issue was happening if a user has never done WETH -> ETH before. For users that did WETH -> ETH before, it seemed to be working. We tested this on a testnet and it seems to be working now. PR looks great once the testnet URL logic is updated and everything is green.

dan437 (approved):

danjm (approved):

PR #12406, opened 2021-10-19
Use ternary conditionals instead of &&

darkwing:
After some discussion with @Gudahtt in Slack (https://consensys.slack.com/archives/GTQAGKY5V/p1634564083331400), one of the patterns we're using for conditionally rendering component is slightly unsafe, in that we can sometimes render unwanted `0`'s.  This should be safer.


mcmire:
Do we know if there is a way to enforce this using ESLint, or is this just something we will have to keep in mind?

mcmire (approved):

adonesky1 (approved):
LGTM!

Issue #12093, opened 2021-09-13
Uncled transaction showed as confirmed


wjmelements:
Metamask 10.0.3

A pending transaction was included in a block, so Metamask UI now shows it is confirmed.
But it is not actually confirmed; it is still pending because that block was uncled.

Metamask should be able to detect this situation via account nonce; if the account nonce is less than a confirmed tx nonce, that tx should return to the pending state.

danjm:
Thank you for the report @wjmelements

jpuri:
Ethereum blockchain can end up generating multiple heads at a point and this is not rare, eventually only one of these will make it to the main chain, details here: https://blog.ethereum.org/2015/08/08/chain-reorganisation-depth-expectations/.
The others blocks get uncled and their transactions are reverted.

As the block get older its chances of getting uncled are reduced, a transaction in a deep enough block can be considered to be final on the chain. This depth is called `number of confirmations` in bitcoin terms. 

Etherscan shows this confirmation count next to the transaction:
<img width="867" alt="Screenshot 2021-09-24 at 3 18 30 PM" src="https://user-images.githubusercontent.com/2182307/134655151-011558a0-4f83-4eeb-9058-427dcdbe61b7.png">

Thus approach to fix this issue could be:

1. We should consider a transaction as confirmed only after it has reached a reasonable `number of confirmations`
Or we show next to transaction the confirmation number and let user derive confidence out of it.
2. If a block eventually gets uncled it will appear as `uncle` of a block on the main chain and at that point, we move all of its transactions back to pending state.

There are 2 implementation level details to be handled here:
1. Calculating `number of confirmations`: this can be calculated from current block number and block number of transaction's block.
2. Checking if certain block has been uncled: as we get latest block on the chain we can go back its parents `number of confirmtions` times and see if a block has uncled block added, all the transactions of uncled block can be rolledback to pending. High chances are that they will eventually get confirmed in some other block.

We need to decide on a reasonable value for `number of confirmations`. Ref: https://ethereum.stackexchange.com/questions/319/what-number-of-confirmations-is-considered-secure-in-ethereum

jpuri:
Related issue: https://github.com/MetaMask/eth-block-tracker/issues/30

jpuri:
More useful resources:
https://blog.ethereum.org/2016/10/31/uncle-rate-transaction-fee-analysis/
[...]

PR #12407, opened 2021-10-19
Uncled transactions

jpuri:
This draft PR is POC for uncled transactions. Uncled transactions are handled by:

1. introducing new state accepted for transactions, as transaction is mined in a block it is in accepted state
2. after 12 confirmations transaction is moved to final state
3. in case block of transaction become uncled before 12 confirmations, transaction is moved back to submitted / pending state

Ref: https://github.com/MetaMask/metamask-extension/issues/12093#issuecomment-944063736

https://user-images.githubusercontent.com/2182307/138050009-5b39a218-d344-4070-ad27-02865f95cdcf.mov


Gudahtt (commented):
Have you considered updating `eth-block-tracker` to detect uncled blocks instead? This solution adds a bunch of extra polling to the extension, and it still leaves plenty of opportunities for transactions to remain stuck with the wrong status.

`eth-block-tracker` should be the only place where we poll the network. Everything else is triggered by the new latest block event. This poll doesn't respect "quiet mode" either, where we stop polling when the UI closes. It would be easier to manage if `eth-block-tracker` handled this detection.

Also this would only  detect when a confirmed transaction gets un-confirmed. Uncled blocks can result in more status changes than just that. For example, an uncled block can make a cancelled transaction confirmed, or it can move from confirmed to pending, or it can result in it being confirmed but with different transaction parameters (e.g. if the original succeeded rather than the speedup). This isn't a complete solution anyway. And I don't think it can be - not if it's implemented here.

jpuri:
The PR was POC closing this and it is not intended to be merged, it is only for reference.

PR #12408, opened 2021-10-19
Avoid storybook-build directory when linting

darkwing:
My `yarn lint:fix` has been hanging for a few weeks now.  After running `./node_modules/eslint/bin/eslint.js . --ext js --debug`, I noticed the last item being used was in the `storybook-build` directory.  Ignoring that directory made the linting finish, albeit in it still took ~80 seconds.


jpuri (approved):

Gudahtt (approved):
Good catch!

PR #12410, opened 2021-10-19
Make showFiatInTestnets only apply to known Testnets, not all non-mainnet networks

### LINKED ISSUE #12119: Make ShowFiatInTestNets preference only apply to test-nets (not all custom networks)
**Describe the Issue**
Currently the `ShowFiatInTestNets` preference in advanced settings disables or enables users to see available fiat conversion rates for all custom networks **in addition to** test-nets.

**Expected behavior**
This setting should only enable and disable available conversion rates for **test-nets**

adonesky1:
Fixes: #12119

Explanation: For a long time there has been a setting/preference called showFiatInTestNets corresponding to 
Screen Shot 2021-10-19 at 5 15 05 PM
in advanced settings.

I've raised and [discussed](https://consensys.slack.com/archives/G1L7H42BT/p1631727338156600) the fact that this setting actually applies to all non-mainnet networks. Its not a large code change so I thought I'd just make it. But I still think we need to be very careful here. While this makes the setting actually do what it claims to do, it reinforces a potentially problematic pattern where we are showing users fiat (or whatever secondary currency they choose) conversion rates for networks purely on the basis of a string match against the user (or API) input value for currency symbol in the network form, which is vulnerable to user error and malicious usages of the `addCustomRpc` method:
Screen Shot 2021-10-19 at 5 20 54 PM
This problem already exists if a user has the preference toggled on but this change would impose it on all users. I'm probably missing some parts of the picture here but wanted to re-raise the topic.

### DIFF
--- a/test/e2e/tests/send-edit.spec.js
+++ b/test/e2e/tests/send-edit.spec.js
@@ -34,7 +34,7 @@ describe('Editing Confirm Transaction', function () {
         const transactionAmount = transactionAmounts[0];
         assert.equal(await transactionAmount.getText(), '1');
 
-        const transactionFee = transactionAmounts[1];
+        const transactionFee = transactionAmounts[3];
         assert.equal(await transactionFee.getText(), '0.00025');
 
         await driver.clickElement(
@@ -62,12 +62,12 @@ describe('Editing Confirm Transaction', function () {
 
         // has correct updated value on the confirm screen the transaction
         const editedTransactionAmounts = await driver.findElements(
-          '.transaction-detail-item__row .transaction-detail-item__detail-text .currency-display-component__text',
+          '.currency-display-component__text',
         );
-        const editedTransactionAmount = editedTransactionAmounts[0];
+        const editedTransactionAmount = editedTransactionAmounts[3];
         assert.equal(await editedTransactionAmount.getText(), '0.0008');
 
-        const editedTransactionFee = editedTransactionAmounts[1];
+        const editedTransactionFee = editedTransactionAmounts[6];
         assert.equal(await editedTransactionFee.getText(), '2.2008');
 
         // confirms the transaction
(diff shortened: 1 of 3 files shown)


darkwing (approved):

jpuri (approved):

Gudahtt (dismissed):
Marking this as "Request changes" for now, so long as the concern about custom networks is outstanding. That could be incredibly confusing for users of custom networks.

adonesky1:
@Gudahtt I suppose the ancillary question you may be raising is should the UI default to showing the ETH ticker symbol even if it doesn't impact the conversionRate question.

Gudahtt:
Phew, that's good :relieved:. If we set ETH in the background state, we'd really be in a bad place. I started tracing that path and found this line: https://github.com/MetaMask/metamask-extension/blob/develop/ui/store/actions.js#L1458
Which made me think it was set in the background too. But I suspect that default is dead code, because `ticker` is initialized with an empty string and is never set to `undefined`.

Gudahtt:
>should the UI default to showing the ETH ticker symbol? even if it doesn't impact the conversionRate question.

Great question. Probably not, but I don't think any default choice makes sense. It would be great if we could make that a mandatory field. It wouldn't help pre-existing custom networks, but at least it would help prevent this from getting worse.

The risk of user-error is still there though. We could probably do more to warn users of the consequences of getting this wrong, and check against the `chainid.network` list for mismatches. But I don't think we can ever be certain.

adonesky1:
@Gudahtt when you see check against `chainid.network`do you mean making a request to `https://chainid.network/chains.json` and checking against that?

Gudahtt:
Yes, exactly. We already do this as part of the "Add Ethereum Chain" flow:
https://github.com/MetaMask/metamask-extension/blob/09d81ac5f2c94d7730d2e981b4f05734858030bf/ui/pages/confirmation/templates/add-ethereum-chain.js#L88

Except in this case, we'd only need to check that one field. I guess we'd find an entry with a matching `chainId`, and ensure the `nativeCurrency` matches the `ticker` value.

segun (approved):
LGTM!

adonesky1:
Very stale, going to close.

PR #12234, opened 2021-09-28
Connect ledger via webhid if that option is available

danjm:
This PR adds a permission request to connect the extension to a ledger hardware wallet device, if that option is available in the current browser.

Associated PRs: https://github.com/MetaMask/eth-ledger-bridge-keyring/pull/106 and https://github.com/MetaMask/eth-ledger-bridge-keyring/pull/107

This PR should also update the version of `eth-ledger-bridge-keyring` once PR 107 is merged and a new version of that module is released.

PR #12411, opened 2021-10-19
Connect Ledger via WebHID

danjm:
This PR improves upon and replaces https://github.com/MetaMask/metamask-extension/pull/12234

The purpose of this PR is to enable users to connect their Ledger devices to metamask via WebHID, if that option is available in the current browser.

When the user goes through the hardware wallet connect flow, by default they will be presented with a browser popup that asks them permission to connect their Ledger device.

In addition to that, other changes/improvements are:

- the settings toggle for ledger live has been changed to a dropdown menu, allowing users to select up to 3 options (ledger live, u2f and webhid)
- if webhid is selected in the aforementioned dropdown, and the ledger device is not connected, then a request to connect the device via webhid is made at that time
- if, in the hardware wallet connection flow, the user cancel's a connection request, an appropriate error message is shown
- every time a user goes to a transaction or message signing confirmation screen, if they have a ledger wallet and have selected the webhid transport type preference, then a check is made whether the ledger device is connect via webhid. If it is not connected, then the user is shown text prompting them to connect
- the confirm screen ledger instructions are now shown on all transaction screens, including erc20 approval confirmations, and message signature confirmations
- note that the webhid connection popup is too big for our notification window and cannot be resized by us, so we have to open the app in a full tab when the user hits the confirmation screen in a notification window and webhid is not connected

Below is a demo video that shows:
- connecting with webhid and successfully sending a transaction
- closing metamask, disconnecting webhib via browser settings, then opening metamask and seeing the connection prompt during the confirmation flow
- the webhid connection prompt/flow when signing a message proposed by a dapp
- fully closing the browser and then reopening and then sending a transaction without having to connect again

https://user-images.githubusercontent.com/7499938/138000820-3a87e350-28cd-4729-bb39-284ef8a88870.mp4


Gudahtt (commented):
>the confirm screen ledger instructions are now shown on all transaction screens, including erc20 approval confirmations, and message signature confirmations

It looks like it's still missing from `signTypedData_v1`, `eth_sign`, and `personal_sign` confirmations (i.e. the `signature-request-original` component).

Edit: This has been addressed

darkwing:
Tiny feedback:  this link is too big:
BigLink

Functionally, however, everything works as expected!  Well done!

Gudahtt (dismissed):
This looks great! Adding the notice to each of the confirmations wasn't as challenging as I had feared it would be. And as a bonus, we're adding the informational text in a few places it was missing as well. Huge improvement.

The design could use some improvements. I agree with David's most recent comment, and also the button in the ledger instruction field for reconnecting to the device is also too large (perhaps you meant to call attention to it over the other bullets? Which would make sense, but I think we can do that in a way that looks less out of place). I don't want to block on that, but a design pass would improve things.

Gudahtt (approved):
LGTM!

darkwing (approved):
This is excellent!  The only bit of unexpected behavior I saw was that I was being prompted each time to connect via WebHID, but I wasn't seeing it before, so it could be browser weirdness.

PR #12414, opened 2021-10-20
Removing unneeded build sections from README.md

ryanml:


rekmarks (approved):
💀

Gudahtt (approved):
LGTM, good catch! Didn't even notice those sections.

PR #12419, opened 2021-10-20
Removing unused prop/code paths from EditContact

ryanml:
The `showingMyAccounts` prop became obsolete as of: https://github.com/MetaMask/metamask-extension/pull/10680 - it's presence was producing the following warning while editing a contact: 
Screen Shot 2021-10-20 at 1 37 36 PM

`setAccountLabel` also became unneeded as a result of this removal


Gudahtt (commented):
LGTM! Except for the unit test failure :thinking:

Gudahtt (approved):

darkwing (approved):
I wonder if there's a tool out there that can detect this type of dead `prop` stuff.

PR #12420, opened 2021-10-20
fix missing conversion rates in swaps token drop down

adonesky1:
Explanation:  Conversion rates were missing in the swaps "from token" searchable dropdown because contractExchangeRates are keyed with addresses in checksum format.  This is yet another case where [address format normalization needs to take place globally](https://github.com/MetaMask/metamask-extension/issues/7985)

additionally the defaultToken (ETH) appeared to only render conversionRates when `swapsTokens` were unavailable so this PR alters logic for deriving the tokensToSearchBuckets such that we guarantee the defaultToken/native asset has data required to show fiat conversion available.

### DIFF
--- a/ui/hooks/useTokensToSearch.js
+++ b/ui/hooks/useTokensToSearch.js
@@ -16,6 +16,7 @@ import { getConversionRate } from '../ducks/metamask/metamask';
 
 import { getSwapsTokens } from '../ducks/swaps/swaps';
 import { isSwapsDefaultTokenSymbol } from '../../shared/modules/swaps.utils';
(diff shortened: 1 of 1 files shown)


darkwing (dismissed):

dan437 (approved):
I've tested the changes locally and I can see fiat prices now + tokens are sorted by their fiat value. Great job!

darkwing (approved):
My only nit is that we run `toChecksumHexAddress(address)` twice, maybe we could store that value in a `const` instead of doing the work twice, but not a huge deal.

jpuri (approved):

Issue #10741, opened 2021-03-26
Update currency conversion is incorrectly showing the current time.


tmashuang:
**Describe the bug**
The updated time for the currency conversion in settings is incorrectly always showing the current time.

**Steps to reproduce (REQUIRED)**
I blocked the specific network request within the developer console, but this can be achieved by going to the `Currency Conversion` setting and turning off the PC wifi. After waiting a about minute for the `Updated` text to update to the current time even though all network requests have been stopped.

**Expected behavior**
The last `conversionDate` should be applied to the text correctly.

danjm:
Note that recreating this requires explicitly blocking https://min-api.cryptocompare.com/ via the network tab in dev tools.

The "Updated" date should only show the date of the last successful update of the currency conversion rate.

PR #12422, opened 2021-10-21
Latest update conversion date  - issue  #10741

igorms-cons:
Fixes: 
(partial fix): see explanation section  below

- Pass good convert date  object to i18n (unix milliseconds)
- In offline mode, a message or the last conversion date is displayed

Explanation:   
The conversion date timestamp from the state was correct.  The date object passing through i18n helper was not converted correctly.  Thus,  the **_convertDate_** state is **always** sent by the library (even in offline mode). The library function is **_updateExchangeRate_**   and the value is   _**Date.now()**_  

Issue:  https://github.com/MetaMask/metamask-extension/issues/10741

Related to controllers PR: https://github.com/MetaMask/controllers/pull/621
**It is required to have this version of controller to get the expected behaviour**

### DIFF
--- a/ui/pages/settings/settings.component.js
+++ b/ui/pages/settings/settings.component.js
@@ -41,11 +41,31 @@ class SettingsPage extends PureComponent {
     initialBreadCrumbKey: PropTypes.string,
     mostRecentOverviewPage: PropTypes.string.isRequired,
     addNewNetwork: PropTypes.bool,
+    conversionDate: PropTypes.number,
   };
 
   static contextTypes = {
     t: PropTypes.func,
   };
+
+  state = {
+    lastFetchedConversionDate: null,
+  };
+
+  componentDidMount() {
+    this.handleConversionDate();
+  }
+
+  componentDidUpdate() {
+    this.handleConversionDate();
+  }
+
+  handleConversionDate() {
+    const { conversionDate } = this.props;
+    if (conversionDate !== null) {
+      this.setState({ lastFetchedConversionDate: conversionDate });
+    }
+  }
 
   render() {
     const {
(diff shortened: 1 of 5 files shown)


igorms-cons:
PR behaviour 
https://user-images.githubusercontent.com/89637087/139861684-8d8b8ea7-f2db-410c-98b0-88d700e75b9a.mov

The date is set on the settings page for now, not in state application. It implies that the date will be reset if we start from from home page.  A message is displayed (in offline mode).

mhasanagic:
Verified by QA - the behavior is as described

david0xd (dismissed):

ui/pages/settings/settings.component.js
```diff
@@ -41,12 +41,32 @@ class SettingsPage extends PureComponent {
...
+
+  componentDidMount() {
+    this.handleConversionDate();
+  }
+
+  componentDidUpdate() {
+    this.handleConversionDate();
+  }
+
+  handleConversionDate() {
```
danjm, line 63: Does this have to happen in this component? Could it just be in `setttings-tab.container.js`?

danjm: oh, I see, we are using client side state to cache the `lastFetchedConversionDate`

igorms-cons: Hello @danjm ,

You are totally right, my first implementation was directly in setttings-tab.container.js. I did it in settings because i wanted the `date` state to be available in all settings tabs. 
Like i said in my above explanations, - in offline mode -  if we leave _settings_  and come back, date state will be lost & message displayed.  Thats normal because `date` is obviously not in appstate. I dont know if we want to put it there.

Let me know which implementation is most relevant. Thx

danjm:
@igorms-cons This is now just blocked by an update to the controllers package. We probably have to handle that in another PR though, as it has to account for some breaking changes

igorms-cons:
@danjm  yep no pb, the PR https://github.com/MetaMask/controllers/pull/621 has been merged yesterday

adonesky1:
Hi @david0xd I have another [PR open](https://github.com/MetaMask/metamask-extension/pull/12443) to integrate the new controller version. We are each addressing different breaking changes separately here. @danjm Can we just merge them sequentially or should we merge our PRs somehow?

igorms-cons:
@danjm  I saw that @adonesky1  [PR](https://github.com/MetaMask/metamask-extension/pull/12443) is merged now (controllers).  How do you want me to proceed from now?

danjm:
@igorms-cons Just rebase this onto `develop` and it should be good to merge

igorms-cons:
@danjm  perfect, thx!

mhasanagic:
Re-verified by QA

david0xd (approved):

danjm (approved):

darkwing (approved):

PR #12424, opened 2021-10-21
Organize Storybook Components

### LINKED ISSUE #12551: Move "Getting Started" section to top in Storybook
**Describe the bug**
Getting Started section is at the bottom of the menu in storybook. It should be at the top.

1. Go to [Metamask storybook](https://metamask.github.io/metamask-storybook/index.html)
2. Scroll down menu to see "Getting Started" section at bottom

**Expected behavior**
Getting Started section should be at the top of the menu in storybook

We can use `storySort` to sort stories https://storybook.js.org/docs/react/writing-stories/naming-components-and-hierarchy#sorting-stories

**Screenshots**
(Screen shot taken from #12424)

Screen Shot 2021-11-01 at 9 30 10 AM

EtDu:
This PR aims to classify and visually organise all storybook components.

Each component is sorted exactly how it appears in the project.

fixes #12551


georgewrmarshall (changes requested):
This is looking good! I like the use of `'App/'`, `'UI/` and `'Pages'/` groupings. I think nesting any further may reduce discoverability? Also some of the groupings do not make complete sense to me e.g. `'UI/Modal/Actionable Message'`

I would suggest at least for this first run only having one level of groupings reflect the location of the stories in the app. This will help:

- Find stories more easily
- Understand how the code is organized 

Could we remove any nesting of groups further than `'App/'`, `'UI/` and `'Pages'/`?

So it would look something like

```
title: 'FOLDER_NAME/COMPONENT_NAME'
```

At least for this addition. We may be able to identify further nested groupings in future but for now I think this change will already make a huge difference!

brad-decker (approved):

georgewrmarshall:
@EtDu also if we could fix this warning about duplicate story titles

Screen Shot 2021-10-26 at 11 28 54 AM

jj-crypto (approved):

jj-crypto (approved):

ui/pages/settings/settings.stories.js
```diff
@@ -19,7 +19,7 @@ import {
 import SettingsPage from './settings.component';
 
 export default {
-  title: 'SettingPage',
+  title: 'Pages/Send/Settings Page',
```
georgewrmarshall, line 22: Should this be 
```suggestion
  title: 'Pages/Settings',
```

georgewrmarshall (commented):
This looks excellent. Nice work! 💯  Just flagged one page but everything else looks great!

georgewrmarshall:
I've also created this issue #12551 if you want to include it in this PR? Also fine as separate.

EtDu:
> I've also created this issue #12551 if you want to include it in this PR? Also fine as separate.

good to go

georgewrmarshall (approved):
LGTM! Nice work 💯

georgewrmarshall:
To fix your lint issue just change `Base` story in `toggle-button.stories.js` to `DefaultStory`. Sorry I forgot I updated that one https://github.com/MetaMask/metamask-extension/blob/develop/ui/components/ui/toggle-button/toggle-button.stories.js#L13

PR #12431, opened 2021-10-21
Add currency symbol validation in the add network form

adonesky1:
- Adds (non-blocking) warning when ticker symbol doesn't match what's found on `https://chainid.network/chains.json`
- Refactors `networks-form.js` validation
- Makes `Currency Symbol` a required field in the Add Network Form
- Updates network form error verbiage to match new figma designs 👇 

[Figma designs](https://www.figma.com/file/EUQ1LPxKPU9tga4YnjISn1/Network-specific-assets?node-id=1407%3A20916)

**Further Context:** This work was originally conceived and discussed in PR #12410

**Steps to Manually Test**
 1. Go to add a network form
 2. Add a valid RPC URL
 3. Add the chainId (if you don't enter the correct on you should see an error indicating the correct one for this RPC endpoint)
 4. See that the save button is still not enabled because currency symbol is not yet complete
 5. Enter any string in the currency symbol field. 
 6. See that a warning below that field pops indicating the expected symbol values for this chain Id.

**Demo:**
(scenario user sees warning but can elect to ignore and proceed to add the network) 
https://user-images.githubusercontent.com/34557516/152047015-c400cb03-2000-44d2-9624-a43b1544de24.mov

Related work to do after this PR:
- [ ] Add prompt for user to add ticker symbol to networks they had previously added without one.
- [ ] Add more tests of new validation logic

### DIFF
--- a/ui/pages/settings/networks-tab/networks-form/networks-form.js
+++ b/ui/pages/settings/networks-tab/networks-form/networks-form.js
@@ -374,9 +374,13 @@ const NetworksForm = ({
   const validateUrlRpcUrl = useCallback(
     (url) => {
       const isValidUrl = validUrl.isWebUri(url);
-      const [{ rpcUrl: matchingRPCUrl = null } = {}] = networksToRender.filter(
-        (e) => e.rpcUrl === url,
-      );
+      const [
+        {
+          rpcUrl: matchingRPCUrl = null,
+          label: matchingRPCLabel,
+          labelKey: matchingRPCLabelKey,
+        } = {},
+      ] = networksToRender.filter((e) => e.rpcUrl === url);
       const { rpcUrl: selectedNetworkRpcUrl } = selectedNetwork;
 
       if (!isValidUrl && url !== '') {
@@ -398,7 +402,7 @@ const NetworksForm = ({
         return {
           key: 'urlExistsErrorMsg',
           msg: t('urlExistsErrorMsg', [
-            matchingRPCUrl.label ?? matchingRPCUrl.labelKey,
+            matchingRPCLabel ?? matchingRPCLabelKey,
           ]),
         };
       }
--- a/ui/pages/settings/networks-tab/networks-form/networks-form.js
+++ b/ui/pages/settings/networks-tab/networks-form/networks-form.js
@@ -5,6 +5,7 @@ import PropTypes from 'prop-types';
 import validUrl from 'valid-url';
 import log from 'loglevel';
 import classnames from 'classnames';
+import { debounce } from 'lodash';
 import { useI18nContext } from '../../../../hooks/useI18nContext';
 import {
   isPrefixedFormattedHexString,
@@ -26,6 +27,7 @@ import {
   DEFAULT_ROUTE,
   NETWORKS_ROUTE,
 } from '../../../../helpers/constants/routes';
+import fetchWithCache from '../../../../helpers/utils/fetch-with-cache';
 
 /**
  * Attempts to convert the given chainId to a decimal string, for display
@@ -83,6 +85,7 @@ const NetworksForm = ({
     selectedNetwork?.blockExplorerUrl || '',
   );
   const [errors, setErrors] = useState({});
+  const [warnings, setWarnings] = useState({});
   const [isSubmitting, setIsSubmitting] = useState(false);
 
   const resetForm = useCallback(() => {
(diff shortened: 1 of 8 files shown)


ui/pages/settings/networks-tab/networks-form/networks-form.js
```diff
@@ -488,39 +571,29 @@ const NetworksForm = ({
...
         <FormField
           error={errors.blockExplorerUrl?.msg || ''}
-          onChange={(value) => {
-            setBlockExplorerUrl(value);
-            validateBlockExplorerURL(value);
-          }}
+          onChange={setBlockExplorerUrl}
           titleText={t('blockExplorerUrl')}
           titleUnit={t('optionalWithParanthesis')}
           value={blockExplorerUrl}
```
jpuri, line 599: @adonesky1 : this file is getting too big, it will be useful to extract validation logic into a separate file and have more unit test coverage.

adonesky1: One problem with this is that the validation methods require the context from `useI18nContext` and it feels wrong to pass it as an argument? And I think we agreed that we shouldn't pass keys to `t()` (from`useI18nContext`) as variables. I suppose I could make a class that holds these methods that instantiates `useI18nContext`, and is itself instantiated in the networks-form component but that isn't a pattern I think we use anywhere else so I'm hesitant. Do you have thoughts on how we should create helper files that need context like `useI18nContext` without passing it in as an argument? cc @jpuri @brad-decker @danjm

adonesky1: I remember @brad-decker also raising whether we should deprecate the recently introduced `<I18nValue>` what was the result of that discussion?

adonesky1: @jpuri added more unit testing here: https://github.com/MetaMask/metamask-extension/pull/12431/commits/ba7b38dbf276a2bfa550fcd5907e5c06deca9dfd

jpuri (dismissed):

brad-decker (dismissed):
LGTM, love the added test coverage

ui/pages/settings/networks-tab/networks-form/networks-form.test.js
```diff
@@ -1,10 +1,42 @@
...
 import configureMockStore from 'redux-mock-store';
-import { fireEvent } from '@testing-library/react';
+import { fireEvent, screen } from '@testing-library/react';
+import nock from 'nock';
 import { renderWithProvider } from '../../../../../test/jest/rendering';
 import { defaultNetworksData } from '../networks-tab.constants';
+import { MAINNET_RPC_URL } from '../../../../../shared/constants/network';
 import NetworksForm from '.';
 
+nock('https://chainid.network:443', { encodedQueryParams: true })
```
mcmire, line 10: This will create a global mock across all tests. I'm afraid this will lead to intermittently failing tests. Should this go within a `beforeEach`, and does it makes sense to call `nock.cleanAll()` within an `afterEach`?

mcmire (changes requested):

ui/pages/settings/networks-tab/networks-form/networks-form.test.js
adonesky1: Yeah good call out!

ui/pages/settings/networks-tab/networks-form/networks-form.test.js
```diff
@@ -68,6 +38,50 @@ const propNetworkDisplay = {
 };
 
 describe('NetworkForm Component', () => {
+  beforeAll(() => {
+    nock.disableNetConnect();
```
mcmire, line 42: Ah yeah I forgot about this!

mcmire (approved):
Thanks!

ui/pages/settings/networks-tab/networks-form/networks-form.test.js
adonesky1: done here @mcmire: https://github.com/MetaMask/metamask-extension/pull/12431/commits/8944afe5c5b3057550529316cd420b33775e61da

adonesky1:
@jpuri @brad-decker just address a small piece of feedback from @mcmire here: https://github.com/MetaMask/metamask-extension/pull/12431/commits/8944afe5c5b3057550529316cd420b33775e61da

brad-decker (approved):

NiranjanaBinoy (approved):

PR #12432, opened 2021-10-22
[Feature Proposal] Allow users to enable Allowlist mode and set allowed hosts in settings

timothycarambat:
### Proposal:

Currently, there has been a large influx of spam and phishing attacks taking place on the Ethereum network. These attacks come in many ways, but the most common is for the attacker to make a carbon copy of a website people are known to interact with and perform a [IDN_homograph_attack](https://en.wikipedia.org/wiki/IDN_homograph_attack).

While MetaMask does provide retroactive protection to its users via the Phishing alert service, we should enable users to whitelist known-good URLs so that in the event they erroneously click on a bad domain - they are not compromising their wallet.

This feature should be non-blocking but provide some indication clearly in the interface that a user is about to interact with a site they have not approved - if this feature is enabled. By default - this setting is not enabled for ease of use.

### Explanation:  

Under `User Settings > Security and Privacy` there is now a new toggle switch + input field.
- This toggle is labeled `Allowlist Mode` and when enabled will prompt users during connection and contract interaction that the hostname they are interacting with is not in the list they can define.
- The input field is a simple text input of comma-separated hostnames without protocols.
- This whitelist filter **does** work with subdomains of the hostnames defined. eg: allowlisting `metamask.io` will also allow `docs.metamask.io`

Allowlist toggle + input field under Settings:
Screen Shot 2021-10-21 at 8 54 19 PM

Allowlist toggle enabled with input defined:
Screen Shot 2021-10-21 at 8 54 53 PM

Allowlist warning on connection:
Screen Shot 2021-10-21 at 8 56 18 PM

Screen Shot 2021-10-21 at 8 57 05 PM

Allowlist warning on transaction interaction:
Screen Shot 2021-10-21 at 8 57 59 PM


timothycarambat:

jj-crypto (approved):

jj-crypto (dismissed):

jj-crypto (approved):

danfinlay:
It's very cool that you were able to implement this feature idea so quickly, but we need to limit merged features to things we expect to have an impact for most users, since we will be taking on any maintenance burden going forward, and we aren't convinced the average user is going to dig into settings to learn to benefit from this, but we are looking at providing similar security benefits in ways that should benefit more users by default.

Closing, but thanks for offering the contribution.

Issue #12342, opened 2021-10-13
Hide top, large eth amount for contract transactions that are not sending any eth


danjm:
Create a transaction that interacts with a smart contract, but does not send any eth. For example, go to https://metamask.github.io/test-dapp/ and click the create token button. When you do so you will see:

![Screenshot from 2021-10-13 05-48-15](https://user-images.githubusercontent.com/7499938/137095027-4f3b8e81-57df-4fdd-b62c-8f3491bafe45.png)

In cases where the user is creating or sending to a contract, but not sending any ETH, hide the large black text that says "0 ETH" at top.

PR #12435, opened 2021-10-22
Hide top, large eth amount for contract transactions that are not sending any eth #12342

dragana8:
Fixes: Removed top eth amount 
https://github.com/MetaMask/metamask-extension/issues/12342
Screenshot from 2021-10-22 12-15-13

### DIFF
--- a/ui/components/app/confirm-page-container/confirm-page-container-content/confirm-page-container-summary/confirm-page-container-summary.component.js
+++ b/ui/components/app/confirm-page-container/confirm-page-container-content/confirm-page-container-summary/confirm-page-container-summary.component.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-negated-condition */
 import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
@@ -14,6 +15,7 @@ const ConfirmPageContainerSummary = (props) => {
     identiconAddress,
     nonce,
     origin,
+    hideTitle,
   } = props;
 
   return (
@@ -37,9 +39,11 @@ const ConfirmPageContainerSummary = (props) => {
             address={identiconAddress}
           />
         )}
-        <div className="confirm-page-container-summary__title-text">
-          {titleComponent || title}
-        </div>
+        {!hideTitle ? (
+          <div className="confirm-page-container-summary__title-text">
+            {titleComponent || title}
+          </div>
+        ) : null}
       </div>
       {hideSubtitle || (
         <div className="confirm-page-container-summary__subtitle">
@@ -60,6 +64,7 @@ ConfirmPageContainerSummary.propTypes = {
   identiconAddress: PropTypes.string,
   nonce: PropTypes.string,
   origin: PropTypes.string.isRequired,
+  hideTitle: PropTypes.boolean,
 };
 
 export default ConfirmPageContainerSummary;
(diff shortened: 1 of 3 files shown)


david0xd (dismissed):

ui/components/app/confirm-page-container/confirm-page-container-content/confirm-page-container-summary/confirm-page-container-summary.component.js
```diff
@@ -37,9 +50,11 @@ const ConfirmPageContainerSummary = (props) => {
             address={identiconAddress}
           />
         )}
-        <div className="confirm-page-container-summary__title-text">
-          {titleComponent || title}
-        </div>
+        {hideTitle || (
```
danjm, line 53: This is a little different from our convention for conditionally rendering components.

There are a few patterns you can follow:
- `booleanVariable && <Component />`, so that `<Component />` will render if `booleanVariable` is true
- `booleanVariable ? <Component /> : null`, so that `<Component />` will render if `booleanVariable` is true
- `componentVariable1 || somethingElse`, so that a component stored at `componentVariable1` will be rendered if defined or `somethingElse` will be render if `componentVariable1` is not defined

The code here is a little different, where the component only renders if the boolean is false. This will work, but is a new pattern. Sorry that we don't have good documentation on this.

I would recommend the following:

```
{!hideTitle
    ? (<div className="confirm-page-container-summary__title-text">
            {titleComponent || title}
          </div>
        )}
    : null
}

File: ui/components/app/confirm-page-container/confirm-page-container-content/confirm-page-container-summary/confirm-page-container-summary.component.js [hunk 2]
```diff
@@ -14,8 +15,20 @@ const ConfirmPageContainerSummary = (props) => {
     identiconAddress,
     nonce,
     origin,
+    type,
   } = props;
 
+  let hideTitle = false;
```
danjm, line 21: Small stylistic comments on this section of code. We try to avoid an `if` block when it is simple and straightforward to do so. If this case, `hideTitle` can just be a `const` set to the logicial statement inside the `if ()` brackets.

Also, we try to use the [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) when it can simplify code.

So this section of code could be simplified to

`const hideTitle = (type === TRANSACTION_TYPES.CONTRACT_INTERACTION || type === TRANSACTION_TYPES.DEPLOY_CONTRACT) && titleComponent.props?.value === '0x0'

danjm: I also think it might be best if this logic is higher up in the component tree, and we pass `hideTitle` down as a prop. Ideally this component would be free of logic related to state. This wasn't the case even before this PR, but it is a direction we want to move.

You could use `currentTransaction` in `ui/components/app/confirm-page-container/confirm-page-container.component.js` to get the value and put this logic in that component. Then pass `hideTitle` down

danjm (changes requested):
Nice work on this. I just had a few suggestions related to code organization and style

ui/components/app/confirm-page-container/confirm-page-container-content/confirm-page-container-summary/confirm-page-container-summary.component.js
dragana8: @danjm updated

ui/components/app/confirm-page-container/confirm-page-container-content/confirm-page-container-summary/confirm-page-container-summary.component.js
dragana8: @danjm  Missed this last comment so I am adding one more commit with appropriate changes.

danjm (commented):
Looks good. Approved.

danjm (dismissed):
approving again, but needs a rebase

danjm (approved):

ryanml (approved):

Issue #11420, opened 2021-06-29
Unable to send ETH with decimal values of more than 6 places


chirag64:
**Describe the bug**
When sending ETH to an address, Metamask tries to round up values after 6 decimal places.

**Steps to reproduce (REQUIRED)**

1. Go to Metamask dashboard when logged in
2. Click on `Send` for ETH
3. Enter any target address to send ETH to
4. In the `Amount` field, try typing the value `1.2345678`. Notice how the value has removed the `7` and is now `1.234568`.
5. Click on 'Next'. Notice how the value is still `1.234568`.

**Expected behavior**
After typing the value `1.2345678`, we would expect to see the same value in the input field as well as on the Confirmation screen when we click Next

**Screenshots**
![image](https://user-images.githubusercontent.com/591424/123850849-22ca2080-d938-11eb-9fbe-cc6c31353ac1.png)

**Browser details (please complete the following information):**
 - OS: Windows
 - Browser: Chrome Canary Version 93.0.4558.0 (Official Build) (64-bit)
 - MetaMask Version: Dev build from lavamoat-dev-optional branch

EHaracic:
**@danjm how many decimal places should we use, 8 places or  more?** 
this issue  is related to the issue: #10167

bschorchit:
Closing this as it has been fixed with: https://github.com/MetaMask/metamask-extension/pull/12437

PR #12437, opened 2021-10-22
Changed ETH decimal values to have 8 places for sending

VSaric:
**Fixes:** #11420

**Explanation for screenshot number one:**  When user want to send ETH to an address, user can insert 8 decimal places and Metamask round up values after 8 decimal places.

**Screenshot number one:**  
Screenshot from 2021-10-22 14-08-35

**Explanation for screenshot number two:** After typing the value 1.23456789, we would expect to see the same value in the input field as well as on the Confirmation screen when we click Next.

**Screenshot number two:** 
Screenshot from 2021-10-22 14-09-33


david0xd (approved):

danjm (approved):
LGTM

NiranjanaBinoy (approved):

ryanml (approved):

PR #12438, opened 2021-10-22
Use separate Infura project ID for production

Gudahtt:
We now use two separate Infura project IDs for production builds, and for all other builds. Previously all CI builds used the production Infura project ID. Separating them will make our Infura dashboard metrics more representative of real production usage.

The new environment variable for production has been setup in CI already, but the old environment variable will remain set to the
production project ID until this commit is included in a release. We can't switch the old environment variable out until we're confident that it won't get used for a production build.

We now use constants for the various different build environments. This was done to improve the JSDoc types of the `getInfuraProjectId` helper method.

The `getConfigValue` function was added to make it easier to validate that required config values are set. This should ensure builds fail early with an informative error message when they are missing the necessary configuration.

### DIFF
--- a/development/build/scripts.js
+++ b/development/build/scripts.js
@@ -83,7 +83,7 @@ function getConfigValue(key) {
  * Get the appropriate Infura project ID.
  *
  * @param {object} options - The Infura project ID options.
- * @param {keyof ENVIRONMENT} options.environment - The build environment.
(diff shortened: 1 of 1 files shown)


adonesky1 (dismissed):
LGTM!

darkwing (dismissed):

adonesky1 (approved):

danjm (approved):
LGTM

PR #12440, opened 2021-10-22
Add secrets for beta and Flask builds

### LINKED ISSUE #11896: Configure API keys for Flask
We need to take an inventory of third-party APIs that should have separate keys for Flask. These include:
- [ ] Infura
- [ ] Sentry
- [ ] Segment / Mixpanel

Gudahtt:
The build script has been updated to embed the correct Infura project ID and Segment write key for beta and Flask builds. These are set via environment variable or config file. They have already been added in CI as environment variables.

The Segment production write key has also been moved into the set of environment variables that can be set in the configuration file. This was to make the way we reference it more consistent.

The new project IDs and keys are only used in the "production" environment, which right now is the merge step into the `master` branch. This is appropriate for Flask, but it doesn't match our plan for how the beta release would get created. In a future PR, when the beta release automation work is completed, the conditions for when the beta secrets are used should be updated to ensure they're used only for the beta builds.

Closes #11896

### DIFF
--- a/development/build/scripts.js
+++ b/development/build/scripts.js
@@ -29,10 +29,15 @@ const bifyModuleGroups = require('bify-module-groups');
 
 const metamaskrc = require('rc')('metamask', {
   INFURA_PROJECT_ID: process.env.INFURA_PROJECT_ID,
+  INFURA_BETA_PROJECT_ID: process.env.INFURA_BETA_PROJECT_ID,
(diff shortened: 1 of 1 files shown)


rekmarks (dismissed):
LGTM!

rekmarks (approved):

darkwing (approved):

PR #12441, opened 2021-10-22
Add build type to Sentry environment

Gudahtt:
The build type (i.e. the distribution) is now included in the Sentry environment during setup, for all builds except the "main" build. This will allow us to track Flask and beta errors separately from other errors.

A constant was created for the build types. The equivalent constant in our build scripts was updated to match it more closely, for consistency. We can't use the same constant in both places because our shared constants are in modules that use ES6 exports, and our build script does not yet support ES6 exports.

The singular `BuildType` was used rather than `BuildTypes` to match our naming conventions elsewhere for enums. We name them like classes or types, rather than like a collection.

Relates to #11896

### DIFF
--- a/app/scripts/lib/setupSentry.js
+++ b/app/scripts/lib/setupSentry.js
@@ -1,6 +1,7 @@
 import * as Sentry from '@sentry/browser';
 import { Dedupe, ExtraErrorData } from '@sentry/integrations';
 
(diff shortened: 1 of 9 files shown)


danjm (approved):
LGTM

ryanml (approved):

jpuri (approved):
👍

PR #12443, opened 2021-10-22
Add CollectiblesController & CollectibleDetectionController

adonesky1:
Explanation:  
- Adds CollectiblesController (and its dependency, AssetsContractController) to the extension. Adds bridge methods in action.js to interface with CollectiblesController.
- Adds CollectibleDetectionController
  
Next Steps:
- [ ] Determine when detection polling should start and stop.


ui/store/actions.js
```diff
@@ -1283,6 +1283,79 @@ export function addToken(
   };
 }
 
+export function addCollectible(address, tokenID, dontShowLoadingIndicator) {
```
darkwing, line 1286: Minor - should we make this `showLoadingIndicator = true`?  I prefer arguments for _taking_ action, not _preventing_ action?  Just a discussion point, not a blocker.

adonesky1: Sure I don't mind changing this. We do use it [elsewhere in this file](https://github.com/MetaMask/metamask-extension/blob/4a90e5bc84567f4a3465fcd33b2f76f79a9e1c83/ui/store/actions.js#L675). But generally I agree with the pattern you're advocating.

adonesky1: It doesn't seem that we destructure arguments anywhere in these action methods... I can't think of a reason why we wouldn't, but without doing so we can't set a default value. 🤔 @darkwing

ryanml (dismissed):
LGTM! thanks for the changes here

danjm (dismissed):
LGTM

adonesky1:
@danjm I've realized that there are still some ways that the OpenSea API will get hit (we attempt to detect collectibles when a user changes accounts) unless we put this behind a feature flag.

danjm (approved):

ryanml (approved):

PR #12446, opened 2021-10-22
Update NumericInput to remove PropTypes warning

mcmire:
Use oneOfType, not oneOf, to specify a union type.


ryanml (approved):

Gudahtt (approved):
LGTM! Though I agree with @darkwing that it would be better to not accept both types here.

darkwing (approved):

PR #12453, opened 2021-10-22
Standardize the use of zero value into a constant

darkwing:
Instead of hardcoding `0x0` everywhere for zero value, we can use a constant.  A different value should be used for `0x0` chain ID, and we should create one constant for a fake address in another PR.


darkwing:
I'm going to close and redo this PR as it's woefully out of date.

PR #12456, opened 2021-10-23
Improve browser detection code

darkwing:
This code looked legacy so I updated with modern standards.


Gudahtt (dismissed):
LGTM! This looks equivalent.

Though it'd be nice if we had unit tests for this as well. There is a `util.test.js` file, but this function isn't covered.

jj-crypto (dismissed):

danjm (dismissed):
LGTM

+1 for it would be nice to have unit tests

jj-crypto (dismissed):

brad-decker (approved):

kumavis (approved):

PR #12459, opened 2021-10-23
Rename the function of the Ledger preference renaming

darkwing:
Since the Ledger transport isn't LedgerLive-centric, we should update our method names.


segun (dismissed):
LGTM

kumavis (approved):

danjm (approved):

PR #12460, opened 2021-10-23
Use gas estimate type constants

darkwing:


ryanml (dismissed):

danjm (dismissed):

ryanml (dismissed):

kumavis (dismissed):

app/scripts/controllers/transactions/index.js
```diff
@@ -1445,8 +1445,8 @@ export default class TransactionController extends EventEmitter {
       sensitiveProperties: {
         status,
         transaction_envelope_type: isEIP1559Transaction(txMeta)
-          ? 'fee-market'
-          : 'legacy',
+          ? GAS_ESTIMATE_TYPES.FEE_MARKET
```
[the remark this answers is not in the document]
kumavis: agree that would be an improvement but doesnt need to happen in this PR

ryanml (approved):

ui/helpers/constants/transactions.js
```diff
@@ -19,3 +19,8 @@ export const TOKEN_CATEGORY_HASH = {
   [TRANSACTION_TYPES.TOKEN_METHOD_TRANSFER]: true,
   [TRANSACTION_TYPES.TOKEN_METHOD_TRANSFER_FROM]: true,
 };
+
+export const TRANSACTION_ENVELOPE_TYPE_NAMES = {
```
mcmire, line 23: Is it worth it, do you think, to put this in `shared/constants/transaction.js` after `TRANSACTION_ENVELOPE_TYPES`, since they're related to each other?

darkwing: The code would then be minified into both the background and the client side; since it's not used in the background, this seems like the right place.

mcmire: Ah, yeah, that's a good point 👍🏻

mcmire (approved):
LGTM.

PR #12461, opened 2021-10-23
Use gas recommendation constants throughout app

darkwing:
Changes out hardcoded values for constants.


ryanml (dismissed):

segun (dismissed):
I think we should change that to use the constant too.

ryanml (approved):

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
```diff
@@ -112,7 +117,8 @@ export default function EditGasPopover({
   );
 
   const txParamsHaveBeenCustomized =
-    estimateToUse === 'custom' || txParamsAreDappSuggested(updatedTransaction);
+    estimateToUse === CUSTOM_GAS_ESTIMATE ||
```
kumavis, line 120:
```suggestion
    (estimateToUse === CUSTOM_GAS_ESTIMATE) ||
```
more explicit order of operations

kumavis (approved):

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
darkwing: `yarn lint:fix` removes those parens, so I think we're good.

PR #12476, opened 2021-10-25
Add fixture server substitutions

Gudahtt:
The fixture server now supports state substitutions. This allows us to embed dynamic values in our fixtures.

The `custom-token` fixture has been updated to include such a fixture. The date that the seed phrase reminder was last shown has been updated to always be the current date, to prevent the reminder from showing up during e2e tests. This fixes the e2e test failure for the test "add-hide-token.spec.js" that we've been seeing on CI lately.

Manual testing steps:  
  - Try running the e2e test that had failed before this PR: `yarn build:test && yarn test:e2e:single --browser chrome --leave-running ./test/e2e/tests/add-hide-token.spec.js` and see that it passes now.

### DIFF
--- a/test/e2e/fixture-server.js
+++ b/test/e2e/fixture-server.js
@@ -1,12 +1,57 @@
 const { promises: fs } = require('fs');
 const path = require('path');
 const Koa = require('koa');
+const { isObject, mapValues } = require('lodash');
 
 const CURRENT_STATE_KEY = '__CURRENT__';
 const DEFAULT_STATE_KEY = '__DEFAULT__';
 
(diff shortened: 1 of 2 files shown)


danjm (approved):
LGTM to me.

As an aside, would it also be okay just to do a match of each `fixtureSubstitutionCommands` against the whole stringified rawState, and just replace those matches in the string? (As opposed to recursively searching and editing the parsed state object)

jpuri (approved):

NiranjanaBinoy (approved):

PR #12488, opened 2021-10-27
Fix production builds

Gudahtt:
The production build was accidentally broken in #12440 because of a merge conflict with a #12441 that wasn't initially noticed. The conflict was the renaming of the `BuildTypes` variable to `BuildType`.

This variable is used to check the current build type, but only for production builds. `BuildTypes` is `undefined`, so this would result in a crash when that enum was used.

Manual testing steps:  
  - Try creating a production build, and ensure the build finishes without crashing.
    - The command I used to test this is `CIRCLE_BRANCH='master' SENTRY_DSN='any string' INFURA_PROD_PROJECT_ID='any string' SEGMENT_PROD_WRITE_KEY='any string' yarn dist`
    - The `CIRCLE_BRANCH` environment variable identifies it as a production build, and the other environment variables are required to be non-empty by validation steps.

### DIFF
--- a/development/build/scripts.js
+++ b/development/build/scripts.js
@@ -55,7 +55,7 @@ const {
 const {
   createRemoveFencedCodeTransform,
 } = require('./transforms/remove-fenced-code');
-const { BuildTypes } = require('./utils');
+const { BuildType } = require('./utils');
 
 /**
  * The build environment. This describes the environment this build was produced in.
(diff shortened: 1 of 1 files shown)


darkwing (approved):

georgewrmarshall (approved):
lgtm

PR #12491, opened 2021-10-27
Fix currency preference and hex data state when in send screen edit mode

### LINKED ISSUE #12347: Fix currency preference and hex data state when in send screen edit mode
1. Turn on the "Show Hex Data" toggle in advanced settings
2. Go to the send flow
3. Enter an amount and then click the currency switcher
4. Enter some hex data
5. Click "Next"
6. On the Confirm screen click "Edit"
7. You will now be on the send screen, but the currency switcher setting will be back to the default and the hex data field will be empty

Instead, the currency preference should be what you switched it to originally, and the hex data field should contain the data you originall set.

Here is a video demonstrating this:

https://user-images.githubusercontent.com/7499938/137104306-e6aa1df3-f07b-4625-901b-19905e054587.mp4

filipsekulic:
Fixes: https://github.com/MetaMask/metamask-extension/issues/12347

When a user changes currency preference (from ETH which is default to USD), enter some amount, then enter hex data, go further and then go back, those changes will be saved (which was not the case before this fix). 

https://user-images.githubusercontent.com/92310504/139071402-2b2b3692-9131-4550-a7df-55d4f603a0c4.mp4

### DIFF
--- a/ui/components/app/user-preferenced-currency-input/user-preferenced-currency-input.component.js
+++ b/ui/components/app/user-preferenced-currency-input/user-preferenced-currency-input.component.js
@@ -6,21 +6,25 @@ export default class UserPreferencedCurrencyInput extends PureComponent {
   static propTypes = {
     useNativeCurrencyAsPrimaryCurrency: PropTypes.bool,
     passDataToSendAmount: PropTypes.func,
-    location: PropTypes.object
+    location: PropTypes.object,
   };
 
   state = {
-    dataFromCurrency: {}
-  }
+    dataFromCurrency: {},
+  };
 
   getDataFromCurrency = (value) => {
-    this.setState({ dataFromCurrency: value })
-  }
+    this.setState({ dataFromCurrency: value });
+  };
 
   render() {
-    const { useNativeCurrencyAsPrimaryCurrency, location, ...restProps } = this.props;
-    
-    this.props.passDataToSendAmount(this.state)
+    const {
+      useNativeCurrencyAsPrimaryCurrency,
+      location,
+      ...restProps
+    } = this.props;
+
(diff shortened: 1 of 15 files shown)


mhasanagic:
Verified by QA

david0xd (dismissed):

david0xd (approved):

danjm:
@filipsekulic Thanks for all the work you put into this.

Unfortunately, I don't think we can accept this PR. This PR uses the location object to manage state, but we try to only use location for things related to routing, deriving state from a route, or possible passing some sort of state when a new window has to be oopened.

Otherwise, state management that does not need to persist sessions should be handled in redux.

I can understand why the approach in the PR may have been taken, and I am to blame here. First of all, the currency preference and hex data issues should not have been grouped together. Me putting them in the same ticket probably indicated that I wanted them to be solved with the same solution. I should have provided more guidance in this regard, and indicated that although the bugs are very similar from a user perspective they still may require different technical solutions.

Second of all, `currency-display.component.js` (as it was written before this PR) is really convoluted and poorly architected. I believe I first implemented that about 4 years ago, and many poor decisions went into it. Trying to work with the state management of that component is really non-intuitive. I can't even understand it's behaviour fully today.

Thirdly, we still don't have great documentation on best practices for things like state management.

Instead of this PR, I suggest that the hex data and currency preference issues be handled separately.

I have created two new tickets which can replace this:
https://github.com/MetaMask/metamask-extension/issues/12601
and
https://github.com/MetaMask/metamask-extension/issues/12602

My apologies for the time you put into this, only to have me tell you that it needs an overhaul after the fact. I think, at least, that the time you spent on this will have given you decent understanding into the code surrounding this component, which will help with the tickets I mention above.

PR #12494, opened 2021-10-27
Don't show ledger live option in advanced settings if using firefox

danjm:
Addresses an issue found during QA of v10.4.0

Ledger Live doesn't work with firefox, so this option should not be selected in the advanced tab of settings when on firefox. This is how it works on prod.


NiranjanaBinoy (dismissed):
LGTM

ryanml (approved):
nit: only if you have to push again, can we group the constant and helper imports together in the top of the file?

darkwing (approved):

PR #12498, opened 2021-10-27
Improve token detection wording

danfinlay:
Previous wording implies that we don't have token detection without this feature enabled.


NiranjanaBinoy (approved):
LGTM

ryanml (approved):

PR #12500, opened 2021-10-27
Turn off all old notifications

danjm:
This PR turns off all existing notifications.

I considered removing those notifications from `shared/notifications/index.js`, but I actually think it would be good to have a record of notifications we have shown in the past.

### DIFF
--- a/test/e2e/metamask-ui.spec.js
+++ b/test/e2e/metamask-ui.spec.js
@@ -176,24 +176,6 @@ describe('MetaMask', function () {
     });
   });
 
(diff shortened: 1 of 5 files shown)


darkwing (dismissed):
LGTM for now -- we should build in an expiration system, whereby notifications no longer display after a given date.

NiranjanaBinoy (dismissed):

darkwing (approved):

NiranjanaBinoy (approved):

PR #12502, opened 2021-10-27
Network tab refactor

### LINKED ISSUE #12404: Refactor Network Tab in Settings to follow Latest Code Patterns
https://github.com/MetaMask/metamask-extension/pull/12302 added new designs for adding custom networks.

That PR fit the new designs into existing code patterns, however review of that PR highlighted many ways in which the existing code can improve.

We should make the following refactors to the networks tab and it child components:
- The `renderAddNetworkForm` and `renderNetworkForm` methods should be combined to a single function, and css should be used to reduce the amount of conditional "view logic" in the component code. Ideally, after refactoring, we won't need divs to specify rows and columns, and instead we can hopefully use css grid
- The networks tab and all its child components, including the network forms, should be refactored into functional components, and refactored to use hooks for state management, lifecycle management and communication with redux
- Make use of the new <FormField /> (found at form-field.js) component which takes a tooltip and label/title prop, rather than the custom ones made here. 
- We can consider whether we want to improve on the route handling as well. It is possible the logic can be simplified.

NiranjanaBinoy:
Fixes: #12404 

- Combined `renderAddNetworkForm` and `renderNetworkForm` methods to a single functional component using of CSS grid.
- Networks tab is split into 4 functional components `network-tab.js`, `network-tab-content.js`, `network-list.js`, and `network-list-item.js`.
- The custom form text fields are replaced by `<FormField/>`
- Improved the routing to path ADD_NETWORK_ROUTE to render with prop `addNewNetwork`.


ui/pages/settings/networks-tab/network-tab.js
```diff
@@ -0,0 +1,157 @@
...
+import { NETWORK_TYPE_RPC } from '../../../../shared/constants/network';
+import { defaultNetworksData } from './networks-tab.constants';
+import NetworkTabContent from './network-tab-content';
+import NetworkForm from './network-form';
+
+const defaultNetworks = defaultNetworksData.map((network) => ({
+  ...network,
+  viewOnly: true,
+}));
+
```
jpuri, line 28: @NiranjanaBinoy : it might be useful to place this in separate file and the component is fairly large.

jpuri (dismissed):
LGTM, just some small feedbacks.

Also, would be nice to have some unit test coverage added for new components.

segun:
LGTM

segun (approved):

jpuri (approved):

PR #12519, opened 2021-10-28
Provide instructions for users switching from Ledger Live to WebHID

darkwing:
This provides a message to the user when they switch from Ledger Live to WebHid.  The message will help them avoid the weird issue of dual transports competing for Ledger control.


danjm (approved):

adonesky1 (approved):
LGTM

PR #12521, opened 2021-10-29
Exclude files from builds by build type

### LINKED ISSUE #12428: Exclude Flask-only JavaScript source files from Flask builds
We should instrument the build system to exclude Flask-only source files from non-Flask builds. We need to decide on some kind of pattern / structure for Flask files so that they're easy to exclude using `browserify.exclude()`, which accepts globs as an argument.

rekmarks:
This PR enables the exclusion of JavaScript and JSON source by `buildType`, and enables the running of `eslint` under LavaMoat. 80-90% of the changes in this PR are `.patch` files and LavaMoat policy additions.

The file exclusion is designed to work in conjunction with our code fencing. If you forget to fence an import statement of an excluded file, the application will now error on boot. **This PR commits us to a particular naming convention for files intended only for certain builds.** Continue reading for details.

### Code Fencing and ESLint

When a file is modified by the code fencing transform, we run ESLint on it to ensure that we fail early for syntax-related issues. This PR adds the first code fences that will be actually be removed in production builds. As a consequence, this was also the first time we attempted to run ESLint under LavaMoat. Making that work required a lot of manual labor because of ESLint's use of dynamic imports, but the manual changes necessary were ultimately quite minor.

### File Exclusion

For all builds, any file in `app/`, `shared/` or `ui/` in a sub-directory matching `**/${otherBuildType}/**` (where `otherBuildType` is any build type except `main`) will be added to the list of excluded files, regardless of its file extension. For example, if we want to add one or more pages to the UI settings in Flask, we'd create the folder `ui/pages/settings/flask`, add any necessary files or sub-folders there, and fence the import statements for anything in that folder. If we wanted the same thing for Beta, we would name the directory `ui/pages/settings/beta`.

As it happens, we already organize some of our source files in this way, namely the logo JSON for Beta and Flask builds. See `ui/helpers/utils/build-types.js` to see how this works in practice.

Because the list of ignored filed is only passed to `browserify.exclude()`, any files not bundled by `browserify` will be ignored. For our purposes, this is mostly relevant for `.scss`. Since we don't have anything like code fencing for SCSS, we'll have to consider how to handle our styles separately.

### Testing

Run the following commands to build the extension locally and observe that everything works. Make sure to reload every time the build script is restarted:
- `yarn start` (builds `main` by default)
- `yarn start --build-type beta`
- `yarn start --build-type flask`

### Future Concerns: SCSS

Regarding SCSS, `gulp.src` accepts globs, can accept an array of strings, and will ignore anything prefixed with `!`. Our best bet may be to create a separate SCSS entry point for Flask files and exclude it depending on the build type.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,4 +1,4 @@
-const reactVersion = require('./package.json').dependencies.react;
+const { version: reactVersion } = require('react/package.json');
 
 module.exports = {
   root: true,
@@ -215,11 +215,11 @@ module.exports = {
 
   settings: {
     react: {
-      // If this is set to 'detect', eslint will import react in order to find
+      // If this is set to 'detect', ESLint will import React in order to find
       // its version. Because we run ESLint in the build system under LavaMoat,
       // this means that detecting the React version requires a LavaMoat policy
       // for all of React, in the build system. That's a no-go, so we grab it
-      // from package.json.
+      // from React's package.json.
       version: reactVersion,
     },
   },
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,5 @@
+const reactVersion = require('./package.json').dependencies.react;
+
 module.exports = {
   root: true,
   parser: '@babel/eslint-parser',
@@ -213,7 +215,12 @@ module.exports = {
 
   settings: {
     react: {
-      version: 'detect',
+      // If this is set to 'detect', eslint will import react in order to find
+      // its version. Because we run ESLint in the build system under LavaMoat,
+      // this means that detecting the React version requires a LavaMoat policy
+      // for all of React, in the build system. That's a no-go, so we grab it
+      // from package.json.
+      version: reactVersion,
     },
   },
 };
--- a/lavamoat/node/policy-override.json
+++ b/lavamoat/node/policy-override.json
@@ -27,6 +27,11 @@
         "eslint-plugin-react-hooks": true
       }
     },
+    "eslint-module-utils": {
+      "packages": {
+        "eslint-import-resolver-node": true
+      }
+    },
     "node-sass": {
       "native": true
     },
--- a/lavamoat/node/policy-override.json
+++ b/lavamoat/node/policy-override.json
@@ -1,13 +1,5 @@
 {
   "resources": {
-    "node-sass": {
-      "native": true
-    },
-    "module-deps": {
-      "packages": {
-        "loose-envify": true
-      }
-    },
     "@babel/core": {
       "packages": {
         "<root>": true,
@@ -18,6 +10,29 @@
         "@babel/plugin-proposal-object-rest-spread": true,
         "@babel/plugin-proposal-optional-chaining": true,
         "@babel/plugin-proposal-nullish-coalescing-operator": true
+      }
+    },
+    "@eslint/eslintrc": {
+      "packages": {
+        "@babel/eslint-parser": true,
+        "@babel/eslint-plugin": true,
+        "@metamask/eslint-config": true,
+        "@metamask/eslint-config-nodejs": true,
+        "eslint": true,
+        "eslint-config-prettier": true,
+        "eslint-plugin-import": true,
+        "eslint-plugin-node": true,
+        "eslint-plugin-prettier": true,
+        "eslint-plugin-react": true,
+        "eslint-plugin-react-hooks": true
+      }
+    },
+    "node-sass": {
+      "native": true
+    },
+    "module-deps": {
+      "packages": {
+        "loose-envify": true
       }
     },
     "sass": {
--- a/package.json
+++ b/package.json
@@ -264,6 +264,7 @@
     "enzyme-adapter-react-16": "^1.15.1",
     "eslint": "^7.23.0",
     "eslint-config-prettier": "^8.1.0",
+    "eslint-import-resolver-node": "^0.3.4",
     "eslint-plugin-import": "^2.22.1",
     "eslint-plugin-jest": "^24.3.4",
     "eslint-plugin-mocha": "^8.1.0",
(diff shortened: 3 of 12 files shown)


package.json
```diff
@@ -176,7 +176,7 @@
     "pump": "^3.0.0",
     "punycode": "^2.1.1",
     "qrcode-generator": "1.4.1",
-    "react": "^16.12.0",
+    "react": "16.14.0",
```
rekmarks, line 179: I had to pin the version so we can dependably extract it in `.eslintrc.js`. See that file for details.

patches/@eslint+eslintrc+0.4.0.patch
```diff
@@ -0,0 +1,21 @@
...
+ const path = require("path");
+-const importFresh = require("import-fresh");
+ const stripComments = require("strip-json-comments");
+ const ConfigValidator = require("./shared/config-validator");
+ const naming = require("./shared/naming");
+@@ -222,7 +221,7 @@ function loadLegacyConfigFile(filePath) {
+ function loadJSConfigFile(filePath) {
+     debug(`Loading JS config file: ${filePath}`);
+     try {
+-        return importFresh(filePath);
```
rekmarks, line 17: `import-fresh` is a Sindre Sorhus package with Sindre Sorhus dependencies with too-clever-by-half Sindre Sorhus hacks that are fundamentally incompatible with LavaMoat. 

`import-fresh` is basically native `require`/`import`, but without retaining any module state created by previous imports of the module. I don't know why the ESLint authors wanted to use it here, but it seems to work just fine without it.

patches/eslint+7.23.0.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/eslint/lib/linter/linter.js b/node_modules/eslint/lib/linter/linter.js
+index adb5c21..4a4be92 100644
+--- a/node_modules/eslint/lib/linter/linter.js
++++ b/node_modules/eslint/lib/linter/linter.js
+@@ -560,7 +560,7 @@ function resolveParserOptions(parserName, providedOptions, enabledEnvironments)
+  */
+ function resolveGlobals(providedGlobals, enabledEnvironments) {
+     return Object.assign(
+-        {},
++        Object.create(null),
```
rekmarks, line 10: For some reason, the properties being assigned include `Object.prototype` properties, and assigning them in this manner appears to walk the prototype chain. This doesn't work under LavaMoat, because the prototype is frozen. Creating an object with a `null` prototype seems to work just fine.

patches/object.values+1.1.1.patch
```diff
@@ -0,0 +1,23 @@
...
+-var polyfill = getPolyfill();
+-
+-define(polyfill, {
+-	getPolyfill: getPolyfill,
+-	implementation: implementation,
+-	shim: shim
+-});
+-
+-module.exports = polyfill;
++module.exports = Object.values;
```
rekmarks, line 23: The polyfill was assigning a bunch of garbage to `Object.values`. It's no longer 1999, so I don't think we need this nonsense.

lavamoat/node/policy-override.json
```diff
@@ -20,6 +12,29 @@
...
+        "@metamask/eslint-config-nodejs": true,
+        "eslint": true,
+        "eslint-config-prettier": true,
+        "eslint-plugin-import": true,
+        "eslint-plugin-node": true,
+        "eslint-plugin-prettier": true,
+        "eslint-plugin-react": true,
+        "eslint-plugin-react-hooks": true
+      }
+    },
```
rekmarks, line 29: These are all dynamically required by `@eslint/eslintrc` at runtime.

lavamoat/node/policy-override.json
```diff
@@ -20,6 +12,34 @@
...
+    "eslint-module-utils": {
+      "packages": {
+        "eslint-import-resolver-node": true
+      }
+    },
```
rekmarks, line 34: For some reason this is only used by `eslint-module-utils` with `--build-type flask` or `beta`.

yarn.lock
```diff
@@ -24517,7 +24517,7 @@ react-transition-group@^4.4.0:
     loose-envify "^1.4.0"
     prop-types "^15.6.2"
 
-react@^16.12.0:
+react@16.14.0:
```
shanejonas, line 24520: 👍

ui/helpers/utils/build-types.js
```diff
@@ -1,17 +1,25 @@
+///: BEGIN:ONLY_INCLUDE_IN(beta)
 import betaJson from '../../../app/build-types/beta/beta-mascot.json';
+///: END:ONLY_INCLUDE_IN
+///: BEGIN:ONLY_INCLUDE_IN(flask)
 import flaskJson from '../../../app/build-types/flask/flask-mascot.json';
+///: END:ONLY_INCLUDE_IN
```
shanejonas, line 6: 🤯

shanejonas (dismissed):

.eslintrc.js
```diff
@@ -213,7 +215,12 @@ module.exports = {
 
   settings: {
     react: {
-      version: 'detect',
+      // If this is set to 'detect', eslint will import react in order to find
+      // its version. Because we run ESLint in the build system under LavaMoat,
+      // this means that detecting the React version requires a LavaMoat policy
+      // for all of React, in the build system. That's a no-go, so we grab it
+      // from package.json.
+      version: reactVersion,
```
kumavis, line 223: can we do `require('react/package.json').version` instead?

rekmarks: That is much better: 61ffca9

kumavis (approved):
We’ve been trying to reach you about your car’s extended warranty :car:

yarn.lock
rekmarks: This was undone: 4e4f08a

package.json
rekmarks: Actually, no: 4e4f08a

ryanml (approved):

PR #12530, opened 2021-10-29
Adding GasFee context

jpuri:
The PR creates a context on top of `useGasFeeInputs` hook, the data from `useGasFeeInputs` will now be available to all 3 screen:

1. confirm-transaction
2. edit gas fee
3. advance gas fee

For implementing V2 designs this is needed. The PR takes care of creating the context and wrapping required elements in that context.


ui/contexts/gasFee.js
```diff
@@ -0,0 +1,37 @@
...
+export const GasFeeContext = createContext({});
+
+export const GasFeeContextProvider = ({
+  children,
+  defaultEstimateToUse,
+  transaction,
+  minimumGasLimit,
+  editGasMode,
+}) => {
+  const gasFeeDetails = useGasFeeInputs(
```
brad-decker, line 14: Is the benefit of having this here instead of multiple hook invocations simply that there can be no discrepancies? I'm not sure I quite follow what the benefit of this context would be over using the hooks return values directly.

jpuri: Hey @brad-decker : please check these following PRs which use the code from this PR
https://github.com/MetaMask/metamask-extension/pull/12545
https://github.com/MetaMask/metamask-extension/pull/12567

Results from the context can be used in any child component like [this](https://github.com/MetaMask/metamask-extension/pull/12545/files#diff-66d162bb86cf769887e2240412f8950a0f5b3c16f562f75640956f344b03de94R53).

danjm: @jpuri Brad does raise a question worth discussing and explaining though:

Why is it better to call `useGasFeeContext` on any child component? Instead of just calling `useGasFeeInputs` on any child component?

jpuri: Sure, gas related information are required in a lot of components, we need it for V2 on 3 screens:
1. confirm-transaction
2. edit gas fee
3. advance gas fee

These 3 pages will in-turn have child components thus in all will be about 20 components (approx including containers). If we do not use context we need to pass down the props across all these components. Context eases this props passing in react and is thus useful to keep transient state.

If there is a data field which we want to keep in redux also we can save it in redox from `GasFeeContext` as we do with other react components.

danjm: Oh, right.

So to further add to the answer to @brad-decker's question is that there is transitional state that is set in `useGasFeeInputs` in response to user action. These are the custom values the user sets for gas fees. We can't just call `useGasFeeInputs` in a bunch of different places, because each `useGasFeeInputs` call is independent of the other. For the `useGasFeeInputs` to have the data we need it to have in all places, we would have to connect it to redux.

This comment from our slack thread explained this further:

> Gotcha. I would say that anything that is part of the "current draft transaction" should be in Redux, in a reducer that represents that draft (e.g. confirm-transaction, though in practice that slice doesn't fully serve that purpose at the moment). But transient data used by gas inputs just while they're being adjusted doesn't belong in Redux, and using component state or React context makes perfect sense for that state (https://consensys.slack.com/archives/GTQAGKY5V/p1635345447101200?thread_ts=1635316808.093100&cid=GTQAGKY5V)

ui/pages/confirm-transaction-base/gas-details-item/gas-details-item.js
```diff
@@ -29,6 +30,9 @@ const GasDetailItem = ({
   useNativeCurrencyAsPrimaryCurrency,
 }) => {
   const t = useI18nContext();
+  const { estimateToUse } = useGasFeeContext();
+  console.log('estimateToUse = ', estimateToUse); // estimateToUse available for use here
```
danjm, line 34: This console.log can be removed

ui/pages/swaps/view-quote/view-quote.js
```diff
@@ -674,143 +675,158 @@ export default function ViewQuote() {
...
+    userFeeLevel: swapsUserFeeLevel || 'high',
+    txParams: {
+      maxFeePerGas,
+      maxPriorityFeePerGas,
+      gas: maxGasLimit,
+    },
+  };
 
+  return (
+    <>
```
danjm, line 688: Is this fragment (`<>`) needed here? If we remove it, it will make the diff in this file a lot simpler

ui/components/app/transaction-list-item/transaction-list-item.component.js
```diff
@@ -140,9 +141,18 @@ export default function TransactionListItem({
   ]);
 
   const showCancelButton = !hasCancelled && isPending && !isUnapproved;
+  let mode;
+  if (showRetryEditGasPopover) {
+    mode = EDIT_GAS_MODES.SPEED_UP;
+  } else if (showCancelEditGasPopover) {
+    mode = EDIT_GAS_MODES.CANCEL;
+  }
```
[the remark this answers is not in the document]
danjm: Hmmm.... we only need to add wrap this in `GasFeeContextProvider` for the sake of `EditGasPopover`, right?

Also, I am realizing that wherever `EditGasPopover` is, we will need `GasFeeContextProvider`. Is there a way we can eliminate the code duplication of always wrapping `EditGasPopover` in `GasFeeContextProvider`? It would also be nice not to have to wrap components that don't need this context.

For instance, could we modify the edit-gas-popover file to export a function that wraps `EditGasPopover` with `GasFeeContextProvider`, so that the wrapping only needs to be done once?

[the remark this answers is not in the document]
jpuri: Hey @danjm : if we just wrap `EditGasPopover` in context it will defeat the purpose of having data available to all 3 screens in V2 flow.

There is no advantage of minimising the scope of context, a lot of times context if added to root of application like we have for i18n. Bigger scope will ensure availability of this data to other components also if needed - no downside actually.

ui/pages/confirm-transaction-base/gas-details-item/gas-details-item.js
jpuri: it is removed in following PR actually, I had added it here just for demoing the use of context in POC

let me remove here also

ui/pages/swaps/view-quote/view-quote.js
jpuri: Good point 👍

jpuri:
@danjm : you feedbacks made me think more about it and I realised as we will get rid of edit gas modal in new designs I will undo changes from old code and use this context for only new flow. That ways chances of regressions will also be minimised.

jpuri:
I added WIP to PR to look more into breaking e2e.

ui/hooks/gasFeeInput/useGasFeeErrors.js
```diff
@@ -21,6 +21,7 @@ import { GAS_FORM_ERRORS } from '../../helpers/constants/gas';
 const HIGH_FEE_WARNING_MULTIPLIER = 1.5;
```
jpuri, line 22: After creating context over `useGasFeeHook` e2e was unstable in FF cox of race condition, these null checks are added for that.

Gudahtt: This is concerning. Why would `gasLimit` be `undefined`? I see that it is supposed to be initialized to zero in `useGasFeeInputs`. Maybe I'm missing a case.

jpuri: `useState` hook is asynchronous, it is less likely but possible that hook to find errors is executed at same time.

For other fields utility methods used for comparison already have check for `undefined` value. But I found that these 3 fields did not had that: `gasLimit`, `minimumCostInHexWei`, `ethBalance`. May be `gasLimit` is not reason here, I can try to remove null check for it and see.

jpuri: From last commit it seems `gasLimit` was not the reason.

Gudahtt: `useState` is definitely synchronous. Everything that happens during the render is synchronous, so it will get initialized. If something is set to `undefined`, it must mean it was not initialized properly in the first place, or it was set to `undefined`.

jpuri: Yep render is definitely synchronous, but I was thinking of this possibility as error code that was throwing exception is run before render. But I am probably wrong here.

ui/contexts/gasFee.js
```diff
@@ -0,0 +1,37 @@
...
+    </GasFeeContext.Provider>
+  );
+};
+
+export function useGasFeeContext() {
+  return useContext(GasFeeContext);
+}
+
+GasFeeContextProvider.propTypes = {
+  children: PropTypes.node,
```
mcmire, line 32: Let's not forget to add `.isRequired` on things that should always be passed. This information is super helpful to know how to write and update unit tests in the absence of TypeScript because you know which props to pass into the component in the test (and which once you can opt to leave out).

ui/hooks/gasFeeInput/useGasFeeErrors.js
```diff
@@ -131,6 +131,8 @@ const getMaxFeeWarning = (
 };
 
 const getBalanceError = (minimumCostInHexWei, transaction, ethBalance) => {
+  if (minimumCostInHexWei === undefined || ethBalance === undefined)
```
mcmire, line 134: What do you think about adding braces here? I believe that @darkwing wanted to add an ESLint rule around this not too long ago — [that PR seems to have been closed](https://github.com/MetaMask/metamask-extension/pull/12578) but I think it's still a good idea.

mcmire: Also does it make sense for this function to return false if there is nothing to return? Seems like we might want `null` or something similar.

ui/hooks/gasFeeInput/useGasFeeInputs.js
```diff
@@ -61,7 +62,7 @@ export function useGasFeeInputs(
   defaultEstimateToUse = 'medium',
   transaction,
   minimumGasLimit = '0x5208',
-  editGasMode,
+  editGasMode = EDIT_GAS_MODES.MODIFY_IN_PLACE,
```
mcmire, line 65: What's the reasoning behind this change? This is probably okay but I'm just curious.

mcmire (changes requested):
Hey, sorry it's taken me so long to review this, I had to take some time to really think about this. I think making a component that wraps `useGasFeeInputs` is a good idea. I think it might lead to better refactors in the future. I'm going to think about it a bit more before I officially approve but in the meantime I have a couple of comments.

ui/hooks/gasFeeInput/useGasFeeInputs.js
jpuri: `EDIT_GAS_MODES.MODIFY_IN_PLACE` is widely used gasMode, thus I added it to avoid passing each time.

mcmire: Okay. I'm wondering if we should watch out when setting default props in the future. The reason is that if for some reason this default gets changed, it may end up breaking tests and code that rely on that default. It may lead to less brittle code in the future if we are more explicit about this than less.

mcmire (approved):
I'm good with this!

danjm (approved):
LGTM

ui/hooks/gasFeeInput/useGasFeeInputs.js
jpuri: I agree default should probably be the locally default value and not the most frequently used value. I will remove this in following PRs.

PR #12535, opened 2021-10-29
Check if ledger was successfully able to establish transport on confirm screen mount

danjm:
For Ledger users who connect with WebHID, on some operating systems the connection of the ledger device to other software can prevent the WebHID transport from successfully being created. This PR adds a message to the user if that state is detected.

Depends on https://github.com/MetaMask/eth-ledger-bridge-keyring/pull/125 and https://github.com/MetaMask/eth-ledger-bridge-keyring/pull/126


app/scripts/metamask-controller.js
```diff
@@ -1549,6 +1553,11 @@ export default class MetamaskController extends EventEmitter {
     return keyring;
   }
 
+  async attemptLedgerTransportCreation() {
+    const keyring = this.getKeyringForDevice('ledger');
```
darkwing, line 1557: This needs an `await`:

```
const keyring = await this.getKeyringForDevice('ledger');
```

danjm: this is addressed since my latest rebase

darkwing (approved):

ryanml (approved):

PR #12550, opened 2021-11-01
Ensure exchange rate is stringified before passing to bignumber in token confirm component

danjm:
With this change, we introduced the possibility of exchange rate numbers having more than 15 significant digits: https://github.com/MetaMask/controllers/pull/585

Our current version of `BigNumber` does not handle that well, but does if the number is stringified.

Currently in prod, users will hit an error screen when this happens. Here is the associated error record in sentry https://sentry.io/organizations/metamask/issues/2745103173/?project=273505&query=is%3Aunresolved+firstRelease%3Alatest&statsPeriod=14d

### DIFF
--- a/ui/pages/confirm-token-transaction-base/confirm-token-transaction-base.component.js
+++ b/ui/pages/confirm-token-transaction-base/confirm-token-transaction-base.component.js
@@ -35,7 +35,9 @@ export default function ConfirmTokenTransactionBase({
     }
 
     const decimalEthValue = new BigNumber(tokenAmount)
-      .times(new BigNumber(contractExchangeRate))
+      .times(
+        new BigNumber(contractExchangeRate ? String(contractExchangeRate) : 0),
+      )
       .toFixed();
 
     return getWeiHexFromDecimalValue({


Gudahtt (commented):
Do we know why `bignumber.js` throws on numbers with more than 15 significant digits? If it's a matter of significant digits beyond 15 being unreliable due to floating point error, maybe we're better off truncating our conversion rate to 15 (e.g. using `toPrecision(15)`) rather than keeping this additional precision.

Longer-term we should consider using BigInt to perform these calculations in the controllers repo so we don't lose precision from floating point errors, and truncating the calculated value to reflect the amount of precious that was input, and returning the value as a BigInt or at least a string.

darkwing (approved):

Gudahtt (approved):
I'll approve for now since this seems like a quick and relatively harmless way to fix the crash, but we should follow up with a better solution that doesn't include spurious digits that give a false impression of how precise this value is.

For more information on that, see https://en.wikipedia.org/wiki/Significant_figures#Rules_to_identify_significant_figures_in_a_number
and https://en.wikipedia.org/wiki/Significant_figures#Multiplication_and_division

danjm:
@Gudahtt A couple notes regarding big number. In the latest versions, it does not throw when significant digits are greater than 15, unless the `DEBUG` flag is set to true.

Before the latest version, stringifying before passing to bignumber was the recommended approach: https://github.com/MikeMcl/bignumber.js/issues/148#issuecomment-384812951

PR #12552, opened 2021-11-01
Add crowdin configuration and github action

rickycodes:
Add crowdin configuration and GitHub action

TODO:

- [x] Add `crowdin/github-action@1.0.4` to "Allow select actions section" in settings/actions (this option can only be set by organisation administrators)
- [x] Add secrets for `CROWDIN_PROJECT_ID` and `CROWDIN_PERSONAL_TOKEN` (these should be regenerated from the crowdin project)
- [x] Test again.

### DIFF
--- a/crowdin.yml
+++ b/crowdin.yml
@@ -1,0 +1,127 @@
+#
+# Your Crowdin credentials
+#
+"project_id_env": CROWDIN_PROJECT_ID
+"api_token_env": CROWDIN_PERSONAL_TOKEN
+"base_path" : "."
+"base_url" : "https://metamask.crowdin.com"
+
+#
+# Choose file structure in Crowdin
+# e.g. true or false
+#
+"preserve_hierarchy": true
+
+#
+# Files configuration
+#
+files: [
+ {
+  #
+  # Source files filter
+  # e.g. "/resources/en/*.json"
+  #
+  "source" : "app/_locales/en/messages.json",
+
+  #
+  # Where translations will be placed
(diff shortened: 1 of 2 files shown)


crowdin.yml
```diff
@@ -0,0 +1,127 @@
...
+  #
+  "source" : "app/_locales/en/messages.json",
+
+  #
+  # Where translations will be placed
+  # e.g. "/resources/%two_letters_code%/%original_file_name%"
+  #
+  "translation" : "/app/_locales/%two_letters_code%/%original_file_name%",
+
+  #
```
adonesky1, line 32: we can probably delete all the commented out configuration for now?

rickycodes: I guess this was just how it was initialised? I thought maybe it could be useful for reference, but I could absolutely delete the comments and then we could just reference back to this (or their docs) if we ever need to adjust.

adonesky1:
@rickycodes I can't remember if you already tested this but can we use [this](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#onpushpull_requestpaths) path property to only have the job run if the source translation file has been changed in the latest commit?

adonesky1:
In the docs I guess it looks more like either a branch or a path but not both?

rickycodes:
@adonesky1 I haven't tried with both, but we can also try that after this is merged? :smiling_imp:

adonesky1 (approved):
Lets go!

darkwing (approved):

PR #12557, opened 2021-11-02
Ensure ledger transport preference is set on import

danjm:
Fixes an issue I found while addressing another problem with the ledger webhid implementation.

This ensures that if a user onboards by importing an existing seed phrase, and then connects their ledger, that `webhid` is correctly set as the transport type in the ledger bridge.

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -1317,6 +1317,13 @@ export default class MetamaskController extends EventEmitter {
         accounts = await keyringController.getAccounts();
       }
 
+      // This must be set as soon as possible to communicate to the
+      // keyring's iframe and have the setting initialized properly
+      // Optimistically called to not block Metamask login due to
+      // Ledger Keyring GitHub downtime
+      const transportPreference = this.preferencesController.getLedgerTransportPreference();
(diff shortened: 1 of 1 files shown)


app/scripts/metamask-controller.js
```diff
@@ -1317,6 +1317,13 @@ export default class MetamaskController extends EventEmitter {
         accounts = await keyringController.getAccounts();
       }
 
+      // This must be set as soon as possible to communicate to the
+      // keyring's iframe and have the setting initialized properly
+      // Optimistically called to not block Metamask login due to
+      // Ledger Keyring GitHub downtime
+      const transportPreference = this.preferencesController.getLedgerTransportPreference();
+      this.setLedgerTransportPreference(transportPreference);
```
darkwing, line 1325: Could this ever be `undefined`?  Should we set a default in `setLedgerTransportPreference`?

danjm: there is a default set in the preferences controller, so I don't think this could be undefined

jpuri (approved):
LGTM

darkwing (approved):

PR #12561, opened 2021-11-02
Sentry - update to v6

kumavis:
update sentry deps across the major version boundary

https://github.com/getsentry/sentry-javascript/releases/tag/6.0.0
> This major version release doesn't contain any breaking API/code changes.

the primary change is the default setting of enabling sessions https://docs.sentry.io/product/releases/health/


jpuri (approved):

rekmarks (changes requested):
>You can opt-out of this behavior by setting `autoSessionTracking: false` option during SDK initialization.

Do we want to do this? I can't find a reference to `autoSessionTracking` in our codebase, so presumably Sentry's behavior will change if we don't disable it.

kumavis:
im curious to take sessionTracking for a spin

rekmarks (approved):
LGTM.

PR #12562, opened 2021-11-02
Add lockdown e2e test

rekmarks:
This PR adds an e2e test to ensure that the background and UI environments are locked down. It reuses the logic from the `protect-intrinsics.test.js`, and runs in both Chrome and Firefox.


Gudahtt (approved):
LGTM!

I really wish I understood why we needed that extra reassignment of `globalThis` on Firefox specifically. But at least the test is working.

kumavis (approved):

PR #12563, opened 2021-11-02
sentry - filter Network Error messages

### LINKED ISSUE #12334: Axios network failures spamming Sentry_Last 30 Days: 0
Sentry Issue: [METAMASK-GN3J](https://sentry.io/organizations/metamask/issues/2050331057/?referrer=github_integration)

```
Error: Network Error
  at createError (node_modules/axios/lib/core/createError.js:16:1)
  at apply (node_modules/axios/lib/adapters/xhr.js:99:1)
  at XMLHttpRequest.c (node_modules/@sentry/browser/dist/helpers.js:75:1)
```
Last 30 Days: 0

kumavis:
attempting to fix https://github.com/MetaMask/metamask-extension/issues/12334

### DIFF
--- a/app/scripts/lib/setupSentry.js
+++ b/app/scripts/lib/setupSentry.js
@@ -100,8 +100,29 @@ export default function setupSentry({ release, getState }) {
     environment,
     integrations: [new Dedupe(), new ExtraErrorData()],
     release,
-    beforeSend: (report) => rewriteReport(report),
+    beforeSend: (report) => {
+      if (shouldFilterReport(report)) {
+        return null;
+      }
+      return rewriteReport(report);
+    },
   });
+
+  function shouldFilterReport(report) {
+    let shouldFilter = false;
+    try {
+      // attempting to reduce segment rate-limiting errors
+      // https://github.com/MetaMask/metamask-extension/issues/12334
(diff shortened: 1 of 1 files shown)


kumavis:
if we want to filter for something more specific we can try to use these parameters
https://github.com/axios/axios/blob/cd7ff042b0b80f6f02e5564d184019131c90cacd/test/specs/core/enhanceError.spec.js#L9-L15

kumavis:
theres a larger issue here that we're allowing this error to bubble up through the app

Gudahtt:
Just as mentioned [here](https://github.com/MetaMask/metamask-extension/pull/12573#issuecomment-960231917), I think we still want to report certain types of network errors to Sentry.

This might also be a good opportunity to extract more information out of these network errors. Ideally we'd be able to send these network error reports in a way that preserves the error information (e.g. response status, error message, URL, etc.)

mcmire:
Taking a look at the comments here, is it fair to say this still needs work?

brad-decker:
@kumavis I believe the issue you were trying to fix with this was resolved, even though the github issue is open the sentry one is closed. going to close this feel free to re-open if you believe this to be in error.

PR #12565, opened 2021-11-03
Remove hexdata field from token send

### LINKED ISSUE #10194: 'Show Hex Data' field ignored for token sends
**Describe the bug**
If the "Show Hex Data" advanced setting is enabled, the "Hex data" field will still appear in the token send flow but it will be completely ineffectual. Any contents submitted will not appear in the final transaction data.

1. Enable "Show Hex Data" in the advanced settings
2. Initiate a token send
3. Enter anything in the "Hex data" field
4. See that on the confirmation page, in the 'Data' tab, the hex data entered will not be shown.

**Expected behavior**
The "Hex data" field should not be shown for token sends. The data sent with a token send needs to reflect the operation taking place.

I'm not sure whether arbitrary data can be added to a token send without breaking it, but if it can, then that would be an OK solution too.

hmalik88:
Fixes: #10194 

Explanation:  Added `asset` as another piece of state mapped to props for the `SendContent` component. Then used `asset` to implement logic to **not** render the `SendHexDataRow` component for assets of type `TOKEN` (ERC-20).

Manual testing steps:  
  - Hit send on main screen (make sure "Show Hex Data" is enabled in Advanced settings prior to this step)
  - Enter recipient address
  - Observe that the Hex Data field is no longer visible

### DIFF
--- a/ui/pages/send/send-content/send-content.component.js
+++ b/ui/pages/send/send-content/send-content.component.js
@@ -55,9 +55,7 @@ export default class SendContent extends Component {
     else if (getIsBalanceInsufficient)
       gasError = INSUFFICIENT_FUNDS_FOR_GAS_ERROR_KEY;
     const showHexData =
-      this.props.showHexData &&
-      asset.type !== ASSET_TYPES.NATIVE &&
-      asset.type !== ASSET_TYPES.TOKEN;
+      this.props.showHexData && asset.type !== ASSET_TYPES.TOKEN;
 
     return (
       <PageContainerContent>
--- a/ui/pages/send/send-content/send-content.component.js
+++ b/ui/pages/send/send-content/send-content.component.js
@@ -9,6 +9,7 @@ import {
   UNSENDABLE_ASSET_ERROR_KEY,
   INSUFFICIENT_FUNDS_FOR_GAS_ERROR_KEY,
 } from '../../../helpers/constants/error-keys';
+import { ASSET_TYPES } from '../../../ducks/send';
 import SendAmountRow from './send-amount-row';
 import SendHexDataRow from './send-hex-data-row';
 import SendAssetRow from './send-asset-row';
@@ -32,6 +33,7 @@ export default class SendContent extends Component {
     noGasPrice: PropTypes.bool,
     networkOrAccountNotSupports1559: PropTypes.bool,
     getIsBalanceInsufficient: PropTypes.bool,
+    asset: PropTypes.string,
   };
 
   render() {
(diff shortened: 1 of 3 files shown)


hmalik88:

ui/pages/send/send-content/send-content.component.js
```diff
@@ -44,13 +46,18 @@ export default class SendContent extends Component {
...
     } = this.props;
 
     let gasError;
     if (gasIsExcessive) gasError = GAS_PRICE_EXCESSIVE_ERROR_KEY;
     else if (noGasPrice) gasError = GAS_PRICE_FETCH_FAILURE_ERROR_KEY;
     else if (getIsBalanceInsufficient)
       gasError = INSUFFICIENT_FUNDS_FOR_GAS_ERROR_KEY;
+    const showHexData =
+      this.props.showHexData &&
+      asset.type !== ASSET_TYPES.NATIVE &&
```
danjm, line 59: Why don't we want to showHexData is the asset type is `NATIVE`?

hmalik88: When I added in hex data for an Ether send it no longer showed the transaction as a transfer on etherscan. I'm also not sure why we would want to allow hex data for a `NATIVE` token send, open to hearing your thoughts :)

image

Gudahtt: But... this would disable custom hex data for all transactions :sweat_smile: There are only two types - native and token.

The _sole purpose_ of this field is to add additional data to simple send transactions. Sometimes people want to attach messages to their transfers, and sometimes they just want a way to embed a message on the blockchain. Maybe there are other reasons, I'm not sure.

Either way, I don't see why we'd care how Etherscan displays such transactions.

danjm: Okay, yeah, you can just hide hex data if `asset.type !== ASSET_TYPES.TOKEN`

hmalik88: @Gudahtt That's true, I suppose hex data wouldn't matter for a native token send, I hadn't realized we were not wanting to enable hex data for ERC-20 was because the hex data would potentially mess up the transfer invocation 😅. A transfer can be successful with hex data if the ERC20 has extra functions with an extra field on `transfer` but I guess that would be too much of a pain to deal with. Update coming in a few!

hmalik88: > To clarify that a little bit, it's not so much that it would mess up the transfer invocation. But rather that right now we _overwrite_ the custom hex data entered by the user with the transfer invocation. So it's misleading to include the filed when it just gets overwritten.

Good to know, thanks!

kumavis (approved):

Gudahtt (approved):
LGTM!

PR #12566, opened 2021-11-03
Lavamoat - enable lavamoat for the webapp background

kumavis:


rekmarks:
🌋🌋🌋

Why does do the resulting policy changes mostly result in authority being removed?

kumavis:
@rekmarks the policy change is unrelated to enabling lavamoat, its either due to optional dependencies (if so, potentially an issue) or code having been merged without updating the policy

Gudahtt (dismissed):
This looks good! Tested on Chrome and Firefox, and it seemed relatively unchanged. The initial background initialization felt slower, but not by enough to matter.

This could use a rebase and a newly generated Lavamoat policy, but we update the policy as part of the release process anyway.

Gudahtt (approved):

mcmire (approved):
This looks good to me as well. I also tested on both Chrome and Firefox and didn't notice anything different. Let's goooooooo!

kumavis:
nuked `node_modules` and generated fresh policy, this is what i get

PR #12570, opened 2021-11-03
add dynamic pollingIntervals for JsonRpcClients

dhrubabasu:
Fixes: #9556

Explanation:
Currently, all chainIds default to the 20s pollingInterval as defined here: https://github.com/MetaMask/eth-block-tracker/blob/main/src/PollingBlockTracker.ts#L45

For chains that have significantly faster blocktimes, this leads to an undesirable UX where the transaction has already been finalized but the injected web3 connector provided by metamask hasn't registered it yet.


dhrubabasu:

danjm (changes requested):
Thanks for the contribution. This PR will need some changes.

The linked issue #9556 says

> Instead, it would be better if we even did a very brief rolling average of recent block times to derive a reasonable polling interval.

Instead of hardcoding intervals, as this PR does, we need to derive block times dynamically.

This will require us adding logic that does necessary polling and computes the average, persists that in state and then passes that state to the blocktracker.

danjm:
@dhrubabasu Thanks for the update. Getting closer, but this will still need to be stored in a metamask state. And there are some optimizations we can make. I will write some notes later to guide how I think this can be refactored.

dhrubabasu:
@danjm Any updates here? It may be easier/faster for you to take it from here as I'm unfamiliar with metamask's state management. Happy to continue iterating if it's not a huge lift.

gdassori:
Is there something that can be done to speed up this PR?

backkem:
There is also a [SubscribeBlockTracker](https://github.com/MetaMask/eth-block-tracker/blob/main/src/SubscribeBlockTracker.ts) in [MetaMask/eth-block-tracker](https://github.com/MetaMask/eth-block-tracker) based on the [newHeads notification](https://geth.ethereum.org/docs/rpc/pubsub). Would reviving this be an option? It takes the guessing and bookkeeping out of the equation. I do wonder if this pubsub endpoint is broadly supported by all node implementations. Maybe someone has insights into that.

If not, it's probably a good idea to put in a limit (maybe 1 or 0.5s) on the calculated polling time to avoid completely DDOSing the API infrastructure.

mcmire:
Revisiting this PR. It seems like it would be a worthwhile change for chains that we don't support on Infura. Two questions:

* It looks like we would only calculate the polling interval when MetaMask starts/loads. So this would not be a performance hit to the network. Is this true? /cc @danjm 
* It appears that the goal of most of the code here is to calculate the average block time of the network. What do you think about extracting that to a separate file?

dhrubabasu:
@mcmire Hey, thanks for taking a look!

> It looks like we would only calculate the polling interval when MetaMask starts/loads. So this would not be a performance hit to the network.

Correct, I felt it would be unnecessary to constantly update as most networks I've worked with have relatively constant block times. The current implementation does 3 additional rpc calls at startup:
- 1 `eth_blockNumber` to get latest block number
- 1 `eth_getBlockByNumber` to get the block timestamp of the latest block
- 1 `eth_getBlockByNumber` to get the block timestamp of a prior block (currently configured for latest - 5)

Happy to change this behavior though if you see merit in periodically updating the polling interval.

> It appears that the goal of most of the code here is to calculate the average block time of the network. What do you think about extracting that to a separate file?

Sounds good to me! I put it all in one file for a PoC as I'm not too familiar with Metamask's architecture. If it's not too much of a lift, I'm happy to continue iterating with your guidance.

mcmire:
I think the added requests here are fine.

As for refactoring, maybe we ought to make a new file in `app/scripts/controllers/network` called `calculateAverageBlockTime.js` which takes an `rpcUrl` and spits out a number? Then your `getChainOpts` would basically be:

``` javascript
import calculateAverageBlockTimeInMs from './calculateAverageBlockTimeInMs';

const getChainOpts = async (rpcUrl) => {
  try {
    const averageBlockTimeInMs = await calculateAverageBlockTimeInMs(rpcUrl);
    return {
      pollingInterval: averageBlockTimeInMs * 0.75,
    };
  } catch {
    return {};
  }
};
```

mcmire:
I think the added requests here are fine.

As for refactoring, maybe we ought to make a new file in `app/scripts/controllers/network` called `calculateAverageBlockTime.js` which takes an `rpcUrl` and spits out a number? Then your `getChainOpts` would basically be:

``` javascript
import calculateAverageBlockTimeInMs from './calculateAverageBlockTimeInMs';

const getChainOpts = async (rpcUrl) => {
  try {
    const averageBlockTimeInMs = await calculateAverageBlockTimeInMs(rpcUrl);
    return {
      pollingInterval: averageBlockTimeInMs * 0.75,
    };
  } catch {
    return {};
  }
};
```

dhrubabasu:
refactored per @mcmire's comments and rebased onto latest `develop` branch

app/scripts/controllers/network/createJsonRpcClient.js
```diff
@@ -9,17 +9,32 @@ import {
...
+    return {};
+  }
+};
+
 const getTestMiddlewares = () => {
   return inTest ? [createEstimateGasDelayTestMiddleware()] : [];
 };
 
-export default function createJsonRpcClient({ rpcUrl, chainId }) {
+export default async function createJsonRpcClient({ rpcUrl, chainId }) {
```
dhrubabasu, line 33: I'm guessing this change to async is causing the CI to fail, any ideas on how to remedy this?

app/scripts/controllers/network/network.js
```diff
@@ -418,8 +418,7 @@ export default class NetworkController extends EventEmitter {
 
   _configureStandardProvider(rpcUrl, chainId) {
     log.info('NetworkController - configureStandardProvider', rpcUrl);
-    const networkClient = createJsonRpcClient({ rpcUrl, chainId });
-    this._setNetworkClient(networkClient);
+    createJsonRpcClient({ rpcUrl, chainId }).then(this._setNetworkClient);
```
mcmire, line 421: It's more idiomatic (and fits with our style) to use `async`/`await` syntax here instead of `.then`. That said, making this method asynchronous is going to make everything up the call stack asynchronous as well. That makes this change a lot trickier than I first thought and indicates that this might not be the right approach for this 🤔

dhrubabasu: I tried using `createAsyncMiddleware` in 11da0a95c55f46e673fc80cfaeabc61cf0646036. Probably did it wrong but is that a workable solution?

mcmire: Unfortunately, I don't think that's going to work. You can't use middleware to build other middleware — it's expected that you'll do something with a request and response (see example [here](https://github.com/MetaMask/json-rpc-engine#async-middleware)). This will require some more thought.

dhrubabasu: Ah got it, back to the drawing board I guess 😅

PR #12573, opened 2021-11-03
metametrics - ensure segment submission failures do not bubble up

kumavis:
Prevents segment submission errors from bubbling up and affecting flows that contain metric events

Does this by separating the metametrics trackEvent call into

`trackEvent`:
  - queue for submission
  - **dont wait** for completion
  - **log, dont throw** errors

`submitEvent`:
  - queue for submission
  - wait for completion
  - error on failure

### DIFF
--- a/app/scripts/controllers/metametrics.js
+++ b/app/scripts/controllers/metametrics.js
@@ -288,6 +288,8 @@ export default class MetaMetricsController {
    * @param {MetaMetricsEventOptions} [options] - options for handling/routing the event
    */
   trackEvent(payload, options) {
+    // validation is not caught and handled
+    this.validatePayload(payload);
     try {
       this.submitEvent(payload, options).catch((err) => {
         console.error(err);
@@ -307,20 +309,7 @@ export default class MetaMetricsController {
    * @returns {Promise<void>}
    */
   async submitEvent(payload, options) {
-    // event and category are required fields for all payloads
-    if (!payload.event || !payload.category) {
-      throw new Error(
-        `Must specify event and category. Event was: ${
-          payload.event
-        }. Category was: ${payload.category}. Payload keys were: ${Object.keys(
-          payload,
-        )}. ${
-          typeof payload.properties === 'object'
-            ? `Payload property keys were: ${Object.keys(payload.properties)}`
-            : ''
-        }`,
-      );
-    }
+    this.validatePayload(payload);
 
     if (!this.state.participateInMetaMetrics && !options?.isOptIn) {
       return;
@@ -359,5 +348,26 @@ export default class MetaMetricsController {
     events.push(this._track(this._buildEventPayload(payload), options));
 
     await Promise.all(events);
+  }
+
+  /**
+   * validates a metametrics event
+   * @param {MetaMetricsEventPayload} payload - details of the event
+   */
+  validatePayload(payload) {
+    // event and category are required fields for all payloads
+    if (!payload.event || !payload.category) {
+      throw new Error(
+        `Must specify event and category. Event was: ${
+          payload.event
+        }. Category was: ${payload.category}. Payload keys were: ${Object.keys(
+          payload,
+        )}. ${
+          typeof payload.properties === 'object'
+            ? `Payload property keys were: ${Object.keys(payload.properties)}`
+            : ''
+        }`,
+      );
+    }
   }
 }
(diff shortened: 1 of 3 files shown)


danjm (dismissed):
LGTM

Gudahtt (changes requested):
This would also swallow validation errors we're throwing, like missing `event` and `category` fields. Those should still bubble up to the callsite.

Ideally we'd check for 4xx errors from Segment and report those to Sentry, as they would indicate our request is invalid.

app/scripts/controllers/metametrics.js
```diff
@@ -283,15 +283,30 @@ export default class MetaMetricsController {
   }
 
   /**
-   * track a metametrics event, performing necessary payload manipulation and
+   * submits a metametrics event, not waiting for it to complete or allowing its error to bubble up
```
Gudahtt, line 286: Why is it desirable to not wait for it to complete? We can catch errors and prevent them from bubbling up without making that change.

kumavis: i think that metric event delivery should have no impact on the submitter, it should simply be something queued for delivery

kumavis:
> This would also swallow validation errors we're throwing, like missing event and category fields. Those should still bubble up to the callsite.

I would consider allowing validation errors to bubble up.

> Ideally we'd check for 4xx errors from Segment and report those to Sentry, as they would indicate our request is invalid.

they would be reported as bread crumbs, but we could also report them properly to sentry as full blown errors.

however, I am currently trying to filter them out as the correspond to a significant portion of events and dont seem to be particularly actionable
image

Gudahtt:
Malformed requests are definitely actionable, and a 4xx status typically represents a malformed request.

Timeouts and blocked requests wouldn't have a 4xx status - those should be fine to log without reporting. If Segment uses the 429 status, we can log and suppress those as well (it would probably represent a per-IP rate limit. We'd find out through the dashboard if we hit a project rate limit surely). But the other 4xx-status network failures could indicate we have a bug.

We can't rely on breadcrumbs for reporting because they get attached to Sentry errors. If there is no Sentry error in the vicinity of the failed request, the breadcrumb won't get sent. It's also not easily searchable on their dashboard (as far as I know anyway).

app/scripts/controllers/metametrics.js
Gudahtt: That probably makes sense in most cases, but the ability to wait until a metric event has been sent before continuing can be useful for events that are a part of a funnel. That's the only way to guarantee they're sent in the correct order. 

The Mixpanel funnels rely upon set event orderings to detect which events correspond to which funnel. It allows funnels to have events that are recieved in different orders in some cases, but the first and last event in the defined funnel would have to come first and last respectively, to ensure that those funnel events don't get mixed up with another funnel instance (e.g. in the case where the user goes through another identical flow, which isn't uncommon for sending, swaps, confirmations, etc.).

We do already call `trackEvent` without `await`-ing it in a few places where we know there's no reason to wait, and we can probably start doing that more often. But to make this function non-blocking would prevent us from waiting in the cases where it matters.

kumavis:
@gudahtt for some reason the http response code is "0" in the breadcrumbs
idk if that means the request was blocked and never got a response or the use of axios is obscuring the code from sentry or what

Gudahtt:
I'm not sure how to account for that. I haven't seen any such examples myself, though I have seen some with no status (presumably because the request was still pending or didn't get a response).

app/scripts/controllers/metametrics.js
kumavis: i dont think you're ever supposed to have unhandled promises, so "simply not awaiting it" seems like a bad pattern

kumavis: added validation to `trackEvent`

app/scripts/controllers/metametrics.js
```diff
@@ -258,54 +268,65 @@ export default class MetaMetricsController {
...
+  trackEvent(payload, options) {
+    // validation is not caught and handled
+    this.validatePayload(payload);
+    try {
+      this.submitEvent(payload, options).catch((err) =>
+        this._captureException(err),
+      );
+    } catch (err) {
+      console.error(err);
     }
```
Gudahtt, line 316: Nit: I don't think this `try...catch` block is needed. Since `submitEvent` is `async`, any thrown error will be a Promise rejection. So this catch block won't be hit

```suggestion
    this.submitEvent(payload, options).catch((err) =>
      this._captureException(err),
    );
```

Gudahtt (dismissed):
LGTM! Left one suggested minor improvement.

Gudahtt (approved):

danjm (approved):
bravo

PR #12577, opened 2021-11-03
EIP-1559 V2 : Adding advancedGasFee property to Preference Controller

### LINKED ISSUE #12447: EIP-1559 V2: Create new preference controller properties required to implement EIP-1559 V2 features
EIP-1559 V2 allows the user to save custom settings as defaults, these will include the `option to use custom Max Base Fee and Priority Fee values and advanced settings as default.`

Also will be needed:
- setters for each of these properties in the preferences controller
- getters for each of these in the redux metamask duck
- necessary functions to be added in actions.js

Screen Shot 2021-11-03 at 6 00 27 PM

NiranjanaBinoy:
Fixes: #12447 

EIP-1559 V2 allows the user to save custom settings as defaults, these will include the option to use custom Max Base Fee and Priority Fee values and advanced settings as default.

### DIFF
--- a/app/scripts/controllers/preferences.js
+++ b/app/scripts/controllers/preferences.js
@@ -37,6 +37,7 @@ export default class PreferencesController {
       // set to true means the dynamic list from the API is being used
       // set to false will be using the static list from contract-metadata
       useTokenDetection: false,
+      useAdvancedGasFee: false,
 
       // WARNING: Do not use feature flags for security-sensitive things.
       // Feature flag toggling is available in the global namespace
@@ -127,6 +128,16 @@ export default class PreferencesController {
    */
   setUseTokenDetection(val) {
     this.store.updateState({ useTokenDetection: val });
+  }
+
+  /**
+   * Setter for the `useAdvancedGasFee` property
+   *
+   * @param {boolean} val - Whether or not the user prefers to use the static token list or dynamic token list from the API
+   *
+   */
+  setUseAdvancedGasFee(val) {
+    this.store.updateState({ useAdvancedGasFee: val });
   }
 
   /**
--- a/app/scripts/controllers/preferences.js
+++ b/app/scripts/controllers/preferences.js
@@ -37,7 +37,10 @@ export default class PreferencesController {
       // set to true means the dynamic list from the API is being used
       // set to false will be using the static list from contract-metadata
       useTokenDetection: false,
-      useAdvancedGasFee: false,
+      advancedGasFee: {
+        maxBaseFee: '',
+        priorityFee: '',
+      },
 
       // WARNING: Do not use feature flags for security-sensitive things.
       // Feature flag toggling is available in the global namespace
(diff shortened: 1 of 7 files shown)


app/scripts/controllers/preferences.js
```diff
@@ -129,6 +130,16 @@ export default class PreferencesController {
     this.store.updateState({ useTokenDetection: val });
   }
 
+  /**
+   * Setter for the `useAdvancedGasFee` property
+   *
+   * @param {boolean} val - Whether or not the user prefers to use the static token list or dynamic token list from the API
```
danjm, line 136: param description here needs to change

danjm (commented):
only one small piece of feedback

jpuri:
@NiranjanaBinoy : we need 2 values here:
1. max base fee - percentile value
2. priority fee - in GWEI

We can probably make this an object with 2 props maxFee, maxPriorityFee instead of boolean.

app/scripts/controllers/preferences.js
```diff
@@ -37,6 +37,10 @@ export default class PreferencesController {
       // set to true means the dynamic list from the API is being used
       // set to false will be using the static list from contract-metadata
       useTokenDetection: false,
+      advancedGasFee: {
+        maxBaseFee: '',
+        priorityFee: '',
+      },
```
jpuri, line 43: `undefined` as default may be useful to know that user has no custom setting, WDYT ?

NiranjanaBinoy: I have updated the values to undefined now, it seems to be working.

jpuri: if user has not set default we can check as

1. `advancedGasFee` === undefined or
2. `advancedGasFee.maxBaseFee  == undefined && advancedGasFee.priorityFee  == undefined`

I am thinking what will be best way to check absence of default here.

NiranjanaBinoy: I would prefer to go with 2, it might be an extra check but I think we can reduce the chances of breaking the code if we miss checking for undefined for advancedGasFee anywhere in the future.

mcmire: I am not sure this is a proper use of `undefined`. I feel that `undefined` should only be used internally by JavaScript to indicate that a property of an object doesn't exist. What do we feel about either leaving these properties out entirely and making `advancedGasFee` an empty object (making them *de facto* `undefined`), or (and this might make more sense from a type perspective) setting them to `null`?

jpuri: Setting `advancedGasFee` to null sounds good to me.

NiranjanaBinoy: updated `advancedGasFee` to default to null

jpuri (dismissed):

segun (dismissed):
LGTM

jpuri (approved):

segun (approved):

PR #12578, opened 2021-11-04
Ensure we use curly braces for single statements

darkwing:
I noticed in a few recent PRs with and without braces for single statements, and I feel like we should have a standard.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -55,6 +55,7 @@ module.exports = {
     'default-param-last': 'off',
     'prefer-object-spread': 'error',
     'require-atomic-updates': 'off',
+    'curly': 'error',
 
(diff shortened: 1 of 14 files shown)


mcmire (approved):
Yes please 👍🏻

ryanml (approved):
🙌

.eslintrc.js
```diff
@@ -55,6 +55,7 @@ module.exports = {
     'default-param-last': 'off',
     'prefer-object-spread': 'error',
     'require-atomic-updates': 'off',
+    'curly': 'error',
```
Gudahtt, line 58: This is already set in `@metamask/eslint-config`, to `['error', 'all']`.  Can we use that instead? It's not clear to me why it's not already enabled :thinking: I don't see it being disabled.

PR #12579, opened 2021-11-04
Password field non standard

### LINKED ISSUE #7355: Password field non-standard (incompatible with password managers)
https://support.1password.com/compatible-website-design/

VSaric:
Fixes: #7355

**Explanation:**  

I did a little more research on this issue. I followed all the guidelines offered on the site https://support.1password.com/compatible-website-design/, but still the password field is incompatible with password managers.

I installed several password managers (RoboForm, 1Password, Dashlane, LastPass and Bitwarden) and on each I tested whether the password filed would be compatible with any of the password managers, but without success.

In each of these password managers it is impossible to do an auto-fill for the password field. Why it can't be done, my guess is, because password managers are extensions just like MetaMask and therefore can't recognize the password field. It doesn't support communication between extension and extension.

If you have any suggestions or additional information so that this problem can be solved, please let me know.

**Example:**

https://user-images.githubusercontent.com/92527393/141992267-05d525dd-ac20-40d1-86c3-4a8723e6b9c3.mp4

### DIFF
--- a/ui/pages/first-time-flow/create-password/import-with-seed-phrase/import-with-seed-phrase.component.js
+++ b/ui/pages/first-time-flow/create-password/import-with-seed-phrase/import-with-seed-phrase.component.js
@@ -285,6 +285,8 @@ export default class ImportWithSeedPhrase extends PureComponent {
         </div>
         <TextField
           id="password"
+          name="password"
+          placeholder="New password"
           label={t('newPassword')}
           type="password"
           className="first-time-flow__input"
@@ -292,11 +294,14 @@ export default class ImportWithSeedPhrase extends PureComponent {
           onChange={(event) => this.handlePasswordChange(event.target.value)}
           error={passwordError}
           autoComplete="new-password"
+          passwordrules="minlength: 8; maxlength: 20; required: lower; required: upper; required: digit;"
           margin="normal"
           largeLabel
         />
         <TextField
           id="confirm-password"
+          name="create-password"
+          placeholder="Confirm password"
           label={t('confirmPassword')}
           type="password"
           className="first-time-flow__input"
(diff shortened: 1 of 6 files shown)


mhasanagic:
Vladimir has already given an explanation of how (and why) this ticket was made in this way. What I checked was the following:

- almost all password input fields have added attributes (placeholder, name, password rules). The password input field that doesn’t have these attributes is the field in the section **Settings -> Security&Privacy -> Reveal Secret Recovery Phrase**
Screenshot 2021-11-08 at 17 51 35

- the password rule doesn’t work as it should, ie the specified rule does not have to be followed
- password manager -> auto-fill is not possible



**_Conclusion_** - Password inputs work with these changes as before (nothing is broken)

ui/pages/first-time-flow/create-password/import-with-seed-phrase/import-with-seed-phrase.component.js
```diff
@@ -306,6 +311,7 @@ export default class ImportWithSeedPhrase extends PureComponent {
           }
           error={confirmPasswordError}
           autoComplete="new-password"
+          passwordrules="minlength: 8; maxlength: 20; required: lower; required: upper; required: digit;"
```
darkwing, line 314: I don't believe `passwordrules` is a widely supported attribute, and I believe all of these requirements can be achieved via `minLength`, `maxLength`, and `pattern` attributes.

VSaric: @darkwing minLength, maxLength and pattern attributes are created and added.
You can see changes here: c58db5fbcc2d8fd7680e2ad17a66d0a81015c642 and dd4811f34029a47e9ead01fd5d4da45ea748f12b
Is it okey now?

brad-decker:
@VSaric should we close this and update the ticket with your findings? The text in this PR says that the changes don't actually fix the issue just want to make sure thats an accurate statement.

VSaric:
@brad-decker You are right, I will close this PR and I updated the ticket with my findings here: https://github.com/MetaMask/metamask-extension/issues/7355#issuecomment-961107686.

PR #12580, opened 2021-11-04
Multi network refactor approach A

danfinlay:
DO NOT MERGE!

A total proof of concept for beginning the multi-simultaneous network refactor, in two parts.

Network.js currently does this this clever work of swapping out properties on a pair of proxies, so it emulates having many networks while only maintaining one in memory. This refactor instead caches many evm-type networks, but strives to maintain a non-breaking interface based on the legacy proxy-swapping behavior, so that multi-network methods can be exposed as needed, without requiring a major refactor.

## 1. Refactor network initialization into reusable network objects

Mostly just moved code from the `network.js` controller to a new `network2.js`, which represents a single network, and modified network.js to reuse those network objects as cached instead of re-initializing them.

## 2. Add chain conditional provider to network controller

Allows a network controller consumer to include a chain_id and have the request routed to an appropriate provider if available, defaulting to the selected network.

### BEWARE/TODO:
I am not persisting any state in this, this is not using the current controller architecture, as it's refactoring an old controller, and so for the proof of concept I didn't even use the old style storage. If we choose to go this path, it might be a good time to also upgrade this controller to the new patterns.

### Outstanding ambiguity problem

  There is of course a problem where the user may have many providers with a common chainId. We could approach this in a few ways:
  - Naively to start, just route to one.
  - Parameterize by networkKey, requiring much more specificity (not very dapp friendly to require knowing what provider source is being used)
  - Expose an additional method to allow users to specify the preferred network for a given chainId.
  - Allow users to select the preferred network when connecting to a dapp, and then just remember that networkKey, and route to that provider for that chainId.

### DIFF
--- a/app/scripts/controllers/network/network.js
+++ b/app/scripts/controllers/network/network.js
@@ -92,10 +92,12 @@ export default class NetworkController extends EventEmitter {
 
     /*
      * type NetworkOps = { "type": string, "rpcUrl": string, "chainId": string };
+     * type networkKey = `JSON.parse([type, rpcUrl, chainId])`
      * type Network = { "provider": Provider, "blockTracker": BlockTracker };
-     * type networks = Map<NetworkOps, Network>;
+     * type networks = Map<NetworkKey, Network>;
      */
     this.networks = new Map();
+    this.selectedNetwork = this._networkKeyForOpts(defaultProviderConfig);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/network/network.js
```diff
@@ -297,47 +316,87 @@ export default class NetworkController extends EventEmitter {
...
   }
 
+  _chainConditionalRequestMiddleware (req, res, next, end) {
+    // If no chainId is present, then behave as usual, using "current network" logic:
+    if (!req?.params?.chainId) {
+      return next();
+    }
+
+    const networkOptsArr = this.networks.keys.map(this._networkOptsForKey)
+    .filter(key => key.chainId === req.params.chainId);
```
danfinlay, line 368: need to parse this key into an object.

danfinlay:
Could also support [CAIP-27](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-27.md) as a chain_id routing method: https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-27.md

danfinlay:
Closing for now, will re-open when it's more ready.

PR #12586, opened 2021-11-04
Update Chinese translations: correct typos.

shichangzhi:
Explanation:  Fixing the wrong translations in Chinese locale file.


mcmire (approved):
Thanks for this! Translation looks good to me (just checked using google translate).

darkwing (approved):

PR #12596, opened 2021-11-05
Ensure transaction parameters from dapps are valid hex #11780

### LINKED ISSUE #11780: Ensure transaction parameters from dapps are valid hex
See metamask-extension/app/scripts/controllers/transactions/lib/util.js

dragana8:
Fixes: #11780
https://github.com/MetaMask/metamask-extension/issues/11780

Explanation:  
Fields 'Gas price', 'Max Fee' and 'Max Priority Fee' if populated with some random characters that are not hex caused extension to crash, like on the screenshot. This ticket covers those cases and user gets error message as for the rest of the fields.

Screenshot from 2021-11-04 09-41-48


mhasanagic:
Verified by QA

app/scripts/controllers/transactions/lib/util.js
```diff
@@ -164,6 +164,11 @@ export function validateTxParams(txParams, eip1559Compatibility = true) {
           'maxPriorityFeePerGas',
         );
         ensureFieldIsString(txParams, 'gasPrice');
+        if (!value.match(/^0x[a-fA-F0-9]+$/u)) {
```
david0xd, line 167: Would it be possible and cleaner if you assign that regex to some variable instead of keeping it as a literal? Since it's repeating. I would call it by some descriptive name (i.e. validHexStringRegExp) and define as a const variable somewhere at the top.

app/scripts/controllers/transactions/lib/util.js
```diff
@@ -173,6 +178,11 @@ export function validateTxParams(txParams, eip1559Compatibility = true) {
           'gasPrice',
         );
         ensureFieldIsString(txParams, 'maxFeePerGas');
+        if (!value.match(/^0x[a-fA-F0-9]+$/u)) {
```
david0xd, line 181: Since this code fragment seems to be duplicated few times, can we extract this into a simple function (i.e. ensureValueIsValidHexString())? For error messages try passing the param for the fieldName, etc.

app/scripts/controllers/transactions/lib/util.test.js
```diff
@@ -231,6 +263,22 @@ describe('txUtils', function () {
         );
       });
 
+      it('should error when maxPriorityFeePerGas is not a hex string', function () {
```
david0xd, line 266: If you manage to extract the validation part into a function, I would propose an additional test for that function (positive & negative). The tests added here are nice to see as well and should stay.

app/scripts/controllers/transactions/lib/util.js
dragana8: moved the logic into separate function.

david0xd (dismissed):

darkwing (dismissed):

danjm (approved):
This code looks good. Approving for now. I just need to verify that requiring `0x` at the beginning of these strings will not be a breaking change.

danjm (changes requested):
Actually, I am not going to approve just so that this doesn't accidentally get merged.

darkwing:
Adding `DO-NOT-MERGE` label until Dan researches what he needs to.

brad-decker:
@danjm do you want me to take over this research?

brad-decker:
I am taking over this PR review process. It may not be today but I'll get to checking the items @danjm had laid out and make sure this is mergeable.

PR #12607, opened 2021-11-05
Fix flaky account removal e2e test

danjm:
An attempt to prevent this occasional failure:

Screenshot from 2021-11-05 17-16-05

### DIFF
--- a/test/e2e/tests/from-import-ui.spec.js
+++ b/test/e2e/tests/from-import-ui.spec.js
@@ -233,6 +233,8 @@ describe('Metamask Import UI', function () {
         // should remove the account
         await driver.clickElement({ text: 'Remove', tag: 'button' });
 
+        await driver.delay(regularDelayMs);
+
         const currentActiveAccountName = await driver.findElement(
           '.selected-account__name',
         );
--- a/test/e2e/tests/from-import-ui.spec.js
+++ b/test/e2e/tests/from-import-ui.spec.js
@@ -233,7 +233,12 @@ describe('Metamask Import UI', function () {
         // should remove the account
         await driver.clickElement({ text: 'Remove', tag: 'button' });
 
-        await driver.delay(regularDelayMs);
+        await driver.wait(async () => {
+          const _currentActiveAccountName = await driver.findElement(
+            '.selected-account__name',
+          );
+          return _currentActiveAccountName !== 'Account 5';
+        }, 10000);
 
         const currentActiveAccountName = await driver.findElement(
           '.selected-account__name',
--- a/test/e2e/tests/from-import-ui.spec.js
+++ b/test/e2e/tests/from-import-ui.spec.js
@@ -233,12 +233,13 @@ describe('Metamask Import UI', function () {
         // should remove the account
         await driver.clickElement({ text: 'Remove', tag: 'button' });
 
-        await driver.wait(async () => {
-          const _currentActiveAccountName = await driver.findElement(
-            '.selected-account__name',
-          );
-          return _currentActiveAccountName !== 'Account 5';
-        }, 10000);
+        // Wait until selected account switches away from removed account to first account
+        await driver.waitForSelector({
+            css: '.selected-account__name',
+            text: 'Account 1'
+          },
+          { timeout: 10000 },
+        );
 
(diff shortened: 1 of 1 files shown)


test/e2e/tests/from-import-ui.spec.js
```diff
@@ -233,6 +233,13 @@ describe('Metamask Import UI', function () {
         // should remove the account
         await driver.clickElement({ text: 'Remove', tag: 'button' });
 
+        await driver.wait(async () => {
+          const _currentActiveAccountName = await driver.findElement(
+            '.selected-account__name',
+          );
+          return _currentActiveAccountName !== 'Account 5';
+        }, 10000);
```
Gudahtt, line 241: Nit: A commend would help explain why we're doing this.

```suggestion
        // Wait until selected account switches away from removed account
        await driver.wait(async () => {
          const _currentActiveAccountName = await driver.findElement(
            '.selected-account__name',
          );
          return _currentActiveAccountName !== 'Account 5';
        }, 10000);
```

Gudahtt: We could also consider using the more-concise `waitForSelector`, and matching the expected next account directly rather than looking for "not account 5" and then "account 1" separately.

```suggestion
        // Wait until selected account switches away from removed account to first account
        await driver.waitForSelector({
            css: '.selected-account__name',
            text: 'Account 5'
          },
          { timeout: 10000 },
        );
```

Then if you apply this suggestion, you'll want to delete next few redundant lines as well.

Gudahtt (dismissed):
LGTM! Left a few suggestions for minor improvements, but I think this solves the problem already as-is.

test/e2e/tests/from-import-ui.spec.js
```diff
@@ -233,6 +233,14 @@ describe('Metamask Import UI', function () {
...
         await driver.clickElement({ text: 'Remove', tag: 'button' });
 
+        // Wait until selected account switches away from removed account to first account
+        await driver.waitForSelector({
+            css: '.selected-account__name',
+            text: 'Account 1'
+          },
+          { timeout: 10000 },
+        );
+
```
danjm, line 243: @Gudahtt I applied your suggestion here. The one concern I have about removing the `assert.equal(await currentActiveAccountName.getText(), 'Account 1');` line below is that I recall a convo with @PeterYinusa where he was saying how a lot of our e2e tests rely on implicit assertions, but it is better to be explicit so that we get clear error reports.

So having `assert.equal(await currentActiveAccountName.getText(), 'Account 1');` is good because if the account name doesn't equal that, the test will tell us what it was instead. Whereas just having the await for the selector will only give us an error message that the await timed out.

I almost want a `waitForSelector` method that, if timed out, include what the selector text was in the error message in the test output :thinking:

Gudahtt: I don't think this case is an implied assertion now that we have this `waitForSelector` step. There was one before (we were assuming the account would switch before we checked), but it's explicit now with the `waitForSelector` step. The comment helps make that clear as well.

But you are raising a legitimate problem, that the error message isn't very useful when that `wait` fails. Agreed that we should improve that. But these lines below remain redundant right now, because if they are false, the test will fail at the `waitForSelector`.

>I almost want a waitForSelector method that, if timed out, include what the selector text was in the error message in the test output thinking

Right now the text is part of the selector. If it times out waiting for a selector, it won't match anything. It doesn't know which element was supposed to have that text, so it can't tell you what the current text is.

We could create another method that has a separate selector and assertion though. Then it could find the element with the selector, assert that it matches the assertion, and explain in the error message what is different if it still doesn't match on the last attempt. That should work.

PeterYinusa (dismissed):

test/e2e/tests/from-import-ui.spec.js
danjm: Okay, I have removed the redundant code here 6409e3cc3

I have created an issue for improving on this helper method here: https://github.com/MetaMask/metamask-extension/issues/12635

Gudahtt (approved):
LGTM!

PeterYinusa (approved):

PR #12613, opened 2021-11-08
Throw, and not just display, ledgerDeviceOpenFailureMessage error

danjm:
This should have been part of 12604.

Instead of the call to `connectHardware` resolving and attempting to get accounts in `connect-hardware/index.js`, an error should be thrown and handled by the catch logic in that file.


ryanml (approved):

tmashuang (approved):
lgtm

PR #12621, opened 2021-11-08
Fix button-group.stories for new Storybook documentation

EtDu:
Includes DefaultStory syntax with variations like Disabled Button and Radio Button. Also includes MDX documentation.

### DIFF
--- a/ui/components/ui/button-group/README.mdx
+++ b/ui/components/ui/button-group/README.mdx
@@ -34,7 +34,7 @@ By changing the `disabled` value to true, the buttons inside the button group co
 
 Rendering radio type button instead
 
-```html
+```jsx
 <ButtonGroup variant="radiogroup">
   <button>cheap</button>
   <button>average</button>
--- a/ui/components/ui/button-group/README.mdx
+++ b/ui/components/ui/button-group/README.mdx
@@ -50,7 +50,7 @@ Rendering radio type button instead
 
 Rendering button group without active button
 
-```html
+```jsx
 <ButtonGroup noButtonActiveByDefault>
   <button>cheap</button>
   <button>average</button>
--- a/ui/components/ui/button-group/README.mdx
+++ b/ui/components/ui/button-group/README.mdx
@@ -14,20 +14,6 @@ Button Group is a wrapper for buttons to align them horizontally
 
 <ArgsTable of={ButtonGroup} />
 
-## Usage
-
-The following describes the props and example usage for this component.
-
-| Prop Type                  | Description                                                                         |
-| -------------------------- | ----------------------------------------------------------------------------------- |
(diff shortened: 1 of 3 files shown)


ui/components/ui/button-group/README.mdx
```diff
@@ -0,0 +1,77 @@
...
+
+| Prop Type                  | Description                                                                         |
+| -------------------------- | ----------------------------------------------------------------------------------- |
+| `defaultActiveButtonIndex` | change button active order                                                          |
+| `noButtonActiveByDefault`  | no buttons are active before clicked by the user                                     |
+| `disabled`                 | disable every button inside button group                                          |
+| `className`                | classname is used for children prop, changing it in storybook won't change anything |
+| `style`                    | add style for button group component                                             |
+| `newActiveButtonIndex`     | update value of active button in button group component                           |
+| `variant`                  | options for rendering types of buttons, consists of 'default' and 'radiogroup'         |
```
georgewrmarshall, line 29: This is awesome and I just realized if we just jsDocs formatting we can get the description to show up in the `ArgsTable` I'll create an issue to update the Button component and contribution docs to include this standard. This should encourage and help maintain good documentation practices. If you move all of your descriptions from here to `ui/components/ui/button-group/button-group.component.js`. You can remove the table and it's one less thing that will need to update if props change.

```
export default class ButtonGroup extends PureComponent {
  static propTypes = {
    /**
     * Change button active order
     */
    defaultActiveButtonIndex: PropTypes.number,
```
> Example of first prop description

How it looks in the `ArgsTable`

Screen Shot 2021-11-09 at 8 36 51 AM

georgewrmarshall (commented):
Fantastic work 💯 Just one little update that I think could improve our templates overall. But apart from that smashed it out of the park

ui/components/ui/button-group/button-group.component.js
```diff
@@ -4,13 +4,37 @@ import classnames from 'classnames';
...
+    /**
+     * disabling every button inside button group
+     */
     disabled: PropTypes.bool,
+    /**
+     * Children must be an array of button components
+     */
     children: PropTypes.array,
+    /**
+     * classname is used for children prop, changing it in storybook won't change anything
```
georgewrmarshall, line 24: `className` seems to add a class to the parent component? Is that not what you've experienced? Also not sure we should mention storybook anymore as it's inside of the component

```suggestion
     * Adds a className to the root div of the of the ButtonGroup component
```

Screen Shot 2021-11-10 at 12 02 42 PM

georgewrmarshall (commented):
Looking really good! Few small updates

georgewrmarshall (approved):
lgtm!

darkwing (approved):

PR #12627, opened 2021-11-08
Add support for eip-1559 on Trezor

danjm:
Blocked by https://github.com/MetaMask/eth-trezor-keyring/pull/107 and https://github.com/MetaMask/eth-trezor-keyring/pull/108

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -2377,7 +2377,7 @@ export default class MetamaskController extends EventEmitter {
     if (keyring.type === KEYRING_TYPES.TREZOR) {
       const model = keyring.getModel();
       return model === 'T';
-    };
+    }
     return true;
   }
 
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -2374,7 +2374,11 @@ export default class MetamaskController extends EventEmitter {
     const address =
       fromAddress || this.preferencesController.getSelectedAddress();
     const keyring = await this.keyringController.getKeyringForAccount(address);
-    return keyring.type !== KEYRING_TYPES.TREZOR;
+    if (keyring.type === KEYRING_TYPES.TREZOR) {
+      const model = keyring.getModel();
+      return model === 'T';
+    };
+    return true;
   }
 
   //=============================================================================
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -1706,6 +1706,11 @@ export default class MetamaskController extends EventEmitter {
     if (deviceName === DEVICE_NAMES.LATTICE) {
       keyring.appName = 'MetaMask';
     }
+    if (deviceName === 'trezor') {
+      const model = keyring.getModel();
+      this.appStateController.setTrezorModel(model);
+    }
+
     keyring.network = this.networkController.getProviderConfig().type;
 
     return keyring;
--- a/ui/pages/create-account/connect-hardware/account-list.js
+++ b/ui/pages/create-account/connect-hardware/account-list.js
@@ -61,10 +61,9 @@ class AccountList extends Component {
 
   renderHeader() {
     const { device } = this.props;
-    const shouldShowHDPaths =
-      device.toLowerCase() === 'ledger' ||
-      device.toLowerCase() === 'lattice' ||
-      device.toLowerCase() === 'trezor';
+    const shouldShowHDPaths = ['ledger', 'lattice', 'device'].includes(
+      device.toLowerCase(),
+    );
     return (
       <div className="hw-connect">
         <h3 className="hw-connect__unlock-title">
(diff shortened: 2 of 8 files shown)


ui/selectors/selectors.js
```diff
@@ -97,9 +97,12 @@ export function getParticipateInMetaMetrics(state) {
 }
 
 export function isEIP1559Account(state) {
-  // Trezor does not support 1559 at this time
-  const currentKeyring = getCurrentKeyring(state);
-  return currentKeyring && currentKeyring.type !== KEYRING_TYPES.TREZOR;
+  const keyring = getCurrentKeyring(state);
+  if (keyring.type === KEYRING_TYPES.TREZOR) {
+    const model = keyring.getModel();
```
darkwing, line 102: This call isn't going to the TrezorKeyring handle; `getCurrentKeyring` in this instance is returning:

```
{
  "type": "Trezor Hardware",
  "accounts": [
    "0x9546aa6dfabfb3974db931096a4e6f84600cdb73"
  ]
}
```

So the `getModel` call is erroring out.

ui/pages/create-account/connect-hardware/account-list.js
```diff
@@ -62,7 +62,9 @@ class AccountList extends Component {
   renderHeader() {
     const { device } = this.props;
     const shouldShowHDPaths =
-      device.toLowerCase() === 'ledger' || device.toLowerCase() === 'lattice';
+      device.toLowerCase() === 'ledger' ||
```
darkwing, line 65: We can probably change this to:

```
['ledger', 'lattice', 'device'].includes(device.toLowerCase())
```

darkwing:
I've analyzing this and summary includes:

The selector `isEIP1559Account` is problematic because `getCurrentKeyring` only provides an object with the `type` and `accounts`, not a handle to the keyring.  The `metamask-controller` change *does* get the keyring, since it's directly interfacing with said keyring.

Also, the Trezor keyring change which calls `getFeatures` can also be problematic because (1) it's an extra iframe open, regardless of when it's called and (2) calling it in the constructor aggressively opens windows way more often than any user would like.  

One idea to make this process better would be detecting the Trezor model type upon adding the account, and storing it in state.  That would work for new Trezor connections moving forward, but not for existing Trezor connections in MetaMask.  

One alternative could be to call `getFeatures()` if we don't have don't have the Trezor model in state, and then we'd not need to do it moving forward.  They downside of this approach could be our Trezor users seeing an un unexpected Trezor window open/close and the user not understanding why it happened or if something untrustworthy has happened. Possibly one way around this unexpected window is adding an Advanced Setting called "Use 1559 Transactions in Trezor", which when toggled, shows the popup and allows us to get the model.

Happy to talk more about this on Monday!

ui/selectors/selectors.js
danjm: I have addressed this in the latest commit 9a53bc270

ui/pages/create-account/connect-hardware/account-list.js
danjm: done in 385b6b713

danjm:
@darkwing With the latest changes on https://github.com/MetaMask/eth-trezor-keyring/pull/108 and on this branch, I have addressed the two issues you describe above.

darkwing (approved):

app/scripts/metamask-controller.js
```diff
@@ -1706,6 +1706,11 @@ export default class MetamaskController extends EventEmitter {
     if (deviceName === DEVICE_NAMES.LATTICE) {
       keyring.appName = 'MetaMask';
     }
+    if (deviceName === 'trezor') {
```
mcmire, line 1709: Should this be a constant?

ui/selectors/selectors.js
```diff
@@ -142,9 +142,12 @@ export function getParticipateInMetaMetrics(state) {
...
 export function isEIP1559Account(state) {
-  // Trezor does not support 1559 at this time
-  const currentKeyring = getCurrentKeyring(state);
-  return currentKeyring && currentKeyring.type !== KEYRING_TYPES.TREZOR;
+  const keyring = getCurrentKeyring(state);
+
+  if (keyring?.type === KEYRING_TYPES.TREZOR) {
+    return state.metamask.trezorModel === 'T';
+  }
+  return true;
```
mcmire, line 150: [nit] A more succinct and possibly more readable way to write this:

``` suggestion
  return keyring == null ||
    keyring.type !== KEYRING_TYPES.TREZOR ||
    state.metamask.trezorModel === 'T'
```

app/scripts/metamask-controller.js
```diff
@@ -2374,7 +2379,11 @@ export default class MetamaskController extends EventEmitter {
     const address =
       fromAddress || this.preferencesController.getSelectedAddress();
     const keyring = await this.keyringController.getKeyringForAccount(address);
-    return keyring.type !== KEYRING_TYPES.TREZOR;
+    if (keyring.type === KEYRING_TYPES.TREZOR) {
+      const model = keyring.getModel();
+      return model === 'T';
+    }
+    return true;
```
mcmire, line 2386: [nit] A more succinct and possibly more readable way to write this:

``` suggestion
    keyring.type !== KEYRING_TYPES.TREZOR || keyring.getModel() === 'T'
```

mcmire (approved):
Have a few nits, but nothing blocking this.

Issue #7135, opened 2019-09-09
MM assumes transaction was confirmed when it first sees the confirmation, not when it was actually confirmed


leafcutterant:
**Describe the bug**
See title.

**To Reproduce**
1. Submit tx such that it doesn't confirm in the next block
2. Log out of MM or exit browser
3. Wait until tx is confirmed (i.e. included in a valid block) (let's call this time t0)
4. Wait some more time
5. Open browser and/or log into MM (let's call this t1)
6. The tx cell says the tx was confirmed at t1

**Expected behavior**
The tx cell should say the tx was confirmed at t0.

**Browser details (please complete the following information):**
 - Browser: Firefox
 - MetaMask Version 7.1.1

danjm:
To recreate:

1. Go through the MetaMask send flow to create and send a transaction. Do this on Ropsten or Rinkeby, not localhost.
2. As soon as the transaction is created (and in a pending state) lock MetaMask and close your browser
3. wait 4-5 minutes
4. Open your browser and log back in to MetaMask
5. Open the transaction in the transaction list
6. It will say "transaction confirmed at 8:00" (or whatever time you opened MetaMask

If you view the transaction on etherscan, you will see that it was actually confirmed at an earlier time

![Screenshot from 2021-10-18 08-01-24](https://user-images.githubusercontent.com/7499938/137715059-d44bcb3b-9f9b-4595-86ae-76162a0b8364.png)

igorms-cons:
More information will be detailled in the next PR but I can say:
Timestamp transaction on etherscan is **only** on the moment the transaction is mined. There (I couldnt find) is no time when all transactions are done.

PR #12633, opened 2021-11-09
Fix confirmed transaction time 

igorms-cons:
Fixes: 
 Update with the timestamp fetched from Etherscan api 

**Explanation:**  

1. All transactions (current, pending, confirmes, etc.) are saved to the state.  Each new entry has a timestamp assigned to Date.now and they will be used to create/fill  an array  of history of transactions.
2. This timestamp is the one used when we display the time on activities logs.
3. Meaning, if we closed the browser and re-open 20min later, the time will be set as mentioned in bullet 1
4. All TXs info are fetched from  our etherscan api call. The timestamp this call api is giving us is the time mined and ofc the one displayed on etherscan view

**Manual testing steps:**  
  - Create a transaction
  - Close the browser
  - Wait 2 or 3min
  - Open browser and metamask, log
  - You will see notification about TX confirmation
  - Click  on the TX, and you will see the time of the confirmed transaction is the one from  etherscan view.
  
 **Breaking:**
- The displayed calculated confirmation time is now the time where the transaction was  mined. (from etherscan)
 
  Issue: https://github.com/MetaMask/metamask-extension/issues/7135

### DIFF
--- a/app/scripts/controllers/transactions/index.js
+++ b/app/scripts/controllers/transactions/index.js
@@ -1011,7 +1011,7 @@ export default class TransactionController extends EventEmitter {
 
       if (submittedTime) {
         metricsParams.completion_time = this._getTransactionCompletionTime(
-          submittedTime,
+          txMeta.time,
         );
       }
 
@@ -1465,8 +1465,8 @@ export default class TransactionController extends EventEmitter {
     });
   }
 
-  _getTransactionCompletionTime(submittedTime) {
-    return Math.round((Date.now() - submittedTime) / 1000).toString();
+  _getTransactionCompletionTime(time) {
+    return Math.round(time / 1000).toString();
   }
 
   _getGasValuesInGWEI(gasParams) {
--- a/app/scripts/controllers/transactions/index.js
+++ b/app/scripts/controllers/transactions/index.js
@@ -977,7 +977,7 @@ export default class TransactionController extends EventEmitter {
    * @param {number} txId - The tx's ID
    * @returns {Promise<void>}
    */
-  async confirmTransaction(txId, txReceipt, baseFeePerGas) {
+  async confirmTransaction(txId, txReceipt, baseFeePerGas, blockTimestamp) {
     // get the txReceipt before marking the transaction confirmed
     // to ensure the receipt is gotten before the ui revives the tx
     const txMeta = this.txStateManager.getTransaction(txId);
@@ -1002,6 +1002,9 @@ export default class TransactionController extends EventEmitter {
       if (baseFeePerGas) {
         txMeta.baseFeePerGas = baseFeePerGas;
       }
+      if (blockTimestamp) {
+        txMeta.blockTimestamp = blockTimestamp;
+      }
 
       this.txStateManager.setTxStatusConfirmed(txId);
       this._markNonceDuplicatesDropped(txId);
@@ -1011,7 +1014,7 @@ export default class TransactionController extends EventEmitter {
 
       if (submittedTime) {
         metricsParams.completion_time = this._getTransactionCompletionTime(
-          txMeta.time,
+          submittedTime,
         );
       }
 
@@ -1183,8 +1186,13 @@ export default class TransactionController extends EventEmitter {
     });
     this.pendingTxTracker.on(
       'tx:confirmed',
-      (txId, transactionReceipt, baseFeePerGas) =>
-        this.confirmTransaction(txId, transactionReceipt, baseFeePerGas),
+      (txId, transactionReceipt, baseFeePerGas, blockTimestamp) =>
+        this.confirmTransaction(
+          txId,
+          transactionReceipt,
+          baseFeePerGas,
+          blockTimestamp,
+        ),
     );
     this.pendingTxTracker.on('tx:dropped', (txId) => {
       this._dropTransaction(txId);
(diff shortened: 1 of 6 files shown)


igorms-cons:
Please read my explanation and comment on the issue to be able to know what is going on.
This PR will avoid wrong confirmation time if the browser is closed and open later BUT the calculation between submitted and confirmed that transaction is gone (`submittedtime - date.now()`).

Best scenario will be: 
- if browser is open =>  `submittedtime - date.now()`)
- if not, when open => display the mined `timestamp` from etherscan
The Transaction polling is only running when browser is open. I dont know if it is possible  to implement the above scenario.

mhasanagic:
Verified by QA - the time of the confirmed transaction is the one from Etherscan view

app/scripts/controllers/transactions/index.js
```diff
@@ -1003,7 +1003,7 @@ export default class TransactionController extends EventEmitter {
 
       if (submittedTime) {
         metricsParams.completion_time = this._getTransactionCompletionTime(
-          submittedTime,
+          txMeta.time,
```
danjm, line 1006: `txMeta.time` is the time the transaction is created

The issue here is that we need to show the actual time the transaction is confirmed

app/scripts/controllers/transactions/index.js
```diff
@@ -1457,8 +1457,8 @@ export default class TransactionController extends EventEmitter {
     });
   }
 
-  _getTransactionCompletionTime(submittedTime) {
-    return Math.round((Date.now() - submittedTime) / 1000).toString();
+  _getTransactionCompletionTime(time) {
```
danjm, line 1460: `metricsParams.completion_time` is meant to track the amount of time it takes for a transaction to go from being in the submitted state to being confirmed. This change would instead make it simply track the time the transaction was created.

app/scripts/controllers/transactions/index.js
danjm: unless... does `txMeta.time` get updated anywhere?

danjm: no, I don't think so

app/scripts/controllers/transactions/lib/tx-state-history-helpers.js
```diff
@@ -35,11 +35,16 @@ export function generateHistoryEntry(previousState, newState, note) {
...
   // Add a note to the first op, since it breaks if we append it to the entry
   if (entry[0]) {
+    entry[0].timestamp = Date.now();
     if (note) {
       entry[0].note = note;
+      if (
+        note.toLowerCase().includes('setting status to confirmed') ||
+        note.toLowerCase().includes('transactions#confirmtransaction')
+      ) {
+        entry[0].timestamp = newState.time;
```
danjm, line 45: I think this will work if the transaction is mined within the same minute that it is created within MetaMask, but I don't think it will work if the mining happens significantly later. This is hard to test on test networks, where the mining time is short.

danjm (changes requested):

danjm:
I have a suggested alternative solution:

- in the `_checkPendingTx` function in `pending-tx-tracker.js`, `getBlockByHash` is called after the transaction is confirmed
- the result of calling `getBlockByHash` will include a timestamp, this timestamp can be used to show the actual confirmation time
- that timestamp should be added to txMeta in the `confirmTransaction` method within the transaction controller. It would be a new property, `blockTimestamp`
- somehow that property has to be used in `ui/components/app/transaction-activity-log/transaction-activity-log.util.js`, so that the correct time is shown on the confirm step

app/scripts/controllers/transactions/index.js
igorms-cons: nope, txMeta.time is the timestamp from etherscan when TX  starts mining.

app/scripts/controllers/transactions/index.js
igorms-cons: yes

app/scripts/controllers/transactions/lib/tx-state-history-helpers.js
igorms-cons: I see mmmh

igorms-cons:
@danjm,
Thanks again for the suggestive solution. I have implemented it & here are my concerns before I commit my code:
 - the `timestamp` value from `getBlockByHash` result data is the same datetime as the one mined (from etherscan)
 - Even If we put a `date.now` as  a timestamp after the  `getBlockByHash`,  it duplicates the previous  confirmation time calculation  present  `controllers/transactions/index.js/_getTransactionCompletionTime`  and the issue is still there: 
         - if MM is open no issue **but** I soon as I close all my browser after I sent the transaction and open MM few minutes later; the confirmation date will be (date.now). Why? because `_checkPendingTx` is called only when the browser is open.

mhasanagic:
Verified by QA

david0xd (dismissed):

ui/components/app/transaction-activity-log/transaction-activity-log.util.js
```diff
@@ -189,7 +189,16 @@ export function getActivities(transaction, isFirstTransaction = false) {
                   gasPrice: cachedGasPrice,
                 });
               }
+              break;
+            }
 
+            case BLOCKTIMESTAMP: {
+              const filteredAcc = acc.filter(
```
darkwing, line 196: We can do `acc.find` so that you don't need to reference `[0]` below

darkwing (changes requested):

danjm (approved):

segun (approved):

darkwing (approved):

PR #12638, opened 2021-11-10
Add platform-specific build type manifest modifications

Gudahtt:
The build system now supports platform-specific modifications to the manifest for each build type. The need to customize the `id` on Firefox motivated this change.

To support this, a new directory was made in each build type directory for manifest changes. The images currently in this directory were moved into an `images` subdirectory.

This new `manifest` directory can include each manifest file currently in `app/manifest`. The `_base.json` file is assumed to exist, but the platform manifest modifications are optional.

Manual testing steps:  
  - Create a non-main build (e.g. `yarn build dev --build-type beta`)
  - Ensure the assets are correct, and that the correct custom ID was set in the Firefox manifest.


darkwing (approved):
I love this!

rekmarks (approved):
LGTM!

Issue #10085, opened 2020-12-16
 Already processing eth_requestAccounts


tre-dev:
**Describe the bug**

When I close the unlock wallet popup while requesting eth accounts, I'm not getting any result. And when I request eth accounts again, I'm getting the following error `MetaMask - RPC Error: Already processing eth_requestAccounts. Please wait.`

**Steps to reproduce (REQUIRED)**

```
 window.web3 = new Web3(window.ethereum)
 const account = await window.ethereum.request({
      method: 'eth_requestAccounts',
 })
```

- Have wallet locked
- Trigger request
- Close unlock popup
- Trigger request again

**Expected behavior**

Receive error with specific code to know that unlock wallet has been closed as well as the transaction to be discharged from the processing pipe.

**Browser details (please complete the following information):**

Version 1.17.75 Chromium: 87.0.4280.88 (Official Build) (x86_64)

tmashuang:
Is the notification number on the icon not present? After unlocking, does the pending connect request not present?

Gudahtt:
Thanks for the report! I was able to reproduce this behaviour.

We did intend to block repeated attempts to connect, so this is working as intended. But, I hadn't anticipated that this would prevent the popup from showing up if it was accidentally closed the first time :thinking:. Ideally a user would click the MetaMask button in their toolbar to unlock, but admittedly not all of our users are in the habit of clicking that button, and some have it hidden.

This definitely seems less than ideal, but I'm not sure how we should handle this.

>Receive error with specific code to know that unlock wallet has been closed as well as the transaction to be discharged from the processing pipe.

The _first_ request to connect should return a Promise to the dapp that is still pending; that was meant to be our signal to dapps that the request is still being processed. The thought was that the _reason_ for the delay wasn't necessarily of concern to the dapp. The only thing the dapp needs to know is that it's still awaiting approval. In an ideal world, anyway. Note that repeated connect attempts will result in this error even when the extension is unlocked.

Note that dapps can check whether MetaMask is unlocked using the [experimental provider method `isUnlocked`](https://docs.metamask.io/guide/ethereum-provider.html#ethereum-metamask-isunlocked), but be warned that we don't provide any stability guarantees for that method.

Returning an error that explains that the extension is locked is an interesting idea. Or it might be worth trying to re-trigger the popup for repeated connect requests while unlocked, if the popup is currently closed. Would making the popup re-appear in this case solve this problem from your perspective @tre-dev ?

[...]

PR #12643, opened 2021-11-10
Reject popup confirmations on close

### LINKED ISSUE #10302: Reject all pending confirmations when notification window closed
We should reject all pending confirmations when the notification window is closed.

We already consider closing the notification window to be a rejection of the confirmation currently being shown. However if the user has multiple pending confirmations, closing the notification window only rejects the first one. This can be confusing the next time the user triggers a new confirmation, because they end up getting presented with the old one instead. This can also be confusing when the user closes the notification window on the lock screen, which currently leaves any confirmations pending without rejecting anything.

Rejecting all pending confirmations upon closing the notification window would further cement the idea that closing the window is a rejection, and it would help prevent the build-up of stale confirmations that are no longer of interest to the user.

ritave:
Fixes: #10302

### DIFF
--- a/app/scripts/background.js
+++ b/app/scripts/background.js
@@ -487,13 +487,32 @@ function setupController(initState, initLangCode) {
     // Metamask can open a popup for a login, we need to check if
     // we're on a confirmation page
     if (controller.isUnlocked()) {
-      controller.txController.txStateManager.clearUnapprovedTxs();
-      controller.messageManager.clearUnapproved();
-      controller.personalMessageManager.clearUnapproved();
-      controller.decryptMessageManager.clearUnapproved();
-      controller.encryptionPublicKeyManager.clearUnapproved();
-      controller.typedMessageManager.clearUnapproved();
-      controller.approvalController.clear();
+      /*
+      controller.txController.txStateManager.store.getState().transactions.filter((tx) => tx.status === "unapproved")
+      controller.txController.txStateManager
+        .getUnapprovedTxList()*/
+      Object.keys(controller.txController.txStateManager
+        .getUnapprovedTxList())
+        .forEach((txId) => controller.txController.txStateManager.setTxStatusRejected(txId));
+      controller.messageManager.messages
+        .filter((msg) => msg.status === 'unapproved')
+        .forEach((tx) => controller.messageManager.rejectMsg(tx.id));
+      controller.personalMessageManager.messages
+        .filter((msg) => msg.status === 'unapproved')
+        .forEach((tx) => controller.personalMessageManager.rejectMsg(tx.id));
+      controller.decryptMessageManager.messages
+        .filter((msg) => msg.status === 'unapproved')
+        .forEach((tx) => controller.decryptMessageManager.rejectMsg(tx.id));
+      controller.encryptionPublicKeyManager.messages
+        .filter((msg) => msg.status === 'unapproved')
+        .forEach((tx) => controller.decryptMessageManager.rejectMsg(tx.id));
+      controller.typedMessageManager.messages
+        .filter((msg) => msg.status === 'unapproved')
+        .forEach((tx) => controller.decryptMessageManager.rejectMsg(tx.id));
+      Object.keys(controller.approvalController.state.pendingApprovals)
+        .forEach((approvalId) => controller.approvalController.reject(approvalId, new Error()));
+
+
       updateBadge();
     }
   }
--- a/app/scripts/background.js
+++ b/app/scripts/background.js
@@ -17,7 +17,10 @@ import {
   ENVIRONMENT_TYPE_FULLSCREEN,
 } from '../../shared/constants/app';
 import { SECOND } from '../../shared/constants/time';
-import { REJECT_NOTFICIATION_CLOSE, REJECT_NOTFICIATION_CLOSE_SIG} from '../../shared/constants/metametrics';
+import {
+  REJECT_NOTFICIATION_CLOSE,
+  REJECT_NOTFICIATION_CLOSE_SIG,
+} from '../../shared/constants/metametrics';
 import migrations from './migrations';
 import Migrator from './lib/migrator';
 import ExtensionPlatform from './platforms/extension';
@@ -494,16 +497,31 @@ function setupController(initState, initLangCode) {
       );
       controller.messageManager.messages
         .filter((msg) => msg.status === 'unapproved')
-        .forEach((tx) => controller.messageManager.rejectMsg(tx.id, REJECT_NOTFICIATION_CLOSE_SIG));
+        .forEach((tx) =>
+          controller.messageManager.rejectMsg(
+            tx.id,
+            REJECT_NOTFICIATION_CLOSE_SIG,
+          ),
+        );
       controller.personalMessageManager.messages
         .filter((msg) => msg.status === 'unapproved')
         .forEach((tx) => controller.personalMessageManager.rejectMsg(tx.id));
       controller.decryptMessageManager.messages
         .filter((msg) => msg.status === 'unapproved')
-        .forEach((tx) => controller.decryptMessageManager.rejectMsg(tx.id, REJECT_NOTFICIATION_CLOSE));
+        .forEach((tx) =>
+          controller.decryptMessageManager.rejectMsg(
+            tx.id,
+            REJECT_NOTFICIATION_CLOSE,
+          ),
+        );
       controller.encryptionPublicKeyManager.messages
         .filter((msg) => msg.status === 'unapproved')
-        .forEach((tx) => controller.encryptionPublicKeyManager.rejectMsg(tx.id, REJECT_NOTFICIATION_CLOSE));
+        .forEach((tx) =>
+          controller.encryptionPublicKeyManager.rejectMsg(
+            tx.id,
+            REJECT_NOTFICIATION_CLOSE,
+          ),
+        );
       controller.typedMessageManager.messages
         .filter((msg) => msg.status === 'unapproved')
         .forEach((tx) => controller.typedMessageManager.rejectMsg(tx.id));
--- a/app/scripts/background.js
+++ b/app/scripts/background.js
@@ -505,7 +505,20 @@ function setupController(initState, initLangCode) {
         );
       controller.personalMessageManager.messages
         .filter((msg) => msg.status === 'unapproved')
-        .forEach((tx) => controller.personalMessageManager.rejectMsg(tx.id));
+        .forEach((tx) =>
+          controller.personalMessageManager.rejectMsg(
+            tx.id,
+            REJECT_NOTFICIATION_CLOSE_SIG,
+          ),
+        );
+      controller.typedMessageManager.messages
+        .filter((msg) => msg.status === 'unapproved')
+        .forEach((tx) =>
+          controller.typedMessageManager.rejectMsg(
+            tx.id,
+            REJECT_NOTFICIATION_CLOSE_SIG,
+          ),
+        );
       controller.decryptMessageManager.messages
         .filter((msg) => msg.status === 'unapproved')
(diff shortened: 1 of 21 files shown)


ritave:

ui/pages/settings/advanced-tab/advanced-tab.container.js
```diff
@@ -12,7 +12,6 @@ import {
   turnThreeBoxSyncingOnAndInitialize,
   setUseNonceField,
   setIpfsGateway,
-  setLedgerTransportPreference,
```
rekmarks, line 15: The changes in this file seem unrelated to the PR.

rekmarks:
@Gudahtt if all pending confirmations are rejected on notification popup close, how should we handle metrics? The `ApprovalController` probably needs a controller messenger event for this purpose, given the current implementation.

Gudahtt:
Good point.

Right now our "on reject" metrics are in the UI. We can move them into the `tx-state-manager` to preserve them with this approach. And similarly for the "message manager"-style classes, we can have them dispatch metrics on reject as well. The state manager for the message seems like a better place than the UI to dispatch such a metric anyway.

As for the confirmations managed by the approval controller, I am not sure why we'd want to dispatch metrics directly in that controller. It doesn't know what these approvals are for, it doesn't have the context. The one controller using that right now is the permissions controller, so to take that as an example, we could dispatch the rejection metric event in `rejectPermissionsRequest`. That's where we have all of the context about what is being rejected.

ui/pages/confirmation/confirmation.js
```diff
@@ -29,9 +29,9 @@ import { getTemplateValues, getTemplateAlerts } from './templates';
  * immutable and painless. This state is not stored in redux state because it
  * should persist only for the lifespan of the current session, and will only
  * be used on this page. Dismissing alerts for confirmations should persist
- * while the user pages back and forth between confirmations. However, if the
- * user closes the confirmation window and later reopens the extension they
- * should be displayed the alerts again.
+ * while the user pages back and forth between confirmations. If the user
```
Gudahtt, line 32: The alerts discussed here are not the same as the confirmations we're rejecting. These should remain unaffected by the changes in this PR, so this comment change is not needed.

ui/components/app/signature-request/signature-request.component.js
```diff
@@ -34,10 +33,8 @@ export default class SignatureRequest extends PureComponent {
     }
   }
 
-  _beforeUnload = (event) => {
+  _beforeUnload = (_) => {
```
Gudahtt, line 36: Nit: This argument can be omitted entirely since we're not using it

```suggestion
  _beforeUnload = () => {
```

ui/components/app/signature-request/signature-request.component.js
```diff
@@ -17,7 +17,6 @@ export default class SignatureRequest extends PureComponent {
       name: PropTypes.string,
     }).isRequired,
     isLedgerWallet: PropTypes.bool,
-    clearConfirmTransaction: PropTypes.func.isRequired,
```
Gudahtt, line 20: This prop should be removed in `signature-request.container.js` as well, since it's no longer needed. The entire `mapDispatchToProps` function can be omitted, since this is the only function it returns.

ui/components/app/signature-request-original/signature-request-original.component.js
```diff
@@ -55,8 +55,7 @@ export default class SignatureRequestOriginal extends Component {
     this._removeBeforeUnload();
   };
 
-  _beforeUnload = (event) => {
-    const { clearConfirmTransaction, cancel } = this.props;
+  _beforeUnload = (_) => {
```
Gudahtt, line 58: Nit: This argument can be omitted
```suggestion
  _beforeUnload = () => {
```

ui/components/app/dropdowns/network-dropdown.js
```diff
@@ -68,7 +68,6 @@ function mapDispatchToProps(dispatch) {
         }),
       );
     },
-    hideTestNetMessage: () => dispatch(actions.hideTestNetMessage()),
```
Gudahtt, line 71: Nit: This also seems unrelated to the other changes in this PR

app/scripts/lib/notification-manager.js
```diff
@@ -12,7 +18,9 @@ export default class NotificationManager {
    */
 
   constructor() {
+    super();
     this.platform = new ExtensionPlatform();
+    extension.windows.onRemoved.addListener(this._onWindowClosed.bind(this));
```
Gudahtt, line 23: Nit: We've been using the `ExtensionPlatform` module as a way to encapsulate our extension API interactions, like this one. In keeping with that, this might be better off as an `ExtensionPlatform` method. Something like `addOnRemovedListener`.

Not a blocker, as we aren't even consistently doing that yet today (we use `extensionizer` directly in a few places). But since `this.platform` is already here it might be easy enough.

ui/pages/confirm-encryption-public-key/confirm-encryption-public-key.component.js
```diff
@@ -46,21 +46,8 @@ export default class ConfirmEncryptionPublicKey extends Component {
...
 
   _beforeUnload = async (event) => {
-    const {
-      clearConfirmTransaction,
-      cancelEncryptionPublicKey,
-      txData,
-    } = this.props;
-    const { metricsEvent } = this.context;
+    const { cancelEncryptionPublicKey, txData } = this.props;
     await cancelEncryptionPublicKey(txData, event);
```
Gudahtt, line 50: `cancelEncryptionPublicKey` results in the message being rejected in the background. Since we're doing this automatically on window close now, we don't need to call this anymore.

The same goes for `cancelDecryptMessage` as well, in the `confirm-decrypt-message.component.js` beforeunload handler. Both of these beforeunload handlers are no longer needed.

app/scripts/lib/message-manager.js
```diff
@@ -218,6 +219,13 @@ export default class MessageManager extends EventEmitter {
    *
    */
   rejectMsg(msgId) {
+    this.metricsEvent({
+      event: 'Reject Sig Request',
```
Gudahtt, line 223: This property corresponds to the "name" option passed into `metricsEvent` in the front end metric event this replaces.

The value for that event was this:

```suggestion
      event: 'Cancel Sig Request Via Notification Close',
```

We should try not to change the event details so that we don't break any reports built on the existing events.

Gudahtt: It looks like some of the other event names have changed as well. We'll want to ensure they all match what they were before.

app/scripts/background.js
```diff
@@ -475,6 +477,45 @@ function setupController(initState, initLangCode) {
...
+      controller.encryptionPublicKeyManager.messages
+        .filter((msg) => msg.status === 'unapproved')
+        .forEach((tx) => controller.decryptMessageManager.rejectMsg(tx.id));
+      controller.typedMessageManager.messages
+        .filter((msg) => msg.status === 'unapproved')
+        .forEach((tx) => controller.decryptMessageManager.rejectMsg(tx.id));
+      Object.keys(
+        controller.approvalController.state.pendingApprovals,
+      ).forEach((approvalId) =>
+        controller.approvalController.reject(approvalId, new Error()),
```
Gudahtt, line 512: This empty error isn't very useful in describing why the request was rejected. This error gets returned to sites awaiting on a response, so we should ensure it matches the provider error spec and that it's informative.

Should we be using the approval controller directly here? (cc @rekmarks ) Right now we're only using this for permission requests, so we could reject each permission request instead via `permissionController.rejectPermissionRequest`, which would set the appropriate error.

In general I don't imagine we'd want to use the approval controller directly, because it's too generic. The controllers making use of it have more context for how to handle rejections and confirmations.

app/scripts/lib/message-manager.js
ritave: The name of the metric changed because the behaviour changed, it will be fired in more cases and doesn’t represent the same thing happening anymore

For example, it will be fired on pressing cancel button, not only on window close

app/scripts/background.js
ritave: I'll change it to permissionController for now

app/scripts/lib/message-manager.js
```diff
@@ -217,7 +218,16 @@ export default class MessageManager extends EventEmitter {
    * @param {number} msgId - The id of the Message to reject.
    *
    */
-  rejectMsg(msgId) {
+  rejectMsg(msgId, reason = undefined) {
+    if (reason) {
+      this.metricsEvent({
```
Gudahtt, line 223: We will need to replicate this pattern in the other signature message managers as well (`personalMessageManager` and `typedMessageManager`). All three of these message managers are used by the signature request confirmations.

ui/components/app/signature-request/signature-request.component.js
```diff
@@ -28,33 +25,6 @@ export default class SignatureRequest extends PureComponent {
...
-      txData: { type },
-    } = this.props;
-    const { metricsEvent } = this.context;
-    metricsEvent({
-      eventOpts: {
-        category: 'Transactions',
-        action: 'Sign Request',
-        name: 'Cancel Sig Request Via Notification Close',
-      },
-      customVariables: {
```
Gudahtt, line 50: Note this extra data  here. This component (`signature-request.component.js`) is used for `eth_signTypedData_v3` and `eth_signTypedData_v4`). So this data should be added in the `typedMessageManager` specifically for types V3 and V4, and not for type V1.

Though honestly the extra data couldn't hurt :thinking: Maybe we can add it for all types. I highly doubt any report is counting on it _not_ being set. These metric events are likely to be replaced with more sensible ones in the not too distant future anyway.

Gudahtt (commented):
This mostly looks good! One more set of changes left, for the remaining message managers.

ui/components/app/signature-request/signature-request.component.js
Gudahtt: It looks like this custom variable is still missing from the `typedMessageManager` metric event. We need it to be present at least for v3 and v4 messages, because this is how we distinguish between them on our metrics dashboard.

Gudahtt: Oh, it looks like I was wrong. This `type` doesn't represent the version at all. It just says that it's a signed type message :disappointed: . Oh well. Sorry about that mixup. I see that you've added it to the metric events already by now though, and that seems useful regardless.

app/scripts/background.js
```diff
@@ -475,6 +481,73 @@ function setupController(initState, initLangCode) {
...
 
+  notificationManager.on(
+    NOTIFICATION_MANAGER_EVENTS.POPUP_CLOSED,
+    rejectUnapprovedNotifications,
+  );
+
+  function rejectUnapprovedNotifications() {
+    // Metamask can open a popup for a login, we need to check if
+    // we're on a confirmation page
+    if (controller.isUnlocked()) {
```
Gudahtt, line 492: I think we want to reject all confirmations even if the wallet is locked. This is critical for ensuring subsequent `request_account` requests trigger the popup to open.

Issue #10085 is what prompted me to write up the issue this PR is addressing, and that case is about users getting stuck after closing the popup window once while the wallet was locked. Once they do that, subsequent attempts by the site to request access to accounts were rejected because we're already processing that request. Interpreting the closing of the popup as a rejection ensures that the next attempt will open the popup again.

It might be unexpected for a user if this ends up rejecting something unexpectedly that they haven't even seen yet. Maybe there's something we can do to guard against that mistake. I don't really understand why a user would choose not to unlock the wallet in this case though, especially if they've just done something that triggered a confirmation that is important to them.

Gudahtt (approved):
LGTM!

danjm (approved):

PR #12647, opened 2021-11-10
Add github action for shellcheck

rickycodes:


darkwing:
@rickycodes This is still in draft but tests are passing -- are we pushing this forward?

rickycodes:
@darkwing this is still missing the linting of `package.json` scripts currently. i think for now i will close until we can sort that out.

PR #12063, opened 2021-09-10
Add  Euskara / Basque language

aitorla:
Fixes: #

Explanation:  

Manual testing steps:  
  - 
  - 
  -

PR #12653, opened 2021-11-10
[i18n:eu] Add Euskara language support

darkwing:
Superecedes https://github.com/MetaMask/metamask-extension/pull/12063 to fix lint.


kumavis (dismissed):

mcmire:
Should probably wait until #13206 is merged before merging this.

mcmire:
@jpuri I think it's okay to have languages not supported by Crowdin. That means that the community will have to maintain them, but I think that's fine. Also, this language shouldn't conflict with any PRs Crowdin gives us. Agree/disagree?

brad-decker (approved):

PR #12657, opened 2021-11-11
shared/modules/hexstring-utils - replace ethereumjs-util with hexstring-utils

kumavis:
progress towards [reduce cryptography dependencies bundled in ui](https://github.com/MetaMask/metamask-extension/pull/11698)

replaces all our `ethereumjs-util` imports with our own hexstring util built on top. this is intended to make it easier to remove `ethereumjs-util` in the frontend


kumavis:
main issue is they take in all of ethereum-cryptography, which we dont need in the ui https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/util/package.json#L41

mcmire:
Hey @kumavis, do you think this is something we can take over from you? Sounds like this would be a beneficial change. Would you have any time next week to meet about this?

kumavis:
going to abandon this in the face of manifest v3 roadmap

PR #12675, opened 2021-11-12
Add migration to set showTestNetworks to true if there is evidence of testnet use

danjm:
Addresses this feedback from 10.6.0 QA:

> When on one of the test networks before the update, the network will not show up on the network list after updating. This is just more of an observation/documenting of an odd behavior. There is the banner that notifies and navigates to the test network setting change to bring back the networks. Also, it doesn't impede on switching networks. Maybe it would be beneficial to conditionally set the setting if the user is currently on a test network for the update.

This migrates the users data so that `showTestNetworks` is set to true if we have evidence they have used a test network

What counts as "evidence":
- if the user is currently on a test network
- if the user has sent a transaction on
- if the user has a "cached balance" on a test network, which means that the user has an account with a greater than 0 balance on a test network, and they have selected that account-network combination within MetaMask in the past

### DIFF
--- a/app/scripts/migrations/067.js
+++ b/app/scripts/migrations/067.js
@@ -24,45 +24,43 @@ export default {
 };
 
 function transformState(state) {
-  let newState = state;
-
   const PreferencesController = state?.PreferencesController || {};
   const preferences = PreferencesController.preferences || {};
 
-  if (!preferences.showTestNetworks) {
-    const transactions = state?.TransactionController?.transactions || {};
-    const provider = state.NetworkController?.provider || {};
-    const cachedBalances = state.CachedBalancesController?.cachedBalances || {};
-
-    const userIsCurrentlyOnATestNet = TEST_CHAINS.includes(provider?.chainId);
-    const userHasMadeATestNetTransaction = Object.values(
-      transactions,
-    ).some(({ chainId }) => TEST_CHAINS.includes(chainId));
-    const userHasACachedBalanceOnATestnet = TEST_CHAINS.some((chainId) => {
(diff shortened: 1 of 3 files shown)


app/scripts/migrations/067.js
```diff
@@ -0,0 +1,68 @@
...
+    const state = versionedData.data;
+    const newState = transformState(state);
+    versionedData.data = newState;
+    return versionedData;
+  },
+};
+
+function transformState(state) {
+  let newState = state;
+  console.log('state', state);
```
ryanml, line 28: nit: leftover from debug perhaps?

ryanml (dismissed):

app/scripts/migrations/067.js
```diff
@@ -0,0 +1,68 @@
...
+  },
+};
+
+function transformState(state) {
+  let newState = state;
+
+  const PreferencesController = state?.PreferencesController || {};
+  const preferences = PreferencesController.preferences || {};
+
+  if (!preferences.showTestNetworks) {
```
darkwing, line 32: Instead of this huge conditional, can we do `if(preferences.showTestNetworks) { return state }`?

tmashuang (dismissed):
lgtm

app/scripts/migrations/067.js
danjm: done in 6c9becbfa

app/scripts/migrations/067.js
danjm: addressed in a4a8d4410

ryanml (approved):

darkwing (approved):

PR #12676, opened 2021-11-12
Add support for Smart Transactions

dan437:
## Description
This PR adds support for Smart Transactions in Swaps on the Ethereum network (Mainnet and Rinkeby). They simulate transactions before they reach public blockchain in order to decrease transaction costs, reduce failures, etc.

## Pending items
- [x] Fix unit tests
- [x] Add new unit tests
- [x] Resolve ESLint issues
- [x] Publish the latest changes of the STX Controller, update package.json here to use it
- [x] Load STX deadline for pending transactions from backend
- [x] Check why we see a big gap between estimated and max fee on Rinkeby
- [x] Animate background .svgs on the status page
- [x] Show an animation before gas fees are loaded for STX
- [x] Show a notification to a user when a fallback to regular txs happened
- [x] Update UI + content with its final version once we have it (now it's about 95% done)

## Testing
You can just get this branch, run it locally, click on "Swap" on the homepage and you will see a modal window for Smart Transactions. Once you enable them, you can test them out!

## Screenshots
One time modal for Smart Transactions opt-in on the Build Quote page:
image

On and Off switch for Smart Transactions on the Build Quote page:
image

STX Status Page - Pending:
image

STX Status Page - Success:
image


package.json
```diff
@@ -114,6 +114,7 @@
     "@metamask/obs-store": "^5.0.0",
     "@metamask/post-message-stream": "^4.0.0",
     "@metamask/providers": "^8.1.1",
+    "@metamask/smart-transactions-controller": "file:./../smart-transactions-controller/metamask-smart-transactions-controller-v0.0.1.tgz",
```
dan437, line 117: We need to publish latest changes of the STX Controller and then we can use the `@metamask/smart-transactions-controller` node module instead of a .tgz file.

ui/ducks/swaps/swaps.js
```diff
@@ -645,6 +723,90 @@ export const fetchQuotesAndSetQuoteState = (
...
+          destinationTokenDecimals,
+          destinationTokenSymbol,
+          sourceTokenSymbol,
+          swapMetaData,
+          swapTokenValue,
+          type,
+        }),
+      );
+    } catch (e) {
+      console.log('signAndSendSwapsSmartTransaction error', e);
```
dan437, line 805: We set up a `smartTransactionError` on state for 5xx errors, which should disable smart transactions, but we need to actually simulate this failure and see what happens.

This could be handled by one of the pending items I mentioned: Show a notification to a user when a fallback to regular txs happened

ui/pages/swaps/smart-transaction-status/canceled-icon.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+export default function CanceledIcon() {
```
dan437, line 3: I've added a few icons for this feature directly in the STX Status folder, but I'm not sure if we already have some of them in some reusable folder. I could also move all of them to a reusable folder (which one?), so they could be used by other features as well in the future.

ui/pages/swaps/view-quote/view-quote.js
```diff
@@ -214,10 +263,21 @@ export default function ViewQuote() {
...
+  ) {
+    const maxFeePerGasDecWEI =
+      unsignedTransactionsAndEstimates.fees[
+        unsignedTransactionsAndEstimates.fees.length - 1
+      ]?.maxFeePerGas;
+    maxFeePerGas = decWEIToHexWEI(maxFeePerGasDecWEI);
+    maxGasLimit = unsignedTransactionsAndEstimates.gasLimit;
+  }
+
+  const gasTotalInWeiHex = calcGasTotal(maxGasLimit, maxFeePerGas || gasPrice);
```
dan437, line 280: I need to make sure that all fee estimates for regular / smart transactions show correct values. We especially need to complete this pending item: Check why we see a big gap between estimated and max fee on Rinkeby

ui/ducks/app/app.js
```diff
@@ -101,6 +102,24 @@ export default function reduceApp(state = {}, action) {
...
+      return {
+        ...appState,
+        smartTransactionsLiveness: action.payload,
+      };
+
+    case actionConstants.SET_SMART_TRANSACTIONS_ERROR:
+      return {
+        ...appState,
+        smartTransactionsError: action.payload,
+      };
```
dan437, line 121: Instead of adding STX state directly into the appState, can I use a namespace for Smart Transactions, e.g.:
```
case actionConstants.SET_UNSIGNED_TRANSACTIONS_AND_ESTIMATES:
  return {
    ...appState,
    smartTransactionsState: {
      ...appState.smartTransactionsState,
      unsignedTransactionsAndEstimates: action.payload,
    }
  };
```

app/_locales/en/messages.json
```diff
@@ -1525,6 +1534,9 @@
   "learnMore": {
     "message": "learn more"
   },
+  "learnMoreWithDot": {
+    "message": "Learn more."
```
darkwing, line 1538: Is this necessary? One with a period and one without?

app/scripts/controllers/transactions/index.js
```diff
@@ -917,6 +924,65 @@ export default class TransactionController extends EventEmitter {
...
 
+  async approveTransactionsWithSameNonce(listOfTxParams) {
+    // This is hacky, need to review with other engineers and decide on best alternative.
+    const uniqueHashOfParams = hashObject(listOfTxParams);
+    if (this.inProcessOfSigning.has(uniqueHashOfParams)) {
+      return '';
+    }
+    let rawTxes;
+    this.inProcessOfSigning.add(uniqueHashOfParams);
+    let nonceLock;
```
darkwing, line 935: Very minor nit but maybe we group the `let`'s here: `let rawTxes, nonceLock`

app/scripts/controllers/transactions/index.js
```diff
@@ -1084,6 +1150,83 @@ export default class TransactionController extends EventEmitter {
...
+
+    if (!txMeta) {
+      return;
+    }
+
+    try {
+      // It seems that sometimes the numerical values being returned from
+      // this.query.getTransactionReceipt are BN instances and not strings.
+      const gasUsed =
+        typeof txReceipt.gasUsed === 'string'
```
darkwing, line 1167: Strings have a `toString` method in JavaScript, so I don't think we need the type check.

app/scripts/controllers/transactions/index.js
```diff
@@ -1084,6 +1150,83 @@ export default class TransactionController extends EventEmitter {
...
+      const metricsParams = { gas_used: gasUsed };
+
+      if (submittedTime) {
+        metricsParams.completion_time = this._getTransactionCompletionTime(
+          submittedTime,
+        );
+      }
+
+      if (txReceipt.status === '0x0') {
+        metricsParams.status = 'failed on-chain';
```
darkwing, line 1193: I think the `failed on-chain` error is used in more than one place now, so we may want to make it a constant.

ui/pages/swaps/build-quote/build-quote.js
```diff
@@ -498,6 +545,93 @@ export default function BuildQuote({
...
+          <Popover
+            title={t('smartTransactionsAreHere')}
+            footer={
+              <>
+                <Button type="primary" onClick={onEnableSmartTransactionsClick}>
+                  {t('enableSmartTransactions')}
+                </Button>
+                <Box marginTop={4}>
+                  <Typography variant={TYPOGRAPHY.H6}>
+                    <a
```
darkwing, line 558: Could this be a `<Button type="link">`?

ui/pages/swaps/smart-transaction-status/index.scss
```diff
@@ -0,0 +1,83 @@
...
+    margin-bottom: 28px;
+  }
+
+  &__loading-bar {
+    height: 3px;
+    background: var(--Blue-500);
+    -webkit-transition: width 0.5s linear;
+    -moz-transition: width 0.5s linear;
+    -o-transition: width 0.5s linear;
+    transition: width 0.5s linear;
```
darkwing, line 28: I don't think we should need the browser prefixes ; `transition` has been around for several years now.

darkwing (changes requested):

app/scripts/controllers/transactions/index.js
meppsilon: I copied these lines from the code for `confirmTransactions` above.

meppsilon: But are you saying to just call `.toString(16)` on all types for `txReceipt.gasUsed`?

ui/pages/swaps/build-quote/build-quote.js
meppsilon: Possibly. Let me look into this.

ui/pages/swaps/smart-transaction-status/index.scss
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  &__content {
+    flex-flow: column;
+    width: 100%;
+  }
+
+  &__background-animation {
+    position: relative;
+    left: -88px;
```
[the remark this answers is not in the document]
meppsilon: RTL?

app/scripts/controllers/swaps.js
```diff
@@ -41,6 +41,8 @@ const POLL_COUNT_LIMIT = 3;
 // If for any reason the MetaSwap API fails to provide a refresh time,
 // provide a reasonable fallback to avoid further errors
 const FALLBACK_QUOTE_REFRESH_TIME = MINUTE;
+const FALLBACK_SMART_TRANSACTION_REFRESH_TIME = SECOND * 10;
+const FALLBACK_SMART_TRANSACTIONS_DEADLINE = 180;
```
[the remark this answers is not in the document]
meppsilon: I don't know if it fits here because the smart transactions deadline is counted in seconds and `MINUTE` is counted in milliseconds.

app/_locales/en/messages.json
dan437: I'm trying to follow the latest content guidelines. We will do a review of all STX content before pushing it to prod and if this is no longer needed, I will remove it.

ui/pages/swaps/smart-transaction-status/index.scss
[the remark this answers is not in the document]
dan437: CSS direction, right to left: https://developer.mozilla.org/en-US/docs/Web/CSS/direction

This `__background-animation` doesn't have any text, so even if we would use RTL somewhere for content, the animation can be the same as it is now:
image

app/scripts/controllers/transactions/index.js
```diff
@@ -1058,7 +1128,84 @@ export default class TransactionController extends EventEmitter {
       }
 
       if (txReceipt.status === '0x0') {
-        metricsParams.status = 'failed on-chain';
+        metricsParams.status = METRICS_STATUS_FAILED;
+        // metricsParams.error = TODO: figure out a way to get the on-chain failure reason
+      }
+
+      this._trackTransactionMetricsEvent(
```
danjm, line 1135: Note: 1135->1166 is not new code... just appears that way in the diff

app/scripts/controllers/transactions/index.js
danjm: The comments on lines 1179 and 1180 explain why the type check is needed

app/scripts/controllers/transactions/index.js
```diff
@@ -1058,7 +1128,84 @@ export default class TransactionController extends EventEmitter {
...
+    }
+  }
+
+  async confirmExternalTransaction(txMeta, txReceipt, baseFeePerGas) {
+    // add external transaction
+    await this.txStateManager.addExternalTransaction(txMeta);
+
+    const txId = txMeta.id;
+
+    if (!txMeta) {
```
danjm, line 1174: If this is true, then there will be an error on line 1172. Maybe this if condition should be placed before the txId variable declaraction?

app/scripts/controllers/transactions/index.js
```diff
@@ -928,6 +940,64 @@ export default class TransactionController extends EventEmitter {
     }
   }
 
+  async approveTransactionsWithSameNonce(listOfTxParams) {
+    // This is hacky, need to review with other engineers and decide on best alternative.
+    const uniqueHashOfParams = hashObject(listOfTxParams);
```
danjm, line 945: Still need to figure out the best alternative to this:

- [ ] how to determine if transaction is in process of signing

ui/helpers/utils/conversions.util.js
```diff
@@ -135,6 +135,15 @@ export function decGWEIToHexWEI(decGWEI) {
   });
 }
 
+export function decWEIToHexWEI(decWEI) {
```
danjm, line 138: We should just be able to use the `decimalToHex` function, already in this file, instead of this new one.

app/scripts/controllers/transactions/index.js
danjm: But we could move some of the identical code between this method and the `confirmTransaction` method to a shared helper method. For example, we could create a `getNormalizedTxReceiptGasUsed` method

ui/components/app/transaction-list/transaction-list.component.js
```diff
@@ -114,38 +115,53 @@ export default function TransactionList({
...
-            {pendingTransactions.map((transactionGroup, index) => (
-              <TransactionListItem
-                isEarliestNonce={index === 0}
-                transactionGroup={transactionGroup}
-                key={`${transactionGroup.nonce}:${index}`}
-              />
-            ))}
+            {pendingTransactions.map((transactionGroup, index) =>
+              transactionGroup.initialTransaction.transactionType ===
+              'smart' ? (
```
danjm, line 128: Let's create a constant for `'smart'`. It can be added to the `TRANSACTION_TYPES` object of `shared/constants/transaction.js`

ui/ducks/swaps/swaps.js
```diff
@@ -263,6 +300,36 @@ const getSwapsState = (state) => state.metamask.swapsState;
...
+export const getSmartTransactionsError = (state) =>
+  state.appState.smartTransactionsError;
+
+export const getSmartTransactionsErrorMessageDismissed = (state) =>
+  state.appState.smartTransactionsErrorMessageDismissed;
+
+export const getSmartTransactionsEnabled = (state) => {
+  const hardwareWalletUsed = isHardwareWallet(state);
+  const chainId = getCurrentChainId(state);
+  const isAllowedNetwork = [MAINNET_CHAIN_ID, RINKEBY_CHAIN_ID].includes(
```
danjm, line 312: Maybe the allowed networks for smart transactions should be saved in `shared/constants/swaps.js` where we also have things like `ALLOWED_SWAPS_CHAIN_IDS`

ui/store/actions.js
```diff
@@ -3110,6 +3119,169 @@ export async function setWeb3ShimUsageAlertDismissed(origin) {
...
+    const unsignedTransactionWithFees = {
+      ...unsignedTransaction,
+      maxFeePerGas: decimalToHex(fee.maxFeePerGas),
+      maxPriorityFeePerGas: decimalToHex(fee.maxPriorityFeePerGas),
+      gas: areCancelTransactions
+        ? decimalToHex(21000) // It has to be 21000 for cancel transactions, otherwise the API would reject it.
+        : unsignedTransaction.gas,
+      value: unsignedTransaction.value,
+    };
+    if (areCancelTransactions) {
```
danjm, line 3166: If it is a cancel transaction, should we also set the `data` field to `'0x'` or an empty string?

ui/ducks/swaps/swaps.js
```diff
@@ -342,6 +409,45 @@ export const getApproveTxParams = (state) => {
...
+    getCurrentChainId(state)
+  ];
+};
+
+export const getPendingSmartTransactions = (state) => {
+  const currentSmartTransactions = getCurrentSmartTransactions(state);
+  if (!currentSmartTransactions || currentSmartTransactions.length === 0) {
+    return [];
+  }
+  return currentSmartTransactions.filter((stx) => stx.status === 'pending');
```
danjm, line 427: Probably use our constant pattern for `'pending'`? Declare a constant in `shared/constants/transaction.js` and import it here

ui/selectors/transactions.js
```diff
@@ -45,13 +46,27 @@ export const unapprovedEncryptionPublicKeyMsgsSelector = (state) =>
...
 
+export const smartTransactionsListSelector = (state) =>
+  state.metamask.smartTransactionsState?.smartTransactions?.[
+    getCurrentChainId(state)
+  ]
+    ?.filter((stx) => !stx.confirmed)
+    .map((stx) => ({
+      ...stx,
+      transactionType: 'smart',
+      status: stx.status?.startsWith('cancelled')
```
danjm, line 57: Any reason we can't do an exact match here?

ui/selectors/transactions.js
```diff
@@ -45,13 +46,27 @@ export const unapprovedEncryptionPublicKeyMsgsSelector = (state) =>
...
   state.metamask.unapprovedTypedMessages;
 
+export const smartTransactionsListSelector = (state) =>
+  state.metamask.smartTransactionsState?.smartTransactions?.[
+    getCurrentChainId(state)
+  ]
+    ?.filter((stx) => !stx.confirmed)
+    .map((stx) => ({
+      ...stx,
+      transactionType: 'smart',
```
danjm, line 56: Related to a previous comment: replace `'smart'` with a `TRANSACTION_TYPES.SMART` from `shared/constants/transaction.js`

Alexalucky:
Job

ui/selectors/transactions.js
Alexalucky: smart

app/scripts/controllers/transactions/index.js
dan437: Let us please know if you figure it out or we can have a quick brainstorming session to find the best solution.

ui/store/actions.js
dan437: Great question, I will check it.

ui/selectors/transactions.js
dan437: Yup, we have multiple statuses with the `cancelled_` prefix, e.g. `cancelled_user_cancelled` or `cancelled_would_revert`.

ui/pages/swaps/smart-transaction-status/canceled-icon.js
danjm: I think this is fine for now. There will be a larger reorganization related to icons in the next couple of months.

ui/store/actions.js
```diff
@@ -3110,6 +3119,170 @@ export async function setWeb3ShimUsageAlertDismissed(origin) {
...
+  await promisifiedBackground.setSmartTransactionsOptInStatus(optInState);
+}
+
+export function fetchSmartTransactionFees(unsignedTransaction) {
+  return async (dispatch) => {
+    try {
+      const smartTransactionFees = await promisifiedBackground.fetchSmartTransactionFees(
+        unsignedTransaction,
+      );
+      dispatch({
```
danjm, line 3133: Do we need to add this to redux state? Could we just add it to `state.metamask.smartTransactionsState` by setting it on state in the controller?

ui/store/actions.js
```diff
@@ -3110,6 +3119,170 @@ export async function setWeb3ShimUsageAlertDismissed(origin) {
...
+      throw e;
+    }
+  };
+}
+
+export function fetchSmartTransactionsLiveness() {
+  return async (dispatch) => {
+    try {
+      const smartTransactionsLiveness = await promisifiedBackground.fetchSmartTransactionsLiveness();
+      dispatch({
```
danjm, line 3266: Do we need to add this to redux state? Could we just add it to state.metamask.smartTransactionsState by setting it on state in the controller?

ui/store/actions.js
dan437: Sure, we can do that.

ui/store/actions.js
```diff
@@ -3135,6 +3144,158 @@ export async function setWeb3ShimUsageAlertDismissed(origin) {
...
+  const signedTransactions = await promisifiedBackground.approveTransactionsWithSameNonce(
+    unsignedTransactionsWithFees,
+  );
+  return signedTransactions;
+};
+
+export function signAndSendSmartTransaction({
+  unsignedTransaction,
+  smartTransactionFees,
+}) {
```
danjm, line 3202: One final improvement to make to this PR: I think the logic in `signAndSendSmartTransaction` and `createSignedTransactions` should be within the smart transactions controller itself. So a `createAndSubmitSignedTransactions` method would be exposed, and passed `unsignedTransaction` and 
  `smartTransactionFees`, and would return `{ uuid: 'dP23W7c2kt4FK9TmXOkz1UM2F20' }`
  
  This can be done in a follow up improvement PR if preferred.

app/scripts/controllers/transactions/index.js
danjm: Maybe we could just get a hash of the first transaction in the array, using TransactionFactory... that would be a bit better I think

danjm: but we might not be able to do that for an unsigned transaction...

danjm: I pushed a commit here that I think resolves my concerns: https://github.com/MetaMask/metamask-extension/commit/54a01beabb0e0d15cf451584982f32923834dddb

app/scripts/controllers/transactions/index.js
```diff
@@ -930,6 +941,73 @@ export default class TransactionController extends EventEmitter {
...
+    const initialTxAsSerializedHex = bufferToHex(initialTxAsEthTx.serialize());
+
+    if (this.inProcessOfSigning.has(initialTxAsSerializedHex)) {
+      return '';
+    }
+    this.inProcessOfSigning.add(initialTxAsSerializedHex);
+    let rawTxes, nonceLock;
+    try {
+      // TODO: we should add a check to verify that all transactions have the same from address
+      const fromAddress = listOfTxParams[0].from;
```
darkwing, line 963: I think we can elevate this to line 950, since we use `listOfTxParams[0].from` via the `initialTx` variable.

app/scripts/controllers/transactions/index.js
```diff
@@ -930,6 +941,73 @@ export default class TransactionController extends EventEmitter {
...
+          return this.signExternalTransaction(txParams);
+        }),
+      );
+      nonceLock.releaseLock();
+    } catch (err) {
+      log.error(err);
+      // must set transaction to submitted/failed before releasing lock
+      if (nonceLock) {
+        nonceLock.releaseLock();
+      }
```
darkwing, line 979: Can this conditional be moved from both the `try` and `catch` to `finally`?

app/scripts/controllers/transactions/lib/util.js
```diff
@@ -277,3 +277,15 @@ export function getFinalStates() {
...
+/**
+ * Normalizes tx receipt gas used to be a hexadecimal string.
+ * It seems that sometimes the numerical values being returned from
+ * this.query.getTransactionReceipt are BN instances and not strings.
+ *
+ * @param {string or BN instance} gasUsed
+ * @returns normalized gas used as hexadecimal string
+ */
+export function normalizeTxReceiptGasUsed(gasUsed) {
+  return typeof gasUsed === 'string' ? gasUsed : gasUsed.toString(16);
```
darkwing, line 290: I think I mentioned this somewhere else but `string` has a `toString` method, so I don't think we need the type check.

ui/components/app/transaction-list-item/smart-transaction-list-item.component.js
```diff
@@ -0,0 +1,89 @@
...
+  } = smartTransaction;
+  const category = TRANSACTION_GROUP_CATEGORIES.SWAP;
+  const title = t('swapTokenToToken', [
+    sourceTokenSymbol,
+    destinationTokenSymbol,
+  ]);
+  const subtitle = 'metamask';
+  const date = formatDateWithYearContext(time);
+  let displayedStatusKey;
+  if (status === 'pending') {
```
darkwing, line 38: Can we just use `TRANSACTION_GROUP_STATUSES.PENDING` here?

ui/components/app/transaction-list-item/smart-transaction-list-item.component.js
```diff
@@ -0,0 +1,89 @@
...
+    displayedStatusKey = TRANSACTION_GROUP_STATUSES.PENDING;
+  } else if (status?.startsWith('cancelled')) {
```
darkwing, line 40: I think we can use `TRANSACTION_GROUP_STATUSES .CANCELLED` here.

ui/components/app/transaction-list-item/smart-transaction-list-item.component.js
meppsilon: Yea, we could switch it over to `TRANSACTION_GROUP_STATUSES.PENDING`. I have been hesitant because 'pending' reflects an enum from the smart transactions controller and `TRANSACTION_GROUP_STATUSES.PENDING` is a constant that reflects the status of normal transactions. So if one of those values changes, this piece of code will break. But I think if they are unlikely to change over time, it should be fine to change it.

ui/components/app/transaction-list-item/smart-transaction-list-item.component.js
meppsilon: Same goes for cancelled. This bit of code is supposed to pick up a cancelled status from an array of potential cancelled statuses, differentiated by them all beginning with `cancelled`. For example, some of the statuses are `cancelled_would_revert`, `cancelled_too_cheap`, `cancelled_deadline_missed`. It works here because `TRANSACTION_GROUP_STATUSES.CANCELLED` starts with the exact value but I worry that the value might change over time for either this constant or smart transactions statuses and this bit of code would break.

app/scripts/controllers/transactions/lib/util.js
meppsilon: You mentioned it in this thread: https://github.com/MetaMask/metamask-extension/pull/12676/files/dddef0a043450bd69d1fbc00bf2d9cc98243163f#r785919603

meppsilon: Let me know if the responses answer why the type check is needed. But I added this code based on @danjm comment to refactor the duplicate code into a separate function.

ui/components/app/transaction-list-item/smart-transaction-list-item.component.js
dan437: If we would be using a constant, I would prefer to have a new one, which is specific to Smart Transactions. E.g. `STX_STATUSES` or `SMART_TRANSACTION_STATUSES`.

ui/store/actions.js
dan437: FYI, I verified that setting it to '0x' is correct. Thanks for the suggestion!

ui/store/actions.js
dan437: I will do that in a follow up PR, so we wouldn't need to repeat too much of this code on the mobile side.

app/scripts/metamask-controller.js
```diff
@@ -734,6 +738,20 @@ export default class MetamaskController extends EventEmitter {
...
       ),
     });
+    this.smartTransactionsController = new SmartTransactionsController({
+      onNetworkStateChange: this.networkController.store.subscribe.bind(
+        this.networkController.store,
+      ),
+      getNetwork: this.networkController.getNetworkState.bind(
+        this.networkController,
+      ),
+      nonceTracker: this.txController.nonceTracker,
```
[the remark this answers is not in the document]
danjm: I see the change made for the nonce tracker, but can you do the same for the transactions controller as well: just pass `confirmExternalTransaction` instead of the whole controller. We try not to pass around whole controllers.

app/scripts/controllers/transactions/tx-state-manager.js
```diff
@@ -266,6 +266,11 @@ export default class TransactionStateManager extends EventEmitter {
     return txMeta;
   }
 
+  addExternalTransaction(txMeta) {
```
danjm, line 269: One more request: could we add validation here to make sure that any external transaction added to state has a confirmed status and that it's nonce does not match the nonce of any current confirmed or pending transaction

dan437: Sure.

dan437: FYI, I added it yesterday, let me know what you think!

ui/pages/swaps/swaps.util.js
```diff
@@ -490,6 +492,34 @@ export async function fetchSwapsGasPrices(chainId) {
...
+  chainId,
+  currentCurrency,
+  conversionRate,
+  nativeCurrencySymbol,
+  feeInWeiDec,
+}) => {
+  const feeInWeiHex = decimalToHex(feeInWeiDec);
+  const ethFee = getValueFromWeiHex({
+    value: feeInWeiHex,
+    toDenomination: 'ETH',
```
darkwing, line 505: Do we have `ETH` as a constant somewhere?

danjm (dismissed):
This is looking good to me.

Let's make sure this improvement suggestion gets followed up on soon https://github.com/MetaMask/metamask-extension/pull/12676/files#r789644150

darkwing (approved):

danjm (approved):

PR #12679, opened 2021-11-12
Fix: eth_sign does not validate input

### LINKED ISSUE #5039: bug: eth_sign does not validate input
i was seeing the metamask approval window and approving it but not seeing the callback called from the dapp

invocation method:
```js
web3.eth.sign(web3.eth.accounts[0], 'yo', console.log)
```

ritave:
Fixes: #5039

### DIFF
--- a/app/scripts/lib/message-manager.js
+++ b/app/scripts/lib/message-manager.js
@@ -93,6 +93,10 @@ export default class MessageManager extends EventEmitter {
                 'MetaMask Message Signature: User denied message signature.',
               ),
             );
+          case 'errored':
+            return reject(
+              new Error(`MetaMask Message Signature: ${data.error}`),
+            );
           default:
             return reject(
               new Error(
@@ -231,6 +235,19 @@ export default class MessageManager extends EventEmitter {
       });
     }
     this._setMsgStatus(msgId, 'rejected');
+  }
+
+  /**
+   * Sets a Message status to 'errored' via a call to this._setMsgStatus.
+   *
+   * @param {number} msgId - The id of the Message to error
+   *
+   */
+  errorMessage(msgId, error) {
+    const msg = this.getMsg(msgId);
+    msg.error = error;
+    this._updateMsg(msg);
+    this._setMsgStatus(msgId, 'errored');
   }
 
   /**
--- a/app/scripts/lib/message-manager.js
+++ b/app/scripts/lib/message-manager.js
@@ -80,8 +80,8 @@ export default class MessageManager extends EventEmitter {
    *
    */
   addUnapprovedMessageAsync(msgParams, req) {
+    const msgId = this.addUnapprovedMessage(msgParams, req);
     return new Promise((resolve, reject) => {
-      const msgId = this.addUnapprovedMessage(msgParams, req);
       // await finished
       this.once(`${msgId}:finished`, (data) => {
         switch (data.status) {
@@ -304,7 +304,7 @@ export default class MessageManager extends EventEmitter {
  * @returns {string} A hex string conversion of the buffer data
  *
  */
-function normalizeMsgData(data) {
+export function normalizeMsgData(data) {
   if (data.slice(0, 2) === '0x') {
     // data is already hex
     return data;
--- a/app/scripts/lib/message-manager.js
+++ b/app/scripts/lib/message-manager.js
@@ -79,9 +79,9 @@ export default class MessageManager extends EventEmitter {
    * @returns {promise} after signature has been
    *
    */
-  addUnapprovedMessageAsync(msgParams, req) {
+  async addUnapprovedMessageAsync(msgParams, req) {
     const msgId = this.addUnapprovedMessage(msgParams, req);
-    return new Promise((resolve, reject) => {
+    return await new Promise((resolve, reject) => {
       // await finished
       this.once(`${msgId}:finished`, (data) => {
         switch (data.status) {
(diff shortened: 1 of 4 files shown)


app/scripts/lib/message-manager.js
```diff
@@ -79,8 +79,8 @@ export default class MessageManager extends EventEmitter {
    *
    */
   addUnapprovedMessageAsync(msgParams, req) {
+    const msgId = this.addUnapprovedMessage(msgParams, req);
```
ritave, line 82: Tests were failing due to race condition, trying to access unapproved list before the Promise even started

Gudahtt: Huh, interesting. So this situation isn't ideal because if `addUnapprovedMessage` throws, it'll do so synchronously, but if anything inside the Promise throws, it'll reject. This might be confusing if anyone tries to handle errors with `.catch` and this throws and slips past that attempt.

Could you remove the `Promise` wrapper altogether and make this function `async`? Or would that trigger the same issue you encountered here I wonder. Which test was it that was failing?

app/scripts/metamask-controller.js
```diff
@@ -1826,12 +1827,18 @@ export default class MetamaskController extends EventEmitter {
    * @param {Function} cb - The callback function called with the signature.
    */
   newUnsignedMessage(msgParams, req) {
-    const promise = this.messageManager.addUnapprovedMessageAsync(
-      msgParams,
-      req,
-    );
-    this.sendUpdate();
-    this.opts.showUserConfirmation();
+    const data = normalizeMsgData(msgParams.data);
```
Gudahtt, line 1830: Just as with the other comment, I think having two styles of throwing errors in this function is confusing and should be avoided. Making this `async` would ensure errors are thrown in a single way.

This was an issue that I had known about for a while, and haven't gotten around to fixing yet.

app/scripts/lib/message-manager.js
ritave: It failed here - https://github.com/MetaMask/metamask-extension/blob/6309d51e359aa0d1788f180fb410e39a1e2bd0d8/app/scripts/metamask-controller.test.js#L857-L864
With `messages` returned being empty.

I don't think this promise can be easily converted into async, since it waits on EventEmitter

app/scripts/metamask-controller.js
```diff
@@ -1825,14 +1826,20 @@ export default class MetamaskController extends EventEmitter {
...
-  newUnsignedMessage(msgParams, req) {
-    const promise = this.messageManager.addUnapprovedMessageAsync(
-      msgParams,
-      req,
-    );
-    this.sendUpdate();
-    this.opts.showUserConfirmation();
-    return promise;
+  async newUnsignedMessage(msgParams, req) {
+    const data = normalizeMsgData(msgParams.data);
```
Gudahtt, line 1830: It looks like this `normalizeMsgData` function is being called twice now - once here, and once in `addUnapprovedMessage` in `message-manager-.js`.

I don't really follow why it was moved to be honest. Also, does it need to be called twice?

app/scripts/metamask-controller.js
```diff
@@ -1825,14 +1826,20 @@ export default class MetamaskController extends EventEmitter {
...
+    let promise;
+    // 64 hex + "0x" at the beginning
+    if (data.length === 66) {
```
Gudahtt, line 1833: Why require messages to be 64 hex characters?

ritave: This is the requirement of ECSign we use behind the scenes, it only works for 32 byte numbers

app/scripts/metamask-controller.js
ritave: We need to check the 32 byte length somehow, and data can be either a Buffer or a hex string. Without this function we would need to check for both types of incoming data. Normalising it is simpler

app/scripts/metamask-controller.js
ritave: I'll add more explicit documentation

app/scripts/metamask-controller.js
Gudahtt: Gotcha. I hadn't realized a length of 32 was a requirement. Testing with prod confirms that it is the case though. Mostly anyway.

I noticed one odd exception: if the message contains an odd number of digits, the last digit is ignored. I see how that would happen, as that would be 32 and a half bytes. I guess it interprets the string as bytes and ignores any remainder.

But imposing this new restriction could break sites that are unintentionally relying on their signatures working even when they add an extra byte. Perhaps we should allow the extra character for this reason, just to make this non-breaking.

Gudahtt:
It's great that we've prevented this case from happening with added validation, but it's still concerning that any failure during signing results in no error being returned to the site. If some other case also resulted in throwing, we'd see the same behaviour of the request staying as pending forever.

I think the reason the request remains pending is that we don't handle the error properly in `signMessage`. We don't handle it in `signPersonalMessage` either. If the signing fails, we bubble the error up to the UI (which is good), but we don't set the state of that message to "failed". You can see this is done with some of the other signatures types, like `decruptMessage` and `signTypedMessage`. Rather than let it bubble up directly, we should be catching the error, setting the status of the signature request as failed, _then_ re-throwing it so the UI knows that it failed.

This would ensure that if something breaks again in the future, at least the site knows it failed.

Gudahtt (commented):
LGTM! I can approve after the lint error is fixed.

Gudahtt (approved):

darkwing (approved):

PR #12685, opened 2021-11-12
Ensure approve button can be enabled when using ledger webhid

danjm:
Fixes a bug in prod with the ledger webhid feature that prevents confirmation of token approvals (when using the webhid connection)

The code was incorrectly using `ledgerWalletRequiredHidConnection` to decide whether to disable the "Confirm" button, and whether to render the ledger instruction component. This is a problem because the ledger instruction component, in its `useEffect`, checks the ledger connection and then changes state to change `ledgerWalletRequiredHidConnection` from `true` to `false`. Once it is set to false, the ledger instruction component no longer renders. But upon dismount, that component updates state again and `ledgerWalletRequiredHidConnection` becomes true. So it is stuck in an infinite loop of mounting and unmounting and toggling `ledgerWalletRequiredHidConnection` between true and false.

The fix is to only use the check whether the current account is a ledger when deciding whether to render the instruction component. This is what is done for the confirm screen for all other transactions, and for the message signing confirmation screens.

### DIFF
--- a/ui/pages/confirm-approve/confirm-approve-content/confirm-approve-content.component.js
+++ b/ui/pages/confirm-approve/confirm-approve-content/confirm-approve-content.component.js
@@ -56,7 +56,7 @@ export default class ConfirmApproveContent extends Component {
     showCustomizeNonceModal: PropTypes.func,
     warning: PropTypes.string,
     txData: PropTypes.object,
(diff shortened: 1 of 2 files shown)


darkwing:
Using our test dapp, I created a token and then was able to approve said token.

darkwing (approved):

ryanml (approved):

PR #12693, opened 2021-11-13
Replace history during onboarding

mcmire:
Manual testing steps:  

* "Create wallet" flow should still work (using onboarding v1)
* "Import wallet" flow should still work (using onboarding v1)
* "Create wallet" flow should still work (using onboarding v2)
* "Import wallet" flow should still work (using onboarding v2)

### DIFF
--- a/ui/pages/first-time-flow/create-password/import-with-seed-phrase/import-with-seed-phrase.component.js
+++ b/ui/pages/first-time-flow/create-password/import-with-seed-phrase/import-with-seed-phrase.component.js
@@ -142,7 +142,7 @@ export default class ImportWithSeedPhrase extends PureComponent {
 
       setSeedPhraseBackedUp(true).then(async () => {
         initializeThreeBox();
-        history.push(INITIALIZE_END_OF_FLOW_ROUTE);
+        history.replace(INITIALIZE_END_OF_FLOW_ROUTE);
       });
     } catch (error) {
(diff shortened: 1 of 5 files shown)


jpuri (dismissed):
LGTM

Gudahtt (approved):
LGTM!

I was testing that "back" worked correctly on the page following each of these changes, and I noticed that it's broken for the "import" flow. If you press back on the last screen, it brings you to the metrics opt-in page, and then into the "confirm seed phrase" flow (which is part of the _create wallet_ flow). But I checked prod, and this is a pre-existing issue that we can solve separately. The create wallet flow works fine before and after this change, as does the V2 import flow.

mcmire:
@Gudahtt I noticed that behavior as well in testing. I wonder if that behavior is intentional — once you enter a seed phrase then it remembers this and so if you attempt to go back it forwards you past that step automatically and doesn't let you go back. That's just a guess though, perhaps it's not supposed to happen. I'll look into it for sure.

Gudahtt:
Maybe :thinking: The asking for verification is a bit weird, since the user has kinda proved already that they have backed it up by virtue of entering it in the first place. But either way, I think we can deal with that in a separate PR if we do decide it should be changed. It may be rendered obsolete by the new onboarding flow soon enough anyway.

rekmarks (approved):

PR #12702, opened 2021-11-14
Add per-build type LavaMoat policies

### LINKED ISSUE #12690: Create and use separate LavaMoat policies for each build type
We exclude code based on the `build-type` specified to the build system, which means that LavaMoat may (and in practice, does) generate different policies depending on the build type. Therefore, we need to store one policy for each policy, keep them up to date, and use the correct policy for every build.

rekmarks:
This PR adds one LavaMoat background script policy or each build type. It also renames the build system policy directory from `node` to `build-system` to make its purpose more clear. Each build type has the original `policy-override.json` for `main` builds.

We need to maintain separate policies for each build type because each type will produce different bundles with different internal and external modules.

In addition to the LavaMoat-related changes, this PR updates the lint scripts to no longer lint the auto-generated LavaMoat policy files. This should save off a significant amount of time from our JSON linting (which we do with `prettier` directly).

Closes #12690


development/generate-lavamoat-policies.sh
```diff
@@ -0,0 +1,12 @@
...
+set -e
+set -u
+set -o pipefail
+
+# Generate LavaMoat policies for the extension background script for each build
+# type.
+concurrently --kill-others-on-fail -n main,beta,flask \
+  "WRITE_AUTO_POLICY=1 yarn dist" \
+  "WRITE_AUTO_POLICY=1 yarn dist --build-type beta" \
+  "WRITE_AUTO_POLICY=1 yarn dist --build-type flask"
```
rekmarks, line 12: Note: Your device may board the Struggle Bus when running this locally.

kumavis:
this sounds like hell for the security engineers charged with reviewing policies and policy changes
but i dont have a better solution on deck

development/generate-lavamoat-policies.sh
kumavis: load testing!

kumavis (dismissed):
so be it

README.md
```diff
@@ -67,9 +67,12 @@ Whenever you change dependencies (adding, removing, or updating, either in `pack
 * The `allow-scripts` configuration in `package.json`
   * Run `yarn allow-scripts auto` to update the `allow-scripts` configuration automatically. This config determines whether the package's install/postinstall scripts are allowed to run. Review each new package to determine whether the install script needs to run or not, testing if necessary.
   * Unfortunately, `yarn allow-scripts auto` will behave inconsistently on different platforms. macOS and Windows users may see extraneous changes relating to optional dependencies.
-* The LavaMoat auto-generated policy in `lavamoat/node/policy.json`
-  * Run `yarn lavamoat:auto` to re-generate this policy file. Review the changes to determine whether the access granted to each package seems appropriate.
+* The production LavaMoat policy files (`lavamoat/browserify/*/policy.json`)
+  * Run `yarn lavamoat:auto` to re-generate these policy files. Review the changes to determine whether the access granted to each package seems appropriate.
```
Gudahtt, line 71: The `lavamoat:auto` script will re-generate both the prod and build system policy files. Is this what you meant?

```suggestion
  * Run `yarn lavamoat:background:auto` to re-generate these policy files. Review the changes to determine whether the access granted to each package seems appropriate.
```

README.md
```diff
@@ -67,9 +67,12 @@ Whenever you change dependencies (adding, removing, or updating, either in `pack
...
   * Unfortunately, `yarn lavamoat:auto` will behave inconsistently on different platforms. macOS and Windows users may see extraneous changes relating to optional dependencies.
+* The build system LavaMoat policy file (`lavamoat/build-system/policy.json`)
+  * Run `yarn lavamoat:build:auto` to re-generate this policy file.
+  * This only applies if the external modules imported by the build system or its dependencies have changed. Keep in mind that some modules may be imported dynamically, and may elude LavaMoat's static analysis. If this occurs, debug the build system with [`ndb`](https://www.npmjs.com/package/ndb) to determine which packages are dynamically imported, add them to the "useless" imports in `development/build/index.js`, and re-run the auto policy script.
```
Gudahtt, line 75: This advice seems... overly specific? Dynamic imports aren't the only Lavamoat-related problem people are likely to run up against, nor is it specific to the build system. Other problems include dynamic global usage, attempts to mutate frozen globals, and incorrect policy generation due to multiple copies of a dependency being in the dependency tree.

development/generate-lavamoat-policies.sh
```diff
@@ -0,0 +1,13 @@
...
+
+set -e
+set -u
+set -o pipefail
+
+# Generate LavaMoat policies for the extension background script for each build
+# type.
+# ATTN: This may tax your device when running it locally.
+concurrently --kill-others-on-fail -n main,beta,flask \
+  "WRITE_AUTO_POLICY=1 yarn dist" \
```
Gudahtt, line 11: Is this safe to run concurrently? All three of these builds will write to the `dist` directory, so I'd expect these to be overwriting each other.

package.json
```diff
@@ -41,8 +41,9 @@
     "test:coverage:path": "nyc --check-coverage yarn test:unit:path",
     "ganache:start": "./development/run-ganache.sh",
     "sentry:publish": "node ./development/sentry-publish.js",
-    "lint": "prettier --check '**/*.json' && eslint . --ext js,snap --cache && yarn lint:styles",
-    "lint:fix": "prettier --write '**/*.json' && eslint . --ext js --cache --fix",
+    "lint:prettier": "prettier '**/*.json' '!lavamoat/**/policy.json'",
```
Gudahtt, line 44: We already ignore these policies in the `.prettierignore` file. Except now that's broken since policy names were introduced. I'd suggest updating that invalid entry instead of adding it here (the `*` needs to be changed to `**`).

development/generate-lavamoat-policies.sh
rekmarks: Hm. For the purposes of validating the LavaMoat config, it doesn't matter what's in the `dist` directory or whether files are being overwritten. The build processes just need to complete normally. We could really use a `--no-emit` option, actually.

My question is whether this could interfere with any of the other build scripts we run, if they also write or read from `dist`?

Gudahtt: My concern was that LavaMoat itself might be generating the policy files from-disk. So the version of the build it was generating a policy for might be reliant on this race condition of which build finishes when.

But on second thought... that's not how any of this works. LavaMoat runs on the JavaScript while it's still in stream form, before it gets written to disk. At least I'm pretty sure. So yeah you're probably right.

And I agree, a `--no-emit` option would be useful here.

package.json
rekmarks: Fixed in 0ff072978

Gudahtt (dismissed):
LGTM! Though regarding that README comment, I think that could be improved by moving that debugging advice to a separate bullet point or section. So that it can be expanded later, and inclusive of more than just the build system policy.

README.md
rekmarks: Yeah, fair enough! Updated here: bb2836752

Gudahtt (approved):
LGTM!

development/generate-lavamoat-policies.sh
kumavis: im not sure if:
- `clean` is safe (removes and creates the dir) 
- inpage/contentscript is safe (creates the inpage bundle and puts it in the contentscript bundle)
but if it works it works

kumavis (approved):

PR #12704, opened 2021-11-15
Adding network status section on edit gas fee modal

### LINKED ISSUE #12452: EIP-1559 V2: Edit gas fee modal - show network status details
Show current network status details in Edit gas fee modal as below:
Screenshot 2021-10-23 at 1 05 45 AM

Screenshot 2021-10-23 at 1 18 03 AM

3 data show are:
1. base fee
2. priority fee range
3. network stability slider
These will be made available in app state as part of this task: https://github.com/MetaMask/metamask-extension/issues/12473

Also, add "Want to learn more about gas?" link at bottom.

jpuri:
Fixes: #12452

Explanation:  Adding network status section on edit gas fee modal.

Screenshot 2021-11-15 at 5 38 34 AM


ui/components/app/edit-gas-fee-popover/network-status/network-status.js
```diff
@@ -0,0 +1,47 @@
...
+        fontWeight="bold"
+        margin={[3, 0]}
+        variant="h6"
+      >
+        <I18nValue messageKey="networkStatus" />
+      </Typography>
+      <div className="network-status__info">
+        <div className="network-status__info-field">
+          <span className="network-status__info-field-data">
+            {gasFeeEstimates?.estimatedBaseFee} GWEI
```
mcmire, line 27: Should we truncate this number? Perhaps to 2 or 3 decimal places?

Also, if `gasFeeEstimates` is null then this will just appear as "GWEI". Should we even render this component unless we are guaranteed to have `gasFeeEstimates`? Or should we render some kind of loader that will appear until we have `gasFeeEstimates`?

ui/components/app/edit-gas-fee-popover/network-status/status-slider/status-slider.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+
+const GradientColors = [
```
mcmire, line 3: nit: Maybe this should be `gradientColors` or `GRADIENT_COLORS`?

ui/components/app/edit-gas-fee-popover/network-status/status-slider/status-slider.js
```diff
@@ -0,0 +1,39 @@
...
+const StatusSlider = () => {
+  const statusValue = 0.5;
+  const sliderValueNumeric = Math.round(statusValue * 10);
+  return (
+    <div className="status-slider">
+      <div
+        className="status-slider__arrow"
+        style={{
+          borderTopColor: GradientColors[sliderValueNumeric],
+          marginLeft: `${sliderValueNumeric * 10 - 10}%`,
```
mcmire, line 25: It looks like `statusValue` is already being multiplied by 10 — did you mean to divide `sliderValueNumeric` by 10 again or not multiply this by 10 here?

Also, what is the reasoning behind subtracting 10 here? If it's because the X coordinate of the arrow starts at the left side of the arrow, maybe it's worth adding a negative margin-left to the arrow? This way the positioning of the arrow is anchored at its center and you may not need an extra -10% here.

ui/components/app/edit-gas-fee-popover/network-status/index.scss
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    &-field {
+      display: flex;
+      flex-direction: column;
+      align-items: center;
+      justify-content: center;
+      width: 30%;
+
+      &__priorityfee {
```
mcmire, line 23: Another nit about BEM, but it looks like the priority fee may be a type of field and not a child of a field. Should this be `&--priority-fee` instead, or perhaps at the least `.network-status__priority-fee-field`?

mcmire (commented):
Nice work, just had some comments/suggestions.

ui/components/app/edit-gas-fee-popover/network-status/status-slider/status-slider.js
```diff
@@ -0,0 +1,39 @@
...
+        }}
+      />
+      <div className="status-slider__line" />
+      <div
+        className="status-slider__label"
+        style={{ color: GradientColors[sliderValueNumeric] }}
+      >
+        {statusValue <= 0.5 ? 'Stable' : 'Unstable'}
```
mcmire, line 33: Maybe this isn't a big deal now as we have to come back to this anyway, but didn't this end up getting changed to three levels? I'm seeing "Not busy", "Stable", and "Busy".

ui/components/app/edit-gas-fee-popover/network-status/status-slider/status-slider.js
jpuri: This is to account for the width of arrow.

ui/components/app/edit-gas-fee-popover/network-status/status-slider/status-slider.js
jpuri: Good point, I need to check this with product team.

ui/components/app/edit-gas-fee-popover/network-status/index.scss
jpuri: I changed it to `&--priority-fee`.

ui/components/app/edit-gas-fee-popover/network-status/status-slider/status-slider.js
jpuri: Fixed 👍

ui/components/app/edit-gas-fee-popover/network-status/network-status.js
jpuri: I added condition to show edit-gas-fee modal only if estimates are present. 

edit-gas-fee modal does show `LoadingHeartBeat` component while gas estimates are loading.

ui/components/app/edit-gas-fee-popover/network-status/status-slider/status-slider.js
mcmire: Gotcha, that's what I figured. Since this is disconnected from the CSS, that might not be obvious. What if you were to add a negative margin-left to the arrow element so that when `marginLeft` here is 0% then the point of the arrow lands at the right spot?

ui/components/app/edit-gas-fee-popover/network-status/network-status.js
mcmire: Okay. So if the modal only appears if `gasFeeEstimates` then do you need the null check here? I get a little sensitive about null checks because it means that we don't trust the upstream code, whatever that is. From a type perspective, it seems that either `gasFeeEstimates` is null or that it's something. But correct me if I'm wrong.

jpuri: LoadingHeartBeat does not hide edit gas modal content but shows a loading effect. Thus check for null is needed to avoid exception being thrown.

ui/components/app/edit-gas-fee-popover/network-status/status-slider/status-slider.js
jpuri: Total value of margin-left depends on `statusValue` so for status value .5 it will be 40%.

Not sure how do you suggest doing it.

ui/components/app/edit-gas-fee-popover/network-status/network-status.js
mcmire: Hmm ok. Just feels like we could do this better. I don't have enough familiarity to know what that is yet though. More food for thought I guess.

ui/components/app/edit-gas-fee-popover/network-status/index.scss
```diff
@@ -0,0 +1,41 @@
+.network-status {
+  margin: 24px 0 12px;
+
+  &__info {
+    border-top: 1px solid $ui-2;
+    border-bottom: 1px solid $ui-2;
+    height: 56px;
+    display: flex;
+    align-items: center;
```
mcmire, line 9: It looks these three columns get shoved to the left and don't take seem to take up the full width (this shows up when base fee is only three digits). What if you add `justify-content: space-between` here?

ui/components/app/edit-gas-fee-popover/network-status/index.scss
```diff
@@ -0,0 +1,41 @@
...
+
+    &-separator {
```
mcmire, line 11: Sorry to be the BEM nazi 😂 but I'm wondering if these shouldn't be nested to avoid specificity issues and should be named:

* `.network-status__separator`
* `.network-status__field`
* `.network-status__field--priority-fee` (this one is good already)
* `.network-status__field-data`
* `.network-status__field-label`

ui/components/app/edit-gas-fee-popover/network-status/status-slider/status-slider.js
mcmire: Okay, I played around with this a bit and came up with an idea where the arrow is wrapped in a couple of divs. The reason for `arrow-border` is that I noticed in the designs that there is a small white border around the arrow, and it gets pressed into the slider a bit. The reason for `arrow-wrapper` is that it allows you to apply a margin-left in a clear way without having to do any subtraction. Let me know what you think about this:

``` jsx
<div
  className="status-slider__arrow-wrapper"
  style={{
    marginLeft: `${sliderValueNumeric * 10}%`,
  }}
>
  <div className="status-slider__arrow-border">
    <div
      className="status-slider__arrow"
      style={{
        borderTopColor: GRADIENT_COLORS[sliderValueNumeric],
      }}
    ></div>
  </div>
</div>
```

``` css
.status-slider {
  &__arrow-wrapper {
    align-self: baseline;
  }

  &__arrow-border {
    width: 0;
    height: 0;
    border-left: 10px solid transparent;
    border-right: 10px solid transparent;
    border-top: 10px solid white;
    position: relative;
    /* presses the arrow into the slider */
    margin-bottom: -2px;
    /* places the point of the arrow at the very beginning of the slider by default so that you don't have to do extra math */
    margin-left: -10px;
  }

  &__arrow {
    width: 0;
    height: 0;
    border-left: 5px solid transparent;
    border-right: 5px solid transparent;
    border-top: 5px solid black;
    position: absolute;
    bottom: 3px;
    left: -5px;
  }
}
```

jpuri: Great idea, I appreciate your eye for details. I fixed it.

ui/components/app/edit-gas-fee-popover/network-status/network-status.js
```diff
@@ -0,0 +1,48 @@
...
+        margin={[3, 0]}
+        variant="h6"
+      >
+        <I18nValue messageKey="networkStatus" />
+      </Typography>
+      <div className="network-status__info">
+        <div className="network-status__info-field">
+          <span className="network-status__info-field-data">
+            {gasFeeEstimates?.estimatedBaseFee &&
+              `${gasFeeEstimates?.estimatedBaseFee} GWEI`}
```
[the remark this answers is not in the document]
jpuri: Yep I also see that locally for test network, but as I understand `gasFeeEstimates?.estimatedBaseFee` is already in GWEI, thus I guess it is only a test network thing.

ui/components/app/edit-gas-fee-popover/network-status/index.scss
jpuri: That was broken coz of some changes I did in PR, I fixed it.

ui/components/app/edit-gas-fee-popover/network-status/index.scss
jpuri: I appreciate your detailed reviews, I will take time to read BEM more. Thanks a lot.

ui/components/app/edit-gas-fee-popover/network-status/status-slider/status-slider.js
```diff
@@ -14,8 +16,15 @@ const GRADIENT_COLORS = [
 ];
 
 const StatusSlider = () => {
+  // todo: value below to be replaced with dynamic values from api once it is available
+  // corresponding test cases also to be added
   const statusValue = 0.5;
   const sliderValueNumeric = Math.round(statusValue * 10);
+
+  let statusLabel = 'stable';
+  if (statusValue <= 0.33) statusLabel = 'notBusy';
```
mcmire, line 25: Do you mind putting these if statements on their own line? I'm not sure why ESLint isn't picking this up, but I think it'll make this logic more obvious.

ui/components/app/edit-gas-fee-popover/network-status/network-status.js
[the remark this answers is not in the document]
mcmire: Hmm. That's odd.

mcmire (dismissed):
Another minor thing, but this is good to go otherwise 👍🏻

ui/components/app/edit-gas-fee-popover/network-status/status-slider/status-slider.js
jpuri: fixed 👍

ui/components/app/edit-gas-fee-popover/network-status/network-status.js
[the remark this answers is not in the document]
jpuri: the values looks odd to you ?

in test networks I noticed it varies from `.00000009` to something like `213.2342353`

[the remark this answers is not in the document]
mcmire: Oh I see. I just wouldn't expect gas fees to be lower than 1 GWEI at all, but if it is _sometimes_ higher then I guess it's a weird quirk of the test net.

Beyond that I think the thing that I'm most worried here is that what happens if the number is more than two decimal places. I wouldn't want the UI to jump in that case. Like, I'm wondering if we do something like:

* If the estimated base fee is <= 0.01, show `<= 0.01 GWEI`.
* If the estimated base fee is >= 0.01, show `${baseFee} GWEI` where the base fee is rounded to 2 decimal places.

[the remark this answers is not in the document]
jpuri: Good idea, let me work on this.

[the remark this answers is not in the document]
jpuri: I made the change.

jpuri:
@mcmire , @darkwing , @NiranjanaBinoy : PR is updated and ready for re-review.

darkwing (dismissed):
LGTM!  One small suggestion would be to assign some of the numeric values as variables so we always know what they represent, but not a blocker.

jpuri:
> LGTM! One small suggestion would be to assign some of the numeric values as variables so we always know what they represent, but not a blocker.

Done 👍 , I hope we soon move to typescript to fix these issues. Thx for the feedback

ui/components/app/edit-gas-fee-popover/network-status/network-status.js
```diff
@@ -0,0 +1,58 @@
...
+
+import StatusSlider from './status-slider';
+
+const NetworkStatus = () => {
+  const { gasFeeEstimates } = useGasFeeContext();
+  let estBaseFee = 0;
+  if (gasFeeEstimates?.estimatedBaseFee) {
+    // estimatedBaseFee is not likely to be below 1, value .01 is used as test networks sometimes
+    // show have small values for it and more decimal places may cause UI to look broken.
+    estBaseFee = parseFloat(gasFeeEstimates?.estimatedBaseFee);
```
mcmire, line 16: I think we have to be careful here, because ether can't be represented properly using floats (for instance, 15 WEI is `15 * 10**18`, which gets corrupted into `1.5000000000000002e-16`). I think the best thing here might be to run this through BigNumber. I feel like you might have run across this code already but I think you can do that by using `toBigNumber.dec()` from `shared/modules/conversion.utils` (unless there's another way you know to do this). Of course if this is already a BigNumber then you're good to go! But once you have a BigNumber you can call `toFixed(2)` on that to format it (while keeping it as a BigNumber).

mcmire (commented):
Sorry, two more comments and I swear I'm done 😂

ui/components/app/edit-gas-fee-popover/network-status/network-status.js
```diff
@@ -0,0 +1,58 @@
...
+        fontWeight="bold"
+        margin={[3, 0]}
+        variant="h6"
+      >
+        <I18nValue messageKey="networkStatus" />
+      </Typography>
+      <div className="network-status__info">
+        <div className="network-status__info__field">
+          <span className="network-status__info__field-data">
+            {estBaseFee >= 0.01 && `${estBaseFee} GWEI`}
```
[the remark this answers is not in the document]
jpuri: That default was actually inspired from this feedback: https://github.com/MetaMask/metamask-extension/pull/12704#issuecomment-974998249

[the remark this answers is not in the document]
jpuri: I made change again

ui/components/app/edit-gas-fee-popover/network-status/network-status.js
```diff
@@ -0,0 +1,60 @@
...
+
+const NetworkStatus = () => {
+  const { gasFeeEstimates } = useGasFeeContext();
+  let estBaseFee = null;
+  if (gasFeeEstimates?.estimatedBaseFee) {
+    // estimatedBaseFee is not likely to be below 1, value .01 is used as test networks sometimes
+    // show have small values for it and more decimal places may cause UI to look broken.
+    estBaseFee = toBigNumber.dec(gasFeeEstimates?.estimatedBaseFee);
+    estBaseFee =
+      estBaseFee.greaterThan(0.01) < 0.01 ? 0.01 : estBaseFee.toFixed(2);
```
mcmire, line 19: `greaterThan`/`lessThan` already return a boolean, so you don't need to combine it with `<` or `>`. But also, should this be `estBaseFee.lessThan(0.01) ? estBaseFee : estBaseFee.toFixed(2)`? So basically, the thinking here is that if it's a very small value we go ahead and show that value, otherwise we chop it off at 2 decimal places.

jpuri: yep that was my bad, just fixed it. thx

darkwing (dismissed):

mcmire (approved):

darkwing (approved):

ui/components/app/edit-gas-fee-popover/edit-gas-fee-popover.js
```diff
@@ -54,6 +57,27 @@ const EditGasFeePopover = ({ onClose }) => {
...
               onClose={onClose}
             />
+            <NetworkStatus />
+            <Typography
+              className="edit-gas-fee-popover__know-more"
+              align="center"
+              color={COLORS.UI4}
+              fontSize="12px"
+            >
+              <I18nValue
```
adonesky1, line 67: Not a blocker, just curious, what was the impetus for creating this component over using the `useI18nContext` hook?

PR #12707, opened 2021-11-15
Allow sending up to 15 decimals on the send screen

danjm:
This PR allows entering up to 15 decimals on the send screen.


darkwing (dismissed):

jpuri:
Hey @danjm , the input internally uses `input` and number types in javascript have a limit of 16 decimal places. As I understand any value above 16 will not be effective.

mcmire (dismissed):
I agree with @jpuri on this. If we want to support this, we will have to change our numeric input. (See these for more explanations of why this is: https://docs.telerik.com/kendo-ui/controls/editors/numerictextbox/limitations, https://stackoverflow.com/questions/10249662/html5-input-type-number-16-digits-or-more-chrome-replaces-with-zeroes, etc.)

ui/components/ui/token-input/token-input.component.js
```diff
@@ -127,7 +129,7 @@ export default class TokenInput extends PureComponent {
     } else {
       // Display ETH
       currency = ETH;
-      numberOfDecimals = 6;
+      numberOfDecimals = primaryNumberOfDecimals;
```
NiranjanaBinoy, line 132: Use of decimal from token details instead of the hardcoding to 6 seems to be a good change to add even if we decide against decimal value as 18

danfinlay:
Why stop at 15 decimals? Why not allow full precision?

danjm:
@danfinlay see this https://github.com/MetaMask/metamask-extension/pull/12707#pullrequestreview-846199271... it is a lot more work to get 16+ decimals working nicely with our input

ui/components/app/currency-input/currency-input.js
```diff
@@ -25,12 +25,14 @@ import { getCurrentCurrency, getShouldShowFiat } from '../../../selectors';
...
  * @param options0.onChange
  * @param options0.onPreferenceToggle
+ * @param options0.primaryNumberOfDecimals
  */
 export default function CurrencyInput({
   hexValue,
   featureSecondary,
   onChange,
   onPreferenceToggle,
+  primaryNumberOfDecimals,
```
mcmire, line 35: Should we throw an error if `primaryNumberOfDecimals` is more than 16?

ui/pages/send/send-content/send-amount-row/send-amount-row.component.js
```diff
@@ -31,12 +31,14 @@ export default class SendAmountRow extends Component {
         onChange={this.handleChange}
         token={asset.details}
         value={amount}
+        primaryNumberOfDecimals={asset.decimals < 15 ? asset.decimals : 15}
```
mcmire, line 34: Should this be `asset.decimals <= 16 ? asset.decimals : 16`? I'm now second guessing my comments earlier — I'm not sure if the cutoff is 14, 15, or 16.

NiranjanaBinoy:
Copying further comments/conversations regarding the isuue :

from @mcmire :
> there is an inherent limitation with <input type="number"> fields. this is because a number is used to hold the value, and numbers in JavaScript can only be accurately represented up to 16 places. after that point, they are subject to rounding errors. so maybe the number you used happened to work because it could be represented correctly but not all numbers will work.

from @jpuri :
This PR has details about number input behavior for large decimal values. I spent much time on it back then: https://github.com/MetaMask/metamask-extension/pull/12140
In short, for decimals > 16, inputs behave weird - they will show all decimals but onChange will fail thus value in-state will be different from what you see.

Check this comment: https://github.com/MetaMask/metamask-extension/pull/12140#discussion_r711862121

ui/components/app/currency-input/currency-input.js
NiranjanaBinoy: @mcmire `primaryNumberOfDecimals` value sent from <SendAmountRow/> will either be 15 ( as we have set at the moment) or any value less than that, and we can also default to 8 (as it was the value we are currently replacing) or to 15 (to avoid confusion).

mcmire: I guess I want to make sure anyone who uses this in the future isn't confused. Is there anything stopping them from saying `<CurrencyInput primaryNumberOfDecimals={100} />`?

ui/components/app/currency-input/currency-input.js
```diff
@@ -127,7 +130,8 @@ export default function CurrencyInput({
     if (shouldUseFiat()) {
       // Display ETH
       currency = preferredCurrency || ETH;
-      numberOfDecimals = 8;
+      numberOfDecimals =
+        primaryNumberOfDecimals <= 16 ? primaryNumberOfDecimals : 16;
     } else {
```
jpuri, line 135: @NiranjanaBinoy : 
Can you plz confirm that the input now does not allow entering > 16 decimal digits ?

I am not sure but may be we should stop user at 15th decimal digit [example](https://github.com/MetaMask/metamask-extension/pull/12140/files#diff-45a7e71158bc214b6ff3d28ad60cc509f25e80efdfff6ad59f7734d0e6246d66R33), at 16th digit input becomes controlled.

mcmire (commented):
This looks good to me pending a rebase.

ui/components/app/currency-input/currency-input.js
jpuri: @NiranjanaBinoy : can you plz check point above, other than that PR looks good to me.

darkwing:
Nit that I hope we can change here -- we're hardcoding `16` in a lot of places -- can we make that a constant?

mcmire (dismissed):

mcmire (dismissed):

danjm (commented):
I approve Niranjana's work on this

NiranjanaBinoy (dismissed):

mcmire (approved):

NiranjanaBinoy (approved):
Approving on behalf of DanM.

PR #12717, opened 2021-11-16
Fix manifest i18n substitutions for Beta and Flask

rekmarks:
We noticed that the Chrome web store wasn't substituting our locale messages in the manifest for Beta and Flask builds. I believe this is because their i18n API expects the entire string value to be of the format `__MSG_messageKey__`. This PR updates the Beta and Flask manifest substitutions to match, by means of new locale messages.


Gudahtt (approved):
LGTM!

ryanml (approved):

PR #12722, opened 2021-11-16
Updates to network dropdown test network messaging

ryanml:
Screen Shot 2021-11-16 at 12 22 01 PM

### DIFF
--- a/ui/components/app/dropdowns/network-dropdown.js
+++ b/ui/components/app/dropdowns/network-dropdown.js
@@ -324,7 +324,7 @@ class NetworkDropdown extends Component {
                     history.push(`${ADVANCED_ROUTE}#show-testnets`);
                   }}
                 >
-                  {t('disableEnable')}
+                  {t('showHide')}
                 </a>,
               ])}
               <button
--- a/ui/components/app/dropdowns/network-dropdown.js
+++ b/ui/components/app/dropdowns/network-dropdown.js
@@ -313,10 +313,7 @@ class NetworkDropdown extends Component {
           <div className="network-dropdown-divider" />
           {showTestnetMessageInDropdown ? (
             <div className="network-dropdown-content">
-              {t('defaultNetwork', [
-                <span key="testNetworksEnabled">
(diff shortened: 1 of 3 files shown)


danjm (dismissed):

danjm:
probably needs a `yarn lint:fix`

Gudahtt:
The `test-lint-shellcheck` error appears to be CircleCI's fault. Hopefully it will work on retry.

tmashuang:
Is disabling/enable test networks terminology correct? Would show/hide be better?

ui/components/app/dropdowns/network-dropdown.js
```diff
@@ -327,14 +324,15 @@ class NetworkDropdown extends Component {
...
-                  {t('here')}
+                  {t('showHide')}
                 </a>,
               ])}
               <button
-                title={t('dismiss')}
                 onClick={hideTestNetMessage}
-                className="fas fa-times network-dropdown-content--close"
-              />
+                className="network-dropdown-content--dismiss"
```
Gudahtt, line 332: Nit: this was a pre-existing issue with the `--close` classname, but this name isn't correct according to BEM. The part after the double-dashes is for _modifiers_, like "hover", "pressed", "focused", "disabled", "invalid", etc. Temporary states.

A BEM-compliant name would be `network-dropdown-content__dismiss

Gudahtt (approved):
LGTM! Left one suggestion but it's not a blocker.

ui/components/app/dropdowns/network-dropdown.js
Gudahtt: Well, technicallty `network-dropdown__dismiss` would be better. The "content" part of the network dropdown should be an _element_, because it's not a standalone entity. The entire network dropdown seems like the better "block" in this component.

But that would require changing all of the classnames in this component, which seems even father outside of scope than my first suggestion. Would not recommend for this PR.

ryanml: Very good suggestions thanks, I can update this in a separate PR

tmashuang (approved):

PR #12729, opened 2021-11-17
Use v2 API for fiat onboarding URL creation, fix wrapping / unwrapping

dan437:
## Description
This PR contains 2 things:
- Use a Swaps v2 APIs to get a fiat onboarding URL, so we can deprecate old APIs
- Fix an issue with wrapping / unwrapping if an address contained uppercase chars

#### v2 API
- Click on "Buy" on the homepage when Ethereum Mainnet network is selected
- Click on "Continue to Wyre"
- New tab with Wyre fiat onboarding is opened and your Ethereum address is prepopulated
- Network tab in MM shows a call to this v2 URL: `https://api2.metaswap.codefi.network/networks/1/fiatOnRampUrl`

#### Wrapping / unwrapping
Homepage:
- On the homepage click on on the "Swap" button 
- Try swaps for ETH -> WETH and WETH -> ETH
- They should both show just one quote from backend (`/trades` API call with the `enableDirectWrapping: true` query param)

Activity tab:
- On the homepage in the Activity tab click on WETH and then on "Swap"
- Try to swap WETH -> ETH
- It should show just one quote from backend (`/trades` API call with the `enableDirectWrapping: true` query param)

### DIFF
--- a/app/scripts/lib/buy-eth-url.test.js
+++ b/app/scripts/lib/buy-eth-url.test.js
@@ -28,8 +28,10 @@ const KOVAN = {
 
 describe('buy-eth-url', function () {
   it('returns Wyre url with an ETH address for Ethereum mainnet', async function () {
-    nock('https://api.metaswap.codefi.network')
-      .get(`/fiatOnRampUrl?serviceName=wyre&destinationAddress=${ETH_ADDRESS}`)
+    nock('https://api2.metaswap.codefi.network')
+      .get(
+        `/networks/1/fiatOnRampUrl?serviceName=wyre&destinationAddress=${ETH_ADDRESS}`,
+      )
       .reply(200, {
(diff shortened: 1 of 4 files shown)


app/scripts/lib/buy-eth-url.test.js
```diff
@@ -28,8 +28,10 @@ const KOVAN = {
 
 describe('buy-eth-url', function () {
   it('returns Wyre url with an ETH address for Ethereum mainnet', async function () {
-    nock('https://api.metaswap.codefi.network')
-      .get(`/fiatOnRampUrl?serviceName=wyre&destinationAddress=${ETH_ADDRESS}`)
+    nock('https://api2.metaswap.codefi.network')
```
darkwing, line 31: Can the `SWAPS_API_V2_BASE_URL` constant be used here?

darkwing (dismissed):
Small suggestion, not a blocker.

danjm (approved):

darkwing (approved):

PR #12731, opened 2021-11-17
Remove unused background API methods

Gudahtt:
These background API methods were not used anywhere in the UI. One of them was called in `actions.js` by a function that itself was never called, so it have been removed. Additionally, one unused `actions.js` function was found and removed as well.

`setAdvancedGasFee` is the only unused background method that remains. It was recently added and will be used in the near future.


rekmarks (approved):
LGTM!

ryanml (approved):

PR #12733, opened 2021-11-17
Confirm transaction screen - wrap lengthier values

### LINKED ISSUE #12598: Confirm transaction screen - wrapping lengthier value
On confirm transaction screen wrap bigger values as shown below:

Screenshot 2021-11-05 at 5 05 20 PM

jpuri:
Fixes: #12598

Wrap lengthier values on confirm transaction screen:

Screenshot 2021-11-18 at 12 46 46 AM

Screenshot 2021-11-18 at 12 45 59 AM

### DIFF
--- a/ui/components/app/transaction-detail-item/index.scss
+++ b/ui/components/app/transaction-detail-item/index.scss
@@ -5,12 +5,15 @@
 
   &__row {
     display: flex;
+    justify-content: space-between;
     grid-gap: 5px;
   }
 
-  &__title {
-    flex-grow: 1;
-    word-break: break-word;
+  &__detail-values {
+    display: flex;
+    flex-wrap: wrap;
+    justify-content: end;
+    width: 65%;
   }
 
   .info-tooltip {
@@ -20,19 +23,6 @@
     path {
       fill: $ui-3;
     }
-  }
-
-  &__total {
-    font-weight: bold;
-    color: $ui-black;
-  }
-
-  &__subtitle {
-    flex-grow: 1;
-  }
-
-  &__subtext {
-    text-align: end;
   }
 
   .currency-display-component {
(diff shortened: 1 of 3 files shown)


ui/components/app/transaction-detail-item/index.scss
```diff
@@ -22,9 +29,10 @@
     }
   }
 
-  &__total {
+  & h6.transaction-detail-item__total {
     font-weight: bold;
     color: $ui-black;
+    margin-left: 4px;
   }
```
jpuri, line 36: Change above is to make this rule more specific.

ui/components/app/transaction-detail-item/index.scss
```diff
@@ -22,9 +29,10 @@
     }
   }
 
-  &__total {
+  & h6.transaction-detail-item__total {
     font-weight: bold;
     color: $ui-black;
```
NiranjanaBinoy, line 34: Hi, I understand that the `font-weight` and `color` might be pre-existing but aren't they already covered in <Typography/> ?

NiranjanaBinoy (dismissed):

ui/components/app/transaction-detail-item/index.scss
mcmire: Hmm. We should be able to write CSS in a way that doesn't require these kinds of specificity overrides. What were you needing to override?

ui/components/app/transaction-detail-item/index.scss
jpuri: yep, it was not noticed as values overriding them were similar.

ui/components/app/transaction-detail-item/index.scss
jpuri: Yep I agree this should ideally not be needed, this classes are styled in couple of components - that makes it hard to fix. Let me spend some more time on it.

jpuri: It is fixed now and code looks cleaner :)

jpuri:
@mcmire , @NiranjanaBinoy : can you plz re-review the PR ?

darkwing (approved):

mcmire (approved):
Looks good!

PR #12737, opened 2021-11-18
ci - enforce yarn lock deduplications

kumavis:
Adds a CI task that requires a PR to have applied all yarn deduplications.

removes 0.7 gig from `node_modules` (!)

By adding this as a workflow requirement, we can ensure we have a minimal number of duplicate deps, which means:
:zap: faster dep installs
:bird: smaller build sizes
:adhesive_bandage: fewer patches
:female_detective: more easily auditable lavamoat policies (for future lavamoat versions)

I recommend we merge this CI task now as an advisory/recommendation and then upgrade it to a requirement later


patches/sass+1.35.2.patch
```diff
@@ -0,0 +1,36 @@
...
+-        Object.defineProperty($constructor, t1, {value: C.UnknownJavaScriptObject_methods, enumerable: false, writable: true, configurable: true});
+-        return C.UnknownJavaScriptObject_methods;
+-      }
++      // if (typeof $constructor == "function") {
++      //   t1 = $._JS_INTEROP_INTERCEPTOR_TAG;
++      //   if (t1 == null)
++      //     t1 = $._JS_INTEROP_INTERCEPTOR_TAG = init.getIsolateTag("_$dart_js");
++      //   Object.defineProperty($constructor, t1, {value: C.UnknownJavaScriptObject_methods, enumerable: false, writable: true, configurable: true});
++      //   return C.UnknownJavaScriptObject_methods;
++      // }
```
rekmarks, line 33: Should we just delete these instead?

kumavis: deleted

rekmarks (approved):
I can't fault it!

ui/helpers/utils/optimism/buildUnserializedTransaction.test.js
```diff
@@ -14,15 +13,13 @@ describe('buildUnserializedTransaction', () => {
...
-    expect(unserializedTransaction).toMatchObject({
-      nonce: new BN('00', 16),
-      gasPrice: new BN('64', 16),
-      gasLimit: new BN('5208', 16),
-      to: expect.objectContaining({
-        buf: Buffer.from(times(20, 0)),
-      }),
-      value: new BN('09184e72a000', 16),
-      data: Buffer.from([0]),
+    expect(unserializedTransaction.toJSON()).toMatchObject({
```
Gudahtt, line 16: What motivated this change? Has `unserializedTransaction` changed with a dependency update?

kumavis: maybe some BN version mismatch? it wasn't obvious why, but the BNs did not match each other any more

Gudahtt: Ahh I see. So that's why `toJSON` is used, to stringify the numbers.

Gudahtt (approved):
LGTM!

PR #12310, opened 2021-10-08
use improved-yarn-audit and exclude 1002401 and 1002581

adonesky1:
Explanation:  There are currently a couple of npm security advisories on develop [1002401](https://www.npmjs.com/advisories/1002401) and [1002581](https://www.npmjs.com/advisories/1002581). Both are Regex denial of service vulnerabilities so not security issues for user funds. 

Per this [slack thread](https://consensys.slack.com/archives/GTQAGKY5V/p1633643644222800) it was decided that we would follow [a precedent set by mobile](https://github.com/MetaMask/metamask-mobile/pull/2742) to exclude certain low level vulnerabilities with an audit script wrapper until the more complex process of fixing the underlying issues can be groomed.

PR #12738, opened 2021-11-18
ci - fix yarn-audit

kumavis:
this seems to have been broken since https://github.com/MetaMask/metamask-extension/pull/12310

though, we employ `shellcheck` to catch this sort of things, not sure why this is not enforced


Gudahtt (approved):
LGTM, good catch!

PR #12742, opened 2021-11-18
Improve code fence transform error handling

rekmarks:
This PR improves the error handling of the code fence removal transform stream by catching errors thrown by the `removeFencedCode` function and passing them to the `end` callback. This appears to resolve a problem where watched builds would blow up whenever a file with fences was reloaded.

### DIFF
--- a/development/build/transforms/remove-fenced-code.js
+++ b/development/build/transforms/remove-fenced-code.js
@@ -58,12 +58,11 @@ class RemoveFencedCodeTransform extends Transform {
     };
 
     if (this.shouldLintTransformedFiles && didModify) {
-      lintTransformedFile(fileContent, this.filePath)
+      return lintTransformedFile(fileContent, this.filePath)
         .then(pushAndEnd)
         .catch((error) => end(error));
-    } else {
-      pushAndEnd();
     }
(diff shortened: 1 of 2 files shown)


development/build/transforms/remove-fenced-code.js
```diff
@@ -41,11 +41,16 @@ class RemoveFencedCodeTransform extends Transform {
...
-    );
+    let fileContent, didModify;
+    try {
+      [fileContent, didModify] = removeFencedCode(
+        this.filePath,
+        this.buildType,
+        Buffer.concat(this._fileBuffers).toString('utf8'),
+      );
+    } catch (error) {
+      end(error);
```
Gudahtt, line 52: We should probably `return` here as well, otherwise `end` will get called again below.

```suggestion
      return end(error);
```

rekmarks: Damn, I'm kind of shocked that our linter didn't blow up on this!

Gudahtt (dismissed):
LGTM!

Edit: Oh, one more lint issue

Gudahtt (approved):

darkwing (approved):

PR #12750, opened 2021-11-18
1559 v2 tooltips

### LINKED ISSUE #12588: EIP-1559 V2: Edit gas modal select estimate section - create required tooltips
Edit gas fee modal required quire a few tooltips in section where user select gas option as shown below:

Screenshot 2021-11-05 at 12 23 53 PM

The task includes building these.

Ref: https://www.figma.com/file/DFlkrM9JRyP0hDrd4ykMZW/1559-Gas-Controls---v2.1?node-id=690%3A11306

adonesky1:
Fixes: #12588

Explanation:  Adds explanatory Tooltips for EIP1559 v2 Gas Modal settings.
Screen Shot 2021-11-19 at 11 31 37 AM

Screen Shot 2021-11-19 at 11 31 49 AM

Screen Shot 2021-11-19 at 11 31 56 AM

Screen Shot 2021-11-19 at 11 32 05 AM

Screen Shot 2021-11-19 at 11 32 18 AM


ui/components/app/edit-gas-fee-popover/edit-gas-item/edit-gas-item.js
```diff
@@ -90,6 +101,41 @@ const EditGasItem = ({ priorityLevel, onClose }) => {
     onClose();
   };
```
jpuri, line 103: @adonesky1 : I would prefer to have here s separate small functional component that returns tooltip.

ui/components/app/edit-gas-fee-popover/edit-gas-item/edit-gas-item.js
```diff
@@ -40,13 +45,19 @@ const EditGasItem = ({ priorityLevel, onClose }) => {
...
-    maxFeePerGas = hexWEIToDecGWEI(dappSuggestedGasFees.maxFeePerGas);
+    maxFeePerGas = hexWEIToDecGWEI(
+      dappSuggestedGasFees.maxFeePerGas ?? dappSuggestedGasFees.gasPrice,
+    );
     maxPriorityFeePerGas = hexWEIToDecGWEI(
-      dappSuggestedGasFees.maxPriorityFeePerGas,
+      dappSuggestedGasFees.maxPriorityFeePerGas ??
+        // TODO should this default back to something else? Will definitely be too high in this case
+        dappSuggestedGasFees.gasPrice,
     );
```
jpuri, line 61: @adonesky1 : new flow should be used for only transactions supporting EIP-1559

ui/components/app/edit-gas-fee-popover/edit-gas-item/edit-gas-item.js
```diff
@@ -136,7 +182,89 @@ const EditGasItem = ({ priorityLevel, onClose }) => {
...
+              priorityLevel !== PRIORITY_LEVELS.DAPP_SUGGESTED ? (
+                <img
+                  alt=""
+                  width={130}
+                  src={`./images/curve-${priorityLevel}.svg`}
+                />
+              ) : null}
+              {priorityLevel === PRIORITY_LEVELS.HIGH ? (
+                <div className="edit-gas-item__tooltip__container__dialog">
+                  <Typography fontSize="12px" color={COLORS.WHITE}>
```
darkwing, line 198: `0.75rem` instead?

ui/components/app/edit-gas-fee-popover/edit-gas-item/index.scss
```diff
@@ -51,6 +51,54 @@
...
+        width: 50%;
+      }
+
+      &__value {
+        width: 50%;
+      }
+
+      p {
+        margin-bottom: 0 !important;
+      }
```
jpuri, line 88: It should be possible to avoid `!important` I think

ui/components/app/edit-gas-fee-popover/edit-gas-item/index.scss
```diff
@@ -51,6 +51,54 @@
     text-align: right;
     width: 10%;
```
jpuri, line 52: I think will be useful to extract styles also like

- edit-gas-tooltip
     - edit-gas-tooltip.js 
     - edit-gas-tooltip.test.js 
     - edit-gas-tooltip.scss

ui/components/app/edit-gas-fee-popover/edit-gas-item/edit-gas-item.js
```diff
@@ -136,7 +144,17 @@ const EditGasItem = ({ priorityLevel, onClose }) => {
...
+          priorityLevel={priorityLevel}
+          gasContext={{
+            gasLimit,
+            maxFeePerGas,
+            maxPriorityFeePerGas,
+            maxFeePerGasValue,
+            maxPriorityFeePerGasValue,
+            origin,
+          }}
+        />
```
jpuri, line 157: we can use `useGasFeeContext` in `EditGasToolTip` to reduce these props to just 2.

app/images/curve-low.svg
```diff
@@ -0,0 +1,10 @@
+<svg width="125" height="31" viewBox="0 0 125 31" fill="none" xmlns="http://www.w3.org/2000/svg">
```
jpuri, line 1: Will be useful to optimise the svg: https://jakearchibald.github.io/svgomg/

ui/components/app/edit-gas-fee-popover/edit-gas-tooltip.js
```diff
@@ -0,0 +1,143 @@
...
+          <span key={priorityLevel}>
+            <b>{t('high')}</b>
+          </span>,
+        ]);
+      case PRIORITY_LEVELS.CUSTOM:
+        return t('customGasSettingToolTipMessage', [
+          <span key={priorityLevel}>
+            <b>{t('custom')}</b>
+          </span>,
+        ]);
```
jpuri, line 46: First 4 cases can be reduced to 

```
case PRIORITY_LEVELS.CUSTOM:
        return t(`${priorityLevel}GasSettingToolTipMessage`, [
          <span key={priorityLevel}>
            <b>{t(priorityLevel)}</b>
          </span>,
        ]);
```

adonesky1: I originally had it like this but `yarn verify-locales` didn't recognize that these keys were being used... so had to use this much less satisfying and concise approach 😢

ui/components/app/edit-gas-fee-popover/edit-gas-item/edit-gas-item.js
adonesky1: I thought it would be better to avoid additional calls to `useGasFeeInputs` (which makes network requests in `useGasFeeEstimates`) and since we have all of the data right here we might as well just pass it along?

app/images/curve-low.svg
adonesky1: Good call!

ui/components/app/edit-gas-fee-popover/edit-gas-item/edit-gas-item.js
adonesky1: unfortunately `rems` don't appear to work as a prop to the `Typography` component...

ui/components/app/edit-gas-fee-popover/edit-gas-item/index.scss
adonesky1: Unfortunately I'm not finding a different way to override this `margin-bottom` setting set here otherwise: https://github.com/MetaMask/metamask-extension/blob/42fea3d32329f55004d9038c93db51e130a4f4da/ui/components/ui/info-tooltip/index.scss#L56

ui/components/app/edit-gas-fee-popover/edit-gas-item/index.scss
adonesky1: 👍

ui/components/app/edit-gas-fee-popover/edit-gas-item/edit-gas-item.js
jpuri: I think including `useGasFeeContext` does not call `useGasFeeInputs` each time. It will only return latest state in the context.

ui/components/app/edit-gas-fee-popover/edit-gas-item/index.scss
jpuri: typography also allow passing margin, does that one helps ?

ui/components/app/edit-gas-fee-popover/edit-gas-tooltip/edit-gas-tooltip.test.js
```diff
@@ -0,0 +1,82 @@
...
+const renderComponent = (props) => {
+  const mockStore = {
+    metamask: {
+      provider: {
+        type: 'test',
+      },
+    },
+  };
+
+  const store = configureMockStore()(mockStore);
```
jpuri, line 43: May be use `configureStore` like here: https://github.com/MetaMask/metamask-extension/blob/develop/ui/components/app/edit-gas-fee-popover/edit-gas-item/edit-gas-item.test.js#L50

@mcmire suggested this for unit tests and I think is better over using mock store.

ui/components/app/edit-gas-fee-popover/edit-gas-item/edit-gas-item.js
mcmire: Yes, can confirm ☝🏻

ui/components/app/edit-gas-fee-popover/edit-gas-tooltip/edit-gas-tooltip.js
```diff
@@ -0,0 +1,145 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { PRIORITY_LEVELS } from '../../../../../shared/constants/gas';
+import {
+  COLORS,
+  FONT_WEIGHT,
+} from '../../../../helpers/constants/design-system';
+import Typography from '../../../ui/typography';
+
+const EditGasToolTip = ({ priorityLevel, gasContext, t }) => {
```
mcmire, line 10: Do you need to pass `t` here? I think we can get this from either the i18n context or using Jyoti's new `I18nValue` component, right?

ui/components/app/edit-gas-fee-popover/edit-gas-tooltip/edit-gas-tooltip.test.js
```diff
@@ -0,0 +1,82 @@
...
+
+  return renderWithProvider(
+    <EditGasToolTip {...props} t={sinon.stub()} />,
```
mcmire, line 46: Do you need to pass `t` into the component? If you `import { renderWithProvider } from '../../../../../test/jest'` rather than `import { renderWithProvider } from '../../../../../test/lib/render-helpers'` I think `t` is already stubbed so you shouldn't have to worry about this.

ui/components/app/edit-gas-fee-popover/edit-gas-tooltip/index.scss
```diff
@@ -0,0 +1,53 @@
+.tooltip {
```
mcmire, line 1: I feel like because our CSS is global, the name of the class should match the name of the component to avoid naming collisions. Is it possible for this to be `.edit-gas-tooltip` instead? Or should this be somewhere else?

mcmire (changes requested):
Had a question about CSS.

ui/components/app/edit-gas-fee-popover/edit-gas-item/edit-gas-item.js
adonesky1: ah ok nice! 👍

ui/components/app/edit-gas-fee-popover/edit-gas-item/index.scss
adonesky1: Unfortunately it doesn't seem to... I'm actually already passing in `margin={0}` to all the typography components, which I honestly don't think is doing anything at this point. Going to remove them in fact.

ui/components/app/edit-gas-fee-popover/edit-gas-item/edit-gas-item.js
adonesky1: I guess the other problem here is that if I use the useGasFeeContext rather than passing along the values, I will have to rerun all [this logic](https://github.com/MetaMask/metamask-extension/blob/f65063ac6ca41bc7ddef78ceda1f6751bc291d3b/ui/components/app/edit-gas-fee-popover/edit-gas-item/edit-gas-item.js#L44) (lines 38 - 64) to get the correct values? cc @jpuri. Am I wrong?

ui/components/app/edit-gas-fee-popover/edit-gas-tooltip/edit-gas-tooltip.js
adonesky1: So this was driving me crazy yesterday. For some reason the context is not accessible in this component? It seems it must have something to do with the `react-tippy` package we're using for this tool-tip. I'll try using Jyoti's I18nValue component, but I might need some help digging into why I can only get it working this way...

ui/components/app/edit-gas-fee-popover/edit-gas-tooltip/index.scss
adonesky1: agreed!

ui/components/app/edit-gas-fee-popover/edit-gas-item/index.scss
adonesky1: Here's what I get without the `!important`
Screen Shot 2021-11-23 at 12 36 21 PM

ui/components/app/edit-gas-fee-popover/edit-gas-tooltip/edit-gas-tooltip.js
mcmire: Ah interesting. I might not have time to help today, but I'll see if I can dive into this tomorrow. Or @jpuri have you seen this before?

adonesky1: Screen Shot 2021-11-23 at 12 40 57 PM
so here is what I'm seeing ('maxBaseFee' => '[maxBaseFee]') when I either use the `I18nValue` component or the `useI18nContext` hook directly...

ui/components/app/edit-gas-fee-popover/edit-gas-tooltip/edit-gas-tooltip.test.js
adonesky1: this is required because of that bug I'm dealing with the translationContext not working in this component, I've had to make the component expect a method prop called `t`.

ui/components/app/edit-gas-fee-popover/edit-gas-item/index.scss
jpuri: Margin in typography does work for me, let me check again

ui/components/app/edit-gas-fee-popover/edit-gas-item/edit-gas-item.js
jpuri: @adonesky1 : values used directly from context can be avoided to be passed, for others it is ok to pass I think

ui/components/app/edit-gas-fee-popover/edit-gas-tooltip/edit-gas-tooltip.js
jpuri: Yep `I18nValue` does not works inside tooltip I guess some context issue, but it should be possible to get `t` from i18nContext here

ui/components/app/edit-gas-fee-popover/edit-gas-item/index.scss
adonesky1: @jpuri Were you able to get it to work in the context of this PR?

ui/components/app/edit-gas-fee-popover/edit-gas-tooltip/edit-gas-tooltip.js
adonesky1: Also fails :(

ui/components/app/edit-gas-fee-popover/edit-gas-item/edit-gas-item.js
adonesky1: @jpuri made this change. Let me know what you think.

ui/components/app/edit-gas-fee-popover/edit-gas-tooltip/index.scss
```diff
@@ -0,0 +1,53 @@
...
+      padding: 4px;
+    }
+
+    &__label {
+      width: 50%;
+    }
+
+    &__value {
+      width: 50%;
+    }
```
darkwing, line 36: Nit:

```
&__label, &__value {
      width: 50%;
    }
```

ui/components/app/edit-gas-fee-popover/edit-gas-item/edit-gas-item.js
jpuri: Looks good

jpuri: Oh I found that even this context is not working in tooltips, we need to revert this change :(

ui/components/app/edit-gas-fee-popover/edit-gas-item/index.scss
jpuri: Yeh it does not works. I see bit of issue here, our typography components are block components, we are using those as inline here, it should rather be a typography component here wrapping inline spans / strong tags wrapped in it.

app/images/curve-high.svg
```diff
@@ -0,0 +1 @@
...-.314-.78-.54a2.623 2.623 0 0 1-.52-.81 2.8 2.8 0 0 1-.18-1.01c0-.354.06-.684.18-.99.126-.314.3-.584.52-.81.22-.234.48-.414.78-.54.3-.134.623-.2.97-.2.373 0 .713.083 1.02.25.306.16.546.366.72.62v-.77h.86v4.71c0 .42-.067.783-.2 1.09-.127.313-.304.57-.53.77-.227.206-.494.36-.8.46-.307.106-.64.16-1 .16Zm4.149-2.19v-7.5h.86v3.27c.173-.3.4-.52.68-.66s.583-.21.91-.21c.28 0 .537.046.77.14.233.093.43.23.59.41.167.173.297.386.39.64.093.246.14.523.14.83v3.08h-.85v-2.95c0-.42-.11-.75-.33-.99-.213-.247-.497-.37-.85-.37-.2 0-.39.04-.57.12-.173.08-.327.2-.46.36-.127.153-.23.343-.31.57-.073.226-.11.486-.11.78v2.48h-.86Z" fill="#F66A0A"/><path opacity=".3" d="M19.506 22.805c-10.763 7.42-19.416 8-19.416 8h110.25s-8.653-.58-19.417-8c-10.763-7.42-23.363-22-35.708-22-12.345 0-24.945 14.58-35.709 22Z" fill="#037DD6"/><mask id="a" style="mask-type:alpha" maskUnits="userSpaceOnUse" x="0" y="0" width="111" height="31"><path d="M19.506 22.672c-10.763 7.42-19.416 8-19.416 8h110.25s-8.653-.58-19.417-8c-10.763-7.42-23.363-22-35.708-22-12.345 0-24.945 14.58-35.709 22Z" fill="#EAF6FF"/></mask><g mask="url(#a)"><path fill="#F66A0A" stroke="#fff" stroke-width="2" d="M91.986-5.143h20.706v39.25H91.986z"/></g></svg>
```
mcmire, line 1: Is it customary for us to add `width` and `height` to our icons? Seems like that may prevent us from properly scaling those icons (by overriding the dimensions via CSS). I guess if it hasn't been a problem up to now then this is fine but I am curious all the same.

ui/components/app/edit-gas-fee-popover/edit-gas-tooltip/edit-gas-tooltip.js
```diff
@@ -0,0 +1,140 @@
...
+  // related to the source of the estimates. We pass these values from the
+  // the parent component (edit-gas-item) rather than recalculate them
+  maxFeePerGas,
+  maxPriorityFeePerGas,
+  t,
+}) => {
+  const {
+    gasLimit,
+    maxFeePerGas: maxFeePerGasValue,
+    maxPriorityFeePerGas: maxPriorityFeePerGasValue,
```
mcmire, line 23: Despite the comment above, it seems that we are still recalculating them, so what is the purpose of also passing them down? Sorry if I'm being dense.

ui/components/app/edit-gas-fee-popover/edit-gas-tooltip/edit-gas-tooltip.test.js
```diff
@@ -0,0 +1,98 @@
...
+  };
+
+  const store = configureStore(mockStore);
+
+  return renderWithProvider(
+    <GasFeeContextProvider
+      transaction={{ txParams: { gas: '0x5208' }, ...transactionProps }}
+      {...gasFeeContextProps}
+    >
+      <EditGasToolTip {...props} t={sinon.stub()} />
```
mcmire, line 61: Do we want to use `jest.fn()` here instead of sinon?

mcmire (dismissed):
A few comments but overall this looks good to me.

jpuri (dismissed):

ui/components/app/edit-gas-fee-popover/edit-gas-tooltip/edit-gas-tooltip.js
adonesky1: So here were getting the `maxFeePerGas` and `maxPriorityFeePerGas` from the gasContext and renaming them to `maxFeePerGasValue` and `maxPriorityFeePerGasValue`  to use them as fallback values to the passed down values that are derived from logic on lines 45-67 in `edit-gas-item.js`

mcmire: Ah, ok, gotcha.

mcmire (approved):
Seems good to me 👍🏻

jpuri (approved):

PR #12770, opened 2021-11-19
Fix issue where contractExchangeRates are not available in SwapsController

adonesky1:
- Fix issue where contractExchangeRates are not available in swaps controller
- Make token address matching case insensitive

Explanation:  I introduced a bug in https://github.com/MetaMask/metamask-extension/commit/9d968bae1ea256e3e50e644f7ec3422a0d98628d# where Token Exchange Rates (from the TokenRatesController) were no longer being correctly accessed in the Swaps controller. This PR corrects this issue and further ensures that use of contractExchangeRates object will work properly regardless of any casing differences of tokenAddresses across the controllers.

Manual testing steps:  
  - Build a quote for a swap
  - click into "quotes" to view details of each quote
Screen Shot 2021-11-19 at 1 21 47 PM
  - Make sure the **best quote** is selected based on a combination of amount of token received **and** fees (might require a calculator)
Screen Shot 2021-11-19 at 1 22 29 PM

### DIFF
--- a/app/scripts/controllers/swaps.js
+++ b/app/scripts/controllers/swaps.js
@@ -28,6 +28,7 @@ import {
 } from '../../../ui/pages/swaps/swaps.util';
 import fetchWithCache from '../../../ui/helpers/utils/fetch-with-cache';
 import { MINUTE, SECOND } from '../../../shared/constants/time';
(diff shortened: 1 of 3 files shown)


dan437 (approved):
I've tried it and it seems to be working correctly. Thank you!

danjm (approved):

PR #12776, opened 2021-11-20
Removing the token detection announcement when the token detection is ON

NiranjanaBinoy:
Currently, the token detection announcement is visible even when the token detection is ON, this PR will change that to make the announcement disappear when once the token detection is turned ON, and re-appear when the user turns OFF the token detection.


danjm (approved):

mcmire (approved):
Makes sense to me.

PR #12784, opened 2021-11-22
Proper calculation of the gas limit

### LINKED ISSUE #12072: [BUG] When sending to a smart contract address, modifying original send amount will result in a failed transaction
When using the Metamask extension for Chrome to send to a smart contract address, the gas limit can be reset unknowingly by the user by modifying the original send amount to something different. This changes the gas limit from 37581 to 21000, thus resulting in a transaction failure "Warning! Error encountered during contract execution [Out of gas]" after submitting. Quite disappointing to lose $10 this way, so I thought I'd submit a bug report about it at least so it can be fixed. Ideally, there should be some sort of warning or indicator confirming the gas limit total matches what one would need for a contract interaction on the final send page before one hits the "Confirm" button.

1. Using MM extension, start a new send to a smart contract recipient address
2. Enter an amount to send and hit "Next"
3. Click the "< Edit" button at the top left
4. Modify the send amount and hit "Next"
5. You'll now see the gas limit has been reset to the default 21000, which is not enough gas for any contract interaction

 - OS: Windows 10
 - Hardware Wallet: N/A
 - Browser: Chrome Version 93.0.4577.63 (Official Build) (64-bit)
 - MetaMask Version: 10.0.3

VSaric:
fixes: #12072

**Explanation:**  Proper calculation of the gas limit filed on the amount change and added warning message when the amount is less then calculated gas limit on manually enter the gas limit. When user manually enter the gas limit, and when that value is less then proper gas limit value, it shows warning message on the top of "Edit priority" popover, when the value is equal or greater then proper gas limit value, it not show warning message on the top of "Edit priority" popover. 

**Steps to reproduce (REQUIRED):**

1. Using MM extension, start a new send to a smart contract recipient address
2. Enter an amount to send and hit "Next"
3. Click the "< Edit" button at the top left
4. Modify the send amount and hit "Next"
5. Click "Edit" in "Estimated gas fee" field
6. Enter manually value in "Gas Limit" field to see warning message

**Screencast where the user has already explicitly set the gas limit:**
This is the second case related to Dan's comment: https://github.com/MetaMask/metamask-extension/pull/12784#discussion_r761031633. 

https://user-images.githubusercontent.com/92527393/145834940-7e6daf08-77d6-4333-9ecc-6886861fb6f2.mp4

**Screencast where gas limit was explicitly proposed by a dapp:**
This is the first case related to Dan's comment: https://github.com/MetaMask/metamask-extension/pull/12784#discussion_r761031633. 

https://user-images.githubusercontent.com/92527393/145835391-8611aadd-583a-4937-b4d0-0955369e9adb.mp4

**Warning message example in screencast:**

Screenshot from 2021-12-15 15-14-26

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -1167,6 +1167,9 @@
   "gasLimitInfoTooltipContent": {
     "message": "Gas limit is the maximum amount of units of gas you are willing to spend."
   },
+  "gasLimitIsLessThenCalculatedGasLimit": {
+    "message": "Gas limit must be equal or greater than $1"
+  },
   "gasLimitTooLow": {
     "message": "Gas limit must be at least 21000"
   },
@@ -1179,9 +1182,6 @@
   },
   "gasOption": {
     "message": "Gas option"
-  },
-  "gasLimitIsLessThenCalculatedGasLimit": {
-    "message": "Gas limit must be equal or greater than $1"
   },
   "gasPrice": {
     "message": "Gas Price (GWEI)"
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -1180,6 +1180,9 @@
   "gasOption": {
     "message": "Gas option"
   },
+  "gasLimitIsLessThenCalculatedGasLimit": {
+    "message": "Gas limit must be equal or greater than $1"
+  },
   "gasPrice": {
     "message": "Gas Price (GWEI)"
   },
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -1167,8 +1167,8 @@
   "gasLimitInfoTooltipContent": {
     "message": "Gas limit is the maximum amount of units of gas you are willing to spend."
   },
-  "gasLimitIsLessThenCalculatedGasLimit": {
-    "message": "Gas limit must be equal or greater than $1"
+  "gasLimitRecommended": {
+    "message": "Recommended gas limit is $1. If the gas limit is less than estimated transaction can fail."
   },
   "gasLimitTooLow": {
     "message": "Gas limit must be at least 21000"
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -1168,7 +1168,7 @@
     "message": "Gas limit is the maximum amount of units of gas you are willing to spend."
   },
   "gasLimitRecommended": {
-    "message": "Recommended gas limit is $1. If the gas limit is less than estimated transaction can fail."
+    "message": "Recommended gas limit is $1. If the gas limit is less that, it may fail."
   },
   "gasLimitTooLow": {
     "message": "Gas limit must be at least 21000"
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -1168,7 +1168,7 @@
     "message": "Gas limit is the maximum amount of units of gas you are willing to spend."
   },
   "gasLimitRecommended": {
-    "message": "Recommended gas limit is $1. If the gas limit is less that, it may fail."
+    "message": "Recommended gas limit is $1. If the gas limit is less than that, it may fail."
   },
   "gasLimitTooLow": {
     "message": "Gas limit must be at least 21000"
--- a/ui/components/app/edit-gas-display/edit-gas-display.component.js
+++ b/ui/components/app/edit-gas-display/edit-gas-display.component.js
@@ -111,9 +111,7 @@ export default function EditGasDisplay({
   let showTopWarning, warningMessage;
   if (gasLimit < properGasLimit) {
     showTopWarning = true;
-    warningMessage = t('gasLimitIsLessThenCalculatedGasLimit', [
-      properGasLimit,
-    ]);
+    warningMessage = t('gasLimitRecommended', [properGasLimit]);
   }
 
   const showTopError =
--- a/ui/components/app/edit-gas-display/edit-gas-display.component.js
+++ b/ui/components/app/edit-gas-display/edit-gas-display.component.js
@@ -111,7 +111,9 @@ export default function EditGasDisplay({
   let showTopWarning, warningMessage;
   if (gasLimit < properGasLimit) {
     showTopWarning = true;
-    warningMessage = t('gasLimitIsLessThenCalculatedGasLimit', [properGasLimit]);
+    warningMessage = t('gasLimitIsLessThenCalculatedGasLimit', [
+      properGasLimit,
+    ]);
   }
 
   const showTopError =
--- a/ui/components/app/edit-gas-display/edit-gas-display.component.js
+++ b/ui/components/app/edit-gas-display/edit-gas-display.component.js
@@ -108,9 +108,8 @@ export default function EditGasDisplay({
       dappSuggestedAndTxParamGasFeesAreTheSame,
   );
 
-  let showTopWarning, warningMessage;
+  let warningMessage;
   if (gasLimit < properGasLimit) {
-    showTopWarning = true;
     warningMessage = t('gasLimitRecommended', [properGasLimit]);
   }
 
@@ -153,7 +152,7 @@ export default function EditGasDisplay({
             <ErrorMessage errorKey={errorKey} />
           </div>
         )}
-        {showTopWarning && (
+        {warningMessage && (
           <div className="edit-gas-display__warning">
             <ActionableMessage
               className="actionable-message--warning"
--- a/ui/components/app/edit-gas-display/edit-gas-display.component.js
+++ b/ui/components/app/edit-gas-display/edit-gas-display.component.js
@@ -2,6 +2,7 @@ import React, { useContext, useLayoutEffect, useRef, useState } from 'react';
 import { useSelector } from 'react-redux';
 import PropTypes from 'prop-types';
 
+import BigNumber from 'bignumber.js';
 import {
   GAS_RECOMMENDATIONS,
   EDIT_GAS_MODES,
@@ -109,7 +110,7 @@ export default function EditGasDisplay({
   );
 
   let warningMessage;
-  if (gasLimit < properGasLimit) {
+  if (new BigNumber(gasLimit).lessThan(new BigNumber(properGasLimit))) {
     warningMessage = t('gasLimitRecommended', [properGasLimit]);
   }
 
--- a/ui/components/app/edit-gas-display/edit-gas-display.component.js
+++ b/ui/components/app/edit-gas-display/edit-gas-display.component.js
@@ -57,6 +57,7 @@ export default function EditGasDisplay({
   setGasPrice,
   gasLimit,
   setGasLimit,
+  properGasLimit,
   estimateToUse,
   setEstimateToUse,
   estimatedMinimumFiat,
@@ -107,6 +108,12 @@ export default function EditGasDisplay({
       dappSuggestedAndTxParamGasFeesAreTheSame,
   );
 
+  let showTopWarning, warningMessage;
+  if (gasLimit < properGasLimit) {
+    showTopWarning = true;
+    warningMessage = t('gasLimitIsLessThenCalculatedGasLimit', [properGasLimit]);
+  }
+
   const showTopError =
     (balanceError || estimatesUnavailableWarning) &&
     (!isGasEstimatesLoading || txParamsHaveBeenCustomized);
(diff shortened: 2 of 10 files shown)


ui/components/ui/actionable-message/index.scss
```diff
@@ -14,7 +14,7 @@
   }
 
   &--with-icon.actionable-message--warning {
-    justify-content: normal;
+    justify-content: center;
```
david0xd, line 17: What's the impact of this on the other components where this is used? 
Is it certain that this change will not affect the other pages in a way that is not suitable?

ui/ducks/send/send.js
```diff
@@ -354,7 +354,7 @@ export const computeEstimatedGasLimit = createAsyncThunk(
     const { send, metamask } = state;
     const isNonStandardEthChain = getIsNonStandardEthChain(state);
     const chainId = getCurrentChainId(state);
-    if (send.stage !== SEND_STAGES.EDIT) {
+    if (send.stage) {
```
david0xd, line 357: Can I get short explanation why this change was required?

david0xd (dismissed):

ui/ducks/send/send.js
VSaric: @david0xd This change was required because when the user wants to modify the original send amount to something different, `gasLimit` will not reset to default 21000, because when the user click the `< Edit` button at the top left, `send.stage` will be in status `EDIT` and than, when the user is modifying the original send amount to something different, function `estimateGasLimitForSend` will be called and recompute the `gasLimit` estimate to be as accurate as possible.

ui/components/ui/actionable-message/index.scss
VSaric: @david0xd I researched the impact of this on the other components in app and you're right about this. 
I changed it immediately, sorry, my mistake.
You can see changes here: 68e87f995e68cad0d644b5e3a68139d0432df9d4
Is it okay now?

david0xd (dismissed):

ui/components/app/edit-gas-display/edit-gas-display.component.js
```diff
@@ -107,6 +108,12 @@ export default function EditGasDisplay({
       dappSuggestedAndTxParamGasFeesAreTheSame,
   );
 
+  let showTopWarning, warningMessage;
+  if (gasLimit < properGasLimit) {
+    showTopWarning = true;
```
darkwing, line 113: I don't think we need a `showTopWarning` variable -- on line 156, we can simply check if `warningMessage` is defined and display based on that.

VSaric: @darkwing Yes, you're right, we don't need a `showTopWarning` variable.
You can see changes here: b32a0fca871461945f501d2e3567d9d2cbb85c5e
Is it okay now?

ui/components/app/edit-gas-display/edit-gas-display.component.js
```diff
@@ -107,6 +108,11 @@ export default function EditGasDisplay({
       dappSuggestedAndTxParamGasFeesAreTheSame,
   );
 
+  let warningMessage;
+  if (gasLimit < properGasLimit) {
```
danjm, line 112: This comparison should be done using `BigNumber`

ui/ducks/send/send.js
danjm: I think there are at least two cases when `send.stage === SEND_STAGES.EDIT` that we don't want new estimation to happen. First, is if the gas limit was explicitly proposed by a dapp. The second is if the user has already explicitly set the gas limit.

For the first case, we can check the transaction data for a `dappSuggestedGasFees.gas` property.

For the second, I think we might have to modify the `onSubmit` function in `ui/components/app/edit-gas-popover/edit-gas-popover.component.js` so that if the gas limit was changed by the user, then we record a "userEditedGasLimit" on the txMeta, so that it can be accessed when editing the transaction.

Both of these will only work if the transaction metadata is available when editing the transaction.

app/_locales/en/messages.json
```diff
@@ -1088,6 +1088,9 @@
   "gasLimitInfoTooltipContent": {
     "message": "Gas limit is the maximum amount of units of gas you are willing to spend."
   },
+  "gasLimitRecommended": {
+    "message": "Recommended gas limit is $1. If the gas limit is less than estimated transaction can fail."
```
danjm, line 1092: I would modify this to " "Recommended gas limit is $1. If the gas limit is less that, it may fail.""

ui/hooks/gasFeeInput/useGasFeeInputs.js
```diff
@@ -136,6 +136,10 @@ export function useGasFeeInputs(
     Number(hexToDecimal(transaction?.txParams?.gas ?? '0x0')),
   );
 
+  const properGasLimit = Number(
```
danjm, line 139: What about this case:

1. Gas limit is estimated to 100,000
2. User opens the gas popover, edits the gas limit to 200,000 and saves the changes
3. `transaction?.txParams?.gas` is now 200,000
4. User opens the gas popover again, and edits the gas limit to 150,000
5. `properGasLimit` will be 200,000 and `gasLimit` will be 150,000 and the user will see a warning, even though 150,000 is higher than the original estimate of 100,000

I think we need to save the original estimate on the transaction metadata.

To do that, the following code in `app/scripts/controllers/transactions/index.js` should be edited

```
    if (defaultGasLimit && !txMeta.txParams.gas) {
      txMeta.txParams.gas = defaultGasLimit;
    }
```

It could be changed to 

```
  if (defaultGasLimit && !txMeta.txParams.gas) {
      txMeta.txParams.gas = defaultGasLimit;
      txMeta.originalGasEstimate = defaultGasLimit;
  }
```

ui/ducks/send/send.js
VSaric: @danjm Thank you Dan for commenting this issue. My understanding of these cases are:

For the first case, I checked the transaction data for a `dappSuggestedGasFees.gas` property in the code. I run this case several times and I think it doesn't estimate new gas limit and gas limit is the same as `dapp` is explicitly proposed.

For the second, let's say gas limit for a transaction is 21000, the user explicitly sets gas limit to 30,000 and click `Save`, should that value of 30,000 remain in the gas limit field if he click `Edit` again where the gas limit field is located? If this is okay and if I understand correctly this second case, it works as in the above exposition and we don't need to modify the `onSubmit` function in `ui/components/app/edit-gas-popover/edit-gas-popover.component.js`. If there is another solution for this second case, please let me know.

app/_locales/en/messages.json
```diff
@@ -1126,7 +1126,7 @@
     "message": "Gas limit is the maximum amount of units of gas you are willing to spend."
   },
   "gasLimitRecommended": {
-    "message": "Recommended gas limit is $1. If the gas limit is less than estimated transaction can fail."
+    "message": "Recommended gas limit is $1. If the gas limit is less that, it may fail."
```
VSaric, line 1129: @danjm You can see changes here realted to: https://github.com/MetaMask/metamask-extension/pull/12784#discussion_r761033281

ui/hooks/gasFeeInput/useGasFeeInputs.js
VSaric: @danjm I added `originalGasEstimate` property into `txMeta` and now it is work correctly.
You can see changes here: 8e8a6ecf4a162029c09ca4616d393bde27dc713a
Is it okay now?

app/_locales/en/messages.json
VSaric: @danjm I modified the message as you suggested.
You can see changes here: 8e8a6ecf4a162029c09ca4616d393bde27dc713a
Is it okay now?

ui/components/app/edit-gas-display/edit-gas-display.component.js
VSaric: @danjm I modified this comparison by using `BigNumber` as you suggested.
You can see changes here: 8e8a6ecf4a162029c09ca4616d393bde27dc713a
Is it okay now?

ui/ducks/send/send.js
danjm: @VSaric The issue I describe can be seen in the two videos below (which I recorded on your branch):


https://user-images.githubusercontent.com/7499938/145382988-a949ef4d-5ef1-494b-b3b0-a2763ef7e1a7.mp4


https://user-images.githubusercontent.com/7499938/145382991-126b7eef-47a6-40ee-a45e-275cd799ae99.mp4

danjm: To recreate:

1) Create a token from the test dapp
2) Create a transaction to transfer the token
3) Edit the gas limit
4) Click 'Edit' in the top left to go to the send/edit screen
5) change the transfer amount to `0` or from `0`
6) Go back to the confirm screen, open the confirm popup, and see that the gas limit has changed from what it was edited to before

danjm: The reason we want to prevent changing the gas limits in these cases is this is that currently, on the send/edit screen, we don't show the user any information about the gas limit. So if the user manually edited the gas limit, we don't want to change it without making it clear to them that we are doing so. We also don't want to change a dapp suggested gas limit, in case a dapp has intentionally set the gas limit very high because they know that estimations will be inaccurate.

danjm: So I was thinking that the `if (send.stage) {` condition needs to change to something like `if (send.stage !== SEND_STAGES.EDIT || !dappSuggestedGasFees?.gas || !userEditedGasLimit) {`

VSaric: Thank you @danjm for added comments here to further explanation of this cases. 
I made changes related to this explanation.
You can see changes here: e2afb797e92694ced2d0a1565681202efd7acd52
Please take a look and review and let me know what do you think about this.

mirjanaKukic:
Verified by QA.

darkwing (changes requested):

app/_locales/en/messages.json
```diff
@@ -1163,6 +1163,9 @@
   "gasLimitInfoTooltipContent": {
     "message": "Gas limit is the maximum amount of units of gas you are willing to spend."
   },
+  "gasLimitRecommended": {
+    "message": "Recommended gas limit is $1. If the gas limit is less that, it may fail."
```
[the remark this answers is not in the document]
VSaric: @darkwing 
Fixed: b2bae5bbe0d93f2c4a8394802a9df3032eaa31bb

ui/hooks/gasFeeInput/useGasFeeInputs.js
```diff
@@ -133,6 +133,14 @@ export function useGasFeeInputs(
...
```
danjm, line 136: Why was the `?? '0x0'` needed? Is it possible to have a transaction without an `originalGasEstimate`?

VSaric: Yes @danjm, you're right, `?? '0x0` is not needed, it is possible to have a transaction without an `originalGasEstimate`. 
I will changed it immediately.
Thanks

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
```diff
@@ -163,6 +165,7 @@ export default function EditGasPopover({
 
     const updatedTxMeta = {
       ...updatedTransaction,
+      userEditedGasLimit,
```
danjm, line 168: Instead of setting this using `setUserEditedGasLimit` and passing it to this component, we should just be able to calculate it here with a comparison to the original gas limit. Something like:
`userEditedGasLimit: gasLimit !== transaction.originalGasEstimate`

danjm (changes requested):
Looks good. I left one question and one suggestion. Once those are resolved this should be good to merge.

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
VSaric: @danjm What do you think, can we use `properGasLimit` instead of `transaction.originalGasEstimate` because `gasLimit` is a number and `transaction.originalGasEstimate` is a hex value and we have that `properGasLimit` set like this: `const properGasLimit = Number(hexToDecimal(transaction?.originalGasEstimate));` in `ui/hooks/gasFeeInput/useGasFeeInputs.js`? We can also use `Number(transaction.originalGasEstimate)` to change the hex value to a number.
I will wait for your response about this and than I will make the necessary changes.
Thank you

VSaric: @danjm I resolved this issue.
You can see changes here: 365a71e56615512f1a28d3b7430538fa94f58178
Is it okay now?

ui/hooks/gasFeeInput/useGasFeeInputs.js
VSaric: @danjm I resolved this issue.
You can see changes here: 365a71e
Is it okay now?

danjm (approved):
LGTM

NiranjanaBinoy (approved):

darkwing (approved):

jpuri:
The PR updated old transaction UI, a similar change will be needed for 1559 V2 implementation also.

PR #12260, opened 2021-09-30
Implement Network Switcher designs

segun:
Fixes: #11473 

Explanation:  

Improve the current network switcher.

[Stories and notes](https://www.notion.so/7af6f9afb8634469ac881acb0c9fd9b4?v=92f5f9ccb9804651a29cea4e313428b5&p=792db403977f49afad30a21e8f47402b)

Designs are ready [here](https://www.figma.com/file/F8NVM9pM8RpIe3YNbHsk5r/Switch-Network?node-id=54%3A364)

Manual testing steps:  
  - 
  - 
  -

PR #12786, opened 2021-11-22
Ensure one and only one localhost option appears in network list

### LINKED ISSUE #12754: Two localhost:8545 in MetaMask 10.6.0, second does not work
There are now two `Localhost 8545` with MetaMask 10.6.0.  The first is selectable, the last is not.  

image

danjm:
Fixes: #12754

That issue applies to people who had MetaMask installed prior to v10.6.0

It fixes another issue with the localhost option in the network dropdown on v10.6.0: upon freshly installing MetaMask, and then switching on the display of test nets, the localhost option does not work

https://github.com/MetaMask/metamask-extension/pull/12260 removed localhost from the `first-time-state` `frequentRpcListDetail`, and also used `renderNetworkEntry` in `network-dropdown` to render it in the network list. `renderNetworkEntry` is meant for use with test networks, but we add localhost as a custom rpc. This change was likely made to ensure that localhost only appears once in the network dropdown and that it is at the bottom of the network list (in accordance with designs).

This PR corrects that by re-adding localhost to first-time-state, and using the correct render method for localhost in `network-dropdown`

With this PR, localhost is behaving correctly in the the network dropdown list.

### DIFF
--- a/ui/components/app/dropdowns/network-dropdown.js
+++ b/ui/components/app/dropdowns/network-dropdown.js
@@ -152,13 +152,21 @@ class NetworkDropdown extends Component {
     );
   }
 
-  renderCustomRpcList(rpcListDetail, provider) {
+  renderCustomRpcList(rpcListDetail, provider, opts = {}) {
     const reversedRpcListDetail = rpcListDetail.slice().reverse();
 
     return reversedRpcListDetail.map((entry) => {
       const { rpcUrl, chainId, ticker = 'ETH', nickname = '' } = entry;
       const isCurrentRpcTarget =
         provider.type === NETWORK_TYPE_RPC && rpcUrl === provider.rpcUrl;
+
+      let borderColor = COLORS.UI2;
+      if (isCurrentRpcTarget) {
+        borderColor = COLORS.WHITE;
+      }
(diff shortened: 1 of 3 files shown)


segun (dismissed):
LGTM!

ui/components/app/dropdowns/network-dropdown.js
```diff
@@ -149,14 +152,22 @@ class NetworkDropdown extends Component {
...
       const { rpcUrl, chainId, ticker = 'ETH', nickname = '' } = entry;
       const isCurrentRpcTarget =
         provider.type === NETWORK_TYPE_RPC && rpcUrl === provider.rpcUrl;
 
+      let borderColor = COLORS.UI2;
+      if (isCurrentRpcTarget) {
+        borderColor = COLORS.WHITE;
+      }
+      if (opts.isLocalHost) {
+        borderColor = 'localhost';
```
adonesky1, line 168: [nit] non-blocking question. Should we use a different color for `$localhost`? Its the same as mainnet.

adonesky1 (dismissed):
LGTM!

segun (approved):

adonesky1 (approved):

PR #12800, opened 2021-11-23
 mainnet tokens autopopulate add token form fields when on custom networks #12087

### LINKED ISSUE #12087: mainnet tokens autopopulate add token form fields when on custom networks
1. Be on a custom network
2. Copy the mainnet address of a popular erc20 token contract
3. Go to the add token flow, paste the erc20 contract address
4. The decimal and symbol fields will populate with mainnet data and the user will be able to add the token

Instead, we should show the user the same warnings we show when they paste a non-contract address.

This might be resolved with our upcoming token detection improvements.

dragana8:
Fixes: https://github.com/MetaMask/metamask-extension/issues/12087

Explanation:  
When trying to import mainnet token user gets error message 'This is a mainnet token' and symbol and decimals fields are not autopopulated but user is able to submit.
Screenshot from 2021-11-23 11-57-18

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -1595,7 +1595,7 @@
     "message": "Ethereum Mainnet"
   },
   "mainnetToken": {
-    "message": "This is a mainnet token address. We do not expect it to work on the current network..."
+    "message": "This address matches a known mainnet token address. Recheck the contract address and network for the token you are trying to add."
   },
   "makeAnotherSwap": {
     "message": "Create a new swap"
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -1594,6 +1594,9 @@
   "mainnet": {
     "message": "Ethereum Mainnet"
   },
+  "mainnetToken": {
+    "message": "This is a mainnet token"
+  },
   "makeAnotherSwap": {
     "message": "Create a new swap"
   },
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -1595,7 +1595,7 @@
     "message": "Ethereum Mainnet"
   },
   "mainnetToken": {
-    "message": "This is a mainnet token"
+    "message": "This is a mainnet token address. We do not expect it to work on the current network..."
   },
   "makeAnotherSwap": {
     "message": "Create a new swap"
--- a/ui/pages/import-token/import-token.component.js
+++ b/ui/pages/import-token/import-token.component.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { getTokenTrackerLink } from '@metamask/etherscan-link';
+import contractMap from '@metamask/contract-metadata';
 import {
   checkExistingAddresses,
   getURLHostName,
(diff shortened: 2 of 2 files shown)


mhasanagic:
Verified by QA

ui/pages/import-token/import-token.component.js
```diff
@@ -191,13 +199,24 @@ class ImportToken extends Component {
...
     });
     const standardAddress = addHexPrefix(customAddress).toLowerCase();
 
+    let isMainnetToken = false;
+
+    Object.keys(contractMap).forEach((key) => {
+      if (key.toLowerCase() === customAddress.toLowerCase()) {
+        isMainnetToken = true;
+      }
+    });
```
danjm, line 216: Nit: this code can be simplified with the `.some` method

```suggestion
    const isMainnetToken = Object.keys(contractMap).some((key) => key.toLowerCase() === customAddress.toLowerCase());
```

ui/pages/import-token/import-token.component.js
```diff
@@ -133,6 +135,12 @@ class ImportToken extends Component {
     );
   }
 
+  hasWarning() {
```
danjm, line 138: It looks like this method is not used... can it be deleted?

danjm (dismissed):
This looks like a good solution. I have left a couple of small comments

ui/pages/import-token/import-token.component.js
dragana8: deleted

ui/pages/import-token/import-token.component.js
dragana8: updated

darkwing:
The code looks good but I feel like from a design perspective, we should probably do more to inform the user about why this may not work.  Let me see what design thinks.

app/_locales/en/messages.json
```diff
@@ -1576,6 +1576,9 @@
   "mainnet": {
     "message": "Ethereum Mainnet"
   },
+  "mainnetToken": {
+    "message": "This is a mainnet token"
```
danjm, line 1580: Perhaps to resolve @darkwing's point about design, we could just modify this text a bit. Change it to "This is a mainnet token address. Adding it to the current network is likely to fail."

danjm (dismissed):
Approving, but we could make the one copy change I suggested, and I think that would address @darkwing's design concern as well.

app/_locales/en/messages.json
dragana8: Done.

danjm (dismissed):
LGTM

app/_locales/en/messages.json
```diff
@@ -1594,6 +1594,9 @@
   "mainnet": {
     "message": "Ethereum Mainnet"
   },
+  "mainnetToken": {
+    "message": "This is a mainnet token address. We do not expect it to work on the current network..."
```
adonesky1, line 1598:
```suggestion
    "message": "This address matches a known mainnet token address. Recheck that you are on the right network and the contract address for the token you are trying to add..."
```
cc @rachelcope @danjm I think this copy is more clear. What do you think?

danjm: looks good to me

adonesky1 (approved):

brad-decker (approved):

PR #12802, opened 2021-11-23
Fix: "BigNumber Error: times() number type has more than 15 significant digits"

dan437:
## Description
It resolves an issue when numbers used in the `.times()` BigNumber function had more than 15 significant digits.

## Testing
- Switch to Polygon
- Try to swap from MATIC to AAVE
- You will see quotes instead of the occasional "Error fetching quotes" page

## Screenshots
image

### DIFF
--- a/app/scripts/controllers/swaps.js
+++ b/app/scripts/controllers/swaps.js
@@ -746,7 +746,7 @@ export default class SwapsController {
       const conversionRateForSorting = tokenConversionRate || 1;
 
       const ethValueOfTokens = decimalAdjustedDestinationAmount.times(
-        conversionRateForSorting,
+        conversionRateForSorting.toString(10),
         10,
       );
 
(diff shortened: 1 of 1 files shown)


app/scripts/controllers/swaps.js
```diff
@@ -768,7 +768,7 @@ export default class SwapsController {
         quote.ethValueOfTokens = ethValueOfTokens.toString(10);
         quote.overallValueOfQuote = overallValueOfQuoteForSorting.toString(10);
         quote.metaMaskFeeInEth = metaMaskFeeInTokens
-          .times(conversionRateForCalculations)
+          .times(conversionRateForCalculations.toString(10))
           .toString(10);
```
adonesky1, line 772: [Q] - is this second `.toString(10)` call required still?

adonesky1 (approved):
LGTM

danjm (approved):

PR #12813, opened 2021-11-23
Refactor currency input component and ensure currency toggle persists between settings

### LINKED ISSUE #12601: Refactor currency input component and ensure currency toggle persists between settings
Replaces part of https://github.com/MetaMask/metamask-extension/issues/12347

### Here is an explanation of the currency settings part of that bug:

1. Turn on the "Show Conversion on Testnets" toggle in advanced settings
2. Go to the send flow
3. Enter an amount and then click the currency switcher
4. Click "Next"
5. On the Confirm screen click "Edit"
6. You will now be on the send screen, but the currency switcher setting will be back to the default

Instead, the currency preference should be what you switched it to originally

Here is a video demonstrating this (ignore parts of the video where hex data is edited):

https://user-images.githubusercontent.com/7499938/137171004-0ee806b1-7a0b-4f99-89c3-e7ebc4351430.mp4

### Notes on fixing this issue

Fixing this is annoyingly complicated because `currency-input.component.js` is convoluted, poorly designed and over-reliant on component state.

We should probably just write that component from scratch, and also use redux to store whether the user has switched the currency display or not.

So here are guidelines for creating a new `currency-input.js` component:

- name  it `currency-input.js`
- it should be a functional component
- the new component should have the exact same behaviour as the current one
- the new component should look exactly the same as the current one, you can copy css as necessary

- the component should receive the following properties:
`hexValue`, type: string, the value to display
`preferredCurrency`, type: string, the symbol of the users currently preferred currency (as set in general settings)
`secondaryCurrency`, type: string, the symbol of the currency that is not primary
`hideSecondary`, type: boolean, whether the conversion from the primary to secondary currency should be hidden and "No conversion rates available" shown instead
`featureSecondary`, type: boolean, whether the top, editable line in the component should be the secondary currency
`conversionRate`, type: number, the rate that can be used to convert the primary to secondary currency
`primarySuffix`, type: string, the suffix that can be shown in line with the primary currency
`secondarySuffix`, type: string, the suffix that can be shown in line with the secondary currency
`onChange`, type: function, a handler for the event of the input changing
`onPreferenceToggle`, type: function, a handler for when the switch/toggle icon is clicked

If the `primaryCurrency` is ETH and `featureSecondary` is false, it should look like:
Screenshot from 2021-11-05 10-15-35
 
If the `primaryCurrency` is ETH and `featureSecondary` is true, it should look like:
Screenshot from 2021-11-05 10-15-40

If `hideSecondary` is true, it should look like:
Screenshot from 2021-11-05 10-16-05

- the new component should continue to use the `UnitInput` and `CurrencyDisplay` components as the current component does

- outside this component, a `sendInputCurrencySwitched` property should be stored in redux. This can be done in the `appState` duck
- `sendInputCurrencySwitched` should be toggled when `onPreferenceToggle` is called
- the parent of the new CurrencyInput component should set `featureSecondary` to true if `(useNativeCurrencyAsPrimaryCurrency && sendInputCurrencySwitched) || (!useNativeCurrencyAsPrimaryCurrency && !sendInputCurrencySwitched)` (`useNativeCurrencyAsPrimaryCurrency` is available in state)
- if `featureSecondary` is true but `hideSecondary` is also true, `hideSecondary` should take precedence, and the component should act as if `featureSecondary` is false

filipsekulic:
Fixes: https://github.com/MetaMask/metamask-extension/issues/12601

Explanation: Refactor currency input component and ensure currency toggle persists between settings

Screencast:

https://user-images.githubusercontent.com/92310504/143141772-77fbcb56-e22d-4089-bd17-0e4c247e3b85.mp4


paffur (dismissed):

paffur (dismissed):

david0xd (dismissed):

ui/components/ui/currency-input/currency-input.js
```diff
@@ -0,0 +1,166 @@
...
+    }
+  }, [isSwapped]);
+
+  const renderConversionComponent = () => {
+    let currency, numberOfDecimals;
+
+    if (hideSecondary) {
+      return (
+        <div className="currency-input__conversion-component">
+          {t('noConversionRateAvailable')}
```
adonesky1, line 106: related comment on the issue ticket here: https://github.com/MetaMask/metamask-extension/issues/12601#issuecomment-985680398

ui/components/ui/currency-input/currency-input.container.js
```diff
@@ -1,7 +1,7 @@
 import { connect } from 'react-redux';
 import { ETH } from '../../../helpers/constants/common';
 import { getShouldShowFiat } from '../../../selectors';
-import CurrencyInput from './currency-input.component';
+import CurrencyInput from './currency-input';
```
adonesky1, line 4: Is this container used anymore? If not, the file should be deleted.

adonesky1: Also the test.

adonesky1: We should use selectors to get redux state in `currency-input.js` rather than keep this container and pass through connect.

ui/components/ui/currency-input/currency-input.js
danjm: @adonesky1 I think that comment is correct, but it should probably be addressed in a separate PR. This PR is just maintaining the current behaviour (currently implemented in https://github.com/MetaMask/metamask-extension/pull/12813/files#diff-6afdf120de27d111574e2c0e237c729a4218101d99419d68773c1c633731323dL126-L132) as it refactors the component

ui/components/ui/currency-input/currency-input.container.js
danjm: Yeah, I think this is a good comment @filipsekulic 

The refactor you've made basically takes the update of the component almost all the way to our latest patterns. Using `useSelector` within the component and deleting the `.container` file would complete the refactor

filipsekulic: Fixed

danjm (dismissed):
This looks good to me.

danjm (approved):

david0xd (approved):

brad-decker (approved):

ui/components/ui/currency-input/currency-input.js
```diff
@@ -0,0 +1,176 @@
+import React, { useContext, useEffect, useState } from 'react';
```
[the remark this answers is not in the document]
filipsekulic: @brad-decker I did it in: https://github.com/MetaMask/metamask-extension/pull/13380. Also, I waited for the pipeline to be fixed, as there were some tests failing, before I rebased this PR.

PR #12814, opened 2021-11-23
Remove legacy node parent detection

darkwing:
We no longer need the old method of walking parent nodes to detection descendants.


Gudahtt (approved):
LGTM!

jpuri (approved):
👍

PR #12832, opened 2021-11-24
Added SiteOrigin and PermissionList components

ritave:
Changes made to the flow:

Screenshot 2021-11-25 at 18 36 31
Screenshot 2021-11-25 at 18 36 38

### DIFF
--- a/ui/components/app/permissions-connect-header/index.scss
+++ b/ui/components/app/permissions-connect-header/index.scss
@@ -9,20 +9,6 @@
     display: flex;
     flex-direction: column;
     align-items: center;
-
-    .icon-with-fallback__identicon-container,
-    .icon-with-fallback__identicon-border {
-      height: 64px;
-      width: 64px;
-    }
-
-    .icon-with-fallback__identicon-border {
-      border: 1px solid $Grey-100;
-    }
-
-    .icon-with-fallback__identicon-container {
-      margin-bottom: 8px;
-    }
   }
 
   &__title {
@@ -33,22 +19,11 @@
     margin-top: 16px;
   }
 
-  &__text,
   &__subtitle {
     @include H6;
 
     text-align: center;
     color: $Grey-500;
-  }
-
-  &__text {
-    width: 100%;
-    text-overflow: ellipsis;
-    overflow: hidden;
-    margin-top: 8px;
-
-    /*rtl:ignore*/
-    direction: rtl;
   }
 
   &__subtitle {
--- a/ui/components/app/permissions-connect-permission-list/index.scss
+++ b/ui/components/app/permissions-connect-permission-list/index.scss
@@ -1,0 +1,7 @@
+.permissions-connect-permission-list {
+  .permission {
+    width: 100%;
+    padding: 14px;
+    border: 0 0 1px 0 $Grey-100;
+  }
+}
--- a/ui/components/app/permissions-connect-permission-list/index.scss
+++ b/ui/components/app/permissions-connect-permission-list/index.scss
@@ -1,7 +1,19 @@
 .permissions-connect-permission-list {
   .permission {
     width: 100%;
-    padding: 14px;
-    border: 0 0 1px 0 $Grey-100;
+    padding-bottom: 16px;
+    border-bottom: 1px solid $Grey-100;
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+    color: $black;
+
+    i {
+      display: block;
+      padding: 16px;
+      min-width: 16px;
+      min-height: 16px;
+      color: $Grey-500;
+    }
   }
 }
--- a/ui/components/app/permissions-connect-permission-list/index.scss
+++ b/ui/components/app/permissions-connect-permission-list/index.scss
@@ -1,12 +1,14 @@
 .permissions-connect-permission-list {
   .permission {
+    @include H6;
+
     width: 100%;
     padding-bottom: 16px;
     border-bottom: 1px solid $Grey-100;
     display: flex;
     flex-direction: row;
     align-items: center;
-    color: $black;
+    color: $Black-100;
 
     i {
       display: block;
@@ -14,6 +16,7 @@
       min-width: 16px;
       min-height: 16px;
       color: $Grey-500;
+      font-size: 1rem;
     }
   }
 }
--- a/ui/components/app/permissions-connect-permission-list/permissions-connect-permission-list.js
+++ b/ui/components/app/permissions-connect-permission-list/permissions-connect-permission-list.js
@@ -17,15 +17,13 @@ export default function PermissionsConnectPermissionList({ permissions }) {
 
   return (
     <div className="permissions-connect-permission-list">
-      {Object.keys(permissions)
-        .map((permission) => PERMISSION_TYPES[permission])
-        .map((permission) => (
-          <div className="permission" key={permission.label}>
-            <i className={permission.leftIcon} />
-            {permission.label}
-            <i className={permission.rightIcon} />
-          </div>
-        ))}
+      {Object.keys(permissions).map((permission) => (
+        <div className="permission" key={PERMISSION_TYPES[permission].label}>
+          <i className={PERMISSION_TYPES[permission].leftIcon} />
+          {PERMISSION_TYPES[permission].label}
+          <i className={PERMISSION_TYPES[permission].rightIcon} />
+        </div>
+      ))}
     </div>
   );
 }
--- a/ui/components/app/permissions-connect-permission-list/permissions-connect-permission-list.js
+++ b/ui/components/app/permissions-connect-permission-list/permissions-connect-permission-list.js
@@ -1,5 +1,6 @@
-import React, { useContext } from 'react';
-import { I18nContext } from '../../../contexts/i18n';
+import React from 'react';
+import PropTypes from 'prop-types';
+import { useI18nContext } from '../../../hooks/useI18nContext';
 
 const PERMISSION_TYPES = {
   eth_accounts: {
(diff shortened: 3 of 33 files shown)


ui/components/app/permissions-connect-permission-list/index.scss
```diff
@@ -0,0 +1,22 @@
...
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+    color: $Black-100;
+
+    i {
+      display: block;
+      padding: 16px;
+      min-width: 16px;
+      min-height: 16px;
```
georgewrmarshall, line 17: Is this so if the font-awesome class is not provided the layout remains the same?

ui/components/app/permissions-connect-permission-list/index.scss
```diff
@@ -0,0 +1,22 @@
+.permissions-connect-permission-list {
+  .permission {
+    @include H6;
+
+    width: 100%;
```
georgewrmarshall, line 5: Is this needed? Should be 100% by default?

ui/components/app/permissions-connect-permission-list/index.scss
```diff
@@ -0,0 +1,22 @@
...
+    padding-bottom: 16px;
```
georgewrmarshall, line 6: Not a big deal but if this was a reusable component I would suggest any outside spacing to be added at the page level or where the layout requires it using a className.

ui/components/app/permissions-connect-permission-list/permissions-connect-permission-list.js
```diff
@@ -0,0 +1,32 @@
...
+  const t = useI18nContext();
+  return (
+    <div className="permissions-connect-permission-list">
+      {Object.keys(permissions)
+        .map((permission) => PERMISSION_TYPES[permission])
+        .map((permission) => (
+          <div className="permission" key={permission.label}>
+            <i className={permission.leftIcon} />
+            {t(permission.label)}
+            <i className={permission.rightIcon} />
```
georgewrmarshall, line 23: Should this be conditional if it's not required? Will align text in layout better?
```suggestion
            {Boolean(permission.rightIcon) && <i className={permission.rightIcon} />}
```

ui/components/app/permissions-connect-permission-list/permissions-connect-permission-list.js
```diff
@@ -0,0 +1,32 @@
...
+          </div>
+        ))}
+    </div>
+  );
+}
+
+PermissionsConnectPermissionList.propTypes = {
+  permissions: PropTypes.object.isRequired,
```
georgewrmarshall, line 31: Should this be a shape proptype?
e.g
```
PropTypes.shape({...}).isRequired
```

ui/components/ui/site-origin/index.scss
```diff
@@ -0,0 +1,5 @@
+.site-origin {
+  .chip__left-icon {
+    padding: 12px 0 12px 12px;
```
georgewrmarshall, line 3: Suggest using 8px grid
```suggestion
    padding: 8px 0 8px 8px;
```

georgewrmarshall (commented):
Looks good!! Some questions and small suggestions. UI components could use some stories.

`ui/components/app/permissions-connect-permission-list/permissions-connect-list.stories.js`

```
import React from 'react';

import PermissionsConnectList from '.';

export default {
  title: 'Components/App/PermissionsConnectList',
  id: __filename,
  component: PermissionsConnectList,
  argTypes: {
    permissions: {
      control: 'object',
    },
  },
};

export const DefaultStory = (args) => <PermissionsConnectList {...args} />;

DefaultStory.storyName = 'Default';

DefaultStory.args = {
  permissions: {
    eth_accounts: {
      leftIcon: 'fas fa-eye',
      label: 'eth_accounts',
      rightIcon: null,
    },
  },
};
```

`ui/components/ui/site-origin/site-origin.stories.js`

```
import React from 'react';

import SiteOrigin from '.';

export default {
  title: 'Components/UI/SiteOrigin',
  id: __filename,
  component: SiteOrigin,
  argTypes: {
    siteOrigin: {
      control: 'text',
    },
    iconSrc: {
      control: 'text',
    },
    iconName: {
      control: 'text',
    },
  },
};

export const DefaultStory = (args) => <SiteOrigin {...args} />;

DefaultStory.storyName = 'Default';

DefaultStory.args = {
  siteOrigin: 'https://metamask.io/',
  iconName: 'MetaMask',
  iconSrc: './metamark.svg',
};
```

ui/components/ui/site-origin/index.scss
Gudahtt: This would be a bit closer to the designs:

```suggestion
    padding: 4px 0 4px 8px;
```

Currently it looks far too tall.

ui/components/app/permissions-connect-permission-list/permissions-connect-permission-list.js
```diff
@@ -0,0 +1,32 @@
...
+export default function PermissionsConnectPermissionList({ permissions }) {
+  const t = useI18nContext();
+  return (
+    <div className="permissions-connect-permission-list">
+      {Object.keys(permissions)
+        .map((permission) => PERMISSION_TYPES[permission])
+        .map((permission) => (
+          <div className="permission" key={permission.label}>
+            <i className={permission.leftIcon} />
+            {t(permission.label)}
```
Gudahtt, line 22: Using variables as the parameter to `t` does not work with the `verify-locales` strict mode, meaning that it's more difficult to detect whether the localized message is used or not. Strict mode requires that a string literal is passed in as the parameter, directly.

You could instead move `PERMISSION_TYPES` into the component, letting you use `t` directly for the `label` property:

e.g.

```
const PERMISSION_TYPES = {
  eth_accounts: {
    leftIcon: 'fas fa-eye',
    label: t('eth_accounts'),
    rightIcon: null,
  },
};
```

ui/components/app/permissions-connect-header/index.scss
```diff
@@ -33,24 +19,13 @@
...
   }
 
-  &__text {
-    width: 100%;
-    text-overflow: ellipsis;
-    overflow: hidden;
-    margin-top: 8px;
-
-    /*rtl:ignore*/
-    direction: rtl;
```
Gudahtt, line 51: This was used to ensure that long domain names were ellipses on the right, rather than the left. This was to prevent phishers from mimicking the domains of legitimate sites.

If we let the ellipsis stay on the right, then the domains `legitimatelongdomainname.com` and `legitimatelongdomainname.malicious.com` might both look the same (e.g. like `legitimatelongdomainname...`).

Putting the ellipses on the _left_ preserves the most important part of the domain, which is the end.

Gudahtt: It looks like the new site origin component doesn't handle long domains very well right now. They just expand the component to exceed the width of the screen. e.g. try with `http://llanfairpwllgwyngyllgogerychwyrndrobwllllantysiliogogogoch.co.uk/` or `https://3.141592653589793238462643383279502884197169399375105820974944592.eu/`

ui/components/app/permissions-connect-permission-list/index.scss
ritave: Exactly

ui/components/app/permissions-connect-permission-list/index.scss
ritave: It's needed. I tried multiple things, haven't found any that works well

ui/components/app/permissions-connect-permission-list/index.scss
ritave: It's internal spacing - the 1px bottom border is part of the item

ui/components/app/permissions-connect-permission-list/permissions-connect-permission-list.js
ritave: It's not conditional to adjust to other permissions that may have icons on the right. Otherwise the text will flow in place where there might be an icon

ui/components/app/permissions-connect-permission-list/permissions-connect-permission-list.js
ritave: It's more akin to a Map of Bools.

ui/components/app/permissions-connect-permission-list/permissions-connect-permission-list.js
```diff
@@ -0,0 +1,35 @@
...
+        rightIcon: null,
+      },
+    };
+  }, [t]);
+
+  return (
+    <div className="permissions-connect-permission-list">
+      {Object.keys(permissions)
+        .map((permission) => PERMISSION_TYPES[permission])
+        .map((permission) => (
```
darkwing, line 22: I'd prefer that we didn't double `map`; it seems reasonable to use `PERMISSION_TYPES[permission].label`, etc.

georgewrmarshall (dismissed):
UI LGTM!

ui/components/app/permissions-connect-permission-list/permissions-connect-permission-list.js
hmalik88: I agree with this ^

hmalik88:
I had two concerns regarding the `SiteOrigin` component:

1. Screen Shot 2021-11-30 at 3 52 13 PM

I think the white space on the right looks a bit jarring, is there any way we can dial this in? I know you need the `max-width` property to ensure the ellipsis starts, but I was wondering if it's possible in the CSS to enact this rule only when the text gets to a certain length?

2. I know @Gudahtt was mentioning the need for the ellipsis to be rtl because of the ending of the url probably being more relevant, but what about scenarios for long URLs where beginning and end are the same but the middle is different? In that case I was wondering if it would be beneficial to add a tooltip that shows the full URL?

ritave:
@hmalik88 Fixed the max-width

ui/components/ui/chip/chip.js
```diff
@@ -35,6 +36,7 @@ export default function Chip({
         'chip--with-right-icon': Boolean(rightIcon),
         [`chip--border-color-${borderColor}`]: true,
         [`chip--background-color-${backgroundColor}`]: true,
+        'chip--max-conent': maxContent,
```
hmalik88, line 39: I think you have a typo here, should be: 'chip--max-content'

ui/components/ui/chip/chip.scss
```diff
@@ -63,4 +63,8 @@
       margin-right: 8px;
     }
   }
+
+  &--max-content {
+    width: max-content;
+  }
```
georgewrmarshall, line 69: Looks like you still need to remove in the base class?

Screen Shot 2021-12-01 at 11 47 21 AM

georgewrmarshall (commented):
Looking good. Left one comment

ui/pages/permissions-connect/permissions-connect.component.js
```diff
@@ -174,7 +174,6 @@ export default class PermissionConnect extends Component {
             className="permissions-connect__back"
             onClick={() => this.goBack()}
           >
-            <i className="fas fa-chevron-left" />
```
hmalik88, line 177: I believe @holantonela updated the design to keep this

darkwing (approved):

ui/components/ui/chip/chip.scss
georgewrmarshall: resolved

ui/pages/permissions-connect/permissions-connect.component.js
```diff
@@ -174,7 +174,7 @@ export default class PermissionConnect extends Component {
...
+            <i className="fas fa-chevron-left"></i>
```
georgewrmarshall, line 177: nit: Could use short hand for closed tag here
```suggestion
            <i className="fas fa-chevron-left" />
```

georgewrmarshall (approved):
LGTM! Tiny comment about self-closing `<i/>` tag. Shouldn't stop from merging though

PR #12834, opened 2021-11-24
Feat/add collectible manually

adonesky1:
Explanation:  
- Hooks up CollectiblesController to the NFT (manually) add collectible view.
- Adds success and failure toasts for the manual add action.

Manual testing steps:  
  - Navigate to the NFTs tab and click on "Add NFT"
  - Add an NFT (contract address/token ID) that you actually own (if none you can get some [here](https://faucet.paradigm.xyz/) for testnets)
  - Check that there is a dismissable "Collectible was successfully added!" toast
  - Redo this process with an NFT that you don't own
  - Check for a dismissable failure toast "Collectible was not added because: This collectible is not owned by the user"
  
  TODO:
- [ ] add unit & e2e tests?


jpuri (dismissed):

ui/store/actions.js
```diff
@@ -1317,6 +1317,43 @@ export function addCollectible(address, tokenID, dontShowLoadingIndicator) {
...
+      dispatch(showLoadingIndication());
+    }
+    try {
+      await promisifiedBackground.addCollectibleVerifyOwnership(
+        address,
+        tokenID,
+      );
+    } catch (error) {
+      if (
+        error.message.includes('This collectible is not owned by the user') ||
```
danjm, line 1342: These look like errors that we would want to display instead of throw? And any other errors would be the ones we want to throw. If not, can you explain why we are throwing these errors and displaying all others?

app/scripts/metamask-controller.js
```diff
@@ -184,6 +184,7 @@ export default class MetamaskController extends EventEmitter {
     });
 
     this.assetsContractController = new AssetsContractController();
+    this.assetsContractController.configure({ provider: this.provider });
```
danjm, line 187: Only a question so I can understand: why does this need to be set with a `configure` call, instead of just passing it with the constructor?

danjm (changes requested):
Looks good overall. I think one change might be needed. And I left a question.

ui/store/actions.js
adonesky1: So perhaps this isn't the best pattern but let me explain how its working:

These two errors are the expected ones, that I forward along to the UI [here](https://github.com/MetaMask/metamask-extension/blob/f54e3747aaaccba1e84ecf1ae6b5994e17079131/ui/pages/add-collectible/add-collectible.component.js#L28) and ultimately show them in a toast indicating that the collectible was not added.

Other unexpected errors are logged and displayed in the way we display (unknown/unexpected?) errors for every other action.

danjm: okay, I get it, I think this is fine

app/scripts/metamask-controller.js
adonesky1: you're right this can be done in the constructor 👍

jpuri (dismissed):

danjm (dismissed):

jpuri (approved):

danjm (approved):

PR #12839, opened 2021-11-24
Add and hook up setting to toggle on/off collectible autodetection

adonesky1:
Explanation:  Adds and hooks up setting to toggle on/off collectible auto-detection per [designs](https://www.figma.com/file/8Xe22jEPgcElS5tWPw8AGM/NFTs-in-extension-(MVP)?node-id=1%3A3)

Screen Shot 2021-11-24 at 3 17 20 PM

Pending question about these designs:
We had discussed having a separate toggle for whether users want to use OpenSea's API, and then only showing this if they say yes to OpenSea: https://consensys.slack.com/archives/G1L7H42BT/p1636155799120300?thread_ts=1635187467.018000&cid=G1L7H42BT

Not sure if we decided against this?

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -3148,6 +3148,12 @@
   "urlExistsErrorMsg": {
     "message": "This URL is currently used by the $1 network."
   },
+  "useCollectibleDetection": {
+    "message": "Autodetect NFTs"
+  },
+  "useCollectibleDetectionDescription": {
+    "message": "Displaying NFTs media & data may expose your IP address to centralized servers. Third-party APIs (like OpenSea) are used to detect NFTs in your wallet. This exposes your account address with those services. Leave this disabled if you don’t want the app to pull data from those those services."
+  },
   "usePhishingDetection": {
     "message": "Use Phishing Detection"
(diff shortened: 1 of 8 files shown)


jpuri (approved):
LGTM

danjm:
I think I am missing something. What affect does this toggle have? Where is `useCollectibleDetection` used?

adonesky1:
> I think I am missing something. What affect does this toggle have? Where is `useCollectibleDetection` used?

sorry it seems I may have left some important context out of the PR description. The useCollectibleDetection value is [used in the CollectiblesDetectionController to enable/disable detection and polling](https://github.com/MetaMask/controllers/blob/3497fa8cad2a206d914bbac5d8914045467a39f5/src/assets/CollectibleDetectionController.ts#L233). This was introduced in a [PR](https://github.com/MetaMask/controllers/pull/638) included in @metamask/controllers v21.0.0 (which is added to the extension in https://github.com/MetaMask/metamask-extension/pull/12834.

danjm:
okay, I see, that makes sense

danjm (approved):

PR #12840, opened 2021-11-24
Remove unnecessary unlock step

Gudahtt:
As part of the unlock process, currently the seed phrase is retrieved from the background then discarded. This step is pointless, so it has been deleted.

Manual testing steps:  
  - Try unlocking the extension, and ensure it acts the same as it did before.


darkwing (approved):
I can login just fine, and looking at the original PR for this, I don't really understand what it was ever added.

danjm (approved):

PR #12844, opened 2021-11-25
Fix LavaMoat background policy generation

Gudahtt:
The LavaMoat policy generation script would sporadically fail because it ran the build concurrently three times, and the build includes steps that delete the `dist` directory and write to it. So if one build process tried to write to the directory after another deleted it, it would fail.

This was solved by adding a new `--policy-only` flag to the build script, and a new `scripts:prod` task. The `scripts:prod` task only runs the script tasks for prod, rather than the entire build process. The `--policy-only` flag stops the script tasks once the policy has been written, and stops any other files from being written to disk.

This prevents the three concurrent build processes from getting in each others way, and it dramatically speeds up the process.

Manual testing steps:  
  - Run `yarn lavamoat:background:auto` a few times, and see that it does not fail.
  - Ensure `lavamoat:background:auto` can still update the LavaMoat policy when adding a new dependency to the background.

### DIFF
--- a/development/build/README.md
+++ b/development/build/README.md
@@ -52,6 +52,9 @@ Options:
                       bundle. Setting this to `false` is useful e.g. when
                       linking dependencies that are incompatible with lockdown.
                                                        [boolean] [default: true]
+  --policy-only       Stops the build after generating the LavaMoat policy,
+                      skipping any writes to disk.
+                                                       [boolean] [deafult: false]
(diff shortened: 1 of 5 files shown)


danjm (dismissed):
LGTM

Gudahtt:
Rebased to resolve a conflict

adonesky1 (approved):
LGTM (tested)

danjm (approved):

Issue #6805, opened 2019-07-04
Supported browser versions


Gudahtt:
We should be explicit about which browser versions we support, and take steps to ensure we remain compatible.

Currently we state on our website that we are compatible with Chrome, Firefox, and Brave, but we don't specify any minimum required browser versions. Opera is listed as well, but that version is very outdated. We don't regularly test which browsers the application works on either, relying instead upon bug reports for compatibility with older browsers.

There was a previous issue for documenting which browsers we support (#3343), but that was closed with an adjustment of our Babel configuration file to support all browser versions with at least 0.25% global usage (according to `browserlist`), excluding IE 11 and Opera Mini. However, this includes some _very old_ browser versions, more than we'd care to support in practice.

Instead we should document explicitly which browsers we support, and add tests to ensure we stay compatible (perhaps to run nightly, so we don't overly burden CI). Then we can get rid of the Babel transformations and polyfills we no longer need. We should also update the manifests to state the minimum browser version, so users don't accidentally install the extension on an unsupported browser.

Steps:

- [ ] Document which browser versions we support
- [ ] Remove unnecessary Babel transformations and polyfills
- [ ] Add nightly tests for minimum supported browser versions
- [ ] Update manifests to include minimum browser version

Gudahtt:
**Choose minimum supported browser versions**
Here are the factors to consider that I've thought of so far
* User impact
  * How many users would be affected?
  * How many users would be unable or unwilling to switch to a supported browser?
* Support / Quality Assurance
  * More browser versions to test means more QA and support
    * We could have a two-tiered system, where "tier 1" platforms (e.g. last 2 versions of each browser) get are included in typical QA efforts, but for "tier 2" platforms (e.g. older versions) we rely upon bug reports from users. It would still be more work than dropping support though.
* Development / Maintenance
  * Transformations and polyfills are required to support features we use on older platforms
    * Complicates build process
    * Slows down build process
    * Increases final bundle size
  * Many of the tools and libraries we use may not work on older platforms. Taking care to ensure things are compatible takes time, and is often not considered by new contributors
  * Certain new JavaScript features come with improvements to the development experience (e.g. async stack traces). Supporting older browsers might mean not benefiting from these improvements
  * Certain new features might be easier/faster to build if we didn't have to worry about supporting older platforms
* Product
  * Certain new JavaScript features come with performance improvements
* Security
[...]

PR #12847, opened 2021-11-25
Update minimum browser versions

Gudahtt:
The minimum browser versions have been updated to Chromium v80 and Firefox v78.

~Chromium v78 is >2 years old, and very few users use older versions than that.~
Updated to v80 to provide support for optional chaining

Firefox v78 is the _previous_ extended support release. The current Firefox extended support release is v91.

Related: #6805

Manual testing steps:  
  - Ensure that the extension still works on versions equal to or above the new minimum versions.

### DIFF
--- a/app/manifest/v2/chrome.json
+++ b/app/manifest/v2/chrome.json
@@ -4,5 +4,5 @@
     "matches": ["https://metamask.io/*"],
     "ids": ["*"]
   },
-  "minimum_chrome_version": "66"
+  "minimum_chrome_version": "80"
 }
--- a/babel.config.js
+++ b/babel.config.js
@@ -5,7 +5,7 @@ module.exports = function (api) {
       strictMode: true,
     },
     targets: {
-      browsers: ['chrome >= 66', 'firefox >= 68'],
+      browsers: ['chrome >= 80', 'firefox >= 78'],
     },
     presets: [
       '@babel/preset-typescript',
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,1 +1,1 @@
-chrome >= 66, firefox >= 68
+chrome >= 80, firefox >= 78
--- a/app/manifest/v2/firefox.json
+++ b/app/manifest/v2/firefox.json
@@ -2,7 +2,7 @@
   "applications": {
     "gecko": {
       "id": "webextension@metamask.io",
-      "strict_min_version": "68.0"
+      "strict_min_version": "78.0"
     }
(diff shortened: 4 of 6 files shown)


hmalik88 (dismissed):

darkwing (dismissed):

rekmarks (dismissed):

Gudahtt:
Apparently this might affect up to 1% of the userbase, though it's likely a bit lower than that. More details here: https://github.com/MetaMask/metamask-extension/issues/6805#issuecomment-979467319

Gudahtt:
I'm marking this as "DO-NOT-MERGE" for now, to give those away on holiday the chance to weigh in about whether we should do this now, or whether we should wait until after we add more in-app messaging asking users to update.

darkwing:
Idea that @Gudahtt and I were discussing was parsing the UA and, if the user is stuck on an old browser, displaying a persistent message on top of the home screen to let the user know they wont receive MM updates.  We'd need that to get merged the release before this gets merged.

Gudahtt:
The issue #16283 has been created to track adding a notification warning users if we're about to drop support for their browser. That will unblock this PR.

app/manifest/v2/chrome.json
```diff
@@ -4,5 +4,5 @@
     "matches": ["https://metamask.io/*"],
     "ids": ["*"]
   },
-  "minimum_chrome_version": "66"
+  "minimum_chrome_version": "78"
```
Gudahtt, line 7: Should probably update this to at least 80, so that we can use the optional chaining operator.

babel.config.js
```diff
@@ -5,7 +5,7 @@ module.exports = function (api) {
       strictMode: true,
     },
     targets: {
-      browsers: ['chrome >= 66', 'firefox >= 68'],
+      browsers: ['chrome >= 78', 'firefox >= 78'],
```
Gudahtt, line 8:
```suggestion
      browsers: ['chrome >= 80', 'firefox >= 78'],
```

digiwand (dismissed):
it appears the `DO-NOT-MERGE` tag was added for the last holiday season and can be removed now - LGTM! 🎄

Gudahtt:
Thanks for the review @digiwand ! This was blocked by #16283, but is unblocked now that that has been included in the v10.26.0 RC.

FrederikBolding (approved):

digiwand (approved):

PR #12855, opened 2021-11-25
Fix environment variables used during test builds

Gudahtt:
The environment variables used for test builds was wrong for certain bundles because the `testing` flag wasn't passed through to the function that determines which environment variables to inject. Effectively this means that test builds on `master` were going to the production `metamask` Sentry project rather than the `test-metamask` project. This has been the case since #11080.

The `testing` flag is now included for all bundles, and test builds now use the `test-metamask` Sentry project in all cases.

Manual testing steps:  
  - Create a test build (`yarn build test`) with the `CIRCLE_BRANCH` environment variable set to various values (e.g. `master`, `Version-v10.8.0`, `feature-branch`, `develop`). Ensure that the environment is always set to `testing` in the Sentry initialization console message in the background.

### DIFF
--- a/development/build/scripts.js
+++ b/development/build/scripts.js
@@ -200,18 +200,18 @@ function createScriptTasks({
     // this can run whenever
     const disableConsoleSubtask = createTask(
       `${taskPrefix}:disable-console`,
-      createTaskForBundleDisableConsole({ devMode }),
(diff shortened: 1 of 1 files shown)


danjm (approved):

adonesky1 (approved):
LGTM

Issue #12421, opened 2021-10-21
Account details edit allows duplicate labels


darkwing:
When editing an account name, the use is allowed to repeat an account name that already exists.

**Steps to reproduce (REQUIRED)**
Steps to reproduce the behavior, libraries used with version number, and/or any setup information to easily reproduce:

1. Create multiple accounts
2. From the three-dot menu, click Account Details
3. Click the pen icon next to the account name
4. Type an account name that matches another existing account, then submit

**Expected behavior**
Prevent submission, show an error that says something to the effect of "This account name already exists"

**Screenshots**

https://user-images.githubusercontent.com/46655/138191673-1eb05cb2-b72e-405c-9a86-820b2aa96363.mp4

EHaracic:
AC:
- On changing the account name - (Edit) we should check the name does not exist
- On creating a new account  (Create Account) we should check the name does not exist

PR #12867, opened 2021-11-26
Fix account name duplicates

filipsekulic:
**Fixes**: https://github.com/MetaMask/metamask-extension/issues/12421

**Explanation**:  
When editing an existing account name or adding a new account, the user is not allowed to repeat an account name that already exists.

**Screencast**:

https://user-images.githubusercontent.com/92310504/143558318-628d2e15-4e2f-43f7-8b37-53db9a7fcdce.mp4

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -74,7 +74,8 @@
     "message": "Account Name"
   },
   "accountNameDuplicate": {
-    "message": "This account name already exists"
+    "message": "This account name already exists",
+    "description": "This is an error message shown when the user enters a new account name that matches an existing account name"
   },
   "accountOptions": {
     "message": "Account Options"
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -73,6 +73,9 @@
   "accountName": {
     "message": "Account Name"
   },
+  "accountNameDuplicate": {
+    "message": "This account name already exists"
+  },
   "accountOptions": {
     "message": "Account Options"
   },
--- a/ui/components/app/modals/account-details-modal/account-details-modal.component.js
+++ b/ui/components/app/modals/account-details-modal/account-details-modal.component.js
@@ -8,6 +8,7 @@ import EditableLabel from '../../../ui/editable-label';
 import Button from '../../../ui/button';
 import { getURLHostName } from '../../../../helpers/utils/util';
 import { isHardwareKeyring } from '../../../../helpers/utils/hardware';
+import { getAccountsNames } from '../../../../selectors';
 
 export default class AccountDetailsModal extends Component {
   static propTypes = {
@@ -37,10 +38,6 @@ export default class AccountDetailsModal extends Component {
     } = this.props;
     const { name, address } = selectedIdentity;
 
-    const accountsNames = Object.values(accounts)
-      .map((item) => item.name)
-      .filter((itemName) => itemName !== name);
-
     const keyring = keyrings.find((kr) => {
       return kr.accounts.includes(address);
     });
@@ -57,7 +54,7 @@ export default class AccountDetailsModal extends Component {
           className="account-details-modal__name"
           defaultValue={name}
           onSubmit={(label) => setAccountLabel(address, label)}
-          accountsNames={accountsNames}
+          accountsNames={getAccountsNames(accounts, name)}
         />
 
         <QrView
--- a/ui/components/app/modals/account-details-modal/account-details-modal.component.js
+++ b/ui/components/app/modals/account-details-modal/account-details-modal.component.js
@@ -8,7 +8,6 @@ import EditableLabel from '../../../ui/editable-label';
 import Button from '../../../ui/button';
 import { getURLHostName } from '../../../../helpers/utils/util';
 import { isHardwareKeyring } from '../../../../helpers/utils/hardware';
-import { getAccountsNames } from '../../../../selectors';
 
 export default class AccountDetailsModal extends Component {
   static propTypes = {
@@ -41,6 +40,12 @@ export default class AccountDetailsModal extends Component {
     const keyring = keyrings.find((kr) => {
       return kr.accounts.includes(address);
     });
+
+    const getAccountsNames = (allAccounts, currentName) => {
+      return Object.values(allAccounts)
+        .map((item) => item.name)
+        .filter((itemName) => itemName !== currentName);
+    };
 
     let exportPrivateKeyFeatureEnabled = true;
     // This feature is disabled for hardware wallets
--- a/ui/components/app/modals/account-details-modal/account-details-modal.component.js
+++ b/ui/components/app/modals/account-details-modal/account-details-modal.component.js
@@ -17,6 +17,7 @@ export default class AccountDetailsModal extends Component {
     setAccountLabel: PropTypes.func,
     keyrings: PropTypes.array,
     rpcPrefs: PropTypes.object,
+    accounts: PropTypes.array,
   };
 
   static contextTypes = {
@@ -32,8 +33,13 @@ export default class AccountDetailsModal extends Component {
       setAccountLabel,
       keyrings,
       rpcPrefs,
+      accounts,
     } = this.props;
     const { name, address } = selectedIdentity;
+
+    const accountsNames = Object.values(accounts)
+      .map((item) => item.name)
+      .filter((itemName) => itemName !== name);
 
     const keyring = keyrings.find((kr) => {
       return kr.accounts.includes(address);
@@ -51,6 +57,7 @@ export default class AccountDetailsModal extends Component {
           className="account-details-modal__name"
           defaultValue={name}
           onSubmit={(label) => setAccountLabel(address, label)}
+          accountsNames={accountsNames}
         />
 
         <QrView
--- a/ui/components/ui/editable-label/editable-label.js
+++ b/ui/components/ui/editable-label/editable-label.js
@@ -85,18 +85,21 @@ class EditableLabel extends Component {
     const { className, accountsNames } = this.props;
 
     return (
-      <div>
+      <>
         <div className={classnames('editable-label', className)}>
           {isEditing ? this.renderEditing() : this.renderReadonly()}
         </div>
         {accountsNames.includes(value) ? (
           <div
-            className={classnames('send-v2__error', 'send-v2__error-amount')}
+            className={classnames(
+              'editable-label__error',
+              'editable-label__error-amount',
+            )}
           >
             {this.context.t('accountNameDuplicate')}
           </div>
         ) : null}
-      </div>
+      </>
     );
   }
 }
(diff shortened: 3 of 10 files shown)


mhasanagic:
Verified by QA

ui/pages/create-account/new-account.component.js
```diff
@@ -72,6 +92,7 @@ export default class NewAccountCreateForm extends Component {
               large
               className="new-account-create-form__button"
               onClick={createClick}
+              disabled={Boolean(accountsNames.includes(newAccountName))}
```
david0xd, line 95: Maybe to wrap this into a function accountCreateBtnDisabled(), or similar.
Something like the pattern used in `this.renderEditing()`, `this.renderReadonly()`.

david0xd (dismissed):

ui/pages/create-account/new-account.component.js
```diff
@@ -16,7 +17,19 @@ export default class NewAccountCreateForm extends Component {
...
     const { newAccountName, defaultAccountName } = this.state;
-    const { history, createAccount, mostRecentOverviewPage } = this.props;
+    const {
+      history,
+      createAccount,
+      mostRecentOverviewPage,
+      accounts,
+    } = this.props;
+
+    const accountsNames = accounts.map((item) => item.name);
```
darkwing, line 27: Can this, and the `map` in `ui/components/app/modals/account-details-modal/account-details-modal.component.js`, be turned into a helper function?  Maybe `accountNameExists(proposedAccountName)`?  I'm always scared when we do similar operations in multiple spots.

ui/pages/create-account/new-account.component.js
```diff
@@ -16,7 +17,19 @@ export default class NewAccountCreateForm extends Component {
...
+
+    const accountCreateButtonDisabled = () => {
+      return Boolean(accountsNames.includes(newAccountName));
```
darkwing, line 30: Includes is already `boolean` type, so we don't need to cast it.

ui/components/app/modals/account-details-modal/account-details-modal.component.js
```diff
@@ -32,9 +33,14 @@ export default class AccountDetailsModal extends Component {
       setAccountLabel,
       keyrings,
       rpcPrefs,
+      accounts,
     } = this.props;
     const { name, address } = selectedIdentity;
 
+    const accountsNames = Object.values(accounts)
+      .map((item) => item.name)
+      .filter((itemName) => itemName !== name);
```
darkwing, line 42: I propose that in the `map` and `filter`, we cast account names to lower case, so that we don't allow `Account 1` and `account 1`

darkwing (changes requested):

ui/components/app/modals/account-details-modal/account-details-modal.component.js
filipsekulic: We agreed with @danjm to support case-sensitive account names. @danjm, do you want us to implement it in the other way as suggested above?

CwanyJest95 (approved):

ui/components/app/modals/account-details-modal/account-details-modal.component.js
danjm: Yeah, I was thinking that it is okay if a user has both `Account 1` and `account 1`. These are only labels, so there is no risk of data corruption. And the user can edit at any time. Some users may want to take advantage of differences in casing... or may already do that intentionally.

Basically: I think there is not risk in allowing this, and if there is any possibility that someone wants to name one account `Triple AAA` and another `triple aaa` or whatever, then I thought it made sense just to stick with a strict comparison.

ui/pages/create-account/new-account.component.js
filipsekulic: Fixed in c71199a60e2949308b7659f613fe02a48f46d79d.

ui/pages/create-account/new-account.component.js
filipsekulic: Fixed in c71199a60e2949308b7659f613fe02a48f46d79d.

ui/pages/create-account/new-account.component.js
filipsekulic: Fixed in c71199a60e2949308b7659f613fe02a48f46d79d.

ui/components/app/modals/account-details-modal/account-details-modal.component.js
filipsekulic: Fixed in c71199a60e2949308b7659f613fe02a48f46d79d.

ui/selectors/selectors.js
```diff
@@ -806,3 +806,25 @@ export function getIsAdvancedGasFeeDefault(state) {
     Boolean(advancedGasFee?.maxBaseFee) && Boolean(advancedGasFee?.priorityFee)
   );
 }
+
+/**
+ *  To check if the specific accountName is included in the accountsNames array which is consisted of the current user's accounts
+ *  @returns Boolean
+ */
+
+export function accountNameExists(accounts, accountName) {
```
adonesky1, line 815: It doesn't seem like this needs to be a selector since its only used in new-account.component? I think you could just add it as a method on that component?

adonesky1: You could use the `getAccountsNames` selector in that component and then just use includes against it.

adonesky1 (changes requested):
Small change requested here

ui/selectors/selectors.js
filipsekulic: Fixed in 0a99efb6150d62886139fc2f1a8f2d71ea3533bd.

danjm (dismissed):

ui/pages/create-account/new-account.component.js
```diff
@@ -43,21 +50,39 @@ export default class NewAccountCreateForm extends Component {
...
     return (
       <div className="new-account-create-form">
         <div className="new-account-create-form__input-label">
           {this.context.t('accountName')}
         </div>
         <div>
           <input
-            className="new-account-create-form__input"
+            className={classnames('new-account-create-form__input', {
+              'new-account-create-form__input__error': accountNameExists(
```
Gudahtt, line 67: Nit: Rather than repeat `accountNameExists(accounts, newAccountName)` three times, it would be easier to read this code if it was called once and the result stored in a variable.

ui/pages/create-account/index.scss
```diff
@@ -87,6 +87,10 @@
     color: $scorpion;
     margin-top: 15px;
     padding: 0 20px;
+
+    &__error {
+      border: 1px solid $monzo;
```
Gudahtt, line 92: This color is listed in `ui/css/design-system/deprecated-colors.scss`, which indicates that it generally should not be used. See the `ui/css/design-system/colors.scss` module for the set of colors we are supposed to be using going forward.

The most appropriate color here is probably `$error-3`

```suggestion
      border: 1px solid $error-3;
```

app/_locales/en/messages.json
```diff
@@ -73,6 +73,9 @@
   "accountName": {
     "message": "Account Name"
   },
+  "accountNameDuplicate": {
+    "message": "This account name already exists"
```
Gudahtt, line 77: Nit: A `description` that explains what this is would be helpful for translators. e.g.

```suggestion
    "message": "This account name already exists",
    "description": "This is an error message shown when the user enters a new account name that matches an existing account name"
```

ui/pages/create-account/new-account.component.js
```diff
@@ -43,21 +50,39 @@ export default class NewAccountCreateForm extends Component {
...
             value={newAccountName}
             placeholder={defaultAccountName}
             onChange={(event) =>
               this.setState({ newAccountName: event.target.value })
             }
             autoFocus
           />
+          {accountNameExists(accounts, newAccountName) ? (
+            <div
+              className={classnames('send-v2__error', 'send-v2__error-amount')}
```
Gudahtt, line 81: These classes were written for the "Send" page, and should not be used here. The styles for this component should be under the `new-account-create-form` namespace.

ui/components/ui/editable-label/editable-label.js
```diff
@@ -73,12 +81,21 @@ class EditableLabel extends Component {
...
     return (
-      <div className={classnames('editable-label', className)}>
-        {isEditing ? this.renderEditing() : this.renderReadonly()}
+      <div>
+        <div className={classnames('editable-label', className)}>
+          {isEditing ? this.renderEditing() : this.renderReadonly()}
+        </div>
+        {accountsNames.includes(value) ? (
+          <div
+            className={classnames('send-v2__error', 'send-v2__error-amount')}
```
Gudahtt, line 94: These classes were written for the "Send" page, and should not be used here. The styles for this component should be under the `editable-label` namespace.

ui/components/ui/editable-label/editable-label.js
```diff
@@ -73,12 +81,21 @@ class EditableLabel extends Component {
...
   render() {
-    const { isEditing } = this.state;
-    const { className } = this.props;
+    const { isEditing, value } = this.state;
+    const { className, accountsNames } = this.props;
 
     return (
-      <div className={classnames('editable-label', className)}>
-        {isEditing ? this.renderEditing() : this.renderReadonly()}
+      <div>
```
Gudahtt, line 88: What is the purpose of this wraoper div? If you just want to return two elements here, you can do that using a React fragment rather than creating a new DOM element.

ui/pages/create-account/new-account.component.js
filipsekulic: Fixed in: d69e6682ef8cce5a6f0d7dd860acabac7ec614e3

ui/pages/create-account/index.scss
filipsekulic: Fixed in: d69e6682ef8cce5a6f0d7dd860acabac7ec614e3

app/_locales/en/messages.json
filipsekulic: Fixed in: d69e6682ef8cce5a6f0d7dd860acabac7ec614e3

ui/pages/create-account/new-account.component.js
filipsekulic: Fixed in: d69e6682ef8cce5a6f0d7dd860acabac7ec614e3

ui/components/ui/editable-label/editable-label.js
filipsekulic: Fixed in: d69e6682ef8cce5a6f0d7dd860acabac7ec614e3

ui/components/ui/editable-label/editable-label.js
filipsekulic: Fixed in: d69e6682ef8cce5a6f0d7dd860acabac7ec614e3

Gudahtt (approved):
LGTM!

darkwing (approved):

PR #12873, opened 2021-11-26
Feat/collectibles display

adonesky1:
Explanation:  Adds remaining required wiring + front-end work to get NFTs MVP in extension

In designs but excluded here
- Send flow (I think we agreed somewhere that this wasn't required for MVP... but I might totally be making this up)
- `Report as Scam` flow (the functionality here is not sufficiently spec'd out to implement)

Screenshots:
Screen Shot 2021-11-30 at 4 09 50 PM
Screen Shot 2021-11-30 at 4 10 54 PM

Manual testing steps:  
  - Purchase or mint or have an (or preferably many) NFT(s)
  Detection:
  - Turn on NFT autodetection in settings -> experimental and wait 2 mins
  - Check that your NFT has been added to the NFTs tab
  Manual:
  - Navigate to the NFTs tab
  - Click "Import NFTs"
  - Check that your NFT has been successfully added


darkwing (dismissed):
1.  We should prevent submission here without validated data, much as we do on other screens:
DisableAdd

2.  For the "learn more" and "Add NFT" links, I think we should use a `<Button>` component.  As of this moment, they aren't keyboard accessible.

adonesky1:
@darkwing good call, I've added form validation here: https://github.com/MetaMask/metamask-extension/pull/12873/commits/1966deeb9bb1a1cacb2f53d6bb2fd8681361ed87

and the "learn more" and "add nft" links are already using the `<Button>` component: https://github.com/MetaMask/metamask-extension/blob/8e6a0ff87959a1619ea7dc9e9e4228d0309d78f0/ui/components/app/collectibles-tab/collectibles-tab.component.js#L59? am I wrong?

georgewrmarshall (commented):
It's looking good! I do have some css suggestions but struggling with changing the ui through the dev build on my local. It takes really long to update a small change. Would be great if we could get the story working? Looks like it may have been before.

Screen Shot 2021-11-30 at 11 18 43 AM

Also looks like there are some prop type errors?
Screen Shot 2021-11-30 at 11 55 54 AM
Screen Shot 2021-11-30 at 11 56 15 AM

*As an aside I wouldn't say my css changes are critical to getting this merged*

ui/pages/settings/advanced-tab/advanced-tab.component.js
```diff
@@ -566,10 +566,7 @@ export default class AdvancedTab extends PureComponent {
   }
 
   handleIpfsGatewaySave() {
```
danjm, line 568: Why this change?

danjm: As discussed on a call, we need a migration to properly support this and an update to the `attemptResolve` function in the background

The reason this change is needed is because the controller now expects the ipfs gateway to be a url

adonesky1: should be addressed in https://github.com/MetaMask/metamask-extension/pull/12873/commits/a3732fcc6bc02e22c12fcd8450660f12963c4ad7

ui/pages/add-collectible/add-collectible.component.js
```diff
@@ -33,6 +35,24 @@ export default function AddCollectible() {
...
+    }
+    setAddress(val);
+  };
+
+  const validateAndSetTokenId = (val) => {
+    if (isValidHexAddress(address) && val) {
+      setDisabled(false);
+    } else {
+      setDisabled(true);
+    }
```
jpuri, line 52: `setDisabled(isValidHexAddress(address) && val)`

ui/components/ui/box/box.js
```diff
@@ -169,6 +169,6 @@ Box.propTypes = {
   display: PropTypes.oneOf(Object.values(DISPLAY)),
   width: PropTypes.oneOf(Object.values(BLOCK_SIZES)),
   height: PropTypes.oneOf(Object.values(BLOCK_SIZES)),
-  backgroundColor: PropTypes.oneOf(Object.values(COLORS)),
+  backgroundColor: PropTypes.string,
```
jpuri, line 172: Color I think should be one of predefined set of colors.

ui/components/app/collectibles-items/index.scss
```diff
@@ -0,0 +1,18 @@
...
+.collection-icon {
+  border-radius: 50%;
+  width: 2rem;
+  height: 2rem;
+  padding: 0.5rem;
+  background: $ui-4;
+  color: $ui-white;
+  text-align: center;
+  line-height: 1;
+}
```
jpuri, line 18: Above styles should be nested I think.

ui/components/app/collectibles-items/collectibles-items.component.js
```diff
@@ -109,15 +149,28 @@ export default function CollectiblesItems({ onAddNFT, onRefreshList }) {
...
+                  >
+                    {t('refreshList')}
+                  </Button>
+                </Box>
+              ) : (
+                <Box justifyContent={JUSTIFY_CONTENT.FLEX_END}>
+                  <Button
+                    type="link"
+                    onClick={onEnableAutoDetect}
+                    style={{ padding: '4px', fontSize: '16px' }}
```
jpuri, line 168: Lets avoid using inline styled like above.

ui/components/app/collectible-details/index.scss
```diff
@@ -0,0 +1,26 @@
...
+    font-size: $font-size-paragraph;
+    color: $Black-100;
+    background-color: inherit;
+    padding: 2px 0 2px 8px;
+  }
+
+  &__icon {
+    font-weight: 900;
+  }
+}
```
jpuri, line 26: It should have a parent classname I think into which all styles should be enclosed

jpuri (dismissed):
@adonesky1 : changes look good to me, I have left few small feedbacks. It will be useful to address those may be in a following PR.

app/scripts/migrations/068.js
```diff
@@ -0,0 +1,60 @@
...
+function transformState(state) {
+  const PreferencesController = state?.PreferencesController || {};
+  const preferences = PreferencesController.preferences || {};
+  const oldIpfsGateWay = preferences.ipfsGateway;
+
+  let newState;
+
+  if (oldIpfsGateWay && oldIpfsGateWay !== 'dweb.link') {
+    const newIpfsGateway = new URL(
+      `${addUrlProtocolPrefix(oldIpfsGateWay)}/ipfs/`,
```
danjm, line 34: do we want to add `/ipfs/` to every gateway?

ui/components/app/collectibles-items/collectibles-items.component.js
danjm: agreed, let's remove inline styles from this file

ui/pages/add-collectible/add-collectible.component.js
```diff
@@ -33,6 +35,24 @@ export default function AddCollectible() {
...
+    if (isValidHexAddress(val) && tokenId) {
+      setDisabled(false);
+    } else {
+      setDisabled(true);
+    }
+    setAddress(val);
+  };
+
+  const validateAndSetTokenId = (val) => {
+    if (isValidHexAddress(address) && val) {
```
danjm, line 48: should this be validating the `tokenId`?

danjm (changes requested):
The migration needs to be fixed, I think. + a few other small comments

app/scripts/migrations/068.js
adonesky1: Yes I believe we do for retrieving assets atleast. Might have to do a bit of research on this though.

ui/pages/add-collectible/add-collectible.component.js
adonesky1: This is just validating that there is a value for TokenId. No other validation I can think of? The TextField is already numeric.

adonesky1:
@jpuri had to rebase because of conflicts the only changes are https://github.com/MetaMask/metamask-extension/pull/12873/commits/5a43ab631ba93f2d993898872d26736a5bfa4917
cc @danjm

app/scripts/migrations/068.js
danjm: Here is a list of different gateways: https://ipfs.github.io/public-gateway-checker/

Maybe the addition of `/ipfs/` should be done in the functions/classes responsible for retrieving assets? Or maybe we do want to store a separate property after all, `nftMetadataGateway` or something like that

adonesky1: this concern should be addressed here: https://github.com/MetaMask/metamask-extension/pull/12873/commits/c68a608351921e2e456bb7e345ae338f71d7bf4e

ui/components/ui/box/box.js
adonesky1: Ok. Will address in subsequent PR!

ui/components/app/collectibles-items/index.scss
adonesky1: Agreed! Thanks! Will address in subsequent PR!

ui/components/app/collectibles-items/collectibles-items.component.js
adonesky1: Agreed! Thanks! Will address in subsequent PR! (I'm noting these requests for a followup ticket!)

ui/components/app/collectible-details/index.scss
adonesky1: Will be addressed in subsequent PR

app/scripts/controllers/preferences.js
```diff
@@ -61,7 +61,7 @@ export default class PreferencesController {
         hideZeroBalanceTokens: false,
       },
       // ENS decentralized website resolution
-      ipfsGateway: 'dweb.link',
+      ipfsGateway: 'https://cloudflare-ipfs.com/ipfs/',
```
jpuri, line 64: Not sure if hard coded url is ok.

ui/components/app/collectibles-items/collectibles-items.component.js
```diff
@@ -109,15 +149,28 @@ export default function CollectiblesItems({ onAddNFT, onRefreshList }) {
...
-              <Box justifyContent={JUSTIFY_CONTENT.FLEX_END}>
-                <Button
-                  type="link"
-                  onClick={onRefreshList}
-                  style={{ padding: '4px' }}
-                >
-                  {t('refreshList')}
-                </Button>
-              </Box>
+              {' '}
```
jpuri, line 152: `{' '}` looks like not needed here

jpuri (approved):
Some small feedbacks

danjm (approved):
LGTM

PR #12884, opened 2021-11-29
Use node-browsers-medium-plus executor for validate-lavamoat-config job

danjm:
The `validate-lavamoat-config` job was failing on the v10.7.0 branch, which is the first RC it's been included in. It seems that the job was taking a long time, so we needed to increase the runner size. I tested this by temporarily pushing it to the Version-v10.7.0 branch


Gudahtt (approved):
LTGM!

adonesky1 (approved):

PR #12886, opened 2021-11-29
Update ESLint config to v8

Gudahtt:
The ESLint config has been updated to v8. The breaking changes are:

* The Prettier rule `quoteProps` has been changed from `consistent` to `as-needed`, meaning that if one key requires quoting, only that key is quoted rather than all keys.
* The ESLint rule `no-shadow` has been made more strict. It now prevents globals from being shadowed as well.

Most of these changes were applied with `yarn lint:fix`. Only the shadowing changes required manual fixing (shadowing variable names were either replaced with destructuring or renamed).

The dependency `globalThis` was added to the list of dynamic dependencies in the build system, where it should have been already. This was causing `depcheck` to fail because the new lint rules required removing the one place where `globalThis` had been erroneously imported previously.

A rule requiring a newline between multiline blocks and expressions has been disabled temporarily to make this PR smaller and to avoid introducing conflicts with other PRs.


darkwing (dismissed):
I agree with the lint changes made!

darkwing:
Adding the `DO-NOT-MERGE` tag on this for now, and will bring it up during standup.  This could cause merge conflicts or existing PRs and our queue is at an eye-bleeding 130+.  We might need to wait until we get the queue under control.

Gudahtt:
Ok! I don't anticipate this creating any difficult to resolve conflicts, because most of the changes are fixable with `lint:fix`. But it will cause conflicts.

Gudahtt:
@darkwing This has been rebased to resolve conflicts, and the most disruptive rule has been disabled temporarily to reduce the size of the diff and avoid conflicts. I don't think this is likely to cause issues with the PR queue anymore. Let me know what you think!

mcmire:
> The Prettier rule `quoteProps` has been changed from `consistent` to
`as-needed`, meaning that if one key requires quoting, only that key is
quoted rather than all keys.

I don't know if I agree with this change. I feel like consistency is fine here, because it makes all keys look uniform. That said, if no keys need be quoted, then no keys should be quoted, so if there is a way to enforce _that_ then that seems more reasonable to me. Was there a problem that this addressed or was it to reduce git churn or something?

Gudahtt:
>if no keys need be quoted, then no keys should be quoted, so if there is a way to enforce that then that seems more reasonable to me. 

That's what the old setting did (`consistent`).

See [this issue](https://github.com/MetaMask/eslint-config/issues/175) and [this Slack thread](https://consensys.slack.com/archives/G1L7H42BT/p1621300185004800). Essentially, some people agree with you but "as-needed" won out the last time we asked the dev team. And yes, reducing git churn was a major factor.

mcmire:
@Gudahtt Ah, okay. Thank you!

mcmire (approved):

darkwing (approved):

PR #12890, opened 2021-11-30
Jestify app/scripts/lib/**/*.test.js

tmashuang:
Jestify app/scripts/lib/**/*.test.js


darkwing (approved):

segun (approved):

app/scripts/lib/metaRPCClientFactory.test.js
```diff
@@ -86,13 +81,12 @@ describe('metaRPCClientFactory', function () {
     });
   });
 
-  it('should be able to handle notifications', function (done) {
+  it('should be able to handle notifications', () => {
```
naugtur, line 84: all tests where done was removed and no other mechanism introduced will always pass.

PR #12905, opened 2021-11-30
"eth_signTypedData" presents fields that do not appear in 'types' filed

### LINKED ISSUE #11457: "eth_signTypedData" presents fields that do not appear in 'types' filed
**Describe the bug**
When executing a call to 'eth_signTypedData', the expectation is that the message that is presented to the user and the data that will be signed must match exactly. In fact, what seems to happen is that MetaMask presents the entire 'message' field, including fields that do not appear in the 'types' field. If that is in fact what is happening, it seems to present some security risk as clients can be expecting to sign on some fields, such as "expirationBlock" and such, but in fact these fields are not part of EIP-712 signature created.

Note that the pseudocode is a simplified version of affected production code:

1. Define an EIP-712 type structure:
```
dataToSign.type = [
  { name: 'operationA', type: 'uint256' },
  { name: 'operationB', type: 'uint256' }
]
``` 

2. Create an object:
```
dataToSign.message = {
  operationA: 123,
  operationA: 456,
  expirationBlock: 12777777
}
```
3. Execute

```
web3.currentProvider.send({
  method: 'eth_signTypedData_v4',
  params: [senderAddress, dataToSign],
  from: senderAddress
})
```

**Expected behavior**
Expectation: `expirationBlock` field does not appear in a signature popup. UPD: preferably, exception is raised indicating that structure does not match the supposed data type.
Actual: `expirationBlock` field appears as usual, but is not part of a signed data.

segun:
Fixes: #11457 

Explanation:  
When executing a call to 'eth_signTypedData', the expectation is that the message that is presented to the user and the data that will be signed must match exactly. In fact, what seems to happen is that MetaMask presents the entire 'message' field, including fields that do not appear in the 'types' field. If that is in fact what is happening, it seems to present some security risk as clients can be expecting to sign on some fields, such as "expirationBlock" and such, but in fact these fields are not part of EIP-712 signature created.

Manual testing steps:  
see #11457 
  - 
  - 
  -

### DIFF
--- a/ui/components/app/signature-request/signature-request.component.js
+++ b/ui/components/app/signature-request/signature-request.component.js
@@ -5,8 +5,6 @@ import LedgerInstructionField from '../ledger-instruction-field';
 import Header from './signature-request-header';
 import Footer from './signature-request-footer';
 import Message from './signature-request-message';
-import { TypedDataUtils } from 'eth-sig-util';
-import { MESSAGE_TYPE } from '../../../../shared/constants/app';
 
 export default class SignatureRequest extends PureComponent {
   static propTypes = {
@@ -53,39 +51,48 @@ export default class SignatureRequest extends PureComponent {
     const mapType = (msgType, definedType) => {
       // try and map it to a solidity type
       if (
-        (
-          definedType.type.indexOf("int") >= 0 ||
-          definedType.type.indexOf("fixed") >= 0
-        ) && msgType === "number") {
+        (definedType.type.indexOf('int') >= 0 ||
+          definedType.type.indexOf('fixed') >= 0) &&
+        msgType === 'number'
+      ) {
         return true;
       } else if (
-        (
-          definedType.type.indexOf("bytes") >= 0 ||
-          definedType.type === "address"
-        ) && msgType === "string") {
+        (definedType.type.indexOf('bytes') >= 0 ||
+          definedType.type === 'address') &&
+        msgType === 'string'
+      ) {
         return true;
       }
 
       return false;
-    }
+    };
 
     const sanitizeMessage = (msg, baseType) => {
-      let sanitizedMessage = {};
+      const sanitizedMessage = {};
       const msgKeys = Object.keys(msg);
-      const msgTypes = Object.values(msg).map(value => typeof (value));
+      const msgTypes = Object.values(msg).map((value) => typeof value);
       const baseTypeType = types[baseType];
       msgKeys.forEach((msgKey, index) => {
         const valueType = msgTypes[index];
         if (baseTypeType) {
-          const definedType = Object.values(baseTypeType).find(ptt => ptt.name === msgKey);          
+          const definedType = Object.values(baseTypeType).find(
+            (ptt) => ptt.name === msgKey,
+          );
 
           if (definedType) {
             if (definedType.type === valueType) {
               sanitizedMessage[msgKey] = msg[msgKey];
-            } else if (definedType.type.indexOf("[]") && valueType === "object" && Array.isArray(msg[msgKey])) {
+            } else if (
+              definedType.type.indexOf('[]') &&
+              valueType === 'object' &&
+              Array.isArray(msg[msgKey])
+            ) {
               sanitizedMessage[msgKey] = msg[msgKey];
-            } else if (valueType === "object") {
-              sanitizedMessage[msgKey] = sanitizeMessage(msg[msgKey], definedType.type);
+            } else if (valueType === 'object') {
+              sanitizedMessage[msgKey] = sanitizeMessage(
+                msg[msgKey],
+                definedType.type,
+              );
             } else if (mapType(valueType, definedType)) {
               sanitizedMessage[msgKey] = msg[msgKey];
             }
@@ -94,7 +101,7 @@ export default class SignatureRequest extends PureComponent {
       });
 
       return sanitizedMessage;
-    }
+    };
 
     const onSign = (event) => {
       sign(event);
@@ -127,7 +134,7 @@ export default class SignatureRequest extends PureComponent {
     };
 
     return (
-      <div className="signature-request page-container" >
+      <div className="signature-request page-container">
         <Header fromAccount={fromAccount} />
         <div className="signature-request-content">
           <div className="signature-request-content__title">
@@ -159,7 +166,7 @@ export default class SignatureRequest extends PureComponent {
           signAction={onSign}
           disabled={hardwareWalletRequiresConnection}
         />
-      </div >
+      </div>
     );
   }
 }
--- a/ui/components/app/signature-request/signature-request.component.js
+++ b/ui/components/app/signature-request/signature-request.component.js
@@ -68,39 +68,43 @@ export default class SignatureRequest extends PureComponent {
     };
 
     const sanitizeMessage = (msg, baseType) => {
-      const sanitizedMessage = {};
-      const msgKeys = Object.keys(msg);
-      const msgTypes = Object.values(msg).map((value) => typeof value);
-      const baseTypeType = types[baseType];
-      msgKeys.forEach((msgKey, index) => {
-        const valueType = msgTypes[index];
-        if (baseTypeType) {
-          const definedType = Object.values(baseTypeType).find(
-            (ptt) => ptt.name === msgKey,
-          );
+      if (version === 'V4') {
+        const sanitizedMessage = {};
+        console.log(version);
+        const msgKeys = Object.keys(msg);
+        const msgTypes = Object.values(msg).map((value) => typeof value);
+        const baseTypeType = types[baseType];
+        msgKeys.forEach((msgKey, index) => {
+          const valueType = msgTypes[index];
+          if (baseTypeType) {
+            const definedType = Object.values(baseTypeType).find(
+              (ptt) => ptt.name === msgKey,
+            );
 
-          if (definedType) {
-            if (definedType.type === valueType) {
-              sanitizedMessage[msgKey] = msg[msgKey];
-            } else if (
-              definedType.type.indexOf('[]') &&
-              valueType === 'object' &&
-              Array.isArray(msg[msgKey])
-            ) {
-              sanitizedMessage[msgKey] = msg[msgKey];
-            } else if (valueType === 'object') {
-              sanitizedMessage[msgKey] = sanitizeMessage(
-                msg[msgKey],
-                definedType.type,
-              );
-            } else if (mapType(valueType, definedType)) {
-              sanitizedMessage[msgKey] = msg[msgKey];
+            if (definedType) {
+              if (definedType.type === valueType) {
+                sanitizedMessage[msgKey] = msg[msgKey];
+              } else if (
+                definedType.type.indexOf('[]') &&
+                valueType === 'object' &&
+                Array.isArray(msg[msgKey])
+              ) {
+                sanitizedMessage[msgKey] = msg[msgKey];
+              } else if (valueType === 'object') {
+                sanitizedMessage[msgKey] = sanitizeMessage(
+                  msg[msgKey],
+                  definedType.type,
+                );
+              } else if (mapType(valueType, definedType)) {
+                sanitizedMessage[msgKey] = msg[msgKey];
+              }
             }
           }
-        }
-      });
+        });
 
-      return sanitizedMessage;
+        return sanitizedMessage;
+      }
+      return msg;
     };
 
     const onSign = (event) => {
--- a/ui/components/app/signature-request/signature-request.component.js
+++ b/ui/components/app/signature-request/signature-request.component.js
@@ -69,29 +69,28 @@ export default class SignatureRequest extends PureComponent {
       return false;
     }
 
-    const sanitizeMessage = () => {
-      let sanitizedMessage = TypedDataUtils.sanitizeData(JSON.parse(data));
-      const msgKeys = Object.keys(message);      
-      const msgTypes = Object.values(message).map(value => typeof (value));            
-      const primaryTypeType = types[primaryType];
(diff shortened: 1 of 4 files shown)


ui/components/app/signature-request/signature-request.component.js
```diff
@@ -45,9 +45,85 @@ export default class SignatureRequest extends PureComponent {
...
+    };
+
+    const sanitizeMessage = (msg, baseType) => {
+      if (version === 'V4') {
+        const sanitizedMessage = {};
+        const msgKeys = Object.keys(msg);
+        const msgTypes = Object.values(msg).map((value) => typeof value);
+        const baseTypeType = types[baseType];
+        msgKeys.forEach((msgKey, index) => {
+          const valueType = msgTypes[index];
```
Gudahtt, line 77: Please use `Object.entries` to iterate over the keys and values of an object, rather than using `Object.keys` and `Object.values` and tying them together with the index. It's much easier to read that way, and it avoids concerns of iteration order not being standardized for objects.

Gudahtt: Also, what is the purpose of looking at the `typeof` of the value? The type of the value is not useful, and is entirely distinct from the type defined in the message.

ui/components/app/signature-request/signature-request.component.js
```diff
@@ -45,9 +45,85 @@ export default class SignatureRequest extends PureComponent {
...
+          if (baseTypeType) {
```
Gudahtt, line 78: This doesn't need to be in the loop. I'd suggest checking this before the loop, and negating this condition and exiting early rather than indenting everything.

ui/components/app/signature-request/signature-request.component.js
```diff
@@ -45,9 +45,85 @@ export default class SignatureRequest extends PureComponent {
...
+        msgType === 'string'
+      ) {
+        return true;
+      }
+
+      return false;
+    };
+
+    const sanitizeMessage = (msg, baseType) => {
+      if (version === 'V4') {
```
Gudahtt, line 71: Why only V4?

segun: I thought only V4 has types...just checked test-app and saw it not.

ui/components/app/signature-request/signature-request.component.js
```diff
@@ -45,9 +45,72 @@ export default class SignatureRequest extends PureComponent {
...
+    const solidityTypes = [
+      'bool',
+      'int',
+      'fixed',
+      'address',
+      'bytes',
+      'string',
+    ];
+    const sanitizeMessage = (msg, baseType) => {
+      let baseTypeType;
```
jpuri, line 60: better name for this `baseTypeType` ?

ui/components/app/signature-request/signature-request.component.js
```diff
@@ -45,9 +45,72 @@ export default class SignatureRequest extends PureComponent {
       hardwareWalletRequiresConnection,
     } = this.props;
     const { address: fromAddress } = fromAccount;
-    const { message, domain = {} } = JSON.parse(data);
+    const { message, domain = {}, types, primaryType } = JSON.parse(data);
     const { metricsEvent } = this.context;
```
jpuri, line 50: It will be easy to understand `sanitizeMessage` if it can be extracted into a util function and test covered.

ui/components/app/signature-request/signature-request.component.js
Gudahtt: Maybe `baseTypeDefinition`?

ui/components/app/signature-request/signature-request.component.js
```diff
@@ -45,9 +45,72 @@ export default class SignatureRequest extends PureComponent {
...
+        }
+      } else {
+        return msg;
+      }
+
+      const sanitizedMessage = {};
+      const msgKeys = Object.keys(msg);
+      msgKeys.forEach((msgKey) => {
+        const definedType = Object.values(baseTypeType).find(
+          (ptt) => ptt.name === msgKey,
```
Gudahtt, line 74: What does `ptt` mean? We should replace this with a better variable name. e.g. `property`

ui/components/app/signature-request/signature-request.component.js
Gudahtt: Note that this has not yet been addressed. You're still using `Object.keys` and `Object.values` instead of `Object.entries`.

ui/components/app/signature-request/signature-request.component.js
```diff
@@ -45,9 +45,72 @@ export default class SignatureRequest extends PureComponent {
...
+              // nested object
+              sanitizedMessage[msgKey] = sanitizeMessage(
+                msg[msgKey],
+                definedType.type,
+              );
+            }
+          } else {
+            // check if it's a valid solidity type
+            const isSolidityType = solidityTypes.some(
+              (solidityType) => definedType.type.indexOf(solidityType) >= 0,
```
Gudahtt, line 103: This would allow invalid types that started with a solidity type. For example `stringfoo` would be recognized as a Solidity type here, despite not being a real type.

ui/components/app/signature-request/signature-request.component.js
Gudahtt: And also the below variable `nestedTypeType` could be `nestedTypeDefinition`

ui/components/app/signature-request/signature-request.component.js
```diff
@@ -45,9 +45,72 @@ export default class SignatureRequest extends PureComponent {
...
+              const sanitizedArrayMessage = {};
+              const definedArrayType = definedType.type.replace('[]', '');
+              const arrayMsg = msg[msgKey];
+              arrayMsg.forEach((msgArray, arrIndex) => {
+                sanitizedArrayMessage[arrIndex] = sanitizeMessage(
+                  msgArray,
+                  definedArrayType,
+                );
+              });
+              sanitizedMessage[msgKey] = sanitizedArrayMessage;
```
Gudahtt, line 92: This seems overly complex. This can be done more simply with `.map`.

e.g.

```suggestion
              sanitizedMessage[msgKey] = msg[msgKey].map((value) => sanitizeMessage(value, definedArrayType));
```

ui/components/app/signature-request/signature-request.component.js
Gudahtt: I must have misunderstood; there is no longer a secondary call to `Object.values` on the same object. I was mistaking the type definition object for the message I think.

You still could use `Object.entries` here but it would be a minimal improvement to readability at best.

ui/helpers/utils/util.js
```diff
@@ -443,3 +443,77 @@ export const toHumanReadableTime = (t, milliseconds) => {
...
+const solidityTypes = () => {
+  const types = [
+    'bool',
+    'address',
+    'string',
+    'bytes',
+    'int',
+    'uint',
+    'fixed',
+    'ufixed',
```
Gudahtt, line 456: We should also include the explicitly sized versions of `fixed` and `ufixed`, i.e. `ufixedMxN` where M is divisible by 8 and between 8 and 256 (inclusive), and N is an integer between 0 and 80.

Details: https://docs.soliditylang.org/en/latest/types.html#fixed-point-numbers

ui/components/app/signature-request/signature-request.component.js
```diff
@@ -105,7 +106,7 @@ export default class SignatureRequest extends PureComponent {
             <LedgerInstructionField showDataInstruction />
           </div>
         ) : null}
-        <Message data={message} />
+        <Message data={sanitizeMessage(message, primaryType)} />
```
Gudahtt, line 109: It looks like the third parameter, `types`, is supposed to be passed in here

ui/helpers/utils/util.js
```diff
@@ -443,3 +443,77 @@ export const toHumanReadableTime = (t, milliseconds) => {
...
+      } else {
+        // check if it's a valid solidity type
+        const isSolidityType = solidityTypes().includes(
+          definedType.type.replace('[]', ''),
+        );
+        if (isSolidityType) {
+          sanitizedMessage[msgKey] = msg[msgKey];
+        }
+      } // endif (nestedType)
+    } // endif (definedType)
```
Gudahtt, line 516: Nit: To avoid lots of indentation, and the need for comments like this, you could invert the condition.

Rather than:

```
if (definedType) {
  ...
}
```
Do something like this:

```
if (!definedType) {
  return;
}
...
```

ui/helpers/utils/util.js
```diff
@@ -443,3 +443,77 @@ export const toHumanReadableTime = (t, milliseconds) => {
...
+          sanitizedMessage[msgKey] = sanitizeMessage(
+            msg[msgKey],
+            definedType.type,
+            types,
+          );
+        }
+      } else {
+        // check if it's a valid solidity type
+        const isSolidityType = solidityTypes().includes(
+          definedType.type.replace('[]', ''),
```
Gudahtt, line 510: We should exclude types with `[]` at the beginning or middle as well:
```suggestion
          definedType.type.replace(\[\]$/, ''),
```

ui/helpers/utils/util.test.js
```diff
@@ -340,4 +340,101 @@ describe('util', () => {
       expect(util.toHumanReadableTime(t, 7200000)).toStrictEqual('2 hrs');
     });
   });
+  describe('sanitizeMessage', () => {
+    const message = {
```
Gudahtt, line 344: Nit: Better to move these fixtures to a function, and call this function in each test to get a fresh copy, just in case one of the tests mutates the object.

If these get mutated by one of the tests, it could invalidate the test result, and it might also be very difficult to track down the issue. It could also lead to dependencies between tests.

ui/helpers/utils/util.test.js
```diff
@@ -340,4 +340,101 @@ describe('util', () => {
...
+      EIP712Domain: [
+        { name: 'name', type: 'string' },
+        { name: 'version', type: 'string' },
+        { name: 'chainId', type: 'uint256' },
+        { name: 'verifyingContract', type: 'address' },
+      ],
+      Group: [
+        { name: 'name', type: 'string' },
+        { name: 'members', type: 'Person[]' },
+      ],
```
Gudahtt, line 375: This type appears to be unused :thinking: 

```suggestion
```

ui/helpers/utils/util.js
```diff
@@ -443,3 +443,77 @@ export const toHumanReadableTime = (t, milliseconds) => {
...
+  );
+  return [...types, ...ints, ...uints, ...bytes];
+};
+
+export const sanitizeMessage = (msg, baseType, types) => {
+  let baseTypeDefinitions;
+  if (types) {
+    baseTypeDefinitions = types[baseType];
+    if (!baseTypeDefinitions) {
+      return msg;
```
Gudahtt, line 476: Is it possible for a signature request to get this far without any types, or with no primary type? I would hope this would be caught by our validation, but maybe not.

If this is possible, showing the user a message preview that could be totally wrong sounds dangerous :thinking:

segun: I agree, but not sure what to do in these instance.

ui/helpers/utils/util.js
```diff
@@ -443,3 +443,109 @@ export const toHumanReadableTime = (t, milliseconds) => {
...
+    // key has a type. check if the definedType is also a type
+    const nestedTypeDefinition = types[definedType.type.replace('[]', '')];
+
+    if (nestedTypeDefinition) {
+      if (definedType.type.indexOf('[]') > 0) {
+        // nested array
+        const definedArrayType = definedType.type.replace('[]', '');
+        sanitizedMessage[msgKey] = msg[msgKey].map((value) =>
+          sanitizeMessage(value, definedArrayType, types),
+        );
```
Gudahtt, line 531: Same thing here re: ensuring the type ends in `[]`
```suggestion
    const nestedTypeDefinition = types[definedType.type.replace(/\[\]$/, '')];

    if (nestedTypeDefinition) {
      if (definedType.type.endsWith('[]')) {
        // nested array
        const definedArrayType = definedType.type.replace(/\[\]$/, '')];
        sanitizedMessage[msgKey] = msg[msgKey].map((value) =>
          sanitizeMessage(value, definedArrayType, types),
        );
```

Or, to cache the nested type name:

```suggestion
    const nestedType = definedType.type.replace(/\[\]$/, '');
    const nestedTypeDefinition = types[nestedType];

    if (nestedTypeDefinition) {
      if (definedType.type.endsWith('[]')) {
        // nested array
        sanitizedMessage[msgKey] = msg[msgKey].map((value) =>
          sanitizeMessage(value, nestedType, types),
        );
```

ui/helpers/utils/util.js
Gudahtt: I have just confirmed that it is safe to assume here that `types` is defined, and that `primaryType` is included in `types`. So we should throw an error here for those cases.

This is guaranteed by the validation performed here: https://github.com/MetaMask/metamask-extension/blob/5e9ea9d8fad692d6b8346ccd72fc65a78e29164c/app/scripts/lib/typed-message-manager.js#L192

This is done before the message gets added to our state as an unapproved message. Validation is done [here](https://github.com/MetaMask/metamask-extension/blob/5e9ea9d8fad692d6b8346ccd72fc65a78e29164c/app/scripts/lib/typed-message-manager.js#L127), whereas the message isn't added until later ([here](https://github.com/MetaMask/metamask-extension/blob/5e9ea9d8fad692d6b8346ccd72fc65a78e29164c/app/scripts/lib/typed-message-manager.js#L143)).

Gudahtt (approved):
LGTM!

jpuri (approved):

PR #12911, opened 2021-11-30
EIP-1559 V2: Adding default settings to advanced gas modal

### LINKED ISSUE #12469: EIP-1559 V2: Advanced gas fee modal - user to be able to set default preferences
In custom gas fee modal user should be able to set default multiplier / GWEI value for:
1. base fee
2. priority fee

User should also be able to set if custom should be used by default.

New value should be saved as user click Save button.

Screen Shot 2021-11-18 at 9 48 26 AM

Screen Shot 2021-11-18 at 9 50 49 AM

Redux state changes and actions to update these preferences will be made available as part of this task: https://github.com/MetaMask/metamask-extension/issues/12447

TDB: Question here is if user edit base fee GWEI value how will be multiplier derived to be saved as default.

NiranjanaBinoy:
Fixes: #12469 

Enabling the default settings will save the `maxBaseFee` (multiplier value) and `priorityFee` to the state so that they can be used as a user preference in future transactions.

Scenarios covered:

- Checking the checkbox will save the current `maxBaseFee` and `priorityFee` to advancedGasFee and the message will be updated to `Always use these values and advanced setting as default.` 
- Updating any of the `maxBaseFee` and `priorityFee`  while the checkbox is checked will uncheck the box and update the message to `Use the “new values” and advanced setting as default.`, but will retain the existing default preference values.
- Rechecking the checkbox after updating the values will save the new values.
- Unchecking the checkbox manually will remove the existing default preference values.

Screenshot:
Screen Shot 2021-12-03 at 9 39 39 PM

Screen recording:

https://user-images.githubusercontent.com/43930900/146818352-fd92558d-444f-4278-b899-f571f823d646.mov

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -171,7 +171,7 @@
     "message": "When your transaction gets included in the block, any difference between your max base fee and the actual base fee will be refunded. Total amount is calculated as max base fee (in GWEI) * gas limit."
   },
   "advancedGasFeeDefaultOptIn": {
-    "message": "Use the “new values” and advanced setting as default."
+    "message": "Update the values used by advanced settings by default."
   },
   "advancedGasFeeDefaultOptOut": {
     "message": "Always use these values and advanced setting as default."
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -171,7 +171,7 @@
     "message": "When your transaction gets included in the block, any difference between your max base fee and the actual base fee will be refunded. Total amount is calculated as max base fee (in GWEI) * gas limit."
   },
   "advancedGasFeeDefaultOptIn": {
-    "message": "Update the values used by advanced settings by default."
+    "message": "Save these $1 as my default for \"Advanced\""
   },
   "advancedGasFeeDefaultOptOut": {
     "message": "Always use these values and advanced setting as default."
@@ -1856,6 +1856,9 @@
   },
   "newTransactionFee": {
     "message": "New Transaction Fee"
+  },
+  "newValues": {
+    "message": "new values"
   },
   "next": {
     "message": "Next"
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -176,6 +176,12 @@
   "advancedGasPriceTitle": {
     "message": "Gas price"
   },
+  "advancedGassFeeDefaultOptIn": {
+    "message": "Use the “new values” and advanced setting as default."
+  },
+  "advancedGassFeeDefaultOptOut": {
+    "message": "Always use these values and advanced setting as default."
+  },
   "advancedOptions": {
     "message": "Advanced Options"
   },
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -170,17 +170,17 @@
   "advancedBaseGasFeeToolTip": {
     "message": "When your transaction gets included in the block, any difference between your max base fee and the actual base fee will be refunded. Total amount is calculated as max base fee (in GWEI) * gas limit."
   },
+  "advancedGasFeeDefaultOptIn": {
+    "message": "Use the “new values” and advanced setting as default."
+  },
+  "advancedGasFeeDefaultOptOut": {
+    "message": "Always use these values and advanced setting as default."
+  },
   "advancedGasFeeModalTitle": {
     "message": "Advanced gas fee"
   },
   "advancedGasPriceTitle": {
     "message": "Gas price"
-  },
-  "advancedGassFeeDefaultOptIn": {
-    "message": "Use the “new values” and advanced setting as default."
-  },
-  "advancedGassFeeDefaultOptOut": {
-    "message": "Always use these values and advanced setting as default."
   },
   "advancedOptions": {
     "message": "Advanced Options"
--- a/ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
+++ b/ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
@@ -1,5 +1,5 @@
-import React, { useState } from 'react';
-import { useSelector } from 'react-redux';
+import React, { useCallback, useState } from 'react';
+import { useSelector, useDispatch } from 'react-redux';
 import Box from '../../../ui/box';
 import Typography from '../../../ui/typography';
 import CheckBox from '../../../ui/check-box';
@@ -10,12 +10,38 @@ import {
   FLEX_DIRECTION,
   TYPOGRAPHY,
 } from '../../../../helpers/constants/design-system';
-
+import { useAdvanceGasFeePopoverContext } from '../context';
 import { getIsAdvancedGasFeeDefault } from '../../../../selectors';
+import { setAdvancedGasFee } from '../../../../store/actions';
 
 const AdvancedGasFeeDefaults = () => {
+  const dispatch = useDispatch();
+
   const isAdvancedGasFeeDefault = useSelector(getIsAdvancedGasFeeDefault);
   const [defaultValues, setDefaultValues] = useState(isAdvancedGasFeeDefault);
+
+  const {
+    maxFeePerGas,
+    maxPriorityFeePerGas,
+  } = useAdvanceGasFeePopoverContext();
+
+  const updateDefaultSettings = useCallback(
+    (value) => {
+      setDefaultValues(value);
+      if (value) {
+        dispatch(
+          setAdvancedGasFee({
+            maxBaseFee: maxFeePerGas,
+            priorityFee: maxPriorityFeePerGas,
+          }),
+        );
+      } else {
+        dispatch(setAdvancedGasFee(null));
+      }
+    },
+    [maxFeePerGas, maxPriorityFeePerGas, setDefaultValues, dispatch],
+  );
+
   return (
     <Box
       display={DISPLAY.FLEX}
@@ -31,7 +57,7 @@ const AdvancedGasFeeDefaults = () => {
         <CheckBox
           checked={defaultValues}
           className="advanced-gas-fee-defaults__checkbox"
-          onClick={() => setDefaultValues((checked) => !checked)}
+          onClick={() => updateDefaultSettings(!defaultValues)}
         />
         <Typography variant={TYPOGRAPHY.H7} color={COLORS.UI4}>
           {defaultValues ? (
--- a/ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
+++ b/ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
@@ -1,5 +1,6 @@
-import React, { useCallback, useState, useEffect } from 'react';
+import React from 'react';
 import { useSelector, useDispatch } from 'react-redux';
+
 import Box from '../../../ui/box';
 import Typography from '../../../ui/typography';
 import CheckBox from '../../../ui/check-box';
@@ -10,12 +11,13 @@ import {
   FLEX_DIRECTION,
   TYPOGRAPHY,
 } from '../../../../helpers/constants/design-system';
-import { useAdvanceGasFeePopoverContext } from '../context';
 import {
   getIsAdvancedGasFeeDefault,
   getAdvancedGasFeeValues,
 } from '../../../../selectors';
 import { setAdvancedGasFee } from '../../../../store/actions';
+
+import { useAdvanceGasFeePopoverContext } from '../context';
 
 const AdvancedGasFeeDefaults = () => {
   const dispatch = useDispatch();
@@ -26,41 +28,23 @@ const AdvancedGasFeeDefaults = () => {
   } = useAdvanceGasFeePopoverContext();
   const isAdvancedGasFeeDefault = useSelector(getIsAdvancedGasFeeDefault);
   const advancedGasFeeValues = useSelector(getAdvancedGasFeeValues);
-  const [defaultValues, setDefaultValues] = useState(isAdvancedGasFeeDefault);
 
-  const updateDefaultSettings = useCallback(
-    (value) => {
-      setDefaultValues(value);
-      if (value) {
-        dispatch(
-          setAdvancedGasFee({
-            maxBaseFee: baseFeeMultiplier,
-            priorityFee: maxPriorityFeePerGas,
-          }),
-        );
-      } else {
-        dispatch(setAdvancedGasFee(null));
-      }
-    },
-    [baseFeeMultiplier, maxPriorityFeePerGas, setDefaultValues, dispatch],
-  );
-
-  useEffect(() => {
-    if (
-      !isAdvancedGasFeeDefault ||
-      advancedGasFeeValues.maxBaseFee !== baseFeeMultiplier ||
-      advancedGasFeeValues.priorityFee !== maxPriorityFeePerGas
-    ) {
-      setDefaultValues(false);
+  const updateDefaultSettings = (value) => {
+    if (value) {
+      dispatch(
+        setAdvancedGasFee({
+          maxBaseFee: baseFeeMultiplier,
+          priorityFee: maxPriorityFeePerGas,
+        }),
+      );
     } else {
-      setDefaultValues(true);
+      dispatch(setAdvancedGasFee(null));
     }
-  }, [
-    baseFeeMultiplier,
-    maxPriorityFeePerGas,
-    advancedGasFeeValues,
-    isAdvancedGasFeeDefault,
-  ]);
+  };
+  const defaultPreference =
+    isAdvancedGasFeeDefault &&
+    advancedGasFeeValues.maxBaseFee === baseFeeMultiplier &&
+    advancedGasFeeValues.priorityFee === maxPriorityFeePerGas;
 
   return (
     <Box
@@ -69,16 +53,18 @@ const AdvancedGasFeeDefaults = () => {
       className="advanced-gas-fee-defaults"
     >
       <CheckBox
-        checked={defaultValues}
+        checked={defaultPreference}
         className="advanced-gas-fee-defaults__checkbox"
-        onClick={() => updateDefaultSettings(!defaultValues)}
+        onClick={() => updateDefaultSettings(!defaultPreference)}
       />
       <Typography variant={TYPOGRAPHY.H7} color={COLORS.UI4}>
-        {defaultValues ? (
-          <I18nValue messageKey="advancedGassFeeDefaultOptOut" />
-        ) : (
-          <I18nValue messageKey="advancedGassFeeDefaultOptIn" />
-        )}
+        <I18nValue
+          messageKey={
+            defaultPreference
+              ? 'advancedGasFeeDefaultOptOut'
+              : 'advancedGasFeeDefaultOptIn'
+          }
+        />
       </Typography>
     </Box>
   );
--- a/ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
+++ b/ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
@@ -54,7 +54,7 @@ const AdvancedGasFeeDefaults = () => {
         className="advanced-gas-fee-defaults__checkbox"
         onClick={() => updateDefaultSettings(!defaultPreference)}
       />
-      <Typography variant={TYPOGRAPHY.H7} color={COLORS.UI4}>
+      <Typography variant={TYPOGRAPHY.H7} color={COLORS.UI4} margin={0}>
         <I18nValue
           messageKey={
             defaultPreference
--- a/ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
+++ b/ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
@@ -40,13 +40,13 @@ const AdvancedGasFeeDefaults = () => {
       dispatch(setAdvancedGasFee(null));
     }
   };
-  const defaultPreference =
+  const isDefaultSettingsSelected =
     Boolean(advancedGasFeeValues) &&
     advancedGasFeeValues.maxBaseFee === baseFeeMultiplier &&
     advancedGasFeeValues.priorityFee === maxPriorityFeePerGas;
 
   const handleUpdateDefaultSettings = () =>
-    updateDefaultSettings(!defaultPreference);
+    updateDefaultSettings(!isDefaultSettingsSelected);
 
   return (
     <Box
@@ -56,13 +56,13 @@ const AdvancedGasFeeDefaults = () => {
       className="advanced-gas-fee-defaults"
     >
       <CheckBox
-        checked={defaultPreference}
+        checked={isDefaultSettingsSelected}
         className="advanced-gas-fee-defaults__checkbox"
         onClick={handleUpdateDefaultSettings}
         disabled={hasErrors}
       />
       <Typography variant={TYPOGRAPHY.H7} color={COLORS.UI4} margin={0}>
-        {!defaultPreference && Boolean(advancedGasFeeValues) ? (
+        {!isDefaultSettingsSelected && Boolean(advancedGasFeeValues) ? (
           <I18nValue
             messageKey="advancedGasFeeDefaultOptIn"
             options={[
--- a/ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
+++ b/ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
@@ -37,11 +37,13 @@ const AdvancedGasFeeDefaults = () => {
       dispatch(setAdvancedGasFee(null));
     }
   };
-  const handleUpdateDefaultSettings = () => updateDefaultSettings(!defaultPreference);
   const defaultPreference =
     Boolean(advancedGasFeeValues) &&
     advancedGasFeeValues.maxBaseFee === baseFeeMultiplier &&
     advancedGasFeeValues.priorityFee === maxPriorityFeePerGas;
+
+  const handleUpdateDefaultSettings = () =>
+    updateDefaultSettings(!defaultPreference);
 
   return (
     <Box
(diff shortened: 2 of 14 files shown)


ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-inputs/base-fee-input/base-fee-input.js
```diff
@@ -116,7 +127,24 @@ const BaseFeeInput = () => {
 
   useEffect(() => {
     setMaxFeePerGas(maxBaseFeeGWEI);
-  }, [maxBaseFeeGWEI, setMaxFeePerGas]);
+    setMaxBaseFee(maxBaseFeeMultiplier);
```
jpuri, line 130: I think default should not be updated as input value change, but rather if input value change we deselect the checkbox and user select it again to update defaults. Updating defaults should happen only from `advanced-gas-fee-defaults.js`.

My understanding is:
- checkbox should be selected if value in inputs is same as defaults
- if user de-select it we remove default
- if user changes input values checkbox is de-delected and as user selects it again defaults are updated

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-inputs/priority-fee-input/priority-fee-input.js
```diff
@@ -46,7 +53,21 @@ const PriorityFeeInput = () => {
...
     setMaxPriorityFeePerGas(priorityFee);
-  }, [priorityFee, setMaxPriorityFeePerGas]);
+    if (isAdvancedGasFeeDefault) {
+      dispatch(
+        setAdvancedGasFee({
+          ...advancedGasFeeValues,
+          priorityFee,
+        }),
+      );
+    }
```
jpuri, line 63: Also here, lets get rid of this dispatch

ui/components/app/advanced-gas-fee-popover/context/advanceGasFeePopover.js
```diff
@@ -7,16 +7,19 @@ export const AdvanceGasFeePopoverContextProvider = ({ children }) => {
   const [maxFeePerGas, setMaxFeePerGas] = useState();
   const [maxPriorityFeePerGas, setMaxPriorityFeePerGas] = useState();
   const [isDirty, setDirty] = useState();
+  const [maxBaseFee, setMaxBaseFee] = useState();
```
jpuri, line 10: Something line `maxBaseFeeMultiplier` will be more intuitive

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-inputs/base-fee-input/base-fee-input.js
NiranjanaBinoy: This was my initial understanding too, but what if the user assumes the new values will be saved if the default is selected? I will confirm this with the design team.

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
```diff
@@ -0,0 +1,76 @@
...
+  DISPLAY,
+  FLEX_DIRECTION,
+  TYPOGRAPHY,
+} from '../../../../helpers/constants/design-system';
+import { useAdvanceGasFeePopoverContext } from '../context';
+import {
+  getIsAdvancedGasFeeDefault,
+  getAdvancedGasFeeValues,
+} from '../../../../selectors';
+import { setAdvancedGasFee } from '../../../../store/actions';
```
jpuri, line 18: Nit: imports can be organised better.

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
```diff
@@ -0,0 +1,76 @@
...
+    maxPriorityFeePerGas,
+  } = useAdvanceGasFeePopoverContext();
+  const isAdvancedGasFeeDefault = useSelector(getIsAdvancedGasFeeDefault);
+  const advancedGasFeeValues = useSelector(getAdvancedGasFeeValues);
+  console.log(
+    isAdvancedGasFeeDefault,
+    advancedGasFeeValues,
+    baseFeeMultiplier,
+    maxPriorityFeePerGas,
+  );
```
jpuri, line 34: 🔪 console.log

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
```diff
@@ -0,0 +1,76 @@
...
+            maxBaseFee: baseFeeMultiplier,
+            priorityFee: maxPriorityFeePerGas,
+          }),
+        );
+      } else {
+        dispatch(setAdvancedGasFee(null));
+      }
+    },
+    [baseFeeMultiplier, maxPriorityFeePerGas, setDefaultValues, dispatch],
+  );
```
jpuri, line 52: `useCallback` is not needed here I think

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
```diff
@@ -0,0 +1,76 @@
...
+        className="advanced-gas-fee-defaults__checkbox"
+        onClick={() => updateDefaultSettings(!defaultValues)}
+      />
+      <Typography variant={TYPOGRAPHY.H7} color={COLORS.UI4}>
+        {defaultValues ? (
+          <I18nValue messageKey="advancedGassFeeDefaultOptOut" />
+        ) : (
+          <I18nValue messageKey="advancedGassFeeDefaultOptIn" />
+        )}
+      </Typography>
```
jpuri, line 71: Here we do not need `defaultValues` in state but chekbox should be checked if these values are same
```
const {
    baseFeeMultiplier,
    maxPriorityFeePerGas,
  } = useAdvanceGasFeePopoverContext();
  const isAdvancedGasFeeDefault = useSelector(getIsAdvancedGasFeeDefault);
```

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
```diff
@@ -0,0 +1,76 @@
...
+
+const AdvancedGasFeeDefaults = () => {
+  const dispatch = useDispatch();
+
+  const {
+    baseFeeMultiplier,
+    maxPriorityFeePerGas,
+  } = useAdvanceGasFeePopoverContext();
+  const isAdvancedGasFeeDefault = useSelector(getIsAdvancedGasFeeDefault);
+  const advancedGasFeeValues = useSelector(getAdvancedGasFeeValues);
```
jpuri, line 28: `const advancedGasFeeValues = useSelector(getAdvancedGasFeeValues);` looks like not needed

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
jpuri: Also 
```
<I18nValue messageKey={condition ?  'advancedGassFeeDefaultOptOut' : 'advancedGassFeeDefaultOptIn'/>
```

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.test.js
```diff
@@ -0,0 +1,72 @@
...
+describe('AdvancedGasFeeDefaults', () => {
+  it('should renders correct message when the default is not set', () => {
+    render({ advancedGasFee: null });
+
+    expect(
+      screen.queryByText(
+        'Use the “new values” and advanced setting as default.',
+      ),
+    ).toBeInTheDocument();
+  });
```
jpuri, line 62: Here we should set default redux state (in actual execution state is updated from preference controller)

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-popover.js
```diff
@@ -28,8 +29,10 @@ const AdvancedGasFeePopover = () => {
         onClose={closeAllModals}
         footer={<AdvancedGasFeeSaveButton />}
       >
-        <Box className="advanced-gas-fee-popover__wrapper">
+        <Box className="advanced-gas-fee-popover__wrapper" margin={4}>
           <AdvancedGasFeeInputs />
+          <div className="advanced-gas-fee-popover__separator" />
+          <AdvancedGasFeeDefaults />
```
jpuri, line 35: I am not sure if this margin is really needed, inputs already have margins. Can you plz check margins are not doubled:
Screenshot 2021-12-03 at 12 11 25 PM

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
NiranjanaBinoy: I did some re-arrangements, if you have a different pattern/order in mind please let me know.

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
NiranjanaBinoy: With the recent changes, we are curremtly using the advancedGasFeeValues

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-popover.js
NiranjanaBinoy: I rebased and pulled in your changes for gasLimit, so now there is a separator between inputs and default and one between default and gasLimi
Screen Shot 2021-12-03 at 4 46 45 PM
t.

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
jpuri: we really do not need these 2 selectors: getAdvancedGasFeeValues, getIsAdvancedGasFeeDefault

`getAdvancedGasFeeValues` should do

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-popover.js
jpuri: 👍 , can you plz also ensure that margins are as in designs, in above image left and right margin should be 16 px.

jpuri (dismissed):
LGTM

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-popover.js
NiranjanaBinoy: Image I shared in the above comment is from the design figma. This is the current implementation .

Screen Shot 2021-12-06 at 3 14 31 PM

jpuri (dismissed):

app/_locales/en/messages.json
```diff
@@ -169,6 +169,12 @@
   "advancedBaseGasFeeToolTip": {
     "message": "When your transaction gets included in the block, any difference between your max base fee and the actual base fee will be refunded. Total amount is calculated as max base fee (in GWEI) * gas limit."
   },
+  "advancedGasFeeDefaultOptIn": {
+    "message": "Use the “new values” and advanced setting as default."
```
adonesky1, line 173: I'm not seeing this copy the designs? And personally I think it's a bit confusing. Maybe something like?: 
```suggestion
    "message": "Set the currently selected values as defaults when using advanced settings."
```

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
```diff
@@ -0,0 +1,70 @@
...
+    <Box
+      display={DISPLAY.FLEX}
+      flexDirection={FLEX_DIRECTION.ROW}
+      marginRight={4}
+      className="advanced-gas-fee-defaults"
+    >
+      <CheckBox
+        checked={defaultPreference}
+        className="advanced-gas-fee-defaults__checkbox"
+        onClick={() => updateDefaultSettings(!defaultPreference)}
```
georgewrmarshall, line 55: Not sure what our stance is on inline arrow functions but I would suggest moving this to a function outside of the return method.

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/index.scss
```diff
@@ -0,0 +1,6 @@
+.advanced-gas-fee-defaults {
+  & &__checkbox {
+    font-size: $font-size-h4;
```
georgewrmarshall, line 3: Is this font-size used for the fontawesome check icon?

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-inputs/index.scss
```diff
@@ -9,6 +9,6 @@
 
   &__separator {
     border-top: 1px solid $ui-grey;
-    margin: 24px 0 16px 0;
+    margin: 20px 0;
```
georgewrmarshall, line 12: Not sure exactly what this spacing effects but would suggest adhering to `8px` grid and changing to either `24px` or `16px`
```suggestion
    margin: 16px 0;
```

ui/components/app/advanced-gas-fee-popover/index.scss
```diff
@@ -1,14 +1,21 @@
 .advanced-gas-fee-popover {
-  &__wrapper {
-    border-top: 1px solid $ui-grey;
-  }
-
   &__separator {
     border-top: 1px solid $ui-grey;
-    margin: 24px 0 16px 0;
+    margin: 20px 0;
```
georgewrmarshall, line 4: 8px grid
```suggestion
    margin: 16px 0;
```

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
```diff
@@ -0,0 +1,70 @@
...
+      />
+      <Typography variant={TYPOGRAPHY.H7} color={COLORS.UI4}>
```
georgewrmarshall, line 57: Add margin 0 to the Typography component here so it aligns better. I think a default margin being added to the Typography component was a mistake. Going to look into how we can resolve this soon

```suggestion
      <Typography variant={TYPOGRAPHY.H7} color={COLORS.UI4} boxProps={{margin: 0}}>
```

Before
Screen Shot 2021-12-07 at 2 52 50 PM

After
Screen Shot 2021-12-07 at 2 50 38 PM

georgewrmarshall (commented):
Excellent PR description and use of design system components 💯 . Couple small suggestions. Excellent work!!

app/_locales/en/messages.json
NiranjanaBinoy: The last image for the advanced gas modal in this [Frame](https://www.figma.com/file/DFlkrM9JRyP0hDrd4ykMZW/1559-Gas-Controls---v2.1?node-id=610%3A7104) shows that copy. It is displayed when the checkbox is not selected. 

@jakehaugen Can I get your input on updating the copy?

adonesky1:
Hey Niranjana so testing this I'm seeing some odd behavior:
https://user-images.githubusercontent.com/34557516/145064938-ecaa5155-8cdb-4739-9a53-056d14f65733.mov

1) The save button is disabled without any errors being shown (maybe this is because the values match the aggressive values? If so I think we should indicate that's the reason why `Save` is disabled. @jakehaugen thoughts on this?)
2) The `Priority fee must be at least 1 GWEI` error is not disabling `Save` as it should.
3) But I am allowed to set the 0 priority fee as a default. It seems the checkbox should be disabled when a true blocking error is present like Priority fee set to 0.
4) It doesn't even appear to save the defaults I set when I switch to a different mode (aggressive) and then switch back to advanced.
5) And this is more of a general UX thing that I believe is on develop right now, but maybe can be addressed in this PR. I think when a user first clicks into the advanced gas fee modal we should show the Max Base fee in GWEI and not the multiplier (as is happening here in the video). @jakehaugen thoughts on this as well?

Sorry to pile on here! Happy to take a look together and discuss further!

jpuri:
> The save button is disabled without any errors being shown (maybe this is because the values match the aggressive values? If so I think we should indicate that's the reason why Save is disabled. @jakehaugen thoughts on this?)

Save button is disabled unless a change is made to any input

> The Priority fee must be at least 1 GWEI error is not disabling Save as it should.

I will fix this

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
```diff
@@ -0,0 +1,73 @@
...
+const AdvancedGasFeeDefaults = () => {
+  const dispatch = useDispatch();
+
+  const {
+    baseFeeMultiplier,
+    maxPriorityFeePerGas,
+  } = useAdvancedGasFeePopoverContext();
+  const advancedGasFeeValues = useSelector(getAdvancedGasFeeValues);
+
+  const updateDefaultSettings = (value) => {
```
mcmire, line 28: I'm wondering if it makes more sense to break this function into two functions? It seems that this does two distinct things: update the default settings (based on whatever the current value of `baseFeeMultiplier` and `maxPriorityFeePerGas` is) and clear the default settings.

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
```diff
@@ -0,0 +1,73 @@
...
+} from '../../../../helpers/constants/design-system';
+import { getAdvancedGasFeeValues } from '../../../../selectors';
+import { setAdvancedGasFee } from '../../../../store/actions';
+
+import { useAdvancedGasFeePopoverContext } from '../context';
+
+const AdvancedGasFeeDefaults = () => {
+  const dispatch = useDispatch();
+
+  const {
```
mcmire, line 22: What is the difference between the advanced settings which are kept in the context and the advanced settings which are kept in Redux? Is the difference that Redux represents the saved version and the context represents the changed-but-unsaved version? If that's the case, then I'm wondering whether it makes more sense to include that in the name somewhere? Maybe we have `unsavedBaseFeeMultiplier` and `unsavedMaxPriorityFeePerGas` vs. `savedBaseFeeMultiplier` and `savedMaxPriorityFeePerGas`. I feel like these names could get confusing.

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
```diff
@@ -0,0 +1,73 @@
...
+        setAdvancedGasFee({
+          maxBaseFee: baseFeeMultiplier,
+          priorityFee: maxPriorityFeePerGas,
+        }),
+      );
+    } else {
+      dispatch(setAdvancedGasFee(null));
+    }
+  };
+  const defaultPreference =
```
mcmire, line 40: Is there another name for this? This sounds like this returns whatever preference the user chose, yet it's a boolean. What does this represent? Whether or not the defaults should be saved or cleared?

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
```diff
@@ -0,0 +1,73 @@
...
+    Boolean(advancedGasFeeValues) &&
+    advancedGasFeeValues.maxBaseFee === baseFeeMultiplier &&
+    advancedGasFeeValues.priorityFee === maxPriorityFeePerGas;
+
+  const handleUpdateDefaultSettings = () =>
+    updateDefaultSettings(!defaultPreference);
```
mcmire, line 46: What is the difference between "default settings" and "preference"?

mcmire (commented):
This looks good except for some naming questions.

georgewrmarshall (commented):
UI is looking good!

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
NiranjanaBinoy: The function basically just sets and clears the `advancedGasFee` object when the checkbox is checked and unchecked.

mcmire (dismissed):
I'm good with this. There are some naming tweaks I think we could to improve readability, but I guess given that this isn't a lot of code, in the context of the file the purpose of each thing is easy enough to understand.

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
NiranjanaBinoy: nothing much, it's the same thing I think, I will update the code to avoid this confusion.

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
NiranjanaBinoy: updated the name

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
NiranjanaBinoy: I was just following the existing pattern in the`AdvancedGasFeePopoverContextProvider` changing them might end up with unrelated changes to this Pr, if necessary we can have a follow-up PR!!

mcmire (dismissed):

mcmire (dismissed):

georgewrmarshall (commented):
UI looks great! Fantastic work. My one comment is that I found the inputs quite finicky as you can't fully clear the input. As a user I was expecting to be able to fully clear the input and then enter my values. Instead it will always start with a 0.


https://user-images.githubusercontent.com/8112138/145890666-ef70ba23-ac13-471b-a0a7-5bae789898a8.mov

NiranjanaBinoy:
@georgewrmarshall  This is because in `onChange` of `<NumericInput/>` component used in `<FormField/>` while the input is numeric. If the `newValue` updated is empty, 0 is set to the value property `onChange?.(parseFloat(newValue || 0, 10));` . I am worried that changing that might result in some unrelated changes in this PR. But I am open to suggestions.

> UI looks great! Fantastic work. My one comment is that I found the inputs quite finicky as you can't fully clear the input. As a user I was expecting to be able to fully clear the input and then enter my values. Instead it will always start with a 0.

georgewrmarshall:
@NiranjanaBinoy that makes sense maybe it can be an improvement if we find that users are having trouble with it.

app/_locales/en/messages.json
```diff
@@ -1832,6 +1838,9 @@
   "newTransactionFee": {
     "message": "New Transaction Fee"
   },
+  "newValues": {
+    "message": "new values"
+  },
```
jpuri, line 1843: May be `advancedGasFeeNewValues `

ui/components/app/advanced-gas-fee-popover/index.scss
```diff
@@ -1,14 +1,21 @@
...
+  .popover-header {
+    border-radius: 0;
+    border-top-left-radius: 10px;
+    border-top-right-radius: 10px;
+    border-bottom: 1px solid $ui-grey;
+  }
+
+  .popover-footer {
+    border-top: none;
+  }
```
jpuri, line 20: I would prefer to have advance gas modal aligned to edit gas modal

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
```diff
@@ -0,0 +1,80 @@
...
+      dispatch(setAdvancedGasFee(null));
+    }
+  };
+  const isDefaultSettingsSelected =
+    Boolean(advancedGasFeeValues) &&
+    advancedGasFeeValues.maxBaseFee === baseFeeMultiplier &&
+    advancedGasFeeValues.priorityFee === maxPriorityFeePerGas;
+
+  const handleUpdateDefaultSettings = () =>
+    updateDefaultSettings(!isDefaultSettingsSelected);
```
jpuri, line 49: It appears that `!isDefaultSettingsSelected` is not required to be passed, it can be made available in `handleUpdateDefaultSettings` also.

app/_locales/en/messages.json
NiranjanaBinoy: If for some scenario for resuing `new values` occur in the future, won't it be better for the copy to have a neutral name?

ui/components/app/advanced-gas-fee-popover/index.scss
NiranjanaBinoy: But this aligns the Advanced gas modal with the design provided by the designers, QA might also be matching the implementation with the design provided.

mcmire:
I'm not completely sold on this user experience. I think it might be unsettling for users if the text for the checkbox changed after the checkbox was checked. Why do we want to change the text to begin with? If we're trying to communicate that changes to base and priority fee will overwrite any preferences you've previously saved, then I think that the text should change after you make a change to the input, not when you check the checkbox.

It also doesn't seem right that the base and priority fee are saved in preferences immediately after checking the checkbox. It seems that this would be done after pressing Save. In other words, if you check the box but don't press Save, then when you come back you'd see the old values.

Thoughts/comments @jakehaugen?

NiranjanaBinoy:
@mcmire Regarding the first part of your message, currently we are changing the copy/text only when the user already has saved default values and then try to change any of the base fee or priority fee. The screen recording g attached with this PR is an old one, I forgot to update that after that change. 

https://user-images.githubusercontent.com/43930900/146818257-090678fa-c5f8-49e9-93d4-c52a02c190fe.mov


> I'm not completely sold on this user experience. I think it might be unsettling for users if the text for the checkbox changed after the checkbox was checked. Why do we want to change the text to begin with? If we're trying to communicate that changes to base and priority fee will overwrite any preferences you've previously saved, then I think that the text should change after you make a change to the input, not when you check the checkbox.

jakehaugen:
@NiranjanaBinoy, yes what is shown in this video is the correct functionality. @mcmire, does that clear anything up for you?

I also have not been 100% sold on this interaction though. I am less concerned about it since it's an "advanced" feature. But, I do think keeping an eye on user feedback and potentially doing more user testing is a good idea.

mcmire:
@NiranjanaBinoy Okay thanks for the updated video. That's a lot better!

@jakehaugen Thanks for the response, sounds good to me.

mcmire (dismissed):

mcmire (approved):

darkwing (approved):

PR #12938, opened 2021-12-01
Enable Lavamoat for Home screen

kumavis:
part of incremental introduction of lavamoat

Adds lavamoat protections to the home screen (does not include popup)


darkwing (approved):
Home screen working for me!

mcmire:
@kumavis Just to clarify, to test this, pulling up the extension in full screen view after entering a password (or going back to the main screen if you're already authenticated) should work?

mcmire:
@kumavis We still want to do this?

kumavis:
@mcmire correct, this only affects the full screen. pr needs to merge with main and have `yarn lavamoat:auto` run to be updated

kumavis:
the next step is #12939 which can be separate or combined

PR #12954, opened 2021-12-02
Remove redundant `EIP_1559_V2_ENABLED` comparison

Gudahtt:
A redundant comparison for the `EIP_1559_V2_ENABLED` variable has been removed. This variable is a boolean, so it can be treated as a boolean directly.

One such comparison has been preserved for the purpose of allowing unit testing, because `process.env` entries are always strings in Node.js. A comment was added to explain this.

Manual testing steps:  
  - This should have no functional impact


Gudahtt:
It turns out that handling the string case was required for unit tests, because `process.env` properties are always strings in Node.js. They can be booleans when using `loose-envify`, but not in Node.js directly. I'll update this to at least remove the redundant `=== true`

NiranjanaBinoy (approved):

segun (approved):

PR #12951, opened 2021-12-02
Fix `isMainnet` propType error

Gudahtt:
A propType error was showing up during e2e tests with a `testDev` build. It was caused by `process.env.IN_TEST` being treated as a boolean, when in fact it is either the string `'true'` or a boolean.

`IN_TEST` has been updated to always be a boolean. `loose-envify` has no trouble injecting boolean values, so there's no reason to treat this as a string.

<details>
<summary>Here is the propType error:</summary>

```
[469511:469511:1202/124122.394917:INFO:CONSOLE(14112)] "Warning: Failed prop type: Invalid prop `isMainnet` of type `string` supplied to `ImportTokenLink`, expected `boolean`.       in ImportTokenLink (created by AssetList)                                                                                                                                     
    in AssetList (created by Home)                                                                                                                                                
    in div (created by Tabs)                                                                                                                                                      
    in div (created by Tabs)                                                                                                                                                      
    in Tabs (created by Home)                                                                                                                                                         in div (created by Home)                                                                                                                                                          in div (created by Home)                                                                                                                                                          in div (created by Home)                                                                                                                                                          in Home (created by ConnectFunction)                                                                                                                                              in ConnectFunction (created by Context.Consumer)                                                                                                                                  in withRouter(Connect(Home)) (created by Context.Consumer)                                                                                                                        in Route (created by Authenticated)                                                                                                                                               in Authenticated (created by ConnectFunction)                                                                                                                                 
    in ConnectFunction (created by Routes)                                                                                                                                            in Switch (created by Routes)                                                                                                                                                     in div (created by Routes)                                                                                                                                                        in div (created by Routes)                                                                                                                                                        in Routes (created by ConnectFunction)                                                                                                                                            in ConnectFunction (created by Context.Consumer)                                                                                                                                  in withRouter(Connect(Routes)) (created by Index)                                                                                                                                 in LegacyI18nProvider (created by Index)                                                                                                                                          in I18nProvider (created by Index)                                                                                                                                                in LegacyMetaMetricsProvider (created by Index)                                                                                                                               
    in MetaMetricsProvider (created by Index)                                                                                                                                     
    in LegacyMetaMetricsProvider (created by Index)                                                                                                                               
    in MetaMetricsProvider (created by Index)                                            
    in Router (created by HashRouter)                                                                                                                                             
    in HashRouter (created by Index)                                                                                                                                              
    in Provider (created by Index)                                                       
[...]

PR #12955, opened 2021-12-02
Disable React dev tools in `testDev` builds

Gudahtt:
The React dev tools can result in console errors if dev tools is not open during the test. Some of our e2e tests fail if there are any console errors, so these errors break those tests.

`react-devtools` has been completely disabled for `testDev` builds to make debugging e2e tests easier. The React dev tools can still be used from development builds.

Manual testing steps:  
  - Create a `testDev` build (`yarn build testDev`)
  - Open the extension build in a browser, and check that there are no Websocket failure console messages from React dev tools
    - The errors are depicted here: https://github.com/facebook/react/issues/20095
  - Ensure that React dev tools still work for development builds

### DIFF
--- a/development/build/scripts.js
+++ b/development/build/scripts.js
@@ -366,6 +366,7 @@ function createFactoredBuild({
       minify,
       reloadOnChange,
       shouldLintFenceFiles,
+      testing,
     });
 
     // set bundle entries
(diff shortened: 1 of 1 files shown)


Gudahtt:
This depends upon #12951

tmashuang (approved):

danjm (approved):

PR #12956, opened 2021-12-02
Disable `remote-redux-devtools` in non-dev builds

Gudahtt:
`remote-redux-devtools` is now explicitly excluded and disabled in non-dev builds, and in the `testDev` build. This was causing console errors in the `testDev` build during e2e tests, which would cause certain tests to fail.

This was already only supposed to be enabled for development builds, but this library used the `NODE_ENV` environment variable to make that determination. This gives us more control over when it's disabled.

Manual testing steps:  
  - Create a `testDev` build (`yarn build testDev`)
  - Open the extension build in a browser, and check that there are no console errors from the remote Redux dev tools
  - Ensure that the Remote Redux dev tools still work for development builds

### DIFF
--- a/development/build/scripts.js
+++ b/development/build/scripts.js
@@ -625,6 +625,7 @@ function setupBundlerDefaults(
   // Ensure react-devtools are not included in non-dev builds
   if (!devMode || testing) {
     bundlerOpts.manualIgnore.push('react-devtools');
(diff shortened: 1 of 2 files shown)


Gudahtt:
This depends upon #12955

segun (approved):

darkwing (approved):

PR #12962, opened 2021-12-03
Adding featureflags to .metamaskrc.dist file

jpuri:
Adding featureflags to .metamaskrc.dist file


segun (dismissed):

.metamaskrc.dist
```diff
@@ -2,3 +2,5 @@
 PASSWORD=METAMASK PASSWORD
 INFURA_PROJECT_ID=00000000000
 SEGMENT_WRITE_KEY=
+ONBOARDING_V2=1
```
Gudahtt, line 5: Should these default to being enabled? Right now they're disabled by default.

jpuri: in dev mode it will be nice to have these enabled.

Gudahtt: I would have found it surprising personally. Unless you're working on one of these features, you wouldn't have reason to enable them. We're still maintaining the old UI after all, fixing bugs and making other improvements.

jpuri: Make sense, I removed the value to disable the feature flag

Gudahtt (dismissed):
LGTM! Though two other flags that might be worth including are `SWAPS_USE_DEV_APIS` and `COLLECTIBLES_V1`.

darkwing:
New flags added

Gudahtt (approved):

darkwing (approved):

PR #12971, opened 2021-12-03
Storybook updating styles so they don't override fontawesome

georgewrmarshall:
Explanation: Previous storybook styles pr overrides fontawesome fonts so this reverts those changes and uses storybook theme api

Manual testing steps:  
  -  Run `yarn storybook`
  -  See documentation guidelines brand fonts still work and code still formatted
  - Go to component that uses fontawesome e.g. "PermissionsConnect" see fontawesome icons work

**Images**

*Fontawesome icons work*
Screen Shot 2021-12-03 at 12 49 12 PM

*Also removing duplicate knobs warning from console*
Screen Shot 2021-12-03 at 12 53 31 PM

*Docs codeblocks still use the monospace font*
docs

### DIFF
--- a/.storybook/metamask-storybook-theme.js
+++ b/.storybook/metamask-storybook-theme.js
@@ -7,4 +7,8 @@ export default create({
   base: 'light',
   brandTitle: 'MetaMask Storybook',
   brandImage: logo,
+
+  // Typography
+  fontBase: 'Euclid, Roboto, Helvetica, Arial, sans-serif',
+  fontCode: 'Inconsolata, monospace',
 });
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -7,50 +7,21 @@
 
 <style>
   * {
-    --blue-500: #037dd6;
-    --blue-400: #1098fc;
     --gray-pre-bg: #f8f8f8;
-    --font-family-base: Euclid, Roboto, Helvetica, Arial, sans-serif;
     --font-family-monospace: Inconsolata, monospace;
     --font-size-code: 0.875rem;
-    --line-height-code: 19px;
-    --font-size-base: 1rem;
-    --line-height-base: 1.7;
-
-    font-family: var(--font-family-base) !important;
-  }
-
-  h1,
-  h2 {
-    font-weight: bold !important;
-  }
-
-  p,
-  li {
-    font-size: var(--font-size-base) !important;
-    line-height: var(--line-height-base) !important;
   }
 
   .docblock-source {
     background: var(--gray-pre-bg) !important;
   }
 
-  code {
+  .docblock-source code {
     font-family: var(--font-family-monospace) !important;
     font-size: var(--font-size-code) !important;
   }
-  code * {
+  .docblock-source code * {
     font-family: var(--font-family-monospace) !important;
     font-size: var(--font-size-code) !important;
-  }
-
-  a {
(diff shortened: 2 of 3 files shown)


.storybook/metamask-storybook-theme.js
```diff
@@ -7,4 +7,8 @@ export default create({
   base: 'light',
   brandTitle: 'MetaMask Storybook',
   brandImage: logo,
+
+  // Typography
+  fontBase: 'Euclid, Roboto, Helvetica, Arial, sans-serif',
+  fontCode: 'Inconsolata, monospace',
 });
```
georgewrmarshall, line 14: Using [storybook theme api](https://storybook.js.org/docs/react/configure/theming) instead of overwriting styles in preview head.

.storybook/preview-head.html
```diff
@@ -7,50 +7,21 @@
...
-  a {
-    color: var(--blue-500) !important;
-  }
-
-  a:hover,
-  a:active,
-  a:focus {
-    color: var(--blue-400) !important;
-  }
 </style>
```
georgewrmarshall, line 27: Removing styles that could possibly overwrite the UI in the storybook iframes and using storybook theme api instead.

.storybook/preview.js
```diff
@@ -82,5 +85,4 @@ const metamaskDecorator = (story, context) => {
   );
 };
 
-addDecorator(withKnobs);
 addDecorator(metamaskDecorator);
```
georgewrmarshall, line 88: Adding custom theme and removing duplicate adding of knobs addon

hmalik88 (approved):

.storybook/preview.js
Gudahtt: Where else is this decorator added?

georgewrmarshall: [in`main.js`](https://github.com/MetaMask/metamask-extension/blob/develop/.storybook/main.js#L11)

georgewrmarshall: Can verify knobs still work in the [storybook build on the Box component](https://410908-42009758-gh.circle-artifacts.com/0/storybook/index.html?path=/story/ui-components-ui-box-box-stories-js--default-story)
Screen Shot 2021-12-07 at 9 49 01 AM

Gudahtt: That's odd. The docs suggest that you're supposed to add this to the list of addons *and* include the `withKnobs` decorator. Check out the examples: https://github.com/storybookjs/addon-knobs#readme They all use the decorator.

georgewrmarshall: I would tentatively suggest that the docs haven't been updated as [addDecorator has been deprecated in v6](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-addparameters-and-adddecorator) knobs as a whole have been deprecated in favour or controls so I would be looking to convert all our stories as well. It is [also mentioned here](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#new-addon-presets)

Screen Shot 2021-12-07 at 12 22 29 PM

Gudahtt: Ah, great find, thanks! I didn't see that.

Gudahtt (approved):
LGTM!

.storybook/preview.js
georgewrmarshall: Thanks for the thorough review 🙏

PR #12881, opened 2021-11-29
Feature: Transaction Insights

alaahd:
### Explanation:

This PR adds the Transactions Insights feature and supporting functionality. This will give users more information about what is happening when sending a transaction to a contract. The core of the functionality is displayed below:

https://user-images.githubusercontent.com/7499938/144235680-024f9aae-586f-4d79-bebc-ad849961d879.mp4

To achieve this, this PR uses https://github.com/trufflesuite/truffle/tree/develop/packages/decoder to decode transaction details which are retrieved from a codefi api

The above screen capture demonstrates the transaction decoding feature and the use of the new modals for viewing and editing account and nickname information.

In addition, the PR adds the transaction decoding to the transaction details modal that is opened when clicking list items in the transaction activity list. This addition also updates the styles/layouts of that modal.

### Manual testing steps:  

This can be done on the Kovan network.

1.   Create a transaction that requires interaction with a contract. Uniswap is a good example
2.   Inspect the "Data" tab to see the transaction details 
3.   Click an address in the transaction details and edit the nickname
4.  Confirm the transaction
5. Click the transaction in the activity list
6. View the transaction information
7. Click "Transaction Data"
8. See the Transaction Data including the nickname you previously edited.

### Screenshots and videos:

https://user-images.githubusercontent.com/7499938/144238081-6a3a2861-25f1-4afa-8ffb-592b1591f087.mp4

PR #12972, opened 2021-12-03
Fix tx insights

alaahd:
Fixes addressing some styling and cosmetic changes related to transaction insights feature and addressbook nicknames based on Dan Miller code review and comments

**CSS**
- classname should be resolvable without having to specify .disclosure as a child of _data, could require the addition of a different class. https://github.com/MetaMask/metamask-extension/pull/12881/files#diff-37b35ddf3d088e4a3707b8fbc05c06ae8aeba47b1ac77905359865d3760231a8R24
- fix usage for optional classNames to comply with standard implementation https://github.com/MetaMask/metamask-extension/pull/12881/files#diff-c85f241de885a39d520383b6f4b7b713b58d7bc333201e3f39584e31f0dc46a4R20
- cleanup the styling and the markup of the disclosure ui component.
- ideally css shouldn't reference classes of other components, instead the other component should be configurable and manage its own css under that configuration https://github.com/MetaMask/metamask-extension/pull/12881/files#diff-503fa60e87628a2c02738c80d14b3fe666a3bc8574eed53af6a41a97f8cf9e62R6
- adding another class, using BEM style, instead of child selectors: https://github.com/MetaMask/metamask-extension/pull/12881/files#diff-f7d836e9f6f1a3588f842b9ee5a55e3f3d282672d2ddf8c3e7cc70eac8fceeccR6

**JavaScript**
- hide activity log in the transaction history view if activity log is empty.
- move fetchWithCache calls to a helper function https://github.com/MetaMask/metamask-extension/pull/12881/files#diff-0d33bff56e4fdad212852eb6845d22ea07ca0ab70437ff224e8489df4e92d35eR37-R58

**Tests**
- add unit test to replace this one that was removed https://github.com/MetaMask/metamask-extension/pull/12881/files#diff-d4ab175dc9f8c76a2bafdfc82421ebe30922f7dddd3b061150ab16e8c0c449a3
- add some unit tests for the transaction decoding util


brad-decker:
@alaahd thank you for your contribution. If you'd like us to consider this for inclusion I would prefer if you were to break up these changes into a few PRs.

PR #9257, opened 2020-08-18
Add sort and search to AddRecipient accounts list

PatrykLucka:
Adds `internalSearch` option to ENS Input component, which allows to disable validation and use input for custom purpose - in this case as a search input for my accounts list.

Also adds sorting for my accounts list.
Fixes #9102 

Screenshot 2020-08-18 at 12 57 08
Screenshot 2020-08-18 at 12 57 15

PR #12981, opened 2021-12-06
update order of accounts list

### LINKED ISSUE #12960: Update the order of account list on the "Send to" screen to match the order from the dropdown list
**User Story**
This issue is a result of this bug report: #12885 where the order of accounts listed doesn't match the user expectation. 

**Steps to reproduce the bug**
Create at least 10 accounts with each respective name being "Account 1", "Account 2", "Account 3", ... , "Account 10"
Click "Send"
Click "Transfer between my accounts"
See mis-ordered accounts (e.g. "Account 10" shows up before "Account 2").

**Task**
Update the order of the account list on the "Send to" screen (1st screenshot) to match the same order from the dropdown list (2nd screenshot)

Send to screen:
Screen Shot 2021-12-02 at 19 38 25

Dropdown list:
Screen Shot 2021-12-02 at 19 38 04

**Acceptance criteria:**
- [ ] Accounts on the "Send to" screen are listed in the same order as the dropdown list
- [ ] Changing account's names doesn't change the order they are shown on the "Send to" screen
- [ ] There are no changes made to the order of accounts on the dropdown list

hsy822:
Fixes: #12960 

Explanation:  I removed some of the array sort code so that the accounts come out in order. Dowe need a sort code?

`
  const ownedAccounts = accountsWithSendEtherInfoSelector(state).sort((a, b) =>
    a.name.localeCompare(b.name),
  );
`

### DIFF
--- a/ui/pages/send/send-content/add-recipient/add-recipient.container.js
+++ b/ui/pages/send/send-content/add-recipient/add-recipient.container.js
@@ -34,9 +34,7 @@ function mapStateToProps(state) {
 
   const addressBook = getAddressBook(state);
 
-  const ownedAccounts = accountsWithSendEtherInfoSelector(state).sort((a, b) =>
-    a.name.localeCompare(b.name),
-  );
+  const ownedAccounts = accountsWithSendEtherInfoSelector(state);
 
   return {
     addressBook,


Gudahtt:
This was added as part of #9257, supposedly because beforehand the accounts were listed in a jumbled order.

It doesn't look like `accountsWithSendEtherInfoSelector` on its own guarantees the correct order is used here. I agree that the way this is being sorted is wrong, but we should at least ensure the order here matches the account order in the keyring.

hsy822:
> Thanks for the contribution!
> 
> This was added as part of #9257, supposedly because beforehand the accounts were listed in a jumbled order.
> 
> It doesn't look like `accountsWithSendEtherInfoSelector` on its own guarantees the correct order is used here. I agree that the way this is being sorted is wrong, but we should at least ensure the order here matches the account order in the keyring.
> 
> Can you verify that this returns accounts in numeric order? Even after some accounts are renamed?

Thank you for the reply. 

From what I've checked, this approach( #12999 ) seems more appropriate. I will close this PR.

PR #12983, opened 2021-12-06
Fix known token contract address warning

### LINKED ISSUE #12642: Warning not being shown when sending tokens to a known token contract address
In `ducks/send/send.js` we have code that is meant to show a warning to users when they are sending to a known contract address:

```
if (
          isSendingToken &&
          isValidHexAddress(recipient.userInput) &&
          (tokenAddressList.find((address) =>
            isEqualCaseInsensitive(address, recipient.userInput),
          ) ||
            checkExistingAddresses(recipient.userInput, tokens))
        ) {
          recipient.warning = KNOWN_RECIPIENT_ADDRESS_WARNING;
        } else {
          recipient.warning = null;
 }
```

This `recipient.warning` property is read in `renderDialogs` of `ui/pages/send/send-content/add-recipient/add-recipient.component.js`

Currently, if I try to send DAI token to the DAI contract, no warning is shown.

Acceptance criteria:
- if a known token contract address is entered in the recipient field in the send flow, a warning should be shown to the user
- this warning can just be the `<Dialog type="warning" className="send__error-dialog">` that `add-recipient.component.js` is currently meant to render
- "known token contract addresses" include any of the addresses in this list: https://github.com/MetaMask/contract-metadata/blob/master/contract-map.json

filipsekulic:
Fixes: https://github.com/MetaMask/metamask-extension/issues/12642

**Screencast**:

https://user-images.githubusercontent.com/92310504/152987260-6dee5ed3-9d53-4b62-8c60-6bf17909695a.mov

### DIFF
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -1073,8 +1073,7 @@ const slice = createSlice({
         recipient.warning = null;
       } else {
         const { chainId, tokens, tokenAddressList } = action.payload;
-        const assetAddress =
-          asset.details && asset.details.address ? asset.details.address : '';
+        const assetAddress = asset?.details?.address ?? '';
 
         if (
           isBurnAddress(recipient.userInput) ||
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -1072,10 +1072,10 @@ const slice = createSlice({
         recipient.error = null;
         recipient.warning = null;
       } else {
-        const isSendingToken =
-          asset.type === ASSET_TYPES.TOKEN ||
-          asset.type === ASSET_TYPES.COLLECTIBLE;
         const { chainId, tokens, tokenAddressList } = action.payload;
+        const assetAddress =
+          asset.details && asset.details.address ? asset.details.address : '';
+
         if (
           isBurnAddress(recipient.userInput) ||
           (!isValidHexAddress(recipient.userInput, {
@@ -1086,16 +1086,12 @@ const slice = createSlice({
           recipient.error = isDefaultMetaMaskChain(chainId)
             ? INVALID_RECIPIENT_ADDRESS_ERROR
             : INVALID_RECIPIENT_ADDRESS_NOT_ETH_NETWORK_ERROR;
-        } else if (
-          isSendingToken &&
-          isOriginContractAddress(recipient.userInput, asset.details.address)
-        ) {
+        } else if (isOriginContractAddress(recipient.userInput, assetAddress)) {
           recipient.error = CONTRACT_ADDRESS_ERROR;
         } else {
           recipient.error = null;
         }
         if (
-          isSendingToken &&
           isValidHexAddress(recipient.userInput) &&
           (tokenAddressList.find((address) =>
             isEqualCaseInsensitive(address, recipient.userInput),
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -937,7 +937,7 @@ const slice = createSlice({
     },
     updateRecipient: (state, action) => {
       state.recipient.error = null;
-      state.recipient.userInput = '';
+      state.recipient.userInput = state.recipient.userInput ?? '';
       state.recipient.address = action.payload.address ?? '';
       state.recipient.nickname = action.payload.nickname ?? '';
 
--- a/ui/pages/send/send-content/send-content.component.js
+++ b/ui/pages/send/send-content/send-content.component.js
@@ -10,7 +10,7 @@ import {
   INSUFFICIENT_FUNDS_FOR_GAS_ERROR_KEY,
 } from '../../../helpers/constants/error-keys';
 import { ASSET_TYPES } from '../../../ducks/send';
-import { getSymbolAndDecimals } from '../../../helpers/utils/token-util';
+import { getTokenMetadata } from '../../../helpers/utils/token-util';
 import SendAmountRow from './send-amount-row';
 import SendHexDataRow from './send-hex-data-row';
 import SendAssetRow from './send-asset-row';
@@ -44,6 +44,16 @@ export default class SendContent extends Component {
     tokenAddressList: PropTypes.object,
   };
 
+  componentDidMount() {
+    this.checkContractAddress();
+  }
+
+  componentDidUpdate(prevProps) {
+    if (prevProps.recipient !== this.props.recipient) {
+      this.checkContractAddress();
+    }
+  }
+
   render() {
     const {
       warning,
@@ -70,10 +80,6 @@ export default class SendContent extends Component {
       asset.type !== ASSET_TYPES.TOKEN &&
       asset.type !== ASSET_TYPES.COLLECTIBLE;
 
-    if (!this.state.isKnownContractAddress) {
-      this.checkContractAddress();
-    }
-
     return (
       <PageContainerContent>
         <div className="send-v2__form">
@@ -94,19 +100,15 @@ export default class SendContent extends Component {
     );
   }
 
-  checkContractAddress = async () => {
+  checkContractAddress = () => {
     const { recipient, tokenAddressList } = this.props;
 
-    const symbolAndDecimals = await getSymbolAndDecimals(
+    const tokenMetadata = getTokenMetadata(
       recipient.userInput,
       tokenAddressList,
     );
 
-    if (
-      (symbolAndDecimals.symbol !== undefined ||
-        symbolAndDecimals.decimals !== undefined) &&
-      symbolAndDecimals.symbol !== ''
-    ) {
+    if (tokenMetadata?.symbol !== undefined && tokenMetadata?.symbol !== '') {
       this.setState({ isKnownContractAddress: true });
     }
   };
--- a/ui/pages/send/send-content/send-content.component.js
+++ b/ui/pages/send/send-content/send-content.component.js
@@ -92,7 +92,6 @@ export default class SendContent extends Component {
 
   checkContractAddress = () => {
     const { recipient, tokenAddressList } = this.props;
-
     getSymbolAndDecimals(recipient.userInput, tokenAddressList).then(
       (result) => {
         if (result.symbol !== '' && result.decimals !== '0') {
--- a/ui/pages/send/send-content/send-content.component.js
+++ b/ui/pages/send/send-content/send-content.component.js
@@ -94,16 +94,21 @@ export default class SendContent extends Component {
     );
   }
 
-  checkContractAddress = () => {
+  checkContractAddress = async () => {
     const { recipient, tokenAddressList } = this.props;
 
-    getSymbolAndDecimals(recipient.userInput, tokenAddressList).then(
-      (result) => {
-        if (result.symbol !== undefined || result.decimals !== undefined) {
-          this.setState({ isKnownContractAddress: true });
-        }
-      },
+    const symbolAndDecimals = await getSymbolAndDecimals(
+      recipient.userInput,
+      tokenAddressList,
     );
+
+    if (
+      (symbolAndDecimals.symbol !== undefined ||
+        symbolAndDecimals.decimals !== undefined) &&
+      symbolAndDecimals.symbol !== ''
+    ) {
+      this.setState({ isKnownContractAddress: true });
+    }
   };
 
   maybeRenderAddContact() {
--- a/ui/pages/send/send-content/send-content.component.js
+++ b/ui/pages/send/send-content/send-content.component.js
@@ -95,7 +95,7 @@ export default class SendContent extends Component {
 
     getSymbolAndDecimals(recipient.userInput, tokenAddressList).then(
       (result) => {
-        if (result.symbol !== '' && result.decimals !== '0') {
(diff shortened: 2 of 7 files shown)


mirjanaKukic:
Verified by QA.

david0xd (dismissed):

ui/pages/send/send-content/send-content.component.js
```diff
@@ -19,6 +20,7 @@ import SendGasRow from './send-gas-row';
 export default class SendContent extends Component {
   state = {
     showNicknamePopovers: false,
+    isContractAddress: false,
```
adonesky1, line 23: I think for clarity we should call this `isKnownContractAddress`

ui/pages/send/send-content/send-content.component.js
```diff
@@ -84,24 +88,52 @@ export default class SendContent extends Component {
     );
   }
 
+  checkContractAddress = () => {
+    const { recipient, tokenAddressList } = this.props;
+
+    getSymbolAndDecimals(recipient.userInput, tokenAddressList).then(
+      (result) => {
+        if (
+          (result.symbol !== undefined && result.symbol !== '') ||
```
adonesky1, line 97: do we require both conditions here? Wouldn't  `result.symbol !== undefined || result.decimals !== undefined` be sufficient to establish that its on the list and what we would consider a know token/contract address?

ui/pages/send/send-content/send-content.component.js
```diff
@@ -84,24 +88,52 @@ export default class SendContent extends Component {
...
+  };
+
   maybeRenderAddContact() {
     const { t } = this.context;
-    const { isOwnedAccount, contact = {}, to } = this.props;
-    const { showNicknamePopovers } = this.state;
+    const { isOwnedAccount, contact = {}, to, recipient } = this.props;
+    const { showNicknamePopovers, isContractAddress } = this.state;
+
+    this.checkContractAddress();
```
adonesky1, line 111: any reason this shouldn't just be called in the `render` method?

ui/pages/send/send-content/send-content.component.js
```diff
@@ -84,24 +88,52 @@ export default class SendContent extends Component {
...
+          {t('knownAddressRecipient')}
+        </Dialog>
+      ) : null;
     }
 
     return (
       <>
+        {recipient.warning === 'knownAddressRecipient' || isContractAddress ? (
+          <Dialog type="warning" className="send__error-dialog">
+            {t('knownAddressRecipient')}
```
adonesky1, line 126: Wondering if we should rename this key to `knownContractAddressWarning`? cc @danjm

adonesky1 (commented):
A few small questions

ui/pages/send/send-content/send-content.component.js
filipsekulic: Fixed in: ad39793344c75085bf0a75435319383fa7f7b514

ui/pages/send/send-content/send-content.component.js
filipsekulic: You have got the point.
Fixed in: ad39793344c75085bf0a75435319383fa7f7b514

ui/pages/send/send-content/send-content.component.js
filipsekulic: Fixed in: ad39793344c75085bf0a75435319383fa7f7b514

ui/pages/send/send-content/send-content.component.js
filipsekulic: Did not change anything. Waiting for Dan's response on this one.

adonesky1 (dismissed):
LGTM. Still think we should change that key name but won't block on it.

ui/ducks/send/send.js
```diff
@@ -992,8 +992,10 @@ const slice = createSlice({
         recipient.error = null;
         recipient.warning = null;
       } else {
-        const isSendingToken = asset.type === ASSET_TYPES.TOKEN;
         const { chainId, tokens, tokenAddressList } = action.payload;
+        const assetAddress =
+          asset.details && asset.details.address ? asset.details.address : '';
```
darkwing, line 997: Can this be `asset?.details?.address ?? ''`?

ui/pages/send/send-content/send-content.component.js
```diff
@@ -84,24 +92,48 @@ export default class SendContent extends Component {
     );
   }
 
+  checkContractAddress = () => {
```
darkwing, line 95: Can we make this method async so we can use `async/await` instead of `then`?

ui/ducks/send/send.js
filipsekulic: Yes, it can be. 
Fixed in: 51054012cfbb3e51b2853adafa9e9d8fd155656d

ui/pages/send/send-content/send-content.component.js
filipsekulic: Yes, we can. I changed it.
Fixed in: 51054012cfbb3e51b2853adafa9e9d8fd155656d

ui/pages/send/send-content/send-content.component.js
filipsekulic: Actually, I had to modify this condition. The reason for doing so is the fact that entering some user's address gives the result: 
{ 
  symbol: '',
  decimals: '0'
}.
If the condition was just `result.symbol !== undefined || result.decimals !== undefined`, it would pass. 
Concerning that, I added another check, so the final condition looks like `if ((symbolAndDecimals.symbol !== undefined || symbolAndDecimals.decimals !== undefined) && symbolAndDecimals.symbol !== '')`.
If the address is contract, it will have the symbol as well.

ui/pages/send/send-content/send-content.component.js
danjm: Let's not change it right now, because then we have to update all of the translations too

danjm (dismissed):
LGTM

ui/ducks/send/send.js
```diff
@@ -1054,10 +1054,9 @@ const slice = createSlice({
         recipient.error = null;
         recipient.warning = null;
       } else {
-        const isSendingToken =
-          asset.type === ASSET_TYPES.TOKEN ||
-          asset.type === ASSET_TYPES.COLLECTIBLE;
```
filipsekulic, line 1059: @adonesky1 
I removed this piece of code as the `isSendingToken` is not necessary here for my PR. In order to make sending ETH possible,`|| asset.type ===  ASSET_TYPES.NATIVE` should be added or `isSendingToken` removed from here and down there. 
ASSET_TYPES looks like this:
`export const ASSET_TYPES = {
  NATIVE: 'NATIVE',
  TOKEN: 'TOKEN',
  COLLECTIBLE: 'COLLECTIBLE',
};
`
In this case, if another condition was added, `isSendingToken` would always be true.  
Agreed with Dan about this one.
// cc @danjm

adonesky1: Looks good to me!

adonesky1: Thanks for pointing this out though!

ui/pages/send/send-content/send-content.component.js
```diff
@@ -83,24 +91,53 @@ export default class SendContent extends Component {
...
     const { t } = this.context;
-    const { isOwnedAccount, contact = {}, to } = this.props;
-    const { showNicknamePopovers } = this.state;
+    const { isOwnedAccount, contact = {}, to, recipient } = this.props;
+    const { showNicknamePopovers, isKnownContractAddress } = this.state;
 
     if (isOwnedAccount || contact.name) {
-      return null;
+      return recipient.warning === 'knownAddressRecipient' ||
+        isKnownContractAddress ? (
```
adonesky1, line 118: These knownContract addresses are only on Mainnet so should we only show this warning when we're on Ethereum mainnet? cc @danjm

ui/pages/send/send-content/send-content.component.js
```diff
@@ -83,24 +91,53 @@ export default class SendContent extends Component {
     );
   }
 
+  checkContractAddress = async () => {
+    const { recipient, tokenAddressList } = this.props;
+
+    const symbolAndDecimals = await getSymbolAndDecimals(
```
adonesky1, line 97: Once controllers v25.0.0 is added to the extension we'll have a way check whether the address is an ERC721 or ERC1155 contract as well. Should we wait and add that functionality? I'm happy to pair program and make that change as well. cc @danjm @david0xd

david0xd (dismissed):

ui/pages/send/send-content/send-content.component.js
```diff
@@ -66,6 +70,10 @@ export default class SendContent extends Component {
       asset.type !== ASSET_TYPES.TOKEN &&
       asset.type !== ASSET_TYPES.COLLECTIBLE;
 
+    if (!this.state.isKnownContractAddress) {
```
danjm, line 73: This should be done in a `componentDidUpdate` call. Ideally, the render method does not have state management side effects. Also, with the `componentDidUpdate` method, we can compare `prevProps.recipient` to `this.props.recipient` and only call `this.checkContractAddress` when they are different. This ensures we don't call it more than necessary, and that we do call it if the recipient changes.

ui/pages/send/send-content/send-content.component.js
danjm: I think that can be a separate PR after this PR merges. We should create a separate github ticket for that:

- [ ] Create github ticket for ensuring the known contract address warning in the send flow also apples to NFT tokens

ui/pages/send/send-content/send-content.component.js
```diff
@@ -86,24 +94,53 @@ export default class SendContent extends Component {
     );
   }
 
+  checkContractAddress = async () => {
+    const { recipient, tokenAddressList } = this.props;
+
+    const symbolAndDecimals = await getSymbolAndDecimals(
```
danjm, line 100: @filipsekulic Rereading this now, I realize a made a mistake in previous advice I gave you. We shouldn't call `getSymbolAndDecimals` because that will return symbol and decimals for all contracts that have symbol and decimals APIs implemented, but we only want to show the warning when it is from our token list.

I think if you use the `getTokenMetadata` from the same file, instead of `getSymbolAndDecimals` this will work correctly. With that change, you will also just need to add an optional chaining operator below (e.g. `symbolAndDecimals?.symbol`)

ui/pages/send/send-content/send-content.component.js
danjm: No, I think it is okay. Because if we are not on mainnet, then we definitely don't want the user sending to those token addresses. I think improve warnings for that case are being developed, but this actually has benefit in the meantime.

danjm (changes requested):
@filipsekulic Sorry, I checked this again and noticed a couple of other changes that are needed.

ui/pages/send/send-content/send-content.component.js
filipsekulic: Fixed in: e3b7ee44fa5b49dbce8251dcb856212b1b34eb85

ui/pages/send/send-content/send-content.component.js
filipsekulic: Fixed in: e3b7ee44fa5b49dbce8251dcb856212b1b34eb85

filipsekulic:
> @filipsekulic Sorry, I checked this again and noticed a couple of other changes that are needed.

@danjm No problem, it's okay. I fixed it like we agreed (e3b7ee44fa5b49dbce8251dcb856212b1b34eb85).

mirjanaKukic:
Verified by QA

david0xd (dismissed):

cyanlink:
Sorry for asking too much but IMO the "warning" is still not enough, at least you should say "WARNING: known contract address, Read More.". Then in the read more you redirect to [this zendesk page](https://metamask.zendesk.com/hc/en-us/articles/360020028092-What-is-the-known-contract-address-warning-) or maybe a newer verbose explanation, like my proposal at #13486 . I really do not know how to contact the metamask team in the most efficient way (the community post and new request ticket does not draw any attention).

bschorchit:
I believe this PR doesn't include the acceptance criteria that @EHaracic [mentioned here](https://github.com/MetaMask/metamask-extension/issues/12642#issuecomment-970235647): "Remove "New address detected ..." when this specific warning is shown"

I'm saying that by looking at the video only, so I might be wrong or this might have been already fixed later.

filipsekulic:
> I believe this PR doesn't include the acceptance criteria that @EHaracic [mentioned here](https://github.com/MetaMask/metamask-extension/issues/12642#issuecomment-970235647): "Remove "New address detected ..." when this specific warning is shown"
> 
> I'm saying that by looking at the video only, so I might be wrong or this might have been already fixed later.

@bschorchit You are right. Somehow I forgot to add that AC to Dan's AC. Good catch. Thanks!

danjm:
Nice work here @filipsekulic

Some new requirements have come in and there are some edge case that I realized are only fixable if we modify the send duck more. Also, the advice I gave before on getting the symbols and decimals wasn't correct for all cases. Instead of making you go back on this again, I started making changes on top your commit and opened a new PR. So we can close this and I will finish off the work their.

https://github.com/MetaMask/metamask-extension/pull/13588

PR #12984, opened 2021-12-06
WIP: Snaps custom confirmation UI

### LINKED ISSUE #11686: Snaps Custom Confirmation
We want to allow snaps to show custom confirmations to the user. This will allow snap devs to ask for user consent before taking certain actions, such as signing.

This custom confirmation can be very limited to start with, with just a custom title, maybe a custom subtitle, a custom text area, and "Confirm" / "Reject" buttons (or something similar). All custom text areas should be specified by the snap.

We should make it clear that the confirmation is coming from the snap, not from MetaMask. We don't know what it happens if the user confirms, whether the text is accurate, or how high or low risk the action being confirmed might be. All we know is that it's internal to the snap.

ritave:
fixes #11686
Screenshot 2021-12-06 at 18 34 37


ritave:
Superseded by https://github.com/MetaMask/metamask-extension/pull/13036

PR #12989, opened 2021-12-06
Make restore vault a form so an user can submit via keyboard

rickycodes:
Similar to "import-with-seed-phrase" I would like to be able to restore my vault by pressing `<enter>` on my keyboard.

Explanation: I noticed the other day that you can't restore from vault using keyboard. We do something similar in `import-with-seed-phrase.component.js`. this should be made keyboard accessible as well.

Manual testing steps:  
  - try to restore vault by filling in seedphrase + passwords.
  - when clicking enter nothing happens

### DIFF
--- a/test/e2e/tests/metamask-responsive-ui.spec.js
+++ b/test/e2e/tests/metamask-responsive-ui.spec.js
@@ -178,10 +178,7 @@ describe('Metamask Responsive UI', function () {
 
         await driver.fill('#password', 'correct horse battery staple');
         await driver.fill('#confirm-password', 'correct horse battery staple');
-        await driver.clickElement({
-          text: enLocaleMessages.restore.message,
-          tag: 'button',
-        });
+        await driver.press('#confirm-password', driver.Key.ENTER);
 
         // balance renders
         await driver.waitForSelector({
--- a/ui/pages/keychains/restore-vault.js
+++ b/ui/pages/keychains/restore-vault.js
@@ -85,7 +85,8 @@ class RestoreVaultPage extends Component {
     this.setState({ confirmPassword, confirmPasswordError });
   }
 
-  handleImport = () => {
+  handleImport = (event) => {
+    event.preventDefault();
     const { password, seedPhrase, disabled } = this.state;
     if (disabled) return;
     const {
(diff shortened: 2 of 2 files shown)


test/e2e/tests/metamask-responsive-ui.spec.js
```diff
@@ -178,10 +178,7 @@ describe('Metamask Responsive UI', function () {
 
         await driver.fill('#password', 'correct horse battery staple');
         await driver.fill('#confirm-password', 'correct horse battery staple');
-        await driver.clickElement({
-          text: enLocaleMessages.restore.message,
-          tag: 'button',
-        });
+        await driver.press('#confirm-password', driver.Key.ENTER);
```
rickycodes, line 181: one thing I don't really like about this is it only tests keyboard input. it might be nice to continue testing both? but idk if there's really an effective way. you could introduce some randomness, but part of testing is making these reproducible so :shrug: 

idea:
```suggestion
        if (Math.floor(Math.random() * 2)) {
          await driver.clickElement({
            text: enLocaleMessages.restore.message,
            tag: 'button',
          });
        } else {
          await driver.press('#confirm-password', driver.Key.ENTER);
        }
```

adonesky1: 🤔 Not sure if we should do this. Seems like this thread has some interesting perspectives: https://sqa.stackexchange.com/questions/33444/how-do-you-design-ui-automated-tests-to-pick-alternative-steps-randomly-in-a-flo

adonesky1 (approved):
LGTM! Tested and seems work!

test/e2e/tests/metamask-responsive-ui.spec.js
Gudahtt: Oh my goodness, please no randomness :weary:. Not unless you're trying to tank team morale. The e2e tests fail intermittently enough as-is.

Gudahtt: Testing both cases would make sense though. Maybe not here in the "responsive layout" test, but somewhere.

rickycodes: hey, it was just an idea, but I agree it's a bad one! I didn't commit the suggestion for a reason. I think we can just leave it as testing with `<enter>` (the way it is now). I'm confident that as long as the button is `submit` it will function as desired.

darkwing (approved):

PR #13003, opened 2021-12-07
Remove `nodeify`

Gudahtt:
The `nodeify` utility is no longer needed for the background API. Instead each method is assumed to be either synchronous or Promise-returning.

The error handling was updated to at least log the error in the case where a method fall fails after the connection is broken.


app/scripts/controllers/preferences.test.js
```diff
@@ -41,6 +42,30 @@ describe('preferences controller', function () {
     sinon.restore();
   });
 
+  describe('useBlockie', function () {
```
Gudahtt, line 45: These tests used to be in `metamask-controller.test.js` even though they were testing the preferences controller. They were moved here.

app/scripts/lib/createMetaRPCHandler.js
```diff
@@ -15,24 +15,35 @@ const createMetaRPCHandler = (api, outStream) => {
...
+    let error;
+    try {
+      result = await api[data.method](...data.params);
+    } catch (err) {
+      error = err;
+    }
+
+    if (outStream._writableState.ended) {
+      if (error) {
+        console.log(error);
```
Gudahtt, line 29: Oops, this was meant to be `console.error`.

```suggestion
        console.error(error);
```

app/scripts/metamask-controller.js
```diff
@@ -3320,74 +3213,6 @@ export default class MetamaskController extends EventEmitter {
     await this.threeBoxController.init();
   }
 
-  /**
-   * Sets whether or not to use the blockie identicon format.
-   * @param {boolean} val - True for bockie, false for jazzicon.
-   * @param {Function} cb - A callback function called when complete.
-   */
-  setUseBlockie(val, cb) {
```
Gudahtt, line 3328: All of these wrapper functions were no longer needed. These functions were manually attempts at "nodeify-ing" the called functions essentially.

app/scripts/metamask-controller.test.js
```diff
@@ -33,22 +33,32 @@ const firstTimeState = {
 const ganacheServer = new Ganache();
 
 const threeBoxSpies = {
+  _registerUpdates: sinon.spy(),
```
Gudahtt, line 36: This mock really isn't ideal, but I needed to add all methods referenced in the background API to it to prevent it from blowing up. Hopefully we can replace it altogether in the Jest migration.

app/scripts/metamask-controller.test.js
```diff
@@ -813,21 +798,6 @@ describe('MetaMaskController', function () {
     });
   });
 
-  describe('#setCurrentLocale', function () {
```
Gudahtt, line 816: This was moved to the preferences controller test module.

ui/store/actions.test.js
```diff
@@ -1485,23 +1485,20 @@ describe('Actions', () => {
 
     it('calls setUseBlockie in background', async () => {
       const store = mockStore();
-
-      const setUseBlockStub = background.setUseBlockie.callsFake((_, cb) =>
```
Gudahtt, line 1489: These changes were needed because the tests referenced old wrapper functions, which were removed.

The fixtures here are still callback-based because the extension API used for cross-process communication is still callback-based.

app/scripts/lib/createMetaRPCHandler.js
```diff
@@ -15,24 +15,35 @@ const createMetaRPCHandler = (api, outStream) => {
...
+    let error;
+    try {
+      result = await api[data.method](...data.params);
+    } catch (err) {
+      error = err;
+    }
+
+    if (outStream._writableState.ended) {
+      if (error) {
+        console.error(error);
```
Gudahtt, line 29: This is the change to error handling mentioned in the description. This is just to ensure that any error is logged _somewhere_ unless explicitly ignored. If the stream is still open, the error is the responsibility of the UI.

mcmire (dismissed):
This looks great! Now if we can somehow make it so that we don't have to `bind` each controller method to the controller that would simplify MetamaskController even more 😅  But this is a great improvement.

Given that this touches a bunch of core code in the extension, is it customary to have QA test everything as much as possible or are we confident that the tests cover all of these changes?

Gudahtt:
I would not be confident in e2e tests touching all of these functions. They aren't complete enough. To properly QA this would require a comprehensive overview of the application. Though we're likely to land other changes before the next release that require a full overview anyway (e.g. enabling LavaMoat in the UI).

Gudahtt:
Rebased to resolve a conflict

mcmire (approved):

darkwing (approved):
I tried numerous operations that this PR affected, as well as reviewed changes, and all is working well.  Happy to see `nodify` gone!

PR #13004, opened 2021-12-07
Add `yarn-deduplicate` step to README

Gudahtt:
The instructions for changing dependencies have been updated to include the `yarn yarn-deduplicate` step. This command removes duplicate dependencies in the lockfile where possible while keeping everything in-range, and it's checked in CI in the `test-yarn-dedupe` job that was added in #12737.


ryanml (approved):

georgewrmarshall (approved):
lgtm

PR #13013, opened 2021-12-07
Fixing EIP-1559 V2 advance gas fee inputs validations

jpuri:
PR fixes issues in advance gas fee save button behaviour:

1. save button is enabled by default
2. save button gets disabled in only 2 scenario:
        1. Priority fee entered is < 1
        2. Base fee in GWEI is less than priority fee

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -845,7 +845,7 @@
     "message": "Max priority fee must be greater than 0 GWEI"
   },
   "editGasMaxPriorityFeeBelowMinimumV2": {
-    "message": "Priority fee must be at least 1 GWEI"
+    "message": "Priority fee must be greater than 0."
   },
   "editGasMaxPriorityFeeHigh": {
     "message": "Max priority fee is higher than necessary. You may pay more than needed."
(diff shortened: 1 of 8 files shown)


NiranjanaBinoy (approved):
LGTM

mcmire (approved):

darkwing:
InpuTerrors

The first error message seems incorrect, since it's not lower than the priority fee.

jpuri:
@darkwing : this is bit misleading, but first input is now not value in GWEI but multiplier and the error message signifies that values in GWEI for base fee is less than priority fee.

Let me discuss this with @jakehaugen

jakehaugen:
> this is bit misleading, but first input is now not value in GWEI but multiplier and the error message signifies that values in GWEI for base fee is less than priority fee.
> 
> Let me discuss this with @jakehaugen

Yes, this logic needs to be updated. When it's a multiplier it should validate on the GWEI amount, not the multiplier itself. 

That also means we should probably update the copy for this specific case. 

I'd propose a copy change to: `Multiplier is low relative to "Priority fee"`

When editing in GWEI, the error copy/validation logic should remain as is.

PR #13016, opened 2021-12-08
Changes in gas loading animation in EIP-1559 V2

### LINKED ISSUE #12976: Reduce EIP-1559 v2 gas fee refresh animation
This issue aims to reduce EIP-1559 gas fee refresh animation in order to unblock users from submitting transactions while fees are refreshing, and reduce anxiety caused by the frequent animation.

- [x] remove overall screen refresh animation from the transaction confirmation, gas edit, and advanced edit screens
- [x] just refresh  gas estimates. Refresh animation of the estimates should stay the same. from each of those screens (noted in red on the screenshots below)
- [x] remove disabling of `confirm` and `save` buttons while gas estimates are refreshed
Screen Shot 2021-12-03 at 7 15 58 PM

Screen Shot 2021-12-03 at 7 13 51 PM

Screen Shot 2021-12-03 at 7 14 01 PM

jpuri:
Fixes: #12976

Explanation:  
PR includes. removing these 2 types of animations:
1. overall screen loading animation while gas is loading
2. disabling of submit button while gas is loading


ui/components/app/edit-gas-fee-popover/edit-gas-item/edit-gas-item.js
```diff
@@ -24,6 +25,8 @@ import EditGasToolTip from '../edit-gas-tooltip/edit-gas-tooltip';
 import InfoTooltip from '../../../ui/info-tooltip';
 import { useCustomTimeEstimate } from './useCustomTimeEstimate';
 
+const HeartBeat = () => (process.env.IN_TEST ? null : <LoadingHeartBeat />);
```
darkwing, line 28: In the `LoadingHeartBeat` component itself, let's just do the `process.env.IN_TEST` check there and return null, so that we don't have to do conditional checks like this

darkwing (changes requested):

ui/components/app/edit-gas-fee-popover/edit-gas-item/edit-gas-item.js
jpuri: good point, I just fixed

adonesky1:
@jpuri @kevinghim reading through the issue and looking at PR its still unclear to me where we landed on disabling the "confirm" button when new gas is loading?
Screen Shot 2021-12-09 at 10 16 35 AM

adonesky1:
Should the total section also blink since it will also update if gasFee's change?
Screen Shot 2021-12-09 at 10 19 17 AM
?

ui/components/ui/loading-heartbeat/index.js
```diff
@@ -11,6 +11,8 @@ export default function LoadingHeartBeat() {
   useShouldAnimateGasEstimations();
   const active = useSelector(getGasLoadingAnimationIsShowing);
 
+  if (process.env.IN_TEST) return null;
```
mcmire, line 14: Should we put this higher? I'm just thinking maybe there is no reason to run `useShouldAnimateGasEstimations` if we don't need to.

jpuri: react warns for conditionally calling hooks

jpuri:
@adonesky1 : yep it is required to disable confirm button for loading estimates, this avoid scenario - that user confirms with an estimate and it changes as user confirms and transaction has a different estimate.

jpuri:
@adonesky1 : I made fixes and this is how it looks now:

https://user-images.githubusercontent.com/2182307/145465978-83863f19-4d0f-4e45-8a41-d089b9f7d06e.mov


In addition confirm button on confirm transaction page is also disabled during loading

app/_locales/en/messages.json
```diff
@@ -820,15 +820,18 @@
...
   },
-  "editGasMaxBaseFeeImbalance": {
-    "message": "Max base fee cannot be lower than priority fee"
-  },
   "editGasMaxBaseFeeLow": {
     "message": "Max base fee is low for current network conditions"
   },
+  "editGasMaxBaseFeeMultiplierImbalance": {
+    "message": "Multiplier is low relative to \"Priority fee\""
+  },
```
jpuri, line 834: @darkwing  : this is more intuitive message

RefL https://github.com/MetaMask/metamask-extension/pull/13013#issuecomment-989179553

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-input-subtext/advanced-gas-fee-input-subtext.js
```diff
@@ -10,18 +10,24 @@ const AdvancedGasFeeInputSubtext = ({ latest, historical }) => {
     <Box className="advanced-gas-fee-input-subtext">
       <Box display="flex" alignItems="center">
         <span className="advanced-gas-fee-input-subtext__label">
-          <LoadingHeartBeat />
           <I18nValue messageKey="currentTitle" />
         </span>
-        <span>{latest}</span>
+        <span className="advanced-gas-fee-input-subtext__value">
+          {' '}
```
adonesky1: `{' '}` probably not required?

jpuri: yep removed

adonesky1 (dismissed):
LGTM!

ui/components/ui/loading-heartbeat/index.js
mcmire: Hmm. Since `useShouldAnimateGasEstimations` calls `useGasFeeEstimates`, and since `useShouldAnimateGasEstimations` includes two `useEffect`s, it seems like it might not be worth it to put this conditional here then? Otherwise it doesn't seem like the heartbeat would be completely disabled.

darkwing (approved):

adonesky1 (approved):

PR #13018, opened 2021-12-08
Dispose the trezor connect iframe upon lock

danjm:
Blocked by merge and release of https://github.com/MetaMask/eth-trezor-keyring/pull/113

This PR addresses an error that would occur if MetaMask is locked and then unlocked, when a Trezor account has been added.

While the error was obscured by an issue with Lavamoat, we were able to determine that it was an error being thrown from within TrezorConnect: `TypeError: Cannot add property default, object is not extensible`

This error is thrown when trying to call `TrezorConnect.init` (within the trezor keyring constructor) if the Trezor iframe already exists. We instantiate a new Trezor keyring whenever we unlock, so if a user was to lock and unlock without closing the browser, this issue could happen.

The `TrezorConnect.dispose` method removes the iframe from the dom. That method is exposed on the keyring in the PR linked above.

This PR makes it so that we call that method upon locking MetaMask.

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -3519,7 +3519,7 @@ export default class MetamaskController extends EventEmitter {
    */
   setLocked() {
     const [trezorKeyring] = this.keyringController.getKeyringsByType(
-      'Trezor Hardware',
+      KEYRING_TYPES.TREZOR,
     );
     if (trezorKeyring) {
       trezorKeyring.dispose();
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -3518,6 +3518,12 @@ export default class MetamaskController extends EventEmitter {
    * Locks MetaMask
    */
   setLocked() {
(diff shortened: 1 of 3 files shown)


app/scripts/metamask-controller.js
```diff
@@ -3518,6 +3518,12 @@ export default class MetamaskController extends EventEmitter {
    * Locks MetaMask
    */
   setLocked() {
+    const [trezorKeyring] = this.keyringController.getKeyringsByType(
+      'Trezor Hardware',
```
darkwing, line 3522: We should use `KEYRING_TYPES.TREZOR` from `shared/constants/hardware-wallets.js` here

danjm: good call, added in latest commit

darkwing:
Tested this and worked well! Really great find on the `dispose`!

darkwing (dismissed):

darkwing (approved):

PeterYinusa (approved):

PeterYinusa:
LGTM!

PR #12097, opened 2021-09-14
add BitBox02 hardware wallet support

benma:
Fixes: #8119

[Shift](https://shiftcrypto.ch/) is receiving requests for a BitBox02 Metamask integration nearly on a daily basis.

Previously we waited with working on the integration until the [plugin system](https://github.com/MetaMask/metamask-snaps-beta/wiki) was ready. 

However, since user requests are piling up, we took another look and saw that the integration is actually quite trivial:

- Add BitBox02 next to Trezor/Ledger in the lists for hardware wallets
- Copy the [Trezor keyring](https://github.com/MetaMask/eth-trezor-keyring/blob/c606543f2bbd4135b8bf1a83c45ceaac71b3461c/index.js) to `eth-bitbox02-keyring.js` and fill in the connection for the BitBox02

As a result, the diff is quite minimal and easy to understand and maintain.

When the plugin system is ready, we will be happy to port the integration to that if desired.

We will also be happy to maintain this integration and fix any issues that might arise. We are generally careful not to introduce any breaking changes.

Manual testing steps:  
  - Install the [BitBoxBridge](https://shiftcrypto.ch/download/)
  - Connect an initialized BitBox02 (firmware v9.7.0 or later)
  - Connect it as a hardware wallet and use it to sign transactions and messages

PR #13023, opened 2021-12-08
Support EIP-1559 on Trezor Model One

danjm:
This PR brings in the version of eth-trezor-keyring (and thereby trezor-connect) that allows EIP-1559 transactions to work on the Trezor Model One.

I will prepare a proper PR later today to remove all dead code, and make the associated changes in the controllers repo, but in the interest of unblocking v10.8.0 I am doing this the most simple way possible right now

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -2519,20 +2519,9 @@ export default class MetamaskController extends EventEmitter {
   }
 
   /**
-   * Method to return a boolean if the keyring for the currently selected
-   * account is a ledger or trezor keyring.
-   * TODO: remove this method when Ledger and Trezor release their supported
-   * client utilities for EIP-1559
    * @returns {boolean} true if the keyring type supports EIP-1559
    */
-  async getCurrentAccountEIP1559Compatibility(fromAddress) {
-    const address =
-      fromAddress || this.preferencesController.getSelectedAddress();
-    const keyring = await this.keyringController.getKeyringForAccount(address);
-    if (keyring.type === KEYRING_TYPES.TREZOR) {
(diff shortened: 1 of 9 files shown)


darkwing (dismissed):

aloisklink (commented):
:+1: I was literally working on the exact same thing, I was just a few hours too slow!

Two minor comments:

~1. There's a pending PR to add a hardware wallet called BitBox02: https://github.com/MetaMask/metamask-extension/pull/12097
  They use the `isEIP1559Account` and `getCurrentAccountEIP1559Compatibility` functions, since they don't support EIP1559, so it might be worth temporarily leaving in the dead code until that's merged.~
  Edit: never mind, I didn't see the latest comments saying that you guys aren't adding anymore USB HW wallets.

2. There's a failing test. Unfortunately, I can't add a comment/suggestion on it, since you didn't edit that code, but essentially, in the `'returns true if network does not support EIP-1559'` test for when `EIPS: {1559: undefined}`, `mockState.metamask.keyrings` is a TREZOR wallet.
   And because that Trezor wallet now supports EIP-1559, `checkNetworkOrAccountNotSupports1559` now returns `false` and the test fails.
   I've got a commit that fixes it here, feel free to cherry-pick it (or just copy-and-paste) https://github.com/aloisklink/metamask-extension/commit/386f3b3d986c929d3020ff9ac2a317461cb2b3e4

darkwing (approved):
LGTM!  We should evaluate uses of `getCurrentAccountEIP1559Compatibility` and remove them if we're only returning `true`; that can happen in a follow up though.

tmashuang (approved):

PR #13026, opened 2021-12-08
Add subject types

rekmarks:
This PR introduces the concept of subject _types_ to be associated with each subject in the `SubjectMetadataController`, and used for control flow in our RPC stack (`setupProviderEngine` and so forth).

We already differentiate between "types" of subjects in various places on an ad hoc basis via boolean flags (e.g. `isInternal` in our RPC stack) or the presence/absence of certain values in the subject's metadata (specifically `metadata.extensionId`). The status quo is manageable if not ideal, but will start to become untenable with the introduction of Snaps in the near future.

Therefore, this PR establishes a `SUBJECT_TYPES` enum and adds the `subjectType` property to the metadata of each subject. A new migration is added to accomplish this. Finally, we specify and `INTERNAL` subject type to distinguish internal from external requests.

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -58,7 +58,10 @@ import {
 import { UI_NOTIFICATIONS } from '../../shared/notifications';
 import { toChecksumHexAddress } from '../../shared/modules/hexstring-utils';
 import { MILLISECOND } from '../../shared/constants/time';
-import { POLLING_TOKEN_ENVIRONMENT_TYPES } from '../../shared/constants/app';
+import {
+  POLLING_TOKEN_ENVIRONMENT_TYPES,
+  SUBJECT_TYPES,
+} from '../../shared/constants/app';
 
 import { hexToDecimal } from '../../ui/helpers/utils/conversions.util';
 import ComposableObservableStore from './lib/ComposableObservableStore';
@@ -2741,7 +2744,12 @@ export default class MetamaskController extends EventEmitter {
    */
   setupProviderConnection(outStream, sender, isInternal) {
     const origin = isInternal ? 'metamask' : new URL(sender.url).origin;
+    let subjectType = isInternal
+      ? SUBJECT_TYPES.INTERNAL
+      : SUBJECT_TYPES.WEBSITE;
+
     if (sender.id !== this.extension.runtime.id) {
+      subjectType = SUBJECT_TYPES.EXTENSION;
       this.subjectMetadataController.addSubjectMetadata(origin, {
         extensionId: sender.id,
       });
@@ -2756,7 +2764,7 @@ export default class MetamaskController extends EventEmitter {
       origin,
       location: sender.url,
       tabId,
-      isInternal,
+      subjectType,
     });
 
     // setup connection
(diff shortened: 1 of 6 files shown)


app/scripts/metamask-controller.js
```diff
@@ -2779,14 +2787,15 @@ export default class MetamaskController extends EventEmitter {
   }
 
   /**
-   * A method for creating a provider that is safely restricted for the requesting domain.
+   * A method for creating a provider that is safely restricted for the requesting subject.
+   *
    * @param {Object} options - Provider engine options
    * @param {string} options.origin - The origin of the sender
    * @param {string} options.location - The full URL of the sender
+   * @param {string} options.subjectType - The type of the sender subject.
```
Gudahtt, line 2795: Nit: It's not really the type of the _object_. It's the type of the sender.

There is a sender object (e.g. in `setupProviderConnection`) but that doesn't differ between subject types.

```suggestion
   * @param {string} options.subjectType - The type of the sender.
```

app/scripts/migrations/069.js
```diff
@@ -0,0 +1,40 @@
...
+  },
+};
+
+function transformState(state) {
+  const { SubjectMetadataController = {} } = state;
+  const { subjectMetadata = {} } = SubjectMetadataController;
+
+  // mutate SubjectMetadataController.subjectMetadata in place
+  Object.values(subjectMetadata).forEach((metadata) => {
+    if (metadata && typeof metadata === 'object' && !Array.isArray(metadata)) {
```
Gudahtt, line 27: Nit: These branches seem untested. You could add a test with malformed subject metadata (e.g. the metadata is an array, or is something other than an object).

And if we want to protect against malformed state in general, it might be worth ensuring `subjectMetadata` is an object before invoking `Object.values` with it.

app/scripts/migrations/index.js
```diff
@@ -72,6 +72,7 @@ import m065 from './065';
 import m066 from './066';
 import m067 from './067';
 import m068 from './068';
+import m069 from './069';
 
 const migrations = [
```
mcmire, line 77: Is it worth replacing this with a dynamic import at some point?

app/scripts/migrations/069.test.js
```diff
@@ -0,0 +1,99 @@
...
+    };
+
+    const newStorage = await migration69.migrate(oldStorage);
+    expect(newStorage).toStrictEqual({
+      meta: {
+        version: 69,
+      },
+      data: {
+        FooController: { a: 'b' },
+        SubjectMetadataController: { subjectMetadata: {} },
```
Gudahtt, line 95: Nit: I guess it doesn't matter, but usually we avoid making unnecessary changes to state in migrations. If there is no subject metadata state to migrate, we could avoid making any changes rather than initializing this empty object.

mcmire (dismissed):
Makes sense to me 👍🏻

Gudahtt (dismissed):
LGTM! Left a few suggested improvements

app/scripts/migrations/index.js
rekmarks: Hm, what do you have in mind? Like, dynamically importing all migrations? That's an improvement I could get behind.

app/scripts/metamask-controller.js
rekmarks: It's sender _subject_, not object!

Gudahtt: Welp. Not sure how I misread that, even during editing.

Gudahtt (approved):

hmalik88 (approved):
LGTM.

app/scripts/migrations/index.js
mcmire: Yeah, looping through all of the files inside of this folder and then dynamically importing them instead of listing them by hand. Not proposing this in the current PR, mind, just an idea as I read this.

mcmire (approved):

PR #13027, opened 2021-12-08
Add `@metamask/rpc-methods` package

Gudahtt:
The `selectHooks` function has been replaced with the equivalent function from the `@metamask/rpc-methods` package, which is functionally equivalent.

The function was included in that package so that it could be used elsewhere in the `snaps-skunkworks` repo. Eventually the goal is to migrate much of our RPC logic into this package so that it can be shared across products, and by our libraries as needed.

Manual testing steps:  
  - This should include no functional changes


rekmarks (dismissed):
LGTM!

Gudahtt:
Rebased to resolve a lockfile conflict

danjm (approved):

adonesky1 (approved):
LGTM

PR #12860, opened 2021-11-25
Added snaps error notification

shanejonas:
This adds the error notification using the `HomeNotification` component. It also adds the `HomeNotification` to storybook.

Fixes: #11897 
  
image

PR #11837, opened 2021-08-13
Snaps Canary Release

rekmarks:
Closes #11626

PR #13035, opened 2021-12-09
Add stories for Home notification component

Gudahtt:
The Home notification component now has a storybook file. This work was extracted from the `snaps` branch, and was originally implemented in PR #12860.

The storybook file was enhanced to include more stories demonstrating each likely usage scenario.

### DIFF
--- a/ui/components/app/home-notification/home-notification.component.js
+++ b/ui/components/app/home-notification/home-notification.component.js
@@ -91,14 +91,55 @@ const HomeNotification = ({
 };
 
 HomeNotification.propTypes = {
+  /**
+   * The text for the "Accept" button. This must be accompanied by the `onAccept` prop.
+   *
+   * The "Accept" button is only rendered if this prop is set.
+   */
   acceptText: PropTypes.node,
+  /**
+   * The text to display alongside the checkbox.
+   *
+   * The checkbox state is passed to the `onIgnore` handler, so this should only be used if the `onIgnore` prop is set.
+   *
+   * The checkbox is only rendered if this prop is set.
+   */
   checkboxText: PropTypes.node,
+  /**
+   * The text to display in the checkbox tooltip.
+   *
+   * The tooltip is only rendered if this prop is set.
+   */
   checkboxTooltipText: PropTypes.node,
+  /**
+   * Custom class names.
+   */
   classNames: PropTypes.array,
+  /**
+   * The notification description.
+   */
   descriptionText: PropTypes.node.isRequired,
+  /**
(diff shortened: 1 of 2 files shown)


mcmire (commented):
The checkbox in the `WithIgnoreCheckbox` story is uncheckable and it looks like there's a double checkbox there. Is that intentional? 

Also, what do you think about using some more real-world text in the stories? Or perhaps a description of when you would use the `OnlyAccept` version vs. `OnlyIgnore`?

Screen Shot 2021-12-09 at 2 17 13 PM

Gudahtt:
Hmm. The highlighted styles do look a bit messed up, but that is a pre-existing problem with the component itself. The checkbox works for me, and I only see a single one.

>Also, what do you think about using some more real-world text in the stories? Or perhaps a description of when you would use the OnlyAccept version vs. OnlyIgnore?

Sure, I can improve the example text. And that reminds me, I was going to try writing an `.mdx` file as well to accompany this. The documentation can live there.

georgewrmarshall (commented):
Looks great! You could add jsDoc comments to explain the propTypes more in depth but apart from that everything looks great! The comments in the propTypes will show up in the ArgsTable on the docs page.

Screen Shot 2021-12-10 at 12 17 16 PM

rekmarks (dismissed):
This was already reviewed by myself, @darkwing, and @georgewrmarshall as part of #12860. This won't be a problem for much longer, but we should avoid double-reviewing work that was merged into #11837.

Gudahtt:
This PR is quite different from #12860. The single story added in that PR was expanded to 6 stories in this one, to represent the different use cases for this component that affect the layout.

Gudahtt:
I have improved the example text, and I added JSDoc comments to the props.

I tried writing a `.mdx` file but it didn't show up anywhere in Storybook for that component :thinking: Our 
Storybook "Getting started" docs made me think that the recommendation was to create an `.mdx` file per-component, but I guess I misunderstood. It looks like that's just for the top-level Storybook documentation instead.

darkwing (approved):

mcmire (approved):
Nice, I like the added JSDoc descriptions.

georgewrmarshall:
Thanks for the feedback on the docs! The intention was to provide 2 options. A simple story without a docs page and a comprehensive version that could include an mdx file. If you'd like to provide an mdx file with further documentation you need to import it in the story. And set the docs parameter. See the Button version in the documentation guidelines page. I'll review the docs and try make this clearer 

image

image

Gudahtt:
Ahh, thanks for that explanation @georgewrmarshall ! That makes sense.

PR #13036, opened 2021-12-09
Added Snaps Custom Confirmation

### LINKED ISSUE #11686: Snaps Custom Confirmation

ritave:
fixes #11686
Screenshot 2021-12-13 at 18 19 32

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -64,6 +64,9 @@ import { UI_NOTIFICATIONS } from '../../shared/notifications';
 import { toChecksumHexAddress } from '../../shared/modules/hexstring-utils';
 import { MILLISECOND } from '../../shared/constants/time';
 import {
+  ///: BEGIN:ONLY_INCLUDE_IN(flask)
+  MESSAGE_TYPE,
+  ///: END:ONLY_INCLUDE_IN
   POLLING_TOKEN_ENVIRONMENT_TYPES,
   SUBJECT_TYPES,
 } from '../../shared/constants/app';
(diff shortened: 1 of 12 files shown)


darkwing:
Hello @ritave !  Could we get a screenshot or video of how this will appear?

ritave:
@darkwing I've updated the PR comment with a screenshot

darkwing:
Code looks good but I feel like, from a design perspective, we should differentiate the basic text from the specific text like `message` and `account`, otherwise the text all just runs together.  Could we differentiate the font or use bold?

hmalik88 (commented):
The screenshot doesn't seem to match the figma design, did we have something to pass in for title and subtitle?

ritave:
@hmalik88 The header and subtitle are empty in the FilSnap example
@darkwing The text is provided by an external Snap, not Metamask, in the current release it's not planned the text to be style-able

rekmarks (changes requested):
This is most of the way there, we just need to fix one thing before we merge, and make another change pending a new version of the `skunkworks` packages.

First, we are missing the disclaimer in the footer, above the buttons. The copy used for this footer we can iterate on, but we need some value there. The `add-ethereum-chain` confirmation template has a similar footer, and can be used as a reference.

<details><summary>Screenshot of designs</summary></details>

The footer is the only thing we need to fix to merge this PR. Secondly, there is the question of the parameters, specifically their names and purpose. I think the Filecoin Snap has lead us astray in terms of how we use the values we receive from `snap_confirm`, compounded by a poor review of mine of the method implementation.

At the moment, the `prompt` is used to populate the scrollable text area, while the title and subtitle components of the confirmation are unused. Instead, the `prompt` should be a short question used as the title of the confirmation, while the subtitle / description and text area contents should be optional.

I created a PR against `skunkworks` updating the method [here](https://github.com/MetaMask/snaps-skunkworks/pull/158), but fixing the footer issue is sufficient to merge this PR. We can update it per the `skunkworks` changes later.

ritave:
Screenshot 2021-12-14 at 13 43 15
@rekmarks I've added the footer. I've rebased onto latest snaps branch and seems that origin and subjectMetadata are empty now, at least for FilSnap.

rekmarks (approved):
LGTM!

hmalik88 (approved):
LGTM.

Issue #11682, opened 2021-07-30
Snaps Installation UI


Gudahtt:
We should allow dapps to request permission to use snaps when requesting permissions or connecting. If the user doesn't have the snap installed yet, this should initiate the snap install process. We will need to download the snap manifest and present the user with enough information to decide whether or not to install the snap. This should include:

- [x] The snap name
- [x] The snap author
- [x] The place the snap was downloaded from
- [x] The snap description
- [x] The permissions the snap requires
        [ ] This should include detailed description of what each permission is, and why it might matter to a user
- [x] The reason I'm being asked to install the snap (e.g. the dapp that asked permission to use the snap).

This UI should be incorporated into the connect flow, so that the dapp is completely ready to use after the connection process.

As a user, I want to be notified of the following snaps errors:
- When a snap fails to install
  - This will probably be because the snap crashed immediately when MetaMask tried to run it, but it could be for other reasons, so a generic "failed to install" error with a more detailed / verbose error message somehow accessible would be helpful.

https://www.figma.com/file/nBKUQx0btfQiqVf30UA6KD/Flask-MVP?node-id=336%3A9134

Gudahtt:
This is blocked by #11626

holantonela:
After our permissions discussion, I consolidated the last version here, ready for review:

https://www.figma.com/file/nBKUQx0btfQiqVf30UA6KD/?node-id=336%3A6342

Converting the requirement's list in a to-do here too.

holantonela:
`The reason I'm being asked to install the snap (e.g. the dapp that asked permission to use the snap).` is it ok to think about this requirement as an step in the connect flow where "the site wants to connect with snap" ? For a reference view flow "Add Snap - Site permissions"

holantonela:
` The place the snap was downloaded from` is this a URL? How we want to show it?

Gudahtt:
>is it ok to think about this requirement as a step in the connect flow where "the site wants to connect with snap" ?

Yes - definitely. I think that would fulfill the requirement anyway. I guess we'll see if users are confused or not :sweat_smile:. We can spell it out for them if it becomes an issue.

>is this a URL? How we want to show it?

See [my comment on the other issue](https://github.com/MetaMask/metamask-extension/issues/11681#issuecomment-896833014).

holantonela:
Final UI here

https://www.figma.com/file/nBKUQx0btfQiqVf30UA6KD/Snaps-MVP?node-id=336%3A6344

PR #13039, opened 2021-12-09
Choose accounts refactor

hmalik88:
Related to: [Snaps Installation UI issue](https://github.com/MetaMask/metamask-extension/issues/11682)

**Background Updates:** Fixes to `subjectMetadata` were made to include the `origin`. An "unknown" subject type was added and integrated into the `useOriginMetadata` hook. Background updates are being made in this PR instead of a separate one, because in interest of time, the changes are relevant to this UI and because there aren't a significant number of changes.

**UI Explanation:**  Flask designs ([Figma](https://www.figma.com/file/nBKUQx0btfQiqVf30UA6KD/Flask-MVP?node-id=336%3A6344)) updated the `ChooseAccount` screen to have a bolded title and black subtitle. The account list was refactored into its own component (`AccountList`) in an effort to tidy up  `ChooseAccount`. A wrapper div was added around the account list to prevent storybook from collapsing the div. `ChooseAccount` was also converted to a functional component. Following pics are of the collapsed account list in storybook before the wrapper, a before of it functioning and after of when the styles/text were updated:

Screen Shot 2021-12-09 at 3 08 31 PM

Screen Shot 2021-12-08 at 3 00 41 PM

Screen Shot 2021-12-09 at 2 45 45 PM

Note: `ChooseAccount` will be updated in the future to change the choose accounts text to conditionally end with either "site" or "snap" when snaps supports `eth_requestAccounts`.

Manual testing steps:  
  - Start storybook
  - Navigate to `ChooseAccount`
  - Additionally, you may run `yarn start`, load the extension into chrome, navigate to a dapp (i.e. Aave) and connect to MetaMask.


georgewrmarshall (dismissed):
UI looks good. There are a few rogue values here that don't adhere to the 8px grid convention but I think this is the UI that I myself tried to refactor and had a very hard time as the different components effect the layout of different pages. So for the sake of keeping things consistent. I'm happy to approve.

ui/components/ui/account-list/account-list.js
```diff
@@ -0,0 +1,175 @@
...
+      checked = UNCHECKED;
+    } else {
+      checked = INDETERMINATE;
+    }
+
+    return (
+      <div
+        className={classnames({
+          'choose-account-list__header--one-item': accounts.length === 1,
+          'choose-account-list__header--two-items': accounts.length > 1,
```
darkwing, line 38: Should we use `multiple` instead of 2?  Not important just a thought.

ui/components/ui/account-list/index.scss
```diff
@@ -0,0 +1,140 @@
...
+  height: 100%;
+  align-items: center;
+
+  &__header--one-item,
+  &__header--two-items {
+    display: flex;
+    flex: 0;
+    margin-top: 36px;
+    width: 100%;
+    padding-left: 15px;
```
darkwing, line 14: Let's use `padding-inline-start` and `padding-inline-end` for better RTL support.

darkwing: We can use `margin-inline-start` and `margin-inline-end`  elsewhere in this file.

ui/pages/permissions-connect/choose-account/choose-account.js
```diff
@@ -0,0 +1,148 @@
...
+    );
+    setSelectedAccounts(newSelectedAccounts);
+  };
+
+  const deselectAll = () => {
+    setSelectedAccounts(new Set());
+  };
+
+  const allAreSelected = () => {
+    return accounts.every(({ address }) => selectedAccounts.has(address));
```
darkwing, line 47: Can we do `length` checks instead of mapping through accounts?

darkwing (commented):
Very well done, just a few nits!

georgewrmarshall:
One thing that's not apart of this PR but might be a good change is if we use sentence case for "Connect with MetaMask" in `app/_locales/en/messages.json`

ui/components/ui/account-list/account-list.js
hmalik88: I agree, I'll change that!

app/_locales/en/messages.json
```diff
@@ -430,6 +430,9 @@
   "chainIdExistsErrorMsg": {
     "message": "This Chain ID is currently used by the $1 network."
   },
+  "chooseAccounts": {
+    "message": "Choose the account(s) to use on this site"
+  },
```
rekmarks, line 435: Although this message matches the designs, I think the verb should be "select" here, not "choose". We overwhelmingly use "select" over "choose" in our existing locale messages (57 vs 2 instances), and we should do it for consistency if no other reason.

The remaining changes to the message are fine. We can probably rename the key to "selectAccounts" and keep the translations if we make this change. We just need to highlight the change to our translators next week.

rekmarks (changes requested):

ui/components/ui/account-list/index.scss
hmalik88: Pushed up some changes, please let me know if that works.

app/_locales/en/messages.json
```diff
@@ -2425,7 +2425,7 @@
     "message": "Select a higher gas fee to accelerate the processing of your transaction.*"
   },
   "selectAccounts": {
-    "message": "Select account(s)"
+    "message": "Select the account(s) to use on this site"
```
shanejonas, line 2428: whats the process for getting the translations for other languages? or should the translation stay the same?

hmalik88: I suppose the idea is to have this change and usher in changes to the other locales once we relay that message to the translators? @rekmarks thoughts?

georgewrmarshall (changes requested):
Looks like <SiteOrigin> isn't receiving the correct data after the refactor.

Testing on [E2E Test Dapp](https://metamask.github.io/test-dapp/)

Screen Shot 2021-12-14 at 9 23 15 AM

Screen Shot 2021-12-14 at 9 25 09 AM

Tested on [uniwap](https://app.uniswap.org/#/swap) as well
Screen Shot 2021-12-14 at 9 25 38 AM
Screen Shot 2021-12-14 at 9 25 54 AM

app/_locales/en/messages.json
rekmarks: Yeah, this change is fine, and we will deal with updating the translations out of band.

georgewrmarshall (dismissed):
lgtm!

georgewrmarshall (dismissed):
lgtm!

rekmarks (approved):
LGTM!

georgewrmarshall (approved):
lgtm

darkwing:
Wow, really excellent work @hmalik88 !

PR #13047, opened 2021-12-10
Move HW wallet tracking in Swaps

dan437:
## Description
Move HW wallet tracking outside of `sensitiveProperties` in Swaps.

## Manual testing steps
  - Open background.html and go to the Network tab
  - Filter it by the `batch` string
  - Check that we send these 2 params now inside the `properties` field


darkwing (approved):

jpuri (approved):

dan437:
This PR is no longer needed, since tracking of sensitive properties was fixed here: https://github.com/MetaMask/metamask-extension/pull/13237

PR #12970, opened 2021-12-03
Feat/collectibles the return

adonesky1:
Explanation:  Add required wiring + front-end work to get viewing/adding NFTs in extension

In designs but excluded here
- Send flow (will be included in fast-follow PR)
- `Report as Scam` flow (the functionality here is not sufficiently spec'd out to implement)

Screenshots:
Screen Shot 2021-11-30 at 4 09 50 PM
Screen Shot 2021-11-30 at 4 10 54 PM

Manual testing steps:  
  - Purchase or mint or have an (or preferably many) NFT(s) 
  
  Detection:
  - (only available on Mainnet)
  - Turn on `OpenSea Api` & `NFT autodetection` in settings -> experimental
  - Check that your NFT has been added to the NFTs tab
  
  Manual:
  - Navigate to the NFTs tab
  - Click "Import NFTs"
  - Enter `address`/`tokenId` in `/add-collectible` form
  
1 (with NFT you own)
  - Check that your NFT has been successfully added (should see green success toast)
 
2 (with NFT you don't own)
 - Check that the NFT has not been added (should see yellow warning toast: "collectible not owned by user")

PR #13091, opened 2021-12-13
Extending Box component to allow style prop

georgewrmarshall:
Fixes: #NAN

Explanation: Extending the `Box` component to allow for style prop. This provides an escape hatch for developers to use if the current prop selection does not meet their needs. It should be used as a last resort and allows things like background colors from third party providers. An example of this is the NFT feature and using background colors from opensea. [Relevant discussion at the bottom of this comment in this PR](https://github.com/MetaMask/metamask-extension/pull/12970#discussion_r768144267)

Example of third party background colors: 
Screen Shot 2021-12-14 at 10 38 48 AM

Need to add story and docs

Manual testing steps:  TBD
  - 
  - 
  -

### DIFF
--- a/ui/components/ui/box/box.js
+++ b/ui/components/ui/box/box.js
@@ -86,6 +86,7 @@ export default function Box({
   children,
   className,
   backgroundColor,
+  style,
 }) {
   const boxClassName = classnames('box', className, {
     // ---Borders---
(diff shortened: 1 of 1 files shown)


georgewrmarshall:
Closing for now as we will need to make a decision on how we handle dynamic inline style values across the board and whether allowing the style prop on our components is not type specific enough

PR #13098, opened 2021-12-14
Implement Dark Mode

darkwing:
Using CSS variables will make implementing a dark mode much, much easier and more performant.

### DIFF
--- a/ui/components/app/network-display/index.scss
+++ b/ui/components/app/network-display/index.scss
@@ -11,37 +11,10 @@
     cursor: not-allowed;
   }
 
-  &--colored {
-    background-color: lighten(rgb(125, 128, 130), 45%);
-  }
-
-  &--mainnet {
-    background-color: lighten($blue-lagoon, 68%);
-  }
-
-  &--ropsten {
-    background-color: lighten($crimson, 45%);
-  }
-
-  &--kovan {
-    background-color: lighten($purple, 65%);
-  }
-
-  &--rinkeby {
-    background-color: lighten($tulip-tree, 35%);
-  }
-
-  &--goerli {
-    background-color: lighten($dodger-blue, 35%);
-  }
-
-  &--localhost {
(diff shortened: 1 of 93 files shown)


georgewrmarshall (commented):
If my understanding is correct this implementation would add a `theme-dark` class to the root element of each page? Then we would have our theme css vars? under `:root.theme-dark`. Seems like a simple and elegant approach to me. Just wondering if this would scale to allow not only colors but other design tokens like space, font-size etc. to be able to create more themes in future?

darkwing:
> If my understanding is correct this implementation would add a `theme-dark` class to the root element of each page? Then we would have our theme css vars? under `:root.theme-dark`. Seems like a simple and elegant approach to me. Just wondering if this would scale to allow not only colors but other design tokens like space, font-size etc. to be able to create more themes in future?

Yep!  Any CSS value can be themed using `var`!

ui/components/app/network-display/index.scss
```diff
@@ -11,37 +11,10 @@
     cursor: not-allowed;
   }
 
-  &--colored {
```
darkwing, line 14: Note to self; this needs to be restored.

Gudahtt:
@darkwing Can this be closed now that #13147 has been merged?

darkwing:
@Gudahtt No -- this PR has the actual dark mode colors.  I've retitled the PR appropriately.

georgewrmarshall:
@darkwing would be great to standardize these with the design system. Have booked a meeting for Thursday to discuss

calebplum:
Any updates on this?

jpuri:
This is great work, code and approach looks good. I think we should be good to merge incremental changes for dark mode behind feature flag.

darkwing:
Closing as @GuillaumeRx and I have mapped out a different strategy moving forward.

PR #13109, opened 2021-12-16
Update `@babel/runtime` patch to fix lockdown error

Gudahtt:
We are currently patching `@babel/runtime` to fix various lockdown errors caused by `@babel/runtime` modifying globals as part of various polyfills. There was one lockdown error that was showing up in Sentry error reports, which is the polyfill used for `Reflect.construct`.

All of our supported browsers include this API, so the polyfill has been replaced with a Proxy that returns the `Reflect.construct`
function directly, except with the addition of the `default` and `__esModule` properties. I don't know what these properties are for (maybe ES5/ES6 interoperability?) but I left them just in case they were being relied upon.

Manual testing steps:  

I don't know of any easy way to reproduce this problem. But here is at least one method that has been shown to work:
  - Revert these two changes:
    - https://github.com/MetaMask/eth-trezor-keyring/pull/113
    - https://github.com/MetaMask/metamask-extension/pull/13018
  - Connect a Trezor account
  - Lock the extension
  - Unlock the extension
  - Before this patch, you should see the error `Cannot add property default, object is not extensible`. After this patch, you should see the error `Error: TrezorConnect has been already initialized` instead (which is fixed by the two changes reverted in the first step).

### DIFF
--- a/patches/@babel+runtime+7.16.3.patch
+++ b/patches/@babel+runtime+7.16.3.patch
@@ -1,3 +1,53 @@
+diff --git a/node_modules/@babel/runtime/helpers/construct.js b/node_modules/@babel/runtime/helpers/construct.js
+index 108b39a..e3c769b 100644
+--- a/node_modules/@babel/runtime/helpers/construct.js
++++ b/node_modules/@babel/runtime/helpers/construct.js
+@@ -1,26 +1,21 @@
+-var setPrototypeOf = require("./setPrototypeOf.js");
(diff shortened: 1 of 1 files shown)


danjm:
Looks good to me

danjm (approved):

kumavis:
> the addition of the default and __esModule properties. I don't know what these properties are for

yeah esm interop. this was authored in esm and transpiled to this

your solution is quite clever 😉 ill use this pattern again

kumavis (approved):

PR #13110, opened 2021-12-16
Feat/check update collectible ownership

adonesky1:
Explanation:  Adds use of the `checkAndUpdateCollectiblesOwnershipStatus` method, added to the CollectibleController in this [PR](https://github.com/MetaMask/controllers/pull/664). When a user clicks the "Refresh List" button
Screen Shot 2021-12-22 at 9 00 56 AM
at the bottom of the NFT tab, this method will query the contract of each the NFTs the user currently has in their collection and check whether they still own each. In the case that they do not still own it, a value on the collectible object `isCurrentlyOwned` will be set to false, and in the UI this collectible will be moved to a **newly added** previously owned section.

Manual testing steps:  
  - Get an NFT on a test network (this [faucet](https://faucet.paradigm.xyz/) will add NFTs to many testnets)
  - Manually add the NFT via the Import NFTs flow 
Screen Shot 2021-12-22 at 9 08 06 AM
  - Make sure its been correctly added as a collection in the NFTs tabs
  - Send the NFT using Etherscan:

https://user-images.githubusercontent.com/34557516/147114538-f32db2ad-afe8-4638-882c-6c95a0b63219.mov

- Once the transaction has been confirmed hit "refresh list". You should see that the NFT has now been moved to a "Previously Owned" Collection.

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,7 +107,7 @@
     "@keystonehq/metamask-airgapped-keyring": "0.2.1",
     "@material-ui/core": "^4.11.0",
     "@metamask/contract-metadata": "^1.31.0",
-    "@metamask/controllers": "^22.0.0",
+    "@metamask/controllers": "^23.0.0",
     "@metamask/eth-ledger-bridge-keyring": "^0.10.0",
     "@metamask/eth-token-tracker": "^3.0.1",
     "@metamask/etherscan-link": "^2.1.0",
--- a/ui/components/app/collectibles-items/collectibles-items.js
+++ b/ui/components/app/collectibles-items/collectibles-items.js
@@ -120,7 +120,11 @@ export default function CollectiblesItems({
               const { image, address, tokenId, backgroundColor } = collectible;
               const collectibleImage = getAssetImageURL(image, ipfsGateway);
               return (
-                <Box width={width} key={`collectible-${i}`} className="collectibles-items__collection-item-wrapper">
+                <Box
+                  width={width}
+                  key={`collectible-${i}`}
+                  className="collectibles-items__collection-item-wrapper"
+                >
                   <div
                     className="collectibles-items__collection-item"
                     style={{
--- a/ui/components/app/collectibles-items/collectibles-items.js
+++ b/ui/components/app/collectibles-items/collectibles-items.js
@@ -31,6 +31,7 @@ export default function CollectiblesItems({
   previouslyOwnedCollection = {},
 }) {
   const defaultDropdownState = { [PREVIOUSLY_OWNED_KEY]: false };
+  const [dropdownState, setDropdownState] = useState(defaultDropdownState);
   const ipfsGateway = useSelector(getIpfsGateway);
 
   Object.keys(collections).forEach((key) => {
@@ -142,7 +143,6 @@ export default function CollectiblesItems({
     );
   };
 
-  const [dropdownState, setDropdownState] = useState(defaultDropdownState);
   return (
     <div className="collectibles-items">
       <Box padding={[6, 4]} flexDirection={FLEX_DIRECTION.COLUMN}>
@@ -175,6 +175,26 @@ export default function CollectiblesItems({
 }
 
 CollectiblesItems.propTypes = {
+  previouslyOwnedCollection: PropTypes.shape({
+    collectibles: PropTypes.arrayOf(
+      PropTypes.shape({
+        address: PropTypes.string.isRequired,
+        tokenId: PropTypes.string.isRequired,
+        name: PropTypes.string,
+        description: PropTypes.string,
+        image: PropTypes.string,
+        standard: PropTypes.string,
+        imageThumbnail: PropTypes.string,
+        imagePreview: PropTypes.string,
+        creator: PropTypes.shape({
+          address: PropTypes.string,
+          config: PropTypes.string,
+          profile_img_url: PropTypes.string,
+        }),
+      }),
+    ),
+    collectionName: PropTypes.string,
+  }),
   collections: PropTypes.shape({
     collectibles: PropTypes.arrayOf(
       PropTypes.shape({
--- a/ui/components/app/collectibles-items/collectibles-items.js
+++ b/ui/components/app/collectibles-items/collectibles-items.js
@@ -26,6 +26,7 @@ const width =
     : BLOCK_SIZES.ONE_SIXTH;
 
 const PREVIOUSLY_OWNED_KEY = 'previouslyOwned';
+
 export default function CollectiblesItems({
   collections = {},
   previouslyOwnedCollection = {},
--- a/ui/components/app/collectibles-items/collectibles-items.js
+++ b/ui/components/app/collectibles-items/collectibles-items.js
@@ -24,8 +24,13 @@ const width =
   getEnvironmentType() === ENVIRONMENT_TYPE_POPUP
     ? BLOCK_SIZES.ONE_THIRD
     : BLOCK_SIZES.ONE_SIXTH;
-export default function CollectiblesItems({ collections = {} }) {
-  const defaultDropdownState = {};
+
+const PREVIOUSLY_OWNED_KEY = 'previouslyOwned';
+export default function CollectiblesItems({
+  collections = {},
+  previouslyOwnedCollection = {},
+}) {
+  const defaultDropdownState = { [PREVIOUSLY_OWNED_KEY]: false };
   const ipfsGateway = useSelector(getIpfsGateway);
 
   Object.keys(collections).forEach((key) => {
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -107,7 +107,7 @@
     "@keystonehq/metamask-airgapped-keyring": "0.2.1",
     "@material-ui/core": "^4.11.0",
     "@metamask/contract-metadata": "^1.31.0",
-    "@metamask/controllers": "^22.0.0",
+    "@metamask/controllers": "^23.0.0",
```
Gudahtt, line 110: This update includes a breaking change to the approval controller, which is that an error is now explicitly required when calling `clear`. We call clear once in `metamask-controller.js` on line 537.

ui/store/actions.js
```diff
@@ -1391,6 +1391,19 @@ export function removeCollectible(address, tokenID, dontShowLoadingIndicator) {
...
 
+export function checkAndUpdateCollectiblesOwnershipStatus() {
+  return async (dispatch) => {
+    try {
+      await promisifiedBackground.checkAndUpdateCollectiblesOwnershipStatus();
+    } catch (error) {
+      log.error(error);
+      dispatch(displayWarning(error.message));
+    } finally {
+      await forceUpdateMetamaskState(dispatch);
```
Gudahtt, line 1402: I don't think this step is needed. Background state changes are automatically propagated to the UI. We only need to "force update" the MetaMask state when we need to perform additional steps _after_ the state update has occurred. Since this is the last step here, that doesn't seem to be the case here.

ui/store/actions.js
```diff
@@ -1391,6 +1391,19 @@ export function removeCollectible(address, tokenID, dontShowLoadingIndicator) {
   };
 }
 
+export function checkAndUpdateCollectiblesOwnershipStatus() {
+  return async (dispatch) => {
+    try {
+      await promisifiedBackground.checkAndUpdateCollectiblesOwnershipStatus();
+    } catch (error) {
+      log.error(error);
```
Gudahtt, line 1399: When is this expected to fail? How does this affect the user? The current approach of logging the error and moving on might make sense of we expect this to fail sometimes but in a way that isn't important to tell the user about. But if we expect this to never fail, we should let the error bubble up to Sentry instead. And if it is important to tell the user, we should consider that in the UI design and expose it somewhere for the user.

The `displayWarning` style of showing the user an error is _awful_ and should be considered as deprecated. It places the error in one of many places around the wallet, where it's shown totally devoid of context, and not translated either.

ui/components/app/collectibles-items/collectibles-items.js
```diff
@@ -24,114 +24,149 @@ const width =
   getEnvironmentType() === ENVIRONMENT_TYPE_POPUP
     ? BLOCK_SIZES.ONE_THIRD
     : BLOCK_SIZES.ONE_SIXTH;
-export default function CollectiblesItems({ collections = {} }) {
-  const defaultDropdownState = {};
+
+const PREVIOUSLY_OWNED_KEY = 'previouslyOwned';
```
mcmire, line 28: [nit] Add an extra line here so it's easier to see where the default export of this file starts?

ui/components/app/collectibles-items/collectibles-items.js
```diff
@@ -140,6 +175,26 @@ export default function CollectiblesItems({ collections = {} }) {
 }
 
 CollectiblesItems.propTypes = {
+  previouslyOwnedCollection: PropTypes.shape({
```
mcmire, line 178: Nice!

ui/components/app/collectibles-tab/collectibles-tab.test.js
```diff
@@ -264,15 +268,33 @@ describe('Collectible Items', () => {
     });
   });
   describe('Collectibles options', () => {
-    it('should render a link "Refresh list" when some collectibles are present and collectible auto-detection preference is set to true, which, when clicked calls a method DetectCollectibles', () => {
+    it('should render a link "Refresh list" when some collectibles are present on mainnet and collectible auto-detection preference is set to true, which, when clicked calls methods DetectCollectibles and checkAndUpdateCollectiblesOwnershipStatus', () => {
```
mcmire, line 271: Nice test name!

mcmire (commented):
Only one little thing, but overall looks pretty good to me, other than the things Mark mentioned. I'll wait until those are addressed before approving.

package.json
adonesky1: Do you know what `code` and `message` we should use for the `EthereumRpcError` we pass?

adonesky1: @rekmarks ☝️ do you know?

Gudahtt: That particular `clear` is triggered by the UI closing, which we interpret as the user rejecting the approval. So you can use `ethErrors.provider.userRejectedRequest()`, with no provided message. The default message is perfectly fine.

ui/store/actions.js
adonesky1: we shouldn't surface any failures to the user. It could potentially fail in an unexpected way that we should bubble to sentry: https://github.com/MetaMask/controllers/blob/main/src/assets/CollectiblesController.ts#L1062. Will remove the `displayWarning` but retain the `log.error(error)`

ui/store/actions.js
adonesky1: removed.

ui/components/app/collectibles-items/collectibles-items.js
adonesky1: done here https://github.com/MetaMask/metamask-extension/pull/13110/commits/f18e1eea775cc11d5395071ec8294f2206b93fac

ui/store/actions.js
adonesky1: done here: https://github.com/MetaMask/metamask-extension/pull/13110/commits/f18e1eea775cc11d5395071ec8294f2206b93fac

package.json
adonesky1: added here: https://github.com/MetaMask/metamask-extension/pull/13110/commits/f18e1eea775cc11d5395071ec8294f2206b93fac

ui/components/app/collectibles-tab/collectibles-tab.test.js
adonesky1: 😄  been doing a bunch of insanely long names lately haha

ui/store/actions.js
Gudahtt: Note that this still won't make it to Sentry. Console logs get included as breadcrumbs (i.e. as context for other errors), but they aren't recorded as errors themselves. You'd need to remove the `try { .. } catch` block altogether to let it bubble up to Sentry. Unhandled exceptions/rejections are captured.

ui/store/actions.js
```diff
@@ -1391,6 +1391,16 @@ export function removeCollectible(address, tokenID, dontShowLoadingIndicator) {
...
 
+export function checkAndUpdateCollectiblesOwnershipStatus() {
+  return async () => {
+    try {
+      await promisifiedBackground.checkAndUpdateCollectiblesOwnershipStatus();
+    } catch (error) {
+      log.error(error);
+    }
+  };
+}
```
Gudahtt, line 1402: Since no action is dispatched here, this doesn't need to be treated as an action. It can be a normal function, like many of the other functions in this file that just call out to the background.

```suggestion
export async function checkAndUpdateCollectiblesOwnershipStatus() {
  await promisifiedBackground.checkAndUpdateCollectiblesOwnershipStatus();
}
```

^ That suggestion also removes the `try/catch` block, which ensures any errors get reported to Sentry.

This suggestion is not enough on its own though - it changes how this has to be called. It'll need to be invoked directly rather than passed to `dispatch` now.

ui/components/app/collectibles-tab/collectibles-tab.js
```diff
@@ -38,31 +41,52 @@ export default function CollectiblesTab({ onAddNFT }) {
...
 
   const onEnableAutoDetect = () => {
     history.push(EXPERIMENTAL_ROUTE);
   };
 
+  const onRefresh = () => {
+    if (isMainnet) {
+      dispatch(detectCollectibles());
+    }
+    dispatch(checkAndUpdateCollectiblesOwnershipStatus());
```
Gudahtt, line 80: Here is the change to accompany [this suggestion](checkAndUpdateCollectiblesOwnershipStatus):
```suggestion
    checkAndUpdateCollectiblesOwnershipStatus();
```

We're not calling `await` here, so the result of this call is unhandled. But that seems OK in this case where we don't expect failure, since if it rejects it will still be captured by Sentry.

adonesky1: updated here: https://github.com/MetaMask/metamask-extension/pull/13110/commits/cf5fa0ce01ba5ef3be07cc4fcd8617f27874e141

ui/store/actions.js
adonesky1: updated here: https://github.com/MetaMask/metamask-extension/pull/13110/commits/cf5fa0ce01ba5ef3be07cc4fcd8617f27874e141

adonesky1:
@Gudahtt had to rebase once [this](https://github.com/MetaMask/metamask-extension/pull/13136) merged. Otherwise just addressed your feedback in https://github.com/MetaMask/metamask-extension/pull/13110/commits/a1b5233d74e3937aefa9a68926d4a336a857fe43. Should be ready for another look

Gudahtt:
Thanks! Just confirmed that all of my feedback has been addressed, but I don't have time to review the rest of this PR today. Will return to it in the new year if it is still around by then.

jpuri (approved):

mcmire (approved):
LGTM!

PR #13114, opened 2021-12-16
Snaps settings page

### LINKED ISSUE #11681: Snap Settings Page
We need a settings page for snaps. This page should list all snaps.

For each snap, it should display:

- [x] The snap name
- [x] The snap author
- [x] The place the snap was downloaded from
- [x] The snap install date
- [x] The snap description
- [x] All permissions that the snap has
- [x] Whether or not the snap is currently running
- [x] Whether or not the snap is currently enabled
- [x] Some way to enable/disable the snap.

hmalik88:
Closes: #11681 

Explanation:  Snaps settings page [[FIGMA](https://www.figma.com/file/nBKUQx0btfQiqVf30UA6KD/Flask-MVP?node-id=336%3A6342)] used to interact with the settings of individual snaps. There are two views: a list view under `SnapListTab` and a singular view with `ViewSnap`.  A user should be able to start & disable a snap and also be able to remove it entirely. They should also be able to disconnect a snap from a connected site.

**NOTE:** We do not have a snaps discover page / repository nor do we have snap verification in place (yet), so that part of the design was omitted for now. Some parts of the work were improvised due to the current design not covering all edge cases. Design will be polished in future iterations.

Manual testing steps:  
  - Start storybook
  - Navigate to `SettingsPage` and click on the "snaps" tab
  - Observe the changes
  - Additionally, it is recommended to observe the changes in a flask build: Run `yarn start --build-type flask` to start a dev build


app/_locales/en/messages.json
```diff
@@ -2597,6 +2597,9 @@
   "slow": {
     "message": "Slow"
   },
+  "snapAccess": {
+    "message": "$1 snap has access to:"
```
darkwing, line 2607: Whenever we use a wildcard like `$1`, we add a `description` property to the object which explains what the variable represents.

In this case, I'd make it:

```
"snapAccess": {
  "message": "$1 snap has access to:",
  "description":  "$1 represents the name of the snap"
},
```

ui/components/app/permissions-connect-permission-list/permissions-connect-permission-list.js
```diff
@@ -10,15 +10,15 @@ export default function PermissionsConnectPermissionList({ permissions }) {
...
         leftIcon: 'fas fa-eye',
         label: t('eth_accounts'),
         rightIcon: null,
       },
       ///: BEGIN:ONLY_INCLUDE_IN(flask)
-      snap_confirm: {
+      'snap_confirm': {
         leftIcon: 'fas fa-eye',
-        label: 'snap_confirm',
+        label: t('snap_confirm'),
```
darkwing, line 21: Do we have a localization with the key `snap_confirm`?  These are generally camelcased.

ui/components/app/flask/icons/verified-icon.js
```diff
@@ -0,0 +1,29 @@
...
+    >
+      <path
+        d="M12.8452 7C12.8452 8.48521 12.2482 9.90959 11.1855 10.9598C10.1228 12.01 8.68144 12.6 7.17855 12.6C5.67565 12.6 4.23432 12.01 3.17161 10.9598C2.1089 9.90959 1.51188 8.48521 1.51188 7C1.51188 5.51479 2.1089 4.09041 3.17161 3.0402C4.23432 1.99 5.67565 1.4 7.17855 1.4C7.71688 1.4 8.24105 1.477 8.73688 1.617L9.84897 0.518C9.0273 0.182 8.12772 0 7.17855 0C6.24835 0 5.32726 0.18106 4.46787 0.532843C3.60848 0.884626 2.82762 1.40024 2.16988 2.05025C0.841492 3.36301 0.0952148 5.14348 0.0952148 7C0.0952148 8.85652 0.841492 10.637 2.16988 11.9497C2.82762 12.5998 3.60848 13.1154 4.46787 13.4672C5.32726 13.8189 6.24835 14 7.17855 14C9.05717 14 10.8588 13.2625 12.1872 11.9497C13.5156 10.637 14.2619 8.85652 14.2619 7H12.8452ZM4.28146 5.656L3.28272 6.65L6.47022 9.8L13.5535 2.8L12.5548 1.806L6.47022 7.819L4.28146 5.656Z"
+        fill={color}
+      />
+    </svg>
+  );
+}
+
+VerifiedIcon.defaultProps = {
```
darkwing, line 22: Instead of using `defaultProps`, I think we can simply set a default value in the function signature:

```
export default function VerifiedIcon({ className, color ='#4CD964' }) {
```

ui/pages/settings/snaps-list-tab/snap-list-tab.stories.js
```diff
@@ -0,0 +1,51 @@
+import React, { useState } from 'react';
+import { Provider } from 'react-redux';
+// import { object, boolean, select } from '@storybook/addon-knobs';
```
darkwing, line 3: Do we need this or should we remove?

ui/pages/settings/snaps-list-tab/view-snap/index.scss
```diff
@@ -0,0 +1,30 @@
+div .view-snap-subheader {
+    @media screen and (max-width: $break-small) {
+        display: flex;
+        margin-right: 0px;
+        padding-right: 24px;
+        padding-left: 24px;
```
darkwing, line 6: `padding-inline-start` and `padding-inline-end` would be better for RTL; if there's no padding top or left, we can do `padding: 0 24px;`

ui/pages/settings/snaps-list-tab/view-snap/view-snap.component.js
```diff
@@ -0,0 +1,85 @@
...
+      </Box>
+    </div>
+  );
+}
+
+ViewSnap.propTypes = {
+  snap: PropTypes.object,
+  onRemove: PropTypes.func,
+  onToggle: PropTypes.func,
+  // listRoute: PropTypes.string.isRequired,
```
darkwing, line 82: Might need to be removed

darkwing (changes requested):
Great stuff @hmalik88 !  Just a few comments!

ui/components/app/permissions-connect-permission-list/permissions-connect-permission-list.js
rekmarks: They're generally correct, but we will establish keys for all permissioned RPC methods we add, in the vein of `eth_accounts`: https://github.com/MetaMask/metamask-extension/blob/develop/app/_locales/en/messages.json/#L1041-L1044

ui/pages/settings/flask/snaps-list-tab/snap-list-tab.container.js
```diff
@@ -0,0 +1,63 @@
...
+const mapStateToProps = (state, ownProps) => {
+  const { location, history } = ownProps;
+  const { pathname } = location;
+
+  const viewingSnap = Boolean(pathname.match(SNAPS_VIEW_ROUTE));
+  const pathNameTail = pathname.match(/[^/]+$/u)[0];
+  const envIsPopup = getEnvironmentType() === ENVIRONMENT_TYPE_POPUP;
+  const onClick = (_, snap) => {
+    const route = `${SNAPS_VIEW_ROUTE}/${window.btoa(
+      unescape(encodeURIComponent(snap.name)),
```
shanejonas, line 23: i think`snap.name` should be `id`

ui/pages/settings/flask/snaps-list-tab/snap-list-tab.container.js
```diff
@@ -0,0 +1,63 @@
...
+    onRemove,
+    currentSnap:
+      state.metamask.snaps && viewingSnap
+        ? Object.entries(state.metamask.snaps)
+            .map(([_, value]) => value)
+            .find((snap) => {
+              const decoded = decodeURIComponent(
+                escape(window.atob(pathNameTail)),
+              );
+              return snap.name === decoded;
```
shanejonas, line 42: and this would also match for `snap.id`

ui/components/app/connected-sites-list/index.scss
```diff
@@ -33,6 +33,14 @@
     margin-left: 6px;
   }
 
+  &__disconnect {
+    @include H7;
+
+    cursor: pointer;
+    color: $primary-1;
+    padding-left: 24px;
```
darkwing, line 41: `padding-inline-start: 24px` would be a bit better for RTL

ui/pages/settings/flask/snaps-list-tab/index.scss
```diff
@@ -0,0 +1,13 @@
+.snaps-list-wrapper {
+  width: auto;
+}
+
+.snap-settings-card {
+  margin-top: 8px;
+  margin-bottom: 8px;
```
darkwing, line 7: If there's no margin on the left or right, we can shorthand with `margin: 8px 0`, but no big deal.

.storybook/test-data.js
```diff
@@ -102,6 +102,43 @@ const state = {
       swapsFeatureIsLive: false,
       swapsQuoteRefreshTime: 60000,
     },
+    snaps: {
+      "foo": {
+        name: 'foo',
```
shanejonas, line 107: this will have `.id` and `.proposedName`? @rekmarks

rekmarks: Nope! Ref:

```json
{
  "snapStates": {},
  "snaps": {
    "local:http://localhost:8080/": {
      "enabled": true,
      "id": "local:http://localhost:8080/",
      "initialPermissions": {
        "snap_confirm": {}
      },
      "manifest": {
        "description": "An example MetaMask Snap.",
        "initialPermissions": {
          "snap_confirm": {}
        },
        "manifestVersion": "0.1",
        "proposedName": "MetaMask Example Snap",
        "repository": {
          "type": "git",
          "url": "https://github.com/MetaMask/snaps-skunkworks.git"
        },
        "source": {
          "location": {
            "npm": {
              "filePath": "dist/bundle.js",
              "iconPath": "images/icon.svg",
              "packageName": "@metamask/example-snap",
              "registry": "https://registry.npmjs.org/"
            }
          },
          "shasum": "3lEt0yUu080DwV78neROaAAIQWXukSkMnP4OBhOhBnE="
        },
        "version": "0.6.0"
      },
      "permissionName": "wallet_snap_local:http://localhost:8080/",
      "sourceCode": "(...)",
      "status": "stopped",
      "svgIcon": "<svg>...</svg>",
      "version": "0.6.0"
    }
  }
}
```

rekmarks: And `SubjectMetadata`:

```json
{
  "subjectMetadata": {
    "http://localhost:8080": {
      "extensionId": null,
      "iconUrl": null,
      "name": "Hello, Snaps!",
      "origin": "http://localhost:8080",
      "subjectType": "website"
    },
    "https://metamask.github.io": {
      "extensionId": null,
      "iconUrl": null,
      "name": "Snaps Iframe Execution Environment",
      "origin": "https://metamask.github.io",
      "subjectType": "website"
    },
    "local:http://localhost:8080/": {
      "extensionId": null,
      "iconUrl": null,
      "name": "MetaMask Example Snap",
      "origin": "local:http://localhost:8080/",
      "subjectType": "snap",
      "svgIcon": "<svg>...</svg>",
      "version": "0.6.0"
    }
  }
}
```

ui/pages/settings/flask/snaps-list-tab/view-snap/view-snap.container.js
```diff
@@ -0,0 +1,35 @@
...
+import { withRouter } from 'react-router-dom';
+import { getSubjectMetadata } from '../../../../../selectors';
+import { removePermittedAccount } from '../../../../../store/actions';
+import ViewSnap from './view-snap.component';
+
+const mapStateToProps = (state, ownProps) => {
+  const { snap } = ownProps;
+  const subjectMetadata = getSubjectMetadata(state);
+  const connectedSubjects = [];
+  Object.entries(subjectMetadata).forEach(([_, entry]) => {
```
darkwing, line 12: This could probably be a `reduce` instead of `forEach + connectedSubjects` but no big deal, we can improve that in the future.

ui/pages/settings/flask/snaps-list-tab/view-snap/view-snap.container.js
```diff
@@ -0,0 +1,35 @@
...
+    onDisconnect: (origin, address) => {
+      dispatch(removePermittedAccount(origin, address));
+    },
+  };
+};
+
+export default compose(
+  withRouter,
+  connect(mapStateToProps, mapDispatchToProps),
+)(ViewSnap);
```
darkwing, line 35: I think we underlying functional components we avoid the `container` pattern and use methods like `useSelector` and `useDispatch`, but given our rush here, I think this is fine for now.

darkwing (dismissed):
Lots of nits but none blocking! Well done!

darkwing (dismissed):

ui/pages/settings/flask/snaps-list-tab/view-snap/view-snap.container.js
```diff
@@ -0,0 +1,38 @@
...
+  };
+};
+
+const mapDispatchToProps = (dispatch) => {
+  return {
+    onDisconnect: (origin, address) => {
+      dispatch(removePermittedAccount(origin, address));
+    },
+  };
+};
```
rekmarks, line 33: This is using the wrong action. We will need to remove the Snap permission from the origin in question, the user's address has nothing to do with it.

rekmarks: We need to use the `removePermissionsFor` action. Something like this:

```suggestion
const mapDispatchToProps = (dispatch) => {
  return {
    onDisconnect: (connectedOrigin, snapPermissionName) => {
      dispatch(removePermissionsFor({ [connectedOrigin]: [snapPermissionName] }));
    },
  };
};
```

app/scripts/metamask-controller.js
```diff
@@ -1453,6 +1453,9 @@ export default class MetamaskController extends EventEmitter {
       removeSnapError: this.snapController.removeSnapError.bind(
         this.snapController,
       ),
+      disableSnap: this.snapController.disableSnap.bind(this.snapController),
+      enableSnap: this.snapController.enableSnap.bind(this.snapController),
+      removeSnap: this.snapController.removeSnap.bind(this.snapController),
```
shanejonas, line 1458:
```suggestion
      removeSnap: this.snapController.removeSnap.bind(this.snapController),
      revokePermission: this.permissionController.revokePermission.bind(this.permissionController),
```

shanejonas: we need to expose reveokPermission so the ui can call it

shanejonas: or `snapController.removeAllPermissionsFor` (they both do the same thing, removeAllPermissionFor calls revoke underneath)

rekmarks (commented):
@hmalik88, I just started playing around with this to get a feel for it, and the Snap settings UI looks really, really good. Great work!

The one major thing we have to figure out is what to do with Snap settings in the popup UI, because it is currently visually broken. I think it should be reasonably easy to get the main view with the cards into shape, but we may have to redirect to the fullscreen view for the "See Details" page and make that fit into the popup post launch.

rekmarks (changes requested):
I found another issue, namely that another Snap is showing up in the connected sites list for a Snap.

image

Looking at the PermissionController state, only a single site (`http://localhost:3000`) has the permission for the Snap I'm looking at in the screenshot, and `npm:@metamask/example-snap` only has the `snap_confirm` permission.

rekmarks (dismissed):
I also noticed that, when a Snap is removed from settings, the connected sites of the Snap still retain their permissions for the Snap. All permissions for a Snap should be revoked when it is removed, via `PermissionController.revokePermissionForAllSubjects(snapPermissionName)` in the appropriate action / background API function.

app/scripts/metamask-controller.js
```diff
@@ -1452,6 +1452,12 @@ export default class MetamaskController extends EventEmitter {
...
+      revokePermissionsForAllSubjects: this.permissionController.revokePermissionForAllSubjects.bind(
```
Gudahtt, line 1458: Nit: It might make sense to create a `removeSnap` function here in `metamask-controller.js`, and call `revokePermissionsForAllSubjects` immediately after `removeSnap` is called to remove those snap permissions.

Right now the _only_ time that we want to call this function is after calling a snap. And we _never_ want to remove a snap without removing permissions associated with that snap. Moving these two steps together simplifies the background API and makes it less likely we will accidentally remove a snap without removing the associated permissions.

ui/components/app/connected-sites-list/connected-sites-list.component.js
```diff
@@ -39,11 +40,20 @@ export default class ConnectedSitesList extends Component {
                 {this.getSubjectDisplayName(subject)}
               </span>
             </div>
-            <i
-              className="fas fa-trash-alt connected-sites-list__trash"
-              title={t('disconnect')}
-              onClick={() => onDisconnect(subject.origin)}
-            />
+            {isSnap ? (
```
Gudahtt, line 43: Rather than make the "Disconnect" button look different for snaps v.s. the connected sites page, can we instead change this to display the full word "Disconnect" in both places? Or use the trash can for now and worry about this later? It seems silly for this component to be passed an `isSnap` prop for this sole purpose. Nothing about snaps makes "Disconnect" preferable to a trash can icon.

Gudahtt: Also, this is a pre-existing mistake that we can fix in a separate PR, but we should use clickable elements for clickable things. Anchors and buttons are built-in elements that the browser already understands are meant to be interacted with, so they automatically get focused by keyboards and interpreted as interactive by screen readers and other assistive technologies. Patching `onClick` handlers onto divs makes accessibility unnecessarily complicated.

ui/components/app/flask/icons/verified-icon.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function VerifiedIcon({ className, color = '#4CD964' }) {
```
Gudahtt, line 4: I guess this is unrelated to the rest of the PR? We don't use this icon anywhere yet. Which is good, since we don't have any verified snaps yet.

ui/components/app/flask/icons/verified-icon.js
```diff
@@ -0,0 +1,25 @@
...
+      <path
+        d="M12.8452 7C12.8452 8.48521 12.2482 9.90959 11.1855 10.9598C10.1228 12.01 8.68144 12.6 7.17855 12.6C5.67565 12.6 4.23432 12.01 3.17161 10.9598C2.1089 9.90959 1.51188 8.48521 1.51188 7C1.51188 5.51479 2.1089 4.09041 3.17161 3.0402C4.23432 1.99 5.67565 1.4 7.17855 1.4C7.71688 1.4 8.24105 1.477 8.73688 1.617L9.84897 0.518C9.0273 0.182 8.12772 0 7.17855 0C6.24835 0 5.32726 0.18106 4.46787 0.532843C3.60848 0.884626 2.82762 1.40024 2.16988 2.05025C0.841492 3.36301 0.0952148 5.14348 0.0952148 7C0.0952148 8.85652 0.841492 10.637 2.16988 11.9497C2.82762 12.5998 3.60848 13.1154 4.46787 13.4672C5.32726 13.8189 6.24835 14 7.17855 14C9.05717 14 10.8588 13.2625 12.1872 11.9497C13.5156 10.637 14.2619 8.85652 14.2619 7H12.8452ZM4.28146 5.656L3.28272 6.65L6.47022 9.8L13.5535 2.8L12.5548 1.806L6.47022 7.819L4.28146 5.656Z"
+        fill={color}
+      />
+    </svg>
+  );
+}
+
+VerifiedIcon.propTypes = {
+  className: PropTypes.string,
```
Gudahtt, line 23: Nit: Including JSDoc comments here for each prop type will improve the docs shown in Storybook for this component.

It's a good practice for all components really, but especially for those in Storybook.

ui/components/app/permissions-connect-permission-list/permissions-connect-permission-list.js
```diff
@@ -113,7 +113,7 @@ export default function PermissionsConnectPermissionList({ permissions }) {
           typeof listValue === 'function' ? listValue(permission) : listValue;
 
         return (
-          <div className="permission" key={label}>
+          <div className="permission" key={idx}>
```
Gudahtt, line 116: This seems strictly worse. Why use the index as the key? The key should uniquely identify the entry, so that React can tell the difference between permissions getting re-arranged v.s. their properties being changed. The label was already a unique representation of the permission (or so I thought). The permission name would work as well. But the index does not serve as an adequate key, it is not unique per permission.

ui/pages/settings/flask/snaps-list-tab/view-snap/index.scss
```diff
@@ -0,0 +1,84 @@
...
+.view-snap__remove__button {
+  @media screen and (max-width: $break-small) {
+    align-self: center;
+  }
+}
+
+.view-snap-section {
+  .connected-sites-list__content-row {
+    border-top: none;
+    border-bottom: 1px solid #d6d9dc;
```
Gudahtt, line 78: There is a variable for this color:
```suggestion
    border-bottom: 1px solid $ui-2;
```

ui/pages/settings/flask/snaps-list-tab/view-snap/index.scss
```diff
@@ -0,0 +1,84 @@
...
+.view-snap-toggle-container {
+  @media screen and (max-width: $break-small) {
+    padding-left: 0;
+    margin-left: auto;
+    display: inline-block;
+  }
+}
+
+.settings-page__content-item.view-snap-section {
+  border-bottom: 1px solid #d8d8d8;
```
Gudahtt, line 52: Nit: We should really avoid hard-coding colors into stylesheets. This doesn't appear to be one of the design system colors either. Can we use another color here instead from our set of design system colors? (in `ui/css/design-system/colors.scss`)

ui/pages/settings/flask/snaps-list-tab/index.scss
```diff
@@ -0,0 +1,28 @@
+.snaps-list-wrapper {
```
Gudahtt, line 1: Nit: These class names do not comply with the naming convention this project uses, which is BEM (Block, Element, Modifier). See here for more information: http://getbem.com/

Typically the component is a "block", and each piece inside the component is an "element". So we would have a top-level class for the block, and sub classes within that block called `&__[element-name]`.

ui/pages/settings/flask/snaps-list-tab/snap-list-tab.container.js
```diff
@@ -0,0 +1,64 @@
...
+  const onRemove = () => {
+    history.push(SNAPS_LIST_ROUTE);
+  };
+  return {
+    snaps: state.metamask.snaps,
+    onRemove,
+    currentSnap:
+      state.metamask.snaps && viewingSnap
+        ? Object.entries(state.metamask.snaps)
+            .map(([_, value]) => value)
```
Gudahtt, line 37: Nit: `value` is not a great variable name. `snap` or `snapState` might be better here.

ui/selectors/permissions.js
```diff
@@ -96,6 +97,27 @@ export function getConnectedSubjectsForSelectedAddress(state) {
   return connectedSubjects;
 }
 
+export function getSubjectsWithPermission(state, permissionName) {
+  const subjects = getPermissionSubjects(state);
+
+  const connectedSubjects = [];
+
+  Object.entries(subjects).forEach(([subjectKey, { permissions }]) => {
```
Gudahtt, line 105: Nit: may as well call this parameter `origin`. That is what it is, and would save reassigning it later on line 112.

ui/pages/settings/settings.component.js
```diff
@@ -152,7 +165,7 @@ class SettingsPage extends PureComponent {
     } else if (initialBreadCrumbKey) {
       subheaderText = t(initialBreadCrumbKey);
     } else {
-      subheaderText = t(pathnameI18nKey || 'contacts');
+      subheaderText = t(pathnameI18nKey || 'general');
```
Gudahtt, line 168: What motivated this change?

ui/pages/settings/flask/snaps-list-tab/snap-list-tab.container.js
```diff
@@ -0,0 +1,64 @@
...
+import SnapListTab from './snap-list-tab.component';
+
+const mapStateToProps = (state, ownProps) => {
+  const { location, history } = ownProps;
+  const { pathname } = location;
+
+  const viewingSnap = Boolean(pathname.match(SNAPS_VIEW_ROUTE));
+  const pathNameTail = pathname.match(/[^/]+$/u)[0];
+  const envIsPopup = getEnvironmentType() === ENVIRONMENT_TYPE_POPUP;
+  const onClick = (_, snap) => {
```
Gudahtt, line 21: Well this is highly unconventional :sweat_smile: 

As @darkwing mentioned, this container/component pattern is deprecated and should not be used going forward. But in that pattern, the `mapStateToProps` function was for mapping state to props, as the title implies. These two functions aren't doing that - they don't use state. They can be embedded directly in the component, there is no reason to construct them here.

ui/components/app/flask/icons/verified-icon.js
hmalik88: Yes, I saw it part of the design and decided to just make the icon now, since it technically will be a part of the settings page at some point.

ui/components/app/flask/icons/verified-icon.js
hmalik88: Will do, overlooked that -- thanks!

ui/components/app/permissions-connect-permission-list/permissions-connect-permission-list.js
hmalik88: The reason I changed this to idx was because you can have multiple permissions returning as `UNKOWN_PERMISSION`, using that label as a key doesn't work. I ran into that problem when the key was as label before.

ui/pages/settings/flask/snaps-list-tab/view-snap/index.scss
hmalik88: Will change, thanks!

ui/pages/settings/flask/snaps-list-tab/view-snap/index.scss
hmalik88: Yeah, no problem -- I'll find an alternative.

ui/pages/settings/flask/snaps-list-tab/index.scss
hmalik88: Ah! I'll make those changes.

ui/pages/settings/flask/snaps-list-tab/snap-list-tab.container.js
hmalik88: I agree.

ui/pages/settings/settings.component.js
hmalik88: In the settings page container, there isn't a mapping for `SETTINGS_ROUTE`, I think whomever wrote this code was trying to default the settings route to a certain key -- "contacts" is the wrong key here. If you look in expanded view, the subheaderText shows as "Contacts" for when you first navigate to settings, it should be "General".

ui/pages/settings/flask/snaps-list-tab/view-snap/view-snap.component.js
```diff
@@ -0,0 +1,136 @@
...
+            variant={TYPOGRAPHY.H3}
+            boxProps={{ textAlign: 'center' }}
+          >
+            {snap.manifest.proposedName}
+          </Typography>
+          <Box className="view-snap-pill-toggle-container">
+            <Box className="view-snap-pill-container" paddingLeft={2}>
+              <SnapsAuthorshipPill
+                packageName={snap.manifest.proposedName}
+                url={snap.manifest.repository.url}
```
Gudahtt, line 41: The `repository` field is technically optional. See the spec: https://github.com/MetaMask/specifications/blob/main/snaps/publishing.md
```suggestion
                url={snap.manifest.repository?.url}
```

ui/pages/settings/flask/snaps-list-tab/snap-list-tab.container.js
```diff
@@ -0,0 +1,64 @@
...
+              return snap.id === decoded;
+            })
+        : undefined,
+    viewingSnap,
+    envIsPopup,
+    onClick,
+  };
+};
+const mapDispatchToProps = (dispatch) => ({
+  onToggle: (_, snap) => {
```
Gudahtt, line 51: Nit: rather than ignore the first parameter, why not remove it completely? It doesn't need to be there.

ui/pages/settings/flask/snaps-list-tab/snap-list-tab.container.js
```diff
@@ -0,0 +1,64 @@
...
+const mapStateToProps = (state, ownProps) => {
+  const { location, history } = ownProps;
+  const { pathname } = location;
+
+  const viewingSnap = Boolean(pathname.match(SNAPS_VIEW_ROUTE));
+  const pathNameTail = pathname.match(/[^/]+$/u)[0];
+  const envIsPopup = getEnvironmentType() === ENVIRONMENT_TYPE_POPUP;
+  const onClick = (_, snap) => {
+    const route = `${SNAPS_VIEW_ROUTE}/${window.btoa(
+      unescape(encodeURIComponent(snap.id)),
```
Gudahtt, line 23: Huh. What is going on here? I have never seen `unescape` used before, and apparently it's deprecated.

Why can't we just use `encodeURIComponent(snap.id)` here?

ui/pages/settings/flask/snaps-list-tab/snap-list-tab.component.js
```diff
@@ -0,0 +1,110 @@
...
+              {t('manageSnaps')}
+            </Typography>
+          </Box>
+          <div className="snaps-list-wrapper">
+            {Object.entries(snaps).map(([key, snap]) => {
+              return (
+                <SnapSettingsCard
+                  className="snap-settings-card"
+                  isEnabled={snap.enabled}
+                  dateAdded={new Date().toDateString()}
```
Gudahtt, line 75: This always shows the current date. That's not right. Also this format isn't ideal because it doesn't match the users current locale.

Better to remove the date if we don't know when the snap was added. If we to do know, we should format the date with the `Intl.DateTimeFormat` API, which respects the user's current locale.

ui/pages/settings/flask/snaps-list-tab/snap-list-tab.container.js
Gudahtt: This applies to the `onClick` function defined in `mapStateToProps` as well

ui/components/app/permissions-connect-permission-list/permissions-connect-permission-list.js
Gudahtt: Ahhh I see. Well, using an index silences the warning, but it doesn't fix the problem. The index is what React already falls back to internally.

Could we use the permission name instead maybe? That shouldn't be duplicated.

Also, which permissions were coming up as unknown? :thinking: They are supposed to all be listed here.

ui/pages/settings/flask/snaps-list-tab/snap-list-tab.component.js
Gudahtt: It looks like there is no install date stored in the snaps controller right now, so we can omit this for now.

ui/components/app/permissions-connect-permission-list/permissions-connect-permission-list.js
hmalik88: I think it may have been part of the test-data, hmmm maybe we don't run into that problem in production since we're cognizant of all the permissions we allow...I'll change it to just `permission` then. Thanks!

ui/pages/settings/flask/snaps-list-tab/snap-list-tab.container.js
hmalik88: Hmm not sure, @shanejonas -- is there a specific reason for the use of unescape?

shanejonas: needs escape to handle all cases properly: https://developer.mozilla.org/en-US/docs/Glossary/Base64#the_unicode_problem

ui/pages/settings/flask/snaps-list-tab/view-snap/index.scss
```diff
@@ -0,0 +1,84 @@
...
+  padding-bottom: 18px;
```
georgewrmarshall, line 53: Should adhere to 8px grid unless specific padding is absolutely necessary for layout.
```suggestion
  padding-bottom: 16px;
```

georgewrmarshall (commented):
UI Looks good! Have a request about the Icon. It doesn't look like it's being used yet? Ping me after you've made @Gudahtt's suggested changes and I will approve

ui/pages/settings/flask/snaps-list-tab/snap-list-tab.container.js
Gudahtt: Yeah but why base64 encode it in the first place? `encodeURIComponent` is all we need to make it URL-safe.

ui/pages/settings/flask/snaps-list-tab/snap-list-tab.component.js
```diff
@@ -51,9 +51,9 @@ const SnapListTab = ({
     );
   }
   return (
-    <>
+    <div className="snap-list-tab">
```
Gudahtt, line 54: Nit: This was fine beforehand - the `snap-list-tab` class doesn't have any rules in it, so it doesn't need to be set anywhere.

ui/pages/settings/flask/snaps-list-tab/view-snap/view-snap.component.js
```diff
@@ -0,0 +1,142 @@
...
+            <Button
+              className="view-snap__remove__button"
+              type="danger"
+              css={{
+                maxWidth: '175px',
+              }}
+              onClick={() => {
+                dispatch(removeSnap(snap));
+                history.push(SNAPS_LIST_ROUTE);
+              }}
```
Gudahtt, line 116: Two issues here: first `removeSnap` is async, and we probably want to wait on that before redirecting so that the user doesn't still see their snap listed. But more importantly, `removeSnap` is not an action creator, so it should not be passed to `dispatch`. It's just a regular function.

```suggestion
              onClick={async () => {
                await removeSnap(snap);
                history.push(SNAPS_LIST_ROUTE);
              }}
```

ui/pages/settings/flask/snaps-list-tab/view-snap/view-snap.container.js
```diff
@@ -0,0 +1,58 @@
...
+    history,
+    connectedSubjects,
+  };
+};
+
+const mapDispatchToProps = (dispatch) => {
+  return {
+    onDisconnect: (connectedOrigin, snapPermissionName) => {
+      dispatch(
+        removePermissionsFor({
```
Gudahtt, line 39: Lots of other examples here of calling `dispatch` on things that aren't action creators. `dispatch` isn't needed in any of these cases. These can be called directly from the component actually, rather than passed as props via `mapDispatchToProps`.

ui/pages/settings/flask/snaps-list-tab/view-snap/view-snap.component.js
Gudahtt: Actually... after trying this out, there was still one render of the old removed snap being listed, and it's my fault :facepalm:. The `forceUpdateMetamaskState` call would prevent this, but since that was removed, the state update doesn't propagate to the UI for a few milliseconds, leaving us 1 render cycle with stale data. Oops.

Not a big deal at all, as a user it's so quick it's not confusing in the slightest. Just not ideal.

ui/pages/settings/flask/snaps-list-tab/view-snap/view-snap.container.js
hmalik88: I fixed the others, namely, `enableSnap` and `disableSnap`. Wasn't sure on changing `removePermissionsFor`, doesn't that use `dispatch`?

hmalik88: On second thought, I just decided to play it safe and revert the changes taking out `forceMetamaskUpdateState`.

hmalik88 (commented):
Reviewed @Gudahtt 's commits, looking good! Things are in a working state.

Gudahtt (approved):
LGTM!

georgewrmarshall (approved):

PR #13117, opened 2021-12-17
Added Snap Install page to the UI

### LINKED ISSUE #11682: Snaps Installation UI

ritave:
Closes: #11682 

Screenshot 2021-12-17 at 16 08 34

### DIFF
--- a/ui/components/app/modals/modal.js
+++ b/ui/components/app/modals/modal.js
@@ -30,6 +30,9 @@ import AddToAddressBookModal from './add-to-addressbook-modal';
 import EditApprovalPermission from './edit-approval-permission';
 import NewAccountModal from './new-account-modal';
 import CustomizeNonceModal from './customize-nonce';
+///: BEGIN:ONLY_INCLUDE_IN(flask)
+import SnapInstallWarning from './snap-install-warning';
+///: END:ONLY_INCLUDE_IN
 
 const modalContainerBaseStyle = {
   transform: 'translate3d(-50%, 0, 0px)',
@@ -355,6 +358,21 @@ const MODALS = {
       borderRadius: '8px',
     },
   },
+
+  ///: BEGIN:ONLY_INCLUDE_IN(flask)
+  SNAP_INSTALL: {
+    contents: <SnapInstallWarning />,
+    mobileModalStyle: {
+      ...modalContainerMobileStyle,
+    },
+    laptopModalStyle: {
+      ...modalContainerLaptopStyle,
+    },
+    contentStyle: {
+      borderRadius: '8px',
+    },
+  },
+  ///: END:ONLY_INCLUDE_IN
 
   CANCEL_TRANSACTION: {
     contents: <CancelTransaction />,
--- a/ui/components/app/modals/snap-install-warning/snap-install-warning.js
+++ b/ui/components/app/modals/snap-install-warning/snap-install-warning.js
@@ -1,0 +1,45 @@
+import React, { useCallback, useState } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import { useI18nContext } from '../../../../hooks/useI18nContext';
+import Modal from '../../modal';
+import CheckBox from '../../../ui/check-box/check-box.component';
+import { hideModal } from '../../../../store/actions';
+
+export default function SnapInstallWarning() {
+  const t = useI18nContext();
+  const dispatch = useDispatch();
+
+  const { onSubmit, snapName } = useSelector(
+    (state) => state.appState.modal.modalState.props,
+  );
+
+  const [isConfirmed, setIsConfirmed] = useState(false);
+
+  const onCheckboxClicked = useCallback(() => setIsConfirmed((i) => !i), []);
+  const onCancel = useCallback(() => dispatch(hideModal()), [dispatch]);
+  return (
+    <Modal
+      onSubmit={onSubmit}
+      onCancel={onCancel}
+      submitText={t('approveAndInstall')}
+      cancelText={t('cancel')}
+      submitDisabled={!isConfirmed}
+    >
+      <div className="snap-install-warning">
+        <div className="header">{t('areYouSure')}</div>
+        <div className="subheader">{t('snapInstallWarningCheck')}</div>
+        <div className="checkbox-label">
+          <CheckBox
+            checked={isConfirmed}
+            disabled={false}
+            id="warning-accept"
+            onClick={onCheckboxClicked}
+          />
+          <label htmlFor="warning-accept">
+            {t('snapInstallWarningKeyAccess', [snapName])}
+          </label>
+        </div>
+      </div>
+    </Modal>
+  );
+}
--- a/ui/components/app/modals/snap-install-warning/snap-install-warning.js
+++ b/ui/components/app/modals/snap-install-warning/snap-install-warning.js
@@ -4,6 +4,11 @@ import { useI18nContext } from '../../../../hooks/useI18nContext';
 import Modal from '../../modal';
 import CheckBox from '../../../ui/check-box/check-box.component';
 import { hideModal } from '../../../../store/actions';
+import Typography from '../../../ui/typography/typography';
+import {
+  TYPOGRAPHY,
+  FONT_WEIGHT,
+} from '../../../../helpers/constants/design-system';
 
 export default function SnapInstallWarning() {
   const t = useI18nContext();
@@ -15,7 +20,10 @@ export default function SnapInstallWarning() {
 
   const [isConfirmed, setIsConfirmed] = useState(false);
 
-  const onCheckboxClicked = useCallback(() => setIsConfirmed((i) => !i), []);
+  const onCheckboxClicked = useCallback(
+    () => setIsConfirmed((confirmedState) => !confirmedState),
+    [],
+  );
   const onCancel = useCallback(() => dispatch(hideModal()), [dispatch]);
   return (
     <Modal
@@ -26,12 +34,15 @@ export default function SnapInstallWarning() {
       submitDisabled={!isConfirmed}
     >
       <div className="snap-install-warning">
-        <div className="header">{t('areYouSure')}</div>
-        <div className="subheader">{t('snapInstallWarningCheck')}</div>
+        <Typography variant={TYPOGRAPHY.H4} fontWeight={FONT_WEIGHT.BOLD}>
+          {t('areYouSure')}
+        </Typography>
+        <Typography variant={TYPOGRAPHY.H6} paddingLeft={16} paddingRight={16}>
+          {t('snapInstallWarningCheck')}
+        </Typography>
         <div className="checkbox-label">
           <CheckBox
             checked={isConfirmed}
-            disabled={false}
             id="warning-accept"
             onClick={onCheckboxClicked}
           />
--- a/ui/components/app/permissions-connect-header/permissions-connect-header.component.js
+++ b/ui/components/app/permissions-connect-header/permissions-connect-header.component.js
@@ -1,6 +1,9 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
-import SiteOrigin from '../../ui/site-origin/site-origin';
+import SiteOrigin from '../../ui/site-origin';
+///: BEGIN:ONLY_INCLUDE_IN(flask)
+import SnapsAuthorshipPill from '../flask/snaps-authorship-pill';
+///: END:ONLY_INCLUDE_IN
 
 export default class PermissionsConnectHeader extends Component {
   static propTypes = {
@@ -9,6 +12,9 @@ export default class PermissionsConnectHeader extends Component {
     siteOrigin: PropTypes.string.isRequired,
     headerTitle: PropTypes.node,
     headerText: PropTypes.string,
+    ///: BEGIN:ONLY_INCLUDE_IN(flask)
+    npmPackageName: PropTypes.string,
+    ///: END:ONLY_INCLUDE_IN
   };
 
   static defaultProps = {
@@ -28,11 +34,30 @@ export default class PermissionsConnectHeader extends Component {
   }
 
   render() {
-    const { headerTitle, headerText } = this.props;
+    const {
+      headerTitle,
+      headerText,
+      ///: BEGIN:ONLY_INCLUDE_IN(flask)
+      npmPackageName,
+      ///: END:ONLY_INCLUDE_IN
+    } = this.props;
+    ///: BEGIN:ONLY_INCLUDE_IN(flask)
+    const npmPackageUrl = `https://www.npmjs.com/package/${npmPackageName}`;
+    ///: END:ONLY_INCLUDE_IN
     return (
       <div className="permissions-connect-header">
         {this.renderHeaderIcon()}
         <div className="permissions-connect-header__title">{headerTitle}</div>
+        {
+          ///: BEGIN:ONLY_INCLUDE_IN(flask)
+          npmPackageName ? (
+            <SnapsAuthorshipPill
+              packageName={npmPackageName}
+              url={npmPackageUrl}
+            />
+          ) : null
+          ///: END:ONLY_INCLUDE_IN
+        }
         <div className="permissions-connect-header__subtitle">{headerText}</div>
       </div>
     );
--- a/ui/helpers/utils/transactions.util.js
+++ b/ui/helpers/utils/transactions.util.js
@@ -212,7 +212,7 @@ export function getTransactionTypeTitle(t, type, nativeCurrency = 'ETH') {
       return t('transferFrom');
     }
     case TRANSACTION_TYPES.TOKEN_METHOD_APPROVE: {
-      return t('approveLimit');
+      return t('approve');
     }
     case TRANSACTION_TYPES.SIMPLE_SEND: {
       return t('sendingNativeAsset', [nativeCurrency]);
--- a/ui/helpers/utils/transactions.util.js
+++ b/ui/helpers/utils/transactions.util.js
@@ -212,7 +212,7 @@ export function getTransactionTypeTitle(t, type, nativeCurrency = 'ETH') {
       return t('transferFrom');
     }
     case TRANSACTION_TYPES.TOKEN_METHOD_APPROVE: {
-      return t('approve');
+      return t('approveLimit');
     }
     case TRANSACTION_TYPES.SIMPLE_SEND: {
       return t('sendingNativeAsset', [nativeCurrency]);
--- a/ui/pages/permissions-connect/permissions-connect.component.js
+++ b/ui/pages/permissions-connect/permissions-connect.component.js
@@ -37,6 +37,7 @@ export default class PermissionConnect extends Component {
     snapInstallPath: PropTypes.string.isRequired,
     isSnap: PropTypes.bool.isRequired,
     ///: END:ONLY_INCLUDE_IN
+    totalPages: PropTypes.string.isRequired,
     page: PropTypes.string.isRequired,
     targetSubjectMetadata: PropTypes.shape({
       extensionId: PropTypes.string,
@@ -192,7 +193,7 @@ export default class PermissionConnect extends Component {
 
   renderTopBar() {
     const { redirecting } = this.state;
-    const { page, isRequestingAccounts } = this.props;
+    const { page, isRequestingAccounts, totalPages } = this.props;
     const { t } = this.context;
     return redirecting ? null : (
       <div className="permissions-connect__top-bar">
@@ -207,7 +208,7 @@ export default class PermissionConnect extends Component {
         ) : null}
         {isRequestingAccounts ? (
           <div className="permissions-connect__page-count">
-            {t('xOfY', [page, '2'])}
+            {t('xOfY', [page, totalPages])}
           </div>
         ) : null}
       </div>
--- a/ui/pages/permissions-connect/permissions-connect.component.js
+++ b/ui/pages/permissions-connect/permissions-connect.component.js
@@ -206,11 +206,9 @@ export default class PermissionConnect extends Component {
             {t('back')}
           </div>
         ) : null}
-        {isRequestingAccounts ? (
-          <div className="permissions-connect__page-count">
-            {t('xOfY', [page, totalPages])}
-          </div>
-        ) : null}
+        <div className="permissions-connect__page-count">
+          {t('xOfY', [page, totalPages])}
+        </div>
       </div>
     );
   }
--- a/ui/pages/permissions-connect/permissions-connect.component.js
+++ b/ui/pages/permissions-connect/permissions-connect.component.js
@@ -86,6 +86,7 @@ export default class PermissionConnect extends Component {
 
   componentDidMount() {
     const {
+      connectPath,
       confirmPermissionPath,
       ///: BEGIN:ONLY_INCLUDE_IN(flask)
       snapInstallPath,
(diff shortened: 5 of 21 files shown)


ui/pages/permissions-connect/permissions-connect.container.js
```diff
@@ -64,6 +65,11 @@ const mapStateToProps = (state, ownProps) => {
     }
   }
 
+  const isSnap = Boolean(
+    targetSubjectMetadata &&
+      targetSubjectMetadata.subjectType === SUBJECT_TYPES.SNAP,
```
darkwing, line 70: Nit: We can do `targetSubjectMetadata?.subjectType === SUBJECT_TYPES.SNAP` and remove the `targetSubjectMetadata &&`

ui/pages/permissions-connect/snap-install/index.scss
```diff
@@ -0,0 +1,47 @@
...
+
+  .snap-requests-permission {
+    @include H7;
+
+    padding: 16px;
+  }
+
+  .permissions-connect-permission-list {
+    padding-left: 24px;
+    padding-right: 24px;
```
darkwing, line 17: Do you know if we have padding top or bottom?  If not, we can do `padding: 0 24px;`

ui/pages/permissions-connect/snap-install/index.scss
```diff
@@ -0,0 +1,47 @@
...
+
+  .source-code {
+    @include H7;
+
+    display: flex;
+    color: $ui-4;
+
+    .link {
+      color: $primary-blue;
+      margin-left: 4px;
```
darkwing, line 28: `margin-inline-start: 4px` would be good for RTL

darkwing:
I'm also seeing `TypeError: can't access property "RestrictedMethod", _snapControllers.PermissionType is undefined`. in the error console

ritave:
> I'm also seeing `TypeError: can't access property "RestrictedMethod", _snapControllers.PermissionType is undefined`. in the error console

This is probably related to Snaps branch in current state, not this PR. I'll check it out in a separate issue

ui/components/app/permissions-connect-header/permissions-connect-header.component.js
```diff
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
-import SiteOrigin from '../../ui/site-origin/site-origin';
+import SiteOrigin from '../../ui/site-origin';
+import SnapsAuthorshipPill from '../flask/snaps-authorship-pill';
```
Gudahtt, line 4: This import will fail on non-Flask builds because modules under `flask` directories are excluded. So this will need to be behind a code fence

ui/components/app/modals/modal.js
```diff
@@ -356,6 +359,21 @@ const MODALS = {
     },
   },
 
+  ///: BEGIN:ONLY_INCLUDE_IN(flask)
+  SNAP_INSTALL: {
```
Gudahtt, line 363: For future reference, this entire modal system is deprecated and shouldn't be used for new designs. We'll leave this for now due to our time constraints, but this modal system predates the "React portal" feature, which lets us embed modals directly in the UI next to the code that triggers them, rather than use global state to control them at-a-distance which we had to do previously (because we need them to be rendered at the bottom of the document, so they would be shown above the other content with minimal z-index hacks).

ui/components/app/modals/snap-install-warning/snap-install-warning.js
```diff
@@ -0,0 +1,45 @@
...
+  const t = useI18nContext();
+  const dispatch = useDispatch();
+
+  const { onSubmit, snapName } = useSelector(
+    (state) => state.appState.modal.modalState.props,
+  );
+
+  const [isConfirmed, setIsConfirmed] = useState(false);
+
+  const onCheckboxClicked = useCallback(() => setIsConfirmed((i) => !i), []);
```
Gudahtt, line 18: Nit: We should use a more descriptive variable name here

```suggestion
  const onCheckboxClicked = useCallback(() => setIsConfirmed((confirmedState) => !confirmedState), []);
```

ui/components/app/modals/snap-install-warning/snap-install-warning.js
```diff
@@ -0,0 +1,45 @@
...
+      cancelText={t('cancel')}
+      submitDisabled={!isConfirmed}
+    >
+      <div className="snap-install-warning">
+        <div className="header">{t('areYouSure')}</div>
+        <div className="subheader">{t('snapInstallWarningCheck')}</div>
+        <div className="checkbox-label">
+          <CheckBox
+            checked={isConfirmed}
+            disabled={false}
```
Gudahtt, line 34: Nit: This is the default for this prop, so we can omit it

```suggestion
```

ui/components/app/modals/snap-install-warning/snap-install-warning.js
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    <Modal
+      onSubmit={onSubmit}
+      onCancel={onCancel}
+      submitText={t('approveAndInstall')}
+      cancelText={t('cancel')}
+      submitDisabled={!isConfirmed}
+    >
+      <div className="snap-install-warning">
+        <div className="header">{t('areYouSure')}</div>
```
darkwing, line 29: We should use our `Typography` component here:

```
<Typography variant={TYPOGRAPHY.H7} fontWeight={FONT_WEIGHT.BOLD}
```

ui/components/app/modals/snap-install-warning/snap-install-warning.js
```diff
@@ -0,0 +1,45 @@
...
+        <div className="subheader">{t('snapInstallWarningCheck')}</div>
```
darkwing, line 30: Same here:  

```
<Typography variant={TYPOGRAPHY.H6} paddingLeft={16} paddingRight={16}
```

darkwing (changes requested):

ui/pages/permissions-connect/permissions-connect.container.js
```diff
@@ -54,6 +57,10 @@ const mapStateToProps = (state, ownProps) => {
     subjectType: SUBJECT_TYPES.UNKNOWN,
   };
 
+  ///: BEGIN:ONLY_INCLUDE_IN(flask)
+  const isSnap = targetSubjectMetadata?.subjectType === SUBJECT_TYPES.SNAP;
```
Gudahtt, line 61: Nit: `targetSubjectMetadata` should always be an object (see the `??` operator on the above line), so this `?` isn't needed

```suggestion
  const isSnap = targetSubjectMetadata.subjectType === SUBJECT_TYPES.SNAP;
```

ui/pages/permissions-connect/permissions-connect.container.js
```diff
@@ -68,18 +75,29 @@ const mapStateToProps = (state, ownProps) => {
...
+  ///: END:ONLY_INCLUDE_IN
 
   let page = '';
   if (pathname === connectPath) {
     page = '1';
   } else if (pathname === confirmPermissionPath) {
     page = '2';
+    ///: BEGIN:ONLY_INCLUDE_IN(flask)
+  } else if (pathname === snapInstallPath) {
+    page = '3';
```
Gudahtt, line 89: This doesn't seem right. The total number of pages is still hard-coded as 2 (in `permissions-connect.component.js`) for starters, so the header would show "page 3 / 2" here.

I would expect that the total number of pages would be 1 as a baseline, plus 1 if `eth_accounts` was requested, and plus 1 if a snap was requested that hasn't been installed yet. Then calculating the _current_ page depends on that total. This should only be 3 if `eth_accounts` is also being requested, otherwise this would be 2.

Gudahtt:
How have you been testing this by the way? The "snaps template" website didn't seem to work with this branch.

ritave:
> How have you been testing this by the way? The "snaps template" website didn't seem to work with this branch.

Just tried it again - works for me.

ui/components/app/modals/modal.js
ritave: Do you have any code snippets on how to build modals in the future?

ui/store/actions.js
```diff
@@ -655,6 +655,20 @@ export function updateCustomNonce(value) {
   };
 }
 
+///: BEGIN:ONLY_INCLUDE_IN(flask)
+export function showSnapInstallWarning(snapName, onSubmit) {
+  return (dispatch) => {
+    dispatch(
+      showModal({
+        name: 'SNAP_INSTALL',
+        onSubmit,
```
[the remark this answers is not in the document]
ritave: This is done by other modals, eg. `ConfirmDeleteNetwork`.

ui/pages/permissions-connect/snap-install/snap-install.js
```diff
@@ -0,0 +1,112 @@
...
+                })
+              }
+            >
+              {t('openSourceCode')}
+            </div>
+          </div>
+        ) : null}
+        <PermissionsConnectFooter />
+        <PageContainerFooter
+          cancelButtonTyype="default"
```
Gudahtt, line 89: Typo:
```suggestion
          cancelButtonType="default"
```

ui/pages/permissions-connect/snap-install/snap-install.js
```diff
@@ -0,0 +1,112 @@
...
+
+  return (
+    <div className="page-container snap-install">
+      <div className="headers">
+        <PermissionConnectHeader
+          icon={targetSubjectMetadata.iconUrl}
+          iconName={targetSubjectMetadata.name}
+          headerTitle={t('snapInstall')}
+          headerText={null} // TODO(ritave): Add header text when snaps support description
+          siteOrigin={targetSubjectMetadata.name}
```
Gudahtt, line 63: The closest thing we have to an origin is the npm ID I think? The name certainly seems like the wrong thing to show here.

ui/pages/permissions-connect/snap-install/snap-install.js
```diff
@@ -0,0 +1,112 @@
...
+          </div>
+        ) : null}
+        <PermissionsConnectFooter />
+        <PageContainerFooter
+          cancelButtonTyype="default"
+          onCancel={onCancel}
+          cancelText={t('cancel')}
+          onSubmit={shouldShowWarning ? showWarning : onSubmit}
+          submitText={t('approveAndInstall')}
+          buttonsSizeLarge={false}
```
Gudahtt, line 94: This is a typo as well (the prop is `buttonSizeLarge`, but also `false` is the default so we don't need to include this at all

```suggestion
```

ui/pages/permissions-connect/snap-install/snap-install.js
```diff
@@ -0,0 +1,112 @@
...
+          headerText={null} // TODO(ritave): Add header text when snaps support description
+          siteOrigin={targetSubjectMetadata.name}
+          npmPackageName={npmId}
+        />
+        <div className="snap-requests-permission">
+          {t('snapRequestsPermission')}
+        </div>
+        <PermissionsConnectPermissionList permissions={request.permissions} />
+      </div>
+      <div className="footers">
```
Gudahtt, line 71: Nit: The format we use for CSS classes is BEM - Block, Element, Modifier. See here for details: http://getbem.com/introduction/

The basic idea is that any component that makes sense on its own, as an independent component, is a "block", and each sub-piece that is a part of a greater whole is an element. And modifiers are for temporary states like "disabled", often not used.

So for the class names in this component to be BEM-compliant, this would be something like `snap-install__footer`.

Not a blocker, these class names work as-is functionally. Just wanted to make a note for future reference at least though, so your class names can be in-line with those used elsewhere in the wallet.

ui/components/app/modals/modal.js
Gudahtt: Yep, we have a dedicated component for this: `Popover`. There are a few examples of how this is used throughout the codebase.

ui/pages/permissions-connect/snap-install/snap-install.js
ritave: This `siteOrigin` is probably not the best name for this property. It's displayed in a Chip as a name of the Snap

ui/pages/permissions-connect/permissions-connect.container.js
```diff
@@ -68,18 +75,30 @@ const mapStateToProps = (state, ownProps) => {
 
   const connectPath = `${CONNECT_ROUTE}/${permissionsRequestId}`;
   const confirmPermissionPath = `${CONNECT_ROUTE}/${permissionsRequestId}${CONNECT_CONFIRM_PERMISSIONS_ROUTE}`;
+  ///: BEGIN:ONLY_INCLUDE_IN(flask)
+  const snapInstallPath = `${CONNECT_ROUTE}/${permissionsRequestId}${CONNECT_SNAP_INSTALL_ROUTE}`;
+  ///: END:ONLY_INCLUDE_IN
 
+  const totalPages = (1 + isRequestingAccounts + isSnap).toString();
```
Gudahtt, line 82: This crashes the UI in a normal build because `isSnap` is not defined. So we need a fence here, for the Flask and non-Flask version of this calculation. Or perhaps it would be simpler to remove the fence around the `isSnap` variable.

Gudahtt: Also, adding a number to a boolean is unintuitive. Maybe it would be simpler for readability to do this the more verbose way.  e.g. something like:

```
const totalPages = 1 + (isRequestingAccounts ? 1 : 0) + (isSnap ? 1 : 0);
```

ui/pages/permissions-connect/snap-install/snap-install.js
Gudahtt: On second thought, I don't think the snap should be shown in the "SiteOrigin" component at all. That design would be inconsistent, and might make it easier to mislead users with misleading snap names. I created a thread about this on Slack.

Gudahtt:
I was able to test this with the "snaps template" website after, I think I must have been using the wrong build type on my last attempt. That template doesn't request key access though, so the warning is not triggered. Do you have any suggestions for testing that warning? Or should I go ahead and clone that template locally and modify it.

darkwing:
From a design/user perspective, `snap_confirm` doesn't really mean anything to me:
Install

I believe we should maybe bold `snap_confirm` and then underneath it, provide a description of what exactly that means, what the extent of the permission does, etc.  Otherwise users find themselves in a spot where it's all or nothing, aka "yolo" or "oh no".

ui/pages/permissions-connect/permissions-connect.container.js
```diff
@@ -68,18 +75,35 @@ const mapStateToProps = (state, ownProps) => {
...
+  ///: BEGIN:ONLY_INCLUDE_IN(flask)
+  totalPages += isSnap;
+  ///: END:ONLY_INCLUDE_IN
+  totalPages = totalPages.toString();
 
   let page = '';
   if (pathname === connectPath) {
     page = '1';
   } else if (pathname === confirmPermissionPath) {
     page = '2';
```
Gudahtt, line 92: If no `eth_accounts` permission is requested, this will be the first page:

```suggestion
    page = isRequestingAccounts ? '2' : '1';
```

ui/pages/permissions-connect/permissions-connect.component.js
```diff
@@ -188,7 +208,7 @@ export default class PermissionConnect extends Component {
         ) : null}
         {isRequestingAccounts ? (
```
Gudahtt, line 209: Why is this hidden if no accounts are requested? :thinking: This results in the page number being hidden for the snaps install flow as well.

Gudahtt: The "back" button should be shown on the snap install screen as well, which would require the `goBack` function being updated to handle either page 2 or 3.

ui/pages/permissions-connect/snap-install/snap-install.js
```diff
@@ -0,0 +1,111 @@
...
+        {targetSubjectMetadata.sourceCode ? (
+          <div className="source-code">
+            <div className="text">{t('areYouDeveloper')}</div>
+            <div
```
Gudahtt, line 75: Nit: we should be using anchors or buttons for clickable elements, rather than divs. Using built-in elements that are interactive from the start makes it easier to ensure the wallet is accessible via keyboard, screen reader, and other accessibility tools.

rekmarks:
@Gudahtt @darkwing @ritave re: the `snap_confirm` text, on the base branch (`snaps`) the permissions descriptions are added correctly so whatever is going on in @darkwing's screenshot shouldn't be happening. Ref:

ui/pages/permissions-connect/permissions-connect.component.js
ritave: The back button is a harder thing to implement - it would require removing already approved permissions, due to time constraints I suggest we leave it without a back button.

ritave:
> @Gudahtt @darkwing @ritave re: the `snap_confirm` text, on the base branch (`snaps`) the permissions descriptions are added correctly so whatever is going on in @darkwing's screenshot shouldn't be happening. Ref:
> 
> 

This shouldn't, and doesn't happen when I test it

Gudahtt (commented):
When I press 'cancel' on the snap install page, the UI crashes with this error:

```
Uncaught TypeError: Cannot convert undefined or null to object
  at Function.keys (<anonymous>)
  at ui-2.js:15311
  at updateMemo (ui-0.js:35031)
  at Object.useMemo (ui-0.js:35511)
  at useMemo (common-5.js:30162)
  at SnapInstall (ui-2.js:15311)
  at renderWithHooks (ui-0.js:34437)
  at updateFunctionComponent (ui-0.js:36524)
  at beginWork$1 (ui-0.js:38353)
  at HTMLUnknownElement.callCallback (ui-0.js:18513)
backend.js:12979 The above error occurred in the <SnapInstall> component:
    in SnapInstall (created by Context.Consumer)
    in Route (created by PermissionConnect)
    in Switch (created by PermissionConnect)
    in div (created by PermissionConnect)
    in PermissionConnect (created by ConnectFunction)
    in ConnectFunction (created by Context.Consumer)
    in Route (created by Authenticated)
    in Authenticated (created by ConnectFunction)
    in ConnectFunction (created by Routes)
    in Switch (created by Routes)
    in div (created by Routes)
    in div (created by Routes)
    in Routes (created by ConnectFunction)
    in ConnectFunction (created by Context.Consumer)
    in withRouter(Connect(Routes)) (created by Index)
    in LegacyI18nProvider (created by Index)
    in I18nProvider (created by Index)
    in LegacyMetaMetricsProvider (created by Index)
    in MetaMetricsProvider (created by Index)
    in LegacyMetaMetricsProvider (created by Index)
    in MetaMetricsProvider (created by Index)
    in Router (created by HashRouter)
    in HashRouter (created by Index)
    in Provider (created by Index)
    in Index

React will try to recreate this component tree from scratch using the error boundary you provided, Index.
```

ui/pages/permissions-connect/permissions-connect.component.js
Gudahtt: I hadn't realized that the snap installation was handled by a separate permission request. I agree that postponing the pagination makes sense for now, including the back functionality.

ritave:
> When I press 'cancel' on the snap install page, the UI crashes with this error:
> 
> ```
> Uncaught TypeError: Cannot convert undefined or null to object
>   at Function.keys (<anonymous>)
>   at ui-2.js:15311
>   at updateMemo (ui-0.js:35031)
>   at Object.useMemo (ui-0.js:35511)
>   at useMemo (common-5.js:30162)
>   at SnapInstall (ui-2.js:15311)
>   at renderWithHooks (ui-0.js:34437)
>   at updateFunctionComponent (ui-0.js:36524)
>   at beginWork$1 (ui-0.js:38353)
>   at HTMLUnknownElement.callCallback (ui-0.js:18513)
> backend.js:12979 The above error occurred in the <SnapInstall> component:
>     in SnapInstall (created by Context.Consumer)
>     in Route (created by PermissionConnect)
>     in Switch (created by PermissionConnect)
>     in div (created by PermissionConnect)
>     in PermissionConnect (created by ConnectFunction)
>     in ConnectFunction (created by Context.Consumer)
>     in Route (created by Authenticated)
>     in Authenticated (created by ConnectFunction)
>     in ConnectFunction (created by Routes)
>     in Switch (created by Routes)
>     in div (created by Routes)
>     in div (created by Routes)
>     in Routes (created by ConnectFunction)
>     in ConnectFunction (created by Context.Consumer)
>     in withRouter(Connect(Routes)) (created by Index)
>     in LegacyI18nProvider (created by Index)
>     in I18nProvider (created by Index)
>     in LegacyMetaMetricsProvider (created by Index)
>     in MetaMetricsProvider (created by Index)
>     in LegacyMetaMetricsProvider (created by Index)
>     in MetaMetricsProvider (created by Index)
>     in Router (created by HashRouter)
>     in HashRouter (created by Index)
>     in Provider (created by Index)
>     in Index
> 
> React will try to recreate this component tree from scratch using the error boundary you provided, Index.
> ```

Fixed

ui/helpers/utils/transactions.util.js
```diff
@@ -212,7 +212,7 @@ export function getTransactionTypeTitle(t, type, nativeCurrency = 'ETH') {
       return t('transferFrom');
     }
     case TRANSACTION_TYPES.TOKEN_METHOD_APPROVE: {
-      return t('approve');
+      return t('approveLimit');
```
Gudahtt, line 215: This change seems unrelated to the snap install page. A few of the locale changes related to approving seem unrelated as well. Merge conflict resolution mistake perhaps?

ritave: It's a fix for snap-confirm.js having wrong button text

ui/pages/permissions-connect/permissions-connect.component.js
```diff
@@ -267,6 +290,29 @@ export default class PermissionConnect extends Component {
...
             />
+            {
+              ///: BEGIN:ONLY_INCLUDE_IN(flask)
+            }
+            <Route
+              path={snapInstallPath}
+              exact
+              render={() => (
+                <SnapInstall
+                  request={permissionsRequest || {}}
```
Gudahtt, line 301: `permissionsRequest` should not be empty :thinking: At least not unless `redirecting` is `true`, in which case this would not be rendered. See the `componentDidUpdate` lifecycle function

ui/helpers/utils/transactions.util.js
Gudahtt: Ahh I see, OK thanks!

ui/pages/permissions-connect/permissions-connect.component.js
ritave: Check PermissionPageContainer route, the same safeguard is there in place.

Gudahtt (approved):
LGTM! A few things to improve still but this seems to work

darkwing (approved):

PR #13136, opened 2021-12-21
maintain nft image aspect ratios and center images in collection view

adonesky1:
Explanation: 
- Add CSS to preserve aspect ratio of image in details view and center (non-square) images in collection view.
- Also reorganize classnames hierarchy for improved semantic legibility.

Before:
Screen Shot 2021-12-21 at 11 19 40 AM
After:
Screen Shot 2021-12-21 at 10 50 09 AM

Before: 
Screen Shot 2021-12-21 at 11 20 08 AM
After:
Screen Shot 2021-12-21 at 10 50 22 AM

### DIFF
--- a/ui/components/app/collectibles-items/collectibles-items.js
+++ b/ui/components/app/collectibles-items/collectibles-items.js
@@ -48,7 +48,7 @@ export default function CollectiblesItems({ collections = {} }) {
             const isExpanded = dropdownState[key];
             return (
               <div
-                className="collectibles-items__item"
+                className="collectibles-items__collection"
                 key={`collection-${index}`}
                 onClick={() => {
                   setDropdownState((_dropdownState) => ({
@@ -62,19 +62,19 @@ export default function CollectiblesItems({ collections = {} }) {
                   display={DISPLAY.FLEX}
                   alignItems={ALIGN_ITEMS.CENTER}
                   justifyContent={JUSTIFY_CONTENT.SPACE_BETWEEN}
-                  className="collectibles-items__item__accordion-title"
+                  className="collectibles-items__collection__accordion-title"
                 >
                   <Box
                     alignItems={ALIGN_ITEMS.CENTER}
-                    className="collectibles-items__item__collection-header"
+                    className="collectibles-items__collection__collection-header"
                   >
                     {collectionImage ? (
                       <img
                         src={collectionImage}
-                        className="collectibles-items__item__collection-image"
+                        className="collectibles-items__collection__collection-image"
                       />
                     ) : (
-                      <div className="collectibles-items__item__collection-image-alt">
+                      <div className="collectibles-items__collection__collection-image-alt">
                         {collectionName[0]}
                       </div>
                     )}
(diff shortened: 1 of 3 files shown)


darkwing (dismissed):

ui/components/app/collectibles-items/collectibles-items.js
```diff
@@ -62,19 +62,19 @@ export default function CollectiblesItems({ collections = {} }) {
                   display={DISPLAY.FLEX}
                   alignItems={ALIGN_ITEMS.CENTER}
                   justifyContent={JUSTIFY_CONTENT.SPACE_BETWEEN}
-                  className="collectibles-items__item__accordion-title"
+                  className="collectibles-items__collection__accordion-title"
```
Gudahtt, line 65: This does not comply with our class name naming convention, which is BEM: http://getbem.com/

The double-underscore separates the "block" from the "element" part of the classname, and the "element" from the "modifier". But nothing here is a modifier. It's not a general purpose indication of "hierarchy". Elements within other elements are still just a part of the block.

e.g. this classname should be something like `collectibles-items__collection-accordian-title`

adonesky1: TBH I've always been a bit confused about how to handle what appear to be nested "blocks". So for instance, what should the `<Box>` element immediately below here with `className="collectibles-items__collection__collection-header"` in this PR be, following your recommendation? It is nested within `collectibles-items__collection-accordian-title` so `className="collectibles-items__collection-accordian-title__header"`? or `collectibles-items__collection-header`?

adonesky1: @Gudahtt @jpuri ☝️

Gudahtt: The BEM way would be to classify every element as either a "Block" or an "element". The difference being that a "block" makes sense on its own, whereas an "element" has no purpose/function except as part of the block. For the most part, for us,  "blocks" are components and anything inside is an "element". 

The nesting shouldn't be considered in the naming really. BEM isn't hierarchical beyond the 3 levels it uses. Anything further is flattened. So that example would be `collectibles-items__collection-header` - it's the collection header, in the collectibles-items block.

If that name isn't sufficient to identify the element (i.e. if there are multiple "collection headers" in this block), the name can be updated to something more clear, but in this case that name seems fine on its own.

adonesky1: Ok. Thanks @Gudahtt for clarifying!

adonesky1: @Gudahtt I believe I've formatted the CSS per BEM schema. Please let me know if further changes are required.

Gudahtt (approved):
LGTM!

ui/components/app/collectibles-items/collectibles-items.js
Gudahtt: Looks good!

darkwing (approved):

PR #13145, opened 2021-12-22
Simplify the `removeSnapError` function

Gudahtt:
The `removeSnapError` function has been changed from an action creator to a normal function, as the only actions dispatched were unnecessary. It was dispatching actions to update the `metamask` state, but the background state will automatically update on its own.

It also uses `async/await` rather than `.then` now, and the callsites have been updated to use `async/await`. Platforms that support async stack traces rely upon `async/await` for that enhancement to errors, so this will result in better stack traces on those platforms.

### DIFF
--- a/ui/store/actions.js
+++ b/ui/store/actions.js
@@ -812,13 +812,8 @@ export function txError(err) {
 }
 
 ///: BEGIN:ONLY_INCLUDE_IN(flask)
-export function removeSnapError(msgData) {
-  return (dispatch) => {
-    return promisifiedBackground
-      .removeSnapError(msgData)
-      .then(() => updateMetamaskStateFromBackground())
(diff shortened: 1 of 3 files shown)


ui/store/actions.js
```diff
@@ -812,13 +812,8 @@ export function txError(err) {
 }
 
 ///: BEGIN:ONLY_INCLUDE_IN(flask)
-export function removeSnapError(msgData) {
-  return (dispatch) => {
-    return promisifiedBackground
-      .removeSnapError(msgData)
-      .then(() => updateMetamaskStateFromBackground())
-      .then((newState) => dispatch(updateMetamaskState(newState)));
-  };
```
shanejonas, line 821: ah i thought the front-end needed to call `updateMetamaskState` after every background call to update the whole ui's state.

shanejonas (approved):

PR #13147, opened 2021-12-22
Dark Mode Part 1: Switch to Using CSS Variables

darkwing:
https://github.com/MetaMask/metamask-extension/pull/13098 was getting a bit large due to the hack week time constraint, so I'm breaking it up.

This PR switches our usage of SASS variables for colors to CSS variables to prove that the design doesn't change.  You'll notice many changes like `$Grey-100` becoming `var(--Grey-100)`.  Important files to review are:

* `ui/css/design-system/colors.scss`
* `ui/css/design-system/deprecated-colors.scss`
* `ui/components/ui/typography/typography.scss`


Gudahtt:
Should we replace the deprecated colors before switching them to variables? Theming won't be practical with such a large set of colors anyway. Or is that planned as part of a future step, after this?

darkwing:
As a future step.  The idea behind this PR is "proving" that we can switch to CSS vars with minimal if any disruption in UI.  I had noticed that many of the colors are no longer used as well; we can address those in the future.

Gudahtt:
That approach makes sense to me! I will review this next year if it's still around by then.

brad-decker:
Much excite, many wow.

jpuri (dismissed):
🙏 thanks for this work 😄

ui/components/ui/color-indicator/color-indicator.scss
```diff
@@ -41,23 +41,23 @@ $sizes: (
...
       &#{$self}--partial-filled #{$self}__inner-circle {
-        background-color: $color;
+        background-color: var($color);
       }
       &#{$self}--filled {
-        background-color: $color;
+        background-color: var($color);
       }
       & #{$self}__icon {
-        color: #{utilities.choose-contrast-color($color)};
```
Gudahtt, line 52: Hmm. I see that we're no longer using this color contrast mixin. This was the sole use, so we may as well remove the mixin if we aren't using ti here.

This is an unfortunate change, as it means icons won't be assigned contrasting colors anymore. Does this always evaluate to black right now in practice? I guess we could solve this later using the theming system without the use of this mixin.

darkwing: Right; unfortunately SASS can't render or work with dynamic values from CSS variables, and the spots I noticed where rendering as black.  

IMO this change is worth it, as CSS variables will make theming miles more extensible moving forward.

Gudahtt (dismissed):
LGTM!

ui/components/ui/button/buttons.scss
```diff
@@ -2,15 +2,15 @@
...
-$hover-secondary: #b0d7f2;
-$hover-default: #b3b3b3;
-$hover-confirm: #0372c3;
-$hover-red: #feb6bf;
-$hover-red-primary: #c72837;
-$hover-orange: #ffd3b5;
-$warning-light-orange: #f8b588;
-
 .button {
+  --hover-secondary: #b0d7f2;
```
mcmire, line 6: Are these new colors or do they match any existing colors?

mcmire (dismissed):
One question but regardless this looks good to me.

ui/components/ui/button/buttons.scss
darkwing: They match the colors cited above.

mcmire (dismissed):

Gudahtt (approved):

mcmire (approved):

PR #13155, opened 2021-12-23
Fix how `version_name` manifest field is used

Gudahtt:
The `version_name` manifest field was being used on Chrome to store the build type. However, Chrome intended this field to be a full representation of the version, for display purposes. This was evident when uploading this version to the Chrome Web Store, because it used `flask` as the entire version.

Instead the `version_name` field now includes the full SemVer version string. The version parsing code within the build script and in the wallet itself have been updated accordingly.

Manual testing steps:  
  - Update the `version` field in `package.json` to include a valid prerelease suffix (e.g. `-beta.0`).
  - Run the build (`yarn start --build-type beta`)
  - See that the version is displayed properly on the "About" page, and in the extension settings "Details" page.

### DIFF
--- a/app/scripts/platforms/extension.js
+++ b/app/scripts/platforms/extension.js
@@ -86,15 +86,12 @@ export default class ExtensionPlatform {
 
     const versionParts = version.split('.');
     if (versionName) {
-      // On Chrome, the build type is stored as `version_name` in the manifest, and the fourth part
-      // of the version is the build version.
-      const buildType = versionName;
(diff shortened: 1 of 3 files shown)


fahdkhan561 (commented):
Simple binary mistakes coded heavily .1 .2 bases can be used like 01 02 03 04 f2.

hmalik88 (approved):

ritave (approved):

PR #13158, opened 2021-12-23
Update the Firefox prerelease version format

Gudahtt:
The Firefox extension version format does not support the version format we use (SemVer), so we have to specially format the extension version to be compatible. The format we chose was `[major].[minor].[patch].[buildType][buildVersion]`. But when we tried to submit a build with a version in that format, it was rejected as invalid for unknown reasons.

The Firefox extension format has been updated to `[major].[minor].[patch][buildType][buildVersion]`. This seems to pass validation.

Manual testing steps:  
  - Update the `version` field of the `package.json` to include a prerelease verison (e.g. `10.8.1-flask.0`).
  - Run `yarn dist`
  - Try to upload the resulting build to Firefox as a new extension (just to see if it passes validation, you don't need to complete the upload)
  - See that the error it shows you is that the addon ID has already been used. If you try before this change, the error would say the version is invalid.

### DIFF
--- a/app/scripts/platforms/extension.js
+++ b/app/scripts/platforms/extension.js
@@ -92,14 +92,16 @@ export default class ExtensionPlatform {
       // On Chrome, a more descriptive representation of the version is stored
       // in the `version_name` field for display purposes.
       return versionName;
-    } else if (versionParts.length === 4) {
+    } else if (versionParts.length !== 3) {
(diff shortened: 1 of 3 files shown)


Gudahtt:
This depends upon #13155

danjm (approved):
LGTM

jpuri (approved):

PR #13194, opened 2022-01-03
Prevent automatic rejection of confirmations

Gudahtt:
Confirmations are now only automatically rejected if a user explicitly closes the notification window. If we close the window programmatically because there are no notifications left to show, nothing gets rejected.

This partially avoids a race condition where a confirmation gets rejected automatically without the user having seen the confirmation first. This could happen if the confirmation was processed just as the notification window was being closed.

It's still possible for a confirmation that the user has never seen to get rejected as a result of the user closing the window. But at least now it's no longer possible for a confirmation to get rejected in this manner after the user resolves the last confirmation in the queue.

If there is a queued confirmation when the popup is automatically closed, the popup is triggered again so that the user is presented with it.

### DIFF
--- a/app/scripts/lib/notification-manager.js
+++ b/app/scripts/lib/notification-manager.js
@@ -23,6 +23,16 @@ export default class NotificationManager extends EventEmitter {
   }
 
   /**
+   * Mark the notification popup as having been automatically closed.
+   *
+   * This lets us differentiate between the cases where we close the
+   * notification popup v.s. when the user closes the popup window directly.
+   */
+  markAsAutomaticallyClosed() {
+    this._popupAutomaticallyClosed = true;
+  }
+
+  /**
    * Either brings an existing MetaMask notification window into focus, or creates a new notification window. New
    * notification windows are given a 'popup' type.
    *
@@ -72,7 +82,10 @@ export default class NotificationManager extends EventEmitter {
   _onWindowClosed(windowId) {
     if (windowId === this._popupId) {
       this._popupId = undefined;
-      this.emit(NOTIFICATION_MANAGER_EVENTS.POPUP_CLOSED);
+      this._popupAutomaticallyClosed = undefined;
+      this.emit(NOTIFICATION_MANAGER_EVENTS.POPUP_CLOSED, {
+        automaticallyClosed: this._popupAutomaticallyClosed,
+      });
     }
   }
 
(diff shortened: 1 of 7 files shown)


Gudahtt:
This depends upon #13202

app/scripts/lib/notification-manager.js
```diff
@@ -72,7 +82,10 @@ export default class NotificationManager extends EventEmitter {
   _onWindowClosed(windowId) {
     if (windowId === this._popupId) {
       this._popupId = undefined;
-      this.emit(NOTIFICATION_MANAGER_EVENTS.POPUP_CLOSED);
+      this._popupAutomaticallyClosed = undefined;
```
rekmarks, line 85: Why not `false`?

rekmarks: Also, doesn't this make it so that `automaticallyClosed` is always `undefined`?

Gudahtt: Oops! Yes, good catch.

`false` works as well I suppose. I left it unassigned by default because when this isn't set, we don't know whether the next "window close" event is automatic or not. The options are "unknown" or "closed automatically", so `undefined` felt more appropriate

ui/pages/home/home.component.js
```diff
@@ -46,6 +46,17 @@ const LEGACY_WEB3_URL =
...
+  isNotification,
+  totalUnapprovedCount,
+  isSigningQRHardwareTransaction,
+}) {
+  return (
+    isNotification &&
+    totalUnapprovedCount === 0 &&
+    !isSigningQRHardwareTransaction
+  );
+}
```
rekmarks, line 59: nit: breathing space
```suggestion
}

```

rekmarks (commented):
Looks good overall!

ui/pages/home/home.component.js
Gudahtt: Done!

app/scripts/lib/notification-manager.js
Gudahtt: Fixed!

Gudahtt:
When testing this with the `snaps` branch, it appeared that the problem was fixed but there was a new problem now: the popup never opened. I realized that it's because if the confirmation was queued while the window was closing, it would never be shown.

I've updated this branch to automatically re-trigger the UI in this circumstance as well. The PR description has been updated accordingly:
>If there is a queued confirmation when the popup is automatically closed, the popup is triggered again so that the user is presented with it.

rekmarks (approved):
LGTM!

ritave (approved):

PR #13196, opened 2022-01-03
EIP-1559 V2: :  Advanced Gas Fee Modal -  Updating the trend arrows

### LINKED ISSUE #13195: Advanced Gas Fee Modal: Updating the trend arrows
- Update the down arrow to point from upper left to lower right.
- Add arrow for level trend, horizontal gray arrow.
- Remove the trend retaining when the trend is level

NiranjanaBinoy:
Fixes: #13195 

Explanation:  

- Updating the down arrow to point from upper left to lower right.
- Adding new arrow for level trend, horizontal gray arrow.
- Removing the trend retaining when the trend is level.

Screen Shot 2022-01-03 at 6 21 21 PM


jpuri (dismissed):

mcmire (commented):
It looks like in the screenshot that there is a bit more space to the left of the "up" and "down" arrows than there is for the "level" arrow. I know that may not actually be the case but I think it's an illusion because when the arrow is rotated then it provides more space. Anyway, what do you think about adding some left-margin on just the level arrow?

darkwing:
I agree with @mcmire with the level arrow.  Side note:
Gear

The gear doesn't appear vertically aligned.

NiranjanaBinoy:
@darkwing it looks aligned when I tried from a new build. Can you verify it once?
Screen Shot 2022-01-06 at 12 21 07 PM

> The gear doesn't appear vertically aligned.

NiranjanaBinoy:
@mcmire I have updated the left margin, can you verify if this looks okay now?
Screen Shot 2022-01-07 at 11 12 05 AM
Screen Shot 2022-01-07 at 11 13 56 AM


> It looks like in the screenshot that there is a bit more space to the left of the "up" and "down" arrows than there is for the "level" arrow. I know that may not actually be the case but I think it's an illusion because when the arrow is rotated then it provides more space. Anyway, what do you think about adding some left-margin on just the level arrow?

mcmire:
@NiranjanaBinoy That's better. It looks like the up and down arrows need to be vertically aligned now, is it possible to do that?

NiranjanaBinoy:
> @NiranjanaBinoy That's better. It looks like the up and down arrows need to be vertically aligned now, is it possible to do that?

@mcmire Does this looks okay?

Screen Shot 2022-01-07 at 5 34 16 PM

mcmire:
@NiranjanaBinoy Yes, that looks better!

mcmire (approved):
Everything looks good to me!

jpuri (approved):

PR #13202, opened 2022-01-04
Remove unnecessary `mounted` state from Home component

Gudahtt:
The `mounted` state was used to derive state from props before the first render of the Home component. Instead this state is now derived in the constructor, which is also run before the first render. This should behave exactly the same, except now we don't need the `mounted` state or the `deriveStateFromProps` function anymore.

The call to `closeCurrentWindow` that was made in `componentDidUpdate` has been moved to the constructor as well. There is no need to delay that call, and this saves us from having to compare current with previous state in that lifecycle function.

Manual testing steps:  
  - All confirmation flows rely upon this logic. I'd suggest testing every confirmation you can think of, and ensuring that the Home screen is rendered correctly in the fullscreen UI when it's open at the same time as the confirmation.

### DIFF
--- a/ui/pages/home/home.component.js
+++ b/ui/pages/home/home.component.js
@@ -97,10 +97,43 @@ export default class Home extends PureComponent {
   };
 
   state = {
-    // eslint-disable-next-line react/no-unused-state
-    mounted: false,
     canShowBlockageNotification: true,
+    closing: false,
+    redirecting: false,
   };
+
+  constructor(props) {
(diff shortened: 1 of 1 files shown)


darkwing:
All confirmations and suggestions worked for me.  Well done!

darkwing (approved):

rekmarks (approved):
The confirmations work on my device. LGTM!

PR #13204, opened 2022-01-04
Ignore 4Byte 502 errors during e2e tests

Gudahtt:
The 4Byte API can sometimes fail during e2e tests with a 502 error. Ideally we would avoid calling it at all during e2e tests, but in the meantime we shouldn't treat this as a reason to fail the e2e test. We have multiple fallbacks for 4Byte, it isn't relied upon by any tests.

Manual testing steps:  
  - Create a test build (`yarn build test`) 
  - Run the "Contract interactions" e2e test suite (`yarn test:e2e:single --leave-running --browser chrome ./test/e2e/tests/contract-interactions.spec.js`)
  - See that it does not fail with a 4Byte error when 4Byte is down (it's down at time of writing).


darkwing (approved):

adonesky1 (approved):
LGTM!

PR #13205, opened 2022-01-04
Add missing permissions middleware hook

rekmarks:
Adds a missing middleware hook for `wallet_requestPermissions` that we failed to add in #12243. Also adds a runtime check that throws an error if any expected hooks are not provided to `createMethodMiddleware`.

### DIFF
--- a/app/scripts/lib/rpc-method-middleware/createMethodMiddleware.js
+++ b/app/scripts/lib/rpc-method-middleware/createMethodMiddleware.js
@@ -15,10 +15,7 @@ const handlerMap = allHandlers.reduce((map, handler) => {
 
 const expectedHookNames = Array.from(
   new Set(
-    Array.from(handlerMap.values()).reduce(
-      (allHooks, { hookNames = {} }) => allHooks.concat(Object.keys(hookNames)),
-      [],
-    ),
+    allHandlers.map(({ hookNames }) => Object.keys(hookNames)).flat(),
   ).values(),
 );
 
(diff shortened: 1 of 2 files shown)


Gudahtt (dismissed):
LGTM!

hmalik88 (dismissed):

hmalik88 (dismissed):

app/scripts/lib/rpc-method-middleware/createMethodMiddleware.js
```diff
@@ -13,6 +13,15 @@ const handlerMap = allHandlers.reduce((map, handler) => {
...
 }, new Map());
 
+const expectedHookNames = Array.from(
+  new Set(
+    Array.from(handlerMap.values()).reduce(
+      (acc, { hookNames = {} }) => acc.concat(Object.keys(hookNames)),
+      [],
+    ),
+  ).values(),
+);
```
Gudahtt, line 23: Nit: Seems a bit silly to take an array of handlers, insert them into a map according to the method names they handle, then take it back out of a back again and deduplicate handlers, when we could use the original handlers array instead

Would something like this be simpler, and equivalent?
```suggestion
const expectedHookNames = Array.from(
  new Set(
    allHandlers.map(({ hookNames }) => Object.keys(hookNames)).flat()
  ).values(),
);
```

Gudahtt (dismissed):
LGTM!

Gudahtt (approved):

hmalik88 (approved):

PR #13211, opened 2022-01-05
Snaps UI fixes

hmalik88:
Explanation:  Made some minor UI improvements per @holantonela's suggestions: https://docs.google.com/document/d/1vQMGaYfmoGwBfkitDkYgKxrr03sWUcGLX2QUxIXD5H0/edit#

Manual testing steps:  
  - Run `yarn start --build-type flask` on the `snaps-ui-fixes` branch
  - Observe changes

### DIFF
--- a/ui/pages/permissions-connect/snap-install/index.scss
+++ b/ui/pages/permissions-connect/snap-install/index.scss
@@ -14,7 +14,7 @@
 
   .permissions-connect-permission-list {
     padding: 0 24px;
-    
+
     .permission {
       padding: 8px 0;
     }
--- a/ui/pages/permissions-connect/snap-install/index.scss
+++ b/ui/pages/permissions-connect/snap-install/index.scss
@@ -14,6 +14,10 @@
 
   .permissions-connect-permission-list {
     padding: 0 24px;
+    
+    .permission {
+      padding: 8px 0;
+    }
   }
 
   .source-code {
@@ -38,6 +42,14 @@
     display: flex;
     flex-direction: column;
     align-items: center;
+  }
+
+  .permissions-connect-footer {
+    padding-bottom: 16px;
+  }
+
+  &__footer--no-source-code {
+    padding-top: 16px;
   }
 }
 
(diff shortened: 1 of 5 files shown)


ritave (approved):

ui/pages/permissions-connect/snap-install/snap-install.js
```diff
@@ -73,21 +73,28 @@ export default function SnapInstall({
...
+                  })
+                }
+              >
+                {t('openSourceCode')}
+              </div>
             </div>
+            <PermissionsConnectFooter />
+          </>
+        ) : (
+          <div className="snap-install__footer--no-source-code">
```
Gudahtt, line 93: Why was this div needed? I don't see anything in the document about this, and it seems incongruous to have a wrapper div for the "no source" case, but not for the "source" case.

ui/pages/permissions-connect/snap-install/index.scss
```diff
@@ -14,6 +14,10 @@
 
   .permissions-connect-permission-list {
     padding: 0 24px;
+
+    .permission {
+      padding: 8px 0;
```
Gudahtt, line 19: Rather than overriding the padding here, can we change it in `.permissions-connect-permission-list .permission` ? I don't think we want the current padding on any version of the permission list.

ui/pages/permissions-connect/snap-install/snap-install.js
hmalik88: I added the wrapper div to introduce some padding on the top of `PermissionsConnectFooter` in the case of no source code, it's not needed when there is source code.

Gudahtt: Ok. I think this would be far simpler to manage in CSS, by setting the padding/margins of the `footers` and `headers` elements, rather than conditionally adding wrapper divs. But this looks like a lot to untangle right now - we can clean this up later. The styles of this component are very unconventional at the moment.

Gudahtt (approved):

PR #13216, opened 2022-01-05
Created "Token details" page

### LINKED ISSUE #12964: Create a new screen/page "Token details"
**User Story**
As a user I want to be able to see/check token details
Figma mockup: https://www.figma.com/file/EUQ1LPxKPU9tga4YnjISn1/Network-specific-assets?node-id=634%3A12028

**Tasks:** 
- Add a new menu item: "Token details" on the tokens menu 
- Create a new screen/page for token details.

A new menu item and token details screens are shown on the Figma mockup below:
image

**Acceptance criteria**

VSaric:
Fixes: #12964

**Explanation:**  Created new screen/page "Token details"

**Screencast:**

https://user-images.githubusercontent.com/92527393/148245261-4010ca9d-5e63-4f93-87f4-684e74658520.mp4

**Screenshot (modal view):**

Screenshot 2022-02-10 at 13 27 04

**Screenshot (expand view):**

Screenshot 2022-02-11 at 10 48 05


mirjanaKukic:
Verified by QA

ui/components/ui/token-details/token-details.component.js
```diff
@@ -0,0 +1,154 @@
+import React, { useCallback, useContext } from 'react';
```
brad-decker, line 1: [REQ] This is actually a page, not a component, and should be placed in the pages folder. It should also drop the .component syntax.

brad-decker (changes requested):

ui/components/ui/token-details/token-details.component.js
```diff
@@ -0,0 +1,154 @@
...
+          margin={[4, 0, 0, 0]}
+          variant={TYPOGRAPHY.H6}
+          color={COLORS.BLACK}
+          onClick={onClose}
+        >
+          {t('tokenDetailsTitle')}
+        </Typography>
+        <Box className="token-details__address">
+          <Typography
+            align="center"
```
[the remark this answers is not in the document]
VSaric: @brad-decker I miss this part, I changed it.
You can see changes here: bb391fbb5cbede0b4ae3a006707c864548afee15
Is it okay now?

ui/components/ui/token-details/token-details.component.js
```diff
@@ -0,0 +1,154 @@
...
+          className="token-details__token-details-title"
+          fontWeight={FONT_WEIGHT.BOLD}
+          margin={[4, 0, 0, 0]}
+          variant={TYPOGRAPHY.H6}
+          color={COLORS.BLACK}
+          onClick={onClose}
+        >
+          {t('tokenDetailsTitle')}
+        </Typography>
+        <Box className="token-details__address">
```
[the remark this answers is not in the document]
VSaric: @brad-decker I added additional props for the Box component.
You can see changes here: bb391fbb5cbede0b4ae3a006707c864548afee15
Is it okay now?

app/_locales/en/messages.json
```diff
@@ -3166,6 +3172,15 @@
   "tokenDecimalFetchFailed": {
     "message": "Token decimal required."
   },
+  "tokenDecimalTitle": {
```
segun, line 3175: A few comments here

1. What's the different between `tokenDetails` and `tokenDetailsTitle`. They seem to show the same message (only capitalisation changed)

2. Do we want to stick with one case type e.g Token Decimal or Token decimal.

ui/pages/token-details/token-details-screen.js
```diff
@@ -0,0 +1,159 @@
...
+            className="token-details__token-address"
+          >
+            {address}
+          </Typography>
+          <Tooltip
+            position="bottom"
+            title={copied ? t('copiedExclamation') : t('copyToClipboard')}
+            containerClassName="token-details__copy-icon"
+          >
+            <button
```
segun, line 92: Should this be Button (capital B)?

brad-decker (changes requested):
Getting there, thanks for doing some more leg work. 

the .container.js syntax is also deprecated, and when it was used it didn't contain UI just the redux container for hooking state/dispatch to props. What I would like to see before this PR can merge is condensing token-details.js, token-details.container.js and token-details-screen.js into ONE component. The former two seem to only do some logic with hooks.... lets just do that logic in the component that uses it.

VSaric:
> Getting there, thanks for doing some more leg work.
> 
> the .container.js syntax is also deprecated, and when it was used it didn't contain UI just the redux container for hooking state/dispatch to props. What I would like to see before this PR can merge is condensing token-details.js, token-details.container.js and token-details-screen.js into ONE component. The former two seem to only do some logic with hooks.... lets just do that logic in the component that uses it.

@brad-decker I added logic with hooks and condensing `token-details.js`, `token-details.container.js` and `token-details-screen.js` into one component (`token-details-page.js`).
You can see changes here: a28aee5e9014c2d1872c083c5e1692440209d540

app/_locales/en/messages.json
VSaric: @segun Yes, we want to stick with one case type, so I deleted `tokenDetailsTitle` and use just `tokenDetails` to display. You can see changed here: a28aee5e9014c2d1872c083c5e1692440209d540

mirjanaKukic:
While on Token Details page and switching from Test network to Main network getting this error (same behaviour when switching from Mainnet to Testnet) : 

image

mirjanaKukic:
Verified by QA.

ui/pages/token-details/token-details-page.test.js
```diff
@@ -0,0 +1,137 @@
...
+        address: '0x6b175474e89094c44da98b954eedeac495271d0f',
+        decimals: 18,
+        image: './images/eth_logo.svg',
+        isERC721: false,
+        symbol: 'ETH',
+      },
+    };
+
+    const store = configureMockStore()(state);
+    const wrapper = mountWithRouter(
```
brad-decker, line 128: All new tests should be written with React Testing Library instead of Enzyme, see the renderWithProvider helper in the render-helpers files. Also this test is just ensuring that the page actually mounts without error? Perhaps a more significant test suite is in order?

ui/pages/token-details/token-details-page.js
```diff
@@ -0,0 +1,229 @@
...
+              data-testid="token-details-page__token-contract-address"
+            >
+              {token.address}
+            </Typography>
+            <Tooltip
+              position="bottom"
+              title={copied ? t('copiedExclamation') : t('copyToClipboard')}
+              containerClassName="token-details__copy-icon"
+            >
+              <button
```
brad-decker, line 144: [REQ] you are using a native button element versus our Button component here, please fix.

ui/pages/token-details/token-details-page.js
```diff
@@ -0,0 +1,229 @@
...
+    if (token) {
+      content = renderTokenDetailsComponent();
+    } else {
+      content = <Redirect to={{ pathname: DEFAULT_ROUTE }} />;
+    }
+
+    return content;
+  };
+
+  return renderTokenDetailsPage();
```
brad-decker, line 228: before the main JSX of this component, check to see if token is defined. If it isn't,:
``` return <Redirect to={{ pathaname: DEFAULT_ROUTE }} />```

otherwise return the JSX. This is cleaner and easier to follow than the function assignments and calls, and might have some optimization wins as React will deal with the different JSX trees internally versus function memory assignment on every render.

**Review by brad-decker — CHANGES_REQUESTED** [review]

**Review by david0xd — DISMISSED** [review]

File: ui/pages/token-details/token-details-page.js [hunk 9]
```diff
@@ -0,0 +1,229 @@
...
+            fontWeight={FONT_WEIGHT.BOLD}
+            data-testid="token-details-page__current-network-title"
+          >
+            {t('network')}
+          </Typography>
+          <Typography
+            variant={TYPOGRAPHY.H7}
+            margin={[1, 0, 0, 0]}
+            color={COLORS.BLACK}
+            data-testid="token-details-page__current-network"
```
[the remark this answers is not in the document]
david0xd: The `data-testid` attributes are added in order to support an efficient way to do e2e testing. At the moment there is no good way to identify and 'pick-up' these elements in the e2e testing suite.

We had a conversation about it here on the Slack: https://consensys.slack.com/archives/C02GH833686/p1643203963098900 and agreed that we're going to use these attributes.

Having these attributes in the reusable components as well, will help with e2e testing in the future.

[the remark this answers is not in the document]
brad-decker: There's the ability to find elements by text, and a number of other resources before we rely on data-test-ids.... @Gudahtt do we now prefer data-test-id over finding elements by css paths and text?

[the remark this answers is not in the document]
Gudahtt: IMO the test ID selectors are preferable to using CSS or ID-based selectors because they are less fragile and error prone. IDs introduced for use in selectors like this can break functionality (if the ID isn't unique), and using CSS selectors can be hazardous if the same class is used elsewhere without your knowledge (I saw a "false positive" e2e test once because of this problem).

Using text as the selector is nice though. I like "what the user sees"-based selectors because it makes the test easier to read. It's more fragile because the text is more likely to change, but the tradeoff is worth it for me. Calling attention to text changes can be useful too.

ui/pages/token-details/token-details-page.js
VSaric: @brad-decker 
Fixed in: fc353df083ca4099424224735ae80831198a6dcd

ui/pages/token-details/token-details-page.js
VSaric: @brad-decker 
Fixed in: fc353df083ca4099424224735ae80831198a6dcd

ui/pages/token-details/token-details-page.test.js
VSaric: @brad-decker 
Fixed in: fc353df083ca4099424224735ae80831198a6dcd

ui/pages/token-details/token-details-page.js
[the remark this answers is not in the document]
danjm: I like using data-test-ids because it is less fragile, so my preferences fall on the other side of the tradeoff Mark describes.. A combination of good test descriptions, well thought out assertions and code comments can help with test readability.

Regarding the point "Calling attention to text changes can be useful too.": we could just write assertions against text when that is important.

@brad-decker Why should we not use data-test-ids?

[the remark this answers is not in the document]
brad-decker: We should use data-testids when we cannot use a selector that selects based on visible elements on the page. Like looking for a Button with the word "Next". If we first try to use text based selectors that work the way our users are using the product then we will get better warning about changes that will impact how our users will use the app. If we use a testid for that same scenario we are essentially saying "what the button says is irrelevant, our users won't be affected by what the button says" -- but as we have learned changing the one orange button in our app to blue was enough to create a viral tweet about how MetaMask may have been hacked.

testids should be used ahead of classNames etc like @Gudahtt pointed out, but no ahead of using the text on the screen as our users will.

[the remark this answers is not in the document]
brad-decker: >  Regarding the point "Calling attention to text changes can be useful too.": we could just write assertions against text when that is important.

we should assume all text is important, and to do so we could simply *try* to find elements by text first and if its impossible fall back to -testids. Considering that the Typography component is specifically used to display Text, this component should probably never need the testid.

ui/pages/token-details/index.scss
```diff
@@ -0,0 +1,50 @@
...
+    font-size: 32px;
+  }
+
+  &__token-address {
+    width: 222px;
+  }
+
+  &__copy-icon {
+    float: right;
+    margin-left: 62px;
```
darkwing, line 38: `margin-inline-start` would be better for RTL support.

ui/pages/token-details/index.scss
```diff
@@ -0,0 +1,50 @@
...
+
+  &__title {
+    text-transform: capitalize;
+  }
+
+  &__closeButton {
+    background: none;
+    float: right;
+    margin-top: -8px;
+    margin-right: -4px;
```
darkwing, line 20: `margin-inline-end` would be better for RTL support.

ui/pages/token-details/index.scss
```diff
@@ -0,0 +1,50 @@
+.token-details {
+  width: 359px;
```
darkwing, line 2: Where does this number come from?

darkwing (changes requested):
Overall looks pretty good!

ui/pages/token-details/token-details-page.js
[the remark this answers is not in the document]
danjm: @Gudahtt @brad-decker So, to confirm, we should establish a standard in our codebase that says: all e2e tests should select elements by text rendered on screen where possible, and by test-ids where that is not possible. ?

ui/pages/token-details/index.scss
VSaric: @darkwing This number come from Figma design: https://www.figma.com/file/EUQ1LPxKPU9tga4YnjISn1/Network-specific-assets?node-id=634%3A12028
Here is screenshots about that property:
Screenshot 2022-02-07 at 13 14 21
Screenshot 2022-02-07 at 13 14 57

P.S. In Figma you can see design only for modal view, but I made the page display on expand view as well.

ui/pages/token-details/index.scss
VSaric: @darkwing Fixed in: ad52edb29d98d72599f0cb307501e19525d57d36

ui/pages/token-details/index.scss
VSaric: @darkwing Fixed in: ad52edb29d98d72599f0cb307501e19525d57d36

ui/pages/token-details/token-details-page.js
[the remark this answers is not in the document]
danjm: Oh, if we are going to talk about e2e test code standards, I'd also like to know if @PeterYinusa has any thoughts on this topic?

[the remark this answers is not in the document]
danjm: @david0xd @VSaric Let's proceed with making changes according to the rule:
- If an element can be selected based on its text content, don't add data-test-ids.
- If an element that needs to be seleccted in e2e tests cannot be selected based on its text content, do add data-test-ids

ui/pages/token-details/token-details-screen.js
VSaric: @segun I'm sorry I didn't answer this request before, I fix it here: https://github.com/MetaMask/metamask-extension/commit/fc353df083ca4099424224735ae80831198a6dcd

ui/pages/token-details/token-details-page.js
```diff
@@ -0,0 +1,199 @@
...
+      <Box marginLeft={5} marginRight={6}>
+        <Typography
+          fontWeight={FONT_WEIGHT.BOLD}
+          margin={[4, 0, 0, 0]}
+          variant={TYPOGRAPHY.H6}
+          color={COLORS.BLACK}
+          className="token-details__title"
+        >
+          {t('tokenDetails')}
+          <button
```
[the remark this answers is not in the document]
VSaric: @segun Fixed in: 6432d927a47e25e8774fe137b0de015ca531bf7f

ui/pages/token-details/index.scss
VSaric: You are right @brad-decker, in this case the popup window is 359 px wide.
I have small fix for expanded view here: 6432d927a47e25e8774fe137b0de015ca531bf7f
Now `Token Details` page is looking and displays good as @segun said. 
Here is screenshot in expanded view:

Screenshot 2022-02-11 at 10 48 05

segun (approved):
LGTM

brad-decker (approved):
LGTM

brad-decker:
@VSaric please rebase and then i'll reapprove.

jpuri (approved):
LGTM

darkwing (approved):
I'll approve under the condition the font sizing and width of copy block gets addressed ASAP

PR #13225, opened 2022-01-05
Suggested edits

rekmarks:
Some suggested edits.

<details>
<summary>Screenshot:</summary>

onboarding-updated-updated-copy

</details>

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -1132,13 +1132,13 @@
     "message": "from",
     "description": "Part of the sentence describing when and where snap was added"
   },
-  "flaskWelcomeDoNotUse": {
-    "message": "we recommend against using this extension",
+  "flaskWelcomeUninstall": {
+    "message": "you should uninstall this extension",
     "description": "This request is shown on the Flask Welcome screen. It is intended for non-developers, and will be bolded."
   },
   "flaskWelcomeWarning1": {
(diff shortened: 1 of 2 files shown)


app/_locales/en/messages.json
```diff
@@ -1132,20 +1132,20 @@
     "message": "from",
     "description": "Part of the sentence describing when and where snap was added"
   },
-  "flaskWelcomeUninstall": {
-    "message": "uninstall this extension immediately",
+  "flaskWelcomeDoNotUse": {
+    "message": "we recommend against using this extension",
```
Gudahtt, line 1136: Why the softer wording? This is more than just a recommendation. This sounds like we're saying that it might be a bad idea. But no, it's definitely a bad idea.

Gudahtt: The POV to consider is that you are an uninformed user following the instructions of a trusted support agent who is actually a phisher. I was hoping the strong wording would jolt some sense into them.

rekmarks: I don't know that that's true; I can think of a variety of technically competent users that aren't developers but may nevertheless want to use Flask despite the risks. That category describes a significant portion of crypto Twitter, for example.

Gudahtt: Discussion resolved on Slack, and we found a compromise wording

PR #13226, opened 2022-01-05
Update Flask support links and home footer

rekmarks:
This PR uses code fencing to establish different support URLs and home page footers depending on the build type. MetaMask Beta already used a different footer and some different support links, and these are now enforced via the new code fencing logic.

Note that the Flask support URLs aren't live yet, but will work as expected when Flask itself is public.

Finally, to facilitate these changes, this PR disables the `prefer-const` ESLint rule during code fence linting, to allow us to use the pattern established in `ui/helpers/constants/common.js`.

### DIFF
--- a/ui/components/app/account-menu/account-menu.component.js
+++ b/ui/components/app/account-menu/account-menu.component.js
@@ -12,7 +12,9 @@ import UserPreferencedCurrencyDisplay from '../user-preferenced-currency-display
 import {
   PRIMARY,
   SUPPORT_LINK,
+  ///: BEGIN:ONLY_INCLUDE_IN(beta,flask)
   SUPPORT_REQUEST_LINK,
+  ///: END:ONLY_INCLUDE_IN
 } from '../../../helpers/constants/common';
 import { KEYRING_TYPES } from '../../../../shared/constants/hardware-wallets';
 import {
@@ -25,8 +27,6 @@ import {
 import TextField from '../../ui/text-field';
 import SearchIcon from '../../ui/search-icon';
 import Button from '../../ui/button';
-
-import { isBeta } from '../../../helpers/utils/build-types';
 
 export function AccountMenuItem(props) {
   const { icon, children, text, subText, className, onClick } = props;
@@ -325,10 +325,10 @@ export default class AccountMenu extends Component {
 
     let supportText = t('support');
     let supportLink = SUPPORT_LINK;
-    if (isBeta()) {
-      supportText = t('needHelpSubmitTicket');
-      supportLink = SUPPORT_REQUEST_LINK;
-    }
+    ///: BEGIN:ONLY_INCLUDE_IN(beta,flask)
+    supportText = t('needHelpSubmitTicket');
+    supportLink = SUPPORT_REQUEST_LINK;
+    ///: END:ONLY_INCLUDE_IN
 
     return (
       <div className="account-menu">
--- a/ui/components/app/account-menu/account-menu.component.js
+++ b/ui/components/app/account-menu/account-menu.component.js
@@ -9,7 +9,11 @@ import { getEnvironmentType } from '../../../../app/scripts/lib/util';
 import Identicon from '../../ui/identicon';
 import SiteIcon from '../../ui/site-icon';
 import UserPreferencedCurrencyDisplay from '../user-preferenced-currency-display';
-import { PRIMARY } from '../../../helpers/constants/common';
+import {
+  PRIMARY,
+  SUPPORT_LINK,
+  SUPPORT_REQUEST_LINK,
+} from '../../../helpers/constants/common';
 import { KEYRING_TYPES } from '../../../../shared/constants/hardware-wallets';
 import {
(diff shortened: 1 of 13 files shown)


ui/helpers/constants/common.js
```diff
@@ -13,3 +13,15 @@ export const GAS_ESTIMATE_TYPES = {
...
   FASTEST: 'FASTEST',
 };
+
+let _supportLink = 'https://support.metamask.io';
+let _supportRequestLink = 'https://metamask.zendesk.com/hc/en-us/requests/new';
+
+///: BEGIN:ONLY_INCLUDE_IN(flask)
+_supportLink = 'https://metamask-flask.zendesk.com/hc';
+_supportRequestLink =
+  'https://metamask-flask.zendesk.com/hc/en-us/requests/new';
```
Gudahtt, line 23: This link does not appear to work :thinking:

Gudahtt: It redirects to some ZenDesk dashboard, if I'm logged in. If not, it asks me to log in. The page is just missing.

Maybe this needs to get enabled on the Flask ZenDesk instance.

rekmarks: Ah yes, I should've noted this in the description. These links exist and will work, but the instance is not open to the public yet!

Gudahtt: Oh, but I was logged in when I tried and there is still no form. Yet I am able to see https://metamask-flask.zendesk.com/hc

It seemed unlikely that I had permission to view one but not the other.

ui/components/app/account-menu/account-menu.component.js
```diff
@@ -320,10 +324,10 @@ export default class AccountMenu extends Component {
     }
 
     let supportText = t('support');
-    let supportLink = 'https://support.metamask.io';
+    let supportLink = SUPPORT_LINK;
     if (isBeta()) {
```
Gudahtt, line 328: This actually might be exactly what we need for Flask as well - another explicit request for feedback.

Maybe we can move this into a code fence, and use `needHelpSubmitTicket` for Beta, but something more general for Flask?

ui/helpers/constants/common.js
Gudahtt: Maybe the form is not accessible until it's open to the public? :shrug: We ought to be able to see it and try it out before it's public though.

ui/pages/home/beta/beta-home-footer.component.js
```diff
@@ -1,21 +1,18 @@
...
-        target="_blank"
-        rel="noopener noreferrer"
-      >
+      <a href={SUPPORT_REQUEST_LINK} target="_blank" rel="noopener noreferrer">
         {t('needHelpSubmitTicket')}
       </a>{' '}
       |{' '}
       <a
-        href="https://community.metamask.io/c/metamask-beta/30"
+        href="https://community.metamask.io/c/metamask-beta"
```
Gudahtt, line 15: Interesting. Not sure what the trailing `/30` was - maybe the category number? Seems to work without it.

rekmarks: I was playing around on the community site and the number seemed to vary with the amount of things displayed on the page? In any event, it does seem inessential to get the user to where we want to go. The category sub-pages appear to appear under their alphabetical names in the URL.

ui/helpers/constants/common.js
rekmarks: I'm 99% certain that the URL is correct and that we will be able to try it out before it goes live. I will confirm with Support that this is the case as well.

ui/components/app/account-menu/account-menu.component.js
rekmarks: Fixed in https://github.com/MetaMask/metamask-extension/pull/13226/commits/6286bfffacb905c16beb9944d1aab091fcbf4b82

Gudahtt: So that commit did this for the Home page footer, but this line is for the account menu. It seems that that hasn't changed yet.

It seems reasonable to make the equivalent change here in the account menu too maybe?

Gudahtt:
From CI build:

>scripts:core:prod:standardEntryPoints: (node:241) UnhandledPromiseRejectionWarning: Error: MetaMask build: Lint errors encountered for transformed file "/home/circleci/project/ui/pages/home/home.component.js":
>
>    no-unused-vars
>    'SUPPORT_LINK' is defined but never used.

rekmarks:
@Gudahtt fixed!

darkwing (dismissed):

ui/components/app/account-menu/account-menu.component.js
rekmarks: Yeah, that makes sense. Change made here: ae63fd6

Gudahtt (dismissed):
LGTM!

darkwing (dismissed):

darkwing (approved):

Gudahtt (approved):

PR #13232, opened 2022-01-06
don't run CLA bot on `l10n_crowdin_action` branch

rickycodes:
since we're having issues getting the status check back from the CLA bot (and we're all scratching our heads) maybe for these PRs we could just not run the CLA bot on that branch? Curious to hear other people's thoughts.

/cc @adonesky1 @Gudahtt

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -9,7 +9,7 @@ jobs:
   CLABot:
     if: |
       (github.event_name == 'pull_request_target' || contains(github.event.comment.html_url, '/pull/')) &&
-      github.ref != 'refs/heads/l10n_crowdin_action'
+      (github.ref != 'refs/heads/l10n_crowdin_action' && github.event.issue.user.login != 'github-actions')
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -9,7 +9,7 @@ jobs:
   CLABot:
     if: |
       (github.event_name == 'pull_request_target' || contains(github.event.comment.html_url, '/pull/')) &&
-      (github.ref != 'refs/heads/l10n_crowdin_action' && github.event.issue.user.login != 'github-actions')
+      !(github.ref == 'refs/heads/l10n_crowdin_action' && github.event.issue.user.login == 'github-actions')
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -7,7 +7,9 @@ on:
 
 jobs:
   CLABot:
-    if: github.event_name == 'pull_request_target' || contains(github.event.comment.html_url, '/pull/')
+    if: |
+      (github.event_name == 'pull_request_target' || contains(github.event.comment.html_url, '/pull/')) &&
+      github.ref != 'refs/heads/l10n_crowdin_action'
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write


mcmire (dismissed):
I'm good with this.

Gudahtt (dismissed):
We should still find out the root problem with the CLA bot action on that PR, but this is a good workaround in the meantime.

.github/workflows/cla.yml
```diff
@@ -7,7 +7,9 @@ on:
 
 jobs:
   CLABot:
-    if: github.event_name == 'pull_request_target' || contains(github.event.comment.html_url, '/pull/')
+    if: |
+      (github.event_name == 'pull_request_target' || contains(github.event.comment.html_url, '/pull/')) &&
+      github.ref != 'refs/heads/l10n_crowdin_action'
```
Gudahtt, line 12: I just dismissed my review because I had the thought - could a third-party contributor name their branch this to skip the CLA bot check? i.e. if someone creates a branch with this name on a fork, would this check pass on their PR?

Gudahtt: Maybe we can also ensure the author is `github-actions` here. Or that it's not a fork. Something like that.

rickycodes:
> We should still find out the root problem with the CLA bot action on that PR
completely agree. I've added the `DO-NOT-MERGE` label in the meantime so I can document this issue and send it through to GitHub support. Once I send that ticket off I'll remove it and merge.

.github/workflows/cla.yml
rickycodes: yes. I believe that is one small caveat. if a third party contributor opens a PR from a branch `l10n_crowdin_action` this check will not run. we could just keep an eye out for that? alternatively there might be a better way to lock that branch down? but it's a good point. idk if there's much we can do to control branch names on forks :shrug:

adonesky1 (dismissed):
LGTM!

rickycodes:
I've submitted a ticket to GitHub. Removing the label now.

.github/workflows/cla.yml
```diff
@@ -7,7 +7,9 @@ on:
 
 jobs:
   CLABot:
-    if: github.event_name == 'pull_request_target' || contains(github.event.comment.html_url, '/pull/')
+    if: |
+      (github.event_name == 'pull_request_target' || contains(github.event.comment.html_url, '/pull/')) &&
+      (github.ref != 'refs/heads/l10n_crowdin_action' && github.event.issue.user.login != 'github-actions')
```
Gudahtt, line 12: I think this should be a OR.

```suggestion
      (github.ref != 'refs/heads/l10n_crowdin_action' || github.event.issue.user.login != 'github-actions')
```

Or we could switch this to a negation, making it easier to read:

```suggestion
      !(github.ref == 'refs/heads/l10n_crowdin_action' && github.event.issue.user.login == 'github-actions')
```

Gudahtt: Right now this would skip the CLA bot for any branches with that name and any PRs created by the GitHub Actions user. So forks would still skip this.

Gudahtt:
Just had a thought. The CLA bot status check is still a required status check. That's what's blocking the merge of that translations PR. If we stop the bot from running, it'll still be blocked because of that missing status.

Gudahtt:
We need to return a status of OK in this circumstance, rather than run nothing at all. There is an example of that here: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks

adonesky1 (dismissed):

Gudahtt:
Converted this to draft until we're sure this works

rickycodes:
closing this now that I've confirmed using a personal access token works for this.

PR #13234, opened 2022-01-06
Remove dupe Prettier config from ESLint config

mcmire:
The ESLint config for the extension explicitly includes support for
Prettier. However, this is already being provided by our global ESLint
config (`@metamask/eslint-config`). Therefore there is no need to
include it here. In fact, this is causing weird issues where the `curly`
option is getting overridden somehow. After this change, these syntaxes
are invalid:

``` javascript
if (foo) return;
```

``` javascript
if (foo) return 'bar';
```

---

Related to https://github.com/MetaMask/metamask-extension/pull/12578.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -39,13 +39,9 @@ module.exports = {
     'storybook-build/**',
   ],
 
-  extends: [
-    '@metamask/eslint-config',
-    '@metamask/eslint-config-nodejs',
-    'prettier',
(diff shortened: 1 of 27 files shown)


darkwing:
Yes. Yes. Yes.

darkwing (approved):

Gudahtt (approved):
LGTM!

PR #13238, opened 2022-01-07
EIP-1559 V2: Use default gas preferences of user for all transactions

### LINKED ISSUE #12587: Use default gas preferences of user for all transactions
As part of code changes for issue https://github.com/MetaMask/metamask-extension/issues/12447 default gas settings are saved for user.

If a user has these defaults configured we need to ensure that the defaults are always used as a new transaction comes into metamask, user can then change the gas selection is user does not wants to use default.

We need to update initialisation of `estimateToUse` here: https://github.com/MetaMask/metamask-extension/blob/develop/ui/hooks/gasFeeInput/useGasFeeInputs.js#L78

and also `isUsingDappSuggestedGasFees` here: https://github.com/MetaMask/metamask-extension/pull/12567/files#diff-34d0d466a71686f66f2b6fd8cad9426791bb1518008487a18497d2c018234394R100

NiranjanaBinoy:
Fixes: #12587 

Explanation:
- Removing the multiplier from the max base fee and related changes
- When the default values for advancedGasFee are saved by selecting the checkbox in the <AdvancedGasFeePopover/> the Advanced gas settings a.k.a custom gas settings have to be selected by default for all future transactions.

https://user-images.githubusercontent.com/43930900/148617833-04aa8cfa-4995-4bb5-8598-84876d964277.mov


jpuri (dismissed):

app/_locales/en/messages.json
```diff
@@ -1293,6 +1284,9 @@
   "grantedToWithColon": {
     "message": "Granted to:"
   },
+  "gweiInParanthesis": {
```
mcmire, line 1287: typo: `gweiInParanthesis` -> `gweiInParentheses`

mcmire (commented):
One little thing, but this looks good otherwise.

app/_locales/en/messages.json
NiranjanaBinoy: 😅👍

NiranjanaBinoy: done 😅

mcmire (dismissed):

app/_locales/en/messages.json
```diff
@@ -1293,6 +1284,9 @@
   "grantedToWithColon": {
     "message": "Granted to:"
   },
+  "gweiInParentheses": {
+    "message": "(GWEI)"
+  },
```
jpuri, line 1289: I think we can leave out parentheses from internationalised message

jpuri (dismissed):

jpuri:
@NiranjanaBinoy : a fix may be needed here also, plz check https://github.com/MetaMask/metamask-extension/blob/develop/ui/components/app/edit-gas-fee-popover/edit-gas-item/useGasItemFeeDetails.js#L62

jpuri (approved):

jpuri:
Will be useful to have some unit test cases added for advance gas configurations.

mcmire (approved):

PR #13257, opened 2022-01-10
Handle missing eth_accounts permission history in getLastConnectedInfo

rekmarks:
This PR ensures that the `getLastConnectedInfo` selector handles missing `eth_accounts` permission history. Historically, `eth_accounts` was the only permission in existence, and the only way that a permission subject ended up with a permission history in the first place. This will no longer the case as of #11837, and we were perhaps never right to bake in this assumption to begin with.

The first commit adds a unit test for `getLastConnectedInfo`, without modifying the selector. Running this test locally should suffice as manual testing. The second commit modifies the selector and updates the unit test to handle the case of missing `eth_accounts` history.

### DIFF
--- a/ui/selectors/permissions.js
+++ b/ui/selectors/permissions.js
@@ -238,16 +238,23 @@ export function activeTabHasPermissions(state) {
   );
 }
 
+/**
+ * Get the connected accounts history for all origins.
+ *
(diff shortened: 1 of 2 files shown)


Gudahtt (approved):
LGTM!

jpuri (approved):
Nice work 👍

PR #13271, opened 2022-01-10
Add error that redirects users to Import NFT page when they attempt to add an NFT on the Import Token page

adonesky1:
Explanation:  
Adds this validation error when a user attempts to add an NFT via the ERC20 token import flow:
Screen Shot 2022-01-12 at 10 08 18 AM
When clicked, the link sends the user to the correct (Import NFT) page:
Screen Shot 2022-01-12 at 10 10 18 AM

Manual testing steps:  
  - Paste the address of any ERC721 or ERC1155 in the Import Token -> Custom Token Page, `Token Contract Address` field
  - See error `This token is an NFT. Add on the Import NFT page`
  - Click  Import NFT page link. 
  - See that you've been directed to the Import NFT page
  - Repeat with an ERC20 token and see that no error occurs and that you are able to add the token as normal.

### DIFF
--- a/ui/components/ui/page-container/page-container-header/page-container-header.component.js
+++ b/ui/components/ui/page-container/page-container-header/page-container-header.component.js
@@ -49,6 +49,7 @@ export default class PageContainerHeader extends Component {
       onClose && (
         <div
           className="page-container__header-close"
+          data-testid="header-close-button"
           onClick={() => onClose()}
         />
       )
--- a/ui/components/ui/page-container/page-container-header/page-container-header.component.js
+++ b/ui/components/ui/page-container/page-container-header/page-container-header.component.js
@@ -47,9 +47,8 @@ export default class PageContainerHeader extends Component {
 
     return (
       onClose && (
-        <div
+        <button
           className="page-container__header-close"
-          data-testid="header-close-button"
           onClick={() => onClose()}
         />
       )
--- a/ui/components/ui/page-container/page-container-header/page-container-header.component.js
+++ b/ui/components/ui/page-container/page-container-header/page-container-header.component.js
@@ -50,6 +50,7 @@ export default class PageContainerHeader extends Component {
         <button
           className="page-container__header-close"
           onClick={() => onClose()}
+          aria-label="close"
         />
       )
     );
--- a/ui/pages/import-token/import-token.component.js
+++ b/ui/pages/import-token/import-token.component.js
@@ -8,6 +8,7 @@ import {
 } from '../../helpers/utils/util';
 import { tokenInfoGetter } from '../../helpers/utils/token-util';
 import {
+  ADD_COLLECTIBLE_ROUTE,
   CONFIRM_IMPORT_TOKEN_ROUTE,
   EXPERIMENTAL_ROUTE,
 } from '../../helpers/constants/routes';
@@ -46,6 +47,8 @@ class ImportToken extends Component {
     rpcPrefs: PropTypes.object,
     tokenList: PropTypes.object,
     useTokenDetection: PropTypes.bool,
+    getTokenStandardAndDetails: PropTypes.func,
+    selectedAddress: PropTypes.string,
   };
 
   static defaultProps = {
@@ -62,6 +65,7 @@ class ImportToken extends Component {
     customAddressError: null,
     customSymbolError: null,
     customDecimalsError: null,
+    collectibleAddressError: null,
     forceEditSymbol: false,
     symbolAutoFilled: false,
     decimalAutoFilled: false,
@@ -126,13 +130,15 @@ class ImportToken extends Component {
       customAddressError,
       customSymbolError,
       customDecimalsError,
+      collectibleAddressError,
     } = this.state;
 
     return (
       tokenSelectorError ||
       customAddressError ||
       customSymbolError ||
-      customDecimalsError
+      customDecimalsError ||
+      collectibleAddressError
     );
   }
 
@@ -186,11 +192,12 @@ class ImportToken extends Component {
     this.handleCustomDecimalsChange(decimals);
   }
 
-  handleCustomAddressChange(value) {
+  async handleCustomAddressChange(value) {
     const customAddress = value.trim();
     this.setState({
       customAddress,
       customAddressError: null,
+      collectibleAddressError: null,
       tokenSelectorError: null,
       symbolAutoFilled: false,
       decimalAutoFilled: false,
@@ -208,14 +215,42 @@ class ImportToken extends Component {
 
     const isMainnetNetwork = this.props.chainId === '0x1';
 
+    let standard;
+    if (addressIsValid) {
+      try {
+        ({ standard } = await this.props.getTokenStandardAndDetails(
+          standardAddress,
+          this.props.selectedAddress,
+        ));
+      } catch (error) {
+        // ignore
+      }
+    }
+
+    const addressIsEmpty = customAddress.length === 0;
+
     switch (true) {
-      case !addressIsValid:
+      case !addressIsValid && !addressIsEmpty:
         this.setState({
           customAddressError: this.context.t('invalidAddress'),
           customSymbol: '',
           customDecimals: 0,
           customSymbolError: null,
           customDecimalsError: null,
+        });
+
+        break;
+      case standard === 'ERC1155' || standard === 'ERC721':
+        this.setState({
+          collectibleAddressError: this.context.t('collectibleAddressError', [
+            <a
+              className="import-token__collectible-address-error-link"
+              onClick={() => this.props.history.push(ADD_COLLECTIBLE_ROUTE)}
+              key="collectibleAddressError"
+            >
+              {this.context.t('importNFTPage')}
+            </a>,
+          ]),
         });
 
         break;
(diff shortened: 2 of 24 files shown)


ui/pages/import-token/import-token.component.js
```diff
@@ -208,8 +215,22 @@ class ImportToken extends Component {
...
+    let standard;
+    if (addressIsValid) {
+      try {
+        ({ standard } = await this.props.getTokenStandardAndDetails(
+          standardAddress,
+          this.props.selectedAddress,
+        ));
+      } catch (error) {
+        // ignore
+      }
```
jpuri, line 227: should we add a log or may be a metrics event in catch block ?

ui/components/ui/page-container/page-container-header/page-container-header.component.js
```diff
@@ -49,6 +49,7 @@ export default class PageContainerHeader extends Component {
       onClose && (
         <div
           className="page-container__header-close"
+          data-testid="header-close-button"
           onClick={() => onClose()}
```
jpuri, line 53: Above looks wrong it should rather be a button, may be a role can be added for unit test.

ui/pages/import-token/import-token.component.js
```diff
@@ -218,6 +239,20 @@ class ImportToken extends Component {
...
+      case standard === 'ERC1155' || standard === 'ERC721':
+        this.setState({
+          collectibleAddressError: this.context.t('collectibleAddressError', [
+            <a
+              className="import-token__collectible-address-error-link"
+              onClick={() => this.props.history.push(ADD_COLLECTIBLE_ROUTE)}
+              key="collectibleAddressError"
+            >
+              {this.context.t('importNFTPage')}
+            </a>,
```
jpuri, line 252: Anchor tag without href does not look like good practice to me, @darkwing / @georgewrmarshall  can you plz suggest if it will be better to use a button here.

ui/pages/import-token/import-token.component.js
```diff
@@ -242,7 +277,7 @@ class ImportToken extends Component {
 
         break;
       default:
-        if (customAddress !== emptyAddr) {
+        if (customAddress !== emptyAddr && !addressIsEmpty) {
           this.attemptToAutoFillTokenParams(customAddress);
         }
```
jpuri, line 282: May be it will be useful 
```
const addressIsEmpty = customAddress !== emptyAddr && customAddress.length === 0;
```

ui/pages/import-token/import-token.component.js
```diff
@@ -330,10 +366,13 @@ class ImportToken extends Component {
...
           onChange={(e) => this.handleCustomAddressChange(e.target.value)}
-          error={customAddressError || mainnetTokenWarning}
+          error={
+            customAddressError || mainnetTokenWarning || collectibleAddressError
+          }
           fullWidth
           autoFocus
           margin="normal"
+          data-testid="custom-address-input"
         />
```
jpuri, line 376: For unit tests we can probably use getByRole for input box and avoid adding `data-testid`

ui/pages/import-token/import-token.component.js
adonesky1: Yes good call! @brad-decker any thoughts on what the metric event could look like here?

ui/pages/import-token/import-token.component.js
adonesky1: fixed here: https://github.com/MetaMask/metamask-extension/pull/13271/commits/2c635b32bc267b68a43bd0683f4be4b997fa1d58

ui/components/ui/page-container/page-container-header/page-container-header.component.js
adonesky1: fixed here: https://github.com/MetaMask/metamask-extension/pull/13271/commits/2c635b32bc267b68a43bd0683f4be4b997fa1d58

ui/pages/import-token/import-token.component.js
adonesky1: fixed here: https://github.com/MetaMask/metamask-extension/pull/13271/commits/bf875f338df92e9d47aad398f4991bd08154bf5b

ui/pages/import-token/import-token.component.js
adonesky1: Having trouble getting a `<Button>` to play nice as a passed variable. Do we do this any where else? Can't style it to be in line?
Screen Shot 2022-01-12 at 2 59 22 PM

georgewrmarshall: Great discussion! Yes looks like there is a requirement for an inline button component that remains a button and has the style of inline. Currently the [type "link" will change the root html to an `<a>` tag](https://metamask.github.io/metamask-storybook/index.html?path=/docs/ui-components-ui-button-button-stories-js--type-link#type-link) I'll create a ticket. In the mean time you may have to create a custom button component.

Inline button component ticket: #13293

georgewrmarshall (commented):
Looks good! Pending @jpuri's suggestions.

ui/pages/import-token/import-token.component.js
adonesky1: https://github.com/MetaMask/metamask-extension/pull/13271/commits/026e802fbb5054f4f8230b6e19b8981a4315be01

ui/pages/import-token/import-token.component.js
jpuri: data-testid can also be removed

adonesky1: missed this one! Thanks!

ui/pages/import-token/import-token.component.js
adonesky1: @jpuri I met with Kevin, Brad and Dan Miller today to discuss NFT tracking events and this is one that was raised. I think we need to define a few things for the schema first so I am intending to batch all the NFT tracking work in an upcoming PR, but if we can hold off on it for this PR that would be great.

ui/pages/import-token/import-token.component.js
adonesky1: Actually had to remove the href because it was messing with routing. Not sure the right answer here? Is is really to make a custom `Button` component for now?

jpuri: Lets not block this PR for it. But will be nice to address this in a future PR.

jpuri (dismissed):

ui/pages/import-token/import-token.test.js
```diff
@@ -1,110 +1,177 @@
...
+      });
+      expect(historyStub).toHaveBeenCalledWith('/confirm-import-token');
     });
 
-    it('cancels', () => {
-      const cancelButton = wrapper.find('.page-container__header-close');
-      cancelButton.simulate('click');
+    it('cancels out of import token flow', () => {
+      const { getAllByRole } = render();
+      const closeButton = getAllByRole('button')[0];
```
georgewrmarshall, line 145: This seems a little fragile do we not have an aria label on the close button?
`<button aria-label="close" />`
`getByRole('button', { name: 'close' })`

ui/pages/import-token/index.scss
```diff
@@ -59,4 +59,9 @@
     margin-bottom: 16px;
     margin-top: 0;
   }
+
+  &__collectible-address-error-link {
+    color: var(--primary-blue);
+    cursor: pointer;
+  }
```
georgewrmarshall, line 66: Should be blue instead of orange on hover.

Screen Shot 2022-01-17 at 10 17 15 AM

Might have to add this extra css to get the link to hover correctly. I've created an issue for this #13335

```suggestion
  &__collectible-address-error-link {
    color: var(--primary-blue);
    cursor: pointer;

    &:hover {
      color: var(--Blue-400);
    }
  }
```

georgewrmarshall (commented):
Works awesomely! Have created an issue for the link color. I could update that now and try get it into the release instead of adding the custom css. Thoughts?

Also I think this is an existing error but getting this prop type error on this page
Screen Shot 2022-01-18 at 8 19 32 AM

ui/pages/import-token/import-token.test.js
adonesky1: I can add one!

adonesky1:
> I could update that now and try get it into the release instead of adding the custom css. Thoughts?
Up to you. We could add the custom CSS and add that link update later on! But if you want to address that issue now I'm cool with that too.
@georgewrmarshall

georgewrmarshall:
@adonesky1 I think we just add the custom css for now. We will need to address the base stylesheet as a larger ticket I think.

ui/pages/import-token/import-token.test.js
adonesky1: actually the cancel button [already has a `data-testid`](https://github.com/MetaMask/metamask-extension/blob/cb815a91ad8de86f3aaf0e36959d412b441498e9/ui/components/ui/page-container/page-container-footer/page-container-footer.component.js#L54) so going to use that

adonesky1: wait nevermind this test was for the top "x" button. Will add the aria label

adonesky1:
@georgewrmarshall addressed your feedback (thanks!). Ready for another look!

@jpuri just pushed a few commits addressing George's comments!

georgewrmarshall (approved):
Verified trying to add an NFT as a token asset ✅ 
Verified link colors are correct ✅ 
Verified unit tests pass without proptype warnings ✅ 
LGTM! 💯

jpuri (approved):

PR #13244, opened 2022-01-07
Update the copy for the Flask welcome page (#13223)

Gudahtt:
Note: This PR was already reviewed and merged into the `snaps` branch [here](https://github.com/MetaMask/metamask-extension/pull/13223). It should have targeted `develop`, that was my mistake.

The copy for the Flask Welcome page has been updated to better dissuade users who are not the target audience, and to better explain the risks of using Flask.

Manual testing steps:  
  - Create a Flask build (`yarn start --build-type flask`) and look at the first page of onboarding
  
<details>
<summary>Screenshot:</summary>

onboarding-updated-updated-copy

</details>

PR #13272, opened 2022-01-10
Remove unused localized messages

Gudahtt:
These messages were removed from the `en` locale in #13244, but they were not deleted because that branch was not up-to-date when it was merged, and the translations were recent additions (#13206)


ryanml (approved):
lgtm!

darkwing (approved):
Good catch!

PR #13281, opened 2022-01-11
#10168 - Fix to Can't open settings if browser zoom level > 100%

### LINKED ISSUE #10168: Can't open settings if browser zoom level > 100%
**Describe the bug**
If zoom level in the browser is set to more than 100% there is no way to open settings in the menu. On 100% it's fine.
The menu is not scrollable.

- set zoom level in browser settings to 125%
- open metamask menu - you won't see setting in the bottom.

**Screenshots**
Screen: http://joxi.net/l2ZM6p3i7Zn33m

 - OS: MacOS
 - Browser: Tested in Chrome, Edge.
 - MetaMask Version: 8.1.11

Gtonizuka:
Fixes: #
10168

Explanation:  
When browser setting was set > 100% the bottom part of the accounts menu was cut off preventing users from clicking settings tab (I could only reproduce this on Chrome). 
I added a fix (for dropdown extension only) to remove the `position:fixed` part, so now users can scroll all the way down.

Manual testing steps:  
Tested on Chrome (full screen view + extension dropdown view). Created multiple accounts to extend that accounts list.
Tested on Firefox (full screen view + extension dropdown view). Created multiple accounts to extend that accounts list.

  Before (first screenshot is zoom 100% on Chrome and second zoom 125% where the bug happens):

 
before_fix_1
before_fix_2

After:

after_1

after_2

### DIFF
--- a/ui/components/app/account-menu/index.scss
+++ b/ui/components/app/account-menu/index.scss
@@ -12,6 +12,7 @@
   @media screen and (max-width: $break-small) {
     position: initial;
     margin-top: -10px;
+    margin-left: calc(((100vw - 100%) / 2) + 8px);
   }
 
   @media screen and (min-width: $break-large) {
--- a/ui/components/app/account-menu/index.scss
+++ b/ui/components/app/account-menu/index.scss
@@ -10,7 +10,8 @@
   color: var(--white);
 
   @media screen and (max-width: $break-small) {
-    right: calc(((100vw - 100%) / 2) + 8px);
+    position: initial;
+    margin-top: -10px;
   }
 
   @media screen and (min-width: $break-large) {


Gtonizuka:

ui/components/app/account-menu/index.scss
```diff
@@ -10,7 +10,9 @@
   color: var(--white);
 
   @media screen and (max-width: $break-small) {
-    right: calc(((100vw - 100%) / 2) + 8px);
+    position: initial;
+    margin-top: -10px;
+    margin-left: calc(((100vw - 100%) / 2) + 8px);
```
darkwing, line 15: Do we want to apply the same strategy for subsequent media queries for the sake of consistency?

danjm (approved):
LGTM, we could address David's comment here or in a separate PR

ui/components/app/account-menu/index.scss
Gtonizuka: @darkwing I think this is actually better applied to only this media-query and leave all the others as `position:fixed` with the position explicitly set.

Otherwise in the full screen view we run into cases like:
Screenshot 2022-01-12 at 16 56 21

(This is obviously a very edge case and most likely no one shrinkes his browser window that much but still)

brad-decker (approved):

digiwand:
Reverted this PR since it created a new issue: https://github.com/MetaMask/metamask-extension/pull/13459

Follow-up fix here: https://github.com/MetaMask/metamask-extension/pull/13460

PR #13298, opened 2022-01-13
Add deprecation notice to modal folder names

rekmarks:
Rumor has it that the modal is deprecated in favor of the popover component. This PR prefixes all modal folders with the string `DEPRECATED-` to make this much harder to miss. The component imported in `routes.component.js` is also prefixed with `DEPRECATED_`. Our ESLint config for the `camelcase` rule is updated to allow this.

Elsewhere, we have decided that our locally implemented modal component should be deprecated in favor of the popover.


jpuri:
Even renaming these components I think will not be very huge effort.

jpuri:
@rekmarks : is this deprecation in material-ui ? Can you plz share some details.

rekmarks:
@jpuri the engineering team decided at some point in the past that our locally implemented modal component should be deprecated in favor of our popover component. I don't remember why, but @Gudahtt probably does.

Gudahtt:
The old "modal system" being deprecated here was necessary at one point because we wanted the modal to appear above any UI shown at the time, and the simplest way to do that was to put the modal at the end of the DOM. That would let it stack on-top of anything before it without using `z-index`, which can quickly become messy. Because the modal had to be in this specific location, there was no way to co-locate the modal with any component that needed to trigger it. Instead we used global state to manage it. The modal type and all props are set in Redux, so that you can trigger a modal from anywhere in the component hierarchy, and it will appear in the correct place.

This is no longer necessary since React introduced [Portals](https://reactjs.org/docs/portals.html) years ago. Now we can have modals co-located with the components that are responsible for triggering them, and React magically places them in the correct place in the DOM for us. This is dramatically simpler. Now we don't need any global state for managing the modals, and the logic can be isolated to the one place that understands their purpose.

Another problem with the old pattern is that we commonly have to inject functions as props to these modals to handle various button actions. This means the functions have to be set in Redux state. The Redux maintainers [strongly recommend against setting non-serializable data in Redux stores](https://redux.js.org/style-guide/style-guide/#do-not-put-non-serializable-values-in-state-or-actions), because it can break the Redux debugger and it can result in components that don't re-render when the function changes, or that re-render too often even when the function doesn't change. So by using React portals (via the newer `Popover` component) we can migrate away from this practice as well.

.eslintrc.js
```diff
@@ -49,6 +49,13 @@ module.exports = {
   },
 
   rules: {
+    camelcase: [
+      'error',
+      {
+        properties: 'never',
```
digiwand, line 55: thanks for making this explicit!

Makes sense to me considering we do deal with a fair share of external APIs. I don't know the codebase well enough to propose enforcing this on certain files or directories

digiwand (approved):
Good, quick(ish)* solution! 

Tagging along to what @Gudahtt said, it would be ideal to use the JSDoc `@deprecated` to see the inline strikethrough in IDEs ([VSCode supported since 2020](https://code.visualstudio.com/updates/v1_49#_deprecated-tag-support-for-javascript-and-typescript))

This said, adding `@deprecated` to all of the modal declarations here would increase the scope of this PR significantly. There are numerous files, including nested files, that would need to be changed. 

I think this is a great first step! We can add `@deprecated` along the way or in a follow-up PR. It's an extra plus being able to identify deprecated modals on the import lines

rekmarks:
Thanks @digiwand! I will add the deprecation tags in a follow-up if this is merged, cc: @Gudahtt

brad-decker:
@rekmarks I'd like to shepherd this into merging today if possible. Could you do a (hopefully quick) rebase?

rekmarks:
This unfortunately died on the vine.

PR #13300, opened 2022-01-13
use object.values to get array for iteration

brad-decker:
Fixes:  error thrown at initialization because fragments is an object, not an array.

### DIFF
--- a/app/scripts/controllers/metametrics.js
+++ b/app/scripts/controllers/metametrics.js
@@ -123,7 +123,7 @@ export default class MetaMetricsController {
     // a timeout can be specified that will cause an abandoned event to be
     // tracked if the event isn't progressed within that amount of time.
     setInterval(() => {
-      this.store.getState().fragments.forEach((fragment) => {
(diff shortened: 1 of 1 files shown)


jpuri:
Changes look good 👍 
We can also check [map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) for such usecases.

brad-decker:
@jpuri -- Good shout. I wanted to confirm something before I replied. https://consensys.slack.com/archives/GTQAGKY5V/p1642088432109700 -- we could convert it to a Map before iterating, but I think in this case it doesn't serve a lot of value to do so. Basically Maps/Sets are not serializable to JSON so we'd have to intercept the store on update and keep a shallow copy in a Map which would end up causing issues. If at some point the store supports serializable data objects we could re-explore this.

NiranjanaBinoy (approved):

jpuri (approved):

digiwand (approved):

rekmarks (approved):

PR #13301, opened 2022-01-13
Add resolution for follow-redirects

rekmarks:
Adds a resolution for `follow-redirects` to resolve https://github.com/advisories/GHSA-74fj-2j2h-c42q. The dependency graph is `analytics-node#axios#follow-redirects`, and neither parent package has published a fix for this vulnerability.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,7 @@
     "3box/**/libp2p-crypto/node-forge": "^0.10.0",
     "3box/**/libp2p-keychain/node-forge": "^0.10.0",
     "analytics-node/axios": "^0.21.2",
+    "analytics-node/axios/follow-redirects": "^1.14.7",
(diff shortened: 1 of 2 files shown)


jpuri (approved):

hmalik88 (approved):

package.json
```diff
@@ -86,6 +86,7 @@
     "3box/**/libp2p-crypto/node-forge": "^0.10.0",
     "3box/**/libp2p-keychain/node-forge": "^0.10.0",
     "analytics-node/axios": "^0.21.2",
+    "analytics-node/axios/follow-redirects": "^1.14.7",
```
Gudahtt, line 89: This was an in-range update, so a resolution is not required. Just the lockfile update is enough

PR #13303, opened 2022-01-13
Adding e2e test for edit gas fee modal

jpuri:
Adding more e2e test coverage for EIP-1559 V2


test/e2e/helpers.js
```diff
@@ -147,9 +147,46 @@ async function withFixtures(options, testSuite) {
   }
 }
 
+const getWindowHandles = async (driver) => {
+  await driver.waitUntilXWindowHandles(3);
```
mcmire, line 151: What's the significance of 3 window handles? Why not 2 or 4?

test/e2e/helpers.js
```diff
@@ -147,9 +147,46 @@ async function withFixtures(options, testSuite) {
...
+    windowHandles,
+  );
+  const popup = windowHandles.find(
+    (handle) => handle !== extension && handle !== dapp,
+  );
+  return { extension, dapp, popup };
+};
+
+const connectDappWithExtensionPopup = async (driver) => {
+  await driver.openNewPage('http://127.0.0.1:8080/');
```
mcmire, line 166: Any way to tie this to starting the dapp server? It seems like you couldn't call this function without first doing that. Maybe this function should at least take the port number for the dapp server in case that ever changes?

test/e2e/tests/edit-gas-fee.spec.js
```diff
@@ -0,0 +1,280 @@
...
+    await withFixtures(
+      {
+        fixtures: 'eip-1559-v2',
+        ganacheOptions,
+        title: this.test.title,
+      },
+      async ({ driver }) => {
+        await driver.navigate();
+
+        if (process.env.EIP_1559_V2 === '1') {
```
mcmire, line 31: Seems that you could wrap the whole `it` block in this, yeah? Otherwise the server will start for nothing. Same goes for other EIP-1559 v2 tests.

test/e2e/tests/edit-gas-fee.spec.js
```diff
@@ -0,0 +1,280 @@
...
+
+describe('Editing Confirm Transaction', function () {
+  it('allow selecting high, medium, low gas estimates on edit gas fee popover', async function () {
+    const ganacheOptions = {
+      hardfork: 'london',
+      accounts: [
+        {
+          secretKey:
+            '0x7C9529A67102755B7E6102D6D950AC5D5863C98713805CEC576B945B15B71EAC',
+          balance: '0x15af1d78b58c40000',
```
mcmire, line 18: Might be nice to see if we can still represent this as a decimal, otherwise we will have to convert this by some other means.

test/e2e/tests/edit-gas-fee.spec.js
```diff
@@ -0,0 +1,280 @@
...
+const {
+  connectDappWithExtensionPopup,
+  getWindowHandles,
+  largeDelayMs,
+  withFixtures,
+  regularDelayMs,
+} = require('../helpers');
+
+describe('Editing Confirm Transaction', function () {
+  it('allow selecting high, medium, low gas estimates on edit gas fee popover', async function () {
```
mcmire, line 11: [typo] allow -> allows

test/e2e/tests/edit-gas-fee.spec.js
```diff
@@ -0,0 +1,280 @@
...
+            '.transaction-list-item__primary-currency',
+          );
+          assert.equal(txValues.length, 1);
+          assert.ok(/-2.2\s*ETH/u.test(await txValues[0].getText()));
+        }
+      },
+    );
+  });
+
+  it('allow accessing advance gas fee popover from edit gas fee popover', async function () {
```
mcmire, line 94: [typo] allow -> allows, advance -> advanced

test/e2e/tests/edit-gas-fee.spec.js
```diff
@@ -0,0 +1,280 @@
...
+            '.transaction-list-item__primary-currency',
+          );
+          assert.equal(txValues.length, 1);
+          assert.ok(/-2.2\s*ETH/u.test(await txValues[0].getText()));
+        }
+      },
+    );
+  });
+
+  it('should use daap suggested estimates for transaction coming from dapp', async function () {
```
mcmire, line 195: [typo] daap -> dapp

mcmire (commented):
A few suggestions and nits.

test/e2e/tests/edit-gas-fee.spec.js
jpuri: fixed

test/e2e/tests/edit-gas-fee.spec.js
jpuri: fixed

test/e2e/tests/edit-gas-fee.spec.js
jpuri: fixed

test/e2e/tests/edit-gas-fee.spec.js
jpuri: done 👍

test/e2e/tests/edit-gas-fee.spec.js
jpuri: Fixed

test/e2e/helpers.js
jpuri: When using test dapp there are 3 windows:
1. main extension window where user metamask login is done
2. text dapp window
3. extension popup

Wait above ensure that all windows are opened.

test/e2e/helpers.js
jpuri: dapp is initialised [here](https://github.com/MetaMask/metamask-extension/blob/develop/test/e2e/helpers.js#L20). I did the code change to re-use port constant here. Ideally it will be better to combine code to initialise and connect dapp, but that would need a bigger refactor in many test cases.

test/e2e/helpers.js
mcmire: What if the dapp is not started, but this is called? The name of this function sounds like it can be used to get the window handles for _any_ kind of test, regardless of whether it involves the test dapp or not.

darkwing: It might make sense to add a code comment as to what each of the windows represents, if only for maintainability.

darkwing (dismissed):

test/e2e/helpers.js
mcmire: I'm wondering if it would help to change the name of this function to reflect its intended use. Maybe something like `getWindowHandlesIncludingDappServer`? Just thinking that if the dapp server is not started, then `waitUntilXWindowHandles(3)` will literally wait for a window to exist that will never appear.

mcmire (commented):
I'm good with this pending the question I have about `getWindowHandles`.

test/e2e/helpers.js
jpuri: I just updated the PR, thanks for the useful feedbacks 👍

mcmire (approved):
That's better, thanks for the updates! Looks good.

brad-decker (approved):

PR #13304, opened 2022-01-13
Create "inline" type for Button component

### LINKED ISSUE #13293: Create "inline" type for Button component
**Description**

There is a requirement for an inline button component. Currently we have the `type="link"` which changes the root html to an `<a>` tag. There are instances where it is preferable for it to remain a `<button>` to use and `onClick` handler instead of an `href`

**Requirements**
- prop `type="inline"` 
- prop type and description
- include a story in storybook and MDX documentation
- css suggestions (Subject to change when testing)
```
    display: inline;
    padding: 0 4px;
    font-size: inherit;
    width: auto;
```

**Example**

inlinebutton

6201:
Fixes: #13293

Explanation:  Create "inline" type for Button component


NiranjanaBinoy (approved):

brad-decker (approved):
LGTM, thanks for the contribution

PR #13306, opened 2022-01-13
Feat/hide 0 balances

### LINKED ISSUE #13239: Hide Tokens without Balance from Send Dropdown
We currently have a setting which allows users to hide tokens with zero balance from the main token listing on the home screen.  The dropdown on the send screen, however, still shows zero balance tokens:
ZeroBalance

We should consider hiding those tokens with 0 balance; if not always, at least when the user has the setting turned on.

Gtonizuka:
Fixes: #13239

Explanation:  

From the "send" screen, hides tokens with 0 balances (same as main screen)

Manual testing steps:  
  Tested on Chrome
   Tested on FF


ui/pages/send/send-content/send-asset-row/send-asset-row.component.js
```diff
@@ -161,9 +162,13 @@ export default class SendAssetRow extends Component {
           />
           <div className="send-v2__asset-dropdown__list">
             {this.renderNativeCurrency(true)}
-            {this.state.sendableTokens.map((token) =>
-              this.renderToken(token, true),
-            )}
+
+            <TokenListDisplay
```
adonesky1, line 166: Just curious why did you decide to make this a new component instead of changing the `sendableTokens` to further filter out tokens without balances?

ui/components/ui/token-list-display/token-list-display.js
```diff
@@ -0,0 +1,77 @@
...
+        }}
+      >
+        {t('loadingTokens')}
+      </div>
+    );
+  }
+
+  return (
+    <>
+      {tokensWithBalances.map((tokenData) => {
```
adonesky1, line 47: This needs to filter out tokens where `isERC721` is false, as we do currently in the `send-asset-row` component: https://github.com/MetaMask/metamask-extension/blob/bd33512c2a04d3855bafdffcee2943f3a925c1cd/ui/pages/send/send-content/send-asset-row/send-asset-row.component.js#L58

Gtonizuka:
@adonesky1 answering here because it does not let me respond to your review comment.

Re _"Just curious why did you decide to make this a new component instead of changing the sendableTokens to further filter out tokens without balances?"_

I was following the same pattern used in the main screen where we have a similar list of tokens with their balances shown. There is a good point to be made that these components should be the same reusable one as they look very similar, but I guess it is out of scope for this PR. 

Furthermore I think its a good idea to start moving out from those `class` based component architecture towards functional components and selectors hooks when possible (it is also mentioned in the FE architecture doc to use func components).

ui/components/ui/token-list-display/token-list-display.js
```diff
@@ -0,0 +1,79 @@
...
+      {sendableTokens.map((tokenData) => {
+        const { address, image, symbol } = tokenData;
+
+        return (
+          <div
+            key={address}
+            className="send-v2__asset-dropdown__asset"
+            onClick={() => handleSelectToken(tokenData)}
+          >
+            <div className="send-v2__asset-dropdown__asset-icon">
```
brad-decker, line 58: [REQ] If we are going to extract these into a UI component we need to update the class names to not reference the send flow, and we should use the <Box> component where it makes sense as well as <Typography> to reduce stylesheet clutter.

ui/components/ui/token-list-display/token-list-display.js
```diff
@@ -0,0 +1,79 @@
...
+import { useI18nContext } from '../../../hooks/useI18nContext';
+import { getTokens } from '../../../ducks/metamask/metamask';
+
+export default function TokenListDisplay(props) {
+  const t = useI18nContext();
+  const shouldHideZeroBalanceTokens = useSelector(
+    getShouldHideZeroBalanceTokens,
+  );
+
+  const { clickHandler } = props;
```
brad-decker, line 19: [REQ] If we could destructure this in the function signature that'd be preferable.

ui/components/ui/token-list-display/token-list-display.js
```diff
@@ -0,0 +1,79 @@
...
+
+  const handleSelectToken = (token) => clickHandler(token);
```
brad-decker, line 21: [REQ] Isn't handleSelectToken now functionally equivalent to just passing clickHandler directly?

ui/components/ui/token-list-display/token-list-display.js
```diff
@@ -0,0 +1,79 @@
...
+  const tokens = useSelector(getTokens, isEqual);
+  const { loading, tokensWithBalances } = useTokenTracker(
+    tokens,
+    true,
+    shouldHideZeroBalanceTokens,
+  );
+  if (loading) {
+    return (
+      <div
+        style={{
```
brad-decker, line 32: [REQ] Could we use css versus inline styles here?

ui/components/ui/token-list-display/token-list-display.js
```diff
@@ -0,0 +1,79 @@
...
+
+import { getShouldHideZeroBalanceTokens } from '../../../selectors';
+import { useTokenTracker } from '../../../hooks/useTokenTracker';
+import Identicon from '../identicon';
+import TokenBalance from '../token-balance';
+import { useI18nContext } from '../../../hooks/useI18nContext';
+import { getTokens } from '../../../ducks/metamask/metamask';
+
+export default function TokenListDisplay(props) {
+  const t = useI18nContext();
```
brad-decker, line 14: [Q] This might be more of a `app/` component? What do you think @georgewrmarshall

brad-decker (changes requested):

ui/components/ui/token-list-display/token-list-display.js
georgewrmarshall: Yes, I agree this is more suited to an `app/` component

ui/components/ui/token-list-display/token-list-display.js
brad-decker: The main reason being we already do it in most places and we should try and find a cohesive style guide so that each engineers preferences don't lead to many different patterns.I'd love to hear more about the react team's recommendations and why if you can find me some links/material -- and we can discuss making this the standard moving forward. But as of now i'd like most react components we write to look as close to the same style as possible.

ui/components/app/token-list-display/token-list-display.scss
```diff
@@ -0,0 +1,43 @@
...
+        font-size: 1rem;
+        line-height: 140%;
+        margin-bottom: 2px;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        white-space: nowrap;
+    }
+
+    &__balance {
+        font-size: 0.75rem;
```
brad-decker, line 35: [REQ] Please use the Typography mixins H1-H9,and Paragraph instead of setting font-size manually. There's also the Typography component that can be used to avoid some commonly needed styles.

brad-decker (changes requested):
Looks good just a best practices req.

ui/components/app/token-list-display/token-list-display.scss
brad-decker: Special note: you may not be able to use the full mixin for the symbol (due to setting font family etc) but there is also a font-size variable that you can use instead of the literal rem value

ui/components/app/token-list-display/token-list-display.scss
```diff
@@ -0,0 +1,43 @@
+.loading-span {
+  display: flex;
+  height: 250px;
+  align-items: center;
+  justify-content: center;
+  padding: 30px;
```
georgewrmarshall, line 6: Adhere to `8px` grid
```suggestion
  padding: 32px;
```

ui/components/app/token-list-display/token-list-display.scss
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  &__balance {
+    @include H7;
+    display: flex;
+    flex-flow: row nowrap;
+  }
+  
+  &__balance__label {
+    margin-right: 0.25rem;
```
georgewrmarshall, line 41: Any reason why you are using rems here? We are following the convention of using px values for margin and padding.

```suggestion
    margin-right: 4px;
```

georgewrmarshall (commented):
A couple tiny css suggestions

ui/components/app/token-list-display/token-list-display.js
```diff
@@ -0,0 +1,61 @@
...
+          <div
+            key={address}
+            className="token-list-item"
+            onClick={() => clickHandler(tokenData)}
+          >
+            <Identicon address={address} diameter={36} image={image} />
+            <div className="token-list-item__data">
+              <div className="token-list-item__symbol">{symbol}</div>
+              <div className="token-list-item__balance">
+                <span className="token-list-item__balance__label">
```
brad-decker, line 46: [req] nesting blocks isn't really valid BEM syntax for class names, if we could make this __balance-label instead that would be better. I know this is a NIT but using this forum as a way of spreading knowledge of best practices until the best practices documentation comes together.

brad-decker (approved):
Looks good to go. just a tiny little NIT and honestly the existing component you replaced put this practice into place and you inherited the class names so if we merge as is no big deal.

TL;DR - LGTM

darkwing (changes requested):
I'm seeing some odd styling -- we should get this addressed before merging:
LayoutError

Functionally I'm also seeing an item in the list with 0 balance.

Gtonizuka:
@darkwing not sure why. you seeing that - it looks fine on my machine. It seems like there is some CSS missing?
Screenshot 2022-01-31 at 16 26 31

danjm (approved):
LGTM

danjm:
@darkwing I tested and it looks okay. Can you double check and update your review when you get a chance?

darkwing (approved):
My apologies @Gtonizuka !  I re-checked and it's working great.  Well done!

georgewrmarshall (approved):
UI LGTM!

PR #13307, opened 2022-01-13
use PAT for `GITHUB_TOKEN`

rickycodes:
I've confirmed that using a personal access token for this action will successfully trigger the CLA bot action:

https://github.com/rickycodes/metamask-extension/pull/2

what we'll want to do is create a PAT for the @metamaskbot and add it as a GitHub secret (`secrets.METAMASKBOT_CROWDIN_TOKEN` to match this change) in the repository and we should be golden. once that's done we can merge this and consider this task done.

ref: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow

> When you use the repository's GITHUB_TOKEN to perform tasks on behalf of the GitHub Actions app, events triggered by the GITHUB_TOKEN will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs. For example, if a workflow run pushes code using the repository's GITHUB_TOKEN, a new workflow will not run even when the repository contains a workflow configured to run when push events occur.

### DIFF
--- a/.github/workflows/crowdin_action.yml
+++ b/.github/workflows/crowdin_action.yml
@@ -26,6 +26,6 @@ jobs:
         upload_translations: true
         download_translations: true
       env:
-        GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+        GITHUB_TOKEN: ${{ secrets.METAMASKBOT_CROWDIN_TOKEN }}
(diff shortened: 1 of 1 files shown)


Gudahtt (approved):
LGTM!

adonesky1 (approved):
🔥

PR #13315, opened 2022-01-14
Snaps key access warning refactor

### LINKED ISSUE #13297: Snaps key access warning refactor
UI improvement to the flask UI for warnings related to key access. Currently, a deprecated modal is being used to display this warning message. The intention is to possibly replace it with the popover component. Current [flask design](https://www.figma.com/file/nBKUQx0btfQiqVf30UA6KD/Flask-MVP?node-id=336%3A6342).

Current UI:

Popover component:

hmalik88:
Fixes: #13297 

Explanation:  `SnapInstallWarning` component was re-written to use the `Popover` component instead of the deprecated `Modal` component. 

See the linked issue for an image of the previous UI and see below for the improved UI per [flask designs](https://www.figma.com/file/nBKUQx0btfQiqVf30UA6KD/Flask-MVP?node-id=336%3A6342)

Manual testing steps:  
  - Run `yarn start --build-type flask`
  - To test a snap with the key access permissions, you can clone our fork of the filecoin snap at: [Filecoin snap](https://github.com/MetaMask/filsnap/tree/prerelease)
  - Checkout the `prerelease` branch and then run `yarn && yarn demo`
  - Follow the install flow for the filecoin snap from the locally hosted dapp and observe the changes when the snap install warning pops up


ritave (dismissed):

ui/components/app/app-components.scss
```diff
@@ -65,12 +75,9 @@
...
-@import 'loading-network-screen/index';
-@import 'flask/experimental-area/index';
-@import 'transaction-decoding/index';
-@import 'advanced-gas-fee-popover/index';
-@import 'advanced-gas-fee-popover/advanced-gas-fee-gas-limit/index';
-@import 'advanced-gas-fee-popover/advanced-gas-fee-inputs/index';
-@import 'advanced-gas-fee-popover/advanced-gas-fee-inputs/base-fee-input/index';
-@import 'advanced-gas-fee-popover/advanced-gas-fee-input-subtext/index';
-@import 'advanced-gas-fee-popover/advanced-gas-fee-defaults/index';
+
```
Gudahtt, line 78: Nit: These newlines appear to have been added accidentally

ritave (dismissed):

ui/components/app/flask/snap-install-warning/index.scss
```diff
@@ -0,0 +1,43 @@
...
+
+  &__footer-button {
+    margin-right: 16px;
+
+    &:last-of-type {
+      margin-right: 0;
+    }
+  }
+
+  .popover-header {
```
Gudahtt, line 27: What are these styling changes for? Typically we try to scope all style changes to the namespace for the component, `snap-install-warning` in this case. Adding rules for other classes can introduce subtle dependencies between componments that can be broken accidentally fairly easily. For example if you rename a class within the `Popover` component, change layout or rearrange classes/components, etc. It makes the styles part of the public API of the component, which we don't want.

hmalik88: Hmmm, I wanted to add in these rules to this usage of the `Popover` component in `SnapInstallWarning`, I thought throwing the rules under `.snap-install-warning` to avoid any conflicts. Not sure how I can have the style changes I need without doing this or either customizing the `Popover` component to accept css rules.

Gudahtt: Ideally we'd use the same style for all popovers so they have a consistent look/feel. What motivated these changes?

hmalik88: The flask designs themselves I suppose, the padding it already had didn't seem sufficient.

Gudahtt: Could you leave those changes out, and make them to the base Popover class instead? I don't think there is anything special about this popover that warrants different padding than anywhere else in the extension.

hmalik88: If I remove the `Popover` css from `SnapInstallWarning`'s css, add a wrapper div ('snap-install-warning__content`) around what get's displayed in the `popover-content` div and add some padding (because the popover-content div doesn't provide that), the result is below. Do you think that's fine, visually?

Screen Shot 2022-01-20 at 10 58 56 AM

hmalik88: I guess this eliminates the need for me to change the `Popover` component and makes the content portion look decent, although I'd argue with 24px padding on the sides, things look better. This way I also don't have to add the `Popover` css in `SnapInstallWarning` at all.

Gudahtt: >Do you think that's fine, visually?

I mean... it doesn't look _terrible_ to me. Certainly passable enough that it's a step forward.

I think I agree that the 24px padding looks better. Just we should use the same popover styles throughout the app to maintain a consistent look and feel. So we should pursue that change in the popover component itself rather than for this specific instance.

hmalik88: Okay, I pushed up the change. I'll customize the `Popover` component further in a separate PR and then loopback and adjust the content padding to 24px in this component.

Gudahtt (approved):
LGTM!

PR #13320, opened 2022-01-14
Fix manually connecting sites

rekmarks:
The manual site connection flow was broken due to an incorrectly implemented background API hook. This PR restores the manual site connection functionality.

To test manually, try connecting to an unconnected site manually via the UI.

### DIFF
--- a/app/scripts/controllers/permissions/background-api.js
+++ b/app/scripts/controllers/permissions/background-api.js
@@ -1,3 +1,4 @@
+import nanoid from 'nanoid';
 import {
   CaveatTypes,
   RestrictedMethods,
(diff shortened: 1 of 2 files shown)


Gudahtt (dismissed):
LGTM!

Gudahtt:
I have confirmed that [the id is optional, and also used `nanoid` when not provided](https://github.com/MetaMask/snaps-skunkworks/blob/main/packages/controllers/src/permissions/PermissionController.ts#L1741), so how the ID is generated has not changed. I also confirmed that [this is how it worked in the old permission controller](https://github.com/MetaMask/metamask-extension/pull/12243/files#diff-dd58fd574b3f61106e9c9952e1c5457ab8ef7acb1949f50c44683c8a706cd6abL129).

Gudahtt (approved):

mcmire (approved):
LGTM as well.

PR #13328, opened 2022-01-17
Editing token to be sent changes the displayed recipient address #13031

### LINKED ISSUE #13031: Editing token to be sent changes the displayed recipient address
**Describe the bug**
While in the process of creating a transaction, the displayed recipient address changes to the ERC-20 token contract address when user edits the token to be sent from ETH to said ERC-20 token.

#### Notes:
* As far as I see, this only affects the **displayed** recipient address. The transaction would still be published with the correct, initially intended recipient address, if it was confirmed.
* This bug happens on various blockchains with various gas tokens initially selected. I tested ETH Mainnet, BSC, and Optimism.

1. Click on "Send" 
2. Choose recipient
3. Choose ETH
4. Click Next
5. On the page that shows the estimated gas fees, click "<EDIT" in the top left
6. Change the asset to send from the gas token to any other token
7. Click Next
8. Observe that the recipient address has changed to the contract address of the token you are about to send

**Expected behavior**
Recipient address does not change

**Screenshots**

 - OS: Ubuntu 20.04
 - Hardware Wallet: no / independent of it
 - Browser: Chromium Version 95.0.4638.69, Firefox Browser 94.0
 - MetaMask Version: 10.7.0 on Firefox and 10.7.1 on Chromium.

dragana8:
Fixes: https://github.com/MetaMask/metamask-extension/issues/13031

Screenshot from 2022-01-17 13-10-22
Screenshot from 2022-01-17 13-10-45

### DIFF
--- a/shared/modules/transaction.utils.js
+++ b/shared/modules/transaction.utils.js
@@ -1,4 +1,9 @@
 import { isHexString } from 'ethereumjs-util';
+import { ethers } from 'ethers';
+import abi from 'human-standard-token-abi';
+import log from 'loglevel';
+import { TRANSACTION_TYPES } from '../constants/transaction';
+import { readAddressAsContract } from './contract-utils';
 
 export function transactionMatchesNetwork(transaction, chainId, networkId) {
   if (typeof transaction.chainId !== 'undefined') {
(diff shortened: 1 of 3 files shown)


mirjanaKukic:
Verified by QA

david0xd (approved):

brad-decker (changes requested):
I'm not sure what this is doing, the type of transaction should already be set in the txMeta... The problem here is that in confirm-send-ether.container.js we have hardcoded the asset type to native, but allowed the user to change from a non native asset in the edit screen. We need to improve the editTransaciton methods in both confirm-send-ether.container.js and confirm-send-token.container.js such that the variability of asset types is not overridden by the hardcoded asset types.

danjm:
@brad-decker See this thread https://consensys.slack.com/archives/C02GH833686/p1641401539229200

Alexalucky (approved):

danjm:
Conversation continued in slack:

Screenshot from 2022-01-24 07-35-24

danjm:
Moving this back to draft as we resolve the block issues

PR #13329, opened 2022-01-17
Adding metrics events for EIP-1559 V2 flow

### LINKED ISSUE #12107: EIP-1559 events tracking: Edit button and Edit screen audit/implementation
Refer to [Notion Success Metrics](https://www.notion.so/EIP-1559-v2-330b809eb95048f9a45fa70c9f9b9303#3dea90b44d7e444ebd74190a2095fa75) for details.

**Background**
1. Better default recommendations: Are our users using our recommended gas settings? Some directional insights can be provided by:
a. Are users clicking on 1) `Confirm` button or 2) clicking on `Reject` or `Edit` buttons on the Transaction Confirmation screen? If users are confirming without editing, we can assume that our gas estimates are acceptable for most transaction use cases.
Screen Shot 2021-09-13 at 4 38 28 PM

b. If the users are editing their transactions, are they selecting the default recommendation based on the transaction type (1)? Or are they customizing (2) the gas inputs? What we want to track is the acceptance rate for the default recommendation for regular transactions and Swaps transactions.
Screen Shot 2021-09-13 at 4 56 36 PM

c. Gas estimation accuracy: What's the difference in the estimated default to the user's manually changed Max Priority fee, Max fee, and Total gas_fees (in gwei)?
- Log where the confirmation happens in transaction screen or advanced option screen.

**Audit**
Send/Swaps screen
1. Verify that the `Clicked on Edit Gas` is is correctly associated with `Edit` in 1a, and in Swaps `Edit` 
2. Verify that the `Confirm` button is triggering with the `Transaction Approved` event in the Send screen as well as the Swaps screen
3. Verify that the `Reject` button is triggering with the `Transaction Rejected` event

**Edit screen**
4. Verify that `Gas Fee Changed` is triggering when users change from the default Medium in Send and High in Swaps
5. Verify that when users change priority from the default above, we're capturing the associated `gas price`
6. Verify that `Clicked Advanced Options` is triggering when user are clicking on the Advanced Options drop down
7. What event is `Save` button is associated with?

**Bonus**
- Capture the breadcrumb steps the user takes from the `Edit` button to the `Save` button

jpuri:
Fixes: #12107

Adding metrics events for EIP-1559 V2


ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-gas-limit/advanced-gas-fee-gas-limit.js
```diff
@@ -36,6 +38,12 @@ const AdvancedGasFeeGasLimit = () => {
     setGasLimit(value);
   };
 
+  useEffect(() => {
+    if (!initialGasValues.gasLimit) {
```
mcmire, line 42: What do you think about using `!Boolean(initialGasValues.gasLimit)` rather than just `!`, in case this is ever a number? Same goes for the other three instances of this.

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
```diff
@@ -25,27 +26,32 @@ const AdvancedGasFeeDefaults = () => {
...
+  const handleUpdateDefaultSettings = () => {
+    if (isDefaultSettingsSelected) {
+      dispatch(setAdvancedGasFee(null));
+    } else {
+      const defaults = {
+        maxBaseFee,
+        priorityFee: maxPriorityFeePerGas,
+      };
+      captureTransactionMetrics({
+        action: 'Advance gas fee modal',
```
mcmire, line 46: Advance -> Advanced?

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-save/advanced-gas-fee-save.js
```diff
@@ -9,22 +10,51 @@ import I18nValue from '../../../ui/i18n-value';
...
       maxFeePerGas: decGWEIToHexWEI(maxFeePerGas),
       maxPriorityFeePerGas: decGWEIToHexWEI(maxPriorityFeePerGas),
       gasLimit,
+    };
+    updateTransaction({
+      estimateUsed: PRIORITY_LEVELS.CUSTOM,
+      ...gasValues,
+    });
+    captureTransactionMetrics({
+      action: 'Advance gas fee modal',
```
mcmire, line 38: Advance -> Advanced? Same for the other instances.

ui/hooks/gasFeeInput/useTransactionMetrics.js
```diff
@@ -0,0 +1,29 @@
...
+      eventOpts: {
+        category: 'Transactions',
+        action,
+        name,
+      },
+      customVariables: {
+        origin,
+        transaction_type: type,
+        source: origin === 'metamask' ? 'user' : 'dapp',
+        EIP_1559_V2: true,
```
mcmire, line 20: This function seems like it could be pretty useful in general, but tying it specifically to EIP-1559 V2 could be a surprise for people. What if this was an explicit thing? So within EIP-1559 V2 features, we say `captureTransactionMetrics({ action: ..., name: ..., EIP_1559_V2: true })`, but in other places we might just say `captureTransactionMetrics({ action: ..., name: ... })`.

mcmire (commented):
Left some comments!

ui/hooks/gasFeeInput/useTransactionMetrics.js
jpuri: This is mostly used in V2 implementation, `variables` can be used to override this.

mcmire: I see. It may be getting used in the EIP-1559 V2 code, but I'm not sure everyone knows that. The name of this function doesn't say anything about where it should be used or what it does by default, so it seems like it could be misleading. One might reasonably assume in fact that EIP-1559 V2 is _not_ the default considering that it's not the default elsewhere (yet).

If you want to keep this line, then what do you think about calling this function `useTransactionMetricsForEIP1559V2` or something and have `EIP_1559_V2: true` be non-overridable? Then if we want to use this in a more general way we can come back later and make a decision about this.

ui/hooks/gasFeeInput/useTransactionMetrics.js
```diff
@@ -0,0 +1,29 @@
+import { useContext } from 'react';
```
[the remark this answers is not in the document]
jpuri: good point, I moved it out of `/gasFeeInput`

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-save/advanced-gas-fee-save.js
jpuri: fixed 👍

ui/hooks/gasFeeInput/useTransactionMetrics.js
jpuri: I changed it to 2 methods `captureTransactionMetrics` which is more generic and `captureTransactionMetricsForEIP1559V2` which uses the previous one.

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
jpuri: fixed

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-gas-limit/advanced-gas-fee-gas-limit.js
jpuri: Weird but it seems to fail lint:

Screenshot 2022-01-19 at 3 39 46 PM

brad-decker (changes requested):
The event signaturess in this PR are the old metrics format, not the new...furthermore we need the fragment system in place to report the metrics in the way that product wants to use then..hold until event fragments for transactions are available

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-gas-limit/advanced-gas-fee-gas-limit.js
mcmire: Hmm, weird. Perhaps `if (initialGasValues.gasLimit === undefined || initialGasValues.gasLimit === null) {`, then?

ui/hooks/gasFeeInput/useTransactionMetrics.js
mcmire: Thanks!

shared/constants/transaction.js
```diff
@@ -249,6 +249,10 @@ export const TRANSACTION_GROUP_CATEGORIES = {
  *  ones, are added to the controller state in an unapproved status. When this
  *  happens we fire the Transaction Added event to show that the transaction
  *  has been added to the user's MetaMask.
+ * @property {'Transaction UI Action'} UI_ACTION - When an unapproved transaction
```
brad-decker, line 252: These types of events should *modify properties of the other events* Like for example, if the user edits gas it should be modifying properties on the -added- fragment ... that fragment will ultimately result in the approved action when the user approves the tx...and then we can notate that the user modified the gas

app/scripts/controllers/transactions/index.js
```diff
@@ -1708,7 +1708,7 @@ export default class TransactionController extends EventEmitter {
       // If the user rejects a transaction, finalize the transaction added
       // event fragment. with the abandoned flag set.
       case TRANSACTION_EVENTS.REJECTED:
-        id = `transaction-added-${txMeta.id}`;
+        id = `transaction-rejected-${txMeta.id}`;
```
brad-decker, line 1711: This is incorrect... it should be finalizing the transaction-added fragment. THere is no transaction-rejected fragment.

app/scripts/controllers/transactions/index.js
```diff
@@ -1721,6 +1721,11 @@ export default class TransactionController extends EventEmitter {
         this.updateEventFragment(id, { properties, sensitiveProperties });
         this.finalizeEventFragment(`transaction-submitted-${txMeta.id}`);
         break;
+      case TRANSACTION_EVENTS.UI_ACTION:
+        id = `transaction-ui_action-${txMeta.id}`;
+        this.updateEventFragment(id, { properties, sensitiveProperties });
+        this.finalizeEventFragment(`transaction-ui_action-${txMeta.id}`);
+        break;
```
brad-decker, line 1728:
```suggestion
```

app/scripts/metamask-controller.js
```diff
@@ -1288,6 +1288,9 @@ export default class MetamaskController extends EventEmitter {
       updateAndApproveTransaction: txController.updateAndApproveTransaction.bind(
         txController,
       ),
+      trackTransactionMetricsEvent: txController.trackTransactionMetricsEvent.bind(
```
brad-decker, line 1291: [REQ] We should not expose this function to the UI. I added the public method createTransactionEventFragment for the UI to verify that the proceeding fragment exists

ui/hooks/useTransactionEventFragment.js
```diff
@@ -0,0 +1,31 @@
+import { useCallback } from 'react';
+import { TRANSACTION_EVENTS } from '../../shared/constants/transaction';
+
+import { useGasFeeContext } from '../contexts/gasFee';
+import { trackTransactionMetricsEvent } from '../store/actions';
+
+export const useTransactionEventFragment = () => {
```
brad-decker, line 7: [REQ] I think this function should ultimately have the same signature as the `useEventFragment` method but with the extra bit of using a known id (everything a user can modify in a normal flow will modify the transaction-added fragment, the only one we might want to modify the transaction-submitted fragment is on the speed up or cancel flows). We also need to use the createTransactionEventFragment to ensure the proper fragment exists before we modify it.

brad-decker (changes requested):
Added some additional direction. Feel free to push back on these -- this is just the design implementation I had envisioned when doing the proceeding work.

shared/constants/transaction.js
jpuri: For some of the events it is possible to have modified properties in fragment but some events are like `user clicked edit link` or  `user open advance modal`. Which are just user action and no property update.

May be we keep only later ones in `UI_ACTION` action and others can have different action like `GAS_UPDATED` with property updates.

app/scripts/controllers/transactions/index.js
jpuri: I see, let me revert this change

app/scripts/metamask-controller.js
jpuri: It will be useful for UI to be able to re-use this whole function otherwise I need to duplicate some of this code in UI. Is it ok if I add a method on top of this one for use by UI ?

ui/hooks/useTransactionEventFragment.js
jpuri: As mentioned above I would prefer to re-use controller methods `trackTransactionMetricsEvent` for UI, thus keeping things simple at UI for transaction events.

brad-decker: using the fragment system will ultimately result in this method being called, that's the issue I'm seeing ... and this method was never meant or intended to be called from the UI.

shared/constants/transaction.js
brad-decker: I think, and could be wrong, that we'd prefer to have a 'user_clicked_edit_gas' property on the 'Transaction Approved' event -- such that we can see of the transactions approved by a user, what percentage attempted to have gas edited. Versus which percent actually modified the gas. @kevinghim could you tell me what you think? If we only track the click event separately it can tell us what percentage of *users* have clicked edit gas while transacting, but if we track it on the transaction event we can see what percentage of transactions users have attempted to edit gas on. Does that make sense?

kevinghim: > we'd prefer to have a 'user_clicked_edit_gas' property on the 'Transaction Approved' event -- such that we can see of the transactions approved by a user, what percentage attempted to have gas edited.
This is correct. Ultimately, at the event of approval, we'd want to track not only the percentage of approval based on where it was approved (transaction screen, edit screen, advanced edit screen, swap), but also what was edited (gas option level, amount in preconfigured levels (L/M/H), or manual).

jpuri: @brad-decker : this is actually not clear to me, here are you suggesting that we capture these events in txMeta and add as property to transaction approved event ?

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
```diff
@@ -44,14 +46,23 @@ const AdvancedGasFeeDefaults = () => {
...
           maxBaseFee,
           priorityFee: maxPriorityFeePerGas,
         }),
       );
-      setDefaultSettingsSelected(true);
+      updateTransactionEventFragment({
+        advanced_gas_defaults_updated: true,
+        advanced_gas_defaults_updated_maxbasefee: maxBaseFee,
+        advanced_gas_defaults_updated_priorityfee: maxPriorityFeePerGas,
+      });
```
jpuri, line 65: Above takes care to add/remove parameter in transaction fragment for advance gas default updates. This is exactly not part of transaction flow but more of user settings. But I still added it to transaction, I also see it included in transaction metrics [here](https://www.notion.so/EIP-1559-v2-330b809eb95048f9a45fa70c9f9b9303#3dea90b44d7e444ebd74190a2095fa75).

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-gas-limit/advanced-gas-fee-gas-limit.js
jpuri: I removed this part of code, thus resolving it

jpuri:
The PR add params in `transaction-added-XXXX` fragment for following UI actions:

1. Confirm transaction page - user open edit gas modal
2. Confirm transaction page - user open advance gas modal
3. Edit gas popup - select a estimate from - `low / medium / high / dapp suggested / swap suggested / 10% Minimum`
4. Edit gas popup - open advance gas popup
5. Advance gas modal - user update default (add new default / remove default)
6. Advance gas modal - user submits advance gas modal

An additional change us for cancel / speedup estimateSuggested is fixed to `medium / tenPercentIncrease` whatever is the transaction initialised with.

ui/hooks/useTransactionEventFragment.js
```diff
@@ -0,0 +1,22 @@
...
+      }
+      updateEventFragment(`transaction-added-${transaction.id}`, params);
+    },
+    [transaction],
+  );
+
+  return {
+    updateTransactionEventFragment,
+  };
+};
```
jpuri, line 22: For transactions this hook provided only the option to update fragments, I found that other fragments changes were not required for transaction, not at least currently.

jpuri:
With this PR I have tried to cover all events detailed here: https://www.notion.so/EIP-1559-v2-330b809eb95048f9a45fa70c9f9b9303#3dea90b44d7e444ebd74190a2095fa75

Fir old transaction screens not updated as part of 1559 V2 implementation we would still need to update them to use fragments.

brad-decker (changes requested):
I think we should make this simpler for tracking:

gas_edit_type: 'none' (default) | 'basic' | 'advanced' (shows if the gas was actually edited and what screen it was done)
gas_edit_attempted: 'none' (default) | 'basic' (shows user opened the basic modal) | 'advanced' (shows user opened advanced options) 

so we can see who opened the modal but didn't save by examining these two properties. and we've condensed the properties required down to 2.

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
```diff
@@ -44,14 +46,23 @@ const AdvancedGasFeeDefaults = () => {
     if (isDefaultSettingsSelected) {
       dispatch(setAdvancedGasFee(null));
       setDefaultSettingsSelected(false);
+      updateTransactionEventFragment({
+        advanced_gas_defaults_updated: true,
```
brad-decker, line 50: You need to specify the properties/sensitiveProperies subkeys otherwise these get written to the fragment options instead of the payload that gets sent to segment.

brad-decker: sensitiveProperties are reserved for things that will show up on chain,

ui/hooks/useTransactionEventFragment.js
```diff
@@ -0,0 +1,22 @@
+import { useCallback } from 'react';
+
+import { useGasFeeContext } from '../contexts/gasFee';
+import { updateEventFragment } from '../store/actions';
+
+export const useTransactionEventFragment = () => {
```
brad-decker, line 6: [Q] out of curiosity how many callsites for this do not have access to txMeta.id?

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
brad-decker: this is fine for now, but as its a user level setting (or is it per account?) we'll want to migrate these to traits whenever we do the work to make that a thing. Sory for a future date. 

I do think gas_default_max_base_fee is sufficient, the presence of it in the payload is good enough to track that it was updated.

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
jpuri: Hey @brad-decker : in case sensitiveProperties is reserved for on chain values, should I updated [this](https://github.com/MetaMask/metamask-extension/pull/13385/files#diff-675c437262d906486cf5436c0b64535347286cc4d1081dbd863fb45b8a01b81aR1513) PR to use properties ?

app/scripts/controllers/transactions/index.js
```diff
@@ -1570,6 +1570,8 @@ export default class TransactionController extends EventEmitter {
       network,
       type,
       eip_1559_version: eip1559Version,
+      gas_edit_type: 'none',
+      gas_edit_attempted: 'none',
```
jpuri, line 1574: These are the defaults

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
jpuri: These is user level and updated preference controller.

ui/hooks/useTransactionEventFragment.js
jpuri: [GasFeeContext](https://github.com/MetaMask/metamask-extension/blob/develop/ui/contexts/gasFee.js) has been added to the root components of all transaction flows - simplesend, swap, cancel/speedup, approve token. Thus this should be available to all of transaction related components.

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-defaults/advanced-gas-fee-defaults.js
brad-decker: @jpuri I think we should keep these where they are because we'll want to compare them to actual values used (on chain) and they don't serve much purpose on the non-anonymized events. so no, lets keep them as sensitive.

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-save/advanced-gas-fee-save.js
```diff
@@ -26,6 +28,11 @@ const AdvancedGasFeeSaveButton = () => {
       maxPriorityFeePerGas: decGWEIToHexWEI(maxPriorityFeePerGas),
       gasLimit,
     });
+    updateTransactionEventFragment({
+      properties: {
+        gas_edit_type: 'advance',
```
mcmire, line 33: `advanced`?

ui/components/app/edit-gas-fee-button/edit-gas-fee-button.js
```diff
@@ -46,12 +48,23 @@ export default function EditGasFeeButton({ userAcknowledgedGasMissing }) {
...
   }
 
+  const openEditGasFeeModal = () => {
+    updateTransactionEventFragment({
+      gas_edit_attempted: 'basic',
+    });
+    openModal('editGasFee');
+  };
+
+  const openAdvanceGasFeeModal = () => {
```
mcmire, line 58: Advance -> Advanced

ui/components/app/edit-gas-fee-button/edit-gas-fee-button.js
```diff
@@ -46,12 +48,23 @@ export default function EditGasFeeButton({ userAcknowledgedGasMissing }) {
...
+    updateTransactionEventFragment({
+      gas_edit_attempted: 'advance',
```
mcmire, line 60: advance -> advanced

mcmire (commented):
This all looks good, just some typos I noticed.

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-save/advanced-gas-fee-save.js
jpuri: Just fixed

ui/components/app/edit-gas-fee-button/edit-gas-fee-button.js
jpuri: Fixed

ui/components/app/edit-gas-fee-button/edit-gas-fee-button.js
jpuri: Also fixed

brad-decker (approved):

darkwing (approved):

adonesky1 (approved):
LGTM!

PR #13331, opened 2022-01-17
implement event fragments for tx controller

brad-decker:
**Q: What does this PR do?**
_A: Uses the new event fragment system in the metametrics controller to create expansion points of transaction metrics_

**Q: What impact will this have on the current analytics?**
_A: It should result in no changes to when, how, or what is tracked. This is where we need to spend a majority of review time_

**Q: If this results in no change, why is it necessary?**
_A: Product and business leaders would like to get an assessment of how well our transaction confirmation screens perform in a variety of cases. As an example, how accurate and friendly are we presenting as estimations? To do this we need to be able to tack on additional details to the transaction events so that we can build reports._

**Q: Couldn't this be achieved by tracking individual events?**
_A: Sort of. We can get an idea of how many times the user clicked edit on the gas section of the confirmation screen, and we can edit the number of times they modified the fields of the gas parameters, but we cannot say out of all transactions submitted in November, 30% ejected from MetaMask gas price suggestions. This is because there is no way to uniquely tie a string of events to a single transaction. This is the problem the event fragment system aims to solve._

**Q: I noticed  some changes to the MetaMetrics controller. What gives?**
_A: I added a uniqueIdentifier optional parameter for the fragment event object. This allows systems that track events for objects or data structures that already have unique identifiers to use a deterministic id for event fragments. This allows the transaction controller to not have to perform any lookups for fragments to find the proper event to update/finalize. I also added a new option 'initialEvent' which allows a fragment to immediately kick off a metric event. This simply saves in code duplication and verbosity._

### DIFF
--- a/app/scripts/controllers/metametrics.js
+++ b/app/scripts/controllers/metametrics.js
@@ -240,7 +240,7 @@ export default class MetaMetricsController {
    *  originated the fragment. This is for fallback only, the fragment referrer
    *  property will take precedence.
    */
-  finalizeEventFragment(id, { abandoned = false, page, referrer }) {
+  finalizeEventFragment(id, { abandoned = false, page, referrer } = {}) {
     const fragment = this.store.getState().fragments[id];
     if (!fragment) {
       throw new Error(`Funnel with id ${id} does not exist.`);
--- a/app/scripts/controllers/metametrics.js
+++ b/app/scripts/controllers/metametrics.js
@@ -200,6 +200,21 @@ export default class MetaMetricsController {
   }
 
   /**
+   * Returns the fragment stored in memory with provided id or undefined if it
+   * does not exist.
+   *
+   * @param {string} id - id of fragment to retrieve
+   * @returns {[MetaMetricsEventFragment]}
+   */
+  getEventFragmentById(id) {
+    const { fragments } = this.store.getState();
+
+    const fragment = fragments[id];
+
+    return fragment;
+  }
+
+  /**
    * Updates an event fragment in state
    *
    * @param {string} id - The fragment id to update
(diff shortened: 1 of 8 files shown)


app/scripts/controllers/transactions/index.js
```diff
@@ -1501,27 +1503,81 @@ export default class TransactionController extends EventEmitter {
...
+          sensitiveProperties,
+          persist: true,
+          uniqueIdentifier: `transaction-submitted-${txMeta.id}`,
+        });
+        break;
+      case TRANSACTION_EVENTS.FINALIZED:
+      default:
+        this.finalizeEventFragment(`transaction-submitted-${txMeta.id}`);
+        break;
+    }
```
jpuri, line 1580: Is it possible to also have a field which tells whether it is EIP-1559 V1 / V2. Also, can you plz take a look at [this](https://www.notion.so/EIP-1559-v2-330b809eb95048f9a45fa70c9f9b9303#3dea90b44d7e444ebd74190a2095fa75), @kevinghim has suggested few more fields for transaction like type, source, referrer.

brad-decker: We can add those values, yes, but the goal of this PR is to just create the hook points -- not to modify behavior. Further changes can be done in stacked PRs.

Alexalucky: 881390

app/scripts/controllers/metametrics.js
```diff
@@ -168,7 +168,7 @@ export default class MetaMetricsController {
     }
     const { fragments } = this.store.getState();
 
-    const id = generateUUID();
+    const id = options.uniqueIdentifier ?? generateUUID();
```
jpuri, line 171: 👍

jpuri (approved):
great work 👍

darkwing (approved):

brad-decker:
Going to validate taht all events we care about still get fully into segment/mixpanel before i merge.

PR #13337, opened 2022-01-18
Incorrect Rounding Down of Balance #10167

### LINKED ISSUE #10167: Incorrect Rounding Down of Balance
**Describe the bug**
0.0048 WBTC is rounded to 0.004 WBTC in the display. This is a non-trivial difference. The extension also displays $161.29 USD rather than the actual ~$190 ish that the correct amount converts to. So it is using the incorrectly rounded number during the conversion.

It is both rounding incorrectly (rounding to 0.004 rather than 0.005) but also with way too few decimal points. From what I can tell if you add a WBTC contract address as a custom token, it should use 8 decimal places. The 0.0008 WBTC rounding error is roughly $30.

Transfer some amount of WBTC (or presumably other currencies), see incorrect rounding in extension on main page and when you click on the token details.
Verify that the actual amount differs from this.

**Expected behavior**
Either allow the user to pick the decimal place to round at or at least round at enough decimal places that the displayed error doesn't matter.

**Screenshots**
image
If you manually lookup the address on etherscan or attempt a transaction and hit max, the correct amount is shown.
image

 - OS: Windows 10
 - Wallet: Metamask
 - Browser: Version 87.0.4280.88 (Official Build) (64-bit)
 - MetaMask Version: Recently installed MetaMask, I am on 8.1.11.

At one point there was a "There was a problem with your balance" message here but it disappeared.
image

On mobile, the bug is not present. Instead, the total is displayed as 0.00479 WBTC with a correctly calculated USD amount also

dragana8:
Fixes: https://github.com/MetaMask/metamask-extension/issues/10167

Screenshot from 2022-01-18 10-31-00
Screenshot from 2022-01-18 10-31-13

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "@metamask/contract-metadata": "^1.31.0",
     "@metamask/controllers": "^24.0.0",
     "@metamask/eth-ledger-bridge-keyring": "^0.10.0",
-    "@metamask/eth-token-tracker": "^3.0.1",
+    "@metamask/eth-token-tracker": "^4.0.0",
     "@metamask/etherscan-link": "^2.1.0",
     "@metamask/jazzicon": "^2.0.0",
     "@metamask/logo": "^3.1.1",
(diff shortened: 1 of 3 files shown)


mirjanaKukic:
Verified by QA

david0xd (approved):

NiranjanaBinoy (approved):

PR #13339, opened 2022-01-18
Feat: Label hardware wallets

### LINKED ISSUE #13123: Label hardware wallet type
Currently the hardware wallet label on accounts just says "hardware", would be nice if it just said which kind instead:
- Trezor
- Ledger
- Lattice1
- QR
Screen Shot 2021-12-18 at 9 20 40 PM

Gtonizuka:
Fixes: #13123

Explanation:  

Instead of showing a generic message, we customize the label shown depending on the hardware type

Manual testing steps:  

I only tested this on Ledger Nano X as I don't have any other physical wallet - please test on other devices

Screenshot 2022-01-18 at 13 48 31


ui/components/app/account-menu/account-menu.test.js
```diff
@@ -96,7 +96,7 @@ describe('Account Menu', () => {
 
     it('render imported account label', () => {
       const importedAccount = wrapper.find('.keyring-label.allcaps');
-      expect(importedAccount.text()).toStrictEqual('imported');
+      expect(importedAccount.text()).toStrictEqual('[imported]');
```
Gtonizuka, line 99: We discussed this with @danjm and due to the way we wrap our `useI18nContext` around the provider, the square brackets are needed in this case

ui/components/app/account-menu/keyring-label.js
```diff
@@ -0,0 +1,37 @@
...
+
+export default function KeyRingLabel(props) {
+  const t = useI18nContext();
+
+  const { keyring } = props;
+
+  let label = null;
+
+  // Keyring value might take a while to get a value
+  if (keyring) {
```
darkwing, line 15: Instead of this large `if`, I think we can simply do:

```
if (!keyring) { return null; }
```

ui/components/app/account-menu/keyring-label.js
```diff
@@ -0,0 +1,37 @@
...
+    const { type } = keyring;
+    if (type === KEYRING_TYPES.QR) {
```
darkwing, line 17: We can also probably use a `switch` instead of the series of `if`'s.

darkwing (changes requested):
Just a few quick comments @Gtonizuka !  I can't wait to see this go in!

ui/components/app/account-menu/keyring-label.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { useI18nContext } from '../../../hooks/useI18nContext';
+import { KEYRING_TYPES } from '../../../../shared/constants/hardware-wallets';
+
+export default function KeyRingLabel(props) {
+  const t = useI18nContext();
+
+  const { keyring } = props;
```
brad-decker, line 10: [REQ] Could we destructure this in function signature?

brad-decker (changes requested):

ui/components/app/account-menu/keyring-label.js
darkwing: In our newer components we've been destructuring in the signature.  I think for now we continue that pattern, and if you have strong feelings about it, we can address it in an engineering meeting.

ui/components/app/account-menu/keyring-label.js
```diff
@@ -0,0 +1,48 @@
...
+  // Keyring value might take a while to get a value
+  if (!keyring) {
+    return null;
+  }
+
+  const { type } = keyring;
+
+  switch (type) {
+    case KEYRING_TYPES.QR:
+      label = 'QR';
```
darkwing, line 23: Maybe in `shared/constants/hardware-wallets.js` we should create a `KEYRING_NAMES` object that has these labels; I feel like we'll end up using these shorthand names more.  I use Ledger, for example, in an upcoming PR.

brad-decker (approved):

Gtonizuka:
@darkwing if you give me a green check I will ship this today

darkwing (approved):

PR #13340, opened 2022-01-18
Maintain leading whitespace in sig req window

### LINKED ISSUE #13251: Signature request window strips leading whitespace
**Describe the bug**
When using the EIP 191 sign message functionality in Metamask, the window that renders the message being signed, strips the whitespace at the beginning of each line.  For example, if passing in an indented JSON, the rendered text removes all the indentations, causing the message to be much harder to parse visually, as well as being a misrepresentation of the message being signed (see image in Screenshots section).  After discussion with @danfinlay, it seems this was not intentional behavior, and that this is likely the result of the fact [that HTML by default renders serial whitespace as a singular space](https://stackoverflow.com/a/12863646/3299736).

**Expected behavior**
Serial whitespace should be maintained in the rendering of the message being signed.  I believe this can be done by making use of [the `white-space` CSS property, and setting its value to `pre`](https://www.w3.org/TR/CSS2/text.html#white-space-prop).

I believe this is the relevant code section:  https://github.com/MetaMask/metamask-extension/blob/develop/ui/components/app/signature-request-original/signature-request-original.component.js#L241

**Screenshots**
image

 - OS: Mac OS
 - Browser: Chrome
 - MetaMask Version: 10.8.1

jonator:
Fixes: #13251

Explanation:  According to @danfinlay, the signature request window unintentionally trims leading whitespace for new lines. This is misleading for what is actually being signed (indented json).

Manual testing steps:  
  - Sign a message containing indentation, ensure whitespace is maintained

### DIFF
--- a/ui/components/app/signature-request-original/index.scss
+++ b/ui/components/app/signature-request-original/index.scss
@@ -222,7 +222,7 @@
     overflow-wrap: break-word;
     border-bottom: 1px solid #d2d8dd;
     padding: 6px 18px 15px;
-    white-space: pre-line;
+    white-space: pre;
   }
 
   &__help-link {


jonator:

NiranjanaBinoy (approved):

brad-decker (approved):

brad-decker:
Thanks for the contribution @jonator

PR #13342, opened 2022-01-18
Update permission confirmation window title

rekmarks:
Adds a generic "Permission Request" title to the permission confirmation that's applied when `eth_accounts` is not requested. This does not address the case when both `eth_accounts` _and_ other permissions are requested, in which case we should arguably fall back to the generic title as well. Either way, this improves upon the status quo.


ryanml (approved):
lgtm!

NiranjanaBinoy (approved):

PR #13344, opened 2022-01-18
Update global link hover color from orange to blue (--Blue-400)

### LINKED ISSUE #13335: [Bug]: Update global link hover color
**Description**
Hovering over a link or `<a>` tag that doesn't have any custom css turns orange rather than a variation of blue.

**Expected behaviour**
Link turns light blue(`--Blue-400` or `#1098fc`) instead of orange.

**Screenshots**
Screen Shot 2022-01-17 at 10 17 15 AM

digiwand:
Fixes: https://github.com/MetaMask/metamask-extension/issues/13335

Explanation: Updates the global `a:hover` color from `#df6b0e` to `--Blue-400` (`#1098fc`). This deprecates `#df6b0e`. 

----
This affects one `a.warning` instance. We will remove the `.warning` class from this link instance and replace it with the `--primary-blue` color we use for other links. Please see George's comment below for more details about this decision: https://github.com/MetaMask/metamask-extension/pull/13344#pullrequestreview-856143632.

Manual testing steps:
  - Click on Profile Icon
  - Click on "Import Account"
  - Change "Select Type" option to "JSON File"
  - Hover over "File import not working? Click here!"
  - Observe hovered text

**Original (`#df6b0e`):**
MM original

**After applying `--Blue-400` (`#1098fc`) to the hover state:**
MM Import Account hover on warning

**After removing the `.warning` class and setting the default state to `--primary-blue` (`#037dd6`):**
MM use regular blue link

### DIFF
--- a/ui/css/base-styles.scss
+++ b/ui/css/base-styles.scss
@@ -52,10 +52,6 @@ html {
   color: #ffae00;
 }
 
-a.warning:hover {
-  color: var(--Orange-500);
-}
-
 /* stylelint-disable */
 #app-content {
   overflow-x: hidden;
--- a/ui/css/base-styles.scss
+++ b/ui/css/base-styles.scss
@@ -45,7 +45,6 @@ html {
 
 /*
   This warning class is used in the following files still:
-  /ui/pages/create-account/import-account/json.js
   /ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.component.js
 */
 .warning {
--- a/ui/css/base-styles.scss
+++ b/ui/css/base-styles.scss
@@ -72,7 +72,7 @@ a {
 }
 
 a:hover {
-  color: #df6b0e;
+  color: var(--Blue-400);
 }
 
 input.large-input,
--- a/ui/css/base-styles.scss
+++ b/ui/css/base-styles.scss
@@ -52,6 +52,10 @@ html {
   color: #ffae00;
 }
 
+a.warning:hover {
+  color: var(--Orange-500);
+}
+
 /* stylelint-disable */
 #app-content {
   overflow-x: hidden;
--- a/ui/pages/create-account/import-account/index.scss
+++ b/ui/pages/create-account/import-account/index.scss
@@ -56,6 +56,10 @@
     margin-top: 34px;
   }
 
+  &__help-link {
+    color: var(--primary-blue);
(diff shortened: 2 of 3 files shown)


digiwand:

georgewrmarshall (commented):
Great PR description and auditing work on checking existing link components! In the case of the warning link I'm wondering if we should remove the className `warning` and add some css to the `ui/pages/create-account/import-account/index.scss` file so it conforms to the general link styles. My thinking is:

- It is an outlier and probably should conform to similar link component styles
- I wouldn't categorize it as a warning per se but rather additional information
- All actionable links should be associated with the primary action color(blue).

That way we aren't adding any more styles to the base styles css. We are updating the outlier to conform to current conventions and we get to remove some deprecated references.

The other two instances listed should also be updated. I'll [create an issue for that](https://github.com/MetaMask/metamask-extension/issues/13347). If they are warnings maybe they should look more like the below
Screen Shot 2022-01-19 at 1 28 46 PM

None of the above is documented anywhere and it is mostly my opinions so apologies I am  working on that!!

The warning link className change could be something like `new-account-import-form__help-link` and the css would be

```
  &__help-link {
    color: var(--primary-blue);
  }
```

digiwand:
Thank you for the clear thoughts! I love that this solution removes the outlier, and I also wouldn't categorize that message as a warning. Updating the code with your proposal. The BEM naming convention you provided LGTM! 

Happy to take over the issue you just made to get rid of the `.warning` class once and for all! :)

digiwand:
@georgewrmarshall 

Thank you for the clear thoughts! I love that this solution removes the outlier, and I also wouldn't categorize that message as a warning. Updating the code with your proposal. The BEM naming convention you provided LGTM!

Happy to take over the issue you just made to get rid of the .warning class once and for all! :)

georgewrmarshall (approved):
LGTM! Nailed it 💯 

Verified link conversion in `json.js` ✅ 
Screen Shot 2022-01-19 at 8 45 01 PM

brad-decker (approved):
Great work, @digiwand

adonesky1 (approved):
Looks good!

ryanml (approved):
awesome work!

digiwand:
Thanks for the reviews y'all!

PR #13360, opened 2022-01-20
Delete transaction stuck in approved state on cancel

### LINKED ISSUE #9686: Submitting a tx for hardware wallet that is disconnected causes metamask to be unusable
**Describe the bug**
I disconnect my hardware wallet quite often. Sometimes I submit a tx when the hardware is disconnected. When I realize, I connect the hardware wallet and try and resubmit the transaction. However it does not allow resubmitting. The UI is completely broken at that point. For example, clicking speedup, choosing a new gas and clicking 'save' does not submit the transaction. The UI stays on the 'save' screen and cannot proceed, almost as if there is an error that is not caught or something.

Logging out / Logging back in or removing the hardware wallet and adding it back does not fix this. Metamask is basically unusable from this point forward. The only solution I've found is to uninstall metamask and reinstall it. That's quite a painful UX for something that happens quite often.

1. Import the hardware wallet to your metamask account.
2. Disconnect your hardware wallet from the usb.
2. Make a transaction on any page and submit it.
3. Realize your hardware wallet is disconnected and connect it.
4. Click the speed up button on the transaction and try resubmitting it. It is not possible.

**Expected behavior**
I expect that when I click speed up button it should resubmit.

I also expect that if my hardware wallet is not connected the transaction should fail. When my hardware wallet is connected, and I don't confirm a tx on the wallet in a given time, metamask will mark the transaction as failed after a given time. However that same user experience when the hardware wallet is disconnected doesn't happen. I would expect that submitting a tx when my wallet isn't connected would fail after X seconds.

 - OS: Macbook Mojave 10.14.6 
 - Hardware Wallet: Ledger Nano S 1.6.0
 - Browser: Chrome Version 86.0.4240.80 (Official Build) (x86_64)
 - MetaMask Version: 8.0.10

**Extra details**
The transaction is not actually submitted to infura or whatever eth nodes being used (because the hardware wallet didn't sign it). However the transaction stays stuck in Metamask so you can't submit a future transaction. (There isn't even a nonce for you to replace). Hence the reason that metamask becomes completely unusable after this point.

danjm:
Fixes: #9686

This PR fixes the following scenario:
1. user goes through confirm transaction flow with a hardware wallet account
2. user confirms in the MetaMask UI
3. user disconnects their hardware wallet or closes the metamask UI before confirming on their device
4. The transaction can then be in a state where it is displayed as "pending" but is not actually submitted on chain and can't be submitted or cancelled

Transactions in the above state have their status stuck as "Approved". This PR corrects this scenario by enabling the deletion/rejection of transactions stuck in that state.

The PR also adds an e2e test for the existing functionality of marking transactions in that state as "Failed" upon boot of metamask. I attempted to add an e2e test for the functionality this PR adds, but because of the onboot functionality, it is impossible to get the app in that state (without either adding mock hardware wallet functions for tests, or adding logic that only applies for certain tests)


test/e2e/tests/delete-approved-tx.spec.js
```diff
@@ -0,0 +1,44 @@
...
+
+        const completedTransactions = await driver.findElements(
+          '.transaction-list__completed-transactions',
+        );
+        assert.equal(completedTransactions.length, 1);
+        const transactionStatus = await driver.findElements(
+          '.transaction-status',
+        );
+
+        assert.equal(await transactionStatus.length, 1);
```
adonesky1, line 40: Does this assertion guarantee that the status is now `failed`? Should we drill into the value of the `transaction-status` element?

test/e2e/tests/delete-approved-tx.spec.js
```diff
@@ -0,0 +1,44 @@
+const { strict: assert } = require('assert');
+const { convertToHexValue, withFixtures } = require('../helpers');
+
+describe('Failed approved but not submitted transaction on boot', function () {
```
adonesky1, line 4:
```suggestion
describe('Fail approved but not submitted transaction on boot', function () {
```

adonesky1:
Is this PR still relevant? Or overcome by other work? @danjm

test/e2e/tests/delete-approved-tx.spec.js
segun: fixed [here](https://github.com/MetaMask/metamask-extension/pull/13360/commits/65f822c84e491070b2ef5a0114256eb921d65aad)

danjm:
@adonesky1 This PR is still relevant

lavamoat/build-system/policy-override.json
```diff
@@ -10,7 +10,9 @@
         "@babel/plugin-proposal-nullish-coalescing-operator": true,
         "@babel/plugin-proposal-object-rest-spread": true,
         "@babel/plugin-proposal-optional-chaining": true,
-        "@babel/preset-typescript": true
+        "@babel/preset-typescript": true,
+        "@babel/plugin-transform-react-jsx": true,
+        "@babel/plugin-syntax-jsx": true
```
jpuri, line 15: Hey @danjm : is above change intended for this PR ?

ui/components/app/transaction-list-item/transaction-list-item.component.js
```diff
@@ -95,14 +99,24 @@ function TransactionListItemInner({
           legacy_event: true,
         },
       });
-      if (supportsEIP1559V2) {
+      if (isApprovedButNotSubmitted) {
+        dispatch(cancelTx(transactionGroup.primaryTransaction));
+      } else if (supportsEIP1559V2) {
```
jpuri, line 104: Hey @danjm : thinking of an edge case that transaction is submitted by the time the request to cancel above reaches the backend.

segun: @jpuri not sure I understand how this could happen. Can you please explain.

PR #13367, opened 2022-01-21
UI fixes in EIP-1559 V2 screens

### LINKED ISSUE #13346: Design tweaks
### Describe the bug

- [x] Proposing removing the orange highlight on the "High probability" copy. It seems too "warning" like? Instead, let's include that message in the the copy below: "High probability, even in volatile markets. Use Aggressive to cover..."  
Screen Shot 2022-01-18 at 3 33 25 PM

- [ ] Bell curve images not centered, will need to re-export with equal whitespace on either side to account for the text. @jakehaugen can help format and export when needed. 
Screen Shot 2022-01-18 at 3 35 22 PM

- [x] Text color of time estimate between the "edit" screen and transaction screens should match (i.e. should be green in both places in screenshot).
Screen Shot 2022-01-18 at 3 37 25 PM

- [x] When you have “Advanced” on by default and click “Edit”, the back arrow closes the modal. This feels a bit unexpected, let's remove the back arrow and just  have the “X” icon to close.  
Screen Shot 2022-01-18 at 3 40 41 PM

- [x] Use same alert style for both messages (keep icon and left justified text)
image

- [x] Let's remove the second sentence in the warning here. We already show the time estimate and the "last seen..." is not really telling the user anything new. So, warning will just say "Future transactions will queue after this one".
Screen Shot 2022-01-18 at 3 43 14 PM

- [x] Copy change: "customise" to "customize"
Screen Shot 2022-01-18 at 3 50 03 PM

- [x] While on the "edit" screen, as prices update the values flash. There is a white background on the numbers that becomes visible on the selected item. Can we remove this? 
Screen Shot 2022-01-18 at 3 51 29 PM

- [x] Tooltip placement for “Aggressive” is off a bit. Is this due to having more content? Can we get this aligned the same way as all the other tooltips? 
Screen Shot 2022-01-18 at 3 52 45 PM

- [x] Can we make the hit areas for "Base fee" and "Priority fee" be both the amount and label? Right now it's just he amount. 
Screen Shot 2022-01-18 at 3 53 35 PM

- [x] Network status block should be aligned with the text above, not the "selected" box"
Screen Shot 2022-01-18 at 4 01 28 PM

- [x] In the designs, the thought was to anchor the “Network status” to the bottom of the modal to better separate gas options vs. network data. Is that possible without mudding up the code? If not, we can increase spacing slightly and size the background of the modal to fit. 
Screen Shot 2022-01-18 at 5 03 47 PM

- [x] Max base fee shown in "Edit" screen tooltips (all three) should follow rounding logic and not show so many digits. 
image

- [x] Fiat estimates in "Advanced" not working? 
image

### Steps to reproduce

n/a

_No response_

### Version

10.8.2

### Build type

Beta

### Browser

Chrome

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

jpuri:
Fixes: #13346

Explanation:  Design tweaks in EIP-1559 V2 functionality

### DIFF
--- a/ui/components/ui/loading-heartbeat/index.js
+++ b/ui/components/ui/loading-heartbeat/index.js
@@ -3,7 +3,6 @@ import classNames from 'classnames';
 import PropTypes from 'prop-types';
 import React from 'react';
 
-import { COLORS } from '../../../helpers/constants/design-system';
 import { getGasLoadingAnimationIsShowing } from '../../../ducks/app/app';
 import { useShouldAnimateGasEstimations } from '../../../hooks/useShouldAnimateGasEstimations';
 
--- a/ui/components/ui/loading-heartbeat/index.js
+++ b/ui/components/ui/loading-heartbeat/index.js
@@ -10,7 +10,7 @@ import { useShouldAnimateGasEstimations } from '../../../hooks/useShouldAnimateG
 const BASE_CLASS = 'loading-heartbeat';
 const LOADING_CLASS = `${BASE_CLASS}--active`;
 
-export default function LoadingHeartBeat({ backgroundColor = COLORS.WHITE }) {
+export default function LoadingHeartBeat({ backgroundColor = '#fff' }) {
   useShouldAnimateGasEstimations();
   const active = useSelector(getGasLoadingAnimationIsShowing);
 
--- a/ui/components/ui/loading-heartbeat/index.js
+++ b/ui/components/ui/loading-heartbeat/index.js
@@ -1,13 +1,16 @@
-import classNames from 'classnames';
-import React from 'react';
 import { useSelector } from 'react-redux';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import { COLORS } from '../../../helpers/constants/design-system';
 import { getGasLoadingAnimationIsShowing } from '../../../ducks/app/app';
 import { useShouldAnimateGasEstimations } from '../../../hooks/useShouldAnimateGasEstimations';
 
 const BASE_CLASS = 'loading-heartbeat';
 const LOADING_CLASS = `${BASE_CLASS}--active`;
 
-export default function LoadingHeartBeat() {
+export default function LoadingHeartBeat({ backgroundColor = COLORS.WHITE }) {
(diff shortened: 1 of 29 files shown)


brad-decker (approved):
LGTM

ui/pages/confirm-transaction-base/transaction-alerts/transaction-alerts.js
```diff
@@ -111,9 +111,19 @@ const TransactionAlerts = ({
       )}
       {isNetworkBusy ? (
         <ActionableMessage
-          message={<I18nValue messageKey="networkIsBusy" />}
+          message={
```
mcmire, line 114: Thanks for adding this, I think I just copied from another place but didn't know how we normally do this 👍🏻

ui/components/ui/loading-heartbeat/index.js
```diff
@@ -24,6 +26,11 @@ export default function LoadingHeartBeat() {
         e.preventDefault();
         e.stopPropagation();
       }}
+      style={{ backgroundColor }}
```
mcmire, line 29: Is there a reason why we have to fill in this div with any kind of color when the item in question is being loaded? I noticed that in the styles for LoadingHeartbeat we set a default background color of `#fff`. Do we need this at all? Can we just fade out whatever text is being loaded and fade it back in?

jpuri: Hey @mcmire : the way this loader currently works is it is absolutely positioned on top of relative container and covers it. It changes its opacity (animates) thus the text or anything in container blinks.

Background colour of loader needs to be same as container this I made this change, this fixes point-8 here: https://github.com/MetaMask/metamask-extension/issues/13346

mcmire: Oh right. I still don't like that, but I guess maybe there's a reason for it. Okay 👍🏻

mcmire (approved):

ui/components/ui/loading-heartbeat/index.js
jpuri: We can refactor to improve this loader, but yep currently that is out of scope.

PR #13378, opened 2022-01-22
fixbug: adding network overwrite the existed network

ForestTigerKing:
Fixes:  
When i add a new custom network，the new nework can overwrite the existed network in extreme cases.

step1：Add Binance Smart Chain Mainnet
bsc

step2:  Add  TEST network with wrong rpc url and same chainid

step3:  Click "Save", the error is "Could not fetch chain ID, is your RPC URL correct?"

step4: input the same rpc url

step5: Click "Save", overwrite the Binance Smart Chain Mainnet

Explanation:  
The cause of the error:   In an event(such as Onclick) , modify the usestate variable multiple times, and only the last time is valid. In the function validateUrlRpcUrl modify errors twice, the urlExistsErrorMsg was ignored（step4）, so the "Save" button is enabled. I split "errors" into three variables to avoid this problem.

### DIFF
--- a/ui/pages/settings/networks-tab/networks-form/networks-form.js
+++ b/ui/pages/settings/networks-tab/networks-form/networks-form.js
@@ -82,7 +82,9 @@ const NetworksForm = ({
   const [blockExplorerUrl, setBlockExplorerUrl] = useState(
     selectedNetwork?.blockExplorerUrl || '',
   );
-  const [errors, setErrors] = useState({});
+  const [rpcUrlError, setRpcUrlError] = useState({});
+  const [chainIdError, setChainIdError] = useState({});
+  const [blockExplorerUrlError, setBlockExplorerUrlError] = useState({});
   const [isSubmitting, setIsSubmitting] = useState(false);
 
   const resetForm = useCallback(() => {
@@ -91,7 +93,9 @@ const NetworksForm = ({
     setChainId(getDisplayChainId(selectedNetwork.chainId));
     setTicker(selectedNetwork?.ticker);
     setBlockExplorerUrl(selectedNetwork?.blockExplorerUrl);
-    setErrors({});
+    setRpcUrlError({});
+    setChainIdError({});
+    setBlockExplorerUrlError({});
     setIsSubmitting(false);
   }, [selectedNetwork, selectedNetworkName]);
 
@@ -125,7 +129,9 @@ const NetworksForm = ({
       setChainId('');
       setTicker('');
       setBlockExplorerUrl('');
-      setErrors({});
+      setRpcUrlError({});
+      setChainIdError({});
+      setBlockExplorerUrlError({});
       setIsSubmitting(false);
     } else if (
       prevNetworkName.current !== selectedNetworkName ||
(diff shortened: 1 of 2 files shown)


ForestTigerKing:
the unit test failed，there should be a problem with the code of develop branch.

darkwing:
Thank you for the submission @ForestTigerKing !  Would it be possible to add some test coverage here?

NiranjanaBinoy (commented):
@ForestTigerKing The changes here look good, but it would be great to have some test coverage as @darkwing  mentioned. Thanks for the submission.

ui/pages/settings/networks-tab/networks-form/networks-form.js
```diff
@@ -166,37 +176,20 @@ const NetworksForm = ({
...
-      // Do not factor in duplicate chain id error for submission disabling
-      if (key === 'chainId' && error.key === 'chainIdExistsErrorMsg') {
-        return false;
-      }
-      return error.key && error.msg;
-    });
+    // Do not factor in duplicate chain id error for submission disabling
+    return (
+      (Boolean(chainIdError?.key) &&
+        chainIdError.key === 'chainIdExistsErrorMsg') ||
```
digiwand, line 189:
```suggestion
        chainIdError.key !== 'chainIdExistsErrorMsg') ||
```
To confirm, we want to not include the duplicate chain id errors here

digiwand, line 189:
```suggestion
        chainIdError.key !== 'chainIdExistsErrorMsg') ||
```
To confirm, we want to not include the duplicate chain id errors here

digiwand (commented):
Great job finding and fixing this bug, and for sweeping out the unused errors!

ForestTigerKing:
@darkwing @digiwand @NiranjanaBinoy  thank you for your review.   I fix the bug and add some unit test.

ForestTigerKing:
@darkwing @digiwand @NiranjanaBinoy  I have added unit tests，judge whether the logic is correct through the state of the “Save” button.

digiwand (changes requested):
Fix is good to merge with me. I think the tests work though they could be improved since I see a couple of missing test cases. I'm okay to get this merged in to fix the issue and return to the test cases later. 

Please update the test phrase, `should button Safe in correct state`, to something like `should render "Save" button in the correct state`, and I will approve! 

Super awesome you found and smashed this bug!

CryptoGodfatherVA2 (approved):

CryptoGodfatherVA2 (approved):

ui/pages/settings/networks-tab/networks-form/networks-form.js
NiranjanaBinoy: good catch @digiwand

digiwand (approved):
[Edit]

Merge conflict seems a bit complicated as it does crossover this fix. I am investigating this. @ForestTigerKing

digiwand (commented):
man... this got messy with one of the latest merges... PR(https://github.com/MetaMask/metamask-extension/pull/12431)

Good news is that the issue is fixed! The unfortunate news is that PR(https://github.com/MetaMask/metamask-extension/pull/12431) overwrites a ton of work done in this PR. There was no mention of this bug being fixed in the other PR, which I think is how the collision happened. This has been great work, and I'm bummed we can't merge this in. Thank you for looking out for this project, responding to all of our feedback, and being an excellent contributor! I hope to be there to see the next merge. 

Something that is still missing is these tests that you have added to test the enabled and disabled state of the "Save" button. We could open a new PR and get that merged

ForestTigerKing:
@digiwand It's all right，I closed the pr.

PR #13382, opened 2022-01-24
Using new graph images for edit gas fee popover

jpuri:
Using new images for graphs

Screenshot 2022-01-24 at 6 11 32 AM


ui/components/app/edit-gas-fee-popover/edit-gas-tooltip/edit-gas-tooltip.js
```diff
@@ -88,7 +88,7 @@ const EditGasToolTip = ({
         editGasMode === EDIT_GAS_MODES.SWAPS
       ) &&
       !estimateGreaterThanGasUse ? (
-        <img alt="" src={`./images/curve-${priorityLevel}.svg`} />
+        <img alt="" src={`./images/curve-${priorityLevel}.png`} />
```
brad-decker, line 91: [REQ] Lets add an alt description using t('').

brad-decker (dismissed):
Everything looks great and i love those graph images, just one accessibility nit that we should solve

mcmire (commented):
Is it possible to keep these as SVGs? I think that would be more maintainable in the long run. If we use PNGs then we have to have Retina-compatible images so they appear crisp on Macs.

ui/components/app/edit-gas-fee-popover/edit-gas-tooltip/edit-gas-tooltip.js
jpuri: Added the fix for this.

jpuri:
@brad-decker , @mcmire : I did the fix to add alt text. For svg images I am checking with Jake.

mcmire (approved):
Looks better, thanks!

adonesky1 (approved):
LGTM!

PR #13385, opened 2022-01-24
Capturing default gas estimates in txMeta and passing it to metrics

jpuri:
PR partially addresses: https://github.com/MetaMask/metamask-extension/issues/12107
It adds default gas estimate for transaction and some other details to transaction metrics fragment.


app/scripts/controllers/transactions/index.js
```diff
@@ -1502,6 +1509,12 @@ export default class TransactionController extends EventEmitter {
       gasParams.gas_price = gasPrice;
     }
 
+    if (defaultGasEstimates) {
+      gasParams.defaultGasEstimates = this._getGasValuesInGWEI(
```
brad-decker, line 1513: [REQ] This is insufficient because segment won't flatten this object. We'll need to do gasParams.default_gas_price etc.

app/scripts/controllers/transactions/index.js
```diff
@@ -543,6 +543,12 @@ export default class TransactionController extends EventEmitter {
       txMeta.txParams.gas = defaultGasLimit;
       txMeta.originalGasEstimate = defaultGasLimit;
     }
+    txMeta.defaultGasEstimates = {
+      gas: txMeta.txParams.gas,
```
brad-decker, line 547: [Q] Wouldn't this only capture the gasPrice / maxFeePerGas etc of the initial default? If it changes in the gas controller it would be out of sync with what the user is actually being presented, yeah?

app/scripts/controllers/transactions/index.js
```diff
@@ -1512,12 +1525,15 @@ export default class TransactionController extends EventEmitter {
...
 
+    const { eip1559V2Enabled: eip1559V2 } = this.preferencesStore.getState();
+
     const properties = {
       chain_id: chainId,
       referrer,
       source,
       network,
       type,
+      eip1559V2,
```
brad-decker, line 1536: [REQ]: `eip_1559_version: eip1559V2 ? '2' : '1';` (opens up better options after v3 goes live.

brad-decker (changes requested):

app/scripts/controllers/transactions/index.js
jpuri: yes you are right in case of estimate low /medium / high this can get out of sync. Let me change this to use gas estimates

app/scripts/controllers/transactions/index.js
```diff
@@ -1502,6 +1510,21 @@ export default class TransactionController extends EventEmitter {
       gasParams.gas_price = gasPrice;
     }
 
+    if (defaultGasEstimates) {
+      const estimateType = defaultGasEstimates.defaultEstimate;
+      const { gasFeeEstimates } = await this._getEIP1559GasFeeEstimates();
+      gasParams.default_gas_estimates = this._getGasValuesInGWEI({
```
brad-decker, line 1516: [REQ] --- this can't be an object, has to be nested properties. gasParams will be passed to _getGasValuesInGWEI on line 1536. Just add the values directly to gasParams here.
gasParams.default_gas_price, gasParams.default_max_priority_fee_per_gas, etc.

app/scripts/controllers/transactions/index.js
```diff
@@ -1512,12 +1535,19 @@ export default class TransactionController extends EventEmitter {
 
     const gasParamsInGwei = this._getGasValuesInGWEI(gasParams);
 
+    let eip1559Version = 'null';
```
brad-decker, line 1538: [req] lets default to '0'.

app/scripts/controllers/transactions/index.js
jpuri: I just updated the PR with fix.

app/scripts/controllers/transactions/index.js
jpuri: I just updated the PR with change.

app/scripts/controllers/transactions/index.js
```diff
@@ -1474,7 +1481,7 @@ export default class TransactionController extends EventEmitter {
     }
   }
 
-  _buildEventFragmentProperties(txMeta, extraParams) {
+  async _buildEventFragmentProperties(txMeta, extraParams) {
```
brad-decker, line 1484: [Q] Were all usages of this function converted to use await?

brad-decker (changes requested):

app/scripts/controllers/transactions/index.js
jpuri: I see just fixed

app/scripts/controllers/transactions/index.js
jpuri: yep

jpuri:
@brad-decker : another thing I noticed is that there are metrics events captured before gas details are calculated for the transaction in `addTxGasDefaults`.

Those events will not have defaults.

brad-decker:
@jpuri the only one that will affect is the `Transaction Added` event which is a top of funnel event. Shouldn't hurt us in our analytics. 

The only other default I would like to see added is what level we are presenting to the user "low/medium/high" as our recommendation.

jpuri:
@brad-decker : I added [this](https://github.com/MetaMask/metamask-extension/pull/13385/files#diff-675c437262d906486cf5436c0b64535347286cc4d1081dbd863fb45b8a01b81aR1516) to capture default estimate.

Metamask always suggests medium gas estimate unless there is a DAPP suggested or Default Advance / Custom estimate saved by user.

`default_estimate` will be default for transaction, it can be low/medium/high/dappSuggested/custom.

brad-decker:
@jpuri code looks good. I'm going to do some tests with segment and mixpanel and see what the data looks like coming through before I give a thumbs up. Great work!

brad-decker (approved):
image

Nana12345678910:
Pull [Pull every thing back to original account every other request is fraudulent ](christina.vongphit.com)

darkwing (approved):

PR #13387, opened 2022-01-24
Fix failing test:unit pipeline

VSaric:
This PR fixes failing `test:unit` on pipeline. Added some more checks to the `edit-gas-display.component.js` to ensure that errors are not thrown in case `gasLimit` or `properGasLimit` are not defined.

### DIFF
--- a/ui/components/app/edit-gas-display/edit-gas-display.component.js
+++ b/ui/components/app/edit-gas-display/edit-gas-display.component.js
@@ -110,7 +110,11 @@ export default function EditGasDisplay({
   );
 
   let warningMessage;
-  if (new BigNumber(gasLimit).lessThan(new BigNumber(properGasLimit))) {
+  if (
+    gasLimit &&
+    properGasLimit &&
+    new BigNumber(gasLimit).lessThan(new BigNumber(properGasLimit))
+  ) {
     warningMessage = t('gasLimitRecommended', [properGasLimit]);
   }
 
--- a/ui/components/app/edit-gas-display/edit-gas-display.component.js
+++ b/ui/components/app/edit-gas-display/edit-gas-display.component.js
@@ -111,8 +111,8 @@ export default function EditGasDisplay({
 
   let warningMessage;
   if (
-    gasLimit &&
-    properGasLimit &&
+    gasLimit !== undefined &&
+    properGasLimit !== undefined &&
     new BigNumber(gasLimit).lessThan(new BigNumber(properGasLimit))
   ) {
     warningMessage = t('gasLimitRecommended', [properGasLimit]);


danjm (dismissed):
LGTM, we can merge this once the other failure is passing

digiwand (dismissed):
LGTM! I wonder if we would like to specify that this is an optional prop in `propTypes`

ui/components/app/edit-gas-display/edit-gas-display.component.js
```diff
@@ -110,7 +110,11 @@ export default function EditGasDisplay({
   );
 
   let warningMessage;
-  if (new BigNumber(gasLimit).lessThan(new BigNumber(properGasLimit))) {
+  if (
+    gasLimit &&
```
mcmire, line 114: Is `gasLimit` a number or a hex value here? Just wondering whether it would make more sense to say `if (Boolean(gasLimit) && Boolean(properGasLimit) && ...)` here.

VSaric: @mcmire `gasLimit` is a number. I tried with this changes that you suggested here `(Boolean(gasLimit) && Boolean(properGasLimit) && ...)`, unit test and coverage are passed also.
@danjm Can you please look at this and comment what do you think about it? I can add this changes if you agree with this.

brad-decker:
@VSaric I Have rebased this onto a branch that hopefully doesn't have the test audit failure. This PR cannot merge until the update-nanoid pr is merged then we'll rebase this on the other branch and finally merge into develop. Assuming @mcmire is satisfied with your response on the feedback

ui/components/app/edit-gas-display/edit-gas-display.component.js
danjm: @mcmire for which case does the `Boolean` call make a difference?

mcmire: `Boolean(...)` would check if the value in question is null or undefined. Without that, if `gasLimit` is 0, but is less than `properGasLimit`, then we wouldn't show a warning. Seems like we might not want to do that.

danjm: `Boolean(0) === false` right?

So do you mean that we need to do something like `gasLimit || gasLimit === 0`... or, what if we do `parseInt(gasLimit) !== NaN`?

danjm: I just pushed a commit that I think resolves the concern

mcmire: Yup, that looks good!

brad-decker (approved):

mcmire (approved):

PR #13401, opened 2022-01-25
Feat/tx insight decoding

ChingizMard:
Fixes: #

Explanation:  
Adding the decoding functionality for the DATA tab of the transaction preview. Now when transaction has a parameter that resembles a token amount it will be rendered with correct number of decimals and with fiat equivalent nearby. Properties that resemble time will be formatted and shown in relative format as well.

Manual testing steps:  
-  Go onto SushiSwap and try swapping tokens on chain 1. 
-  When MetaMask asks to preview transaction, go onto data tab.

## Function Signature Adding Guide
Contributors can now very easily add new signature decoding patters that are not included with this PR. Please see the examples bellow on how to do so.

## Example One

Let's cover a simple example first. We will make sure that a "deadline" parameter for the transaction to DEXes is presented in a human readable way. 

A lot of DEX smart contracts would have a function similar to this:

`function swapTokensForExactTokens(
        uint amountOut,
        uint amountInMax,
        address[] calldata path,
        address to,
        uint deadline
) external returns (uint[] memory amounts);`

The "deadline" parameter is usually an epoch time and looks something like this 1643042830, which is not very useful to a human.

In order to fix that we have added a rule below  so let's now look what it does. 

```
    {
        field: 'deadline',  
        transform: (deadline) => {
            return <TimestampRow date={deadline.toNumber()} from={true} />
        }
    },
```

First we need to select the parameter name that we want to work with. We put it under property called "field". Note that we can use a regex here. If we new that some DEXes use parameter name "deadline" and others use "deadlineToSwap" we would put "deadline.*" under the field property to match both cases. 

In order to tell MetaMask what to do with the value of the parameter we also need to write a "transform" function. Transform function will take the value of the "field" property as an argument and should return a React component. In this case we created a TimestampRow component. Feel free to reuse components we have created in your own rules. 

It is worth mentioning here that parameters are returned as the following types:

   - uint => BigNumber
   - address => String
   - int => BigNumber
   - uint[] => [BigNumber]
   - string[] => [String]

That means that in our case the deadline parameter is going to be passed to us as BigNumber and we can use a toNumber() function like we did on line:

## Example Two

Let's now consider a slightly more complex scenario. Let's show the token amounts for our swap transactions with correct number of decimal along side a fiat conversion rate. The trick here is that we would need to access the value of a different parameter in order to correctly determine token address and fetch fiat price and decimal places.
We have already made a component that will do all the conversation for you if you supply it with the original amount as BigNumber and the token address as a string. Component is called DecimalsDisplay.

```
    {
        field: 'amountIn.*',
        detect: ["path"],
        transform: (amountIn, path) => {
            return <DecimalsDisplay tokenAmount={amountIn} tokenAddress={path[0]} />
        }
    }
```

Now back to writing the rule. First of all we want to modify the "amountOut" parameter. Since we know that some of the other DEXes use "amountOutMinimum" we will use the "amountOut.*" regex. In order to properly display the values we also need the last address in the "path" parameter. To obtain that we will put the name of that parameter in the array under the "detect" property. Note that items under detect could also be regexes. 

After we have done this we write the transform function with the only difference that values of the detect field are also passed as function arguments in the order they appear in the detect list.

We then pass this information to the DecimalDisplay component and it will take care of the rest. 

### Callback Parameter Types

Here is the list of how different Solidity types are passed into the transform function. Anything that is not listed here is passed through form the 
tx-inisght backend, so please print the value to see how you can get the data you want from them. 

    uint => BigNumber
    address => String
    int => BigNumber
    uint[] => [BigNumber]
    string[] => [String]

## Availiable Components

- **TimestampRow** - converts epoch time into human readable string with relative time counter. 
Ex: 1643042830 => Jan 24 2022, 11:47 am (in 28 minutes)

@prop date: Number - epoch time as a regular integer.

- **DecimalsDisplay** - converts uint into a human readable token amount and adds fiat conversion on certain chains.
Respects local currency selection of the user.
Ex: 1372930994980681733 => 1.3729 ($5.72)

@prop tokenAmount: BigNumber - raw parameter value to format
@prop tokenAddress: String - address to fetch the  decimal count and fiat price.

### DIFF
--- a/package.json
+++ b/package.json
@@ -176,6 +176,7 @@
     "lodash": "^4.17.21",
     "loglevel": "^1.4.1",
     "luxon": "^1.26.0",
+    "moment": "^2.29.1",
     "nanoid": "^2.1.6",
     "nonce-tracker": "^1.0.0",
     "obj-multiplex": "^1.0.0",
--- a/ui/components/app/transaction-decoding/components/row-components/decimals/decimals.component.js
+++ b/ui/components/app/transaction-decoding/components/row-components/decimals/decimals.component.js
@@ -1,0 +1,66 @@
+import React, { useState, useEffect } from 'react';
+import { useSelector } from 'react-redux';
+import PropTypes from 'prop-types';
+import BigNumber from 'bignumber.js';
+import {
+  fetchTokenDecimal,
+  fetchTokenPrice,
+} from '../../../../../../pages/swaps/swaps.util';
+import { roundToDecimalPlacesRemovingExtraZeroes } from '../../../../advanced-gas-fee-popover/advanced-gas-fee-inputs/utils';
+import { hexToDecimal } from '../../../../../../helpers/utils/conversions.util';
+import {
+  getCurrentChainId,
+  getCurrentCurrency,
+} from '../../../../../../selectors';
+
+import { useCurrencyDisplay } from '../../../../../../hooks/useCurrencyDisplay';
+
+export default function DecimalsDisplay({ tokenAddress, tokenAmount }) {
+  const [formatedCryptoAmount, setFormatedCryptoAmount] = useState('');
+  const [formatedFiatAmount, setFormatedFiatAmount] = useState('');
+  const network = hexToDecimal(useSelector(getCurrentChainId));
+  const currentCurrency = useSelector(getCurrentCurrency);
+
+  const [, parts] = useCurrencyDisplay('1', {
+    numberOfDecimals: 2,
+    hideLabel: false,
+    currency: currentCurrency,
+  });
+
+  const getSymbol = (value) => {
+    const symbol = value.replace(/[0-9,. ]/gu, '');
+    return symbol;
+  };
+
+  useEffect(() => {
+    (async () => {
+      const decimals = await fetchTokenDecimal(tokenAddress, network);
+      const amount = new BigNumber(tokenAmount).dividedBy(
+        new BigNumber(10 ** decimals),
+      );
+      const correctedAmount = roundToDecimalPlacesRemovingExtraZeroes(
+        amount.toString(),
+        4,
+      );
+      const price = await fetchTokenPrice(tokenAddress, currentCurrency);
+      if (price) {
+        const fiatAmountBN = correctedAmount.mul(new BigNumber(price));
+        const fiatAmount = roundToDecimalPlacesRemovingExtraZeroes(
+          fiatAmountBN.toString(),
+          2,
+        );
+        setFormatedFiatAmount(
+          `(${fiatAmount.toString()}${getSymbol(parts.value.toString())})`,
+        );
+      }
+      setFormatedCryptoAmount(correctedAmount.toString());
+    })();
+  });
+
+  return <div>{`${formatedCryptoAmount} ${formatedFiatAmount}`}</div>;
+}
+
+DecimalsDisplay.propTypes = {
+  tokenAddress: PropTypes.string,
+  tokenAmount: PropTypes.BigNumber,
+};
--- a/ui/components/app/transaction-decoding/transaction-decoding.component.js
+++ b/ui/components/app/transaction-decoding/transaction-decoding.component.js
@@ -11,7 +11,7 @@ import { getSelectedAccount, getCurrentChainId } from '../../../selectors';
 import { hexToDecimal } from '../../../helpers/utils/conversions.util';
 import { I18nContext } from '../../../contexts/i18n';
 import { toChecksumHexAddress } from '../../../../shared/modules/hexstring-utils';
-import { transformTxDecoding } from './transaction-decoding.util';
+import { transformTxDecoding, parameterProcessing } from './transaction-decoding.util';
 import {
   FETCH_PROJECT_INFO_URI,
   FETCH_SUPPORTED_NETWORKS_URI,
@@ -20,6 +20,7 @@ import {
 import Address from './components/decoding/address';
 import CopyRawData from './components/ui/copy-raw-data';
 import Accreditation from './components/ui/accreditation';
+
 
 export default function TransactionDecoding({ to = '', inputData: data = '' }) {
   const t = useContext(I18nContext);
@@ -33,6 +34,8 @@ export default function TransactionDecoding({ to = '', inputData: data = '' }) {
   const [loading, setLoading] = useState(false);
   const [hasError, setError] = useState(false);
   const [errorMessage, setErrorMessage] = useState('');
+
+  const [processedValues, setProcessedValues] = useState({})
 
   useEffect(() => {
     (async () => {
@@ -84,9 +87,21 @@ export default function TransactionDecoding({ to = '', inputData: data = '' }) {
           blockNumber: null,
         });
 
+        let functionName = decoding.abi.name
+
         // transform tx decoding arguments into tree data
         const params = transformTxDecoding(decoding?.arguments);
+        
         setTx(params);
+
+        let dict = await parameterProcessing(params)
+
+        // This should become a function that simply takes an array of params and then returns the dictionary with name => what to display
+        if (functionName === "swapExactETHForTokens"){
+          
+          setProcessedValues(dict)
+        }
+        
 
         setLoading(false);
       } catch (error) {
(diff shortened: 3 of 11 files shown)


chingiz:
Wrong mention :)

ChingizMard:

package.json
```diff
@@ -176,6 +176,7 @@
     "lodash": "^4.17.21",
     "loglevel": "^1.4.1",
     "luxon": "^1.26.0",
+    "moment": "^2.29.1",
```
VikChawla, line 179: Is moment.js an acceptable package to add?

ui/pages/swaps/swaps.util.js
```diff
@@ -440,15 +453,14 @@ export async function fetchSwapsFeatureFlags() {
   return response;
 }
 
-export async function fetchTokenPrice(address) {
-  const query = `contract_addresses=${address}&vs_currencies=eth`;
-
+export async function fetchTokenPrice(address, currency = 'eth') {
```
VikChawla, line 456: Might want someone from swaps to look over these changes to their functions

package.json
dylanbutler1: I think date-fns is preferred due to its ability to be tree-shaken

https://blog.bitsrc.io/date-fns-vs-momentjs-9833f7463751

FrederikBolding: `luxon` is the current Date management package of choice. I would say use that over moment.

ui/pages/swaps/swaps.util.js
```diff
@@ -368,6 +369,18 @@ export async function fetchToken(contractAddress, chainId) {
   return token;
 }
 
+export async function fetchTokenDecimal(contractAddress, chainId) {
+  const tokenInfo = await fetchToken(contractAddress, chainId);
+  if (tokenInfo.error) {
+    const decimalsStr = await new AssetsContractController({
```
FrederikBolding, line 375: getTokenStandardAndDetails

FrederikBolding, line 375: Controllers should not be used in the frontend, instead you'll need to ask the backend to fetch the data for you

FrederikBolding: Usually those functions are passed via props iirc

ui/components/app/transaction-decoding/transaction-decoding.util.js
```diff
@@ -28,3 +30,85 @@ export const transformTxDecoding = (params) => {
...
+  }
+}
+
+function applyRule(params, rule) {
+  const regName = rule.field;
+  const mainParameter = params[rule.paramMapping[regName]];
+  const suppliments = [getFormattedParameter(mainParameter)];
+  const fieldName = mainParameter.name;
+  const detect = rule.detect === undefined ? [] : rule.detect;
+  for (const regex of detect) {
```
FrederikBolding, line 99: Perhaps this could be a reduce instead?

ui/components/app/transaction-decoding/transaction-decoding.util.js
```diff
@@ -28,3 +30,85 @@ export const transformTxDecoding = (params) => {
...
+    suppliments.push(getFormattedParameter(params[rule.paramMapping[regex]]));
+  }
+
+  return { fieldName, suppliments };
+}
+
+export const parameterProcessing = async (params) => {
+  const dict = {};
+  const rules = getMathingRules(params);
+  for (const rule of rules) {
```
FrederikBolding, line 109: This could also be a reduce

ui/pages/swaps/swaps.util.js
FrederikBolding: I would assume there is another way to do this using a controller.

ui/components/app/transaction-decoding/transaction-decoding.util.js
```diff
@@ -28,3 +30,85 @@ export const transformTxDecoding = (params) => {
...
+function ruleRegexSet(rule) {
+  const ruleMatches = [rule.field].concat(
+    rule.detect === undefined ? [] : rule.detect,
+  );
+  const rgxSet = new Set(ruleMatches);
+  return rgxSet;
+}
+
+function matchParam(param, reSet) {
+  for (const regx of reSet) {
```
FrederikBolding, line 43: Replace with `find`?

ui/components/app/transaction-decoding/components/row-components/decimals/decimals.component.js
```diff
@@ -0,0 +1,66 @@
...
+import { hexToDecimal } from '../../../../../../helpers/utils/conversions.util';
+import {
+  getCurrentChainId,
+  getCurrentCurrency,
+} from '../../../../../../selectors';
+
+import { useCurrencyDisplay } from '../../../../../../hooks/useCurrencyDisplay';
+
+export default function DecimalsDisplay({ tokenAddress, tokenAmount }) {
+  const [formatedCryptoAmount, setFormatedCryptoAmount] = useState('');
```
FrederikBolding, line 19: Typos

FrederikBolding (changes requested):

ui/components/app/transaction-decoding/transaction-decoding.util.js
```diff
@@ -28,3 +30,85 @@ export const transformTxDecoding = (params) => {
     };
   });
 };
+
+function ruleRegexSet(rule) {
+  const ruleMatches = [rule.field].concat(
+    rule.detect === undefined ? [] : rule.detect,
```
darkwing, line 36: Can this be `rule.detect ?? []`?

ui/components/app/transaction-decoding/transaction-decoding.component.js
```diff
@@ -124,9 +135,20 @@ export default function TransactionDecoding({ to = '', inputData: data = '' }) {
 
           case 'uint':
             return (
-              <span className="sol-item solidity-uint">
-                {[value.asBN || value.asString].toString()}
-              </span>
+              <Tooltip
+                title={[value.asBN || value.asString].toString()}
+                interactive="true"
```
darkwing, line 140: Can this be just the attribute name `interactive`?  I think with React you don't need to add the `"true"` in this case.

ui/components/app/transaction-decoding/transaction-decoding.component.js
```diff
@@ -86,8 +92,13 @@ export default function TransactionDecoding({ to = '', inputData: data = '' }) {
...
+
         setTx(params);
 
+        const dict = await parameterProcessing(params);
+
+        setProcessedValues(dict);
+
         setLoading(false);
       } catch (error) {
         setLoading(false);
```
darkwing, line 104: I know this isn't new to this PR but I think `setLoading(false);` can be removed from the `try` and `catch`, and instead be added with a `finally` block

ui/components/app/transaction-decoding/transaction-decoding.util.js
```diff
@@ -28,3 +30,85 @@ export const transformTxDecoding = (params) => {
...
+    .filter((rule) => rule !== null);
+  return rules;
+}
+
+function getFormattedParameter(param) {
+  switch (param.type.typeClass) {
+    case 'uint':
+      return param.value.asBN;
+    case 'int':
+      return param.value.asBN;
```
darkwing, line 83: I think this can be:

```
case 'unit':
case: 'int":
  return param.value.asBN;
```

ui/components/app/transaction-decoding/transaction-decoding.util.js
```diff
@@ -28,3 +30,85 @@ export const transformTxDecoding = (params) => {
...
+      return param;
+  }
+}
+
+function applyRule(params, rule) {
+  const regName = rule.field;
+  const mainParameter = params[rule.paramMapping[regName]];
+  const suppliments = [getFormattedParameter(mainParameter)];
+  const fieldName = mainParameter.name;
+  const detect = rule.detect === undefined ? [] : rule.detect;
```
darkwing, line 98: Not sure but could we do `const detect = rule.detect ?? []`?

darkwing (changes requested):
Just a few tiny code ideas.  Nice work!

PR #13428, opened 2022-01-27
lavamoat - @lavamoat/allow-scripts v2

kumavis:
Updates `@lavamoat/allow-scripts` to use the new secure `'a'a` package naming convention

The `'a'a` package naming convention uses the "shortest logical dependency path" (`a>b>c`) to name a package, instead of using the package's filesystem location based name (`c`). Using standard node resolution logic helps avoid bugs associated with edgecase package locations and security issues arising from flattened package path based names.


brad-decker (approved):
very nice

mcmire:
@kumavis For the uninformed, what is the significance of the changes to `allow-scripts` here? I see that it solves some problems mentioned in Slack — what are those problems and how do they solve them?

rekmarks (approved):
LGTM, but we should answer @mcmire's question in the PR description.

kumavis:
@mcmire added description. ping me with more questions

PR #13433, opened 2022-01-27
added extension changes from snaps branch

hmalik88:
The changes in this PR were split off from the `snaps` branch as they pertain to changes in the extension

  -  A hex value was changed to a CSS var
  -  The `SiteOrigin` component was used in the confirmation template to replace the `Chip` and `SiteIcon` components
  -  Network display is added as an allowed template key -- this allows network display to be optional (because snap confirmation doesn't pertain to any network)


jpuri (approved):

rekmarks (approved):
LGTM!

PR #13435, opened 2022-01-27
Yet more NFT UX cleanups

adonesky1:
- Organize NFT collection collapse/opened state by account/chainId to consistently preserve this state across accounts which may hold NFTs from the same contracts
- Add functionality/messaging for loading NFT collections when user changes accounts/chainIds, and wrap it in a hook
  before:
  https://user-images.githubusercontent.com/34557516/151612992-b6c9e5bc-b723-499b-b820-28ef02d776cf.mov
  after:
  https://user-images.githubusercontent.com/34557516/151611919-6a17ff8a-772b-4171-8be2-7d99b64221f1.mov

- Centers letter logo for collections without a proper logo:
before:
Screen Shot 2022-01-28 at 1 49 45 PM
after:
Screen Shot 2022-01-28 at 1 47 39 PM

### DIFF
--- a/app/scripts/controllers/preferences.js
+++ b/app/scripts/controllers/preferences.js
@@ -137,12 +137,6 @@ export default class PreferencesController {
    * @param {boolean} val - Whether or not the user prefers to autodetect collectibles.
    */
   setUseCollectibleDetection(val) {
-    const { openSeaEnabled } = this.store.getState();
-    if (val && !openSeaEnabled) {
-      throw new Error(
-        'useCollectibleDetection cannot be enabled if openSeaEnabled is false',
-      );
-    }
     this.store.updateState({ useCollectibleDetection: val });
   }
 
--- a/app/scripts/controllers/preferences.js
+++ b/app/scripts/controllers/preferences.js
@@ -134,10 +134,10 @@ export default class PreferencesController {
   /**
    * Setter for the `useCollectibleDetection` property
    *
-   * @param {boolean} val - Whether or not the user prefers to autodetect collectibles.
+   * @param {boolean} useCollectibleDetection - Whether or not the user prefers to autodetect collectibles.
    */
-  setUseCollectibleDetection(val) {
-    this.store.updateState({ useCollectibleDetection: val });
+  setUseCollectibleDetection(useCollectibleDetection) {
+    this.store.updateState({ useCollectibleDetection });
   }
 
   /**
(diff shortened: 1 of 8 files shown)


jpuri (dismissed):

app/scripts/controllers/preferences.js
```diff
@@ -137,12 +137,6 @@ export default class PreferencesController {
    * @param {boolean} val - Whether or not the user prefers to autodetect collectibles.
    */
   setUseCollectibleDetection(val) {
-    const { openSeaEnabled } = this.store.getState();
-    if (val && !openSeaEnabled) {
-      throw new Error(
-        'useCollectibleDetection cannot be enabled if openSeaEnabled is false',
-      );
-    }
     this.store.updateState({ useCollectibleDetection: val });
```
darkwing, line 140: Lamest nit of all time but we could call the argument `useCollectibleDetection` and get rid of the `: val`, but not a big deal

ui/components/app/collectibles-items/collectibles-items.js
```diff
@@ -39,22 +43,44 @@ export default function CollectiblesItems({
...
+  const chainId = useSelector(getCurrentChainId);
 
   useEffect(() => {
     if (
-      !Object.keys(collectiblesDropdownState).length &&
-      previousCollectionKeys !== collectionsKeys
+      selectedAddress !== undefined &&
+      collectiblesDropdownState?.[selectedAddress] === undefined
+    ) {
+      collectiblesDropdownState[selectedAddress] = {};
```
darkwing, line 54: What does this empty object represent?

adonesky1: We reach this block if collectibleDropDown state has not been initialized for the currently active address. This just initializes that state so that we can then safely initialize the state a layer deeper for the chainId as well. Because this state can be different across accounts and chains this state structure is nested two layers: selectedAddress -> chainId. This is also the shape for how we store state for collectibles data across accounts and chains. An example of what this looks like:

```{
    "0x806627172af48bd5b0765d3449a7def80d6576ff": {
        "0x1": {
            "0x495f947276749Ce646f68AC8c248420045cb7b5e": false
        },
        "0x4": {
            "0xf5de760f2e916647fd766B4AD9E85ff943cE3A2b": true
        }
    },
    "0xe7d522230eff653bb0a9b4385f0be0815420dd98": {
        "0x4": {
            "0xf5de760f2e916647fd766B4AD9E85ff943cE3A2b": false
        }
    }
}
```

adonesky1: All that being said, you're right that we don't really need to initialize it like that. I've updated the logic to do this more cleanly: https://github.com/MetaMask/metamask-extension/pull/13435/commits/cbcbac0b785285b21ad2f79b7ea36bddeca433ee

jpuri (approved):

darkwing (approved):

PR #13464, opened 2022-01-28
Avoid repeating CSS color vars

darkwing:
@georgewrmarshall and I noticed that the CSS variables were being repeated in the eventual output, likely because the design system is included in multiple places.  This wasn't a problem in the past because `colors.scss` only contained variables.

To test:
1.  Load the extension, ensure all the colors are as they should be.
2. Manually open the generated `index.css` file and ensure the vars are only declared once (you can search for `:root`)


digiwand (approved):
LGTM! 🌈

adonesky1 (approved):
Nice!

PR #13473, opened 2022-01-31
Add shell.nix for Nix environment

yellowgh0st:
Fixes: no issue opened

Explanation:  
Adds _shell.nix_ expression which allows to use `nix-shell` on system with Nix package manager to start an interactive shell in which the environment has corresponding dependencies.

Respectively:
Node.js 14_x
Yarn on Node.js 14_x

Adds the information about it into readme.

Manual testing steps:  
  1) Execute `nix-shell` in the project root
  2) yarn setup


darkwing:
Thank you for your contribution @yellowgh0st !  It appears that this is a local developer dependency fix that the wider project doesn't need at this moment.  For now we'll close this PR but if it because a compelling need for us, will re-open and merge.  Thank you again!

PR #13481, opened 2022-02-01
Removing EIP_1559_V2 feature flag

jpuri:
Remove feature flag for EIP_1559_V2


test/e2e/fixtures/send-edit-v2/state.json
```diff
@@ -130,6 +130,19 @@
       "useNonceField": false,
       "usePhishDetect": true
     },
+    "MetaMetricsController": {
+      "fragments": {
+        "transaction-added-4046084157914634": {
```
brad-decker, line 135: [REQ] I think we should be doubly safe and make sure that fragments exist in userland before calling them. Scenario: User submits a transaction with low max fee (doesn't care how long it takes) before the new version goes out... they don't have a fragment for transaction-added. This would cause a failure.

There's a function on the transaction controller called createTransactionEventFragment that I had intended the UI hook to call to ensure the existence of the fragment. If you call this method with the event as TRANSACTION_EVENTS.APPROVED constant it'll create the fragment if it doesn't exist. Its a safety net. Just ensure we only call this method the first time the hook is executed to avoid over calling the method.

brad-decker (changes requested):

brad-decker (approved):

ui/hooks/useTransactionEventFragment.js
```diff
@@ -1,10 +1,28 @@
...
+  const fragment = useSelector((state) =>
+    selectMatchingFragment(state, {
+      fragmentOptions: {},
+      existingId: `transaction-added-${transaction?.id}`,
+    }),
+  );
+
+  useEffect(() => {
+    if (!fragment && transaction) {
+      createTransactionEventFragment(transaction.id, TRANSACTION_EVENTS.ADDED);
```
brad-decker, line 23: Oops just noticed this should be .APPROVED -- .ADDED will end up firing the ADDED event which we don't want to capture when creating missing fragments. In case it was already fired.

brad-decker (changes requested):

ui/hooks/useTransactionEventFragment.js
```diff
@@ -1,10 +1,28 @@
-import { useCallback } from 'react';
+import { useCallback, useEffect } from 'react';
```
mcmire, line 1: Should these changes be in a separate PR? Are these just things that got missed or things that we couldn't do while the feature flag was there?

ui/hooks/useTransactionEventFragment.js
jpuri: Just updated the PR as discussed.

ui/hooks/useTransactionEventFragment.js
jpuri: @mcmire : this was thing that was missed out previously. These were actually review feedbacks on this PR in fact.

Ideally yes we should have planned a different PR for this. But work is completed on this PR,  thus I would prefer to not spend time now on splitting it.

brad-decker (approved):

ui/hooks/useTransactionEventFragment.js
mcmire: Okay. It might be good to split this off into a separate commit at least if you can, so if someone is trying to do a `git blame` on this later they don't get confused.

mcmire (approved):

PR #13482, opened 2022-02-01
Refactor ESLint config

mcmire:
We would like to insert TypeScript into the ESLint configuration, and
because of the way that the current config is organized, that is not
easy to do.

Most files are assumed to be files that are suited for running in a
browser context. This isn't correct, as we should expect most files to
work in a Node context instead. This is because all browser-based files
will be run through a transpiler that is able to make use of
Node-specific variables anyway.

There are a couple of important ways we can categories files which our
ESLint config should be capable of handling well:

* Is the file a script or a module? In other words, does the file run
  procedurally or is the file intended to be brought into an existing
  file?
* If the file is a module, does it use the CommonJS syntax (`require()`)
  or does it use the ES syntax (`import`/`export`)?

When we introduce TypeScript, this set of questions will become:

* Is the file a script or a module?
* If the file is a module, is it a JavaScript module or a TypeScript
  module?
* If the file is a JavaScript module, does it use the CommonJS syntax
  (`require()`) or does it use the ES syntax (`import`/`export`)?

To represent these divisions, this commit removes global rules — so now
all of the rules are kept in `overrides` for explicitness — and sets up
rules for CommonJS- and ES-module-compatible files that intentionally do
not overlap with each other. This way TypeScript (which has its own set
of rules independent from JavaScript and therefore shouldn't overlap
with the other rules either) can be easily added later.

Finally, this commit splits up the ESLint config into separate files and
adds documentation to each section. This way sets of rules which are
connected to a particular plugin (`jsdoc`, `@babel`, etc.) can be easily
understood instead of being obscured.

---

Testing steps:

* Run `yarn lint`. This command should not fail. (This is probably checked on CI anyway.)
* Run `yarn start`. This command should not fail with a linting error.
* Run `yarn dist`. This command should not fail with a linting error.

### DIFF
--- a/.eslintrc.babel.js
+++ b/.eslintrc.babel.js
@@ -1,0 +1,9 @@
+module.exports = {
+  parser: '@babel/eslint-parser',
+  plugins: ['@babel'],
+  rules: {
+    '@babel/no-invalid-this': 'error',
+    // Prettier handles this
+    '@babel/semi': 'off',
+  },
+};
--- a/.eslintrc.base.js
+++ b/.eslintrc.base.js
@@ -1,0 +1,67 @@
+const path = require('path');
+
+module.exports = {
+  extends: [
+    '@metamask/eslint-config',
+    path.resolve(__dirname, '.eslintrc.jsdoc.js'),
+  ],
+
+  globals: {
+    document: 'readonly',
+    window: 'readonly',
+  },
+
+  rules: {
+    'default-param-last': 'off',
+    'prefer-object-spread': 'error',
+    'require-atomic-updates': 'off',
+
+    // This is the same as our default config, but for the noted exceptions
+    'spaced-comment': [
+      'error',
+      'always',
+      {
+        markers: [
+          'global',
+          'globals',
+          'eslint',
+          'eslint-disable',
+          '*package',
+          '!',
+          ',',
+          // Local additions
+          '/:', // This is for our code fences
+        ],
+        exceptions: ['=', '-'],
+      },
+    ],
+
+    'no-invalid-this': 'off',
+
+    // TODO: remove this override
+    'padding-line-between-statements': [
+      'error',
+      {
+        blankLine: 'always',
+        prev: 'directive',
+        next: '*',
+      },
+      {
+        blankLine: 'any',
+        prev: 'directive',
+        next: 'directive',
+      },
+      // Disabled temporarily to reduce conflicts while PR queue is large
+      // {
+      //   blankLine: 'always',
+      //   prev: ['multiline-block-like', 'multiline-expression'],
+      //   next: ['multiline-block-like', 'multiline-expression'],
+      // },
+    ],
+
+    // It is common to import modules without assigning them to variables in
+    // a browser context. For instance, we may import polyfills which change
+    // global variables, or we may import stylesheets.
+    'import/no-unassigned-import': 'off',
+  },
+};
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,132 +1,107 @@
+const path = require('path');
 const { version: reactVersion } = require('react/package.json');
 
 module.exports = {
   root: true,
-  parser: '@babel/eslint-parser',
-  parserOptions: {
-    sourceType: 'module',
-    ecmaVersion: 2017,
-    ecmaFeatures: {
-      experimentalObjectRestSpread: true,
-      impliedStrict: true,
-      modules: true,
-      blockBindings: true,
-      arrowFunctions: true,
-      objectLiteralShorthandMethods: true,
-      objectLiteralShorthandProperties: true,
-      templateStrings: true,
-      classes: true,
-      jsx: true,
-    },
-  },
-
-  ignorePatterns: [
-    '!.eslintrc.js',
-    '!.mocharc.js',
-    'node_modules/**',
-    'dist/**',
-    'builds/**',
-    'test-*/**',
-    'docs/**',
-    'coverage/',
-    'jest-coverage/',
-    'development/chromereload.js',
-    'app/vendor/**',
-    'test/e2e/send-eth-with-private-key-test/**',
-    'nyc_output/**',
-    '.vscode/**',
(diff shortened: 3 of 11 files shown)


.eslintrc.js
```diff
@@ -1,132 +1,107 @@
...
-    'padding-line-between-statements': [
-      'error',
-      {
-        blankLine: 'always',
-        prev: 'directive',
-        next: '*',
+      rules: {
+        // This rule does not work with CommonJS modules. We will just have to
+        // trust that all of the files specified above are indeed modules.
+        'import/unambiguous': 'off',
```
mcmire, line 46: In refactoring this config, I discovered that ESLint was confused over whether this file was a script or a module. It seemed that this was not possible because I thought that [we'd configured `eslint-plugin-import` to work with CommonJS](https://github.com/MetaMask/eslint-config/blob/ec27156931a0cda1b34ff46937b8296742dcb1c0/packages/base/src/index.js#L278). That is true, but that setting only applies to the `no-unresolved` rule. The `no-ambiguous` rule is separate and [only seems to work on ES module files](https://github.com/import-js/eslint-plugin-import/blob/41d4500d83e969a0b24be837625f754b9eadada6/utils/unambiguous.js) . As far as I can tell, this is a long-standing bug and not a feature. In the meantime, we just have to turn this rule off for CommonJS files.

lavamoat/build-system/policy-override.json
```diff
@@ -14,6 +14,7 @@
     },
     "@eslint/eslintrc": {
       "packages": {
+        "<root>": true,
```
brad-decker, line 17: [Q] What does this mean? Why was <root> added and is it a good thing?

brad-decker (dismissed):
LGTM.

lavamoat/build-system/policy-override.json
Gudahtt: That doesn't sound right. LavaMoat isn't asking for access to anything, it has access already. All packages have access to everything by default, that's what LavaMoat is preventing. LavaMoat is _gating_ access, it's controlling which packages and APIs each package can access.

Here is the relevant portion of the quoted error message:

`package "@eslint/eslintrc" requested "<root>" as "/Users/elliot/code/metamask/metamask-extension/.eslintrc.base.js"`

Which means pretty much what it sounds like - the `@eslint/eslintrc` package is requesting the `<root>` package, but it wasn't explicitly granted access to that package so it was denied. The `<root>` package here _is not_ `@eslint/eslintrc`, it's the root _project_, the MetaMask extension itself.

I'm guessing this is the case because now the `.eslintrc.js` file is _extending_ another module in this repository. So when `@eslint/eslintrc` tries parsing that config file, it will have to import the base config as well, thus requiring access to the MetaMask extension itself. It didn't need that before, because we weren't extending from another config in this repo before. We were just extending from `@metamask/eslint-config` (which was already added to the policy override a few lines down from here, for this same reason).

mcmire: Ahh, okay. That makes more sense.

mcmire:
There were some good suggestions provided in the walkthrough call today on how we should add TS to ESLint in a way that doesn't decrease performance. I'm going to move this back into draft status as this will prompt some changes.

test/e2e/tests/send-eth.spec.js
```diff
@@ -91,6 +91,8 @@ describe('Send ETH from inside MetaMask using default gas', function () {
   });
 });
 
+// TODO wrap all of this in a describe
+/* eslint-disable mocha/max-top-level-suites */
```
mcmire, line 95: With the updates to the ESLint config, these test files are now failing the lint step. I suspect this is the case because not all of the Mocha rules were getting applied to the e2e tests previously and now they are.

.eslintrc.babel.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
```
mcmire, line 1: Babel-specific configuration was pulled out into its own file from `.eslintrc`, but there should be no changes with what was there before.

.eslintrc.base.js
```diff
@@ -0,0 +1,67 @@
+const path = require('path');
```
mcmire, line 1: Base configuration was pulled out into its own file from `.eslintrc`, but there should be no changes with what was there before.

.eslintrc.jsdoc.js
```diff
@@ -0,0 +1,23 @@
+module.exports = {
```
mcmire, line 1: JSDoc-specific configuration was pulled out into its own file from `.eslintrc`, but there should be no changes with what was there before.

development/build/transforms/utils.js
```diff
@@ -1,11 +1,17 @@
 const { ESLint } = require('eslint');
 const eslintrc = require('../../../.eslintrc.js');
 
-// We don't want linting to fail for purely stylistic reasons.
-eslintrc.rules['prettier/prettier'] = 'off';
-// Sometimes we use `let` instead of `const` to assign variables depending on
-// the build type.
-eslintrc.rules['prefer-const'] = 'off';
+eslintrc.overrides.forEach((override) => {
```
mcmire, line 4: Changes are necessary to this file (which is part of the build system) because we run ESLint during the build system. To do this we pull in the ESLint configuration object from `.eslintrc.js` and we explicitly modify that object to disable some rules as this would cause failures which we are not interested in. Because ALL ESLint rules are now in `overrides`, I have to change how these modifications work.

test/e2e/tests/add-hide-token.spec.js
```diff
@@ -55,6 +55,7 @@ describe('Hide token', function () {
   });
 });
 
+/* eslint-disable-next-line mocha/max-top-level-suites */
```
mcmire, line 58: With the updates to the ESLint config, these test files are now failing the lint step. I suspect this is the case because not all of the Mocha rules were getting applied to the e2e tests previously and now they are

brad-decker (approved):
LGTM

darkwing (approved):

PR #13484, opened 2022-02-01
Handling array type values for image in Identicon component

ryanml:
Sentry error: https://sentry.io/organizations/metamask/issues/2977654075/?project=273505&query=is%3Aunresolved+release%3A10.9.0+toLowerCase&sort=new&statsPeriod=14d

It is possible that some assets will supply an array of images for the `asset.image` property. In the case of ONE (Harmony):

Screen Shot 2022-02-01 at 1 37 34 PM

Rather than just type check and ignore, this PR modifies the `Identicon` component to make use of the first image in the list.

Test plan:
1. Via sushiswap, initiate an ETH -> ONE swap
2. Add the ONE token to metamask
3. Open the Assets tab
4. Observe the app error state

With the fix:
Screen Shot 2022-02-01 at 1 44 16 PM

### DIFF
--- a/ui/components/ui/identicon/identicon.component.js
+++ b/ui/components/ui/identicon/identicon.component.js
@@ -34,7 +34,7 @@ export default class Identicon extends PureComponent {
     /**
      * Used as the image source of the Identicon
      */
-    image: PropTypes.string,
+    image: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
     /**
      * Use the blockie type random image generator
      */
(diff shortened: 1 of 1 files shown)


ui/components/ui/identicon/identicon.component.js
```diff
@@ -76,6 +76,10 @@ export default class Identicon extends PureComponent {
     const { className, diameter, alt, imageBorder, ipfsGateway } = this.props;
     let { image } = this.props;
 
+    if (Array.isArray(image) && image.length) {
+      image = image[0];
+    }
+
     if (image.toLowerCase().startsWith('ipfs://')) {
```
adonesky1, line 83: maybe just in case let's do a type check here too? 
```suggestion
    if (typeof image === 'string' && image.toLowerCase().startsWith('ipfs://')) {
```

ryanml: sgtm, done

adonesky1 (approved):
LGTM

danjm (approved):

PR #13489, opened 2022-02-02
Add TypeScript to the build system

mcmire:
This commit modifies the build system so that TypeScript files can be
transpiled into ES5 just like JavaScript files.

Note that this commit does NOT change the build system to run TypeScript
files through the TypeScript compiler. In other words, no files will be
type-checked at the build stage, as we expect type-checking to be
handled elsewhere (live, via your editor integration with `tsserver`,
and before a PR is merged, via `yarn lint`). Rather, we merely instruct
Babel to strip TypeScript-specific syntax from any files that have it,
as if those files had been written using JavaScript syntax alone.

Why take this approach? Because it prevents the build process from being
negatively impacted with respect to performance (as TypeScript takes a
significant amount of time to run).

It's worth noting the downside of this approach: because we aren't
running files through TypeScript, but relying on Babel's [TypeScript
transform][1] to identify TypeScript syntax, this transform has to keep
up with any syntax changes that TypeScript adds in the future. In fact
there are a few syntactical forms that Babel already does not recognize.
These forms are rare or are deprecated by TypeScript, so I don't
consider them to be a blocker, but it's worth noting just in case it
comes up later. Also, any settings we place in `tsconfig.json` will be
completely ignored by Babel. Again, this isn't a blocker because there
are some analogs for the most important settings reflected in the
options we can pass to the transform. These and other caveats are
detailed in the [documentation for the transform][2].

[1]: https://babeljs.io/docs/en/babel-plugin-transform-typescript
[2]: https://babeljs.io/docs/en/babel-plugin-transform-typescript#caveats

---

Manual testing steps:

- Check out this branch locally.
- Look up the commit called "Add some sample files so linting can be tested" and cherry-pick it (if the commit disappears, try `git cherry-pick 9af9b6bbf1b26df408a40101e7eb8e727d334e05`).
  - This commit converts two files to TypeScript, `ui/helpers/utils/tx-helper.js` and `ui/helpers/utils/storage-helpers.ts`. Both of these files have dependents which are JavaScript files, and `ui/helpers/utils/tx-helper.js` has dependencies which are JavaScript files.
- Run `yarn setup`.
- Open `ui/helpers/utils/storage-helpers.ts` in your editor. Remove the semicolon from the first line; your editor should tell you (via ESLint) that this is invalid.
- Now open `ui/index.js`. This file imports `tx-helper.ts`. Mouse over `txHelper` in the `import txHelper` line; you should see proper types for this function which come from the TypeScript file.
- Run `yarn lint`. All files should pass.
- Run `yarn start`. You shouldn't see any errors.
- Open MetaMask in your browser (this assumes that you have the dev version installed). Click around. You shouldn't see any UI changes.
- Run `yarn dist`. You shouldn't see any errors.
- Open MetaMask in your browser (this assumes that you have the dev version installed). Click around. You shouldn't see any UI changes.
- If you like, open `dist/chrome/common-2.js` and search for "getStorageItem" (one of the functions in `ui/helpers/utils/storage-helpers.ts`). **Warning: this is a big file.** You should find a reference to the function here.


lavamoat/build-system/policy-override.json
```diff
@@ -3,13 +3,14 @@
...
-        "@babel/plugin-transform-runtime": true,
         "@babel/plugin-proposal-class-properties": true,
+        "@babel/plugin-proposal-nullish-coalescing-operator": true,
         "@babel/plugin-proposal-object-rest-spread": true,
         "@babel/plugin-proposal-optional-chaining": true,
-        "@babel/plugin-proposal-nullish-coalescing-operator": true
+        "@babel/plugin-transform-runtime": true,
+        "@babel/preset-env": true,
+        "@babel/preset-react": true,
+        "@babel/preset-typescript": true
```
mcmire, line 13: I re-alphabetized this list, but `@babel/preset-typescript` is the new entry here.

babel.config.js
```diff
@@ -1,7 +1,11 @@
 module.exports = function (api) {
   api.cache(false);
   return {
+    parserOpts: {
+      strictMode: true,
```
mcmire, line 5: Make sure TypeScript files are parsed using strict mode. See https://babeljs.io/docs/en/babel-plugin-transform-typescript#typescript-compiler-options.

brad-decker (approved):
LGTM

digiwand (approved):
LGTM 🙌 One of the questions asked at Ethereum Rio 2022 was if or when we are supporting TypeScript in our repo? I told them you've been making tremendous headway here in Q1!

PR #13495, opened 2022-02-02
Add TypeScript to the linting process

mcmire:
This commit allows developers to write TypeScript files and lint them
(either via a language server in their editor of choice or through the
`yarn lint` command).

The new TypeScript configuration as well as the updated ESLint
configuration not only includes support for parsing TypeScript files,
but also provides some compatibility between JavaScript and TypeScript.
That is, it makes it possible for a TypeScript file that imports a
JavaScript file or a JavaScript file that imports a TypeScript file to
be linted.

Note that this commit does not integrate TypeScript into the build
system yet, so we cannot start converting files to TypeScript and
pushing them to the repo until that final step is complete.

---

Manual testing steps:

- Check out this branch locally.
- Run `git cherry-pick 4bde1eb1d0bf0e84d76f83bb0bb065cfc1f45e58`.
  - This commit (which is disconnected from this branch) converts two files to TypeScript, `ui/helpers/utils/tx-helper.ts` and `ui/helpers/utils/storage-helpers.ts`. Both of these files have dependents which are JavaScript files, and `ui/helpers/utils/tx-helper.ts` has dependencies which are JavaScript files.
- Run `yarn setup`.
- Open `ui/helpers/utils/storage-helpers.ts` in your editor. Remove the semicolon from the first line; your editor should tell you (via ESLint) that this is invalid.
- Now open `ui/index.js`. This file imports `tx-helper.ts`. Mouse over `txHelper` in the `import txHelper` line; you should see proper types for this function which come from the TypeScript file.
- Run `yarn lint`. All files should pass.


patches/typescript+4.4.4.patch
```diff
@@ -0,0 +1,194 @@
+diff --git a/node_modules/typescript/lib/typescript.js b/node_modules/typescript/lib/typescript.js
+index 323de6f..bb3e6b4 100644
+--- a/node_modules/typescript/lib/typescript.js
++++ b/node_modules/typescript/lib/typescript.js
+@@ -24,11 +24,11 @@ var __spreadArray = (this && this.__spreadArray) || function (to, from, pack) {
+     return to.concat(ar || Array.prototype.slice.call(from));
+ };
+ var __assign = (this && this.__assign) || function () {
+-    __assign = Object.assign || function(t) {
```
mcmire, line 9: These changes to the `typescript` package were motivated by failures I was getting from LavaMoat when attempting to generate LavaMoat policy files for the build system (via `yarn lavamoat:auto`. That said, this commit is concerned with adding TypeScript to ESLint, not the build system, so why are these changes necessary?

Because, as I learned, the build system actually runs ESLint. As I've been going through the codebase, I've occasionally come across sections of code that are surrounded with `///: BEGIN:ONLY_INCLUDE_IN` and `///: END:ONLY_INCLUDE_IN` (for example, [`account-menu.component.js`](https://github.com/MetaMask/metamask-extension/blob/2b010710aec96f3bf74bb07d7c56cfa2c1c59d7a/ui/components/app/account-menu/account-menu.component.js#L15-L17). I've always wondered about them, and now I know that they get removed [during a step in the build process](https://github.com/MetaMask/metamask-extension/blob/2b010710aec96f3bf74bb07d7c56cfa2c1c59d7a/development/build/transforms/remove-fenced-code.js). After this transform is applied to a file, [that file is linted](https://github.com/MetaMask/metamask-extension/blob/2b010710aec96f3bf74bb07d7c56cfa2c1c59d7a/development/build/transforms/utils.js#L61).

So, when running `yarn lavamoat:auto`, LavaMoat sees these files and follows the path to `eslint.config.js`, then follows the path to the `typescript` package. Why does it trip up on this file in particular?

One of the things that LavaMoat, or more accurately the `lockdown` function in SES, does is to protect against unwanted and unexpected modifications of global objects, such as `Object.prototype`. Admittedly I am unable to explain much beyond this, but you can learn more about this function [here](https://github.com/endojs/endo/blob/master/packages/ses/README.md#lockdown). The `lockdown` function, as it uses `Object.freeze` to lock down objects, suffers from a ["mistake"](https://web.archive.org/web/20141230041441/http://wiki.ecmascript.org/doku.php?id=strawman:fixing_override_mistake) in the ECMAScript spec where a property inherited from a frozen object is itself frozen. Plain objects inherit `constructor` from `Object.prototype`, so if `Object.prototype` is frozen, this code will not work:

``` javascript
const foo = {};
foo["constructor"] = "bar";
```

But that is what is happening in this file. The `ts.textToKeywordObj` object maps names of keywords in TypeScript to token ids that represent those keywords (defined elsewhere), and one of the keywords in TypeScript is `constructor`. For some reason, whatever ended up producing this file (probably TypeScript itself) defined this object essentially like this:

``` javascript
ts.textToKeywordObj = (
  _a = { /* ... */ }, 
  _a["constructor"] = 133,
  // ...
  _a
);
```

That doesn't work, so one of the changes made here is to clean up the definition of this object.

The other change is to the `__assign` function, which you can see here. One of the ways that `__assign` is used is to make a `textToToken` variable on line 9904. This looks something like:

``` javascript
var textToToken = __assign(
  {},
  ts.textToKeywordObj,
  {
    "{": 18 /* OpenBraceToken */,
    "}": 19 /* CloseBraceToken */,
    // ...
  }
)
```

As you can see here, `__assign` defaults to `Object.assign` here if it exists, or otherwise falls back to a polyfill of `Object.assign`. Whether or not the polyfill is used, `__assign` creates a new object by iterating over the given objects and manually copying their keys and values to the new object by setting them directly on the new object. This will attempt to set `constructor` on the new object again, which as we've seen, is not allowed. So this change fixes `__assign` by using `Object.defineProperty` to set properties instead of direct assignment.

tsconfig.json
```diff
@@ -0,0 +1,33 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "allowSyntheticDefaultImports": true,
+    "inlineSources": true,
+    "isolatedModules": true,
+    "jsx": "react",
+    "lib": ["dom", "es2020"],
```
mcmire, line 8: These settings were copied from the [`tsconfig.json`](https://github.com/MetaMask/controllers/blob/81bcbae1c4abe71a9b07819606e155a33ec4a6b4/tsconfig.json) in `controllers` but then adjusted to account for JSX as well as browser globals. ES2020 was chosen because Node v14 supports it, so that should make it less surprising for developers who wish to be able to write frontend code in a similar fashion as Node code. (The JavaScript that TypeScript generates will end up getting processed by Babel in the build system so we should be safe here.)

tsconfig.json
```diff
@@ -0,0 +1,33 @@
...
+    "moduleResolution": "node",
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "outDir": "tsout",
```
mcmire, line 12: This setting will not end up getting used by the build system — [`tsify`](https://github.com/TypeStrong/tsify) will output JavaScript in-memory and then funnel the result through the build pipeline. This is only here so that if you want to test things out you can say `yarn tsc --project .` like [this](https://github.com/MetaMask/controllers/blob/81bcbae1c4abe71a9b07819606e155a33ec4a6b4/package.json#L32).

mcmire:
There were some good suggestions provided in the walkthrough call today on how we should add TS to ESLint in a way that doesn't decrease performance. I'm going to move this back into draft status as this will prompt some changes.

patches/eslint-import-resolver-typescript+2.5.0.patch
```diff
@@ -0,0 +1,23 @@
...
+@@ -49,13 +49,13 @@ function __spreadArray(to, from) {
+ 
+ var IMPORTER_NAME = 'eslint-import-resolver-typescript';
+ var log = debug__default['default'](IMPORTER_NAME);
+-var defaultExtensions = __spreadArray(__spreadArray([
++var defaultExtensions = [
+     '.ts',
+     '.tsx',
+-    '.d.ts'
+-], Object.keys(require.extensions)), [
```
mcmire, line 14: When I opened this file, I got an ESLint warning that `require.extensions` was deprecated. When I ran `yarn lavamoat:auto`, then `require.extensions` was undefined. I am not sure what exactly the correct fix here is, but I don't think we need to make this complicated. The list of extensions we want to support are JavaScript and TypeScript.

patches/@keystonehq+bc-ur-registry+0.4.4.patch
```diff
@@ -0,0 +1,1830 @@
+diff --git a/node_modules/@keystonehq/bc-ur-registry/src/Bytes.ts b/node_modules/@keystonehq/bc-ur-registry/src/Bytes.ts
```
mcmire, line 1: The `@keystonehq/bc-ur-registry` package contains TypeScript files in its distributed form. This is not necessary or recommended for TypeScript-compatible libraries (rather, it is enough to publish type _definition_ files). The issue for this is logged here: https://github.com/KeystoneHQ/ur-registry/issues/13. In the meantime, as our code works fine presently without any TypeScript files, we do not need these in order to use this library. This makes the new `tsc` step in `yarn lint` pass.

patches/await-semaphore+0.1.3.patch
```diff
@@ -0,0 +1,68 @@
+diff --git a/node_modules/await-semaphore/index.ts b/node_modules/await-semaphore/index.ts
```
mcmire, line 1: The `await-semaphore` package contains TypeScript files in its distributed form. This is not necessary or recommended for TypeScript-compatible libraries (rather, it is enough to publish type _definition_ files). The issue for this is logged here: https://github.com/notenoughneon/await-semaphore/issues/6. In the meantime, as our code works fine presently without any TypeScript files, we do not need these in order to use this library. This makes the new `tsc` step in `yarn lint` pass.

patches/@keystonehq+bc-ur-registry+0.4.4.patch
danjm: So, to make sure I understand correctly, this code is not used by the extension, but we need this patch to get lint passing?

patches/await-semaphore+0.1.3.patch
danjm: So, to make sure I understand correctly, this code is not used by the extension, but we need this patch to get lint passing?

patches/@keystonehq+bc-ur-registry+0.4.4.patch
mcmire: Sorry maybe my explanation was confusing. We do use this package (it looks like a [component that is involved with the QR code stuff](https://github.com/MetaMask/metamask-extension/blob/b8fc1ae671c153d84de6caa254de3e55c153de75/ui/components/app/qr-hardware-popover/qr-hardware-sign-request/reader.js#L2) imports `@keystonehq/bc-ur-registry-eth`, which imports `@keystonehq/bc-ur-registry`). However, the way this package has been bundled and published to NPM is incorrect as it includes TypeScript files.

patches/await-semaphore+0.1.3.patch
mcmire: We do use this package (it looks like [at the top of MetamaskController](https://github.com/MetaMask/metamask-extension/blob/b8fc1ae671c153d84de6caa254de3e55c153de75/app/scripts/metamask-controller.js#L13)). However, the way this package has been bundled and published to NPM is incorrect as it includes TypeScript files.

darkwing (dismissed):
These changes look great to me, and thank you so much for all the code commenting!  Hugely useful!

tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+    "allowSyntheticDefaultImports": true,
+    "inlineSources": true,
+    "isolatedModules": true,
+    "jsx": "react",
+    "lib": ["dom", "es2020"],
+    "moduleResolution": "node",
+    "outDir": "tsout",
+    "rootDir": ".",
+    "sourceMap": true,
+    "strict": true
```
digiwand, line 13: Taking a quick peek of some other tsconfig.json file configs for chrome-extensions:
- https://github.com/chibat/chrome-extension-typescript-starter/blob/master/tsconfig.json
- https://betterprogramming.pub/creating-chrome-extensions-with-typescript-914873467b65

I think you might have explained this last Friday, but I forgot. Sorry if this was the case. Why are we not including
```
"module": "commonjs",
"target": "es6",
```
 again? 

----

"noEmitOnError" - might be something to consider later. This holds back from generating js files to `outDir` if there are ts errors found

patches/await-semaphore+0.1.3.patch
digiwand: It looks like we're removing the Semaphore class that Mutex extends in this patch. Does this not break when `this.createVaultMutex.acquire();` runs in the MetamaskController?

digiwand (commented):
Super! 🙌  I had one question about one of the patches and some other optional comments.

Thanks for walking through this PR with me last Friday and enlightening me on TypeScript + patching files in relation to our codebase.

Looking forward to removing the following paths from the excluded list someday too: 
```
    "**/*.test.js",
    "**/*.test.ts",
    "**/*.test.tsx",
    ".storybook/**/*",
```

tsconfig.json
mcmire: > I think you might have explained this last Friday, but I forgot. Sorry if this was the case. Why are we not including
>
> ```
> "module": "commonjs",
> "target": "es6",
> ```

We are, but I admit it's a bit hidden: that is being done by the `"extends": "@tsconfig/node14/tsconfig.json"` line. You can see the source for that [here](https://github.com/tsconfig/bases/blob/main/bases/node14.json).

> "noEmitOnError" - might be something to consider later. This holds back from generating js files to outDir if there are ts errors found

I don't mind adding it now! Added in 0ccb400a500e1d0683aa63a7c37c1f04b34978e5

.eslintrc.typescript-compat.js
```diff
@@ -0,0 +1,24 @@
...
+    // With this extension, ESLint will look for TypeScript files when it
+    // encounters an import. For some reason, this is now introducing
+    // `import/no-named-as-default-member` errors. The `typescript-eslint` docs
+    // recommend turning this off anyway, as TypeScript already provides this
+    // check.
+    // TODO: Move this into our shared config
+    'import/no-named-as-default-member': 'off',
+    // Turn these off as per the `typescript-eslint` docs, as TypeScript already
+    // provides this check.
+    // TODO: Move these into our shared config
```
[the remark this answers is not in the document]
mcmire: Yes, it is. I've removed this line, but in the future I'll make sure to be more explicit about this.

.eslintrc.js
```diff
@@ -75,10 +94,85 @@ module.exports = {
...
+      rules: {
+        /**
+         * Turn these off, as it's recommended by typescript-eslint.
+         *
+         * @see {@link https://typescript-eslint.io/docs/linting/troubleshooting#eslint-plugin-import}
+         */
+        'import/named': 'off',
+        'import/namespace': 'off',
+        'import/default': 'off',
+        'import/no-named-as-default-member': 'off',
```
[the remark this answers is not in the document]
mcmire: No, although arguably it shouldn't be in `typescript-compat.js` as it shouldn't apply to every kind of file, just TypeScript files. Removed in cec7044ed7c6eeb12364945a7ee950e75eb30c2f.

patches/await-semaphore+0.1.3.patch
mcmire: So, we aren't removing all the code in `await-semaphore`. You can see [here](https://unpkg.com/browse/await-semaphore@0.1.3/) that the version of this package published to NPM includes both JavaScript and TypeScript files. We are just removing the TypeScript files, but we are still using the JavaScript files. So `this.createVaultMutex.acquire();` should still work as before.

mcmire:
@digiwand Hopefully I've addressed all your concerns! Would you mind taking another look?

digiwand (commented):
Look good! Thank you. 

Last question I have: Is it possible to add comments to patch files? I think it would be helpful to provide context for each patch.

digiwand (approved):
Great work! Thank you for this! 🎉

mcmire:
@darkwing May I get one more approval?

brad-decker (approved):
Looks good to me, and loving learning from @digiwand 's exploration as well and the questions asked. Great work @mcmire and @digiwand thanks for your thorough review and questions.

PR #13496, opened 2022-02-02
Replace uses of updateTransaction (from the tx controller) with less powerful methods for each specific state transition

### LINKED ISSUE #13386: Replace uses of `updateTransaction` (from the tx controller) with less powerful methods for each specific state transition
This is a follow up to this comment: https://github.com/MetaMask/metamask-extension/pull/13328#issuecomment-1019977808

Currently, the transaction controller has a `updateTransaction` method that can be called externally. This is ultimately called in the UI in at least three places: in the swaps approve transaction flow, in the send edit flow, and when editing the gas params from the confirm screen.

Because this method can overwrite the entire txMeta object, exposing it as a public method is a dangerous design flaw. We should make these updates safe by (as suggested in the first comment below) "exposing methods for each individual state transition"

To resolve this task, we should first document each use of `updateTransaction` and the state transition it needs to make happen. Once we have done that, we can specify a list of methods to be added, and start replacing the current calls to `updateTransaction` one by one.

segun:
Fixes: #13386

Explanation:  

`Rather than expose a generic "update transaction" method with additional safety checks, it would be better to expose methods for each individual state transition. That would be safer than having to imagine each possible state transition and add safety checks for each one.` - Mark.

  - 
  - 
  -


app/scripts/controllers/transactions/index.js
```diff
@@ -331,6 +331,137 @@ export default class TransactionController extends EventEmitter {
...
+   * @param {number} txId
+   * @returns {TransactionMeta} the txMeta who matches the given id if none found
+   * for the network returns undefined
+   */
+  _getTransaction(txId) {
+    const { transactions } = this.store.getState();
+    return transactions[txId];
+  }
+
+  _validateTransaction(txId) {
```
danjm, line 346: I think it would be good if the name of this method reflected the specific purpose: to check if the transaction status is `unapproved`

Perhaps rename it to `_checkIfTxStatusIsUnapproved`

app/scripts/controllers/transactions/index.js
```diff
@@ -331,6 +331,137 @@ export default class TransactionController extends EventEmitter {
...
+    return (
+      this._getTransaction(txId).status === TRANSACTION_STATUSES.UNAPPROVED
+    );
+  }
+
+  _updateTransaction(txParams, note) {
+    const normalizedTxParams = txUtils.normalizeTxParams(txParams);
+    const withAdditionalMeta = this.txStateManager.generateTxMeta({
```
danjm, line 354: `generateTxMeta` is used when we are creating a new transaction object

Also, `txParams` is just one property of `txMeta` that we might need to update. And we might not need to normalize the txParams for every call to _updateTransaction

It is possible that we don't need this `_updateTransaction` method, and can instead just call `this.txStateManager.updateTransaction` in each of the other methods added here

app/scripts/controllers/transactions/index.js
```diff
@@ -331,6 +331,137 @@ export default class TransactionController extends EventEmitter {
...
+   * @param {object} txGasFees - holds the gas fees parameters
+   * {
+   * gasLimit,
+   * gasPrice,
+   * maxPriorityFeePerGas,
+   * maxFeePerGas,
+   * estimateUsed
+   * }
+   */
+  updateTransactionGasFees(txId, txGasFees) {
```
danjm, line 374: The documentation on each of these methods is good.

I think we could go a step further and strictly enforce that only those properties can be updated via this method. This could be done by destructuring the parameter:

```
updateTransactionGasFees(txId, {
   gasLimit,
   gasPrice,
   maxPriorityFeePerGas,
   maxFeePerGas,
   estimateUsed
}) {
```

That way, no additional properties could accidentally be included in the update

app/scripts/controllers/transactions/index.js
```diff
@@ -331,6 +331,137 @@ export default class TransactionController extends EventEmitter {
     });
```
danjm, line 331: General comment:

When updating a transaction, this is the general pattern that will have to be followed:

```
const txMeta = this._getTransaction(txId);
txMeta.propertyA = newValueForA;
txMeta.propertyB = newValueForB;
this.txStateManager.updateTransaction(txMeta, note);
```

darkwing:
Hello @segun !  I used the GitHub functionality to convert this to a draft, per your PR title.

app/scripts/controllers/transactions/index.js
segun: I can't see the following properties here

gasLimit
gasPrice
estimateUsed
decEstimateBaseFee
sourceTokenSymbol
destinationTokenSymbol
destinationTokenDecimals
destinationTokenAddress
swapMetaData
swapTokenValue
userEditedGasLimit

I think the question is are these properties of txMeta or txMeta.txParams

@danjm 

> Also, for reference, here is an example txMeta object:
> 
> ```
> {
>   "id": 655970127251907,
>   "time": 1642690106761,
>   "status": "approved",
>   "metamaskNetworkId": "4",
>   "chainId": "0x4",
>   "loadingDefaults": false,
>   "dappSuggestedGasFees": null,
>   "txParams": {
>     "from": "0xb16f2b68800ea8adae005216ff3604e38e3884a6",
>     "to": "0xe18035bf8712672935fdb4e5e431b1a0183d2dfc",
>     "nonce": "0x28",
>     "value": "0x0",
>     "gas": "0x5208",
>     "maxFeePerGas": "0x59682f2a",
>     "maxPriorityFeePerGas": "0x59682f00",
>     "type": "0x2"
>   },
>   "origin": "metamask",
>   "type": "simpleSend",
>   "history": [
>     {
>       "id": 655970127251907,
>       "time": 1642690106761,
>       "status": "unapproved",
>       "metamaskNetworkId": "4",
>       "chainId": "0x4",
>       "loadingDefaults": true,
>       "dappSuggestedGasFees": null,
>       "txParams": {
>         "from": "0xb16f2b68800ea8adae005216ff3604e38e3884a6",
>         "to": "0xe18035bf8712672935fdb4e5e431b1a0183d2dfc",
>         "value": "0x0",
>         "gas": "0x5208",
>         "maxFeePerGas": "0x59682f2b",
>         "maxPriorityFeePerGas": "0x59682f00",
>         "type": "0x2"
>       },
>       "origin": "metamask",
>       "type": "simpleSend"
>     },
>     [
>       {
>         "op": "replace",
>         "path": "/loadingDefaults",
>         "value": false,
>         "note": "Added new unapproved transaction.",
>         "timestamp": 1642690106771
>       },
>       {
>         "op": "add",
>         "path": "/userFeeLevel",
>         "value": "medium"
>       }
>     ],
>     [
>       {
>         "op": "replace",
>         "path": "/txParams/maxFeePerGas",
>         "value": "0x59682f2a",
>         "note": "confTx: user approved transaction",
>         "timestamp": 1642690121224
>       },
>       {
>         "op": "add",
>         "path": "/estimatedBaseFee",
>         "value": "21"
>       }
>     ],
>     [
>       {
>         "op": "replace",
>         "path": "/status",
>         "value": "approved",
>         "note": "txStateManager: setting status to approved",
>         "timestamp": 1642690121224
>       }
>     ],
>     [
>       {
>         "op": "add",
>         "path": "/txParams/nonce",
>         "value": "0x28",
>         "note": "transactions#approveTransaction",
>         "timestamp": 1642690121236
>       },
>       {
>         "op": "add",
>         "path": "/nonceDetails",
>         "value": {
>           "params": {
>             "highestLocallyConfirmed": 0,
>             "highestSuggested": 40,
>             "nextNetworkNonce": 40
>           },
>           "local": {
>             "name": "local",
>             "nonce": 40,
>             "details": {
>               "startPoint": 40,
>               "highest": 40
>             }
>           },
>           "network": {
>             "name": "network",
>             "nonce": 40,
>             "details": {
>               "blockNumber": "0x98f680",
>               "baseCount": 40
>             }
>           }
>         }
>       }
>     ]
>   ],
>   "userFeeLevel": "medium",
>   "estimatedBaseFee": "21",
>   "nonceDetails": {
>     "params": {
>       "highestLocallyConfirmed": 0,
>       "highestSuggested": 40,
>       "nextNetworkNonce": 40
>     },
>     "local": {
>       "name": "local",
>       "nonce": 40,
>       "details": {
>         "startPoint": 40,
>         "highest": 40
>       }
>     },
>     "network": {
>       "name": "network",
>       "nonce": 40,
>       "details": {
>         "blockNumber": "0x98f680",
>         "baseCount": 40
>       }
>     }
>   }
> }
> ```

segun: Hi @danjm 

Maybe I don't need to worry much which property belongs to where. 

I can do this for instance in _updateTransaction

```
    const txMeta = this._getTransaction(txId);
    const updatedMeta = {...txMeta, ...txParams}
    this.txStateManager.updateTransaction(updatedMeta, note);
```

And leave it up to the implementer to put the properties in the proper object...for instance the method `updateTransactionGasFees` could be called with an object that look like this

```
const toUpdate = {
   userFeeLevel: "medium",
   estimatedBaseFee: "21",
   txParams: {
     "gas": "0x5208",
     "maxFeePerGas": "0x59682f2a"
   }
}
```

But to `strictly enforce that only those properties can be updated` I need to know this structure.

danjm: @segun `gasLimit` and `gasPrice` are part of `txParams`, all of the other properties you listed are not

app/scripts/controllers/transactions/index.js
```diff
@@ -331,6 +331,195 @@ export default class TransactionController extends EventEmitter {
...
+
+  _checkIfTxStatusIsUnapproved(txId) {
+    return (
+      this._getTransaction(txId).status === TRANSACTION_STATUSES.UNAPPROVED
+    );
+  }
+
+  _updateTransaction(txId, txParams, note) {
+    const txMeta = this._getTransaction(txId);
+    const updatedMeta = { ...txMeta, ...txParams };
```
danjm, line 354: `txParams` is a property of `txMeta`. `{ ...txMeta, ...txParams }` will cause the properties of `txParams` to be flattened into `txMeta.
If you just want to update txParams, then this would work: `{ ...txMeta, txParams }` (without the elippses before txParams, so that it's properties are not spread)

danjm: oh, but I see that the other methods are not just passing what we call `txParams`, but whole objects that can be spread properly

danjm: In that case, I just suggest renaming `txParams` to `newTxMeta` so that there is not confusion with the `txMeta.txParams` property.

app/scripts/controllers/transactions/index.js
```diff
@@ -331,6 +331,195 @@ export default class TransactionController extends EventEmitter {
...
+  updateTransactionGasFees(
+    txId,
+    { gasLimit, gasPrice, maxPriorityFeePerGas, maxFeePerGas, estimateUsed },
+  ) {
+    if (!this._validateTransaction(txId)) {
+      return;
+    }
+
+    const txGasFees = {
+      txParams: {
```
danjm, line 385: In this case, we need to be careful not to overwrite the other properties of `txParams`. So we should probably add the existing txParams properties within `txGasFees`. Something like:
```
const { txParams: currentTxParams, estimateUsed: currentEstimateUsed } = this._getTransaction(txId);
    const txGasFees = {
        txParams: {
            ...currentTxParams
            gasLimit: gasLimit || currentTxParams.gasLimit,
            gasPrice: gasPrice || currentTxParams.gasPrice,
            maxPriorityFeePerGas: maxPriorityFeePerGas || currentTxParams.maxPriorityFeePerGas,
            maxFeePerGas: maxFeePerGas || currentTxParams.maxFeePerGas,
          },
          estimateUsed: estimateUsed || currentEstimateUsed,
    };
```

app/scripts/controllers/transactions/index.js
```diff
@@ -331,6 +331,195 @@ export default class TransactionController extends EventEmitter {
...
+   *}
+   * @param swapTransaction.sourceTokenSymbol
+   * @param swapTransaction.destinationTokenSymbol
+   * @param swapTransaction.type
+   * @param swapTransaction.destinationTokenDecimals
+   * @param swapTransaction.destinationTokenAddress
+   * @param swapTransaction.swapMetaData
+   * @param swapTransaction.swapTokenValue
+   */
+  updateSwapTransaction(
```
danjm, line 473: We should make a swap transaction in the app and then get the state logs to verify the shape of the state logs

app/scripts/controllers/transactions/index.js
```diff
@@ -331,6 +331,195 @@ export default class TransactionController extends EventEmitter {
...
+   */
+  updateTransactionEstimateBaseFee(
+    txId,
+    { estimatedBaseFee, decEstimatedBaseFee },
+  ) {
+    if (!this._validateTransaction(txId)) {
+      return;
+    }
+
+    const txEstimateBaseFees = {
```
danjm, line 418: I think with all these methods we might want to allow optional updates of the properties, so that the method can just be used to update one of the properties. In such a case, we should make sure not to overwrite a properties with undefined (as I did in the above example)

app/scripts/controllers/transactions/index.js
```diff
@@ -331,6 +331,195 @@ export default class TransactionController extends EventEmitter {
...
+   * updates a transaction's user settings only if the transaction state is unapproved
+   *
+   * @param {string} txId
+   * @param {object} userSettings - holds the metadata
+   * { userEditedGasLimit, userFeeLevel }
+   * @param userSettings.userEditedGasLimit
+   * @param userSettings.userFeeLevel
+   */
+  updateTransactionUserSettings(txId, { userEditedGasLimit, userFeeLevel }) {
+    if (!this._validateTransaction(txId)) {
```
danjm, line 512: We need to add another type of validation to all of these transactions. We do not want to allow any of these edits (in any of the methods added here) if the transaction status is not `unapproved`. In other words, `unapproved` in the only status a transaction can have if we are editing it with these methods. Otherwise we should return.

segun: Oh, yeah. I just forgot to rename the call to `_checkIfTxStatusIsUnapproved`

app/scripts/controllers/transactions/index.js
segun: The test cases caught this ;) 
I used lodash `merge` instead. One liner and cleaner.

```
    const updated = merge(txMeta, proposedUpdate);
    this.txStateManager.updateTransaction(updated, note);
```

app/scripts/controllers/transactions/index.js
segun: This is Swap Data

```
{
    "id": 5280220671892615,
    "time": 1644423917210,
    "status": "confirmed",
    "metamaskNetworkId": "4",
    "originalGasEstimate": "0x9810",
    "userEditedGasLimit": false,
    "chainId": "0x4",
    "loadingDefaults": false,
    "dappSuggestedGasFees": null,
    "txParams": {
        "from": "0xdec94d958b0da848d8cbfa0a9015451a4625452e",
        "to": "0x881d40237659c251811cec9c364ef91dc08d300c",
        "nonce": "0x0",
        "value": "0xde0b6b3a7640000",
        "gas": "0x9810",
        "maxFeePerGas": "0x178ee1aa8",
        "maxPriorityFeePerGas": "0x77359400"
    },
    "origin": "metamask",
    "type": "swap",
    "userFeeLevel": "medium",
    "defaultGasEstimates": {
        "estimateType": "medium",
        "gas": "0x9810",
        "maxFeePerGas": "0x178ee1aa8",
        "maxPriorityFeePerGas": "0x77359400"
    },
    "estimatedBaseFee": "c63f2c81",
    "sourceTokenSymbol": "ETH",
    "destinationTokenSymbol": "LUNA",
    "destinationTokenDecimals": 18,
    "destinationTokenAddress": "0xd2877702675e6ceb975b4a1dff9fb7baf4c91ea9",
    "swapMetaData": {
        "token_from": "ETH",
        "token_from_amount": "1",
        "token_to": "LUNA",
        "token_to_amount": "55.540352",
        "slippage": 3,
        "custom_slippage": true,
        "best_quote_source": "zeroEx",
        "other_quote_selected": false,
        "other_quote_selected_source": "",
        "gas_fees": "0.550904",
        "estimated_gas": "32440",
        "used_gas_price": "0",
        "is_hardware_wallet": false,
        "max_fee_per_gas": "178ee1aa8",
        "max_priority_fee_per_gas": "77359400",
        "base_and_priority_fee_per_gas": "13d74c081"
    },
    "swapTokenValue": "1",
    "preTxBalance": "29a2241af62c0000",
    "hash": "0x51e4270966c41b8ef39d227ae53902ab0c5574e2d57b383e1ad7bc32a30a5949",
    "submittedTime": 1644423919031,
    "firstRetryBlockNumber": "0x9ab9be",
    "baseFeePerGas": "0xb3e84c6a",
    "blockTimestamp": "6203eaf3",
    "postTxBalance": "1bc0d9579cda87d0"
}
```

segun: @danjm I see that the method takes a swapMetaData 
```
    "swapMetaData": {
        "token_from": "ETH",
        "token_from_amount": "1",
        "token_to": "LUNA",
        "token_to_amount": "55.540352",
        "slippage": 3,
        "custom_slippage": true,
        "best_quote_source": "zeroEx",
        "other_quote_selected": false,
        "other_quote_selected_source": "",
        "gas_fees": "0.550904",
        "estimated_gas": "32440",
        "used_gas_price": "0",
        "is_hardware_wallet": false,
        "max_fee_per_gas": "178ee1aa8",
        "max_priority_fee_per_gas": "77359400",
        "base_and_priority_fee_per_gas": "13d74c081"
    },
 ```
 
 Do we need to validate this too?

Comment (danjm) [reply app/scripts/controllers/transactions/index.js:480]: I don't think we need to validate that

File: app/scripts/controllers/transactions/index.js [hunk 10]
```diff
@@ -331,6 +332,208 @@ export default class TransactionController extends EventEmitter {
...
+    txId,
+    {
+      sourceTokenSymbol,
+      destinationTokenSymbol,
+      type,
+      destinationTokenDecimals,
+      destinationTokenAddress,
+      swapMetaData,
+      swapTokenValue,
+    },
```
danjm, line 490: I think we might also need a `approvalTxId` here

app/scripts/controllers/transactions/index.js
danjm: Thanks for sharing the transaction metadata. Based on that, it looks like the structure of the data here is correct

app/scripts/controllers/transactions/index.js
```diff
@@ -331,6 +332,216 @@ export default class TransactionController extends EventEmitter {
     });
   }
```
jpuri, line 334: Hey @segun , I think we should also capture transaction metrics in update function.

segun: @danjm see my latest push. I added some specific updates already. Let me know if I should leave them in or remove them.

jpuri: I agree with @danjm, we probably do not need event capturing at this stage.

jpuri: We probably should capture fields in some of events like gas fees changing. I see bit of issue in how metrics have been implemented in this PR.

We should not be adding `TRANSACTION_EVENTS` only capturing new fields. We can just call `_createTransactionEventFragment` to add new fields for transaction metrics.

cc @brad-decker can you plz suggest.

app/scripts/controllers/transactions/index.js
```diff
@@ -336,6 +337,278 @@ export default class TransactionController extends EventEmitter {
...
+    );
+  }
+
+  _updateTransaction(txId, proposedUpdate, note, eventType) {
+    const txMeta = this.txStateManager.getTransaction(txId);
+    const updated = merge(txMeta, proposedUpdate);
+    this.txStateManager.updateTransaction(updated, note);
+    this._trackTransactionMetricsEvent(txMeta, eventType);
+  }
+
```
jpuri, line 365: I am bit doubtful here, is it possible for these fields of transaction to be updated `data, from, to`

segun: Yes, there's a method `updateEIP1559Params` that handles this.

app/scripts/controllers/transactions/index.js
segun: As agreed, I'll leave out metrics from this PR>

jpuri (dismissed):

danjm (dismissed):

app/scripts/controllers/transactions/index.js
```diff
@@ -336,6 +337,257 @@ export default class TransactionController extends EventEmitter {
...
+    const txEIP1559 = {
+      txParams: {
+        data,
+        from,
+        to,
+        value,
+        gas,
+      },
+    };
```
brad-decker, line 388: None of these are explicitly EIP1559 params so I\m confused by this function name. it seems like this allows for updating of pretty much every txParam field *except* maxBaseFeePerGas and maxPriorityFeePerGas?

brad-decker (changes requested):

app/scripts/controllers/transactions/index.js
segun: The function names is because of where the function will be used. So this will be used in 
`ui/ducks/send/send.js:1703` 

```
const eip1559OnlyTxParamsToUpdate = {
        data: txParams.data,
        from: txParams.from,
        to: txParams.to,
        value: txParams.value,
        gas: unapprovedTx.userEditedGasLimit
          ? unapprovedTx.txParams.gas
          : txParams.gas,
      };
      unapprovedTx.originalGasEstimate = eip1559OnlyTxParamsToUpdate.gas;
      const editingTx = {
        ...unapprovedTx,
        txParams: Object.assign(
          unapprovedTx.txParams,
          eip1559support ? eip1559OnlyTxParamsToUpdate : txParams,
        ),
      };
      dispatch(updateTransaction(editingTx));
```

So `dispatch(updateTransaction(editingTx));` will become `dispatch(updateEIP1559Params(editingTx));`

danjm: `eip1559OnlyTxParamsToUpdate` in the context of https://github.com/MetaMask/metamask-extension/commit/668fd2db660ef43295de74d85034d65c0b14080d should probably be `updateTheseTxParamsIfWeAreOnAnEIP1559Network
`

danjm: Oh, but by reviewing that commit, I am realizing that this method needs to change a bit...

danjm: The place where this is meant to be used is https://github.com/MetaMask/metamask-extension/commit/668fd2db660ef43295de74d85034d65c0b14080d#diff-bad76d7279418f3520a0ca93cfaeeb6fcbb1f0d4782f6ff942d422632a1c2780R1521-R1534

For the case where `eip1559support` is false, then that call to `updateTransaction` needs to update more params than this method supports

We need to allow this to update `gasPrice` as well.

Then perhaps we could name it `updateEditableParams`

brad-decker: I like updateEditableParams.

segun: @danjm any easy way to know the additional params needed when `eip1559support` is false? I see the update function just takes all of the txParams. If all of the txParams should be updated, can we make 2 calls, 

updateEditableParams if `eip1559support` is true
updateTransactionGasFees if false.

danjm: when it is false, all the same params as what you've currently identified can be updated + only one more: `gasPrice`

app/scripts/controllers/transactions/index.js
```diff
@@ -336,6 +337,268 @@ export default class TransactionController extends EventEmitter {
...
+        to,
+        value,
+        gas,
+        gasPrice,
+      },
+    };
+
+    // only update what is defined
+    editableParams.txParams = pickBy(editableParams.txParams);
+    const note = `Update Editable Params Params for ${txId}`;
```
brad-decker, line 394: Is the repetition of the word 'Params' necessary?

brad-decker (changes requested):
Just a minor question about log notes

app/scripts/controllers/transactions/index.js
segun: Not really. Fixed.

danjm (changes requested):
Only improvement left is some increased test coverage. Then I think this is good to merge.

jpuri (approved):

danjm (approved):
LGTM!

brad-decker (approved):

PR #13511, opened 2022-02-04
Handle a negative gas fee

dan437:


darkwing:
How do we get a negative gas value in the first place?

Gtonizuka:
@darkwing I had the same question, as per @danjm "we are subtracting one value from another to compute an estimate, so some incorrect numbers from the api are causing a negative"

darkwing (approved):
I'm good with this as long as we investigate the API side!

adonesky1:
> I'm good with this as long as we investigate the API side!

Sam is on it: https://consensys.slack.com/archives/G0150V3D4HX/p1643991649380189?thread_ts=1643978152.050779&cid=G0150V3D4HX

adonesky1 (approved):

PR #13512, opened 2022-02-04
Alter broken identicon images when the image errors

darkwing:
This is sort of a conversation starter.  Right now, in Firefox, a broken identicon image shows the alt text, which looks kinda gross:

boooo

Instead we could do something like:
Broken

Happy for design and others to weigh in!

### DIFF
--- a/ui/components/app/asset-list-item/asset-list-item.js
+++ b/ui/components/app/asset-list-item/asset-list-item.js
@@ -131,6 +131,7 @@ const AssetListItem = ({
           image={tokenImage}
           alt={`${primary} ${tokenSymbol}`}
           imageBorder={identiconBorder}
+          tokenSymbol={tokenSymbol}
         />
       }
       midContent={midContent}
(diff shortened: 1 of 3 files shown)


GuillaumeRx:
Maybe we could have a default image to replace the broken identicon ? Like a little token symbol if this only appends on the token identicons

Montoya:
Follow on request is if the broken token image could be per-blockchain, so Ethereum tokens will have a certain image and Polygon tokens will have a certain image and so on.

brad-decker:
@georgewrmarshall @garrettbear @brianacnguyen -- i think y'all are talking about icons, avatars, etc. You may want to take to design team what to do in these cases and align on the fallback.

georgewrmarshall:
We have a component for this! https://metamask.github.io/metamask-storybook/?path=/story/ui-components-component-library-avatar-token-avatar-token-stories-js--default-story&args=tokenImageUrl: but we need to implement the `onError` handler so it shows the first letter fallback instead of the alt tag text

PR #13521, opened 2022-02-07
captureKeyringTypesWithMissingIdentities() when 'Missing identity for address' in permissions/specifications

danjm:
This is in response to https://sentry.io/organizations/metamask/issues/2979064019

Which throws the error `Missing identity for address:` from here https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/controllers/permissions/specifications.js#L122

It seems that somehow it is possible for their to be accounts in the keyring that are not in the `identities` object. After a few hours of investigation, I have not been able to determine the cause of this.

I am adding this PR so that we can gather more information about this error case. The information captured as a result of this PR may tell us some important things:
- whether or not the accounts that cannot be found in the identities object are hardware wallet accounts, HD wallet accounts or imported accounts
- whether the accounts are also missing from the `AccountTracker` state
- an indication as to whether the account isn't missing, but perhaps its address is formatted differently

This should help us narrow the search for the root cause.

Note that this PR is intentionally not capturing any addresses with the capture error.

### DIFF
--- a/app/scripts/controllers/permissions/specifications.js
+++ b/app/scripts/controllers/permissions/specifications.js
@@ -78,11 +78,15 @@ export const getCaveatSpecifications = ({ getIdentities }) => {
  * in the current MetaMask instance.
  * @param options.getIdentities - A function that returns the
  * `PreferencesController` identity objects for all Ethereum accounts in the
+ * @param options.captureKeyringTypesWithMissingIdentities - A function that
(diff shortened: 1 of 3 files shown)


Gtonizuka (dismissed):

jpuri (approved):
LGTM

Gtonizuka (approved):

PR #13523, opened 2022-02-07
RPC URL input whitespace trim

Gtonizuka:
Fixes: N/A

Explanation:  

When I copy/pase a RPC url, or manually add one, into the "add network" view the field validation gives me an error if the URL contains a white space.

As far as I know it shouldn't ever be possible to have a space in there so why don't just trim the value all together?

Manual testing steps:  
 Try to add white spaces into the RPC URL field in the add network view

Before:

Screenshot 2022-02-07 at 16 53 46

After:

Screenshot 2022-02-07 at 16 44 19


ui/pages/settings/networks-tab/networks-form/networks-form.test.js
```diff
@@ -163,11 +163,6 @@ describe('NetworkForm Component', () => {
       ),
     ).toBeInTheDocument();
 
-    await fireEvent.change(rpcUrlField, {
```
Gtonizuka, line 166: This assertion is not needed anymore as this case is now not reproducible

ui/pages/settings/networks-tab/networks-form/networks-form.js
```diff
@@ -571,7 +571,7 @@ const NetworksForm = ({
         />
         <FormField
           error={errors.rpcUrl?.msg || ''}
-          onChange={setRpcUrl}
+          onChange={(e) => setRpcUrl(e.replace(/\s+/gu, ''))}
```
GuillaumeRx, line 574: Arrow function definition should be avoided in props. Could you define it outside ?

GuillaumeRx (changes requested):

Gtonizuka:
@brad-decker can I have your eyes on this? 

I want to make sure this won't affect any specific edge case (can't think of any atm)

danjm (approved):
LGTM

ui/pages/settings/networks-tab/networks-form/networks-form.js
brad-decker: @GuillaumeRx defining it outside or in props has the same effect, from a re-rendering point of view. Could you explain more about what you are attempting to avoid by not having the arrow function in props?

brad-decker:
I think this is fine as is @Gtonizuka -- we may want to only legtimately trim (spaces from beginning/end) instead of replacing spaces in the middle. Spaces in the middle of an URL are obviously invalid but might mean typos, missing characters, and/or shady copy and paste instructions from websites.

darkwing:
Hello @Gtonizuka !  Could you get this one across the finish line?

Gtonizuka:
This bug was fixed in another PR if I recall correctly - closing for now

PR #13402, opened 2022-01-25
Use ActionableMessage for ConfirmAddSuggestionToken warnings

digiwand:
**Fixes:** #13347

**Explanation:**  Deprecate the last instances of the global `.warning` class used in the ConfirmAddSuggestionToken component. We replaced these instances using our ActionableMessage component. 

Additionally: 
- We also brought over the "Learn more about scams and security risks." link from the "Import Tokens" page, here to the "Add Suggested Tokens" page.
- Updated the symbol detection logic by treating symbols as case-insensitive
- Updated storybook to include controls and different warning states

**Manual testing steps (For EtherScan. Can replicate on another platform, e.g., CoinGecko):**  
  1. View an ERC20 token on EtherScan ([example](https://etherscan.io/token/0xdac17f958d2ee523a2206206994597c13d831ec7))
  2. Click on the "More" dropdown in "Profile Summary"
  3. Click on "Add Token to MetaMask (Web3)"
  4. Add token using MetaMask
  5. Repeat step 3
  6. Observe the "already listed" warning message. This should appear when the user is attempting to re-add a token with a token address already in their wallet.

To reproduce the reused symbol warning message, add a token with the same token name, e.g., "USDT," and a different address.

**Before:**
add-suggested-tokens-old

**Screenshot showing both warning messages:** 
Screenshot 2022-01-25 at 6 40 57 PM

**Updated Storybook:**
Screenshot 2022-01-29 at 12 34 59 AM

PR #13377, opened 2022-01-21
Convert ConfirmAddSuggestedToken to a functional component + cleanup

digiwand:
**Fixes:** n/a
This is cleanup before fixing https://github.com/MetaMask/metamask-extension/issues/13347

Follow-up PR: https://github.com/MetaMask/metamask-extension/pull/13402

**Explanation:**  
Convert ConfirmAddSuggestedToken component to functional component. I did a slight cleanup, but I didn't want to do too much to mitigate risk.

**For reviewers:**
I think it might be easier to walk through each commit and hide whitespace changes in the first commit. Please feel free to suggest more updates! 

The `checkTokenDuplicates` and `checkNameReuse` logic will be updated in a a follow-up PR to fix: https://github.com/MetaMask/metamask-extension/issues/13347

**Manual testing steps (For EtherScan. Can replicate on another platform e.g. CoinGecko):**  
  - View an ERC20 token on EtherScan ([example](https://etherscan.io/token/0xdac17f958d2ee523a2206206994597c13d831ec7))
  - Click on the "More" dropdown in "Profile Summary"
  - Click on "Add Token to MetaMask (Web3)"
  - Test "Add Suggested Token" modal works as expected

Issue #11377, opened 2021-06-24
Function wallet_watchAsset not working as expected in Firefox.


jgarciaIOB:
Hi,

**Describe the bug**
With Firefox, when you try to add a token to a network with `wallet_watchAsset`, it does not wait for the promise to be solved. Looks like it is returning if the token already exists in the network or not but without waiting for the promise to be solved by the user ('Cancel' or 'Add Token').
It works fine in Chrome, Brave and Edge.

**Steps to reproduce**
1. Open Firefox
2. Go to a page with ethereum connected
3. Paste this code in the console (https://docs.metamask.io/guide/rpc-api.html#example-5):
```
ethereum.request({
  method: 'wallet_watchAsset',
  params: {
    type: 'ERC20',
    options: {
      address: '0xb60e8dd61c5d32be8058bb8eb970870f07233155',
      symbol: 'FOO',
      decimals: 18,
      image: 'https://foo.io/token-image.svg',
    },
  },
})
  .then((success) => {
    if (success) {
      console.log('FOO successfully added to wallet!')
    } else {
      throw new Error('Something went wrong.')
    }
  })
  .catch(console.error)
```
4. As you can see, it solves the promise before you click on accept or cancel the add token proccess.

**Expected behavior**
It should wait the user interaction to solve the promise (Add the token or cancel the proccess). It works fine in the other browsers.

**Screenshots**
When you try to add the token the first time:
![imagen](https://user-images.githubusercontent.com/85491102/123284007-3296ce80-d50c-11eb-8d83-e17fd5de65b7.png)

When the token is already added:
![imagen](https://user-images.githubusercontent.com/85491102/123284072-3f1b2700-d50c-11eb-8f39-394cd2be3fa9.png)

**Browser details:**
 - OS: Ubuntu 20.04.2 LTS / Ubuntu 18.04.5 LTS
 - Browser: Firefox 89.0.1 64bits
 - MetaMask Version: 9.6.1

hbarcelos:
I'm facing this issue too, but contrary to the author, I'm experiencing the promise being resolved even if the user clicks on "Cancel" in every browser:

- OS: Manjaro Linux (5.11.22-2-MANJARO)
- Firefox: 90.0b4 / MetaMask: 9.7.1
- Chrome: 92.0.4515.80 beta / MetaMask: 9.7.1

kevinghim:
Possibly related to #11383

danjm:
[...]

PR #13526, opened 2022-02-07
ConfirmAddSuggestToken: replace mapDispatchToProps & mapStateToProps; Add Tests

digiwand:
**Fixes:** #

No issue

**Explanation:**  

Update ConfirmAddSuggestToken. No logic changes. We recently converted this component to a functional component. This work is a follow-up work for https://github.com/MetaMask/metamask-extension/pull/13377 as mentioned here https://github.com/MetaMask/metamask-extension/pull/13402#issuecomment-1026430441

- replace mapDispatchToProps with useDispatch
- replace mapStateToProps with useSelector
-  [In Progress] write tests

**Manual testing steps (For EtherScan. Can replicate on another platform, e.g., CoinGecko):**

- View an ERC20 token on EtherScan ([example](https://etherscan.io/token/0xdac17f958d2ee523a2206206994597c13d831ec7))
- Click on the "More" dropdown in "Profile Summary"
- Click on "Add Token to MetaMask (Web3)"
- Add token using MetaMask
- Repeat step 3
- Observe the "already listed" warning message. This should appear when the user is attempting to re-add a token with a token address already in their wallet.
- Test "cancel" button closes MetaMask


ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.js
```diff
@@ -51,18 +54,19 @@ function hasDuplicateSymbolAndDiffAddress(suggestedAssets, tokens) {
...
   const metricsEvent = useContext(MetaMetricsContext);
   const t = useContext(I18nContext);
+  const dispatch = useDispatch();
+  const history = useHistory();
+
+  const mostRecentOverviewPage = useSelector((state) =>
+    getMostRecentOverviewPage(state),
+  );
+  const suggestedAssets = useSelector(
+    (state) => state.metamask.suggestedAssets,
```
adonesky1, line 67: you could make a new selector for this:
in `ui/selectors/selectors.js`
```
export function getSuggestedAssets(state) {
  return state.metamask.suggestedAssets;
}

```

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.js
```diff
@@ -51,18 +54,19 @@ function hasDuplicateSymbolAndDiffAddress(suggestedAssets, tokens) {
...
-
+const ConfirmAddSuggestedToken = () => {
   const metricsEvent = useContext(MetaMetricsContext);
   const t = useContext(I18nContext);
+  const dispatch = useDispatch();
+  const history = useHistory();
+
+  const mostRecentOverviewPage = useSelector((state) =>
+    getMostRecentOverviewPage(state),
+  );
```
adonesky1, line 65: This can be simplified to this:
```suggestion
  const mostRecentOverviewPage = useSelector(getMostRecentOverviewPage);
  );
```

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.js
```diff
@@ -51,18 +54,19 @@ function hasDuplicateSymbolAndDiffAddress(suggestedAssets, tokens) {
...
+  const suggestedAssets = useSelector(
+    (state) => state.metamask.suggestedAssets,
+  );
+  const tokens = useSelector((state) => state.metamask.tokens);
```
adonesky1, line 69: A selector for this already exists in 
`ui/ducks/metamask/metamask.js`

```suggestion
  const tokens = useSelector(getTokens);
```

digiwand:
@adonesky1 

Thank you for the review! This makes sense. I applied the suggestions to the code with the import statements. 

re: https://github.com/MetaMask/metamask-extension/pull/13402#issuecomment-1026430441 
I thought more about the single selector in the newest `confirm-add-suggested-token/index.scss` file. We should not move the style into `confirm-import-token/index.scss` since the style does not exist for the `ConfirmImportToken`. Then I considered moving the `ConfirmImportToken` styles into this `.scss` file. Even though I'm usually okay duplicating or even triplicating code before making it reusable code, I'm afraid that copying the styles will require more maintenance and can more easily lead us to create style disparities. With these thoughts, what do you think about keeping the styles fragmented as they exist now? Or, is there another idea you had in mind to consolidate the styles?

adonesky1:
> Then I considered moving the ConfirmImportToken styles into this .scss file. Even though I'm usually okay duplicating or even triplicating code before making it reusable code, I'm afraid that copying the styles will require more maintenance and can more easily lead us to create style disparities

I think my preference for the short term would be to duplicate these styles in `confirm-add-suggested-token/index.scss`. Since these are considered different components for now they should have there own styles. We're duplicating the `react` so it feels messy to me to not duplicate the styles. If we determine that these components are almost identical minus a few conditional changes, we ought to consolidate them into one react component with some conditional rendering to vary across use cases. This is just my opinion 😄

digiwand:
@adonesky1  Thank you for explaining your opinion here because I was close to being indifferent about duplicating the styles and reusing the styles. I went with duplicating the styles! I also went with not consolidating the two components and styles with a new react component. I can see a benefit in creating a page abstraction if we reuse this page UI in the future. 

What I think could be helpful right now, and in a different PR, would be to consolidate the token or asset lists. Thing is, there seem to be a few variations of lists and list items in our repo. We currently have an AssetList and TokenList Component/UI that is not used on these pages. I can keep this in mind for the future as I learn more about this codebase.

digiwand:
It might be worth mentioning here that by reading the code, it would appear we could expect the app to redirect to the overview page, `history.push(mostRecentOverviewPage);`, after clicking "Add Tokens". Instead, through CoinGecko or EtherScan, the popup closes after clicking "Add Tokens". I suspect this may be expected behavior with the way CoinGecko and EtherScan use the MM API. 

Again, no UI changes have been introduced in this PR. These GIFs below show that the previous and newer "Add Tokens" click remains the same.

previous MM UI w/ "Add Tokens" click:


newer MM UI w/ "Add Tokens" click (same click behavior):

danjm:
> It might be worth mentioning here that by reading the code, it would appear we could expect the app to redirect to the overview page, history.push(mostRecentOverviewPage);, after clicking "Add Tokens".

This is what happens if you have the add tokens confirmation open in the full screen tab or in the extension "popup" view (by clicking the fox icon in the top right of your browser). When interacting via the notification window, it should close upon confirmation as your screen capture displays

digiwand:
@danjm Thanks for clarifying this! I'd also like to clarify that the MM is opened in a popup window and not the popup extension in the GIFs above. What Dan mentions above is in line with expected behaviors.

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.js
```diff
@@ -117,11 +118,18 @@ const ConfirmAddSuggestedToken = (props) => {
     );
   }, [suggestedAssets, tokens, t]);
 
+  useEffect(() => {
+    if (Object.keys(addedToken).length) {
```
danjm, line 122: This sustains the behaviour of tracking the metrics event when the submit button is clicked, but it does so indirectly. Previously, the metrics event would be tracked as part of the onclick method call, but not it is tracked as a result of a state change that is created by the onclick method call.

This could cause confusion for a reader of the code, who might expect `useEffect` to be responding to a state change that is outside of the control of this component. Also, it increases the number of places in which a bug could be introduced in the future. Previously, the path was `onclick call -> metric event method call`, now the path is `onclick call -> state update -> useEffect recognition of state update -> metric event method call`. So there are more potential steps that could be accidentally broken in the future. Finally, as a general rule, if useEffect can be avoided for something simpler, that generally make components more readable and maintainable.

And in this case, there is some risk that this might not always work as desired when there are multiple `suggestedAssets` 

I would suggest that we continue just to call the metrics event method within the onclick handler. To do this, you could put the whole onclick handler function in a function defined in the body of the component function and then pass that to `onClick`.

I think this would require using the existing `tokenAddedEvent` method

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.js
```diff
@@ -117,11 +118,18 @@ const ConfirmAddSuggestedToken = (props) => {
...
+      trackTokenAddedEvent();
+    }
+  }, [addedToken, trackTokenAddedEvent]);
+
   useEffect(() => {
     if (!suggestedAssets.length) {
       history.push(mostRecentOverviewPage);
     }
-  }, [history, suggestedAssets, mostRecentOverviewPage]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
danjm, line 131: Why was `// eslint-disable-next-line react-hooks/exhaustive-deps` needed?

danjm (changes requested):
Left one suggestion and one question

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.js
digiwand: Passing an empty array will ensure this hook is only called once. This is the closest I think we could come to replicating the `componentDidMount` which is what was used when this was a class component. 

see "Note:" at the bottom of ref: https://reactjs.org/docs/hooks-effect.html

If we were to pass `suggestedAssets` as a dependent property here, `history.push(mostRecentOverviewPage);` could override other redirects that happen after altering the `suggestedAssets`. For instance, in the `ConfirmImportToken` page, we redirect the user to the token page after the token has been added. If we were to replicate that behavior here, using `useEffect` with `[suggestedAssets]` dependency list would cause a race condition.

It is recommended that we keep the ESLint warning for `react-hooks/exhaustive-deps` and not use this rule as a last resort.

ref: https://reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often
> Only do this if you couldn’t find a better alternative, as relying on mutation makes components less predictable.

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.js
```diff
@@ -117,11 +106,40 @@ const ConfirmAddSuggestedToken = (props) => {
     );
   }, [suggestedAssets, tokens, t]);
 
+  const handleAddTokensClick = useCallback(async () => {
+    await Promise.all(
+      suggestedAssets.map(async ({ asset, id }) => {
+        await dispatch(acceptWatchAsset(id));
+
+        metricsEvent({
```
brad-decker, line 114: [REQ] change this to `trackEvent` which is the name of the context function for the new context, but also the useNewMetricsEvent hook calls it metricsEvent which is wrong so i get why there is like 5 layers of misdirection. 

Maybe i was trying to encode job security here? 🤷🏻

brad-decker (dismissed):

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.js
digiwand: updated, thank you! @brad-decker

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.js
digiwand: @danjm 

Thank you for calling this out! As mentioned in this past Wednesday's standup, I have updated the logic to call `trackEvent` instead of `useNewMetricEvent` directly. 

I agree using `useEffect` can cause unwanted side-effects as `suggestedAssets` is susceptible to change. 

I did go with calling the click handler from the `onClick` function. I think the biggest hurdle was finding a way to use `useNewMetricEvent`, and the key was not to use this when we have dynamic params.  Since `useNewMetricEvent` returns a custom hook, we cannot call it inside another callback which would cause some limitations.

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.js
danjm: Okay, I think you can do something like this:

```
const goBackIfNoSuggestedAssetsOnFirstRender = () => {
    if (!suggestedAssets.length) {
      history.push(mostRecentOverviewPage);
    }
}

useEffect(() => {
    goBackIfNoSuggestedAssetsOnFirstRender();
}, []);
```

danjm (dismissed):
Looking good! One needed change left that I have commented on

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.js
digiwand: lgtm! Thanks for the function name suggestion. Updated the repo with this suggestion

digiwand: @danjm 

lgtm! Thanks for the function name suggestion. Updated the repo with this suggestion

danjm: I think `// eslint-disable-next-line react-hooks/exhaustive-deps` can be removed now?

digiwand: @danjm @adonesky1 

Removing `// eslint-disable-next-line react-hooks/exhaustive-deps` will enable the warning message again on `[]`.  Adding `goBackIfNoSuggestedAssetsOnFirstRender` as a dependency would cause useEffect to rerun again on each render unless we wrapped it in useCallback, or declare the function inside the useEffect hook. With either of these options, we would run into the need for the empty dependency list, `[]`, either on the useCallback hook or the useEffect hook. 

Is there another way we could go about removing `// eslint-disable-next-line react-hooks/exhaustive-deps` and only rendering this check once on initial load? 

ESLint warning after replacing `[]` with `[goBackIfNoSuggestedAssetsOnFirstRender]` in useEffect:
Screenshot 2022-03-03 at 1 20 56 PM

adonesky1: Does it not work to wrap `goBackIfNoSuggestedAssetsOnFirstRender` in `useCallback` add all dependencies to the dependency array and then add `goBackIfNoSuggestedAssetsOnFirstRender` to the dependency array of the `useEffect`?

digiwand: I don't think so. The initial logic lives in `componentDidMount`, so the thing is, we only want to run this logic on initial load or initial render in this case. I think with this suggestion, if the list of suggestions changes, for instance on `acceptWatchAsset`, then this `useEffect` will run because one of the dependencies of `useCallback` has changed. This causes a race condition between `history.push` here, and the one in the `Add Tokens` click handler.

digiwand: @adonesky1 

I don't think so. The initial logic lives in componentDidMount, so the thing is, we only want to run this logic on initial load or initial render in this case. I think with this suggestion, if the list of suggestions changes, for instance on acceptWatchAsset, then this useEffect will run because one of the dependencies of useCallback has changed. This causes a race condition between history.push here, and the one in the Add Tokens click handler.

jpuri: I can think of an edge case here, what is `suggestedAssets` are not loaded on first render ?

digiwand: @jpuri 

This is a good question. 

What I see happening is when the `wallet_watchAsset` [API request](https://docs.metamask.io/guide/registering-your-token.html) is requested, the `watchAssetHandler` RPC middleware calls [TokensController#watchAsset](https://github.com/MetaMask/controllers/blob/main/src/assets/TokensController.ts#L393). This method validates the asset, updates the redux `suggestedAssets` state with the new asset, and then emits `pendingSuggestedAsset`. `pendingSuggestedAsset` will then call [background.js#openPopup](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/background.js#L605). Unless I am missing something, I think we can safely assume that it will be there before the popup is opened. 

I also think it is safe to make this change due to the order that the hooks run.  `componentDidMount` runs after mounting and before painting. `useEffect` runs after the first paint. We had replaced `componentDidMount` in https://github.com/MetaMask/metamask-extension/pull/13377/files, so the logic is near the same here.

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.test.js
```diff
@@ -0,0 +1,142 @@
+import React from 'react';
```
jpuri, line 1: Good work on test coverage 👍

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.js
```diff
@@ -1,16 +1,21 @@
...
+import { getTokens } from '../../ducks/metamask/metamask';
 import ZENDESK_URLS from '../../helpers/constants/zendesk-url';
 import { isEqualCaseInsensitive } from '../../helpers/utils/util';
+import { getSuggestedAssets } from '../../selectors';
+import { rejectWatchAsset, acceptWatchAsset } from '../../store/actions';
 
 function getTokenName(name, symbol) {
-  return typeof name === 'undefined' ? symbol : `${name} (${symbol})`;
+  return name === undefined ? symbol : `${name} (${symbol})`;
 }
```
jpuri, line 19: 👍

jpuri (dismissed):

adonesky1 (dismissed):
Looks like you need a rebase 😢  but otherwise LGTM!

adonesky1 (dismissed):

digiwand:
@jpuri @adonesky1 Thank you! I rebased and verified this feature works.

ui/pages/confirm-add-suggested-token/index.scss
```diff
@@ -1,9 +1,77 @@
...
+    .token-balance {
+      display: flex;
+      flex-flow: row nowrap;
+      align-items: flex-start;
+
+      &__amount {
+        @include H1;
+
+        color: var(--scorpion);
+        margin-right: 8px;
```
darkwing, line 41: `margin-inline-end` would be good for RTL.

ui/pages/confirm-add-suggested-token/index.scss
```diff
@@ -1,9 +1,77 @@
...
+
+  &__name {
+    min-width: 0;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+
+  &__token-icon {
+    margin-right: 12px;
```
darkwing, line 74: `margin-inline-end` would be good for RTL.

darkwing (changes requested):
This looks awesome!  Firefox was spotty but I noticed it's spotty on `develop` branch.  The popup usually appears on the first click, but doesn't appear to ever do so again.  I don't think it's related to your changes.

digiwand:
@darkwing 

Thanks for mentioning this! I am confirming your observation and that this does not appear to be related to any of these changes.

I also observed this page is spotty in Firefox for this branch and `develop` while it works fine in Brave, Chrome, and Opera. Then, I checked out the last commit before https://github.com/MetaMask/metamask-extension/pull/13377 was merged, d25f0e5817ecdc8e37bfb144e4b7ebf299233421. This was the commit before I made any refactoring changes. I verified this page also works in Opera, Chrome, and Brave. Firefox behaves the same for this commit as it does on this branch and on `develop`. 

The behavior that we are observing in Firefox is that after adding the token once, the popup will not open again. On CoinGecko, after adding the token once via MM, each time the user clicks "Add to MetaMask" again, the "Add Token" notification toast will appear, and the popup will not work again.

ui/pages/confirm-add-suggested-token/index.scss
digiwand: It turns out that the `.token-balance` class is not summoned here. `TokenBalance` component exists in this component. I removed these styles. We should be removing `.token-balance` from `ui/pages/confirm-import-token/index.scss` in a separate PR.

ui/pages/confirm-add-suggested-token/index.scss
digiwand: @darkwing  
`margin-inline-end` is new to me. I see the benefit when the direction changes. How does it help here?

adonesky1 (approved):

darkwing (approved):
This is awesome!  Well done!

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.js
digiwand: per @mcmire's suggestion here https://github.com/MetaMask/metamask-extension/pull/13594#discussion_r805027489 🙂 ✨

ui/pages/confirm-add-suggested-token/index.scss
digiwand: Update: spoke with @darkwing  about this and we don't think `margin-inline-end` is necessary here. We are creating a follow-up task to update the spacing around this element after some margin/padding concerns we noticed.

digiwand:
Thanks for the patience and thoughtful reviews everyone!

digiwand:
@darkwing  

Found something. Looks like this issue, opened in June 2021, may be related to the finicky behavior we were observing in Firefox here: 
https://github.com/MetaMask/metamask-extension/issues/11377

ui/pages/confirm-add-suggested-token/index.scss
digiwand: Update: spoke with @darkwing about this and we don't think margin-inline-end is necessary here

PR #13131, opened 2021-12-20
Refactor token send/method confirmation flow

adonesky1:
Fixes: #11611 and https://github.com/MetaMask/metamask-extension/security/advisories/GHSA-jwfv-7854-fmc3

**Overview**
- Refactors base confirmation routing component - `confirm-transaction.js` - from class to functional (includes providing all required redux state/methods via selectors/dispatch, and changing lifecycle methods to `useEffect`) while preserving all functionality.
- Creates a sub-component `ConfirmTokenTransactionSwitch` through which all token/nft specific confirmation screens are routed.
- `ConfirmTokenTransactionSwitch` instantiates a newly created hook `useAssetDetails` which passes the pending transaction to the new method [getTokenStandardAndDetails](https://github.com/MetaMask/controllers/blob/81bcbae1c4abe71a9b07819606e155a33ec4a6b4/src/assets/AssetsContractController.ts#L151) in the `AssetsContractController` and retrieves all the data ( `assetName`, `userBalance`, `tokenSymbol`, `decimals`, `image` etc.) we need for the three **token** specific method pages we currently support (`approve`, `transfer`, and `transferFrom`) but also includes `assetStandard` data to enable improved support and branching logic depending on the contract standard (ERC20, ERC721 or ERC1155). 
    - **Importantly** this eliminates the [unwanted behavior](https://github.com/MetaMask/metamask-extension/security/advisories/GHSA-jwfv-7854-fmc3) of adding the token to the `TokensController` state whenever a user interacts with a confirmation transaction of a token not already in `TokensController` state

In this PR behavior/UX should be unaltered for the ERC20 method calls listed above but be more accurate and clear for the same method calls for ERC721 and ERC1155 contracts.

**For Example**

**ERC721 `approve` call before:**
Screen Shot 2022-02-09 at 5 31 42 PM
the tokenId is incorrectly being used as token amount ☝️ 👇  
Screen Shot 2022-02-09 at 5 26 07 PM

**after:**
the option to change the amount should not be present for ERC721 tokens 👇 
Screen Shot 2022-01-14 at 2 36 02 PM

**ERC721 `transferFrom` call before:**
Screen Shot 2022-02-09 at 5 34 32 PM

**after:**
Screen Shot 2022-02-09 at 5 53 34 PM

PR #13535, opened 2022-02-07
Fix issue where we show contract address as recipient when calling safe transfer method on erc721 or erc1155 contracts

### LINKED ISSUE #13855: Add `safeTransferFrom` as a supported NFT method screen, reusing the `transferFrom` view
`safeTransferFrom` is a version of the `transferFrom` method that was added to the ERC721 contract standard (improving on the `transferFrom` method introduced in the ERC20 contract standard) that checks for a valid recipient.

adonesky1:
Fixes: https://github.com/MetaMask/metamask-extension/issues/13855

This fixes a [UX issue that OpenSea raised](https://consensys.slack.com/archives/CN2H44RAM/p1639676146047200) with us a couple months back. IIRC it fell of our radar because after reviewing the behavior it appeared to work. This was a red herring because we derive recipient addresses differently in sends that are initiated within Metamask and those initiated via API. There is indeed still an issue with NFT sends initiated via API:
Current Behavior for in-app NFT send (ignore the image not rendering, separate issue): 

https://user-images.githubusercontent.com/34557516/152865105-41b4bde2-2999-4053-902b-b5e8b74e28fa.mov

Current Behavior for NFT send initialized from OpenSea:

https://user-images.githubusercontent.com/34557516/152865181-40c4f85b-9e89-46e2-a1c4-dd25d9a5be6c.mov

With this fix:

https://user-images.githubusercontent.com/34557516/152865686-1b4b5221-764f-496e-98d8-b67ff0108373.mov

- Moves `getTransactionData` method to a shared module and make use of it in the method `determineTransactionType` already in that shared module.

Manual testing steps:  
  - Attempt to transfer an NFT via OpenSea (or Etherscan)
  - Make sure that the "recipient address" in the upper right hand corner matches the recipient you entered, rather than the contract address.
  
Next Steps:
- I opened this ticket to migrate our ABI usage towards the pattern that is used here (importing and making use of all of our known/standard ABIs - ERC20, ERC721, ERC1155). Alternatively we may want to expose methods on AssetsContractController and only instantiate the ABI interfaces there.


Gudahtt:
Where is the NFT contract shown on this confirmation screen? I can't tell from the recording. It makes sense for the 'recipient' to be the actual recipient and not the contract, but we should still be communicating to the user what they're confirming.

adonesky1:
> Where is the NFT contract shown on this confirmation screen?
>  we should still be communicating to the user what they're confirming.

Agreed. We may need some designs for this but currently this is what we show for contract methods that aren't included in the specific subset of "known" methods we have specifically designed for. Like the approval screen.
Screen Shot 2022-02-08 at 8 35 02 AM

🤔  I suppose in this case it would probably makes sense for the "safe transfer from" method to have a similar or the same design as our existing "transfer from" confirmation page (this is what it looks like for an ERC721 `transferFrom` call at the moment):
Screen Shot 2022-02-08 at 8 38 11 AM

adonesky1:
@Gudahtt Actually that image is taken from a refactor I'm [actively working on](https://github.com/MetaMask/metamask-extension/pull/13131) to make our method call confirmations clearer for ERC721 and ERC1155 contracts. Currently our "transferFrom" method call for these contracts is unfortunately even more confusing:

https://user-images.githubusercontent.com/34557516/153011437-9244fd04-f329-49a1-b68c-3602c331f501.mov

As you can see here we _appear_ to represent the tokenId being transferred as the "amount". Unfortunately until my refactor (or similar work) lands I think it might be more confusing to use the "transferFrom" (designed for ERC20 tokens) for ERC721 and ERC1155 contracts

ui/helpers/utils/transactions.util.js
```diff
@@ -35,11 +37,24 @@ const hstInterface = new ethers.utils.Interface(abi);
...
+    // ignore and next try to parse with erc1155 ABI
   }
+
+  try {
+    return erc1155Interface.parseTransaction({ data });
+  } catch {
+    // ignore and return undefined
+  }
+
+  return undefined;
```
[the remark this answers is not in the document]
adonesky1: The data is unfortunately a long hex string that is only interpretable with this helper method and the correct ABI (as far as I know). We do a similar "brute" check in the controllers [here](https://github.com/MetaMask/controllers/blob/81bcbae1c4abe71a9b07819606e155a33ec4a6b4/src/assets/AssetsContractController.ts#L151) and [here](https://github.com/MetaMask/controllers/blob/81bcbae1c4abe71a9b07819606e155a33ec4a6b4/src/assets/CollectiblesController.ts#L899) for example

adonesky1:
While still not ideal, on mainnet alteast the `data` tab will show the contract being interacted with. @Gudahtt

adonesky1:
@Gudahtt thoughts on a proper resolution here? I am getting close to putting this [PR](https://github.com/MetaMask/metamask-extension/pull/13131) - which will improve the UX for NFT confirmations - into review so perhaps we should wait for that?

adonesky1:
With latest updates:

https://user-images.githubusercontent.com/34557516/157741692-cf2b0169-2294-4ce1-98a3-4a23a5f733ff.mov

adonesky1:
@Gudahtt per my most comment above ☝️  we are now showing the contract address on this confirm page.

shared/modules/transaction.utils.js
```diff
@@ -82,6 +84,32 @@ export function txParamsAreDappSuggested(transaction) {
   );
 }
 
+/**
+ * @param data
```
mcmire, line 88: Is it accurate to say that this function takes encoded transaction data and matches it against different token standard ABIs to decode that data properly? Wondering if it'd be good to add documentation for this function for people who aren't familiar. In that vein, should this be renamed? Perhaps something like `getTokenTransactionData`, `getStandardTokenTransactionData`,  `decodeTokenTransactionData`, `decodeStandardTokenTransactionData`, etc.

mcmire (commented):
One suggestion but otherwise this seems pretty straightforward!

shared/modules/transaction.utils.js
adonesky1: > Is it accurate to say that this function takes encoded transaction data and matches it against different token standard ABIs to decode that data properly?

yes I'll add some documentation. And I think seomthing like `decodeStandardTokenTransactionData` or `parseStandardTokenTransactionData` might be good?

adonesky1: Done here: https://github.com/MetaMask/metamask-extension/pull/13535/commits/8c6eda77418a327553bdad6f17c4516df39fdebc

mcmire (approved):
Thanks for the change, looks good 👍🏻

segun (approved):
LGTM

PR #13537, opened 2022-02-07
Suppress diffs for translation files by default

mcmire:
In feature branches, it is very common to make ancillary changes to
translation files. For instance, a key may be no longer in use, so in
order to satisfy CI, it must be removed across all translation files.
These changes can make reviewing PRs difficult, as one has to scroll
past them in order to find the more impactful changes. To address this
problem, this commit updates `.gitattributes` to suppress diffs
involving translation files by default.

A tradeoff to this change is that it will make it more difficult to
review PRs whose sole purpose is to update translations. That said, for
small PRs submitted by contributors, it is likely that only one or two
translation files will be updated, so it is easy to expand the diffs for
such files. For larger PRs such as those submitted by Crowdin, these
diffs are so large that GitHub ends up suppressing them anyway.

If this ends up being bothersome, one way to work around this is to use
the [Refined GitHub][1] browser extension, which allows you to
Alt/Opt-click on a single "Load diff" message to load all diffs which
are currently suppressed. So at most this would take one click to solve.

[1]: https://github.com/refined-github/refined-github

---

With this change, here's what a PR will look like when translation files are present:

Screen Shot 2022-02-07 at 2 55 49 PM


NiranjanaBinoy (approved):

jpuri (approved):
Nice change

adonesky1 (approved):
Yes please!

digiwand (approved):
I was holding off on approving this PR because I'm still fairly new to the codebase and I personally don't see the full benefits yet. Seems to be desired change.

I can roll with this. If we find out we want to undo this for any reasons, we can revert this change no problem 👍

PR #13555, opened 2022-02-08
Resolve a BigNumber issue

dan437:
## Explanation
We started seeing an issue where the extension was throwing an error when a hex value in the `minimumGasLimit` param
was passed into the `decimalToHex` function for Swaps. `decimalToHex` shouldn't be called in the first place for this.

## Screenshots
Before the fix, after submitting the form on the Build Quote page:
image

After:
No issue


ui/hooks/gasFeeInput/useGasEstimates.js
```diff
@@ -111,7 +111,7 @@ export function useGasEstimates({
   const maximumCostInHexWei = getMaximumGasTotalInHexWei(gasSettings);
 
   if (editGasMode === EDIT_GAS_MODES.SWAPS) {
-    gasSettings = { ...gasSettings, gasLimit: decimalToHex(minimumGasLimit) };
+    gasSettings = { ...gasSettings, gasLimit: minimumGasLimit };
```
jpuri, line 114: For gas limit we usually use decimal value but seems minimumGasLimit in case of this swaps is an exception.
@dan437  can we be sure that swaps will always provide a hex value (and we never need to use default which is decimal).

It would I think be better to change [this](https://github.com/MetaMask/metamask-extension/blob/develop/ui/pages/swaps/view-quote/view-quote.js#L695) code to pass decimal.

danjm: However, here we default it to hex https://github.com/MetaMask/metamask-extension/blob/develop/ui/hooks/gasFeeInput/useGasFeeInputs.js#L94

danjm: And here we default it to a hex value as well https://github.com/MetaMask/metamask-extension/blob/develop/ui/components/app/edit-gas-popover/edit-gas-popover.component.js#L47

danjm: We definitely treat it as a decimal when rendering in the UI (e.g. in the `EditGasDisplay`), but we have a separate `minimumGasLimitDec` variable that we use for that

danjm: Might be good to verify that all `minimumGasLimit` passed to `useGasFeeEstimates` are hex

danjm:
Any idea why this is not happening on mainnet?

danjm (dismissed):
LGTM

I agree with the general point of Jyoti's comment, that we need to standardize the format and numeric base of these properties, but as of today I think the specific property changed in this PR is a hex once passed to this file, although we convert it to decimal elsewhere. Standardization across these files and uses should probably be handled in a separate PR.

I also did manual QA with the standard send flow, using both EIP1559 V1 and V2 UIs

danjm (approved):

dan437:
Thanks for checking, @jpuri and @danjm! As discussed internally, we shouldn't be passing a hex value into the `decimalToHex` fn. It seems the reason it worked for Ethereum mainnet is that we were passing a hex with the `0x` prefix into the `decimalToHex` fn, but for BSC / Polygon sometimes the `0x` prefix was missing. I've tested this on multiple networks and it seems to be working well now.

I also agree with the point that we should make the usage of hex / dec values in the extension more consistent. I believe that to make it clear, we could use the `Hex` or `Dec` postfix in naming.

darkwing (approved):

ui/hooks/gasFeeInput/useGasEstimates.js
jpuri: Good point @danjm , I will spend some time on this.

jpuri (approved):

Issue #13525, opened 2022-02-07
NFT UX Audit


adonesky1:
Sweep through the recently added NFT features to ensure we're ready for release on the UX/UI front.

SayaGT:
I have found couple of bugs and some UX issues. Listing them below with some design updates wherever necessary.

**1. NFT auto detection alert in NFT tab**
On chrome, I did not get the NFT detection alert when I first logged in. Auto detect NFT & OpenSea API settings were auto enabled. The alert only appeared after I manually turned the setting off. It worked fine on Firefox. 

**2. Bug in toggle setting**
I stress tested the toggles for OpenSea API & NFT token detection setting. After few clicks, NFT detection toggle remained enabled while the OpenSea API toggle was disabled. See img below
<img width="1887" alt="Screenshot 2022-02-08 at 13 07 03" src="https://user-images.githubusercontent.com/87030673/153004130-10cc7804-f35f-4ff5-8bfc-293b248bcb96.png">


3. The NFT's didn't auto load after turning the setting on. I had to manually import them. Refreshing didn't load them either.

**4. Trouble importing NFT**
I could not import my ENS (ERC721 token) in NFT tab. I kept getting an error (see screenshot below). This appears in my mobile wallet.
<img width="1580" alt="Screenshot 2022-02-08 at 13 33 14" src="https://user-images.githubusercontent.com/87030673/153002439-91cc014a-63d9-41f9-8aed-ed76943d6ec8.png">
![Mob_NFT](https://user-images.githubusercontent.com/87030673/153046520-231626b1-0b1b-4f88-9d02-2205978a3a46.jpg)

 
**5. We need to update the UX for the NFT token displayed in Assets tab.**
Currently if a user clicks on the token in the assets tab, a message is displayed asking user to import it as an NFT. This is confusing if a user has already added that NFT. We should update the message suggesting that NFT has already been imported and this token will disappear from the asset list. [See the Figma file](https://www.figma.com/file/8Xe22jEPgcElS5tWPw8AGM/NFTs-in-extension-(MVP)?node-id=1212%3A12893).

6. Alert/ Error messages look weird in the Extension pop up. Need to adjust the padding.
[See the design suggestions in Figma file](https://www.figma.com/file/8Xe22jEPgcElS5tWPw8AGM/NFTs-in-extension-(MVP)?node-id=1212%3A12894)

7. Import NFT error state - Updated UX & design
[...]

PR #13575, opened 2022-02-09
Fix opensea/autodetect preference setting dependency relation

adonesky1:
Fixes: Issue #2 in https://github.com/MetaMask/metamask-extension/issues/13525

Explanation:  The dependency relationship between the OpenSea API and NFT AutoDetection settings was screwed up. `AutoDetect NFTs` setting toggled on in conjunction with `Enable OpenSea API` setting. If `Enable OpenSea API` is off and the user toggles on `AutoDetect NFTs`, `Enable OpenSea API` should toggle on as well. If both are enabled and the user toggles off `Enable OpenSea API`, `AutoDetect NFTs` should also toggle off. If both are enabled and the user toggles off `AutoDetect NFTs`, `Enable OpenSea API` should remain enabled.

Manual testing steps:  
  - Test the flows described above
  - Comment if they don't work or something doesn't feel intuitive.


danjm (approved):
This makes sense :+1:

jpuri (approved):

PR #13577, opened 2022-02-09
Snaps UI fixes

hmalik88:
Explanation:  This PR addresses comments made RE: snaps/flask UI work in #13462. In some places, the `Box` component was used in place of existing elements to reduce the amount of CSS.


rekmarks (commented):
LGTM, but @brad-decker is the person who's approval we really need.

brad-decker (approved):
LGTM thank you for fixing all that up

PR #13584, opened 2022-02-10
[DARKMODE] Initial new color design tokens and docs

### LINKED ISSUE #13564: Import color design tokens from design token library
**Description**
Now that the new color style naming conventions have been decided on, we can start adding them to the code base. These variables will be color agnostic, theme compatible and semantically correct. This would be the first stage of making sure we are using consistent colors across the app and make all of our styles theme ready.

The end goal is for the design tokens to be updated in Figma, pushed to the design-tokens repo and then updated in the different apps. For MVP we can add the css vars and automate the process in future.

**Technical Details**
- import design token stylesheet from [`@metamask/design-tokens`](https://github.com/MetaMask/design-tokens) repo
- create documentation for their use in storybook

georgewrmarshall:
Fixes: #13564

Explanation:  Adding new design token color style sheets and docs. This is the first step towards theming and dark mode. The next step would be to swap out all of the static hex and current color variables to the ones in this PR.

**Notes**
- Values of the tokens may be subject to change. This should only "tweak" the colors and shouldn't effect how they are to be used throughout the codebase.
- Using HSLA color mode values as we thought it might help consistency with the transparent values and in my opinion is  superior and most logical and of the colormodes. Might be better to use hex values until we adopt HSLA as a company.. I did only find out that you can do alpha with HEX values now so those could be updated too.
- Ideally we should be using the [`design-tokens`](https://github.com/MetaMask/design-tokens) package to generate these stylesheets but for MVP and to help with automation we could start using the end result and then hook everything up to be automated.
- Would be nice to auto generate the documentation based on the design-tokens repo

**TO DO**
- [ ] Update `ui/helpers/constants/design-system.js` with all the new colors

Manual testing steps:  
  - Run `yarn storybook`
  - Go to color docs at http://localhost:6006/?path=/story/design-tokens-color-getting-started--page
  - Read docs and check against code
 
## Screenshots

colors getting started
Screen Shot 2022-02-10 at 12 37 05 PM
Screen Shot 2022-02-10 at 12 37 17 PM


.storybook/preview-head.html
```diff
@@ -24,4 +24,7 @@
     font-family: var(--font-family-monospace) !important;
     font-size: var(--font-size-code) !important;
   }
+  .docblock-colorpalette div:not(:first-child) {
+    grid-column-gap: 16px;
+  }
 </style>
```
georgewrmarshall, line 30: Adds gap to `<ColorPaletteItem /> storybook docs component. Admittedly I'm probably not using it as intended and trying to squeeze the description into the subtitle but I think it works for now.

.storybook/preview.js
```diff
@@ -27,7 +27,15 @@ addParameters({
...
+        'Design Tokens',
+        ['Color'],
+        ['Getting Started', 'Light Theme Colors', 'Dark Theme Colors'],
+        'Components',
+        ['UI', 'App'],
+        'Pages',
+      ],
     },
   },
 });
```
georgewrmarshall, line 41: Reordering stories so "Design Tokens" comes after "Getting Started" and doesn't get pushed to the bottom

PR #13585, opened 2022-02-10
Adding MEGO (Metaverse EGO) compatibility on Polygon

turinglabsorg:
Adding feature: MEGO (Metaverse EGO) compatibility.

Explanation:  
MEGO (Metaverse EGO) is a naming service living on Polygon Blockchain. It's very similar to ENS with the only difference that ERC-721 are served on-chain, not relying on IPFS. It supports multiple "extensions" and names are registered forever. Here our website to have a look at how it works: https://mego.cx.

To achieve the result I just added the Polygon constants (not every constant was previously registered) and created an if statement to check if the blockchain is Polygon or not. If Polygon the contract is called, using `mego.js` file. I used the same naming as ENS, so the function is `mego.lookup`.

Manual testing steps:  
  - Connect to Polygon network
  - Click on `Send` button
  - Search for `turinglabs.dao` to see how the address is extracted from the contract

Video explaining the integration:

https://user-images.githubusercontent.com/14848252/153343445-979e685c-86e2-4541-8dc2-6505e9c68edf.mp4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Add MEGO compatibility for Polygon Network
 ## [10.9.1]
 ### Fixed
 - Fixed application error when adding certain tokens ([#13484](https://github.com/MetaMask/metamask-extension/pull/13484))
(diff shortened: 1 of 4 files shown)


turinglabsorg:

danjm:
Hello,

Thanks for your interest in MetaMask.

Unfortunately, we are not accepting additions of other name services at this time. In the future we will provide a standardized path for adding such functionality to MetaMask, quite possibly through our Snaps system (https://docs.metamask.io/guide/snaps.html#what-is-snaps). However, we cannot yet estimate when that will be.

You can follow MetaMask and the MetaMask Snaps project for relevant updates.

PR #13587, opened 2022-02-10
Adding check for transaction.id in useTransactionEventFragment hook

jpuri:
PR addresses issue-4 detailed here: https://docs.google.com/document/d/1i2xw4U8qJSf9v_MbzBtH154F9_DYbR-4nR79CCAWhHI/edit

Reason of the issue is that as part of 1559 V2 we have added more metrics event capturing to transaction flow, swap transaction on transaction confirmation page still does not have a transaction_id which was causing this exception.


brad-decker (approved):
LGTM

adonesky1 (approved):
LGTM

PR #13588, opened 2022-02-10
Show users a warning when they are sending directly to a token contract

### LINKED ISSUE #12642: Warning not being shown when sending tokens to a known token contract address

### LINKED ISSUE #13486: Improve "Known Contract Address" warning
### Describe the bug

Background: See [My reddit post](https://www.reddit.com/r/CryptoCurrency/comments/sgdxtg/psa_we_should_do_something_for_todays_half/
)
And [See this tweet](https://twitter.com/alex_kroeger/status/1487821077669552132?s=20&t=gzYlmQVEG7k5vw6PEphA3A)

TL;DR: Metamask should display a warning dialog whenever user try to send ERC-20 token to ANY Contract Address directly. Many other wallet software already added this feature e.g. rainbow.me, block wallet. 
The current ["known contract address" solution](https://metamask.zendesk.com/hc/en-us/articles/360020028092-What-is-the-known-contract-address-warning-) is far from enough to protect users. 

Here is my proposed warning text:

> WARNING
> The recipient address you typed in is a Contract address.
> Typically, if you want to give asset to a contract, you should use the dApp of the contract, usually in the form of a website, then follow the instructions there. NOT transferring it here.
> We do not know if the recipient contract can handle a direct token transfer like what you are trying to do, or not. Usually, if it is an DeFi contract, or a token contract, it does not have the ability.
> If that’s the case and you proceed, ALL ASSET SENT will be PERMANENTLY LOST, there is NO WAY to RECOVER.
> Are you sure you want to proceed?
> Cancel (In Big Red Button).
> I want to proceed anyway.
> Option 1: It’s my smart wallet. Mark it as my smart wallet address. Proceed.
> Option 2: I’m very sure this is what I want to do. Proceed.

Full article:
A man just lost 195 WETH, by sending it back to WETH contract address, thinking it would return him raw ETH. He was mislead by introductions on Binance Academy (see original post on r/ethereum). Three more victims showed up today, losing 0.1 0.05 0.03 WETH respectively, check WETH contract addresses’ own WETH holding.
Since ERC-20 standard does not provide a way to notify contracts when they receive token transfers, and it does not mandate the contract the ability to send tokens back out, people who bring the real world common sense may think it wrongly as “I sent it to the account get Wrapped token, now I send it back I should get raw token back”, result in permanent unrecoverable loss.
Vitalik pointed out that Smart, Self-Custodial Wallets are Smart Contracts too, so we cannot ban all transfers to Contract Address all together. I suggest Metamask on all platforms show a warning as shown above when user try to send any ERC-20 token to a Contract Address.
It might be the last chance before unrecoverable loss happens, for many users to realize that it is not the correct operation to do what they want to achieve.

### Steps to reproduce

The current "known contract address" solution at https://metamask.zendesk.com/hc/en-us/articles/360020028092-What-is-the-known-contract-address-warning- is far from enough to protect users. 
This mechanism must detect any contract address by querying ethereum backend, because you can't whitelist all of them.
The contract address of WETH9 token is 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2 . You can all try it out to copy paste it into your Metamask wallet and there will be absolutely ZERO warning about it being a contract address.

_No response_

### Version

10.8.1

### Build type

_No response_

### Browser

Chrome

### Operating system

MacOS

### Hardware wallet

Ledger

### Additional context

_No response_

danjm:
Fixes: #12642
Fixes: #13486
Replaces #12983

This PR ensures we are showing sufficient warnings to users when they are sending to token contracts on their current network or when they are sending to known contract addresses on mainnet.

Screen captures of new behaviour

https://user-images.githubusercontent.com/7499938/153390932-93227c67-a299-4a33-9a7a-e88a8d7d9f6d.mp4

https://user-images.githubusercontent.com/7499938/153390934-2df9b0d2-69d1-4a76-992c-909bad4ab9f7.mp4

### DIFF
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -954,14 +954,6 @@ const slice = createSlice({
           // are no longer valid when sending native currency.
           draftTransaction.recipient.error = null;
         }
-
-        if (
-          draftTransaction.recipient.warning === KNOWN_RECIPIENT_ADDRESS_WARNING
-        ) {
-          // Warning related to sending tokens to a known contract address
-          // are no longer valid when sending native currency.
-          draftTransaction.recipient.warning = null;
-        }
       }
       // if amount mode is MAX update amount to max of new asset, otherwise set
       // to zero. This will revalidate the send amount field.
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -2567,8 +2567,10 @@ export function getRecipientUserInput(state) {
 }
 
 export function getRecipientWarningAcknowledgement(state) {
-  return getCurrentDraftTransaction(state).recipient
-    .recipientWarningAcknowledged;
+  return (
+    getCurrentDraftTransaction(state).recipient?.recipientWarningAcknowledged ??
+    false
+  );
 }
 
 // Overall validity and stage selectors
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -69,6 +69,7 @@ import {
   calcTokenAmount,
   getTokenAddressParam,
   getTokenValueParam,
+  getTokenMetadata,
 } from '../../helpers/utils/token-util';
 import {
   checkExistingAddresses,
@@ -382,6 +383,7 @@ export const draftTransactionInitialState = {
     error: null,
     nickname: '',
     warning: null,
+    recipientWarningAcknowledged: false,
   },
   status: SEND_STATUSES.VALID,
   transactionType: TRANSACTION_ENVELOPE_TYPES.LEGACY,
@@ -1154,6 +1156,20 @@ const slice = createSlice({
       state.recipientInput = '';
       state.recipientMode = action.payload;
     },
+
+    updateRecipientWarning: (state, action) => {
+      const draftTransaction =
+        state.draftTransactions[state.currentTransactionUUID];
+      draftTransaction.recipient.warning = action.payload;
+    },
+
+    acknowledgeRecipientWarning: (state) => {
+      const draftTransaction =
+        state.draftTransactions[state.currentTransactionUUID];
+      draftTransaction.recipient.recipientWarningAcknowledged = true;
+      slice.caseReducers.validateSendState(state);
+    },
+
     /**
      * Updates the value of the recipientInput key with what the user has
      * typed into the recipient input field in the UI.
@@ -1316,10 +1332,13 @@ const slice = createSlice({
           draftTransaction.recipient.error = null;
           draftTransaction.recipient.warning = null;
         } else {
-          const isSendingToken =
-            draftTransaction.asset.type === ASSET_TYPES.TOKEN ||
-            draftTransaction.asset.type === ASSET_TYPES.COLLECTIBLE;
-          const { chainId, tokens, tokenAddressList } = action.payload;
+          const {
+            chainId,
+            tokens,
+            tokenAddressList,
+            isProbablyAnAssetContract,
+          } = action.payload;
+
           if (
             isBurnAddress(state.recipientInput) ||
             (!isValidHexAddress(state.recipientInput, {
@@ -1331,10 +1350,9 @@ const slice = createSlice({
               ? INVALID_RECIPIENT_ADDRESS_ERROR
               : INVALID_RECIPIENT_ADDRESS_NOT_ETH_NETWORK_ERROR;
           } else if (
-            isSendingToken &&
             isOriginContractAddress(
               state.recipientInput,
-              draftTransaction.asset.details.address,
+              draftTransaction.asset?.details?.address,
             )
           ) {
             draftTransaction.recipient.error = CONTRACT_ADDRESS_ERROR;
@@ -1342,12 +1360,12 @@ const slice = createSlice({
             draftTransaction.recipient.error = null;
           }
           if (
-            isSendingToken &&
-            isValidHexAddress(state.recipientInput) &&
-            (tokenAddressList.find((address) =>
-              isEqualCaseInsensitive(address, state.recipientInput),
-            ) ||
-              checkExistingAddresses(state.recipientInput, tokens))
+            (isValidHexAddress(state.recipientInput) &&
+              (tokenAddressList.find((address) =>
+                isEqualCaseInsensitive(address, state.recipientInput),
+              ) ||
+                checkExistingAddresses(state.recipientInput, tokens))) ||
+            isProbablyAnAssetContract
           ) {
             draftTransaction.recipient.warning = KNOWN_RECIPIENT_ADDRESS_WARNING;
           } else {
@@ -1355,6 +1373,7 @@ const slice = createSlice({
           }
         }
       }
+      slice.caseReducers.validateSendState(state);
     },
     /**
      * Checks if the draftTransaction is currently valid. The following list of
@@ -1390,6 +1409,12 @@ const slice = createSlice({
         case new BigNumber(draftTransaction.gas.gasLimit, 16).lessThan(
           new BigNumber(state.gasLimitMinimum),
         ):
+          draftTransaction.status = SEND_STATUSES.INVALID;
+          break;
+        case draftTransaction.recipient.warning === 'loading':
+        case draftTransaction.recipient.warning ===
+          KNOWN_RECIPIENT_ADDRESS_WARNING &&
+          draftTransaction.recipient.recipientWarningAcknowledged === false:
           draftTransaction.status = SEND_STATUSES.INVALID;
           break;
         default:
@@ -1589,9 +1614,16 @@ const {
   validateRecipientUserInput,
   updateRecipientSearchMode,
   addHistoryEntry,
+  acknowledgeRecipientWarning,
 } = actions;
 
-export { useDefaultGas, useCustomGas, updateGasLimit, addHistoryEntry };
+export {
+  useDefaultGas,
+  useCustomGas,
+  updateGasLimit,
+  addHistoryEntry,
+  acknowledgeRecipientWarning,
+};
 
 // Action Creators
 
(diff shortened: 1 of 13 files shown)


darkwing (dismissed):
This looks good, but needs tests passing.

ryanml:
Now using V2 of `ActionableMessage`:
Screen Shot 2022-02-17 at 11 34 22 AM (1)

darkwing:
From a code perspective this is awesome.  I do have a suggestion about messaging though.  Could we change to something like`Warning: you are about to send to a token contract which could result in a loss of funds. Click to proceed. $1`.  I think we could be a bit more alarmist here.

holantonela:
Thanks for this suggestion @darkwing. Let's move on with that copy @ryanml.

ryanml:
Copy updated

ui/pages/send/send-content/send-content.component.js
```diff
@@ -124,6 +140,36 @@ export default class SendContent extends Component {
...
+    const { t } = this.context;
+    return (
+      <div className="send__warning-container">
+        <ActionableMessage
+          type="danger"
+          useIcon
+          iconFillColor="#d73a49"
+          primaryActionV2={{
+            label: t('tryAnywayOption'),
+            onClick: () => acknowledgeRecipientWarning(),
```
brad-decker, line 154: [REQ] just pass method name as it takes no props, the intermediate method is unnecessary.

ui/pages/send/send-content/send-content.component.js
```diff
@@ -124,6 +140,36 @@ export default class SendContent extends Component {
...
+          }}
+          message={t('sendingToTokenContractWarning', [
+            <a
+              key="contractWarningSupport"
+              target="_blank"
+              rel="noopener noreferrer"
+              className="send__warning-container__link"
+              href="https://metamask.zendesk.com/hc/en-us/articles/360020028092-What-is-the-known-contract-address-warning-"
```
brad-decker, line 162: [REQ] There's a new zendesk urls constant in the constants folder we should use for this.

ui/ducks/send/send.js
```diff
@@ -1061,52 +1068,57 @@ const slice = createSlice({
       // input field
       state.recipient.userInput = action.payload;
     },
-    validateRecipientUserInput: (state, action) => {
+    updateRecipientWarning: (state, action) => {
+      state.recipient.warning = action.payload;
+    },
+    validateRecipientAddress: (state, action) => {
```
brad-decker, line 1074: [Q] I'm not against it just curious why this was changed? its been a while but i thought that the original intent was to validate the user's input *Before setting it as the actual address of the recipient in state*

brad-decker (changes requested):

ui/pages/send/send-content/send-content.component.js
ryanml: Fixed

ui/pages/send/send-content/send-content.component.js
ryanml: done

ui/ducks/send/send.js
ryanml: cc: @danjm

brad-decker (dismissed):

ui/ducks/send/send.js
```diff
@@ -1061,52 +1068,57 @@ const slice = createSlice({
...
       const { asset, recipient } = state;
 
       if (
         recipient.mode === RECIPIENT_SEARCH_MODES.MY_ACCOUNTS ||
-        recipient.userInput === '' ||
-        recipient.userInput === null
+        recipient.address === '' ||
```
Gtonizuka, line 1079: can this ever be undefined? If so we could just get away with `if(Boolean(recipient.address )`

ui/ducks/send/send.js
```diff
@@ -1530,31 +1552,38 @@ export function updateSendAsset({ type, details }) {
...
     const state = getState();
     const chainId = getCurrentChainId(state);
     const tokens = getTokens(state);
     const useTokenDetection = getUseTokenDetection(state);
     const tokenAddressList = Object.keys(getTokenList(state));
-    debouncedValidateRecipientUserInput(dispatch, {
+    const { symbol, decimals } = await getSymbolAndDecimals(
+      userInput,
+      tokenAddressList,
+    );
```
adonesky1, line 1579: We might consider using `[getTokenStandardAndDetails](https://github.com/MetaMask/controllers/blob/a799f42ad7448a951ae02fd00be7fbfe6904aa8e/src/assets/AssetsContractController.ts#L167)` since it actually returns the contract standard and a more details with which we can inform the user more clearly about the contract they are sending to.

adonesky1: ☝️  This would also allow us to tell users if they are sending to an ERC721 or ERC1155 contract, which this current approach doesn't enable.

ui/ducks/send/send.js
adonesky1: ➕

adonesky1: It looks like you're setting recipient.address to `''` right before calling into here and then checking whether its `''`
on line [1642](https://github.com/MetaMask/metamask-extension/pull/13588/files#diff-bad76d7279418f3520a0ca93cfaeeb6fcbb1f0d4782f6ff942d422632a1c2780L1611)

adonesky1: oh wait nvm thats in a reset method

DELJUJAZI (dismissed):

ui/ducks/send/send.js
darkwing: This appears to still require an answer.

ui/ducks/send/send.js
darkwing: I like this idea!

ui/ducks/send/send.js
```diff
@@ -1313,10 +1323,14 @@ const slice = createSlice({
         recipient.error = null;
         recipient.warning = null;
       } else {
```
danjm, line 1325: `isSendingToken` is removed because we want to show the user a warning regardless of what currency they are sending to a token contract

ui/ducks/send/send.js
```diff
@@ -1821,14 +1844,18 @@ export function updateSendAsset({ type, details }) {
...
  */
-const debouncedValidateRecipientUserInput = debounce((dispatch, payload) => {
-  dispatch(
-    addHistoryEntry(
-      `sendFlow - user typed ${payload.userInput} into recipient input field`,
-    ),
-  );
-  dispatch(validateRecipientUserInput(payload));
-}, 300);
+const debouncedValidateRecipientUserInput = debounce(
```
danjm, line 1847: The addition of resolve as a parameter allows the caller to asynchronously await the debounced call

ui/pages/send/send.js
```diff
@@ -111,7 +112,6 @@ export default function SendTransactionScreen() {
               `sendFlow - User pasted ${text} into address field`,
             ),
           );
```
danjm, line 114: This update recipient call is being removed because pastes also trigger the change listener on the input in the EnsInput component, which results in `onValidAddressTyped` being called, which (as you can see above) also calls `updateRecipient`. We don't want to call that twice.

ui/ducks/send/send.js
```diff
@@ -2486,6 +2559,11 @@ export function getRecipientUserInput(state) {
   return state[name].recipientInput;
 }
 
+export function getRecipientWarningAcknowledgement(state) {
+  return getCurrentDraftTransaction(state).recipient
+    .recipientWarningAcknowledged;
```
brad-decker, line 2564: Should probably use optional chaining with a sensible default (false?) because the draftTransaction *could be an empty object* if no draft transaction is currently found. This is an edge case but its better to guard against it.

brad-decker (changes requested):

darkwing:
I was able to trigger a weird error with this:

STR:
1.  Click "Send" from the home screen
2.  Use the Chainlink contract address:  `0x514910771AF9Ca656af840dff83E8264EcF986CA`
3.  See the contract warning message (yay!)
4.  Click the currency box
5. See that the contract warning goes away, and the screen shows a "new address" block


https://user-images.githubusercontent.com/46655/178345905-79cfd968-d4bd-4112-a8ef-3ed50e8c14fc.mp4



I was also able to trigger an actual error by clicking on something else, but don't recall how:

WeirdError2

danjm:
> Click "Send" from the home screen
> Use the Chainlink contract address: 0x514910771AF9Ca656af840dff83E8264EcF986CA
> See the contract warning message (yay!)
> Click the currency box
> See that the contract warning goes away, and the screen shows a "new address" block

@darkwing That's not a bug. We always show that "new address" block when sending to a new address. We could claim that the UX of showing one block after another is not good, but I don't think that is a bug.

Meanwhile, the type error identified is a bug on develop: https://github.com/MetaMask/metamask-extension/issues/15188

ui/pages/send/send.js
```diff
@@ -96,10 +96,11 @@ export default function SendTransactionScreen() {
         userInput={userInput}
         className="send__to-row"
         onChange={(address) => dispatch(updateRecipientUserInput(address))}
-        onValidAddressTyped={(address) => {
+        onValidAddressTyped={async (address) => {
           dispatch(
             addHistoryEntry(`sendFlow - Valid address typed ${address}`),
           );
```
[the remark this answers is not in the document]
danjm: @brad-decker Can you add more clarification/detail to this. `updateRecipient` is not called in the debounce handler called from `updateRecipientUserInput` so I'm not exactly sure what you are suggesting.

ui/ducks/send/send.js
danjm: done in ca614adcb

danjm:
> That's not a bug. We always show that "new address" block when sending to a new address. We could claim that the UX of showing one block after another is not good, but I don't think that is a bug.

I was wrong about that, I misunderstood the bug.

The bug was that the recipient warning went away went clicking the Asset dropdown. I fixed this with 89428b9ab

ryanml (approved):

brad-decker (approved):

PR #13594, opened 2022-02-10
ConfirmImportToken: Convert to Functional Component; Add storybook controls

### LINKED ISSUE #13071: ConfirmImportToken story: convert knobs and actions to controls / args
**Describe the bug**
As part of a push to convert all stories that use deprecated `@storybook/addon-knobs` and `@storybook/addon-actions` to use [controls](https://storybook.js.org/docs/react/essentials/controls) and [action argType annotation](https://storybook.js.org/docs/react/essentials/actions#action-argtype-annotation)

Convert `ConfirmImportToken ` story in `ui/pages/confirm-import-token/confirm-import-token.stories.js` to use [controls](https://storybook.js.org/docs/react/essentials/controls) and [action argType annotation](https://storybook.js.org/docs/react/essentials/actions#action-argtype-annotation)

Read the [MetaMask extension storybook documentation guidelines](https://metamask.github.io/metamask-storybook/?path=/story/getting-started-documentation-guidelines--page) to get a better understanding of writing stories. Providing MDX documentation although helpful, is a bonus and not a requirement for this issue.

To start storybook:
```
yarn storybook
```

**Requirements**
- [ ] Story has `argTypes` that align with component api / props
- [ ] All instances of `@storybook/addon-knobs` have been removed in favour of control args
- [ ] All instances of `@storybook/addon-actions` have been removed in favour of [action argType annotation](https://storybook.js.org/docs/react/essentials/actions#action-argtype-annotation)

Any further improvements to the story interaction or documentation is a bonus

**Images**

Screen Shot 2021-12-12 at 6 18 01 PM

digiwand:
Fixes: n/a

Explanation:  Similarly to https://github.com/MetaMask/metamask-extension/pull/13377 and https://github.com/MetaMask/metamask-extension/pull/13526, we are converting this component to a Functional Component.

Additionally:
- add jest tests
- add storybook controls

Manual testing steps:  
  - On the main page and underneath the list of accounts, click "import tokens"
  - Search any token and select it
  - Click "Next" 
  - Test this "Import Tokens" page works as expected


ui/pages/confirm-import-token/confirm-import-token.stories.js
```diff
@@ -11,45 +9,39 @@ import ConfirmAddToken from '.';
...
+  const { metamask: state } = store.getState();
 
-const PageSet = ({ children }) => {
-  const symbol = text('symbol', 'TRDT');
-  const state = store.getState();
-  const pendingTokensState = state.metamask.pendingTokens;
-  // only change the first token in the list
   useEffect(() => {
-    const pendingTokens = { ...pendingTokensState };
-    pendingTokens['0x33f90dee07c6e8b9682dd20f73e6c358b2ed0f03'].symbol = symbol;
```
digiwand, line 25: I didn't see the benefit of having this here. I think the intention of using the `text` knob was to hook into the controls, but there was no control setup for this.  The user is now about to update the entire list of `pendingTokens` with the controls added in this PR.

tmashuang:
For the e2e failure, the regression happening is that after adding the token from the search list it should route to the token-overview screen. Instead, it is navigating to the home screen.

<details>
<summary>Before</summary>

</details>

<details>
<summary>E2E issue</summary>

</details>

AndreySttar (approved):

ui/pages/confirm-import-token/confirm-import-token.test.js
```diff
@@ -0,0 +1,103 @@
...
+import reactRouterDom from 'react-router-dom';
+import configureMockStore from 'redux-mock-store';
+import thunk from 'redux-thunk';
+import { fireEvent, screen } from '@testing-library/react';
+import {
+  ASSET_ROUTE,
+  IMPORT_TOKEN_ROUTE,
+} from '../../helpers/constants/routes';
+import { addTokens, clearPendingTokens } from '../../store/actions';
+import { renderWithProvider } from '../../../test/jest/rendering';
```
mcmire, line 11: [nit] You can drop `/rendering` from this path if you like. `test/jest` exports helpers from `test/jest/rendering` as well as `screen` and `fireEvent` from RTL.

ui/pages/confirm-import-token/confirm-import-token.js
```diff
@@ -0,0 +1,134 @@
...
+  const dispatch = useDispatch();
+  const history = useHistory();
+
+  const mostRecentOverviewPage = useSelector((state) =>
+    getMostRecentOverviewPage(state),
+  );
+  const pendingTokens = useSelector((state) => state.metamask.pendingTokens);
+
+  const tokenAddedEvent = (pendingToken) => {
+    metricsEvent({
```
mcmire, line 32: I think this action you're using isn't equivalent to the one that was being used before and this one requires different arguments. That's why the e2e test is failing — because this fails before `history.push` is called. You'll want to take a look at other places that use `metricsEvent` to know what to pass. I think it may be something like

``` javascript
metricsEvent({
  eventOpts: {
    name: 'Token Added',
    category: 'Wallet',
  }
})
```

I am not sure that `sensitiveProperties` is being used in this case so I am not sure it is necessary. But then again I am not completely familiar with how the metrics stuff works. Perhaps someone else can chime in here.

ui/pages/confirm-import-token/confirm-import-token.test.js
```diff
@@ -0,0 +1,103 @@
+import React from 'react';
+import reactRouterDom from 'react-router-dom';
+import configureMockStore from 'redux-mock-store';
```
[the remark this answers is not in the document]
digiwand: Thank you for bringing this up! From the same link you posted, the way to test Redux is to

> render a <Provider> with a real store instance wrapping the component/s being tested.

I am with you. I agree that using a mock store is not necessary. I think we should move forward using the real Redux store to preserve the integrity of our integration tests. It would be nice to deprecate the 49 `redux-mock-store` instances in this repo as we come across them and move onto RTL and hook-based components.

Made the change

ui/pages/confirm-import-token/confirm-import-token.js
brad-decker: So the eventOpts parameter is for the old metametrics.The new one, which the old implementation used too, has the payload that is mentioned here. So what is in the code is correct but the wrong context is being used. This is because we don't export the context from the new provider to prevent it from being used. Instead, using the useNewMetricEvent hook from the 'useMetricEvent'  file should be used. 

The sensitiveProperties here is used to keep on chain data anonymous in our metrics payload. Technically source, and unlisted should be hoisted into the `properties` payload. If we tracked the others to a metrics id then we could, in theory, identify a wallet -> metametrics id by examining the tokens for which wallets have balances and aligning them to the ids we know have those tokens in their metamask. By specifying in sensitiveProperties we make this orders of mangitude more difficult (preserving user privacy)

ui/pages/confirm-import-token/confirm-import-token.test.js
```diff
@@ -0,0 +1,103 @@
...
+import thunk from 'redux-thunk';
+import { fireEvent, screen } from '@testing-library/react';
+import {
+  ASSET_ROUTE,
+  IMPORT_TOKEN_ROUTE,
+} from '../../helpers/constants/routes';
+import { addTokens, clearPendingTokens } from '../../store/actions';
+import { renderWithProvider } from '../../../test/jest/rendering';
+import mockState from '../../../test/data/mock-state.json';
```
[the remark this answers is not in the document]
digiwand: @mcmire 

Passing the state into `renderComponent` is a great idea 👍  

Mocking the data directly in the file will keep our separation of concerns cleaner, and we can avoid complexities, as you mentioned. I agree that we don't need to pass all the states from `mock-state.json`. Also, on second thought, implying `pendingTokens` share `token` attributes is not a great approach as its model can be changed separately from the token model. 

I'd love to hear more of your thoughts on fixtures. I think tests that break due to changes to the shared data objects can also be beneficial, so I'm not exactly sure to what extent you mean here
> In my experience using shared objects (also known as fixtures) in tests can lead to brittle tests. 

This said, applying TypeScript can also catch data disparities without fixtures. Are you proposing no shared data, or is there a subset of shared mock data that you think could be helpful in the codebase?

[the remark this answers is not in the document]
mcmire: > I'd love to hear more of your thoughts on fixtures. I think tests that break due to changes to the shared data objects can also be beneficial, so I'm not exactly sure to what extent you mean here
> Are you proposing no shared data, or is there a subset of shared mock data that you think could be helpful in the codebase?

For components that rely upon state to do their job, either that state needs to be supplied in the test or the selector functions used to access pieces of that state need to be mocked. Assuming that the state is not mocked, then the state that is supplied for one component's tests will probably be different than the state supplied for another component's tests. Even for a single component, the state requirements may differ from test to test depending on what the test is testing.





Yes, to be honest, I don't think we need shared data at all. I think if we look at tests for each component

ui/pages/confirm-import-token/confirm-import-token.js
```diff
@@ -0,0 +1,134 @@
...
+  const t = useContext(I18nContext);
+  const dispatch = useDispatch();
+  const history = useHistory();
+
+  const mostRecentOverviewPage = useSelector((state) =>
+    getMostRecentOverviewPage(state),
+  );
+  const pendingTokens = useSelector((state) => state.metamask.pendingTokens);
+
+  const tokenAddedEvent = (pendingToken) => {
```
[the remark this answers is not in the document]
digiwand: I actually had this same thought to prepend this with `track`. I went with conforming to existing patterns. Thanks for bringing this up! :) 

I think it might be sufficient to use `trackTokenAdded`, but I don't think I know enough context of the app to propose this at this time. I can go with `trackTokenAddedEvent` in the meantime.

ui/pages/confirm-import-token/confirm-import-token.js
```diff
@@ -0,0 +1,134 @@
...
+import Button from '../../components/ui/button';
+import Identicon from '../../components/ui/identicon';
+import TokenBalance from '../../components/ui/token-balance';
+import { I18nContext } from '../../contexts/i18n';
+import { MetaMetricsContext } from '../../contexts/metametrics';
+import { getMostRecentOverviewPage } from '../../ducks/history/history';
+import { addTokens, clearPendingTokens } from '../../store/actions';
+
+const getTokenName = (name, symbol) => {
+  return typeof name === 'undefined' ? symbol : `${name} (${symbol})`;
```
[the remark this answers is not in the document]
digiwand: LGTM 👍

ui/pages/confirm-import-token/confirm-import-token.test.js
digiwand: sweeet. Thank you!

ui/pages/confirm-import-token/confirm-import-token.js
digiwand: @mcmire @brad-decker 

I believe I understand. I initially tried to maintain logic by preserving the context used by the previous code, and to do this I used this `metricsEvent` here. Thank you two for all of this info. Glad to know we can transition to `useNewMetricEvent` with the same args from our older `metricsEvent` method and to get this clarification with `sensitiveProperties`.

`useNewMetricEvent` was a bit tricky to use with dynamic props since it returns a `useCallback` hook. I consulted with @ryanml a bit about what we should do in this case and we thought utilizing `useState` might be fine here. Then I updated the code. Please let me know what y'all think.

During these updates, I found the `history.push` issue. My mistake was replacing `componentDidMount` with `useEffect` and forgetting to force this hook to run just once by passing an empty dependency list. This should be fixed now.

digiwand:
@tmashuang

Thank you for the investigation and helpful gifs! Issue is now fixed

darkwing (approved):

jpuri (approved):
🚀

digiwand:
Thanks, @darkwing & @jpuri! Merging with the two approvals. 

@mcmire @brad-decker  Thank you for the reviews! I can create a new PR if there are any further change requests

ui/pages/confirm-import-token/confirm-import-token.js
brad-decker: @mcmire @digiwand `metricsEvent` is the context method name for the old analytics system context that was converted to translate the old metrics payloads into ones that work with the new ones. those payloads have the `eventOpts` object definition in them like @mcmire pointed out. 

`trackEvent` is the context value that the old code used, which is the correct new method for the new metametrics context. Yes its confusing.

in terms of the useCallback problem, this has been expressed before. As long as your current implementation is working and sending the right event payload then this is fine. I'll take into consideration how we might make the context better for consumption directly. To be fair, your original implementation using context would have been fine if it was the NEW metametrics context and the 'trackEvent' method. 😬 sorry about that.

digiwand: @brad-decker @mcmire Ooh like using 
```
import { MetaMetricsContext } from '../../contexts/metametrics.new';
```
instead of 
```
import { MetaMetricsContext } from '../../contexts/metametrics';
```

I was trying to preserve the context, but maybe I misunderstood the context used in the older, class implementation. I don't fully understand the differences between each of the providers, but I think I understand we should stick to using this newer context; ideally using `useNewMetricEvent` or possibly `import { MetaMetricsContext } from '../../contexts/metametrics.new';` depending on the circumstances. Thanks for elaborating on this!

brad-decker: No problem. With the class based approach its literally just the *name* of the context item. If you review the two metrics contexts one is trackEvent the other is metricsEvent or something ... which name you used indicates which context you get lol.

PR #13599, opened 2022-02-11
Include transactions from all networks in state logs

### LINKED ISSUE #13548: Include transactions from all networks in state logs
Currently our state logs only contain the `currentNetworkTxList`. This, as the name implies, only contains transactions from the selected network. This can be a problem because when users submit state logs to us, they may not have the network selected that contains the transaction(s) we need to investigate. It would be better just to include all transactions from all networks in state logs.

NiranjanaBinoy:
Fixes: #13548 

Explanation:  Adding a new property `completeTxList` to the state log, which holds the list of all transactions done by the user irrespective of chainId.

Manual testing steps:  
  - Open the Extension application; right-click and select inspect.
  - In the console, type in `copy(await logState(true))` to copy the state log to clipboard.
  - Search for `completeTxList` and verify all your transactions are present on the list irrespective of the chainId.

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -1325,6 +1325,7 @@ export default class MetamaskController extends EventEmitter {
       createTransactionEventFragment: txController.createTransactionEventFragment.bind(
         txController,
       ),
+      getTransactions: txController.getTransactions.bind(txController),
 
       // messageManager
       signMessage: this.signMessage.bind(this),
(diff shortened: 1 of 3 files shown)


darkwing (approved):
Looks good to me.  I have two ideas or suggestions though:

1.  Should we make this method name much more specific, like 'getAllTransctionsOnAllNetworks'?
2.  Do we need to coordinate documentation updates with the support team?

danjm:
I don't think we should rename this, because we have to pass the `filterToCurrentNetwork: false,` flag for this to get all transcations.

And I think it is okay not to modify support documents because this just adds additional informaiton to the state logs, it doesn't change the method of retrieving the logs at all.

danjm (approved):

PR #13617, opened 2022-02-14
Fix: advance defaults should never be used for swaps

jpuri:
Fixes: RC v10.10.0 RC issue: The advanced gas fee default setting is used instead of swap suggested as displayed

### DIFF
--- a/app/scripts/controllers/transactions/index.js
+++ b/app/scripts/controllers/transactions/index.js
@@ -55,6 +55,11 @@ const hstInterface = new ethers.utils.Interface(abi);
 
 const MAX_MEMSTORE_TX_LIST_SIZE = 100; // Number of transactions (by unique nonces) to keep in memory
 
+const SWAP_TRANSACTION_TYPES = [
+  TRANSACTION_TYPES.SWAP,
+  TRANSACTION_TYPES.SWAP_APPROVAL,
+];
+
 /**
  * @typedef {import('../../../../shared/constants/transaction').TransactionMeta} TransactionMeta
  * @typedef {import('../../../../shared/constants/transaction').TransactionMetaMetricsEventString} TransactionMetaMetricsEventString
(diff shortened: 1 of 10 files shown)


app/scripts/controllers/transactions/index.js
```diff
@@ -337,9 +337,10 @@ export default class TransactionController extends EventEmitter {
    *
    * @param txParams
    * @param origin
+   * @param transactionType
    * @returns {txMeta}
    */
-  async addUnapprovedTransaction(txParams, origin) {
+  async addUnapprovedTransaction(txParams, origin, transactionType) {
```
danjm, line 343: To make sure that this new parameter is not accidentally misused, could we validate this third parameter by checking that, `if (transactionType !== undefined)` then `transactionType` should equal either `TRANSACTION_TYPES.SWAP_APPROVAL` or `TRANSACTION_TYPES.SWAP`, and otherwise an error should be thrown

ui/ducks/swaps/swaps.js
```diff
@@ -881,7 +882,11 @@ export const signAndSendTransactions = (history, metaMetricsEvent) => {
     }
 
     const tradeTxMeta = await dispatch(
-      addUnapprovedTransaction(usedTradeTxParams, 'metamask'),
+      addUnapprovedTransaction(
+        usedTradeTxParams,
+        'metamask',
+        TRANSACTION_TYPES.SWAP,
```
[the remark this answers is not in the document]
jpuri: This file has 2 calls of `addUnapprovedTransaction` both of which I updated.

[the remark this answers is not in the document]
jpuri: fixed

app/scripts/controllers/transactions/index.js
```diff
@@ -337,9 +337,19 @@ export default class TransactionController extends EventEmitter {
...
+    if (
```
dan437, line 344: You could do this instead:
```
const allowedTransactionTypes = [TRANSACTION_TYPES.SWAP, TRANSACTION_TYPES.SWAP_APPROVAL];
if (transactionType !== undefined && !allowedTransactionTypes.includes(transactionType)) {
throw new Error(
    `TransactionController - invalid transactionType value: ${transactionType}`,
  );
}
```

jpuri: Just updated

adonesky1: total nit but `allowedTransactionTypes` feels a bit misleading in the context of this method. Maybe `allowedTransactionTypeParam`. I don't know?

danjm (dismissed):
LGTM

danjm (dismissed):

app/scripts/controllers/transactions/index.js
jpuri: I added a new SWAP_TRANSACTION_TYPES constant.

dan437 (approved):

adonesky1 (approved):
LGTM

PR #13621, opened 2022-02-14
Ensure pre-selected account on connect screen is visible in list

### LINKED ISSUE #8903: Ensure pre-selected account on connect screen is visible in list
Scenario:
- Have many accounts
- Have a later account selected
- Connects to a new site

Problem:
- The pre-selected (checked) account is not visible on the connect screen, and so the user thinks no accounts are selected.

Ideally:
- A pre-selected account will either be shown at the top by default, or the list will be automatically scrolled to show that account.

ryanml:
Fixes: https://github.com/MetaMask/metamask-extension/issues/8903

Test Plan:

1. With Account numbers ranging from 1-15, use our test dapp to attempt to connect your account
2. Ensure that the selected account is show in to view, and there are no clipping issues with the bottom confirm and cancel buttons

https://user-images.githubusercontent.com/8732757/153953789-d594ad23-0b6c-4e18-97f3-782575948edb.mov

### DIFF
--- a/ui/components/ui/account-list/account-list.js
+++ b/ui/components/ui/account-list/account-list.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useLayoutEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import { useI18nContext } from '../../../hooks/useI18nContext';
@@ -20,6 +20,13 @@ const AccountList = ({
   handleAccountClick,
 }) => {
   const t = useI18nContext();
+  const selectedAccountScrollRef = useRef(null);
+
+  useLayoutEffect(() => {
+    if (selectedAccountScrollRef.current) {
+      selectedAccountScrollRef.current.scrollIntoView?.();
+    }
+  }, []);
 
   const Header = () => {
     let checked;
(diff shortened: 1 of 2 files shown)


NiranjanaBinoy (dismissed):

ui/components/ui/account-list/account-list.js
```diff
@@ -20,6 +20,13 @@ const AccountList = ({
   handleAccountClick,
 }) => {
   const t = useI18nContext();
+  const selectedAccountScrollRef = useRef(null);
+
+  useLayoutEffect(() => {
+    if (selectedAccountScrollRef.current) {
+      selectedAccountScrollRef.current.scrollIntoView({ behavior: 'smooth' });
```
darkwing, line 27: Not sure if we have a style preference but we could avoid the `if` and just do:

```
selectedAccountScrollRef.current?.scrollIntoView({ behavior: 'smooth' });
```

jpuri (dismissed):

ui/components/ui/account-list/account-list.js
ryanml: Updated!

darkwing (approved):
This is a good intermediate fix for the hovering over content issue but we desperately need to get the buttons pinned back to the bottom of the window, and make only the list scroll.

jpuri (approved):

PR #13633, opened 2022-02-15
Fixed Add Recipient Stories and migrated to use control args.

### LINKED ISSUE #12931: AddRecipient story broken
**Describe the bug**
AddRecipient story needs to be fixed

1. Go to [Metamask storybook](https://metamask.github.io/metamask-storybook/?path=/story/ui-pages-send-send-content-add-recipient-add-recipient-stories-js--add-recipient-component)
2. Search "Add Recipient" if not already there
3. See error

**Expected behavior**
Add recipient UI should display 

**Screenshots**
Screen Shot 2021-12-01 at 9 53 33 AM

rojotek:
Fixes: #12931

Explanation:  
Fixes the compile error on the story, and migrates it to use control args.

Manual testing steps:  
  - `yarn storybook`
  - open the story here: /?path=/story/ui-pages-send-send-content-add-recipient-add-recipient-stories-js--default-story

### DIFF
--- a/ui/pages/send/send-content/add-recipient/add-recipient.stories.js
+++ b/ui/pages/send/send-content/add-recipient/add-recipient.stories.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import { Provider } from 'react-redux';
-import { text } from '@storybook/addon-knobs';
 
 import configureStore from '../../../../store/store';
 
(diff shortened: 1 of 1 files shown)


ui/pages/send/send-content/add-recipient/add-recipient.stories.js
```diff
@@ -9,31 +8,71 @@ import AddRecipient from './add-recipient.component';
...
 
+const { metamask } = store.getState();
+const { addressBook } = metamask;
+const recipient = metamask.accountArray[0];
+
 export default {
   title: 'Pages/Send/SendContent/AddRecipient',
   id: __filename,
   decorators: [(story) => <Provider store={store}>{story()}</Provider>],
+  argTypes: {
```
georgewrmarshall, line 19: One improvement you could make is to move the args for `ensError` and `ensWarning` here also. I feel like the default story should contain all component prop options.

Ahh so I see the dilemma here it needs to be a translation key. Maybe we could add the argType name and description options this will let the user know it should be a key. Kinda not great it shows up in the name column of the argsTable but I think it's clear not to include the brackets in the prop name...

Screen Shot 2022-02-17 at 9 39 07 AM
Screen Shot 2022-02-17 at 9 39 46 AM

```suggestion
  argTypes: {
    ensError: {
      control: {
        type: 'select',
      },
      options: ['loading'],
      name: 'ensError(must be the key for a translation)',
      description: 'must be the key for a translation',
    },
    ensWarning: {
      control: {
        type: 'select',
      },
      options: ['loading'],
      name: 'ensWarning(must be the key for a translation)',
      description: 'must be the key for a translation',
    },
```

georgewrmarshall (approved):
LGTM! Left one optional improvement

ui/pages/send/send-content/add-recipient/add-recipient.stories.js
digiwand: I like the addition of the descriptions here. We might want to keep in mind that we will likely be undoing some component prop args as class components are refactored into functional components. I am making a mental note for myself to refactor class components to functional components before writing stories when possible moving forward.

digiwand (approved):
I can get on board with adding all component props as an argType (https://github.com/MetaMask/metamask-extension/pull/13633/files#r808531484). (Just learning this myself.) In the future, I see `ensWarning` and `ensError` being removed as a passed component prop when we convert this component to a functional component. 

No blockers. LGTM!

PR #13636, opened 2022-02-15
Drop transaction after 800 retries

jasonzhouu:
Fixes: 
infinite resubmit pending transaction if gas price too low.

Explanation:  
drop transaction after enough retries.


darkwing:
Thank you for the contribution @jasonzhouu !  `yarn lint:fix` should fix the lint issue.  Also, any specific reason for `800`?

jasonzhouu:
> Thank you for the contribution @jasonzhouu ! `yarn lint:fix` should fix the lint issue. Also, any specific reason for `800`?

@darkwing Thank you for your reply. It takes about 4 hours for 800 retries. I think it's long enough to drop transaction.


```
800 retries * 20 polling interval = 16,000 seconds = 4.4 hours
```

PR #13643, opened 2022-02-15
Confirm transaction page: use method name only for contract transactions

jpuri:
Use methodname from transaction data as title on confirm transaction page for only contract transaction.


brad-decker:
@jpuri is there an issue tracking what this resolves?

danjm (approved):

adonesky1 (approved):
LGTM

PR #13646, opened 2022-02-15
update gas fees

segun:
Fixes: #13496

Explanation:  This PR uses a direct method to update transaction gas fees instead of calling the updateTransaction method.

Manual testing steps:  
  - 
  - 
  -


ui/hooks/gasFeeInput/useTransactionFunctions.js
```diff
@@ -87,7 +88,11 @@ export const useTransactionFunctions = ({
         );
         dispatch(updateCustomSwapsEIP1559GasParams(newGasSettings));
       } else {
-        dispatch(updateTransactionFn(updatedTxMeta));
+        const userSettings = {};
+        userSettings.userEditedGasLimit = updatedTxMeta.userEditedGasLimit;
+        userSettings.userFeeLevel = updatedTxMeta.userFeeLevel;
+        dispatch(updateTransactionGasFees(updatedTxMeta.id, newGasSettings));
+        dispatch(updateTransactionUserSettings(updatedTxMeta.id, userSettings));
```
jpuri, line 95: `txMeta` may have more fields, do we ensure that all is saved ?

https://github.com/MetaMask/metamask-extension/pull/13646/files#diff-a37b7544a0b876163dca658b96d70b11e99c7d7c33cb64ef2295c2244343b981R73

app/scripts/controllers/transactions/index.js
```diff
@@ -568,7 +568,6 @@ export default class TransactionController extends EventEmitter {
     if (!this._checkIfTxStatusIsUnapproved(txId)) {
       return;
     }
-
```
darkwing, line 571: This change is unrelated and can probably be reverted.

darkwing (changes requested):

ui/hooks/gasFeeInput/useTransactionFunctions.js
segun: @jpuri The essence of this change is that we don't want to update `all` fields, just the ones that should be updated. In this case the gas settings.

segun: @darkwing can't do that because of [this](https://eslint.org/docs/rules/no-case-declarations)

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
```diff
@@ -187,7 +189,11 @@ export default function EditGasPopover({
         );
         break;
       case EDIT_GAS_MODES.MODIFY_IN_PLACE:
-        dispatch(updateTransaction(updatedTxMeta));
+        userSettings.userEditedGasLimit = updatedTxMeta.userEditedGasLimit;
+        userSettings.userFeeLevel = updatedTxMeta.userFeeLevel;
+
+        dispatch(updateTransactionGasFees(updatedTxMeta.id, newGasSettings));
+        dispatch(updateTransactionUserSettings(updatedTxMeta.id, userSettings));
```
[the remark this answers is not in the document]
segun: @darkwing can't do that because of [this](https://eslint.org/docs/rules/no-case-declarations)

app/scripts/controllers/transactions/index.js
segun: fixed here [4f107a9](https://github.com/MetaMask/metamask-extension/pull/13646/commits/4f107a903a983d58be8f2b905927939894804c4e)

ui/ducks/metamask/metamask.js
```diff
@@ -209,7 +209,7 @@ const toHexWei = (value, expectHexWei) => {
 };
 
 // Action Creators
-export function updateTransactionGasFees({
+export function updateTransaction({
```
brad-decker, line 212: [Q] Why is this function being renamed? I am not a fan of this change because it mimics the naming of a function that previously updated the transaction in state in the back-end... which made me think all of its usages in this were that old function.  And it seems like this is still only updating gas related fields? So why was the name changed?

brad-decker (changes requested):

ui/ducks/metamask/metamask.js
segun: I renamed it to avoid name collision (in this file) with updateTransactionGasFees of txController.

ui/hooks/gasFeeInput/useTransactionFunctions.js
jpuri: @segun : but in this case there are more fields than gas settings in txMeta which will be lost.

jpuri (changes requested):
This change can effect cancel / speed-up flow: https://github.com/MetaMask/metamask-extension/pull/13646/files#r814856524

ui/ducks/metamask/metamask.js
segun: I changed the name to `updateGasFees` [here](https://github.com/MetaMask/metamask-extension/pull/13646/commits/70f994ad10acfb26327d4dd1398c0195e91ab258)

ui/hooks/gasFeeInput/useTransactionFunctions.js
segun: @jpuri I am updating gas fees and userSettingsFees based on 

1. the parameters passed to `useTransactionFunction`
2. the computed values in the method.

ui/hooks/gasFeeInput/useTransactionFunctions.js
```diff
@@ -87,7 +89,18 @@ export const useTransactionFunctions = ({
...
+        userSettings.userEditedGasLimit = updatedTxMeta.userEditedGasLimit;
+        userSettings.userFeeLevel = updatedTxMeta.userFeeLevel;
+
+        if (txMeta && txMeta.previousGas) {
+          dispatch(
+            updatePreviousGasParams(updatedTxMeta.id, txMeta.previousGas),
+          );
+        }
+
+        dispatch(updateTransactionGasFees(updatedTxMeta.id, newGasSettings));
```
danjm, line 102: I think we need to `await` the first dispatch before we call the second

ui/hooks/gasFeeInput/useTransactionFunctions.js
```diff
@@ -87,7 +89,18 @@ export const useTransactionFunctions = ({
         );
         dispatch(updateCustomSwapsEIP1559GasParams(newGasSettings));
       } else {
-        dispatch(updateTransactionFn(updatedTxMeta));
+        const userSettings = {};
+        userSettings.userEditedGasLimit = updatedTxMeta.userEditedGasLimit;
+        userSettings.userFeeLevel = updatedTxMeta.userFeeLevel;
+
+        if (txMeta && txMeta.previousGas) {
+          dispatch(
```
danjm, line 97: We should await this

ui/store/actions.js
```diff
@@ -673,6 +673,70 @@ const updateMetamaskStateFromBackground = () => {
   });
 };
 
+export function updateTransactionGasFees(txId, txGasFees) {
+  return async (dispatch) => {
+    try {
+      await promisifiedBackground.updateTransactionGasFees(txId, txGasFees);
+    } catch (error) {
+      dispatch(txError(error));
+      dispatch(goHome());
```
danjm, line 682: I don't think we need `dispatch(goHome());` anymore.

ui/store/actions.js
```diff
@@ -673,6 +673,70 @@ const updateMetamaskStateFromBackground = () => {
...
+      log.error(error.message);
+      throw error;
+    }
+
+    dispatch(updateTransactionParams(txGasFees.id, txGasFees.txParams));
+    const newState = await updateMetamaskStateFromBackground();
+    dispatch(updateMetamaskState(newState));
+    dispatch(showConfTxPage({ id: txId }));
```
danjm, line 690: We don't need `dispatch(showConfTxPage({ id: txId }));` anymore

ui/hooks/gasFeeInput/useTransactionFunctions.js
jpuri: @danjm : in EIP-1559 V2 we added this field `previousGas` in txMeta. This is used in case of cancel/speedup and saves the gas used in original transaction. This field is used to derive the new estimate type `10% Minimum` for the transaction.

Commit here [fixes](https://github.com/MetaMask/metamask-extension/pull/13646/commits/83652f84fa8a631e6bcabf8a0af2d4ef9d5159a9#diff-675c437262d906486cf5436c0b64535347286cc4d1081dbd863fb45b8a01b81aR386) this issue.

ui/store/actions.js
danjm: same for the other methods as well

ui/hooks/gasFeeInput/useTransactionFunctions.js
segun: Fixed [here](https://github.com/MetaMask/metamask-extension/pull/13646/commits/1669aac0914ec5681b651a1e3a76546f5108bd3e)

ui/hooks/gasFeeInput/useTransactionFunctions.js
segun: Fixed [here](https://github.com/MetaMask/metamask-extension/pull/13646/commits/1669aac0914ec5681b651a1e3a76546f5108bd3e)

app/scripts/controllers/transactions/index.js
```diff
@@ -406,6 +436,17 @@ export default class TransactionController extends EventEmitter {
...
+   * =======
+   * {
+   * gasLimit,
+   * gasPrice,
+   * maxPriorityFeePerGas,
+   * maxFeePerGas,
+   * estimateUsed,
+   * estimateSuggested
+   * }
+   * >>>>>>> 83652f84f (fix:)
```
jpuri, line 449: 🔪

test/e2e/metamask-ui.spec.js
```diff
@@ -279,6 +279,7 @@ describe('MetaMask', function () {
       await gasPriceInput.fill('20');
       await driver.delay(veryLargeDelayMs);
       await driver.clickElement({ text: 'Save', tag: 'button' });
+      await driver.delay(veryLargeDelayMs);
```
segun, line 282: You'll see some of these pattern in e2e tests in this PR where I have to call delay after Save is clicked. This is because it takes some milliseconds between when Save is clicked and when the values are updated and if the next line is executed immediately, it will use the old values.

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
```diff
@@ -187,7 +189,13 @@ export default function EditGasPopover({
         );
         break;
```
jpuri, line 190: Hey @segun : sorry for this delayed feedback, but the current update transaction function results in loading spinner being shown on UI till changes are reflected in app state. Will there be same behaviour with these new update methods.

`userFeeLevel` and `gas values` are kind of related, I think it would be better to update them in single method. `await` below is helping in achieving this. But if we can have a single methods at controller level, that will be even  better.

segun: There was a comment (a long time ago...I will try and find it) that asked that I remove the spinners.

On 2. Yeah, I can have a method that combine both. But again the crux of the PR is to separate these calls. 

@danjm what do you think about combining these 2?

danjm: @segun I think combining them is fine. What's most important is the removal of the `updateTransaction` function, which has the ability to update _every_ parameter. But I think it is fine to combine these

jpuri: Also, for edit gas fee popup, spinner is important, as this spinner shows up on confirm transaction page. It is important for this spinner to be visible till estimates are updated.

danjm: okay, @segun let's re-add the logic for showing the spinner

ui/components/app/edit-gas-popover/edit-gas-popover.component.js
```diff
@@ -187,7 +189,14 @@ export default function EditGasPopover({
...
       case EDIT_GAS_MODES.MODIFY_IN_PLACE:
-        dispatch(updateTransaction(updatedTxMeta));
+        newGasSettings.userEditedGasLimit = updatedTxMeta.userEditedGasLimit;
+        newGasSettings.userFeeLevel = updatedTxMeta.userFeeLevel;
+
+        dispatch(showLoadingIndication());
+        await dispatch(
+          updateTransactionGasFees(updatedTxMeta.id, newGasSettings),
+        );
+        dispatch(hideLoadingIndication());
```
jpuri, line 199: It will be nice to wrap this whole thing 
```
        await dispatch(
          updateTransactionGasFees(updatedTxMeta.id, newGasSettings),
        );
        dispatch(hideLoadingIndication());
```
in hook so that all place in UI which use this function will not have to replicate the logic to show / hide loader.

jpuri (dismissed):

danjm (dismissed):

darkwing (dismissed):

brad-decker (dismissed):

brad-decker (dismissed):

brad-decker (dismissed):

app/scripts/controllers/transactions/index.js
```diff
@@ -378,14 +378,42 @@ export default class TransactionController extends EventEmitter {
    * updates the params that are editible in the send edit flow
    *
    * @param {string} txId - transaction id
-   * @param {object} editableParams - holds the editable parameters
+   * @param {object} previousGasParams - holds the parameter to update
+   * @param {string} previousGasParams.maxFeePerGas
+   * @param {string} previousGasParams.maxPriorityFeePerGas
+   * @param {string} previousGasParams.gasLimit
+   */
+  updatePreviousGasParams(
```
danjm, line 386: This method should be updated to reflect the changes I made in the other methods: returning the full updated txMeta

ui/store/actions.js
```diff
@@ -671,6 +671,23 @@ const updateMetamaskStateFromBackground = () => {
   });
 };
 
+export function updatePreviousGasParams(txId, previousGasParams) {
```
danjm, line 674: This should be updated to reflect the changes I made in the other method, and return the updated transaction

danjm (commented):
LGTM

darkwing (approved):
Make sense to me.  Might be good to get @PeterYinusa/QA's thoughts on the `driver.delay` calls.  Nice job @segun !

brad-decker (approved):

jpuri (approved):

test/e2e/metamask-ui.spec.js
PeterYinusa: In future cases, where possible we can use the following method to wait for an expected condition (e.g. updated gas estimate), instead of adding a delay

```
	css: 'GAS_ESTIMATE_CSS_SELECTOR',
	text: 'GAS_ESTIMATE_UPDATED_VALUE',
});
```

PR #13652, opened 2022-02-16
Update EIP1559 params.

segun:
Fixes: #13496

Explanation:  This PR uses a direct method to update eip1559 params  instead of calling the updateTransaction method.

Manual testing steps:  
  - 
  - 
  -


app/scripts/controllers/transactions/index.js
```diff
@@ -336,6 +337,268 @@ export default class TransactionController extends EventEmitter {
...
+   * @param {string} txId - transaction id
+   * @param {object} editableParams - holds the eip1559 fees parameters
+   * @param editableParams.data
+   * @param editableParams.from
+   * @param editableParams.to
+   * @param editableParams.value
+   * @param editableParams.gas
+   * @param editableParams.gasPrice
+   */
+  updateEditableParams(txId, { data, from, to, value, gas, gasPrice }) {
```
brad-decker, line 376: I know we agreed upon this name in another thread. If we feel good about this I 'm not gonna block on it. However, all params are "editable"

Also gasPrice is not a part of this when we have explicit gas update methods? I'm confused.

app/scripts/controllers/transactions/index.js
```diff
@@ -336,6 +337,268 @@ export default class TransactionController extends EventEmitter {
...
+   * @param {string} txId - transaction id
+   * @param {object} txGasFees - holds the gas fees parameters
+   * {
+   * gasLimit,
+   * gasPrice,
+   * maxPriorityFeePerGas,
+   * maxFeePerGas,
+   * estimateUsed,
+   * estimateSuggested
+   * }
```
brad-decker, line 410: I am unfamiliar with this syntax for jsdocs can you enlighten me on its effect?

app/scripts/controllers/transactions/index.js
```diff
@@ -336,6 +337,268 @@ export default class TransactionController extends EventEmitter {
...
+   * @param txGasFees.gasLimit
+   * @param txGasFees.gasPrice
+   * @param txGasFees.maxPriorityFeePerGas
+   * @param txGasFees.maxFeePerGas
+   * @param txGasFees.estimateUsed
+   * @param txGasFees.estimateSuggested
+   * @param txGasFees.defaultGasEstimates
+   * @param txGasFees.gas
+   * @param txGasFees.originalGasEstimate
```
brad-decker, line 419: All of these docs are missing types.

brad-decker (changes requested):
@segun a number of your prs seem to have unrelated changes in them, likely cause they were pulled off from the same branch... but its making it difficult to review in full. I've left some notes on a few prs related to the transaction update changes ... lemme know if any of them are just a result of splitting off the work. Sorry man!

app/scripts/controllers/transactions/index.js
segun: I added the gasPrice because of [this comment](https://github.com/MetaMask/metamask-extension/pull/13496#discussion_r812346125).

Thinking about it, now I can split this into two calls. Where the second call will use the dedicated gasUpdate method.

app/scripts/controllers/transactions/index.js
segun: No I don't think this is specifically JSDoc. I started by just trying to list the parameters the methods take and linter formats it out like that.

Since I have proper JSDoc @param comments now, I think I can remove these.

app/scripts/controllers/transactions/index.js
segun: Fixed [here] (https://github.com/MetaMask/metamask-extension/pull/13652/commits/6d30f36f8a2657cc5dda74da3d17ea8c1890547f)

app/scripts/controllers/transactions/index.js
segun: Fixed [here](https://github.com/MetaMask/metamask-extension/commit/6d30f36f8a2657cc5dda74da3d17ea8c1890547f)

app/scripts/controllers/transactions/index.js
segun: Fixed [here](https://github.com/MetaMask/metamask-extension/pull/13652/commits/c0c75ea0e1df26d077ce5b1186d2fc8462b79d9d)

brad-decker (approved):

jpuri (approved):
👍

PR #13655, opened 2022-02-16
[UPDATE] NFT Auto-detect copy change to align with mobile

sethkfman:
Explanation:  Updated copy to align with improved mobile copy for NFT auto-detect

Related: https://github.com/MetaMask/metamask-mobile/pull/3762

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -980,7 +980,7 @@
     "message": "Enable OpenSea API"
   },
   "enableOpenSeaAPIDescription": {
-    "message": "Use OpenSea's API to fetch NFT data. NFT auto-detection relies on OpenSea's API, and will not be available when this is turned off."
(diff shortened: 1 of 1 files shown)


darkwing (approved):

PeterYinusa (approved):

adonesky1 (changes requested):
@sethkfman we already [have language similar to this](https://github.com/MetaMask/metamask-extension/blob/fa467d1b8cdce0e5690497baf1c2568df0385e1a/app/_locales/en/messages.json#L3512) under the `AutoDetect NFTs` preferences. In this PR you're adding this description about the risks of auto-detection to the preference that is only about using OpenSea's API. Perhaps we want to tweak the AutoDetect NFT's language but I think it should be under that preference and not under Enable OpenSea's API

brad-decker:
@sethkfman is this something we should close or do you still see a need for this change?

brad-decker:
Closing, if this is something we still want we can reopen the PR and update it. Thanks @sethkfman and @adonesky1

PR #13663, opened 2022-02-17
Change "Quote X of Y" to "Fetching quote X of Y"

dan437:
## Explanation
Sometimes the total number of quotes displayed on the loading page doesn't match what is on the View Quote page. Using the word `Fetching` suggests that work is being done in the background, without making any indirect and unintentional promises.

## Testing Steps
- Open swaps
- Select token from and to, amount and quickly submit it
- You should see a loading page with updated content

## Screenshots
image


darkwing (approved):
I love this!  Small details like this make the experience better!

PeterYinusa (approved):

NiranjanaBinoy (approved):

PR #13664, opened 2022-02-17
Fix hard code account menu and settings page buttons style 

egorlem:
Explanation:  Buttons did't adapt to text size in different languages. This PR fixes the styles of it 

**Production**

Снимок экрана 2022-02-17 в 17 04 22

**Fix**

Снимок экрана 2022-02-17 в 17 02 23


guesswhozzz:

brad-decker (changes requested):

darkwing (approved):
Thank you for the contribution, @guesswhozzz !  I was testing this wife a few changes:

https://gist.github.com/darkwing/236cc6765a5c0b6417b2dabb45ea64f0

I don't think the `<div>` with `width: auto` is necessary .

darkwing (changes requested):
Thank you for the contribution, @guesswhozzz !  I was testing this wife a few changes:

https://gist.github.com/darkwing/236cc6765a5c0b6417b2dabb45ea64f0

I don't think the `<div>` with `width: auto` is necessary .

guesswhozzz:
> Thank you for the contribution, @guesswhozzz ! I was testing this wife a few changes:
> 
> https://gist.github.com/darkwing/236cc6765a5c0b6417b2dabb45ea64f0
> 
> I don't think the `<div>` with `width: auto` is necessary . 

Oh, I agree the `with: auto` property on a bare div element is weird. I removed it. I also removed wrapping the title in a <div>. I noticed a BEM notation in your layout and I still think even if the element does not have styles, it still needs to be wrapped in a tag so as't to violate BEM.  

Thank you for your response

PR #13668, opened 2022-02-17
add e2e test for deterministic account address generation

adonesky1:
Addresses one bullet point on https://github.com/MetaMask/pm-security/issues/38

### DIFF
--- a/test/e2e/tests/add-account.spec.js
+++ b/test/e2e/tests/add-account.spec.js
@@ -3,12 +3,14 @@ const {
   convertToHexValue,
   withFixtures,
   regularDelayMs,
+  completeImportSRPOnboardingFlow,
 } = require('../helpers');
 const enLocaleMessages = require('../../../app/_locales/en/messages.json');
 
 describe('Add account', function () {
   const testSeedPhrase =
     'forum vessel pink push lonely enact gentle tail admit parrot grunt dress';
+  const testPassword = 'correct horse battery staple';
   const ganacheOptions = {
     accounts: [
       {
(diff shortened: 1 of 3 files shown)


test/e2e/tests/add-account.spec.js
```diff
@@ -36,4 +44,216 @@ describe('Add account', function () {
       },
     );
   });
+
```
jpuri, line 47: test description looks confusing `should add the same account addresses if added and then the same mnemonic is imported`

test/e2e/tests/add-account.spec.js
```diff
@@ -36,4 +44,216 @@ describe('Add account', function () {
...
+    await withFixtures(
+      {
+        fixtures: 'onboarding',
+        ganacheOptions,
+        title: this.test.title,
+        failOnConsoleError: false,
+      },
+      async ({ driver }) => {
+        await driver.navigate();
+
```
jpuri, line 58: It might be useful to extract onboarding part into separate method that can be reused in other account / onboarding tests also.

test/e2e/tests/add-account.spec.js
adonesky1: maybe something like: `'should add the same account addresses when a secret recovery phrase is imported, the account is locked, and the same secret recovery phrase is imported again'`?

test/e2e/tests/add-account.spec.js
adonesky1: done here: https://github.com/MetaMask/metamask-extension/pull/13668/commits/578c813a526cd61d974e706cc3839bcee7ef1c12

darkwing (approved):

jpuri (approved):

PR #13669, opened 2022-02-17
Deleting transactions from `currentNetworkTxnList` based on unique address along with nonce and chainId

### LINKED ISSUE #13550: Make transaction deletion logic in tx-state-manager account for unique addresses when deciding which txes to preserve
Currently, we have the following logic in the tx state manager for deleting transactions from state upon the addition of a new transaction:

```
    // checks if the length of the tx history is longer then desired persistence
    // limit and then if it is removes the oldest confirmed or rejected tx.
    // Pending or unapproved transactions will not be removed by this
    // operation. For safety of presenting a fully functional transaction UI
    // representation, this function will not break apart transactions with the
    // same nonce, per network. Not accounting for transactions of the same
    // nonce and network combo can result in confusing or broken experiences
    // in the UI.
    //
    // TODO: we are already limiting what we send to the UI, and in the future
    // we will send UI only collected groups of transactions *per page* so at
    // some point in the future, this persistence limit can be adjusted. When
    // we do that I think we should figure out a better storage solution for
    // transaction history entries.
    const nonceNetworkSet = new Set();
    const txsToDelete = transactions
      .reverse()
      .filter((tx) => {
        const { nonce } = tx.txParams;
        const { chainId, metamaskNetworkId, status } = tx;
        const key = `${nonce}-${chainId ?? metamaskNetworkId}`;
        if (nonceNetworkSet.has(key)) {
          return false;
        } else if (
          nonceNetworkSet.size < txHistoryLimit - 1 ||
          getFinalStates().includes(status) === false
        ) {
          nonceNetworkSet.add(key);
          return false;
        }
        return true;
      })
      .map((tx) => tx.id);

    this._deleteTransactions(txsToDelete);
```

This logic will prevent a transaction from being deleted if it shares both a nonce and a chainId with another transaction. This prevents many transactions from being deleted from state. If a user creates transactions on two or more accounts, and any two transactions from different accounts share nonces, then those transactions will never be deleted. This was never intended and there is no good reason for it. If we further distinguish transactions within the deletion logic by account address, then more individual transactions will be preserved for longer in state, while transactions of different accounts that share nonces won't be unnecessarily preserved in state forever.

NiranjanaBinoy:
Fixes: #13550 

Explanation:  
Adding `from` address will prevent a transaction from being deleted if it shares the same nonce and chainId with another transaction. Suppose a user creates transactions on two or more accounts, and any two transactions from different accounts share nonces. In that case, the old transaction stays longer in our state unnecessarily until the new old reaches the end of the list. This was never intended, and there is no good reason for it. By adding the `from` account address, more individual transactions will be preserved for longer in the state, while transactions of different accounts that share nonces won't be unnecessarily preserved in the state forever.

### DIFF
--- a/app/scripts/controllers/transactions/tx-state-manager.js
+++ b/app/scripts/controllers/transactions/tx-state-manager.js
@@ -245,9 +245,9 @@ export default class TransactionStateManager extends EventEmitter {
     const txsToDelete = transactions
       .reverse()
       .filter((tx) => {
-        const { nonce } = tx.txParams;
+        const { nonce, from } = tx.txParams;
         const { chainId, metamaskNetworkId, status } = tx;
-        const key = `${nonce}-${chainId ?? metamaskNetworkId}`;
+        const key = `${nonce}-${chainId ?? metamaskNetworkId}-${from}`;
         if (nonceNetworkSet.has(key)) {
           return false;
         } else if (


jpuri (approved):

jpuri:
Change look safe, but I am curious what can be the scenario when we get same nonce again ?

NiranjanaBinoy:
> Change look safe, but I am curious what can be the scenario when we get same nonce again ?

@jpuri I believe chances of getting the same nonce again from the same network and same account is rare to none. If the same nonce is generated from a different account or different network the current change will be able to generate a unique key.

brad-decker (approved):

PR #13673, opened 2022-02-18
Add test: Import Account using private key of an already active account should result in an error

adonesky1:
Fixes: One test coverage improvement raised [here](https://github.com/MetaMask/pm-security/issues/38):
> It should not be possible to import an existing account

### DIFF
--- a/test/e2e/tests/from-import-ui.spec.js
+++ b/test/e2e/tests/from-import-ui.spec.js
@@ -293,6 +293,47 @@ describe('Metamask Import UI', function () {
       },
     );
   });
+
+  it('Import Account using private key of an already active account should result in an error', async function () {
+    const testPrivateKey =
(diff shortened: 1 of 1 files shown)


test/e2e/tests/from-import-ui.spec.js
```diff
@@ -293,6 +293,48 @@ describe('Metamask Import UI', function () {
...
+        await driver.clickElement({ text: 'Import Account', tag: 'div' });
+
+        // enter private key',
+        await driver.fill('#private-key-box', testPrivateKey1);
+        await driver.clickElement({ text: 'Import', tag: 'button' });
+
+        // error should occur
+        await driver.waitForSelector({
+          css: '.error',
+          text: "The account you're are trying to import is a dlicate",
```
darkwing, line 332: Is this supposed to be `duplicate`?

jpuri (approved):

ryanml (approved):

PR #13682, opened 2022-02-18
add account_type and device_model to tx metrics

### LINKED ISSUE #13357: Add wallet_type parameter to transaction MetaMetrics events
We would like to be able to segment transactions by the wallet's type. This way we can see percentage of all transactions that are issued from a ledger, etc. To do this we need to modify the transaction controller event payload to include a 'wallet_type' parameter in properties. This wallet_type parameter should be populated with one of the following types based on the address of the from parameter ( we cannot assume that the transaction in question belongs to the currently selected wallet )

1. Ledger
2. Trezor
3. Lattice1
4. QR
5. Imported
6. ? -- 

Note that it might be best to record this value on the txMeta object, which would prevent having to get this value multiple times throughout the transaction lifecycle.

brad-decker:
Fixes: #13357 

Explanation:  adds account_type and device_model to the metrics payload. 

Manual testing steps:  
  - set the environment variables for segment to point to development segment source
  - test with each type of account sending a transaction
  - check segment debugger to confirm that tx properties includes account_type and device_model

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -700,6 +700,8 @@ export default class MetamaskController extends EventEmitter {
       getExternalPendingTransactions: this.getExternalPendingTransactions.bind(
         this,
       ),
+      getAccountType: this.getAccountType.bind(this),
+      getDeviceModel: this.getDeviceModel.bind(this),
     });
     this.txController.on('newUnapprovedTx', () => opts.showUserConfirmation());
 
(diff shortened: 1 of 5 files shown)


app/scripts/metamask-controller.js
```diff
@@ -2125,6 +2127,54 @@ export default class MetamaskController extends EventEmitter {
...
+        // TODO: get model after ledger keyring exposes method
+        return DEVICE_NAMES.LEDGER;
+      case KEYRING_TYPES.LATTICE:
+        // TODO: get model after lattice keyring exposes method
+        return DEVICE_NAMES.LATTICE;
+      default:
+        return 'N/A';
+    }
+  }
+
```
jpuri, line 2177: 👍

darkwing (approved):

ryanml (approved):
lgtm!

PR #13705, opened 2022-02-21
Fixes in signTypedData domain header

### LINKED ISSUE #13703: Make signTypedData domain headers more consistent/less confusing
In the OpenSea migration to the Wyvern contracts, they've switched to using EIP-712 signTypedData, which improves signature readability, but the current header is confusing to their users, because it shows info about the contract instead of their site (like their users are used to).

image

This has led to [OpenSea requesting a special header](https://github.com/MetaMask/metamask-extension/pull/13685) on EIP-712 signatures when requested from their site. A hard-coded special case is a huge code smell imo, and we should be looking at the root of the problem and addressing it in a way that alleviates this future problem for other sites as well.

## Immediate, consumer-side mitigation

First off, @Gudahtt notes that the contract name shown here is actually already dapp-provided, and is part of the `domain separator` part of the signature request [as shown here](https://eips.ethereum.org/EIPS/eip-712#definition-of-domainseparator). This field is not part of the signed payload, so OpenSea can change the name today with no extra issues.

## More fundamental issue

Why did this suddenly surprise them? Probably because EIP-712 signatures currently have a header that is inconsistent with our other signature views. Those views currently show data purely reflecting the requesting site (the page `title`, and the page metadata defined preview image). So that means `personal_sign` (what they were previously using) and `signTypedData` (what they are now using) are both using different ways of rendering data provided by the dapp, and so the simplest fix would be to simply change the `signTypedData` header to behave the same way the `personal_sign` header works, to totally side-step the inconsistency confusion.

## Longer term questions

Is it safe to show arbitrary data from an untrustworthy dapp? Well, it doesn't help users recognize a phishing site, that's for sure, and we have work to do in that direction, but I think we can treat it as a separate issue for now. This issue is just about header consistency, and then we can talk about improving header safety.

## Acceptance criteria

- EIP-712 signTypedData signatures should use the same header rendering logic as `personal_sign`.
- These changes should also be reflected on mobile.

jpuri:
Fixes: #13703

### DIFF
--- a/ui/components/app/signature-request/signature-request.component.js
+++ b/ui/components/app/signature-request/signature-request.component.js
@@ -76,7 +76,7 @@ export default class SignatureRequest extends PureComponent {
       type,
     } = txData;
     const { address: fromAddress } = fromAccount;
-    const { message, domain, primaryType, types } = JSON.parse(data);
+    const { message, domain = {}, primaryType, types } = JSON.parse(data);
     const { metricsEvent } = this.context;
     const targetSubjectMetadata = origin ? subjectMetadata?.[origin] : null;
 
--- a/ui/components/app/signature-request/signature-request.component.js
+++ b/ui/components/app/signature-request/signature-request.component.js
@@ -1,6 +1,5 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
-import Identicon from '../../ui/identicon';
 import SiteIcon from '../../ui/site-icon';
 import LedgerInstructionField from '../ledger-instruction-field';
 import { getURLHostName, sanitizeMessage } from '../../../helpers/utils/util';
@@ -122,23 +121,23 @@ export default class SignatureRequest extends PureComponent {
             {this.context.t('sigRequest')}
           </div>
           <div className="signature-request-content__identicon-container">
-            {targetSubjectMetadata?.iconUrl ? (
-              <SiteIcon
-                icon={targetSubjectMetadata.iconUrl}
-                name={
-                  getURLHostName(targetSubjectMetadata.origin) ||
-                  targetSubjectMetadata.origin
-                }
-                size={24}
-              />
-            ) : (
-              <div className="signature-request-content__identicon-initial">
-                {/* {domain.name && domain.name[0]} */}
-                {originHostNameCharacter}
-              </div>
-            )}
-            <div className="signature-request-content__identicon-border" />
-            <Identicon address={fromAddress} diameter={70} />
+            <div className="signature-request-content__identicon-border">
+              {targetSubjectMetadata?.iconUrl ? (
+                <SiteIcon
+                  icon={targetSubjectMetadata.iconUrl}
+                  name={
+                    getURLHostName(targetSubjectMetadata.origin) ||
+                    targetSubjectMetadata.origin
+                  }
+                  size={32}
+                />
+              ) : (
+                <div className="signature-request-content__identicon-initial">
+                  {/* {domain.name && domain.name[0]} */}
+                  {originHostNameCharacter}
+                </div>
+              )}
+            </div>
           </div>
           <div className="signature-request-content__info--bolded">
             {origin}
--- a/ui/components/app/signature-request/signature-request.component.js
+++ b/ui/components/app/signature-request/signature-request.component.js
@@ -75,7 +75,7 @@ export default class SignatureRequest extends PureComponent {
       type,
     } = txData;
     const { address: fromAddress } = fromAccount;
-    const { message, domain = {}, primaryType, types } = JSON.parse(data);
+    const { message, primaryType, types } = JSON.parse(data);
     const { metricsEvent } = this.context;
     const targetSubjectMetadata = origin ? subjectMetadata?.[origin] : null;
 
@@ -153,7 +153,7 @@ export default class SignatureRequest extends PureComponent {
           </div>
         ) : null}
         <Message
-          data={{ ...sanitizeMessage(message, primaryType, types), domain }}
+          data={sanitizeMessage(message, primaryType, types)}
           onMessageScrolled={() => this.setState({ hasScrolledMessage: true })}
           setMessageRootRef={this.setMessageRootRef.bind(this)}
           messageRootRef={this.messageRootRef}
--- a/ui/components/app/signature-request/signature-request.component.js
+++ b/ui/components/app/signature-request/signature-request.component.js
@@ -76,11 +76,9 @@ export default class SignatureRequest extends PureComponent {
       type,
     } = txData;
     const { address: fromAddress } = fromAccount;
-    const { message, primaryType, types } = JSON.parse(data);
+    const { message, domain, primaryType, types } = JSON.parse(data);
     const { metricsEvent } = this.context;
-    const targetSubjectMetadata = txData.msgParams.origin
-      ? subjectMetadata?.[txData.msgParams.origin]
-      : null;
+    const targetSubjectMetadata = origin ? subjectMetadata?.[origin] : null;
 
     const onSign = (event) => {
       sign(event);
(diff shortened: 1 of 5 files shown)


ui/components/app/signature-request/signature-request.component.js
```diff
@@ -2,7 +2,7 @@ import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
```
jpuri, line 2: This is place in code where we use this component for v3, v4 requests https://github.com/MetaMask/metamask-extension/blob/develop/ui/pages/confirm-transaction/conf-tx.js#L121

ui/components/app/signature-request/signature-request.component.js
```diff
@@ -117,15 +118,16 @@ export default class SignatureRequest extends PureComponent {
           </div>
           <div className="signature-request-content__identicon-container">
             <div className="signature-request-content__identicon-initial">
-              {domain.name && domain.name[0]}
+              {/* {domain.name && domain.name[0]} */}
+              {originHostNameCharacter}
```
jpuri, line 122: In old component `subjectMetadata` is used to get icon site, https://github.com/MetaMask/metamask-extension/blob/develop/ui/components/app/signature-request-original/signature-request-original.component.js#L139

Somehow, for v3/v4 requests `subjectMetadata` is not available in state.

danfinlay: Sounds like we have to ensure this method is passing subjectMetadata with its invocation. This might be in the inpage script.

jpuri: I was playing around with this fix and it looked like this on a site. May be we should remove icon:

Screenshot 2022-02-22 at 1 16 12 AM

jpuri: I could not find this validation but I have changed code in this component to use icon from `subjectMetadata` is available.

I am not sure if I should also use `subjectMetadata.name` over `origin` if available ?

Gudahtt: We stopped using `subjectMetadata.name` in favor of just using the `origin` because the name can be set to anything (so it can't be trusted), and because in practice it was typically set to something that didn't make for a good title (because we get it from the site's title, and page titles are apparently often full of junk like ads, notifications, etc.).

ui/components/app/signature-request/signature-request.component.js
```diff
@@ -62,18 +64,23 @@ export default class SignatureRequest extends PureComponent {
...
       isLedgerWallet,
       hardwareWalletRequiresConnection,
+      subjectMetadata,
     } = this.props;
+    const {
+      msgParams: { data, origin, version },
+      type,
+    } = txData;
     const { address: fromAddress } = fromAccount;
-    const { message, domain = {}, primaryType, types } = JSON.parse(data);
```
Gudahtt, line 75: Huh. I had not realized that the domain information isn't present anywhere :thinking: It is effectively a part of the message payload, so I would expect it to be rendered there. But it is not.

Maybe we should pass `domain` along to the `Message` component to be displayed alongside `data`.

ui/components/app/signature-request/signature-request.component.js
```diff
@@ -116,16 +124,30 @@ export default class SignatureRequest extends PureComponent {
...
           </div>
           <div className="signature-request-content__identicon-container">
-            <div className="signature-request-content__identicon-initial">
-              {domain.name && domain.name[0]}
-            </div>
-            <div className="signature-request-content__identicon-border" />
-            <Identicon address={fromAddress} diameter={70} />
+            {targetSubjectMetadata?.iconUrl ? (
+              <SiteIcon
+                icon={targetSubjectMetadata.iconUrl}
```
archana-s, line 129: @jpuri Where do you get the subjectMetadata from? Is it from the contract-metadata [here](https://github.com/MetaMask/contract-metadata)?

archana-s: I'm trying to see where I should I add OpenSea's logo for the contract id and any other associated info.

Gudahtt: It is populated by our inpage provider. You can see how the image is found here: https://github.com/MetaMask/providers/blob/main/src/siteMetadata.ts#L78

archana-s: So this is the favicon of the origin site.
This poses a security risk: If a malicious site adds OpenSea logo as their favicon, this would look legitimate upon signing.

archana-s: Is it possible to get this from the contract-metadata.json? May not be an ideal fit for OpenSea, but worth a try? 

Or could we just special case for OS for the short term. we can work together on a longer term fix.

archana-s (changes requested):

ui/components/app/signature-request/signature-request.component.js
archana-s: Circling back after conversations on slack. This looks good to us for the short term.

archana-s (dismissed):

jpuri:
@holantonela : a small design concern in this task. If user provides icon we show it like this:
Screenshot 2022-02-22 at 3 05 09 PM
This icon is taken from header of user's page, if I increase size of the icon it looks broken.

If user does not provide icon we show first letter of origin capitalised as this: 

Screenshot 2022-02-22 at 2 57 26 PM
This is how it is currently showing W [here](https://github.com/MetaMask/metamask-extension/issues/13703#issue-1146052114).

Can you plz confirm if this looks good ?

jpuri:
@holantonela , @Gudahtt : I have added domain details to bottom of message:
Screenshot 2022-02-22 at 3 22 07 PM

ui/components/app/signature-request/signature-request.component.js
```diff
@@ -116,16 +122,28 @@ export default class SignatureRequest extends PureComponent {
...
+                size={24}
+              />
+            ) : (
+              <div className="signature-request-content__identicon-initial">
+                {/* {domain.name && domain.name[0]} */}
+                {originHostNameCharacter}
+              </div>
+            )}
             <div className="signature-request-content__identicon-border" />
             <Identicon address={fromAddress} diameter={70} />
```
Gudahtt, line 141: I don't think this identicon should be here. This represents the signing account, but it's not being shown in a way that communicates to the user which account is being used to sign. The signing account is already being shown in the top-left corner (with both the identicon and name)

Gudahtt:
I was expecting this to look more like our old connect screen designs (which this was based on). e.g. like this: https://www.figma.com/file/hm2RQXkH2QVRJnnTnArLM7/Confirmations?node-id=129%3A467

archana-s:
@jpuri Based on the screenshots posted above, do we know why the code is not pulling the favicon for opensea?

jpuri:
I made changes to icon, this is how it will look if there is an icon:
Screenshot 2022-02-22 at 11 36 43 PM

If site has no icon, first letter of origin is show in blue:
Screenshot 2022-02-22 at 11 25 52 PM

ui/components/app/signature-request/signature-request.component.js
```diff
@@ -136,7 +153,7 @@ export default class SignatureRequest extends PureComponent {
           </div>
         ) : null}
         <Message
-          data={sanitizeMessage(message, primaryType, types)}
+          data={{ ...sanitizeMessage(message, primaryType, types), domain }}
```
Gudahtt, line 156: This seems dangerous. `domain` is a valid field for the message itself, and this would override that field (if it existed).

This also would display the domain separator data as part of the message, which it is not.

When I suggested displaying the domain information as part of the `Message`, I meant that it could be passed as a separate prop and displayed separately (_somehow_ - I don't know how we should display it).

jpuri: domain is not displayed in old component even

Gudahtt: I had brought it up here because previously the `domain.name` was shown where the site icon is now shown.

Overall this is a step forward because the way we were displaying `domain.name` was misleading, but there is a sense in which this is a regression because now we're not showing it anywhere. So if a user was looking to verify the `domain.name` of the message, that would be impossible now.

But yes, maybe you are right, we can deal with it in a separate PR. Hiding it seems less bad than displaying it in a misleading way.

ui/components/app/signature-request/signature-request.component.js
jpuri: Fixed

PR #13717, opened 2022-02-22
Fix issue where token details page unnecessarily relies on send asset state

adonesky1:
Fixes: [This issue](https://docs.google.com/document/d/1OvpkZbvRet33XQVplIc32zlm-6PgSasJcj1XFv9kh9s/edit?disco=AAAAU7k9FMs) in the 10.11.0 RC

Explanation:  In its initial implementation the TokenDetails page unnecessarily uses the `sendAsset` state in the `send duck` to determine which asset to show details for. This change simply pushes the `tokenAddress` to use to fetch the token state for this page in the history object when the user navigates to this page.

Manual testing steps:  
- Switch to Rinkeby
- Complete a swap, so you have a mainnet token add to your token list
- Select the asset
- Select the three dots, then click Token details
- The token details page should show correctly

### DIFF
--- a/ui/pages/asset/components/token-asset.js
+++ b/ui/pages/asset/components/token-asset.js
@@ -17,8 +17,6 @@ import {
 import { getURLHostName } from '../../../helpers/utils/util';
 import { showModal } from '../../../store/actions';
 import { useNewMetricEvent } from '../../../hooks/useMetricEvent';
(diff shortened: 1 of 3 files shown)


ryanml (approved):
LGTM!

VSaric (approved):
LGTM!

PR #13726, opened 2022-02-23
Fix token details icon when 'Use Token Detection' is enabled

ryanml:
Fixes an issue found during the QA of `v10.11.0`

When searching the `tokenList` for `tokenMetadata`, a case insensitive address search should be performed. The behavior of the icon on the token details page is now consistent whether or not "Use Token Detection" is enabled

### DIFF
--- a/ui/pages/token-details/token-details-page.js
+++ b/ui/pages/token-details/token-details-page.js
@@ -35,8 +35,9 @@ export default function TokenDetailsPage() {
   const useTokenDetection = useSelector(getUseTokenDetection);
 
   const tokenAddress = history?.location?.state?.tokenAddress;
-
-  const tokenMetadata = tokenList?.[tokenAddress];
+  const tokenMetadata = Object.values(tokenList).find((token) =>
+    isEqualCaseInsensitive(token.address, tokenAddress),
+  );
   const fileName = tokenMetadata?.iconUrl;
   const imagePath = useTokenDetection
     ? fileName


ui/pages/token-details/token-details-page.js
```diff
@@ -42,7 +42,9 @@ export default function TokenDetailsPage() {
 
   const { asset: tokenAddress } = assetAddress;
 
-  const tokenMetadata = tokenList[tokenAddress];
+  const tokenMetadata = Object.values(tokenList).find((token) =>
```
ryanml, line 45: Actually, I think it's safe here to also do:

```
const tokenMetadata = useTokenDetection ? tokenAddress.toLowerCase() : tokenAddress;
```

ryanml, line 45: I think it may be safe to alternatively do:

```
const tokenMetadataKey = useTokenDetection ? tokenAddress.toLowerCase() : tokenAddress;
const tokenMetadata = tokenList[tokenMetadataKey];
```

darkwing (dismissed):

danjm (dismissed):

digiwand (dismissed):

digiwand (dismissed):

jpuri (dismissed):

ui/pages/token-details/token-details-page.js
digiwand: if we knew for certain that `tokenList` only had lowercase addresses, and if we are not concerned about `useTokenDetection` for this logic, then I think we could use 
```
const tokenMetadata = tokenList[tokenAddress?.toLowerCase()];
```

This said I think the implementation you created in this PR is more robust since it doesn't involve assuming, or ensuring, `tokenList` addresses are also lowercase.

digiwand (approved):

adonesky1 (approved):
LGTM

Issue #13422, opened 2022-01-27
Create utility method for determining if a transaction is sending base asset for current network


brad-decker:
### Description
in #13420 and #13423 it will need to be known if the transaction in question is dealing in the base asset for the network. This is easy to determine by examining `txParams`, and the code for doing so exists in the transaction controller as `_determineTransactionType`. If the result of that method is SIMPLE_SEND then we know we are dealing with the base asset.

#### Requirements/Suggested Implementation details
1. Extract the _determineTransactionType code to a helper method that can be used in either the IncomingTransactionController or TransactionController. Do not put this in shared as I believe it shouldn't be used on frontend.
2. The new function takes a txMeta object as input.
3. Function will check if the type parameter is already set, if it is it'll check if its a SIMPLE_SEND and return boolean based on that 
4. If not it'll call the new helper method with txMeta.txParams to get the type prior to checking it against SIMPLE_SEND

PR #13737, opened 2022-02-23
extract determineTransactionType from tx controller

brad-decker:
partially addresses #13422 

- extracts the _determineTransactionType class method from transaction controller. It will additionally be used in the incomingTransactionController too
- no functional changes
- issue states that not putting the method in shared is a requirement, however after careful consideration and confirming that EthQuery, Ethers, etc are already in the frontend bundle this isn't an issue.
- Creates a new string-utils global helper and moves the isEqualCaseInsensitive method to it.


shared/modules/transaction.utils.js
```diff
@@ -61,3 +81,53 @@ export function txParamsAreDappSuggested(transaction) {
...
+      isContractAddress,
+    } = await readAddressAsContract(query, to);
+
+    contractCode = resultCode;
+    result = isContractAddress
+      ? TRANSACTION_TYPES.CONTRACT_INTERACTION
+      : TRANSACTION_TYPES.SIMPLE_SEND;
+  }
+
+  return { type: result, getCodeResponse: contractCode };
```
darkwing, line 132: I know `getCodeResponse` was used in the previous code but it seems like a weird property name, unless it's a function.

darkwing (approved):

shared/modules/transaction.utils.js
```diff
@@ -61,3 +81,53 @@ export function txParamsAreDappSuggested(transaction) {
...
+  try {
+    name = data && hstInterface.parseTransaction({ data }).name;
+  } catch (error) {
+    log.debug('Failed to parse transaction data.', error, data);
+  }
+
+  const tokenMethodName = [
+    TRANSACTION_TYPES.TOKEN_METHOD_APPROVE,
+    TRANSACTION_TYPES.TOKEN_METHOD_TRANSFER,
+    TRANSACTION_TYPES.TOKEN_METHOD_TRANSFER_FROM,
```
adonesky1, line 108: also not something we need to do in this PR but I just created this [issue](https://github.com/MetaMask/metamask-extension/issues/13855) and we should consider adding `safeTransferFrom` here soon.

brad-decker: Yeah, I don't want to make this about anything other than extracting this method. A future PR should be launched to change this behavior.

shared/modules/transaction.utils.js
```diff
@@ -1,4 +1,24 @@
 import { isHexString } from 'ethereumjs-util';
+import { ethers } from 'ethers';
+import abi from 'human-standard-token-abi';
```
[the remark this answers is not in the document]
brad-decker: In the interest of narrowing scope of this PR I feel we should make that change in a subsequent PR

DELJUJAZI (approved):

adonesky1 (approved):
LGTM

jpuri (approved):
👏   thanks for looking into improving code quality.

PR #13740, opened 2022-02-23
Add custom CSS to Popover

hmalik88:
Explanation:  This PR is a result of the following conversation: https://github.com/MetaMask/metamask-extension/pull/13315#discussion_r788063156

The idea is to add the `Box` component into the `Popover` component and have it accept `headerProps`, `contentProps` and `footerProps`. This ensures that we can add custom CSS into the `Popover` component without having to reference the `Popover` component's classes directly inside of other component's CSS files.

NOTE: `README.mdx` was removed from the `Popover` folder because it currently served no purpose apart from showing the component's props which can already be done without the need for an `.mdx` file.


Jochy1873:
Were is my wallets in CoinGecko that's is block unblock I have evidence your team take my Eth take in my wallet moving other time. My value that take other wallet y have evidence  o have  the token I'd  you erase from my folder I have more copy y have e for pay in crypto .com

Jochy1873:
And other time my Mco leave from my Crypto.com wallet  I need my currency now or I'm go reported to fcc

Jochy1873:
I have evidence your team delete. My evidence were is my shivas my leash usdt  Dai usdc 1 inch uniswap when my make swat in Shiba swat you altered and errace in my google the file  I have evidence

ui/components/ui/popover/popover.component.js
```diff
@@ -35,44 +78,31 @@ const Popover = ({
...
         ) : null}
         {footer ? (
           <footer className={classnames('popover-footer', footerClassName)}>
-            {footer}
+            {hasStyles(footerProps) ? (
+              <Box {...footerProps}>{footer}</Box>
+            ) : (
+              footer
+            )}
           </footer>
```
georgewrmarshall, line 106: Would we have specificity issues if we just changed the footer element to be a Box? e.g.

```
<Box {...footerProps} tag="footer" className={classnames('popover-footer', footerClassName, footerProps?.className)}>{footer}</Box>
```

Not sure if the optional chaining operator works or if it would list "undefined" as a className but you get the idea. I think we would need to add "footer" to `tag` prop options in `Box` too...

georgewrmarshall (commented):
I like the intention of this PR a lot. I'm just not sure about the pattern of adding an element based on if it should contain props I think it should already exist? So instead of `headerProps ? <Box {...headerProps} : <Header />` Just include the Box regardless of headerProps existing or not? I understand it's less html rendered if it does not exist but I think this pattern may create some confusion.

We should also add docs about how the headerProps, contentProps and footerProps work in the MDX docs in storybook. Don't have to document every prop of the component just the ones that are being introduced in this PR.

e.g.

### headerProps, contentProps, footerProps

The wrapping element for the header, content and footer is the Box component and all Box props can be passed to them using `headerProps`, `contentProps` or `footerProps`

ui/components/ui/popover/popover.component.js
hmalik88: Yeah, good point. I was on the fence of just straight up changing the tags themselves to `Box` but felt iffy on removing the tags, I had forgot about the `tag` prop. I think I'll add in the tags that are missing and change the elements to `Box` instead of nesting them.

hmalik88: @georgewrmarshall actually, I think you might have been thinking about the `Typography` component, the `Box` component does not have a tag prop...I can add one though, but that opens the question of what tags we think should classify as a box. Also, why do you think the HTML rendering of `Box` only when there are actually props is confusing?

hmalik88: Even then, I think it's fine to change the elements to `Box` themselves even though we don't specify a tag for them within the component itself..

georgewrmarshall: @hmalik88 yes you're right! I was. IMO all components should have a polymorphic tag prop that should allow any html element to be rendered as it's base. For type validation I think we would just add to the list as we go. Starting with `'div', 'footer'`. As to the second point. Personally, if I were to add a className or set of props using the `headerProps` or other prop object I would think that I'm adding it to the already existing root element without changing the DOM structure but that is just my opinion. Which ever pattern we choose I think we should just be consistent across the board.

hmalik88: Oh I see your point! In that case, I'll take it out. Thanks!

ui/components/ui/popover/popover.component.js
```diff
@@ -18,9 +19,47 @@ const Popover = ({
...
   centerTitle,
+  headerProps = {},
+  contentProps = {},
+  footerProps = {},
 }) => {
   const t = useI18nContext();
   const showHeader = title || onBack || subtitle || onClose;
+  const Header = () => {
+    return (
+      <Box {...headerProps} className="popover-header" tag="header">
```
georgewrmarshall, line 30: I guess we should update the Box component to accept a `tag` prop as well and include the types used here as well as div?

georgewrmarshall (commented):
This is looking a lot cleaner to me! I'm just wondering if we will be able to update the `Box` component to have a polymorphic prop similar to the Typography component.

ui/components/ui/popover/popover.component.js
hmalik88: Removed the tag prop, that was unintentional. I say that we make a separate issue/PR for making it polymorphic.

georgewrmarshall (commented):
Not sure about the ArgsTable. There is a workaround that I think would be ok for now. Delete these lines in `popover.stories.js`
```
parameters: {
    docs: {
      page: README,
    },
  },
```

Then you can remove the README.mdx import and also delete the README.mdx file as it doesn't really contain much info about the component API. Then in the docs page you should get something like:

Screen Shot 2022-03-03 at 2 16 10 PM

georgewrmarshall (approved):
LGTM! May want to updates the PR description with some screenshots or comment on why the README.mdx has been removed for future reviewers?

jpuri (approved):

PR #13651, opened 2022-02-16
Dark Theme setup and button theming

GuillaumeRx:
Explanation: Add `@metamask/design-tokens` that includes dark theme color variables and apply it on the button component

PR #13742, opened 2022-02-23
Updating some styles and removing some deprecated buttons

georgewrmarshall:
Explanation: Some style suggestion updates and removing some deprecated buttons.

Manual testing steps:  
  - `yarn storybook`
  -  Go to button type story

## Video

https://user-images.githubusercontent.com/8112138/155430696-fbb78fd6-7a9e-4ac6-84be-925304958f11.mov

### DIFF
--- a/ui/components/ui/button/button.component.js
+++ b/ui/components/ui/button/button.component.js
@@ -5,11 +5,9 @@ import classnames from 'classnames';
 const CLASSNAME_DEFAULT = 'btn-default';
 const CLASSNAME_PRIMARY = 'btn-primary';
 const CLASSNAME_SECONDARY = 'btn-secondary';
-const CLASSNAME_CONFIRM = 'btn-primary';
 const CLASSNAME_RAISED = 'btn-raised';
 const CLASSNAME_LARGE = 'btn--large';
 const CLASSNAME_ROUNDED = 'btn--rounded';
-const CLASSNAME_FIRST_TIME = 'btn--first-time';
 const CLASSNAME_INLINE = 'btn--inline';
 
 const typeHash = {
@@ -21,10 +19,7 @@ const typeHash = {
   'danger-primary': 'btn-danger-primary',
   link: 'btn-link',
   inline: CLASSNAME_INLINE,
-  // TODO: Legacy button type to be deprecated
-  confirm: CLASSNAME_CONFIRM,
   raised: CLASSNAME_RAISED,
-  'first-time': CLASSNAME_FIRST_TIME,
 };
 
 const Button = ({
--- a/ui/components/ui/button/button.stories.js
+++ b/ui/components/ui/button/button.stories.js
@@ -91,6 +91,10 @@ export const Type = (args) => (
       {args.children || 'Danger primary'}
     </Button>
     <br />
+    <Button {...args} type="raised">
+      {args.children || 'Raised'}
+    </Button>
+    <br />
     <Button {...args} type="link">
       {args.children || 'Link'}
     </Button>
--- a/ui/components/ui/button/buttons.scss
+++ b/ui/components/ui/button/buttons.scss
@@ -51,7 +51,7 @@
 }
 
 .btn-warning {
-  color: var(--color-warning-inverse);
+  color: var(--color-text-default);
   border: 1px solid var(--color-warning-default);
   background-color: var(--color-background-default);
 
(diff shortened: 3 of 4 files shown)


ui/components/ui/button/button.component.js
```diff
@@ -21,10 +19,7 @@ const typeHash = {
   'danger-primary': 'btn-danger-primary',
   link: 'btn-link',
   inline: CLASSNAME_INLINE,
-  // TODO: Legacy button type to be deprecated
-  confirm: CLASSNAME_CONFIRM,
   raised: CLASSNAME_RAISED,
-  'first-time': CLASSNAME_FIRST_TIME,
 };
 
 const Button = ({
```
georgewrmarshall, line 25: Consolidating and removing unused types. `raised` is used but only once that I can find going to leave it for now and add it to stories.

ui/components/ui/button/button.stories.js
```diff
@@ -91,6 +91,10 @@ export const Type = (args) => (
       {args.children || 'Danger primary'}
     </Button>
     <br />
+    <Button {...args} type="raised">
+      {args.children || 'Raised'}
+    </Button>
+    <br />
     <Button {...args} type="link">
       {args.children || 'Link'}
     </Button>
```
georgewrmarshall, line 100: Adding raised type for reference

ui/components/ui/button/buttons.scss
```diff
@@ -3,14 +3,6 @@
  */
 
 .button {
-  --hover-secondary: #b0d7f2;
-  --hover-default: #b3b3b3;
-  --hover-confirm: #0372c3;
-  --hover-red: #feb6bf;
-  --hover-red-primary: #c72837;
-  --hover-orange: #ffd3b5;
-  --warning-light-orange: #f8b588;
```
georgewrmarshall, line 12: Removing unused css variables.

ui/components/ui/button/buttons.scss
```diff
@@ -59,23 +51,23 @@
...
-    border-color: var(--color-secondary-default);
+    background: var(--color-warning-alternative);
+    border: 1px solid var(--color-warning-alternative);
   }
 
   &--disabled,
   &[disabled] {
     opacity: 1;
-    color: var(--color-secondary-disabled);
+    color: var(--color-warning-disabled);
```
georgewrmarshall, line 70: Updating warning button to use warning colors. We want to stay away from using the orange that is associated with the metamask fox logo with any negative connotations e.g. warning, error etc

ui/components/ui/button/buttons.scss
```diff
@@ -101,7 +93,7 @@
 }
 
 .btn-danger-primary {
-  color: var(--color-background-default);
+  color: var(--color-error-inverse);
```
georgewrmarshall, line 96: The `inverse` category should be used for any foreground elements that go on top of the `default` category all the below text examples use the "inverse" in its respective category. 

inverse

ui/pages/swaps/searchable-item-list/item-list/item-list.component.js
```diff
@@ -142,7 +142,7 @@ export default function ItemList({
                 ) : null}
               </div>
               {result.notImported && (
-                <Button type="confirm" onClick={onClick}>
+                <Button type="primary" onClick={onClick}>
                   {t('import')}
                 </Button>
               )}
```
georgewrmarshall, line 148: This is the only instance of the confirm button that I could find. @NiranjanaBinoy is there any reason that this type is used as opposed to primary type?

GuillaumeRx:
Thanks for the clarification on the colors usage ! I would say it's a bad idea to change the color of the warning button, it's a drastic change and sometimes it's not well interpreted on the user side (cf. our discution with @darkwing)

GuillaumeRx:
image

ui/pages/swaps/searchable-item-list/item-list/item-list.component.js
NiranjanaBinoy: I don't think so; I might have missed updating the type since it was already using the <Buttton/> component; my attention was on getting the rounded removed while updating that component.

georgewrmarshall: All good! Thanks for the update @NiranjanaBinoy 🙏

GuillaumeRx:
Merged in https://github.com/MetaMask/metamask-extension/pull/13651

PR #13745, opened 2022-02-24
Swapping the order of ui and app stylesheets

### LINKED ISSUE #13744: Swap imports of stylesheets to improve specificity
**Description**
Currently our stylesheets are ordered so the `ui` component stylesheet comes after the `app` component stylesheet. This creates higher specificity for the `ui` component styles when in most cases the `app` components should override the `ui` css. This is a big change and will need to be thoroughly tested. However I **cannot** see when it would be beneficial for a `ui` component styles to override an `app` components styles...

**Technical Details**
swap order of imports of stylesheets in `ui/css/index.scss`
From
```
@import '../components/app/app-components';
@import '../components/ui/ui-components';
```

To
```
@import '../components/ui/ui-components';
@import '../components/app/app-components';
```

## Screenshots
Before
Screen Shot 2022-02-24 at 3 33 20 PM

After
Screen Shot 2022-02-24 at 3 36 30 PM

georgewrmarshall:
Fixes: #13744 

Explanation:  Swaps the order of imports of the `ui` and `app` component stylesheets to improve css class order and overriding styles. This saves devs from using `!important` in their `app` component styles when trying to override some UI component styles.

This **needs be thoroughly tested** but I can't see when there would be a case that a `ui` component styles should override an `app` components styles.

Manual testing steps:  
  - This is kind of hard because it effects all `ui/components/ui` and `ui/components/app` component styles.
  - Run `yarn storybook`
  - Check if any `UI` or `App` component styles look broken?
  
## Screenshots
**BEFORE** Example of custom `app` component styles being overwritten by a `ui` component styles

Screen Shot 2022-02-24 at 3 33 20 PM

**AFTER** Custom styles move up in order

Screen Shot 2022-02-24 at 3 36 30 PM

### DIFF
--- a/ui/css/index.scss
+++ b/ui/css/index.scss
@@ -8,8 +8,8 @@
 @import './design-system/index';
 @import './utilities/colors.scss';
 @import './base-styles.scss';
-@import '../components/app/app-components';
 @import '../components/ui/ui-components';
+@import '../components/app/app-components';
 @import '../pages/pages';
 
 /*


darkwing:
I think this idea is really good.  I navigated around a bit and did see these two issues:

Arrow outside of its container:
SendAsset

Slim-height network dropdown:
DifferentDropdown

We'll definitely need more eyes/QA to identify other potential problems.

PR #13759, opened 2022-02-25
Handling undefined case for token balance (Token Details)

ryanml:
This produces a console error on the token details page:
Screen Shot 2022-02-24 at 11 40 21 PM

The initial value for `tokenBalance` can come in as `undefined` before being available with the correct value. This handles the display of this value similar to `tokenCurrencyBalance`: https://github.com/MetaMask/metamask-extension/blob/develop/ui/pages/token-details/token-details-page.js#L111 - so we don't have any undefined children


darkwing (approved):

georgewrmarshall (approved):
LGTM!

PR #13760, opened 2022-02-25
Improving identicon settings accessibility

### LINKED ISSUE #13753: Make Identicon Settings More Accessible
### Describe the bug

I noticed that the identicon settings makes you specifically click the circular image but you can't also click the text, which is what you'd expect as form `label`.

### Steps to reproduce

1. Go to Settings > General
2. Click the verbiage next to either circular image
3. The preference doesn't change

_No response_

### Version

10.10.0

### Build type

_No response_

### Browser

Firefox

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

ryanml:
Fixes: https://github.com/MetaMask/metamask-extension/issues/13753

https://user-images.githubusercontent.com/8732757/155673998-7170a82b-0dc6-4a4e-bab9-acfb138f9310.mov

### DIFF
--- a/ui/pages/settings/index.scss
+++ b/ui/pages/settings/index.scss
@@ -298,7 +298,8 @@
         flex-direction: row;
         align-items: center;
         justify-content: space-between;
-        cursor: pointer;
+        background: unset;
+        border: 1px solid transparent;
 
         &__icon {
           &--active {
--- a/ui/pages/settings/index.scss
+++ b/ui/pages/settings/index.scss
@@ -298,6 +298,7 @@
         flex-direction: row;
         align-items: center;
         justify-content: space-between;
+        cursor: pointer;
 
         &__icon {
           &--active {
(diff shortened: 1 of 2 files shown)


NiranjanaBinoy (dismissed):

darkwing (changes requested):
This is a slight improvement but I think we need to make the element a `<button>` instead of a `div` so that it's keyboard accessible.

ryanml:
@darkwing - I've made the change for keyboard accessibility:
https://user-images.githubusercontent.com/8732757/155790514-56443f0c-0366-4a4d-b08c-f92069826628.mov

darkwing (dismissed):
Nice improvement!  I think we can improve this in 2 ways:

1.  `background: unset` will get rid of the gray background
2.   We should add a transparent border around the unselected item so that there's no height jump

ryanml (commented):
Thanks @darkwing - updated

brad-decker (approved):

jpuri (approved):

PR #13762, opened 2022-02-25
Improvements in connectRemote function in background.js

jpuri:
Improvements in connectRemote function in background.js


app/scripts/background.js
```diff
@@ -345,12 +341,22 @@ function setupController(initState, initLangCode) {
    */
   function connectRemote(remotePort) {
     const processName = remotePort.name;
-    const isMetaMaskInternalProcess = metamaskInternalProcessHash[processName];
 
     if (metamaskBlockedPorts.includes(remotePort.name)) {
       return;
     }
 
+    let isMetaMaskInternalProcess = false;
```
darkwing, line 349: I don't think we need to initialize this value to `false` since the `else` below will set that value.

darkwing (changes requested):

app/scripts/background.js
rekmarks: The safe, explicit default is nice though. I think we should keep it!

darkwing: Cool with me

darkwing (dismissed):

jpuri:
I found that e2e on FF is broken even after updating the branch, I am working on it.

darkwing (dismissed):

app/scripts/background.js
```diff
@@ -345,12 +347,22 @@ function setupController(initState, initLangCode) {
...
+    const sourcePlatform = getPlatform();
+
+    if (sourcePlatform === PLATFORM_FIREFOX) {
+      isMetaMaskInternalProcess = metamaskInternalProcessHash[processName];
+    } else {
```
jpuri, line 360: I have reverted the changes for FF
```
if (sourcePlatform === PLATFORM_FIREFOX) {
      isMetaMaskInternalProcess = remotePort.sender.id === extension.runtime.id;
    }
    ```
    
    as it was not working for dapps

adonesky1:
@jpuri could you add a bit more context to the description to contextualize the goals/approach of these changes?

app/scripts/background.js
adonesky1, line 360: Are there other `sourcePlatforms` that are not `PLATFORM_FIREFOX` for which this will still work? Opera? Brave? cc @jpuri

jpuri: Other browsers are chromium based so should work well. But good point, I can test it once to confirm.

adonesky1: Yea just want to make sure that this format is the same across them.

jpuri: I tested it on brave and it works as expected.

adonesky1: Should we test on Opera too since it would be pretty consequential for those users if for some reason this didn't work?

jpuri: Yep just tested in opera also and works as expected.

adonesky1 (approved):

segun (approved):

PR #13775, opened 2022-02-25
ConfirmImportToken: Update Metrics Logic

digiwand:
Fixes: # 
follow-up for https://github.com/MetaMask/metamask-extension/pull/13594

Explanation:  
We want to make the metrics logic in the ConfirmImportToken component more robust. Currently, it is using a solution that is susceptible to breaking changes. We are using `useEffect` to call `useNewMetricsEvent`. This change removes to use of `useEffect` and `useNewMetricsEvents` by calling `trackEvent` directly.

Manual testing steps:  
- On the main page and underneath the list of accounts, click "import tokens"
- Search any token and select it
- Click "Next"
- Test this "Import Tokens" page works as expected
- Check "Token Added" event in Mixpanel is being tracked as expected


darkwing (approved):
Well done!

segun (approved):
LGTM

digiwand:
Thanks for the reviews!

PR #13683, opened 2022-02-18
Confirm transaction page: show contract address details in title

jpuri:
Improvements in contract name display on transaction confirmation pages.

----
Simple transaction not a contract:

Screenshot 2022-02-19 at 12 12 15 AM

----
Contract deployment:

Screenshot 2022-02-19 at 12 10 36 AM

----
Contract whose name is not saved, user can click address to save it:

Screenshot 2022-02-19 at 12 10 55 AM

----
Contract whose name is saved in address book:

Screenshot 2022-02-19 at 12 11 13 AM

PR #13776, opened 2022-02-26
Add contract address as prefix to method display in confirmation screen for token contract calls `transfer` and `transferFrom`

adonesky1:
In https://github.com/MetaMask/metamask-extension/pull/13683, @jpuri added the contract address as a prefix to the component where we display the name of the method being called on a contract, for example:
154743478-20f7d146-d097-4af8-88f0-edbd9f9b0412

However the initial implementation excluded contract interactions with the methods `transfer` and `transferFrom`, I suggested we include these methods (related threads [here](https://github.com/MetaMask/metamask-extension/pull/13683#discussion_r813158404) and [here](https://github.com/MetaMask/metamask-extension/pull/13683#discussion_r814898366)), and we ultimately settled on making those additions in a separate PR - this work brings in those changes: 
ERC721 `transferFrom`:
Screen Shot 2022-03-02 at 10 42 45 AM
ERC20 `transfer`:
Screen Shot 2022-03-02 at 10 43 28 AM

This work changes the naming of the prop `identiconAddress` up the stack of components from which it was being pass (it was being renamed from `tokenAddress` to `identiconAddress` in `confirm-page-container-content.component.js`) back to `tokenAddress` for clarity, and then uses the `tokenAddress` as the `contractAddress` to pass to `useAddressDetails`.  We need to use the `tokenAddress` rather than the `toAddress` when the transaction is a `transfer` or `transferFrom` type because in these cases the `toAddress` is the `recipient` rather than the contract address itself.

### DIFF
--- a/ui/components/app/confirm-page-container/confirm-page-container-content/confirm-page-container-content.component.js
+++ b/ui/components/app/confirm-page-container/confirm-page-container-content/confirm-page-container-content.component.js
@@ -28,7 +28,7 @@ export default class ConfirmPageContainerContent extends Component {
     errorMessage: PropTypes.string,
     hasSimulationError: PropTypes.bool,
     hideSubtitle: PropTypes.bool,
-    identiconAddress: PropTypes.string,
+    tokenAddress: PropTypes.string,
     nonce: PropTypes.string,
     subtitleComponent: PropTypes.node,
     title: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
@@ -106,7 +106,7 @@ export default class ConfirmPageContainerContent extends Component {
       titleComponent,
       subtitleComponent,
       hideSubtitle,
-      identiconAddress,
+      tokenAddress,
       nonce,
       detailsComponent,
       dataComponent,
@@ -179,7 +179,7 @@ export default class ConfirmPageContainerContent extends Component {
           titleComponent={titleComponent}
           subtitleComponent={subtitleComponent}
           hideSubtitle={hideSubtitle}
-          identiconAddress={identiconAddress}
+          tokenAddress={tokenAddress}
           nonce={nonce}
           origin={origin}
           hideTitle={hideTitle}
--- a/ui/components/app/confirm-page-container/confirm-page-container-content/confirm-page-container-summary/confirm-page-container-summary.component.js
+++ b/ui/components/app/confirm-page-container/confirm-page-container-content/confirm-page-container-summary/confirm-page-container-summary.component.js
@@ -20,22 +20,40 @@ const ConfirmPageContainerSummary = (props) => {
     subtitleComponent,
     hideSubtitle,
     className,
-    identiconAddress,
+    tokenAddress,
+    toAddress,
     nonce,
     origin,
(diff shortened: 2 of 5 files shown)


jpuri:
@adonesky1 : I dropped my feedback here: https://github.com/MetaMask/metamask-extension/pull/13683/files#r816559593

holantonela:
> We need to use the `tokenAddress` rather than the `toAddress` when the transaction is a `transfer` or `transferFrom` type because in these cases the `toAddress` is the recipient rather than the contract address itself.

It makes sense for me. What will we do with `approve` type of confirmations?

jpuri:
@adonesky1 , @holantonela : I have a question here, I am still not sure if we really need this change for transaction other than contract_interaction. As described here: https://github.com/MetaMask/metamask-extension/pull/13683#issue-1143570316 this change was done only for transactions where method name is derived from data.

In case of `transfer and transferFrom` the action is already clear to the user.

adonesky1:
> It makes sense for me. What will we do with approve type of confirmations?

Where this goes for the `approve` confirmation screen isn't as clear, given its current designs:

Screen Shot 2022-03-02 at 12 47 28 PM

Maybe after the token symbol here? cc @holantonela

adonesky1:
> In case of transfer and transferFrom the action is already clear to the user.

Per my comments [here](https://github.com/MetaMask/metamask-extension/pull/13683#discussion_r813158404), I had previously discussed with @Gudahtt and @SayaGT the ways in which our confirmation screens for `transfer` and `transferFrom` were inadequately clear about the contract the user is interacting with (the token symbol and/or token image can be faked and should not be the only indication we provide the user of what contract they are interfacing with).  That is why I think we should implement this same solution for these types of contract interactions as well.

cc @jpuri

jpuri (approved):
Code changes look good 👍

holantonela (approved):

adonesky1:
@holantonela @sethkfman here are a bunch of screenshots of how confirmation screens for various methods are handled as of this PR:
ERC721 `transferFrom`:
Screen Shot 2022-03-03 at 1 50 05 PM

ERC20 `transfer`:
Screen Shot 2022-03-03 at 1 50 21 PM

ERC20 `approve`:
Screen Shot 2022-03-03 at 1 51 03 PM

`Contract Deployment`:
Screen Shot 2022-03-03 at 1 52 20 PM

Then all other unrecognized contract interaction types:
`mint`:
Screen Shot 2022-03-03 at 1 51 52 PM

`drain` on this [NFT contract](https://rinkeby.etherscan.io/address/0xf5de760f2e916647fd766B4AD9E85ff943cE3A2b#writeContract)) :
Screen Shot 2022-03-03 at 1 53 27 PM

`pull` (on this ERC20 [contract](https://rinkeby.etherscan.io/address/0x6A9865aDE2B6207dAAC49f8bCba9705dEB0B0e6D#writeContract))
Screen Shot 2022-03-03 at 1 54 13 PM

All other methods for which we don't have a special case/screen will follow this pattern as far as representing the method (with the contract address beside it)

PR #13780, opened 2022-02-27
fix: chooseAccountList will be scrollable on overflow

CeamKrier:
Fixes: #

This PR will be fixing the overflow issue on connecting an account if there is a long list of available accounts.

Issue: 
image

Proposed fix:
image

Explanation:  
Account list component's last child gets under the action buttons if the list's height is larger than the pop-up window. To fix the issue, I propose to make the account list component overflow if its height is greater than its container

Manual testing steps:  
  - A metamask account that has 7+ accounts in it
  - Try to connect your account to a web3 application
  - Check the bottom action buttons to see the last account on the list is under the action buttons

### DIFF
--- a/ui/pages/permissions-connect/choose-account/index.scss
+++ b/ui/pages/permissions-connect/choose-account/index.scss
@@ -1,4 +1,5 @@
 .permissions-connect-choose-account {
+  position: fixed;
   display: flex;
   flex-direction: column;
   align-items: center;
--- a/ui/components/ui/account-list/index.scss
+++ b/ui/components/ui/account-list/index.scss
@@ -3,6 +3,7 @@
   flex-direction: column;
   width: 100%;
   align-items: center;
+  overflow-y: auto;
 
   &__header--one-item,
   &__header--multiple-items {


darkwing (changes requested):
Wow, this is excellent @CeamKrier !  This makes me sooo happy!

The only issue I'm seeing on my Mac / Firefox is scrolling and overflow:

HasScrolling

If we can get that tightened up we can merge ASAP!  Thank you!

CeamKrier:
Hey @darkwing 

I could not reproduce the exact issue but I think it is originating from the parent's overflow so I disabled it

mcmire (approved):
CI is failing, but I think it's an intermittent failure. I tested this out manually and it looks good to me.

ui/pages/permissions-connect/choose-account/index.scss
```diff
@@ -1,4 +1,5 @@
 .permissions-connect-choose-account {
+  position: fixed;
```
mcmire, line 2: Actually, I'm not certain we want this. This will end up placing the site being connected to right up against the top of the screen. Trying to figure out an alternative.

mcmire (changes requested):

mcmire:
In reference to my comment above, I think instead of affixing one of the children of the page container to the top of the window, it would be better to simplify the structure of the `.permissions-connect` div. What if you:

* Update `ui/pages/permissions-connect/choose-account/choose-account.js` so that instead of wrapping everything in `<div className="permissions-connect-choose-account">...</div>` we just wrap it in `<>...</>`
* Remove the styles for `permissions-connect-choose-account` from `ui/pages/permissions-connect/choose-account/index.scss` (not the entire file; just the bit before `@media screen and (min-width: $break-large) {`
* Update `ui/components/app/permissions-connect-header/index.scss` and remove `width: 92%`

If you make these changes you will see that the parts of the window will naturally expand vertically without the need for `position: fixed`. This should work on Chrome and Firefox and regardless of whether there is 1 account or 20 accounts.

What do you think about this @darkwing?

darkwing:
Superceded by https://github.com/MetaMask/metamask-extension/pull/13859 ; thank you @CeamKrier !

PR #13788, opened 2022-02-28
Refactor token send/method confirmation flow (trimmed down)

adonesky1:
Fixes: #11611 (and generally improves the UX of NFT confirmation screens)

**Overview**
- Creates a sub-component `ConfirmTokenTransactionSwitch` through which token/NF confirmation screens for  the methods `approve`, `transfer`, and `transferFrom` are routed.
- `ConfirmTokenTransactionSwitch` instantiates a newly created hook `useAssetDetails` which passes the pending transaction to the new method [`getTokenStandardAndDetails`](https://github.com/MetaMask/controllers/blob/81bcbae1c4abe71a9b07819606e155a33ec4a6b4/src/assets/AssetsContractController.ts#L151) in the `AssetsContractController` and retrieves all the data ( `assetName`, `userBalance`, `tokenSymbol`, `decimals`, `image` etc.) we need for the three **token** specific method pages we currently support (`approve`, `transfer`, and `transferFrom`). This method call also returns the `assetStandard` to enable improved  branching logic to provide suitable language depending on the token contract standard (`ERC20`, `ERC721` or `ERC1155`). 
    - **Importantly** this eliminates the [unwanted behavior](https://github.com/MetaMask/metamask-extension/security/advisories/GHSA-jwfv-7854-fmc3) of adding the token to the `TokensController` state whenever a user interacts with a confirmation transaction of a token not already in `TokensController` state

**Additionally (though less importantly)**
- Refactors `confirm-token-transaction-base` and `confirm-send-token` from class based to functional components for improved ease of use.

### The behavior/UX should be unaltered for `ERC20` method calls listed above but be more accurate and clear for the same method calls for `ERC721` and `ERC1155` contracts.

**For Example**
**ERC721 `approve` call before:**
Screen Shot 2022-02-09 at 5 31 42 PM
the tokenId is incorrectly being used as token amount ☝️ 👇  
Screen Shot 2022-02-09 at 5 26 07 PM

**after:**
the option to change the amount should not be present for ERC721 tokens 👇 
Screen Shot 2022-01-14 at 2 36 02 PM

**ERC721 `transferFrom` call before:**
Screen Shot 2022-02-09 at 5 34 32 PM

**after:**
Screen Shot 2022-02-09 at 5 53 34 PM

PR #13789, opened 2022-02-28
refactor confirm-transaction component from class to functional

adonesky1:
Explanation:  Originally part of larger refactor (https://github.com/MetaMask/metamask-extension/pull/13131), extracted out because it is not necessary to the core of that work, and may distract from the core changes (new trimmed down version here: https://github.com/MetaMask/metamask-extension/pull/13788).

 - Refactors base confirmation routing component - `confirm-transaction.js` - from class to functional (includes providing all required redux state/methods via selectors/dispatch, and changing lifecycle methods to `useEffect`) while preserving all functionality.


adonesky1:
Very stale. probably not worth it at this point.

PR #13792, opened 2022-03-01
Create password page

### LINKED ISSUE #13443: Create Password page
- [x] "learn more" should be capitalized and lead to a link with more information about passwords in metamask rather than terms of service.*

- [ ] The create password screen does not adequately emphasize that we cannot recover the password, and it doesn't communicate what the password is used for very well.*

- [ ] Clicking on the password input doesn't work if you've recently typed a key into the other input. e.g. type something into "confirm password" then try clicking "password" right away. It does not work until you wait a few seconds. (personally I'm not able to reproduce this issue)

- [ ] 	After creating a password, it shows the login page briefly with a loading overlay before navigation to the success screen. This shouldn't happen. It should lead directly to the success screen.

- [ ] The "Passwords Don't Match" error has every word capitalized, but our convention is to only capitalize the first word of each phrase.

*requires design input (cc @rachelcope)

VSaric:
fixes: #13443

**Explanation:**  Create password page, added zxcvbn module and check for password strength

**Screencast:**

https://user-images.githubusercontent.com/92527393/156155816-c1f84827-2a3b-4a10-9c1b-fe5be4073c69.mov

### DIFF
--- a/lavamoat/build-system/policy.json
+++ b/lavamoat/build-system/policy.json
@@ -1052,6 +1052,16 @@
         "buffer-equal": true
       }
     },
+    "are-we-there-yet": {
+      "builtin": {
+        "events.EventEmitter": true,
+        "util.inherits": true
+      },
+      "packages": {
+        "delegates": true,
+        "readable-stream": true
+      }
+    },
     "arr-diff": {
       "packages": {
         "arr-flatten": true,
@@ -1460,6 +1470,7 @@
         "anymatch": true,
         "async-each": true,
         "braces": true,
+        "fsevents": true,
         "glob-parent": true,
         "inherits": true,
         "is-binary-path": true,
@@ -1724,6 +1735,16 @@
       "packages": {
         "shasum-object": true,
         "through2": true
+      }
+    },
+    "detect-libc": {
+      "builtin": {
+        "child_process.spawnSync": true,
+        "fs.readdirSync": true,
+        "os.platform": true
+      },
+      "globals": {
+        "process.env": true
       }
     },
     "detective": {
@@ -2429,6 +2450,45 @@
         "process.version": true
       }
     },
+    "fsevents": {
+      "builtin": {
+        "events.EventEmitter": true,
+        "fs.stat": true,
+        "path.join": true,
+        "util.inherits": true
+      },
+      "globals": {
+        "__dirname": true,
+        "process.nextTick": true,
+        "process.platform": true,
+        "setImmediate": true
+      },
+      "native": true,
+      "packages": {
+        "node-pre-gyp": true
+      }
+    },
+    "gauge": {
+      "builtin": {
+        "util.format": true
+      },
+      "globals": {
+        "clearInterval": true,
+        "process": true,
+        "setImmediate": true,
+        "setInterval": true
+      },
+      "packages": {
+        "aproba": true,
+        "console-control-strings": true,
+        "has-unicode": true,
+        "object-assign": true,
+        "signal-exit": true,
+        "string-width": true,
+        "strip-ansi": true,
+        "wide-align": true
+      }
+    },
     "get-assigned-identifiers": {
       "builtin": {
         "assert.equal": true
@@ -2807,6 +2867,16 @@
         "process.argv": true
       }
     },
+    "has-unicode": {
+      "builtin": {
+        "os.type": true
+      },
+      "globals": {
+        "process.env.LANG": true,
+        "process.env.LC_ALL": true,
+        "process.env.LC_CTYPE": true
+      }
+    },
     "has-value": {
       "packages": {
         "get-value": true,
@@ -2976,6 +3046,11 @@
     "is-extendable": {
       "packages": {
         "is-plain-object": true
+      }
+    },
+    "is-fullwidth-code-point": {
+      "packages": {
+        "number-is-nan": true
       }
     },
     "is-glob": {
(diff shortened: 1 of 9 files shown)


mirjanaKukic:
Verified by QA

ui/pages/onboarding-flow/create-password/create-password.js
```diff
@@ -58,18 +62,77 @@ export default function CreatePassword({
...
+      case 3:
+        passwordStrengthInput = t('passwordStrength', [
+          <span
+            key={passwordEvaluation.score}
+            className="create-password__form__average"
+          >
+            {t('average')}
+          </span>,
+        ]);
+        passwordStrengthDescription = t('passwordStrengthDescription');
```
adonesky1, line 113: [nit] since this is the same for all cases except `strong` maybe we should initialize it to `t('passwordStrengthDescription');` and set it to `''` in case 4?

adonesky1 (dismissed):
Small nit but otherwise LGTM!

ui/pages/onboarding-flow/create-password/create-password.js
```diff
@@ -58,18 +62,77 @@ export default function CreatePassword({
...
   const handlePasswordChange = (passwordInput) => {
-    let error = '';
     let confirmError = '';
-    if (passwordInput && passwordInput.length < 8) {
-      error = t('passwordNotLongEnough');
+    let passwordStrengthDescription = '';
+    let passwordStrengthInput;
+    const passwordEvaluation = zxcvbn(passwordInput);
+
+    switch (passwordEvaluation.score) {
```
mcmire, line 70: Since the only two things that change here are class name and text, it seems like we can refactor this a bit. What do you think about this?

``` javascript
// You could define these outside of the component since they don't depend on anything, if you wanted

function getPasswordStrengthLabel(score, t) {
  if (score >= 4) {
    return {
      className: "create-password__form__strong",
      text: t("strong")
    }
  } else if (score === 3) {
    return {
      className: "create-password__form__average",
      text: t("average")
    };
  } else {
    return {
      className: "create-password__form__weak",
      text: t("weak")
    }
  } 
}

// Later on, inside the component...

const passwordStrengthLabel = getPasswordStrengthLabel(passwordEvaluation.score, t);
const passwordStrengthInput = t('passwordStrength', [
  <span
    key={passwordEvaluation.score}
    className={passwordStrengthLabel.className}
  >
    {passwordStrengthLabel.text}
  </span>,
]);
```

ui/pages/onboarding-flow/onboarding-flow.js
```diff
@@ -66,7 +66,7 @@ export default function OnboardingFlow() {
       return;
     }
 
-    if (isInitialized && !isUnlocked) {
+    if (!isInitialized && isUnlocked) {
```
mcmire, line 69: Was this incorrect? This seems like an important change if so. Maybe we ought to extract this to a separate PR for visibility purposes and write a test to catch this in the future?

ui/pages/onboarding-flow/create-password/index.scss
```diff
@@ -29,6 +29,18 @@
       }
     }
 
+    &__weak {
```
mcmire, line 32: [nit] According to BEM, nesting elements like this (`.foo__bar__baz`) is [discouraged](https://en.bem.info/methodology/faq/#why-not-create-elements-of-elements-block__elem1__elem2). What do you think about placing this as a direct child of `.create-password` instead?

mcmire (commented):
This looks good, but I had a question about the change to `onboarding-flow.js` that seems like it could get missed easily.

ui/pages/onboarding-flow/onboarding-flow.js
VSaric: @mcmire I made this change because @adonesky1 write one bullet task for this. Here is description of it:

`After creating a password, it shows the login page briefly with a loading overlay before navigation to the success screen. This shouldn't happen. It should lead directly to the success screen.`

When we have `isInitialized && !isUnlocked` it shows the login page briefly with loading overlay before navigation to the success screen. I will share you screencast what I made when we have this condition in `onboarding-flow.js` file.

https://user-images.githubusercontent.com/92527393/156780038-94855e2f-48d0-4a5e-91f4-6d54c98cd24e.mov

Maybe is good to extract this to separate PR. What do you think about this @adonesky1?

ui/pages/onboarding-flow/create-password/create-password.js
VSaric: @adonesky1 Fixed in: 31b1950541693058091725c294105bf99a39b4fa

ui/pages/onboarding-flow/create-password/create-password.js
VSaric: @mcmire Fixed in: 31b1950541693058091725c294105bf99a39b4fa

ui/pages/onboarding-flow/create-password/index.scss
VSaric: @mcmire Fixed in: 31b1950541693058091725c294105bf99a39b4fa

ui/pages/onboarding-flow/onboarding-flow.js
VSaric: @mcmire @adonesky1 I didn't revert changes in `onboarding-flow.js` file. I will wait for your comments about this.

adonesky1: I'm messing around with this and I agree with @mcmire that we should extract this into a separate PR, since its a bit tricky and requires a bit more direct thought.

adonesky1: Good call out @mcmire

VSaric: Thank you @adonesky1 for respond on this. I revert changes in `onboarding-flow.js` file here: ea104e72152e04834f36cd41dac5754ad20b2190

lavamoat/build-system/policy.json
```diff
@@ -2429,6 +2450,45 @@
         "process.version": true
       }
     },
+    "fsevents": {
```
adonesky1, line 2453: @kumavis @mcmire do we have a good way to evaluate changes to the lavamoat policy file? Or should we assume they're safu?

mcmire: I think as long as you can say `yarn start` and as long as CI works (since it runs `yarn build prod`) then you can assume these changes are safe? 🤔

mcmire (approved):
LGTM!

adonesky1 (approved):
LGTM (cc @VSaric please make sure to open a new PR to resolve the unlock page issue separately)

0ex-d (approved):

PR #13803, opened 2022-03-01
Improve unit tests for `SrpInput` component

Gudahtt:
The library `@testing-library/user-event` has been updated to the latest beta version, so that our unit tests better model real user interactions. In particular, I found that previously the `paste` event was missing the `clipboardData` API, so it was impossible to implement any custom handling of paste events (which we will need in later PRs).

See the `v14.0.0-beta.1` release notes for a [list of all breaking changes](https://github.com/testing-library/user-event/releases/tag/v14.0.0-beta.1). The main change is that all methods now return Promises. The `paste` method has also been dramatically simplified.

The unit tests have also been updated to reset all mocks before each test. These tests don't have any shared mocks, but this is generally a good practice, to ensure that tests don't develop accidental inter-dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -262,7 +262,7 @@
     "@testing-library/jest-dom": "^5.11.10",
     "@testing-library/react": "^10.4.8",
     "@testing-library/react-hooks": "^3.2.1",
-    "@testing-library/user-event": "^13.5.0",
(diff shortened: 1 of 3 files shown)


mcmire (approved):
Makes sense to me 👍🏻

adonesky1 (approved):
LGTM

PR #13810, opened 2022-03-02
Fix colorIndicator color render asynchronously

leonzchang:
Fixes: ColorIndicator in `network-dropdown` and `networks-list-item` color are not the same

before:
Screen Shot 2022-03-02 at 6 01 46 PM

after:
Screen Shot 2022-03-02 at 6 14 50 PM


leonzchang:

darkwing:
@leonzchang Hello!  Could you rebase?

@georgewrmarshall Do you have color recommendations here?

georgewrmarshall:
Hmm.. ok wow thanks for pointing how many discrepancies we have with our network display @leonzchang!! The ideal outcome would be to have the network icon if we have it e.g. ethereum. Then display it anywhere we display a network. Then for the fallback display the first letter of the network. The only exception are the test networks where we display the color and white letter. This update would add a considerable amount of work than just updating the color though.

Screen Shot 2022-05-16 at 3 06 44 PM
Screen Shot 2022-05-16 at 3 13 37 PM

PR #13818, opened 2022-03-02
Add e2e testcase for token details functionality

seaona:
**Explanation**:  
After [Token Details](https://github.com/MetaMask/metamask-extension/pull/13216) page has been added, we are increasing e2e coverage, for testing token details functionality.
This testcase imports a token and checks that token details page is correctly loaded, by looking at the contract address field on the token Details page.

### DIFF
--- a/test/e2e/tests/token-details.spec.js
+++ b/test/e2e/tests/token-details.spec.js
@@ -1,0 +1,56 @@
+const { strict: assert } = require('assert');
+const {
+  convertToHexValue,
+  withFixtures,
+  regularDelayMs,
+} = require('../helpers');
+
+describe('Token Details', function () {
+  const ganacheOptions = {
+    accounts: [
+      {
+        secretKey:
+          '0x7C9529A67102755B7E6102D6D950AC5D5863C98713805CEC576B945B15B71EAC',
+        balance: convertToHexValue(25000000000000000000),
+      },
+    ],
+  };
+  it('should show token details for an imported token', async function () {
+    await withFixtures(
+      {
+        fixtures: 'imported-account',
+        ganacheOptions,
+        title: this.test.title,
+      },
+      async ({ driver }) => {
+        await driver.navigate();
+        await driver.fill('#password', 'correct horse battery staple');
+        await driver.press('#password', driver.Key.ENTER);
+
+        await driver.clickElement({ text: 'import tokens', tag: 'a' });
+        await driver.clickElement({ text: 'Custom Token', tag: 'button' });
+
+        const tokenAddress = '0x2EFA2Cb29C2341d8E5Ba7D3262C9e9d6f1Bf3711';
+        const tokenSymbol = 'AAVE';
+
+        await driver.fill('#custom-address', tokenAddress);
+        await driver.delay(regularDelayMs);
+        await driver.fill('#custom-symbol', tokenSymbol);
+        await driver.clickElement({ text: 'Add Custom Token', tag: 'button' });
+        await driver.clickElement({ text: 'Import Tokens', tag: 'button' });
+        await driver.waitForSelector('[title="Asset options"]');
+        await driver.clickElement('[title="Asset options"]');
+        await driver.waitForSelector({ text: 'Token details', tag: 'span' });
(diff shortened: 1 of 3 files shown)


PeterYinusa (dismissed):

test/e2e/tests/token-details.spec.js
```diff
@@ -0,0 +1,56 @@
...
+        await driver.press('#password', driver.Key.ENTER);
+
+        await driver.clickElement({ text: 'import tokens', tag: 'a' });
+        await driver.clickElement({ text: 'Custom Token', tag: 'button' });
+
+        const tokenAddress = '0x2EFA2Cb29C2341d8E5Ba7D3262C9e9d6f1Bf3711';
+        const tokenSymbol = 'AAVE';
+
+        await driver.fill('#custom-address', tokenAddress);
+        await driver.delay(regularDelayMs);
```
mcmire, line 37: Is there a way to avoid the delay here or is this necessary in this case?

test/e2e/tests/token-details.spec.js
```diff
@@ -0,0 +1,56 @@
...
+        await driver.fill('#custom-symbol', tokenSymbol);
+        await driver.clickElement({ text: 'Add Custom Token', tag: 'button' });
+        await driver.clickElement({ text: 'Import Tokens', tag: 'button' });
+        await driver.waitForSelector('[title="Asset options"]');
```
mcmire, line 41: `clickElement` does an implicit wait, so you shouldn't need to wait for it first.

seaona: right, I confirm it is safe to remove the wait. Fixed on the new commit. Ty

test/e2e/tests/token-details.spec.js
seaona: Good point. Since we must wait for the address check, I've changed the delay for a wait for selector.

test/e2e/tests/token-details.spec.js
```diff
@@ -0,0 +1,50 @@
...
+
+        await driver.fill('#custom-address', tokenAddress);
+        await driver.waitForSelector('#custom-symbol-helper-text');
+        await driver.fill('#custom-symbol', tokenSymbol);
+        await driver.clickElement({ text: 'Add Custom Token', tag: 'button' });
+        await driver.clickElement({ text: 'Import Tokens', tag: 'button' });
+        await driver.clickElement('[title="Asset options"]');
+        await driver.clickElement({ text: 'Token details', tag: 'span' });
+
+        // Using findElements for avoiding error handling, if element is not present
```
mcmire, line 40: This tells me we should consider either adding another helper OR adding a custom assertion to make this easier, but this is good for now!

mcmire (dismissed):
LGTM!

test/e2e/tests/token-details.spec.js
mcmire: Oh great!

test/e2e/tests/token-details.spec.js
```diff
@@ -0,0 +1,54 @@
...
+        await driver.clickElement({ text: 'Add Custom Token', tag: 'button' });
+        await driver.clickElement({ text: 'Import Tokens', tag: 'button' });
+        await driver.clickElement('[title="Asset options"]');
+        await driver.clickElement({ text: 'Token details', tag: 'span' });
+
+        const tokenAddressFound = {
+          text: tokenAddress,
+        };
+
+        const exists = await elementExists(driver, tokenAddressFound);
```
mcmire, line 48: Actually :) should we add this to `driver.js`?

test/e2e/tests/token-details.spec.js
```diff
@@ -0,0 +1,54 @@
...
+
+        assert.equal(exists, true);
```
mcmire, line 50: And I think this can be `assert.ok(exists, "error message that should be shown if it doesn't exist")`

mcmire (commented):
Two small nits :)

PeterYinusa (approved):

mcmire (approved):
Thanks for the updates, looks good!

PR #13820, opened 2022-03-02
Add TypeScript migration dashboard

mcmire:
👉 [**View a live version of the dashboard here!**](https://metamask.github.io/metamask-extension-ts-migration-dashboard/) 👈

---

As we convert parts of the codebase to TypeScript, we will want a way to
track progress. This commit adds a dashboard which displays all of the
files that we wish to convert to TypeScript and which files we've
already converted.

The list of all possible files to convert is predetermined by walking
the dependency graph of each entrypoint the build system uses to compile
the extension (the files that the entrypoint imports, the files that the
imports import, etc). The list should not need to be regenerated, but
you can do it by running:

    yarn ts-migration:enumerate

The dashboard is implemented as a separate React app. The CircleCI
configuration has been updated so that when a new commit is pushed, the
React app is built and stored in the CircleCI artifacts. When a PR is
merged, the built files will be pushed to a separate repo whose sole
purpose is to serve the dashboard via GitHub Pages (this is the same
way that the Storybook works). All of the app code and script to build
the app are self-contained under
`development/ts-migration-dashboard`. To build this app yourself, you
can run:

    yarn ts-migration:dashboard:build

or if you want to build automatically as you change files, run:

    yarn ts-migration:dashboard:watch

Then open the following file in your browser (there is no server
component):

    development/ts-migration-dashboard/build/index.html

Finally, although you shouldn't have to do this, to manually deploy the
dashboard once built, you can run:

    git remote add ts-migration-dashboard git@github.com:MetaMask/metamask-extension-ts-migration-dashboard.git
    yarn ts-migration:dashboard:deploy

---

Screen Shot 2022-03-01 at 11 16 48 AM


package.json
```diff
@@ -70,11 +70,15 @@
...
+    "lavamoat:auto": "yarn lavamoat:build:auto && yarn lavamoat:background:auto",
+    "ts-migration:enumerate": "ts-node development/ts-migration-dashboard/scripts/writeListOfFilesToConvert.ts",
+    "ts-migration:dashboard:watch": "ts-node development/ts-migration-dashboard/scripts/build.ts --watch",
+    "ts-migration:dashboard:build": "ts-node development/ts-migration-dashboard/scripts/build.ts",
+    "ts-migration:dashboard:deploy": "gh-pages --dist development/ts-migration-dashboard/build --remote ts-migration-dashboard"
   },
   "resolutions": {
     "**/regenerator-runtime": "^0.13.7",
-    "**/caniuse-lite": "1.0.30001265",
+    "**/caniuse-lite": "^1.0.30001312",
```
mcmire, line 81: I was getting warnings about a mismatch between the expected version of `caniuse-lite` and the one that was installed so I bumped this version.

tsconfig.json
```diff
@@ -22,5 +23,8 @@
     "dist/**/*",
     "node_modules/**"
   ],
-  "extends": "@tsconfig/node14/tsconfig.json"
+  "extends": "@tsconfig/node14/tsconfig.json",
+  "paths": {
```
mcmire, line 27: This configuration option allows us to specify backfill TypeScript types for modules which lack them. The way this option works is explained here: https://github.com/MetaMask/metamask-module-template/pull/95#issuecomment-1079484410

mcmire:
Placing this PR back in draft status until conflicts are resolved.

development/ts-migration-dashboard/files-to-convert.json
```diff
@@ -0,0 +1,1737 @@
+[
```
mcmire, line 1: This file holds all of the files that we want to convert to TypeScript. It is generated from the [file patterns defined in `development/ts-migration-dashboard/scripts/constants.ts`](https://github.com/MetaMask/metamask-extension/pull/13820/files#diff-7403df2bdf0769e46840036cc85a72d36a7254adcf04e68be71ca9288ea1dd11R4), which includes the same entrypoints that the build system uses as well as patterns to represent tests and Storybook files. The files matched by the patterns are used as starting points and the dependency graphs of these files are traversed via the [`madge`](https://github.com/pahen/madge) package to produce this file list.

Why this approach? Because if we were to just use a glob like `{app,development,shared,ui}/**/*.js` then we might catch files that don't end up being used anywhere. By traversing the dependency graph we are essentially guaranteeing that all the files listed here are being referred to somewhere.

Why commit this file to the codebase? Because as we convert files to TypeScript, we may end up introducing new files, and we don't want those files to appear in the dashboard. By maintaining a list of the original files to convert we can ensure that the files listed on the dashboard don't change.

development/ts-migration-dashboard/scripts/build-module-partitions.ts
```diff
@@ -0,0 +1,336 @@
+import fs from 'fs';
```
mcmire, line 1: This file looks at the current state of the repo to determine all of the files that have been converted to TypeScript, filtered by the list of original files to convert, and produces the data structure the dashboard uses to render the UI. To render the various levels displayed on the UI (explained in the ["What is this?" section](https://metamask.github.io/metamask-extension-ts-migration-dashboard/)) we use the [`madge`](https://github.com/pahen/madge) package to iterate through the dependencies and subdependencies of a set of starting files.

As the code here is expensive to run, it doesn't get run when you load the dashboard, but rather during the build process. More specifically, it's run once when you run `yarn ts-migration:dashboard:build` or `yarn ts-migration:dashboard:watch`. The build script writes the output of this file to a temporary JSON file, and that JSON file is then read by the React app.

development/ts-migration-dashboard/scripts/build-module-partitions.ts
```diff
@@ -0,0 +1,336 @@
...
+  //
+  // These are called circular dependencies, and we detect them by tracking the
+  // files that depend on a file (dependents) in addition to the files on which
+  // that file depends (dependencies). In the example above, `baz.js` has a
+  // dependency `foo.js`, and its chain of dependents is `bar.js` -> `foo.js`
+  // (working backward from `baz.js`). The chain of dependents of `baz.js`
+  // includes `foo.js`, so we say `foo.js` is a circular dependency of `baz.js`,
+  // and we don't need to follow it.
+
+  const modulesToFill: Module[] = entryModuleIds.map((moduleId) => {
```
mcmire, line 199: I regret to say, upon looking back at this, that there are no tests for this file (for anything, but especially this). I don't know that I have time to add tests right now for this but I'm happy to explain what this is doing over a call.

mcmire:
Conflicts resolved and ready for review again!

mcmire:
There are LavaMoat issues that are causing CI to fail. I think we need to run:

```
yarn setup
yarn yarn-deduplicate
yarn lavamoat:auto
```

brad-decker (dismissed):
😍

mcmire:
New section of the metamaskbot message:

Screen Shot 2022-08-08 at 2 52 02 PM

mcmire:
@brad-decker Should be good to re-review now!

brad-decker (dismissed):
Great work, @mcmire

darkwing (dismissed):

brad-decker (approved):

Abadiq313 (approved):

darkwing (approved):

Issue #12811, opened 2021-11-23
Swap Component | Switching between ETH (or token) and USD Equivalent 


jrbgit:
**Description**
When sending funds  swapping back and forth from ETH/USD as the input amount, the component simply swaps the input type but leaves the original value there, not the estimated value of the new input type.

**Steps**
1. Click on Send
2. Enter an address or swap between accounts
3. Enter a value in Token (Like ETH)
4. Press the ^v icon to change to USD as the input

**Expected**
The value should calculate the correct amount of the other currency to equal what the original input was.

**Example**
Enter 0.12 ETH which is about $525. Swap the input to USD, and the value becomes 0.12 USD. It should be $525. Sometimes it's nice to swap back and forth between ETH/USD multiple times when choosing an ETH amount to send.

**Screenshots**
Available upon request

**My System**
 - OS: OS X Monterey - Latest Version - All Updates Done
 - Browser: Latest version - doesn't matter which browser since the issue is in metamask code. Brave/Chrome latest version
 - MetaMask Version: 10.6.1

**Other**
This could be considered a feature request, however any other crypto platform like Uniswap for example, does the conversion the way I think it should work. To me this is the industry standard expectation on how you would think it would work.

PR #13827, opened 2022-03-03
Fix switching between ETH and USD

### LINKED ISSUE #13275: Fix switching between ETH and USD
### **User Story**
As a user I want on switching the token the token value will be recalculated (estimated value of the new token).

**Context**: When sending funds swapping back and forth from ETH/USD as the input amount, the component simply swaps the input type but leaves the original value there, not the estimated value of the new input type.
On mobile it already works like that.

**Figma/Notion links:** 

**Notion**:https://www.notion.so/Switching-between-ETH-or-token-and-USD-Equivalent-on-send-screen-2463d1c7992e4705b190a0ea951ede4a

### Steps to Reproduce

1. Click on Send
2. Enter an address or swap between accounts
3. Enter a value in Token (Like ETH)
4. Press the ^v icon to change to USD as the input

### Acceptance Criteria:
- When a user clicks the swap arrow icon the new value should be the previously calculated conversion value for that currency (like it is on mobile video). The example is shown on the screenshots below
image

VSaric:
fixes: #13275, #12811

**Explanation:**  Fix switching between ETH and USD and handling invalid prop `currentTransaction`. 

**Steps to Reproduce**

1. Click on Send
2. Enter an address or swap between accounts
3. Enter a value in Token (Like ETH)
4. Press the ^v icon to change to USD as the input

**Screencast (Fix switching between ETH and USD):**

https://user-images.githubusercontent.com/92527393/157023398-42190e7d-a88d-41e1-b034-672b02bac790.mov

Invalid prop `currentTransaction` produces a console error on the transaction confirmation page:

Screenshot 2022-03-03 at 10 06 33

This console error is now handled.

### DIFF
--- a/ui/components/app/currency-input/currency-input.js
+++ b/ui/components/app/currency-input/currency-input.js
@@ -44,7 +44,7 @@ export default function CurrencyInput({
 
   const [isSwapped, setSwapped] = useState(false);
   const [newHexValue, setNewHexValue] = useState(hexValue);
-  const [newFeatureSecondary, setNewFeatureSecondary] = useState(
+  const [shouldDisplayFiat, setShouldDisplayFiat] = useState(
     featureSecondary,
   );
 
@@ -53,7 +53,7 @@ export default function CurrencyInput({
       return false;
     }
 
-    return Boolean(newFeatureSecondary);
+    return Boolean(shouldDisplayFiat);
   };
 
   const getDecimalValue = () => {
@@ -78,7 +78,7 @@ export default function CurrencyInput({
   const swap = async () => {
     await onPreferenceToggle();
     setSwapped(!isSwapped);
-    setNewFeatureSecondary(!newFeatureSecondary);
+    setShouldDisplayFiat(!shouldDisplayFiat);
   };
 
   const handleChange = async (newDecimalValue) => {
--- a/ui/components/app/currency-input/currency-input.js
+++ b/ui/components/app/currency-input/currency-input.js
@@ -79,7 +79,7 @@ export default function CurrencyInput({
     setShouldDisplayFiat(!shouldDisplayFiat);
   };
 
-  const handleChange = async (newDecimalValue) => {
+  const handleChange = (newDecimalValue) => {
     const hexValueNew = shouldUseFiat()
       ? getWeiHexFromDecimalValue({
           value: newDecimalValue,
@@ -95,7 +95,7 @@ export default function CurrencyInput({
         });
 
     setNewHexValue(hexValueNew);
-    await onChange(hexValueNew);
+    onChange(hexValueNew);
     setSwapped(!isSwapped);
   };
 
--- a/ui/components/app/currency-input/currency-input.js
+++ b/ui/components/app/currency-input/currency-input.js
@@ -44,7 +44,7 @@ export default function CurrencyInput({
 
   const [isSwapped, setSwapped] = useState(false);
   const [newHexValue, setNewHexValue] = useState(hexValue);
-  const [newfeatureSecondary, setNewFeatureSecondary] = useState(
+  const [newFeatureSecondary, setNewFeatureSecondary] = useState(
     featureSecondary,
   );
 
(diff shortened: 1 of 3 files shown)


ui/components/app/currency-input/currency-input.js
```diff
@@ -44,13 +44,16 @@ export default function CurrencyInput({
 
   const [isSwapped, setSwapped] = useState(false);
   const [newHexValue, setNewHexValue] = useState(hexValue);
+  const [newFeatureSecondary, setNewFeatureSecondary] = useState(
```
segun, line 47: Can you give a brief comment what this state does?

VSaric: @segun This state use `featureSecondary` prop to show what currency is displayed when user click on `Send` button. When user press the swap (^v) icon `newFeatureSecondary` changed and display the estimated value of the new input type.

ui/components/app/currency-input/currency-input.js
```diff
@@ -44,13 +44,16 @@ export default function CurrencyInput({
...
+    featureSecondary,
+  );
 
   const shouldUseFiat = () => {
     if (hideSecondary) {
       return false;
     }
 
-    return Boolean(featureSecondary);
+    return Boolean(newFeatureSecondary);
```
segun, line 56: I think it's the name that's confusing me. In this place `newFeatureSecondary` is used to show either fiat or crypto, but the name doesn't reflect anything like that.

VSaric: @segun [Renamed](https://github.com/MetaMask/metamask-extension/pull/13827/commits/9ea5bb73f19d162b2ffbf0d2764aa14ff2d448e1) as we agreed!

segun (dismissed):
LGTM

ui/components/app/currency-input/currency-input.js
```diff
@@ -101,17 +95,20 @@ export default function CurrencyInput({
         });
 
     setNewHexValue(hexValueNew);
-    setDecimalValue(newDecimalValue);
-    onChange(hexValueNew);
+    await onChange(hexValueNew);
```
brad-decker, line 98: Can you help me understand why we are awaiting here? While there's nothing technically wrong with async event handlers, having  multiple that are passed between components seems like a great way to have unintended side effects such as memory leaks or attempts to update state in unmounted components. I did a fairly quick look into  this and it seems the only place this component is even used is in the UserPreferencedCurrencyInput component, which doesn't appear to ever be given an onChange callback. I could be wrong, so try and investigate and see if you can find where its used.

brad-decker (changes requested):

ui/components/app/currency-input/currency-input.js
VSaric: @brad-decker Thank's for this catch! I investigate and searched in code where this component is used, and you are right, it is used only in `UserPreferencedCurrencyInput`. Because there is no attempts to update state in component  I removed `async/await` type and check if everything is working fine when switching between ETH and USD.
You can see changes [here](https://github.com/MetaMask/metamask-extension/pull/13827/commits/5890e75fad25cd7bb54f029bfcdc22b9c0baf3c7)

segun (dismissed):
LGTM

VSaric:
@segun @brad-decker Can I get re-review?

ui/components/app/currency-input/currency-input.js
```diff
@@ -44,13 +44,14 @@ export default function CurrencyInput({
 
   const [isSwapped, setSwapped] = useState(false);
   const [newHexValue, setNewHexValue] = useState(hexValue);
+  const [shouldDisplayFiat, setShouldDisplayFiat] = useState(featureSecondary);
 
   const shouldUseFiat = () => {
```
darkwing, line 49: I know this function existed before now but does `shouldUseFiat` need to be a function?  Can it just be a ternary?

VSaric: @darkwing Changed `shouldUseFiat` function to be a ternary [here](https://github.com/MetaMask/metamask-extension/pull/13827/commits/0d2eba062b090bef6d230583d0568efe8885ce83).

VSaric:
Can I get re-review on this? @segun @brad-decker @darkwing

segun (approved):
LGTM!

darkwing (approved):

brad-decker (approved):

Issue #10461, opened 2021-02-16
Require descriptions for all new localized messages


Gudahtt:
We should require descriptions for all new localized messages. Descriptions help provide context for what each message is used for, and would be of great help to translators. It would also help ensure developers don't use re-use messages in contexts in which they might not be appropriate.

PR #13833, opened 2022-03-03
Fix v10.11: Unable to find value of key "eth_accounts" for locale "en" in Permissions Modal

digiwand:
**Fixes:** [13842](https://github.com/MetaMask/metamask-extension/issues/13842)
Fixes an issue found in v10.11

**Explanation:** 
Unable to find value of key "eth_accounts" for locale "en". 

**What is in this PR:**
- convert ConnectedAccountsPermissions to a functional component
- extract logic from PermissionsConnectPermissionsList into a new custom hook, `usePermissionDescriptions`
- use `usePermissionDescriptions` for both ConnectedAccountsPermissions and PermissionsConnectPermissionsList

**Actual:**
Error: Unable to find value of key "eth_accounts" for locale "en"
No description next to the checkbox, under the title Permissions

**Expected:**
Should display “See address, account balance, activity and initiate transactions” in the connected accounts permissions popover, next to the checkbox

**Manual Testing Steps:**
- Connect to the test dapp https://metamask.github.io/test-dapp/
- Open MetaMask Extension
- Click "Connect" 
- Expand "Permissions"

https://user-images.githubusercontent.com/20778143/156645726-2deec90f-5b22-45f4-a1a7-ad0871a57540.mov

image


ui/components/app/connected-accounts-permissions/connected-accounts-permissions.component.js
```diff
@@ -64,20 +64,25 @@ export default class ConnectedAccountsPermissions extends PureComponent {
...
-                  disabled
-                  id={permissionName}
-                  className="connected-accounts-permissions__checkbox"
-                />
-                <label htmlFor={permissionName}>{t(permissionName)}</label>
-              </li>
-            ))}
+            {permissions.map(({ key }) => {
+              const permissionName =
+                key === 'eth_accounts' ? 'permission_ethereumAccounts' : key;
```
digiwand, line 69: I kept the arbitrary string(s) here for now. We might want to create reusable constant(s) here. 

We have constants for "RestrictedMethods" and I would need more time to understand the correlation between permissions and the related restricted methods.

app/_locales/de/messages.json
```diff
@@ -2061,7 +2061,7 @@
   },
   "permission_ethereumAccounts": {
     "message": "Siehe Adresse, Kontostand, Aktivität und Einleitung von Transaktionen",
-    "description": "The description for the `eth_accounts` permission"
+    "description": "The description for the `permission_ethereumAccounts` permission"
```
mcmire, line 2064: [nit] I don't think these messages need descriptions. We typically use descriptions to give some context around placeholders (`$1`, `$2`, etc.) that we place in messages, but these messages don't have placeholders, so I think the description is redundant and you could remove them if you wanted to.

mcmire (dismissed):
This makes sense to me.

app/_locales/de/messages.json
Gudahtt: The description field should be encouraged for all messages, because they provide more context for translators on how the message is used. There is an open issue for requiring them for all new messages: https://github.com/MetaMask/metamask-extension/issues/10461

Gudahtt: Though in this case, this description is not being added in this PR. It looks like it wasn't even meant to be changed, and that this was done by accident. This is what's causing the lint failure on CI.

The permission is still called `eth_accounts`, so the old version is correct.

ui/components/app/connected-accounts-permissions/connected-accounts-permissions.component.js
Gudahtt, line 69: It looks like this only fixes the localized message for the `eth_accounts` permission. All of the other permissions would result in the same crash, because none of their descriptions match the permission name either.

ui/components/app/connected-accounts-permissions/connected-accounts-permissions.component.js
```diff
@@ -64,20 +64,25 @@ export default class ConnectedAccountsPermissions extends PureComponent {
...
+                  key={permissionName}
+                  className="connected-accounts-permissions__list-item"
+                >
+                  <CheckBox
+                    checked
+                    disabled
+                    id={permissionName}
+                    className="connected-accounts-permissions__checkbox"
+                  />
+                  <label htmlFor={permissionName}>{t(permissionName)}</label>
```
Gudahtt, line 82: Using a variable with the `t` function is discouraged, as it makes it difficult for our `verified-locales` lint script to determine whether or not the variable name is used. That script looks for string literals passed to `t` to ensure unused messages are detected and removed.

This is also probably why this mistake was introduced. Somebody searched for the old message key (`eth_accounts`) for any references to update, and this didn't show up in the search because that string isn't present here.

app/_locales/de/messages.json
mcmire: > The description field should be encouraged for all messages

Ah okay, good to know. Hadn't seen that issue.

ui/components/app/connected-accounts-permissions/connected-accounts-permissions.component.js
digiwand: That is good to know. Thank you! Working on creating a lookup table. First confirming we want to add flask support to this component...

digiwand:
I've reset to the base branch on my local machine to undo the current work here. Asking @rekmarks to see if we want to support Flask here

digiwand:
Confirmed we want to add flask support here. Spoke with @rekmarks. Wiped the current changes to work on: 
- converting `ui/components/app/connected-accounts-permissions/connected-accounts-permissions.component.js` to a FC
- create sharable permission key utility function
- create custom hook to lookup permission label.. or possible config 
- use new utility function and hook for both ConnectedAccountsPermissions and PermissionsConnectPermissionList components

ui/components/app/connected-accounts-permissions/connected-accounts-permissions.component.js
digiwand: After reading this and looking over more code, I realized that there was a system in `ui/components/app/permissions-connect-permission-list/permissions-connect-permission-list.js` that was retrieving permissions labels. The two files are now sharing the logic to retrieve labels. We preserved the `key`/`permissionName` logic.

ui/components/app/connected-accounts-permissions/connected-accounts-permissions.js
```diff
@@ -0,0 +1,83 @@
+import classnames from 'classnames';
```
jpuri, line 1: Hey @digiwand : going forward I would suggest avoid refactoring for code cleanup and bug fixing in same PR. This makes it confusing to review.

jpuri (dismissed):

DELJUJAZI (approved):

DELJUJAZI (dismissed):

ui/components/app/connected-accounts-permissions/connected-accounts-permissions.js
```diff
@@ -0,0 +1,83 @@
...
+import { usePermissionDescriptions } from '../../../hooks/usePermissionDescriptions';
+
+const ConnectedAccountsPermissions = ({ permissions }) => {
+  const t = useI18nContext();
+  const [expanded, setExpanded] = useState(false);
+
+  const getPermissionDescription = usePermissionDescriptions();
+
+  const toggleExpanded = () => {
+    setExpanded(!expanded);
```
Gudahtt, line 15: When state changes in React, it effectively asynchronous. The `expanded` state here is the one from the most recent render, so this can result in the expanded state only toggling _once_ if you click it twice, for example.

Generally when making any state changes that depend upon the previous state value, like a toggle, you should pass a function into the setter that operates on the previous state value. This ensures the behavior is always what we'd expect, avoiding that race condition case.

```
    setExpanded((_expanded) => !_expanded);
```

Gudahtt (dismissed):
LGTM! Couple of minor suggestions but this looks like a good solution overall.

ui/components/app/connected-accounts-permissions/connected-accounts-permissions.js
digiwand: I hear you and I agree. Thanks for the callout! Upon the first round of reviews yesterday, I thought to fix the specific i18n translation, we could simply change 1 line:
```
<label htmlFor={permissionName}>{t(permissionName)}</label>
```

to something like
```
<label htmlFor={permissionName}>
  {permissionName === 'eth_accounts' ? {t('permission_ethereumAccounts')} : t(permissionName)}
</label>
```

However, please correct me if I'm wrong, I believe we realised that this may be a bigger bug. The existing logic would not be able to handle permissions keys that don't map to translations and may break, particularly using flask. We added flask support. To do this, we realised converting the file to a FC and sharing the logic using custom hooks would be cleaner and take roughly the same amount of work as opposed to duplicating code here. @rekmarks would be able to shed more light on this

ui/components/app/connected-accounts-permissions/connected-accounts-permissions.js
digiwand: This makes sense. Thank you for the explanation!

Gudahtt (approved):

jpuri (approved):

ui/hooks/usePermissionDescriptions.js
```diff
@@ -0,0 +1,98 @@
...
+ * @property {string} rightIcon - The right icon.
+ */
+
+/**
+ * @returns {(permissionName:string) => PermissionLabelObject}
+ */
+export const usePermissionDescriptions = () => {
+  const t = useI18nContext();
+
+  return useMemo(() => {
```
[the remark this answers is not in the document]
digiwand: @Gudahtt 

I think I see. "`t` never changes" because it is the `useMemo` function itself. It's the return value from calling the function that may change, not `t` itself. 
`t` is currently:
```
  const t = useMemo(() => {
    return (key, ...args) =>
      getMessage(currentLocale, current, key, ...args) ||
      getMessage(currentLocale, en, key, ...args);
  }, [currentLocale, current, en]);
```
So let's say this is now a regular function in which we pass `t` and the permission name. If `currentLocale` changes, would these translations here update? or should we keep the hook to watch for `currentLocale`?

digiwand:
@mcmire @jpuri @Gudahtt 

Thanks for the reviews y'all and thanks @rekmarks for the pair programming sesh!

We might have a follow-up ticket to replace the new custom hook with a utility function https://github.com/MetaMask/metamask-extension/pull/13833#discussion_r819614766. 

For the sake of fixing the release bug, we can merge this in. I don't know the protocol to merge release fixes. I am thinking these are the next steps I would take:
- merge this in after @rekmarks has a final look
- inform #metamask-release-updates that a hotfix for `v10.11` has been merged in and tag @PeterYinusa and Mariona 
  - or inform a release manager (?) that can inform #metamask-release-updates

ui/hooks/usePermissionDescriptions.js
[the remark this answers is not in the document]
Gudahtt: Sorry you're right, `t` does change when the locale changes. I had forgotten that.

> If `currentLocale` changes, would these translations here update? or should we keep the hook to watch for `currentLocale`?

Maybe best explained by example. Right now `ConnectedAccountsPermissions` looks like this:

```javascript
const ConnectedAccountsPermissions = ({ permissions }) => {
  const t = useI18nContext();
  ...
  const getPermissionDescription = usePermissionDescriptions();

  return (
     ...
      <label htmlFor={permissionName}>
        {getPermissionDescription(permissionName).label}
      </label>
    ...
  );
};
```

If we were to create a regular function called `getPermissionDescription` to replace this hook, it could instead look like this:

```javascript
const ConnectedAccountsPermissions = ({ permissions }) => {
  const t = useI18nContext();
  ...

  return (
     ...
      <label htmlFor={permissionName}>
        {getPermissionDescription(t, permissionName).label}
      </label>
    ...
  );
};
```

`getPermissionDescription` is being called each render with the `t` obtained from the `useI18nContext` hook. That hook ensures that the `t` function always uses the correct locale.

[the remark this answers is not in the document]
digiwand: Thanks for the explanations and suggestions @Gudahtt! All looks good to me.

I'm creating a new PR now. Tiny concern - I noticed you suggested using singular `description` instead of `descriptions`... @rekmarks, and I, spoke about using `descriptions` vs. `description`. I think we were kinda in the middle about this, but we ended up using the plural version to represent multiple descriptions being returned. I'm not sure if you feel strongly about either-or, but please let me know if you do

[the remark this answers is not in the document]
digiwand: @Gudahtt @rekmarks 

on second thought... the function is a bit chunky so it might be nice to keep the memoization despite `t` as a dependency is negligible with `useMemo` here. any thoughts on this?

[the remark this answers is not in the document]
rekmarks: Haha, define "chunky"?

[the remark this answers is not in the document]
digiwand: Lolll just a bit of chunk 🤣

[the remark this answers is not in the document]
digiwand: For the record, the good kind of chunk though this is all very subjective. haha. On my machine, calling `getPermissionDescription(permissionName)` is really just about 0.025 ms on its first call. When memoized with `useMemo`, the time it takes reduces to about 0.005 ms, so we could save about 0.02 ms, or if there are 5 permissions in the list, 0.1 ms each time we rerender this list. Here, this method is called each time we click "Connect" and each time we expand or collapse the "Permissions" list. I guess this could add up to impact performance, but is likely negligible, by itself, to the user. I'm up for keeping it as a hook. After knowing this info, if we think changing this into util function will add simplicity and more value, I'll be happy to make the change too. 

initial call:
Screenshot 2022-03-04 at 10 40 15 PM

initial call and memoized calls:
Screenshot 2022-03-04 at 10 37 30 PM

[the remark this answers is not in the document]
Gudahtt: >we ended up using the plural version to represent multiple descriptions being returned

In that example code snippet, a single description is being returned.

>the function is a bit chunky so it might be nice to keep the memoization

The function is basically a switch statement and an object literal. I'm not sure that memoization is helping a whole lot. Cant draw any firm conclusions from those timings you posted because any function gets faster as it is executed repeatedly, because both SpiderMonkey and V8 optimize at runtime (JIT compilation). And even if we do want to memoize it, that can be done more efficiently using Lodash's `memoize` utility function. Basically it does the same thing, except it saves having to construct a new function definition and throw it away every render, which is what's happening with the `useMemo` hook.

PR #13836, opened 2022-03-04
Warn about multiple MetaMask instances running

### LINKED ISSUE #13407: Warn users about running multiple MetaMask
Having multiple MM running is not recommended. 

Options:

- [ ] Put a note in the flask onboarding page
- [ ] Use the extension messaging API and the immutable, platform-dependent extension IDs to communicate with other versions of MetaMask to detect and display a warning to the user
  - We can use the extension-provider as a reference implementation for how to use the messaging API:
https://github.com/MetaMask/extension-provider
  - We should store the IDs as enums somewhere. The IDs can be retrieved by installing the main extension and/or Flask from the extension store of a particular browser and just check what the ID is. The ID of every extension in the Chrome and Firefox stores are unique, permanent, and set by the browser vendors.

hivtushok:
Fixes: #13407

Explanation:  This PR introduces warnings for running multiple MetaMask instances simultaneously (Prod + Flask). 
To warn the user, a new text is added on the startup screen, and extensions have a simple ping-messaging set up in order to determine if another version is running. A console warning is fired in that case. 

**Startup screen:** (last line added)

image

Manual testing steps:  
  - 
  - 
  -


app/scripts/detect-multiple-instances.js
```diff
@@ -0,0 +1,44 @@
...
+ * mainline version of extension and Flask build
+ * in order to detect & warn if there are two different
+ * versions running simultaneously.
+ */
+
+import extension from 'extensionizer';
+
+// TODO: add IDs from other browsers
+const PROD_BUILD_ID = 'nkbihfbeogaeaoehlefnkodbefgpgknn';
+const FLASK_BUILD_ID = 'ljfoeinjpaedjfecbmggjgodbgkmjkjk';
```
rekmarks, line 12: We should move these to `shared/constants/app.js`.

app/scripts/detect-multiple-instances.js
```diff
@@ -0,0 +1,44 @@
...
+
+const MESSAGE_TEXT = 'isRunning';
+
+/**
+ * Handles the ping message sent from other extension.
+ */
+export const onMessageReceived = (message, sender, sendResponse) => {
+  if (message === MESSAGE_TEXT) {
+    sendResponse({
+      type: 'success',
```
FrederikBolding, line 22: Is there a specific reason for this response format? It doesn't look like the data is used anywhere.

app/scripts/detect-multiple-instances.js
```diff
@@ -0,0 +1,44 @@
...
+    return true;
+  }
+  return false;
+};
+
+/**
+ * Sends the ping message sent to other extension to detect whether it's active or not.
+ * Displays console warning if it's active.
+ */
+export const onExtensionConnect = () => {
```
FrederikBolding, line 34: Shouldn't this just run when the application loads and not `onConnect`? 

According to the docs it's trigged by:
> Fired when a connection is made from either an extension process or a content script (by [runtime.connect](https://developer.chrome.com/docs/extensions/reference/runtime/#method-connect)).

So with this current setup I assume yours is running when we connect to other parts of the MM app? 🤔

app/scripts/detect-multiple-instances.js
```diff
@@ -0,0 +1,44 @@
...
+  const idToPing =
+    extension.runtime.id === FLASK_BUILD_ID ? PROD_BUILD_ID : FLASK_BUILD_ID;
+
+  extension.runtime.sendMessage(idToPing, MESSAGE_TEXT, (response) => {
+    if (extension.runtime.lastError || !response) {
+      return;
+    }
+    console.warn('Warning! You have multiple instances of MetaMask running!');
```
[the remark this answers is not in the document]
hivtushok: The console warning was part of the requirements (https://github.com/MetaMask/metamask-extension/issues/13407)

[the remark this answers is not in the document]
FrederikBolding: Yep yep - just asking whether we want to show this in the UI as well?

app/scripts/detect-multiple-instances.js
hivtushok: I was thinking of checking for multiple instances every time extension is opened/expanded (after losing focus) – so this is used for that

app/scripts/detect-multiple-instances.js
hivtushok: Let me check if I can simplify that 👀

app/scripts/detect-multiple-instances.js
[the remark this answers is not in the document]
hivtushok: Yes, UI notification should be added too

app/scripts/detect-multiple-instances.js
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Handles the ping message sent from other extension.
+ */
+export const onMessageReceived = (message, sender, sendResponse) => {
+  if (message === MESSAGE_TEXT) {
+    sendResponse({
+      type: 'success',
+      isRunning: true,
+    });
+    return true;
```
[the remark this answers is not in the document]
hivtushok: Simplified 👍🏼

app/scripts/detect-multiple-instances.js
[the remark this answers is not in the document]
rekmarks: We should to the UI notification in a separate PR to keep this small.

app/scripts/detect-multiple-instances.js
hivtushok: Removed running a check on connection. Now it's startup-only. Also not responding to a ping message (but warning will be now printed only in one console instead of both).

FrederikBolding (dismissed):

shared/constants/app.js
```diff
@@ -70,3 +70,6 @@ export const POLLING_TOKEN_ENVIRONMENT_TYPES = {
   [ENVIRONMENT_TYPE_NOTIFICATION]: 'notificationGasPollTokens',
   [ENVIRONMENT_TYPE_FULLSCREEN]: 'fullScreenGasPollTokens',
 };
+
+export const METAMASK_PROD_BUILD_ID = 'nkbihfbeogaeaoehlefnkodbefgpgknn';
```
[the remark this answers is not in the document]
hivtushok: Since Flask is only supported in Chrome, this only works for Chrome. @rekmarks please, correct me if I'm wrong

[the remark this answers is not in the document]
hivtushok: However, I think it worth to update naming to reflect that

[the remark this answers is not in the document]
FrederikBolding: That's true, but it shouldn't hurt to add the FF ids as well for the future! Is there an easy way to detect whether we are in FF or Chrome @Gudahtt ?

app/scripts/detect-multiple-instances.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+};
+
+/**
+ * Sends the ping message sent to other extension to detect whether it's active or not.
+ * Displays console warning if it's active.
+ */
+export const checkForMultipleVersionsRunning = () => {
+  const idToPing =
+    browser.runtime.id === METAMASK_FLASK_BUILD_ID
```
Gudahtt, line 36: If the current ID matches neither Flask nor prod (e.g. in a development environment), this will default to only messaging Flask. Instead it might be better to message _both_ in this case. Really if either is running, the warning applies.

app/scripts/detect-multiple-instances.js
```diff
@@ -0,0 +1,46 @@
...
+import browser from 'webextension-polyfill';
+import {
+  METAMASK_PROD_BUILD_ID,
+  METAMASK_FLASK_BUILD_ID,
+} from '../../shared/constants/app';
+
+const MESSAGE_TEXT = 'isRunning';
+
+const showWarning = () =>
+  console.warn('Warning! You have multiple instances of MetaMask running!');
```
Gudahtt, line 17: It would be better to alert the user in this case, in the UI. A console warning seems like a good start though, and the user-facing alert could be handled in a later PR.

FrederikBolding: afaik that is planned in a follow-up PR, @hivtushok wanted to keep the scope of this one smaller!

shared/constants/app.js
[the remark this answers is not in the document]
hivtushok: @FrederikBolding I'd rather don't do that in order not to create confusion for a code reader, but it's good idea to check if we're in Chrome before running this code

app/scripts/detect-multiple-instances.js
hivtushok: Good idea.

app/scripts/detect-multiple-instances.js
```diff
@@ -0,0 +1,46 @@
...
+ */
+export const checkForMultipleVersionsRunning = () => {
+  const idToPing =
+    browser.runtime.id === METAMASK_FLASK_BUILD_ID
+      ? METAMASK_PROD_BUILD_ID
+      : METAMASK_FLASK_BUILD_ID;
+
+  browser.runtime
+    .sendMessage(idToPing, MESSAGE_TEXT)
+    .then(() => {
```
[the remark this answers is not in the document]
hivtushok: I think we don't need this piece now

shared/constants/app.js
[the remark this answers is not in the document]
hivtushok: 👍🏼

shared/constants/app.js
```diff
@@ -70,3 +70,7 @@ export const POLLING_TOKEN_ENVIRONMENT_TYPES = {
   [ENVIRONMENT_TYPE_NOTIFICATION]: 'notificationGasPollTokens',
   [ENVIRONMENT_TYPE_FULLSCREEN]: 'fullScreenGasPollTokens',
 };
+
+export const METAMASK_PROD_CHROME_ID = 'nkbihfbeogaeaoehlefnkodbefgpgknn';
+export const METAMASK_FLASK_CHROME_ID = 'ljfoeinjpaedjfecbmggjgodbgkmjkjk';
```
Gudahtt, line 75: This is not the Firefox ID. On Firefox the IDs are set in the manifest, ours is here: https://github.com/MetaMask/metamask-extension/blob/develop/app/manifest/firefox.json#L4

Firefox uses a GUID generated on install as the actual hostname of the extension, so this GUID will be different on each install and thus useless for identification.

Gudahtt: We already have a Flask ID as well, here: https://github.com/MetaMask/metamask-extension/blob/develop/app/build-types/flask/manifest/firefox.json#L4

Gudahtt: Also beta, which is here: https://github.com/MetaMask/metamask-extension/blob/develop/app/build-types/beta/manifest/firefox.json#L4

The Chrome ID for the beta is `pbbkamfgmaedccnfkmjcofcecjhfgldn`

hivtushok: So this cannot work in Firefox, is that right?

hivtushok: Thanks, added Beta for Chrome

FrederikBolding: @hivtushok According to the docs it should work on FF, as it is valid to use the ID that the extension has provided for itself:
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/sendMessage#parameters

hivtushok: @Gudahtt do you know how can we extract the Firefox GUID to send messages? Or can we use the manifest ids for that?

FrederikBolding: We should be able to use the manifest ID according to the docs!

Gudahtt:
This looks great! I was considering suggesting that you just loop through the ID list but it seems you had the same idea first. 
The only thing missing is unit tests, the new code doesn't appear to be tested yet.

app/scripts/detect-multiple-instances.test.js
```diff
@@ -0,0 +1,58 @@
...
+  checkForMultipleVersionsRunning,
+  onMessageReceived,
+} from './detect-multiple-instances';
+import * as util from './lib/util';
+
+describe('multiple instances running detector', function () {
+  const PING_MESSAGE = 'isRunning';
+
+  it('should send ping message to multiple instances', async function () {
+    const sandbox = sinon.createSandbox();
```
Gudahtt, line 20: Generally we shouldn't use sandboxes. Sinon sandboxes are meant for cases where we want to manage multiple separate sets of mocks independently from each other. In all other cases, the default sandbox is easier and safer to use.

You should also call `sinon.restore()` in the `beforeEach` hook, to ensure that all stubs are properly cleaned up after each test. In addition to helping keep tests isolated, this is recommend in the Sinon general setup docs as a way to prevent memory leaks.

app/scripts/detect-multiple-instances.js
```diff
@@ -0,0 +1,55 @@
...
+    showWarning();
+  }
+};
+
+/**
+ * Sends the ping message sent to other extensions to detect whether it's active or not.
+ */
+export const checkForMultipleVersionsRunning = async () => {
+  if (getPlatform() !== PLATFORM_CHROME && getPlatform() !== PLATFORM_FIREFOX) {
+    return;
```
Gudahtt, line 39: This branch appears to be untested

app/scripts/detect-multiple-instances.js
```diff
@@ -0,0 +1,55 @@
...
+    getPlatform() === PLATFORM_CHROME ? CHROME_BUILD_IDS : FIREFOX_BUILD_IDS;
+
+  const thisBuild = browser.runtime.id;
+
+  for (const id of buildIds) {
+    if (id !== thisBuild) {
+      try {
+        await browser.runtime.sendMessage(id, MESSAGE_TEXT);
+      } catch (error) {
+        // Should do nothing if receiving end was not reached (no other instances running)
```
Gudahtt, line 51: This branch appears to be untested

app/scripts/detect-multiple-instances.js
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Handles the ping message sent from other extension.
+ * Displays console warning if it's active.
+ *
+ * @param message - The message received from the other extension
+ */
+export const onMessageReceived = (message) => {
+  if (message === MESSAGE_TEXT) {
+    showWarning();
+  }
```
Gudahtt, line 31: The `else` branch appears to be untested

app/scripts/metamask-controller.js
```diff
@@ -1061,6 +1065,11 @@ export default class MetamaskController extends EventEmitter {
 
     // TODO:LegacyProvider: Delete
     this.publicConfigStore = this.createPublicConfigStore();
+
+    // Multiple MetaMask instances launched warning
+    this.extension.runtime.onMessageExternal.addListener(onMessageReceived);
```
Gudahtt, line 1070: This is probably challenging to test, but it would be great to confirm this is wired up correctly in the `metamask-controller` tests.

app/scripts/detect-multiple-instances.test.js
hivtushok: Good point!

app/scripts/metamask-controller.js
hivtushok: Hmm not sure how valuable such test would be, but could you suggest the best way of looking into listeners to verify?

app/scripts/detect-multiple-instances.test.js
Gudahtt: It looks like you're still using a sandbox. The "default sandbox" refers to using the global `sinon` instance directly. No use of `createSandbox` at all.

hivtushok: Done! Using sanbox now only for `metamask-controller.test.js` – where it's already used

Gudahtt: Perfect, thanks!

app/scripts/metamask-controller.js
Gudahtt: I didn't get time to look into this, but I agree that maybe the benefit is minimal given how things are currently setup. We can tackle this later.

app/scripts/detect-multiple-instances.test.js
```diff
@@ -0,0 +1,104 @@
...
+      sinon.replace(browser, 'runtime', {
+        sendMessage: sinon.stub().throws(),
+        id: METAMASK_BETA_CHROME_ID,
+      });
+
+      const spy = sinon.spy(checkForMultipleVersionsRunning);
+
+      await checkForMultipleVersionsRunning();
+
+      assert(!spy.threw());
```
Gudahtt, line 79: :sweat_smile: is this just to pass the lint rule about each test needing an assertion? This is a bit silly because this would have thrown on line 77. 

The better way to do this would probably be to use Jest's `expect(...).not.toThrow()` matcher. But I'm guessing this file is still running on mocha - it seems to be anyway. I should have suggested using Jest instead but I won't block on it.

app/scripts/detect-multiple-instances.js
```diff
@@ -0,0 +1,55 @@
...
+  const buildIds =
+    getPlatform() === PLATFORM_CHROME ? CHROME_BUILD_IDS : FIREFOX_BUILD_IDS;
+
+  const thisBuild = browser.runtime.id;
+
+  for (const id of buildIds) {
+    if (id !== thisBuild) {
+      try {
+        await browser.runtime.sendMessage(id, MESSAGE_TEXT);
+      } catch (error) {
```
Gudahtt, line 50: Hmm. I'm surprised the linter isn't complaining about this unused variable.

Gudahtt: In general, not checking the `error` type is a bad habit. We're assuming here that if this error is thrown, the receiving end does not exist. But we're not confirming that that's what this error is. It could be some unexpected error instead, indicating a bug somewhere in our code.

The practice I'd generally recommend is to test all expected flows, and ensure the thrown error is one of the expected ones. If not, re-throw it and we'll end up seeing it in Sentry and we can address it from there (either by fixing the bug or adding a new variant of expected error to the list).

Gudahtt (dismissed):
LGTM! I left a few comments about how the tests and error handling could be improved, but they didn't seem like severe enough issues to warrant blocking the PR.

FrederikBolding (dismissed):

app/scripts/detect-multiple-instances.test.js
hivtushok: Yes, this one is running on mocha, so I couldn't use more suitable matcher 🙂 
The idea was to simulate throw on `sendMessage` and verify that it was handled inside and not exposed out, so this method is fail-safe

app/scripts/detect-multiple-instances.js
hivtushok: That's all totally reasonable. I did it this way to make this method secure from failing some other flows, since this functionality is pure informative, not operational. I think, adding logging of errors would be a useful improvement!

Gudahtt: Hmm. Logging could be useful, but I'm not sure about doing so indiscriminately. This is _expected_ to fail in most scenarios. Our startup logs are pretty concise right now so I don't think this would go unnoticed either. Also this message sounds like it's saying something went _wrong_.

hivtushok: Yeah, you're right, that's probably too much logging. Ok, postponing this for now

Gudahtt (approved):

rekmarks (approved):

PR #13838, opened 2022-03-04
Showing new message while converting token to NFT

### LINKED ISSUE #13837: 1. We need to update the UX for the NFT token displayed in Assets tab.Currently if a user clicks on the token in the assets tab, a message is displayed asking user to import it as an NFT. This is confusing if a user has already added that NFT. We should update the message suggesting that NFT has already been imported and this token will disappear from the asset list. See the Figma file.

NiranjanaBinoy:
Fixes: #13837 

If a user clicks on the token in the assets tab and the same NFT is already imported via the import NFT page, then a new copy is displayed on the modal, and upon clicking `yes` the entry from the Asset tab will be deleted and the user will be redirected to the NFT first entry with the token Address.

https://user-images.githubusercontent.com/43930900/156698180-cc0fa0ba-2187-42aa-a0cc-f017f16421db.mov

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -635,7 +635,7 @@
     "message": "We've detected that this asset is an NFT. Metamask now has full native support for NFTs. Would you like to remove it from your token list and add it as an NFT?"
   },
   "convertTokenToNFTExistDescription": {
-    "message": "We’ve detected that this asset is an NFT and has been added as an NFT. Would you like to remove it from your token list?"
+    "message": "We’ve detected that this asset has been added as an NFT. Would you like to remove it from your token list?"
   },
   "copiedExclamation": {
     "message": "Copied!"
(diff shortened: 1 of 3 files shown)


SayaGT:
Looks good to me! 👍

jpuri (dismissed):

app/_locales/en/messages.json
```diff
@@ -634,6 +634,9 @@
   "convertTokenToNFTDescription": {
     "message": "We've detected that this asset is an NFT. Metamask now has full native support for NFTs. Would you like to remove it from your token list and add it as an NFT?"
   },
+  "convertTokenToNFTExistDescription": {
+    "message": "We’ve detected that this asset is an NFT and has been added as an NFT. Would you like to remove it from your token list?"
```
adonesky1, line 638: [nit]
```suggestion
    "message": "We’ve detected that this asset has been added as an NFT. Would you like to remove it from your token list?"
```

adonesky1 (dismissed):
Verbiage nit otherwise LGTM!

0ex-d (dismissed):
Clean solution!

app/_locales/en/messages.json
ThatEazy: Yes

brad-decker (approved):

adonesky1 (approved):
LGTM

PR #13839, opened 2022-03-04
LavaMoat@6

kumavis:
update build process to lavamoat@6 with new `'a'a` secure package naming

`lavamoat-browserify` upgrade to follow


kumavis:
happy to see tests passing, keeping as draft to due to perf issue. investigating

package.json
```diff
@@ -328,7 +328,7 @@
     "jest-canvas-mock": "^2.3.1",
     "jsdom": "^11.2.0",
     "koa": "^2.7.0",
-    "lavamoat": "^5.3.5",
+    "lavamoat": "6.1.0",
```
Gudahtt, line 331: Was this meant to be pinned on this specific version?

```suggestion
    "lavamoat": "^6.1.0",
```

package.json
```diff
@@ -358,7 +358,7 @@
     "semver": "^7.3.5",
     "serve-handler": "^6.1.2",
     "sinon": "^9.0.0",
-    "source-map": "^0.7.2",
+    "source-map": "^0.6.1",
```
Gudahtt, line 361: Why was this downgraded?

package.json
```diff
@@ -385,18 +385,12 @@
   },
   "lavamoat": {
     "allowScripts": {
-      "3box>3box-orbitdb-plugins>ipfs-log>orbit-db-identity-provider>orbit-db-keystore>leveldown": false,
```
Gudahtt, line 388: What motivated these changes?

package.json
kumavis: no

package.json
kumavis: hmm i dont remember removing this

package.json
kumavis: weird - source-map was last changed 4 years ago, not sure how this got in here

kumavis:
changing `source-map` back led to uncovering a resolution bug in lavamoat, fixed and updated

lavamoat/build-system/policy.json
```diff
@@ -1,12 +1,26 @@
 {
   "resources": {
+    "3box>ipfs>fsm-event>fsm>split": {
```
Gudahtt, line 3: I'm surprised to see 3box here. We don't use it in the build system.

Gudahtt: Seems like a bug. We do use `split` in the build system, but this path is incorrect.

kumavis: the name used is the shortest dependency path of all dependency paths

Gudahtt: Hmm.... if this path doesn't represent the actual path used in the bundle, what is its purpose? i.e. how does it improve security.

package.json
```diff
@@ -328,7 +328,7 @@
     "jest-canvas-mock": "^2.3.1",
     "jsdom": "^11.2.0",
     "koa": "^2.7.0",
-    "lavamoat": "^5.3.5",
+    "lavamoat": "^6.1.1",
```
mcmire, line 331: What new changes are in this release? Does this now rely on `@lavamoat/aa`? Are there other improvements we should know about? If you have a secret changelog then I'd love to see it :)

kumavis: some highlights
1. policy's use package paths
2. policy override is only used when generated the policy, so the policy file is exactly what gets used
3. packages added in the override are walked automatically for policy (though this has been observed not working in this metamask branch)

lavamoat/build-system/policy.json
kumavis: @Gudahtt lavamoat doesn't look at versions specifically, the package manager does the deduplication. lavamoat will load the dependency graph and find the multiple paths pointing at the same location on disk. it will use the shortest name based on the dependency paths for that disk location.

Gudahtt: I see, so it is in effect listing one entry per package install location. And it uses the shortest dep path for each location as the identifier.

package.json
mcmire: Cool, thanks!

package.json
Gudahtt: I see - this change was caused by `@lavamoat/allow-scripts@2.0.1`. Any idea why that patch would have caused these changes?

Gudahtt: Oh, that patch included the introduction of `@lavamoat/aa`, which totally changed the identifier used for each package. Hmm. Surprising that you'd publish that as a patch, that seems like a breaking change. It causes install to fail unless the config is updated.

kumavis: should have been added in `@lavamoat/allow-scripts@2.0.0`

mcmire (dismissed):

kumavis:
trying on a larger resource class

mcmire (approved):

Gudahtt (approved):
LGTM!

kumavis:
migrating to cleaner branch on https://github.com/MetaMask/metamask-extension/pull/14488

PR #13844, opened 2022-03-04
Update snap install warning per designs and fix for popover component

hmalik88:
With custom CSS ability merged for `Popover`, I decided to update `SnapInstallWarning` to use these new props and have it in line with its [design](https://www.figma.com/file/nBKUQx0btfQiqVf30UA6KD/Flask-MVP?node-id=336%3A6344). This PR also addresses the fact that `headerProps`, `contentProps`, `footerProps` being passed to the `Popover` component were being overwritten because of selector specificity. I moved whatever CSS rules that were conflicting with the `Box` component's props out of the `.popover-header`, `.popover-content` and `.popover-footer` classes and into the `Popover` component as `defaultHeaderProps`, `defaultContentProps` and `defaultFooterProps`.

**Thoughts:** In removing some of these CSS rules from `Popover`, it made me think how much we want to allow customizing a component? I think allowing for `Box` props for these containers is cool since having some inline styling would be better than none. When we start looking to customize header, content and footer's children then that might be a bit too flexible...I suppose with design there needs to be _some_ rigidity.

**Before:** 

**After:**

Screen Shot 2022-03-04 at 5 14 33 PM

Manual testing steps:  
  - Run `yarn start --build-type flask`
  - To test a snap with the key access permissions, you can clone our fork of the filecoin snap at: [Filecoin snap](https://github.com/MetaMask/filsnap/tree/prerelease)
  - Checkout the `prerelease` branch and then run `yarn && yarn demo`
  - Follow the install flow for the filecoin snap from the locally hosted dapp and observe the changes when the snap install warning pops up

### DIFF
--- a/ui/components/ui/popover/popover.component.js
+++ b/ui/components/ui/popover/popover.component.js
@@ -29,7 +29,7 @@ const defaultContentProps = {
 const defaultFooterProps = {
   display: 'flex',
   justifyContent: JUSTIFY_CONTENT.SPACE_BETWEEN,
-  padding: [4, 6],
+  padding: [4, 6, 6],
 };
 
 const Popover = ({
@@ -114,7 +114,7 @@ const Popover = ({
         {footer ? (
           <Box
             className={classnames('popover-footer', footerClassName)}
-            {...{ ...defaultFooterProps, footerProps }}
+            {...{ ...defaultFooterProps, ...footerProps }}
           >
             {footer}
           </Box>
--- a/ui/components/ui/popover/popover.component.js
+++ b/ui/components/ui/popover/popover.component.js
@@ -4,6 +4,33 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import { useI18nContext } from '../../../hooks/useI18nContext';
 import Box from '../box';
+import {
+  ALIGN_ITEMS,
+  FLEX_DIRECTION,
+  JUSTIFY_CONTENT,
+} from '../../../helpers/constants/design-system';
+
+const defaultHeaderProps = {
+  padding: [6, 4, 4],
+  display: 'flex',
+  flexDirection: FLEX_DIRECTION.COLUMN,
+  backgroundColor: 'white',
+  borderRadius: 'xl',
+};
+
+const defaultContentProps = {
+  display: 'flex',
+  flexDirection: FLEX_DIRECTION.COLUMN,
+  justifyContent: JUSTIFY_CONTENT.FLEX_START,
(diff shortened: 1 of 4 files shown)


ui/components/ui/popover/popover.component.js
```diff
@@ -19,15 +46,18 @@ const Popover = ({
...
+  contentProps = defaultContentProps,
+  footerProps = defaultFooterProps,
 }) => {
   const t = useI18nContext();
   const showHeader = title || onBack || subtitle || onClose;
   const Header = () => {
     return (
-      <Box {...headerProps} className="popover-header">
+      <Box
+        {...{ ...defaultHeaderProps, ...headerProps }}
```
mcmire, line 58: [nit] You shouldn't have to spread the spread; once should be enough:

```suggestion
        { ...defaultHeaderProps, ...headerProps }
```

ui/components/ui/popover/popover.component.js
```diff
@@ -76,15 +106,15 @@ const Popover = ({
         {children ? (
           <Box
             className={classnames('popover-content', contentClassName)}
-            {...contentProps}
+            {...{ ...defaultContentProps, ...contentProps }}
```
mcmire, line 109: Same for this

``` suggestion
            { ...defaultContentProps, ...contentProps }
```

ui/components/ui/popover/popover.component.js
```diff
@@ -76,15 +106,15 @@ const Popover = ({
...
           >
             {children}
           </Box>
         ) : null}
         {footer ? (
           <Box
             className={classnames('popover-footer', footerClassName)}
-            {...footerProps}
+            {...{ ...defaultFooterProps, ...footerProps }}
```
mcmire, line 117: And this:

``` suggestion
            {...{ ...defaultFooterProps, ...footerProps }}
```

mcmire (commented):
Looks good. Would you mind posting screenshots of before and after so it's easier to see the visual changes?

ui/components/ui/popover/popover.component.js
hmalik88: That doesn't work, it gives an "expression expected" error.

hmalik88: That doesn't work, it gives an "expression expected" error.

hmalik88:
> Looks good. Would you mind posting screenshots of before and after so it's easier to see the visual changes?

Just updated with some screenshots, lmk if you need anything else to help review!

jpuri (approved):

hivtushok (approved):
LGTM

PR #13847, opened 2022-03-04
Fixed serialization of error parameter for `rejectPendingApproval` across JSON-RPC

### LINKED ISSUE #11383: window opened by `wallet_switchEthereumChain` does not respond to input
**Describe the bug**
*Occasionally* when I try to switch networks both action buttons do nothing. 

Unable to reliably reproduce

**Expected behavior**
Cancel should dismiss the window
Confirm should switch networks

**Screenshots**

https://user-images.githubusercontent.com/5773490/123332510-95896500-d50e-11eb-8490-23c8b88c47da.mov

 - OS: macOS Big Sur
 - Browser: Firefox 89.0.1 (64-bit)
 - MetaMask Version: 9.6.1

image

shanejonas:
Fixes: #11383

Explanation:  
Looks like sending `Error` objects across postMessage on firefox kills the `port` connection, this properly serializes the error which is a parameter to `rejectPendingApproval(id, error)` on both sides

I'm sure this isn't the only place where we do this that needs to be fixed.

### DIFF
--- a/ui/pages/confirmation/templates/switch-ethereum-chain.js
+++ b/ui/pages/confirmation/templates/switch-ethereum-chain.js
@@ -83,7 +83,7 @@ function getValues(pendingApproval, t, actions) {
     onCancel: () =>
       actions.rejectPendingApproval(
         pendingApproval.id,
-        ethErrors.provider.userRejectedRequest(),
+        ethErrors.provider.userRejectedRequest().serialize(),
       ),
     networkDisplay: true,
   };
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -9,7 +9,11 @@ import createFilterMiddleware from 'eth-json-rpc-filters';
 import createSubscriptionManager from 'eth-json-rpc-filters/subscriptionManager';
 import { providerAsMiddleware } from 'eth-json-rpc-middleware';
 import KeyringController from 'eth-keyring-controller';
-import { errorCodes as rpcErrorCodes, EthereumRpcError, ethErrors } from 'eth-rpc-errors';
+import {
+  errorCodes as rpcErrorCodes,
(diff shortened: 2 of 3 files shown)


Gtonizuka:
@shanejonas this looks good to me - do we want to apply these changes to other parts of the codebase as part of this PR?

Gudahtt:
Assuming that it's true that sending an Error terminates the session, we should be serializing this in the RPC handler code so that it applies to all actions.

brad-decker (dismissed):

Gudahtt:
>Assuming that it's true that sending an Error terminates the session, we should be serializing this in the RPC handler code so that it applies to all actions.

I misunderstood, I thought this was being thrown but it's being passed as a value. Nevermind, it's on us to ensure the API accepts only JSON.

Though.... is this ever called with something other than "user rejected request"? Maybe we can assume that will be the error rather than serializing it and passing it through.

ui/pages/confirmation/templates/switch-ethereum-chain.js
```diff
@@ -83,7 +83,7 @@ function getValues(pendingApproval, t, actions) {
     onCancel: () =>
       actions.rejectPendingApproval(
         pendingApproval.id,
-        ethErrors.provider.userRejectedRequest(),
+        ethErrors.provider.userRejectedRequest().serialize(),
```
Gudahtt, line 86: This is called in `add-ethereum-chain.js` as well. So if we're still going the route of sending the error from the UI, it'll need to be serialized there as well. Presumably that confirmation suffers from the same problem.

Gudahtt (approved):
LGTM! Though we should consider embedding the error directly in the background in the future, as I think that would be simpler.

brad-decker (approved):

PR #13851, opened 2022-03-07
Add browser recommendation to README

asaneh:
##@We have recently dropped support for certain older browsers, and we're
planning to have a larger conversation soon about which browsers to
support going forward. In preparation for this, it might be worth
recommending that users use the latest browser version.

Fixes: #

Explanation:  

Manual testing steps:  
  - 
  - 
  -


Gudahtt:
Please stop spamming our repositories

PR #13858, opened 2022-03-07
add asset_type to transactions

### LINKED ISSUE #13423: Add is_base_asset property to all transaction events in transaction funnel
### Description
Product and business leaders would like to get a sense of how often MetaMask users send transactions of the base asset for the network. Essentially this means tracking any received Ether, BNB, etc but not ERC-20, ERC-721, etc. To do this we will need to modify the events in the TransactionController such that a new property is added to the event called `is_base_asset`

#### Requirements
1. Modify the buildEventFragmentProperties method in the TransactionController to use the method created in #13422 to determine if it is a simple send.
2. if the transaction is a simple send, set is_base_asset to true (default to false).

brad-decker:
Fixes: #13423

### DIFF
--- a/shared/constants/transaction.js
+++ b/shared/constants/transaction.js
@@ -310,7 +310,9 @@ export const TRANSACTION_EVENTS = {
  * The types of assets that a user can send
  * 1. NATIVE - The native asset for the current network, such as ETH
  * 2. TOKEN - An ERC20 token.
- * 2. COLLECTIBLE - An ERC721 or ERC1155 token.
+ * 3. COLLECTIBLE - An ERC721 or ERC1155 token.
+ * 4. UNKNOWN - A transaction interacting with a contract that isn't a token
+ *  method interaction will be marked as dealing with an unknown asset type.
  */
 export const ASSET_TYPES = {
   NATIVE: 'NATIVE',
--- a/shared/constants/transaction.js
+++ b/shared/constants/transaction.js
@@ -305,3 +305,16 @@ export const TRANSACTION_EVENTS = {
   REJECTED: 'Transaction Rejected',
   SUBMITTED: 'Transaction Submitted',
 };
+
+/**
+ * The types of assets that a user can send
+ * 1. NATIVE - The native asset for the current network, such as ETH
+ * 2. TOKEN - An ERC20 token.
+ * 2. COLLECTIBLE - An ERC721 or ERC1155 token.
+ */
+export const ASSET_TYPES = {
+  NATIVE: 'NATIVE',
+  TOKEN: 'TOKEN',
+  COLLECTIBLE: 'COLLECTIBLE',
+  UNKNOWN: 'UNKNOWN',
+};
--- a/shared/modules/transaction.utils.js
+++ b/shared/modules/transaction.utils.js
@@ -169,18 +169,26 @@ export async function determineTransactionAssetType(
     inferrableType = result.type;
   }
 
+  // If the inferred type of the transaction is one of those that are part of
+  // the token contract standards, we can use the getTokenStandardAndDetails
+  // method to get the asset type.
   const isTokenMethod = [
     TRANSACTION_TYPES.TOKEN_METHOD_APPROVE,
     TRANSACTION_TYPES.TOKEN_METHOD_TRANSFER,
     TRANSACTION_TYPES.TOKEN_METHOD_TRANSFER_FROM,
   ].find((methodName) => isEqualCaseInsensitive(methodName, inferrableType));
 
-  if (isTokenMethod) {
+  if (
+    isTokenMethod ||
+    // We can also check any contract interaction type to see if the to address
+    // is a token contract. If it isn't, then the method will throw and we can
+    // fall through to the other checks.
+    inferrableType === TRANSACTION_TYPES.CONTRACT_INTERACTION
+  ) {
     try {
-      const details = await getTokenStandardAndDetails(
-        txMeta.txParams.to,
-        txMeta.txParams.from,
-      );
+      // We don't need a balance check, so the second parameter to
+      // getTokenStandardAndDetails is omitted.
+      const details = await getTokenStandardAndDetails(txMeta.txParams.to);
       if (details.standard) {
         return {
           assetType:
@@ -191,7 +199,8 @@ export async function determineTransactionAssetType(
         };
       }
     } catch {
-      // noop
+      // noop, We expect errors here but we don't need to report them or do
+      // anything in response.
     }
   }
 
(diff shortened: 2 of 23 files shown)


shared/constants/transaction.js
```diff
@@ -305,3 +305,16 @@ export const TRANSACTION_EVENTS = {
   REJECTED: 'Transaction Rejected',
   SUBMITTED: 'Transaction Submitted',
 };
+
+/**
+ * The types of assets that a user can send
+ * 1. NATIVE - The native asset for the current network, such as ETH
+ * 2. TOKEN - An ERC20 token.
+ * 2. COLLECTIBLE - An ERC721 or ERC1155 token.
+ */
```
brad-decker, line 314: Document the UNKNOWN entry

shared/modules/transaction.utils.js
```diff
@@ -91,6 +92,8 @@ export function txParamsAreDappSuggested(transaction) {
  *
  * @param {Object} txParams - Parameters for the transaction
  * @param {EthQuery} query - EthQuery instance
+ * @param {Function} getTokenStandardAndDetails - Function to fetch token
```
brad-decker, line 95: Remove

shared/modules/transaction.utils.js
```diff
@@ -131,3 +134,47 @@ export async function determineTransactionType(txParams, query) {
...
+  query,
+  getTokenStandardAndDetails,
+) {
+  // Because we will deal with all types of transactions (including swaps)
+  // we want to get an inferrable type of transaction that isn't special cased
+  // that way we can narrow the number of logic gates required.
+  const { type: inferrableType } = await determineTransactionType(
+    txMeta.txParams,
+    query,
+  );
```
brad-decker, line 149: @dan437 @danjm -- should we do something else for swaps to determine the asset_type? If we are recording the asset_type, how can we get the asset the user is swapping *from*

shared/modules/transaction.utils.js
```diff
@@ -131,3 +134,47 @@ export async function determineTransactionType(txParams, query) {
...
+
+  if (isTokenMethod) {
+    const details = await getTokenStandardAndDetails(
+      txMeta.txParams.to,
+      txMeta.txParams.from,
+    );
+
+    if (details.standard) {
+      return {
+        assetType: (details.standard = TOKEN_STANDARDS.ERC20
```
brad-decker, line 165: Accidental assignment instead of equality check

shared/modules/transaction.utils.js
```diff
@@ -131,3 +132,79 @@ export async function determineTransactionType(txParams, query) {
...
+    try {
+      details = await getTokenStandardAndDetails(
+        txMeta.txParams.to,
+        txMeta.txParams.from,
+      );
+    } catch {
+      // noop
+    }
+
+    if (details.standard) {
```
darkwing, line 189: Since we wont have `details.standard` unless the `try` succeeds, should we move this block into the `try`?  Especially if our catch does nothing?

darkwing:
This PR looks awesome, just had one question.

shared/modules/transaction.utils.js
```diff
@@ -131,3 +132,79 @@ export async function determineTransactionType(txParams, query) {
...
+    inferrableType = result.type;
+  }
+
+  const isTokenMethod = [
+    TRANSACTION_TYPES.TOKEN_METHOD_APPROVE,
+    TRANSACTION_TYPES.TOKEN_METHOD_TRANSFER,
+    TRANSACTION_TYPES.TOKEN_METHOD_TRANSFER_FROM,
+  ].find((methodName) => isEqualCaseInsensitive(methodName, inferrableType));
+
+  if (isTokenMethod) {
```
adonesky1, line 178: There are many possible cases where we could still get the `tokenStandard` where the particular method being called is not included in `isTokenMethod`. If a user calls any standard or non-standard method on an ERC20, ERC721 or ERC1155 token contract we can derive the assetType just by passing the tokenAddress (the second arg on getTokenStandardAndDetails should be optional and I will open a PR so that the function signature reflects that). `getTokenStandardAndDetails` will throw the error `Unable to determine contract standard` if the contract is not one of these asset types and then we can set it to unknown.

adonesky1: PR for the optional arg piece here: https://github.com/MetaMask/controllers/pull/717

brad-decker: @adonesky1 i'm confused by this comment. Are you suggesting there are other transaction types that are in fact transactions sending or receiving tokens? I know you suggested a new item for the array that we are looking for in isTokenMethod.

brad-decker: are you suggesting I could just call getTokenStandardAndDetails without first checking the transaction type?

adonesky1: Ok so running back through this the solution I'm proposing is simply to add `TRANSACTION_TYPES.CONTRACT_INTERACTION` to the `isTokenMethod` array. There are many methods on the token standards that move through the transaction controller that are not `transfer` and `transferFrom`, as just a quick example `DRIP` or `DRAIN` on this [contract](https://rinkeby.etherscan.io/token/0xf5de760f2e916647fd766b4ad9e85ff943ce3a2b?a=0x806627172aF48bD5B0765D3449A7DEf80d6576ff#writeContract). There will of course be contract interactions that are not with contracts conforming to the ERC721, ERC1155, ERC20 standards and for those `getTokenStandardAndDetails` will throw an error, `details.standard` will be undefined and we will fall through to the `if (inferrableType === TRANSACTION_TYPES.CONTRACT_INTERACTION) {` block on line 203. But we might as well check for the token standard on all contract interactions and add the standard to our metrics where we can?

brad-decker: Gotcha. Sounds good to me.

brad-decker: @adonesky1 fixed this in the latest, let me know what you think. Added a few more inline comments too

adonesky1: Looks like you've got to rebase and then 💯

brad-decker: Rebased @adonesky1

ui/helpers/constants/common.js
```diff
@@ -9,6 +9,13 @@ export const ERC20 = 'ERC20';
 export const ERC721 = 'ERC721';
 export const ERC1155 = 'ERC1155';
 
+export const TOKEN_STANDARDS = {
+  ERC20,
+  ERC721,
+  ERC1155,
+  NONE: 'NONE',
+};
```
adonesky1, line 17: Feels like these should eventually live in a module that can be shared across controllers and products. But obviously not a blocker here.

adonesky1 (dismissed):

adonesky1 (approved):

darkwing (approved):

PR #13868, opened 2022-03-08
Fix permission site origin overflow

### LINKED ISSUE #260: Snap Install: Long URL causes horizontal overflow
When installing a Snap, if the source site has a long URL, this will cause a horizontal scrollbar in the UI

long_url

FrederikBolding:
Fixes: https://github.com/MetaMask/snaps-skunkworks/issues/260

Explanation:  

Content would overflow when an origins URL was longer than could fit in the viewport. Was originally reported as a Snaps bug, but seems to be present in the regular connect flow as well.

**When working on this PR I also noticed other inconsistencies in the connect flow that I fixed.**

Manual testing steps:  
  - Go to https://deploy-preview-3--evm-chainlist.netlify.app/ (or some other site with a long URL) and connect
  - Make sure there is no overflow
  - Go to https://app.uniswap.org/ and connect
  - See that the "chip" shrinks
  - Go to https://filsnap.chainsafe.io/ and connect
  - See that there are no overflows

### DIFF
--- a/ui/components/app/permissions-connect-header/index.scss
+++ b/ui/components/app/permissions-connect-header/index.scss
@@ -2,6 +2,7 @@
   flex: 0;
 
   &__icon {
+    width: 100%;
     display: flex;
     flex-direction: column;
     align-items: center;
--- a/ui/components/app/permissions-connect-header/index.scss
+++ b/ui/components/app/permissions-connect-header/index.scss
@@ -1,5 +1,6 @@
 .permissions-connect-header {
   flex: 0;
+  width: 100%;
 
   &__icon {
     width: 100%;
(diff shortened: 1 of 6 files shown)


Gtonizuka:
@FrederikBolding I actually can't reproduce, the whole URL is shown on my MM for the given site (see screenshot). 

Have you got another dapp with a very long URL where I can test?
Screen Shot 2022-03-08 at 10 04 06 PM

FrederikBolding:
@Gtonizuka Is this one better? https://6224da936a44480007d48726--evm-chainlist.netlify.app/ but either way the "chip" is also overflowing on your screenshot.

You can also artificially change it in the inspector to see the bug.

Gtonizuka (approved):
This looks good to me  for now - it would be great if we have that URL input scrolling left/right somehow so people can see the entire URL (they might want to copy and paste it somewhere else to check they are connecting to the right place)

brad-decker (approved):

PR #13876, opened 2022-03-08
ToggleButton: Use label so users can click text

darkwing:
At present you need to click the literal ToggleButton to update the value, and clicking the text does nothing, which is bad.  With this PR, you can click the text next to the ToggleButton and the value changes

Explanation:  

Manual testing steps:  
  - Go to Settings
  - Click toggle button text at the bottom of the screen
  - Value should change!

### DIFF
--- a/ui/components/ui/toggle-button/toggle-button.component.js
+++ b/ui/components/ui/toggle-button/toggle-button.component.js
@@ -52,7 +52,7 @@ const ToggleButton = (props) => {
   const modifier = value ? 'on' : 'off';
 
   return (
-    <div
+    <label
       tabIndex="0"
       onKeyDown={(e) => {
         if (e.key === 'Enter') {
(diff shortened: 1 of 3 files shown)


Gtonizuka:
this looks good to me

Gtonizuka (dismissed):

adonesky1:
@darkwing did this UX change ask come down from design? If so is it documented somewhere?

darkwing:
No, but to me it comes across a bit like having a `<label >` with a checkbox or radio button, which is what they are under the hood, essentially.  Clicking text next to a form control to change it, or at the very least select it, is an age old web paradigm.

ryanml (dismissed):

ui/components/ui/toggle-button/toggle-button.component.js
```diff
@@ -52,7 +52,7 @@ const ToggleButton = (props) => {
   const modifier = value ? 'on' : 'off';
 
   return (
-    <div
+    <label
```
mcmire, line 55: It looks like the cursor changes to `pointer` when you hover over the toggle. What do you think about adding `cursor: pointer` to `toggle-button` so that the cursor changes when you hover over the text too?

mcmire (approved):

adonesky1 (approved):

PR #13888, opened 2022-03-09
Make Edit Gas Fee testcase independent of arbitrary Wait time for Gas Editing

seaona:
**What:** Make Edit Gas Fee testcase independent of arbitrary Wait time for Gas Editing

**Fixes:** Edit Gas Fee testcase was not robust, and sometimes failed due to wait times issues. See example of a run on Linux env:
image

**Explanation:**  arbitrary wait times were used in between changing gas preferences, which made that the test failed at some runs. Now this arbitrary wait times have been replaced by a waiting for a certain value ("sec"), which means that API response has been received and rendered to the UI

**Note:** this testcase will be further improved by adding a mock API response, for making sure we get a response. This is being developed in another branch

### DIFF
--- a/test/e2e/tests/edit-gas-fee.spec.js
+++ b/test/e2e/tests/edit-gas-fee.spec.js
@@ -40,9 +40,12 @@ describe('Editing Confirm Transaction', function () {
 
         // update estimates to high
         await driver.clickElement('[data-testid="edit-gas-fee-button"]');
-        await driver.delay(regularDelayMs);
+        const timeElement = await driver.findElement({
+          text: '--',
+          tag: 'span',
+        });
+        await driver.waitForElementContainingValue(timeElement, 'sec');
         await driver.clickElement('[data-testid="edit-gas-fee-item-high"]');
-        await driver.delay(regularDelayMs);
         await driver.waitForSelector({ text: '🦍' });
         await driver.waitForSelector({
           text: 'Aggressive',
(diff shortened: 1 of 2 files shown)


test/e2e/tests/edit-gas-fee.spec.js
```diff
@@ -40,29 +40,28 @@ describe('Editing Confirm Transaction', function () {
 
         // update estimates to high
         await driver.clickElement('[data-testid="edit-gas-fee-button"]');
-        await driver.delay(regularDelayMs);
+        const timeElement = await driver.findElement({
+          text: '--',
+          tag: 'span',
+        });
+        await driver.waitForElementContainingValue(timeElement, 'sec');
```
PeterYinusa, line 47: can we use `waitForSelector` here, instead of introducing a this new method `waitForElementContainingValue`?

test/e2e/webdriver/driver.js
```diff
@@ -257,6 +257,10 @@ class Driver {
     }
   }
 
+  async waitForElementContainingValue(element, value) {
```
PeterYinusa, line 260: same as above

test/e2e/tests/edit-gas-fee.spec.js
seaona: well seen. I've included some extra `waitForSelectors` in further steps, as it was intermittently failing otherwise

PeterYinusa (approved):

segun (approved):
LGTM

Issue #13427, opened 2022-01-27
Flask release automation


Gudahtt:
We should automatically create a Flask build in CI as part of the release automation process, and include it on the release page.

rekmarks:
@Gudahtt this is ready to go since #13462 was merged!

PR #13895, opened 2022-03-09
Derive version suffix from build type and version

Gudahtt:
The version of a build is now derived from both the `version` field in `package.json` and the requested build type and version. The build type and version are added onto the manifest version as a suffix, according to the SemVer prerelease format.

We already have support in the extension for versions of this format, but to apply a Flask or Beta version required manual updates to `package.json`. Now it can be done just with build arguments.

A `get-version` module was created to make it easier to generate the version in the various places we do that during the build. It was created in the `development/lib` directory because it will be used by other non-build development scripts in a future PR.

The `BuildType` constant was extracted to its own module as well, and moved to the `development/lib` directory. This was to make it clear that it's used by various different development scripts, not just the build.

Manual testing steps:  
  - Check the "About" page of the extension to see that the version is set correctly. It should match the `package.json` version for main builds (i.e. not Flask or beta). For Flask and beta, it should be `${version}-${build type}.${build version}`
  - Check that a build version can be set as a build parameter (e.g. `yarn build prod --build-type beta --build-version 1`)
  - Ensure that a build version is not accepted it build type is not set
  - Ensure that only valid integers are accepted for the build version

### DIFF
--- a/development/build/utils.js
+++ b/development/build/utils.js
@@ -1,16 +1,5 @@
 const semver = require('semver');
-const { version } = require('../../package.json');
-
-/**
- * The distribution this build is intended for.
- *
- * This should be kept in-sync with the `BuildType` map in `shared/constants/app.js`.
- */
-const BuildType = {
-  beta: 'beta',
-  flask: 'flask',
-  main: 'main',
-};
+const { BuildType } = require('../lib/build-type');
 
 /**
  * Map the current version to a format that is compatible with each browser.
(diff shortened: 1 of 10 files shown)


Gudahtt:
This is needed for #13427

darkwing (dismissed):

development/lib/build-type.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * The distribution this build is intended for.
+ *
+ * This should be kept in-sync with the `BuildType` map in `shared/constants/app.js`.
```
mcmire, line 4: Is it possible to reuse that file instead of having to keep it in sync? I realize that that file is written in ESM syntax, so Node would not recognize it, but could we move this file into `shared/constants`?

mcmire (dismissed):
One nit, but I went through the testing steps and everything looks good to me.

development/lib/build-type.js
Gudahtt: That would break our lint rules. We could add an exception for it though. Doesn't seem great to have mixed module formats scattered throughout though. At least now we have a simple division between the non-ESM scripts, and ESM everything else.

I was hoping we'd find a way to import ESM modules in Node.js. But I don't know what the status is of that these days.

mcmire: That's fair!

development/build/utils.js
```diff
@@ -1,16 +1,5 @@
 const semver = require('semver');
-const { version } = require('../../package.json');
-
-/**
- * The distribution this build is intended for.
- *
- * This should be kept in-sync with the `BuildType` map in `shared/constants/app.js`.
- */
-const BuildType = {
```
Gudahtt, line 9: Oops. This is still being exported. Missed that.

Gudahtt: Fixed.

mcmire (approved):

rekmarks (approved):
LGTM!

PR #13896, opened 2022-03-09
Refactor stylelint config; upgrade stylelint

mcmire:
I noticed when adding stylelint integration with my editor that our
stylelint configuration could use a little refactoring. It looked like
we were including a copy of `stylelint-config-standard` with certain
rules turned off. While I can understand why this might be desirable as
it ensures that our stylelint configuration is as protected from
stylelint upgrades as possible, it means that we have to maintain a
large list of rules. We don't follow this strategy for ESLint; rather,
we inherit our set of rules from a set that is maintained by the
community. So it seems better to follow the same philosophy here.

In addition, as I looked into it further, I also noticed that we aren't
using any SCSS-specific stylelint rules.

So, this commit:

* Adds `stylelint-config-standard-scss` (which extends
  `stylelint-config-standard`) and updates the stylelint config to
  extend `stylelint-config-standard-scss`
* Removes any rules which are already included in
  `stylelint-config-standard`
* Disables new rules in the current version of
  `stylelint-config-standard` and `stylelint-config-standard-scss` that
  weren't present in the original configuration and would likely cause
  lint violations if they were enabled
* Upgrades `stylelint`
  * Note that I had to switch to a fork of `gulp-stylelint` that is
  compatible with the new version of `stylelint`
* Fixes valid lint violations, which includes:
  * Shortening `padding` and `margin` values in cases where the top and
    bottom or the left and right were the same
  * Removing `.scss` from `@imports` (as this is the default behavior)
  * Removing duplicate properties or selector blocks (which for some
    reason wasn't caught before)
  * Using new Sass modules (`color.adjust` over `lighten` and `map.get`
    over `map.get`)
  * Adding spaces after `//` in comments
  * Updating variable names to use kebab-case instead of camelCase
  * Removing duplicate semicolons
  * Adding extra lines after `@include`s and `@extend`s
  * Removing empty stylesheets

---

**Manual testing steps:**

This commit covers a lot of changes across all files, so I imagine the best way to test this is to do a quick glance at a few different types of screens and make sure that nothing is wildly off.


stylelint.config.js
```diff
@@ -1,134 +1,44 @@
...
-    'selector-pseudo-element-case': 'lower',
-    'selector-pseudo-element-colon-notation': 'double',
-    'selector-type-case': 'lower',
-    'unit-case': 'lower',
-    // 'value-keyword-case': 'lower',
-    'value-list-comma-newline-after': 'always-multi-line',
-    'value-list-comma-space-after': 'always-single-line',
-    'value-list-comma-space-before': 'never',
-    'value-list-max-empty-lines': 0,
+    // style-config-standard overrides
```
adonesky1, line 8: Sorry I don't follow, why are we overriding these rules?

ui/components/app/app-header/index.scss
```diff
@@ -81,7 +81,6 @@
 
   &__account-menu-container {
     display: flex;
-    flex-flow: row nowrap;
```
adonesky1, line 84: is this because `row nowrap` is default?

ui/components/app/modals/cancel-transaction/cancel-transaction-gas-fee/index.scss
```diff
@@ -1,6 +1,5 @@
 .cancel-transaction-gas-fee {
   background: #f1f4f9;
-  padding: 16px;
```
adonesky1, line 3: 🤔

adonesky1: Oh I see there is another padding on line 6...

stylelint.config.js
adonesky1: > Disables new rules in the current version of
stylelint-config-standard and stylelint-config-standard-scss that
weren't present in the original configuration and would likely cause
lint violations if they were enabled

ah ok I see.

ui/components/app/app-header/index.scss
adonesky1: Screen Shot 2022-03-09 at 2 22 54 PM
yes that seems to be it

adonesky1:
All makes sense to me! So satisfying to see all that linting cleanup. Looks like you've got a couple of CI tests to fix but otherwise LGTM!

stylelint.config.js
mcmire: Yes, that's right!

ui/components/app/app-header/index.scss
mcmire: There were many cases I found in which selectors listed duplicate properties (not sure why this wasn't picked up by stylelint before). This is one of them (note how `flex-flow` is already listed on line 87).

ui/components/app/modals/cancel-transaction/cancel-transaction-gas-fee/index.scss
mcmire: Yup this is another one of these cases.

ui/components/app/signature-request-original/index.scss
```diff
@@ -145,7 +145,6 @@
 
   &__body {
     width: 100%;
-    height: 100%;
```
mcmire, line 148: This property is duplicated. `height: 0` overrides `height: 100%`.

ui/components/app/signature-request-original/index.scss
```diff
@@ -207,7 +206,6 @@
   &__row-title {
     @include H5;
 
-    width: 80px;
```
mcmire, line 210: This property is duplicated. `width: 100%` overrides `width: 80px`.

ui/components/app/signature-request/signature-request-header/index.scss
```diff
@@ -18,17 +18,14 @@
...
+        margin: 0;
       }
 
       &__account-name {
         @include H7;
 
         font-weight: 500;
       }
-
-      &__top-row {
```
mcmire, line 29: This selector is duplicated and its properties were moved above.

ui/components/app/transaction-list-item-details/index.scss
```diff
@@ -125,9 +125,8 @@
 
   &__transaction-breakdown {
     flex: 1;
-    margin-right: 8px;
     min-width: 0;
-    margin: 0 0 8px 0;
+    margin: 0 8px 8px;
```
mcmire, line 129: `margin-right` is redundant if we're using `margin`.

ui/pages/first-time-flow/metametrics-opt-in/index.scss
```diff
@@ -51,8 +51,8 @@
   }
 
   &__content {
-    overflow-y: scroll;
```
mcmire, line 54: This property is redundant and is being overridden by `overflow-y: auto` below (which is being pulled up from the duplicate `&__content`).

ui/pages/keychains/index.scss
```diff
@@ -56,23 +71,6 @@
   }
 }
 
-.import-account {
```
mcmire, line 59: `.import-account` is already specified above, so this code was merged with that above.

ui/pages/pages.scss
```diff
@@ -9,7 +9,6 @@
 @import 'confirm-encryption-public-key/confirm-encryption-public-key';
 @import 'confirm-transaction-base/transaction-alerts/transaction-alerts';
 @import 'confirmation/confirmation';
-@import 'connected-sites/index';
```
mcmire, line 12: This file was empty.

ui/pages/settings/alerts-tab/alerts-tab.scss
```diff
@@ -2,7 +2,6 @@
   &__body {
     @include H6;
 
-    display: grid;
```
mcmire, line 5: This property is redundant and is being overridden by `display: block`, below.

ui/pages/swaps/dropdown-search-list/index.scss
```diff
@@ -57,7 +57,6 @@
   &__selector-closed {
     display: flex;
     flex-flow: row nowrap;
-    align-items: center;
```
mcmire, line 60: This property is being duplicated.

ui/pages/swaps/dropdown-search-list/index.scss
```diff
@@ -126,7 +125,6 @@
 
     padding: 16px;
     color: var(--Grey-500);
-    min-height: 300px;
```
mcmire, line 129: This property is redundant and is being overridden by `min-height: 194px`, below.

ui/pages/swaps/searchable-item-list/index.scss
```diff
@@ -192,14 +196,6 @@
     white-space: nowrap;
   }
 
-  &__list-container {
```
mcmire, line 195: This selector is duplicated. All of its properties were moved above.

ui/components/app/app-header/index.scss
adonesky1: Oh ok yeah I over thought it in this case. But yeah I noticed a lot of the duplication getting cleaned-up otherwise.

adonesky1 (dismissed):
LGTM!

georgewrmarshall (commented):
This is great!! Looks like you need to rebase from latest `develop` would be happy to help get this across the line in anyway I can :D

ui/components/app/hold-to-reveal-button/index.scss
```diff
@@ -36,8 +37,7 @@ $circle-stroke-width: 2px;
   }
 }
 
-.hold-to-reveal-button {
-  // Shared styles
+%hold-to-reveal-button {
```
mcmire, line 40: stylelint now complains if `@extend` isn't used with a `%`-selector, so I've moved some things in this file.

mcmire:
@georgewrmarshall @adonesky1 I've rebased this branch. Want to re-review?

ui/components/app/network-display/index.scss
```diff
@@ -1,3 +1,5 @@
+@use "sass:color";
```
georgewrmarshall, line 1: Is this still needed? I don't see a reference to it? Although I guess if it wasn't it should have been picked up by the linter 🤔

georgewrmarshall (dismissed):
Really nice! Ran `yarn lint:styles` it works well. I can't comment on the lavamoat changes so may need @kumavis for that expertise and just left one question

ui/components/app/network-display/index.scss
mcmire: Removed in [15a862a](https://github.com/MetaMask/metamask-extension/pull/13896/commits/15a862ad7ec17f97abc4586d79b7eaa95bae317b).

mcmire:
Looks like there were new changes on develop which caused conflicts 😅 . I've merged those in accordingly.

mcmire:
Hmm. I am not sure why CI is failing. I will revisit this some time later today.

ui/components/app/network-display/index.scss
georgewrmarshall: Weird that the linter didn't pick that up haha

brad-decker:
Spoke with @mcmire and I will be taking this on to finish it up. Thanks @mcmire for your initiative in doing this!

mcmire:
I am going to close this PR as it is something that I will probably never finish, and it has become pretty stale now. It would probably be easier at this point for whoever picks this up to start over. Maybe this can be done in smaller chunks or something.

PR #13906, opened 2022-03-10
Permissions: convert hook to util

digiwand:
Fixes: n/a

clean-up ticket following https://github.com/MetaMask/metamask-extension/pull/13833#discussion_r819614766.

Explanation:  
In https://github.com/MetaMask/metamask-extension/pull/13833, we created a custom hook to get the permission details for the ConnectedAccountsPermissions and PermissionsConnectPermissionsList components. It's been proposed that we replace the hook with a regular util function in the discussion here https://github.com/MetaMask/metamask-extension/pull/13833#discussion_r819614766. 

Manual Testing Steps:

- Connect to the test dapp https://metamask.github.io/test-dapp/
- Open MetaMask Extension
- Click "Connect"
- Expand "Permissions"


rekmarks (changes requested):
Nice work on the performance metrics! I'll let @Gudahtt and @danjm contradict me, but I don't think the performance hit is worth more than well-organized / simple code. On that note, now that this is a plain utility function and not a memoized hook, the organization of this code makes less sense.

If we had written this function as an unmemoized utility from the start, we would never have initialized this big object inside of it every time the function is called. With that in mind, I propose that we:

1. Initialize the `permissionDescriptions` object once, at the top of the `utils/permissions.js` file. Since it's a permission-related object, let's maybe `Object.freeze` or recursively freeze it using `deep-freeze-strict` (or whatever deep freeze package we use in the extension).
2. For all of the properties in the object that are function, add `t` as a parameter to them, to ensure that they are always called with the correct `t`.
3. Optionally, if we want to get a bit fancier, it's really just the nested `label` property that needs to be a function, since it's the only thing that changes dynamically along with `t`. Therefore, we could just turn that into a function and change the end of this function to:

```js
  return typeof value.label === 'function' ? { ...value, label: value.label(t) } : value;
```

...or something like that.

digiwand:
@rekmarks 

Honestly, I didn't quite think to write objects in this way with functions returning computed results until now. This is great for instantiating the object once as it should be! 

Thank you very much for this approach. Is this good? I've turned all of the label props into functions since we need to at least pass `t`. I passed `permissionName` to all of the label functions even when they aren't being used -- passing the unused `permissionName` is a bit odd to me, but I'm not sure if there's a better way going about this part.

rekmarks (approved):
Awesome, LGTM!

rekmarks:
Btw, we should remove or optionally update any outdated performance benchmarks in the PR description cc: @digiwand

darkwing (approved):

digiwand:
@rekmarks Thanks! Good call removing performance benchmark information.

For documentation purposes I will move this info here: 

This PR replaces the permission hook with a util function. It appears to be slightly less performant, but the trade-off is only about a .003 - 0.01 ms cost on Chrome browser on my MacOS machine. Using lodash#memoize isn't too much of a performance boost here, so we didn't add it to this PR. See performance recordings from my Chrome browser on my MacOS machine below. 

**Recorded performance using hook (before this PR):**
Screenshot 2022-03-10 at 12 37 49 PM

**Recorded performance using util (NEW with this PR):** 
Screenshot 2022-03-10 at 1 40 58 PM

**Recorded performance using util and memoize lodash function (not used):** 
Screenshot 2022-03-10 at 1 39 01 PM

PR #13908, opened 2022-03-10
Update pull request template

mcmire:
In order to properly review a pull request, it is necessary to not only
understand the details about the changes presented, but also the
*context* behind those changes. Oftentimes, this context is difficult to
locate or even absent altogether. Providing a link to an issue or a user
story helps, but only if there are sufficient details listed in the
resource itself, which may not always be the case. Ultimately, I feel
that the best way to provide such information is to briefly explain it
in the PR description.

Additionally, for changes that involve UI/UX improvements, screenshots
or screencaps can really help the review process, as one can comprehend
the changes much faster than reading the relevant code.

With that in mind, this commit updates the pull request template to
include some more prompts and a better template for the description
area.

---

You can see what this looks like for a new pull request here: https://github.com/mcmire/metamask-extension/compare/develop...mcmire:new-branch?expand=1

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,6 +7,8 @@ This is a freeform area, but there are 3 questions you should try to answer:
 * What is the current state of things and why does it need to change?
 * What is the solution your changes offer?
 
+Below is a template to give you some ideas. Feel free to use your own words!
+
 (Hint: all of this should ideally go in your commit message so that people don't have to find this pull request in the future ;))
 -->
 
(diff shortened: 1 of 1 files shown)


darkwing:
I like the additions.  I'd really love if we could move the `fixes` higher though; since many PRs are specific to an issue, one of the first things I like doing is reading the issue to understand the problem before the fix is described to me.

darkwing (dismissed):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,8 +1,45 @@
...
+* What information do reviewers need to have in order to understand your changes?
+* What is the current state of things and why does it need to change?
+* What is the solution your changes offer?
+
+Below is a template to give you some ideas. Feel free to use your own words!
+
+(Hint: all of this should ideally go in your commit message so that people don't have to find this pull request in the future ;))
+-->
+
+Currently, ...
```
Gudahtt, line 15: Wary of including un-commented text like this in the templates because of how commonly users just leave it unaltered.

e.g. how many PRs have you seen with:

```
Manual testing steps:
- 
- 
- 
```

Bets on half the community PRs having the manual testing steps of "Go to this screen, do this, then do this"? :sweat_smile:

Gudahtt: Though we could instead agree as a team to start being strict about this. That's another option.

mcmire: Good call! As for whether we should be strict, I do want people to try and give more context around PRs as sometimes issues don't tell the full story but since everyone has their own style I'm trying not to completely dictate how people do that, just nudge them in a certain direction! But you're right that this should be designed for contributors as well. To that end I moved the prompts inside of the comments.

Gudahtt (approved):
LGTM!

brad-decker (approved):

PR #13671, opened 2022-02-17
Fixed firefox Snaps + add initial end-to-end tests

shanejonas:
This fixes snaps in firefox. Also adds e2e tests to test against https://github.com/MetaMask/test-snaps/

PR #13912, opened 2022-03-10
Fix snap e2e tests

Gudahtt:
The snap e2e tests introduced in #13671 were broken due to a conflict with the changes in #13895. The latter PR changed the version name for non-main builds so that it always includes the build type.

The Firefox webdriver has been updated to use the new `getVersion` utility, ensuring that it always looks for the correct build filename.


georgewrmarshall (approved):
lgtm!

rekmarks (approved):

PR #13954, opened 2022-03-15
add e2e test for custom token send transfer action

### LINKED ISSUE #13676: [e2e] Refactor custom token tests (send | metamask-ui.spec.js)
e2e test suite refactoring as part of the https://github.com/MetaMask/metamask-extension/issues/11136

This ticket should cover three batches of the tests identified by describe block in the current code within metamask-ui.spec.js file:
1. Send token from inside MetaMask
2. Send a custom token from dapp
3. Transfers a custom token from dapp when no gas value is specified

AC:
- Migrate e2e tests so that they use fixtures to set up test state, and test specific pieces of functionality in isolation.
- Separate these tests into a new file `custom-token-send.spec.js`

Notes:
* The custom token tests interact with the test dapp. The test dapp is controlled (started/stopped) in a different way than in metamask-ui.spec.js, you can see the new way in test/e2e/tests/personal-sign.spec.js for example.
* Currently each tests in metamask-ui.spec.js depend on the previous one. When migrating the tests we’re moving away from this pattern by setting up the state of the extension using fixtures which live here test/e2e/fixtures.

mirjanaKukic:
## Explanation
We need to refactor these tests since they already exist in _metamask-ui.spec.js_ file and we want them to use fixtures.
New tests will be moved to _custom-token-send-transfer.spec.js_ file. 
Once all refactoring is done these tests will be removed from metamask-ui.spec.js

Resolves [#13676](https://github.com/MetaMask/metamask-extension/issues/13676)

## More Information

## Manual Testing Steps

`yarn test:e2e:single test/e2e/tests/custom-token-send-transfer.spec.js`

## Pre-Merge Checklist

### DIFF
--- a/test/e2e/tests/custom-token-send-transfer.spec.js
+++ b/test/e2e/tests/custom-token-send-transfer.spec.js
@@ -1,0 +1,450 @@
+const { strict: assert } = require('assert');
+const {
+  convertToHexValue,
+  withFixtures,
+  getWindowHandles,
+} = require('../helpers');
+
+describe('Send token from inside MetaMask', function () {
+  const ganacheOptions = {
+    accounts: [
+      {
+        secretKey:
+          '0x7C9529A67102755B7E6102D6D950AC5D5863C98713805CEC576B945B15B71EAC',
+        balance: convertToHexValue(25000000000000000000),
+      },
+    ],
+  };
+  it('starts to send a transaction, transitions to the confirm screen, displays the token transfer data, customizes gas, submits the transaction, finds the transaction in the transactions list', async function () {
+    await withFixtures(
+      {
+        dapp: true,
+        fixtures: 'connected-state',
+        ganacheOptions,
+        title: this.test.title,
+        failOnConsoleError: false,
+      },
+      async ({ driver }) => {
+        await driver.navigate();
+        await driver.fill('#password', 'correct horse battery staple');
+        await driver.press('#password', driver.Key.ENTER);
+
+        // create token
+        await driver.openNewPage(`http://127.0.0.1:8080/`);
+        await driver.waitForSelector({ text: 'Create Token', tag: 'button' });
+        await driver.clickElement({ text: 'Create Token', tag: 'button' });
+
+        let windowHandles = await getWindowHandles(driver, 3);
+        await driver.switchToWindow(windowHandles.popup);
+
+        await driver.clickElement({ text: 'Confirm', tag: 'button' });
+
+        await driver.switchToWindow(windowHandles.dapp);
+
+        await driver.waitForSelector({
+          css: '#tokenAddress',
+          text: '0x',
+        });
+
+        await driver.clickElement({
+          text: 'Add Token to Wallet',
+          tag: 'button',
+        });
+
+        windowHandles = await getWindowHandles(driver, 3);
+        await driver.switchToWindow(windowHandles.popup);
+        await driver.clickElement({ text: 'Add Token', tag: 'button' });
+
+        await driver.switchToWindow(windowHandles.extension);
+
+        const asset = await driver.waitForSelector({
+          css: '.list-item',
+          text: '10 TST',
+        });
+        asset.click();
+
+        await driver.waitForSelector('[data-testid="eth-overview-send"]');
+        await driver.clickElement('[data-testid="eth-overview-send"]');
+
+        await driver.fill(
+          'input[placeholder="Search, public address (0x), or ENS"]',
+          '0x2f318C334780961FB129D2a6c30D0763d9a5C970',
+        );
+        await driver.fill('.unit-input__input', '1');
+        // Continue to next screen
+
+        await driver.waitForSelector(
+          '[data-testid="page-container-footer-next"]',
+        );
+        await driver.clickElement('[data-testid="page-container-footer-next"]');
+
+        const estimatedGasFee = await driver.waitForSelector({
+          css: '.currency-display-component__text',
+          text: '0.000154',
+        });
+        assert.notEqual(
+          await estimatedGasFee.getText(),
+          '0',
+          'Estimated gas fee should not be 0',
+        );
+
+        await driver.waitForSelector({
+          text: '1 TST',
+          tag: 'h1',
+        });
+        await driver.delay(1000);
+        await driver.clickElement({
+          text: 'Hex',
+          tag: 'button',
+        });
+
+        const functionType = await driver.findElement(
+          '.confirm-page-container-content__function-type',
+        );
+        const functionTypeText = await functionType.getText();
+        assert(functionTypeText.match('Transfer'));
+
+        const tokenAmount = await driver.findElement(
+          '.confirm-page-container-summary__title-text',
+        );
+
+        const tokenAmountText = await tokenAmount.getText();
+        assert.equal(tokenAmountText, '1 TST', 'Token amount is not correct');
+
+        const confirmDataDiv = await driver.findElement(
+          '.confirm-page-container-content__data-box',
+        );
+        const confirmDataText = await confirmDataDiv.getText();
+        assert(
+          confirmDataText.match(
+            /0xa9059cbb0000000000000000000000002f318c334780961fb129d2a6c30d0763d9a5c97/u,
+          ),
+        );
+        await driver.clickElement({ text: 'Details', tag: 'button' });
+        await driver.clickElement({ text: 'Edit', tag: 'button' });
+        const inputs = await driver.findElements('input[type="number"]');
+        const gasLimitInput = inputs[0];
+        const gasPriceInput = inputs[1];
+        await gasLimitInput.fill('100000');
+        await gasPriceInput.fill('100');
+        await driver.clickElement({ text: 'Save', tag: 'button' });
+
+        await driver.clickElement({ text: 'Confirm', tag: 'button' });
+        await driver.waitForSelector(
+          {
+            css:
+              '.transaction-list__completed-transactions .transaction-list-item__primary-currency',
+            text: '-1 TST',
+          },
+          { timeout: 10000 },
+        );
+        const transactionTxt = await driver.waitForSelector({
+          css: '.list-item__heading',
+          text: 'Send TST',
+        });
+        assert(
+          transactionTxt.getText(),
+          'Send TST',
+          'Transaction is not done correctly',
+        );
+      },
+    );
+  });
+});
+
+describe('Send a custom token from dapp', function () {
+  const ganacheOptions = {
+    accounts: [
+      {
+        secretKey:
+          '0x7C9529A67102755B7E6102D6D950AC5D5863C98713805CEC576B945B15B71EAC',
+        balance: convertToHexValue(25000000000000000000),
+      },
+    ],
+  };
+  it('sends an already created token', async function () {
+    await withFixtures(
+      {
+        dapp: true,
+        fixtures: 'connected-state',
+        ganacheOptions,
+        title: this.test.title,
+      },
+      async ({ driver }) => {
+        await driver.navigate();
+        await driver.fill('#password', 'correct horse battery staple');
+        await driver.press('#password', driver.Key.ENTER);
+
+        // create token
+        await driver.openNewPage(`http://127.0.0.1:8080/`);
+        await driver.waitForSelector({ text: 'Create Token', tag: 'button' });
+        await driver.clickElement({ text: 'Create Token', tag: 'button' });
+
+        let windowHandles = await getWindowHandles(driver, 3);
+        await driver.switchToWindow(windowHandles.popup);
+
+        await driver.clickElement({ text: 'Confirm', tag: 'button' });
+
+        await driver.switchToWindow(windowHandles.dapp);
+
+        await driver.waitForSelector({
+          css: '#tokenAddress',
+          text: '0x',
+        });
+
+        await driver.clickElement({
+          text: 'Add Token to Wallet',
+          tag: 'button',
+        });
+
+        windowHandles = await getWindowHandles(driver, 3);
+        await driver.switchToWindow(windowHandles.popup);
+        await driver.clickElement({ text: 'Add Token', tag: 'button' });
+
+        windowHandles = await getWindowHandles(driver, 2);
+        await driver.switchToWindow(windowHandles.dapp);
+
+        await driver.clickElement({ text: 'Transfer Tokens', tag: 'button' });
+        windowHandles = await getWindowHandles(driver, 3);
+        await driver.switchToWindow(windowHandles.popup);
+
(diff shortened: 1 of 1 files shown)


test/e2e/tests/custom-token-send-transfer.spec.js
```diff
@@ -0,0 +1,414 @@
...
+          tag: 'button',
+        });
+
+        windowHandles = await getWindowHandles(driver, 3);
+        await driver.switchToWindow(windowHandles.popup);
+        await driver.clickElement({ text: 'Add Token', tag: 'button' });
+
+        await driver.switchToWindow(windowHandles.extension);
+
+        const assets = await driver.waitForSelector({
```
PeterYinusa, line 59: Should this be `asset`? since its only one

test/e2e/tests/custom-token-send-transfer.spec.js
```diff
@@ -0,0 +1,414 @@
...
+        );
+        await driver.fill('.unit-input__input', '1');
+        // Continue to next screen
+        await driver.delay(1000);
+        await driver.waitForSelector(
+          '[data-testid="page-container-footer-next"]',
+        );
+        await driver.clickElement('[data-testid="page-container-footer-next"]');
+
+        await driver.delay(2000);
```
PeterYinusa, line 80: Why do we need this delay? Can't we use waitForSelector?

test/e2e/tests/custom-token-send-transfer.spec.js
```diff
@@ -0,0 +1,414 @@
...
+        await driver.clickElement({ text: 'Add Token', tag: 'button' });
+
+        windowHandles = await getWindowHandles(driver, 2);
+        await driver.switchToWindow(windowHandles.dapp);
+
+        await driver.clickElement({ text: 'Transfer Tokens', tag: 'button' });
+        windowHandles = await getWindowHandles(driver, 3);
+        await driver.switchToWindow(windowHandles.popup);
+
+        await driver.delay(4000);
```
PeterYinusa, line 191: Why do we need this delay?

test/e2e/tests/custom-token-send-transfer.spec.js
```diff
@@ -0,0 +1,414 @@
...
+        await driver.switchToWindow(windowHandles.popup);
+        await driver.clickElement({ text: 'Add Token', tag: 'button' });
+
+        await driver.switchToWindow(windowHandles.dapp);
+
+        await driver.clickElement({ text: 'Transfer Tokens', tag: 'button' });
+        windowHandles = await getWindowHandles(driver, 3);
+        await driver.switchToWindow(windowHandles.popup);
+
+        await driver.delay(4000);
```
PeterYinusa, line 264: Why do we need this delay?

test/e2e/tests/custom-token-send-transfer.spec.js
```diff
@@ -0,0 +1,414 @@
...
+
+        await driver.waitForSelector({
+          css: '.asset-list-item__token-button',
+          text: '8.5 TST',
+        });
+
+        await driver.clickElement({
+          text: 'Activity',
+          tag: 'button',
+        });
```
PeterYinusa, line 318: Any reason why we are ending the test with an action instead of an assertion? Why do we need to click on this element?

PeterYinusa (changes requested):

darkwing (dismissed):

PeterYinusa (dismissed):
Can you use the [ganache seeder](https://github.com/MetaMask/metamask-extension/blob/develop/test/e2e/seeder/ganache-seeder.js) to deploy the hst contract, so you do not have to deploy the contract using the test-dapp in each test. This will also help remove a lot of code from each test

seaona:
Same comment as [here](https://github.com/MetaMask/metamask-extension/pull/15728#issuecomment-1238141179). What do you think @PeterYinusa @mirjanaKukic

test/e2e/tests/custom-token-send-transfer.spec.js
```diff
@@ -0,0 +1,445 @@
...
+        // transfer token from dapp
+        await driver.clickElement({ text: 'Transfer Tokens', tag: 'button' });
+        await driver.waitUntilXWindowHandles(3);
+        windowHandles = await driver.getAllWindowHandles();
+
+        await driver.switchToWindowWithTitle(
+          'MetaMask Notification',
+          windowHandles,
+        );
+        await driver.waitForSelector({ text: '1.5 TST', tag: 'h1' });
```
seaona, line 290: I think we can remove this line

test/e2e/tests/custom-token-send-transfer.spec.js
```diff
@@ -0,0 +1,445 @@
...
+          '0x7C9529A67102755B7E6102D6D950AC5D5863C98713805CEC576B945B15B71EAC',
+        balance: convertToHexValue(25000000000000000000),
+      },
+    ],
+  };
+  it('starts to send a transaction, transitions to the confirm screen, displays the token transfer data, customizes gas, submits the transaction, finds the transaction in the transactions list', async function () {
+    await withFixtures(
+      {
+        dapp: true,
+        fixtures: 'connected-state',
```
seaona, line 24: there is a fixture for a `custom-token`, could we use it instead of this one? This might help reduce the test steps, as we won't need to add the token to the wallet then

seaona:
Maybe we could summarize `it` blocks and `describes`? At the moment this is the current text and hierarchy:
```
describe('Send token from inside MetaMask')
    it('starts to send a transaction, transitions to the confirm screen, displays the token transfer data, customizes gas, submits the transaction, finds the transaction in the transactions list'

describe('Send a custom token from dapp')
    it('sends an already created token')
    it('customizes gas, submits the transaction and finds the transaction in the transactions list')
    
describe('Transfers a custom token from dapp when no gas value is specified')
    it('transfers an already created token without specifying gas, submits the transaction, finds the transaction in the transactions list')
```
A proposal for simplifying tests (we could reduce code by removing common values on the different `describes` by having just one) and improve test-redability a little bit. 100% open for discussion @mirjanaKukic @PeterYinusa  

```
describe('Custom Token Transactions')
    it('can be sent from inside MetaMask')
    it('can be transferred from a dapp')
    it('can be transferred from a dapp customizing gas values')
    it('can be transferred without specifying gas from a dapp'')
```

test/e2e/tests/custom-token-send-transfer.spec.js
```diff
@@ -0,0 +1,445 @@
...
+        windowHandles = await driver.getAllWindowHandles();
+        await driver.switchToWindowWithTitle(
+          'MetaMask Notification',
+          windowHandles,
+        );
+        await driver.waitForSelector({
+          tag: 'h1',
+          text: '1.5 TST',
+        });
+        await driver.waitForSelector({ text: 'Confirm', tag: 'button' });
```
seaona, line 404: I think we can remove this line

test/e2e/tests/custom-token-send-transfer.spec.js
mirjanaKukic: If I remove it I'm getting error on the line after it : 
 _ElementClickInterceptedError: Element <button class="button btn--rounded btn-primary page-container__footer-button"> is not clickable at point (266,567) because another element <div class="loading-overlay"> obscures it_

So I put this waitForSelector here since it seems like it needs a little bit more time to load popup properly. I could use delay but we all agreed to avoid it. Hope this makes sense.

test/e2e/tests/custom-token-send-transfer.spec.js
mirjanaKukic: Good idea. I tried it but when I run test it shows me 0 TST so I'm not able to send tokens since I don't have enough tokens. Maybe there is way to change custom-token fixture in order to have more TST tokens?

test/e2e/tests/custom-token-send-transfer.spec.js
mirjanaKukic: Same as above. If I remove it test fails since it needs more time to load popup.

test/e2e/tests/custom-token-send-transfer.spec.js
seaona: Thank you, that make sense! Let's leave it as it is then! @mirjanaKukic

test/e2e/tests/custom-token-send-transfer.spec.js
seaona: Thank you, that make sense! Let's leave it as it is then! @mirjanaKukic

test/e2e/tests/custom-token-send-transfer.spec.js
```diff
@@ -0,0 +1,420 @@
...
+        await driver.waitUntilXWindowHandles(2);
+        await driver.switchToWindow(extension);
+
+        const asset = await driver.waitForSelector({
+          css: '.list-item',
+          text: '10 TST',
+        });
+        asset.click();
+        // send token from extension
+        await driver.waitForSelector('[data-testid="eth-overview-send"]');
```
PeterYinusa, line 66: You shouldn't need to use waitForSelector followed by clickElement, as clickElement waits until the element is located

test/e2e/tests/custom-token-send-transfer.spec.js
```diff
@@ -0,0 +1,420 @@
...
+
+        await driver.fill(
+          'input[placeholder="Search, public address (0x), or ENS"]',
+          '0x2f318C334780961FB129D2a6c30D0763d9a5C970',
+        );
+        await driver.fill('.unit-input__input', '1');
+        // continue to next screen
+        await driver.waitForSelector(
+          '[data-testid="page-container-footer-next"]',
+        );
```
PeterYinusa, line 77: same as above

mirjanaKukic: Done

brad-decker:
@PeterYinusa @seaona this is ready for another look.

test/e2e/tests/custom-token-send-transfer.spec.js
```diff
@@ -0,0 +1,426 @@
...
+        await driver.clickElement({ text: 'Add token', tag: 'button' });
+        await driver.waitUntilXWindowHandles(2);
+        await driver.switchToWindow(extension);
+
+        const asset = await driver.waitForSelector({
+          css: '.list-item',
+          text: '10 TST',
+        });
+        asset.click();
```
PeterYinusa, line 67: Instead of using `waitForSelector` followed by `click`, can we just use `clickElement` as the asset variable is only used once

test/e2e/tests/custom-token-send-transfer.spec.js
```diff
@@ -0,0 +1,426 @@
...
+        await driver.fill(
+          'input[placeholder="Search, public address (0x), or ENS"]',
+          '0x2f318C334780961FB129D2a6c30D0763d9a5C970',
+        );
+        await driver.fill('.unit-input__input', '1');
+        // continue to next screen
+        await driver.clickElement('[data-testid="page-container-footer-next"]');
+        // added wait for selector since it needs some time to load this page
+        // otherwise would have to use delays
+        await driver.waitForSelector({ text: 'Edit', tag: 'button' });
```
PeterYinusa, line 81: we should be able to remove `await driver.waitForSelector({ text: 'Edit', tag: 'button' });`

test/e2e/tests/custom-token-send-transfer.spec.js
```diff
@@ -0,0 +1,426 @@
...
+        // checks if transaction finish correctly from extension
+        await driver.clickElement({ tag: 'button', text: 'Activity' });
+        await driver.waitForSelector({
+          css: '.transaction-list__completed-transactions .transaction-list-item__primary-currency',
+          text: '-1.5 TST',
+        });
+        await driver.waitForSelector({
+          css: '.list-item__heading',
+          text: 'Send TST',
+        });
```
PeterYinusa, line 318: We should be able to remove the code block below, as theres already a wait for the completed transaction
```
          css: '.list-item__heading',
          text: 'Send TST',
        });
```

test/e2e/tests/custom-token-send-transfer.spec.js
```diff
@@ -0,0 +1,426 @@
...
+          const pendingTxes = await driver.findElements(
+            '.transaction-list__pending-transactions .transaction-list-item',
+          );
+          return pendingTxes.length === 1;
+        }, 10000);
+
+        await driver.waitForSelector({
+          css: '.transaction-list-item__primary-currency',
+          text: '-1.5 TST',
+        });
```
PeterYinusa, line 400: We should be able to remove the following code block, as the line that follows waits for the completed transaction list item
```
          const pendingTxes = await driver.findElements(
            '.transaction-list__pending-transactions .transaction-list-item',
          );
          return pendingTxes.length === 1;
        }, 10000);

          css: '.transaction-list-item__primary-currency',
          text: '-1.5 TST',
        });
```

test/e2e/tests/custom-token-send-transfer.spec.js
PeterYinusa: Thanks for the explanation, this issue occurs because we use the same selector on both the overview page and the asset overview page. Instead of `await driver.waitForSelector('[data-testid="eth-overview-send"]');` can we wait for the overview page to transition to the asset overview page e.g. wait for the balance to show 10 TST?

test/e2e/tests/custom-token-send-transfer.spec.js
mirjanaKukic: Done

test/e2e/tests/custom-token-send-transfer.spec.js
mirjanaKukic: Done

test/e2e/tests/custom-token-send-transfer.spec.js
mirjanaKukic: Done

test/e2e/tests/custom-token-send-transfer.spec.js
mirjanaKukic: Done

test/e2e/tests/custom-token-send-transfer.spec.js
mirjanaKukic: Hmm, I removed it and it works now. Tested it locally and also works after removing waitForSelector.

chloeYue (approved):
LGTM !

brad-decker:
@PeterYinusa do we want this PR still and if so who should own it?

PeterYinusa:
> @PeterYinusa do we want this PR still and if so who should own it?

Yes, we should get this in. I'll review it shortly and get it merged. We can do a follow-up PR to clean up anything outstanding if needed.

PR #13957, opened 2022-03-15
Adding ShowHideToggle component

georgewrmarshall:
## Explanation

Adding ShowHideToggle component and eye icon

Screen Shot 2022-03-15 at 8 17 14 AM
Screen Shot 2022-03-15 at 8 21 20 AM
Screen Shot 2022-03-15 at 8 21 36 AM


ui/components/ui/icon/icon-eye.js
```diff
@@ -0,0 +1,42 @@
...
+  <svg
+    width={size}
+    height={size}
+    fill={color}
+    className={className}
+    aria-label={ariaLabel}
+    xmlns="http://www.w3.org/2000/svg"
+    viewBox="0 0 512 512"
+  >
+    <path d="m444 247c-36-71-108-119-190-119-82 0-154 48-190 119-1 2-2 6-2 9 0 4 1 7 2 10 36 71 108 118 190 118 82 0 154-47 190-118 1-3 2-6 2-9 0-4-1-8-2-10z m-190 105c-53 0-96-43-96-96 0-53 43-96 96-96 53 0 96 43 96 96l0 1c0 52-43 96-96 96z m0-160c-6 1-12 1-17 3 4 5 6 12 6 19 0 17-14 31-31 31-8 0-14-2-20-6-1 6-2 12-2 17 0 35 29 64 64 64 36 0 64-29 64-64 0-35-28-63-64-63z" />
```
Gudahtt, line 19: Where did this SVG come from? Is this fontawesome?

ui/components/ui/show-hide-toggle/show-hide-toggle.js
```diff
@@ -0,0 +1,65 @@
...
+        className="show-hide-toggle__input"
+        id={id}
+        type="checkbox"
+        checked={checked}
+        onChange={onChange}
+        data-testid={dataTestId}
+        disabled={disabled}
+      />
+      <label htmlFor={id} className="show-hide-toggle__label">
+        <IconEye ariaLabel={ariaLabel} className="show-hide-toggle__icon" />
```
Gudahtt, line 28: Looks like this doesn't display any visual indication when it is focused. Or maybe it's the checkbox that is getting focused, and it's not shown because it's off screen. Either way, we have to ensure there is a visual representation of focus for the sake of helping keyboard users.

Curious why you chose to go this route of separating the checkbox component from the label, hiding the checkbox, and using the label instead. This seems like a lot more work than using CSS to display the checkbox as an eye. It's not clear this is any better for accessibility, readability, or even compatibility.

ui/components/ui/icon/icon-eye.js
georgewrmarshall: Yes, I just grabbed it from the Figma file but I wasn't sure where it was from. I think ideally we are using svgs over font icons. But do you think this could be breaking a usage agreement?

Gudahtt: If it's from the "FontAwesome Free" pack, this is allowed under their creative commons license. We're allowed to embed it and change it however we like. But it's dangerous to embed IP we don't own directly in our product without documenting what it is or where it came from. _Minimally_ we should document what it is and where it came from. That way we can be assured that it's allowed.

ui/components/ui/show-hide-toggle/show-hide-toggle.js
georgewrmarshall: I should double check this is an acceptable standard for a custom checkbox but I believed it was. I do see a visual representation of focus in my storybook do you see something different. I did change the component api in another update


https://user-images.githubusercontent.com/8112138/158444898-ba3d5072-4b14-4f85-b713-bf11341ac4b8.mov

ui/components/ui/icon/icon-eye.js
Gudahtt: I'd strongly recommend against _ever_ copying an SVG directly from Figma. Our Figma designs have included "FontAwesome Pro" icons in the past (which we cannot use, because our usage is incompatible with their license, they won't even let us pay for it), and they have contained other proprietary icons from other sources.  Big no-no. Please don't ever do this ,and speak up if you ever see somebody else doing it.

georgewrmarshall: OK thanks. That's good to know. This should be flagged at the design level or at least before design handover to dev. Is this common knowledge in design? Might be worth reminding them. As for using font icons they can render unexpectedly in different browsers so we should be pushing for svg icons. I've seen a few features using svgs go into the codebase with svgs being pulled from Figma so I think we should definitely flag this in the right channels.

ui/components/ui/show-hide-toggle/show-hide-toggle.js
georgewrmarshall: No haven't browser tested and haven't written unit tests yet. I'll do that now

ui/components/ui/icon/icon-eye.js
Gudahtt: We do have all of the SVG icons as a dependency (`@fortawesome/fontawesome-free`, see the `svgs` directory). These are the official SVGs. We could copy them direct from there and leave a comment saying where they came from (including the version). Or we could use the FontAwesome React library, where they've already packaged each SVG as a component. See here for details: https://fontawesome.com/v6/docs/web/use-with/react/

ui/components/ui/icon/icon-eye.js
```diff
@@ -6,17 +6,24 @@ const IconEye = ({
...
+  // It is a combination of the SVGs `regular/eye.svg` and `regular/eye-slash.svg`.
   <svg
     width={size}
     height={size}
     fill={color}
     className={className}
     aria-label={ariaLabel}
     xmlns="http://www.w3.org/2000/svg"
-    viewBox="0 0 512 512"
+    viewBox="0 0 640 512"
```
georgewrmarshall, line 20: What is the purpose of changing the viewBox here?

ui/components/ui/icon/icon-eye.js
```diff
@@ -6,17 +6,24 @@ const IconEye = ({
...
+    ) : (
+      <path d="M288 144a110.94 110.94 0 0 0-31.24 5 55.4 55.4 0 0 1 7.24 27 56 56 0 0 1-56 56 55.4 55.4 0 0 1-27-7.24A111.71 111.71 0 1 0 288 144zm284.52 97.4C518.29 135.59 410.93 64 288 64S57.68 135.64 3.48 241.41a32.35 32.35 0 0 0 0 29.19C57.71 376.41 165.07 448 288 448s230.32-71.64 284.52-177.41a32.35 32.35 0 0 0 0-29.19zM288 400c-98.65 0-189.09-55-237.93-144C98.91 167 189.34 112 288 112s189.09 55 237.93 144C477.1 345 386.66 400 288 400z" />
+    )}
   </svg>
```
georgewrmarshall, line 27: The intention was that would simply be the eye icon as it could be used in different context.

georgewrmarshall (commented):
Left some comments.

ui/components/ui/icon/icon-eye.js
Gudahtt: The fontawesome SVGs each used a different viewbox. Both were an increase from the Figma version seemingly, but they had different widths (the eye-slash one was slightly wider), so I went with the wider one to ensure it didn't get cut off.

As we just discussed I'm going to move this to a separate component, so I'll restore the exact viewbox that the SVGs use.

ui/components/ui/icon/icon-eye.js
Gudahtt: Resolved by splitting into two components.

ui/components/ui/show-hide-toggle/show-hide-toggle.js
Gudahtt: This was fixed by replacing the webkit-specific highlight color referenced in the styles with `Highlight`, which should work across browsers.

Also George explained to me the evils of icon fonts, so I'm now convinced that this is a much better approach than what I had put together before this :sweat_smile:

ui/components/ui/icon/icon-eye.js
Gudahtt: I copied over the official SVGs and added a comment explaining where we got them

georgewrmarshall:
Closing as this is addressed in https://github.com/MetaMask/metamask-extension/pull/13979

PR #13865, opened 2022-03-08
Degrade gracefully when gas API is down

mcmire:
When the gas API is down, the logic we use will no longer compute all of
the data that the gas API returns in order to reduce the burden on
Infura. Specifically, only estimated fees for different priority levels,
as well as the latest base fee, will be available; all other data
points, such as the latest and historical priority fee range and network
stability, will be missing. This commit updates the frontend logic to
account for this lack of data by merely hiding the relevant pieces of
the UI that would otherwise be shown.

---

## Related issues

* https://github.com/MetaMask/controllers/pull/712

## Manual testing steps

* Open the MetaMask popup.
* Go to send a transaction.
* Proceed to the confirm transaction screen.
* Click on "Market" (or whatever priority level is currently selected).
* Note that all of the data under Network Status should be present.
* Click on "Advanced".
* Note that all of the information under each field should be present ("current" and "12hr").
* Close the Advanced modal.
* Find the extension in your browser and open up the inspector for `background.html`.
* Go to the Network tab.
* Find a request that looks like `suggestedGasFees`. Right-click on it and choose "Block request URL".
* Open the MetaMask popup again.
* Note that 2 out of the 3 sections in Network Status should be gone (only Base Fee should be there).
* Open Advanced.
* Note that most of the information under each field should be gone (only "current" should be there).

## Screencaps

**When gas API is up:**

This should be the same behavior as today.

https://user-images.githubusercontent.com/7371/157169150-d55fd1df-8c01-45e9-8e24-68dd1b905730.mov

**When gas API is down:**

Note that the number of requests stays the same, i.e. there shouldn't be excessive `591f0dce...` lines in the request log, and specifically there shouldn't be excessive requests that call `eth_feeHistory`. Also note that some data disappears from the gas and advanced gas fee modals.

https://user-images.githubusercontent.com/7371/157169187-8c33ab9d-72ca-488d-849d-137b56c432fa.mov

PR #13962, opened 2022-03-15
Bump controllers to ^27.0.0

mcmire:
## Explanation

The new version of `controllers` includes an important change which
greatly relieves pressure on Infura when the Gas API goes down.

## More information

* https://github.com/MetaMask/controllers/pull/712
* https://github.com/MetaMask/metamask-extension/pull/13865
* https://consensys.slack.com/archives/GV90KPED9/p1646339312789189 (and other related issues in that channel)

## Manual testing steps

- Open MetaMask (popup or full-screen view).
- Go to Settings and turn on the Enhanced Gas Fee UI setting.
- Go to send a transaction.
- Proceed to the confirm transaction screen.
- Click on "Market" (or whatever priority level is currently selected).
- Note that all of the data under Network Status should be present.
- Click on "Advanced".
- Note that all of the information under each field should be present ("current" and "12hr").
- Close the Advanced modal.
- Find the extension in your browser and open up the inspector for background.html.
- Go to the Network tab.
- Find a request that looks like `suggestedGasFees`. Right-click on it and choose "Block request URL".
- Open the MetaMask popup again.
- Note that 2 out of the 3 sections in Network Status should be gone (only Base Fee should be there).
- Open Advanced.
- Note that most of the information under each field should be gone (only "current" should be there).


danjm (dismissed):

jpuri (dismissed):

adonesky1 (dismissed):

ui/components/ui/box/box.js
```diff
@@ -264,6 +264,7 @@ Box.propTypes = {
   display: PropTypes.oneOf(Object.values(DISPLAY)),
   width: PropTypes.oneOf(Object.values(BLOCK_SIZES)),
   height: PropTypes.oneOf(Object.values(BLOCK_SIZES)),
-  backgroundColor: PropTypes.oneOf(Object.values(ValidBackgroundColors)),
+  // backgroundColor: PropTypes.oneOf(Object.values(ValidBackgroundColors)),
```
adonesky1, line 267: 🤔

mcmire: Yeah I have no clue why this didn't cause e2e to fail before, but upon running the e2e test that was failing on this branch, which centers around editing gas fees, I am now getting:

```
Warning: Failed prop type: Invalid prop `backgroundColor` of value `white` supplied to `Box`, expected one of ["background-default","background-alternative","overlay-default","primary-default","primary-alternative","primary-muted","primary-disabled","secondary-default","secondary-alternative","secondary-muted","secondary-disabled","error-default","error-alternative","error-muted","error-disabled","warning-default","warning-alternative","warning-muted","warning-disabled","success-default","success-alternative","success-muted","success-disabled","info-default","info-alternative","info-muted","info-disabled","mainnet","ropsten","kovan","rinkeby","goerli","transparent","localhost"]
```

This appears to be happening because the default `backgroundColor` is `"white"` when it should be one of the values specified above. I think we should probably change the default to `"background-default"` but I'm afraid that'll break things.

Commenting this line out is definitely not the best solution but I can look at this in a future PR.

ui/components/app/confirm-page-container/confirm-page-container-content/confirm-page-container-content.component.js
```diff
@@ -51,7 +51,10 @@ export default class ConfirmPageContainerContent extends Component {
     hideTitle: PropTypes.bool,
     supportsEIP1559V2: PropTypes.bool,
     hasTopBorder: PropTypes.bool,
-    currentTransaction: PropTypes.string,
+    currentTransaction: PropTypes.shape({
```
mcmire, line 54: This was causing a warning to get generated during e2e tests, I thought I would just fix it so that they're easier to read in the future.

ui/components/app/gas-timing/gas-timing.component.js
```diff
@@ -205,7 +205,10 @@ export default function GasTiming({
 }
 
 GasTiming.propTypes = {
-  maxPriorityFeePerGas: PropTypes.string,
-  maxFeePerGas: PropTypes.string,
+  maxPriorityFeePerGas: PropTypes.oneOfType([
```
mcmire, line 208: This was causing a warning to get generated during e2e tests, I thought I would just fix it so that they're easier to read in the future.

mcmire:
Of course the error I get on CI is not the same error I get locally 🤦🏻

ui/components/ui/box/box.js
adonesky1: Yeah I think we swallow the warning for now and figure out whether/when we change the default to `background-default`. It shouldn't be causing a failure? I don't think we should do this...

test/e2e/tests/edit-gas-fee.spec.js
```diff
@@ -8,7 +8,8 @@ const {
   regularDelayMs,
 } = require('../helpers');
 
-describe('Editing Confirm Transaction', function () {
+/* eslint-disable-next-line mocha/no-skipped-tests */
+describe.skip('Editing Confirm Transaction', function () {
```
adonesky1, line 12: the changes you made to this test suite were not enough to get any of these tests to pass?

ui/components/app/confirm-page-container/confirm-page-container-content/confirm-page-container-content.component.js
adonesky1: I wonder if some might prefer these changes to be in separate PR/commit, I'm not too opinionated on this matter but just calling it out.

test/e2e/tests/edit-gas-fee.spec.js
mcmire: No, unfortunately. When I ran these tests locally, I was getting a certain set of errors — something about not being able to find certain elements. I chalked this up to the test not waiting for the right elements to show up before continuing. So those are fixed now, which is good, but unfortunately on CI I didn't get these errors at all. It seemed that an element was covering another element. You see that error here: https://app.circleci.com/pipelines/github/MetaMask/metamask-extension/20144/workflows/64fd2c0c-23d5-4247-bbb7-04a021728baf/jobs/517843. I am very perplexed as to what could be going on here. 

I am short on time in getting this Infura fix out the door so that's why I opted to skip this test for now, but there is a [similar conversation](https://consensys.slack.com/archives/GTQAGKY5V/p1647369888707269) happening in Slack and it seems Peter might be working on the same thing. I can hit him up tomorrow to see if he might know what's up with this.

danjm:
I don't think the e2e test changes are needed. Here is the same PR, but with all changes to the `test` and `ui` directories removed, and the build is passing: https://github.com/MetaMask/metamask-extension/pull/13985

Gudahtt:
Closing this as obsolete. @mcmire feel free to open a separate PR for any test changes that might still be useful.

mcmire:
@Gudahtt Thanks!

PR #13973, opened 2022-03-15
Add `data-testid` prop to `TextField` component

Gudahtt:
## Explanation

The `data-testid` prop has been added to `TextField`. This prop allows selecting this component more easily from e2e and unit tests. This is required for a subsequent PR that includes a text field that cannot be selected for by placeholder or contents or label (any of which would have been preferable).

### DIFF
--- a/ui/pages/swaps/searchable-item-list/__snapshots__/searchable-item-list.test.js.snap
+++ b/ui/pages/swaps/searchable-item-list/__snapshots__/searchable-item-list.test.js.snap
@@ -3,7 +3,6 @@
 exports[`SearchableItemList renders the component with initial props 1`] = `
 <div
   class="MuiFormControl-root MuiTextField-root searchable-item-list__search MuiFormControl-fullWidth"
-  data-testid="search-list-items"
 >
   <div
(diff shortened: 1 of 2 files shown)


brad-decker (approved):

georgewrmarshall (approved):
lgtm!

ui/pages/swaps/searchable-item-list/__snapshots__/searchable-item-list.test.js.snap
```diff
@@ -3,7 +3,6 @@
 exports[`SearchableItemList renders the component with initial props 1`] = `
 <div
   class="MuiFormControl-root MuiTextField-root searchable-item-list__search MuiFormControl-fullWidth"
-  data-testid="search-list-items"
```
Gudahtt, line 6: Previously the `data-testid` prop was being passed through by our component to the wrapper div of the Material UI Text Field component. Moving it to the input component results in this fixture changing to reflect that.

This has no impact on any tests, because this `search-list-items` ID was never referenced anywhere. I do not know why it was added.

PR #13974, opened 2022-03-15
Fix for connected sites throwing error if name is null.

segun:
## Explanation

There's an issue with connected sites where if the name attr of site metadata is null, extension throws an exception,
below is an example of a connected site with name=null from `state.metamask.subjectMetadata`

```
chrome-extension://ahbgkigagdiopfeihifjapallnjepehi]()": {
        "origin": "[chrome-extension://ahbgkigagdiopfeihifjapallnjepehi]()",
        "extensionId": "ahbgkigagdiopfeihifjapallnjepehi",
        "subjectType": "extension",
        "iconUrl": null,
        "name": null
      }
```

The fix is to check name before calling name.length. 

### Before

Screenshot 2022-03-15 at 22 21 18

### After

Screenshot 2022-03-15 at 22 21 58

## Manual testing steps

I can't reproduce this because I am not sure how to add extensions with null names, but I can simulate it.

Goto [here](https://github.com/MetaMask/metamask-extension/blob/99da6a9283d257c10151c7de267c9c219ef48d0d/ui/selectors/permissions.js#L92)

change `name` to `name: null`

Run Metamask to see the error.

### DIFF
--- a/ui/components/ui/icon-with-fallback/icon-with-fallback.component.js
+++ b/ui/components/ui/icon-with-fallback/icon-with-fallback.component.js
@@ -23,14 +23,14 @@ const IconWithFallback = ({
       src={icon}
       style={style}
       className={className}
-      alt={name.length ? name : 'icon'}
+      alt={name || 'icon'}
       {...props}
     />
   ) : (
     <span
       className={classnames('icon-with-fallback__fallback', fallbackClassName)}
     >
-      {name.length ? name.charAt(0).toUpperCase() : ''}
+      {name && name.length ? name.charAt(0).toUpperCase() : ''}
     </span>
   );
 };
--- a/yarn.lock
+++ b/yarn.lock
@@ -1462,6 +1462,21 @@
     "@ethersproject/properties" "^5.5.0"
     "@ethersproject/strings" "^5.5.0"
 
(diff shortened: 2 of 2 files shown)


mcmire (approved):

ui/components/ui/icon-with-fallback/icon-with-fallback.component.js
```diff
@@ -23,14 +23,14 @@ const IconWithFallback = ({
...
-      alt={name.length ? name : 'icon'}
+      alt={name || 'icon'}
       {...props}
     />
   ) : (
     <span
       className={classnames('icon-with-fallback__fallback', fallbackClassName)}
     >
-      {name.length ? name.charAt(0).toUpperCase() : ''}
+      {name && name.length ? name.charAt(0).toUpperCase() : ''}
```
darkwing, line 33: Could this be `name?.length` instead?

adonesky1 (approved):

rekmarks (approved):

ui/components/ui/icon-with-fallback/icon-with-fallback.component.js
rekmarks: Initially I thought not but that would be equivalent. I think we should merge this anyway because really our linter should handle this sort of thing :thinking:

PR #13979, opened 2022-03-16
Add `ShowHideToggle` component

Gudahtt:
Add a new component for controlling whether a field should be shown or hidden. This will be used in later PRs as a control for sensitive fields that are hidden by default.

This component should be fully accessible. Both mouse and keyboard interactions have been tested, and `aria-label` attributes have been added to explain the two "eye" icons that don't have any corresponding text.

Thorough unit tests have been written, testing all props except `className` (I don't know how to test that using Jest/`jsdom`).

This was adapted from https://github.com/MetaMask/metamask-extension/pull/13957, so most of the credit here goes to @georgewrmarshall

## Screenshots/Screencaps

show-hide

## Manual testing steps

This has not yet used anywhere, so the best you can do is check that it works correctly in Storybook.

### DIFF
--- a/ui/components/ui/show-hide-toggle/index.scss
+++ b/ui/components/ui/show-hide-toggle/index.scss
@@ -1,0 +1,34 @@
+.show-hide-toggle {
+  position: relative;
+  display: inline-flex;
+
+  &__input {
+    appearance: none;
+
+    + .show-hide-toggle__label {
+      cursor: pointer;
+      user-select: none;
+    }
+
+    /* Focused when tabbing with keyboard */
+    &:focus,
+    &:focus-visible {
+      outline: none;
+
+      + .show-hide-toggle__label {
+        outline: Highlight auto 1px;
+      }
+    }
+
+    &:disabled {
+      + label {
+        opacity: 0.5;
+        cursor: auto;
+      }
+    }
+  }
+
+  &__icon {
+    color: var(--color-icon-default);
+  }
+}
--- a/ui/components/ui/show-hide-toggle/show-hide-toggle.js
+++ b/ui/components/ui/show-hide-toggle/show-hide-toggle.js
@@ -1,0 +1,86 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+import IconEye from '../icon/icon-eye';
+import IconEyeSlash from '../icon/icon-eye-slash';
+
+const ShowHideToggle = ({
+  id,
+  shown,
+  onChange,
+  ariaLabelHidden,
+  ariaLabelShown,
+  className,
+  'data-testid': dataTestId,
+  disabled,
+  title,
+}) => {
+  return (
+    <div className={classnames('show-hide-toggle', className)}>
+      <input
+        className="show-hide-toggle__input"
+        id={id}
+        type="checkbox"
+        checked={shown}
+        onChange={onChange}
+        data-testid={dataTestId}
+        disabled={disabled}
+      />
+      <label htmlFor={id} className="show-hide-toggle__label" title={title}>
+        {shown ? (
+          <IconEye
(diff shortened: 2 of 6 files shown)


Gudahtt:
This depends upon #13978

ui/components/ui/show-hide-toggle/show-hide-toggle.stories.js
```diff
@@ -0,0 +1,51 @@
...
+    },
+    disabled: {
+      control: 'boolean',
+    },
+    onChange: {
+      action: 'onChange',
+    },
+    shown: {
+      control: 'boolean',
+    },
```
georgewrmarshall, line 32: Could add title here. I believe the story should show all available props in the controls
```suggestion
    },
    title: {
      control: 'text',
    },
```

ui/components/ui/show-hide-toggle/show-hide-toggle.js
```diff
@@ -0,0 +1,86 @@
...
+
+import IconEye from '../icon/icon-eye';
+import IconEyeSlash from '../icon/icon-eye-slash';
+
+const ShowHideToggle = ({
+  id,
+  shown,
+  onChange,
+  ariaLabelHidden,
+  ariaLabelShown,
```
georgewrmarshall, line 13: Do you think these aria labels should have defaults? I would imagine that most devs wouldn't need to change them seeing as the component is to specifically communicate show and hide? Could possibly save some cognitive load on the developer using this component?

```
  ariaLabelHidden = 'hidden',
  ariaLabelShown = 'shown',
```

georgewrmarshall (commented):
Fantastic work 💯 made a couple small suggestions.

ui/components/ui/show-hide-toggle/show-hide-toggle.js
Gudahtt: They are required because they need to be localized. Unless you think we should bring the `t` context into this? I figured it would be best to avoid a dependence upon Redux state, but the locale is sort of a special case... for some UI components it would seem silly to ask all text to be passed in. But anyway.

georgewrmarshall: Ahh makes sense! Hmm I definitely need to do some research into how to best to manage translations in regards to reusable ui components. I think keeping it external makes a lot of sense. Some more overhead for the developer but ensures proper translations and keeps the component light. Nice.

georgewrmarshall (dismissed):
LGTM! 💯

georgewrmarshall (approved):

ui/components/ui/show-hide-toggle/index.scss
```diff
@@ -0,0 +1,34 @@
...
+      user-select: none;
+    }
+
+    /* Focused when tabbing with keyboard */
+    &:focus,
+    &:focus-visible {
+      outline: none;
+
+      + .show-hide-toggle__label {
+        outline: Highlight auto 1px;
```
darkwing, line 19: Where does `Highlight` come from?  Never seen this before, would love to know what the reference is!

darkwing (approved):

ui/components/ui/show-hide-toggle/index.scss
Gudahtt: Oh, my bad! Should have left a comment for this. This is a "System color": https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#system_colors

Gudahtt: And here's the W3C spec for it: https://www.w3.org/TR/css-color-4/#css-system-colors

I saw a few references to this color having been deprecated, but it seems that W3C went ahead and un-deprecated it. So it should be safe to use. It works on Chrome and Firefox as well.

PR #13916, opened 2022-03-11
Dark Mode: ENS Input

darkwing:
Fixes the ENS input for dark mode.

ENSInput

You can see this component after clicking the SEND button.

PR #13984, opened 2022-03-16
Dark Mode: ENS Input

darkwing:
Supercedes: https://github.com/MetaMask/metamask-extension/pull/13916

Fixes the ENS input for dark mode.

ENSInput

You can see this component after clicking the SEND button.


georgewrmarshall (dismissed):
Eventually I want to update all these icons and I suggest we stop using background images with svgs but this is good for now!

mcmire (approved):

brad-decker (approved):

PR #13986, opened 2022-03-16
Ensure accounts still line up in dropdown

darkwing:
We recently changed the selected account checkmark in from static image to `<IconCheck />` in `ui/components/app/account-menu/account-menu.component.js`.  Unfortunately accounts don't line up when one is checked; this fixes that issue.
LineUp


georgewrmarshall (approved):
Oh that is totally my bad! Thanks for the fix!!

brad-decker (approved):

PR #14002, opened 2022-03-16
Enable Token Detection by default on supported networks

### LINKED ISSUE #13942: Enable Token Detection by default on supported networks
Token Detection Functionality:

- Enable token detection by default
- Enable token detection on Binance, Polygon, and Avalanche.

NiranjanaBinoy:
Fix #13942 
Enabling token detection by default on supported networks.


app/scripts/controllers/preferences.js
```diff
@@ -38,7 +38,8 @@ export default class PreferencesController {
 
       // set to true means the dynamic list from the API is being used
       // set to false will be using the static list from contract-metadata
-      useTokenDetection: false,
+      // eslint-disable-next-line
+      useTokenDetection: process.env.TOKEN_DETECTION_V2 ? true : false,
```
darkwing, line 42: I think `Boolean(process.env.TOKEN_DETECTION_V2)` might be cleaner

app/scripts/controllers/preferences.test.js
```diff
@@ -274,21 +274,37 @@ describe('preferences controller', function () {
     });
   });
   describe('setUseTokenDetection', function () {
-    it('should default to false', function () {
+    it('should default to true', function () {
       const state = preferencesController.store.getState();
-      assert.equal(state.useTokenDetection, false);
+      if (process.env.TOKEN_DETECTION_V2) {
```
darkwing, line 279: Are these tests run multiple times with the flag on and then off?  If not, we should have two separate tests where the flag is shimmed on, and then off.

darkwing (changes requested):

digiwand (dismissed):

app/scripts/controllers/preferences.test.js
```diff
@@ -39,6 +40,7 @@ describe('preferences controller', function () {
   });
 
   afterEach(function () {
+    process.env.TOKEN_DETECTION_V2 = false;
```
digiwand, line 43: I think it could be better to restore the process.env value to its original value. 

Something like.
https://stackoverflow.com/questions/48033841/test-process-env-with-jest

For our case, I think we could use the `beforeAll` and `afterAll` hooks instead of `beforeEach` and `afterEach` since we won't be updating the values in the individual tests.

app/scripts/controllers/preferences.test.js
```diff
@@ -274,20 +277,21 @@ describe('preferences controller', function () {
     });
   });
   describe('setUseTokenDetection', function () {
-    it('should default to false', function () {
+    process.env.TOKEN_DETECTION_V2 = true;
```
digiwand, line 280: We should move this into the test case itself, `it`, or in the `before`/`after` hooks. The logic directly placed into the `describe` block won't be part of the setup and teardown of the test suite itself.

```
  describe.only('Testing order of calls', function () {
    console.log('1 - describe');

    before(function () {
      return console.log('1 - before');
    });
    beforeEach(function () {
      return console.log('1 - beforeEach');
    });
    afterEach(function () {
      return console.log('1 - afterEach');
    });
    after(function () {
      return console.log('1 - after');
    });

    it('Test in describe 1', function () {
      return console.log('1 - test');
    });

    context('Scoped / Nested block', function () {
      console.log('2 - describe');

      it('Test in describe 2', function () {
        return console.log('2 - test');
      });
    });
  });
  ```
  will produce
  ```
1 - describe
2 - describe


  Transaction Controller
    Testing order of calls
1 - before
1 - beforeEach
1 - test
      ✓ Test in describe 1
1 - afterEach
      Scoped / Nested block
1 - beforeEach
2 - test
        ✓ Test in describe 2
1 - afterEach
1 - after
  ```

**Review by digiwand — CHANGES_REQUESTED** [review]
requesting to update the test logic. To clarify, it seems that by default, this flag should still be false if the environment variable is not found. Is this expected behavior?

File: app/scripts/controllers/preferences.test.js [hunk 5]
```diff
@@ -274,7 +277,7 @@ describe('preferences controller', function () {
     });
   });
   describe('setUseTokenDetection', function () {
-    it('should default to false', function () {
+    it('should default to true', function () {
```
digiwand, line 280:
```suggestion
    it('should default to false', function () {
```

digiwand (approved):
✅

darkwing (approved):

PR #14004, opened 2022-03-16
Add `pasteIntoField` webdriver method

Gudahtt:
A new method has been added to the e2e webdriver for pasting text into a field. This will be required to properly test a change to the SRP input, which will be coming in a separate PR.

A few existing e2e tests have been updated to use this method to input the SRP, to show that it works properly.


danjm (approved):
LGTM

Gtonizuka (approved):

darkwing (approved):

Issue #13856, opened 2022-03-07
[Bug]: Design breaks on the Settings Navbar when Internationalization


seaona:
### Describe the bug

When I choose another language, I see that the design breaks on the left navbar for Advance Settings.

**Example1: Spanish**
![image](https://user-images.githubusercontent.com/54408225/157067087-0b0b1ccb-658b-4449-aca5-c6f0651d8178.png)

**Example2: Catalan**
![image](https://user-images.githubusercontent.com/54408225/157067279-6c084574-e1c2-4fc1-b869-88bacb1ae16e.png)

**Note**: this happens with most of the languages

### Steps to reproduce

1. Open the Expanded View
2. Go to Menu // Settings
3. Change the Language (i.e. Spanish)

### Version

10.10.2

### Browser

Firefox

### Operating system

Linux

PR #14012, opened 2022-03-17
fix design breaks on the Settings Navbar when Internationalization

Hyun2:
**Fixes**: #13856 

**Explanation:**
No need `flex-flow: row wrap;` in `.tab-bar__tab__content`

- `flex-direction` is already row .
- It seems that `flex-wrap` was set to `wrap` to express in multiple lines when the text in `.tab-bar__tab__content__title` becomes long. But this has no effect if the text is long.

<br/>

**Manual testing steps:**
1. Open the Expanded View
2. Go to Menu // Settings
3. Change the Language (i.e. Spanish)

**Before**
스크린샷 2022-03-17 오후 12 37 05

**After**
스크린샷 2022-03-17 오후 12 36 54


Hyun2:

brad-decker (approved):
LGTM! Thanks for the contribution!

darkwing (approved):

PR #14013, opened 2022-03-17
Remove the useSegmentContext function from metametrics.new.js

### LINKED ISSUE #13768: Remove the useSegmentContext function from metametrics.new.js
In `ui/contexts/metametrics.new.js`  a hook is defined called `useSegmentContext` this context function was later extracted to the `useMetricEvent` hook file as `useMetaMetricsContext` hook. The goal was to import that hook in the context itself instead of using the old `useSegmentContext` but it was never done. The goal of this ticket is to reduce confusion with what this hook is used for

## Requirements
1. delete the `useSegmentContext` hook from the `metametrics.new.js` file
2. rename `useMetaMetricsContext` to something that doesn't use the word context. (these are passed as the `context` property to the segment payload, but react also has contexts and without proper context, its hard to know which context you're contexting... yes i'm being purposefully obtuse here).
3. Move the above mentioned hook into its own file, update imports.
4. Import that hook into `metametrics.new.js` and replace usages of `useSegmentContext` with it.

VSaric:
fixes: #13768

## Explanation
Remove the `useSegmentContext` function from `metametrics.new.js` and create new file for `useMetaMetrics` hook.

### DIFF
--- a/ui/hooks/useMetaMetrics.js
+++ b/ui/hooks/useMetaMetrics.js
@@ -16,7 +16,7 @@ const PATHS_TO_CHECK = Object.keys(PATH_NAME_MAP);
  *  referrer?: MetaMetricsReferrerObject
  * }}
  */
-export function useMetaMetricsContext() {
+export function useMetaMetrics() {
   const match = useRouteMatch({
     path: PATHS_TO_CHECK,
     exact: true,
--- a/ui/contexts/metametrics.new.js
+++ b/ui/contexts/metametrics.new.js
@@ -10,15 +10,14 @@ import React, {
   useRef,
   useCallback,
 } from 'react';
-import { useSelector } from 'react-redux';
 import PropTypes from 'prop-types';
-import { matchPath, useLocation, useRouteMatch } from 'react-router-dom';
+import { matchPath, useLocation } from 'react-router-dom';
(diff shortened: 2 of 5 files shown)


ui/hooks/useMetaMetrics.js
```diff
@@ -0,0 +1,46 @@
...
+ * if there is a confirmation that was triggered by a dapp. These values are
+ * not required but add valuable context to events, and should be included in
+ * the context object on the event payload.
+ *
+ * @returns {{
+ *  page?: MetaMetricsPageObject
+ *  referrer?: MetaMetricsReferrerObject
+ * }}
+ */
+export function useMetaMetrics() {
```
brad-decker, line 19: [REQ] This name doesn't match the outcome, or rather it gives the impression that this in some way is "using" MetaMetrics.... after talking with a few others, lets use `useSegmentContext` as it is the best name for the outcome.

brad-decker (changes requested):

brad-decker (dismissed):

mirjanaKukic:
Verified by QA

digiwand (dismissed):

brad-decker (approved):

digiwand (approved):

PR #14016, opened 2022-03-17
Split SRP input by word

Gudahtt:
## Explanation

Split the secret recovery phrase input field into one field per word. Each word uses a `type="password"` field. There is a "show/hide toggle" next to each word allowing you to reveal it, but only one word can be revealed at a time.

This is a new design aimed at making our SRP input component more secure. This design makes it impossible to accidentally reveal your entire SRP to anybody nearby. It also acts as additional protection against the complete SRP being stored in a browser autocomplete database (though to be clear, we could find no evidence this was ever occurring).

## More information

Figma design: https://www.figma.com/file/8z4Csc3wlOIkUz2OP1eTZW/Secure-UX-Improvements?node-id=1241%3A12046

There have been some changes from the original design (e.g. the addition of the tip about pasting). Also that design was for the "Onboarding V2" flow, so it was not followed exactly. For now this PR targets the old onboarding flow. I will add this new component to the new onboarding flow in the next PR after this.

### Before

old

### After

new

https://user-images.githubusercontent.com/2459287/158829027-1f21e96f-f100-4d74-98b7-c734fe582074.mp4

## Manual testing steps

The two screens that have changed are the "Import" page of onboarding, and the "Restore wallet" page accessible from the lock screen. The component and how it behaves should be identical in both places.

Suggestions on which behaviors to test:
 * It should allow manually typing in the SRP, or pasting the entire thing, or pasting each word individually.
 * The clipboard should be cleared on paste, *unless* the contents are too long (>24 words). In that case, the clipboard is preserved and the paste event is ignored (i.e. no fields are changed).
 * This should be fully navigable via keyboard, and the show/hide toggle should be toggle-able using the spacebar while focused.
 * All fields should be hidden on paste.
 * Only one field should be revealed at a time. Revealing any field hides the others.


Gudahtt:
>Could this possibly be annoying for some users who want to be able to visually confirm the whole phrase? Or atleast in larger chunks than 1 at a time?

Potentially, yes. But this was the tradeoff for making it safer to input. Unsure how we could make the process of checking the entire SRP easier. Right now the process would be "tab + tab + space" to navigate to the next word and reveal it.

holantonela:
> Or atleast in larger chunks than 1 at a time?
This is interesting.

> Unsure how we could make the process of checking the entire SRP easier
We could do that by rendering the entire as an image.

brad-decker (approved):
um okay i love this. Great work

brad-decker:
I checked all steps on chrome and brave and it worked like a charm.

adonesky1 (approved):
LGTM!

PR #14019, opened 2022-03-17
add space between sentences in insufficient currency warning

adonesky1:
Just noticed we need a space here:
Screen Shot 2022-03-17 at 12 01 05 PM

After these changes:
Screen Shot 2022-03-17 at 12 10 42 PM

After some further comments/discussion this PR grew to include the addition of the inline button/link 'Buy <nativeCurrent>' that pops the buy modal for **all chains on which we support a fiat onramp**. For instance on `polygon`:
159537034-6506cfe6-1f56-4315-ba1e-4bd97c2da593

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -420,18 +420,6 @@
     "message": "Transak supports debit card and bank transfers (depending on location) in 59+ countries. $1 deposits into your MetaMask account.",
     "description": "$1 represents the cypto symbol to be purchased"
   },
-  "insufficientCurrency": {
-    "message": "You do not have enough $1 in your account to pay for transaction fees on $2 network.",
-    "description": "$1 is the native currency of the network, $2 is the name of the current network"
-  },
-  "insufficientCurrencyBuyOrDeposit": {
-    "message": "You do not have enough $1 in your account to pay for transaction fees on $2 network. $3 Deposit from another account.",
-    "description": "$1 is the native currency of the network, $2 is the name of the current network, $3 is the key 'buy' + the ticker symbol of the native currency of the chain + the key 'or' all wrapped in a button"
-  },
-  "depositFromOtherAccount": {
-    "message": "Deposit $1 from another account.",
-    "description": "$1 is the ticker symbol of the native currency of the chain the user is currently on"
-  },
   "buyWithWyre": {
     "message": "Buy ETH with Wyre"
   },
@@ -1529,6 +1517,10 @@
   },
   "insufficientBalance": {
     "message": "Insufficient balance."
+  },
+  "insufficientCurrencyBuyOrDeposit": {
+    "message": "You do not have enough $1 in your account to pay for transaction fees on $2 network. $3 Deposit from another account.",
+    "description": "$1 is the native currency of the network, $2 is the name of the current network, $3 is the key 'buy' + the ticker symbol of the native currency of the chain + the key 'or' all wrapped in a button"
   },
   "insufficientFunds": {
     "message": "Insufficient funds."
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -420,12 +420,13 @@
     "message": "Transak supports debit card and bank transfers (depending on location) in 59+ countries. $1 deposits into your MetaMask account.",
     "description": "$1 represents the cypto symbol to be purchased"
   },
-  "buyEth": {
-    "message": "Buy ETH"
+  "buyNativeCurrency": {
+    "message": "$1 or deposit from another account.",
+    "description": "$1 is the key 'buy' and the ticker symbol of the native currency of the chain the user is currently on wrapped in a button"
   },
-  "buyOther": {
-    "message": "Buy $1 or deposit from another account.",
-    "description": "$1 is a token symbol"
+  "depositFromOtherAccount": {
+    "message": "Deposit $1 from another account.",
+    "description": "$1 is the ticker symbol of the native currency of the chain the user is currently on"
   },
   "buyWithWyre": {
     "message": "Buy ETH with Wyre"
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -405,6 +405,10 @@
   "buy": {
     "message": "Buy"
   },
+  "buyAsset": {
+    "message": "Buy $1",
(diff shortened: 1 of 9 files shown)


danjm (commented):
LGTM

ui/components/app/confirm-page-container/confirm-page-container-content/confirm-page-container-content.component.js
```diff
@@ -224,7 +224,7 @@ export default class ConfirmPageContainerContent extends Component {
                 className="actionable-message--warning"
                 message={
                   <Typography variant={TYPOGRAPHY.H7} align="left">
-                    {t('insufficientCurrency', [nativeCurrency, networkName])}
+                    {t('insufficientCurrency', [nativeCurrency, networkName])}{' '}
```
Gudahtt, line 227: Could this be merged into one message? Splitting adjacent text into multiple messages is generally bad for localization because in some locales the messages might be changed by the surrounding context. Even the number of sentences might differ between locales.

adonesky1: Yea I don't see why not... Not sure why it was done this way originally.

ui/components/app/confirm-page-container/confirm-page-container-content/confirm-page-container-content.component.js
```diff
@@ -224,7 +224,7 @@ export default class ConfirmPageContainerContent extends Component {
...
                     {t('buyOther', [nativeCurrency])}
```
adonesky1, line 228: So currently for non mainnet networks we show this same message telling the user to `buy` the native currency of the network but we don't link them to any resource to do so:
Screen Shot 2022-03-18 at 9 36 30 AM

In cases like `MATIC` and `AVAX` we should probably link to the buy modal since we now support it. But for networks where we don't support a fiat onramp should we remove the "Buy __ or deposit from another account." for just "Deposit ___ from another account."? cc @rachelcope @SayaGT

SayaGT: @adonesky1 yes that sounds correct. For the network that don't support a fiat onramp, just show "Deposit ___ from another account."

On separate note, can we activate **Buy** button for these networks without on ramp option. We can show "**Directly Deposit _Token_**". It basically let's user see their public account address. It will be a consistent experience across networks. What do you think @rachelcope 
Screenshot 2022-03-18 at 15 01 52

adonesky1:
Did some further tweaking here to provide a "buy" link that opens the buy/fiat onramp modal as an inline link for all networks where we provide such an onramp:
Screen Shot 2022-03-22 at 12 11 28 PM


And then this is what it looks like for networks where no onramp is available:
Screen Shot 2022-03-22 at 11 00 45 AM

Gudahtt:
We should use two separate messages for these two cases. This is making too many assumptions about how this message can be structured in other locales. This is forcing translators to use the "or" message to convey that there are two choices, but that might put them in the position where they can't make both options grammatically correct (i.e. if the existence of the first option changes the rest of the sentence). Moreover, we want our translators to express the idea in whatever way is most natural in that locale. Even if it is possible to express this in a way that is grammatically correct this is too constraining.

adonesky1:
@Gudahtt thanks for your feedback here. You're totally right that by breaking these phrases up too much translators lose all the required context. I have separated the two cases - when the users is on a "buyable" chain, and not - into two strings `insufficientCurrencyBuyOrDeposit` and `insuffienceCurrencyDeposit`.

ui/components/ui/button/buttons.scss
```diff
@@ -302,7 +302,7 @@ input[type="submit"][disabled] {
 
 .btn--inline {
   display: inline;
-  padding: 0 4px;
+  padding: 0;
```
adonesky1, line 305: @georgewrmarshall What do you think of this? I'm not sure why we would want r-l padding on inline buttons? It makes blending them with other texts more awkward.

georgewrmarshall: Hmm initially I thought "yah you're right!" but then I checked storybook and the inline button type. It seems the padding is required for context in which it is used in this story. Removing the padding removes the space altogether. Are you seeing it differently in your context?

[Inline type button in storybook](https://metamask.github.io/metamask-storybook/index.html?path=/story/ui-components-ui-button-button-stories-js--type-inline)

Removing the padding

Screen Shot 2022-03-23 at 2 59 13 PM

ui/pages/confirm-transaction-base/transaction-alerts/transaction-alerts.js
```diff
@@ -96,39 +95,33 @@ const TransactionAlerts = ({
...
+              <Typography variant={TYPOGRAPHY.H7} align="left">
+                {t('insufficientCurrencyDeposit', [
+                  nativeCurrency,
+                  networkName,
+                ])}
+              </Typography>
+            )
           }
           useIcon
           iconFillColor="#d73a49"
```
georgewrmarshall, line 127: Can we update this one while we are at it?

```suggestion
          iconFillColor="var(--color-error-default)"
```

georgewrmarshall (changes requested):
Looking good! Just adding some color suggestions here. Down with static hex values ✊

ui/components/ui/button/buttons.scss
georgewrmarshall: Ok to remove the padding of the inline type button after our discussions and some exploration 👍  🙏

georgewrmarshall (approved):
LGTM! Thorough as always 💯

app/_locales/en/messages.json
```diff
@@ -1525,9 +1522,13 @@
   "insufficientBalance": {
     "message": "Insufficient balance."
   },
-  "insufficientCurrency": {
-    "message": "You do not have enough $1 in your account to pay for transaction fees on $2 network.",
-    "description": "$1 is currency, $2 is network"
+  "insufficientCurrencyBuyOrDeposit": {
+    "message": "You do not have enough $1 in your account to pay for transaction fees on $2 network. $3 or deposit from another account.",
```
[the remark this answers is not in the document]
adonesky1: Yes this introduces the first instances of the button with the `inline` type.

segun (approved):
LGTM

PR #14025, opened 2022-03-17
Dark Mode:  Remove unused colors

darkwing:
In an effort to help identify colors we've completely remove, I did a search to remove the ones not used.

### DIFF
--- a/ui/css/utilities/colors.scss
+++ b/ui/css/utilities/colors.scss
@@ -2,15 +2,11 @@
   // These are the colors of the MetaMask design system
   // Only design system colors should be added, no superfluous variables
   --Blue-000: #eaf6ff;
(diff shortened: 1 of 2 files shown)


georgewrmarshall (dismissed):
This is awesome! Did a search for all proposed color removals and came back without any results so LGTM!!

digiwand (dismissed):
Looks good to me! Extra confidence in this approval since George did a double-check on each of the variable removals 🎉 

I think another option would be to comment out unused variables, but I also imagine we can recover the various shades from somewhere if we wanted to add them back in

georgewrmarshall (approved):

brad-decker (approved):
The colors just had their very own `--red-wedding`

PR #14027, opened 2022-03-17
Dark Mode: Implement 3-Dot Menu Hover

darkwing:
Right now there's no menu hover effect which isn't great; this hover is an improvement:
hovers


georgewrmarshall (approved):
LGTM!

brad-decker (approved):

PR #14030, opened 2022-03-18
Dark Mode: Modals

darkwing:
Modals have hardcoded white colors, so I've transitioned colors for dark mode.
ResetAcct

Also includes QR reader styles:
QR

### DIFF
--- a/ui/components/app/modals/index.scss
+++ b/ui/components/app/modals/index.scss
@@ -24,7 +24,7 @@
 
   &__content {
     margin: 0;
-    background-color: white;
+    background-color: transparent;
     animation-fill-mode: forwards;
   }
 
@@ -35,7 +35,7 @@
     bottom: 0;
     left: 0;
     z-index: 1040;
-    background-color: #373a47;
+    background-color: #373a47; /* TODO: What color here? */
     animation-fill-mode: forwards;
     animation-duration: 0.3s;
   }
--- a/ui/components/app/modals/index.scss
+++ b/ui/components/app/modals/index.scss
@@ -35,7 +35,7 @@
     bottom: 0;
     left: 0;
     z-index: 1040;
-    background-color: #373a47; /* TODO: What color here? */
+    background-color: var(--color-overlay-alterantive);
     animation-fill-mode: forwards;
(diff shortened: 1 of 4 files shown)


ui/components/app/modals/index.scss
```diff
@@ -35,7 +35,7 @@
     bottom: 0;
     left: 0;
     z-index: 1040;
-    background-color: #373a47;
+    background-color: #373a47; /* TODO: What color here? */
```
darkwing, line 38: @georgewrmarshall @GuillaumeRx What color would be best here?

ui/components/app/modals/qr-scanner/index.scss
```diff
@@ -67,16 +67,16 @@
 
     button:last-of-type {
       margin-right: 0;
-      background-color: #009eec;
+      background-color: var(--color-primary-default);
       border: none;
-      color: #fff;
+      color: var(--color-text-default);
```
GuillaumeRx, line 72:
```suggestion
      color: var(--color-primary-inverse);
```

GuillaumeRx (changes requested):

GuillaumeRx (dismissed):

darkwing:
I'm cool with the background color changes; I used `transparent` because it inherited the same background.

ui/components/app/modals/index.scss
georgewrmarshall: As it's a background color I would suggest `var(--color-overlay-alterantive)` `--color-text-default` shouldn't be used for any background-color really. I'm trying to find out how and where this is used.

georgewrmarshall: Doesn't look like this is even being used but I think it should be and overlay. I would suggest updating it to
```suggestion
    background-color: var(--color-overlay-alterantive);
```

georgewrmarshall (commented):
Looks good. I have one suggestion

darkwing:
Updated!

georgewrmarshall (approved):

brad-decker (approved):
Changes look good to me.

PR #14032, opened 2022-03-18
Remove AddToAddressBook modal

darkwing:
In as far as I can tell, this modal isn't actually used anymore, so we can delete its modal key and components.


georgewrmarshall (approved):
lgtm

brad-decker (approved):
love code deletion PRs

PR #14045, opened 2022-03-18
Fix flaky test: add account

seaona:
## Problem
- Add Account is a flaky test, that sometimes fails on our circleCI and local runs

## Solution
- Included `waitForSelectors` before grabbing the Account addresses, for avoiding intermitent failures, as we ensure that the element is there, before accessing to its value.

## Screenshots/Screencaps
image


darkwing (approved):
¡Que golazo!

BlockDaRock (approved):

PeterYinusa (approved):

PR #14046, opened 2022-03-18
Updating other colors

georgewrmarshall:
Other css color variables in `ui/css/itcss/components/send.scss`


ui/components/ui/unit-input/index.scss
```diff
@@ -65,8 +61,4 @@
   &--error {
     border-color: var(--color-error-default);
   }
-
-  &__disabled {
-    background-color: #f2f3f4;
-  }
 }
```
georgewrmarshall, line 64: This don't seem to be used anywhere and don't make sense to me.

Issue #13606, opened 2022-02-11
Add friendly error handling when background throws an error before listening for connection


mcmire:
## Overview

In the architecture of the extension there are two main kinds of scripts: content scripts (ephemeral scripts that are loaded on pages that power the UI for the popup and full-screen views as well as the integration code that runs on dapps) and the background script (a persistent script that handles state changes in response to the UI). Since we are only concerned with the UI views, colloquially we will call these the "UI" and the "background".

The UI and background need to talk to each other. For instance, when the user wants to send a transaction, they need to see gas estimates, so the UI asks the background for this information. On the other side, the background may need to tell the UI layer to force refresh what the user sees based on the current state held in the background.

Both sides use a port exposed by the extension to communicate with each other. This [bidirectional communication](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Content_scripts#communicating_with_background_scripts) works as follows:

1. The background and UI start loading simultaneously.
2. The background adds a handler that will be triggered when the UI connects.
3. The UI connects to the port, triggering the handler that the background has set up.
4. The background adds a handler to receive messages from the UI and handle them appropriately.
5. The UI adds a handler to receive messages from the background and handle them appropriately.

Here's the problem: What happens if the background throws an error before it can start listening to the UI? The UI will proceed to send messages to the background, but they will go nowhere and there will be no response from the background. If the UI layer needs information from the background to initialize the UI and that information is not there, then the UI will be blank.

Some errors are caused by glitches or race conditions (the user clicks on the MetaMask icon to bring up the popup before the background completely loads). In this case the user can simply reload the extension to fix the issue, so no real problem.

But some errors are irrecoverable. In #9196, for the majority of users reporting a blank screen, the problem is that the [database which the extension uses to persist the app state to disk](https://blog.mozilla.org/addons/2018/08/03/new-backend-for-storage-local-api/) gets corrupted in Firefox somehow. We are currently researching solutions for recovering this data, but we need to handle this properly in the meantime.

[...]

PR #14047, opened 2022-03-18
Fix delayed loading of UI in MV3

### LINKED ISSUE #14001: MV3: It takes refreshing browser for extension UI to be loaded
We are noticing an issue in MV3 code in branch `mv3`. It takes reloading to see the wallet in browser.

Reason I found was that UI was loaded before extension was ready to send / receive RPC messages.

jpuri:
Fixes: #14001

At times we see extension page is blank and there is error in console that `receiving end does not exist`. It takes refreshing browser for error to go away. This issue was more evident in MV3 where extension was not loading unless we refresh the browser.

I found that reason is that UI is loaded before extension is ready for RPC method calls. The PR tries to address the issue by sending a message from extension to UI when it is ready to receive messages.

Here is a following improvement task planned: https://github.com/MetaMask/metamask-extension/issues/14050

### DIFF
--- a/app/scripts/background.js
+++ b/app/scripts/background.js
@@ -45,6 +45,14 @@ import { getPlatform } from './lib/util';
 const { sentry } = global;
 const firstTimeState = { ...rawFirstTimeState };
 
+const metamaskBlockedPorts = ['trezor-connect'];
+
+const metamaskInternalProcessHash = {
+  [ENVIRONMENT_TYPE_POPUP]: true,
+  [ENVIRONMENT_TYPE_NOTIFICATION]: true,
+  [ENVIRONMENT_TYPE_FULLSCREEN]: true,
+};
+
 log.setDefaultLevel(process.env.METAMASK_DEBUG ? 'debug' : 'info');
 
 const platform = new ExtensionPlatform();
@@ -67,8 +75,19 @@ if (inTest || process.env.METAMASK_DEBUG) {
   global.metamaskGetState = localStore.get.bind(localStore);
 }
 
-// initialization flow
-initialize().catch(log.error);
+let initialState;
+let initialLangCode;
+const initApp = (remotePort) => initialize(remotePort).catch(log.error);
+
+if (process.env.ENABLE_MV3) {
+  browser.runtime.onConnect.addListener(initApp);
+  (async () => {
+    initialState = await loadStateFromPersistence();
+    initialLangCode = await getFirstPreferredLangCode();
+  })();
+} else {
+  initialize().catch(log.error);
+}
 
 /**
  * @typedef {import('../../shared/constants/transaction').TransactionMeta} TransactionMeta
@@ -128,12 +147,17 @@ initialize().catch(log.error);
 /**
  * Initializes the MetaMask controller, and sets up all platform configuration.
  *
+ * @param {string} remotePort - remote application port connecting to extension.
  * @returns {Promise} Setup complete.
  */
-async function initialize() {
-  const initState = await loadStateFromPersistence();
-  const initLangCode = await getFirstPreferredLangCode();
-  await setupController(initState, initLangCode);
+async function initialize(remotePort) {
+  if (!initialState) {
+    initialState = await loadStateFromPersistence();
+  }
+  if (!initialLangCode) {
+    initialLangCode = await getFirstPreferredLangCode();
+  }
+  await setupController(initialState, initialLangCode, remotePort);
   log.info('MetaMask initialization complete.');
 }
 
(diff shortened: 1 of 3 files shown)


app/scripts/background.js
```diff
@@ -67,8 +67,8 @@ if (inTest || process.env.METAMASK_DEBUG) {
   global.metamaskGetState = localStore.get.bind(localStore);
 }
 
-// initialization flow
-initialize().catch(log.error);
+const initApp = initialize().catch(log.error);
+extension.runtime.onConnect.addListener(initApp);
```
jpuri, line 71: Having an event handler at top level is helpful for MV3, I found that at times in MV3 the first event was missed is user opened extension immediately after installing it.

The code above is good for MV2 also, An optimisation we should do here is, initialise the controllers outside the listeners. Reactive programming will be useful here.

mcmire:
Could this partially address the ideas I proposed in #13606 as well?

jpuri:
> Could this partially address the ideas I proposed in #13606 as well?

Thanks for sharing this @mcmire , yes it is related and will partially address #13606.

app/scripts/background.js
```diff
@@ -367,7 +386,7 @@ function setupController(initState, initLangCode) {
       // communication with popup
       controller.isClientOpen = true;
       controller.setupTrustedCommunication(portStream, remotePort.sender);
-
+      remotePort.postMessage({ name: 'CONNECTION_READY' });
```
jpuri, line 389: This message is causing warning:
Screenshot 2022-03-22 at 6 22 58 AM

app/scripts/background.js
```diff
@@ -67,8 +75,16 @@ if (inTest || process.env.METAMASK_DEBUG) {
...
+let initialState;
+let initialLangCode;
+
+const initApp = (remotePort) => initialize(remotePort).catch(log.error);
+browser.runtime.onConnect.addListener(initApp);
+
+(async () => {
+  initialState = await loadStateFromPersistence();
+  initialLangCode = await getFirstPreferredLangCode();
+})();
```
jpuri, line 87: Event handler at root level is required for mv3, otherwise in mv3 if user open UI very fast the first event to connect is missed out.

app/scripts/background.js
```diff
@@ -67,8 +75,19 @@ if (inTest || process.env.METAMASK_DEBUG) {
...
 }
 
-// initialization flow
-initialize().catch(log.error);
+let initialState;
+let initialLangCode;
+const initApp = (remotePort) => initialize(remotePort).catch(log.error);
+
+if (process.env.ENABLE_MV3) {
+  browser.runtime.onConnect.addListener(initApp);
```
adonesky1, line 83: Can you help me understand why we would only want to do this (pass along the remotePort as an argument) in the context of MV3?  cc @jpuri

app/scripts/background.js
```diff
@@ -291,20 +316,19 @@ function setupController(initState, initLangCode) {
     }
   }
 
+  if (process.env.ENABLE_MV3 && remoteSourcePort) {
+    connectRemote(remoteSourcePort);
```
adonesky1, line 320: same question here. Why not simply add the listener (`browser.runtime.onConnect.addListener(connectRemote);`) at the root of the file?

app/scripts/background.js
jpuri: In MV3 it is better to have event listener at root level else it is possible that first event connect is lost if user open extension immediately after installing the extension.
This parameter is required for connect method to work.

app/scripts/background.js
jpuri: `connectRemote` use controller object which is initialised in method containing it [here](https://github.com/MetaMask/metamask-extension/pull/14047/files/146e7efdec2f18c24441fbe2d77d44b02ddcdbf9#diff-215d2cf6fb64ee62b9f484e7bbdd33246739e39b854e3cc8c2245b026352c12aR240).

app/scripts/background.js
adonesky1: > In MV3 it is better to have event listener at root level

Is there any reason not to have it at the root level in MV2 though? I'm wondering why we want to initialize without passing the `remotePort` in MV2?

adonesky1: So I tested, removing all the feature flags for MV3 
For example:
```
let initialState;
let initialLangCode;
const initApp = (remotePort) => initialize(remotePort).catch(log.error);

// if (process.env.ENABLE_MV3) {
browser.runtime.onConnect.addListener(initApp);
(async () => {
  initialState = await loadStateFromPersistence();
  initialLangCode = await getFirstPreferredLangCode();
})();
// } else {
//   initialize().catch(log.error);
// }
```
and everything seems to work quite well in MV2. I'm wondering whether we need to make these changes for only MV3 (is my point with the comments above)

adonesky1:
I added these changes on top of the changes in https://github.com/MetaMask/metamask-extension/pull/14181 and ran as an MV3 build and am still seeing the UI fail to load on the first load. It does load after a refresh.

app/scripts/background.js
jpuri: @adonesky1 : I would prefer to keep all of these changes behind flag. Then for MV3 we can have a flask release and have these features tested more throughly before enabling for everyone.

PR #14054, opened 2022-03-18
Dark Mode: Select Quote Popover

darkwing:
This is the popover at the bottom of a swaps quote where it says "5 sources ..."
qdlight
qddark

This also removes a weird jump effect because rows go from 0 border to 1px border, which results in a 2px "jump"


georgewrmarshall (approved):
lgtm

brad-decker (approved):

PR #14056, opened 2022-03-18
Dark Mode: Token Details

darkwing:
yayay
tokendetailsdark

### DIFF
--- a/ui/pages/token-details/index.scss
+++ b/ui/pages/token-details/index.scss
@@ -42,9 +42,4 @@
       margin-inline-start: 20px;
     }
   }
-
-  .btn--rounded.btn-primary {
-    background-color: var(--color-primary-inverse);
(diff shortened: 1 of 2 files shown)


ui/pages/token-details/index.scss
```diff
@@ -44,6 +44,7 @@
   }
 
   .btn--rounded.btn-primary {
-    background-color: #fff;
+    background-color: var(--color-primary-inverse);
+    color: var(--color-primary-default);
   }
```
georgewrmarshall, line 49: Why aren't we just using `<Button type="secondary" />` here?

georgewrmarshall (commented):
Can we use type="secondary" on the Button component here instead?

georgewrmarshall (approved):
lgtm

brad-decker (approved):

Issue #13719, opened 2022-02-22
Implement Dark Mode


darkwing:
### Describe the bug

The following are tasks required for implementing dark mode:

- [x] Import new colors library from @georgewrmarshall #13657
- [x] Buttons https://github.com/MetaMask/metamask-extension/pull/13651
- [x] Box #13725
- [x] Typography #13758
- [x] AccountList #13864 @georgewrmarshall 
- [x] Callout @georgewrmarshall https://github.com/MetaMask/metamask-extension/pull/13831
- [x] SnapsSettingCard @georgewrmarshall 
- [x] Card @georgewrmarshall https://github.com/MetaMask/metamask-extension/pull/13831
- [x] Chip @georgewrmarshall https://github.com/MetaMask/metamask-extension/pull/13831
- [ ] Icons @georgewrmarshall
    - [x] https://github.com/MetaMask/metamask-extension/pull/13861
- [x] Form fields @darkwing 
    - [x] Text inputs https://github.com/MetaMask/metamask-extension/pull/13821
    - [x] Password inputs https://github.com/MetaMask/metamask-extension/pull/13821
    - [x] Select dropdowns https://github.com/MetaMask/metamask-extension/pull/13875
    - [x] Textareas https://github.com/MetaMask/metamask-extension/pull/13821
    - [x] Settings toggles
- [x] Basic content background and border colors https://github.com/MetaMask/metamask-extension/pull/13764
- [x] Basic content text colors
    - [x] Headings
    - [x] Paragraphs
    - [x] Links
- [x] Dropdowns, modals, and popovers
    - [x] Networks dropdown
    - [x] Accounts dropdown
    - [x] Account details modal
    - [x] Connected account popover https://github.com/MetaMask/metamask-extension/pull/13932
- [x] Home activity feed and transaction list
- [x] Settings screen
- [x] Hardware wallet connectivity screens
- [ ] Tooltips
- [x] Confirmation screens
- [ ] ActionableMessages & warnings
- [x] Swaps screens

## Color list

- [x]   --Blue-300 @GuillaumeRx 
- [x]   --Blue-400 @GuillaumeRx 
- [x]   --Blue-500 / --primary-blue @GuillaumeRx 
- [x]   --Grey-000 @GuillaumeRx 
- [x]   --Grey-100 @GuillaumeRx 
- [x]   --Grey-200 @GuillaumeRx 
- [x]   --Grey-400 @GuillaumeRx 
[...]

PR #14074, opened 2022-03-21
Dark Mode : ConfirmEncryptionPublicKey 

GuillaumeRx:
## More information

* Fixes #13719

## Screenshots/Screencaps

Screenshot 2022-03-21 181820
Screenshot 2022-03-21 181834

## Manual testing steps

- Open latest Storybook build
- Look at ConfirmEncryptionPublicKey

### DIFF
--- a/ui/pages/confirm-encryption-public-key/confirm-encryption-public-key.scss
+++ b/ui/pages/confirm-encryption-public-key/confirm-encryption-public-key.scss
@@ -2,7 +2,7 @@
   &__container {
     width: 380px;
     border-radius: 8px;
-    background-color: var(--white);
+    background-color: var(--color-background-default);
     box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.08);
     display: flex;
     flex-flow: column nowrap;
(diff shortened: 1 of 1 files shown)


ui/pages/confirm-encryption-public-key/confirm-encryption-public-key.scss
```diff
@@ -205,8 +205,8 @@
       height: 48px;
 
       &--default {
-        background-color: #777a87;
-        color: white;
+        background-color: var(--color-background-alternative);
+        color: var(--color-text-muted); //@TODO: Not sure
```
darkwing, line 209: Thoughts @georgewrmarshall ?

georgewrmarshall: I think we are starting to see if an outlier exists and it doesn't fit into the system. Instead of trying to push the system to fit the component we can make the component adhere to the system. e.g the account and network menus. They were awkward to work with, with the dark colors inside light theme. Once we converted it to use light theme colors it was a dream to work with. Same goes for this instance. What does it look like if we just convert it to use the light. theme colors. Can we achieve the same UI goals(contrast, visibility, accessibility etc) if it's in light mode? Probably fine 😁 

```suggestion
        color: var(--color-text-alternative);
```

georgewrmarshall (commented):
Looks good! Provided a suggestion for the TODO

georgewrmarshall (approved):

darkwing (approved):

PR #14075, opened 2022-03-21
Dark Mode: Snackbar component

darkwing:
Use in onboarding:
Snackbar

### DIFF
--- a/ui/components/ui/snackbar/index.scss
+++ b/ui/components/ui/snackbar/index.scss
@@ -2,10 +2,10 @@
   @include H7;
 
   padding: 0.75rem 1rem;
-  color: var(--Blue-600);
+  color: var(--color-text-default);
   min-width: 360px;
   width: fit-content;
-  background: var(--Blue-000);
-  border: 1px solid var(--Blue-200);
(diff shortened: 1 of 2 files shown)


ui/components/ui/snackbar/index.scss
```diff
@@ -2,10 +2,10 @@
   @include H7;
 
   padding: 0.75rem 1rem;
-  color: var(--Blue-600);
+  color: var(--color-text-default);
```
GuillaumeRx, line 5:
```suggestion
  color: var(--color-primary-alternative);
```

If we want to stay with the original blue text

GuillaumeRx (changes requested):

ui/components/ui/snackbar/index.scss
georgewrmarshall: So for alerts we are using `--color-text-default` this is correct! Sorry I should have notified the DM team

georgewrmarshall (approved):
lgtm!

GuillaumeRx (approved):

PR #14097, opened 2022-03-22
Dark Mode: Fix Actionable Message Error Text

darkwing:
## Before:

error-before

## After:

error-after

### DIFF
--- a/ui/components/ui/actionable-message/index.scss
+++ b/ui/components/ui/actionable-message/index.scss
@@ -119,6 +119,7 @@
 
     .actionable-message__message {
       text-align: left;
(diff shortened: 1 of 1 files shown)


georgewrmarshall (commented):
I don't understand why the ActionableMessage type="danger" is not displaying like it is in storybook? I checked `ui/pages/import-token/index.scss` and there doesn't seem to be any css that effects the ActionableMessage component. It should use the `var(--color-error-muted)` as the background color and `var(--color-text-default)` as the text color.

Screen Shot 2022-03-21 at 9 22 47 PM

segun (approved):

PR #13861, opened 2022-03-07
Updating all caret icons for dark mode 🥕

georgewrmarshall:
Explanation: Updating all caret icons to use newly create caret icon components.
- Removed old svg images
- Created components for all caret icons up, down, left, right
- Created stories for any components that were updated for easy visual review

Manual testing steps:  
  - Go to latest build of storybook and go through the list of updated components using search bar
  - Look at only the caret icons
  - Check dark mode by clicking the 🌙  icon in the toolbar of storybook
  
List of updated components:
- `Icons`
- `AddNetwork`
- `ConfirmPageContainerHeader` (controls need to be like the below)
Screen Shot 2022-03-08 at 8 20 33 AM

- `NetworkDisplay`
- `TabBar` (needs to be mobile view)
Screen Shot 2022-03-08 at 8 23 50 AM

- `Dropdown`
- `SenderToRecipient` (needs to be `CARDS_VARIANT`)
Screen Shot 2022-03-08 at 8 25 09 AM

- `AddRecpient` NOT IN STORY LOOK AT SCREENSHOT BELOW with "Back to all" text

- `NetworksListItem`
- `SettingsSearchList` NO STORY
Screen Shot 2022-03-08 at 9 22 28 AM
Screen Shot 2022-03-08 at 9 22 37 AM

## Screenshots
Screen Shot 2022-03-06 at 6 55 58 PM
Screen Shot 2022-03-06 at 6 56 04 PM
Screen Shot 2022-03-06 at 9 36 07 AM
Screen Shot 2022-03-06 at 9 36 29 AM
Screen Shot 2022-03-06 at 9 37 01 AM
Screen Shot 2022-03-06 at 9 55 53 AM
Screen Shot 2022-03-06 at 9 57 00 AM
Screen Shot 2022-03-06 at 11 06 05 AM
Screen Shot 2022-03-06 at 11 06 40 AM
Screen Shot 2022-03-07 at 1 48 37 PM
Screen Shot 2022-03-07 at 1 48 40 PM
Screen Shot 2022-03-07 at 1 59 55 PM
Screen Shot 2022-03-07 at 1 59 59 PM
Screen Shot 2022-03-07 at 2 09 02 PM
Screen Shot 2022-03-07 at 2 09 06 PM
Screen Shot 2022-03-07 at 2 52 46 PM
Screen Shot 2022-03-07 at 2 53 01 PM

PR #14103, opened 2022-03-22
Fix Settings Header Back Button and update "Add Contacts" position

digiwand:
## Explanation

2 things related to the Settings Header:
1. Add missed caret 🥕 following `IconCaretLeft` addition related to Dark Mode. 
- Updated `IconCaretLeft` to use `onClick`
- Related to https://github.com/MetaMask/metamask-extension/pull/13861
2. Update "Add Contacts" button position on Settings > Contacts page

## More information

Fixes: V10.12.0
QA: Release Candidate Rolling Doc:  https://docs.google.com/document/d/1OvpkZbvRet33XQVplIc32zlm-6PgSasJcj1XFv9kh9s/edit#
Slack thread: https://consensys.slack.com/archives/GTQAGKY5V/p1647884761174739?thread_ts=1647856853.318179&cid=GTQAGKY5V

### Before
Screenshot 2022-03-21 at 10 51 04 PM
Screenshot 2022-03-21 at 10 52 34 PM

### After
(Search bar should have a white background here. This is
Screenshot 2022-03-21 at 10 30 32 PM
Screenshot 2022-03-21 at 10 31 40 PM

### DIFF
--- a/ui/pages/settings/contact-list-tab/index.scss
+++ b/ui/pages/settings/contact-list-tab/index.scss
@@ -227,12 +227,12 @@
 .address-book-add-button {
   &__button {
     position: absolute;
-    top: 85px;
+    top: 80px;
     right: 16px;
     width: auto;
 
     @media screen and (max-width: $break-small) {
-      top: 16px;
+      top: 8px;
       right: 60px;
 
       &--hidden {
--- a/ui/pages/settings/index.scss
+++ b/ui/pages/settings/index.scss
@@ -193,13 +193,6 @@
 
     @media screen and (max-width: $break-small) {
       display: block;
-      background-image: url('/images/caret-left-black.svg');
-      width: 18px;
-      height: 18px;
-      opacity: 0.5;
-      background-size: contain;
-      background-repeat: no-repeat;
-      background-position: center;
       margin-right: 16px;
       cursor: pointer;
 
--- a/ui/pages/settings/index.scss
+++ b/ui/pages/settings/index.scss
@@ -193,10 +193,11 @@
 
     @media screen and (max-width: $break-small) {
       display: block;
-      margin-right: 16px;
+      margin-right: 8px;
       cursor: pointer;
 
       [dir='rtl'] & {
+        margin: 0 0 0 8px;
         transform: rotate(180deg);
       }
     }
--- a/ui/components/ui/icon/icon-caret-left.js
+++ b/ui/components/ui/icon/icon-caret-left.js
@@ -6,6 +6,7 @@ const IconCaretLeft = ({
   color = 'currentColor',
   ariaLabel,
   className,
+  onClick,
 }) => (
   <svg
     width={size}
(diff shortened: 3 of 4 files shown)


ui/pages/settings/index.scss
```diff
@@ -193,19 +193,8 @@
...
-      width: 18px;
-      height: 18px;
-      opacity: 0.5;
-      background-size: contain;
-      background-repeat: no-repeat;
-      background-position: center;
       margin-right: 16px;
       cursor: pointer;
-
-      [dir='rtl'] & {
```
danjm, line 206: I think we still want this, or something like it, so that the caret appears correctly in languages that are read right to left

danjm (changes requested):
Only one suggestion, looks good otherwise

ui/pages/settings/contact-list-tab/index.scss
```diff
@@ -227,12 +227,12 @@
...
   &__button {
     position: absolute;
-    top: 85px;
+    top: 80px;
     right: 16px;
     width: auto;
 
     @media screen and (max-width: $break-small) {
-      top: 16px;
+      top: 9px;
```
digiwand, line 235: oops, thought I made this 8px which looked fine. Will update this to 8px

ui/pages/settings/index.scss
digiwand: oh 🤔  interesting. I didn't see how the direction was added here. Looking into this

digiwand: I see now. Thanks for the catch! 
Screenshot 2022-03-22 at 12 58 12 AM

georgewrmarshall: This is just for `[dir='rtl'] & {` right? The rest of the removed css is okay to remove!

georgewrmarshall (commented):
Ahh this is my bad! I also have a fix for this too. I actually prefer yours but either or is fine. Our icon library is going to be updated soon anyways. https://github.com/MetaMask/metamask-extension/pull/14113

ui/pages/settings/index.scss
digiwand: yep! Thanks for confirming @georgewrmarshall! 

There's 1 more thing that's a little off after these changes. The caret no longer lines up with the content. These caret buttons have a larger target area which is great for usability. Is it okay to leave it as is?

digiwand:
@georgewrmarshall oh! I was wondering about the icon sitting on top of the setting headers. I'm good with either as well.

ui/pages/settings/index.scss
georgewrmarshall: I think this is ok to leave like this!

georgewrmarshall:
Let's merge this I will close https://github.com/MetaMask/metamask-extension/pull/14113

georgewrmarshall (approved):
LGTM!

danjm (approved):

Issue #13972, opened 2022-03-15
MV3: fix injecting provider into DAPP


jpuri:
Connecting extension from DAPP fails with this error in MV3

![Screenshot 2022-03-14 at 11 53 59 AM](https://user-images.githubusercontent.com/2182307/158116656-e6a41c2b-86c3-4e75-bd75-739918552e19.png)

We need to find an appropriate solution for it and fix the issue. Solution may be:
1. appropriate change in manifest.json may fix the issue
2. we replace inline script injection with a file in src of script tag

jpuri:
Issue is fixed in PR https://github.com/MetaMask/metamask-extension/pull/14396

mib-hashflow:
It seems like the fix never got rolled out? Can you confirm this is fixed?

jpuri:
Yes this fix is there in the latest code.

PR #14108, opened 2022-03-22
Fix script tag injection in MV3

jpuri:
#13972

Fix script tag injection in MV3.

### DIFF
--- a/app/scripts/contentscript.js
+++ b/app/scripts/contentscript.js
@@ -42,7 +42,11 @@ function injectScript(content) {
     const container = document.head || document.documentElement;
     const scriptTag = document.createElement('script');
     scriptTag.setAttribute('async', 'false');
-    scriptTag.textContent = content;
+    if (process.env.ENABLE_MV3) {
+      scriptTag.setAttribute('src', browser.runtime.getURL('inpage.js'));
+    } else {
+      scriptTag.textContent = content;
+    }
     container.insertBefore(scriptTag, container.children[0]);
     container.removeChild(scriptTag);
   } catch (error) {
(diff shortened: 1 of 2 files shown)


app/scripts/background.js
```diff
@@ -480,8 +480,8 @@ function setupController(initState, initLangCode) {
     if (count) {
       label = String(count);
     }
-    browser.browserAction.setBadgeText({ text: label });
-    browser.browserAction.setBadgeBackgroundColor({ color: '#037DD6' });
+    // browser.browserAction.setBadgeText({ text: label });
+    // browser.browserAction.setBadgeBackgroundColor({ color: '#037DD6' });
```
jpuri, line 484: Ignore above as it is different mv3 issue we plan to address in separate task.

app/scripts/contentscript.js
```diff
@@ -37,12 +37,12 @@ if (shouldInjectProvider()) {
  *
  * @param {string} content - Code to be executed in the current document
  */
-function injectScript(content) {
+function injectScript() {
```
adonesky1, line 40: Should `injectScript(inpageBundle);` become `injectScript();` on line 31?

adonesky1: And remove `inpageSuffix` and `inpageBundle` consts?

jpuri: I am thinking of adding a feature flag for MV3 and using it here to conditionally inject script text or source.

adonesky1 (dismissed):

jpuri:
Additionally script injection using src attribute is async and thus use of events like `DOMContentLoaded` can fail

https://github.com/MetaMask/test-dapp/blob/main/src/index.js#L1274

Using `window.onload` will be better option.

PR #14111, opened 2022-03-22
Fixed lock screen showing briefly issue

VSaric:
## Explanation

This PR is part of #13443 ticket. After creating a password, it shows the login page briefly with a loading overlay before navigation to the success screen. This shouldn't happen. It should lead directly to the success screen.

## More information

We realized that this bug is in `useEffect` in `onboarding-flow.js`, because `isUnlocked` and `isInitialized` is part of dependency array of `useEffect`. 

The dependency array basically tells the hook to "only trigger when the dependency array changes". In the above example, it means "run the callback every time `isInitialized` and `isUnlocked` variable changes". And because of that it shows the login page briefly with a loading overlay before navigation to the success screen.

@adonesky1 Please review this PR and comment if you think this is okay or if you have better solution to fix this bug.  

### Before

https://user-images.githubusercontent.com/92527393/159492688-8c77d723-b0a0-472c-98c8-f1c3db8098d1.mov

https://user-images.githubusercontent.com/92527393/159492737-ac459756-97fb-45ab-80d8-4b5381da84b1.mov

### After

https://user-images.githubusercontent.com/92527393/159492860-c062f73f-f047-4ebc-bb93-45a2b2aa4fbf.mov

https://user-images.githubusercontent.com/92527393/159492965-f783fa28-ef61-4e43-87ef-9acdc19b2757.mov


adonesky1:
Good work. Seems right to me!

adonesky1 (approved):
LGTM!

mcmire (approved):
Makes sense to me too. Nice catch.

PR #14115, opened 2022-03-22
On send/edit page, disable app header so that networks can't be changed

segun:
## Explanation
Disabled app header in send.edit page, so that users can't change networks when editing a transaction.

Currently, I’m able to change the Network while I’m editing a transaction, leading to a workflow break, not being able to move forward again. [See video](https://drive.google.com/file/d/18yn5bnhxBvOy5IPd-ynLAmf9CZx5THFn/view?usp=sharing)

### Before
[See video](https://drive.google.com/file/d/18yn5bnhxBvOy5IPd-ynLAmf9CZx5THFn/view?usp=sharing)

### After

https://user-images.githubusercontent.com/44811/159545345-a03ea016-bf90-4c49-9a5d-0a62ed30a2ca.mov

## Manual testing steps

1. Go to Send
2. Introduce an amount of ETH
3. Click on change network
4. Click on Edit Transaction
5. Click on change network
6. Change the network to a different one
7. Click Next
8. Change network back the the original one

Actual:
After “v” and “vi” network is changed
After “Next”, console log error appears “Uncaught (in promise) TypeError: t is undefined”
After “viii” I cannot continue to the next screen, for confirming the Transaction

Expected:
I should not be able to change network in the middle of the “Edit Transaction” screen

### DIFF
--- a/ui/pages/routes/routes.component.js
+++ b/ui/pages/routes/routes.component.js
@@ -242,6 +242,16 @@ export default class Routes extends Component {
     );
   }
 
+  onEditTransactionPage() {
+    const { location } = this.props;
+    return Boolean(
+      matchPath(location.pathname, {
+        path: SEND_ROUTE,
+        exact: false,
+      }),
+    );
+  }
+
   onSwapsPage() {
     const { location } = this.props;
     return Boolean(
@@ -359,6 +369,7 @@ export default class Routes extends Component {
             onClick={this.onAppHeaderClick}
             disabled={
               this.onConfirmPage() ||
+              this.onEditTransactionPage() ||
               (this.onSwapsPage() && !this.onSwapsBuildQuotePage())
             }
           />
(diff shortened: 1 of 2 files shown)


ui/pages/routes/routes.component.js
```diff
@@ -242,6 +242,16 @@ export default class Routes extends Component {
     );
   }
 
+  onEditTransactionPage() {
```
danjm, line 245: We only want to disable this when a transaction is being edited, not when it is first being created

danjm: ah, you can use `getSendStage(state) === SEND_STAGES.EDIT` to determine this

danjm (dismissed):
LGTM

adonesky1 (dismissed):
LGTM!

danjm (approved):

adonesky1 (approved):

PR #14126, opened 2022-03-22
Added getAccounts suppressUnauthorized param

### LINKED ISSUE #13467: Incorrectly rejecting valid personal_sign requests
When trying to call `personal_sign` under some easy to reproduce but normal-seeming conditions, we are throwing the error `Invalid parameters: must provide an Ethereum address.`.

## Repro steps:

First try invoking personal sign from a site that you are not connected to:
```
await window.ethereum.request({
  "method": "personal_sign",
    "params": [
        "0x666f6f",
        "0x30746D407B71C1CBc159dC6355D1B949F2F3e2Cf",
    ],
})
```
It will throw an error like:
```
{
  "jsonrpc": "2.0",
  "id": 0,
  "error": {
    "code": 32329,
    "message": "Invalid parameters: must provide an Ethereum address."
  }
}
```
Sometimes the error is even thrown when you *are* connected to the site, or connected to it manually.

## Investigation

This error comes from [here](https://github.com/MetaMask/eth-json-rpc-middleware/blob/eb649a8d3fad24a63fe51b7f5d10cb1d807601f5/src/wallet.ts#L393), where for some reason the requesting account is being checked against an array of addresses to validate "whether it is a valid address" or not.

Basically this is probably just the wrong place to be enforcing this validation.

- If we want to ensure the method is restricted, and will only be signed by a connected user, then ideally we're simply making it a `restrictedMethod` in [our permissions controller](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/controllers/permissions/specifications.js).
- If we just want to check for address-ness, then we should check for address format.

We found that if we requested permissions first then it would work

```
      const perms = await ethereum.request({
        method: 'wallet_requestPermissions',
        params: [{eth_accounts: {}}],
      });
```
so this seems like it may be caused by the permissions controller interfering with [the getAccounts request](https://github.com/MetaMask/eth-json-rpc-middleware/blob/eb649a8d3fad24a63fe51b7f5d10cb1d807601f5/src/wallet.ts#L382).

What's really weird is manually setting the connection does not seem to allow this, it seems to have to be initiated in dapp.

shanejonas:
## Explanation

This adds a param to be able to have `getAccounts` throw an error, in cases like `eth-json-rpc-middleware` where it truely is a permission error. 

This fixes #13467 and other methods like `signTypedData` from returning misleading error `Invalid Address`

Relies on: https://github.com/MetaMask/eth-json-rpc-middleware/pull/116


app/scripts/metamask-controller.js
```diff
@@ -2456,18 +2456,21 @@ export default class MetamaskController extends EventEmitter {
    * array if no accounts are permitted.
    *
    * @param {string} origin - The origin whose exposed accounts to retrieve.
+   * @param {boolean} suppressUnauthorized - suppresses the unauthorized error. defaults to `true`.
```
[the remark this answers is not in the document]
digiwand: Optional 2: 
What if we renamed `suppressUnauthorized` to `suppressUnauthorizedErrorCode`?

I'm proposing this because one of my first thoughts upon reading`suppressUnauthorized` is that the return value would not include unauthorized accounts (if this could be a thing). However, I have observed that `suppressUnauthorized` is specific to the error logic.

digiwand (commented):
Optional comments added. No blockers ✨

naugtur:
Yarn audit points out a dependency on a malicious package. Any chance that dependency could be dropped? Maybe the direct dep that pulls it could be updated?

app/scripts/metamask-controller.js
```diff
@@ -2456,18 +2456,21 @@ export default class MetamaskController extends EventEmitter {
...
       return await this.permissionController.executeRestrictedMethod(
         origin,
         RestrictedMethods.eth_accounts,
       );
     } catch (error) {
-      if (error.code === rpcErrorCodes.provider.unauthorized) {
-        return [];
+      if (suppressUnauthorized) {
+        if (error.code === rpcErrorCodes.provider.unauthorized) {
+          return [];
```
Gtonizuka, line 2472: should we throw a different error in case `error.code === rpcErrorCodes.provider.unauthorized` is not true?

app/scripts/metamask-controller.js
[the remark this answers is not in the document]
Gtonizuka: I agree with Option 2 above ^ I am not a  big fan of passing bool as params but I guess in this case its the less painful path

shanejonas:
> Yarn audit points out a dependency on a malicious package. Any chance that dependency could be dropped? Maybe the direct dep that pulls it could be updated?

I didn't change any deps so this is an issue on develop.

app/scripts/metamask-controller.js
shanejonas: it needs to keep the original behaviour to not break other uses of `getAccounts` like when you call `eth_accounts` without permissions it needs to return `[]` for legacy reasons.

app/scripts/metamask-controller.js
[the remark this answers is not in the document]
shanejonas: renamed to `suppressUnauthorizedError`

Gtonizuka (dismissed):

app/scripts/metamask-controller.js
```diff
@@ -1187,12 +1187,15 @@ export default class MetamaskController extends EventEmitter {
       },
       version,
       // account mgmt
-      getAccounts: async ({ origin }) => {
+      getAccounts: async ({ origin }, suppressUnauthorizedError = true) => {
```
mcmire, line 1190: Since this method already takes an object, should `suppressUnauthorizedError` be put inside the object? That way if other arguments get added in the future we do not have to worry about supplying a value for `suppressUnauthorizedError`, etc.

app/scripts/metamask-controller.js
```diff
@@ -2456,18 +2459,21 @@ export default class MetamaskController extends EventEmitter {
...
     try {
       return await this.permissionController.executeRestrictedMethod(
         origin,
         RestrictedMethods.eth_accounts,
       );
     } catch (error) {
-      if (error.code === rpcErrorCodes.provider.unauthorized) {
-        return [];
+      if (suppressUnauthorizedError) {
+        if (error.code === rpcErrorCodes.provider.unauthorized) {
```
mcmire, line 2474: These can be combined, no? `if suppressUnauthorizedError && error.code === rpcErrorCodes.provider.unauthorized`.

app/scripts/metamask-controller.js
mcmire, line 1190: Should `suppressUnauthorizedError` be wrapped in an object so that when calling `getAccounts` it's easy to see what the boolean refers to?

shanejonas:
the object destructuring doesnt really work with the default values without adding a bunch more code to check existence: 

image

mcmire (dismissed):
LGTM!

mcmire (dismissed):
Looks good when compared to https://github.com/MetaMask/eth-json-rpc-middleware/commit/86915dc1a262278c45b1806c646b097aad458256.

mcmire (approved):
LGTM!

Gtonizuka (approved):

PR #14149, opened 2022-03-23
E2e dapp interactions

seaona:
## Explanation
* This PR introduces a new testcase file `Dapp Interactions` to test Metamask notifications / connection disconnection of Dapps etc.
* In this specs file we test 2 scenarios:

**Scenario1**:
1. connect to metamask from a dapp
2. lock metamask
3. connect to metamask from another dapp

**Scenario 2:**

1. from a dapp, take an action requiring a confirmation while metamask is locked

## How
- `dappOptions` has been introduced as a fixture. Here we can indicate the number of dapps we use.
- If we don't pass any `dappOptions`, pre default value of dapp servers is 1
- Dapp servers will be set up and terminated, according to the number of servers we indicate in `dappOptions: {numberOfDapps}`


test/e2e/helpers.js
```diff
@@ -13,7 +13,7 @@ const tinyDelayMs = 200;
 const regularDelayMs = tinyDelayMs * 2;
 const largeDelayMs = regularDelayMs * 2;
 const veryLargeDelayMs = largeDelayMs * 2;
-const dappPort = 8080;
+const dappBasePort = 808;
```
danjm, line 16: I think it could be a little confusing to see the number `808` as a base port, because that is later coerced to a string and concatenated with another string. And this causes the the default and most common case (8080) to never be explicitly written in the code. Maybe setting it to the number `8080` and then using arithmetic to set the port below (e.g. dappServer[i].listen(`${dappBasePort + i`);`

test/e2e/tests/dapp-interactions.spec.js
```diff
@@ -0,0 +1,125 @@
...
+          windowHandles,
+        );
+        await driver.fill('#password', 'correct horse battery staple');
+        await driver.press('#password', driver.Key.ENTER);
+        const notification = await driver.findElement({
+          text: 'Allow this site to add a network?',
+          tag: 'h3',
+        });
+
+        assert.ok(notification, 'Dapp action does not appear in Metamask');
```
danjm, line 58: Should this be `'Dapp action _does_ appear in Metamask'`? This should verify that after unlocking, the user does actually see the add ethereum chain notification page.

test/e2e/tests/dapp-interactions.spec.js
```diff
@@ -0,0 +1,125 @@
...
+        const connectedDapp1 = await driver.findElement({
+          text: 'http://127.0.0.1:8080',
+          tag: 'span',
+        });
+        const connectedDapp2 = await driver.findElement({
+          text: 'http://127.0.0.1:8081',
+          tag: 'span',
+        });
+
+        assert.ok(connectedDapp1, 'Account not connected to Dapp1');
```
danjm, line 120: Same comment as above, should the `'not'` be removed from this assertion message?

danjm (changes requested):
Looks good overall. A few small comments

test/e2e/tests/dapp-interactions.spec.js
```diff
@@ -0,0 +1,125 @@
...
+  const ganacheOptions = {
+    accounts: [
+      {
+        secretKey:
+          '0x7C9529A67102755B7E6102D6D950AC5D5863C98713805CEC576B945B15B71EAC',
+        balance: convertToHexValue(25000000000000000000),
+      },
+    ],
+  };
+  it('should a notification be triggered despite Metamask being locked', async function () {
```
PeterYinusa, line 21: Nit: can we rename the test so it flows when you read the `it` along with the test name e.g. ... It `should trigger the add chain confirmation despite MetaMask being locked`

test/e2e/tests/dapp-interactions.spec.js
```diff
@@ -0,0 +1,125 @@
...
+          text: 'Allow this site to add a network?',
+          tag: 'h3',
+        });
+
+        assert.ok(notification, 'Dapp action does not appear in Metamask');
+      },
+    );
+  });
+
+  it('should a second dapp be able to connect to Metamask after being locked', async function () {
```
PeterYinusa, line 63: same as above

test/e2e/helpers.js
seaona: sure, I've made a few tweaks, so we are now utilizing the same pattern (using `dappOptions`, like `ganacheOptions`) @PeterYinusa

test/e2e/tests/dapp-interactions.spec.js
seaona: @PeterYinusa fixed

test/e2e/tests/dapp-interactions.spec.js
seaona: same as above

test/e2e/helpers.js
PeterYinusa: Oops, I hadn't realised you had already made the arithmetic change. I was suggesting using the same pattern as ganacheOptions too explicitly specify the port of the second instance as an alternative to the loop and port calculations. Sorry for unnecessary extra work proposed

seaona: ohh no problem! I guess the new changes also simplify a bit the approach :) thanks!

test/e2e/tests/dapp-interactions.spec.js
danjm: ah, okay, that makes sense. Thanks for explaining

danjm (approved):
LGTM

PeterYinusa (approved):

Issue #13926, opened 2022-03-14
[Bug]: wacsdc


jpuri:
### Describe the bug

----cwdec

### Steps to reproduce

1.
csdcsdc

### Version

Latest

### Browser

Chrome

### Operating system

Windows, MacOS, Linux

PR #14166, opened 2022-03-23
fix browser action calls for mv3

adonesky1:
Fixes: https://github.com/MetaMask/metamask-extension/issues/13926

This adds back the ability to click on the extension icon to open app that is not currently functional on the `mv3` branch. It also adds back the color and "pending transaction" number to the icon.


Gtonizuka (approved):

adonesky1:
Addressed elsewhere

Issue #13559, opened 2022-02-08
Add preferences state listener to MetaMetrics Controller


brad-decker:
### Requirements
1. The MetaMetrics controller must be modified to use the existing preferences state listener, or add a new one, that will call the method created in #14048 with the new preferences state.

Issue #13557, opened 2022-02-08
Identify user traits on construction of MetaMetrics controller


brad-decker:
### Requirements
1. Modify the MetaMetrics controller constructor so that it calls the method created in #14048 with the preferences state object (preferencesStore.getState()).
2. Add test coverage to ensure the method in #14048 is called upon construction.

Issue #13534, opened 2022-02-07
identify user trait ledger_connection_type


brad-decker:
### Requirements
1. Modify the preferences state update listener created in #13559 such that when the ledger live connnection type is changed we identify the `ledger_connection_type` trait.

Issue #13533, opened 2022-02-07
Identify user trait 3box_enabled


brad-decker:
### Requirements
1. Modify the preferences state update listener created in #13559 such that when the 3box feature is toggled we identify the `3box_enabled trait.

PR #14192, opened 2022-03-24
Build user traits object when metamask state changes

### LINKED ISSUE #14048: Create method for deriving segment user traits from preferences state

brad-decker:
## Explanation

* Adds a listener to MetaMask state tree changes and calls into the MetaMetrics controller. Originally I had intended this to be only the preferences controller but @danjm suggested all of MetaMask state. I agreed, but said for MVP just preferences store. However, three box state is stored on three box controller. This necessitated changing the strategy. Instead of hooking into preferences store we let the MetaMask controller tell the metametrics controller that state changed.
* This method handles caching the previous user traits, and compares them to the new one so that we only call .identify with new traits

## More information
* Fixes #14048, #13532, #13559, #13557, #13534, #13531, #13533

### Remaining items:
- [x] Tests

### DIFF
--- a/app/scripts/controllers/metametrics.js
+++ b/app/scripts/controllers/metametrics.js
@@ -1,4 +1,4 @@
-import { merge, omit, omitBy } from 'lodash';
+import { isEqual, merge, omit, omitBy, pickBy } from 'lodash';
 import { ObservableStore } from '@metamask/obs-store';
 import { bufferToHex, keccak } from 'ethereumjs-util';
 import { generateUUID } from 'pubnub';
@@ -6,6 +6,7 @@ import { ENVIRONMENT_TYPE_BACKGROUND } from '../../../shared/constants/app';
 import {
   METAMETRICS_ANONYMOUS_ID,
   METAMETRICS_BACKGROUND_PAGE_OBJECT,
+  TRAITS,
 } from '../../../shared/constants/metametrics';
 import { SECOND } from '../../../shared/constants/time';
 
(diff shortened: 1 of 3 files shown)


app/scripts/controllers/metametrics.js
```diff
@@ -303,6 +304,142 @@ export default class MetaMetricsController {
...
+        }. Category was: ${payload.category}. Payload keys were: ${Object.keys(
+          payload,
+        )}. ${
+          typeof payload.properties === 'object'
+            ? `Payload property keys were: ${Object.keys(payload.properties)}`
+            : ''
+        }`,
+      );
+    }
+  }
```
brad-decker, line 432: These were just moved above the private method line in the code, no changes

segun (approved):
LGTM

app/scripts/controllers/metametrics.js
```diff
@@ -370,6 +507,34 @@ export default class MetaMetricsController {
...
+      [TRAITS.THREE_BOX_ENABLED]: metamaskState.threeBoxSyncingAllowed,
+    };
+
+    if (!this.previousTraits) {
+      this.previousTraits = currentTraits;
+      return currentTraits;
+    }
+
+    if (this.previousTraits && !isEqual(this.previousTraits, currentTraits)) {
+      const updates = pickBy(
```
digiwand, line 527: super nit & optional: what if we renamed this `updatedTraits`?

digiwand (approved):
Nice 👍  

I like how we have a section for our "private" methods now and these are in alphabetical order

ryanml (approved):
LGTM! nice work

PR #14195, opened 2022-03-24
MetaMetrics: add segment.identify

### LINKED ISSUE #13529: Implement the identify method in MetaMetrics controller
### Helpful Links
[Segment Identify Spec](https://segment.com/docs/connections/spec/identify/)
[Mixpanel  profile property data types](https://help.mixpanel.com/hc/en-us/articles/115004547063-Properties-Supported-Data-Types)
[Segment node analytics identify method](https://segment.com/docs/connections/sources/catalog/libraries/server/node/#identify)

### Requirements
1. identify method should be added to MetaMetrics controller
2. identify method should accept an object of traits that will be passed into the .identify method of the segment api object. Note that traits can be strings, numbers, booleans or dates, or an array of any of those values. The array data type is what mixpanel calls a list. Some type checking should be done, specifically traits cannot be nested objects except for some special ones.
3. segment mock should be updated to include identify method
4. unit tests for the controller must be updated to test that identify is called appropriately
5. userId must be the user's metametrics id. (do not supply anonymousId, we always pass userId as the field except when tracking truly anonymous events.)
6. If the user has opted out of metametrics, the identify call should do nothing and never hit segment.
7. Ensure that identify cannot be called with the 0x000 burn address as 'anonymousId' or 'userId' because that will lead to records being created and tied together for many users.

digiwand:
## Explanation

Fixes https://github.com/MetaMask/metamask-extension/issues/13529

## More information

Support `segment.identify`, analytics-node method in our MetaMetrics controller

## Manual testing steps

1. Ensure Settings > Security & Privacy > "Participate in MetaMetrics" is turned on
2. Update one of the four settings:
  - metamaskState.ledgerTransportType,
  - number of accounts
  - network chains
  - metamaskState.threeBoxSyncingAllowed
3. Go to MixPanel and ensure related user has the updated settings


app/scripts/controllers/metametrics.js
```diff
@@ -303,6 +303,122 @@ export default class MetaMetricsController {
...
+
+  /**
+   * Returns true if the value is an accepted date type
+   *
+   * @param {*} value
+   * @returns {boolean}
+   */
+  _isValidTraitDate = (value) => {
+    return Object.prototype.toString.call(value) === '[object Date]';
+  };
```
brad-decker, line 401: If we allow `Date` objects as input to the .identify method then we'll need to convert them to strings before sending to segment.

ui/store/actions.js
```diff
@@ -3190,6 +3190,14 @@ export async function closeNotificationPopup() {
...
  */
 
+/**
+ * @param {Object} traits
+ * @returns {Promise<void>}
+ */
+export function identifyMetaMetricsUser(traits) {
+  return promisifiedBackground.identifyMetaMetricsUser(traits);
+}
+
```
brad-decker, line 3200: For MVP I think we can skip making this method available to the UI

brad-decker (changes requested):

brad-decker (changes requested):
LGTM! Only thing to do is uncomment the call to .identify in handleMetaMaskStateUpdate

brad-decker (dismissed):
Yay LGTM

brad-decker (approved):

georgewrmarshall (approved):

georgewrmarshall (commented):
Apologies only saw the spelling fix. As I don't have context going to remove myself as a reviewer. Github was doing something weird 


https://user-images.githubusercontent.com/8112138/161090023-df693f76-67cf-4bbc-a596-196c91591c20.mov

brad-decker:
@georgewrmarshall
EyxhFqCvQhVrW

PR #14147, opened 2022-03-23
Transaction updates methods return whole tx meta

danjm:
This PR was made in response to QA on v10.12.0 which found that swaps is not working on that branch and on develop.

The reason is that our newly introduced transaction update methods did not return the fully updated transaction, only the data used for update. This PR corrects that across all such update methods

PR #14196, opened 2022-03-24
TransactionsControllerTest: catch uncaught errors

digiwand:
## Explanation

Slight update to test to ensure the error is thrown, if not, the test should fail 

## More information

Related to: https://github.com/MetaMask/metamask-extension/pull/14147


brad-decker (dismissed):

brad-decker (dismissed):

app/scripts/controllers/transactions/index.test.js
```diff
@@ -2200,14 +2200,17 @@ describe('Transaction Controller', function () {
...
-          'Cannot call updateTransactionGasFees on a transaction that is not in an unapproved state',
-        );
-      }
+      assert.throws(
+        () =>
+          txController.updateTransactionGasFees('4', {
+            maxFeePerGas: '0x0088',
+          }),
+        Error,
+        'Cannot call updateTransactionGasFees on a transaction that is not in an unapproved state',
```
adonesky1, line 2209: [nit] could we make this error `Cannot call updateTransactionGasFees on a transaction that is in an approved state` 
(not a double negative)

adonesky1 (dismissed):
small nit but otherwise LGTM!

brad-decker (dismissed):

app/scripts/controllers/transactions/index.test.js
digiwand: @adonesky1 I agree these messages were confusing. Thanks for calling this out!

Currently, the transactions can have these statuses: 
``` 
  UNAPPROVED: 'unapproved',
  APPROVED: 'approved',
  REJECTED: 'rejected',
  SIGNED: 'signed',
  SUBMITTED: 'submitted',
  FAILED: 'failed',
  DROPPED: 'dropped',
  CONFIRMED: 'confirmed',
  PENDING: 'pending',
```

The calls should be blocked for other transactions that are not `approved`, like `dropped`. I updated the related function name and the error messages to hopefully add some clarity.

adonesky1:
@digiwand looks like you need a rebase 😅

app/scripts/controllers/transactions/index.test.js
```diff
@@ -2186,10 +2186,10 @@ describe('Transaction Controller', function () {
       assert.equal(result.userFeeLevel, 'high');
     });
 
-    it('throws error if status is not unapproved', function () {
+    it('should not update and should throw error if status is not unapproved', function () {
```
adonesky1, line 2189: again just a tiny nit for clarity (might just be for me haha)
```suggestion
    it('should not update and should throw error if status is not type "unapproved"', function () {
```

digiwand: Works for me! Thanks and updated

adonesky1 (approved):
LGTM!

DELJUJAZI (approved):

brad-decker (approved):

PR #14205, opened 2022-03-25
Update pasteIntoField method

PeterYinusa:
## Explanation
Update pasteIntoField method so it works on Mac OS also

### Before

before

### After

after

## Manual testing steps

yarn setup
yarn build:test
yarn test:e2e:single test/e2e/tests/add-account.spec.js --browser=chrome


digiwand (approved):
💻

brad-decker (approved):

PR #14206, opened 2022-03-25
Add 'Enhanced Gas UI' setting to search index

### LINKED ISSUE #13998: Settings search for EIP-1559 V2 - Enhanced Gas UI is broken
### Describe the bug

It is not possible to search setting option to enable enhanced GAS UI.

### Version

10.10.0

ryanml:
Fixes: https://github.com/MetaMask/metamask-extension/issues/13998

Screen Shot 2022-03-25 at 4 46 38 AM

### DIFF
--- a/ui/helpers/utils/settings-search.js
+++ b/ui/helpers/utils/settings-search.js
@@ -326,6 +326,13 @@ export function getSettingsRoutes(t) {
       route: `${ABOUT_US_ROUTE}#contactus`,
       icon: 'fa fa-info-circle',
     },
+    {
+      tab: t('experimental'),
+      section: t('enableEIP1559V2'),
+      description: t('enableEIP1559V2Description'),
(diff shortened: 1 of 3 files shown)


ui/helpers/utils/settings-search.js
```diff
@@ -350,6 +350,13 @@ export function getSettingsRoutes(t) {
         route: `${EXPERIMENTAL_ROUTE}#token-description`,
         icon: 'fa fa-flask',
       },
+      {
```
danjm, line 353: I think this should just be added as part of the initial definition of the settingRouteList array. This if else block should probably just be used for these two token detection related settings.

ryanml: Good catch, fixed

danjm (approved):

darkwing (approved):

PR #14212, opened 2022-03-25
Dark Mode: Set to default based on user preferred scheme

darkwing:
This code sets the default theme to dark if the user's system signals it prefers dark.

## Manual testing
1. Set your OS system preference to dark mode
2. Install MM
3. See dark onboarding
4. Set your OS system preference to light mode
5. Install MM
6. See light onboarding


NiranjanaBinoy (approved):
Changes look good to be, but the branch needs to be rebased.

digiwand (commented):
I think we need to update this branch. It appears to load as version 10.11.4

I created a new MetaMask account with dark mode turn on on my macOS. Dark-mode appears to be set as expected.

PR #14216, opened 2022-03-26
Update DetectTokenController for token detection v2

### LINKED ISSUE #13940: Use AssetContractController for balance check for ERC20 tokens.
Integrating AssetContractController in DetectTokenController for balance check for ERC20 Token will help reduce the redundancy in work when we add more custom network support in the future.

### LINKED ISSUE #13963: Update DetectTokenController for token detection v2
The token detection functionality has to be updated to use the new `detectToken` from TokensController, so that we can hold the detected tokens in an intermittent state before the user decides to add or ignore the token.

NiranjanaBinoy:
Fixed #13940, #13963 

- The DetectTokensController is updated to add detected tokens to a new intermittent state to hold the tokens before the user decides to add or ignore them using the method 
 `TokensController.addDetectedTokens()`.

- Enabling token detection for all supported networks (Polygon, Mainnet, BSC and Avalanche)

-   Added two new methods for adding and ignoring tokens in batches based on user choices.

- Updating the initialization of AssetContractController based on the new constructor definition 

- The single call balance check in DetectTokensController is replaced with the `getBalancesInSingleCall` in AssetContractController.

The controller changes for token detection v2 is available in `feature/3157-token-detection-v2-controllers` branch of @metamask/controllers, and reviewers might want to link or pack this repo to our extension repo for a full review.


NiranjanaBinoy:
I made the mistake of merging the child PR to this branch once the child PR is approved. So this PR is currently loaded with some unrelated changes, anyone who is reviewing can skip the `ui` folder and the `message.json` file as they are already reviewed and approved as part of child PR (anything related to this commit https://github.com/MetaMask/metamask-extension/pull/14216/commits/4f49443ba030825fd0cd8b7509466fde9927ebde)

brad-decker (dismissed):

tmashuang (dismissed):

app/scripts/controllers/detect-tokens.js
```diff
@@ -84,14 +101,23 @@ export default class DetectTokensController {
...
+      return;
+    }
     const { tokenList } = this._tokenList.state;
     // since the token detection is currently enabled only on Mainnet
     // we can use the chainId check to ensure token detection is not triggered for any other network
     // but once the balance check contract for other networks are deploayed and ready to use, we need to update this check.
     if (
-      this._network.store.getState().provider.chainId !== MAINNET_CHAIN_ID ||
-      Object.keys(tokenList).length === 0
+      !process.env.TOKEN_DETECTION_V2 &&
```
danjm, line 118: What about the case of `process.env.TOKEN_DETECTION_V2 && Object.keys(tokenList).length === 0` do we need to be returning in that case?

danjm (changes requested):
I left one question. Other than that, everything looks good

app/scripts/controllers/detect-tokens.js
NiranjanaBinoy: This was the existing condition for the previous version, token detection was activated for Mainnet only but the tokenList will be available for other networks as well. This will be removed during flag clean-up.

danjm: Is it possible for `Object.keys(tokenList).length === 0)` to be true if `process.env.TOKEN_DETECTION_V2` is true?

tmashuang (approved):

brad-decker (approved):

danjm (approved):
okay, LGTM

PR #14218, opened 2022-03-27
Set default getFetchWithTimeout default timeout value

darkwing:
We're passing the same value multiple times, so we may as well set a default value.


digiwand (approved):
LGTM!

brad-decker (approved):
LGTM!

PR #14232, opened 2022-03-28
updating prettier and eslint

brad-decker:
## Explanation

Prettier has made some improvements to the way that it handles line breaks and object/array bracket placement that I feel will be desirable for most. To make this upgrade I had to upgrade the prettier plugin for eslint, which resulted in eslint needing to be upgraded as well. I have added rule updates to prevent *most* changes unrelated to the prettier upgrade. There are a few notable exceptions:

1. no-unsafe-optional-chaining. In two cases we were destructuring the result of an optional chain. This could result in a type error because you cannot destructure undefined. To fix I used `?? {}` so that the destructuring happens on the empty object instead of undefined.
2. jsdocs/check-types was upgraded to check for casing of 'object'. I ignored 99% of these casing problems using the rules, but some instances were too specific to override. In these cases I made sure that there were no other lint issues of different types in the file and auto fixed those files. Result is things like : `Object<string, Object>` -> `Object<string, object>`
3. no-precision-loss, in some test cases we were using hex numbers as object keys, which results in precision loss at runtime. These were converted to strings. 

Example cases of why I feel this update is beneficial:
image
image
image


.eslintrc.base.js
```diff
@@ -63,5 +63,21 @@ module.exports = {
     // a browser context. For instance, we may import polyfills which change
     // global variables, or we may import stylesheets.
     'import/no-unassigned-import': 'off',
+
+    // import/no-named-as-default-member checks if default imports also have
+    // named exports matching properties used on the default import. Example:
+    // in confirm-seed-phrase-component.test.js we import sinon from 'sinon'
+    // and later access sinon.spy. spy is also exported from sinon directly and
+    // thus triggers the error. Turning this rule off to prevent churn when
```
mcmire, line 71: Can you go into more detail about what you mean about reducing churn when upgrading eslint? Do you mean that your editor slows to a crawl or something after eslint is upgraded? I guess I'm wondering if this is a temporary performance problem or a consistent one.

brad-decker: Not performance - churn in code review. IE trying to minimize the amount of changes in one PR and focusing them on just prettier related changes for reviewing purposes

mcmire: Ohh gotcha, so this should be turned on, but we're disabling it for now since it's going to cause a lot more violations, is that what you mean?

brad-decker: Exactly

brad-decker:
Closing in favor of #15293

PR #14233, opened 2022-03-28
Modify CSP headers to allow hash of injected script

### LINKED ISSUE #31: Allow working with strict CSPs
TIL: Our current mobile script injection technique violates some CSPs!

I just got off a call with a team that currently can't use MetaMask because their page's CSP refuses to interact with an inline script.

It would help them if we also allowed connecting via a non-injected provider. Maybe we should set up detect-provider to bring its own inpage-provider, so it allows side-stepping this issue.

I'm reaching out to them to see what CSP this might be, I think a safely strict testing csp might be `script-src: 'none'`.

### LINKED ISSUE #3133: Inpage injection fails in Firefox under some CSP settings
Hey, as far as I can tell, my content security policy is preventing MetaMask from injecting its scripts. This is only happening in Firefox. It works correctly in Chrome, Safari, Opera, and Brave. My `script-src` directive looks like this:
```
script-src 'self';
```
and I'm seeing this csp violation in the js console:
```
Content Security Policy: The page’s settings blocked the loading of a resource at self. Source: (function e(t,n,r){function s(o,u){if(!n ....
```

Unfortunately Firefox only shows a preview of the blocked script. I've tried sha256 hashing each of the scripts in the latest Metamask release and adding them to the CSP, but that did not work. If you have any ideas that would be great!

<b>Browser:</b> Firefox 58.0.1
<b>Operating System:</b> Mac OSX 10.13.2

Pandapip1:
Fixes #3133
Fixes https://github.com/MetaMask/detect-provider/issues/31


Pandapip1:

Pandapip1:
This needs testing. It's possible it will work first-time, but I find that
unlikely.

Ref: https://github.com/metamask/detect-provider/issues/31

Pandapip1:
Hold on CI is stuck. Rerunning...

darkwing:
@Pandapip1 Could you provide context for this change?  Otherwise this PR isn't something we can accept.

mrnerdhair:
The context here is https://github.com/MetaMask/detect-provider/issues/31.

This is very interesting code and a novel approach, and I appreciate it for that -- I can see myself referring back to this work on occasion.

However, I'm not yet convinced this approach can work in all cases. CSPs can be -- and in many notable usecases, are -- set by meta tags in addition to headers. In fact, CSP meta tags can be constructed, injected, and altered *after page load* by scripts, which is a technique used by all-in-one cookie consent toolkits -- if you turn off consent for a certain site, they'll just update the CSP to block it.

Preemptively adding a CSP when none exists will probably break pages that don't use one, and reliably detecting meta tag CSP's doesn't seem possible. (I'd also want to verify where this filter is applied during the ServiceWorker lifecycle; if the modifications are visible to the ServiceWorker itself, that could cause problems too.) The best solution to this issue that I can see at the moment would be to update MetaMask/detect-provider to use its own inpage-provider if it doesn't see one injected by the extension.

Pandapip1:
I'll add code to fix that. I think the approach may still work.

Pandapip1:
Well, it doesn't really matter where the meta tag is applied during the
service worker if the content script itself modifies the CSP meta tag.

I'll see why some of those chrome tests aren't working, and submit a fix
for them.

Pandapip1:
I'd also like to add that if there is no CSP meta or header present, it
isn't modified, so pages without a CSP will still work properly.

digiwand:
Hello @Pandapip1! I will need to allocate more time to review this thoroughly.`MetaMask/detect-provider` is outside my current domain and is related to this change. In the meantime, I've bumped this PR in our Slack channel for others to review.

Pandapip1:
Do not merge this yet. I've tried out the latest build, but it doesn't seem to correctly be modifying the headers. I'll get back once it seems to be working again.

Pandapip1:
Okay, so it correctly injects with the meta tags:
image
But, patching the headers still needs work:
image

Pandapip1:
Haha, found the issue:
image

Pandapip1:
I can't reproduce the error on Circle CI, so I've just put the code that's erroring in a try/catch. If this passes CI, I'll test it.

Pandapip1:
CI just passed. I'm testing to see if this works.

mrnerdhair:
I don't think this will work. Note from https://w3c.github.io/webappsec-csp/#meta-element:
> Note: Modifications to the [content](https://html.spec.whatwg.org/multipage/semantics.html#attr-meta-content) attribute of a [meta](https://html.spec.whatwg.org/multipage/semantics.html#meta) element after the element has been parsed will be ignored.

Pandapip1:
> I don't think this will work. Note from https://w3c.github.io/webappsec-csp/#meta-element

Well then... firefox doesn't follow the spec :)

I've tested the meta-modifying code, and it works correctly (see https://github.com/MetaMask/metamask-extension/pull/14233#issuecomment-1083785727)

Pandapip1:
Well, this is interesting:
image
It looks like it should be correctly modifying the headers, but it isn't. I wonder if it's to do with the following:

> However, the Content-Security-Policy header is treated differently; its values are combined to apply all the specified policies. But, if two extensions set a CSP value that conflicts, the CSP service makes the restriction more strict to resolve the conflict. For example, if one extension sets img-src: example.com, and another extension sets img-src: example.org, the result is img-src: 'none'. Merged modifications always lean towards being more restrictive, though an extension may remove the original CSP header.
(https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/onHeadersReceived)

I might have to do a bit of tweaking to make this work.

Pandapip1:
Okay, so my plan of action is to have the background page, instead of modifying the CSP header, delete it and send it to the content script, which then re-adds it.

Pandapip1:
I **think** that this should work. Can someone test this?

I've got example webpages set up at:
https://CSPTest.pandapip1.repl.co/cspheader (CSP Header)
https://CSPTest.pandapip1.repl.co/cspmeta (CSP Meta Tag)

Pandapip1:
Why is the CLA bot failing?

Pandapip1:
Good news: it injects web3 all the time
Bad news: it isn't re-adding the CSP

PR #14237, opened 2022-03-28
Fix imported component name and change size proptype type to string

tmashuang:
## Explanation

Fixes imported component name to align with its exported name, and fixes size proptype warning by changing from number to string.

## More information

Follow up PR for https://github.com/MetaMask/metamask-extension/pull/14213

## Screenshots/Screencaps
Screen Shot 2022-03-28 at 4 06 30 PM

### DIFF
--- a/ui/components/ui/icon/icon-token-search.js
+++ b/ui/components/ui/icon/icon-token-search.js
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 const IconTokenSearch = ({
-  size = 24,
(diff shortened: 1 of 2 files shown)


darkwing (changes requested):
With all of our new icons, we're using a `number` prop and passing it as `size={64}`.  I think that when we pass representation of numbers, we should try to stick to `Proptypes.number` and make sure we pass in the number

darkwing (approved):

ryanml (approved):

PR #14238, opened 2022-03-28
MV3: Revive service worker after it goes idle

adonesky1:
Revive service worker on reconnect + a few small tweaks to `manifest.json` and `background.js` for actions to work in MV3


adonesky1:
Overtaken by https://github.com/MetaMask/metamask-extension/pull/14343

PR #14251, opened 2022-03-29
Update token detection logic to only control auto-detection

### LINKED ISSUE #13945: Update token detection logic to only control auto-detection
If the token detection toggle is disabled, it should only affect the automatic token detection in the supported network. For the supported networks the dynamic token list should be available regardless of the state of if the token detection is ON or OFF

NiranjanaBinoy:
Fix: #13945

When the token detection toggle is disabled, it should only affect the automatic token detection in the supported network. For the supported networks, the dynamic token list should be available regardless of the state of if the token detection is ON or OFF

### DIFF
--- a/ui/components/ui/identicon/identicon.component.js
+++ b/ui/components/ui/identicon/identicon.component.js
@@ -151,14 +151,22 @@ export default class Identicon extends PureComponent {
     }
 
     if (address) {
-      // token from dynamic api list is fetched when useTokenDetection is true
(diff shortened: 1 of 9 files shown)


georgewrmarshall:
Removed myself as reviewer as this doesn't look like it requires any UI updates?

tmashuang (approved):

GuillaumeRx (approved):

PR #14256, opened 2022-03-30
Extend timeout

veronicaz41:
## Explanation

We are trying to generate zk proof in snap, and it keeps timeout. Extending timeout works for us.

## Manual testing steps

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -583,6 +583,7 @@ export default class MetamaskController extends EventEmitter {
         name: 'ExecutionService',
       }),
       setupSnapProvider: this.setupSnapProvider.bind(this),
+      unresponsiveTimeout: 120000,
     });
 
     const snapControllerMessenger = this.controllerMessenger.getRestricted({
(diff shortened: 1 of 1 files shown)


veronicaz41:

rekmarks:
Thank you for submitting this, @veronicaz41. I'm marking this as "do not merge" since we need to evaluate this before making a decision about how to solve your issue.

A few questions:

1. How frequently do you need to perform the expensive operation? For example, is it only at initialization or can it occur with arbitrary frequency?
2. How you checked exactly how long the operations take?

veronicaz41:
FYI, this is our snap: https://github.com/0xPARC/cabal/tree/main/snap
To answer your question:
1. It is not at initialization, it can occur with arbitrary frequency
2. I didn't exactly time it, but I tested a few numbers, anything  much lower (e.g. 90000) would timeout. I also want to give it a bit room in case we change our circuits.
This is what takes a long time: https://github.com/0xPARC/cabal/blob/main/snap/src/snark_utils/generate_proof.js#L9-L21

Happy to change a different number, or do something entirely different.  Just let us know.

veronicaz41:
Actually maybe we can make this configurable in the snap somehow, rather than hard code like this? @rekmarks

PR #13214, opened 2022-01-05
Feat/settings search

igorms-cons:
Feature: creating the search feature for settings.

Explanation:   Be able to search for a menu or section in all settings and navigate to it

Manual testing steps:  
  - Open Metamask 
  - Go to Settings 
  - type the section / tab (or its description) you want to go
  - Click et get navigate to it
  
Fix: https://github.com/MetaMask/metamask-extension/issues/12760

PR #14259, opened 2022-03-30
Add e2e testcase for Settings Search functionality

### LINKED ISSUE #14190: e2e: Add Coverage for Search Feature for Settings
Add testcase for covering "Search Feature for Settings" introduced in V10.12
See https://github.com/MetaMask/metamask-extension/pull/13214 for specs.

seaona:
## Explanation
This testcase covers the Settings Search functionality, asserting the following cases:
- `should find element inside the General tab`
- `should find element inside the Advanced tab`
- `should find element inside the Contacts tab`
- `should find element inside the Security tab`
- `should find element inside the Alerts tab`
- `should find element inside the Networks tab`
- `should find element inside the Experimental tab`
- `should find element inside the About tab`
- `should display "Element not found" for a non-existing element`

Furthermore, it checks that the found elements redirect you to the correct Settings page.

## Context
- Feature specs #13214
- Closes #14190

### DIFF
--- a/test/e2e/tests/settings-search.spec.js
+++ b/test/e2e/tests/settings-search.spec.js
@@ -21,7 +21,6 @@ describe('Settings Search', function () {
     experimental: 'Token Detection',
     about: 'Terms of Use',
   };
-  let found;
 
   it('should find element inside the General tab', async function () {
     await withFixtures(
@@ -255,7 +254,7 @@ describe('Settings Search', function () {
         await driver.clickElement({ text: 'Settings', tag: 'div' });
         await driver.fill('#search-settings', 'Lorem ipsum');
 
-        found = await driver.isElementPresent({
+        const found = await driver.isElementPresent({
           text: 'No matching results found',
           tag: 'span',
         });
--- a/test/e2e/tests/settings-search.spec.js
+++ b/test/e2e/tests/settings-search.spec.js
@@ -65,11 +65,6 @@ describe('Settings Search', function () {
         await driver.clickElement({ text: 'Settings', tag: 'div' });
         await driver.fill('#search-settings', settings.general.target);
 
-        // Check if element was found
-        found = await driver.isElementPresent({ text: 'General', tag: 'span' });
-        assert.equal(found, true, 'Element was not found');
-
-        // Check if element redirects to the correct page
         await driver.clickElement({ text: 'General', tag: 'span' });
         url = await driver.getUrl();
         assert.equal(
@@ -96,13 +91,6 @@ describe('Settings Search', function () {
         await driver.clickElement('.account-menu__icon');
         await driver.clickElement({ text: 'Settings', tag: 'div' });
         await driver.fill('#search-settings', settings.advanced.target);
-
-        // Check if element was found
-        found = await driver.isElementPresent({
-          text: 'Advanced',
-          tag: 'span',
-        });
-        assert.equal(found, true, 'Element was not found');
 
         // Check if element redirects to the correct page
         await driver.clickElement({ text: 'Advanced', tag: 'span' });
@@ -132,13 +120,6 @@ describe('Settings Search', function () {
         await driver.clickElement({ text: 'Settings', tag: 'div' });
         await driver.fill('#search-settings', settings.contacts.target);
 
-        // Check if element was found
-        found = await driver.isElementPresent({
-          text: 'Contacts',
-          tag: 'span',
-        });
-        assert.equal(found, true, 'Element was not found');
-
         // Check if element redirects to the correct page
         await driver.clickElement({ text: 'Contacts', tag: 'span' });
         url = await driver.getUrl();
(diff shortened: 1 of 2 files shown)


darkwing (dismissed):
Overall this looks pretty good.  I do see a lot of repeated code, a la:

```
await driver.navigate();
```

Not sure if maybe we could have an `async function login(driver)` we can make to cut down on that, but I don't consider it a blocker

test/e2e/tests/settings-search.spec.js
```diff
@@ -0,0 +1,347 @@
...
+        await driver.fill('#password', 'correct horse battery staple');
+        await driver.press('#password', driver.Key.ENTER);
+
+        await driver.clickElement('.account-menu__icon');
+        await driver.clickElement({ text: 'Settings', tag: 'div' });
+        await driver.fill('#search-settings', settings.general.target);
+
+        // Check if element was found
+        found = await driver.isElementPresent({ text: 'General', tag: 'span' });
+        assert.equal(found, true, 'Element was not found');
```
mcmire, line 70: You can also use `assert.ok` to assert whether something is true or not :)

test/e2e/tests/settings-search.spec.js
```diff
@@ -0,0 +1,347 @@
...
+
+        // Check if element redirects to the correct page
+        await driver.clickElement({ text: 'General', tag: 'span' });
```
mcmire, line 73: Do you need to test for the existence of the element above? `clickElement` will already throw if the element can't be found.

test/e2e/tests/settings-search.spec.js
```diff
@@ -0,0 +1,347 @@
...
+
+        // Check if element was found
+        found = await driver.isElementPresent({
+          text: 'Advanced',
+          tag: 'span',
+        });
+        assert.equal(found, true, 'Element was not found');
+
+        // Check if element redirects to the correct page
+        await driver.clickElement({ text: 'Advanced', tag: 'span' });
```
mcmire, line 108: Similar comment as above, if you use `clickElement` you're making an implicit assertion.

test/e2e/tests/settings-search.spec.js
```diff
@@ -0,0 +1,347 @@
...
+        assert.equal(found, true, 'Element was not found');
+
+        // Check if element redirects to the correct page
+        await driver.clickElement({ text: 'General', tag: 'span' });
+        url = await driver.getUrl();
+        assert.equal(
+          url.includes(settings.general.urlSubstring),
+          true,
+          'Item does not redirect to correct page',
+        );
```
mcmire, line 79: This is a philosophical thing, so take it or leave it, but to me e2e tests should always test what users see and care about as much as possible. Users don't tend to care about URLs so it might be better to test something on the page instead of the URL.

mcmire (commented):
Left some comments!

test/e2e/tests/settings-search.spec.js
seaona: That's true. I opted to make an explicit assertion, since this was the only assertion for each "it" block. A commit later I added some extra step and another assertion, so I will remove these ones, as they are already implicit on `clickElement`

seaona:
> Overall this looks pretty good. I do see a lot of repeated code, a la:
> 
> ```
> await driver.navigate();
> await driver.fill('#password', 'correct horse battery staple');
> await driver.press('#password', driver.Key.ENTER);
> ```
> 
> Not sure if maybe we could have an `async function login(driver)` we can make to cut down on that, but I don't consider it a blocker

Thank you for your review @darkwing. That's right, we have a lot of repeated code. Since this is something that applies not only to this one, but to all testcases, it needs to be solved globally. Page Objects approach would solve this, but its implementation it's still something to decide.

FYI @PeterYinusa

test/e2e/tests/settings-search.spec.js
seaona: fixed in all places

test/e2e/tests/settings-search.spec.js
seaona: mmm I see what you mean. I could check the main header (Advanced, General...) to make sure we are in the right page, without the need to check the url. What do you think? @mcmire

mcmire: Yeah, I think checking the main header makes sense!

seaona: updated. Thank you @mcmire

mcmire (dismissed):
Nice, that looks better!

0ex-d (dismissed):
Nice tests! should we add something like `afterAll` to reset some primitives.

PeterYinusa (dismissed):

seaona:
> Nice tests! should we add something like `afterAll` to reset some primitives.

Thank you for your comment! That's a good observation. At the moment we are not using afterAll / afterEach for any e2e testcase. A clean webdriver is built for every testcase instead (see `helpers.js` and `buildWebDriver`).

Seems something that could be discussed more globally on the e2e infra level, maybe this would spare some execution time for the test runs FYI @PeterYinusa

mcmire (approved):

darkwing (approved):

PR #14269, opened 2022-03-30
track rpc method usage

### LINKED ISSUE #7329: Add metrics for provider
As we publish a new version of our provider, we need a rough understanding of which API features are being used, by which websites. This includes:

```
ethereum.send
ethereum.sendAsync
eth_signTypedData
ethereum.enable
ethereum._metamask.isApproved
ethereum._metamask.isUnlocked
ethereum.selectedAddress
ethereum.networkVersion
ethereum.chainId
```

No deprecation will happen without generous warning!

### LINKED ISSUE #13582: Add metric event tracking for RPC methods called

brad-decker:
## Explanation
This adds a new RPC method middleware that tracks the inpage provider usage. This will happen once every ten seconds *per method* unless the method is one of those that are special cased because they display prompts to the user. Note that tracking the adding/switching networks is handled in its RPC method handler. Note also that all transactions are already tracked in the transaction controller.

## More information
Fixes #13582
Fixes #7329

### DIFF
--- a/app/scripts/lib/createRPCMethodTrackingMiddleware.js
+++ b/app/scripts/lib/createRPCMethodTrackingMiddleware.js
@@ -1,0 +1,81 @@
+import { SECOND } from '../../../shared/constants/time';
+
+const USER_PROMPTED_EVENT_NAME_MAP = {
+  eth_signTypedData_v4: 'Signature Requested',
+  eth_signTypedData_v3: 'Signature Requested',
+  eth_signTypedData: 'Signature Requested',
+  eth_personal_sign: 'Signature Requested',
+  eth_sign: 'Signature Requested',
+  eth_getEncryptionPublicKey: 'Encryption Public Key Requested',
+  eth_decrypt: 'Decryption Requsted',
+  wallet_requestPermissions: 'Permissions Requested',
+  eth_requestAccounts: 'Permissions Requested',
+};
+
+const samplingTimeouts = {};
+
+/**
+ * Returns a middleware that tracks inpage_provider usage using sampling for
+ * each type of event except those that require user interaction, such as
+ * signature requests
+ *
+ * @param {object} opts - options for the rpc method tracking middleware
+ * @param {Function} opts.trackEvent - trackEvent method from MetaMetricsController
+ * @param {Function} opts.getMetricsState - get the state of MetaMetricsController
+ * @returns {Function}
+ */
+export default function createRPCMethodTrackingMiddleware({
+  trackEvent,
+  getMetricsState,
+}) {
+  return function rpcMethodTrackingMiddleware(
+    /** @type {any} */ req,
+    /** @type {any} */ res,
+    /** @type {Function} */ next,
+  ) {
+    const { origin } = req;
+
+    next((callback) => {
+      if (!getMetricsState().participateInMetaMetrics) {
+        return callback();
+      }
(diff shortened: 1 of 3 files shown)


app/scripts/lib/createRPCMethodTrackingMiddleware.js
```diff
@@ -0,0 +1,81 @@
...
+  trackEvent,
+  getMetricsState,
+}) {
+  return function rpcMethodTrackingMiddleware(
+    /** @type {any} */ req,
+    /** @type {any} */ res,
+    /** @type {Function} */ next,
+  ) {
+    const { origin } = req;
+
```
shanejonas, line 37: just a suggestion: we could track the request time

```suggestion
const timeStart = Date.now();
```

app/scripts/lib/createRPCMethodTrackingMiddleware.js
```diff
@@ -0,0 +1,81 @@
...
+          category: 'inpage_provider',
+          referrer: {
+            url: origin,
+          },
+          properties: {
+            method: req.method,
+            status: userRejected ? 'rejected' : 'approved',
+            error_code: res.error?.code,
+            error_message: res.error?.message,
+            has_result: typeof res.result !== 'undefined',
```
shanejonas, line 55: should we add the params? or is that too much info in the event

app/scripts/lib/createRPCMethodTrackingMiddleware.js
brad-decker: request time like as in the time it takes from when the request comes in to when the response occurs? Or like the actual time the event occurred?

app/scripts/lib/createRPCMethodTrackingMiddleware.js
brad-decker: I think the best we could do is stringify it... or somehow loop through the args and add each as its on attribute on the event. you can send arrays but if you do they have to be all the same type.. you can't send an object.. I'd suggest we skip it for now.

rekmarks: Grabbing the params is potentially too invasive, as well.

app/scripts/lib/createRPCMethodTrackingMiddleware.js
shanejonas: the time it takes from when the request comes in to when the response occurs

brad-decker:
@shanejonas added duration property which is endTime - starrtTime, we can use mixpanel functions to manipualte the value for graphs.

app/scripts/lib/createRPCMethodTrackingMiddleware.js
```diff
@@ -0,0 +1,85 @@
+import { SECOND } from '../../../shared/constants/time';
+
+const USER_PROMPTED_EVENT_NAME_MAP = {
+  eth_signTypedData_v4: 'Signature Requested',
```
darkwing, line 4: Super minor but can we make these strings a constant so that we don't accidentally misspell one?

brad-decker: Resolved in latest commit :)

darkwing (approved):

NiranjanaBinoy (approved):

PR #14272, opened 2022-03-30
Allow locally hosted RPC and Block Explorer Urls with `wallet_addEthereumChain`

MicaiahReid:
## Explanation
Currently, adding a chain using the `wallet_addEthereumChain` RPC method requires that the supplied chain has an array of RPC urls, with at least one of them being an HTTPS url. 

This is a problem because locally hosted resources are [considered secure](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts#when_is_a_context_considered_secure), and users may want to programmatically add their own local chains (e.g. Ganache) to MetaMask.

This change instead requires that at least one of the urls is either HTTPS or locally hosted (`url.hostname === "localhost" || url.hostname === "127.0.0.1"`).

The same change is made for the block explorer urls.

## Manual testing steps

To test, run MetaMask and run `ganache --port 8544 --chain.chainId 1338`. Send the `wallet_addEthereumChain` RPC method to MetaMask with the payload:
```
{
    chainId: "0x53A",
    chainName: 'Ganache Test',
    nativeCurrency: { symbol: 'ETH', decimals: 18 },
    rpcUrls: [ 'http://127.0.0.1:8544' ],
    blockExplorerUrls: [ 'http://localhost:3000' ]
  }
```
Verify that the chain is successfully added to MetaMask.

### DIFF
--- a/app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
+++ b/app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
@@ -83,7 +83,9 @@ async function addEthereumChainHandler(
   };
 
   const firstValidRPCUrl = Array.isArray(rpcUrls)
-    ? rpcUrls.find((rpcUrl) => validUrl.isHttpsUri(rpcUrl))
+    ? rpcUrls.find(
+        (rpcUrl) => isLocalhost(rpcUrl) || validUrl.isHttpsUri(rpcUrl),
+      )
     : null;
 
   const firstValidBlockExplorerUrl =
--- a/app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
+++ b/app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
@@ -90,8 +90,10 @@ async function addEthereumChainHandler(
 
   const firstValidBlockExplorerUrl =
     blockExplorerUrls !== null && Array.isArray(blockExplorerUrls)
-      ? blockExplorerUrls.find((blockExplorerUrl) =>
-          validUrl.isHttpsUri(blockExplorerUrl),
+      ? blockExplorerUrls.find(
+          (blockExplorerUrl) =>
+            isLocalhost(blockExplorerUrl) ||
+            validUrl.isHttpsUri(blockExplorerUrl),
         )
       : null;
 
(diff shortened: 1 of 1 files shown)


MicaiahReid:

app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
```diff
@@ -76,14 +76,23 @@ async function addEthereumChainHandler(
     );
   }
 
+  const isLocalhost = (strUrl) => {
+    const url = new URL(strUrl);
+    return url.hostname === 'localhost' || url.hostname === '127.0.0.1';
```
darkwing, line 81: This PR looks good.  The only thing I'd mention is that `new URL` can throw an error if the input is a malformed URL, which is why we usually `try/catch` the `new URL` line.  Of course a malformed URL should never make it here, so maybe I'll see what others think.

brad-decker (changes requested):

app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
MicaiahReid: Done, thanks for the review!

brad-decker (dismissed):

brad-decker (dismissed):

Amphaal:
Many thanks for this PR @MicaiahReid ! Eager to use this patch ASAP 👍

darkwing (dismissed):

AS-Pando:
Just wanted to mention that this feature would really help me out. Right now I'm running a small hardhat instance for testing. I don't seem to have a problem adding my chain from the extension network page (`settings/networks/add-network`). As MicaiahReid mentioned, I'd rather add it programmatically because I'd like to allow the user to direct my d.app to whatever local chain they have installed.

MicaiahReid:
@danjm Is there anything else needed to merge this? Merging is still blocked for me, but it looks like I've had two approvals?

joani24:
> **Ação de Assinatura do CLA:** Todos os autores assinaram o CLA. Pode ser necessário reexecutar manualmente a verificação de PR de bloqueio se ela não passar em alguns minutos.

darkwing (dismissed):

adonesky1 (dismissed):
In general would ask for tests here, but doesn't seem we currently have any test suites for our RPC middleware methods...

LGTM

darkwing (approved):

adonesky1 (approved):

BelfordZ:
fixes #14416

PR #14279, opened 2022-03-30
Add new token added event (duplicating the existing event structure) when collectible is manually added

### LINKED ISSUE #13477: Add event metrics on NFT support in extension
### Describe the bug

Notion reference: https://www.notion.so/NFT-support-in-extension-v1-844f8c3e72e044c4b8d6f7da7e24b646#c9bf0bdb95e044378b6b382953e76a20

### Acceptance Criteria

~~1. **Usage of NFTs. Goal: create asset value per user metric.** i.e. NFT: $15 per user, Non-NFT: $80 per user
    a. Assets vs NFTs
        i. Asset types and value
        ii. Number of NFTs and NFT types (floor value if possible)
    b. Number of NFT collections~~
        
nft 1a
nft 1b

        
~~2. **Detection utility. Goal: elevate usage of detection that’ll lead to improvements of detection overall.**
    a. NFT detection allowed vs disabled
        i. Centralized detection
        ii. Non-centralized detection~~
            
nft 2a
            
b. NFTs added via detection
        i. Added via NFT detection
        ii. Added via Token detection
            
nft 2b
nft 2bii

c. Manually added
        
nft 2c

        
3. **Flow analysis. Goal: track flow of user journey to determine drop-offs and exits**
    a. Adding flow
        i. NFT detected (2bii) → NFT detection (centralized (2ai) / non-centralized (2aii)) → Added via NFT detection (2bi)
        ii. Manual add (2c) → Added via manual 
   b. Sending flow
        i. NFT Details
        ii. Address Added 
        iii. NFT Sent
        
nft 3bi
nft 3bii
nft 3biii

### Tasks
- [x] Add new user trait for 'NFT Autodetection Enabled'. (done #14367) 
- [x] Add new user trait for 'OpenSeas API Enabled'. (done #14367) 
- [x] Add new user trait for 'Number of NFTs'. (done #14495)
- [x] Add new user trait for 'Number of Tokens'. (done in #14427)
- [x] Add new user trait for 'Number of NFT collections'. (done #14377)
- [x] Modify existing token added events to record the `asset_type` using `ASSET_TYPES` from transactions const
- [x] Modify existing token added events to record the `token_standard` using `TOKEN_STANDARDS` from 'ui/helpers/constants/common'
- [x] Add new token added event (duplicating the existing event structure) when auto detection occurs for NFTs (in progress #14279)
- [x] Add metric tracking for NFT detection. See `CollectionDetectionController` and acceptance criteria # 3a
   - Set source property to detected on the new token added event (In Progress https://github.com/MetaMask/controllers/pull/814)
- [x] Add metric tracking for NFT send flow. See `CollectionDetectionController` and acceptance criteria # 3b
- [x] Add `asset_type` to transaction events (done in #13858)
- [x] Add `token_standard` to transaction events (done in #13858)

#### Collectibles state help
Collectibles state comes from the CollectiblesController from metamask/controllers. The key in our state tree we'll want to use is `allCollectibles`. So in the handleMetaMaskUpdate function on the MetaMetrics controller introduced in #14192 you will add some new keys to the traits object for the various tasks above specifying add new user trait.

```js
/**
 * @type Collectible
 *
 * Collectible representation
 * @property address - Hex address of a ERC721 contract
 * @property description - The collectible description
 * @property image - URI of custom collectible image associated with this tokenId
 * @property name - Name associated with this tokenId and contract address
 * @property tokenId - The collectible identifier
 * @property numberOfSales - Number of sales
 * @property backgroundColor - The background color to be displayed with the item
 * @property imagePreview - URI of a smaller image associated with this collectible
 * @property imageThumbnail - URI of a thumbnail image associated with this collectible
 * @property imageOriginal - URI of the original image associated with this collectible
 * @property animation - URI of a animation associated with this collectible
 * @property animationOriginal - URI of the original animation associated with this collectible
 * @property externalLink - External link containing additional information
 * @property creator - The collectible owner information object
 * @property isCurrentlyOwned - Boolean indicating whether the address/chainId combination where it's currently stored currently owns this collectible
 */
```

All collectibles has this shape:
```js
```

This is an object whose keys are accounts and the values are objects where the keys are chainIds. So, for example, in order to get an aggregate number of NFTs imported you'll have to go through each chainId in each account in this object and sum the length of the Collectibles array.

In order to get the number of collections you'll need to do the same sort of thing but you'll want the number of unique 'address' properties on the entries. Each collectible has an address property that indicates which collection it is a part of.

segun:
Add new token added event (duplicating the existing event structure) when collectible is manually added

Set source property to custom on the new token added event,

## More information

* Progresses #13477

### DIFF
--- a/ui/pages/add-collectible/add-collectible.js
+++ b/ui/pages/add-collectible/add-collectible.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useContext, useState } from 'react';
 import { useHistory } from 'react-router-dom';
 import { useDispatch, useSelector } from 'react-redux';
 import { util } from '@metamask/controllers';
@@ -24,6 +24,7 @@ import FormField from '../../components/ui/form-field';
 import { getIsMainnet, getUseCollectibleDetection } from '../../selectors';
 import { getCollectiblesDetectionNoticeDismissed } from '../../ducks/metamask/metamask';
 import CollectiblesDetectionNotice from '../../components/app/collectibles-detection-notice';
+import { MetaMetricsContext } from '../../contexts/metametrics';
 
 export default function AddCollectible() {
   const t = useI18nContext();
@@ -47,6 +48,7 @@ export default function AddCollectible() {
   const [tokenId, setTokenId] = useState('');
   const [disabled, setDisabled] = useState(true);
   const [collectibleAddFailed, setCollectibleAddFailed] = useState(false);
+  const trackEvent = useContext(MetaMetricsContext);
 
   const handleAddCollectible = async () => {
     try {
@@ -65,6 +67,17 @@ export default function AddCollectible() {
       );
     }
     dispatch(setNewCollectibleAddedMessage('success'));
+
+    trackEvent({
+      event: 'Collectible Added',
+      category: 'Wallet',
+      sensitiveProperties: {
+        token_contract_address: address,
+        tokenId: tokenId.toString(),
+        source: 'manual',
+      },
+    });
+
     history.push(DEFAULT_ROUTE);
   };
 
--- a/ui/pages/add-collectible/add-collectible.js
+++ b/ui/pages/add-collectible/add-collectible.js
@@ -17,6 +17,8 @@ import ActionableMessage from '../../components/ui/actionable-message';
 import PageContainer from '../../components/ui/page-container';
 import {
   addCollectibleVerifyOwnership,
+  getERC1155TokenURI,
+  getERC721TokenURI,
   removeToken,
   setNewCollectibleAddedMessage,
 } from '../../store/actions';
@@ -25,6 +27,8 @@ import { getIsMainnet, getUseCollectibleDetection } from '../../selectors';
 import { getCollectiblesDetectionNoticeDismissed } from '../../ducks/metamask/metamask';
 import CollectiblesDetectionNotice from '../../components/app/collectibles-detection-notice';
 import { MetaMetricsContext } from '../../contexts/metametrics';
+import { ASSET_TYPES } from '../../../shared/constants/transaction';
+import { TOKEN_STANDARDS } from '../../helpers/constants/common';
 
 export default function AddCollectible() {
   const t = useI18nContext();
@@ -50,6 +54,29 @@ export default function AddCollectible() {
   const [collectibleAddFailed, setCollectibleAddFailed] = useState(false);
   const trackEvent = useContext(MetaMetricsContext);
 
+  const getCollectibleStandard = async () => {
+    // try ERC721 uri
+    let tokenStandard;
+    try {
+      const result = await dispatch(getERC721TokenURI(address, tokenId));
+      if (result) {
+        tokenStandard = TOKEN_STANDARDS.ERC721;
+      }
+    } catch {
+      // try ERC1155 uri
+      try {
+        const result = await dispatch(getERC1155TokenURI(address, tokenId));
+        if (result) {
+          tokenStandard = TOKEN_STANDARDS.ERC1155;
+        }
+      } catch (err) {
+        tokenStandard = '';
+      }
+    }
+
+    return tokenStandard;
+  };
+
   const handleAddCollectible = async () => {
     try {
       await dispatch(
@@ -68,13 +95,20 @@ export default function AddCollectible() {
     }
     dispatch(setNewCollectibleAddedMessage('success'));
 
+    const tokenStandard = await getCollectibleStandard(
+      address,
+      tokenId.toString(),
+    );
+
     trackEvent({
       event: 'Collectible Added',
       category: 'Wallet',
       sensitiveProperties: {
         token_contract_address: address,
         tokenId: tokenId.toString(),
-        source: 'manual',
+        asset_type: ASSET_TYPES.COLLECTIBLE,
+        token_standard: tokenStandard,
+        source: 'custom',
       },
     });
 
--- a/ui/pages/add-collectible/add-collectible.js
+++ b/ui/pages/add-collectible/add-collectible.js
@@ -22,11 +22,7 @@ import {
   setNewCollectibleAddedMessage,
 } from '../../store/actions';
 import FormField from '../../components/ui/form-field';
-import {
-  getIsMainnet,
-  getSelectedAddress,
-  getUseCollectibleDetection,
-} from '../../selectors';
+import { getIsMainnet, getUseCollectibleDetection } from '../../selectors';
 import { getCollectiblesDetectionNoticeDismissed } from '../../ducks/metamask/metamask';
 import CollectiblesDetectionNotice from '../../components/app/collectibles-detection-notice';
 import { MetaMetricsContext } from '../../contexts/metametrics';
--- a/ui/pages/add-collectible/add-collectible.js
+++ b/ui/pages/add-collectible/add-collectible.js
@@ -17,18 +17,20 @@ import ActionableMessage from '../../components/ui/actionable-message';
 import PageContainer from '../../components/ui/page-container';
 import {
   addCollectibleVerifyOwnership,
-  getERC1155TokenURI,
-  getERC721TokenURI,
+  getTokenStandardAndDetails,
   removeToken,
   setNewCollectibleAddedMessage,
 } from '../../store/actions';
 import FormField from '../../components/ui/form-field';
-import { getIsMainnet, getUseCollectibleDetection } from '../../selectors';
+import {
+  getIsMainnet,
+  getSelectedAddress,
+  getUseCollectibleDetection,
+} from '../../selectors';
 import { getCollectiblesDetectionNoticeDismissed } from '../../ducks/metamask/metamask';
 import CollectiblesDetectionNotice from '../../components/app/collectibles-detection-notice';
 import { MetaMetricsContext } from '../../contexts/metametrics';
 import { ASSET_TYPES } from '../../../shared/constants/transaction';
-import { TOKEN_STANDARDS } from '../../helpers/constants/common';
 
 export default function AddCollectible() {
   const t = useI18nContext();
@@ -53,29 +55,7 @@ export default function AddCollectible() {
   const [disabled, setDisabled] = useState(true);
   const [collectibleAddFailed, setCollectibleAddFailed] = useState(false);
   const trackEvent = useContext(MetaMetricsContext);
-
-  const getCollectibleStandard = async () => {
-    // try ERC721 uri
-    let tokenStandard;
-    try {
-      const result = await dispatch(getERC721TokenURI(address, tokenId));
-      if (result) {
-        tokenStandard = TOKEN_STANDARDS.ERC721;
-      }
-    } catch {
-      // try ERC1155 uri
-      try {
-        const result = await dispatch(getERC1155TokenURI(address, tokenId));
-        if (result) {
-          tokenStandard = TOKEN_STANDARDS.ERC1155;
-        }
-      } catch (err) {
-        tokenStandard = '';
-      }
-    }
-
-    return tokenStandard;
-  };
+  const selectedAddress = useSelector(getSelectedAddress);
 
   const handleAddCollectible = async () => {
     try {
@@ -95,8 +75,9 @@ export default function AddCollectible() {
     }
     dispatch(setNewCollectibleAddedMessage('success'));
 
-    const tokenStandard = await getCollectibleStandard(
+    const tokenDetails = await getTokenStandardAndDetails(
       address,
+      selectedAddress,
       tokenId.toString(),
     );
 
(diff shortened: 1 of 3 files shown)


ui/pages/add-collectible/add-collectible.js
```diff
@@ -46,8 +50,71 @@ export default function AddCollectible() {
...
+
+    if (alreadyTracked) {
+      return;
+    }
+
+    trackEvent({
+      event: 'Token Added',
+      category: 'Wallet',
+      sensitiveProperties: {
+        token_symbol: '',
```
brad-decker, line 93: best to leave these empty strings as undefined (just omit the properties), segment will strip them anyways.

digiwand (commented):
LGTM after updating the code based on what Brad [commented](https://github.com/MetaMask/metamask-extension/pull/14279#discussion_r838950067)

I see that the reused pattern from the codebase has been designed to support more events 👍 Can we get a reference to the related, event tracking pattern you found in the codebase?

ui/pages/add-collectible/add-collectible.js
segun: Fixed [here](https://github.com/MetaMask/metamask-extension/pull/14279/commits/1fd076232dab0526122622ef0968e9f664518c6d)

ui/pages/add-collectible/add-collectible.js
```diff
@@ -46,8 +50,68 @@ export default function AddCollectible() {
...
+      return;
+    }
+
+    trackEvent({
+      event: 'Token Added',
+      category: 'Wallet',
+      sensitiveProperties: {
+        token_contract_address: newCollectible.address,        
+        tokenId: newCollectible.tokenId,
+        name: newCollectible.name ? newCollectible.name : null,
```
digiwand, line 95:
```suggestion
        name: newCollectible.name || null,
```
nit: we could shorten this. Same comment for the `description` prop below

digiwand (dismissed):
LGTM! left just a minor suggestion related to [lines 95 & 96](https://github.com/MetaMask/metamask-extension/pull/14279/files#diff-d1e7bb97394326d6e78a5d8631e4d88531660b2f9e21ce61f5f3906dd556eac5R95)

ui/pages/add-collectible/add-collectible.js
segun: fixed [here](https://github.com/MetaMask/metamask-extension/pull/14279/commits/2dff6f9ebd06242a63bd80eec250ec59d245e5e4)

ui/pages/add-collectible/add-collectible.js
```diff
@@ -46,8 +50,66 @@ export default function AddCollectible() {
...
+  const collectibles = useSelector(getCollectibles);
+
+  useEffect(() => {
+    if (!tokenId || !address) {
+      return;
+    }
+
+    const newCollectible = collectibles.find(
+      (collectible) =>
+        collectible.address === address &&
```
adonesky1, line 73: might be worth making this a case-insensitive check until we have thoroughly standardized checksumming across controllers.

segun: fixed [here](https://github.com/MetaMask/metamask-extension/pull/14279/commits/d68dc9564e5ee8b6f7989ccbbb799b3604389567)

ui/pages/add-collectible/add-collectible.js
```diff
@@ -46,8 +50,66 @@ export default function AddCollectible() {
...
+        collectible.tokenId === tokenId.toString(),
+    );
+
+    if (!newCollectible) {
+      return;
+    }
+
+    const alreadyTracked = eventTracked.find(
+      (et) => et.address === address && et.tracked === true,
```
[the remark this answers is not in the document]
segun: fixed [here](https://github.com/MetaMask/metamask-extension/pull/14279/commits/d68dc9564e5ee8b6f7989ccbbb799b3604389567)

digiwand (dismissed):
Awesome, LGTM! 

Good to know about [@adonesky1's comment above](https://github.com/MetaMask/metamask-extension/pull/14279#discussion_r842893856):
> might be worth making this a case-insensitive check until we have thoroughly standardized storing addresses as checksummed across controllers.

now I'm wondering if we want to normalize the addresses here and/or through Segment.
cc: @brad-decker

jpuri (dismissed):

ui/pages/add-collectible/add-collectible.js
```diff
@@ -46,8 +50,68 @@ export default function AddCollectible() {
...
+  const collectibles = useSelector(getCollectibles);
+
+  useEffect(() => {
+    if (!tokenId || !address) {
+      return;
+    }
+
+    const newCollectible = collectibles.find(
+      (collectible) =>
+        collectible.address.toLowerCase() === address.toLowerCase() &&
```
adonesky1, line 73: [nit] - you could/should use the `isEqualCaseInsensitive` util method

brad-decker (changes requested):
potentially blocking question: ... Is this tracking token added when the token is detected or when the token is added to the wallet? :

ui/pages/add-collectible/add-collectible.js
```diff
@@ -46,8 +50,68 @@ export default function AddCollectible() {
   );
   const [tokenId, setTokenId] = useState('');
   const [disabled, setDisabled] = useState(true);
+  const [eventTracked, setEventTracked] = useState([
+    {
+      event: 'Token Added',
```
adonesky1, line 55: Should this be collectible/NFT added? Sorry I don't have context on the schema we're using, are we somehow indicating that this is an collectible/NFT elsewhere as a subcategory of token?

ui/pages/add-collectible/add-collectible.js
```diff
@@ -46,8 +50,68 @@ export default function AddCollectible() {
...
+
+    trackEvent({
+      event: 'Token Added',
+      category: 'Wallet',
+      sensitiveProperties: {
+        token_contract_address: newCollectible.address,
+        tokenId: newCollectible.tokenId,
+        name: newCollectible.name || null,
+        description: newCollectible.description || null,
+        source: 'detected',
```
adonesky1, line 99: A couple of broader questions adding to @brad-decker's point.
1. This is not a "detected" add token event. This is a manually added token. 
and 2. I'm a bit confused about why this track event is being called in a `useEffect` on the page where the user is adding the collectible. Is the idea that the value of `collectibles` will change when the user submits the form to add the collectible and this effect will happen before the component dismounts from the history.push event called on line 132 (at the end of the submission handler)?

segun: For 2: Yes. I did test this. The useEffect is called before the component dismounts. 
For 1: Do we have a "detected" flow for NFTs?

segun:
> potentially blocking question: ... Is this tracking token added when the token is detected or when the token is added to the wallet? :

It's added when the token is added to the wallet.

ui/pages/add-collectible/add-collectible.js
digiwand: oh. interesting. I was also confused by the original request, "when auto detection occurs for NFTs".

The existing patterns for "Token Added" do not live in `useEffect` hooks. If we were following the pattern used for "Token Added", we could call `trackEvent` in the same method or callback that adds the token. In this case, could we call track in `handleAddCollectible`?  I'm wondering where you found this pattern original pattern with `useEffect`

ui/pages/add-collectible/add-collectible.js
brad-decker: No we should use the asset_type and token_standard properties that match the transaction events to indicate that this 'Added Token' is a collectible/NFT

ui/pages/add-collectible/add-collectible.js
```diff
@@ -65,6 +67,17 @@ export default function AddCollectible() {
...
     }
     dispatch(setNewCollectibleAddedMessage('success'));
+
+    trackEvent({
+      event: 'Collectible Added',
+      category: 'Wallet',
+      sensitiveProperties: {
+        token_contract_address: address,
+        tokenId: tokenId.toString(),
+        source: 'manual',
```
brad-decker, line 77: we're missing properties here. asset_type, token_standard... and source should 'custom' (if this is when the user manually types in all the details of the token)

brad-decker (changes requested):

ui/pages/add-collectible/add-collectible.js
segun: fixed [here](https://github.com/MetaMask/metamask-extension/pull/14279/commits/a82b0a8b6ff401b1be075480a26a729d3ba46c24)

ui/pages/add-collectible/add-collectible.js
```diff
@@ -47,6 +52,30 @@ export default function AddCollectible() {
   const [tokenId, setTokenId] = useState('');
   const [disabled, setDisabled] = useState(true);
   const [collectibleAddFailed, setCollectibleAddFailed] = useState(false);
+  const trackEvent = useContext(MetaMetricsContext);
+
+  const getCollectibleStandard = async () => {
```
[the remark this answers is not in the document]
segun: fixed [here](https://github.com/MetaMask/metamask-extension/pull/14279/commits/06de1bf378dce041c50a35dcf20691fa6e123d49)

ui/pages/add-collectible/add-collectible.js
```diff
@@ -95,8 +75,9 @@ export default function AddCollectible() {
     }
     dispatch(setNewCollectibleAddedMessage('success'));
 
-    const tokenStandard = await getCollectibleStandard(
+    const tokenDetails = await getTokenStandardAndDetails(
       address,
+      selectedAddress,
```
adonesky1, line 80: We actually don't need this here since it is not used for non-ERC20 tokens. Its a weird function signature that we will change (to accept an object) but for now you can and should put the second arg as `null` and no longer need to get the selectedAddress from state.

ui/pages/add-collectible/add-collectible.js
adonesky1: You could add this field back now since some NFTs have a symbol and it is returned by the `getTokenStandardAndDetails` method that you are using now.

adonesky1: cc @segun ☝️

adonesky1:
@segun just want to make sure you see my comment [here](https://github.com/MetaMask/metamask-extension/pull/14279#discussion_r848775987)

chimerahedonista:
> @segun just want to make sure you see my comment [here](https://github.com/MetaMask/metamask-extension/pull/14279#discussion_r848775987)

Oh I missed that comment.

ui/pages/add-collectible/add-collectible.js
chimerahedonista: Somehow missed this comment....fixed [here](https://github.com/MetaMask/metamask-extension/pull/14279/commits/b2b69c05d13ded9b2180dcda2e1afe06c9a004bf)

segun: Somehow missed that....fixed [here](https://github.com/MetaMask/metamask-extension/pull/14279/commits/b2b69c05d13ded9b2180dcda2e1afe06c9a004bf)

adonesky1: @segun that commit doesn't see to re-add the `token_symbol` to the event?

adonesky1:
@segun another comment still to address here: https://github.com/MetaMask/metamask-extension/pull/14279#discussion_r850491673

ui/pages/add-collectible/add-collectible.js
```diff
@@ -65,6 +69,26 @@ export default function AddCollectible() {
...
+    trackEvent({
+      event: 'Token Added',
+      category: 'Wallet',
+      sensitiveProperties: {
+        token_contract_address: address,
+        token_symbol: tokenDetails?.symbol,
+        tokenId: tokenId.toString(),
+        asset_type: ASSET_TYPES.COLLECTIBLE,
+        token_standard: tokenDetails?.standard,
+        source: 'custom',
```
adonesky1, line 88: just a question what does `source: 'custom'` mean?

adonesky1 (approved):
On non blocking question but LGTM

ui/pages/add-collectible/add-collectible.js
brad-decker: custom means its added manually. possible options right now are 'custom' | 'dapp' | 'list' | 'detected'

digiwand (approved):
👍  can we update the description too please? 

currently, it says: 
> Set source property to manual on the new token added event

and we've updated the code to set the source to "custom"

ui/pages/add-collectible/add-collectible.js
digiwand: @segun 

should we set this to `detected`? In [#13477](https://github.com/MetaMask/metamask-extension/issues/13477), one of our tasks is to 
>  Set source property to detected on the new token added event

adonesky1:
@brad-decker have your "requested changes" been addressed? Currently this PR is only blocked by that request.

brad-decker:
me: Welp, this is good to go..why aren't they merging it.
also me: .... 🤦🏻‍♂️

brad-decker (approved):

PR #14286, opened 2022-03-30
Updating design-tokens to 1.5.1

georgewrmarshall:
## Explanation

Updating `@metamask/design-tokens` to `v1.5.1` which includes these updates:

- dark theme background.default and background.alternative values are swapped
- dark theme overlay colors are black instead of white. Same as light mode

**NOTE: This does not effect any default theme / light theme colors ONLY dark theme**

### Before

Screen Shot 2022-03-30 at 3 52 53 PM

### After

Screen Shot 2022-03-30 at 3 57 07 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,7 @@
     "@material-ui/core": "^4.11.0",
     "@metamask/contract-metadata": "^1.31.0",
     "@metamask/controllers": "^27.0.0",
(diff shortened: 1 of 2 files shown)


darkwing (approved):
Wow, I love this!

brad-decker (approved):

PR #14288, opened 2022-03-30
Update version parsing to allow rollback release

Gudahtt:
## Explanation

When we want to rollback a release on Chrome, sometimes we use the fourth part of the version for the rollback release. This is because the Chrome web stores does not directly allow rolling back, but instead requires us to re-submit the release we want to roll back to with a higher version number.

The manifest version parsing now allows for a fourth version part.

The comments have also been updated to be more descriptive, and to fix a minor inaccuracy.

## Manual testing steps

- Try adding a fourth part to the version number in `package.json` (e.g. "v10.12.2.1").
- Ensure that the extension can connect to dapps, that there are no background console errors, and that the "About" page renders the version properly.


ryanml (dismissed):

app/scripts/platforms/extension.js
```diff
@@ -90,13 +90,17 @@ export default class ExtensionPlatform {
       if (versionParts.length < 4) {
         throw new Error(`Version missing build number: '${version}'`);
       }
-      // On Chrome, a more descriptive representation of the version is stored
-      // in the `version_name` field for display purposes.
+      // On Chrome, a more descriptive representation of the version is stored in the
+      // `version_name` field for display purposes. We use this field instead of the `version`
+      // field on Chrome for for non-main builds (i.e. Flask, Beta) because we want to show the
```
darkwing, line 95: `for for`

darkwing (changes requested):
Tiniest, most infuriating nit ever.

app/scripts/platforms/extension.js
Gudahtt:
```suggestion
      // field on Chrome for non-main builds (i.e. Flask, Beta) because we want to show the
```

Gudahtt: Thanks!

darkwing (approved):

brad-decker (approved):
LGTM. Tested locally as well.

PR #14318, opened 2022-03-31
Adding the link component to load Tokens found page

### LINKED ISSUE #13949: Add tokens detected link below token list on `Home` page.
Add the `x new tokens found in this account` link below the Asset list tab, where x is the number of tokens detected.

Screen Shot 2022-03-14 at 10 18 39 PM

NiranjanaBinoy:
Fix #13949 

Adding a new link with number of tokens detected is added to the Home page.

Screen Shot 2022-03-31 at 5 06 18 PM
.

### DIFF
--- a/ui/components/app/asset-list/detetcted-tokens-link/index.scss
+++ b/ui/components/app/asset-list/detetcted-tokens-link/index.scss
@@ -1,0 +1,5 @@
+.detected-tokens-link {
+  & &__link {
+    @include H6;
+  }
+}
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -2220,6 +2220,10 @@
   "notifications9Title": {
     "message": "👓 We are making transactions easier to read."
   },
+  "numberOfNewTokensDetected": {
+    "message": "$1 new tokens found in this account",
+    "description": "$1 is the number of new tokens detected"
+  },
   "ofTextNofM": {
     "message": "of"
   },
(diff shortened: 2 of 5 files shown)


ui/components/app/asset-list/detetcted-tokens-link/index.scss
```diff
@@ -0,0 +1,5 @@
+.detected-tokens-link {
+  & &__link {
+    @include H6;
+  }
+}
```
brad-decker, line 5: @georgewrmarshall we might consider adding a "typographyVariant" or some such prop to the Button component if this becomes a standard use case.  (not a change request or NIT @NiranjanaBinoy , just calling attention to possible design system component updates :))

brad-decker (approved):

ui/components/app/asset-list/detetcted-tokens-link/index.scss
georgewrmarshall: Right. The other option is to use `type="inline"` which inherits the font-size from it's parent. I see your point though if the component isn't inside a parent with a font-size how does the developer match the designs.

georgewrmarshall (changes requested):
Part of me wonders if this needs to be a component. It is quite fractal sometimes its good to weigh up the performance concerns of wrapping a component in another function or just using the foundational components to build it in context. I can see an argument for both. I think this is ok. I would add a `className` prop so it can be accessed via css in its context incase it needs positioning or any bounding box properties(margin, padding). Which should be the parent component's responsibility

georgewrmarshall (approved):

PR #14330, opened 2022-04-01
Replace `ReadOnlyInput`

Gudahtt:
## Explanation

Replace the last two uses of the `ReadOnlyInput` with a div. The idea of using an input for read-only data is silly. We can just put it in the DOM directly instead.

## Screenshots/Screencaps

Both displayed private keys are empty accounts from new wallets.

### Before

before

### After

after

## Manual testing steps

- Open the "Account Options" menu (i.e. the three-veritical-dots menu)
- Click "Account Details"
- Click "Export Private Key"
- Ensure that the account address is displayed properly (this is the first replaced element)
- Enter your password
- Ensure the private key is displayed properly (this is the second replaced element)
- Ensure that you can copy the private key by clicking on it


georgewrmarshall (dismissed):
Changes LGTM! I have a couple questions that could be out of the scope of this ticket..

Screen Shot 2022-04-01 at 11 31 38 AM

Gudahtt:
Thanks @georgewrmarshall !

For the address, users can see the full address on hover, and can copy by selecting or double-clicking. Since the full value is in the DOM that should work mos of the time. We should definitely have a copy button but that seems out of scope for this issue, this is meant to be a non-user-facing change.

For the "click to copy" text, that remains true so seems best to keep it for now. That functionality is... not intuitive and we should replace that with a dedicated copy button, but again best left to a future PR.

danjm (approved):
LGTM

digiwand (changes requested):
[edit] Change request not needed. We are losing the ability for our users to focus on the elements here, but this will not be needed when we add the copy button as mentioned in the above and below comments.  Focusing on the copy button in the future will solve the functionality concern I had. 👍 

Thanks for the explanation @Gudahtt!

Gudahtt:
These fields are for reading. You don't need to be able to focus an element in order to read it. Focus would help with copying the value, but as mentioned in the previous comment, the standard solution for that in MetaMask is a dedicated "copy" button and that will be added in a later PR.

brad-decker (approved):

darkwing:
I was going to hold this up but Mark's suggestion of adding a copy button was exactly what I wanted.  We should probably think about redesigning this at some point.

digiwand (approved):
Okay, thanks for the info and consideration! I'd be up for assisting with redesigns to this in the future. LGTM

Gudahtt:
Here is the issue for adding a copy button: https://github.com/MetaMask/metamask-extension/issues/14342

Issue #13672, opened 2022-02-17
Refactor Settings page to use template system


brad-decker:
If the following are true
1. Settings are *usually* toggles, dropdowns, textfields, etc
2. We need a JSON object structure to power search capabilities #13214 
3. That JSON structure is relied upon in many places to make sure search capabilities work
4. Adding a new setting in code requires touching many different files now to keep search working
5. The current way of scrolling a section into view by registering refs with sections could potentially lead to memory leaks or at least make memory leaks harder to diagnose
6. Settings page is among some of the oldest CSS currently maintained 

It makes sense to me to refactor the settings page to use the template system such that a single JSON tree can be used to both render search and the settings under it. We could greatly reduce complexity of the current implementation as well as make adding new settings relatively painless. We would need to make sure that the settings page has UI components for things like toggles, textfields, etc, but this would make updating the look and feel of the settings page profoundly easier in the future. 

I would say that we would simply have the contacts/address book and network tabs as exceptions to this rule given that they do not follow the pattern of being predominately individual toggles/fields

mohamadrezakheradmand:
> If the following are true
> 1. Settings are *usually* toggles, dropdowns, textfields, etc
> 2. We need a JSON object structure to power search capabilities #13214 
> 3. That JSON structure is relied upon in many places to make sure search capabilities work
> 4. Adding a new setting in code requires touching many different files now to keep search working
> 5. The current way of scrolling a section into view by registering refs with sections could potentially lead to memory leaks or at least make memory leaks harder to diagnose
> 6. Settings page is among some of the oldest CSS currently maintained 
> 
> It makes sense to me to refactor the settings page to use the template system such that a single JSON tree can be used to both render search and the settings under it. We could greatly reduce complexity of the current implementation as well as make adding new settings relatively painless. We would need to make sure that the settings page has UI components for things like toggles, textfields, etc, but this would make updating the look and feel of the settings page profoundly easier in the future. 
> 
> I would say that we would simply have the contacts/address book and network tabs as exceptions to this rule given that they do not follow the pattern of being predominately individual toggles/fields

mohamadrezakheradmand:
> If the following are true
> 1. Settings are *usually* toggles, dropdowns, textfields, etc
> 2. We need a JSON object structure to power search capabilities #13214 
> 3. That JSON structure is relied upon in many places to make sure search capabilities work
> 4. Adding a new setting in code requires touching many different files now to keep search working
[...]

PR #14350, opened 2022-04-05
Settings search refactor

hmalik88:
Refactored settings search code.

The purpose of this refactor is to introduce a cleaner, more efficient implementation of settings search. As it stands, the route objects getter function is creating a new array of route objects every time it is called. The idea is to extract something that should be a constant into its own file for easy updating in the future and to introduce some caching when it comes to calculating a filtered array of route objects. This update also facilitates in tackling issue #13672. Unnecessary util functions were removed and logic for feature flagged settings was simplified to a `featureFlag` property on the routeObject itself. This refactor is also making it easier to implement snaps view search.

**Note:** This also solves the following bugs: 

1. Ampersands weren't allowed in search input (when there's a setting that uses an ampersand)
2. Duplicate entries show in search results for certain settings when typing in an exact match into search input.


ui/helpers/utils/settings-search.js
```diff
@@ -1,420 +1,54 @@
...
+    }
+    return true;
+  });
+  return settingsRoutes;
 }
 
-function getFilteredSettingsRoutes(t, tabName) {
-  return getSettingsRoutes(t).filter((s) => s.tab === tabName);
+function getFilteredSettingsRoutes(tabName) {
+  return getSettingsRoutes().filter((s) => s.tab === tabName);
```
rekmarks, line 20: Single-character variable names are bad and we should do something different, like in my suggestion above.

ui/helpers/constants/settings.js
```diff
@@ -0,0 +1,342 @@
...
+  ABOUT_US_ROUTE,
+  NETWORKS_ROUTE,
+  CONTACT_LIST_ROUTE,
+  EXPERIMENTAL_ROUTE,
+  ///: BEGIN:ONLY_INCLUDE_IN(flask)
+  SNAPS_LIST_ROUTE,
+  ///: END:ONLY_INCLUDE_IN
+} from './routes';
+
+export const SETTINGS_ROUTES = [
```
rekmarks, line 15: These aren't just routes. Maybe something like SETTINGS_CONSTANTS would be better?

rekmarks (dismissed):
Looks good! Did just a quick review, we'll dig in further tomorrow.

ui/helpers/constants/settings.js
hmalik88: sure

ui/pages/settings/settings-search-list/settings-search-list.js
```diff
@@ -29,32 +29,32 @@ export default function SettingsSearchList({ results, onClickSetting }) {
...
                     'settings-page__header__search__list__item__tab',
                     {
                       'settings-page__header__search__list__item__tab-multiple-lines':
-                        tab === t('securityAndPrivacy'),
+                        tabKey === 'securityAndPrivacy',
                     },
                   )}
                 >
-                  {tab}
+                  {t(tabKey)}
```
brad-decker, line 41: This is problematic, and the reason that the original implementation had passed the t method around. Setting variables to message keys and passing those around become a nightmare for figuring out how  messages are being used, and they result in our strict evaluation of where message keys are being used to fail.

brad-decker (changes requested):

ui/pages/settings/settings-search-list/settings-search-list.js
hmalik88: Cool, made some changes

brad-decker (dismissed):
Very nice

danjm:
Can you put some comments in the description of the motivation for this refactor and a summary of the changes (and where appropriate, the benefit/impact of the changes)?

hmalik88:
@danjm updated the description.

rekmarks (approved):
LGTM!

brad-decker (approved):

PR #14362, opened 2022-04-05
[DO NOT MERGE] Test #14233's compatibility with MV3

### LINKED ISSUE #13972: MV3: fix injecting provider into DAPP

Pandapip1:
Do not merge. I am testing if #14233 fixes #13972.


Pandapip1:
Oops, sorry for the mass ping

PR #14366, opened 2022-04-05
Add New Tokens Found page

### LINKED ISSUE #13950: Add New Tokens Found page
Add ‘x’ new tokens found page to display detected tokens with the ability to ignore or add token

- [ ] Token address component - #14104 
Screen Shot 2022-03-18 at 4 31 41 PM

- [ ] Token list component - #14157 
Screen Shot 2022-03-18 at 4 32 17 PM
Screen Shot 2022-03-18 at 4 32 41 PM

- [ ] Token values component - #14274 
Screen Shot 2022-03-30 at 1 22 21 PM

- [ ] Token  component - #14169 
Screen Shot 2022-03-18 at 4 34 02 PM

- [ ] Parent component
Screen Shot 2022-03-18 at 4 33 44 PM

- [ ] Ignore Token Modal
Screen Shot 2022-03-18 at 4 35 14 PM

- [ ] Token added notification
Screen Shot 2022-03-18 at 4 35 52 PM

NiranjanaBinoy:
Fixes #13950 

Adding a new component to add newly detected tokens to the state.

https://user-images.githubusercontent.com/43930900/163239269-c5630860-0496-430d-9bba-030c613a7478.mov


ui/components/app/detected-token/detected-token-values/detected-token-values.js
```diff
@@ -13,32 +13,47 @@ import {
...
+const DetectedTokenValues = ({
+  token,
+  handleTokenSelection,
+  selectedTokens,
+}) => {
+  const [selectToken, setSelectToken] = useState(() => {
+    return Boolean(
+      selectedTokens.filter((address) => token.address === address),
+    );
+  });
```
brad-decker, line 25: This will always be true, because if it finds no match it'll return an empty array which Boolean([]) is true

```
Boolean([].filter(x => x === 1))
> true
```

ui/components/app/detected-token/detected-token.js
```diff
@@ -0,0 +1,109 @@
...
+import DetectedTokenSelectionPopover from './detected-token-selection-popover/detected-token-selection-popover';
+import DetectedTokenIgnoredPopover from './detected-token-ignored-popover/detected-token-ignored-popover';
+
+const DetectedToken = () => {
+  const dispatch = useDispatch();
+  const history = useHistory();
+
+  const detectedTokens = useSelector(getDetectedTokensInCurrentNetwork);
+
+  const [selectedTokens, setSelectedTokens] = useState(detectedTokens);
```
brad-decker, line 23: [Suggestion] You might consider changing the shape of this state object from an array to an object, keyed by address. If you do this then you can avoid array lookups everywhere down the chain where you're dealing with a single token. You can also avoid, in most cases, passing this state array down the tree and instead derive the specific values you need from it and pass those. Your state initialization would look like

```js
const [selectedTokens, setSelectedTokens] = useState(() => _.chain(detectedTokens).keyBy('address').mapValues(() => true).value())
```

chain is a lodash method that creates a lodash functional chain around a value. keyBy creates an object by iterating over the value and using the suppplied key to use as the key for the new object, mapValues iterates over an object and updates the values, and .value() is the end of the chain returning the value computed thus far.

Another way to write this without lodash:

```js
const [selectedTokens, setSelectedTokens] = useState(() => detectedTokens.reduce((tokenObj, token) => {
  tokenObj[token.address] = true;
  return tokenObj;
}, {}));
```

Either way is acceptable. 

Now, later in your code you'll have to do things like where you were previously doing ```dispatch(importTokens(selectedTokens))``` you'll do ```dispatch(importToken(Object.keys(selectedTokens).filter(address => selectedTokens[address]).map(address => detectedTokens[address]);```


but also this:

```js
  const handleTokenSelection = (token) => {
    let newSelectedTokens = [...selectedTokens];
    let newUnSelectedTokens = [...unSelectedTokens];

    if (
      newSelectedTokens.find(({ address }) =>
        isEqualCaseInsensitive(address, token.address),
      )
    ) {
      newSelectedTokens = newSelectedTokens.filter(
        ({ address }) => address !== token.address,
      );
      newUnSelectedTokens.push(token);
    } else {
      newSelectedTokens.push(token);
      newUnSelectedTokens = newUnSelectedTokens.filter(
        ({ address }) => address !== token.address,
      );
    }
    setSelectedTokens(newSelectedTokens);
    setUnSelectedTokens(newUnSelectedTokens);
  };
```

can become

```js
  const handleTokenSelection = (token) => {
    setSelectedTokens((prevSelectedTokens => ({
      ....prevSelectedTokens,
     [token.address]: prevSelectedTokens[token.address] ? false : true,
   )}
 };
```

You can then get rid of the 'unselectedTokens' state because that is just the token addresses that are currently false.

Let me know if that makes sense? Also in things like detected-token-values you don't have to pass in the selectedTokens state and do a lookup. You can simply check if the address of the token is 'true' of 'false' in the selected token state and pass in the value of whether its checked or not, greatly simplifying logic.

brad-decker (changes requested):

tmashuang (dismissed):
lgtm

ui/pages/home/index.scss
```diff
@@ -176,4 +176,35 @@
       margin-left: 44px;
     }
   }
+
+  &__new-tokens-imported-notification {
+    margin-bottom: 4px;
+    margin-left: 8px;
```
darkwing, line 182: Have you tried looking at these changes in `RTL` direction?  It may be the case that using `margin-inline-start` would make positioning look good in both RTL and LTR.

ui/components/app/detected-token/detected-token.js
```diff
@@ -0,0 +1,105 @@
...
+
+  const handleTokenSelection = (token) => {
+    const newSelectedTokens = { ...selectedTokens };
+
+    if (selectedTokens[token.address]) {
+      delete newSelectedTokens[token.address];
+    } else {
+      newSelectedTokens[token.address] = token;
+    }
+    setSelectedTokens(newSelectedTokens);
```
brad-decker, line 56: Getting better but  I still think we could make it better by 

1. `selectedTokensList` -> `tokensList`
2. `[token.address]: token` -> `[token.address]: true/false`
3. delete unSelectedTokensList because a false value indicates it is not selected

then inside handleTokenSelection you can just flip the boolean value instead of adding/deleting entries. 

If you *need* the token itself too, you can switch this to: 

```
[token.address]: { token. selected: true/false}
```

brad-decker (changes requested):

tmashuang (dismissed):

ui/pages/home/index.scss
```diff
@@ -176,4 +176,35 @@
...
+
+    @media screen and (min-width: $break-large) {
+      margin-bottom: 50px;
+      margin-inline-end: 8px;
+      left: 70%;
+    }
+
+    &-title,
+    &-message {
+      text-align: left;
```
darkwing, line 193: Could `text-align: start` work better here for RTL, or should this always be on the left?

ui/pages/home/index.scss
```diff
@@ -176,4 +176,35 @@
...
+
+  &__new-tokens-imported-notification {
+    margin-bottom: 4px;
+    margin-inline-start: 8px;
+    padding: 8px;
+
+    @media screen and (min-width: $break-large) {
+      margin-bottom: 50px;
+      margin-inline-end: 8px;
+      left: 70%;
```
darkwing, line 188: Should this be on the left for RTL?

ui/components/app/detected-token/detected-token.js
```diff
@@ -0,0 +1,127 @@
...
+  const handleTokenSelection = (token) => {
+    const newTokensListDetected = { ...tokensListDetected };
+
+    if (tokensListDetected[token.address].selected) {
+      newTokensListDetected[token.address].selected = false;
+    } else {
+      newTokensListDetected[token.address].selected = true;
+    }
+
+    setTokensListDetected(newTokensListDetected);
```
brad-decker, line 67:
```suggestion
    setTokensListDetected((prevState) => ({
       ...prevState,
       [prevState[token.address]]: {
         ...prevState[token.address],
         selected: prevState[token.address].selected ? false : true
      }
    });
```

ui/components/app/detected-token/detected-token.js
```diff
@@ -0,0 +1,127 @@
...
+      .map(({ token }) => {
+        return token;
+      });
+    const unSelectedTokens = Object.values(tokensListDetected)
+      .filter(({ selected }) => {
+        return !selected;
+      })
+      .map(({ token }) => {
+        return token;
+      });
```
brad-decker, line 45:
```suggestion
    // create a lodash chain on this object
    const { selected, deselected } = chain(tokensListDetected)
      // get the values
      .values()
      // create a new object with keys 'selected', 'deselected' and group the tokens
      .groupBy((token) => (token.selected ? 'selected' : 'deselected'))
       // ditch the 'selected' property and get just the tokens'
      .mapValues((group) => group.map(({ token }) => token))n
      // Exit the chain and get the underlying value, an object.
      .value();
```

you could also do a reduce call if you prefer, but lets reduce the amount of times we have to iterate over the tokens. right now there are 4n operations. The above approach is 2n + 2, a reduce statement would be n.

brad-decker (changes requested):

ui/pages/home/index.scss
NiranjanaBinoy: @darkwing I am facing this issue when I remove the left alignment here, can you help me tackle this?

https://user-images.githubusercontent.com/43930900/164549440-bfacb624-b5c5-47da-88a7-3585f071e5b7.mov

brad-decker (dismissed):

tmashuang (dismissed):

app/_locales/en/messages.json
```diff
@@ -2027,6 +2033,12 @@
   "newToMetaMask": {
     "message": "New to MetaMask?"
   },
+  "newTokensImportedMessage": {
+    "message": "You’ve successfully imported $1."
```
darkwing, line 2037: This needs a `description` key.

ui/components/app/asset-list/asset-list.js
```diff
@@ -64,7 +66,7 @@ const AssetList = ({ onClickAsset }) => {
   });
 
   const primaryTokenImage = useSelector(getNativeCurrencyImage);
-  const isMainnet = useSelector(getIsMainnet) || process.env.IN_TEST;
+  const detectedTokens = useSelector(getDetectedTokensInCurrentNetwork) || -1;
```
darkwing, line 69: Could this be `0` instead of `-1`?

ui/components/app/detected-token/detected-token-aggregators/detected-token-aggregators.js
```diff
@@ -11,18 +12,22 @@ import {
   TYPOGRAPHY,
 } from '../../../../helpers/constants/design-system';
 
-const DetectedTokenAggregators = ({ aggregatorsList }) => {
-  const t = useContext(I18nContext);
-  const numOfHiddenAggregators = parseInt(aggregatorsList.length, 10) - 2;
+const DetectedTokenAggregators = ({ aggregators }) => {
+  const t = useI18nContext();
+  const numOfHiddenAggregators = parseInt(aggregators.length, 10) - 2;
```
darkwing, line 17: Can we make `2` a constant with a meaningful name?  When looking at this, I don't immediately understand what it represents.

ui/components/app/detected-token/detected-token-ignored-popover/index.scss
```diff
@@ -0,0 +1,13 @@
+.detected-token-ignored-popover {
+  &__ignore-button {
+    margin-right: 8px;
```
darkwing, line 3: Does this popover look as it should in RTL?  Just wondering because of the left/right properties.

ui/components/app/detected-token/detected-token.js
```diff
@@ -0,0 +1,123 @@
...
+      [token.address]: {
+        ...prevState[token.address],
+        selected: !prevState[token.address].selected,
+      },
+    }));
+  };
+
+  const onImport = async () => {
+    // create a lodash chain on this object
+    const { selected: selectedTokens } = chain(tokensListDetected)
```
darkwing, line 69: Is this the same chain routine as the previous?  Maybe we create a helper function for this?

darkwing (changes requested):
Just a few comments but this is looking great!

ui/components/app/detected-token/detected-token-ignored-popover/index.scss
NiranjanaBinoy: updated

darkwing (dismissed):

brad-decker (dismissed):

darkwing (approved):

brad-decker (approved):

PR #13325, opened 2022-01-16
Fix Issue: When Editing a network it would auto delete whatever I type

ForestTigerKing:
Fixes: #13308 

Explanation:  
When editing a network, it would reset the form by the network configuration before modification.
The useRef variables didn't work, it causes resest the form frequently, you had to click save really fast.

Add isEditing flag, inited by false, when you edit the network, set it true, and when cancel or submit,set it false.
reset from only isEditing === false.  In this case, this problem can be solved and the refresh frequency can be greatly reduced

PR #14370, opened 2022-04-06
Fix for randomly resets of updated values in network edit form

### LINKED ISSUE #13168: Custom chain editing form randomly resets updated values
Repro steps:

1. Add a chain on https://chainlist.org/ like Optimistic Ethereum
2. Go to Settings > Networks > [the chain you just added]
3. Try editing the RPC to be https://localhost:8871.
4. Wait ~10 seconds
5. Notice that your updated settings have been reset.

I think on new block we're emitting a state update that is populating the form or something. We should not throw away current form values on an occasional state update.

amerkadicE:
## Explanation

Currently, in the network edit form, the field value is resetting randomly

This is a problem because the selectedNetwork prop is updated randomly which will cause calling the resetForm function and all field variables will reset.

In order to solve this problem, this pull request will add a new state **isEditing** which will not allow resetting of the form if the user changes something in the input field.

## More information

Fixes #13168

### After

image

## Manual testing steps

Add a chain on https://chainlist.org/ like Optimistic Ethereum
Go to Settings > Networks > [the chain you just added]
Try editing the RPC to be https://localhost:8871/.
Wait ~10 seconds
Notice that your updated settings didn't change.

Cancel, Delete, Save, and switching between networks should also be tested.


mirjanaKukic:
Verified by QA

Gtonizuka:
@amerkadicE I know this is still a draft but I think we could simplify the code quite a bit.

Look at https://github.com/MetaMask/metamask-extension/pull/13325/files
(not sure why TigerKing closed that PR but it is doing pretty much the same I think)

amerkadicE:
> @amerkadicE I know this is still a draft but I think we could simplify the code quite a bit.
> 
> Look at https://github.com/MetaMask/metamask-extension/pull/13325/files (not sure why TigerKing closed that PR but it is doing pretty much the same I think)
Hi @Gtonizuka, this PR will break the logic for switching between the networks. 
I updated my PR to look more similar to this one.

darkwing (dismissed):

ui/pages/settings/networks-tab/networks-form/networks-form.js
```diff
@@ -466,6 +473,7 @@ const NetworksForm = ({
   ]);
 
   const onSubmit = async () => {
+    setIsEditing(false);
```
digiwand, line 476: In case onSubmit ends up failing, should we move this after the updates have been confirmed?

darkwing:
I think what you have makes sense but I'm seeing some weird jank when I test it out:

https://user-images.githubusercontent.com/46655/173689246-389ae40a-c408-4578-b294-9c44c9270622.mp4

Could you investigate if we could avoid the message weirdness upon save?

darkwing (dismissed):
Noted the jank, hoping we can clean it up

amerkadicE:
> Noted the jank, hoping we can clean it up

It is fixed.

danjm (dismissed):

brad-decker:
@amerkadicE can you rebase this please and resolve conflicts?

amerkadicE:
@brad-decker Done

adonesky1 (approved):
LGTM

digiwand (approved):
LGTM2

PR #14372, opened 2022-04-06
Disable Swaps on Rinkeby for production

dan437:
## Explanation

We sometimes use Rinkeby to test the Swaps flow, but since the actual swapping doesn't happen on Rinkeby, we should only  do this in the dev env and disable Swaps on Rinkeby for production.


georgewrmarshall (dismissed):
Tested and swaps flow still works on rinkerby on local in dev and code looks good but please get a second review from someone familiar with the selectors

darkwing (dismissed):
Side note: Why isn't this just an array?

dan437:
> Side note: Why isn't this just an array?

Great suggestion, I've changed it to arrays.

georgewrmarshall (approved):

darkwing (approved):

PR #14367, opened 2022-04-06
MetaMetrics: Identify 'nft_autodetection_enabled' &  'opensea_api_enabled'

digiwand:
## Explanation

Add two new user traits to Segment:
- nft_autodetection_enabled - NFT Autodetection Enabled
- opensea_api_enabled - OpenSeas API Enabled

## More information

Progresses https://github.com/MetaMask/metamask-extension/issues/13477

PR #14377, opened 2022-04-06
User trait - number of nft collections

segun:
Add new user trait for 'Number of NFT collections'.

Explanation

Adds a new user trait for number of unique nft collections (nft addresses) in metamask state

More information

Progresses #13477

### DIFF
--- a/app/scripts/controllers/metametrics.js
+++ b/app/scripts/controllers/metametrics.js
@@ -591,20 +591,19 @@ export default class MetaMetricsController {
    */
   _getNumberOfNFtCollection(metamaskState) {
     const { allCollectibles } = metamaskState;
-    if (allCollectibles) {
-      const allAccounts = Object.keys(allCollectibles);
-      const allAddresses = allAccounts.flatMap((account) =>
-        Object.keys(allCollectibles[account]).flatMap((chainId) =>
-          allCollectibles[account][chainId].map(
-            (collectible) => collectible.address,
-          ),
-        ),
-      );
-      const unique = [...new Set(allAddresses)];
-      return unique.length;
+    if (!allCollectibles) {
+      return 0;
     }
-
-    return 0;
+    const allAccounts = Object.keys(allCollectibles);
+    const allAddresses = allAccounts.flatMap((account) =>
+      Object.keys(allCollectibles[account]).flatMap((chainId) =>
+        allCollectibles[account][chainId].map(
+          (collectible) => collectible.address,
+        ),
+      ),
+    );
+    const unique = [...new Set(allAddresses)];
+    return unique.length;
   }
 
   /**
--- a/app/scripts/controllers/metametrics.js
+++ b/app/scripts/controllers/metametrics.js
@@ -540,6 +540,9 @@ export default class MetaMetricsController {
         (rpc) => rpc.chainId,
       ),
       [TRAITS.THREE_BOX_ENABLED]: metamaskState.threeBoxSyncingAllowed,
+      [TRAITS.NUMBER_OF_NFT_COLLECTIONS]: this._getNumberOfNFtCollection(
+        metamaskState,
+      ),
     };
 
     if (!this.previousTraits) {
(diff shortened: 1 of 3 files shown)


app/scripts/controllers/metametrics.js
```diff
@@ -581,6 +584,29 @@ export default class MetaMetricsController {
...
   }
 
+  /**
+   *
+   * @param {*} metamaskState
+   * @returns number of unique collectible addresses
+   */
+  _getNumberOfNFtCollection(metamaskState) {
+    const { allCollectibles } = metamaskState;
+    if (allCollectibles) {
```
darkwing, line 594: Could we do the following instead?

```
if (!allCollectables) { return 0; }
```

app/scripts/controllers/metametrics.js
```diff
@@ -540,6 +540,9 @@ export default class MetaMetricsController {
         (rpc) => rpc.chainId,
       ),
       [TRAITS.THREE_BOX_ENABLED]: metamaskState.threeBoxSyncingAllowed,
+      [TRAITS.NUMBER_OF_NFT_COLLECTIONS]: this._getNumberOfNFtCollection(
+        metamaskState,
+      ),
```
digiwand, line 545: Should we only track `number_of_nft_collections` if `metamaskState.useCollectibleDetection` is enabled?

app/scripts/controllers/metametrics.js
```diff
@@ -581,6 +584,28 @@ export default class MetaMetricsController {
     }, {});
   }
 
+  /**
+   *
+   * @param {*} metamaskState
```
digiwand, line 589: nit:
```suggestion
   * @param {object} metamaskState
```

app/scripts/controllers/metametrics.js
```diff
@@ -581,6 +584,28 @@ export default class MetaMetricsController {
...
+   * @returns number of unique collectible addresses
+   */
+  _getNumberOfNFtCollection(metamaskState) {
```
digiwand, line 592: I imagine this function can be a good candidate for lodash's `memoize` method. Can we add this in to ensure we do our best to optimize performance for our tracking methods?

digiwand, line 592: I wonder if this function is a good candidate for lodash's `memoize` method.

shared/constants/metametrics.js
```diff
@@ -164,6 +164,8 @@
  * identify the new number_of_accounts trait
  * @property {string} [THREE_BOX_ENABLED] - when 3box feature is toggled we
  * identify the 3box_enabled trait
+ * @property {string} [NUMBER_OF_NFT_COLLECTIONS] - user trait for number of
```
digiwand, line 167:
```suggestion
 * @property {'number_of_nft_collections'} [NUMBER_OF_NFT_COLLECTIONS] - user trait for number of
```

New change. Replacing the type with the string literal as mentioned by Brad in this comment. 
https://github.com/MetaMask/metamask-extension/pull/14367#discussion_r845270295

app/scripts/controllers/metametrics.js
segun: You mean memoize the whole function or some parts of it?

app/scripts/controllers/metametrics.js
segun: I don't think so....collectibles can be added manually too.

brad-decker (approved):

app/scripts/controllers/metametrics.js
digiwand: The whole function in which we'd pass only `allCollectibles` to the function. If the input is the same, it would return the same computed value and not need to recompute the function.

digiwand (approved):

app/scripts/controllers/metametrics.js
digiwand: On second thought, we could use `memoize` to compute `allAddresses`. 

What I'm thinking is that `allAddresses.length` will give us the `number_of_nfts` based on the description in the issue (please double-check me here):

> All collectibles has this shape:

> This is an object whose keys are accounts and the values are objects where the keys are chainIds. So, for example, in order to get an aggregate number of NFTs imported you'll have to go through each chainId in each account in this object and sum the length of the Collectibles array.

> In order to get the number of collections you'll need to do the same sort of thing but you'll want the number of unique 'address' properties on the entries. Each collectible has an address property that indicates which collection it is a part of.

Then maybe we can share `allAddresses` for both `number_of_nfts` and `number_of_nft_collections`. Let me know if this makes sense. Always up for hopping on a call as well.

PR #14379, opened 2022-04-06
Dark Mode : Add OS option in theme selection dropdown

GuillaumeRx:
## Explanation

Adds an OS option in the theme selection dropdown.

## Screenshots/Screencaps

image

## Manual testing steps

- Go to experimental tab
- Switch to `OS`
- Theme should match your OS theme

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -804,9 +804,6 @@
   "decryptRequest": {
     "message": "Decrypt request"
   },
-  "defaultTheme": {
-    "message": "Default"
-  },
   "delete": {
     "message": "Delete"
   },
@@ -1713,6 +1710,9 @@
   },
   "levelArrow": {
     "message": "level arrow"
+  },
+  "lightTheme": {
+    "message": "Light"
   },
   "likeToImportTokens": {
     "message": "Would you like to import these tokens?"
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -2330,6 +2330,9 @@
   "origin": {
     "message": "Origin"
   },
+  "osTheme": {
+    "message": "OS"
+  },
   "padlock": {
     "message": "Padlock"
   },
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -2331,7 +2331,7 @@
     "message": "Origin"
   },
   "osTheme": {
-    "message": "OS"
+    "message": "Match system"
   },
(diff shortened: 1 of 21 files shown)


darkwing:
Should we change `Default` to `Light` if we're doing this?

GuillaumeRx:
> Should we change `Default` to `Light` if we're doing this?

Yes that's a good idea !

GuillaumeRx:
> I'm a bit hesitant to merge acronyms. Have we considered using "Automatic" or "Sync with System"?

`Match system` or any other would work yes :)

digiwand (dismissed):

app/scripts/controllers/preferences.js
```diff
@@ -68,7 +68,7 @@ export default class PreferencesController {
       ledgerTransportType: window.navigator.hid
         ? LEDGER_TRANSPORT_TYPES.WEBHID
         : LEDGER_TRANSPORT_TYPES.U2F,
-      theme: 'default',
+      theme: 'light',
```
darkwing, line 71: I think we'll want this to be `os`

darkwing (changes requested):
Just a tiny idea; @kevinghim do you agree that for new users we'll want to default to OS?

GuillaumeRx:
> Just a tiny idea; @kevinghim do you agree that for new users we'll want to default to OS?

After today's quick meeting we've agreed not to default to OS setting while in experimental

darkwing (dismissed):

app/_locales/en/messages.json
```diff
@@ -2330,6 +2330,9 @@
   "origin": {
     "message": "Origin"
   },
+  "osTheme": {
+    "message": "System Preference"
```
digiwand, line 2334: Is this copy something we have spoken to design about? 

Since I last commented, I observed another website using "Light", "Dark", and "System". I think "System" rather than "System Preference" could be a nice option here. It might pan out better with translations as well.

jpuri (dismissed):

digiwand (dismissed):
Excited for these full-system vibes! 🌒

No blockers. I left a minor inquiry in regards to the OS option name above.

app/_locales/en/messages.json
kevinghim: I think `System` should be intuitive enough, and be easily translatable.

digiwand (approved):

darkwing (approved):

PR #14380, opened 2022-04-06
Fix ESLint config around files w/ CommonJS imports

mcmire:
## Explanation

The `parserOptions.sourceType` config option in `.eslintrc.js` is used
to distinguish files that use ESM imports vs. files that do not.
The `import` plugin, specifically the `import/ambiguous` rule, behaves
differently depending on this value. If a file is marked with
`sourceType` of `"module"`, then this rule will attempt to ensure that
the file does indeed have ESM imports and/or exports; otherwise it does
nothing.

In other words, files that use CJS imports are *not* "modules" according
to this setting, and therefore should not be marked with a `sourceType`
of `"module"`. This means we do not have to turn off the
`import/ambiguous` rule, as it will no longer trip up on these types of
files.

## More information

* This change comes from a comment @Gudahtt left in another repo: https://github.com/MetaMask/eth-json-rpc-infura/pull/55#discussion_r844115513
* Also see the official ESLint docs which explain what `sourceType` means: https://eslint.org/docs/user-guide/configuring/language-options#specifying-parser-options

## Manual testing steps

* Run `yarn lint`. You shouldn't see any errors.


Gudahtt (approved):
LGTM!

darkwing (approved):

PR #14385, opened 2022-04-06
Updating info ActionableMessage type to default

NiranjanaBinoy:
- Updating ActionableMessage for the token detection message to use the `default` type instead of the old `info` type.
- Removing the `showSearchTab` prop from required as the non-supported networks ended up getting the value undefined and errors are thrown on the console.
- Updating the `showSearchTab` to show only on Ethereum Mainnet when TOKEN_DETECTION_V2 is disabled.


digiwand (dismissed):

brad-decker (dismissed):

digiwand (approved):
👍

brad-decker (approved):

PR #14386, opened 2022-04-07
E2e settings search msgs

seaona:
## Explanation
Improving error messages, so if the test fail, we know at which point it was.

## How
Included variables using template literals on the error message assertions.

### DIFF
--- a/test/e2e/tests/settings-search.spec.js
+++ b/test/e2e/tests/settings-search.spec.js
@@ -22,7 +22,7 @@ describe('Settings Search', function () {
     about: 'Terms of Use',
   };
 
-  var page;
+  let page;
 
   it('should find element inside the General tab', async function () {
     await withFixtures(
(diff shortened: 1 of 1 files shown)


jpuri (dismissed):

test/e2e/tests/settings-search.spec.js
```diff
@@ -22,6 +22,8 @@ describe('Settings Search', function () {
     about: 'Terms of Use',
   };
 
+  let page;
```
brad-decker, line 25: [Q] Non blocking, but why use reassignment on this variable versus just declaring it as a constant in each test case?

brad-decker (dismissed):

test/e2e/tests/settings-search.spec.js
seaona: that's a good point! I've updated it with `const` variables inside each testcase scope @brad-decker  ty

seaona:
@brad-decker @jpuri made a small update following brad's proposal. Could you double-check again? Ty

jpuri (approved):

digiwand (approved):

PR #14407, opened 2022-04-07
Update segment instantiation check. Only check if SEGMENT_WRITE_KEY exists

digiwand:
## Explanation

We can remove the `SEGMENT_HOST` check for our developer environments. If `SEGMENT_HOST` does not exist, then the `analytics-node` library will resolve the host to a default host. 

## Manual testing steps

1. Setup related environment variables
  a. Go to the `.metamaskrc` file in our repo
  b. Ensure `SEGMENT_WRITE_KEY` is set to developer, Segment write key
  c. Ensure `SEGMENT_HOST` is not set
2. Restart developer environment with `yarn start` 
3. Ensure you have no ad blockers running (uBlock Origin, AdBlock)
4. Open the MetaMask Extension
5. Ensure you are Participating in MetaMetrics in "Settings > Security & Privacy"
6. Open Segment and verify that events are being tracked


app/scripts/lib/segment.js
```diff
@@ -1,9 +1,6 @@
 import Analytics from 'analytics-node';
 import { SECOND } from '../../../shared/constants/time';
 
-const isDevEnvironment = Boolean(
-  process.env.METAMASK_DEBUG && !process.env.IN_TEST,
-);
```
brad-decker, line 6: @Gudahtt i could be mistaken, but we have a segment write key set in CIRCLE CI right? Is there any risk if we don't do the IN_TEST check that this hits segment?

brad-decker (changes requested):

app/scripts/lib/segment.js
Gudahtt: Ah, great point. We do indeed have this set in CircleCI, so that the metamaskbot builds send events to the dev segment instance.

Removing this check would send the test build events to the dev segment instance as well, which seems bad.

Gudahtt: Could be fixed by changing the condition below from `SEGMENT_WRITE_KEY` to `SEGMENT_WRITE_KEY && !process.env.IN_TEST`

digiwand: Sorry, trying to wrap my head around this more. Does this mean we want to send events for metamaskbot builds in CircleCL and not our test build? In which case, adding back
```
const isDevEnvironment = Boolean(
  process.env.METAMASK_DEBUG && !process.env.IN_TEST,
);
```
and checking `SEGMENT_WRITE_KEY && !isDevEnvironment` would be what we want? 

@brad-decker @Gudahtt

digiwand: I made the update in case the above is what we want. I can change it to `SEGMENT_WRITE_KEY && !process.env.IN_TEST` [as mentioned](https://github.com/MetaMask/metamask-extension/pull/14407#discussion_r846330536) if I understand what the env variables are doing some more

app/scripts/lib/segment.js
```diff
@@ -82,10 +83,10 @@ export const createSegmentMock = (flushAt = SEGMENT_FLUSH_AT) => {
 };
 
 export const segment =
-  !SEGMENT_WRITE_KEY || (isDevEnvironment && !SEGMENT_HOST)
-    ? createSegmentMock(SEGMENT_FLUSH_AT, SEGMENT_FLUSH_INTERVAL)
-    : new Analytics(SEGMENT_WRITE_KEY, {
+  SEGMENT_WRITE_KEY && !isDevEnvironment
```
brad-decker, line 86: i could be wrong here, but wouldn't this result in the mock being used in development mode? Which means now even with SEGMENT_HOST set events don't make it to segment. METAMASK_DEBUG is true in dev mode. I think we only want to check against IN_TEST

brad-decker (dismissed):

app/scripts/lib/segment.js
digiwand: @brad-decker 

Oh, I'm just now realizing that the METAMASK_DEBUG flag can be tied with dev mode and we want to negate this.

Now I see the purpose of `!SEGMENT_HOST` in the original code is a flag to allow segment in dev mode for the case you just mentioned.

So we're back at what @Gudahtt  mentioned in the above comment and what you are mentioning now - we could use
```SEGMENT_WRITE_KEY && !process.env.IN_TEST```

Gudahtt: Yeah, I think that makes sense.

Thinking about this in a different way.... what we want is to make `SEGMENT_WRITE_KEY` determine if Segment is used locally, but use something else for determining when Segment is used on CI, because `SEGMENT_WRITE_KEY` is always set on CI.

Right now there are two CI build types, prod and testing. Segment is enabled on prod but not on testing. So this preserves that behavior.

The one problem right now is that it's impossible to enable segment with a test build now. Maybe this doesn't matter, but it doesn't seem ideal.

One idea would be to use the condition `SEGMENT_WRITE_KEY && (process.env.METAMASK_DEBUG || !process.env.IN_TEST)` instead. i.e. `SEGMENT_WRITE_KEY &&  isProdOrDevMode`. That would let us use segment locally for a _dev_ test build, but still not for a non-dev test build. That would probably be good enough for all circumstances though.

Gudahtt: Oh darn, this breaks the metrics tests. We _do_ need to enable Segment on CI test builds. Just the host is set to localhost in that environment. I forgot about that. Hmmm....

Gudahtt: Right, so when I said "Currently this would send e2e metrics to Segment, which definitely isn't what we want.", I was wrong. That is what we want, because we're intercepting those requests during e2e test runs using the new e2e test mock server, so they will never reach Segment. Sorry about that!

So maybe the condition `SEGMENT_WRITE_KEY` is all we need. We want Segment to always be enabled on CI, because the e2e test mocks will intercept the requests anyway. And locally the presence of that key is enough.

app/scripts/lib/segment.js
digiwand: Thank you very much for explaining this, and sorry for the mixup here in the last couple of comments. Honestly, I reread this a few times to fully understand it, and I think I have it now. 

To confirm, we want to check `!IN_TEST` because this is related to the test builds, `Test dev build` and `Test build` that you are mentioning. 

From what I understand, another way to look at the other builds is:
- Dev
- Test
  - Dev
  - CI 
- Environment
  - Prod
  - Beta 
  - Prod - Flask
  - etc.
  
I've updated the code to implement your [original suggestion](https://github.com/MetaMask/metamask-extension/pull/14407#discussion_r846330536)

digiwand: 😅  oops. Okay, continuing this convo in [the other thread](https://github.com/MetaMask/metamask-extension/pull/14407#discussion_r850498894)

Gudahtt (approved):
LGTM!

jpuri (approved):

PR #14408, opened 2022-04-07
inject an interception method into the content-script stream piping t…

adonesky1:
Not sure why this got deleted...


adonesky1:
closing because it is superceded elsewhere

PR #14419, opened 2022-04-08
Add snaps view search

hmalik88:
Search for individual snaps is added with this PR. Route objects are created and merged with `SETTINGS_CONSTANTS` to create a list that includes snaps for the search list to display. No refs were used in this implementation as we're simply routing to the page itself, specific section search within a snap view page will be added in future iterations.

Testing:
- Run `yarn start --build-type flask` to create a build
- Load into chrome
- Install a snap (or more) via [snaplist](snaplist.org)
- In settings search, search for an installed snap using its name
- Observe changes

### DIFF
--- a/ui/helpers/utils/settings-search.js
+++ b/ui/helpers/utils/settings-search.js
@@ -15,9 +15,7 @@ export function getSettingsRoutes() {
 
 function getFilteredSettingsRoutes(t, tabMessage) {
   return getSettingsRoutes().filter((routeObject) => {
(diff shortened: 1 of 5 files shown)


darkwing (dismissed):
Code is good and this worked great!  Awesome work!

ritave (changes requested):

ritave (approved):

darkwing (approved):

PR #14425, opened 2022-04-12
Modify import SRP page

### LINKED ISSUE #14316: Modify Import SRP page in Onboarding V2 Flow
We need to modify the import SRP page in the (now long overdue) Onboarding V2 flow:

[Figma design](https://www.figma.com/file/8z4Csc3wlOIkUz2OP1eTZW/Secure-UX-Improvements?node-id=249%3A6062):

Screen Shot 2022-03-31 at 4 14 17 PM

This was done for the current onboarding flow in this [PR](https://github.com/MetaMask/metamask-extension/pull/14016).

VSaric:
## Explanation

Modified import SRP page to use new `SrpInput` component in the Onboarding V2 flow.

## More information

* Fixes #14316

### Before

https://user-images.githubusercontent.com/92527393/162947742-e9c33b7b-5830-4dba-a00d-d3c47efef50a.mov

### After

https://user-images.githubusercontent.com/92527393/162946717-075a1b86-21a4-47a3-8500-93307ac42ace.mov

### DIFF
--- a/ui/components/app/srp-input/srp-input.js
+++ b/ui/components/app/srp-input/srp-input.js
@@ -8,14 +8,17 @@ import ActionableMessage from '../../ui/actionable-message';
 import Dropdown from '../../ui/dropdown';
 import Typography from '../../ui/typography';
 import ShowHideToggle from '../../ui/show-hide-toggle';
-import { TYPOGRAPHY } from '../../../helpers/constants/design-system';
+import {
+  FONT_WEIGHT,
+  TYPOGRAPHY,
+} from '../../../helpers/constants/design-system';
 import { parseSecretRecoveryPhrase } from './parse-secret-recovery-phrase';
 
 const { isValidMnemonic } = ethers.utils;
 
 const defaultNumberOfWords = 12;
 
-export default function SrpInput({ onChange }) {
+export default function SrpInput({ onChange, srpText }) {
   const [srpError, setSrpError] = useState('');
   const [pasteFailed, setPasteFailed] = useState(false);
   const [draftSrp, setDraftSrp] = useState(
@@ -121,8 +124,8 @@ export default function SrpInput({ onChange }) {
   return (
     <div className="import-srp__container">
       <label className="import-srp__srp-label">
-        <Typography variant={TYPOGRAPHY.H4}>
-          {t('secretRecoveryPhrase')}
+        <Typography variant={TYPOGRAPHY.H4} fontWeight={FONT_WEIGHT.BOLD}>
+          {srpText}
         </Typography>
       </label>
       <ActionableMessage
@@ -232,4 +235,8 @@ SrpInput.propTypes = {
    * Otherwise, this is called with an empty string.
    */
   onChange: PropTypes.func.isRequired,
+  /**
+   * Text to show on the left of the Dropdown component. Wrapped in Typography component.
+   */
+  srpText: PropTypes.string,
 };
--- a/ui/components/app/srp-input/srp-input.js
+++ b/ui/components/app/srp-input/srp-input.js
@@ -238,5 +238,5 @@ SrpInput.propTypes = {
   /**
    * Text to show on the left of the Dropdown component. Wrapped in Typography component.
    */
-  srpText: PropTypes.string,
+  srpText: PropTypes.string.isRequired,
 };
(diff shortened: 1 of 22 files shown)


ui/components/app/srp-input/srp-input.js
```diff
@@ -232,4 +235,8 @@ SrpInput.propTypes = {
    * Otherwise, this is called with an empty string.
    */
   onChange: PropTypes.func.isRequired,
+  /**
+   * Text to show on the left of the Dropdown component. Wrapped in Typography component.
+   */
+  srpText: PropTypes.string,
```
Gudahtt, line 241: This should probably be a required prop. The render method doesn't seem to accommodate this being omitted very well - it still renders the `label` and `Typography` elements but with no contents. Plus I don't see a legitimate use case for omitting it.

Gudahtt: Making this required would also require adding this to each render call in the srp input unit tests

ui/pages/onboarding-flow/import-srp/import-srp.js
```diff
@@ -16,44 +14,58 @@ import {
...
+   * If we choose "I have 15-word phrase" in Dropdown component than,
+   * secretRecoveryPhrase.length = 14 and the button will be disabled
+   * If we choose "I have 18-word phrase" in Dropdown component than,
+   * secretRecoveryPhrase.length = 17 and the button will be disabled
+   * If we choose "I have 21-word phrase" in Dropdown component than,
+   * secretRecoveryPhrase.length = 20 and the button will be disabled
+   * If we choose "I have 24-word phrase" in Dropdown component than,
+   * secretRecoveryPhrase.length = 23 and the button will be disabled
+   */
+  const checkButtonVisibility = () => {
```
Gudahtt, line 41: Is this function needed? The `SrpInput` component should already guarantee that the SRP is a valid length. The change handler is only called with either an empty string, or a guaranteed valid SRP.

ui/components/app/srp-input/srp-input.js
VSaric: @Gudahtt Done [here](https://github.com/MetaMask/metamask-extension/pull/14425/commits/b08521f00fbcc0ca018296084d50fbfe951269fa)

ui/pages/onboarding-flow/import-srp/import-srp.js
VSaric: @Gudahtt This function is not needed, you are right. I checked again how to solve button visibility, when the button is disabled and enabled and I found better solution. [Here](https://github.com/MetaMask/metamask-extension/pull/14425/commits/b08521f00fbcc0ca018296084d50fbfe951269fa) are changes that I made for this button.

ui/pages/onboarding-flow/import-srp/index.scss
```diff
@@ -1,5 +1,21 @@
 .import-srp {
-  width: 625px;
+  width: 840px;
+
+  @media screen and (max-width: 349px) {
```
Gudahtt, line 4: What is the origin of these screen sizes? I thought we were using `767px`, `1270px` as our media query breakpoints, with `357px` being our minimum supported screen size.

ui/pages/onboarding-flow/import-srp/import-srp.js
```diff
@@ -68,45 +48,20 @@ export default function ImportSRP({ submitSecretRecoveryPhrase }) {
...
             type="primary"
             data-testid="import-srp-confirm"
             large
+            className="import-srp__confirm-button"
             onClick={() => {
               submitSecretRecoveryPhrase(secretRecoveryPhrase);
               history.replace(ONBOARDING_CREATE_PASSWORD_ROUTE);
             }}
-            disabled={error || secretRecoveryPhrase.length === 0}
+            disabled={!secretRecoveryPhrase.trim()}
```
Gudahtt, line 64: is the `trim()` necessary? This value should always be either an empty string or a valid SRP. It will never be just whitespace.

```suggestion
            disabled={secretRecoveryPhrase === ''}
```

Gudahtt (approved):
LGTM! Left two more comments but they are not blocking

adonesky1 (approved):
LGTM!

PR #14421, opened 2022-04-09
MetaMetrics: Identify 'number_of_tokens' user trait

digiwand:
## Explanation

Add two new user traits to Segment:

number_of_tokens - Number of Tokens

## More information

Progresses https://github.com/MetaMask/metamask-extension/issues/13477

PR #14427, opened 2022-04-12
MetaMetrics: Identify 'number_of_tokens' user trait

digiwand:
## Explanation

Add two new user traits to Segment:

number_of_tokens - Number of Tokens

**Note:** We want the count of all of the tokens across all of the chains as mentioned by @brad-decker  here: https://github.com/MetaMask/metamask-extension/pull/14421#discussion_r847346696

## More information

Progresses https://github.com/MetaMask/metamask-extension/issues/13477

### DIFF
--- a/app/scripts/controllers/metametrics.js
+++ b/app/scripts/controllers/metametrics.js
@@ -625,20 +625,12 @@ export default class MetaMetricsController {
    * @returns number of unique token addresses
    */
   _getNumberOfTokens(metamaskState) {
-    const allUniqueAddresses = Object.values(metamaskState.allTokens).reduce(
-      (result, tokensByAccountByChain) => {
-        Object.values(tokensByAccountByChain).forEach((tokensByAccount) => {
-          tokensByAccount.forEach((token) => {
-            result.add(token.address);
-          });
-        });
-
-        return result;
+    return Object.values(metamaskState.allTokens).reduce(
+      (result, accountsByChain) => {
+        return result + sum(Object.values(accountsByChain).map(size));
       },
-      new Set(),
+      0,
     );
-
-    return allUniqueAddresses.size;
   }
 
   /**
--- a/app/scripts/controllers/metametrics.js
+++ b/app/scripts/controllers/metametrics.js
@@ -554,10 +554,10 @@ export default class MetaMetricsController {
       [TRAITS.NFT_AUTODETECTION_ENABLED]: metamaskState.useCollectibleDetection,
       [TRAITS.NUMBER_OF_ACCOUNTS]: Object.values(metamaskState.identities)
         .length,
-      [TRAITS.NUMBER_OF_TOKENS]: this._getNumberOfTokens(metamaskState),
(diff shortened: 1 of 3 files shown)


app/scripts/controllers/metametrics.js
```diff
@@ -605,6 +606,27 @@ export default class MetaMetricsController {
...
+            result.add(token.address);
+          });
+        });
+
+        return result;
+      },
+      new Set(),
+    );
+
+    return allUniqueAddresses.size;
```
brad-decker, line 627: Its not necessary to do a unique check here, The addresses represent contracts, even if a token manages to have the same address on many different networks (possible, I think, but difficult) the user would have different balances on each chain effectively making them different tokens. It's also possible for two different tokens to have the same address on different chains.

brad-decker (changes requested):

app/scripts/controllers/metametrics.js
```diff
@@ -605,6 +606,27 @@ export default class MetaMetricsController {
...
+  }
```
digiwand, line 628: I think I got this right... "tokensByAccountByChain" is what I presumed from looking through the TokensController and the data produced. Please double-check me here 🙏 

I am also wondering two other things:
1. How large could the data set get, and would it be worth it to use lodash memoize? 
2. May we want to move this logic into the TokensController someday?

app/scripts/controllers/metametrics.js
digiwand: 😮 Thank you for the catch and for explaining this.  Maybe there is a guide I can read to help me better understand the structure of token contracts 

Updating the code to not do the unique filter here.

digiwand: @brad-decker updated here: https://github.com/MetaMask/metamask-extension/pull/14427/commits/0bd0772b8e3173b5593cf109e54377f703e993a1

brad-decker (dismissed):

digiwand:
There was a moment I forgot this branch was branching off https://github.com/MetaMask/metamask-extension/pull/14367. After I force pushed on https://github.com/MetaMask/metamask-extension/pull/14367, I had duplicate commits that I had to clean up. I force pushed a new version. There have been no changes in logic in the last couple of days; cc: @brad-decker

brad-decker (approved):

app/scripts/controllers/metametrics.js
brad-decker: 1. memoizing functions might be worth while, but that can happen down the road. We should discuss the proper level of memoization (should we memoize the entire build user trait function, only portions of the traits, etc)
2. No, although it seems like we will end up with a bunch of methods related to other parts of state, they are building user traits which is specific to meta metrics. I feel all user traits should be built and established inside the metametrics controller regardless of where in state they come from

digiwand: Thank you for the thoughts! 

1. 👍 
2. I thought, maybe, the total number of tokens could be shown in other places. Thanks for confirming that this method wouldn't be desirable in the TokenController. I am all for keeping the user trait logic inside the MetaMetrics controller

jpuri (approved):

PR #14436, opened 2022-04-13
Swaps: Sort "token_from" dropdown tokens by their fiat value first and "token_to" by top tokens

dan437:
## Explanation
Swap From - As a user, I want to easily access and select the tokens that I currently hold, so it makes sense to sort the tokens according to their fiat value.
Swap To - As a user, the odds that I will want to swap into a ‘popular’ token is higher, so if we sort tokens according to their popularity we can make it easier for users to select — and even discover — tokens.

## Manual testing steps
- Click on the "Swap" button on the main page
- Check that in the `Swap from` field tokens are sorted first by user owned tokens and their fiat value
- Check that in the `Swap to` field tokens are sorted by top assets first (https://swap.metaswap.codefi.network/networks/1/topAssets for Ethereum mainnet)


darkwing (approved):

danjm (approved):
LGTM

PR #14438, opened 2022-04-13
Add Sign-In with Ethereum

samgbafa:
## Explanation
[(EIP-4361) Sign-in with Ethereum or SIWE](https://eips.ethereum.org/EIPS/eip-4361) is a standard message format that wallets can parse and this PR implements this spec for Metamask. There are verification steps that can be taken to make sure the the SIWE message matches the specification and that is what this PR aims to do. 

It intercepts a signing message and detects if it's a SIWE message. If it is, it validates that the domain in the message comes from the site sending the request. If domain validation fails, it warns the user. If the message is not detected to be a SIWE message, the normal sign message flow is followed.

It is worth noting that this page does not have detection and communication regarding using a hardware wallet (however this does not impair any functionality, tested with a Ledger device).

New npm package added:
name: @spruceid/siwe-parser 
version: v^1.1.3
code: https://github.com/spruceid/siwe/tree/main/packages/siwe-parser

### Before
Screen Shot 2022-04-13 at 2 41 26 PM

### After
Screen Shot 2022-04-13 at 2 41 43 PM
Screen Shot 2022-04-13 at 2 42 01 PMScreen Shot 2022-04-13 at 2 42 13 PM

## Manual testing steps
1. Turn on the feature flag by adding `SIWE_V1=1` to `metamask-extension/.metamaskrc`
2. Build and run this version of the app.
3. Go to Metamask's [Test Dapp](https://metamask.github.io/test-dapp/) and scroll down to the Sign-In with Ethereum section. (If the dapp doesn't have this [change](https://github.com/MetaMask/test-dapp/pull/164), [use this version](https://spruceid.github.io/test-dapp/)). 
4. Try the 4 use cases by clicking the button for each.
	1. A normal SIWE message
	2. A SIWE message with optional parameters included
	3. A SIWE message with a different domain than the one sending the message (simulating a malicious use of SIWE)
	5. A malformed SIWE message that doesn't meet the spec (missing a required field)


skgbafa:

darkwing:
The EIP states `⚠️ This EIP is not recommended for general use or implementation as it is likely to change.`

ui/components/app/signature-request-siwe/index.scss
```diff
@@ -0,0 +1,140 @@
...
+
+    font-weight: 500;
+  }
+
+  p {
+    @include H6;
+
+    color: var(--color-text-muted);
+  }
+
```
georgewrmarshall, line 83: Did a search for this class name inside of the folder and doesn't look like any of these styles are being used?

Screen Shot 2022-04-26 at 5 59 53 PM

ui/components/app/signature-request-siwe/index.scss
```diff
@@ -0,0 +1,140 @@
...
+  height: 13px;
+  width: 13px;
+  font-size: 16px;
+  color: var(--color-error-default);
+  cursor: pointer;
+}
+
+.color-red {
+  color: var(--color-error-default) !important;
+}
```
georgewrmarshall, line 114: Better BEM and semantic naming

```suggestion

  &--error {
    color: var(--color-error-default) !important;
  }
}
```

Then update your component

Screen Shot 2022-04-26 at 6 04 29 PM

ui/components/app/signature-request-siwe/index.scss
```diff
@@ -0,0 +1,140 @@
...
+.signature-request-header {
+  flex: 1;
+
+  .network-display {
+    justify-content: flex-end;
+    margin-left: auto;
+  }
+}
+
+.signature-request-content {
```
georgewrmarshall, line 24: Not sure these styles are used at all? Did a search for the class name in the component folder and it only came back with this instance

Screen Shot 2022-04-26 at 5 59 53 PM

ui/components/app/signature-request-siwe/index.scss
```diff
@@ -0,0 +1,140 @@
...
+.signature-request {
+  display: flex;
+  flex: 1 1 auto;
+  flex-direction: column;
+  min-width: 0;
+
+  @media screen and (min-width: $break-large) {
+    flex: initial;
+  }
+}
```
georgewrmarshall, line 13: Looks like the only line of these styles that isn't being overwritten is `min-width: 0` you can probably remove them

Screen Shot 2022-04-26 at 6 07 24 PM

georgewrmarshall (changes requested):
Great PR description! Demoed the app and it works perfectly with the test dapp. UI looks great as well! 💯  Some things could be cleaned up under the hood. Have made a few suggestions. Reach out if you have any questions. Nice work!

ui/components/app/signature-request-siwe/signature-request-header/signature-request-header.component.js
```diff
@@ -0,0 +1,55 @@
...
+    </div>
+  );
+}
+
+SIWERequestHeader.propTypes = {
+  fromAccount: PropTypes.object,
+  domain: PropTypes.string,
+  isSIWEDomainValid: PropTypes.bool,
+  subjectMetadata: PropTypes.object,
+};
```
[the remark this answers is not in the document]
skgbafa: Added

ui/components/ui/site-origin/site-origin.js
```diff
@@ -19,4 +25,5 @@ SiteOrigin.propTypes = {
   siteOrigin: PropTypes.string.isRequired,
   iconName: PropTypes.string,
   iconSrc: PropTypes.string,
+  rightIcon: PropTypes.object,
```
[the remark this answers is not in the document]
skgbafa: Fixed

ui/components/app/signature-request-siwe/index.scss
skgbafa: This is leftover from my first implementation, will remove

ui/components/app/signature-request-siwe/signature-request-header/domain-header.scss
```diff
@@ -0,0 +1,421 @@
...
+  & &__custom-nonce-edit {
+    @include H7;
+
+    width: auto;
+  }
+}
+
+.confirm-approve-content--full {
+  height: auto;
+}
```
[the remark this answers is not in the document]
skgbafa: Removed!

ui/components/app/signature-request-siwe/index.scss
```diff
@@ -0,0 +1,140 @@
...
+
+.signature-request-header {
+  flex: 1;
+
+  .network-display {
+    justify-content: flex-end;
+    margin-left: auto;
+  }
+}
```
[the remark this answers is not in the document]
skgbafa: Removed

ui/components/app/signature-request-siwe/index.scss
skgbafa: Removed

ui/components/app/signature-request-siwe/index.scss
skgbafa: changed this to `signature-request-siwe` to keep the components orthogonal

skgbafa:
This PR is ready for review. There is not additional logic on this page for guiding users with a hardware wallet, but a hardware wallet was tested and works as expected.

georgewrmarshall (commented):
Hey Sam, this is looking really good. I started reviewing but decided to create a PR with all my suggestions as it might be quicker. You can find that PR here https://github.com/spruceid/metamask-extension-siwe/pull/1

georgewrmarshall (dismissed):
UI LGTM. Nice work! 💯 Please also get this reviewed from a contract interaction and redux state point of view

skgbafa:
> UI LGTM. Nice work! 💯 Please also get this reviewed from a contract interaction and redux state point of view

Regarding contract interaction: There is no contract interaction with SIWE, it is an extension of the EIP-191 signing functionality. This can be confirmed by a reviewer!

segun (dismissed):
LGTM

app/_locales/en/messages.json
```diff
@@ -44,6 +44,24 @@
   "QRHardwareWalletSteps2Description": {
     "message": "Ngrave (Coming Soon)"
   },
+  "SIWEDomainInvalid": {
+    "message": "The website you are attempting to sign-in to ($1) does not match the domain in the sign-in request. Proceed with caution."
```
digiwand, line 48:
```suggestion
    "message": "The website you are attempting to sign in to ($1) does not match the domain in the sign-in request. Proceed with caution."
```

nit: I believe we want to use "sign in" instead of "sign-in" in the first instance here to be consistent with the `SIWEWarningBody` message and because we are using it as a verb here. ref: https://english.stackexchange.com/questions/49850/sign-in-signin-or-sign-in

FrederikBolding:
@skgbafa Looks like this PR needs to rebased as well before it can be merged as it has conflicts.

ui/helpers/utils/format-message-params.js
```diff
@@ -0,0 +1,105 @@
...
+      issuedAt,
+      expirationTime,
+      notBefore,
+      requestId,
+      resources,
+    } = parsedMessage;
+
+    if (statement) {
+      output.push({
+        label: 'Message:',
```
digiwand, line 27: May we add translation support for these labels please?

ui/components/app/signature-request-siwe/signature-request-siwe.js
```diff
@@ -0,0 +1,150 @@
+import React, { useContext, useState } from 'react';
+import PropTypes from 'prop-types';
+import convertMsg from '../../../helpers/utils/format-message-params';
```
digiwand, line 3: I recommend avoiding synonyms. We should keep naming consistent. Can change `convertMsg` to `formatMessageParams` to match the filename?

ui/helpers/utils/format-message-params.js
```diff
@@ -0,0 +1,105 @@
...
+        label: `Resources: ${resources.length}`,
+        value: resources
+          .reduce((previous, resource) => `${previous}${resource}\n`, '')
+          .trim(),
+      });
+    }
+
+    return output;
+  } catch (error) {
+    return [];
```
digiwand, line 101: Q: In what case could this catch statement run? I think we might not need the `try/catch` blocks

ui/components/app/signature-request-siwe/signature-request-siwe-message/index.scss
```diff
@@ -0,0 +1,35 @@
+.signature-request-siwe-message {
+  flex: 1 100%;
+  padding-bottom: 8px;
```
digiwand, line 3:
```suggestion
```

This should be overridden by `padding: 8px 16px;` below

digiwand (changes requested):
Hello 👋  I have a couple of requests and I plan to check this branch out to run the code as well. The main change requests are:
- move, remove, or rename `ui/helpers/utils/format-message-params.js`
   - if this util is kept, we should add a test file for this
- add i18n translation support 

I'm available to hop on a call if it helps with these requests too

seaona:
@skgbafa great feature! from the QA side, I did some validations on the message fields and have a couple of questions (most of them are conceptual):

- on the `Chain ID`, is it okay that the user is on a different network than the indicated Chain ID on the message? should we perform any validation on our side?

image

- on the `Issued At` and `Not Before`, should we perform any validation on our side? or this should come on the 3rd party dapp side?

 image

- on the malformed Message `"current-domain" wants you to sign in with your Ethereum account:
0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2` is there a reason why are we grabbing the address value passed by the Dapp, instead of taking the current selected Account on Metamask? 

image

ui/components/app/signature-request-siwe/signature-request-siwe.js
```diff
@@ -0,0 +1,150 @@
...
+  const t = useContext(I18nContext);
+  const trackEvent = useContext(MetaMetricsContext);
+
+  const [isShowingDomainWarning, setIsShowingDomainWarning] = useState(false);
+  const [agreeToDomainWarning, setAgreeToDomainWarning] = useState(false);
+
+  const onSign = (event) => {
+    sign(event);
+    trackEvent({
+      category: 'Transactions',
```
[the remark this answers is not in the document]
skgbafa: Addressed!

ui/components/app/signature-request-siwe/signature-request-siwe.js
```diff
@@ -0,0 +1,150 @@
...
+  };
+
+  const onCancel = (event) => {
+    cancel(event);
+    trackEvent({
+      category: 'Transactions',
+      event: 'Cancel',
+      properties: {
+        action: 'SIWE Request',
+        legacy_event: true,
```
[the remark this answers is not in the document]
skgbafa: Addressed!

ui/helpers/utils/format-message-params.js
```diff
@@ -0,0 +1,105 @@
+/**
+ * Takes in aparsed Sign-In with Ethereum Message (EIP-4361)
```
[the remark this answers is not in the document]
skgbafa: Makes sense to move. Moved it.

ui/components/app/signature-request-siwe/signature-request-siwe.js
skgbafa: Done!

ui/helpers/utils/format-message-params.js
skgbafa: After considering this, I think any errors that could cause issues are handled elsewhere. Removed try/catch block.

skgbafa:
Hey MM team! I've made the changes requested and rebased it against `develop`.

@digiwand This PR is ready for review

@seaona The issues you brought up were related to the test-dapp, which are hardcoded siwe request to force the MM UI into certain states for testing. I've updated that repo based on your feedback. Also some more notes, related to your questions:
- Per the spec, only the SIWE message formatting and domain are validated in SIWE requests. Other validation is left to 3rd party receiving the signed message.
- The Chain ID is hardcoded in the test-dapp SIWE request instead of taken from the user. This can be updated in the test-dapp.
- Validation this and other fields takes place with 3rd parties.
- The address in SIWE messages is passed by the dapp, but should be grabbed from the active wallet. Updated the test-dapp to do this.

digiwand:
In case anyone else was wondering if we were supporting this feature following [DarkWing's comment above](https://github.com/MetaMask/metamask-extension/pull/14438#issuecomment-1101536365)
> The EIP states ⚠️ This EIP is not recommended for general use or implementation as it is likely to change.

we have an [internal Slack thread](https://consensys.slack.com/archives/GTQAGKY5V/p1651604302489839) running to support this feature.

----

Q: Do we have a design spec for this? I am thinking for a sign-in page, it would be handy to include the abbreviated copy of the address
Screenshot 2022-05-26 at 4 35 43 PM

ui/components/app/signature-request-siwe/signature-request-siwe-header/signature-request-siwe-header.js
```diff
@@ -0,0 +1,71 @@
...
+        iconName={subjectMetadata.name}
+        headerTitle={t('SIWESiteRequestTitle')}
+        headerText={t('SIWESiteRequestSubtitle')}
+        siteOrigin={domain}
+        className={isSIWEDomainValid ? '' : 'bad-domain'}
+        rightIcon={
+          !isSIWEDomainValid && (
+            <Tooltip
+              position="bottom"
+              html={<p>{t('SIWEWarningBody', [domain])}</p>}
```
digiwand, line 32: `SIWEWarningBody` is missing from the translations file  
Screenshot 2022-05-26 at 4 33 42 PM

ui/components/app/signature-request-siwe/signature-request-siwe.container.js
```diff
@@ -0,0 +1,41 @@
+import { connect } from 'react-redux';
+import { accountsWithSendEtherInfoSelector } from '../../../selectors';
+import { getAccountByAddress } from '../../../helpers/utils/util';
+import SignatureRequest from './signature-request-siwe';
+
+function mapStateToProps(state) {
+  const {
+    metamask: { subjectMetadata = {} },
```
digiwand, line 8: There's no need for this container file since we are using functional components. We can access redux using useSelector. We can also shave off some complexities by working with the passed in params, `ownProps`, directly in the component file itself. 

After deleting this container file, we can do something like 

```
import { useSelector } from 'react-redux';
import {
  accountsWithSendEtherInfoSelector,
  getSubjectMetadata,
} from '../../../selectors';
import { getAccountByAddress } from '../../../helpers/utils/util';

export default function SignatureRequestSIWE({
  txData,
  cancelPersonalMessage,
  signPersonalMessage,
}) {
  // ...

  const allAccounts = useSelector(accountsWithSendEtherInfoSelector);
  const subjectMetadata = useSelector(getSubjectMetadata);

  const {
    msgParams: {
      from,
      origin,
      siwe: { isSIWEDomainValid, isMatchingAddress, messageData },
      version,
    },
    type,
  } = txData;

  const fromAccount = getAccountByAddress(allAccounts, from);
  const targetSubjectMetadata = subjectMetadata[origin];
```

ui/components/app/signature-request-siwe/signature-request-siwe.js
```diff
@@ -0,0 +1,185 @@
...
+        messageData.domain,
+      ])} `;
+    }
+
+    if (!isMatchingAddress) {
+      showError = true;
+      errorMessage += `${t('SIWEAddressInvalid', [
+        messageData.address,
+        fromAccount.address,
+      ])} `;
```
digiwand, line 77: Double-checking, is this a pattern we have in our application to combine error messages? In this case, we are also mentioning `Proceed with caution` twice to the user. See attachment below
cc: @georgewrmarshall 

Screenshot 2022-05-26 at 5 22 09 PM

ui/components/app/signature-request-siwe/signature-request-siwe.js
```diff
@@ -0,0 +1,185 @@
...
+    };
+  };
+
+  const {
+    showError,
+    errorMessage,
+    confirmErrorMessage,
+  } = generateSIWEWarning();
+
+  console.log(showError, errorMessage, confirmErrorMessage);
```
digiwand, line 97:
```suggestion
```

ui/components/app/signature-request-siwe/signature-request-siwe.js
```diff
@@ -0,0 +1,185 @@
...
+              onSubmit={onSign}
+              submitText={t('confirm')}
+              submitButtonType="danger-primary"
+              disabled={!agreeToDomainWarning}
+            />
+          }
+        >
+          <div className="signature-request-siwe__popover__checkbox-wrapper">
+            <Checkbox
+              id="domainWarning_checkbox"
```
digiwand, line 142:
```suggestion
              id="signature-request-siwe_domain-checkbox"
```
To avoid name collisions, I am suggesting that we include the name of the fileName in the id name. Reminder that will need to update `htmlFor` below with the name as well

digiwand (changes requested):
Awesome! It's looking like we're almost there. Left some requests and comments. Please feel free to let me know if you have any questions or if you'd like me to elaborate on any thoughts

ui/components/app/signature-request-siwe/index.scss
```diff
@@ -0,0 +1,67 @@
...
+    flex: 0 0 auto;
+    border-radius: 8px;
+    box-shadow: 0 0 7px 0 rgba(0, 0, 0, 0.08);
+  }
+
+  &__domain-mismatch-warning {
+    padding: 4px 16px;
+  }
+
+  &__page-container-footer {
```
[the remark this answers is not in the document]
georgewrmarshall: This might have been my bad! Had a momentary lapse of judgement on CSS.. that's embarrassing 🤦

Apologies @skgbafa!

And thanks so much for the catch and the resources @digiwand 🙏 💯

ui/components/app/signature-request-siwe/signature-request-siwe.js
georgewrmarshall: Great catch! Hmm I have not seen his before. I think it should be probably stack 2 alerts on top of each other. I'm also aware that this might push the buttons further down the page. But I think it's the optimal move here. That error message doesn't read well...

ui/components/app/signature-request-siwe/index.scss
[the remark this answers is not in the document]
digiwand: All good! 🙏  Honestly, it works fine and it's an existing pattern in our codebase. I think in an app like this, it's like to not cause problems. Thing is, I have run into `!important` takeover battles in the past and I think it's good practice to choose specificity over `!important` when applicable. 

[This StackOverflow answer](https://stackoverflow.com/a/3706876/4053142) has a good mention of when it's okay to use `!important` as well

I appreciate you receiving the feedback/suggestion and getting on board with this too! Thank you! 😊 🤜🏼🤛🏼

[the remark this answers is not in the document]
skgbafa: Addressed! I left in a single `!important` as I was unable to prevent a line from showing any other way. I think the high degree of specificity will prevent collisions.
Screen Shot 2022-06-07 at 6 24 39 PM
Screen Shot 2022-06-07 at 6 25 58 PM

ui/components/app/signature-request-siwe/signature-request-siwe.js
skgbafa: Addressed

ui/components/app/signature-request-siwe/signature-request-siwe.js
skgbafa: Addressed.

ui/components/app/signature-request-siwe/signature-request-siwe.js
skgbafa: Removed.

ui/components/app/signature-request-siwe/signature-request-siwe.container.js
skgbafa: Addressed!

ui/components/app/signature-request-siwe/index.scss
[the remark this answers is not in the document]
skgbafa: Addressed and removed all `!important`. Thank you very much for the assistance @digiwand 😄

ui/components/app/signature-request-siwe/signature-request-siwe-header/signature-request-siwe-header.js
skgbafa: Renamed, but didn't update everywhere! Updated.

skgbafa:
Updated (and rebased) PR based on feedback. Ready for review!

ui/components/app/signature-request-siwe/signature-request-siwe-message/signature-request-siwe-message.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const SignatureRequestSIWEMessage = ({ data }) => {
+  return (
+    <div className="signature-request-siwe-message">
+      <div className="signature-request-siwe-message__list">
+        {data.map(({ label, value }, i) => (
+          <div className="signature-request-siwe-message__list__item" key={i}>
```
digiwand, line 9: We should ensure the key is a string value
ref: https://reactjs.org/docs/lists-and-keys.html#keys

digiwand (commented):
Thanks for the updates! I added a bit more comments. Feel free to ping me for any questions 🙏

app/scripts/lib/siwe/siwe-validation.js
```diff
@@ -0,0 +1,43 @@
+import { stripHexPrefix } from 'ethereumjs-util';
+import { ParsedMessage } from '@spruceid/siwe-parser';
+
+const msgHexToText = (hex) => {
+  try {
+    const stripped = stripHexPrefix(hex);
+    const buff = Buffer.from(stripped, 'hex');
+    return buff.length === 32 ? hex : buff.toString('utf8');
+  } catch (e) {
+    return hex;
```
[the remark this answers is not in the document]
skgbafa: Done!

app/scripts/lib/siwe/siwe-validation.js
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    return {
+      isSIWEMessage: true,
+      isSIWEDomainValid,
+      isMatchingAddress,
+      messageData,
+    };
+  } catch (error) {
+    // ignore error, it's not a valid SIWE message
```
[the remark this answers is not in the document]
skgbafa: I don't think so, as any normal sign message will throw an error.

app/_locales/en/messages.json
```diff
@@ -44,6 +44,57 @@
...
+    "message": "Sign-in request"
+  },
+  "SIWEWarningSubtitle": {
+    "message": "To confirm you understand, check:"
+  },
+  "SIWEWarningTitle": {
+    "message": "Are you sure?"
+  },
+  "SIWELabelMessage": {
+    "message": "Message:"
```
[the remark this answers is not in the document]
skgbafa: This was updated to have colons after an internal review of the design and opting for consistency in colon usage.

ui/components/app/signature-request-siwe/signature-request-siwe-message/signature-request-siwe-message.js
skgbafa: Done

ui/components/app/signature-request-siwe/signature-request-siwe-header/index.scss
```diff
@@ -0,0 +1,39 @@
...
+
+  &__account {
+    flex: 1;
+    display: flex;
+    align-items: center;
+    margin-top: 8px;
+
+    &__item.account-list-item {
+      &__top-row {
+        display: flex;
```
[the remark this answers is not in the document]
skgbafa: Done

ui/components/app/signature-request-siwe/signature-request-siwe-header/signature-request-siwe-header.js
```diff
@@ -0,0 +1,71 @@
...
+              html={<p>{t('SIWEDomainWarningBody', [domain])}</p>}
+              wrapperClassName="signature-request-siwe-header__tooltip"
+              containerClassName="signature-request-siwe-header__tooltip__container"
+            >
+              <InfoIcon severity={SEVERITIES.DANGER} />
+            </Tooltip>
+          )
+        }
+      />
+      <div className="signature-request-siwe-header__account">
```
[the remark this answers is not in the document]
skgbafa: Updated

ui/components/app/signature-request-siwe/signature-request-siwe-header/index.scss
```diff
@@ -0,0 +1,39 @@
...
+
+    &__container {
+      display: flex !important;
+    }
+  }
+
+  &__account {
+    flex: 1;
+    display: flex;
+    align-items: center;
```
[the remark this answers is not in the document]
skgbafa: I think I got it! Made multiple changes to this section that I think addresses this

ui/components/app/signature-request-siwe/signature-request-siwe-header/index.scss
```diff
@@ -0,0 +1,39 @@
...
+    margin-top: 8px;
+
+    &__item.account-list-item {
```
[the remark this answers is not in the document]
skgbafa: Got it, fixed the selectors and there are visible improvements as the styles are being applied correctly. Thanks for the guidance and feedback!

skgbafa:
@digiwand Incorporated your feedback and am ready for your review

digiwand (commented):
Thanks for making the updates! Updates look good to me. almost forgot... I think we should add tests for the new logic, new components, and script updates.

I'd also like to revisit these two messages since the alignment and size of the icons are inconsistent here 
Screenshot 2022-06-10 at 9 52 42 PM

app/scripts/lib/siwe/siwe-validation.js
```diff
@@ -0,0 +1,45 @@
...
+    const stripped = stripHexPrefix(hex);
+    const buff = Buffer.from(stripped, 'hex');
+    return buff.length === 32 ? hex : buff.toString('utf8');
+  } catch (e) {
+    log.error(e);
+    return hex;
+  }
+};
+
+const detectSIWE = (msgParams) => {
```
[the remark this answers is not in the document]
skgbafa: This makes sense, the filename was leftover from early testing. Updated

app/scripts/lib/siwe/siwe-validation.js
```diff
@@ -0,0 +1,45 @@
...
+  try {
+    const stripped = stripHexPrefix(hex);
+    const buff = Buffer.from(stripped, 'hex');
+    return buff.length === 32 ? hex : buff.toString('utf8');
+  } catch (e) {
+    log.error(e);
+    return hex;
+  }
+};
+
```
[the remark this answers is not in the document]
skgbafa: Added!

app/scripts/lib/siwe/siwe-validation.js
```diff
@@ -0,0 +1,45 @@
...
+    if (origin) {
+      const { host } = new URL(origin);
+      isSIWEDomainValid = messageData.domain === host;
+    }
+
+    return {
+      isSIWEMessage: true,
+      isSIWEDomainValid,
+      isMatchingAddress,
```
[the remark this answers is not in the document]
skgbafa: Done!

ui/components/app/signature-request-siwe/format-message-params.js
```diff
@@ -0,0 +1,102 @@
...
+
+  // not in design
+  if (notBefore) {
+    output.push({
+      label: t('SIWELabelNotBefore'),
+      value: notBefore,
+    });
+  }
+
+  // not in design
```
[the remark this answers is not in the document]
skgbafa: Left in accidentally! Great catch. Removed!

ui/components/app/signature-request-siwe/signature-request-siwe-message/signature-request-siwe-message.stories.js
```diff
@@ -0,0 +1,30 @@
...
+  title: 'Components/App/SignatureRequestSIWE/SignatureRequestMessage',
+  id: __filename,
+  argTypes: {
+    data: {
+      controls: 'object',
+    },
+  },
+  args: {
+    data: [
+      { label: 'Label', value: 'value' },
```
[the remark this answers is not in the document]
skgbafa: Updated!

ui/components/app/signature-request-siwe/format-message-params.js
```diff
@@ -0,0 +1,102 @@
+/**
+ * Takes in a parsed Sign-In with Ethereum Message (EIP-4361)
+ * and generates an array of label-value pairs
+ *
+ * @param {Object} parsedMessage - A parsed SIWE message with message contents
+ * @param {Function} t - i18n function
+ * @returns {Array} An array of label-value pairs with the type of the value as the label
+ */
+export const formatParams = (parsedMessage, t) => {
```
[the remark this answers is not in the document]
skgbafa: Done!

digiwand (commented):
Great updates and good job getting rid of the `important!` styles! I noted you kept one `important!` style to overwrite an inline style which is a valid use case for the `important!` style.

re: tests:
I spoke with @adonesky1  earlier and he mentioned that there are tests for the test-dapp that were added to this, so I think we can add additional tests in a follow-up PR if not this one.

ui/components/app/signature-request-siwe/index.scss
```diff
@@ -0,0 +1,77 @@
...
+
+.signature-request-siwe {
+  display: flex;
+  flex-direction: column;
+  flex: 1;
+  height: 100%;
+  width: 100%;
+  min-width: 0;
+  background-color: var(--color-background-default);
+  overflow-y: auto;
```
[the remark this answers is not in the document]
skgbafa: Noted! Updated the styles

ui/components/app/signature-request-siwe/signature-request-siwe-header/index.scss
```diff
@@ -0,0 +1,33 @@
...
+  align-items: center;
+  flex-direction: column;
+
+  &__tooltip {
+    display: flex;
+
+    &__container {
+      display: flex !important;
+    }
+  }
```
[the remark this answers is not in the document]
skgbafa: Done!

ui/components/app/signature-request-siwe/signature-request-siwe-header/index.scss
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    &__account-name {
+      @include H5;
+
+      font-weight: 500;
+    }
+
+    &__top-row {
+      margin: 0;
```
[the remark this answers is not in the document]
skgbafa: Cleaned up!

app/_locales/en/messages.json
```diff
@@ -44,6 +44,57 @@
   "QRHardwareWalletSteps2Description": {
     "message": "Ngrave (Coming Soon)"
   },
+  "SIWEAddressInvalid": {
+    "message": "The address in the sign-in request does not match the address of the account you are using to sign in."
+  },
+  "SIWEDomainInvalid": {
+    "message": "The website you are attempting to sign in to ($1) does not match the domain in the sign-in request. Proceed with caution."
```
darkwing, line 51:
```suggestion
    "message": "The website you are attempting to sign in to ($1) does not match the domain in the sign-in request. Proceed with caution.",
    "description": "$1 represents the website domain"
```

app/_locales/en/messages.json
```diff
@@ -44,6 +44,57 @@
...
+  },
+  "SIWEDomainWarningBody": {
+    "message": "The website ($1) is asking you to sign in to the wrong domain. This may be a phishing attack."
```
darkwing, line 54:
```suggestion
    "message": "The website ($1) is asking you to sign in to the wrong domain. This may be a phishing attack.",
    "description": "$1 represents the website domain"
```

ui/components/app/signature-request-siwe/index.scss
```diff
@@ -0,0 +1,80 @@
...
+    display: flex;
+    flex-direction: row;
+    align-items: flex-start;
+    padding: 8px 16px 24px;
+
+    &__label {
+      @include H7;
+
+      color: var(--color-text-default);
+      margin-left: 8px;
```
darkwing, line 68: `margin-inline-start` may be better for RTL support.

ui/pages/confirm-transaction/conf-tx.js
```diff
@@ -123,6 +129,10 @@ class ConfirmTxScreen extends Component {
       return SignatureRequest;
     }
 
+    if (siwe && siwe.isSIWEMessage) {
```
darkwing, line 132:
```suggestion
    if (siwe?.isSIWEMessage) {
```

darkwing (dismissed):
This is getting super close!

app/_locales/en/messages.json
skgbafa: Updated!

ui/components/app/signature-request-siwe/index.scss
skgbafa: Updated

ui/pages/confirm-transaction/conf-tx.js
skgbafa: Updated

ui/components/app/signature-request-siwe/signature-request-siwe.js
```diff
@@ -0,0 +1,180 @@
...
+            parsedMessage.address,
+            fromAccount.address,
+          ])}
+          iconFillColor="var(--color-warning-default)"
+          useIcon
+          withRightButton
+        />
+      )}
+      {!isSIWEDomainValid && (
+        <ActionableMessage
```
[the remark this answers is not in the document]
skgbafa: No problem! Thanks for cleaning that up and keeping the PR up to date with repo changes

.metamaskrc.dist
```diff
@@ -7,5 +7,8 @@ SWAPS_USE_DEV_APIS=
 COLLECTIBLES_V1=
 TOKEN_DETECTION_V2=
 
+; Set this to true to enable support for SIWE [EIP-4361](https://eips.ethereum.org/EIPS/eip-4361)
+SIWE_VALIDATION=
```
digiwand, line 11: It looks like our other features prepend temporary feature flags with VX. Can we update this to something like `SIWE_V1`? I am reading the versioning as an indication to remove these flags later.

app/scripts/lib/siwe/detectSIWE.js
```diff
@@ -0,0 +1,52 @@
...
+    log.error(e);
+    return hex;
+  }
+};
+
+/**
+ * A locally defined object used to provide data to identify a Sign-In With Ethereum (SIWE)(EIP-4361) message and provide the parsed message
+ *
+ * @typedef localSIWEObject
+ * @param {bool} isSIWEMessage - Does the intercepted message conform to the SIWE specification?
```
digiwand, line 20:
```suggestion
 * @param {boolean} isSIWEMessage - Does the intercepted message conform to the SIWE specification?
```

ui/components/app/signature-request-siwe/signature-request-siwe.js
```diff
@@ -0,0 +1,180 @@
...
+    return isSIWEDomainValid;
+  };
+
+  const isSIWEDomainValid = checkSIWEDomain();
+
+  const [isShowingDomainWarning, setIsShowingDomainWarning] = useState(false);
+  const [agreeToDomainWarning, setAgreeToDomainWarning] = useState(false);
+  const onSign = (event) => {
+    signPersonalMessage(event);
+    trackEvent({
```
digiwand, line 62: Is the following thread proposing that we replace these trackEvents? 
https://consensys.slack.com/archives/C031PSFHQER/p1654811895007879?thread_ts=1654805714.261689&cid=C031PSFHQER

I think we can add a try/catch to handle `signPersonalMessage` and `cancelPersonalMessage`. These methods should be returning promises, and we'll want to await for the promises and confirm they have succeeded before tracking the events.

digiwand: Non-blocking. updating MetaMetrics in a future PR. For now, we updated the code to handle errors and include a @ todo comment
https://github.com/MetaMask/metamask-extension/pull/14438/commits/7b37892103297663236fb2aa869c5337525ad881

digiwand (dismissed):
LGTM! Great work putting this together! & Thanks for letting me hop on board to throw in a couple of minor updates. 

Now we just need to get coverage and e2e tests working

digiwand:
Verified on Chrome x macOS with generic MM account, and Ledger Nano X

seaona:
checking e2e tests

app/_locales/en/messages.json
```diff
@@ -44,6 +44,59 @@
...
+    "message": "Nonce:"
+  },
+  "SIWELabelNotBefore": {
+    "message": "Not Before:"
+  },
+  "SIWELabelRequestID": {
+    "message": "Request ID:"
+  },
+  "SIWELabelResources": {
+    "message": "Resources: $1"
```
darkwing, line 80: Apologies for not catching this earlier -- we need a `description` since we're using `$1` here.

seaona:
@skgbafa I could fix the e2e tests in my local by running `yarn lavamoat:auto `:) It was a problem with lavamoat policies not being updated. 
The issue is that I don't seem to have access to your branch and cannot push the changes (maybe I am doing something wrong?).
If you want, you can either run yourself `yarn lavamoat:auto` and push the changes (there will be changes on the policies files) or you can check if you could give me access and I can push the changes. Whichever you prefer.
Ty!!

FYI @digiwand

georgewrmarshall (commented):
UI in storybook look good to me! I couldn't simulate the bad actors ui by using the [test dApp](https://metamask.github.io/test-dapp/) is there an extra step or some configuration that I'm missing? I have the feature flag turned on and made sure to rebuild.

This is how the UI looks for me

### Bad domain
Screen Shot 2022-07-04 at 11 23 48 AM
Screen Shot 2022-07-04 at 11 22 04 AM

### Bad account
Screen Shot 2022-07-04 at 11 24 07 AM
Screen Shot 2022-07-04 at 11 24 32 AM

### Malformed

Screen Shot 2022-07-04 at 11 25 09 AM
Screen Shot 2022-07-04 at 11 25 04 AM

app/_locales/en/messages.json
skgbafa: Done!

skgbafa:
> @skgbafa I could fix the e2e tests in my local by running `yarn lavamoat:auto `:) It was a problem with lavamoat policies not being updated. The issue is that I don't seem to have access to your branch and cannot push the changes (maybe I am doing something wrong?). If you want, you can either run yourself `yarn lavamoat:auto` and push the changes (there will be changes on the policies files) or you can check if you could give me access and I can push the changes. Whichever you prefer. Ty!!
> 
> FYI @digiwand

I've added you as an editor to the project! I am also running `yarn lavamoat:auto` and will push those changes up. In case that doesn't work, you'll be able to access. I've also given @digiwand the rights to add collaborators on my local repo.

skgbafa:
> UI in storybook look good to me! I couldn't simulate the bad actors ui by using the [test dApp](https://metamask.github.io/test-dapp/) is there an extra step or some configuration that I'm missing? I have the feature flag turned on and made sure to rebuild.
> 
> This is how the UI looks for me

@georgewrmarshall Do you have this value set in your `.metamaskrc`? This should turn the feature flag on!
```
; Set this to '1' to enable support for Sign-In with Ethereum [EIP-4361](https://eips.ethereum.org/EIPS/eip-4361)
SIWE_V1=1
```

georgewrmarshall (dismissed):
Ah yup my mistake see the correct UI now. UI LGTM

digiwand (dismissed):
👍 
One last thing was adding bullet points to the resources list ([internal Slack](https://consensys.slack.com/archives/C031PSFHQER/p1655852059209099)). We can add this in a future PR

digiwand:
note: Sign-in With Ethereum reuses the `personal_sign` RPC method name. This PR introduces the `@spruceid/siwe-parser` npm package to differentiate Sign-in With Ethereum messages from the original `personal_sign` messages.

ui/components/app/signature-request-siwe/format-message-params.js
```diff
@@ -0,0 +1,100 @@
+/**
+ * Takes in a parsed Sign-In with Ethereum Message (EIP-4361)
+ * and generates an array of label-value pairs
+ *
+ * @param {Object} parsedMessage - A parsed SIWE message with message contents
+ * @param {Function} t - i18n function
+ * @returns {Array} An array of label-value pairs with the type of the value as the label
+ */
+export const formatMessageParams = (parsedMessage, t) => {
```
brad-decker, line 9: [REQ] This file throws me off just a tad because I expect mostly react components in the components folder. I do believe we have some dated examples of having helper method files inside component folders but we should try and prevent extending this pattern. I think it would be helpful to create a SIWE file in shared/modules at the root of the repository that contains this method, and potentially the contents of scripts/lib/siwe/detecteSIWE.js. I know in the case of the later its only used in the backend but I think bundling these things together into one SIWE module file would be beneficial

ui/components/app/signature-request-siwe/index.scss
```diff
@@ -0,0 +1,74 @@
+@import 'signature-request-siwe-header/index';
+@import 'signature-request-siwe-message/index';
+
+.signature-request-siwe {
+  display: flex;
```
brad-decker, line 5: [REQ] I know that you largely just followed the other confirmations patterns but I think it would be good to get this confirmation into shape with new patterns such that we aren't furthering our debt in refactoring / updating these.

We introduced the Box component that is one of the basic building blocks of our design component system. A lot of the styles related to positioning and flex properties can be achieved using those components without adding additional css.

ui/components/app/signature-request-siwe/signature-request-siwe-header/index.scss
```diff
@@ -0,0 +1,26 @@
...
+    &__top-row {
+      align-items: center;
+      margin: 0;
+    }
+
+    &__account-name {
+      @include H5;
+
+      font-weight: 500;
+    }
```
brad-decker, line 24: [REQ] For any style where you are reaching to add class names only to style the text to one of our Typography levels (H5, here) then you can use the <Typography> component instead of writing additional styles.

ui/components/app/signature-request-siwe/signature-request-siwe-message/index.scss
```diff
@@ -0,0 +1,34 @@
...
+
+    &__item {
+      flex: 1;
+      flex-direction: column;
+
+      &__label {
+        @include H4;
+
+        margin: 8px auto;
+      }
```
brad-decker, line 22: the <Typography> component also has a `boxProps` prop that can be used for margin, etc.

ui/components/app/signature-request-siwe/signature-request-siwe-message/index.scss
```diff
@@ -0,0 +1,34 @@
...
+
+      &__value {
+        @include H6;
+
+        margin: 8px auto;
+        white-space: pre-line;
+        overflow: hidden;
+        word-wrap: break-word;
```
brad-decker, line 30: I believe white-space, overflow, and word-wrap are not cared for by our design system, so you'll need the custom css here, but you can still use the Typograpy component and remove the duplicated styles such that as the design system is extended it becomes increasingly easier to remove duplicated styles.

ui/components/app/signature-request-siwe/signature-request-siwe.js
```diff
@@ -0,0 +1,206 @@
...
+  const [agreeToDomainWarning, setAgreeToDomainWarning] = useState(false);
+
+  /**
+   * @todo update MetaMetrics for onSign and onCancel
+   * Currently, we're tracking the events similarily to how we track the events with the SignatureRequest component
+   * in which the action names differ. We want to revisit this to add properties mentioned in the related, internal Slack thread.
+   * We would also want to ensure that we are tracking the event appropriatedly should an error be thrown.
+   * @see {@link ui/components/app/signature-request/signature-request.component.js}
+   * @see {@link https://consensys.slack.com/archives/C031PSFHQER/p1654811895007879?thread_ts=1654805714.261689&cid=C031PSFHQER}
+   */
```
brad-decker, line 69: In my opinion not tracking metrics is better than tracking them based on the outdated schema. We can revisit later. I have a pretty good idea on how we can track these in the future.

brad-decker (changes requested):
Thank you for this contribution. I have a few items I think we should address to get this more inline with emerging patterns and the design system. I talked with @digiwand at length about metrics and have a plan for adding these items after the SIWE work lands.

ui/components/app/signature-request-siwe/signature-request-siwe.js
digiwand: @brad-decker heads up, SIWE reuses the same RPC method call name as `personal_sign` making our current tracking solution in createRPCMethodTrackingMiddleware.js unable to support SIWE out of the box. 

I'm okay with removing the current tracking metrics in this PR. It can almost always, if not always, be a good idea to separate metric logic into its own PR. I can also see the benefit of keeping these trackEvents in while we reprioritize and implement an updated version of the metrics here.

cc: @kevinghim

brad-decker: we can add it using the detect method that is being added here, it will be trivial. I already scoped this out when I was re-doing #15082

ui/components/app/signature-request-siwe/format-message-params.js
skgbafa: This was originally in [`ui/helpers/utils/format-message-params.js`](https://github.com/MetaMask/metamask-extension/pull/14438#discussion_r881110964). I could update this to exist in `shared/modules`

ui/components/app/signature-request-siwe/signature-request-siwe.js
```diff
@@ -0,0 +1,206 @@
...
+        trackEvent({
+          category: EVENT.CATEGORIES.TRANSACTIONS,
+          event: 'Confirm',
+          properties: {
+            action: 'SIWE Request',
+            type,
+            version,
+          },
+        });
+      }
```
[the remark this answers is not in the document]
skgbafa: Removed all references to `trackEvent` in signature-request-siwe

ui/components/app/signature-request-siwe/format-message-params.js
skgbafa: Moved SIWE modules

ui/components/app/signature-request-siwe/signature-request-siwe.js
skgbafa: Removed all references to `trackEvent` in signature-request-siwe. We can add this back in via the improved metrics flow

ui/components/app/signature-request-siwe/index.scss
skgbafa: Because this new page reuses existing components, the css changes are made to make these existing components match the design, instead of touching existing components and having to test the changes across the repo. For all new components, I'll rebuild them with the box components and make the accompanying css changes!

ui/components/app/signature-request-siwe/signature-request-siwe-header/index.scss
skgbafa: This reuses existing components, so I'll leave the changes in as css

ui/components/app/signature-request-siwe/signature-request-siwe-message/index.scss
skgbafa: Updated

ui/components/app/signature-request-siwe/signature-request-siwe-message/index.scss
skgbafa: Updated

ui/components/app/signature-request-siwe/signature-request-siwe-message/signature-request-siwe-message.js
```diff
@@ -0,0 +1,42 @@
...
+const SignatureRequestSIWEMessage = ({ data }) => {
+  return (
+    <div className="signature-request-siwe-message">
+      <div className="signature-request-siwe-message__list">
+        {data.map(({ label, value }, i) => (
+          <div
+            className="signature-request-siwe-message__list__item"
+            key={i.toString()}
+          >
+            <div className="signature-request-siwe-message__list__item__label">
```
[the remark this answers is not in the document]
skgbafa: Updated

brad-decker (dismissed):

brad-decker:
@skgbafa need to run lint and what not locally and push those changes up but this is looking good. thanks for doing the additional work.

ui/components/app/signature-request-siwe/format-message-params.js
digiwand: I like the new file created in `shared/modules` to hold all SIWE utils. My issue is that this `formatMessageParams` logic is specific to one component file and does not look like it will be reused. I think this logic can be moved directly into the component that uses this. 

This is a non-blocker

cc: @brad-decker @skgbafa

brad-decker (dismissed):

ui/components/app/signature-request-siwe/signature-request-siwe-message/index.scss
```diff
@@ -0,0 +1,13 @@
+.signature-request-siwe-message {
+  flex: 1 100%;
+  border-radius: 8px;
+  padding: 8px 16px;
+  margin: 16px;
+  border: 1px solid var(--color-border-muted);
+
+  .sub-text {
```
digiwand, line 8: can we update this to `.signature-request-siwe-message_sub-text`? The reason is to avoid name collision. Best to stay away from generic names unless we're writing utility classes or we're working with them through third-party libraries

ui/components/app/signature-request-siwe/signature-request-siwe.js
```diff
@@ -0,0 +1,182 @@
...
+  };
+
+  const isSIWEDomainValid = checkSIWEDomain();
+
+  const [isShowingDomainWarning, setIsShowingDomainWarning] = useState(false);
+  const [agreeToDomainWarning, setAgreeToDomainWarning] = useState(false);
+
+  /**
+   * @todo update MetaMetrics for onSign and onCancel
+   * Currently, we're tracking the events similarily to how we track the events with the SignatureRequest component
```
digiwand, line 60: This comment is outdated now

digiwand (commented):
Nice. I left two small comments. I'm up for approving this PR after these 2 minor comments are addressed

ui/components/app/signature-request-siwe/signature-request-siwe-message/index.scss
skgbafa: Updated

digiwand (approved):
🎉

brad-decker (approved):

seaona:
Oops was doing last QA and just saw it's merged. Just had a couple of small last comments.

- I saw that info message is cut on the left side (or there is no margin) when hovering on the Bad Domain:

https://user-images.githubusercontent.com/54408225/182645940-a4d90521-7384-4c00-80c9-05890c9c3ed1.mp4

- What is this circle representing? Just to confirm that's the expected design or if we are expecting to parse something here (token icon?) @georgewrmarshall 

image

georgewrmarshall:
@seaona I'm assuming it should be the site icon / favicon of the site requesting the signature but to be honest I'm not 100% sure. I'll investigate further.

georgewrmarshall:
OK looks like the `requesterAddress` is returning null which is getting passed to the `<Identicon diameter={40} address={requesterAddress} />` component that generates the blockies or jazzicon graphics for an ethereum address. Not sure if this was the intention @skgbafa?

https://github.com/MetaMask/metamask-extension/blob/develop/ui/components/app/signature-request-original/signature-request-original.component.js#L109-L117

holantonela:
hi team, i've been reviewing the SIWE implementation, amazing work @skgbafa 

To-do (discussed with Sam on Aug 24th)

- [ ] Implement security indicators for the origin site component (warning only) - [Figma](https://www.figma.com/file/llV58Rh6Y3ZBPwTUkCTwfu/Sign-in-with-Ethereum?node-id=703%3A1530)
- Replace site icon with a white warning icon, red background, add [unsafe] pill, make main CTA red, update warning copy to `Deceptive site request. The site you're attempting to sign into doesn't match the domain in the request. Proceed with caution.` 

- [ ] Fix footer CTA since it gets hidden under the scroll - [Figma](https://www.figma.com/file/llV58Rh6Y3ZBPwTUkCTwfu/Sign-in-with-Ethereum?node-id=703%3A1528)
- Add :arrow_down: component, make footer fixed. I also included a mockup of [how it should look](https://www.figma.com/file/llV58Rh6Y3ZBPwTUkCTwfu/Sign-in-with-Ethereum?node-id=703%3A1525) if a warning is triggered.

- [ ] Unify sign-in vs Sign-In vs Sign-in usage
- Must use `Sign in` everywhere (no hyphen). Update title, copy and buttons.

PR #14441, opened 2022-04-14
Event tracking for Token Detection V2

### LINKED ISSUE #13951: Event tracking for Token Detection V2
Adding event tracking on improved token detection.

https://www.notion.so/896022ba921347be8a7058d0641b6d00?v=224f52511d34479695d770a61b20d914

**Tasks**
- [ ] [Token Autodetection](https://www.notion.so/Token-Autodetection-7735db4678c248d8bd80ef21ef4c1c0b)

- [ ] [Token Added](https://www.notion.so/Token-Added-7c29eb0a7dbe49bea75b9a863b18b039)

- [ ] [Token Not Added](https://www.notion.so/Token-Not-Added-7dfbc0f511494993aff3553e728bb6dd)

NiranjanaBinoy:
Fixes #13951 

- Adding a new trait for token detection enabled.
- Adding a new `Token Detected` event when the token is detected at the backend.
- Adding `Token Import Clicked` event to track when the detected tokens modal link is clicked.
- Adding  `Token Import Canceled` event when the detected tokens modal link is closed without performing import or ignore.
- Adding `Token Added` event when detected tokens are imported.
- Adding `Token Hidden` event when detected token are ignored.


app/scripts/controllers/detect-tokens.js
```diff
@@ -51,6 +56,7 @@ export default class DetectTokensController {
     this.detectedTokens = process.env.TOKEN_DETECTION_V2
       ? this.tokensController?.state.detectedTokens
       : [];
+    this._trackMetaMetricsEvent = trackMetaMetricsEvent;
```
jpuri, line 59: May be we can just directly use `trackMetaMetricsEvent` rather than adding `this._trackMetaMetricsEvent`.

ui/components/app/asset-list/asset-list.js
```diff
@@ -96,7 +99,19 @@ const AssetList = ({ onClickAsset }) => {
...
+                trackEvent({
+                  event: 'Clicked "Import Tokens"',
+                  category: EVENT.CATEGORIES.WALLET,
+                  properties: {
+                    token_quantity: detectedTokens.length,
+                    tokens: detectedTokensDetails,
+                  },
+                });
+              }}
+            />
```
jpuri, line 114: Function above looks much bigger now, it will be useful to extract it out I think and not have it inline, also will be nice to have test coverage.

app/scripts/controllers/detect-tokens.js
```diff
@@ -172,6 +178,20 @@ export default class DetectTokensController {
...
             } = tokenList[nonZeroTokenAddress];
+
+            this._trackMetaMetricsEvent({
+              event: 'Token Detected',
+              category: EVENT.CATEGORIES.WALLET,
+              sensitiveProperties: {
+                token_symbol: symbol,
+                token_contract_address: address,
+                token_decimal_precision: decimals,
+                source: 'detected',
```
brad-decker, line 189: We can omit the source property on this one because it will never be a different value

shared/constants/metametrics.js
```diff
@@ -178,6 +178,8 @@
  * @property {'three_box_enabled'} THREE_BOX_ENABLED - when 3box feature is
  *  toggled we identify the 3box_enabled trait
  * @property {'theme'} THEME - when the user's theme changes we identify the theme trait
+ * @property {'token_detection_enabled'} [TOKEN_DETECTION_ENABLED] - when token detection feature is toggled we
+ * identify the token_detection_enabled trait
```
brad-decker, line 182: There is another jsdoc type that needs to be updated in this file that describes the shape of the traits object versus this one that describes the shape of the TRAITS constant object

ui/components/app/detected-token/detected-token-selection-popover/detected-token-selection-popover.js
```diff
@@ -18,11 +20,23 @@ const DetectedTokenSelectionPopover = ({
...
 
   const detectedTokens = useSelector(getDetectedTokensInCurrentNetwork);
+  const detectedTokensDetails = detectedTokens.map(
+    ({ address, symbol }) => `${symbol} - ${address}`,
+  );
 
   const onClose = () => {
     setShowDetectedTokens(false);
+    trackEvent({
+      event: '"Import Tokens" cancelled',
```
brad-decker, line 33: I'm not sold on using quotations around the Import Tokens here or in the Clicked one. can you confirm the source of this so I can explore?

brad-decker (changes requested):
Overall very good work, just a few comments/questions

app/scripts/controllers/detect-tokens.js
```diff
@@ -172,6 +178,20 @@ export default class DetectTokensController {
               iconUrl,
               aggregators,
             } = tokenList[nonZeroTokenAddress];
+
+            this._trackMetaMetricsEvent({
+              event: 'Token Detected',
```
digiwand, line 183: Q: Should we be adding these event names to `EVENT_NAMES` in `shared/constants/metametrics.js?

cc: @brad-decker

ui/components/app/detected-token/detected-token-selection-popover/detected-token-selection-popover.js
NiranjanaBinoy: That was a mistake on my part; but I do have a small query, the popover we are opening and canceling on click are actually `Tokens found`  not the `Import Token` page; so should we have the event to something like `Clicked Tokens found` and `Cancelled Tokens found`?

digiwand: To avoid interchanging synonyms, I suggest incorporating "detect" over "found" here. I think sometimes we can get away with this if it makes sense to present synonyms to the user. Maybe we could update these to "Clicked Detected Tokens" and "Canceled Detect Tokens". 

In regards to the quotes in the event name, we do have some existing examples with this pattern:

- `/ui/components/app/import-token-link/import-token-link.component.js`
  - 'Clicked "Add Token"'
- `/ui/pages/swaps/fee-card/fee-card.js`:
  - 'Clicked "Gas Fees: Learn More" Link'
  
I'm personally okay with the quotes. I could see how it can be easier to read with quotes. It could cause some adjustable complications if we changed the copy, but I also don't imagine the copy would be changed often.

digiwand: New information found: 

John [shared a doc](https://docs.google.com/document/d/1WQ3WzRGamAAtO8g3JtQnztjxV8B-EiVK4DcjkDUY1co/edit#heading=h.qnasv64mcgjm) in [our metrics Slack channel](https://consensys.slack.com/archives/G01FXS64DFA/p1651014166856429),  which proposes either the object-action or context-object-action naming convention. We have some instances in our app that already use this convention e.g. 'Token Added'. 

More information on the object-action naming convention here: 
[ Segment naming convention for clean data article](https://segment.com/academy/collecting-data/naming-conventions-for-clean-data/) 

I have updated the following suggestions in my comment above this with this convention.

app/scripts/controllers/detect-tokens.js
brad-decker: @digiwand -- That's a good idea and perhaps a practice we should start doing now even though it isn't widely used yet.

ui/components/app/detected-token/detected-token-selection-popover/detected-token-selection-popover.js
brad-decker: The two prior examples containing quotations are from our legacy schema, so let's avoid those. 

I want to avoid using the word "Detect" in this situation so it doesn't get confused with Token Detected. 

The user is shown a phrase, "X tokens detected, click here to import them" or something to that effect.

Maybe we should get John to weigh in on this. I"ll start a thread in metametrics channel

app/scripts/controllers/detect-tokens.js
NiranjanaBinoy: While we are on the topic of adding constants, shall I go ahead and add constants for `source` property values? Currently, we have 8 values assigned for the `source` property.

app/scripts/controllers/detect-tokens.js
brad-decker: @jpuri , the method is passed in from the constructor and is invoked outside of the constructor. The only way to make this method available to class methods is by attaching it to the class instance. I think, and could be wrong, that the collapsed code in github made it appear to you (as it did me) that the method was invoked in the constructor, but once you expand the code or pull it locally you'll see it is not invoked in the constructor.

brad-decker (dismissed):

digiwand (changes requested):
This is looking great! I had a couple of suggestions and one possibly blocking event tracking in `handleClearTokensSelection` unless I'm mistaken.

ui/components/app/detected-token/detected-token.js
```diff
@@ -43,12 +48,40 @@ const DetectedToken = ({ setShowDetectedTokens }) => {
...
     const {
-      selected: selectedTokens,
-      deselected: deSelectedTokens,
+      selected: selectedTokens = [],
+      deselected: deSelectedTokens = [],
     } = sortingBasedOnTokenSelection(tokensListDetected);
 
+    selectedTokens.forEach((importedToken) => {
+      trackEvent({
+        event: EVENT_NAMES.TOKEN_ADDED,
```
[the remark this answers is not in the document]
NiranjanaBinoy: Inside `onImport`, we are checking for two scenarios, one where all the tokens are selected to be imported and another where some are deselected. If some of the tokens are deselected, we will call a modal where users have to confirm that they are okay with ignoring the deselected tokens. Upon clicking confirm button in the modal, `handleClearTokensSelection` will be called, where we make a separate list of selected and deselected tokens. If there is no selected tokens `ignoreToken()` will be dispatched and only the TOKEN_HIDDEN event will be fired, if is a mix of selected and deselected tokens, the both `ignoreTokens` and `importTokens` will be dispatched and both TOKEN_ADDED and TOKEN_HIDDEN events will be fired.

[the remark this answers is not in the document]
digiwand: I see. Thank you for the explanation! 🙏  Should we also add the `EVENT_NAMES.TOKEN_ADDED` to the second scenario of `onImport` where `await dispatch(importTokens(selectedTokens));` is called?

ui/components/app/detected-token/detected-token.js
```diff
@@ -82,7 +68,22 @@ const DetectedToken = ({ setShowDetectedTokens }) => {
         },
       });
     }
+
     if (deSelectedTokens.length < detectedTokens.length) {
+      selectedTokens.forEach((importedToken) => {
```
digiwand, line 73: NIT: I think we can improve the readability of this code by extracting the add token logic into its own function

```
      // const selectedTokens = ... 
      
      const tokenSymbols = selectedTokens.map(({ symbol }) => symbol);
      await dispatch(importTokens(selectedTokens));
      dispatch(setNewTokensImported(tokenSymbols.join(', ')));

      // trackEvent logic
```

ui/components/app/detected-token/detected-token.js
```diff
@@ -43,13 +48,42 @@ const DetectedToken = ({ setShowDetectedTokens }) => {
...
+            token_symbol: importedToken.symbol,
+            token_contract_address: importedToken.address,
+            token_decimal_precision: importedToken.decimals,
+            source: EVENT.SOURCE.TOKEN.DETECTED,
+            token_standard: TOKEN_STANDARDS.ERC20,
+            asset_type: ASSET_TYPES.TOKEN,
+          },
+        });
+      });
       await dispatch(ignoreTokens(deSelectedTokens));
```
digiwand, line 87: NIT: kinda out of the scope of this PR, but I think we could remove the `else` statement and move `await dispatch(ignoreTokens(deSelectedTokens));` underneath

ui/components/app/detected-token/detected-token-selection-popover/detected-token-selection-popover.js
```diff
@@ -28,9 +34,20 @@ const DetectedTokenSelectionPopover = ({
     selectedTokens.length === detectedTokens.length
       ? `All`
       : `(${selectedTokens.length})`;
+  const detectedTokensDetails = detectedTokens.map(
```
digiwand, line 37: NIT: we can move this const into the `onClose` function since it's not used outside of the function. This will also prevent `detectedTokens.map` from being called until it's requested

ui/components/app/detected-token/detected-token.js
```diff
@@ -43,13 +48,42 @@ const DetectedToken = ({ setShowDetectedTokens }) => {
...
+          tokens: tokensDetailsList,
+          location: EVENT.LOCATION.TOKEN_DETECTION,
+          token_standard: TOKEN_STANDARDS.ERC20,
+          asset_type: ASSET_TYPES.TOKEN,
+        },
+      });
+    }
+
     if (deSelectedTokens.length < detectedTokens.length) {
+      selectedTokens.forEach((importedToken) => {
```
digiwand, line 73: NIT: I think we can improve the readability of this code by extracting the add token logic into its own function

```
      // const selectedTokens = ... 
      
      const tokenSymbols = selectedTokens.map(({ symbol }) => symbol);
      await dispatch(importTokens(selectedTokens));
      dispatch(setNewTokensImported(tokenSymbols.join(', ')));

      // trackEvent logic
```

ui/components/app/detected-token/detected-token.js
digiwand, line 87: NIT: kinda out of the scope of this PR, but I think we could remove the `else` statement and move `await dispatch(ignoreTokens(deSelectedTokens));` underneath

digiwand (dismissed):
No blockers, only some optional suggestions added.  Thank you for explaining parts of the code to me! :)

digiwand (approved):
🙌

darkwing (approved):

app/scripts/controllers/detect-tokens.js
```diff
@@ -172,6 +179,9 @@ export default class DetectTokensController {
               iconUrl,
               aggregators,
             } = tokenList[nonZeroTokenAddress];
+
+            eventTokensDetails.push(`${symbol} - ${address}`);
```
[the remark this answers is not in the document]
darkwing: Ignore this -- we talked about it.

PR #14443, opened 2022-04-14
GetLocale API method

Gtonizuka:
Implements a new API endpoint to get wallet locale

## Screenshots/Screencaps

Screenshot 2022-04-25 at 13 44 46

## Manual testing steps

Tested on flask trying to get current locale and then changed it to a different language. (see screenshot)


shanejonas:
I think we should remove the restricted part for now at least.

app/scripts/lib/rpc-method-middleware/handlers/index.js
```diff
@@ -16,5 +17,6 @@ const handlers = [
   sendMetadata,
   switchEthereumChain,
   watchAsset,
+  wallet_getLanguage
```
shanejonas, line 20: if we want it to only be included in flask we should wrap this line with the:

```
```

app/scripts/lib/rpc-method-middleware/handlers/index.js
```diff
@@ -16,5 +19,8 @@ const handlers = [
   sendMetadata,
   switchEthereumChain,
   watchAsset,
+  ///: BEGIN:ONLY_INCLUDE_IN(flask)
+  walletgetLocale,
```
FrederikBolding, line 23: Perhaps this should be cased like ``walletGetLocale`?

shared/constants/permissions.js
```diff
@@ -5,6 +5,7 @@ export const CaveatTypes = Object.freeze({
 export const RestrictedMethods = Object.freeze({
   eth_accounts: 'eth_accounts',
   ///: BEGIN:ONLY_INCLUDE_IN(flask)
+  wallet_getLocale: 'wallet_getLocale',
```
FrederikBolding, line 8: If it is restricted it needs to be added to https://github.com/MetaMask/metamask-extension/blob/8b9b9af13fabcef4882d4423e35acb4cd3aef4e9/ui/helpers/utils/permission.js

app/scripts/lib/rpc-method-middleware/handlers/get-locale.js
```diff
@@ -0,0 +1,56 @@
...
+  try {
+    const walletLocale = await getWalletLocale();
+
+    sendMetrics({
+      event: 'Get locale',
+      category: 'Permission',
+      referrer: {
+        url: origin,
+      },
+    });
```
brad-decker, line 49: [REQ] Unless there are specific props etc we want to track in RPC method handlers, the createRPCMethodTrackingMiddleware file sets up tracking for most RPC methods. If its important to know everytime this method is called (versus a sampling), we would add it to the USER_PROMPTED_EVENT_NAME_MAP in that file ... i don't believe it is necessary to capture every event in this case but I could be wrong.

brad-decker (dismissed):

Gtonizuka:
Moving this back into draft as we need more info from @rekmarks

app/scripts/lib/rpc-method-middleware/handlers/get-locale.js
```diff
@@ -0,0 +1,57 @@
...
+
+
+async function getLocaleHandler(
+  req,
+  res,
+  _next,
+  end,
+  { getWalletLocale, hasPermission },
+) {
+    if (hasPermission(MESSAGE_TYPE.GET_LOCALE)) {
```
Gtonizuka, line 42: Can you clarify why is this needed @rekmarks ?

app/scripts/controllers/permissions/specifications.js
```diff
@@ -162,6 +162,21 @@ export const getPermissionSpecifications = ({
         }
       },
     },
+    [PermissionKeys['wallet_getLocale']]: {
+      permissionType: PermissionType.RestrictedMethod,
```
Gtonizuka, line 166: If `permissionType` is `restrictedMethod` why do we need `if(hasPermission...` in the rpc method implementation (i.e. handlers/get-locale.js below) @rekmarks

app/scripts/metamask-controller.js
```diff
@@ -593,6 +593,7 @@ export default class MetamaskController extends EventEmitter {
           getAllAccounts: this.keyringController.getAccounts.bind(
             this.keyringController,
           ),
+          getWalletLocale: this.getWalletLocale.bind(this),
```
BelfordZ, line 596: Should this be wrapped in `BEGIN:ONLY_INCLUDE_IN(flask)` ?

BelfordZ:
functionally it seems to work great. Nice work. Just left a q about the flask inclusion stuff, but otherwise lgtm!

brad-decker:
@BelfordZ @Gtonizuka do we still want this? If so can we get a rebase and another round of reviews? If not lets close it (for PR queue cleanliness)

brad-decker:
@Gtonizuka I am going to close this, if it ends up being something we want we should reopen and rebase it. Thanks!

PR #14454, opened 2022-04-15
Add testnet name as prefix for native currency of preloaded testnets

### LINKED ISSUE #14452: Add prefix to ticker symbols for native currency on testnets
Currently we show the ticker symbol for native currencies on the Ethereum testnets (Rinkeby, Ropsten, Goerli, Kovan) that we preload as `ETH`. 

This both causes confusion for users who are unfamiliar with test networks, and opens a window for scammers to trick unwitting users into trading assets with real value for worthless testnet ETH. 

We should indicate that testnet ETH is not interchangeable with mainnet ETH by prefixing the ticker symbol with the name of the testnet.

adonesky1:
Fixes: #14452

~Not sure if we would prefer a different format for these ticker symbols? cc @holantonela @rachelcope @SayaGT~

~As @kevinghim [pointed out](https://consensys.slack.com/archives/G015BV486HG/p1650035398081129?~ ~thread_ts=1649949932.722959&cid=G015BV486HG), simply prefixing all of these with the first letter of the testnet~ ~name (e.g. `RETH` for rinkeby and ropsten ETH or `KETH` for kovan eth) could also be confusing.~

Examples:
Screen Shot 2022-04-19 at 1 30 10 PM

Screen Shot 2022-04-21 at 9 31 17 AM

Screen Shot 2022-04-21 at 9 31 33 AM

### DIFF
--- a/app/scripts/controllers/network/network.js
+++ b/app/scripts/controllers/network/network.js
@@ -18,6 +18,7 @@ import {
   MAINNET_CHAIN_ID,
   RINKEBY_CHAIN_ID,
   INFURA_BLOCKED_KEY,
+  TEST_NETWORK_TICKER_MAP,
 } from '../../../../shared/constants/network';
 import { SECOND } from '../../../../shared/constants/time';
 import {
@@ -41,7 +42,11 @@ if (process.env.IN_TEST) {
     nickname: 'Localhost 8545',
   };
 } else if (process.env.METAMASK_DEBUG || env === 'test') {
-  defaultProviderConfigOpts = { type: RINKEBY, chainId: RINKEBY_CHAIN_ID };
+  defaultProviderConfigOpts = {
+    type: RINKEBY,
+    chainId: RINKEBY_CHAIN_ID,
+    ticker: TEST_NETWORK_TICKER_MAP.rinkeby,
+  };
 } else {
   defaultProviderConfigOpts = { type: MAINNET, chainId: MAINNET_CHAIN_ID };
 }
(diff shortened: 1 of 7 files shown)


jpuri (dismissed):

darkwing (dismissed):

holantonela:
Thanks for opening this PR Alex! Some comments:

- If we go regular case, it looks more legible eg. `ROPSTENETH` vs `RopstenETH`
Screen Shot 2022-04-19 at 10 19 02
Screen Shot 2022-04-19 at 10 19 13

- I've stressed this UI to see edge cases, it is overlapping. We may want to break that line in two. 

Screen Shot 2022-04-19 at 10 19 40

adonesky1:
Ok I've updated so that we allow lowercase characters for these testnet currencies but preserve the force to uppercase that we were previously doing for all native currencies regardless of how they are input:

Screen Shot 2022-04-20 at 1 19 51 PM

Example of input in lowercase but displayed in uppercase (same as previous behavior)
Screen Shot 2022-04-20 at 1 20 18 PM
Screen Shot 2022-04-20 at 1 20 59 PM

adonesky1:
I've also added wrapping so that very long balance + currency tickers display on separate lines rather than running off the page.

Before:
Screen Shot 2022-04-20 at 1 26 09 PM
Screen Shot 2022-04-20 at 1 26 02 PM

After:
Screen Shot 2022-04-20 at 1 23 09 PM

jpuri (dismissed):

darkwing (dismissed):

adonesky1:
@NiranjanaBinoy fixed the issue [you raised](https://github.com/MetaMask/metamask-extension/pull/14454#issuecomment-1105334638) ☝️  [here](https://github.com/MetaMask/metamask-extension/pull/14454/commits/cd2b64187346b25f3e4d7bb722616421c4afb08e)

jpuri (approved):

darkwing (approved):

PR #14461, opened 2022-04-18
Add friendly error handling when background throws an error before listening for connection

segun:
## Explanation

Currently, if the ui failed to connect to background `browser.runtime.connect({ name: windowType });` for any reason, users only see a blank white screen. This PR fixes that.

## More information

* see #13606 for a detailed description of this issue

### Before

Screenshot 2022-04-18 at 13 29 08

### After

Screenshot 2022-04-18 at 13 29 38

## Manual testing steps

To test this, you can simulate an error in[ background.js:135](https://github.com/MetaMask/metamask-extension/blob/40095cce67807dc89e46917a6422ad1694b2a2c5/app/scripts/background.js#L135) by causing an error like below

```
async function initialize() {
  const initState = await loadStateFromPersistence();
  const initLangCode = await getFirstPreferredLangCode();
  throw 'up'
  await setupController(initState, initLangCode);
  log.info('MetaMask initialization complete.');
```

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -2653,7 +2653,7 @@
     "message": "Make sure you’re using the correct Secret Recovery Phrase before proceeding. You will not be able to undo this."
   },
   "restartMetamask": {
-    "message": "Restart Metamask"
+    "message": "Restart MetaMask"
   },
   "restore": {
     "message": "Restore"
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -2652,6 +2652,9 @@
   "resetWalletWarning": {
     "message": "Make sure you’re using the correct Secret Recovery Phrase before proceeding. You will not be able to undo this."
   },
+  "restartMetamask": {
+    "message": "Restart Metamask"
+  },
   "restore": {
     "message": "Restore"
   },
@@ -2835,6 +2838,9 @@
   },
   "sendAmount": {
     "message": "Send Amount"
+  },
+  "sendBugReport": {
+    "message": "Send us a bug report."
   },
   "sendSpecifiedTokens": {
     "message": "Send $1",
@@ -3108,6 +3114,9 @@
   "step2LedgerWalletMsg": {
     "message": "Connect your wallet directly to your computer.  Unlock your Ledger and open the Ethereum app. For more on using your hardware wallet device, $1.",
     "description": "$1 represents the `hardwareWalletSupportLinkConversion` localization key"
+  },
+  "stillGettingMessage": {
+    "message": "Still getting this message?"
   },
   "storePhrase": {
     "message": "Store this phrase in a password manager like 1Password."
@@ -3801,6 +3810,9 @@
   "troubleConnectingToWallet": {
     "message": "We had trouble connecting to your $1, try reviewing $2 and try again.",
     "description": "$1 is the wallet device name; $2 is a link to wallet connection guide"
+  },
+  "troubleStarting": {
+    "message": "MetaMask had trouble starting. This error could be intermittent, so try restarting the extension."
   },
   "troubleTokenBalances": {
     "message": "We had trouble loading your token balances. You can view them ",
--- a/app/home.html
+++ b/app/home.html
@@ -8,7 +8,10 @@
     <link rel="stylesheet" type="text/css" href="./index-rtl.css" title="rtl" disabled>
   </head>
   <body>
-    <div id="app-content"><div id="app-loader">Loading...</div></div>
+    <div id="app-content">
+      <img class="loading-logo" src="./images/logo/metamask-fox.svg" alt="" />
+      <img class="loading-spinner" src="./images/spinner.gif" alt="" />      
+    </div>
     <div id="popover-content"></div>
     <script src="./globalthis.js" type="text/javascript" charset="utf-8"></script>
     <script src="./sentry-install.js" type="text/javascript" charset="utf-8"></script>
--- a/app/popup.html
+++ b/app/popup.html
@@ -8,7 +8,10 @@
     <link rel="stylesheet" type="text/css" href="./index-rtl.css" title="rtl" disabled>
   </head>
   <body style="width:357px; height:600px;">
-    <div id="app-content"></div>
+    <div id="app-content">
+      <img class="loading-logo" src="./images/logo/metamask-fox.svg" alt="" />
+      <img class="loading-spinner" src="./images/spinner.gif" alt="" />      
+    </div>
     <div id="popover-content"></div>
     <script src="./globalthis.js" type="text/javascript" charset="utf-8"></script>
     <script src="./sentry-install.js" type="text/javascript" charset="utf-8"></script>
--- a/app/scripts/lib/metaRPCClientFactory.js
+++ b/app/scripts/lib/metaRPCClientFactory.js
@@ -24,6 +24,7 @@ class MetaRPCClient {
         return cb(new Error('No response from RPC'), null);
       }
 
+      delete this.responseHandled[id];
       // needed for linter to pass
       return true;
     }, TEN_SECONDS_IN_MILLISECONDS);
--- a/app/scripts/lib/metaRPCClientFactory.js
+++ b/app/scripts/lib/metaRPCClientFactory.js
@@ -1,6 +1,7 @@
 import { EthereumRpcError } from 'eth-rpc-errors';
 import SafeEventEmitter from 'safe-event-emitter';
 import createRandomId from '../../../shared/modules/random-id';
+import { TEN_SECONDS_IN_MILLISECONDS } from '../../../ui/helpers/constants/critical-error';
 
 class MetaRPCClient {
   constructor(connectionStream) {
@@ -10,6 +11,21 @@ class MetaRPCClient {
     this.requests = new Map();
     this.connectionStream.on('data', this.handleResponse.bind(this));
     this.connectionStream.on('end', this.close.bind(this));
+    this.responseHandled = {};
+  }
+
+  send(id, payload, cb) {
+    this.requests.set(id, cb);
+    this.connectionStream.write(payload);
+    this.responseHandled[id] = false;
+    setTimeout(() => {
+      if (!this.responseHandled[id] && cb) {
+        return cb(new Error('No response from RPC'), null);
+      }
+
+      // needed for linter to pass
+      return true;
+    }, TEN_SECONDS_IN_MILLISECONDS);
   }
 
   onNotification(handler) {
@@ -33,6 +49,8 @@ class MetaRPCClient {
     const { id, result, error, method, params } = data;
     const isNotification = id === undefined && error === undefined;
     const cb = this.requests.get(id);
+
+    this.responseHandled[id] = true;
 
     if (method && params && !isNotification) {
       // dont handle server-side to client-side requests
@@ -79,14 +97,13 @@ const metaRPCClientFactory = (connectionStream) => {
         const cb = p[p.length - 1];
         const params = p.slice(0, -1);
         const id = createRandomId();
-
-        object.requests.set(id, cb);
-        object.connectionStream.write({
+        const payload = {
           jsonrpc: '2.0',
           method: property,
           params,
           id,
-        });
+        };
+        object.send(id, payload, cb);
       };
     },
   });
--- a/app/scripts/lib/metaRPCClientFactory.js
+++ b/app/scripts/lib/metaRPCClientFactory.js
@@ -20,6 +20,7 @@ class MetaRPCClient {
     this.responseHandled[id] = false;
     setTimeout(() => {
       if (!this.responseHandled[id] && cb) {
+        delete this.responseHandled[id];
         return cb(new Error('No response from RPC'), null);
       }
 
--- a/app/scripts/lib/metaRPCClientFactory.test.js
+++ b/app/scripts/lib/metaRPCClientFactory.test.js
@@ -131,4 +131,16 @@ describe('metaRPCClientFactory', () => {
       },
     });
   });
+
+  it('should be able to handle no message within TIMEOUT secs', async () => {
+    const streamTest = createThoughStream();
+    const metaRPCClient = metaRPCClientFactory(streamTest);
+    metaRPCClient.foo('bad', (error, _) => {
+      expect(error.message).toStrictEqual('No response from RPC');
+    });
+
+    const delay = (ms) => new Promise((res) => setTimeout(res, ms));
+
+    await delay(11000);
+  }, 20000);
 });
--- a/app/scripts/lib/metaRPCClientFactory.test.js
+++ b/app/scripts/lib/metaRPCClientFactory.test.js
@@ -133,14 +133,19 @@ describe('metaRPCClientFactory', () => {
   });
 
   it('should be able to handle no message within TIMEOUT secs', async () => {
+    jest.useFakeTimers();
     const streamTest = createThoughStream();
     const metaRPCClient = metaRPCClientFactory(streamTest);
-    metaRPCClient.foo('bad', (error, _) => {
-      expect(error.message).toStrictEqual('No response from RPC');
-    });
 
-    const delay = (ms) => new Promise((res) => setTimeout(res, ms));
+    const errorPromise = new Promise((_resolve, reject) =>
+      metaRPCClient.foo('bad', (error, _) => {
+        reject(error);
+      }),
+    );
 
-    await delay(11000);
-  }, 20000);
+    jest.runOnlyPendingTimers();
+    await expect(errorPromise).rejects.toThrow('No response from RPC');
+
+    jest.useRealTimers();
+  });
 });
--- a/app/scripts/ui.js
+++ b/app/scripts/ui.js
@@ -17,6 +17,8 @@ import {
   ENVIRONMENT_TYPE_POPUP,
 } from '../../shared/constants/app';
 import { isManifestV3 } from '../../shared/modules/mv3.utils';
+import { SUPPORT_LINK } from '../../ui/helpers/constants/common';
+import { getErrorHtml } from '../../ui/helpers/utils/error-utils';
 import ExtensionPlatform from './platforms/extension';
 import { setupMultiplex } from './lib/stream-utils';
 import { getEnvironmentType } from './lib/util';
@@ -25,6 +27,21 @@ import metaRPCClientFactory from './lib/metaRPCClientFactory';
 start().catch(log.error);
 
 async function start() {
+  async function displayCriticalError(container, err, metamaskState) {
+    const html = await getErrorHtml(SUPPORT_LINK, metamaskState);
+
+    container.innerHTML = html;
+
+    const button = document.getElementById('critical-error-button');
+
+    button.addEventListener('click', (_) => {
+      browser.runtime.reload();
+    });
+
+    log.error(err.stack);
+    throw err;
+  }
+
   // create platform global
   global.platform = new ExtensionPlatform();
 
@@ -33,6 +50,7 @@ async function start() {
 
   // setup stream to background
   const extensionPort = browser.runtime.connect({ name: windowType });
+
   const connectionStream = new PortStream(extensionPort);
 
   const activeTab = await queryCurrentActiveTab(windowType);
@@ -51,19 +69,12 @@ async function start() {
     initializeUiWithTab(activeTab);
   }
 
-  function displayCriticalError(container, err) {
-    container.innerHTML =
-      '<div class="critical-error">The MetaMask app failed to load: please open and close MetaMask again to restart.</div>';
-    container.style.height = '80px';
-    log.error(err.stack);
-    throw err;
-  }
-
   function initializeUiWithTab(tab) {
     const container = document.getElementById('app-content');
     initializeUi(tab, container, connectionStream, (err, store) => {
       if (err) {
-        displayCriticalError(container, err);
+        // if there's an error, store will be = metamaskState
+        displayCriticalError(container, err, store);
         return;
       }
 
@@ -104,7 +115,7 @@ async function queryCurrentActiveTab(windowType) {
 function initializeUi(activeTab, container, connectionStream, cb) {
   connectToAccountManager(connectionStream, (err, backgroundConnection) => {
     if (err) {
-      cb(err);
+      cb(err, null);
       return;
     }
 
--- a/ui/css/errors.scss
+++ b/ui/css/errors.scss
@@ -1,0 +1,44 @@
+.critical-error {
+  padding: 16px;
+  max-width: 600px;
+  margin: 0 auto;
+
+  &__alert {
+    color: var(--color-text-default);
+    background-color: var(--color-error-muted);
+    border: 1px solid var(--color-error-default);
+    border-radius: 8px;
+    padding: 16px;
+    display: flex;
+    flex-direction: column;
+    margin-bottom: 16px;
+
+    &__message {
+      margin-bottom: 16px;
+    }
+
+    &__button {
+      height: 40px;
+      border-radius: 20px;
+      padding-left: 16px;
+      padding-right: 16px;
+      background-color: var(--color-primary-default);
+      color: var(--color-primary-inverse);
+      border: 1px solid var(--color-primary-default);
+      margin: 0 auto;
+    }
+  }
+
+  &__paragraph {
+    color: var(--color-text-default);
+    text-align: center;
+
+    &__link {
+      color: var(--color-primary-default);
+
+      &:hover {
+        color: var(--color-primary-alternative);
+      }
+    }
+  }
+}
--- a/ui/css/loading.scss
+++ b/ui/css/loading.scss
@@ -1,0 +1,13 @@
+.loading-logo {
+  width: 10rem;
+  height: 10rem;
+  align-self: center;
+  margin: 10rem 0 0 0;
+}
+
+.loading-spinner {
+  width: 2rem;
+  height: 2rem;
+  align-self: center;
+  margin-top: 1rem;
+}
--- a/ui/helpers/utils/error-utils.js
+++ b/ui/helpers/utils/error-utils.js
@@ -1,0 +1,55 @@
+import getFirstPreferredLangCode from '../../../app/scripts/lib/get-first-preferred-lang-code';
+import { setupLocale } from '../..';
+import switchDirection from './switch-direction';
+
+const getLocaleContext = (currentLocaleMessages, enLocaleMessages) => {
+  return (key) => {
+    let message = currentLocaleMessages[key]?.message;
+    if (!message && enLocaleMessages[key]) {
+      message = enLocaleMessages[key].message;
+    }
+    return message;
+  };
+};
+
+export async function getErrorHtml(supportLink, metamaskState) {
+  let response, preferredLocale;
+  if (metamaskState?.currentLocale) {
+    preferredLocale = metamaskState.currentLocale;
+    response = await setupLocale(metamaskState.currentLocale);
+  } else {
+    preferredLocale = await getFirstPreferredLangCode();
+    response = await setupLocale(preferredLocale);
+  }
+
+  const textDirection = ['ar', 'dv', 'fa', 'he', 'ku'].includes(preferredLocale)
+    ? 'rtl'
+    : 'auto';
+
+  switchDirection(textDirection);
+  const { currentLocaleMessages, enLocaleMessages } = response;
+  const t = getLocaleContext(currentLocaleMessages, enLocaleMessages);
+
+  return `
+    <div class="critical-error">
+      <div class="critical-error__alert">
+        <p class="critical-error__alert__message">
+          ${t('troubleStarting')}        
+        </p>   
+        <button id='critical-error-button' class="critical-error__alert__button">
+          ${t('restartMetamask')}
+        </button>
+      </div>    
+      <p class="critical-error__paragraph">    
+        ${t('stillGettingMessage')}
+        <a           
+          href=${supportLink} 
+          class="critical-error__paragraph__link" 
+          target="_blank" 
+          rel="noopener noreferrer">
+            ${t('sendBugReport')}
+          </a>  
+      </p>
+    </div>
+    `;
+}
--- a/ui/helpers/utils/error-utils.test.js
+++ b/ui/helpers/utils/error-utils.test.js
@@ -1,0 +1,47 @@
+import { SUPPORT_LINK } from '../constants/common';
+import { getErrorHtml } from './error-utils';
+
+const defaultState = {
+  localeMessages: {
+    current: {
+      troubleStarting: {
+        message:
+          'MetaMask had trouble starting. This error could be intermittent, so try restarting the extension.',
+      },
+      restartMetamask: {
+        message: 'Restart Metamask',
+      },
+      stillGettingMessage: {
+        message: 'Still getting this message?',
+      },
+      sendBugReport: {
+        message: 'Send us a bug report.',
+      },
+    },
+  },
+  metamask: {
+    currentLocale: 'en',
+  },
+};
+
+jest.mock('./i18n-helper', () => ({
+  fetchLocale: jest.fn((_locale) => defaultState.localeMessages.current),
+  loadRelativeTimeFormatLocaleData: jest.fn(),
+}));
+
+describe('Error utils Tests', () => {
+  it('should get error html', async () => {
+    const errorHtml = await getErrorHtml(SUPPORT_LINK, defaultState.metamask);
+    const currentLocale = defaultState.localeMessages.current;
+    const troubleStartingMessage = currentLocale.troubleStarting.message;
+    const restartMetamaskMessage = currentLocale.restartMetamask.message;
+    const stillGettingMessageMessage =
+      currentLocale.stillGettingMessage.message;
+    const sendBugReportMessage = currentLocale.sendBugReport.message;
+
+    expect(errorHtml).toContain(troubleStartingMessage);
+    expect(errorHtml).toContain(restartMetamaskMessage);
+    expect(errorHtml).toContain(stillGettingMessageMessage);
+    expect(errorHtml).toContain(sendBugReportMessage);
+  });
+});
--- a/ui/helpers/utils/error-utils.test.js
+++ b/ui/helpers/utils/error-utils.test.js
@@ -1,36 +1,38 @@
 import { SUPPORT_LINK } from '../constants/common';
 import { getErrorHtml } from './error-utils';
-
-const mockStore = {
-  localeMessages: {
-    current: {
-      troubleStarting: {
-        message:
-          'MetaMask had trouble starting. This error could be intermittent, so try restarting the extension.',
-      },
-      restartMetamask: {
-        message: 'Restart MetaMask',
-      },
-      stillGettingMessage: {
-        message: 'Still getting this message?',
-      },
-      sendBugReport: {
-        message: 'Send us a bug report.',
-      },
-    },
-  },
-  metamask: {
-    currentLocale: 'en',
-  },
-};
+import { fetchLocale } from './i18n-helper';
 
 jest.mock('./i18n-helper', () => ({
-  fetchLocale: jest.fn((_locale) => mockStore.localeMessages.current),
+  fetchLocale: jest.fn(),
   loadRelativeTimeFormatLocaleData: jest.fn(),
 }));
 
 describe('Error utils Tests', () => {
   it('should get error html', async () => {
+    const mockStore = {
+      localeMessages: {
+        current: {
+          troubleStarting: {
+            message:
+              'MetaMask had trouble starting. This error could be intermittent, so try restarting the extension.',
+          },
+          restartMetamask: {
+            message: 'Restart MetaMask',
+          },
+          stillGettingMessage: {
+            message: 'Still getting this message?',
+          },
+          sendBugReport: {
+            message: 'Send us a bug report.',
+          },
+        },
+      },
+      metamask: {
+        currentLocale: 'en',
+      },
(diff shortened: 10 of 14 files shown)


app/scripts/ui.js
```diff
@@ -11,19 +11,63 @@ import Eth from 'ethjs';
...
+      ${messages[msg].message}
+    </div>
+    <blockquote>
+      ${err.stack}
+      </blockquote>
+    <a href=${SUPPORT_LINK} target="_blank" rel="noopener noreferrer">
+      ${messages.needHelpLinkText.message}    
+    </a>  
+    `;
+    console.log('HTML: ', html);
```
jpuri, line 64: 🔪  `console.log('HTML: ', html);`

app/scripts/ui.js
```diff
@@ -32,24 +76,26 @@ async function start() {
...
-  }
-
   function initializeUiWithTab(tab) {
     const container = document.getElementById('app-content');
     initializeUi(tab, container, connectionStream, (err, store) => {
       if (err) {
-        displayCriticalError(container, err);
+        displayCriticalError(container, err, 'failedToLoadMessage');
         return;
       }
```
jpuri, line 100: At times frontend does not throw error but there is white screen for long, if RPC connection can not be established between background and UI. I am not sure if these changes take care of that scenario also.

app/scripts/ui.js
```diff
@@ -11,19 +11,62 @@ import Eth from 'ethjs';
...
+    </div>
+    <blockquote>
+      ${err.stack}
+      </blockquote>
+    <a href=${SUPPORT_LINK} target="_blank" rel="noopener noreferrer">
+      ${messages.needHelpLinkText.message}    
+    </a>  
+    `;
+    container.innerHTML = html;
+    // container.style.height = '80px';
```
darkwing, line 65: Looks like we could probably remove this and accomplish it with CSS.

app/scripts/ui.js
```diff
@@ -11,19 +11,62 @@ import Eth from 'ethjs';
...
+        background-color:yellow; 
+        text-align: left
+      }
+      blockquote {
+        margin: 20px 0 20px 5px; 
+        padding-left: 5px; 
+        border-left: 1px solid grey
+      }  
+      a {color: blue}
+      a:hover {color: green}
```
darkwing, line 52: Could we borrow CSS styles from the `ActionableMessage` component so this error looks uniform to other errors we present to the user?

darkwing (changes requested):

app/scripts/ui.js
danjm: I don't think so, because I don't think we can use React here.

app/scripts/ui.js
```diff
@@ -11,19 +11,62 @@ import Eth from 'ethjs';
...
+    </style>
+    <div class="critical-error">
+      ${messages[msg].message}
```
danjm, line 55: @Gudahtt We could use your advice here. We are displaying an error message to the user by directly setter `innerHTML` (see below). This is happening in a place where we don't have access to react. A new message has been added.

Recommendations for how to get this to pass the locale verification script?

app/scripts/ui.js
Gudahtt: The CSS isn't added by React. They're unrelated. The CSS is bundled into a single file and added to the document separately from the JS, in the `<head>` actually so it should load first I think.

Gudahtt: It does break conventions to use a component classname outside of that component though. Maybe we can lift or copy the error styles into a global class, something in `ui/css`

app/scripts/ui.js
Gudahtt: The verify locales script look for `t('[key]')`. It doesn't know what `t` is, it just looks for a function call with that name. We can create an equivalent function here.

See here for the definition of `t`: https://github.com/MetaMask/metamask-extension/blob/develop/ui/contexts/i18n.js#L18

Gudahtt: Also it's critical that we have a fallback for if the localization fails, or if the attempt to download the localized messages fails, or anything like that. All errors need to be caught, otherwise we'll still get the blank screen. Showing the wrong language is better than showing a blank screen.

app/scripts/ui.js
```diff
@@ -11,19 +11,62 @@ import Eth from 'ethjs';
...
+      }
+      blockquote {
+        margin: 20px 0 20px 5px; 
+        padding-left: 5px; 
+        border-left: 1px solid grey
+      }  
+      a {color: blue}
+      a:hover {color: green}
+    </style>
+    <div class="critical-error">
```
Gudahtt, line 54: We can add these elements directly to `home.html`. I don't think there's any need to set `innerHTML` at runtime except to set a localized message.

app/scripts/ui.js
```diff
@@ -11,19 +11,62 @@ import Eth from 'ethjs';
...
 import ExtensionPlatform from './platforms/extension';
 import { setupMultiplex } from './lib/stream-utils';
 import { getEnvironmentType } from './lib/util';
 import metaRPCClientFactory from './lib/metaRPCClientFactory';
+import getFirstPreferredLangCode from './lib/get-first-preferred-lang-code';
 
 start().catch(log.error);
 
 async function start() {
+  const preferredLocale = await getFirstPreferredLangCode();
```
Gudahtt, line 30: This locale setup is repeated again in `ui/index.js`. There is no need for us to do this twice. Also this doesn't handle RTL locales correctly.

app/scripts/ui.js
Gudahtt: Showing a loading screen initially instead of a blank screen should address this problem. Presumably the connection attempt will eventually timeout. If not, we can add a timeout.

segun:
@danjm @Gudahtt @mcmire @jpuri 

After my 2 day discussion with @jpuri I realise this issue can occur in many other instances than captured here in the PR. 

One such instance is if `controller.setupTrustedCommunication(portStream, remotePort.sender);` [here](https://github.com/MetaMask/metamask-extension/blob/eb241ae68f66e5cca01266112bfaf07c1f7c29e0/app/scripts/background.js#L369) fails. What happens in that instance is UI sends a request to background and never receives a response. Commenting out that line results in a white screen.

One solution we came up with is adding a timeout that checks <app-content> and shows error if UI is not loaded after XXX seconds like below

`ui.js:88`
```
  setTimeout(() => {
    const container = document.getElementById('app-content');
    if(container.innerHTML === 'Loading...') {
      displayCriticalError(
        container,
        new Error('Connection lost to background script'),
        'backgroundPortClosedError',
      );      
    }
  }, 10000);
  ```
This will catch issues when the UI goes blank when launching the app. 

Of course the major concern is what is the best time to wait for the UI before showing this error.
   
A second question I asked is if the dis-connection between UI and background happens after the initial UI has been initialised, then the above fix won't catch this scenario. I'm not sure how to simulate this but we can catch it (which also covers for the blank screen at launch) by applying fix-2 below

Fix-2.
It also involves setting a timeout, but this time we apply the timeout in 

[metaRPCClientFactory.js](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/lib/metaRPCClientFactory.js)
1. Add a method timeoutAfterWait

[metaRPCClientFactory.js:33](https://github.com/MetaMask/metamask-extension/blob/eb241ae68f66e5cca01266112bfaf07c1f7c29e0/app/scripts/lib/metaRPCClientFactory.js#L33)
```
  timeoutAfterWait(cb) {
    setTimeout(() => {
      if (!this.responseHandled) {
        cb(new Error('No response from RPC after 10000 secs'), null);
      }
    }, 10000);
  }
```

2. Add code to handle responseHandled properly.
[metaRPCClientFactory.js:33](https://github.com/MetaMask/metamask-extension/blob/eb241ae68f66e5cca01266112bfaf07c1f7c29e0/app/scripts/lib/metaRPCClientFactory.js#L13)
```
this.responseHandled = false;
```
[metaRPCClientFactory.js:33](https://github.com/MetaMask/metamask-extension/blob/eb241ae68f66e5cca01266112bfaf07c1f7c29e0/app/scripts/lib/metaRPCClientFactory.js#L42)
```
this.responseHandled = true;
```
[metaRPCClientFactory.js:33](https://github.com/MetaMask/metamask-extension/blob/eb241ae68f66e5cca01266112bfaf07c1f7c29e0/app/scripts/lib/metaRPCClientFactory.js#L94)
```
this.responseHandled = false;
```

This will catch all instances when UI can't receive response from background.

Again the big question is what value is best for timeout.

File: app/scripts/ui.js (hunk 2 above)
Comment (segun) [reply app/scripts/ui.js:99]: https://github.com/MetaMask/metamask-extension/pull/14461#issuecomment-1105069726

File: app/scripts/ui.js (hunk 4 above)
Comment (danjm) [reply app/scripts/ui.js:52]: good point, so disregard my comment. Good suggestions here from Mark

**Comment by danjm** [comment]
I think a timeout of 10 seconds is fine. Also, adding a loading screen as @Gudahtt suggested would be a good addition to this PR

File: app/scripts/lib/metaRPCClientFactory.js [hunk 8]
```diff
@@ -87,6 +101,8 @@ const metaRPCClientFactory = (connectionStream) => {
           params,
           id,
         });
+        metaRPCClient.responseHandled = false;
+        metaRPCClient.timeoutAfterWait(cb);
```
mcmire, line 105: This method seems disconnected from the other things that MetaRPCClient is doing. The class already has `cb` because you're passing it when you call `object.requests.set(id, cb)`, so it seems like we ought to be able to use that. Perhaps we should extract the `requests.set` and `connectionStream.write` lines to a method and then place the `setTimeout` after that. This method could look something like:

``` javascript
class MetaRPCClient {
  // ...
  send(id, payload, cb) {
    this.requests.set(id, cb);
    this.connectionStream.write(payload);
    setTimeout(() => {
      if (!this.responseHandled) {
        cb(new Error('No response from RPC'), null);
      }
    }, 10000);
  }
  // ...
}
```

And now here in `metaRPCClientFactory` we could say:

``` javascript
object.send(id, {
  jsonrpc: '2.0',
  method: property,
  params,
  id,
}, cb);
```

ui/css/errors.scss
```diff
@@ -0,0 +1,47 @@
...
+
+.critical-error-anchor {
+  color: blue;
+}
+
+.critical-error-anchor:hover {
+  color: green;
+}
+
+#loading__logo {
```
mcmire, line 35: These selectors seem to be related to loading rather than errors; should they be extracted to a new file?

app/_locales/en/messages.json
```diff
@@ -326,7 +326,7 @@
     "message": "Back to All"
   },
   "backgroundPortClosedError": {
-    "message": "Your connection to Metamask was interrupted. Quit your browser and try again. If issue persists, please log a bug report by clicking on Metamask link below"
+    "message": "Your connection to MetaMask was interrupted. Quit your browser and try again. If this issue persists, please log a bug report by clicking on the link below"
```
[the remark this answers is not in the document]
jpuri: Agreed

jpuri:
@segun : regarding your point:
> dis-connection between UI and background happens after the initial UI has been initialised

I think major issue is UI blank on initial load, I have not seen reports of disconnection later.

app/_locales/en/messages.json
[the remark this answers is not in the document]
segun: I'm having a real hard time adding this button
```
["Restart MetaMask" button]``
```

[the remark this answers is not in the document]
segun: @mcmire @jpuri really really having a difficult time adding that Button. Main issue is we can't use onClick in extension, we have to add an event listener and that doesn't seem to the working for this either.

I noticed though that closing the extension window and re-launching it basically restarts the extension after this error. Can we reframe the message to reflect this e.g

`....so try restarting the extension by closing the extension window and re-opening it.`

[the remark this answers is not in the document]
mcmire: > Main issue is we can't use onClick in extension, we have to add an event listener and that doesn't seem to the working for this either.

What's not working? Like you can't make the button do anything when clicked?

> Can we reframe the message to reflect this e.g
>
> `....so try restarting the extension by closing the extension window and re-opening it.`

I'm not sure that users will understand what that means. Also, closing the extension window may reload the UI process, but it doesn't reload the background process. Wouldn't we need to do that, if the background process fails to start correctly? It's been a while since I've thought about this so it's possible I'm not understanding the problem correctly anymore.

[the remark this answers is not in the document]
segun: > What's not working? Like you can't make the button do anything when clicked?

Yeah, because inline js is not supported. Usually you setup an event listener for a button using the ID or classname and when the button gets clicked, the event listener is fired. In this case, the button is dynamic and the inline-js is not executed.

>  Also, closing the extension window may reload the UI process, but it doesn't reload the background process
In the cases I've simulated, when background process is not properly started you close and re-open the UI and it restarts the background process. Even in MV3, when you open a UI and the service worker is dead, it restarts the service worker.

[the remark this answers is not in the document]
mcmire: > In the cases I've simulated, when background process is not properly started you close and re-open the UI and it restarts the background process. Even in MV3, when you open a UI and the service worker is dead, it restarts the service worker.

Ah okay. I learned something new today then :)

> In this case, the button is dynamic

You should still be able to attach an event listener to a button that's added dynamically, using `addEventListener`. I think you have to find the button after it's added to the DOM though, say, after here: https://github.com/MetaMask/metamask-extension/pull/14461/files#diff-5a9d964db9d7c2fe207e21470884b2710e54723e0c1e6e6f7fd264706604fd6cR32.

danjm:
This looks pretty good to me. We need to investigate and resolve the failing e2e test, and see if we can resolve Elliot's suggestion above https://github.com/MetaMask/metamask-extension/pull/14461#discussion_r858996223

app/scripts/lib/metaRPCClientFactory.js
```diff
@@ -10,6 +10,21 @@ class MetaRPCClient {
     this.requests = new Map();
     this.connectionStream.on('data', this.handleResponse.bind(this));
     this.connectionStream.on('end', this.close.bind(this));
+    this.responseHandled = false;
```
Gudahtt, line 13: This client is written to handle multiple requests simultaneously. That's why the request callbacks are tracked with a map, with ids as the key. This looks like it assumes there is only one request happening at a time. As such this seems to be totally broken.

Gudahtt (commented):
There appear to be no new tests for these changes. These should be fully tested before being merged.

app/scripts/lib/metaRPCClientFactory.js
```diff
@@ -10,6 +10,21 @@ class MetaRPCClient {
...
+    this.requests.set(id, cb);
+    this.connectionStream.write(payload);
+    setTimeout(() => {
+      if (!this.responseHandled) {
+        return cb(new Error('No response from RPC'), null);
+      }
+
+      // needed for linter to pass
+      return true;
+    }, 10000);
```
Gudahtt, line 27: Nit: for readability, I'd suggesting pulling this out into a constant with a nicer name, e.g. `TEN_SECONDS_IN_MILLISECONDS`. Also we can use numeric separators now, e.g. `10_000` to make the literal number easier to read.

shared/modules/error-utils.js
```diff
@@ -0,0 +1,46 @@
+import getFirstPreferredLangCode from '../../app/scripts/lib/get-first-preferred-lang-code';
+import { setupLocale } from '../../ui';
+
+let currentLocaleMessages, enLocaleMessages;
+
+getFirstPreferredLangCode().then((preferredLocale) => {
```
Gudahtt, line 6: Generally we should not be using `.then` anymore, because async stack traces don't work with `.then`. Instead move this into an async function and use `await`.

Also, the error case isn't handled here. What should we do if this fails? When would we expect this to fail? This goes for both operations.

Right now, if this fails, it would bubble up and get sent to Sentry. It would also cause each invocation of the `t` function below to throw an exception.

ui/index.js
```diff
@@ -46,21 +46,29 @@ export default function launchMetamaskUi(opts, cb) {
...
-  await loadRelativeTimeFormatLocaleData('en');
-  if (metamaskState.currentLocale) {
-    await loadRelativeTimeFormatLocaleData(metamaskState.currentLocale);
-  }
+  const { currentLocaleMessages, enLocaleMessages } = await setupLocale(
+    metamaskState.currentLocale,
+  );
 
   if (metamaskState.textDirection === 'rtl') {
     await switchDirection('rtl');
```
Gudahtt, line 74: RTL locales still won't work property now on the error page because this step is missed. This is a part of the locale setup.

shared/modules/error-utils.js
```diff
@@ -0,0 +1,46 @@
+import getFirstPreferredLangCode from '../../app/scripts/lib/get-first-preferred-lang-code';
+import { setupLocale } from '../../ui';
+
+let currentLocaleMessages, enLocaleMessages;
```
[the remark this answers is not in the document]
danjm: But one of the important error cases this PR is meant to handle is when attempting to intialize the UI before the background connection has been established, or if the attempt to connect to the background fails. In those cases, redux won't be instantiated yet, and we probably can't get the background state.

So it seems unavoidable that we have to directly read the messages from the file in some cases.

Maybe we can refactor this so that these messages are not left in memory?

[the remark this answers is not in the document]
Gudahtt: If the UI fails to get the initial message from the backgound, with the initial background state, we won't know the users locale. Defaulting to english or the browser preferred locale might be fine there (honestly english might be a safer default, given how often the browser preference is wrong, and given that we don't have any way to let the user change the setting on the error screen).

Once we get that initial message from the background, we have the users locale. That gets stored in Redux, but even if there is a crash before Redux is constructed, we have the locale in-memory and we can use it. It's just a matter of how to share that reference between this error code and the UI setup code. I don't think there's any reason to have two copies of the localized messages in-memory in any scenario.

ui/index.js
```diff
@@ -36,7 +36,7 @@ export default function launchMetamaskUi(opts, cb) {
   // check if we are unlocked first
   backgroundConnection.getState(function (err, metamaskState) {
     if (err) {
-      cb(err);
+      cb(err, metamaskState);
```
danjm, line 39: As discussed, this callback is expecting a store. Need to modify the callback so that metamaskState can be passed down

ui/index.js
```diff
@@ -46,21 +46,31 @@ export default function launchMetamaskUi(opts, cb) {
...
+
+  await loadRelativeTimeFormatLocaleData('en');
+  if (currentLocale) {
+    await loadRelativeTimeFormatLocaleData(currentLocale);
+  }
+
+  return { currentLocaleMessages, enLocaleMessages };
+};
+
+export const setupLocale = memoize(_setupLocale);
```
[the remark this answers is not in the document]
segun: added [here](https://github.com/MetaMask/metamask-extension/pull/14461/commits/23828dfb101c3b9a6ab7ed585f158696edffa91b)

ui/index.js
segun: fixed [here](https://github.com/MetaMask/metamask-extension/pull/14461/commits/23828dfb101c3b9a6ab7ed585f158696edffa91b)

ui/index.js
segun: fixed [here](https://github.com/MetaMask/metamask-extension/pull/14461/commits/23828dfb101c3b9a6ab7ed585f158696edffa91b)

app/_locales/en/messages.json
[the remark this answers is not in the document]
mcmire: Hi @segun sorry for the bump — were you able to look into this?

georgewrmarshall (commented):
Hey @segun, Is there a way I can test this easily? I just want to check out the UI

segun:
> Hey @segun, Is there a way I can test this easily? I just want to check out the UI

@georgewrmarshall the easiest way to test is to comment out line 369 in app/scripts/background.js

app/home.html
```diff
@@ -8,7 +8,10 @@
     <link rel="stylesheet" type="text/css" href="./index-rtl.css" title="rtl" disabled>
   </head>
   <body>
-    <div id="app-content"></div>
+    <div id="app-content">
+      <img id="loading__logo" src="./images/logo/metamask-fox.svg" alt="" />
+      <img id="loading__spinner" src="./images/spinner.gif" alt="" />
```
georgewrmarshall, line 13: Can we use classes here instead of ids if we are using the identifier to add styles?

```suggestion
      <img class="loading-logo" src="./images/logo/metamask-fox.svg" alt="MetaMask" />
      <img class="loading-spinner" src="./images/spinner.gif" alt="Loading" />      
```

app/popup.html
```diff
@@ -8,7 +8,10 @@
     <link rel="stylesheet" type="text/css" href="./index-rtl.css" title="rtl" disabled>
   </head>
   <body style="width:357px; height:600px;">
-    <div id="app-content"></div>
+    <div id="app-content">
+      <img id="loading__logo" src="./images/logo/metamask-fox.svg" alt="" />
+      <img id="loading__spinner" src="./images/spinner.gif" alt="" />
```
georgewrmarshall, line 13: Same here?
```suggestion
      <img class="loading-logo" src="./images/logo/metamask-fox.svg" alt="MetaMask" />
      <img class="loading-spinner" src="./images/spinner.gif" alt="Loading" />      
```

ui/css/loading.scss
```diff
@@ -0,0 +1,13 @@
+#loading__logo {
```
georgewrmarshall, line 1: Update to class
```suggestion
.loading-logo {
```

ui/css/loading.scss
```diff
@@ -0,0 +1,13 @@
...
+  width: 10rem;
+  height: 10rem;
+  align-self: center;
+  margin: 10rem 0 0 0;
+}
+
+#loading__spinner {
```
georgewrmarshall, line 8: Update to class
```suggestion
.loading-spinner {
```

ui/css/errors.scss
```diff
@@ -0,0 +1,44 @@
...
+
+.critical-error-anchor:hover {
+  color: green;
+}
+
+.critical-error-button {
+  color: var(--color-primary-default);
+  background: transparent;
+  white-space: pre;
+}
```
georgewrmarshall, line 44: Could we use BEM naming conventions here just so it's better understood that all these styles belong to one component? I've updated some of the styling here to follow our current design system guidelines hope that's ok. You could replace the above with the below.

```scss
.critical-error {
  padding: 16px;
  max-width: 600px;
  margin: 0 auto;

  &__alert {
    color: var(--color-text-default);
    background-color: var(--color-error-muted);
    border: 1px solid var(--color-error-default);
    border-radius: 8px;
    padding: 16px;
    display: flex;
    flex-direction: column;
    margin-bottom: 16px;

    &__message {
      margin-bottom: 16px;
    }

    &__button {
      height: 40px;
      border-radius: 20px;
      padding-left: 16px;
      padding-right: 16px;
      background-color: var(--color-primary-default);
      color: var(--color-primary-inverse);
      border: 1px solid var(--color-primary-default);
      margin: 0 auto;
    }
  }

  &__paragraph {
    color: var(--color-text-default);
    text-align: center;

    &__link {
      color: var(--color-primary-default);

      &:hover {
        color: var(--color-primary-alternative);
      }
    }
  }
}


```

ui/helpers/utils/error-utils.js
```diff
@@ -0,0 +1,55 @@
...
+        ${t('stillGettingMessage')}
+        <a           
+          href=${supportLink} 
+          class="critical-error-anchor" 
+          target="_blank" 
+          rel="noopener noreferrer">
+            ${t('sendBugReport')}
+          </a>  
+      </p>
+    </div>
```
georgewrmarshall, line 53: Updating html based on BEM naming convention and new UI layout

```suggestion
    <div class="critical-error">
      <div class="critical-error__alert">
        <p class="critical-error__alert__message">
          ${t('troubleStarting')}
        </p>
        <button
          id="critical-error-button"
          class="critical-error__alert__button"
        >
          ${t('restartMetamask')}
        </button>
      </div>
      <p class="critical-error__paragraph">
        ${t('stillGettingMessage')}
        <a
          href=${supportLink}
          class="critical-error__paragraph__link"
          target="_blank"
          rel="noopener noreferrer"
        >
          ${t('sendBugReport')}
        </a>
      </p>
    </div>
```

georgewrmarshall (changes requested):
Love this update! I hate the blank screen of death 😄 I've made some css style and naming convention suggestions 🙏 

### Before

Screen Shot 2022-05-20 at 11 46 26 AM
Screen Shot 2022-05-20 at 11 46 44 AM


###  After
Screen Shot 2022-05-20 at 11 40 39 AM

Screen Shot 2022-05-20 at 11 41 04 AM

ui/helpers/utils/error-utils.js
segun: @georgewrmarshall thanks for the UI/CSS suggestions. All applied [here](https://github.com/MetaMask/metamask-extension/pull/14461/commits/af15b73ad4ca96b85ca2a75aa6b9b9040ec40bbb)

darkwing (changes requested):

app/scripts/ui.js
```diff
@@ -24,6 +26,21 @@ import metaRPCClientFactory from './lib/metaRPCClientFactory';
...
 async function start() {
+  async function displayCriticalError(container, err, metamaskState) {
+    const html = await getErrorHtml(SUPPORT_LINK, metamaskState);
+
+    container.innerHTML = html;
+
+    const button = document.getElementById('critical-error-button');
+
+    button.addEventListener('click', (_) => {
+      browser.runtime.reload();
```
jpuri, line 37: Here we can all method to initialise UI again rather than reloading the browser.

app/home.html
joani24: > Can we use classes here instead of ids if we are using the identifier to add styles?

georgewrmarshall (commented):
Looks great. Nice work! 💯 Once @darkwing's comment has been resolved I will approve for UI 🙏

danjm:
@segun Do you know why the tests are failing?

app/scripts/ui.js
danjm: What method can we call to initialise the UI?

app/scripts/background.js
```diff
@@ -366,7 +366,7 @@ function setupController(initState, initLangCode) {
       const portStream = new PortStream(remotePort);
       // communication with popup
       controller.isClientOpen = true;
-      controller.setupTrustedCommunication(portStream, remotePort.sender);
+      // controller.setupTrustedCommunication(portStream, remotePort.sender);
```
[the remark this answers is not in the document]
danjm: What does `controller.setupTrustedCommunication(portStream, remotePort.sender);` do? And how do we know that it is safe to remove this?

danjm (changes requested):

app/scripts/background.js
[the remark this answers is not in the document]
segun: shit...I needed to do this to simulate the failure, I usually removes it before I push...forgot it there this time...probably why the tests are also failing.

app/scripts/ui.js
segun: > @segun Do you know why the tests are failing?

Looking into it.

jpuri: https://github.com/MetaMask/metamask-extension/pull/14461/files/af15b73ad4ca96b85ca2a75aa6b9b9040ec40bbb#diff-5a9d964db9d7c2fe207e21470884b2710e54723e0c1e6e6f7fd264706604fd6cR58 can help

ui/index.js
segun: so if you take a look at ui/index.js:33 the launchMetamaskUi() function, you'll notice that if there's an error, callback will be called with metamaskState (at this point the store has not been initialised). line42 is where the store is initialised and then the callback is called with null as error..

So basically, if there's an error, store will be = metamaskState

app/scripts/lib/metaRPCClientFactory.test.js
```diff
@@ -131,4 +131,16 @@ describe('metaRPCClientFactory', () => {
...
+    const streamTest = createThoughStream();
+    const metaRPCClient = metaRPCClientFactory(streamTest);
+    metaRPCClient.foo('bad', (error, _) => {
+      expect(error.message).toStrictEqual('No response from RPC');
+    });
+
+    const delay = (ms) => new Promise((res) => setTimeout(res, ms));
+
+    await delay(11000);
+  }, 20000);
```
jpuri, line 145: 👍

ui/index.js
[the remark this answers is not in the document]
jpuri: I am not sure if `memoize` is very useful here as this is done only once.

jpuri (dismissed):
👍  great work, tested locally and works fine

app/scripts/lib/metaRPCClientFactory.js
```diff
@@ -10,6 +11,21 @@ class MetaRPCClient {
...
+    this.responseHandled = {};
+  }
+
+  send(id, payload, cb) {
+    this.requests.set(id, cb);
+    this.connectionStream.write(payload);
+    this.responseHandled[id] = false;
+    setTimeout(() => {
+      if (!this.responseHandled[id] && cb) {
+        return cb(new Error('No response from RPC'), null);
```
mcmire, line 23: This is interesting. You're inside of an arrow function, which will get executed asynchronously, after the `send` function completes. Judging by the comment below, maybe you need to give the `send` method a return type of `void`? Then maybe you wouldn't get a linter error.

app/scripts/lib/metaRPCClientFactory.test.js
```diff
@@ -131,4 +131,16 @@ describe('metaRPCClientFactory', () => {
...
+  it('should be able to handle no message within TIMEOUT secs', async () => {
+    const streamTest = createThoughStream();
+    const metaRPCClient = metaRPCClientFactory(streamTest);
+    metaRPCClient.foo('bad', (error, _) => {
+      expect(error.message).toStrictEqual('No response from RPC');
+    });
+
+    const delay = (ms) => new Promise((res) => setTimeout(res, ms));
+
+    await delay(11000);
```
mcmire, line 144: Can we use Jest and/or Sinon's fake timers feature to fast-forward through time instead of waiting 11 seconds?

ui/helpers/utils/error-utils.test.js
```diff
@@ -0,0 +1,47 @@
+import { SUPPORT_LINK } from '../constants/common';
+import { getErrorHtml } from './error-utils';
+
+const defaultState = {
```
mcmire, line 4: What do you think about putting these strings in the test itself? It seems like it'd be more readable. There's also not really a concept of "state" here so the variable seems misleading.

app/scripts/lib/metaRPCClientFactory.test.js
mcmire, line 144: Instead of putting an `expect` in a callback (which may never get called, thus creating a false positive), what do you think about creating a promise that gets resolved when this callback is called, then awaiting that promise? Something like:

``` suggestion
    const promise = new Promise((resolve) => {
      metaRPCClient.foo('bad', (error) => {
        resolve(error);
      });
    });

    const delay = (ms) => new Promise((res) => setTimeout(res, ms));

    await delay(11000);

    const error = await promiseForError;
    expect(error.message).toStrictEqual('No response from RPC');
```

app/_locales/en/messages.json
```diff
@@ -2636,6 +2636,9 @@
   "resetWalletWarning": {
     "message": "Make sure you’re using the correct Secret Recovery Phrase before proceeding. You will not be able to undo this."
   },
+  "restartMetamask": {
+    "message": "Restart Metamask"
```
mcmire, line 2640:
``` suggestion
    "message": "Restart MetaMask"
```

app/popup.html
```diff
@@ -8,7 +8,10 @@
     <link rel="stylesheet" type="text/css" href="./index-rtl.css" title="rtl" disabled>
   </head>
   <body style="width:357px; height:600px;">
-    <div id="app-content"></div>
+    <div id="app-content">
+      <img class="loading-logo" src="./images/logo/metamask-fox.svg" alt="" />
```
mcmire, line 12: I think it would liven up the experience if this were the 3D, interactive fox that appears on the login screen, but that can be tweaked later.

app/scripts/ui.js
mcmire: Nice that you were able to figure this out, thanks for that. The one thing about this button is that when MetaMask reloads, the extension closes, and then nothing happens. I know you can click on the popup to open MetaMask again but I think people might wait for the popup to open again. Do we know if there is a way to automatically do that? If this is something that can be added later just to get this out the door, that's fine.

mcmire (commented):
Getting close! A few more comments.

app/popup.html
mcmire: Actually, besides this, one thing I am noticing is that in the usual case when no error occurs upon initialization, when you open the popup, the loading screen briefly flashes on the screen before showing the home screen. I feel like we could tighten this up a bit. What do you think about wrapping these two divs in another div with a class of `hidden` (not sure if this is defined anywhere, but it would be `display: none`), and adding a `setTimeout` which will wait 1 second or so before removing the `hidden` class? This should remove that flash.

app/scripts/lib/metaRPCClientFactory.js
```diff
@@ -10,6 +11,21 @@ class MetaRPCClient {
...
+      }
+
```
shanejonas, line 25: we should clean up `this.responseHandled[id]` and in the error case.

```suggestion
      delete this.responseHandled[id];
```

georgewrmarshall (commented):
UI looks great! Looks like there are some further questions that need answering. Will leave my approval as just a comment. UI Approved ✅ 💯

app/popup.html
segun: @mcmire can we do this in another PR, so we can involve the design team too. The current behaviour is that the text "Loading" briefly flashes....this inherits that behaviour and would be nice to resolve it properly.

app/scripts/lib/metaRPCClientFactory.test.js
segun: @mcmire I see this pattern used in this test (expect in callback). I'm torn between fixing all the occurrences or creating a new PR to fix them...I'm leaning towards creating a new PR to fix them all. Thoughts?

ui/helpers/utils/error-utils.test.js
segun: Yeah, I noticed that `jest.mock` has to be called before the test is initialised or I get a weird error that locale can't be initialised because only absolute path is supported (calling jest.mock inside a beforeAll or directly inside describe give this error). 

I renamed the variable.

app/scripts/lib/metaRPCClientFactory.test.js
segun: Using fake timers, I can only check that setTimeout was called. I can't check that the callback is called with the error.

If I use fakeTimer I get this error
```
    TypeError: cb is not a function

      21 |     setTimeout(() => {
      22 |       if (!this.responseHandled[id] && cb) {
    > 23 |         return cb(new Error('No response from RPC'), null);
         |                ^
      24 |       }
```

If I console-log cb before setTimeout, it's properly set, but with fakeTimer, by the time the body of settimeout function is being called, it's somehow no longer in scope.

app/scripts/lib/metaRPCClientFactory.js
segun: The error is actually within the setTimeout function

`Expected to return a value at the end of arrow function`

ui/index.js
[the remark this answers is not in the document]
segun: @jpuri in case of an error, it's done twice.

app/scripts/ui.js
segun: I'm thinking to have one or two extra PRs to fix some of this issue that are not necessarily blockers for this PR.

app/scripts/lib/metaRPCClientFactory.js
segun: Fixed [here](https://github.com/MetaMask/metamask-extension/pull/14461/commits/a0f0705168558eae4532761b74529f6c01b0e086)

app/popup.html
mcmire: Ah ok no problem!

app/scripts/lib/metaRPCClientFactory.js
mcmire: Ah, interesting. No worries then.

app/scripts/lib/metaRPCClientFactory.test.js
mcmire: Actually I wonder if we could just change this one test. My main concern is that by adding a delay like this you're adding 11s to the time it takes to run the test suite. Considering that the test suite already takes quite a bit of time to run anything we can do to keep it from growing significantly would help in the long run I think.

app/scripts/ui.js
mcmire: Ok no problem.

mcmire (dismissed):
LGTM!

georgewrmarshall (dismissed):
UI LGTM! Nice work.
- verified in expanded view and action/popover view

Screen Shot 2022-06-03 at 8 31 30 AM
Screen Shot 2022-06-03 at 8 31 53 AM

darkwing (dismissed):
Wow @segun , this is excellent!  Great work!

ui/index.js
danjm: oh, I understand now, thanks

shanejonas (approved):

danjm (approved):

PR #14465, opened 2022-04-18
call controller methods directly in send duck

brad-decker:
## Explanation
There is no need for us to call provider methods in our send duck to create transactions, we already generate the txParams.data for transactions but strip it from the call to these provider methods so that the provider methods generate them for us. However the logic we use for generating the txParams.data in the send duck is the same as the logic these methods invoke.

This change will allow us to get a handle on the created transactionMeta and update it when necessary. We don't do that yet, but in a future PR I will add additional logging to the transaction history to show the steps that a user took to create the transaction. This will help us further debug transactions that are flagged in support tickets. 

## Manual testing steps
1. Test whether sending ETH/native asset works appropriately
2. Test whether sending ERC20s work appropriately
3. Test whether sending NFTs work appropriately

### DIFF
--- a/app/scripts/controllers/transactions/index.js
+++ b/app/scripts/controllers/transactions/index.js
@@ -50,6 +50,7 @@ import {
   determineTransactionType,
   isEIP1559Transaction,
 } from '../../../../shared/modules/transaction.utils';
+import { ORIGIN_METAMASK } from '../../../../shared/constants/app';
 import TransactionStateManager from './tx-state-manager';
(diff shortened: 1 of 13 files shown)


digiwand (dismissed):
🙌  Nice cleanup

brad-decker:
@digiwand used the constant for origin. Thanks. @darkwing can I get a re-review

darkwing (approved):
Code looks good to me and I was able to send a native ETH and ERC-20!

segun (approved):
LGTM

PR #14470, opened 2022-04-19
Run main, flask and beta in sequence in generate-lavamoat-policies.sh

danjm:
Aims to fix the build failures in https://github.com/MetaMask/metamask-extension/pull/14254

It seems that recent changes have caused memory usage issues in our CI builds. Likely introduced in this commit https://github.com/MetaMask/metamask-extension/pull/14254/commits/4447727eb62f72943caf0ab1dfb80bfc17e4a7fa

As a temporary workaround, this PR reduces the number of memory intensive processes running in parallel, specifically for the job that is failing on the v10.13.0 branch


Gudahtt (approved):
LGTM!

darkwing (approved):

PR #14479, opened 2022-04-19
increase test coverage of nonce sorted transaction selector

brad-decker:
## Explanation
This is part one of attempting to cover the nonce sorted transaction selector with tests that make it less brittle to future changes aimed at resolving grouping problems. I broke this into its own file as I expect it to become quite verbose and hopefully some future project will have us moving this logic and potentially changing it. 

While doing this work i realized that in the case of a multiple retry/cancel chain the group could wind up with a dropped retry (or cancel). In that case the dropped transaction would not mark the group accordingly (hasRetried/hasCanceled) porentially allowing a retry of a transaction that was canceled which doesn't seem like something we want.

### DIFF
--- a/ui/selectors/nonce-sorted-transactions-selector.test.js
+++ b/ui/selectors/nonce-sorted-transactions-selector.test.js
@@ -1,0 +1,214 @@
+import { head, last } from 'lodash';
+import { MAINNET_CHAIN_ID } from '../../shared/constants/network';
+import {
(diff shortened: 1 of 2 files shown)


jpuri (approved):

digiwand (approved):
Looks good to me! 

Can we update the PR description to mention that we are enhancing the dropped transactions logic here as well?

PR #14482, opened 2022-04-20
fix: send tokens with decimals > 20

think-in-universe:
## Explanation

Now MetaMask cannot send tokens with decimals larger than 20, this fix allows MetaMask to send tokens with tokens whose decimals <= 32.

## More information

Fixes: #14481

## Manual testing steps

Please test following the reproduce steps in #14481

### DIFF
--- a/shared/modules/conversion.utils.js
+++ b/shared/modules/conversion.utils.js
@@ -26,6 +26,10 @@ import BigNumber from 'bignumber.js';
 
 import { stripHexPrefix, BN } from 'ethereumjs-util';
 
+BigNumber.config({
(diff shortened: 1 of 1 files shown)


think-in-universe:

jpuri:
I do not think this code change will be sufficient to support `<=32` decimal places. It will require larger application wide change and much test coverage to support more decimal places.

brad-decker:
@think-in-universe thank you for your contribution. As per @jpuri 's comment on April 28th this is subject to a lot of other fundamental changes in the extension. If you'd like to pursue those changes we would welcome your contribution.

PR #14485, opened 2022-04-20
Stop storing request and response objects in the permission activity log

rekmarks:
We currently store the JSON-RPC request and response objects in the permission activity log. The utility of doing this was always rather dubious, but never problematic. Until now.

In Flask, as the restricted methods have expanded in number, user secrets may be included on JSON-RPC message objects. This PR removes these properties from the permission activity log, and adds a migration which does the same to existing log objects. We don't interact with the log objects anywhere in our codebase, but we don't want unexpected properties to cause errors in the future should any log objects be retained.

This PR also updates relevant tests and test data. It makes a minor functional change to how a request is designated as a success or failure, but this should not change any behavior in practice.


app/scripts/controllers/permissions/permission-log.js
```diff
@@ -181,9 +178,8 @@ export class PermissionLogController {
       return;
     }
 
-    entry.response = stringify(response, null, 2);
     entry.responseTime = time;
-    entry.success = !response.error;
+    entry.success = Reflect.hasOwnProperty.call(response, 'result');
```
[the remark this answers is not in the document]
ritave: This is not obvious in the code - please add a comment explaining

app/scripts/controllers/permissions/permission-log.test.js
```diff
@@ -1,6 +1,5 @@
 import nanoid from 'nanoid';
```
ritave, line 1: Please add a regression test confirming that there are specifically no `request` and `response` properties.

ritave (changes requested):

app/scripts/migrations/070.test.js
```diff
@@ -0,0 +1,250 @@
...
+              responseTime: 1620710712075,
+              success: true,
+            },
+          ],
+        },
+      },
+    });
+  });
+
+  it('should handle missing PermissionLogController', async () => {
```
ritave, line 230: Should also handle missing `PermissionLogController.permissionActivityLog`

ritave (dismissed):
Minor change, don't need to see afterwards

ritave (approved):

brad-decker (approved):

PR #14489, opened 2022-04-20
Feature/remove bitmask

rickycodes:
I recently discovered that the bitmask that was here had the unintentional effect of [skipping some high and critical vulnerabilities in mobile](https://github.com/MetaMask/metamask-mobile/pull/4134).

This PR makes the same changes here.

Thankfully, there weren't any vulnerabilities being unintentionally skipped here.


Gudahtt (approved):
LGTM!

brad-decker (approved):

PR #14495, opened 2022-04-21
MetaMetrics: add 'number_of_nfts' user trait

### LINKED ISSUE #13477: Add event metrics on NFT support in extension

digiwand:
## Explanation
#13477 Task:
> Add new user trait for 'Number of NFTs'

New user trait: `number_of_nfts`

This involves a slight redesign to the code for `number_of_nft_collections`

## More information

Progresses #13477
Relates to #https://github.com/MetaMask/metamask-extension/pull/14377

### DIFF
--- a/app/scripts/controllers/metametrics.js
+++ b/app/scripts/controllers/metametrics.js
@@ -540,9 +540,7 @@ export default class MetaMetricsController {
    * @returns {MetaMetricsTraits | null} traits that have changed since last update
    */
   _buildUserTraitsObject(metamaskState) {
-    /**
-     * @type {MetaMetricsTraits}
-     */
+    /** @type {MetaMetricsTraits} */
     const currentTraits = {
       [TRAITS.ADDRESS_BOOK_ENTRIES]: sum(
         Object.values(metamaskState.addressBook).map(size),
@@ -554,9 +552,12 @@ export default class MetaMetricsController {
       [TRAITS.NFT_AUTODETECTION_ENABLED]: metamaskState.useCollectibleDetection,
       [TRAITS.NUMBER_OF_ACCOUNTS]: Object.values(metamaskState.identities)
         .length,
-      [TRAITS.NUMBER_OF_NFT_COLLECTIONS]: this._getNumberOfNFtCollection(
-        metamaskState,
+      [TRAITS.NUMBER_OF_NFT_COLLECTIONS]: this._getAllUniqueNFTAddressesLength(
+        metamaskState.allCollectibles,
       ),
+      [TRAITS.NUMBER_OF_NFTS]: this._getAllNFTAddresses(
+        metamaskState.allCollectibles,
+      ).length,
       [TRAITS.NUMBER_OF_TOKENS]: this._getNumberOfTokens(metamaskState),
       [TRAITS.OPENSEA_API_ENABLED]: metamaskState.openSeaEnabled,
       [TRAITS.THREE_BOX_ENABLED]: metamaskState.threeBoxSyncingAllowed,
@@ -605,20 +606,26 @@ export default class MetaMetricsController {
   /**
    *
    * @param {object} metamaskState
-   * @returns number of unique collectible addresses
+   * @param allCollectibles
+   * @returns {[]} Array of all collectible/NFT addresses
    */
-  _getNumberOfNFtCollection(metamaskState) {
-    const { allCollectibles } = metamaskState;
-    if (!allCollectibles) {
-      return 0;
-    }
-
-    const allAddresses = Object.values(allCollectibles)
+  _getAllNFTAddresses(allCollectibles = {}) {
+    return Object.values(allCollectibles)
       .flatMap((chainCollectibles) => Object.values(chainCollectibles))
       .flat()
       .map((collectible) => collectible.address);
-    const unique = [...new Set(allAddresses)];
-    return unique.length;
+  }
+
+  /**
+   *
+   * @param {object} metamaskState
+   * @param allCollectibles
+   * @returns {[]} Array of unique collectible/NFT addresses
+   */
+  _getAllUniqueNFTAddressesLength(allCollectibles = {}) {
+    const allAddresses = this._getAllNFTAddresses(allCollectibles);
+    const uniqueAddresses = new Set(allAddresses);
+    return uniqueAddresses.size;
   }
 
   /**
--- a/app/scripts/controllers/metametrics.js
+++ b/app/scripts/controllers/metametrics.js
@@ -564,9 +564,8 @@ export default class MetaMetricsController {
       [TRAITS.NUMBER_OF_NFT_COLLECTIONS]: this._getAllUniqueNFTAddressesLength(
         metamaskState.allCollectibles,
       ),
-      [TRAITS.NUMBER_OF_NFTS]: this._getAllNFTAddresses(
-        metamaskState.allCollectibles,
-      ).length,
+      [TRAITS.NUMBER_OF_NFTS]: this._getAllNFTs(metamaskState.allCollectibles)
+        .length,
       [TRAITS.NUMBER_OF_TOKENS]: this._getNumberOfTokens(metamaskState),
       [TRAITS.OPENSEA_API_ENABLED]: metamaskState.openSeaEnabled,
       [TRAITS.THREE_BOX_ENABLED]: metamaskState.threeBoxSyncingAllowed,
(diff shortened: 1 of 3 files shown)


app/scripts/controllers/metametrics.js
```diff
@@ -603,22 +613,30 @@ export default class MetaMetricsController {
...
+   * possesses across all networks and accounts.
+   *
+   * @param {Object} allCollectibles
+   * @returns {number}
+   */
+  _getAllUniqueNFTAddressesLength(allCollectibles = {}) {
+    const allAddresses = this._getAllNFTAddresses(allCollectibles);
+    const uniqueAddresses = new Set(allAddresses);
+    return uniqueAddresses.size;
   }
```
brad-decker, line 640: I don't think this is correct...A user may have multiple nfts from the same collection (address) @adonesky1 is that correct?

brad-decker (changes requested):

app/scripts/controllers/metametrics.js
adonesky1: Happy to jump on a call @digiwand and clarify these designations if that would help!

digiwand: Makes sense. I think this code is still correct. Maybe there is confusion because updates to both `number_of_nft_collections` and `number_of_nfts` are made in this PR. When I created this logic, I was going off of 

> All collectibles has this shape:

> This is an object whose keys are accounts and the values are objects where the keys are chainIds. So, for example, in order to get an aggregate number of NFTs imported you'll have to go through each chainId in each account in this object and sum the length of the Collectibles array.

> In order to get the number of collections you'll need to do the same sort of thing but you'll want the number of unique 'address' properties on the entries. Each collectible has an address property that indicates which collection it is a part of.

from the description in #13477. What you two mentioned sounds aligned with this description, and I think the code has been designed for this. 

This particular method called out here is used to compute the value for `number_of_nft_collections`, while `number_of_nfts` is using 
```
this._getAllNFTAddresses(
    metamaskState.allCollectibles,
).length,
```

After mentioning this, do you two still think we need to update the code? Maybe I'm missing something. Thank you! I also love hopping on calls with you two to pick your brains! 😁  

@brad-decker @adonesky1

brad-decker: Sorry @digiwand -- i misread the code changes. So this would increment for each address regardless of whether it was unique. I conflated the changes with collections here.

brad-decker (dismissed):

app/scripts/controllers/metametrics.js
adonesky1: Ah ok. Yes this is correct. Though I think the method _getAllNFTAddresses is a bit of a confusing abstraction even if it is technically working to help get these two traits. I think maybe instead this could be called _getAllNFTs and it could be an array of tuples [address, tokenId], and then _getAllUniqueNFTAddressesLength would filter it down to an array of unique addresses in this array of tuples. Does that make sense?

jpuri (dismissed):

app/scripts/controllers/metametrics.js
digiwand: @brad-decker np! easy mix-up. I think it could have helped if I mentioned this in the description too. Updated the description 👍  

@adonesky1 I like this! Thanks. Update the code with your suggestion here: https://github.com/MetaMask/metamask-extension/pull/14495/commits/57b2bc62fe8891bcb753b1c5aeae35b90188dcc2

app/scripts/controllers/metametrics.js
```diff
@@ -603,22 +612,31 @@ export default class MetaMetricsController {
...
    */
-  _getNumberOfNFtCollection(metamaskState) {
-    const { allCollectibles } = metamaskState;
-    if (!allCollectibles) {
-      return 0;
-    }
+  _getAllNFTs = memoize((allCollectibles = {}) => {
+    return Object.values(allCollectibles)
+      .flatMap((chainNFTs) => Object.values(chainNFTs))
+      .flat();
```
adonesky1, line 624: Is `flat` a `lodash` method? I'm seeing `flatten` in their docs?

adonesky1: Oh it's a native `array` method nvm!

adonesky1 (dismissed):
Great work! Left one more nit comment. But not blocking

adonesky1 (approved):
LGTM! Great Work!

app/scripts/controllers/metametrics.js
```diff
@@ -603,22 +612,31 @@ export default class MetaMetricsController {
...
-   * @returns number of unique collectible addresses
+   * @param {Object} allCollectibles
+   * @returns {[]}
    */
-  _getNumberOfNFtCollection(metamaskState) {
-    const { allCollectibles } = metamaskState;
-    if (!allCollectibles) {
-      return 0;
-    }
+  _getAllNFTs = memoize((allCollectibles = {}) => {
```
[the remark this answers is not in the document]
digiwand: Thanks for hopping on a call with me to explain this! Updated the name to `_getAllNFTsFlattened`. Let me know if you feel like `_getAllNFTsFlat` or another name would suit this better. 

For documentation purposes: 
We had a discussion to deprecate `collection` in favor of `nft` in Slack: https://consensys.slack.com/archives/GTQAGKY5V/p1649274404848719

Which means, `allCollectibles` will be converted to `allNFTs` in the future. Changing `_getAllNFTs` to `_getAllNFTsFlattened` or `_getAllNFTsFlat` will help distinguish the return value from the `allCollectibles`/`allNFTs` state name.

brad-decker (approved):

Issue #14424, opened 2022-04-12
[Bug]: Metamask not displaying all the Ledger accounts available and displaying a repeated account


seaona:
### Describe the bug

**Problem**: seems that there is an incorrect behaviour on how Metamask is finding/displaying the Ledger available accounts, when importing more than 1 Ledger Addresses.
One address appears duplicated, while the other does no appear

Maybe this might cause other open issues related to Ledger (keyring not detected, wrong address etc) - to investigate

https://user-images.githubusercontent.com/54408225/162943441-65c2fe72-e5c9-4bd8-a672-023ff5438623.mp4

### Steps to reproduce

1. Unlock Metamask
2. Import a couple of Ledger accounts (more than 1)
3. Click on import Hardware Wallet again
4. See the available addresses on the list 

### Version

10.13.0

### Browser

Firefox

### Operating system

Linux

Mosillino1:
Hi 
Can anyone help me with "not well-formed currency" error on metamask?
When I try to sell, send and or buy etc. I face this error attached.

Please help me all my cryptos and tokens are in my metamask and I can not do anything with them.
![Screenshot-4221378](https://user-images.githubusercontent.com/104128100/164400338-02d3b84e-9788-4782-9f34-3465597a6f10.jpg)
![Screenshot-4221378](https://user-images.githubusercontent.com/104128100/164402438-5c73c366-3cdf-4d48-9141-9f8f076e352f.jpg)

anaamolnar:
@Mosillino1 Are you able to clear the error after restarting the App or clicking Try again? What currency conversion are you using?

anaamolnar:
@Mosillino1 You should be able to clear the error after restarting the app. After that, please go to Settings->General and change the currency conversion to another currency such as USD or EUR.

Gudahtt:
@Mosillino1 That error has nothing to do with this issue. This issue is about a bug on the extension, but you have posted about an error on MetaMask mobile. Please contact our support team here: https://metamask.zendesk.com/hc/en-us/

Mosillino1:
Hi
How do I clear the error? I press try again but nothing happens. I have
USDC and ETH. I can not send them or do anything else. I wanted to
reinstall the app, but I thought a better idea to install it the app first
if I had the same assets and there was no problem with that, then I delete
previous one. But when I installed it on another device, I didn't have my
assets on the new app. I only have my assets on the previous, but I can't
[...]

PR #14508, opened 2022-04-22
Fix issue with legacy ledger account list

### LINKED ISSUE #6012: Ledger Nano: Metamask Address does not correspond to Ledger
**Describe the bug**
Somehow a Metamask account was created in association with a Ledger Nano that did not correspond to any account supported by that ledger. 

**To Reproduce**
We have not been able to reproduce this issue reliably, but here are the steps we took to encounter it in the first place:
1. Connect a ledger nano, unlock it, and navigate to the Ethereum app on the Ledger
2. In Metamask, select "Connect Hardware Wallet" from the accounts menu
3. Select Ledger and Connect
4. Select an account and click "Unlock"
5. Click the copy button next to the address to copy the wallet address.
6. Send ETH to that address.
7. Reboot, disconnect the ledger, come back later.
8. Reconnect the ledger nano, unlock it, etc.
9. Attempt to send a transaction from the Metamask account. 
10. Get an error similar to the effects of #6011 - Metamask is sending from a different address than the attached ledger.
11. If you attempt to import another account from the ledger, the original address is not available.

**Expected behavior**
If I connect an account from a ledger nano, I expect that account / address to exist on the ledger nano.

 - OS: Windows 10
 - Browser Chrome
 - MetaMask Version 5.3.0

**Additional context**
I'm not certain how this happened. My understanding is that when a ledger is connected, Metamask asks it for a list of addresses, and those addresses are provided by the nano without any proof that it can sign messages with it. I suspect that some kind of interference (static electricity, gamma rays, a butterfly flapping its wings in China, etc.) caused the initial transmission of the address to be sent incorrectly, but it's hard to say for sure. 

Ledger Live does not allow a user to connect an account without first confirming the address by clicking a button on the ledger (I'm not sure how this works under the hood, but I have the impression it's signing a message). It strikes me that this could have been avoided if metamask had required the user to sign a message before considering the account connected. In our case we lost about $2 worth of ETH, but it could have been a lot worse.

jpuri:
Fixes: https://github.com/MetaMask/metamask-extension/issues/6012

For ledger legacy the account list can get out of sync:
Screenshot 2022-04-23 at 12 32 04 AM

User can thus end up sending funds to wrong account.

I found that the reason was that we save the selected ledger type in app state:
Screenshot 2022-04-23 at 12 36 37 AM
Thus though the account list was for `Ledger Live` the dropdown was showing `Ledger (Mew / MyCrypto)`. PR fixes the issue by resetting state when connect hardware component un-mounts.


jpuri:
Hey @seaona : can you plz test this fix.

jpuri:
I am not sure why we are saving `defaultHdPaths` in app state. I could not find it being used anywhere except this component. A better approach may be to refactor use of `defaultHdPaths` and move it to this component rather.

seaona:
@jpuri the fix is working for me.
This PR also fixes this issue https://github.com/MetaMask/metamask-extension/issues/14424

seaona:
**Observations**: at the moment,when you import some Hardware Wallet accounts, and later on import some more, the first ones disappear from the Metamask wallet. This happens with Trezor, Ledger and combined too.

**Steps**:
1. Import a couple of Trezor/Ledger accounts and accept
2. Import again 2 other Trezor/Ledger accounts and accept.

**Result**: the first imported accounts have disappeared from Metamask wallet.

https://user-images.githubusercontent.com/54408225/165470427-26dd8686-cd4e-4921-b305-8a9fcf452548.mp4

Maybe that was the reason for keeping the "state" @jpuri ?

danjm:
Adding to @seaona's most recent comment, I think the default path in state aimed to ensure that if changing the path, the new selection would be the selection upon later returning to the flow

Using ledger I am able to generate at least 4 different account lists. I want to understand exactly what inputs are producing those

jpuri:
@danjm: I see in that case when we return to connect hardware flow we should try to get accounts for saved default path. Yep not sure what can cause the 4 different lists of accounts.

@seaona : I expect [this](https://github.com/MetaMask/metamask-extension/pull/14508#issuecomment-1110681913) behaviour to be in current develop branch also.

jpuri:
@seaona : the issue mentioned [here](https://github.com/MetaMask/metamask-extension/pull/14508#issuecomment-1110681913) does not look related to code changes in this PR - as in the PR I only reset the dropdown selection.

In fact I do expect metamask to forget account if hardware wallet is not more connected.

seaona:
@jpuri from what I've seen, on develop branch the behaviour is different: the accounts are not removed, when I add new ones. See below:

https://user-images.githubusercontent.com/54408225/165905305-9798dfa2-6135-4861-865a-3e6897721e10.mp4

All this process is without disconnecting the device. Just importing 2 accounts and then importing 2 more.

kumavis:
resolved here https://github.com/MetaMask/eth-ledger-bridge-keyring/pull/146

jpuri:
Issue has been fixed with PR here: https://github.com/MetaMask/eth-ledger-bridge-keyring/pull/146

I will create a separate PR in extension repo to update eth-ledger-bridge-keyring.

Issue #11677, opened 2021-07-29
Extension sends random cached addressed instead of the one entered


roromi13:
**Describe the bug**
1. Used the send function from my wallet to my other wallet
2. Transfer address was checked 2-3 times and was deemed to be correct
3. Wallet history show it was sent to the correct address but opening the linked block explorer transaction, it was sent to a different account

**Steps to reproduce (REQUIRED)**
Steps to reproduce the behavior, libraries used with version number, and/or any setup information to easily reproduce:

1. Go to Metamask extension
2. Click on send
3. Enter recipient
4. Click on Next
5. Click on Confirm

**Expected behavior**
A clear description of what you expected to happen.

**Screenshots**
![image](https://user-images.githubusercontent.com/83640918/127571472-c478a1f4-7e68-4e9c-96bc-bd46a359b6d2.png)
Failed to save the expanded view of this transaction but all values were consistent including the nonceswith the block explorer except the address

Block explorer: https://bscscan.com/tx/0x9766e7d7f6724b82c41a72f289d6bbb08c5adc8607c37447250371cdb6c9f12a

**Browser details (please complete the following information):**
 - OS: Windows 10
 - Hardware Wallet: NA
 - Browser: Chrome
 - MetaMask Version: 9.8.2

**Additional context (Error Messages, etc.)**
Most likely it is related to how the extension handles the caching or idk. Luckily, I knew the owner of the wallet and had my little funds sent back to me. I suspect it is closely related to the "Contacts" feature of the extension

danjm:
Thanks for the report. Can you privately submit your state logs via a customer support ticket?

To download your state logs:
(1) Go to "Settings" in MetaMask
(2) Click "Advanced"
(3) Click "Download state logs"

To submit a support ticket, go to https://consensys.zendesk.com/agent/dashboard, and select "MetaMask Extension" in the first drop you see. In the ticket, just copy a link to this github issue, and attach the state logs json file that you downloaded.

This will help us debug this issue

roromi13:
Not sure if it is still available in state logs, I'll be sending it now

roromi13:
I'm getting access denied in zendesk. Sent it to your gmail instead

dpazdan:
@roromi13 I highly recommend scanning the device for possible clipboard malware. 

You can determine if this is the cause by copying another address, and paste it to see if it changes every time to a random address (a malware address)

roromi13:
[...]

PR #14510, opened 2022-04-22
Track send flow history on txMeta

### LINKED ISSUE #14512: Hibee
https://github.com/github/choosealicense.com/blob/a40ef42140d137770161addf4fefc715709d8ccd/no-permission.md

brad-decker:
## Explanation
progresses #11677 

This pull request adds logs from the user interaction with the send flow to the transaction meta so that future reports of the #11677 bug can be further analyzed and debugged. 

things tracked:
1. input into the recipient field
2. pasting into the recipient field
3. selecting from either contacts list, my accounts or ENS resolution
4. asset changes (type, symbol and contract address)
5. gas changes for legacy transaction UI
6. send amount changes
7. switching between contact list and my accounts on recipient screen
8. clearing the recipient
9. supplying hexData if custom hex data is enabled
10. toggling send max mode
11. editing a transaction
12. Clicking next to either create or update a transaction

```json
        "sendFlowHistory": [
          {
            "entry": "sendFlow - user selected transfer to my accounts on recipient screen",
            "timestamp": 1650663928211
          },
          {
            "entry": "sendFlow - User clicked recipient from my accounts. address: 0x0ccc8aeeaf5ce790f3b448325981a143fdef8848, nickname Account 3",
            "timestamp": 1650663928979
          },
          {
            "entry": "sendFlow - user set amount to 1 ETH",
            "timestamp": 1650663931539
          },
          {
            "entry": "sendFlow - user set amount to 1 ETH",
            "timestamp": 1650663931544
          },
          {
            "entry": "sendFlow - user set asset type to NATIVE",
            "timestamp": 1650663933036
          },
          {
            "entry": "sendFlow - user set asset symbol to undefined",
            "timestamp": 1650663933037
          },
          {
            "entry": "sendFlow - user set asset address to undefined",
            "timestamp": 1650663933038
          },
          {
            "entry": "sendFlow - user set asset type to TOKEN",
            "timestamp": 1650663934108
          },
          {
            "entry": "sendFlow - user set asset symbol to TST",
            "timestamp": 1650663934109
          },
          {
            "entry": "sendFlow - user set asset address to 0xbC130FAFF2798dcCec987DF98Bfdda2A1c4B8470",
            "timestamp": 1650663934110
          },
          {
            "entry": "sendFlow - user set amount to 1 TST",
            "timestamp": 1650663935058
          },
          {
            "entry": "sendFlow - user set amount to 1.5 TST",
            "timestamp": 1650663937578
          },
          {
            "entry": "sendFlow - user set amount to 1.5 TST",
            "timestamp": 1650663938716
          }
        ],
```

### DISCLAIMER
None of these logs are visible to anyone at MetaMask or ConsenSys UNLESS the user explicitly downloads state logs from the MetaMask advanced settings menu and send them to our support team. These are not transmitted to any third parties, analytics, reporting or other tools.


danjm (changes requested):
This looks good, but there is one question I have that needs to be resolved

ui/ducks/send/send.js
```diff
@@ -1638,6 +1707,8 @@ export function useMyAccountsForRecipientSearch() {
  */
 export function updateRecipient({ address, nickname }) {
   return async (dispatch, getState) => {
+    // Do not addHistoryEntry here as this is called from a number of places
```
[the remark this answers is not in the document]
brad-decker: Because the *way* the recipient is added is important and not known here, so in each place this is used in the code I have added logging that specifies how/why the recipient is updated.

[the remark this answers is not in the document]
brad-decker: I'll update the comment to be more specific... Alternatively I could create variations of updateRecipient with specifics to be used downstream.

Je1952:
What is the question all about ?

ui/ducks/send/send.js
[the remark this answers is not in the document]
danjm: oh, yes, I see
I hadn't seen the changes in pages/send/send.js before

ui/pages/send/send.js
```diff
@@ -95,13 +96,23 @@ export default function SendTransactionScreen() {
...
         selectedAddress={recipient.address}
         selectedName={recipient.nickname}
-        onPaste={(text) => updateRecipient({ address: text, nickname: '' })}
+        onPaste={(text) => {
+          dispatch(
+            addHistoryEntry(
+              `sendFlow - User pasted ${text} into address field`,
+            ),
+          );
+          updateRecipient({ address: text, nickname: '' });
```
brad-decker: so.........uh, i'm not sure how this was working before because it wasn't being dispatched. The likely situation is that the onValidAddressTyped handler is also fired on a paste...

ui/ducks/send/send.js
```diff
@@ -1482,6 +1523,19 @@ export function updateSendAmount(amount) {
...
+        `sendFlow - user set asset symbol to ${details?.symbol ?? 'undefined'}`,
+      ),
+    );
+    dispatch(
+      addHistoryEntry(
+        `sendFlow - user set asset address to ${
+          details?.address ?? 'undefined'
+        }`,
+      ),
+    );
```
jpuri, line 1538: Can we improve this method `addHistoryEntry` to take multiple entries ?

ui/ducks/send/send.js
```diff
@@ -1452,8 +1465,36 @@ export function resetSendState() {
...
+        value: amount,
+        toCurrency: ETH,
+        numberOfDecimals: 8,
+      });
+      logAmount = `${ethValue} ${ETH}`;
+    }
+    await dispatch(
+      addHistoryEntry(`sendFlow - user set amount to ${logAmount}`),
+    );
+    await dispatch(actions.updateSendAmount(amount));
```
jpuri, line 1497: Above calculation for `logAmount` looks not required to me, we can save raw data in history log.

ui/ducks/send/send.test.js
```diff
@@ -2419,7 +2554,7 @@ describe('Send Slice', () => {
         nickname: '',
       });
 
-      const action = actionResult[6];
+      const action = actionResult[10];
 
       const result = sendReducer(initialState, action);
```
jpuri, line 2560: Great work with test coverage 👍

ui/ducks/send/send.js
```diff
@@ -1746,6 +1824,12 @@ export function signTransaction() {
           eip1559support ? eip1559OnlyTxParamsToUpdate : txParams,
         ),
       };
+      await dispatch(
+        addHistoryEntry(
+          `sendFlow - user clicked next and transaction should be updated in controller`,
+        ),
+      );
+      await dispatch(updateTransactionSendFlowHistory(id, state[name].history));
```
jpuri, line 1832: Can this rather be an implicit part of signTransaction `updateTransactionSendFlowHistory` so that we do not need to call it from UI.

brad-decker: i'm unclear as to what you mean. If its a new transaction we send it with the transaction params to create the new transaction. If its an existing one (edited) then we call this method.

app/scripts/controllers/transactions/index.js
```diff
@@ -647,6 +647,35 @@ export default class TransactionController extends EventEmitter {
     return this._getTransaction(txId);
   }
```
danjm, line 649: This file looks good

ui/ducks/send/send.js
```diff
@@ -60,6 +60,7 @@ import {
   getTokenStandardAndDetails,
   showModal,
   addUnapprovedTransactionAndRouteToConfirmationPage,
+  updateTransactionSendFlowHistory,
 } from '../../store/actions';
 import { setCustomGasLimit } from '../gas/gas.duck';
 import {
```
danjm, line 66: This file looks correct

danjm (approved):
LGTM

ui/ducks/send/send.js
jpuri: I mean can this be part of other action at the backend may be `addUnapprovedTransactionAndRouteToConfirmationPage`

jpuri (approved):
👍  nice work

ui/ducks/send/send.js
brad-decker: that already is the case for that flow, but we also need the update flow for an edit transaction, and its possible to encapsulate the different update methods into a single action but I was following the pattern in place (updateEditableParams/updateTransactionGasFees)

PR #14531, opened 2022-04-26
snaps-skunkworks@0.11.1

FrederikBolding:
Updates Snaps skunkworks packages to `0.11.1` and adds needed hooks for new features


rekmarks (dismissed):
LGTM!

rekmarks (approved):

package.json
```diff
@@ -435,7 +435,10 @@
       "ganache>keccak": false,
       "ganache>leveldown": false,
       "ganache>secp256k1": false,
-      "ganache>utf-8-validate": false
+      "ganache>utf-8-validate": false,
+      "gulp-watch>chokidar>fsevents": false,
```
mcmire, line 439: Curious how these started showing up now, but `fsevents` ensures that when we are watching files on macOS, we use the native file watcher built into the OS, so we may want to enable these as it might lead to less CPU load there.

mcmire (approved):
One small thing, but looks good anyway.

PR #14536, opened 2022-04-26
Adding option to update lavamoat policies in parallel

ryanml:
Recently on CI, the scripts to update lavamoat policies changed to run sequentially: https://github.com/MetaMask/metamask-extension/pull/14470

This is fine for CI, but can be very slow for local development. This change adds an additional command, `yarn lavamoat:background:auto:dev` to run the jobs in parallel.

### DIFF
--- a/development/generate-lavamoat-policies.sh
+++ b/development/generate-lavamoat-policies.sh
@@ -7,6 +7,21 @@ set -o pipefail
 # Generate LavaMoat policies for the extension background script for each build
 # type.
 # ATTN: This may tax your device when running it locally.
-WRITE_AUTO_POLICY=1 yarn build scripts:prod --policy-only &&
-WRITE_AUTO_POLICY=1 yarn build scripts:prod --policy-only --build-type beta &&
-WRITE_AUTO_POLICY=1 yarn build scripts:prod --policy-only --build-type flask
+
+function run_sequentially() {
+  WRITE_AUTO_POLICY=1 yarn build scripts:prod --policy-only &&
+  WRITE_AUTO_POLICY=1 yarn build scripts:prod --policy-only --build-type beta &&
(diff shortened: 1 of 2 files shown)


kumavis:
I think we can alternatively handle this by just setting concurrently's max processes to 1, that way we can define the jobs in just one place

kumavis:
under "help" https://github.com/open-cli-tools/concurrently#usage

FrederikBolding:
> I think we can alternatively handle this by just setting concurrently's max processes to 1, that way we can define the jobs in just one place

That would be slightly cleaner and if possible preferred.

FrederikBolding:
```
#!/usr/bin/env bash

set -e
set -u
set -o pipefail

extraArgs=()
if [[ $# -lt 1 ]]; then
    extraArgs+=(-m 1)
fi

# Generate LavaMoat policies for the extension background script for each build
# type.
# ATTN: This may tax your device when running it locally.
concurrently --kill-others-on-fail -n main,beta,flask \
    "${extraArgs[@]}" \
    "WRITE_AUTO_POLICY=1 yarn build scripts:prod --policy-only" \
    "WRITE_AUTO_POLICY=1 yarn build scripts:prod --policy-only --build-type beta" \
    "WRITE_AUTO_POLICY=1 yarn build scripts:prod --policy-only --build-type flask"
```

@kumavis @ryanml I'm no bash expert, but I tried this, and it seems that with `-m 1` killing the first process doesn't kill the others correctly 🤔

brad-decker:
I tried @FrederikBolding 's change locally and killing first script for me killed all. I applied your change and pushed. I would like to get this landed because doing dependency updates is very painful with the sequential operation.

brad-decker (approved):

NiranjanaBinoy (approved):

PR #14537, opened 2022-04-26
Lavamoat - protect all UI contexts

kumavis:
replaces https://github.com/MetaMask/metamask-extension/pull/12938
replaces https://github.com/MetaMask/metamask-extension/pull/12959

add lavamoat protections to the ui

### DIFF
--- a/patches/zxcvbn+4.4.2.patch
+++ b/patches/zxcvbn+4.4.2.patch
@@ -1,0 +1,13 @@
+diff --git a/node_modules/zxcvbn/lib/matching.js b/node_modules/zxcvbn/lib/matching.js
+index 3940bad..748da8b 100644
+--- a/node_modules/zxcvbn/lib/matching.js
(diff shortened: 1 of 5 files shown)


brad-decker (approved):
LGTM

patches/zxcvbn+4.4.2.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/zxcvbn/lib/matching.js b/node_modules/zxcvbn/lib/matching.js
```
digiwand, line 1: Q: Do you think it would be helpful for us to add description comments at the top of the patch files?

digiwand (approved):
No blockers. Small question above

patches/zxcvbn+4.4.2.patch
kumavis: yeah thats a good idea, ill review our existing patches

PR #14545, opened 2022-04-27
Fixed API for watchAsset

### LINKED ISSUE #12416: Method "wallet_watchAsset" does't work correctly
Hi,

**Describe the bug**
1. In the **FireFox** browser, I see the problem that this method does not handle the promise. It always hangs in the pending status, even after the user selects one of the options. It is impossible to understand what the user has chosen. If you call the method again, the window for adding a token will no longer open.

2. In the **Chrome** browser, the problem is that immediately after calling the method, an object is returned. Although a boolean value should be returned depending on the user's choice, the object is returned immediately before  the user chooses. Because of this, there is no way to hang up the loader and find out what the user has chosen.

**Steps to reproduce**
1. Open Firefox / Chrome
2. Go to a page with ethereum connected
3. Paste this code in the console:
```
ethereum.request({
  method: 'wallet_watchAsset',
  params: {
    type: 'ERC20',
    options: {
      address: '0xb60e8dd61c5d32be8058bb8eb970870f07233155',
      symbol: 'FOO',
      decimals: 18,
      image: 'https://foo.io/token-image.svg',
    },
  },
})
  .then(
    (result) => console.log('then', result),
    (err) => console.log('catch', err)
  )
```

**Expected behavior**
Different browsers should have the same behavior. The method should return a Promise with a boolean value after the user has added or has not added a token to the MetaMask.

**Screenshots**
Chrome: https://prnt.sc/1wunm6y (No tokens have been added, but the promise has already resolved)
FireFox: https://prnt.sc/1wunwk3 (Tokens added, but the promise is in pending status)

 - OS: macOS
 - Firefox: 93.0 / MetaMask: 10.2.2
 - Chrome: 94.0.4606.81 / MetaMask: 10.2.2

### LINKED ISSUE #14656: Error: User rejected to watch the asset.
Sentry Issue: [METAMASK-MA4Z](https://sentry.io/organizations/metamask/issues/2977475863/?referrer=github_integration)

```
Error: User rejected to watch the asset.
  at call (app/scripts/lib/createMetaRPCHandler.js:22:31)
...
(18 additional frame(s) were not displayed)
```
We should not see this error in metamask. The error should be sent to the dapp

shanejonas:
fixes #12416 
may also fix #14656

This just fixes the API. the controller seems to return a promise for `result` so trying to pass that across the background JSON-RPC errors or has unexpected behaviour on different browsers.

It was also passing `result: {result, suggestedAssetMeta}` to go across the wire when the API should just be result: boolean. (spec: https://eips.ethereum.org/EIPS/eip-747)

This PR fixes it inside the extension. Another approach would be returning something close to the current API from the controller instead.

### DIFF
--- a/app/scripts/lib/rpc-method-middleware/handlers/watch-asset.js
+++ b/app/scripts/lib/rpc-method-middleware/handlers/watch-asset.js
@@ -1,3 +1,4 @@
+import { ethErrors } from 'eth-rpc-errors';
 import { MESSAGE_TYPE } from '../../../../../shared/constants/app';
 
 const watchAsset = {
@@ -36,9 +37,11 @@ async function watchAssetHandler(
 ) {
   try {
     const { options: asset, type } = req.params;
-    res.result = await handleWatchAssetRequest(asset, type);
+    const handleWatchAssetResult = await handleWatchAssetRequest(asset, type);
+    const result = await handleWatchAssetResult.result;
+    res.result = Boolean(result);
     return end();
   } catch (error) {
-    return end(error);
+    return end(ethErrors.provider.userRejectedRequest());
   }
 }
--- a/app/scripts/lib/rpc-method-middleware/handlers/watch-asset.js
+++ b/app/scripts/lib/rpc-method-middleware/handlers/watch-asset.js
@@ -38,10 +38,13 @@ async function watchAssetHandler(
   try {
     const { options: asset, type } = req.params;
     const handleWatchAssetResult = await handleWatchAssetRequest(asset, type);
-    const result = await handleWatchAssetResult.result;
-    res.result = Boolean(result);
+    await handleWatchAssetResult.result;
+    res.result = true;
     return end();
   } catch (error) {
-    return end(ethErrors.provider.userRejectedRequest());
+    if (error.message === 'User rejected to watch the asset.') {
+      return end(ethErrors.provider.userRejectedRequest());
+    }
+    return end(error);
   }
 }


app/scripts/lib/rpc-method-middleware/handlers/watch-asset.js
```diff
@@ -36,9 +37,11 @@ async function watchAssetHandler(
...
   try {
     const { options: asset, type } = req.params;
-    res.result = await handleWatchAssetRequest(asset, type);
+    const handleWatchAssetResult = await handleWatchAssetRequest(asset, type);
+    const result = await handleWatchAssetResult.result;
+    res.result = Boolean(result);
     return end();
   } catch (error) {
-    return end(error);
+    return end(ethErrors.provider.userRejectedRequest());
```
Gudahtt, line 45: We can look for this error specifically. Right now the controller returns an error with the message "User rejected to watch the asset." if the user rejects.

Right now this is effectively swallowing the `error`, hiding any potential unexpected error that might be thrown here. Better to bubble it up as an internal error if we don't recognize it.

app/scripts/lib/rpc-method-middleware/handlers/watch-asset.js
```diff
@@ -36,9 +37,11 @@ async function watchAssetHandler(
 ) {
   try {
     const { options: asset, type } = req.params;
-    res.result = await handleWatchAssetRequest(asset, type);
+    const handleWatchAssetResult = await handleWatchAssetRequest(asset, type);
+    const result = await handleWatchAssetResult.result;
+    res.result = Boolean(result);
```
Gudahtt, line 42: Nit: According to the spec, this should _always_ either return `true` or return an error. Casting the result as a boolean is _almost_ the same thing because the result in this case is the asset address, which should be guaranteed by our validation to be a non-empty string. But why bother obscure the expectation here.

```suggestion
    res.result = true;
```

Gudahtt:
Great catch! I think this is how this method was meant to be used. This must have been overlooked when we incorporated the token controller last year.

Gudahtt (approved):
LGTM!

FrederikBolding (approved):

PR #14548, opened 2022-04-27
Feat/get wallet currency

Gtonizuka:


brad-decker:
@Gtonizuka is this still something you'd like to work on / move towards merging? If so could you update the pr with a description and identify the steps needed to move this from draft? Otherwise feel free to close and keep the branch around for future development.

PR #13960, opened 2022-03-15
Deprecate extensionizer for webextension-polyfill

adonesky1:
Fixes: https://github.com/MetaMask/extensionizer/issues/21

Deprecates our use of our own `extensionizer` package for [`webextension-polyfill`](https://github.com/mozilla/webextension-polyfill) as called for [here](https://github.com/MetaMask/extensionizer/issues/21). 

This work is part of [our ongoing project to support Manifest Version 3](https://www.notion.so/Chrome-Extension-Version-MV3-Migration-19063a3ee9af46c48f18bf8e8f7aa546)

PR #14550, opened 2022-04-27
Fix browser.tab.get call

adonesky1:
Fix bug introduced when we [deprecated our `extensionizer` package for a `webextension-polyfill`
](https://github.com/MetaMask/metamask-extension/pull/13960) where browser API calls are promisified instead of taking callbacks. In this particular instance we were getting a [sentry error](https://sentry.io/organizations/metamask/issues/3219266910/?project=273505&query=is%3Aunresolved+firstRelease%3Alatest&sort=freq&statsPeriod=14d) that the browser api method we were calling didn't expect the extra (callback) argument.

Removes method `returnToOnboardingInitiator` because it is redundant. `returnToOnboardingInitiatorTab` can achieve the same behavior by itself.

### DIFF
--- a/ui/pages/first-time-flow/onboarding-initiator-util.js
+++ b/ui/pages/first-time-flow/onboarding-initiator-util.js
@@ -2,21 +2,8 @@ import browser from 'webextension-polyfill';
 import log from 'loglevel';
 
 const returnToOnboardingInitiatorTab = async (onboardingInitiator) => {
-  const tab = await new Promise((resolve) => {
-    browser.tabs.update(onboardingInitiator.tabId, { active: true }).then(
-      // eslint-disable-next-line no-shadow
-      (tab) => {
-        if (tab) {
-          resolve(tab);
-        } else {
-          // silence console message about unchecked error
-          if (browser.runtime.lastError) {
-            log.debug(browser.runtime.lastError);
-          }
-          resolve();
-        }
-      },
-    );
+  const tab = await browser.tabs.update(onboardingInitiator.tabId, {
+    active: true,
   });
 
   if (tab) {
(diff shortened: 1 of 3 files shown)


ui/pages/first-time-flow/onboarding-initiator-util.js
```diff
@@ -28,29 +13,22 @@ const returnToOnboardingInitiatorTab = async (onboardingInitiator) => {
     log.warn(
       `Setting current tab to onboarding initiator has failed; falling back to redirect`,
     );
+
+    if (browser.runtime.lastError) {
```
Gudahtt, line 17: I thought the polyfill checked `lastError` for us. This was part of how errors were thrown in the Chrome API. I would expect the pollyfill to check this and reject the call with the error if it's present.

adonesky1: Yeah it sounds like you're right:

from https://github.com/mozilla/webextension-polyfill:

>Rather than checking the chrome.runtime.lastError property from every callback, code which needs to explicitly deal with errors registers a separate Promise rejection handler.

adonesky1: updated here: https://github.com/MetaMask/metamask-extension/pull/14550/commits/eb9504738b1b35b383d5c9c2bce708fae2b1cfa4

ui/pages/first-time-flow/onboarding-initiator-util.js
```diff
@@ -1,25 +1,18 @@
...
-          }
-          resolve();
-        }
-      },
+export const returnToOnboardingInitiatorTab = async (onboardingInitiator) => {
+  let tab;
+  try {
+    tab = await browser.tabs.update(onboardingInitiator.id, {
+      active: true,
+      highlighted: true,
```
Gudahtt, line 9: Why was this option added?

adonesky1: Added while I was testing and forgot to remove. It's not necessary. Will remove.

adonesky1: updated here: https://github.com/MetaMask/metamask-extension/pull/14550/commits/2da95286b25184a93cfe9d9691913e48f91aa25d

Gudahtt (approved):
LGTM!

jpuri (approved):

jpuri:
@adonesky1 : changes look good, also I could run onboarding flow smoothly. But I am not sure how should I test this specific scenario - may be its already covered in onboarding.

PR #14565, opened 2022-04-28
LavaMoat - UI upgrade - secure package naming

kumavis:
Upgrades :volcano:  `lavamoat-browserify` :volcano:  as used in the ui, to the latest version that uses secure package naming


development/build/scripts.js
```diff
@@ -399,7 +399,7 @@ function createFactoredBuild({
       policyName: buildType,
       policyOverride: path.resolve(
         __dirname,
-        `../../lavamoat/browserify/${buildType}/policy-override.json`,
+        `../../lavamoat/browserify/policy-override.json`,
```
rekmarks, line 402: Why the move to a single policy override?

kumavis: less work maintaining a single override file

rekmarks: So I guess it doesn't matter if we override modules that exist for one build type but not others?

rekmarks (dismissed):
LGTM.

development/build/scripts.js
kumavis: correct. `lavapack` will even remove policy for packages that dont appear in the build.

brad-decker (dismissed):
LGTM

brad-decker (approved):

PeterYinusa:
LGTM

rekmarks (approved):

PR #14568, opened 2022-04-29
E2e send tests

PeterYinusa:
## Explanation

- Update the fill method in the `driver.js` to use `sendkeys` to clear an input. 
  - This change was made to remove the input from the gas fields as the `clear` method only removes the value temporarily 
- Remove arbitrary delays from send tests

### Before

The value that had been cleared from the input reappears, so the update is not as expected 

https://user-images.githubusercontent.com/53189696/166667966-8d337c7d-6fc8-40df-8b45-c87c50503d75.mov

### After

The value is correctly cleared and updated

https://user-images.githubusercontent.com/53189696/166668001-03f99c5a-5fc8-42de-a991-d23446d6e9b0.mov

### DIFF
--- a/test/e2e/tests/edit-gas-fee.spec.js
+++ b/test/e2e/tests/edit-gas-fee.spec.js
@@ -3,9 +3,7 @@ const {
   convertToHexValue,
   connectDappWithExtensionPopup,
   getWindowHandles,
-  largeDelayMs,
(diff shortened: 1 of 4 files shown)


jpuri (approved):
Nice improvements 👍

darkwing (approved):

PR #14571, opened 2022-04-29
Remove 3box feature and delete ThreeBoxController

rekmarks:
Removes the 3box sync feature and deletes the `ThreeBoxController`. Adds a migration to ensure that the `ThreeBoxController` is removed from persisted state.

This shaves 600kb or 3.2% from our bundle (18.5mb -> 17.9mb).


kumavis:
:+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1:

darkwing (dismissed):
Oddly enough somehow the messages from `app/_locales/zh/messages.json` didn't get removed; I'm still seeing `syncWithThreeBox`, `syncWithThreeBoxDescription`, and `syncWithThreeBoxDisabled` keys.

app/scripts/controllers/metametrics.js
```diff
@@ -569,7 +569,6 @@ export default class MetaMetricsController {
       ).length,
       [TRAITS.NUMBER_OF_TOKENS]: this._getNumberOfTokens(metamaskState),
       [TRAITS.OPENSEA_API_ENABLED]: metamaskState.openSeaEnabled,
-      [TRAITS.THREE_BOX_ENABLED]: metamaskState.threeBoxSyncingAllowed,
```
brad-decker, line 572: Could we, instead, set this to false and put a comment that it was set to false due to 3box removal? In a future PR we can remove it entirely but i'd like to see the user trait change gradually..If we delete this it'll be stuck in 'true' mode for those users who had it enabled before this is rolled out

shared/constants/metametrics.js
```diff
@@ -175,8 +175,6 @@
  * identify the new number_of_tokens trait
  * @property {'opensea_api_enabled'} OPENSEA_API_ENABLED - when the OpenSea API is enabled
  * we identify the opensea_api_enabled trait
- * @property {'three_box_enabled'} THREE_BOX_ENABLED - when 3box feature is
```
brad-decker, line 178: if my suggestion to turn the trait t ofalse instead of removing it is accepted, then keep these constants/types but notate they are deprecated.

brad-decker (dismissed):
Hero. 🙏🏻 Non blocking suggestion for metrics cleanliness.

Gudahtt:
We need a deprecation plan before we can merge this, including messaging to any existing users

darkwing:
Marking is `DO-NOT-MERGE` until we resolve Mark's comments.

kumavis:
@Gudahtt i would claim that we dont need a migration path bc the feature was only ever experimental. i agree that informing users that had it enabled is worth blocking on.

app/scripts/controllers/metametrics.js
Gudahtt: Good idea! Done here: 46ab5a404450c4ec7b893afb5aa8fb6af42109cd

shared/constants/metametrics.js
Gudahtt: Done: 3c3475bf20f1a2ac4445b30bc743632ca2969f23

legobeat:
I assume https://github.com/MetaMask/metamask-extension/blob/794395dc48842385f9eecf5e3c2ae162b578d721/test/e2e/fixture-builder.js#L596 should go as well?  That'd be the last lines AFAICT.

danjm (approved):
LGTM

Gtonizuka (approved):

PR #14583, opened 2022-04-29
Add applyLavaMoat build flag

rekmarks:
Adds a new flag, `--apply-lavamoat`, to the main build script. The flag controls whether LavaMoat is actually applied to the output of the build process. The flag defaults to `true`, but we explicitly set it to `false` in the `start` package script. Meanwhile, the `start:lavamoat` script is modified such that it applies LavaMoat to the build output in development mode, but it no longer runs the build process itself under LavaMoat as there aren't very compelling reasons to do so.

This change is motivated by the fact that development builds do not have their own dedicated LavaMoat policies, which causes development builds to fail since #14537. The downside of this change is that LavaMoat-related failures will not be detected when running `yarn start`. @kumavis has plans for fixing this problem in a future major version of the `@lavamoat` suite.

### Testing

It's important that we manually test this in order to verify that the flags work correctly and LavaMoat is in fact applied when `yarn dist` is run, and when `yarn start:lavamoat` is run.


app/background.html
```diff
@@ -6,7 +6,7 @@
   <body>
     <script src="./globalthis.js" type="text/javascript" charset="utf-8"></script>
     <script src="./sentry-install.js" type="text/javascript" charset="utf-8"></script>
-    {{@if(it.useLavamoat)}}
+    {{@if(it.applyLavaMoat)}}
```
rekmarks, line 9: We liked `applyLavaMoat` better than `useLavamoat`, which was improperly cased in any event.

kumavis (approved):

adonesky1 (approved):

PR #14594, opened 2022-05-02
Fix development build scripts

rekmarks:
#14583 broke the development build scripts (e.g. `yarn start`) by adding a positional argument to a package script (`build:dev`) that is used and passed positional arguments in the build script itself. This PR removes the positional argument from the `build:dev` script and `yarn start` now works again. In addition, the `--apply-lavamoat` flag is properly forwarded to child processes, which was not the case in the original implementation.

To test, `yarn start` should work and LavaMoat should _not_ be applied, in distinction to `yarn build:dev dev --apply-lavamoat=true`. Whether LavaMoat is applied can be determined by checking whether `Object.isFrozen(Object.prototype)` is `true` (with LavaMoat) or `false` (without LavaMoat).


jpuri (dismissed):

brad-decker (dismissed):

jpuri:
Tested by running `yarn start` locally and works well.

dan437 (approved):
Awesome, `yarn start` works now and there is no white screen after opening the extension. Thank you!

brad-decker:
@rekmarks @dan437 rebasing on develop now that the fix for the e2e test failure has landed

brad-decker (approved):

PR #14598, opened 2022-05-02
Stop using 4bytes for contract deployment

FrederikBolding:
## Explanation
This PR fixes a bug where we would try to show called method names for contract deployments. Since contract deployments have the bytecode as the data we can't look up method signatures using the first 4 bytes as you can with a normal contract interaction.

## Manual Testing Steps

1. Go to the test dapp
2. Deploy a contract
3. See that the transaction title is "contract deployment"

## Pre-Merge Checklist


kumavis (dismissed):

jpuri:
Can you plz add some description to the PR and if possible test coverage.

ui/hooks/useTransactionDisplayData.js
```diff
@@ -217,17 +217,20 @@ export function useTransactionDisplayData(transactionGroup) {
...
-  } else if (
-    type === TRANSACTION_TYPES.DEPLOY_CONTRACT ||
-    type === TRANSACTION_TYPES.CONTRACT_INTERACTION
-  ) {
+  } else if (type === TRANSACTION_TYPES.CONTRACT_INTERACTION) {
     category = TRANSACTION_GROUP_CATEGORIES.INTERACTION;
     const transactionTypeTitle = getTransactionTypeTitle(t, type);
     title =
       (methodData?.name && camelCaseToCapitalize(methodData.name)) ||
       transactionTypeTitle;
```
adonesky1, line 225: Do we ever add `methodData` to a transaction where the `to` address is `null`/`undefined`?

FrederikBolding: We shouldn't. We should only use it for valid method calls to a contract.

FrederikBolding:
> Can you plz add some description to the PR and if possible test coverage.

Added a description. I would add test coverage, but I don't see any fixtures for it? 🤔

brad-decker (approved):
LGTM

ui/hooks/useTransactionDisplayData.js
adonesky1: I feel like ideally we would guarantee that methodData won't be attached to a `transaction` where address is empty or zero address further upstream (maybe in the `TransactionsController`), but won't block on it here...

adonesky1: I feel like ideally we would guarantee that methodData won't be attached to a transaction where address is empty or zero address further upstream (maybe in the TransactionsController), but won't block on it here...

adonesky1 (approved):

ui/hooks/useTransactionDisplayData.js
brad-decker: in what scenario would we have a transaction without a 'to' address?

adonesky1: I think when you deploy a contract no?

brad-decker: 🤦🏻 yeah. that makes sense...

brad-decker: We don't attach methoData anywhere, its fetched at that top of the useTransactionDisplayData hook.

FrederikBolding: Yeah, it's not affiliated with the transaction outside of this hook afaik @adonesky1

PR #14610, opened 2022-05-04
build - update bify-module-groups for build determinism

kumavis:
metamask can have a little build determinism, as a treat

relevant change in `bify-module-groups` https://github.com/MetaMask/bify-module-groups/pull/3/files

### DIFF
--- a/package.json
+++ b/package.json
@@ -276,7 +276,7 @@
     "@typescript-eslint/parser": "^4.20.0",
     "addons-linter": "1.14.0",
     "babelify": "^10.0.0",
-    "bify-module-groups": "^1.0.0",
+    "bify-module-groups": "^2.0.0",
     "brfs": "^2.0.2",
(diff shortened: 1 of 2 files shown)


Gudahtt (approved):
LGTM! I've tested 3 builds so far, fourth is still running. All bundles are identical between builds. I'm seeing differences in the `builds/` zip file that I don't yet understand and will continue to investigate, but all of the files in `dist` appear to be the same between builds.

Gudahtt:
The zip inconsistencies is probably due to file metadata (e.g. https://github.com/sindresorhus/gulp-zip/issues/103). Investigating a solution to that now, can fix in a separate PR. This PR makes the JS bundles deterministic at least.

brad-decker (approved):

brad-decker:
I tested two builds, so didn't go as far as Mark did, but the code looks good and the builds are identical. Good work @kumavis

PR #14613, opened 2022-05-04
Fix contact address pet name 

### LINKED ISSUE #13736: [Bug]: Address Pet name and currency conversion unavailable in "Edit" mode in the Send flow
### Describe the bug

After initiating a transfer, if a user clicks "edit" and goes back to the screen with recipients address and token. The inputs are broken. The contact pet name gets replaced by the address and it appears twice. As well as the token fiat value is unavailable. See the video below.

https://user-images.githubusercontent.com/87030673/155383141-77b3579e-c1e8-42ad-85a0-4e21a96057be.mov

 

### Steps to reproduce

1. Start a transfer with an asset
2. Select a saved contact (with a pet name) 
3. Select a token with a fiat value
4. Click Next 
5. Click Edit and the currency conversion rate isn't shown + the pet name is replaced by the recipients address

_No response_

### Version

10.9.3

### Build type

_No response_

### Browser

Firefox

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

VSaric:
## Explanation

In this PR the contact address pet name isn't replaced by the recipients address. 

This fix is just first of two steps that needs to be fixed in #13736. For second step which is fixing currency conversion rate, I can't reproduce it on `Ropsten Test Network`. Like @SayaGT said it isn't sure how to reproduce it on anything except `Ethereum Mainnet`. 

We are waiting for @danjm respond on that and more informations about how can we test and reproduce the second step (currency coversion rate).

## More Information

* Fixes #13736

### Before

https://user-images.githubusercontent.com/92527393/166682098-012c7801-1a43-476a-84af-74c4d9f54715.mov

### After

https://user-images.githubusercontent.com/92527393/166682165-ebb1fcc9-f183-4b97-9803-1ed5090be0c2.mov

## Manual Testing Steps

1. Start a transfer with an asset
2. Select a saved contact (with a pet name)
3. Select a token
4. Click Next
5. Click Edit

### DIFF
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -2265,10 +2265,15 @@ export function getRecipient(state) {
   const checksummedAddress = toChecksumHexAddress(
     state[name].recipient.address,
   );
-  return {
-    ...state[name].recipient,
-    nickname: state[name].recipient.nickname || getEnsResolutionByAddress(state, checksummedAddress),
+  if (state.metamask.ensResolutionsByAddress) {
+    return {
+      ...state[name].recipient,
+      nickname:
+        state[name].recipient.nickname ||
+        getEnsResolutionByAddress(state, checksummedAddress),
+    };
   }
+  return state[name].recipient;
 }
 
 // Overall validity and stage selectors
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -2265,17 +2265,10 @@ export function getRecipient(state) {
   const checksummedAddress = toChecksumHexAddress(
     state[name].recipient.address,
   );
-  if (
-    state[name].recipient.nickname === '' &&
-    state.metamask.ensResolutionsByAddress
-  ) {
-    const recipient = {
-      ...state[name].recipient,
-      nickname: getEnsResolutionByAddress(state, checksummedAddress),
-    };
-    return recipient;
+  return {
+    ...state[name].recipient,
+    nickname: state[name].recipient.nickname || getEnsResolutionByAddress(state, checksummedAddress),
   }
-  return state[name].recipient;
 }
 
 // Overall validity and stage selectors
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -44,6 +44,7 @@ import {
   getTokenList,
   getAddressBookEntryOrAccountName,
   getIsMultiLayerFeeNetwork,
+  getEnsResolutionByAddress,
 } from '../../selectors';
 import {
   disconnectGasFeeEstimatePoller,
@@ -91,6 +92,7 @@ import { resetEnsResolution } from '../ens';
 import {
   isBurnAddress,
   isValidHexAddress,
+  toChecksumHexAddress,
 } from '../../../shared/modules/hexstring-utils';
 import { sumHexes } from '../../helpers/utils/transactions.util';
 import fetchEstimatedL1Fee from '../../helpers/utils/optimism/fetchEstimatedL1Fee';
@@ -2260,6 +2262,19 @@ export function getRecipientUserInput(state) {
 }
 
 export function getRecipient(state) {
+  const checksummedAddress = toChecksumHexAddress(
+    state[name].recipient.address,
+  );
+  if (
+    state[name].recipient.nickname === '' &&
+    state.metamask.ensResolutionsByAddress
+  ) {
+    const recipient = {
+      ...state[name].recipient,
+      nickname: getEnsResolutionByAddress(state, checksummedAddress),
+    };
+    return recipient;
+  }
   return state[name].recipient;
 }
 
--- a/ui/selectors/selectors.js
+++ b/ui/selectors/selectors.js
@@ -46,6 +46,7 @@ import {
   getValueFromWeiHex,
   hexToDecimal,
 } from '../helpers/utils/conversions.util';
+import { getSendTo } from '../ducks/send';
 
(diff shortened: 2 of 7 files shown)


VSaric:
After todays meeting with @danjm we discussed about second step for fixing this issue (currency conversion rate). It can be reproduced with `ganache` and we agreed to create another PR for fixing second step (currency conversion rate).

mirjanaKukic:
Verified by QA

ui/selectors/selectors.js
```diff
@@ -375,6 +376,14 @@ export function getAddressBook(state) {
   return Object.values(state.metamask.addressBook[chainId]);
 }
 
+export function getRecipientEnsName(state) {
+  const recipientAddress = getSendTo(state);
+  const checksummedAddress = toChecksumHexAddress(recipientAddress);
+  const recipientEns =
+    state.metamask.ensResolutionsByAddress[checksummedAddress] || '';
```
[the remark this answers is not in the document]
VSaric: @darkwing [Here](https://github.com/MetaMask/metamask-extension/pull/14613/commits/b87bdfc5e9898236622200c643fdbc8ecb40b84c) is created selector and used in the codebase.

darkwing (dismissed):
This looks awesome and worked great!

0ex-d (dismissed):
Clean

segun (dismissed):
LGTM

brad-decker (changes requested):
This isn't the right fix for this, the fix should be made in the send slice, not ejecting from the send slice state. I'll add a more thorough review soon and ideas but blocking for now

ui/selectors/selectors.js
```diff
@@ -375,6 +376,17 @@ export function getAddressBook(state) {
...
+export function getEnsResolutionByAddress(state, address) {
+  return state.metamask.ensResolutionsByAddress[address] || '';
+}
+
+export function getRecipientEnsName(state) {
+  const recipientAddress = getSendTo(state);
+  const checksummedAddress = toChecksumHexAddress(recipientAddress);
+  const recipientEns = getEnsResolutionByAddress(state, checksummedAddress);
+  return recipientEns;
+}
```
brad-decker, line 388: Let's bake this functionality into the getRecipient selector in the send duck.

You can keep the genEnsResolutionByAddress selector you have here, but in the send duck modify the getRecipient selector such that it returns an ens resolution as the nickname if a nickname isn't present but and ens resolution is. Anywhere we display recipient.nickname then will benefit from this treatment.

brad-decker (changes requested):

ui/selectors/selectors.js
VSaric: @brad-decker Fixed [here](https://github.com/MetaMask/metamask-extension/pull/14613/commits/b9250908d9228c50586ebd8b8a8c561864a12169). I used `state.metamask.ensResolutionsByAddress` in the `getRecipient` selector in the send duck to check if the ens resolusion is present. If you think this approach isn't right, I would like some new suggestion if you have about this.
Thank you in advance!

ui/ducks/send/send.js
```diff
@@ -2259,6 +2261,19 @@ export function getRecipientUserInput(state) {
...
+    state[name].recipient.nickname === '' &&
+    state.metamask.ensResolutionsByAddress
+  ) {
+    const recipient = {
+      ...state[name].recipient,
+      nickname: getEnsResolutionByAddress(state, checksummedAddress),
+    };
+    return recipient;
+  }
   return state[name].recipient;
```
brad-decker, line 2277: We can simplify this by using `||` which will evaluate an empty string to false, so ``` '' || 'hello' === 'hello'``` so we don't need to check for the presence of a ensResolution first, just checksum the address, and try to get it and let the selector return ```''``` if nothing is found. 
```suggestion
  const checksummedAddress = toChecksumHexAddress(
    state[name].recipient.address,
  );
  return {
    ...state[name].recipient,
    nickname: state[name].recipient.nickname || getEnsResolutionByAddress(state, checksummedAddress),
  }
```

brad-decker (dismissed):
What you have works, so i'm gonna mark it as approved, but i've got a suggestion for you if you wanna take it. :)

ui/ducks/send/send.js
VSaric: Thank you @brad-decker for a suggestion, this is much better than the `if` statement that I used. :)

VSaric: @brad-decker As you can see, `test-unit` are failing. The reason why they failing is that because on the `send.test.js` in `/ui/pages/send` we don't have `ensResolutionsByAddress` added in `baseStore.metamask`.
This is the error that is shown in Circle-CI: `TypeError: Cannot read property '' of undefined`.
When I added that property and run unit tests locally I run into infinity loop and `send.test.js` never finished. 
So I think that we need to use `state.metamask.ensResolutionsByAddress` in the `getRecipient` selector in the send duck to check if the ens resolusion is present. If ens resolusion is present then we return an ens resolution as the nickname. When I added that statement all tests are passed and everything works fine.[Here](https://github.com/MetaMask/metamask-extension/pull/14613/commits/3d3850488b4251f1044fc290a128115e5960302f) is the changes that I made to fix `test-unit` and pipeline to be green.

brad-decker: You probably need to update the store object we are using in the failing test to have the ensResolutionByAddress object, which will be present in state in production. @VSaric

darkwing (approved):
This is awesome!  Great work!

brad-decker (approved):

segun (approved):

PR #14612, opened 2022-05-04
Jest tests - incremental coverage

PeterYinusa:
## Explanation

Automatically bump up global Jest thresholds whenever coverage goes above them
The minimum threshold increase is set to `5`, to avoid friction as multiple people contribute to the repository.
We can monitor this and tweak this number as needed

## More Information

See: #14555 

### After

Check fails (`ci/circleci: test-unit`) if the coverage needs updating

Failure

Check passes (`ci/circleci: test-unit`) if the coverage does not need updating

Success

## Manual Testing Steps

- yarn setup
- yarn test:coverage:jest

## Pre-Merge Checklist

- [x] PR template is filled out
- [x] Manual testing complete & passed
- [x] **IF** this PR fixes a bug, a test that _would have_ caught the bug has been added
- [x] PR is linked to the appropriate GitHub issue
- [x] **IF** QA attention is required, "QA Board" label has been applied
- [x] PR has been added to the appropriate release Milestone

PR #14616, opened 2022-05-04
Fix redundant entry in lockfile

Gudahtt:
The lockfile would change upon install due to a redundant entry. This seems to have been introduced in #14612 as a result of running `yarn-deduplicate`.


adonesky1 (approved):

Gudahtt:
I'm guessing that CI didn't catch this because the `--frozen-lockfile` flag we use will prevent lockfile changes like this from being written. Hmm. Maybe we should update the "lint lockfile" step to check for this.

PeterYinusa (approved):

PR #14623, opened 2022-05-05
Reproducible `.zip` files

Gudahtt:
Our build system now creates `.zip` archives in a reproducible way. Previously the `.zip` file would differ between builds even when the files being archived were identical. This was because the order the files were passed in was non-deterministic, and the `mtime` for each file was different between builds.

The files are now sorted before being zipped, and the `mtime` for each file has been set to the unix epoch.

## Manual Testing Steps

* Run `yarn dist`
* Rename the `builds` directory (e.g. `mv builds builds2`)
* Run `yarn dist` again
* Compare the hashes of both runs  (`e.g. `md5sum builds/* builds2/*`)

## Pre-Merge Checklist

- [ ] **IF** QA attention is required, "QA Board" label has been applied


development/build/etc.js
```diff
@@ -45,7 +46,9 @@ function createZipTask(platform, buildType, version) {
         : `metamask-${buildType}-${platform}-${version}`;
     await pump(
       gulp.src(`dist/${platform}/**`),
-      gulpZip(`${path}.zip`),
+      // sort files and set `mtime` to epoch to ensure zip build is deterministic
+      sort(),
+      gulpZip(`${path}.zip`, { modifiedTime: new Date(0) }),
```
FrederikBolding, line 51: Why can't we use the current unix timestamp? I.e. `Date.now()`?

Gudahtt: Because the `gulp-zip` accepts a `Date` object for the `modifiedTime` option: https://www.npmjs.com/package/gulp-zip

Gudahtt: That would make the build non-deterministic. It would be different each time.

Gudahtt: Sorry I guess that's not true, they would be "deterministic" for a known timestamp. They would not be reproducible though. Reproducible is the word I should have used. That is the requirement this is trying to meet.

darkwing (approved):
I followed the manual testing directions and got matching md5 checksums for build file between the directories.

brad-decker (approved):
LGTM, matching checksum!

PR #14631, opened 2022-05-06
[e2e] Deploy contracts directly seeding Ganache

david0xd:
### Goal
The main goal is to deploy the smart contracts directly on Ganache, instead of using the Dapp Test interface for achieving the same actions.

### Approach
We are going to split this task in 2 different PRs. The first one (this one) will take care of preparing the contracts and the seeder, so it deploys correctly any contract we pass them into Ganache, from the background.

**First Part in Detail**
- Inside `e2e/seeder` we have a contracts folder, with the different byte code & abi contracts from Test-Dapp. Having separate contracts helps us to be more granular whenever we want to deploy just 1 contract specifically. It can also be expanded by adding more contracts, following the same approach
- `ganache-seeder.js` will deploy a contract to ganache, by contract name

### Testing Steps
- Include smartContract fixture on any testcase with a contract name. I.e.
image
- Run test `yarn test:e2e:single test/e2e/tests/test-name.spec.js --browser=chrome`. 
- Check smart contracthas been deployed to ganache by looking at the console.


test/e2e/seeder/contracts/collectibles-factory.js
```diff
@@ -0,0 +1,356 @@
+const collectiblesFactory = {
```
PeterYinusa, line 1: can we import this contract instead of copying it all across? I think it's accessible in the `test-dapp` package

seaona: I've tried but I kept getting this error `window is not defined`.

Investigating further I think at the moment it is not possible, due to the way test-dapp package is built, as contract variables are not accessible from bundle. Here an explanation:
https://stackoverflow.com/questions/58434256/how-do-i-access-variables-functions-in-bundle-js

I think we could do this if I modify the test-dapp repo and made contract variables accessible from the outside. What do you think @PeterYinusa ? Should I try this option?

PeterYinusa: I think you should be able to get the file containing the abi and bytecode in `smart-contracts.js` with the following:

```
const constants = require('../../../node_modules/@metamask/test-dapp/dist/constants.json');
```

Then we can remove entire contracts directory.

seaona: thank you v much @PeterYinusa that work successfully!!

PeterYinusa (dismissed):

PeterYinusa (dismissed):

test/e2e/helpers.js
```diff
@@ -46,6 +48,15 @@ async function withFixtures(options, testSuite) {
   let failed = false;
   try {
     await ganacheServer.start(ganacheOptions);
+    let contractRegistry;
+
+    if (smartContract) {
+      const debugGanacheSeeder = true;
+      const ganacheSeeder = new GanacheSeeder(debugGanacheSeeder);
```
darkwing, line 55: Nit: Can we avoid the `debugGanacheSeeder` var and simply pass `true` here?

seaona: I think the intention of this approach is for clarity reasons: so code is more understandable, as we explicitly say we are activating the debugger. But surely we can pass true directly, and avoid one extra line. (this part of code was initially done by @david0xd). Ty!

PeterYinusa (approved):

darkwing (approved):

Issue #13087, opened 2021-12-13
Unable to proceed with tx bc of inaccurate/overly aggressive "Insufficient funds for gas" warning


dpazdan:
**Describe the bug**
Cannot proceed with erc20 send when user has "just enough" ETH for gas.

**Steps to reproduce (REQUIRED)**
Try sending a token with a wallet that has ~0.02 ETH or less.

**Expected behavior**
The next screen should be allowed, so the user can edit gas lower than the suggested gas estimate.  Oftentimes a tx will still be sent eventually with 20 - 40% lower gwei than current estimates.

**Screenshots**

![image](https://user-images.githubusercontent.com/16330048/145847094-781a483d-76ae-4e6f-bb5d-8dcae2bc62ee.png)

![image](https://user-images.githubusercontent.com/16330048/145847109-014c85f0-c298-4866-8f36-705e1bb840e3.png)

![image](https://user-images.githubusercontent.com/16330048/145847128-3f475bf4-dc06-4d33-be68-a67fc63255b5.png)

Notice above, this wallet has plenty enough ETH for current estimates.  I'm not sure why I'm even getting the warning to be honest.

MM v.10.7.1

Related to GH issue #10276 and #12029

dpazdan:
cc @danjm @kevinghim

bschorchit:
Here's the [design for the fix](https://www.figma.com/file/hdpaC4Hprc2oaPSMx2fs2a/Insufficient-funds-warning?node-id=47%3A423)

danjm:
![Screenshot from 2022-03-18 09-15-33](https://user-images.githubusercontent.com/7499938/158997492-e160b03f-2acc-441d-91d9-a99788e1c8e2.png)

EHaracic:
### **User Story**
As a user, I want a more informative “Insufficient funds for gas” warning on the send screen.

On the send screen, the user is shown an "Insufficient funds for gas" warning, but no other information about gas. So the user doesn't know if they have insufficient funds because prices are currently high or because the gas limit of the particular contract interaction they are attempting is high.



### Figma/Notion links
Figma link: https://www.figma.com/file/hdpaC4Hprc2oaPSMx2fs2a/Insufficient-funds-warning?node-id=47%3A423

### Acceptance criteria
- Update the Send screen according the Figma designs, screenshots below.
Consider two different scenarios:
1. The user **does not have enough** balance for gas fee before entering the amount
[...]

PR #14634, opened 2022-05-06
Unable to proceed with tx bc of inaccurate/overly aggressive "Insufficient funds for gas" warning #13087

dragana8:
As a user, I want a more informative “Insufficient funds for gas” warning on the send screen.

On the send screen, the user is shown an "Insufficient funds for gas" warning, but no other information about gas. So the user doesn't know if they have insufficient funds because prices are currently high or because the gas limit of the particular contract interaction they are attempting is high. So we show the estimated gas fee and warning message on the Send screen if user does not have enough tokens for the fee or total transaction amount is greater than the balance.

More info: https://github.com/MetaMask/metamask-extension/issues/13087

* First part for #13087 

### BEFORE:

**Expand view (GoerliETH):**

Screenshot 2022-11-04 at 11 13 30

### AFTER:

**Expand view when Enable enhanced gas fee UI is off (GoerliETH):**

Screenshot 2022-11-04 at 10 55 02

**Expand view  when Enable enhanced gas fee UI is on (GoerliETH):**

Screenshot 2022-11-04 at 10 59 05

### Popup view:

https://user-images.githubusercontent.com/92527393/199946441-1a8edda3-973c-40ea-9f3e-56d7d4103539.mov

### Expand view:

https://user-images.githubusercontent.com/92527393/194052167-6446f07c-7346-4aaa-9fbf-35a852bc4e80.mov

### Switching between networks (Ethereum Mainnet, Avalanche C-Chain, Polygon Mainnet, Arbitrum One):

https://user-images.githubusercontent.com/92527393/199950577-9f09b8ba-07d1-4c24-8249-1808d52198d8.mov


segun:
can we have more information in the description? especially the before and after screenshots.

ui/pages/send/gas-display/gas-display.component.js
```diff
@@ -0,0 +1,371 @@
...
+import LoadingHeartBeat from '../../../components/ui/loading-heartbeat';
+
+import TransactionDetailItem from '../../../components/app/transaction-detail-item/transaction-detail-item.component';
+import { NETWORK_TO_NAME_MAP } from '../../../../shared/constants/network';
+import TransactionDetail from '../../../components/app/transaction-detail/transaction-detail.component';
+import ActionableMessage from '../../../components/ui/actionable-message/actionable-message';
+
+const renderHeartBeatIfNotInTest = () =>
+  process.env.IN_TEST ? null : <LoadingHeartBeat />;
+export default class GasDisplay extends Component {
```
brad-decker, line 26: [REQ] This should be refactored to be a functional component

brad-decker (changes requested):
We need to rebase this and resolve conflicts as well.

ui/pages/send/gas-display/gas-display.component.js
brad-decker: Also drop the .component part of the filename, please :)

VSaric: @brad-decker Done [here](https://github.com/MetaMask/metamask-extension/pull/14634/commits/ce133455d78ea7c4671ac747b61bc6f926d76360).

VSaric:
@danjm I pushed changes like we talked in the meeting today. [Here](https://github.com/MetaMask/metamask-extension/pull/14634/commits/ce133455d78ea7c4671ac747b61bc6f926d76360) are those changes. When you have time please review. Also @brad-decker, please review. Thank you! 😄

ui/pages/send/gas-display/gas-display.js
```diff
@@ -0,0 +1,367 @@
...
+            }}
+            fullWidth
+            margin="dense"
+            value={customNonceValue || ''}
+          />
+        </div>
+      </div>
+    </div>
+  ) : null;
+  const renderTotalMaxAmount = () => {
```
brad-decker, line 83: Why is there a nonce field in the gas-display component?

brad-decker (changes requested):

mirjanaKukic:
Verified by QA

ui/pages/send/send-content/send-content.component.js
```diff
@@ -76,12 +100,15 @@ export default class SendContent extends Component {
     const showKnownRecipientWarning =
       recipient.warning === 'knownAddressRecipient';
     const hideAddContactDialog = recipient.warning === 'loading';
+    const title = '';
```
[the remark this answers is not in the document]
VSaric: @brad-decker Sorry about that, `title` is now used when we want to send some custom tokens (like DAI, UNI) in send flow. If we set amount greater then the token balance, in total field will be displayed `title` in the same way like is displayed on confirmation screen. Done [here](https://github.com/MetaMask/metamask-extension/pull/14634/commits/71c1459bb901b933d843ba0af7206893f7c988f8).

ui/pages/send/gas-display/gas-display.js
VSaric: @brad-decker You are right, a nonce field doesn't need to be in gas-display component. I deleted it [here](https://github.com/MetaMask/metamask-extension/pull/14634/commits/71c1459bb901b933d843ba0af7206893f7c988f8).

ui/pages/send/gas-display/gas-display.js
```diff
@@ -0,0 +1,337 @@
...
+            value={hexTransactionTotal}
+            hideLabel={!useNativeCurrencyAsPrimaryCurrency}
+          />
+        </div>
+      );
+    }
+    return useNativeCurrencyAsPrimaryCurrency
+      ? primaryTotalTextOverrideMaxAmount
+      : secondaryTotalTextOverride;
+  };
```
brad-decker, line 101: I would prefer these to go inline in the props instead of as intermediary functions that are invoked.

ui/pages/send/gas-display/gas-display.js
```diff
@@ -0,0 +1,337 @@
...
+        </div>
+      );
+    }
+    return useNativeCurrencyAsPrimaryCurrency
+      ? secondaryTotalTextOverride
+      : primaryTotalTextOverride;
+  };
+  const renderGasDetailsItem = () => {
+    return (
+      <div className="transaction-item">
```
brad-decker, line 127: Why does this TransactionDetailItem get wrapped in 'transaction-item' div?

ui/pages/send/gas-display/gas-display.js
```diff
@@ -0,0 +1,337 @@
...
+                  </>
+                }
+                position="top"
+              >
+                <i className="fa fa-info-circle" />
+              </InfoTooltip>
+            </>
+          }
+          detailText={
+            <div className="confirm-page-container-content__currency-container test">
```
brad-decker, line 160: I can't say that I like using confirm-page styles for this component, its muddling the scss and using styles from another part of the codebase... also why the 'test' class? 

My preference would be to copy the pertinent styles into this gas-display component (replacing as much of them with the Box component as possible) .. this might be duplicative but it makes this component more  resilient to changes in the other component.

brad-decker (changes requested):

ui/pages/send/gas-display/gas-display.js
VSaric: @brad-decker Done [here](https://github.com/MetaMask/metamask-extension/pull/14634/commits/30e2c90ae66cf2a12aeb11e5ea35daa7cfd60340).

ui/pages/send/gas-display/gas-display.js
VSaric: @brad-decker It is not needed this 'transaction-item' div here. I deleted it [here](https://github.com/MetaMask/metamask-extension/pull/14634/commits/30e2c90ae66cf2a12aeb11e5ea35daa7cfd60340).

ui/pages/send/gas-display/gas-display.js
VSaric: @brad-decker Done [here](https://github.com/MetaMask/metamask-extension/pull/14634/commits/30e2c90ae66cf2a12aeb11e5ea35daa7cfd60340).

ui/pages/send/send-content/send-content.component.js
```diff
@@ -77,11 +121,90 @@ export default class SendContent extends Component {
...
+              hexMaximumTransactionFee,
+            )}
+            hideLabel={!useNativeCurrencyAsPrimaryCurrency}
+          />
+        );
+      }
+      return useNativeCurrencyAsPrimaryCurrency
+        ? primaryTotalTextOverrideMaxAmount
+        : undefined;
+    };
```
brad-decker, line 202: these should not be functions. They should either be inline in the props, or just variables.

brad-decker (changes requested):

ui/pages/send/send-content/send-content.component.js
VSaric: @brad-decker Done [here](https://github.com/MetaMask/metamask-extension/pull/14634/commits/3a34638c9acaf99d1d84076857a756c6b30c0a01).

ui/pages/send/gas-display/gas-display.js
```diff
@@ -0,0 +1,290 @@
...
+  return (
+    <>
+      {showDepositPopover && (
+        <DepositPopover onClose={() => setShowDepositPopover(false)} />
+      )}
+      <Box className="gas-display">
+        <TransactionDetail
+          userAcknowledgedGasMissing={false}
+          rows={[
+            renderGasDetailsItem(),
```
brad-decker, line 168: Either make this a new component, or put the JSX inline. I realize its a lot of JSX but if you want to break it out you should use components not functions. for example have a gas-details.js file that does this portion and then render it instead of calling a function.

ui/pages/send/send-content/send-content.component.js
```diff
@@ -77,11 +121,83 @@ export default class SendContent extends Component {
...
+      >
+        <LoadingHeartBeat estimateUsed={draftTransaction?.userFeeLevel} />
+        <UserPreferencedCurrencyDisplay
+          type={SECONDARY}
+          key="total-detail-text"
+          value={hexTransactionTotal}
+          hideLabel={Boolean(useNativeCurrencyAsPrimaryCurrency)}
+        />
+      </Box>
+    );
```
brad-decker, line 157: I would prefer to just see this inline in the JSX then to have to scroll around to see what it is. In general the only time I think we should advise to set a variable to a JSX element is when that element will be manipulated (ie if it was let detailText and we were modifying what component to render based on logic). For example, detailTotal and maxAmount are actually perfectly fine to do this way because they are being modified and it makes the code more readable to have that logic clearly broken apart out of the main JSX statement.

ui/pages/send/send-content/send-content.component.js
```diff
@@ -77,11 +121,83 @@ export default class SendContent extends Component {
...
+          hideLabel={!useNativeCurrencyAsPrimaryCurrency}
+        />
+      );
+    } else if (useNativeCurrencyAsPrimaryCurrency) {
+      detailTotal = primaryTotalTextOverrideMaxAmount;
+      maxAmount = primaryTotalTextOverrideMaxAmount;
+    } else {
+      detailTotal = undefined;
+      maxAmount = undefined;
+    }
```
brad-decker, line 195: This else statement is unnecessary, they will be undefined if neither case hits.

ui/pages/send/send-content/send-content.container.js
```diff
@@ -24,6 +31,43 @@ function mapStateToProps(state) {
...
+  };
+
+  const {
+    hexTransactionAmount,
+    hexMinimumTransactionFee,
+    hexMaximumTransactionFee,
+    hexTransactionTotal,
+  } = transactionFeeSelector(state, txData);
+
+  const { useNativeCurrencyAsPrimaryCurrency } = getPreferences(state);
```
brad-decker, line 69: Rather than adding all of these to sent-content, just to pass them to the new gas-display component...couldn't we just use hooks in the gas-display component to get that information? Also, when a transaction is edited its details are plugged into the draftTransaction object and it becomes the current draft. You *shouldn't* have to use either or. Can you let me know if you're experiencing something different?

brad-decker (changes requested):

ui/pages/send/send-content/send-content.container.js
VSaric: @brad-decker Done [here](https://github.com/MetaMask/metamask-extension/pull/14634/commits/c64948a3c0494308f824cdc443a9b04b010b9054).

ui/pages/send/send-content/send-content.component.js
VSaric: @brad-decker Done [here](https://github.com/MetaMask/metamask-extension/pull/14634/commits/c64948a3c0494308f824cdc443a9b04b010b9054).

ui/pages/send/send-content/send-content.component.js
VSaric: @brad-decker Done [here](https://github.com/MetaMask/metamask-extension/pull/14634/commits/c64948a3c0494308f824cdc443a9b04b010b9054).

ui/pages/send/gas-display/gas-display.js
VSaric: @brad-decker Done [here](https://github.com/MetaMask/metamask-extension/pull/14634/commits/c64948a3c0494308f824cdc443a9b04b010b9054).

brad-decker (dismissed):

brad-decker (approved):

VSaric:
Can I get one more eyes/review on this PR @danjm? 👀 
Thank you in advance! 😄

pedronfigueiredo (approved):

PR #14637, opened 2022-05-06
Update phishing detection so that the click through to the page is properly tested

danjm:
This PR improves the coverage of our e2e tests for the phishing detection page.

In particular, it confirms that the click through to the detected domain is successful. It tests this by setting the test dapp as the site to attempt to access, and putting localhost on the blacklist.

### DIFF
--- a/test/e2e/tests/phishing-detection.spec.js
+++ b/test/e2e/tests/phishing-detection.spec.js
@@ -15,7 +15,7 @@ describe('Phishing Detection', function () {
             tolerance: 2,
             fuzzylist: [],
             whitelist: [],
-            blacklist: ['example.com'],
+            blacklist: ['localhost'],
           },
         };
       });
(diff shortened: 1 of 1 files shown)


test/e2e/tests/phishing-detection.spec.js
```diff
@@ -36,15 +36,20 @@ describe('Phishing Detection', function () {
...
         testSpecificMock: mockPhishingDetection,
+        dapp: true,
       },
       async ({ driver }) => {
         await driver.navigate();
         await driver.fill('#password', 'correct horse battery staple');
         await driver.press('#password', driver.Key.ENTER);
-        await driver.openNewPage('http://example.com');
-        await driver.waitForSelector({ text: 'continuing at your own risk' });
+        await driver.navigate();
```
Gudahtt, line 45: Nit: Is this line necessary? I don't follow what it does, if it is necessary.

test/e2e/tests/phishing-detection.spec.js
```diff
@@ -36,15 +36,20 @@ describe('Phishing Detection', function () {
...
+        await driver.openNewPage('http://localhost:8080');
+        const continueLink = await driver.waitForSelector({
+          text: 'continuing at your own risk',
+        });
+        await continueLink.click();
```
Gudahtt, line 50: Nit: This seems simpler

```suggestion
        await driver.clickElement({
          text: 'continuing at your own risk',
        });
```

The `waitForSelector` method is for when you _just_ want to wait. All of the "interact" functions have built-in functionality to wait until the default timeout until the selector is found.

Gudahtt (commented):
The test mostly looks good! Except the test name is "should display the MetaMask Phishing Detection page", and that's no longer what this test shows. So maybe the test description should change. Unless you think this is worth capturing in two separate tests (one for display, one for continuing).

Gudahtt:
These changes have already landed as part of https://github.com/MetaMask/metamask-extension/commit/7199d9c56775111f85225fe15297e47de8e2bc96

PR #14649, opened 2022-05-09
[e2e] Enabling/Disabling token conversion on testnets

### LINKED ISSUE #14606: Create an e2e test for checking "Show testnet conversion to Fiat" functionality
**Why**: a [bug](https://github.com/MetaMask/metamask-extension/issues/14569) was introduced due to a recent change on the ticker symbols for ETH on testnets. This issue was not catched by any test.

**What**: include an e2e test that checks the toggle functionality for activating/deactivating the "Show Conversion on test networks".

seaona:
## Explanation
New e2e testcase for checking that "Enabling/Disabling Token Conversion" toggle works correctly. This means testing that:

- shows conversion on test network when toggle is activated
- don't show conversion on test network when deactivated

## More Information

* Resolves #14606

## Manual Testing Steps
- Build test environment: yarn build:test
- Run test
  - `yarn test:e2e:single test/e2e/tests/advanced-settings.spec.js --browser=firefox`
  - `yarn test:e2e:single test/e2e/tests/advanced-settings.spec.js --browser=chrome` 

## Pre-Merge Checklist

- [X ] PR template is filled out
- [X ] PR is linked to the appropriate GitHub issue


brad-decker (dismissed):

adonesky1 (commented):
Not a big deal, but would it make more sense to modify these tests so that there is a secondary currency for the test nets? I guess this would involve mocking the coversion rate api?

seaona:
@adonesky1 what do you mean by `modify these tests so that there is a secondary currency for the test nets`? So we are testing both scenarios, if it is present when activated and if it is not present when deactivated. Do you mean something else?

For the mocking part, that is a great point, I could test only Localnetwork with API response mocked, instead of testing real testnets. Ty!!

darkwing (dismissed):

test/e2e/tests/advanced-settings.spec.js
```diff
@@ -0,0 +1,76 @@
...
+
+        await driver.clickElement(
+          '.settings-page__header__title-container__close-button',
+        );
+
+        const secondaryCurrency = await driver.isElementPresent(
+          '[data-testid="eth-overview__secondary-currency"]',
+        );
+        assert.equal(
+          await secondaryCurrency,
```
adonesky1, line 44: Do you have to await here too, even though you did so on line 40? Also instead of checking for `isElementPresent` should we check for the text we expect to be present? In this case `USD` right?

seaona: great catch! I removed the `await` for both `it` blocks as there was not needed. Thank you @adonesky1!

For the `isElementPresent`, I believe there is no difference. I chose this approach as we already had a function on the driver, and was straightforward to apply, as this element represents the secondary currency. Would you recommend me to write another function for `isTextPresent` in the driver file and use this one instead? Ty!!

adonesky1: @seaona I guess I'm just thinking of something like:
```
const secondaryCurrency = await driver.findElement(
          '[data-testid="eth-overview__secondary-currency"]',
        );
 assert.equal(secondaryCurrency.getText(), 'USD')
```

or something like that?

seaona: oh the thing is that `data-testid="eth-overview__secondary-currency` element only appears when the functionality is active, so in the case is deactivated, the `findElement` would fail as this element is not there anymore, so this `findElement` would fail @adonesky1 

image

adonesky1: Could we use the `.findElement` pattern in the `should show conversion on test network when activated` test and use `isElementPresent` on `should not show conversion on test network when deactivated`? Or is that a testing anti-pattern, since you should use the same condition for on and off?

seaona: I think it is okay to do this! I will update it right now @adonesky1 thank you!!!

adonesky1 (approved):
LGTM! Great work. Thanks for bearing with my change requests!

seaona:
thank you for your suggestions @adonesky1 !!

PeterYinusa (approved):

PR #14673, opened 2022-05-10
Remove gas editing in Swaps

dan437:
## Explanation
We should prevent users from editing gas parameters when using Swaps. Sometimes users are submitting transactions that will never get picked up, and by doing so, they unintentionally block themselves from submitting additional transactions.

## Screenshots
Before:
image

Before (Edit gas modal):
image

After:
image


darkwing (approved):
Sad to see it go :(

adonesky1 (approved):

PR #14526, opened 2022-04-25
Support for ENS wildcard and offchain resolution

makoto:
## Explanation

This PR addresses https://github.com/MetaMask/specifications/issues/9

NOTE: Support for contenthash will be separate PR.

## More information

Please read https://medium.com/the-ethereum-name-service/upgrade-ethers-js-to-5-6-1-to-activate-ens-l2-offchain-integration-40ee1a0fdf2a for more deail.

### Display forward lookup

Screenshot 2022-04-25 at 16 36 05

### Display reverse record (aka primary name)

Screenshot 2022-04-25 at 16 35 48

## Manual testing steps

- 1. Select testnet (Ropsten/Rinkeby/Goerli network)
- 2. Lookup 1.offchainexample.eth to lookup ENS name
- 3. Send ETH to 1.offchainexample.eth and check it shows in the transaction log

PR #14675, opened 2022-05-10
Support for ENS wildcard and offchain resolution

makoto:
NOTE: This is retry of https://github.com/MetaMask/metamask-extension/pull/14526 as it included some commits I made but for some reason it didn't show my name 

## Explanation

This PR addresses https://github.com/MetaMask/specifications/issues/9

NOTE: Support for contenthash will be separate PR.

## More information

Please read https://medium.com/the-ethereum-name-service/upgrade-ethers-js-to-5-6-1-to-activate-ens-l2-offchain-integration-40ee1a0fdf2a for more deail.

### Display forward lookup

Screenshot 2022-04-25 at 16 36 05

### Display reverse record (aka primary name)

Screenshot 2022-04-25 at 16 35 48

## Manual testing steps

- 1. Select testnet (Mainnet/Ropsten/Rinkeby/Goerli network)
- 2. Lookup 1.offchainexample.eth to lookup ENS name
- 3. Send ETH to 1.offchainexample.eth and check it shows in the transaction log


0ex-d (dismissed):

seaona:
Some findings/questions from QA:

- It looks like ENS resolution stopped working with emojis in this branch. See difference between this branch (left) and production Extension (right)
ens-not-resolving

- On Mainnet, when we are in the Confirmation Tx screen, we can see that ENS subdomains are displayed as the resolved address instead of the ENS name. On the tx details ENS subdomains are also displayed as the resolved address. Example below using `1.offchainexample.eth`. I suspect here we want to show the ENS name instead?

image

image


- When I look for a subdomain containing emojis, the error `Error: STRINGPREP_CONTAINS_UNASSIGNED` is thrown on the console and I see `ENS Lookup failed.` on the UI. Are emojis allowed on subdomains?

image


- When an address is not registered (on mainnet), we used to saw the message `No address has been set` (right). Now we don't see any message, but we cannot proceed with the tx neither.
I read that `you can actually send assets to an ENS name that has never been registered`. So I'm wondering in this case, how should we proceed on the UI. Should we allow to go to the next screen and send eth to that ENS address? and/or should we display any message/warning? At the moment, none of these options is happening and maybe it confuses the user

image

- FYI: opened [this bug](https://github.com/MetaMask/metamask-extension/issues/14800), related to red char highlight, as a separate issue, as it's an old one

makoto:
Hi @seaona thank you for the detailed feedback.
I will look into 3~5.
For 1 and 2, the current ENS name resolution follows a rule specified by [IDNA](https://datatracker.ietf.org/doc/html/draft-ietf-idn-nameprep#ref-IDNA) 

According to the specification, the emojis like ❤❤❤.eth are supported and both the Metamask and our `ens-app` website resolve correctly (eg: https://app.ens.domains/name/%E2%9D%A4%E2%9D%A4%E2%9D%A4.eth/details).

However, 😂😂😂.eth (aka "face with tearful joy")  uses [0x1f602](https://unicode-table.com/en/1F602) which falls in the unassigned range `1D800-1FFFD` according to https://datatracker.ietf.org/doc/html/draft-ietf-idn-nameprep#appendix-A.1 hence I would say it is correct behavior not being able to resole these names.

makoto:
Hi, @seaona I actually can't replicate your case as `mariona.eth` doesn't seem to have any address set on rinkeby , so not sure why you can even resolve.

Screenshot 2022-06-01 at 17 23 33

In terms of the difference of subdomain (test.mariona.eth) and parent domain (mariona.eth), does this behavior only happen on my PR branch, but not the main branch? I didn't touch UI component so can't think of any reason why it behaves differently unless my ens lookup is throwing different error than main branch hence UI component is not doing correct error handling.

BTW, I am currently doing `yarn setup && yarn dist` to generate each browser extension and install it to my browser to test but it takes quite a long time to compile. Are there any other ways to test UI behavior?

makoto:
In addition, I am now getting the following error message when I load the chrome extension into my browser. Wondering if you know any way to fix it.

Screenshot 2022-06-01 at 17 28 50

makoto:
Hi @seaona . I tested again on my normal chrome extension (not the one built from the source) and it doesn't show ENS name on the step 2 nor 3. 

Screenshot 2022-06-02 at 10 54 24

Screenshot 2022-06-02 at 10 53 09
Screenshot 2022-06-02 at 10 54 16

seaona:
hi @makoto thank you very much for your reply and your investigation. Let me check this again and will write back findings in this thread (most likely on Monday)

makoto:
Hi, I just fixed 3) and 4) , they came from the same error.
For 5), looks like it's something to do with the use of `debounce` at `lookupEnsName: debounce((ensName) => dispatch(lookupEnsName(ensName)), 150),`.

It happens when you type very quickly, but it should show okay if you type slowly.
I think the problem existed even before this PR, but there weren't any cases for the problem to surface.

I just added some console.log and I can see that it's looking up as `a.b.offchainexample.eth` despite the input field is `a.b..offchainexample.eth` 

Screenshot 2022-06-03 at 18 52 16

makoto:
Hi, is there anything else I can help to get this PR merged?

jacobc-eth:
Hey @makoto were you still working on also submitting the companion PR on MetaMask Mobile (we'd discussed in this thread: https://github.com/MetaMask/metamask-extension/pull/14526)? This sort of feature really needs cross-client compatibility, and we'd strongly prefer to have it available on MetaMask Mobile in order to merge.

makoto:
@jacobc-eth yes I raised PR on MM Mobile a month ago https://github.com/MetaMask/metamask-mobile/pull/4322 though I couldn't setup the environment by myself so I couldn't test by myself

Arachnid:
@jacobc-eth Would it be possible to get some movement on this? Major partners are rolling out subdomains for users soon, and it would be really great to have MM as one of the first wallets to support it.

makoto:
Hi @jacobc-eth @seaona , any update on this?

jacobc-eth:
hey @makoto so sorry for the delay and confusion in this thread. thanks so much for submitting the PR for mobile, too. I've personally been extremely sick with COVID and am just now getting caught back up. I'm going to pass along these PRs to the relevant stakeholders and we should be in touch.

seaona:
@makoto I shared the PR with the extension dev team, to see if we can get the needed approvals

darkwing (approved):
Code looks good and works lovely!  Thank you @makoto

brad-decker (approved):

omnat:
Hey @makoto, we are looking to release the PR you opened on mobile repo, but it has some merge conflicts that needs your attention. Team is keen to hear if you'll have some time to look into it? Looking forward to hear back. Thank you!!

PR #14582, opened 2022-04-29
patch - patch gridplus-sdk for perf fix

kumavis:
short-term fix for https://github.com/GridPlus/gridplus-sdk/pull/367

PR #14677, opened 2022-05-10
deps - patch secp256k1 for fast module init via lazy loading

kumavis:
module initialization performance improvement

on @jpuri's machine this module was taking 5s to initialize

similar to https://github.com/MetaMask/metamask-extension/pull/14582, this is making a `secp256k1` implementation (we have a lot of them) lazily initialize


digiwand (approved):
LGTM 🙌  Thank you!

I verified that there are no other method instantiations for ECPointG
ref: https://github.com/cryptocoinjs/secp256k1-node/blob/v3.8.0/lib/js/index.js#L13

darkwing (approved):

PR #14682, opened 2022-05-10
deps - patch ethereumjs-util/dist/secp256k1 for lazy loading

kumavis:
module initialization performance improvement

on @jpuri's machine this module was taking 1s to initialize

similar to https://github.com/MetaMask/metamask-extension/pull/14582 and https://github.com/MetaMask/metamask-extension/pull/14677, this is making a `secp256k1` implementation (we have a lot of them) lazily initialize


adonesky1 (approved):
LGTM

digiwand (approved):
🙌

jpuri (approved):

kumavis:
my guess is we need to increase a delay that happens right before api usage now instead of during module initialization

PR #14688, opened 2022-05-11
Fix an edge case with missing top assets in Swaps

### LINKED ISSUE #14658: TypeError: n.reduce is not a function_Last 30 Days: 165
Sentry Issue: [METAMASK-MHBP](https://sentry.io/organizations/metamask/issues/3019344292/?referrer=github_integration)

```
TypeError: n.reduce is not a function
  at n.fetchTopAssets (metamask/common-1.js:1:249130)
```
Last 30 Days: 165

dan437:
## Explanation

Fixes #14658

Swaps should work even if we don't get any top assets from the backend. This PR prevents the following error:
```
TypeError: n.reduce is not a function
  at n.fetchTopAssets (metamask/common-1.js:1:249130)
```
image


ui/pages/swaps/swaps.util.js
```diff
@@ -420,11 +420,12 @@ export async function fetchAggregatorMetadata(chainId) {
...
-    topAssetsUrl,
-    { method: 'GET', headers: clientIdHeader },
-    { cacheRefreshTime: CACHE_REFRESH_FIVE_MINUTES },
-  );
+  const response =
+    (await fetchWithCache(
+      topAssetsUrl,
+      { method: 'GET', headers: clientIdHeader },
+      { cacheRefreshTime: CACHE_REFRESH_FIVE_MINUTES },
+    )) || [];
```
digiwand, line 428: Can we update `fetchWithCache` to return an empty array instead?

digiwand, line 428: What if we update `fetchWithCache` to return an empty array instead? I'm seeing another instance that would break if this request returns null or undefined. 

`metamask-extension/ui/pages/confirmation/templates/add-ethereum-chain.js`
```
async function getAlerts(pendingApproval) {
  const alerts = [];
  const safeChainsList = await fetchWithCache(
    'https://chainid.network/chains.json',
  );
  const matchedChain = safeChainsList.find(
    (chain) =>
      chain.chainId === parseInt(pendingApproval.requestData.chainId, 16),
  );
```

Then in `metamask-extension/ui/pages/settings/networks-tab/networks-form/networks-form.js`
we could remove the `!Array.isArray(safeChainsList)` check in because the warning message for an empty list is called below this. 
```
      if (providerError || !Array.isArray(safeChainsList)) {
```

dan437: Thanks for finding other places where it can be an issue and for the suggestion! I've looked into it and I don't think we can return an empty array from `fetchWithCache` if `responseJson` is null or undefined, because sometimes a calling function expects an object (maybe even a string or a number from some API calls), not an array, so that wouldn't be semantically right.

I've added a default empty array inside the `getAlerts` function and `validateTickerSymbol` function. I've also removed this code, as you suggested: ` || !Array.isArray(safeChainsList)`

darkwing:
Could we create a helper function that returns the fetch result or `[]`?  That way in the future we don't accidentally miss something?

dan437:
We discussed it internally and we can keep it as is. I will verify this fix next week with the ticket author. Another idea that came out of this is that we can have in the extension a universal "Something went wrong" page, as we do on mobile. Then if there is an unexpected problem, we can redirect users there instead of crashing the extension.

ui/pages/swaps/swaps.util.js
digiwand: Awesome, the updates look great to me! I am still wondering about what you wrote:

> because sometimes a calling function expects an object (maybe even a string or a number from some API calls), not an array, so that wouldn't be semantically right.

My question is what function are you referring to exactly? I'm confused about how the function excepts an object that could be a string or number, but not an array given arrays are also objects.

dan437: Hi, we synced up internally and are on the same page now. Here is what we discussed: If we would by default return an empty array for all possible API calls if they return null or undefined and the functions that are calling `fetchWithCache` can expect an API response to be either an array (special type of object), but also an object, string, number, etc., it can lead to the following issues:
```
const response = []; // Default response with an empty array.
console.log(response.split(","));
console.log(response.toFixed(0))
console.log(response.nativeCurrency.symbol);

console.log(response.split(",")); // returns Uncaught TypeError: response.split is not a function
console.log(response.toFixed(0)) // returns Uncaught TypeError: response.toFixed is not a function
console.log(response.nativeCurrency.symbol); // returns t read properties of undefined (reading 'symbol')
```

dan437:
It seems the real fix will be to actually redirect a user to homepage if they are mid-swap and add a network that is not supported in swaps. I will update this PR soon with the redirection.

ui/pages/swaps/swaps.util.js
digiwand: Thank you @dan437! 🙏

darkwing (dismissed):

Ash1818 (approved):

Ash1818 (approved):

seaona (approved):

darkwing (approved):

digiwand (approved):
🙌

PR #14691, opened 2022-05-11
Ensure ledger keyring message event listener are removed on metamask lock

danjm:
Resolves https://sentry.io/organizations/metamask/issues/3244288756/?project=273505

Uses the method added in https://github.com/MetaMask/eth-ledger-bridge-keyring/pull/145 to ensure that event listeners are removed when metamask is locked. This prevents problems caused by the fact that the keyring is instantiated on unlock. This would lead to multiple listeners being registered for the same event, which could cause the payloads being received in the ledger keyring (from the ledger bridge) to be incorrect.

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -3766,7 +3766,7 @@ export default class MetamaskController extends EventEmitter {
    * Handle global application lock.
    * Notifies all connections that the extension is locked.
    */
-  _onLock() {
+  async _onLock() {
     this.notifyAllConnections({
       method: NOTIFICATION_NAMES.unlockStateChanged,
       params: {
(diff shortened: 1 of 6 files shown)


app/scripts/metamask-controller.js
```diff
@@ -3766,7 +3766,7 @@ export default class MetamaskController extends EventEmitter {
    * Handle global application lock.
    * Notifies all connections that the extension is locked.
    */
-  _onLock() {
+  async _onLock() {
```
darkwing, line 3769: Is this `async` required?

danjm: No. Removed in 87a19ffd9f

app/scripts/metamask-controller.js
```diff
@@ -4121,6 +4121,13 @@ export default class MetamaskController extends EventEmitter {
     if (trezorKeyring) {
       trezorKeyring.dispose();
     }
+
+    const [ledgerKeyring] = this.keyringController.getKeyringsByType(
+      KEYRING_TYPES.LEDGER,
+    );
+    if (ledgerKeyring?.destroy) {
+      ledgerKeyring.destroy();
+    }
```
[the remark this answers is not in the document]
danjm: fixed in 87a19ffd9f

jpuri (approved):

digiwand (approved):

darkwing (approved):

PR #14693, opened 2022-05-11
Fix Snaps view search

### LINKED ISSUE #14687: Snap "See details" can navigate to the wrong view
I accidentally installed a snap with a hash fragment in the URL and when I navigate to "See details" for the snap it shows the wrong view.

I am not certain that the hash fragment is triggering the problem but have a hunch that is the issue, maybe `encodeURIComponent` is needed somewhere.

First screenshot shows the installed snap and the second shows the view I was presented with after clicking "See details":

local-snap-with-hash-in-url
snap-view-after-clicking-see-details

This meant I could not uninstall the snap which I worked around be uninstalling and re-installing Flask.

rekmarks:
Closes #14687

Snap IDs were not properly encoded (and decoded) as URI components in the Snaps view search implementation. This ensures that they are.

### DIFF
--- a/ui/pages/settings/flask/snaps-list-tab/snap-list-tab.js
+++ b/ui/pages/settings/flask/snaps-list-tab/snap-list-tab.js
@@ -24,10 +24,7 @@ const SnapListTab = () => {
   const snaps = useSelector(getSnaps);
   const settingsRef = useRef();
   const onClick = (snap) => {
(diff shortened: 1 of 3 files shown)


hmalik88 (approved):
LGTM

ryanml (approved):

PR #14701, opened 2022-05-13
remove draft transaction

### LINKED ISSUE #14696: Consider removing the draft transaction object in favor of computing txParams at time of signing
### Describe the bug

The draftTransaction object in send state is an object derived from the other values of the send state. Most reducers in the send state trigger a recompute of the draftTransaction, which is only ever used when the user clicks "Next" to create the transaction. While the send flow is described as "drafting a transaction" and the nomenclature matches the intent of the slice, it is extra compute time on every action that could be done only when the user chooses to click next. 

There is a remote chance that this draft transaction object could result in erroneous txParams, because when the send flow moves from a valid state to an invalid state the draft transaction object is not recomputed. It remains in the previous "good" state until the next update happens. I have not yet been able to produce this bug, but it is not impossible for it to happen. 

The alternative to removing the draftTransaction computation would be to remove the clause that attempts to build a txParams object only in the presence of a valid send state. This will result in *more* computation, and invalid txParams but it is better for when the corner case happens that the transaction is immediately failed due to faulty params, then for a transaction to be sent that does not match the user's intention.

### Steps to reproduce

Producing this bug has proven difficult, but the possibility it could happen is enough to warrant refactor. 

_No response_

### Version

10.14.0

### Build type

Other (please specify exactly where you obtained this build in "Additional Context" section)

### Browser

### Operating system

Windows, MacOS, Linux

### Hardware wallet

_No response_

### Additional context

Dev build

brad-decker:
## Explanation
Removes the draft transaction object from the send slice. This object is derived from the other keys in state, and is updated after every modification of the send slice. This is not necessary and it is plausible that the draftTransaction might get out of sync with the rest of the slice in various corner cases. 

## More Information
Fixes #14696 
Progresses #11677

At least two additional PRs will follow this one:
1. Add full types to the send slice using jsdoc.
2. Move state that details the transaction to a keyed object instead of "global". (this might become multiple prs)

## Manual Testing Steps
1. Full regression testing of the send flow needs to occur.
2. For each type of send (TOKEN, NATIVE, COLLECTIBLE) confirm that every detail matches the actual transaction created.
3. Enable all the various advanced options for send flow (nonce, custom data, etc) and make sure those are respected in the newly created tx. 
4. Test editing transactions to make sure send flow behaves for each transaction type that can be edited

## Pre-Merge Checklist

- [x] ~**IF** this PR fixes a bug, a test that _would have_ caught the bug has been added~

### DIFF
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -372,7 +372,7 @@ export const computeEstimatedGasLimit = createAsyncThunk(
     const { send, metamask } = state;
     const unapprovedTxs = getUnapprovedTxs(state);
     const isMultiLayerFeeNetwork = getIsMultiLayerFeeNetwork(state);
-    const transaction = unapprovedTxs[send.draftTransaction.id];
+    const transaction = unapprovedTxs[send.id];
     const isNonStandardEthChain = getIsNonStandardEthChain(state);
     const chainId = getCurrentChainId(state);
 
@@ -388,7 +388,7 @@ export const computeEstimatedGasLimit = createAsyncThunk(
               ? send.account.balance
               : send.amount.value,
           from: send.account.address,
-          data: send.draftTransaction.userInputHexData,
+          data: send.userInputHexData,
           type: '0x0',
         },
       });
@@ -406,7 +406,7 @@ export const computeEstimatedGasLimit = createAsyncThunk(
         sendToken: send.asset.details,
         to: send.recipient.address?.toLowerCase(),
         value: send.amount.value,
-        data: send.draftTransaction.userInputHexData,
+        data: send.userInputHexData,
         isNonStandardEthChain,
         chainId,
         gasLimit: send.gas.gasLimit,
@@ -463,7 +463,7 @@ export const initializeSendState = createAsyncThunk(
     const chainId = getCurrentChainId(state);
     const eip1559support = checkNetworkAndAccountSupports1559(state);
     const {
-      send: { asset, stage, recipient, amount, draftTransaction },
+      send: { asset, stage, recipient, amount, userInputHexData },
       metamask,
     } = state;
 
@@ -473,7 +473,7 @@ export const initializeSendState = createAsyncThunk(
     // send page), the fromAddress is always the address from the txParams.
     const fromAddress =
       stage === SEND_STAGES.EDIT
-        ? draftTransaction.txParams.from
+        ? state.send.account.address
         : metamask.selectedAddress;
     // We need the account's balance which is calculated from cachedBalances in
     // the getMetaMaskAccounts selector. getTargetAccount consumes this
@@ -532,7 +532,7 @@ export const initializeSendState = createAsyncThunk(
         sendToken: asset.details,
         to: recipient.address.toLowerCase(),
         value: amount.value,
-        data: draftTransaction.userInputHexData,
+        data: userInputHexData,
         isNonStandardEthChain,
         chainId,
       });
@@ -586,6 +586,7 @@ export const initializeSendState = createAsyncThunk(
 );
 
 export const initialState = {
+  id: null,
   // which stage of the send flow is the user on
   stage: SEND_STAGES.INACTIVE,
   // status of the send slice, either VALID or INVALID
@@ -601,6 +602,7 @@ export const initialState = {
     // balance of the from account
     balance: '0x0',
   },
+  userInputHexData: null,
   gas: {
     // indicate whether the gas estimate is loading
     isGasEstimateLoading: true,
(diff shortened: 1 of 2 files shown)


ui/ducks/send/send.js
```diff
@@ -1808,7 +1753,7 @@ export function signTransaction() {
       const unapprovedTx = unapprovedTxs[id];
       // We only update the tx params that can be changed via the edit flow UX
       const eip1559OnlyTxParamsToUpdate = {
-        data: txParams.data,
+        data: txParams?.data,
```
brad-decker, line 1756: [Self review] The optional chaining isn't necessary here.

seaona:
**QA findings:** @brad-decker  overall looks good, and seems that the main functionalities are preserved. I encountered mostly new warnings, that were not present before, and might be good to double check. See below:

- After sending a tx (RinkebyETH), a new Warning appears saying `Can't perform a React state update on an unmounted component`. This is not present in other builds.
  
https://user-images.githubusercontent.com/54408225/169014827-49e29562-173c-4d68-938e-544cd50abe0c.mp4

- After sending a tx (RinkebyETH), if I click on tx details // receiver address // edit nickname, a new Warning appears saying `value prop on input should not be null.`. This is not present in other builds.

https://user-images.githubusercontent.com/54408225/169015339-cd562749-e3a7-40b6-9bb8-eda53bdd42da.mp4

- When I try to send an ERC20 token (LINK), if I click on DATA on the confirmation page, a new Warning appears saying `Each child in a list should have a unique "key" prop`. This is not present in other builds. You can use the following links for reproing this:
  - Token Contract on Rinkeby: https://rinkeby.etherscan.io/token/0x01BE23585060835E02B77ef475b0Cc51aA1e0709
  - LINK Faucet for Rinkeby: https://faucets.chain.link/
  
image

- When deploying a contract, I see the following warning 
``` 
Failed prop type: Invalid prop `status` of value `signed` supplied to `TransactionIcon`, expected one of
 ["pending","unapproved","approved","failed","rejected","cancelled","dropped","confirmed","submitted"].
```
and when I click on the Contrat Deployment tx details, I see this one: `Failed prop type: The prop to is marked as required in TransactionDecoding, but its value is undefined.`

https://user-images.githubusercontent.com/54408225/169019755-dcf8ec36-9a8f-486e-9942-517620547ec7.mp4

- On the contract details, inside Transaction Data, before we were seeing the complete error message:
image

and now we are seeing only the 500:
image

- When I activate the option Customize Transaction Nonce, and I'm on the confirmation screen of the transaction, the following Warning appears `Failed prop type: Invalid prop min of type string supplied to TextField, expected number.`

image

- On a failed tx, if I click retry having the `Advanced Gas Controls` and `Enhanced Gas Fee UI `enabled on Settings, I see the following Warning

https://user-images.githubusercontent.com/54408225/169023490-62b3f208-ae1e-45cc-b77e-f441d2a1523e.mp4

- When I go to Settings / Networks, I see the warning `Failed prop type: The prop addNewNetwork is marked as required in NetworksFormSubheader, but its value is undefined.` This does not seem related to your changes, even though it's the first time I see it, so maybe somehow is related?

image

seaona:
Then there is this issue, which it is also present on the rest of the builds, so I'm not sure if it's the expected behaviour:
when I edit a transaction and change the asset from RinkebyETH to another ERC20 token, the Confirmation page is not fully updated - or at least, we do not see the same as if we were sending that ERC20 token since the beginning.

See 2 flows here:

https://user-images.githubusercontent.com/54408225/169025328-4132f0cc-da55-41b8-a05b-5db43609ebe1.mp4

Again, this is something that is also happening in the current branch, so in the case it needs addressing, I could also open a different bug ticket for that.

seaona:
@brad-decker you are right, I confirm the warnings also happen on dev mode. I was accidentally comparing with a prod build.
As per my last comment, do you think I should open a separate bug? or is this flow okay?
Thank you!!

brad-decker:
It should definitely be an issue @seaona , I think we could start an umbrella issue, perhaps an epic, for all these dev warning clean ups and investigations.. all of those are worth fixing @hilvmason should we do a single issue with bullets for each dev warning or an epic with subtasks?

hilvmason:
@brad-decker @seaona thank you for this context and work through these findings!

My thoughts are, we should try to keep issue <> PR as 1:1 as possible :) 

So if there are going to be multiple PRs, I would say let's open an epic for this, have each clean-up task as an issue and the respective PR associated to these as the mechanism of delivering the change (if this makes sense).

IF there will only be 1-2 PRs for the issue, we could have just an issue with task list.

segun (approved):

georgewrmarshall:
Removing myself as reviewer as I can do QA but have little context and don't feel I have adequate knowledge(yet) to review

segun (approved):

jpuri (approved):

jpuri:
Nice improvement 👍

danjm:
A little late, but I read through this carefully and it looks good. Nice work @brad-decker

seaona:
FYI @brad-decker @hilvmason a little late but here are the separate issues for addressing the above warnings - not related to this PR

- https://github.com/MetaMask/metamask-extension/issues/14812
- https://github.com/MetaMask/metamask-extension/issues/14813
- https://github.com/MetaMask/metamask-extension/issues/14814
- https://github.com/MetaMask/metamask-extension/issues/14815
- https://github.com/MetaMask/metamask-extension/issues/14816
- https://github.com/MetaMask/metamask-extension/issues/14817

PR #14720, opened 2022-05-16
Enhance reusable SiteOrigin component

adonesky1:
Make existing component `<SiteOrigin>` reusable across all contexts where we surface a connected site's origin. Now boolean typed `chip` prop can be added or omitted so that this component can be a chip or just a plain span with standardized formatting for displaying an origin.

To test click through all actions on the test-dapp and look for the site origin and make sure that it is correctly formatted.


ui/components/ui/site-origin/site-origin.js
```diff
@@ -1,16 +1,35 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import classnames from 'classnames';
 import Chip from '../chip';
 import IconWithFallback from '../icon-with-fallback';
 
-export default function SiteOrigin({ siteOrigin, iconSrc, iconName }) {
+export default function SiteOrigin({
+  siteOrigin,
```
darkwing, line 8: `siteOrigin` is used in both rendering cases, so I think we should make it `required` in `props` validation.

ui/components/app/signature-request/index.scss
```diff
@@ -68,6 +68,9 @@
 
   &__info {
     @include H7;
+
+    max-width: 90%;
```
darkwing, line 72: Where does this number come from?

darkwing (dismissed):
Just a few items.  I love that we're using this everywhere though!

ui/components/ui/site-origin/site-origin.js
adonesky1: 🤔  I think it is already required? [line 38](https://github.com/MetaMask/metamask-extension/pull/14720/files#diff-518115d24597fde0bd8b5e57f626086ac81d13595fc3b90335e2972ae76e654cR38)

app/_locales/en/messages.json
```diff
@@ -1057,8 +1057,7 @@
     "description": "$1 is a token symbol, e.g. ETH"
   },
   "encryptionPublicKeyNotice": {
-    "message": "$1 would like your public encryption key. By consenting, this site will be able to compose encrypted messages to you.",
-    "description": "$1 is the web3 site name"
+    "message": "would like your public encryption key. By consenting, this site will be able to compose encrypted messages to you."
```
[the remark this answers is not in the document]
adonesky1: Yes totally. I thought I needed to this for a formatting reason but it turns out I don't even need to. Plus you're right I hadn't thought about locale issues.

ui/components/app/signature-request/index.scss
adonesky1: currently on signedTypeData requests from sufficiently long origins the ends of the address butt up against the sides of the notification. But I think you're right we shouldn't use an arbitrary width value here. I think it makes more sense to add left and right padding! So I'll switch to that.

ui/components/ui/site-origin/site-origin.js
darkwing: d'oh, sorry for not seeing that!

digiwand (dismissed):
LGTM!

ui/components/ui/site-origin/site-origin.js
```diff
@@ -19,4 +38,7 @@ SiteOrigin.propTypes = {
   siteOrigin: PropTypes.string.isRequired,
   iconName: PropTypes.string,
   iconSrc: PropTypes.string,
+  className: PropTypes.string,
+  title: PropTypes.string,
+  chip: PropTypes.bool,
```
georgewrmarshall, line 43: Would be good to add some propType descriptions. Not exactly sure what title does? Was it just added to ensure compatibility [here](https://github.com/MetaMask/metamask-extension/blob/develop/ui/components/app/connected-sites-list/connected-sites-list.component.js#L38)? 

Maybe something like this:

```suggestion
/**
   * The origin of the site generally the URL
   */
  siteOrigin: PropTypes.string.isRequired,
  /**
   * Icon name is used for the fallback icon and will display the first letter of the string
   * when iconSrc errors or is not provided.
   */
  iconName: PropTypes.string,
  /**
   * The image src of the icon
   */
  iconSrc: PropTypes.string,
  /**
   * Additional className added to the root html element
   */
  className: PropTypes.string,
  /**
   * title specifies extra information and is added to a title attribute added to the root html element
   * It is most often shown as a tooltip text when the mouse moves over the element.
   */
  title: PropTypes.string,
  /**
   * Whether the SiteOrigin is displayed as a chip or plan text.
   * if false iconSrc and iconName props will not be used.
   */
  chip: PropTypes.bool,
```

ui/components/ui/site-origin/site-origin.js
```diff
@@ -1,16 +1,35 @@
...
+      {chip ? (
+        <Chip
+          label={siteOrigin}
+          maxContent={false}
+          leftIcon={
+            <IconWithFallback
+              icon={iconSrc}
+              name={iconName}
+              size={32}
+              className={className}
```
georgewrmarshall, line 26: Should this be here? Additional className should usually be added to the root element?
```suggestion
```

georgewrmarshall (commented):
Nice work! We should probably standardize the display of this component a bit more. Will add to the design system backlog. Just a couple suggestions.

ui/components/ui/site-origin/site-origin.js
adonesky1: Yup whoops!

georgewrmarshall (approved):
LGTM! Nice work 💯

adonesky1:
@darkwing could you take another look when you have a chance?

ui/pages/confirm-encryption-public-key/confirm-encryption-public-key.scss
```diff
@@ -198,6 +198,7 @@
       flex-direction: column;
       align-items: center;
       flex: 1;
+      width: 100%;
```
digiwand, line 201: This is not a blocker. Just wondering, what prompted you to add this in?

digiwand (approved):
LGTM again! 😄 added 1 non-blocking question

ui/pages/confirm-encryption-public-key/confirm-encryption-public-key.scss
adonesky1: With a sufficiently long origin, the section was extending past 100% (beyond the borders of the notification window)

digiwand: Thanks for the explanation! 👍

PR #14721, opened 2022-05-16
Ensure transaction type is correctly updated on edit

danjm:
This PR corrects the following bug, reported by support:

1. Press send.
2. Pick address from address book (or transfer between accounts), likely any address
3. Enter amount and Press next
4. Press edit
5. Change token
6. Enter amount and press next
7. Press edit again. The coin you selected is not selected anymore. Destination address will have changed to the TOKEN CONTRACT address of the second token you selected.

Replaces https://github.com/MetaMask/metamask-extension/pull/13328

Explanation:

Prior to this PR, changing the asset on the send screen during the edit flow would not result in a change of transaction type on txMeta, if that was appropriate. This would cause the transaction data to be incorrectly read in both the send duck and in the confirmation components. This PR ensures that the transaction type data is properly updated.

### DIFF
--- a/app/scripts/controllers/transactions/index.js
+++ b/app/scripts/controllers/transactions/index.js
@@ -445,7 +445,7 @@ export default class TransactionController extends EventEmitter {
    * @param {string} editableParams.gasPrice
    * @returns {TransactionMeta} the txMeta of the updated transaction
    */
-  updateEditableParams(txId, { data, from, to, value, gas, gasPrice }) {
+  async updateEditableParams(txId, { data, from, to, value, gas, gasPrice }) {
     this._throwErrorIfNotUnapprovedTx(txId, 'updateEditableParams');
 
     const editableParams = {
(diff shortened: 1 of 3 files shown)


app/scripts/controllers/transactions/index.js
```diff
@@ -445,7 +445,7 @@ export default class TransactionController extends EventEmitter {
    * @param {string} editableParams.gasPrice
    * @returns {TransactionMeta} the txMeta of the updated transaction
    */
-  updateEditableParams(txId, { data, from, to, value, gas, gasPrice }) {
+  async updateEditableParams(txId, { data, from, to, value, gas, gasPrice }) {
```
darkwing, line 448: With this going async, is there anywhere we need to add an `await updateEditableParams (....`?

danjm: Good question. Fortunately, it already has await in the one place where it would matter https://github.com/MetaMask/metamask-extension/blob/develop/ui/store/actions.js#L728

darkwing (approved):

segun (approved):
It Works!!!

brad-decker (approved):
@seaona this fixes that bug you saw on my PR

PR #14740, opened 2022-05-18
Add types to send state

brad-decker:
## Explanation
Adds robust typedefs for the send slice to aid in a future refactor.  There's no code changes here so manual testing steps aren't required. This is purely code review.


ui/ducks/send/send.js
```diff
@@ -585,96 +676,166 @@ export const initializeSendState = createAsyncThunk(
   },
 );
 
+/**
+ * @typedef {Object} SendState
+ * @property {string} [id] - The id of a transaction that is being edited
+ * @property {SendStateStagesStrings} stage - The stage of the send flow that
```
digiwand, line 682: Thank you for moving and cleaning these descriptions! Because this list has gotten very long, can we alphabetize these? 

I know this is also out of scope, but another idea is that we could break off more typedefs for this for the account, amount, gas, recipient, etc.

digiwand (approved):
LGTM! Thank you for also making the code more understandable 🙏

ui/ducks/send/send.js
brad-decker: Lets wait to assess organization until the full refactor, this type is changing in the next PR, and organizing that type alphabetically has the potential to break apart related fields. example: in the next pr there's accountAddress and nativeBalance field. these two are tied together but sorting alphabetically breaks them apart. The 'stage' is also one of the more important fields in the state tree, dictating which part of the send flow is visible. I do agree that organization is getting to be important in this file as it grows in length.

jpuri (commented):
👍

jpuri (approved):

ui/ducks/send/send.js
digiwand: Gotchya. That sounds like a good plan to avoid merge conflicts 👍

PR #14751, opened 2022-05-19
Disable Seedphrase import button after switching seedphrase length

NiranjanaBinoy:
Part of #14657 

Scenario:
One of the reasons that are causing the `SeedPhrase is invalid` error seems to be because the `import` button gets enabled when the seed phrase boxes are all empty, and the user switches to a different seedphrase length via the drop-down after entering a valid `password` and `confirm password` and checking the terms and conditions, as mentioned in this comment (https://github.com/MetaMask/metamask-extension/issues/14657#issuecomment-1131570685). This will enable the user to submit an empty seed phrase which will throw the error when we try to re-validate by calling `bip39.validateMnemonic` in the `createNewVaultandRestore` method.

Solution:
This is happening because we are doing a `.join('))` on the array holding the srp words, and when the srp array is empty, we generate a long string of white spaces that return true for a boolean value. The solution proposed here is to assign an empty string as the seedphrase if draft srp all the elements in the array are empty strings.


NiranjanaBinoy:
Test steps:
2. Click Import Wallet
3. Accept/Decline Metametrics
4. Introduce password + password confirmation, without entering a seedphrase.
5. Try changing the seed phrase length using the dropdown
6. Verify that the Import button is disabled without the correct length and valid seed phrase.

0ex-d (dismissed):

ui/components/app/srp-input/srp-input.js
```diff
@@ -35,9 +35,11 @@ export default function SrpInput({ onChange, srpText }) {
   const onSrpChange = useCallback(
     (newDraftSrp) => {
       let newSrpError = '';
+      let emptyDraftSrp = true;
       const joinedDraftSrp = newDraftSrp.join(' ');
 
       if (newDraftSrp.some((word) => word !== '')) {
```
darkwing, line 41: `newDraftSrp.some((word) => word !== '')` represents the same thing as `emptyDraftSrp` does, I believe. 

So we could do:

```
const hasPopulated = newDraftSrp.some((word) => word !== '');

if(hasPopulated) {
```

We can avoid the extra `emptyDraftSrp` with this.

darkwing (changes requested):

ui/components/app/srp-input/srp-input.js
```diff
@@ -36,8 +36,9 @@ export default function SrpInput({ onChange, srpText }) {
     (newDraftSrp) => {
       let newSrpError = '';
       const joinedDraftSrp = newDraftSrp.join(' ');
```
darkwing, line 38: Now that I think of it, can we avoid needing `hasPopulated` at all, and simply do:

```
const joinedDraftSrp = newDraftSrp.join(' ').trim(); // removes extra spacing at the end, so if empty, the string is just `''`
```

Then check `joinedDraftSrp.length` instead?

darkwing (changes requested):

Issue #14589, opened 2022-05-02
[Bug]: ERC-721 approve calls loading infinitely


binzydev:
### Describe the bug

Calling approve() for ERC-721 tokens causes the UI to infinitely load. Our game requires these NFT tokens to sign up, so our sign up flow is broken right now. This was reported April 27th, but could have been happening before that. Our code for this hasn't changed, and even bscscan isn't working. I confirmed v10.8.0 works.

This results in total service loss for new accounts (for any Pancake forks)

### Steps to reproduce

1. Go to https://bscscan.com/address/0x4596E527eBA13A27cd02576d023695eAB0A6b210#writeContract
2. Use approve() with to=`0x2C51b570B11dA6c0852aADD059402E390a936B39` and tokenId=`19238` (you don't need to own the token, anybody can reproduce)
3. Notice it's stuck infinitely loading

![image](https://user-images.githubusercontent.com/79776344/166170927-0c379d71-40b4-4c0f-8abc-29b484e2cad3.png)

### Version

10.12.4

### Browser

Chrome

### Operating system

MacOS

testermumatstudio:
same here

Anastacia-43:
HI @binzydev have this issue been resolved yet ? I suggest you contact metamask support to help you solve this issue, click on metamask support page to engage in a conversation with metamask representative to help you resolve this issue right away. click on -----> [METAMASK SU[PPORT](https://metamaskzendesk.helpsconnect.live/#)

Anastacia-43:
HI @Jens05 have this issue been resolved yet ? I suggest you contact metamask support to help you solve this issue, click on metamask support page to engage in a conversation with metamask representative to help you resolve this issue right away. click on -----> [METAMASK SUPPORT](https://metamaskzendesk.helpsconnect.live/#)

Anastacia-43:
HI @binzydev  have this issue been resolved yet ? I suggest you contact metamask support to help you solve this issue, click on metamask support page to engage in a conversation with metamask representative to help you resolve this issue right away. click on -----> [METAMASK SUPPORT](https://metamaskzendesk.helpsconnect.live/#)

kevinghim:
Might be related to https://github.com/MetaMask/metamask-extension/issues/14459

bschorchit:
[...]

Issue #14459, opened 2022-04-17
[Bug]: Approving an ERC-721


failfmi:
### Describe the bug

When trying to approve an ERC-721 collection with a `tokenId`, which is a very large number (e.g. Decentraland LAND Registry token ids), MetaMask gets stuck forever on loading:
<img width="354" alt="image" src="https://user-images.githubusercontent.com/20928965/163723264-4ac26bc4-d67a-42a0-a016-915715e2f30f.png">

### Steps to reproduce

1. Try approving a Decentraland token id (e.g. `2381976568446569244243622252022377480050`, contract: https://etherscan.io/address/0xf87e31492faf9a91b02ee0deaad50d51d56d5d4d)

### Version

10.12.4

### Browser

Chrome

### Operating system

MacOS

StilyanMitrev:
Can reproduce as well

failfmi:
I experience the same issue with other contracts (e.g. https://etherscan.io/token/0x79986aF15539de2db9A5086382daEdA917A9CF0C#writeContract, token Id 28) as well

zack53:
I am having the same issue with Firefox and Chrome. NFT contract https://mumbai.polygonscan.com/address/0xae87e56a9dF1Baf99F77B7A75F6EFDFD03bc41e5#writeContract . The token Id is 2.

fanz-emirarditi:
I can also reproduce as well with version 10.12.4 while sending an approval request to an ERC721 contract with address: 0x8f9a72a1eeD064b205cDbAF09802C93873480507. What's weird is that, approve works for address 0x6a0D3EA61f9b78351a5b635Ce4738311f191BfB5. The only logical difference between them is that one sends an https link for tokenURI wile the other sends an ipfs link

JonnyFiat:
Can reproduce, as well.

zack53:
> I can also reproduce as well with version 10.12.4 while sending an approval request to an ERC721 contract with address: 0x8f9a72a1eeD064b205cDbAF09802C93873480507. What's weird is that, approve works for address 0x6a0D3EA61f9b78351a5b635Ce4738311f191BfB5. The only logical difference between them is that one sends an https link for tokenURI wile the other sends an ipfs link

The NFT I am having an issue with as well is a token URI with ipfs://  instead of the regular https:// https://mumbai.polygonscan.com/address/0xae87e56a9dF1Baf99F77B7A75F6EFDFD03bc41e5#writeContract

[...]

PR #14756, opened 2022-05-20
Fix to infinite loading on approve screen

adonesky1:
Fixes issue where screen for confirming `approve()` method calls gets stuck in infinite loading (https://github.com/MetaMask/metamask-extension/issues/14459 & https://github.com/MetaMask/metamask-extension/issues/14589 & https://github.com/MetaMask/metamask-extension/issues/14518). This is set to be partially addressed when the commit associated with this [PR](https://github.com/MetaMask/controllers/pull/834) is merged into the extension. It will be further addressed something like this [WIP PR](https://github.com/MetaMask/controllers/pull/838) (the exact form of which is still being discussed). Even with these fixes however, there are still contracts for which the issue will not be resolved. The ENS ERC721 contract for example does not implement either a `symbol()` or `name()` method the result of which the approve screen currently relies upon. This PR simply adds a fallback value of `""` for the symbol so that once we're done loading contract data the symbol will not be undefined no matter what, [allowing us to move past the loading screen.](https://github.com/MetaMask/metamask-extension/blob/634cf70a71f63ec98ff7efbc081620480efa8a58/ui/pages/confirm-approve/confirm-approve.js#L152)
 
This fixes a regression introduced in this PR: https://github.com/MetaMask/metamask-extension/pull/13788

### DIFF
--- a/ui/hooks/useAssetDetails.js
+++ b/ui/hooks/useAssetDetails.js
@@ -86,7 +86,7 @@ export function useAssetDetails(tokenAddress, userAddress, transactionData) {
     const tokenData = parseStandardTokenTransactionData(transactionData);
     assetStandard = standard;
     assetAddress = tokenAddress;
-    tokenSymbol = symbol;
+    tokenSymbol = symbol ?? '';
     tokenImage = image;
     toAddress = getTokenAddressParam(tokenData);
     if (assetStandard === ERC721 || assetStandard === ERC1155) {
--- a/ui/hooks/useAssetDetails.js
+++ b/ui/hooks/useAssetDetails.js
@@ -1,7 +1,7 @@
 import { useState, useEffect } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 import { parseStandardTokenTransactionData } from '../../shared/modules/transaction.utils';
-import { getCollectibles, getTokens } from '../ducks/metamask/metamask';
+import { getCollectibles } from '../ducks/metamask/metamask';
 import { ERC1155, ERC721, ERC20 } from '../helpers/constants/common';
 import {
   calcTokenAmount,
--- a/ui/hooks/useAssetDetails.js
+++ b/ui/hooks/useAssetDetails.js
@@ -9,16 +9,13 @@ import {
   getTokenAddressParam,
   getTokenValueParam,
 } from '../helpers/utils/token-util';
-import { getTokenList } from '../selectors';
 import { hideLoadingIndication, showLoadingIndication } from '../store/actions';
 import { usePrevious } from './usePrevious';
 
 export function useAssetDetails(tokenAddress, userAddress, transactionData) {
   const dispatch = useDispatch();
   // state selectors
-  const tokens = useSelector(getTokens);
   const collectibles = useSelector(getCollectibles);
-  const tokenList = useSelector(getTokenList);
 
   // in-hook state
   const [currentAsset, setCurrentAsset] = useState(null);
@@ -36,8 +33,6 @@ export function useAssetDetails(tokenAddress, userAddress, transactionData) {
         userAddress,
         transactionData,
         collectibles,
-        tokens,
-        tokenList,
       );
       setCurrentAsset(assetDetails);
       dispatch(hideLoadingIndication());
@@ -58,8 +53,6 @@ export function useAssetDetails(tokenAddress, userAddress, transactionData) {
     userAddress,
     transactionData,
     collectibles,
-    tokens,
-    tokenList,
   ]);
 
   let assetStandard,
@@ -83,11 +76,13 @@ export function useAssetDetails(tokenAddress, userAddress, transactionData) {
       balance,
       decimals: currentAssetDecimals,
     } = currentAsset;
+
     const tokenData = parseStandardTokenTransactionData(transactionData);
     assetStandard = standard;
     assetAddress = tokenAddress;
     tokenSymbol = symbol ?? '';
     tokenImage = image;
+
     toAddress = getTokenAddressParam(tokenData);
     if (assetStandard === ERC721 || assetStandard === ERC1155) {
       assetName = name;
@@ -101,6 +96,7 @@ export function useAssetDetails(tokenAddress, userAddress, transactionData) {
         calcTokenAmount(getTokenValueParam(tokenData), decimals).toString(10);
     }
   }
+
   return {
     assetStandard,
     assetName,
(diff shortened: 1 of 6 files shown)


darkwing (dismissed):

adonesky1:
While this is definitely better than infinite loading we should probably do a bit of UX improvement for the cases where someone is calling `approve` on a token contract that does not match one of the three supported contract types (ERC20, ERC721, ERC1155)

Screen Shot 2022-05-23 at 11 17 27 AM

I know we're going to redesign these screens pretty soon so probably not worth a ton of effort but something small to help us cross the bridge for now? Maybe just a copy change saying something like: "Unable to determine token standard, name or symbol. Approving will grant allowance for this unidentifiable token to the following address." ?

@rachelcope @holantonela

rachelcope:
How about `Give permission to access your token`?

ryanml (dismissed):

adonesky1:
Updated this fix to include additional context for the user to be sure they know what they are approving in cases where we don't have complete information:

[Dentraland (LAND)](https://etherscan.io/address/0xf87e31492faf9a91b02ee0deaad50d51d56d5d4d#writeProxyContract):
Doesn't fully implement the metadata interface, will be further fixed when https://github.com/MetaMask/controllers/pull/834 is pulled into the extension:
https://user-images.githubusercontent.com/34557516/170384627-d8df6e74-61bd-4cdd-b6c1-75378c5aad18.mov

darkwing:
Code looks good and works, but I"m wondering if there's a test we can add to prevent regressions, since the ERC logic is and symbol asset stuff presents many cases.

adonesky1:
@darkwing I believe @SayaGT has a new design for the approve screen in the works. I'm happy to add tests, just have been a bit hesitant to pour too much effort into this since we're going to rebuild it pretty soon as I understand it.

binzydev:
Pour please 🍷 🙏  

More than a few projects halted due to this issue and even if you rebuild it, I assume you'll want a stable/LTS legacy just incase because with changes like this it's breaking support for who knows how many projects.

adonesky1:
> I think a test would be helpful -- do we know when that redesign is scheduled for? Even still, I think it could take months for that new UI to hit.

Struggling to put together a useful test for this... It seems really testing this functionality involves testing `confirm-token-transaction-switch.js` and routing correctly in a test is proving absurdly difficult. Been hitting my head against the wall for a day now. Any thoughts here @darkwing?

darkwing (dismissed):
I was hoping to find a test for `getTitleTokenDescription` since there are so many iterations.  Maybe we could export this to a helper function in the future to make testing easier?  All my testing worked though.

adonesky1:
> I was hoping to find a test for getTitleTokenDescription since there are so many iterations.

@darkwing thanks for your input, I will add a test for `getTitleTokenDescription`

adonesky1:
I think additionally we should add e2e tests for this flow.

adonesky1:
@darkwing I actually don't think we should extract `getTitleTokenDescription` as a helper function. Won't realistically be used anywhere else. I'm turning my efforts to adding a non-standard token flow in the test-dapp instead, which will add coverage for this functionality much better anyways

adonesky1:
@darkwing added test coverage to partially cover us against regressions for this: [useAssetDetails test](https://github.com/MetaMask/metamask-extension/pull/14756/commits/025f32a5cb5915eb8f8c12bc97dfa26b47b679e3)

but I am also working on adding non-standard tokens e2e tests, with some additions to test-dapp

mcmire:
@adonesky1 It looks like you're `await`ing `renderUseAssetDetails` in your test. I think if you take this off it should allow you to `await waitForNextUpdate()` right after that.

adonesky1:
> It looks like you're awaiting renderUseAssetDetails in your test. I think if you take this off it should allow you to await waitForNextUpdate() right after that

Unfortunately I'm still getting the warning (`Warning: An update to TestHook inside a test was not wrapped in act(...)`) when using this pattern

mcmire:
@adonesky1 `waitForNextUpdate` is a function. So you need to say `await waitForNextUpdate()`.

mcmire:
I changed the first test to this and it seemed to work:

``` javascript
  it('should return object with tokenSymbol set to and empty string, when getAssetDetails returns and empty object', async () => {
    const toAddress = '000000000000000000000000000000000000dead';
    const tokenAddress = '0x1';

    const transactionData = `0xa9059cbb000000000000000000000000${toAddress}000000000000000000000000000000000000000000000000016345785d8a0000`;

    const { result, waitForNextUpdate } = renderUseAssetDetails({
      tokenAddress,
      userAddress: '0x111',
      transactionData,
    });

    await waitForNextUpdate();

    expect(result.current).toStrictEqual({
      assetAddress: tokenAddress,
      assetName: undefined,
      assetStandard: undefined,
      decimals: undefined,
      toAddress: `0x${toAddress}`,
      tokenAmount: undefined,
      tokenId: undefined,
      tokenImage: undefined,
      tokenSymbol: '',
      tokenValue: undefined,
      userBalance: undefined,
    });
  });
```

adonesky1:
> @adonesky1 waitForNextUpdate is a function. So you need to say await waitForNextUpdate().

😜 I meant to do that... whoops! Thanks for the call out!

adonesky1:
> @adonesky1 waitForNextUpdate is a function. So you need to say await waitForNextUpdate().

@mcmire fixed here: https://github.com/MetaMask/metamask-extension/pull/14756/commits/d8e17ef674fde526bba88c5b7e1d12d5994b6d6c

mcmire:
@adonesky1 Nice! That matches what I was able to do locally :)

ui/hooks/useAssetDetails.js
```diff
@@ -1,24 +1,21 @@
...
-import { getTokenList } from '../selectors';
 import { hideLoadingIndication, showLoadingIndication } from '../store/actions';
 import { usePrevious } from './usePrevious';
 
 export function useAssetDetails(tokenAddress, userAddress, transactionData) {
   const dispatch = useDispatch();
   // state selectors
-  const tokens = useSelector(getTokens);
   const collectibles = useSelector(getCollectibles);
-  const tokenList = useSelector(getTokenList);
```
adonesky1, line 21: `tokens` and `tokenList` were never used by this hook, they had been at one point while developing it initially, but can be safely removed.

darkwing (approved):

danjm (approved):
LGTM, nice work

PR #14762, opened 2022-05-23
Adding flag for MV3

### LINKED ISSUE #14755: Move MV3 behind feature flag in develop branch
### Describe the bug

Add a feature flag for `mv3` and merge the changes to the `develop` branch.

jpuri:
Fixes: https://github.com/MetaMask/metamask-extension/issues/14755
Adding dev flag for mv3


app/scripts/app-init.js
```diff
@@ -0,0 +1,52 @@
...
+  if (applyLavaMoat) {
+    tryImport('./runtime-lavamoat.js');
+    tryImport('./lockdown-more.js');
+    tryImport('./policy-load.js');
+  } else {
+    tryImport('./lockdown-install.js');
+    tryImport('./lockdown-more.js');
+    tryImport('./lockdown-run.js');
+    tryImport('./runtime-cjs.js');
+  }
```
jpuri, line 31: Above I have hard coded `applyLavaMoat` to true as else condition is breaking in `tryImport('./runtime-cjs.js');` with this error:

Screenshot 2022-05-23 at 5 21 07 PM

app/scripts/ui.js
```diff
@@ -37,6 +37,17 @@ async function start() {
...
+    extensionPort.onMessage.addListener((message) => {
+      if (
+        document.getElementById('app-loader') &&
+        message?.name === 'CONNECTION_READY'
+      ) {
+        initializeUiWithTab(activeTab);
+      }
+    });
+  }
+
```
jpuri, line 50: I have added this message above to initialise UI when backend is ready, otherwise in if UI initialises before backend is ready it results in blank screen in extension and this problem was very frequent with MV3.

app/scripts/app-init.js
kumavis: i gues xhr is not available in service workers? its web3 junk we're not using anyways. can fix with a patch.

app/scripts/app-init.js
```diff
@@ -0,0 +1,52 @@
...
+
+  const fileList = [
+    /** FILE NAMES */
```
kumavis, line 34: should add a comment indicating the fileList is injected during build time

app/scripts/ui.js
kumavis: calling `initializeUiWithTab` twice in this case, is that intentional?

development/build/scripts.js
```diff
@@ -834,6 +893,7 @@ function getEnvironmentVariables({ buildType, devMode, testing, version }) {
     ONBOARDING_V2: metamaskrc.ONBOARDING_V2 === '1',
     COLLECTIBLES_V1: metamaskrc.COLLECTIBLES_V1 === '1',
     TOKEN_DETECTION_V2: metamaskrc.TOKEN_DETECTION_V2 === '1',
+    ENABLE_MV3: metamaskrc.ENABLE_MV3 === '1',
```
kumavis, line 896: we should not required setting this value in the rc, it should all be set via the yarn script

kumavis (changes requested):

app/scripts/background.js
```diff
@@ -128,12 +148,17 @@ initialize().catch(log.error);
...
-async function initialize() {
-  const initState = await loadStateFromPersistence();
-  const initLangCode = await getFirstPreferredLangCode();
-  await setupController(initState, initLangCode);
+async function initialize(remotePort) {
+  if (!initialState) {
+    initialState = await loadStateFromPersistence();
+  }
+  if (!initialLangCode) {
+    initialLangCode = await getFirstPreferredLangCode();
```
kumavis, line 159: breakout `loadStateFromPersistence` and `getFirstPreferredLangCode` into a function and use it above
inside of `if (process.env.ENABLE_MV3) {...}`

app/manifest/mv3/_base.json
```diff
@@ -0,0 +1,81 @@
+{
```
kumavis, line 1: move manifest v2 files to `manifest/v2/`

kumavis:
looks great. add more comments in general. if you felt the need to add context here in the pull request, might as well put that context in the code

development/build/scripts.js
Gudahtt, line 896: This is checking for a value of `1`, but the yarn script is setting it to `true`

development/build/scripts.js
```diff
@@ -345,6 +346,48 @@ function createScriptTasks({
...
+    buildType,
+    destFilepath: 'app-init.js',
+    devMode,
+    entryFilepath: './app/scripts/app-init.js',
+    ignoredFiles,
+    label,
+    testing,
+    policyOnly,
+    shouldLintFenceFiles,
+  })().then(() => {
```
Gudahtt, line 380: Nit: Please use `async/await` over `.then`, because it enables async stack traces.

app/scripts/background.js
jpuri: I did some improvements here , plz check.

development/build/scripts.js
jpuri: fixed

app/scripts/background.js
```diff
@@ -291,20 +319,17 @@ function setupController(initState, initLangCode) {
     }
   }
 
+  if (process.env.ENABLE_MV3 && remoteSourcePort) {
+    connectRemote(remoteSourcePort);
+    browser.runtime.onConnect.removeListener(initApp);
+  }
```
kumavis, line 325: move below the section header
```
  //
  // connect to other contexts
  //
  ```

File: app/scripts/contentscript.js [hunk 9]
```diff
@@ -42,7 +42,12 @@ function injectScript(content) {
     const container = document.head || document.documentElement;
     const scriptTag = document.createElement('script');
     scriptTag.setAttribute('async', 'false');
-    scriptTag.textContent = content;
+    // Inline scripts do not work in MV3 due to more strict security policy
+    if (process.env.ENABLE_MV3) {
+      scriptTag.setAttribute('src', browser.runtime.getURL('inpage.js'));
```
kumavis, line 47: this will make the ethereum api injection late.
i think theres a new mv3 way of doing this, mentioned in this thread https://bugs.chromium.org/p/chromium/issues/detail?id=634381#c38
```suggestion
      // for sync injection see https://bugs.chromium.org/p/chromium/issues/detail?id=634381#c38
      scriptTag.setAttribute('src', browser.runtime.getURL('inpage.js'));
```
we should at least link to this thread in a comment
```js
if (_browser.runtime.getManifest().manifest_version == 3) {
    _browser.scripting.unregisterContentScripts().then(() => {
        var scripts = [{
             id: "inject",
             js: ["inject.js"],
             matches: ["<all_urls>"],
             world: "MAIN",
             runAt: "document_start",
         }];
        _browser.scripting.registerContentScripts(scripts);
    });
```

kumavis:
I just noticed we can do this to check if we're in version 3 if we want
```js
_browser.runtime.getManifest().manifest_version == 3
```

app/scripts/contentscript.js
jpuri: Documents say that MV3 support this, but I could not get this to work:

Screenshot 2022-05-25 at 7 52 36 AM

kumavis (approved):

app/home.html
```diff
@@ -8,7 +8,7 @@
     <link rel="stylesheet" type="text/css" href="./index-rtl.css" title="rtl" disabled>
   </head>
   <body>
-    <div id="app-content"></div>
+    <div id="app-content"><div id="app-loader">Loading...</div></div>
```
adonesky1, line 11: What is this for?

adonesky1 (approved):
Just a few non-blocking questions. Otherwise LGTM!

PR #14763, opened 2022-05-23
Remove module paths from bundle

Gudahtt:
## Explanation

A patch has been added to ensure lavapack no longer includes the path for each module as part of each serialized module. This path was originally added for debugging purposes, and is not used for anything at runtime. The module path was an absolute path, not a relative one, so it was an obstacle to having reproducible builds between environments.

## Manual Testing Steps

This should have no functional impact upon the wallet, so the QA steps would be whatever is typically done for regression testing.

For ensuring the filename is no longer present in the bundle:
* Run `yarn dist`
* Look in any of the bundles created using lavapack (e.g. the factored bundles like `background-[number].js`, `ui-[number].js`, etc), and search for the directory the project is in on your local machine to ensure it's not included in the bundle.

## Pre-Merge Checklist

- [ ] **IF** QA attention is required, "QA Board" label has been applied


brad-decker (approved):
LGTM was curious when doing the manual build process about seeing portions of my personal file structure in the build output.

darkwing (approved):

PR #14767, opened 2022-05-24
Migrate the "estimateGas" API call to "getFees" for STX

dan437:
## Explanation
The "getFees" endpoint now supports multiple transactions, so we don't need to make two API calls to "estimateGas" when an approval tx is needed together with a trade tx.

## Testing
- Open Swaps
- On the Build Quote page make sure that Smart Transactions are on
- Once you click on "Review Swap" on the Build Quote page, in the DevTools for background.html you will see a `getFees` API call. If you would choose a source token (token from) that requires an approval, the API payload will have 2 transactions. One for the approval and another one for your trade


app/scripts/metamask-controller.js
```diff
@@ -1819,9 +1819,6 @@ export default class MetamaskController extends EventEmitter {
       fetchSmartTransactionFees: smartTransactionsController.getFees.bind(
         smartTransactionsController,
       ),
-      estimateSmartTransactionsGas: smartTransactionsController.estimateGas.bind(
-        smartTransactionsController,
-      ),
```
jpuri, line 1824: 👍

jpuri (dismissed):
Changes look good to me, though I did not tested swaps functionality.

jpuri (approved):

ui/ducks/swaps/swaps.js
```diff
@@ -923,36 +922,44 @@ export const signAndSendSwapsSmartTransaction = ({
...
+      if (!fees) {
+        log.error('"fetchSwapsSmartTransactionFees" failed');
+        dispatch(setSwapsSTXSubmitLoading(false));
+        dispatch(setCurrentSmartTransactionsError(stxErrorTypes.UNAVAILABLE));
+        return;
+      }
+      if (approveTxParams) {
         updatedApproveTxParams.gas = `0x${decimalToHex(
-          smartTransactionApprovalFees?.gasLimit || 0,
+          fees.approvalTxFees?.gasLimit || 0,
```
adonesky1, line 947: could `approvalTxFees` ever be undefined here even if `fees` is defined ? Should we make this `fees?.approvalTxFees?.gasLimit`?

ui/ducks/swaps/swaps.js
```diff
@@ -923,36 +922,44 @@ export const signAndSendSwapsSmartTransaction = ({
...
           }),
         );
       }
-
-      const smartTransactionFees = await dispatch(
-        fetchSwapsSmartTransactionFees(unsignedTransaction),
-      );
       unsignedTransaction.gas = `0x${decimalToHex(
-        smartTransactionFees?.gasLimit || 0,
+        fees.tradeTxFees?.gasLimit || 0,
```
adonesky1, line 957: same here

ui/pages/swaps/view-quote/view-quote.js
```diff
@@ -391,11 +389,11 @@ export default function ViewQuote() {
   if (
     currentSmartTransactionsEnabled &&
     smartTransactionsOptInStatus &&
-    smartTransactionEstimatedGas?.txData
+    smartTransactionFees?.tradeTxFees
   ) {
     const stxEstimatedFeeInWeiDec =
-      smartTransactionEstimatedGas.txData.feeEstimate +
-      (smartTransactionEstimatedGas.approvalTxData?.feeEstimate || 0);
+      smartTransactionFees?.tradeTxFees.feeEstimate +
```
adonesky1, line 395: maybe?
```suggestion
      smartTransactionFees?.tradeTxFees?.feeEstimate +
```

ui/ducks/swaps/swaps.js
dan437: If `fees` is falsy, we exit early, so then here we don't need to do `fees?`:
```
if (!fees) {
        log.error('"fetchSwapsSmartTransactionFees" failed');
        dispatch(setSwapsSTXSubmitLoading(false));
        dispatch(setCurrentSmartTransactionsError(stxErrorTypes.UNAVAILABLE));
        return;
      }
```

ui/pages/swaps/view-quote/view-quote.test.js
```diff
@@ -87,7 +88,7 @@ describe('ViewQuote', () => {
       getByTestId('main-quote-summary__exchange-rate-container'),
     ).toMatchSnapshot();
     expect(getByText('Estimated gas fee')).toBeInTheDocument();
-    expect(getByText('0.00544 ETH')).toBeInTheDocument();
+    expect(getByText('0.00008 ETH')).toBeInTheDocument();
```
adonesky1, line 91: just curious why did this change?

adonesky1 (approved):
LGTM

ui/pages/swaps/view-quote/view-quote.test.js
dan437: I've changed mock data with `feeEstimate` as part of this PR and because of that it's showing a little different estimate.

PR #14777, opened 2022-05-25
Remove global transaction state from send flow

### LINKED ISSUE #11677: Extension sends random cached addressed instead of the one entered

brad-decker:
## Explanation

As part of exploring possible causes of #11677 and came to the conclusion that, even though I could not reproduce the bug, that it was possible for the send slice to get polluted with bad data. There were two main problems with the structure of the send slice:

1. Upon every update to the transaction state we generated the txParams and kept them in the slice as well. txParams are a derivative of the other state tree values but under certain scenarios it was plausible that the txParams in state would become stale and not be an accurate representation of the rest of the state tree. To resolve this issue I removed the derivative state from the slice in #14701 
2. The transaction state in the send slice is global to the slice, meaning that the slice is "initiated" when navigating to the send page and the send page would clear the send slice state after the send flow ended...but it was plausible that a new send flow could be kicked off in some way while there was lingering state from the previous send in the tree. This would cause an issue where the transaction would get corrupted with the previous transaction's intent. 

This PR aims to alleviate the second point.

Send flow before this change:
1. User clicks send in the extension main screen.
2. If the user was on an asset page or clicked send from a token  asset, we'd dispatch `updateSendAsset` with the token details before routing the user to the send page.
3. The send page had a `useEffect` that would dispatch `initializeSendState` this would do things like fetch gas price estimations, set the validity of the state, etc. 
4. The transaction values such as recipient, amount, from and to were stored in a global state tree.
5. Upon leaving the send flow the send page would dispatch the `resetSendState` action that would set the tree back to default.

The same basic flow would happen upon editing a transaction, except that in addition to setting some global values the editTransaction method would be dispatched that also set the details of the edited transaction into the global state.

Send flow after this change:
1. user clicks send in the extension main screen
2. we dispatch `startNewTransaction` and include asset details in that call.
3. The send slice clears all previous draft transactions from state, which are now stored in an object in state that is no longer global. 
4. The send slice generates a new UUID and sets as the `currentTransactionUUID` in state, then creates a blank transaction in the `draftTransactions` object that is keyed by UUID.
5. The `initializeSendState` thunk is dispatched.
6. Once complete the user is routed to the send page. 
7. Upon leaving the send flow the state is cleared.

if editing a transaction a new 'editExistingTransaction' method is dispatched with the asset type and id of the transaction, but the rest of the flow is the same. 

### How does this help? 
1. All transaction values are kept inside an object that is keyed by a non deterministic UUID generated at the time the draft is created.  
2. When a new draft is created, all previous drafts are discarded as an additional fail safe to make sure that the new transaction has no chance to be confused with an older, stale one. 
3. We no longer manipulate the send state ahead of kicking off the send flow which makes the send flow less reliant upon side effects and makes it idempotent, in that it is less likely to have a different outcome based on the order of operations.

### Related Changes
1. A new 'getCurrentDraftTransaction' selector was added as a tool for all other selectors. Most selectors now call this selector to get either an empty object or the current draftTransaciton.
2. A lot of selectors have been updated to reflect the possibility that a draftTransaction does not exist. 
3. For some reason in some of our confirmation components we use the send slice selectors, I did not change this but we can be reasonably assured now that they aren't grabbing unrelated send state. 
4. The send slice has been reorganized a great deal to try and keep things alphabetical where possible. It is not always possible, especially in the selectors area, due to things being used before they are declared. 
5. More types and safer types have been added.
6. A number of helper methods were extracted from the send slice file to a helpers.js file in the send slice folder. This is to make the send slice less difficult to read.

## More Information
Fixes #11677 

### Before & After
The functionality and flow of the send page is unchanged, no E2E tests needed to be modified to make this work. 

## Manual Testing Steps
1. Test sending ETH, Tokens, Collectibles.
2. Test changing accounts, networks, and assets on the send flow.
3. Test changing the send amount in the send flow.
4. Test editing transactions.
5. Ensure that every transaction you create through the send flow results in a confirmation with matching details.
7. Check ENS, copy and paste, and manual address entry in the recipient fielld.
8. Test adding nicknames to addresses and see that the nickname is reflected in the send flow. 
9. Any other details related to the send flow.

## Pre-Merge Checklist

- [x] ~**IF** this PR fixes a bug, a test that _would have_ caught the bug has been added~

### DIFF
--- a/ui/components/app/wallet-overview/token-overview.js
+++ b/ui/components/app/wallet-overview/token-overview.js
@@ -14,7 +14,7 @@ import {
 } from '../../../helpers/constants/routes';
 import { useTokenTracker } from '../../../hooks/useTokenTracker';
 import { useTokenFiatAmount } from '../../../hooks/useTokenFiatAmount';
-import { updateSendAsset } from '../../../ducks/send';
+import { startNewDraftTransaction } from '../../../ducks/send';
 import { setSwapsFromToken } from '../../../ducks/swaps/swaps';
 import {
   getCurrentKeyring,
@@ -93,7 +93,7 @@ const TokenOverview = ({ className, token }) => {
               });
               try {
                 await dispatch(
-                  updateSendAsset({
+                  startNewDraftTransaction({
                     type: ASSET_TYPES.TOKEN,
                     details: token,
                   }),
--- a/ui/ducks/send/helpers.js
+++ b/ui/ducks/send/helpers.js
@@ -1,0 +1,295 @@
+import { addHexPrefix } from 'ethereumjs-util';
+import abi from 'human-standard-token-abi';
+import { GAS_LIMITS } from '../../../shared/constants/gas';
+import { CHAIN_ID_TO_GAS_LIMIT_BUFFER_MAP } from '../../../shared/constants/network';
+import {
+  ASSET_TYPES,
+  TRANSACTION_ENVELOPE_TYPES,
+} from '../../../shared/constants/transaction';
+import { readAddressAsContract } from '../../../shared/modules/contract-utils';
+import {
+  conversionUtil,
+  multiplyCurrencies,
+} from '../../../shared/modules/conversion.utils';
+import { ETH, GWEI } from '../../helpers/constants/common';
+import { calcTokenAmount } from '../../helpers/utils/token-util';
+import { MIN_GAS_LIMIT_HEX } from '../../pages/send/send.constants';
+import {
+  addGasBuffer,
+  generateERC20TransferData,
+  generateERC721TransferData,
+  getAssetTransferData,
+} from '../../pages/send/send.utils';
+import { getGasPriceInHexWei } from '../../selectors';
+import { estimateGas } from '../../store/actions';
+
+export async function estimateGasLimitForSend({
+  selectedAddress,
+  value,
+  gasPrice,
+  sendToken,
+  to,
+  data,
+  isNonStandardEthChain,
+  chainId,
+  gasLimit,
+  ...options
+}) {
+  let isSimpleSendOnNonStandardNetwork = false;
+
+  // blockGasLimit may be a falsy, but defined, value when we receive it from
+  // state, so we use logical or to fall back to MIN_GAS_LIMIT_HEX. Some
+  // network implementations check the gas parameter supplied to
+  // eth_estimateGas for validity. For this reason, we set token sends
+  // blockGasLimit default to a higher number. Note that the current gasLimit
+  // on a BLOCK is 15,000,000 and will be 30,000,000 on mainnet after London.
+  // Meanwhile, MIN_GAS_LIMIT_HEX is 0x5208.
+  let blockGasLimit = MIN_GAS_LIMIT_HEX;
+  if (options.blockGasLimit) {
+    blockGasLimit = options.blockGasLimit;
+  } else if (sendToken) {
+    blockGasLimit = GAS_LIMITS.BASE_TOKEN_ESTIMATE;
+  }
+
+  // The parameters below will be sent to our background process to estimate
+  // how much gas will be used for a transaction. That background process is
+  // located in tx-gas-utils.js in the transaction controller folder.
+  const paramsForGasEstimate = { from: selectedAddress, value, gasPrice };
+
+  if (sendToken) {
+    if (!to) {
+      // If no to address is provided, we cannot generate the token transfer
+      // hexData. hexData in a transaction largely dictates how much gas will
+      // be consumed by a transaction. We must use our best guess, which is
+      // represented in the gas shared constants.
+      return GAS_LIMITS.BASE_TOKEN_ESTIMATE;
+    }
+    paramsForGasEstimate.value = '0x0';
+
+    // We have to generate the erc20/erc721 contract call to transfer tokens in
+    // order to get a proper estimate for gasLimit.
+    paramsForGasEstimate.data = getAssetTransferData({
+      sendToken,
+      fromAddress: selectedAddress,
+      toAddress: to,
+      amount: value,
+    });
+
+    paramsForGasEstimate.to = sendToken.address;
+  } else {
+    if (!data) {
+      // eth.getCode will return the compiled smart contract code at the
+      // address. If this returns 0x, 0x0 or a nullish value then the address
+      // is an externally owned account (NOT a contract account). For these
+      // types of transactions the gasLimit will always be 21,000 or 0x5208
+      const { isContractAddress } = to
+        ? await readAddressAsContract(global.eth, to)
+        : {};
+      if (!isContractAddress && !isNonStandardEthChain) {
+        return GAS_LIMITS.SIMPLE;
+      } else if (!isContractAddress && isNonStandardEthChain) {
+        isSimpleSendOnNonStandardNetwork = true;
+      }
+    }
+
+    paramsForGasEstimate.data = data;
+
+    if (to) {
+      paramsForGasEstimate.to = to;
+    }
+
+    if (!value || value === '0') {
+      // TODO: Figure out what's going on here. According to eth_estimateGas
+      // docs this value can be zero, or undefined, yet we are setting it to a
+      // value here when the value is undefined or zero. For more context:
+      // https://github.com/MetaMask/metamask-extension/pull/6195
+      paramsForGasEstimate.value = '0xff';
+    }
+  }
+
+  if (!isSimpleSendOnNonStandardNetwork) {
+    // If we do not yet have a gasLimit, we must call into our background
+    // process to get an estimate for gasLimit based on known parameters.
+
+    paramsForGasEstimate.gas = addHexPrefix(
+      multiplyCurrencies(blockGasLimit, 0.95, {
+        multiplicandBase: 16,
+        multiplierBase: 10,
+        roundDown: '0',
+        toNumericBase: 'hex',
+      }),
+    );
+  }
+
+  // The buffer multipler reduces transaction failures by ensuring that the
+  // estimated gas is always sufficient. Without the multiplier, estimates
+  // for contract interactions can become inaccurate over time. This is because
+  // gas estimation is non-deterministic. The gas required for the exact same
+  // transaction call can change based on state of a contract or changes in the
+  // contracts environment (blockchain data or contracts it interacts with).
+  // Applying the 1.5 buffer has proven to be a useful guard against this non-
+  // deterministic behaviour.
+  //
+  // Gas estimation of simple sends should, however, be deterministic. As such
+  // no buffer is needed in those cases.
+  let bufferMultiplier = 1.5;
+  if (isSimpleSendOnNonStandardNetwork) {
+    bufferMultiplier = 1;
+  } else if (CHAIN_ID_TO_GAS_LIMIT_BUFFER_MAP[chainId]) {
+    bufferMultiplier = CHAIN_ID_TO_GAS_LIMIT_BUFFER_MAP[chainId];
+  }
+
+  try {
+    // Call into the background process that will simulate transaction
+    // execution on the node and return an estimate of gasLimit
+    const estimatedGasLimit = await estimateGas(paramsForGasEstimate);
+    const estimateWithBuffer = addGasBuffer(
+      estimatedGasLimit,
+      blockGasLimit,
+      bufferMultiplier,
+    );
+    return addHexPrefix(estimateWithBuffer);
+  } catch (error) {
+    const simulationFailed =
+      error.message.includes('Transaction execution error.') ||
+      error.message.includes(
+        'gas required exceeds allowance or always failing transaction',
+      ) ||
+      (CHAIN_ID_TO_GAS_LIMIT_BUFFER_MAP[chainId] &&
+        error.message.includes('gas required exceeds allowance'));
+    if (simulationFailed) {
+      const estimateWithBuffer = addGasBuffer(
+        paramsForGasEstimate?.gas ?? gasLimit,
+        blockGasLimit,
+        bufferMultiplier,
+      );
+      return addHexPrefix(estimateWithBuffer);
+    }
+    throw error;
+  }
+}
+
+/**
+ * Generates a txParams from the send slice.
+ *
+ * @param {import('.').SendState} sendState - the state of the send slice
+ * @returns {import(
+ *  '../../../shared/constants/transaction'
+ * ).TxParams} A txParams object that can be used to create a transaction or
+ *  update an existing transaction.
+ */
+export function generateTransactionParams(sendState) {
+  const draftTransaction =
+    sendState.draftTransactions[sendState.currentTransactionUUID];
+  const txParams = {
+    // If the fromAccount has been specified we use that, if not we use the
+    // selected account.
+    from:
+      draftTransaction.fromAccount?.address ||
+      sendState.selectedAccount.address,
+    // gasLimit always needs to be set regardless of the asset being sent
+    // or the type of transaction.
+    gas: draftTransaction.gas.gasLimit,
+  };
+  switch (draftTransaction.asset.type) {
+    case ASSET_TYPES.TOKEN:
+      // When sending a token the to address is the contract address of
+      // the token being sent. The value is set to '0x0' and the data
+      // is generated from the recipient address, token being sent and
+      // amount.
+      txParams.to = draftTransaction.asset.details.address;
+      txParams.value = '0x0';
+      txParams.data = generateERC20TransferData({
+        toAddress: draftTransaction.recipient.address,
+        amount: draftTransaction.amount.value,
+        sendToken: draftTransaction.asset.details,
+      });
+      break;
+    case ASSET_TYPES.COLLECTIBLE:
+      // When sending a token the to address is the contract address of
+      // the token being sent. The value is set to '0x0' and the data
+      // is generated from the recipient address, token being sent and
+      // amount.
+      txParams.to = draftTransaction.asset.details.address;
+      txParams.value = '0x0';
+      txParams.data = generateERC721TransferData({
+        toAddress: draftTransaction.recipient.address,
+        fromAddress:
+          draftTransaction.fromAccount?.address ??
+          sendState.selectedAccount.address,
+        tokenId: draftTransaction.asset.details.tokenId,
+      });
+      break;
+    case ASSET_TYPES.NATIVE:
+    default:
+      // When sending native currency the to and value fields use the
+      // recipient and amount values and the data key is either null or
+      // populated with the user input provided in hex field.
+      txParams.to = draftTransaction.recipient.address;
+      txParams.value = draftTransaction.amount.value;
+      txParams.data = draftTransaction.userInputHexData ?? undefined;
+  }
+
+  // We need to make sure that we only include the right gas fee fields
+  // based on the type of transaction the network supports. We will also set
+  // the type param here.
+  if (sendState.eip1559support) {
+    txParams.type = TRANSACTION_ENVELOPE_TYPES.FEE_MARKET;
+
+    txParams.maxFeePerGas = draftTransaction.gas.maxFeePerGas;
+    txParams.maxPriorityFeePerGas = draftTransaction.gas.maxPriorityFeePerGas;
+
+    if (!txParams.maxFeePerGas || txParams.maxFeePerGas === '0x0') {
+      txParams.maxFeePerGas = draftTransaction.gas.gasPrice;
+    }
+
+    if (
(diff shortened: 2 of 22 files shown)


ui/ducks/send/send.js
```diff
@@ -280,164 +281,181 @@ export const RECIPIENT_SEARCH_MODES = {
...
-  }
+/**
+ * @type {DraftTransaction}
+ */
+export const draftTransactionInitialState = {
+  id: null,
+  status: SEND_STATUSES.VALID,
+  transactionType: TRANSACTION_ENVELOPE_TYPES.LEGACY,
+  userInputHexData: null,
+  editAccount: null,
```
brad-decker, line 362: [self review] remove editAccount in favor of 'fromAccount'

ui/ducks/send/send.js
```diff
@@ -513,27 +533,145 @@ export const computeEstimatedGasLimit = createAsyncThunk(
...
-    numberOfDecimals: 9,
-    toDenomination: GWEI,
-    fromNumericBase: 'dec',
-    toNumericBase: 'dec',
-    fromCurrency: ETH,
-    fromDenomination: GWEI,
-  });
-  const gasPriceAsNumber = Number(gasPriceInDecGwei);
-  return getGasPriceInHexWei(gasPriceAsNumber);
+function getAssetDetailsAndBalance(
```
brad-decker, line 565: [self review] I would love for this function to be non side effect producing (currently it opens modals, shows loading indication, etc)... The only way I thought to do that was to interpret the error message in the consumer... but then not only are we using errors for control flow, we're also requiring duplication of that flow in multiple places (every callsite). @adonesky1 if you get a minute could you tell me if the logic in here matches the old upsetSendAsset flow?  You did most of the work on collectibles which is where i'm the least confident.

ui/ducks/send/send.js
```diff
@@ -549,27 +687,23 @@ function getRoundedGasPrice(gasPriceEstimate) {
...
   async (_, thunkApi) => {
+    /**
+     * @typedef {Object} ReduxState
+     * @property {Object} metamask - Half baked type for the MetaMask object
+     * @property {SendState} send - the send state
+     */
+
+    /**
+     * @type {ReduxState}
+     */
```
brad-decker, line 698: This can be typed using the AsyncThunk type from redux toolkit

joani24 (dismissed):
Mto bom

lavamoat/build-system/policy-override.json
```diff
@@ -55,6 +55,793 @@
         "eslint-plugin-jest>@typescript-eslint/experimental-utils>eslint-utils>eslint-visitor-keys": true
       }
     },
+    "gulp-watch>chokidar": {
```
brad-decker, line 58: this stuff was needed for my local e2e, but not necessary here. I'll remove this file.

test/e2e/helpers.js
```diff
@@ -110,7 +110,10 @@ async function withFixtures(options, testSuite) {
           '\n',
         )}`;
         if (failOnConsoleError) {
-          throw new Error(errorMessage);
+          errors.forEach((err) => {
```
brad-decker, line 113: this change was not helpful for debugging e2e console errors so i'll remove

ui/pages/send/send.js
```diff
@@ -49,11 +47,8 @@ export default function SendTransactionScreen() {
   }, [dispatch]);
 
   useEffect(() => {
-    if (chainId !== undefined) {
-      dispatch(initializeSendState());
```
brad-decker, line 53: This is now handled in the actions file instead of here.

ui/ducks/send/helpers.js
```diff
@@ -0,0 +1,201 @@
...
+ * after receiving a gasPrice from eth_gasPrice. First, the returned gasPrice
+ * was converted to GWEI, then it was converted to a Number, then in the send
+ * duck (here) we would use getGasPriceInHexWei to get back to hexWei. Now that
+ * we receive a GWEI estimate from the controller, we still need to do this
+ * weird conversion to get the proper rounding.
+ *
+ * @param {string} gasPriceEstimate
+ * @returns {string}
+ */
+export function getRoundedGasPrice(gasPriceEstimate) {
```
brad-decker, line 31: This file contains helpers that were previously in the send duck file, these do not directly modify state or interact with redux.

ui/ducks/send/send.js
brad-decker: still true, Brad.

ui/ducks/send/send.js
```diff
@@ -587,43 +615,49 @@ export const initializeSendState = createAsyncThunk(
...
-      gasPrice = getRoundedGasPrice(gasFeeEstimates.gasPrice);
-    } else if (gasEstimateType === GAS_ESTIMATE_TYPES.FEE_MARKET) {
-      gasPrice = getGasPriceInHexWei(
-        gasFeeEstimates.medium.suggestedMaxFeePerGas,
-      );
-    } else {
-      gasPrice = gasFeeEstimates.gasPrice
-        ? getRoundedGasPrice(gasFeeEstimates.gasPrice)
-        : '0x0';
+    if (sendState.stage !== SEND_STAGES.EDIT) {
```
brad-decker, line 618: This logic prevents the gasPrice ans gasLimit from an edit transaction from being ovewritten immediately on initialization.

ui/ducks/send/send.js
```diff
@@ -587,43 +615,49 @@ export const initializeSendState = createAsyncThunk(
...
+          : '0x0';
+      }
     }
 
     // Set a basic gasLimit in the event that other estimation fails
-    let gasLimit =
-      asset.type === ASSET_TYPES.TOKEN || asset.type === ASSET_TYPES.COLLECTIBLE
-        ? GAS_LIMITS.BASE_TOKEN_ESTIMATE
-        : GAS_LIMITS.SIMPLE;
+    let { gasLimit } = draftTransaction.gas;
```
brad-decker, line 638: default to the gasLimit set on the draftTransaction, which is either '0x0' or an actual limit from an edit tx

ui/ducks/send/send.js
```diff
@@ -632,36 +666,8 @@ export const initializeSendState = createAsyncThunk(
...
-    if (asset.type === ASSET_TYPES.COLLECTIBLE) {
-      if (asset.details === null) {
-        // If we're sending a collectible but details have not been provided we must
-        // abort and set the send slice into invalid status.
-        throw new Error(
-          'Send slice initialized as collectibles send without token details',
-        );
-      }
-      balance = '0x1';
-    }
```
brad-decker, line 660: this portion is now handled in the updateSendAsset, which now is never called from outside the send flow (it was previously called before routing to the send page to set the asset to be sent)

ui/ducks/send/helpers.js
```diff
@@ -0,0 +1,295 @@
...
+  }
+
+  // The parameters below will be sent to our background process to estimate
+  // how much gas will be used for a transaction. That background process is
+  // located in tx-gas-utils.js in the transaction controller folder.
+  const paramsForGasEstimate = { from: selectedAddress, value, gasPrice };
+
+  if (sendToken) {
+    if (!to) {
+      // if no to address is provided, we cannot generate the token transfer
```
darkwing, line 61:
```suggestion
      // If no to address is provided, we cannot generate the token transfer
```

ui/ducks/send/helpers.js
```diff
@@ -0,0 +1,295 @@
...
+  } else {
+    if (!data) {
+      // eth.getCode will return the compiled smart contract code at the
+      // address. If this returns 0x, 0x0 or a nullish value then the address
+      // is an externally owned account (NOT a contract account). For these
+      // types of transactions the gasLimit will always be 21,000 or 0x5208
+      const { isContractAddress } = to
+        ? await readAddressAsContract(global.eth, to)
+        : {};
+      if (!isContractAddress && !isNonStandardEthChain) {
```
darkwing, line 88: Is something like this better?

```suggestion
      if (!isContractAddress) {
        if(isNonStandardEthChain) {
          isSimpleSendOnNonStandardNetwork = true;
        }
        else {
          return GAS_LIMITS.SIMPLE;
        }
      }
```

ui/ducks/send/helpers.test.js
```diff
@@ -0,0 +1,160 @@
...
+          sendToken: tokenDetails,
+        }),
+        to: '0xToken',
+        type: '0x0',
+        value: '0x0',
+        gas: '0x0',
+        gasPrice: '0x0',
+      });
+    });
+    it('should generate a txParams for a collectible transfer', () => {
```
darkwing, line 56: Many of our test files have empty lines before each `it` test.

ui/pages/confirm-send-token/confirm-send-token.container.js
```diff
@@ -18,18 +18,11 @@ const mapStateToProps = (state) => {
...
-        editTransaction(
-          ASSET_TYPES.TOKEN,
-          id.toString(),
-          tokenData,
-          assetDetails,
-        ),
-      );
-      dispatch(clearConfirmTransaction());
-      dispatch(showSendTokenPage());
+      await dispatch(editExistingTransaction(ASSET_TYPES.TOKEN, id.toString()));
```
darkwing, line 23: Out of curiosity, I know this wasn't changed, but what could `id` be that we need to typecast it to string?

darkwing (changes requested):

ui/ducks/send/helpers.js
brad-decker: I like what you have, but i'm hesitant to change it as this function simply moved unchanged from its previous location.

ui/pages/confirm-send-token/confirm-send-token.container.js
brad-decker: I honestly have no clue... it should be a string and is a string in the state tree. i'm not sure if there is some higher level area where it is manipulated away from being a JavaScript string object. The .toString() was copied from the confirmation base when the id is manipulated.

tmashuang:
One minor thing I found was that no matter what I couldn't get the `send.recipientMode` to change from `CONTACT_LIST` to `MY_ACCOUNTS`. I don't know if that matters somewhere, but I was able to send the txs with no issues.

lavamoat/build-system/policy-override.json
Nunnaphat1234: Hello, brothers and sisters. From the bottom of my heart. I haven't replied to many things. Yes, what should I do, the brothers who have been helping me all this time come to my account and replace them? I give you the right to do every action with autonomy.

Nunnaphat1234:
1/ No, I only have one wallet.0xC8eac206a4e511BA244531C01C91Df5B1BD967e1

ui/ducks/send/helpers.js
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    paramsForGasEstimate.data = data;
+
+    if (to) {
+      paramsForGasEstimate.to = to;
+    }
+
+    if (!value || value === '0') {
+      // TODO: Figure out what's going on here. According to eth_estimateGas
```
darkwing, line 102: Is this ToDo blocking?

brad-decker: Nah, this function just got relocated. It has been a To do since my last refacotr. 😭

tmashuang (dismissed):

ui/ducks/send/helpers.js
```diff
@@ -0,0 +1,295 @@
...
+      // When sending a token the to address is the contract address of
+      // the token being sent. The value is set to '0x0' and the data
+      // is generated from the recipient address, token being sent and
+      // amount.
+      txParams.to = draftTransaction.asset.details.address;
+      txParams.value = '0x0';
+      txParams.data = generateERC721TransferData({
+        toAddress: draftTransaction.recipient.address,
+        fromAddress:
+          draftTransaction.fromAccount.address ??
```
adonesky1, line 218:
```suggestion
          draftTransaction.fromAccount?.address ??
```

adonesky1:
I'm not all the way through, but I want to note this thought: I'm not entirely clear on why `draftTransactions` should be an object keyed by a `UUID`. It should only ever contain one transaction at a time right? Couldn't it just be 

```
draftTransaction: {
  amount: {
    error: null,
    value: '0x0',
  },
  asset: {
    balance: '0x0',
    details: null,
    error: null,
    type: ASSET_TYPES.NATIVE,
  },
  fromAccount: null,
  gas: {
    error: null,
    gasLimit: '0x0',
    gasPrice: '0x0',
    gasTotal: '0x0',
    maxFeePerGas: '0x0',
    maxPriorityFeePerGas: '0x0',
  },
  history: [],
  id: null,
  recipient: {
    address: '',
    error: null,
    nickname: '',
    warning: null,
  },
  status: SEND_STATUSES.VALID,
  transactionType: TRANSACTION_ENVELOPE_TYPES.LEGACY,
  userInputHexData: null,
};
``` ?

**Comment by adonesky1** [comment]
Seeing a pretty funny bug:

So if you try to change the network in the middle of a token send you get an error in the `tokenInput` component because it's trying to find token data in the `tokenExchangeRates` object. I've created [an issue for this bug](https://github.com/MetaMask/metamask-extension/issues/15036)(which exists on develop too). 

But so if you do a naive fix for that bug: 
`const tokenExchangeRate = tokenExchangeRates?.[existingToken.address] || 0;`
 -> 
 `const tokenExchangeRate = tokenExchangeRates?.[existingToken**?**.address] || 0;`
 
 you get this funny scenario:

https://user-images.githubusercontent.com/34557516/175351058-92d96467-4e7a-44f7-8c00-c0a5ba941846.mov

where the send state asset is not cleared and instead we just get a token that is associated with the same address on the new network. I'm guessing this is not the behavior we want...

**Comment by brad-decker** [comment]
> I'm not all the way through, but I want to note this thought: I'm not entirely clear on why draftTransactions should be an object keyed by a UUID. It should only ever contain one transaction at a time right? Couldn't it just be?:

This same assumption occurs in the current send state, that there will only ever be one "in flight" transaction in the send flow at a time, we do clear this state when the user closes the window, etc but still we saw possibility of being corrupted. The idea of the UUID here is to generate it during the send flow being triggered, such that anytime the send flow is entered a new UUID is created and thus we will never accidentally reference the state of a different transaction as the previous UUID is replaced and can no longer be recalled.

So to answer your question -- it is simply another layer of safety in case the clearing of a draft out of the object is partial, or doesn't happen, as we expect it to. I realize that it might not be immediately clear how that could happen, but that sort of quality applies to the current send state as well.

**Comment by brad-decker** [comment]
> But so if you do a naive fix for that bug:
const tokenExchangeRate = tokenExchangeRates?.[existingToken.address] || 0;
->
const tokenExchangeRate = tokenExchangeRates?.[existingToken**?**.address] || 0;
you get this funny scenario:

I'm unclear on this feedback if it needs to be addressed here ... if the bug you are outlining exists in this PR's current state?

**Comment by adonesky1** [comment]
> I'm unclear on this feedback if it needs to be addressed here ... if the bug you are outlining exists in this PR's current state?

I think the critical bug - which is relevant to the implementation here - is that when a user changes networks during the send flow we don't appear to be clearing/reset the `asset` value in the `draftTransaction` so you end up with strange situations (like the video linked above) where, after changing networks, `asset` is pointing to a different token (while the name in the input remains unchanged).

**Comment by brad-decker** [comment]
We should set the asset back to the native asset on a network change. I'll make that change.

File: ui/ducks/send/send.js [hunk 17]
```diff
@@ -278,168 +259,216 @@ export const AMOUNT_MODES = {
...
-  // deterministic behaviour.
-  //
-  // Gas estimation of simple sends should, however, be deterministic. As such
-  // no buffer is needed in those cases.
-  let bufferMultiplier = 1.5;
-  if (isSimpleSendOnNonStandardNetwork) {
-    bufferMultiplier = 1;
-  } else if (CHAIN_ID_TO_GAS_LIMIT_BUFFER_MAP[chainId]) {
-    bufferMultiplier = CHAIN_ID_TO_GAS_LIMIT_BUFFER_MAP[chainId];
-  }
```
adonesky1, line 399: Did you intend to fully remove buffering for the gas limit? If so why?

adonesky1: Oh nvm I see this whole method was moved to helpers 🤦

adonesky1 (commented):
Amazing (and ambitious) work as always Brad! I've left comments on everything that caught my eye after going through the whole thing 😌

ui/pages/confirm-send-token/confirm-send-token.component.js
```diff
@@ -12,8 +12,9 @@ export default class ConfirmSendToken extends Component {
 
   handleEdit(confirmTransactionData) {
     const { editTransaction, history } = this.props;
-    editTransaction(confirmTransactionData);
-    history.push(SEND_ROUTE);
+    editTransaction(confirmTransactionData).then(() => {
```
danjm, line 15: In `ui/pages/confirm-send-ether/confirm-send-ether.container.js` we made a change to use `editExistingTransaction` instead of `editTransaction`. but not here

was that intentional?

ui/components/app/wallet-overview/token-overview.js
```diff
@@ -93,7 +93,7 @@ const TokenOverview = ({ className, token }) => {
               });
               try {
                 await dispatch(
-                  updateSendAsset({
+                  startNewDraftTransaction({
```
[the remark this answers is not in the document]
brad-decker: we are `await`'ing it here, so we do wait.

ui/pages/confirm-send-token/confirm-send-token.component.js
brad-decker: the container wraps editExistingTransaction and calls it 'editTransaction' i'm changing this to avoid confusion and making debugging better.

ui/components/app/wallet-overview/token-overview.js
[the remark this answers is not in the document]
danjm: oh yes

PeterYinusa:
Hey @brad-decker, I'm seeing some weird behaviour with the `max amount` button in the send flow for ERC20 tokens, it seems to work as expected for the native currency. See below:

https://user-images.githubusercontent.com/53189696/175666453-b5f9a294-3d72-4a14-a92b-5bf8264b4e50.mov

brad-decker:
@PeterYinusa omg good catch, sir! I changed how we got the asset balance to avoid multiple calls to the client but the result wasn't formatted the same way. This should be functionally equivalent now. Please test with several ERC20's

adonesky1 (dismissed):
🥳 Great work Brad!

tmashuang (dismissed):

seaona:
Just wondering if these changes facilitate the fix for the "`Undefined`" value when editing a Send tx, or it's totally unrelated topic /  must be fixed separately. @brad-decker this bug appears both in this branch and in develop/master, but since it's related to the Send flow, thought I might share here just in case it's related. See [more](https://github.com/MetaMask/metamask-extension/issues/15047).

image

brad-decker:
> Just wondering if these changes facilitate the fix for the "Undefined" value when editing a Send tx, or it's totally unrelated topic / must be fixed separately. @brad-decker this bug appears both in this branch and in develop/master, but since it's related to the Send flow, thought I might share here just in case it's related. See https://github.com/MetaMask/metamask-extension/issues/15047.

@seaona if they appear in both places lets keep it as a new bug to deal with after this lands just to avoid holding this PR up for longer.

darkwing (dismissed):

PeterYinusa:
> @PeterYinusa omg good catch, sir! I changed how we got the asset balance to avoid multiple calls to the client but the result wasn't formatted the same way. This should be functionally equivalent now. Please test with several ERC20's

Looking good 🙌

tmashuang (approved):

darkwing (approved):

PR #14781, opened 2022-05-25
MV3: Re-activate service worker and reconnect UI streams

### LINKED ISSUE #14559: Reconnect Streams when service worker is re-loaded

jpuri:
Fixes: https://github.com/MetaMask/metamask-extension/issues/14559

1. reactivate in-active service worker
2. re-connect UI stream on full-page implementation of extension (this still needs to be fixed for extension popup)

How to test:
1.  Use MV3 via `yarn start:mv3`
2. Open home screen wait a five minutes, where service worker should disconnect, and keep using the extension -- it should work
3. Open DevTools in the UI 
4. Choose Application -> Service Workers and kill the service worker
5. Notice that the service worker restarts


app/scripts/app-init.js
```diff
@@ -12,9 +12,11 @@ function tryImport(...fileNames) {
 
 function importAllScripts() {
   const startImportScriptsTime = Date.now();
-  // applyLavaMoat has been hard coded to "true" as
-  // tryImport('./runtime-cjs.js') is giving issue with XMLHttpRequest object which is not avaialble to service worker.
-  // we need to dynamically inject values of applyLavaMoat once this is fixed.
+  /*
+   * applyLavaMoat has been hard coded to "true" as
+   * tryImport('./runtime-cjs.js') is giving issue with XMLHttpRequest object which is not avaialble to service worker.
```
kumavis, line 17: issue is not actually in runtime-cjs but in web3

app/scripts/app-init.js
```diff
@@ -46,8 +50,26 @@ function importAllScripts() {
   );
 }
 
-// Placing script import call here ensures that scripts are inported each time service worker is activated.
-importAllScripts();
+// eslint-disable-next-line
+self.oninstall = () => {
```
kumavis, line 54: is there an alternative, like an event listener? who calls this?

jpuri: I took it from this example: https://stackoverflow.com/a/66408379/3696652

It is a service worker lifecycle callback, called once for MV3 service worker.

jpuri: As I am looking into fixing this issue, I found that we are on pretty old version of Web3.js `0.20.0`. Latest version of Web3.js already has this issue fixed. Should I just upgrade Web3 rather than patching it. It will be major version upgrade.

app/scripts/app-init.js
jpuri: As I am looking into fixing this issue, I found that we are on pretty old version of Web3.js 0.20.0. Latest version of Web3.js already has this issue fixed. Should I just upgrade Web3 rather than patching it. It will be major version upgrade.

jpuri: I am working on fixing Web3 using patch.

app/scripts/app-init.js
```diff
@@ -30,6 +32,8 @@ function importAllScripts() {
     tryImport('./lockdown-run.js');
     tryImport('./runtime-cjs.js');
   }
+  // eslint-disable-next-line
+  self.scriptLoaded = true;
```
adonesky1, line 36: hmmm I didn't know you could write to an arbitrary property to `self` 🤔

app/scripts/app-init.js
```diff
@@ -46,8 +50,26 @@ function importAllScripts() {
...
+ * but there is issue in importing webextension-polyfill into service worker.
+ * chrome does seems to work in at-least all chromium based browsers
+ */
+// eslint-disable-next-line
+chrome.runtime.onMessage.addListener((_1, _2, sendResponse) => {
+  // eslint-disable-next-line
+  if (!self.scriptLoaded) {
+    importAllScripts();
+  }
+  // Response below if not required but there is an error if message listener does not send response.
```
adonesky1, line 70: could you explain this a bit? It's not required but there's a non blocking error logged to the console if we don't send a response?

app/scripts/app-init.js
adonesky1: This is called the first time the service worker is setup, but not after revived correct? Never understood why that was the case...

ui/index.js
```diff
@@ -30,16 +30,38 @@ import txHelper from './helpers/utils/tx-helper';
...
+          data,
+        )}`,
+      );
+    }
+  });
+};
+
 export default function launchMetamaskUi(opts, cb) {
   const { backgroundConnection } = opts;
-  actions._setBackgroundConnection(backgroundConnection);
```
adonesky1, line 35: We no longer have to call this here?

app/scripts/ui.js
```diff
@@ -101,7 +156,7 @@ async function queryCurrentActiveTab(windowType) {
   });
 }
 
-function initializeUi(activeTab, container, connectionStream, cb) {
+function initializeUi(activeTab, connectionStream, cb) {
```
adonesky1, line 159: looks like `container` is still used on line 169?

app/scripts/app-init.js
jpuri: I found, it does works :)

ui/index.js
jpuri: This is now done [here](https://github.com/MetaMask/metamask-extension/pull/14781/files#diff-bb36c8e9593a9516a59e8710c563c14ae46a1a58515a5b7859233da7a0868ba3R165).

app/scripts/ui.js
jpuri: `container` is now moved [here](https://github.com/MetaMask/metamask-extension/pull/14781/files#diff-5a9d964db9d7c2fe207e21470884b2710e54723e0c1e6e6f7fd264706604fd6cR27).

app/scripts/app-init.js
jpuri: I removed that line plz ignore.

ui/index.js
adonesky1: Could you explain a bit what moving this accomplished?

jpuri: It is just code cleanup, as this functionality is now invoked from 2 places.

app/scripts/ui.js
```diff
@@ -60,23 +60,50 @@ async function start() {
...
+    };
+
+    // disconnectListener takes care to remove listeners from closed streams
+    // it also creates new streams and attach event listeners to them
+    const disconnectListener = () => {
+      extensionPort.onMessage.removeListener(messageListener);
+      extensionPort.onDisconnect.removeListener(disconnectListener);
+      // message below will try to activate service worker
+      // in MV3 is likely that reason of stream closing is service worker going in-active
+      browser.runtime.sendMessage({ name: 'UI_OPEN' });
```
adonesky1, line 85: is this message specifically listened for anywhere?

adonesky1 (commented):
A few small questions, but generally looks good!

ui/index.js
jpuri: As I tested it seems to work well.

app/scripts/ui.js
jpuri: This is captured by service worker and is used to activate it in case it becomes in-active.

adonesky1: [Here](https://github.com/MetaMask/metamask-extension/pull/14781/files#diff-cd37b653f9862ccada12e440c6ef67e8b2b2c5b906a320870ed71232d244e660R63)? Should we specifically listen for this message?

ui/index.js
adonesky1: Were you able to test hitting [this if block](https://github.com/MetaMask/metamask-extension/blob/051254dffdcaacbb8cb7639f5bcd537a85cc255a/ui/index.js#L107)?

kumavis (changes requested):
- add a big comment at the top saying this is used only by the mv3 service worker to initialize the app

app/scripts/app-init.js
```diff
@@ -29,6 +29,8 @@ function importAllScripts() {
     tryImport('./lockdown-more.js');
     tryImport('./runtime-cjs.js');
   }
+  // eslint-disable-next-line
+  self.scriptLoaded = true;
```
kumavis, line 33: make variable

app/scripts/app-init.js
kumavis: prefer add listener is available

digiwand (commented):
It's working great on my end when I run this without my cold wallet addresses, using Chrome browser and macOS

app/scripts/app-init.js
jpuri: Fixed now

app/scripts/app-init.js
```diff
@@ -6,37 +12,49 @@ function tryImport(...fileNames) {
...
+    );
   }
 
-  const fileList = [
-    // The list of files is injected at build time by replacing comment below with comma separated strings of file names
-    // https://github.com/MetaMask/metamask-extension/blob/496d9d81c3367931031edc11402552690c771acf/development/build/scripts.js#L406
-    /** FILE NAMES */
-  ];
+  // Mark scripts as loaded
+  scriptsLoaded = true;
```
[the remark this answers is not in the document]
jpuri: This should be set as soon as import start to avoid calling method again.

app/scripts/ui.js
jpuri: Hey @adonesky1 : sorry for delay, it seems link above does not works. Can you plz update.

app/scripts/ui.js
```diff
@@ -24,34 +24,37 @@ import { setupMultiplex } from './lib/stream-utils';
...
 import metaRPCClientFactory from './lib/metaRPCClientFactory';
 
-start().catch(log.error);
-
-async function start() {
-  async function displayCriticalError(container, err, metamaskState) {
-    const html = await getErrorHtml(SUPPORT_LINK, metamaskState);
-
-    container.innerHTML = html;
+const container = document.getElementById('app-content');
```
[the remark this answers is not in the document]
jpuri: `ui.js` in imported by `home.html` and it runs as the html is loaded.

Also this guarantees that `#app-content` is always present when the file is loaded.

ui/index.js
jpuri: yep I did and it seems to work well

app/scripts/ui.js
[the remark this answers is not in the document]
digiwand: ooh I see now - through `jsBundles`. Thank you for the explanation! 🙏

digiwand (dismissed):

app/scripts/app-init.js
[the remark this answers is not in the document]
naugtur: In that case it could be called `scriptsLoadPending` and be set to true as initial value.

app/scripts/app-init.js
```diff
@@ -46,5 +64,14 @@ function importAllScripts() {
...
+self.addEventListener('install', importAllScripts);
+
+/*
+ * Message event listener below loads script if they are no longer available.
+ * chrome below needs to be replaced by cross-browser object,
+ * but there is issue in importing webextension-polyfill into service worker.
+ * chrome does seems to work in at-least all chromium based browsers
+ */
+// eslint-disable-next-line no-undef
+chrome.runtime.onMessage.addListener(importAllScripts);
```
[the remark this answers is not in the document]
jpuri: Hey @naugtur : this message listener is added to revive service worker as it becomes inactive. As service worker becomes inactive the background files are no longer available and we need to reload them.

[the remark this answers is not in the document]
naugtur: If scriptsLoaded is true, it doesn't do anything. So why wait to load the scripts if they can be installed once, synchronously? What do we gain from doing it either on install or on message instead of always and right away?

Does it have anything to do with google folks suggesting we move importScripts to 'install' event?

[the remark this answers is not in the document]
digiwand: I was wondering the same thing as you @naugtur. It appears that when I manually terminate the service worker, `scriptsLoadedInitiated` gets reset. This implies to me that `app-init.js` is torn down and reloaded. The app continues to run throughout this process. 

However, after initially importing the scripts and then terminating the service worker, I'm running into this error (see the second error). I believe this error is related to concerns in this thread:
Screenshot 2022-07-23 at 2 49 56 AM

app/scripts/ui.js
```diff
@@ -60,23 +63,52 @@ async function start() {
    * Code below ensures that UI is rendered only after background is ready.
    */
   if (isManifestV3()) {
-    extensionPort.onMessage.addListener((message) => {
+    /*
+     * In case of MV3 the issue of blank screen was very frequent, it is caused by UI initialising before background is ready to send state.
+     * Code below ensures that UI is rendered only after CONNECTION_READY message is received thus background is ready.
+     * In case UI is already renderd only stream will be updated.
```
digiwand, line 69:
```suggestion
     * In case the UI is already rendered, only update the streams.
```

app/scripts/app-init.js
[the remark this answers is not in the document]
jpuri: Good point @naugtur , @digiwand : I have updated the branch plz check.

Thetoan1 (approved):

app/scripts/app-init.js
```diff
@@ -111,14 +99,13 @@ function importAllScripts() {
...
 /*
  * Message event listener below loads script if they are no longer available.
  * chrome below needs to be replaced by cross-browser object,
  * but there is issue in importing webextension-polyfill into service worker.
  * chrome does seems to work in at-least all chromium based browsers
  */
-// eslint-disable-next-line no-undef
-chrome.runtime.onMessage.addListener(importAllScripts);
+// eslint-disable-next-line
+chrome.runtime.onMessage.addListener(() => {});
```
digiwand, line 111: After the latest changes to `importAllScripts`, can we remove this line and the comments above?

digiwand, line 111: After the latest changes around `importAllScripts`, can we remove this line and the comments above?

jpuri: reactivating service worker does not work if I remove this line

kumavis: you need to add a comment explaining that the otherwise meaningless looking line is important

kumavis (dismissed):
add comment to 
```js
chrome.runtime.onMessage.addListener(()
```

app/scripts/app-init.js
```diff
@@ -111,14 +99,13 @@ function importAllScripts() {
...
 }
 
-// eslint-disable-next-line no-undef
-self.addEventListener('install', importAllScripts);
+importAllScripts();
 
 /*
  * Message event listener below loads script if they are no longer available.
  * chrome below needs to be replaced by cross-browser object,
  * but there is issue in importing webextension-polyfill into service worker.
```
naugtur, line 107: The part of the polyfill that gives you this particular functionality should be under 20 lines. It'd be wise to copy-paste it here.

app/scripts/app-init.js
naugtur:
```js
chrome.runtime.onMessage.addListener((message) => {
  return message.name === WORKER_KEEPALIVE_MESSAGE;
});
```
Something like this, while useless, would accomplish following things:
1. share a reference to the message name so that it's easy to navigate the codebase to find where the keepalive is sent
2. explain what the line is for
3. avoid future breakage if a browser decides to unregister/ignore empty listeners to prevent unnecessary keepalive. (If you find that impossible, read up on the beforeunload event)

app/scripts/ui.js
```diff
@@ -24,34 +24,37 @@ import { setupMultiplex } from './lib/stream-utils';
...
-
-async function start() {
-  async function displayCriticalError(container, err, metamaskState) {
-    const html = await getErrorHtml(SUPPORT_LINK, metamaskState);
-
-    container.innerHTML = html;
+const container = document.getElementById('app-content');
 
-    const button = document.getElementById('critical-error-button');
+const WORKER_KEEP_ALIVE_INTERVAL = 1000;
```
naugtur, line 29: If the worker is killed after 5 minutes of inactivity, do we need to ping it every second? What's the reason behind this?

ui/index.js
naugtur: I agree it ooks brittle.

- backgroundConnection.getState calls its callback, which calls startApp, which is asynchronous and reduxStore is only set in a .then
- meanwhile, synchronously after, so before reduxStore is set, updateBackgroundConnection(backgroundConnection); is called

It will only work if we can guarantee startApp always resolves before backgroundConnection.onNotification event is triggered.  

Also, startApp makes 2 calls to actions, which means startApp can use actions but it's secretly limited to only the actions that don't depend on background connection.

I'd recommend keeping the additional early call and inside the _setBackgroundConnection function do nothing if argument is `===` to what's already set.

app/scripts/app-init.js
```diff
@@ -111,14 +99,13 @@ function importAllScripts() {
   }
 }
 
-// eslint-disable-next-line no-undef
-self.addEventListener('install', importAllScripts);
+importAllScripts();
```
naugtur, line 102: +1 to it being easier to reason about.

We should see if there's a measurable difference in loading behavior (via performance/JS profiler/tracing) between importing top-level vs importing on install.

app/scripts/ui.js
jpuri: I want it to receive this ping as soon as it goes inactive. 5 minutes is the approximate time after which service worker dies but it is not guaranteed interval.

app/scripts/app-init.js
jpuri: Actually service worker documents also suggest to use `install` event for loading assets. I have doubt if this approach is the best.

app/scripts/app-init.js
jpuri: We are using this polyfill library https://github.com/MetaMask/metamask-extension/blob/develop/package.json#L380

We need to figure out something different for service worker to avoid pasting whole thing here.

app/scripts/app-init.js
digiwand: ah makes sense now. I didn't realize there was a dual purpose in the previous iteration. We need to add the listener on the service worker to receive the messages for  `WORKER_KEEP_ALIVE_MESSAGE` 👍  

Thanks for updating the comment. I'm also a fan of @naugtur's comment and code to receive the message in hand.

digiwand: ah makes sense now. I didn't realize there was a dual purpose in the previous iteration. We need to add the listener on the service worker to receive the messages for  `WORKER_KEEP_ALIVE_MESSAGE` 👍  

Thanks for updating the comment! I'm also a fan of @naugtur's comment and code to receive the message in hand.

app/scripts/ui.js
digiwand: My thought was that sending the `WORKER_KEEP_ALIVE_MESSAGE` message resets the 5 minute-ish timer. If this is the case then I think it can be safe to move it up to 10 seconds, 30 seconds, or even a minute. Is this not the case?

app/scripts/app-init.js
naugtur: Yes, and here we only need a function to add a message listener, which seems to be a matter of getting the `runtime` reference depending on the environment.

we don't seem to be using anything else here and rightfully so.

app/scripts/ui.js
naugtur: Are you sure it dies after 5 minutes even if it's receiving messages during that period of time? If not, even a message every 4 minutes makes it immortal.

digiwand: Jyoti says it's not guaranteed, so 4 minutes seems like we'd be living on the edge

app/scripts/app-init.js
jpuri: I made the update

ui/index.js
jpuri: As discussed in code walkthrough a race condition is unlikely here

app/scripts/app-init.js
```diff
@@ -1,11 +1,12 @@
 // This file is used only for manifest version 3
 
-// Represents if importAllScripts has been run
 // eslint-disable-next-line
-let scriptsLoaded = false;
+const WORKER_KEEP_ALIVE_MESSAGE = 'UI_OPEN';
 
+// Represents if importAllScripts has been run
```
digiwand, line 6: I don't understand this comment with respect to `testMode`. Should we remove this? 
```suggestion
```

app/scripts/app-init.js
```diff
@@ -1,11 +1,12 @@
...
 // Variable testMode is set to true when preparing test build.
 // This helps in changing service worker execution in test environment.
+// eslint-disable-next-line
```
digiwand, line 9:
```suggestion
```

app/scripts/app-init.js
```diff
@@ -111,14 +104,15 @@ function importAllScripts() {
...
 
 /*
  * Message event listener below loads script if they are no longer available.
  * chrome below needs to be replaced by cross-browser object,
  * but there is issue in importing webextension-polyfill into service worker.
  * chrome does seems to work in at-least all chromium based browsers
  */
-// eslint-disable-next-line no-undef
-chrome.runtime.onMessage.addListener(importAllScripts);
+// eslint-disable-next-line
```
digiwand, line 115: We could add 
```
/* global chrome */
```
to the top of this file to allow other rules to continue working on this line

app/scripts/app-init.js
```diff
@@ -111,14 +104,15 @@ function importAllScripts() {
   }
 }
 
-// eslint-disable-next-line no-undef
-self.addEventListener('install', importAllScripts);
+importAllScripts();
 
 /*
  * Message event listener below loads script if they are no longer available.
```
digiwand, line 110: Can we update this to @naugtur's [suggestion earlier](https://github.com/MetaMask/metamask-extension/pull/14781#discussion_r930724118)? I think the proposed comment provides a bit more context. 

```
A keepalive message receiver to avoid Service Worker getting shut down due to inactivity. UI sends the message periodically.
```

jpuri:
Hey @digiwand : can you plz add a commit with above suggestions.

app/scripts/app-init.js
```diff
@@ -1,11 +1,12 @@
 // This file is used only for manifest version 3
 
-// Represents if importAllScripts has been run
 // eslint-disable-next-line
-let scriptsLoaded = false;
+const WORKER_KEEP_ALIVE_MESSAGE = 'UI_OPEN';
```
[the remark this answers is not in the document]
naugtur: +1 to that suggestion

app/scripts/app-init.js
```diff
@@ -111,14 +107,20 @@ function importAllScripts() {
...
  * chrome below needs to be replaced by cross-browser object,
  * but there is issue in importing webextension-polyfill into service worker.
  * chrome does seems to work in at-least all chromium based browsers
  */
-// eslint-disable-next-line no-undef
-chrome.runtime.onMessage.addListener(importAllScripts);
+// eslint-disable-next-line
+chrome.runtime.onMessage.addListener((message) => {
+  importAllScripts();
+  return false;
```
naugtur, line 125: I originally suggested to put something like  `return message.name === WORKER_KEEP_ALIVE_MESSAGE;` here because it'd help people find out where the message is coming from. I see a change where you remove the `const WORKER_KEEP_ALIVE_MESSAGE` line and replaced it with `return false;`. To be clear, there's no reason this function needs to return boolean, it was just means to get `WORKER_KEEP_ALIVE_MESSAGE` here. If you don't want that, there's no need to have the return statement.

jpuri: `return false` works well

I removed `return message.name === WORKER_KEEP_ALIVE_MESSAGE` as it is throwing error if this function returns true.

naugtur: It was not about returning anything, but pointing out the fact that message.name should be `WORKER_KEEP_ALIVE_MESSAGE` to the person wondering where the message is coming from. returning anything was not the point :) 

If it didn't work as an explanation for you, it's not gonna work on others :sweat_smile:

app/scripts/app-init.js
```diff
@@ -111,14 +108,20 @@ function importAllScripts() {
...
 }
 
+// Ref: https://stackoverflow.com/questions/66406672/chrome-extension-mv3-modularize-service-worker-js-file
 // eslint-disable-next-line no-undef
 self.addEventListener('install', importAllScripts);
 
 /*
- * Message event listener below loads script if they are no longer available.
+ * A keepalive message receiver to avoid Service Worker getting shut down due to inactivity.
+ * UI sends the message periodically.
```
naugtur, line 117:
```suggestion
 * A keepalive message listener to prevent Service Worker getting shut down due to inactivity.
 * UI sends the message periodically, in a setInterval. 
 * Chrome will revive the service worker if it was shut down, whenever a new message is sent, but only if a listener was defined here.
```

app/scripts/app-init.js
naugtur: I've added a suggestion to the comment above. Feel free to remove the return statement or keep it - whichever you find less surprising.

app/scripts/app-init.js
naugtur: I've added a suggestion to the comment above. Feel free to remove the return statement or keep it - whichever you find less surprising.

app/scripts/ui.js
naugtur: I've seen situations where 1s was helpful. Can confirm.

naugtur (dismissed):

app/scripts/app-init.js
digiwand: I think what @naugtur is trying to say is that it's not straightforward to know what this event listener is listening to. If someone were to read this file, they would need to put a debugger here and hope it lands (which we happen to know it currently does every second), or try to interpret it from the comment above which doesn't provide a silver bullet answer either. I think it's also better to enable the code to speak for itself rather than adding comments.


What if we did something similar to the `updateBackgroundConnection` logic below?
something like
```
if (message.name === 'WORKER_KEEP_ALIVE_MESSAGE') { return; }

console.warn('ServiceWorker: unknown message handled by onMessage event: ', message.name);
```

app/scripts/app-init.js
```diff
@@ -111,14 +110,20 @@ function importAllScripts() {
   }
 }
 
+// Ref: https://stackoverflow.com/questions/66406672/chrome-extension-mv3-modularize-service-worker-js-file
 // eslint-disable-next-line no-undef
 self.addEventListener('install', importAllScripts);
```
digiwand, line 115: I am wondering why we use `self.addEventListener('install', ...` here instead of `chrome.runtime.onInstalled.addListener`, and `chrome.runtime.onMessage.addListener` instead of  `self.addEventListener('message', ...`

app/scripts/app-init.js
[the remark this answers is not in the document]
jpuri: I did this change

naugtur (approved):

digiwand (approved):

PR #14795, opened 2022-05-27
MV3: Fix dynamic file list injection on service worker reload

jpuri:
MV3: Fix dynamic file list injection on service worker reload


kumavis (dismissed):

development/build/scripts.js
```diff
@@ -386,6 +386,21 @@ async function bundleMV3AppInitialiser({
...
+
+  let prevChromeFileContent;
+  watch('./dist/chrome/app-init.js', () => {
+    const chromeFileContent = readFileSync('./dist/chrome/app-init.js', 'utf8');
+    if (chromeFileContent !== prevChromeFileContent) {
+      mv3BrowserPlatforms.forEach((browser) => {
+        const appInitFile = `./dist/${browser}/app-init.js`;
+        const fileContent = readFileSync('./app/scripts/app-init.js', 'utf8');
+        const fileOutput = fileContent.replace('/** FILE NAMES */', fileList);
+        writeFileSync(appInitFile, fileOutput);
```
darkwing, line 398: I wont' block on it but this block looks identical to lines 384-387 -- in the future it might be good to create a helper function to avoid the duplication.

darkwing (dismissed):

adonesky1:
@jpuri could you explain a bit the approach being taken here?

development/build/scripts.js
jpuri: I updated the PR

jpuri:
> @jpuri could you explain a bit the approach being taken here?

Hey @adonesky1 : the issue this PR is solving is when a change is made to service worker file `app-init.js` while extension is running in development mode, service worker code was babelified but list of files was not injected dynamically here: https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/app-init.js#L36

I added a gulp watch tash which will check if `app-init.js` in `dist` is changed and if yes it dynamically injects file list.

adonesky1 (approved):
LGTM!

darkwing (approved):

PR #14799, opened 2022-05-30
Upgrade @metamask/eth-ledger-bridge-keyring

### LINKED ISSUE #6012: Ledger Nano: Metamask Address does not correspond to Ledger

jpuri:
Fixes: https://github.com/MetaMask/metamask-extension/issues/6012

Upgrade `@metamask/eth-ledger-bridge-keyring` to fix issue with ledger accounts.


digiwand (dismissed):

darkwing (approved):
Tested, everything works!

digiwand (approved):

PR #14803, opened 2022-05-31
Include snap version in pill

FrederikBolding:
## Explanation

Moves the snap version to the pill component for snap install.

### Before

image

### After

image

## Manual Testing Steps

1. Go to https://metamask.github.io/test-snaps/0.3.0/
2. Connect and install a snap
3. See that the snap version is included in the pill


eriknson:

hmalik88 (approved):
LGTM.

darkwing (approved):
Approving but we may want to try `padding-inline-end: 8px;` instead of `padding-right`;  did you check in RTL?

eriknson:
> 

It seems like we add a `dir="rtl"` property by the top of the DOM when the setting is set to an RTL language, flipping the suitable styling of the included elements automatically. So `padding-right` turns into `padding-left` in this case.

PR #14821, opened 2022-06-01
Handle an RPC provider delay in Swaps

dan437:
## Explanation
**Return an estimated amount for a completed swap if an RPC provider has a delay**
Right now if an RPC provider has a delay with updating the `postTxBalance` field and the destination token is a native one, we show a user much lower amount, which is equal to the gas fee. Thanks to this PR we will show a user an estimated amount if the `postTxBalance` hasn't been updated.

**Create a recursive function for updating post tx balance**
If `preTxBalance` and `postTxBalance` are equal, we will try to fetch `postTxBalance` again multiple times.

## Testing
- Add the Polygon network with the following RPC, which is sometimes delayed: `https://polygon-rpc.com/`
- Try to swap some token into MATIC
- We sometimes show you an estimated amount first and then we make a tiny amount update within ~30s


darkwing (changes requested):
Can we mock a test here?  I think this functionality could really use it.

darkwing (approved):

app/scripts/controllers/transactions/index.js
```diff
@@ -1465,6 +1467,39 @@ export default class TransactionController extends EventEmitter {
...
+      txMeta.chainId,
+    );
+    if (
+      isDefaultTokenAddress &&
+      txMeta.preTxBalance === latestTxMeta.postTxBalance &&
+      numberOfAttempts > 0
+    ) {
+      setTimeout(() => {
+        // If postTxBalance is the same as preTxBalance, try it again.
+        this.updatePostTxBalance({
```
adonesky1, line 1488: Seems unlikely but is there a world where we go through this recursively and they resolve out of order of when they were called and we end up with less fresh results?

adonesky1 (approved):
LGTM!

PR #14822, opened 2022-06-01
Fixed eth_getTransactionCount pending to return hex

### LINKED ISSUE #5845: eth_getTransactionCount of "pending" returns number instead of hex string
Calling eth_getTransactionCount against the "pending" block returns transaction count as a number. Asking for "latest" or any other block returns hex string of the transaction count.

In your browser console:
```
window.ethereum.sendAsync({"jsonrpc":"2.0","method":"eth_getTransactionCount","params":
 ["0x003bbce1eac59b406dd0e143e856542df3659075", "pending"],"id":1}, console.log)

{id: 1, jsonrpc: "2.0", result: 35693}
```

```
window.ethereum.sendAsync({"jsonrpc":"2.0","method":"eth_getTransactionCount","params":
 ["0x003bbce1eac59b406dd0e143e856542df3659075", "latest"],"id":1}, console.log)

{id: 1, jsonrpc: "2.0", result: "0x8b6d"}
```

**Expected behavior**:  The JSON-RPC interface standard uses hex-string for transactionCount result

 - OS: OS X
 - Chrome
 - MetaMask Version 5.0.3

Found while fixing this eth library issue: https://github.com/ethereumjs/ethrpc/pull/122
Discovered while using a hardware wallet to sign using custom code, but metamask is loaded and the `window.ethereum` object is used to query for nonce.

shanejonas:
## Explanation
`eth_getTransactionCount` had an issue where only for `"pending"` it was returning our internal nonce which is fine, but it was returning it as a number not as hex as described in the [spec](https://metamask.github.io/api-playground/api-documentation/#eth_getTransactionCount)

Fixes #5845

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -15,7 +15,7 @@ import {
   ethErrors,
 } from 'eth-rpc-errors';
 import { Mutex } from 'await-semaphore';
-import { stripHexPrefix } from 'ethereumjs-util';
+import { bnToHex, stripHexPrefix } from 'ethereumjs-util';
 import log from 'loglevel';
 import TrezorKeyring from 'eth-trezor-keyring';
 import LedgerBridgeKeyring from '@metamask/eth-ledger-bridge-keyring';
@@ -53,6 +53,7 @@ import {
 } from '@metamask/snap-controllers';
 ///: END:ONLY_INCLUDE_IN
 
+import { BN } from 'bn.js';
 import {
   TRANSACTION_STATUSES,
   TRANSACTION_TYPES,
@@ -3895,7 +3896,7 @@ export default class MetamaskController extends EventEmitter {
     const pendingNonce = nonceDetails.params.highestSuggested;
 
     releaseLock();
-    return pendingNonce;
+    return bnToHex(new BN(pendingNonce));
   }
 
   /**


danjm:
An annoying question: might existing dapps rely on it being a number?

shanejonas:
this is an example of https://www.hyrumslaw.com/

> With a sufficient number of users of an API,
it does not matter what you promise in the contract:
all observable behaviors of your system
will be depended on by somebody.

danjm (approved):
Code change looks good. I think we should figure out how we communicate this to developers. If there are even just a few dapps that depend on current behaviour, it would be best if we communicated with them ahead of time.

Even just:
- a short post somewhere
- an e-mail to our developer e-mail list that links to the post
- a tweet that links to the post

shanejonas:
@danjm can we use the "Whats new?" feature for this?

app/scripts/metamask-controller.js
```diff
@@ -3895,7 +3896,7 @@ export default class MetamaskController extends EventEmitter {
     const pendingNonce = nonceDetails.params.highestSuggested;
 
     releaseLock();
-    return pendingNonce;
+    return bnToHex(new BN(pendingNonce));
```
segun, line 3899: Can this be replaced with

```suggestion
    return `0x${BigInt(pendingNonce).toString(16)}`
```

This uses the browsers inbuilt bigint.

This is in line with your discussion about reducing build size cc @danjm

shanejonas:
I think we could add a page on docs.metamask.io that shows API changes for versions of MetaMask, similar to this page: https://docs.metamask.io/guide/provider-migration.html

danjm:
That makes sense @shanejonas 

I think we just need to establish a standard communication process for when we make changes that could break dapps. We probably should start an active e-mail list and make it known that it should be monitored if you want to stay on top of breaking changes.

So in addition to adding notices to the page you linked, we should have some way of pushing updates to that page out to developers

brad-decker:
@danjm @shanejonas did the notice ever go out and can we move forward with this PR?

adonesky1:
@shanejonas do we still want this?

PR #14823, opened 2022-06-01
Fix for: Invalid transaction envelope type: specified type "0x0" but including...requires type: "0x2"

### LINKED ISSUE #14761: Sentry Error: Invalid transaction envelope type: specified type "0x0" but including maxFeePerGas and maxPriorit..._Last 30 Days: 536
Sentry Issue: [METAMASK-NT4V](https://sentry.io/organizations/metamask/issues/3287354575/?referrer=github_integration)

```
Error: Invalid transaction envelope type: specified type "0x0" but including maxFeePerGas and maxPriorityFeePerGas requires type: "0x2"
  at "stack": "Error: Invalid transaction envelope type: specified type \"0x0\" but including maxFeePerGas and maxPriorityFeePerGas requires type: \"0x2\"\n    at new n ( (metamask/common-1.js:27:365128)\n    at a ()
  at ensureProperTransactionEnvelopeTypeProvided (app/scripts/controllers/transactions/lib/util.js:108:29)
  at None (app/scripts/controllers/transactions/lib/util.js:174:9)
  at Array.forEach (<anonymous>)
  at validateTxParams (app/scripts/controllers/transactions/lib/util.js:150:28)
...
(10 additional frame(s) were not displayed)
```

Last 30 Days: 536

segun:
## Explanation
This PR fixes sentry issue #14761 

# Description of the problem
We now have validations in place that checks that transaction params are "proper" before adding them to the pending tx list. But for some clients (mostly edge browsers), there were some transactions in pending without "proper" tx params but these clients have the new validations. Best guess is that they have these transactions in the pending list before updating to the latest client with validations.

Extension `app/scripts/controllers/transactions/pending-tx-tracker.js` will then try to re-submit these transactions which will fail validations then try again which will fail and so on and so forth, in my test, I was able to retry about 2K times before unit tests timed out.

The fix for this is in multiple layers

1. If we try to update a transaction (update calls `normalizeAndValidateTxParams`), and there's an error, we just don't throw the error, we first check if the txMeta warning message is `There was an error when resubmitting this transaction.` which means the transaction has been sent for resubmission and failed. In this case, we set transaction status to failed.

2. `setTxStatusFailed` calls `updateTransaction`, this will then call `normalizeAndValidateTxParams` and fail and we are stuck in an endless loop again. So we need to refactor out the parts really needed for update (`_updateTransactionHistory`). so `setTxStatusFailed` will call `_updateTransactionHistory` instead.

3. `setTxStatusFailed` also calls `_setTransactionStatus` which also calls `updateTransaction`. Again we are stuck in an endless loop, so we call our refactored `_updateTransactionHistory` from `_setTransactionStatus` instead.

## More Information

"proper" txParams
```
 {
            from: '0xe04ab39684a24d8d4124b114f3bd6fbeb779caca',
            to: '0x0c54FcCd2e384b4BB6f2E405Bf5Cbc15a017AaFb',
            value: '0x0',
            gasLimit: '0x5028',
            maxFeePerGas: '0x2540be400',
            maxPriorityFeePerGas: '0x3b9aca00',
          },
```

"failing" txParams
```
 {
            from: '0xe04ab39684a24d8d4124b114f3bd6fbeb779caca',
            to: '0x0c54FcCd2e384b4BB6f2E405Bf5Cbc15a017AaFb',
            value: '0x0',
            gasLimit: '0x5028',
            maxFeePerGas: '0x2540be400',
            maxPriorityFeePerGas: '0x3b9aca00',
            type: '0x0',
 },
```

Fixes #14761

I couldn't replicate this issue because there's currently no way to get a tx with an "improper" tx params into state. But I have included unit tests to cover this scenario.

## Pre-Merge Checklist

### DIFF
--- a/app/scripts/controllers/transactions/pending-tx-tracker.js
+++ b/app/scripts/controllers/transactions/pending-tx-tracker.js
@@ -2,6 +2,7 @@ import EventEmitter from 'safe-event-emitter';
 import log from 'loglevel';
 import EthQuery from 'ethjs-query';
 import { TRANSACTION_STATUSES } from '../../../../shared/constants/transaction';
+import { ERROR_SUBMITTING } from './tx-state-manager';
 
 /**
  * Event emitter utility class for tracking the transactions as they
@@ -106,7 +107,7 @@ export default class PendingTransactionTracker extends EventEmitter {
         // encountered real error - transition to error state
         txMeta.warning = {
           error: errorMessage,
-          message: 'There was an error when resubmitting this transaction.',
+          message: ERROR_SUBMITTING,
         };
         this.emit('tx:warning', txMeta, err);
       }
--- a/app/scripts/controllers/transactions/tx-state-manager.js
+++ b/app/scripts/controllers/transactions/tx-state-manager.js
@@ -18,6 +18,8 @@ import {
   validateConfirmedExternalTransaction,
 } from './lib/util';
 
+export const ERROR_SUBMITTING =
+  'There was an error when resubmitting this transaction.';
 /**
  * TransactionStatuses reimported from the shared transaction constants file
  *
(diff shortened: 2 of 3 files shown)


darkwing (approved):
This looks good to me, Olu.  Everything seems to work well and the test makes sense.

danjm (approved):
LGTM

jpuri (approved):

Issue #14829, opened 2022-06-02
[Bug]: Warning - A component is changing an uncontrolled input of type text to be controlled


seaona:
### Describe the bug

**Problem**: when saving a contact info from a tx details we see the following warning on the console `Warning: A component is changing an uncontrolled input of type text to be controlled`

https://user-images.githubusercontent.com/54408225/171586023-4fc21556-3c48-4783-ad87-1235da62bdb0.mp4

### Steps to reproduce

1. Unlock Metamask
2. Perform a tx to another address
3. Click on Activity tab
4. Click on tx details
5. Click on the receiver address
6. Click Add a nickname
7. Introduce text on input fields
8. Check console

### Version

10.14.6

### Browser

Chrome

### Operating system

Linux

Issue #14828, opened 2022-06-02
[Bug]: Warning - Failed prop type: Invalid prop `maxPriorityFee` and `maxFee` of type `string` supplied to `AdvancedGasControls`, expected `number`


seaona:
### Describe the bug

**Problem**: when opening advance gas fee edit screen we can see these 2 warning errors in the console:
- [Bug]: Warning - Failed prop type: Invalid prop `maxPriorityFee` of type `string` supplied to `AdvancedGasControls`, expected `number`
- [Bug]: Warning - Failed prop type: Invalid prop `maxFee` of type `string` supplied to `AdvancedGasControls`, expected `number`

https://user-images.githubusercontent.com/54408225/171584195-5f30eb06-db1c-4632-8f08-a45c4f306905.mp4

### Steps to reproduce

1. Unlock Metamask
2. Click Send ETH
3. Introduce amount and receiver address and proceed
4. Click Edit gas fees
5. Click Advanced Options
6. Check console

### Version

10.14.6

### Browser

Chrome

### Operating system

Linux

bschorchit:
https://github.com/MetaMask/metamask-extension/pull/14837

Issue #14826, opened 2022-06-02
[Bug]: Warning - Failed prop type: seedPhraseBackedUp is marked as required at Home but its value is null


seaona:
### Describe the bug

**Problem**: on home page we see this warning error `Failed prop type: seedPhraseBackedUp is marked as required at Home but its value is null`

![image](https://user-images.githubusercontent.com/54408225/171582054-f74c927c-adf3-49ee-a62c-71008c82335e.png)

Reported by @ger

### Steps to reproduce

 tbd

### Version

10.14.6

### Browser

Chrome

### Operating system

Linux

bschorchit:
https://github.com/MetaMask/metamask-extension/pull/14837

Issue #14825, opened 2022-06-02
[Bug]: Warning - Failed prop type: onRemove and onViewTokenDetails


seaona:
### Describe the bug

**Problem**: when initiating Metamask for the first time, and clicking on ETH asset, we trigger the following warnings:
- `Warning: Failed prop type: The prop onRemove is marked as required in AssetOptions, but its value is undefined`
- `Warning: Failed prop type: The prop onViewTokenDetails is marked as required in AssetOptions, but its value is undefined`

https://user-images.githubusercontent.com/54408225/171581467-bd9f2e3e-2ceb-4176-8f91-de2b47ce34fe.mp4

### Steps to reproduce

1. Import Metamask on your browser
2. Start a new Wallet
3. Click on ETH Asset
4. Check console

### Version

10.14.6

### Browser

Chrome

### Operating system

Linux

bschorchit:
https://github.com/MetaMask/metamask-extension/pull/14837

PR #14837, opened 2022-06-02
Prop type fixups

BelfordZ:
## Explanation

This PR includes fixes for various proptype issues. Each commit addresses one issue (except there was some extra junk that snuck into the first commit and was plucked out in the last).

issues fixed: 
- #14764
- #14825
- #14826 
- #14829 
- #14828
- #14812
- #14813
- #14814
- #14815 
- #14816
- #14817

Pretty simple and straightforward solutions to each issue.

## Manual Testing Steps

Reproduction steps can be found inside each issue.

## Pre-Merge Checklist


ui/pages/first-time-flow/first-time-flow-switch/first-time-flow-switch.component.js
```diff
@@ -34,7 +34,7 @@ export default class FirstTimeFlowSwitch extends PureComponent {
       return <Redirect to={{ pathname: DEFAULT_ROUTE }} />;
     }
 
-    if (seedPhraseBackedUp !== null) {
+    if (seedPhraseBackedUp) {
```
BelfordZ, line 37: From what I can tell, there is no case where seedphrase is used as either true | false | null. It seems to be only true | null.

digiwand: I'm hesitant to change this because we have a test to support `seedPhraseBackedUp` as false  in `ui/pages/first-time-flow/first-time-flow-switch/first-time-flow-switch.test.js`. See "redirects to end of flow route when seedPhraseBackedUp is false"

ui/pages/settings/settings.component.js
```diff
@@ -342,6 +342,10 @@ class SettingsPage extends PureComponent {
           path={ADD_NETWORK_ROUTE}
           render={() => <NetworksTab addNewNetwork />}
         />
+        <Route
+          path={NETWORKS_ROUTE}
+          render={() => <NetworksTab addNewNetwork={false} />}
+        />
         <Route path={NETWORKS_ROUTE} component={NetworksTab} />
```
digiwand, line 349:
```suggestion
```

looks like we can remove this now that we have the new Route you created 🧹🙂

ui/components/app/gas-details-item/gas-details-item.js
```diff
@@ -90,8 +90,8 @@ const GasDetailsItem = ({ userAcknowledgedGasMissing = false }) => {
       }
       subTitle={
         <GasTiming
-          maxPriorityFeePerGas={maxPriorityFeePerGas}
-          maxFeePerGas={maxFeePerGas}
+          maxPriorityFeePerGas={maxPriorityFeePerGas.toString()}
```
digiwand, line 93: LGTM! Heads up, we might want to do future work on these two props, `maxPriorityFeePerGas` and `maxFeePerGas`, to keep data type passed around our app consistent, as strings. Some details in this ticket https://github.com/MetaMask/metamask-extension/pull/14665

digiwand (commented):
Yay! The console looks much better with all of these prop-type errors and warnings fixed! Thank you

1 request to reconsider supporting `seedPhraseBackedUp` as a false value

seaona:
@BelfordZ I've QAd this PR and all the warnings are now fixed! Amazing work :100: A couple of inputs:
- I've realized that there is another Warning I didn't report, so I've opened a new [github issue](https://github.com/MetaMask/metamask-extension/issues/14844). Don't have to be fixed here necessarily, just FYI. Sorry about that!
- There is some [audit failure](https://app.circleci.com/pipelines/github/MetaMask/metamask-extension/23576/workflows/bd549165-76b6-4bf9-b733-966d65b24ab6/jobs/605536), which I think it's fixed if you pull the last changes.
- For the[ lint errors](https://app.circleci.com/pipelines/github/MetaMask/metamask-extension/23576/workflows/bd549165-76b6-4bf9-b733-966d65b24ab6/jobs/605541) you can just do `yarn lint` on your local and fix the highlighted issues

ui/pages/first-time-flow/first-time-flow-switch/first-time-flow-switch.component.js
BelfordZ: Amazing, thank you for pointing this out. I will revert this.

BelfordZ:
@seaona @digiwand should be all sorted now! Thank you for the thorough review!

ui/components/ui/update-nickname-popover/update-nickname-popover.js
```diff
@@ -20,8 +20,10 @@ export default function UpdateNicknamePopover({
 }) {
   const t = useContext(I18nContext);
 
-  const [nicknameInput, setNicknameInput] = useState(nickname);
-  const [memoInput, setMemoInput] = useState(memo);
+  const [nicknameInput, setNicknameInput] = useState(
+    nickname === null ? '' : nickname,
```
digiwand, line 24: is it possible nickname could be undefined as well? This should happen if no nickname is passed to this component. What if we reverse the change to this line and set a default value to nickname on line 15?

Same comment for `memoInput` below.

```
nickname = '',
```

ui/pages/asset/components/asset-options.js
```diff
@@ -85,14 +85,39 @@ const AssetOptions = ({
   );
 };
 
+const isntFunc = (p) => {
```
digiwand, line 88: REQ: This is kind of silly and opinionated, but could we change this to `isNotFunc`? 

Reasons: 
- `isNot` is known to our codebase while `isnt` isn't. Keeping `isNot` will maintain consistency which will help with readibility
- the missing apostrophe may throw someone off one day, especially if English is not their first language

danjm:
Needs a rebase on latest develop to fix the test-deps-audit failure

ui/components/ui/update-nickname-popover/update-nickname-popover.js
BelfordZ: While this greatly improve readability, we would still need to handle the case where `nickname` or `memo` are passed in explicitly as `null`. Setting a default value inside of object destructuring will treat `null` as being defined, thus not defaulting it. For example:
```
function foo({ a, b = 2, c }) { console.log(a, b, c); }
foo({a: 1, b: null, c: 3})
// 1 null 3
foo({a: 1, c: 3})
// 1 2 3
```
I think there is a broader issue here: we are allowing those props to be `null | undefined | string`, where null and undefined are treated logically the same. I can add the `nickname = ''` to handle the `undefined` case, though we still need to handle the `null` case (and so I would need to keep the `nickname === null ? '' : nickname` stuff).

BelfordZ: Also I think this components' `Address` prop should be `PropTypes.string.isRequired`, but that should be done as a seperate issue so that we can keep the scope minimized for this already widely scoped PR =D

ui/pages/asset/components/asset-options.js
BelfordZ: agreed, will fix.

BelfordZ:
@danjm thank you for the tip, all rebased now.
@digiwand thx again for the feedback! Hopefully the lastest commit addresses the issues brought up.

ui/components/ui/update-nickname-popover/update-nickname-popover.js
digiwand: Ooh right. Passing in null will not take the default value provided. Glad you caught this. What if we used the nullish operator here? 
```suggestion
    nickname ?? ''
```
ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

BelfordZ: I've added the handling of `undefined` (since it is *possible*)

digiwand: I'm also not sure if there would be a case that we pass null, but I like that we're keeping this robust. Sweet :)

digiwand (approved):
LGTM!

ui/components/ui/update-nickname-popover/update-nickname-popover.js
BelfordZ: never hurts to make things stronger!

BelfordZ: > Ooh right. Passing in null will not take the default value provided. Glad you caught this. What if we used the nullish operator here?
> 
> ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

you learn something every day!!

darkwing (approved):
Well done @BelfordZ !

PR #14870, opened 2022-06-06
Enhancement to e2e test for  phishing detection

danjm:
This assertion helps verify that the extension has taken the user to the appropriate error screen if the phishing detection screen is inside an iframe and is attempting to take the user to a blocked resource.


test/e2e/tests/phishing-detection.spec.js
```diff
@@ -124,6 +124,10 @@ describe('Phishing Detection', function () {
 
         // Ensure we're not on the wallet home page
         await driver.assertElementNotPresent('[data-testid="wallet-balance"]');
+
+        // Check that the user has been directed to the appropriate warning screen
+        const currentUrl = await driver.getCurrentUrl();
+        assert.equal(currentUrl, 'chrome-extension://invalid/');
```
Gudahtt, line 130: This if course will not work on Firefox

Gudahtt: Maybe we can check that the current URL is still the warning page OR it's this URL. It'd be the warning page on Firefox, and on older versions of Chrome. Both still demonstrate this was successful.

danjm: > This if course will not work on Firefox

:man_facepalming: 

I swear I made a note to myself before I stashed my changes on this hours ago, and then unstashed committed and pushed without thinking after coming back to it

Your suggestion is good, thanks

Ash1818:
Will be contacting bounty bug to go over it’s source. Websites have failed
like Uswap allowing users like myself to connect to inknown wallets causing
damages that get hacked . Hackers sent out spam it’s get clicked on because
of no security on Thai platforms is the real question I have looked Ivey
exactly what is out there. Hacker gets access to the users email cause
Another device to connect to its network mirroring there decide and
controlling it getting access to photos including personal pictures and
that’s in invasion of price at  and getting access to users wallets and
bank information. Apple is also responsible and will be handled according
to let’s fix this and act accordingly thanks Ash

brad-decker:
@danjm updated branch just to check, i realize there are changes necessary still but this bumps the PR into the limelight again and maybe we can get this landed without too much additional effort.

brad-decker:
Converted to draft to give time for this to be updated and made ready for review again @danjm

PR #14894, opened 2022-06-08
Create e2e test for snap in-app notification

bowensanders:
This creates an e2e test for snaps in-app notifications, as seen in the metamask.github.io/test-snaps/0.4.0 page. 
This test is based on the previous work on the error and manageState e2e tests.


test/e2e/snaps/test-snap-notification.spec.js
```diff
@@ -0,0 +1,112 @@
...
+        // check to see that there is one notification
+        const notificationResult = await driver.findElement(
+          '.account-menu__icon__notification-count',
+        );
+        assert.equal(await notificationResult.getText(), '1');
+
+        // try to click on the account menu icon (via xpath)
+        await driver.clickElement({
+          xpath:
+            '//*[@id="app-content"]/div/div[1]/div/div[2]/div[2]/div[1]/div',
```
FrederikBolding, line 91: Could you not just click `.account-menu__icon`?

test/e2e/snaps/test-snap-notification.spec.js
```diff
@@ -0,0 +1,112 @@
...
+        });
+        await driver.delay(500);
+
+        // try to click on the notification item (via xpath)
+        await driver.clickElement({
+          xpath: '//*[@id="app-content"]/div/div[3]/div[10]',
```
FrederikBolding, line 97: It may be beneficial that we add a `testid` for this. If we add a `testid` prop to the element we are looking for we can use:

```
await driver.clickElement(
'[data-testid="notifications-list-item"]',
);
```
or something like that.

test/e2e/snaps/test-snap-notification.spec.js
```diff
@@ -0,0 +1,112 @@
...
+        });
+        await driver.delay(500);
+
+        // look for the correct text in notifications (via xpath)
+        const notificationResultMessage = await driver.findElement({
+          xpath: '//*[@id="app-content"]/div/div[3]/div/div[2]/div/div[2]/p[1]',
```
FrederikBolding, line 103: May also need a `testid` or a better selector.

FrederikBolding (changes requested):
Overall looks good. I think we should use better selectors than the hardcoded xpath ones since those are likely to break with newly introduced HTML.

test/e2e/snaps/test-snap-notification.spec.js
bowensanders: ✅

joani24 (dismissed):

test/e2e/snaps/test-snap-notification.spec.js
bowensanders: I replaced this with a search for text/tag

test/e2e/snaps/test-snap-notification.spec.js
bowensanders: This was also solved by just looking for the generic '.notifications__item__details__message' tag as it would only be triggered the first time by a test correctly identified already as 1 of 1 message

FrederikBolding (approved):

hmalik88 (approved):
LGTM.

PR #14903, opened 2022-06-09
Add internal EIP-1193 provider

rekmarks:
# WIP


Nana12345678910:
I can't access my wallet and all my card are not working it keeps saying code error iv been trying to link my new number over but it's been impossible

brad-decker:
@rekmarks is this something that we should see through to completion? Is there a ticket tracking the underlying issue somewhere we can use to prioritize the work for this with?

PR #14904, opened 2022-06-09
@metamask/providers@9.0.0

rekmarks:
Bump `@metamask/providers` to `9.0.0`. Should be completely non-breaking for our purposes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
     "@metamask/metamask-eth-abis": "^3.0.0",
     "@metamask/obs-store": "^5.0.0",
     "@metamask/post-message-stream": "^4.0.0",
-    "@metamask/providers": "^8.1.1",
+    "@metamask/providers": "^9.0.0",
     "@metamask/rpc-methods": "^0.15.0",
     "@metamask/slip44": "^2.1.0",
     "@metamask/smart-transactions-controller": "^2.0.1",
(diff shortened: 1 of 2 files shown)


joani24:
> Bata `@metamask/providers`para `9.0.0`. Deve ser completamente inquebrável para nossos propósitos.

FrederikBolding (approved):

darkwing:
Looking at the changelog (https://github.com/MetaMask/providers/commit/a6e3b8a50d3e33d2147deeee85366f2c2fe3eac2), I'm seeing lots of breaking changes.  Anything specific we should be manually testing for?

FrederikBolding:
@darkwing The only breaking change in `9.0.0` is moving some logic from `BaseProvider` to a new class. But `BaseProvider` isn't used directly anywhere I could find.

rekmarks:
@darkwing all of the breaking changes should be in the `BaseProvider` / `StreamProvider`, there should be no breaking changes whatsoever in the `MetaMaskInpageProvider`, which is what we inject via our content script. It is not impossible that we missed something during review so I think we'll have to rely on careful QA of the provider behavior on the test dapp while doing lots of network switches and account switches etc. to see how it does.

cc: @PeterYinusa @tmashuang

darkwing (approved):

PR #14913, opened 2022-06-09
mv3/app-init - import all scripts at the same time

kumavis:
calls `importScripts` with all scripts at the same time. does not show a strong immediate performance gain on my machine, but gives chrome the best shot of handling it optimally :shrug:

### DIFF
--- a/app/scripts/app-init.js
+++ b/app/scripts/app-init.js
@@ -17,18 +17,20 @@ function importAllScripts() {
   // we need to dynamically inject values of applyLavaMoat once this is fixed.
   const applyLavaMoat = true;
 
-  tryImport('./globalthis.js');
-  tryImport('./sentry-install.js');
+  const systemFiles = [];
+
+  systemFiles.push('./globalthis.js');
+  systemFiles.push('./sentry-install.js');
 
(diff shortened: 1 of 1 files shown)


naugtur:
importScripts pauses all execution for the time it takes to load and execute the scripts. 
The implementation in chromium unfortunately just iterates over scripts one by one.

https://chromium.googlesource.com/chromium/src/third_party/+/refs/heads/main/blink/renderer/core/workers/worker_global_scope.cc#298

This article says importScripts can run in the install hook as well 
https://developers.google.com/web/updates/2018/10/tweaks-to-addAll-importScripts

Which gave me and idea to maybe try to first populate the cache with the contents of the files (which the browser might do with some degree of parallelization) and then proceed to importScripts in install, so the for loop would not wait fro each fetch one by one, but use cache. 

This is the function used to load the script synchronously https://chromium.googlesource.com/chromium/src/third_party/+/refs/heads/main/blink/renderer/core/workers/worker_classic_script_loader.cc#103  
Unfortunately, I can't seem to find out if it's using a cache for network at all.

jpuri:
The change is already in develop thus closing this PR.

PR #14914, opened 2022-06-09
Changed accounts fuzzy search threshhold to match all

### LINKED ISSUE #14899: [Bug]: Account Search functionality not returning correct results
### Describe the bug

**Problem**: looks like the Account Search functionality is having some strange behaviour, as it's returning some "non-matching" results.
For example, if I search for `3d` I see 1 Result, whereas if I add a letter and search for `3da`, I now see 3 results. This seems not possible/wrong.

https://user-images.githubusercontent.com/54408225/172599347-8149de10-3685-466f-9192-0b75df441403.mp4

### Steps to reproduce

1. Unlock Metamask
2. Add a couple of accounts (Create Account)
3. Start searching for some chars (depending on each wallet will see one results or others, but with very few tries you can already see that some results are not consistent with the search)

```shell
n/a
```

### Version

10.14.7

### Build type

_No response_

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

BelfordZ:
## Explanation

This PR is to resolve https://github.com/MetaMask/metamask-extension/issues/14899

I've taken a look at [the fuzzy search options provided to fuse](https://fusejs.io/api/options.html) vs. [what is being used in the extension](https://github.com/MetaMask/metamask-extension/blob/b599035a12010e31b85c3733fb573ba05b840b3c/ui/components/app/account-menu/account-menu.component.js#L102), and tried a couple variations. 

A more detailed explanation of the way fuse ranks searches can be found [here](https://fusejs.io/concepts/scoring-theory.html#scoring-theory)

From what I can tell, this issue creeps up as a result of having the search scoring weighted across 2 fields. When you add a 3rd character, you get more search matches which are above the threshold provided. Secondly, it's a little extra odd because the fuse lib normalizes the scores based on the length of the searched strings. In our case, this means that if you use the default title or anything shorter than the address, matches for title end up scored higher. 

I've adjusted the settings slightly, and tested against some addresses of my own. @seaona if you could please check it against your example and let me know if its better for you, that would be greatly appreciated :).

### DIFF
--- a/ui/components/app/account-menu/account-menu.component.js
+++ b/ui/components/app/account-menu/account-menu.component.js
@@ -100,11 +100,12 @@ export default class AccountMenu extends Component {
   };
 
   addressFuse = new Fuse([], {
-    threshold: 0.45,
+    threshold: 0.55,
     location: 0,
     distance: 100,
     maxPatternLength: 32,
     minMatchCharLength: 1,
+    ignoreFieldNorm: true,
     keys: [
       { name: 'name', weight: 0.5 },
       { name: 'address', weight: 0.5 },
--- a/ui/components/app/account-menu/account-menu.component.js
+++ b/ui/components/app/account-menu/account-menu.component.js
@@ -132,7 +132,7 @@ export default class AccountMenu extends Component {
 
   renderAccountsSearch() {
     const handleChange = (e) => {
-      const val = e.target.value.length > 1 ? e.target.value : "";
+      const val = e.target.value.length > 1 ? e.target.value : '';
       this.setSearchQuery(val);
     };
 
--- a/ui/components/app/account-menu/account-menu.component.js
+++ b/ui/components/app/account-menu/account-menu.component.js
@@ -150,8 +150,7 @@ export default class AccountMenu extends Component {
         id="search-accounts"
         placeholder={this.context.t('searchAccounts')}
         type="text"
-        value={this.state.searchQuery}
-        onChange={(e) => this.setSearchQuery(e.target.value)}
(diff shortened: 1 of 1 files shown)


seaona:
@BelfordZ it works much better now. Looks that this new parameters give some more consistent matching results.
I've identified occasionally some "weirdness" while using 1 char. See example below, where looking for `p` has no matchings, but if I look for `pf` returns matchings.

https://user-images.githubusercontent.com/54408225/173334808-8bbf0480-2de9-44cc-89af-06eed41a55a0.mp4

I've seen that for the Settings Search, the function does not return any result unless we write at least 3 or more chars. Maybe we could take the same approach for 1-2 chars? What do you think?

https://user-images.githubusercontent.com/54408225/173334933-a4d16493-3e9b-4fdb-8665-0ab49616dc4d.mp4

BelfordZ:
@seaona delaying the search until the first 3 chars are written would for sure fix a lot of the oddities around search, especially the single-character case.

I will update the PR with your great suggestion, thanks for that.

ui/components/app/account-menu/account-menu.component.js
```diff
@@ -130,6 +131,11 @@ export default class AccountMenu extends Component {
   }
 
   renderAccountsSearch() {
+    const handleChange = (e) => {
+      const val = e.target.value.length > 1 ? e.target.value : "";
```
BelfordZ, line 135:
```suggestion
      const val = e.target.value.length > 1 ? e.target.value : '';
```

seaona (approved):

seaona:
@BelfordZ this sounds good to me! The Search consolidation could be done on a separate PR, but I would discuss it first with the UI/UX team or @danjm

brad-decker (approved):

PR #14922, opened 2022-06-10
Updated Dutch language

JasperAlexander:
## Explanation

Current Dutch language support is deprecated. This is a problem because not all Dutch people can speak English. I added and translated all missing messages from the English language file and removed the deprecated messages.

It is not clear to me where to put this change into the CHANGELOG.md file.


JasperAlexander:
I'm going to open a clean PR.

PR #14936, opened 2022-06-13
Make the accounts menu accessible

darkwing:
Users that are sighted and unsighted should be able to navigate the accounts dropdown, which is incredibly important, with their keyboard.

https://user-images.githubusercontent.com/46655/173465041-8ee2ac94-41ed-4ba6-b266-a162c41802c8.mp4


ryanml (approved):
++

digiwand (approved):

PR #14937, opened 2022-06-14
Increase likelyhood of valid method signatures being returned by `getMethodData`

FrederikBolding:
## Explanation
Less than ideal method signatures have been added to the decentralized method registry that we use along with the 4byte directory. This PR changes how we choose which of these signatures to use. 

The new logic boils down to: Try both the decentralized method registry and the 4byte directory, for the 4byte directory picking the oldest method signature entry. When choosing between the signature from the decentralized registry and the 4byte registry, pick the shortest signature.

## Manual Testing Steps
For the specific bug:
- Complete an ERC721 approval and see that the method signature is correct

In general:
- Complete a few contract interactions to verify that method signatures are correct

## Pre-Merge Checklist

### DIFF
--- a/app/scripts/migrations/072.js
+++ b/app/scripts/migrations/072.js
@@ -3,7 +3,7 @@ import { cloneDeep } from 'lodash';
 const version = 72;
 
 /**
- * Renames NotificationController to AnnouncementController
+ * Should empty the `knownMethodData` object in PreferencesController
  */
 export default {
   version,
--- a/app/scripts/migrations/072.js
+++ b/app/scripts/migrations/072.js
@@ -1,0 +1,30 @@
+import { cloneDeep } from 'lodash';
+
+const version = 72;
+
+/**
+ * Renames NotificationController to AnnouncementController
+ */
+export default {
+  version,
+  async migrate(originalVersionedData) {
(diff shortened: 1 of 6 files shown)


Nana12345678910:
I don't have access to my old phone number any more witch was 0405161403 my new number is 0405433015

adonesky1:
Looks right to me. As [mentioned in that slack thread](https://consensys.slack.com/archives/GTQAGKY5V/p1655211587717079?thread_ts=1651507012.193889&cid=GTQAGKY5V), we will need to do migration to clear out the "knownMethodData" so these changes can be seen for users with incorrect data already persisted to state.

adonesky1:
Happy to add the migration here if you'd like the assist @FrederikBolding

FrederikBolding:
> Happy to add the migration here if you'd like the assist @FrederikBolding 

Feel free 👍

danjm:
I see @adonesky1 offered to add a migration. That would be good. Alex: I think we can just clear the `knownMethodData` for all users. It is just a cache to reduce network calls, and will be repopulated.

adonesky1 (approved):
👍

kumavis (approved):

PR #14943, opened 2022-06-14
Add missing subtitles for secret recovery phrase

rickycodes:
## Explanation

This syncs translations for the seedphrase video from what's in mobile. This includes several missing languages:
French, Greek, Turkish, Chinese

The was German added previously, but the map was never updated to include it so I don't think that would have been working?

If we decide that replacing all these with what's in mobile is too heavy handed (it might be?) we could just add what's missing _instead_. That said, I think we should keep these subtitles consistent across products.


adonesky1 (dismissed):
LGTM!

darkwing (dismissed):

adonesky1 (approved):

darkwing (approved):

PR #14944, opened 2022-06-14
change getMethodFrom4Byte to return oldest entry when multiple exist for a given 4byte key

adonesky1:
## Explanation
Currently when we query [4byte.directory](https://www.4byte.directory/) for the name of a method associated with the 4byte hex string in the data of a given transaction, we only return a value if only one entry is returned. This directory is public and appears to be manipulated at times: 

## Pre-Merge Checklist

### DIFF
--- a/ui/helpers/utils/transactions.util.js
+++ b/ui/helpers/utils/transactions.util.js
@@ -38,7 +38,13 @@ async function getMethodFrom4Byte(fourBytePrefix) {
 
   if (fourByteResponse.count === 1) {
     return fourByteResponse.results[0].text_signature;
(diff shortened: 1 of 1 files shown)


FrederikBolding:
@adonesky1 Can probably be closed in favor of #14937 ?

FrederikBolding:
> Example manipulation: https://www.4byte.directory/signatures/?bytes4_signature=0xa22cb465
> 
> When the method is called, MetaMask shows `niceFunctionHerePlzClick943230089`.

Thanks for reporting! We are aware and it should be fixed by https://github.com/MetaMask/metamask-extension/pull/14937

PR #14966, opened 2022-06-15
Add support for non-ssl rpc urls to be added via api

### LINKED ISSUE #14416: [Bug]: wallet_addEthereumChain() does not support HTTP RPCs
### Describe the bug

When you run

```
await ethereum.request({
        method: 'wallet_addEthereumChain',
        params: [
          {
            chainId: '0x539',
            chainName: 'MyTestnet',
            rpcUrls: ['http://1.1.1.1:8545'],
          },
        ],
      });
```

it errors out with `Expected an array with at least one valid string HTTPS url 'rpcUrls', Received: http://1.1.1.1:8545`.

However, the user is able to manually add an HTTP url within their own RPC settings, even though it fails on programmatic adding. For consistency, Metamask should allow programmatically adding an HTTP url.

Additionally, [EIP-3085]([](https://eips.ethereum.org/EIPS/eip-3085#returns)All URL strings MUST include the protocol component of the URL. HTTPS SHOULD always be used over HTTP.) specifies HTTPS urls as a SHOULD, not a MUST. Given the prevalence of running local clients and exposing ports without SSL certificates, this would be a great improvement for users interested in self-hosting.

The change would need to be made here: https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js#L86

The fix is simple: check for an HTTP or HTTPS uri, instead of just an HTTPS uri.

### Steps to reproduce

Run within the JS console:

```
await ethereum.request({
        method: 'wallet_addEthereumChain',
        params: [
          {
            chainId: '0x539',
            chainName: 'MyTestnet',
            rpcUrls: ['http://1.1.1.1:8545'],
          },
        ],
      });
```

_No response_

### Version

10.12.3

### Build type

_No response_

### Browser

Chrome

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

BelfordZ:
fixes #14416 

WIP WIP

todo:

- add warning messaging to confirmation dialog

### DIFF
--- a/app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
+++ b/app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
@@ -24,6 +24,21 @@ const addEthereumChain = {
 };
 export default addEthereumChain;
 
+const isLocalhost = (strUrl) => {
+  try {
+    const url = new URL(strUrl);
+    return url.hostname === 'localhost' || url.hostname === '127.0.0.1';
+  } catch (error) {
+    return false;
+  }
+};
+
+const firstValidUrl = (urls) => urls.find(
+  (rpcUrl) => isLocalhost(rpcUrl) ||
+    validUrl.isHttpsUri(rpcUrl) ||
+    validUrl.isHttpUri(rpcUrl)
+);
+
 async function addEthereumChainHandler(
   req,
   res,
@@ -53,7 +68,7 @@ async function addEthereumChainHandler(
   const {
     chainId,
     chainName = null,
-    blockExplorerUrls = null,
+    blockExplorerUrls = [],
     nativeCurrency = null,
     rpcUrls,
   } = req.params[0];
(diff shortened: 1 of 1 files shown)


app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
```diff
@@ -77,29 +92,9 @@ async function addEthereumChainHandler(
...
-    blockExplorerUrls !== null && Array.isArray(blockExplorerUrls)
-      ? blockExplorerUrls.find(
-          (blockExplorerUrl) =>
-            isLocalhost(blockExplorerUrl) ||
-            validUrl.isHttpsUri(blockExplorerUrl),
-        )
-      : null;
+  const rpcs = Array.isArray(rpcUrls) ? rpcUrls : [rpcUrls];
+  const firstValidRPCUrl = firstValidUrl(rpcs);
+  const firstValidBlockExplorerUrl = firstValidUrl(blockExplorerUrls);
```
BelfordZ, line 97: allowing non-ssl explorer urls as well.

darkwing:
Is this ready for review (re: todo todo)?

PR #14967, opened 2022-06-16
Manage accounts feature for snaps

tmpfs:
## Explanation

Implements the `snap_manageAccounts` permission and RPC method handler.

## More Information

Uses a new type of keyring called [snap-keyring](https://github.com/LavaMoat/snap-keyring/) which has been added as a dependency using `snap-keyring` but we should consider moving this to the `@metamask` scope, advice on setting up Github CI to publish to that scope would be appreciated.

We need to isolate the accounts for each snap by identifier and as such we need a stable identifier, currently when the `methodImplementation` is called the `origin` passed in the `context` is the `snapId` so at the moment we **cannot consider the account isolation safe** until the identifier is the actual origin of the dapp rather than the user-supplied `snapId`.

For testing purposes please use the [snap-test-manage-accounts](https://github.com/LavaMoat/snap-test-manage-accounts) repository.

## Screenshots/Screencaps

The file was too big to attach here so I uploaded it to [snap-manage-accounts.mov](https://s3.ap-southeast-1.amazonaws.com/files.tmpfs.org/snap-manage-accounts.mov).

## Manual Testing Steps

1) Start flask `yarn start --build-type=flask`
2) Clone the [snap-test-manage-accounts](https://github.com/LavaMoat/snap-test-manage-accounts) repository.
3) Run `yarn setup`
4) Run `yarn serve` and visit `http://localhost:8080` to manually test the `snap_manageAccounts` permission and actions.

## Pre-Merge Checklist


tmpfs:
Note the only way I will be able to get CI to pass (as the test spec is bound to the `@metamask/rpc-methods` package) is to migrate the code to `@metamask/rpc-methods`, see: https://github.com/MetaMask/snaps-skunkworks/pull/549.

Once I can get that merged I should be able to address the CI problems here.

Issue #14968, opened 2022-06-16
[Bug]: Ten second rpc timeout disrupts hw connectivity


PeterYinusa:
### Describe the bug

When connecting a hw wallet, if no response if returned within 10 seconds the `select an account` screen is not displayed

https://drive.google.com/file/d/1f44Y_T37ndes3b-6C4bIqsb7SC46E59C/view?usp=sharing

### Steps to reproduce

1. Open the account menu and select `connect hardware wallet`
2. Select a hw, click continue
3. Wait 10 seconds
4. Complete the process to connect the device

### Version

10.16.0

### Browser

Chrome

### Operating system

MacOS

### Hardware wallet

Ledger, Trezor, Keystone, GridPlus Lattice1

seaona:
@PeterYinusa @segun can we close this issue now? Because of #14972 fix

PeterYinusa:
Fixed by #14972

PR #14972, opened 2022-06-16
Only have timeout for when payload is getState

### LINKED ISSUE #14970: [Bug]: Uncaught (in promise) Error: No response from RPC
### Describe the bug

Unhandled exception is thrown when taking too long sign a transaction on your hw

https://drive.google.com/file/d/18uU4mej-QR9lKYc-7NNHN2tOUTby21k4/view?usp=sharing

### Steps to reproduce

1. Connect a hw
2. Send a transaction
3. Wait 10 secs before signing it on your hw
4. Speed up the transaction
5. Again wait 10 secs before signing with your hw

```shell
ui-5.js:1 Uncaught (in promise) Error: No response from RPC
  at ui-5.js:1:67964
  at a (sentry-install.js:1:10465)
```

### Version

10.16.0

### Build type

_No response_

### Browser

Chrome

### Operating system

MacOS

### Hardware wallet

Ledger, Trezor, Keystone, GridPlus Lattice1

### Additional context

_No response_

segun:
## Explanation
To fix the `blank screen of death` on MetaMask, we added a loading screen and a timeout on the UI startup, so when MetaMask starts up, is shows a loading screen and waits for 10s for the background to startup and respond to the `getState` request, if the background doesn't respond within that timeframe, the timeout kicks in and the UI shows an error to the user.

The problem is that the timeout is in place for every request to the background and this is generally OK as the background responds within 10s, but for hardware, this is an issue because the user might not be able to complete the signing process on the hardware in 10s as in this case #14970 or the hardware connectivity itself might take longer than 10s as in this case #14968.

The fix as suggested by @danjm [here](https://consensys.slack.com/archives/C029JG63136/p1655386318719959?thread_ts=1655373624.323409&cid=C029JG63136) is to only have the timeout in place until the UI is loaded...this basically translates to having the timeout for `getState` requests only because once we have state UI is initialised and the timeout error will not be shown anyways.

The PR makes sure timeout is setup only for `getState`

## More Information

* Fixes #14970
* See: #14968

## Manual Testing Steps

The issues linked have videos and steps on how to reproduce the bug.

## Pre-Merge Checklist

### DIFF
--- a/app/scripts/lib/metaRPCClientFactory.js
+++ b/app/scripts/lib/metaRPCClientFactory.js
@@ -18,16 +18,18 @@ class MetaRPCClient {
     this.requests.set(id, cb);
     this.connectionStream.write(payload);
     this.responseHandled[id] = false;
-    setTimeout(() => {
-      if (!this.responseHandled[id] && cb) {
-        delete this.responseHandled[id];
-        return cb(new Error('No response from RPC'), null);
-      }
+    if (payload.method === 'getState') {
+      setTimeout(() => {
+        if (!this.responseHandled[id] && cb) {
+          delete this.responseHandled[id];
+          return cb(new Error('No response from RPC'), null);
+        }
 
-      delete this.responseHandled[id];
-      // needed for linter to pass
-      return true;
-    }, TEN_SECONDS_IN_MILLISECONDS);
+        delete this.responseHandled[id];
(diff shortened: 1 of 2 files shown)


darkwing (approved):
This worked great for me!

ryanml (approved):
awesome!

PR #14993, opened 2022-06-21
Destroy ledger keyring when removing last account in ledger keyring

### LINKED ISSUE #14954: [Bug]: Cannot read properties of undefined (reading 'publicKey')
### Describe the bug

An unhandled error is thrown when trying to connect a ledger device while the device is on standby

### Steps to reproduce

https://drive.google.com/file/d/1vUVYEVd2vIObtXOSOOfbDvzI5l8b4tOt/view?usp=sharing

Prerequisites:
- Default preferred ledger connection type: webhid
- Hardware device: Ledger Nano S, Ethereum App version 1.9.18, Firmware version 2.1.0
- Reduce the screen saver timeout on your ledger device to 1 min

Steps:

1. Unlock the extension
2. Connect your Ledger to your computer, unlock the device and open the Ethereum app
3. Open the `Account menu`, then click `Connect Hardware Wallet`
4. Ensure the HW is connected to your computer with the Ethereum App open
5. Select the `Ledger` option, then click `Continue`, once the HID popup appears select the device and click `Connect`
6. Select an account and click the `Unlock` button
7. On the overview page click the `Account options menu`, then click `Remove Account` followed by the `Remove` button to remove the recently added HW account from your wallet, leaving you with your original MM account
8. Wait for the device to go into standby mode (1 min: see prerequisites)
9. Open the `Account menu`, then click `Connect Hardware Wallet`
10. Select Ledger, then press Continue, once the HID popup appears click Cancel
11. Repeat this step 2 more times

```shell
TypeError: Cannot read properties of undefined (reading 'publicKey')
  at messageId(node_modules/@metamask/eth-ledger-bridge-keyring/index.js:136:1)
  at apply(node_modules/@metamask/eth-ledger-bridge-keyring/index.js:496:1)
  at ignoreOnError(node_modules/@sentry/browser/dist/helpers.js:76:1)
```

### Version

21.16.0

### Build type

_No response_

### Browser

Chrome

### Operating system

MacOS

### Hardware wallet

Ledger Nano S, Ethereum App version 1.9.18, Firmware version 2.1.0

### Additional context

_No response_

danjm:
Fixes https://github.com/MetaMask/metamask-extension/issues/14954

Whenever the user enters the connect hardware flow, `addNewKeyring` is called if no keyring for a given advice exists in the keyring controller. This will instantiate a ledger keyring, which sets up listeners (within the ledger keyring class). When removing the last account from a ledger keyring, the eth-keyring-controller will remove that keyring from its store of keyrings, but the listeners instantiated by the keyring class will not be removed. This can result in multiple listeners being created if further ledger keyrings are instantiated in the future. This will result in listeners handling events incorrectly, as each listener will match messages with handlers according to previously set ids, and ids and handlers across keyrings instances will not match correctly.

The fix is to `destroy` the ledger keyring if it no longer contains any accounts, which will result in all listeners being removed.

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -2673,8 +2673,17 @@ export default class MetamaskController extends EventEmitter {
     // Remove account from the account tracker controller
     this.accountTracker.removeAccount([address]);
 
+    const keyring = await this.keyringController.getKeyringForAccount(address);
+
     // Remove account from the keyring
     await this.keyringController.removeAccount(address);
+
+    const updatedKeyringAccounts = await keyring.getAccounts();
+
+    if (updatedKeyringAccounts.length === 0) {
+      keyring.destroy?.();
+    }
+
     return address;
   }
 
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -2674,13 +2674,10 @@ export default class MetamaskController extends EventEmitter {
     this.accountTracker.removeAccount([address]);
 
     const keyring = await this.keyringController.getKeyringForAccount(address);
-
     // Remove account from the keyring
     await this.keyringController.removeAccount(address);
-
-    const updatedKeyringAccounts = await keyring.getAccounts();
-
-    if (updatedKeyringAccounts.length === 0) {
+    const updatedKeyringAccounts = keyring ? await keyring.getAccounts() : {};
+    if (updatedKeyringAccounts?.length === 0) {
       keyring.destroy?.();
     }
 
--- a/app/scripts/metamask-controller.test.js
+++ b/app/scripts/metamask-controller.test.js
@@ -782,12 +782,20 @@ describe('MetaMaskController', function () {
   describe('#removeAccount', function () {
     let ret;
     const addressToRemove = '0x1';
+    let mockKeyring;
 
     beforeEach(async function () {
+      mockKeyring = {
+        getAccounts: sinon.stub().returns(Promise.resolve([])),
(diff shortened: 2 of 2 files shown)


jpuri (dismissed):
👍 
Broken unit test needs to be addressed.

darkwing:
Wow, excellent sleuthing Dan!  Destroying makes sense.

Looking at `removeEmptyKeyrings` in the KeyringController (https://github.com/MetaMask/KeyringController/blob/main/index.js#L365), the keyring should get removed, so I think this is really good cleanup.  Maybe in the future we move the destroy call to the KeyringController

darkwing (dismissed):

Gudahtt:
>When removing the last account from a ledger keyring, the eth-keyring-controller will remove that keyring from its store of keyrings, but the listeners instantiated by the keyring class will not be removed.

If the keyring controller is doing the _removal_, it would seem reasonable for it to handle calling `destroy` as well. Otherwise keyrings might not get destroyed if they get removed via other code paths.

jpuri (approved):

darkwing (approved):

seaona:
No sure if it's totally related but I've noticed that when we remove a Ledger account and go back to the Import Hardware Wallet screen we get this error `TrezorConnect has been already initialized`, which seems totally strange as no action related to Trezor has been performed. Could it be that there is some action triggering Trezor while removing Ledger accounts and destroying ledger keyring? @danjm @Gudahtt 

See steps and further explanation of this behavior here:
https://github.com/MetaMask/metamask-extension/issues/15007

PR #15011, opened 2022-06-22
Show how much a cancellation of a smart transaction will cost

dan437:
## Explanation
Before on the STX status page we were showing a link `Cancel swap`. However, users had no idea how much the cancellation would cost. Now we show if it's either for free or for a fee.

## Screenshots/Screencaps
Before:
image

After (0 fee):
image

After (fee > 0):
image

## Manual Testing Steps
- Do a swap with Smart Transactions enabled. On the STX status page you can see the new, dynamic cancellation link


darkwing (approved):

digiwand (approved):
Code lgtm! 

note: this code implies that there is no way for the cancellationFeeWei to be missing if the user will be charged a Swap cancellation fee.

PR #15016, opened 2022-06-22
Fix tokenIds larger than MAX_SAFE_INTEGER converted to scientific notation and failing to import

### LINKED ISSUE #13581: tokenIds larger than Number.MAX_SAFE_INTEGER getting translated into scientific notation
TokenIds input by a user in the import NFT form need to be handled using `BigInt` or `bn.js`, because they can be higher than `Number.MAX_SAFE_INTEGER` and imports with these high `tokenId`s will currently fail since the numbers are not handled properly.

adonesky1:
Fixes #13581

Currently the `AddCollectibles` component used to manually import NFTs uses a numeric input for the `tokenID` field. The problem is that when a user inputs a tokenId larger than `MAX_SAFE_INTEGER`: 

Screen Shot 2022-06-22 at 5 08 56 PM

the numeric input automatically converts it to scientific notation. This PR changes the `tokenID` field to a normal text input field which can then pass along the input `tokenID` as a string, avoiding any transformation issues. 

Also adds validation that prevents users from submitting the add attempt if the tokenIds is not a number.

Adds tests to verify a few critical paths for the `AddCollectibles` component flow.

### DIFF
--- a/ui/pages/add-collectible/add-collectible.js
+++ b/ui/pages/add-collectible/add-collectible.js
@@ -4,7 +4,6 @@ import { useDispatch, useSelector } from 'react-redux';
 import { util } from '@metamask/controllers';
 import { useI18nContext } from '../../hooks/useI18nContext';
 import { DEFAULT_ROUTE } from '../../helpers/constants/routes';
-
 import {
   DISPLAY,
   FONT_WEIGHT,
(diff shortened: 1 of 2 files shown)


darkwing (approved):

segun (approved):
LGTM

PR #15034, opened 2022-06-23
Add fallback image/card for NFTs when image was not fetched correctly or does not exist

adonesky1:
Adds a fallback card when an NFT image is either not fetched correctly or does not exist. 

Figma designs [here](https://www.figma.com/file/8Xe22jEPgcElS5tWPw8AGM/NFTs-in-extension-(MVP)?node-id=1%3A3).

Before:
Screen Shot 2022-06-23 at 9 37 32 AM

After:
Screen Shot 2022-06-23 at 9 27 29 AM


ui/components/app/collectibles-items/collectibles-items.js
```diff
@@ -164,7 +167,9 @@ export default function CollectiblesItems({
                 variant={TYPOGRAPHY.H5}
                 margin={[0, 0, 0, 2]}
               >
-                {`${collectionName} (${collectibles.length})`}
+                {`${collectionName ?? t('unknownCollection')} (${
```
adonesky1, line 170: I'm using `Unknown Collection` as fallback when a `collectionName` either does not exist or is not fetched correctly. @rachelcope @SayaGT Would you prefer a different fallback here?

ui/components/app/collectible-default-image/collectible-default-image.js
```diff
@@ -0,0 +1,39 @@
...
+}) {
+  const t = useI18nContext();
+  return (
+    <div
+      className={classnames('collectible-default__wrapper', {
+        'collectible-default__clickable': handleImageClick,
+      })}
+      onClick={handleImageClick}
+    >
+      <img className="collectible-default__image" src="images/empty-nft.png" />
```
georgewrmarshall, line 21: Instead of adding an image could we use a background color instead? That should save us a few Bytes 🔍  Could we add `background-color: var(--color-background-alternative);` to `.collectible-default__wrapper`?

georgewrmarshall (commented):
Great update! Will add a more in-depth review start of next week 🤙

ui/components/app/collectible-default-image/collectible-default-image.js
adonesky1: I tried this at first but had trouble keeping the div retain the correct size/aspect ratio across different breakpoints without using some cheap padding tricks. But I'm not a CSS wizard so I'm sure you know better how to achieve this without using an img!

georgewrmarshall (commented):
One tiny little update. You're gonna sigh at it lol it's incredibly subtle 😅 

Otherwise looks great!! Here is how it looks with my big ass ENS ID

Screen Shot 2022-06-29 at 9 23 48 AM
Screen Shot 2022-06-29 at 9 23 58 AM

ui/components/app/collectible-default-image/collectible-default-image.js
georgewrmarshall: Addressed https://github.com/MetaMask/metamask-extension/pull/15034/commits/a83bcd39ea3373d179051116dd9ee5009fde5788

georgewrmarshall (dismissed):
LGTM!

ui/components/app/collectibles-items/index.scss
```diff
@@ -27,29 +27,29 @@
...
       text-align: center;
     }
+  }
 
-    &-item-wrapper {
-      align-self: center;
-    }
+  &__item-wrapper {
+    align-self: center;
+  }
```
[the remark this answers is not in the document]
georgewrmarshall: Resolved in https://github.com/MetaMask/metamask-extension/pull/15034/commits/53cc18baab65ba5e3c7896a747f18e6484aa9742

ui/components/app/collectibles-items/collectibles-items.js
georgewrmarshall: Resolved in https://github.com/MetaMask/metamask-extension/pull/15034/commits/0db81311e9ff31b96f69750747e46e9c8bc09958

Screen Shot 2022-06-29 at 12 11 57 PM

ryanml (approved):
++

georgewrmarshall (approved):

PR #15063, opened 2022-06-28
Revert v10.16.0

Gudahtt:
This reverts commit f09ab8889148c406551dea1643966e3331fde4aa, reversing changes made to effc761e0ee4ea7ffb77f275b5ed650a7098d6f8.

This is being temporarily reverted to make it easier to release an urgent fix for v10.15.1.

PR #15064, opened 2022-06-28
Ensure that addresses are properly hex-prefixed in the generate ERC token transfer functions

### LINKED ISSUE #15058: [Bug]: Metamask sending TX to random wallet addresses - repeatable bug
### Describe the bug

I was deposing funds to a Coinbase deposit address I've used in the past. Coinbase deposit page only shows 64 digits (eg doesn't include the 0x). In the past, Metamask has just added it when attempting to send funds to a 64 character ethereum address, but some recent change broke that and it doesn't work every time now.

My funds were sent to a random wallet instead of the one I entered. I spoke with Metamask support through email who initially thought my computer had a "clipboard virus" but after investigating all day and testing other machines / wallets / having friends test they were all able to replicate the issue.

This is a serious problem and it's so easy to replicate that anyone could have run into it

### Steps to reproduce

1. initiate a transaction to any wallet address without the 0x at the start (eg 64 characters instead of 66) - here's vitalik's for example. Ab5801a7D398351b8bE11C439e05C5B3259aeC9B
2. Send USDC, USDT, YFI or WETH
3. Metamask converts the address to 0x00000020CA5cA4F9236e618f7c3A8bD45b59eB45 instead of adding the 0x

If you attempt to do this with ETH or most other tokens, it adds the 0x. But for some reason when sending USDT, USDC, or WETH it converts it to something else. It most likely applies to other tokens as well, but I didn't have anything else in my wallet so I couldn't test it

I've had many other people test this and they all get the same issue

_No response_

### Version

10.15.1

### Build type

_No response_

### Browser

### Operating system

Windows, MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

danjm:
Fixes https://github.com/MetaMask/metamask-extension/issues/15058

This PR is currently comparing to the branch for https://github.com/MetaMask/metamask-extension/pull/15063, we need to merge that PR first, and then rebase this PR's branch onto master.

we used to do this to create a token transfer transaction from the send screen:
https://github.com/MetaMask/metamask-extension/pull/14465/files#diff-bad76d7279418f3520a0ca93cfaeeb6fcbb1f0d4782f6ff942d422632a1c2780L1762-L1766
```
        token.transfer(address, value, {
          ...txParams,
          to: undefined,
          data: undefined,
        });
```

we now do this:
https://github.com/MetaMask/metamask-extension/pull/14465/files#diff-3c58b916cefff81e4927a9afd392acadbf4515f99f41783964a9851b7774791bR815-R833
```
const txMeta = await promisifiedBackground.addUnapprovedTransaction(
        txParams,
        ORIGIN_METAMASK,
        type,
      );
    });
```

previously we would create the token transfer via libraries that depend on this from ethjs-abi https://github.com/ethjs/ethjs-abi/blob/master/src/utils/index.js#L166, so validation of the hex string would fail there, and the user wouldn't be able to send the transaction, with the change referenced above, that validation stopped

### DIFF
--- a/ui/ducks/send/send.test.js
+++ b/ui/ducks/send/send.test.js
@@ -88,11 +88,16 @@ setBackgroundConnection({
 
 describe('Send Slice', () => {
   let getTokenStandardAndDetailsStub;
+  let addUnapprovedTransactionAndRouteToConfirmationPageStub;
   beforeEach(() => {
     jest.useFakeTimers();
     getTokenStandardAndDetailsStub = jest
       .spyOn(Actions, 'getTokenStandardAndDetails')
       .mockImplementation(() => Promise.resolve({ standard: 'ERC20' }));
+    addUnapprovedTransactionAndRouteToConfirmationPageStub = jest.spyOn(
+      Actions,
+      'addUnapprovedTransactionAndRouteToConfirmationPage',
+    );
     jest
       .spyOn(Actions, 'estimateGas')
       .mockImplementation(() => Promise.resolve('0x0'));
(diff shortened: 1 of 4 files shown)


Gudahtt (dismissed):
LGTM!

We should follow up with additional unit tests, an e2e test, and inline docs for these functions.

Gudahtt:
Actually, maybe better to target this at develop instead of master.

The unit test failure appears unrelated, it's one that intermittently fails.

Gudahtt (dismissed):

brad-decker (dismissed):

adonesky1 (dismissed):
LGTM!

darkwing:
Needs unit test update, but I sent USDC on MainNet to an address without the `0x` prefix and it made it to the appropriate address.

adonesky1 (approved):

Gudahtt (approved):

PR #15077, opened 2022-06-29
Fixes double account creation when using keyboard controls

### LINKED ISSUE #15061: [Bug]: Create account using keyboard generates 2 accounts instead of 1
### Describe the bug

**Problem**: when I use the keyboard to create an account, I see that 2 accounts are created instead of 1.

https://user-images.githubusercontent.com/54408225/176143394-21e14cea-5b53-4c54-903b-d5703fd47dae.mp4

### Steps to reproduce

1. Unlock MetaMask
2. Click icon menu
3. Click Create Account
4. Insert account name
5. Click TAB 2 times (so Create button is selected)
6. Click Enter -- see that this has generated 2 accounts

_No response_

### Version

10.17.0

### Build type

_No response_

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

ryanml:
Fixes: https://github.com/MetaMask/metamask-extension/issues/15061

## Explanation

Defined in issue

## Manual Testing Steps
Follow the steps mentioned in the issue, ensure that only a single account is created when using keyboard shortcuts


Atandanny (approved):

digiwand (approved):
👍🏼

darkwing (changes requested):
I might be crazy but pressing `[ENTER]` doesn't submit the form; we may need to event type check here

digiwand:
@darkwing What browser were you using? To confirm, was the target element the "Confirm" button after pressing tab to get to the button? 

This fix is working on my end on Chrome. I was able to see the double accounts being created before the fix, and then a single account being created with this fix.

digiwand:
This fix is working on my end on Chrome. I was able to see the double accounts being created before the fix, and then a single account being created with this fix.

@darkwing what browser were you using? To confirm, was the target element the "Confirm" button after pressing tab to get to the button? I can see this could be easily mistaken since the focus state could be mistaken for the default state

darkwing:
Just tested on Chrome + OSX.  If I click the "Create" button, the account gets created.  If click "Create account" and immediately press ENTER, nothing happens.

digiwand:
Echoing what @jpuri was mentioning, this PR tests navigating user flow with a keyboard and not a mouse.

@darkwing 
> Just tested on Chrome + OSX. If I click the "Create" button, the account gets created. If click "Create account" and immediately press ENTER, nothing happens.

This is existing behavior though would be nice to add form submit on error. After clicking "Create account", the input is in focus. This PR is testing when we hit enter when the "Create" button is in focus, and not the input in focus. You will need to tab twice to focus on the create button from the input, and then click enter to trigger the action

ryanml:
@darkwing, would you be fine with a separate PR for adding submit on 'Enter' behavior?

darkwing (approved):

PR #15082, opened 2022-06-29
Fix Provider Tracking Metrics

brad-decker:
## Explanation
in #14269 we started tracking provider method usage. When this went out to production we realized we were tracking way too many events. This PR does the following:
1. Prevents tracking RPC usage of our own extension
2. Prevents tracking RPC usage of the inpage provider init
3. Increases rate limiting from 10 seconds to 60 seconds
4. Moves permission requests into the rate limited category

## QA Steps

Do the following on develop so you can see the problem: 

1. Either set your .metamaskrc up to use segment dev environment, or run a local metrics server and point segment at it. 
2. Make sure you have opted into metrics
3. Inspect the background view, open the segment debugger and/or look at the console of the local metrics server
4. Load any page in your browser after reloading / unlocking extension
5. See tracks for method 'metamask_sendDomainMetadata' 
6. Go to the test dapp, load the page and see upwards of 6 tracks on page load.
7. Open the extension UI and start a send flow and see tracks with origin 'metamask'

then do the above again, notice now on page loads that are not dapps there are no events. Also see that loading the test dapp only tracks 2-3 events and we no longer track our own (origin metamask) usage of the provider when in the UI.


shared/constants/metametrics.js
```diff
@@ -276,9 +276,18 @@ export const REJECT_NOTFICIATION_CLOSE_SIG =
  */
 
 export const EVENT_NAMES = {
+  ENCRYPTION_PUBLIC_KEY_APPROVED: 'Encryption Public Key Approved',
+  ENCRYPTION_PUBLIC_KEY_REJECTED: 'Encryption Public Key Rejected',
   ENCRYPTION_PUBLIC_KEY_REQUESTED: 'Encryption Public Key Requested',
+  DECRYPTION_APPROVED: 'Decryption Approved',
+  DECRYPTION_REJECTED: 'Decryption Requested',
```
worldlyjohn, line 283: @brad-decker this should read `DECRYPTION_REJECTED: 'Decryption Rejected`' not `Requested`

app/scripts/lib/createRPCMethodTrackingMiddleware.js
```diff
@@ -1,85 +1,205 @@
...
+      getMetricsState().participateInMetaMetrics === true;
+
+    // Get the event type, each of which has APPROVED, REJECTED and REQUESTED
+    // keys for the various events in the flow.
+    const eventType = EVENT_NAME_MAP[method];
+
+    // Boolean variable that reduces code duplication and increases legibility
+    const shouldTrackEvent =
+      // Don't track if the request came from our own UI or background
+      origin !== 'metamask' &&
```
darkwing, line 138: Should we use `ORIGIN_METAMASK` here?

shared/constants/metametrics.js
brad-decker: Done in latest :)

darkwing (approved):

ryanml (approved):
lgtm!

PR #14171, opened 2022-03-23
Unable to download state logs from Settings > Advanced. Fix using getPlatformInfo browser polyfill

digiwand:
## Explanation

Unable to download state logs from Settings > Advanced. This PR updates `getPlatformInfo` the method to use the polyfill version

## More information

Slack: https://consensys.slack.com/archives/GTQAGKY5V/p1648069151587819
Related PR: https://github.com/MetaMask/metamask-extension/pull/13960/files

## Screenshots/Screencaps

Screen Shot 2022-03-23 at 4 52 08 PM
Screenshot 2022-03-23 at 7 05 28 PM

## Manual testing steps

- Go to Settings > Advanced
- Click "Download State Logs"

PR #15015, opened 2022-06-22
Chromedriver v103

PeterYinusa:
## Explanation

Update ChromeDriver to v103
Install Chrome stable channel update for desktop v103.0.5060.53

## Pre-Merge Checklist

- [ ] PR template is filled out
- [ ] **IF** this PR fixes a bug, a test that _would have_ caught the bug has been added
- [ ] PR is linked to the appropriate GitHub issue
- [ ] PR has been added to the appropriate release Milestone

### + If there are functional changes:

- [ ] Manual testing complete & passed
- [ ] "Extension QA Board" label has been applied

PR #15123, opened 2022-07-01
State logs e2e

PeterYinusa:
## Explanation

This PR adds an e2e test to ensure that the state logs can be downloaded from Settings -> Advanced

## More Information

See: https://github.com/MetaMask/metamask-extension/pull/14171

## Screenshot

Screenshot showing the downloaded file in the CI pipeline

Chrome
State Logs

Firefox
State Logs FF

### DIFF
--- a/.circleci/scripts/firefox-install.sh
+++ b/.circleci/scripts/firefox-install.sh
@@ -4,7 +4,7 @@ set -e
 set -u
 set -o pipefail
 
-FIREFOX_VERSION='83.0'
+FIREFOX_VERSION='102.0'
 FIREFOX_BINARY="firefox-${FIREFOX_VERSION}.tar.bz2"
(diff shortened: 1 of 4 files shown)


.circleci/scripts/firefox-install.sh
```diff
@@ -4,7 +4,7 @@ set -e
 set -u
 set -o pipefail
 
-FIREFOX_VERSION='83.0'
```
darkwing, line 7: Are we sure we want to update this value?  Shouldn't we keep this value at our min supported version ?

PeterYinusa: Yes this was intentional, our min supported version is 68 for FF.
I updated this to be the latest FF release as most users are on the latest version(s). 
I also needed to update the browser to be able to control the download directory of FF.
I've recently updated Chrome to the latest stable version #15015

darkwing (approved):

jpuri (approved):

PR #15138, opened 2022-07-05
Token detection V2 Flag Removal and Re-introducing the use of legacy token list when token detection is OFF

NiranjanaBinoy:
Changes addressed in this PR:

- Removing TOKEN_DETECTION_V2 flag

- Reverting the `token Detection to OFF by default` to avoid the third-party token fetch on application loading.

- Updated the copy/locale message in the token detection toggle description in the Advanced settings tab.

- In MAINNET, when the token detection is turned OFF, the user will be able to search for the token in the Legacy token list (token data from the contract-metadata repo); automatic token detection will also be performed in Mainnet on the legacy token list in this scenario.

- For all the other networks (even those that support the dynamic toke list), search and token detection will be disabled when token detection is turned OFF.

- Adding a different message in the popover when only a few selected tokens are imported and the rest are ignored.

Screen Shot 2022-07-05 at 1 27 52 PM


darkwing:
Test failure looks legit.

NiranjanaBinoy:
> Test failure looks legit.

This might be because the related controller changes are yet to be published and added to the extension.

app/scripts/metamask-controller.js
```diff
@@ -2239,13 +2236,8 @@ export default class MetamaskController extends EventEmitter {
...
         ].filter((asset) => {
           if (asset.isERC721 === undefined) {
-            // since the token.address from allTokens is checksumaddress
-            // asset.address have to be changed to lowercase when we are using dynamic list
-            const address = useTokenDetection
-              ? asset.address.toLowerCase()
-              : asset.address;
             // the tokenList will be holding only erc20 tokens
-            if (tokenList[address] !== undefined) {
+            if (tokenList[asset.address.toLowerCase()] !== undefined) {
```
danjm, line 2240: Is it ever possible for asset.address to be undefined?

NiranjanaBinoy: Chances are rare, but to be on the safe side, I will add a conditional check; I am also updating the tokenList to use the static list as well, which I missed earlier.

liteflows (dismissed):

adonesky1:
@NiranjanaBinoy looks like this needs a rebase

app/scripts/controllers/detect-tokens.js
```diff
@@ -50,14 +49,15 @@ export default class DetectTokensController {
...
     this.selectedAddress = this.preferences?.store.getState().selectedAddress;
     this.tokenAddresses = this.tokensController?.state.tokens.map((token) => {
       return token.address;
     });
     this.hiddenTokens = this.tokensController?.state.ignoredTokens;
-    this.detectedTokens = process.env.TOKEN_DETECTION_V2
-      ? this.tokensController?.state.detectedTokens
-      : [];
+    this.detectedTokens = this.tokensController?.state.detectedTokens;
+    this.chainId = network?.store.getState().provider.chainId;
```
danjm, line 60: I see there is an optional chaining operator on `network` here. Should we also have it on all references to `this._network`?

danjm: There are multiple places in this file where we try to get chainId by doing this `network?.store.getState().provider.chainId`

I'd suggest adding a private method, or a utility function:

```
function getChainIdFromNetworkController(network) {
    return network?.store?.getState().provider?.chainId
}
```
And use that throughout the file

ui/components/ui/jazzicon/jazzicon.component.js
```diff
@@ -48,7 +48,11 @@ export default class Jazzicon extends PureComponent {
 
   appendJazzicon() {
     const { address, diameter, tokenList } = this.props;
-    const image = iconFactory.iconForAddress(address, diameter, tokenList);
+    const image = iconFactory.iconForAddress(
+      address,
+      diameter,
+      tokenList[address.toLowerCase()],
```
danjm, line 54: I don't think this change is needed anymore right? Now that the static token list has been modified to match the structure of the remote token list, I think the iconFactory can remained unchanged?

ui/components/ui/nickname-popover/nickname-popover.component.js
```diff
@@ -54,8 +48,7 @@ const NicknamePopover = ({
           address={address}
           diameter={36}
           className="nickname-popover__identicon"
-          useTokenDetection={useTokenDetection}
-          tokenList={tokenList}
+          image={tokenList[address.toLowerCase()]?.iconUrl}
```
danjm, line 51: Do we need to do this here anymore? Now that the static token list matches the data structure of the dynamic list, can we just  pass the token list to the identicon?

ui/selectors/selectors.js
```diff
@@ -828,8 +828,8 @@ function getAllowedAnnouncementIds(state) {
     7: false,
     8: supportsWebHid && currentKeyringIsLedger && currentlyUsingLedgerLive,
     9: false,
-    10: Boolean(process.env.TOKEN_DETECTION_V2) && !process.env.IN_TEST,
-    11: Boolean(process.env.TOKEN_DETECTION_V2) && !process.env.IN_TEST,
+    10: true,
```
danjm, line 831: Why are there two separate notifications for token detection?

ui/components/ui/jazzicon/jazzicon.component.js
NiranjanaBinoy: Previously we were sending in the whole tokenList as part of Token Detection V1;  with this change, we are just passing in the metadata for the specified token.

ui/selectors/selectors.js
NiranjanaBinoy: There is a `Scam and Security risk` announcement and `Improved token Detection` announcement.
Screen Shot 2022-08-09 at 11 06 55 AM

NiranjanaBinoy: As I was reading through this, I realized that we might want to update the description for the Token detection announcement in extension as we are not keeping token detection ON by default, and token detection with static token still works in MAINNET when token detection toggle is turned OFF.

ui/components/ui/jazzicon/jazzicon.component.js
danjm: Okay, that makes sense. This does seem like an improvement

danjm (dismissed):

brad-decker (dismissed):

brad-decker (approved):

danjm (approved):

ryanml (approved):

PR #15154, opened 2022-07-07
Bumps eth-lattice-keyring to v0.10.0

douglance:
## Explanation

GridPlus has shipped a new version of firmware for the Lattice1. This new version of the firmware includes support for automatic decoding of ABI data, which is then displayed for users on their device.

In order to accomplish this for MetaMask users, we will need to bump the version of our eth-lattice-keyring. This [updated version of the eth-lattice-keyring](https://github.com/GridPlus/eth-lattice-keyring) uses [a function of the gridplus-sdk](https://github.com/GridPlus/gridplus-sdk/blob/v2.1.0/src/util.ts?rgh-link-date=2022-06-23T16%3A35%3A51Z#L455=) that fetches ABI data and bundles it with requests to the Lattice1.


Anastasia1510 (approved):

douglance:
Is there anything I need to do to get the circleci tests to run? Does someone need to approve my account to run those? 

Thanks!

PR #15175, opened 2022-07-11
Modify transaction type and add contract method data to metrics

### LINKED ISSUE #15150: Modify transaction type, and add contract method, data to metrics
## Acceptance Criteria

The `type` property for `Transaction Approved`, `Transaction Rejected`, `Transaction Finalized` and the `Transaction Submitted` events are now replaced by the below defined `transaction_type` property with the following values and additional properties:

**`transaction_type`** (string): *send, contract_interaction, cancel*

- if *transaction_type* = *contract_interaction*:
    
    **`transaction_contract_method`** (string): *name_of_contract_method_here*

VSaric:
## Explanation

Currently, on extension we track token allowances within `type` property of transactions events in MetaMetrics. But we don't know much about it other than the fact that the transaction was a token allowance.

We're updating the `type` property name on transactions events to `transaction_type` and limiting the possible values to `send`, `contract_interaction` and `cancel`. Additionally, we're also adding a new property called `transaction_contract_method` where we will track the contract method that is being called so we can continuously improve the user experience for the most common ones, specifically for token allowances.

[Here](https://github.com/MetaMask/metamask-extension/issues/15150#issuecomment-1177533555) we have some notes on technical implementation. I added new property `contractMethodName` in `TransactionMeta` to represent a name of transaction contract method. I noticed that the `name` property in `methodData` object in `ConfirmTransactionBase` component is a contract method name. I passed it to the transaction controller from `handleSubmit` method.

In transaction controller, I get `transactions` object from `this.store.getState` in `_buildEventFragmentProperties` and if the `transaction_type` is a `contract_interaction`, we set `transaction_contract_method` from corresponding transaction created by the user (`transactions[id]?.contractMethodName`). If the `transaction_type` is not a `contract_interaction` then we set `transaction_contract_method` to be `undefined`.

## More Information

* Fixes #15150


VSaric:
I saw your question @brad-decker, we will wait until @danjm answers

brad-decker (approved):

NiranjanaBinoy (dismissed):

brad-decker (changes requested):
req changing to block pr merge

app/scripts/controllers/transactions/index.js
```diff
@@ -2078,19 +2080,41 @@ export default class TransactionController extends EventEmitter {
...
+    } else if (type === TRANSACTION_TYPES.SIMPLE_SEND) {
+      transactionType = TRANSACTION_TYPES.SIMPLE_SEND;
+    } else if (contractInteractionTypes) {
+      transactionType = TRANSACTION_TYPES.CONTRACT_INTERACTION;
+    }
+
+    let transactionContractMethod;
+    if (transactionType === TRANSACTION_TYPES.CONTRACT_INTERACTION) {
+      transactionContractMethod = transactions[id]?.contractMethodName;
+    }
```
brad-decker, line 2103: @darkwing pointed out to me that we aren't caring for all types here, and transactionType will be undefined in the following cases with this code:

SWAP
SWAP_APPROVAL
SMART
TOKEN_METHOD_SET_APPROVAL_FOR_ALL
RETRY
DEPLOY_CONTRACT

are these meant to be stripped from metrics? @danjm @bschorchit

brad-decker (changes requested):

app/scripts/controllers/transactions/index.js
brad-decker: I know there is a separate PR for the speed up, but the way this code is wirtten if a retry comes through in this state it'll have no 'transaction_type'

bschorchit: Thaanks Brad and David for catching that! Here's how we're addressing it:

- SWAP, SWAP_APPROVAL, SMART, TOKEN_METHOD_SET_APPROVAL_FOR_ALL will now have transaction_type = contract_interaction

- DEPLOY_CONTRACT will now have transaction_type = contract_deployment

- RETRY will have transaction_type = the original transaction type which we're doing in this separate PR [here](https://github.com/MetaMask/metamask-extension/pull/15197)

VSaric: Thank you @brad-decker @darkwing for catching that! I made changes [here](https://github.com/MetaMask/metamask-extension/pull/15175/commits/d4c7e4d5eda07230006b494dc6ef92a38ba34970) and added missing transaction types like @bschorchit mentioned above.

app/scripts/controllers/transactions/index.js
```diff
@@ -2081,19 +2083,47 @@ export default class TransactionController extends EventEmitter {
...
+    let transactionType;
+    if (type === TRANSACTION_TYPES.CANCEL) {
+      transactionType = TRANSACTION_TYPES.CANCEL;
+    } else if (type === TRANSACTION_TYPES.SIMPLE_SEND) {
+      transactionType = TRANSACTION_TYPES.SIMPLE_SEND;
+    } else if (type === TRANSACTION_TYPES.DEPLOY_CONTRACT) {
+      transactionType = TRANSACTION_TYPES.DEPLOY_CONTRACT;
+    } else if (contractInteractionTypes) {
+      transactionType = TRANSACTION_TYPES.CONTRACT_INTERACTION;
+    }
```
brad-decker, line 2107:
```suggestion
    let transactionType = TRANSACTION_TYPES.SIMPLE_SEND;
    if (type === TRANSACTION_TYPES.CANCEL) {
      transactionType = TRANSACTION_TYPES.CANCEL;
    } else if (type === TRANSACTION_TYPES.DEPLOY_CONTRACT) {
      transactionType = TRANSACTION_TYPES.DEPLOY_CONTRACT;
    } else if (contractInteractionTypes) {
      transactionType = TRANSACTION_TYPES.CONTRACT_INTERACTION;
    }
```


This prevents us from having potentially undefined types, however with the way the code currently works and without the originalType yet being recorded this will result in retry being labeld as "Simple Send" I suggest that we first keep RETRY as a type then remove it in the PR that is handling original type.

brad-decker (changes requested):

app/scripts/controllers/transactions/index.js
VSaric: @brad-decker So you are saying that we first keep RETRY as a `transaction_type` in this PR and than when this get merged to remove it in the PR that is handling original type? I need to add new if case:

>  if (type === TRANSACTION_TYPES.RETRY) {
>     transactionType = TRANSACTION_TYPES.RETRY;
> }

VSaric: @brad-decker [Here](https://github.com/MetaMask/metamask-extension/pull/15175/commits/ebeb2392fc54ee7274ba0bb161473f067be67d33) I made this requested changes.

brad-decker: We should also perserve the retry type until the retry work lands.

bschorchit: Let's have RETRY have transaction_type = retry and ask @adnansahovic to include the respective change to his PR. cc: @danjm

brad-decker (approved):

app/scripts/controllers/transactions/index.js
```diff
@@ -2081,19 +2083,47 @@ export default class TransactionController extends EventEmitter {
...
+
+    let transactionType = TRANSACTION_TYPES.SIMPLE_SEND;
+    if (type === TRANSACTION_TYPES.CANCEL) {
+      transactionType = TRANSACTION_TYPES.CANCEL;
+    } else if (type === TRANSACTION_TYPES.RETRY) {
+      transactionType = TRANSACTION_TYPES.RETRY;
+    } else if (type === TRANSACTION_TYPES.DEPLOY_CONTRACT) {
+      transactionType = TRANSACTION_TYPES.DEPLOY_CONTRACT;
+    } else if (contractInteractionTypes) {
+      transactionType = TRANSACTION_TYPES.CONTRACT_INTERACTION;
```
NiranjanaBinoy, line 2106: Why can't we assign `transactionContractMethod = transactions[id]?.contractMethodName;` inside this `else if` block?

darkwing (approved):

app/scripts/controllers/transactions/index.js
brad-decker: Good suggestion, @NiranjanaBinoy.

brad-decker: https://github.com/MetaMask/metamask-extension/pull/15329 @NiranjanaBinoy made change after merge

VSaric: Good suggestion @NiranjanaBinoy! Thank you for reviewing this! 
Thank you @brad-decker for creating new PR to modify this! It is very cleaner code now with this additional change.

PR #15186, opened 2022-07-11
Disable Seedphrase import button if any of the characters is in uppercase

NiranjanaBinoy:
Part of https://github.com/MetaMask/metamask-extension/issues/14657

The main reason behind the `Seed phrase invalid` error seems to be that some cases accepted by `isValidMnemonic` from ethers.utils in srp-inpiut.js are rejected when the SRP is revalidated at bip39.validateMnemonic in the createNewVaultandRestore method.
One of the scenarios is covered in #15139 

Second scenario: 
As mentioned in the comment https://github.com/MetaMask/metamask-extension/issues/14657#issuecomment-1158845402 by @seaona, when a user enters an uppercase letter by mistake it will be accepted by `isValidMnemonic` in srp-input but reject at `validateMnemonic` during revalidation.

Solution:
To enable the import button only if all the words on the SRP are in lower case.

Test steps:

2. Click Import Wallet
3. Accept/Decline Metametrics
4. Type-in/copy-paste a valid seedphrase and then change any of the letters to uppercase
5.  Verify that the Import button is disabled with the error message `Invalid Secret Recovery Phrase` showing (with and without entering the password and checking the terms & conditions checkbox).

### DIFF
--- a/ui/components/app/srp-input/srp-input.js
+++ b/ui/components/app/srp-input/srp-input.js
@@ -44,10 +44,9 @@ export default function SrpInput({ onChange, srpText }) {
       if (newDraftSrp.some((word) => word !== '')) {
         if (newDraftSrp.some((word) => word === '')) {
           newSrpError = t('seedPhraseReq');
-        } else if (
-          isLowerCase(joinedDraftSrp) ||
-          !isValidMnemonic(joinedDraftSrp)
-        ) {
+        } else if (isLowerCase(joinedDraftSrp)) {
+          newSrpError = t('invalidSeedPhraseCaseSensitive');
+        } else if (!isValidMnemonic(joinedDraftSrp)) {
           newSrpError = t('invalidSeedPhrase');
         }
       }
--- a/ui/components/app/srp-input/srp-input.js
+++ b/ui/components/app/srp-input/srp-input.js
@@ -19,6 +19,10 @@ const { isValidMnemonic } = ethers.utils;
 
 const defaultNumberOfWords = 12;
 
+const isLowerCase = (draftSrp) => {
+  return draftSrp !== draftSrp.toLowerCase();
+};
+
 export default function SrpInput({ onChange, srpText }) {
   const [srpError, setSrpError] = useState('');
   const [pasteFailed, setPasteFailed] = useState(false);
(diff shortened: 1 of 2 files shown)


ui/components/app/srp-input/srp-input.js
```diff
@@ -40,7 +44,10 @@ export default function SrpInput({ onChange, srpText }) {
       if (newDraftSrp.some((word) => word !== '')) {
         if (newDraftSrp.some((word) => word === '')) {
           newSrpError = t('seedPhraseReq');
-        } else if (!isValidMnemonic(joinedDraftSrp)) {
+        } else if (
+          isLowerCase(joinedDraftSrp) ||
```
darkwing, line 48: Do we want to provide a different error message for cases where the user inputs an uppercase word?  Is "Invalid seed phrase" too generic?

NiranjanaBinoy: Yes, it seems a bit too generic; @SayaGT, Can you please help us with a new appropriate message?

SayaGT: @NiranjanaBinoy We could say something like 
1. Invalid input! Secret Recovery Phrase is case sensitive.  OR
2. Input is case sensitive. OR
3. Secret Recovery Phrase is case sensitive.

NiranjanaBinoy: @danjm and @darkwing Shall we proceed with `Invalid input! Secret Recovery Phrase is case sensitive.` as @SayaGT suggested ?

ui/components/app/srp-input/srp-input.js
```diff
@@ -19,6 +19,10 @@ const { isValidMnemonic } = ethers.utils;
 
 const defaultNumberOfWords = 12;
 
+const isLowerCase = (draftSrp) => {
```
darkwing, line 22: Sorry for not noticing the first time but should this be "hasUpperCase"?  I think `isLowerCase` would be a good thing in this case.

NiranjanaBinoy: done

darkwing (approved):

seaona:
From QA point of view, fix is working correctly. Now button is disabled when there is an uppercase on the seed-phrase. @NiranjanaBinoy

brad-decker (approved):

PR #15190, opened 2022-07-12
[FLASK] Fix snaps authorship pill

### LINKED ISSUE #15161: Truncate snaps authorship pill
Long snap package names are not truncated and as a result can offset the UI.

hmalik88:
Fixes #15161 

An arbitrary `max-width` was chosen to enforce a limit at which a snap's package name is truncated in the pill. A title was added so that a user can still hover and view the complete package name.

Steps to reproduce:

1. Run a flask build using `yarn start --build-type flask` and add it to chrome.
2. Go to `snaplist.org` to test out the install flow.
3. Once you get to the install screen, you can inspect the page and change the package name in the DOM to a length that would typically offset the UI.
4. Observe that the pill truncates the package name.

Screenshot:
Screen Shot 2022-07-12 at 12 05 31 AM

### DIFF
--- a/ui/components/app/flask/snaps-authorship-pill/index.scss
+++ b/ui/components/app/flask/snaps-authorship-pill/index.scss
@@ -1,14 +1,18 @@
 @import "design-system";
 
 .snaps-authorship-pill {
-  display: inline-block;
+  width: 100%;
+  display: inline-flex;
+  justify-content: center;
 
   .chip {
     padding-right: 8px;
+    max-width: 100%;
+    margin-left: 0;
+    margin-right: 0;
   }
 
   .chip__label {
-    max-width: 136px;
     text-overflow: ellipsis;
     overflow: hidden;
   }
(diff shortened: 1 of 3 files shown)


GuillaumeRx (dismissed):

FrederikBolding:
@hmalik88 I tried something to remove the hardcoded `max-width`. We can remove if you don't like it, but I would probably prefer this.

cc @GuillaumeRx

GuillaumeRx:
> @hmalik88 I tried something to remove the hardcoded `max-width`. We can remove if you don't like it, but I would probably prefer this.
> 
> cc @GuillaumeRx

I prefer @FrederikBolding implementation. It lets the parent `div` constrain the pill at the desired width improving the responsiveness of the pill.

FrederikBolding (approved):

GuillaumeRx (approved):

PR #15193, opened 2022-07-12
Removed the Infura v3 key from rpcUrl display

### LINKED ISSUE #15174: Remove Infura v3 key from the appearing in network url on the add popular networks feature UI
On the add network modal (screenshot) and when editing network details, for Infura supported urls, we might be exposing the Infura keys in the network url. If we are doing that, we want to hide the key and only show the first part of the url.

- Network url we might be showing now: ‘[https://arbitrum-mainnet.infura.io/v3/${INFURA_API_KEY}](https://arbitrum-mainnet.infura.io/v3/$%7BINFURA_API_KEY%7D)‘

- Network url we want to show: ‘[https://arbitrum-mainnet.infura.io](https://arbitrum-mainnet.infura.io/v3/$%7BINFURA_API_KEY%7D)‘

This applies for Arbitrum, Optimism, Palm and Polygon.

Screenshots: 

Screen Shot 2022-07-11 at 08 49 34

Screen Shot 2022-07-11 at 09 56 55

filipsekulic:
## Explanation

Removed `Infura project ID` from displaying when adding a popular custom network (add network modal) and when editing network details.

## More Information

Fixes: https://github.com/MetaMask/metamask-extension/issues/15174

### Before

Screenshot 2022-07-12 at 11 20 41

Screenshot 2022-07-12 at 11 22 23

### After

Screenshot 2022-07-12 at 11 10 01
Screenshot 2022-07-12 at 11 11 08

## Manual Testing Steps
`1st case`

1. From experimental tab enable showing the popular custom network list.
2. Go to networks tab.
3. Add a network.
4. Click to add a network that relies on Infura ID (without the warning icon).
5. When a popup shows up, `Network URL` contains the Infura ID (after infura.io/v3/).

`2nd case`

1. Go to networks tab.
2. See `New RPC URL` box which contains the Infura ID (after infura.io/v3/).

### DIFF
--- a/shared/constants/network.js
+++ b/shared/constants/network.js
@@ -56,7 +56,7 @@ export const FANTOM_DISPLAY_NAME = 'Fantom Opera';
 export const HARMONY_DISPLAY_NAME = 'Harmony Mainnet Shard 0';
 export const PALM_DISPLAY_NAME = 'Palm';
 
-const infuraProjectId = process.env.INFURA_PROJECT_ID;
+export const infuraProjectId = process.env.INFURA_PROJECT_ID;
 export const getRpcUrl = ({ network, excludeProjectId = false }) =>
   `https://${network}.infura.io/v3/${excludeProjectId ? '' : infuraProjectId}`;
 
--- a/ui/pages/confirmation/templates/add-ethereum-chain.js
+++ b/ui/pages/confirmation/templates/add-ethereum-chain.js
@@ -1,5 +1,6 @@
 import { ethErrors } from 'eth-rpc-errors';
 import React from 'react';
+import { infuraProjectId } from '../../../../shared/constants/network';
 import {
   SEVERITIES,
   TYPOGRAPHY,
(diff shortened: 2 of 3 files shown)


danjm:
Nice work!

I have added two commits to this. One just to prevent type errors by adding optional chaining, because I found that we could hit an error screen after clicking delete/cancel

The other is to prevent saving the the form if there is a chain id error when editing one of the popular networks. This was to prevent a scenario where a user accidentally makes a small change to the RPC url and then saves, and then loses the data about the infura key. This situation would result in nothing working on that network. It would be fixable by the user by deleting the network and then adding it again, but the `chainIdErrorOnFeaturedRpcDuringEdit` just avoids that possibility. More generally, if there is a chainId error/warning of some sort while editing a popular network, it definitely means that something has gone wrong, so the user should be prevented from saving their changes and forced to cancel and start again.

danjm:
I have rebased onto develop to resolve conflicts

danjm (approved):

brad-decker (approved):

PR #15198, opened 2022-07-12
Show a block explorer URL hostname, hide the "add token" notification if block explorer URL is not available

dan437:
## Explanation
Show a block explorer URL hostname, hide the "add token" notification if block explorer URL is not available

## Screenshots/Screencaps
Show a block explorer URL hostname:
image

Hide the "add token" notification if block explorer URL is not available:
image


adonesky1 (approved):
LGTM!

darkwing (approved):

PR #15201, opened 2022-07-12
Fix default currency symbol for `wallet_addEthereumChain` + improve warnings for data that doesn't match our validation expectations

adonesky1:
## Explanation

Currently if the `wallet_addEthereumChain` API is used and no `nativeCurrency` symbol is specified, [we default to calling the native currency for the newly added network `ETH`](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js#L239). We should make `nativeCurrency.symbol` a required field for this API, [as we have done when adding a new network manually](https://github.com/MetaMask/metamask-extension/pull/12431). But making this a required field may break some usage, so we will need to publicly announce the change so dapps can adapt, and in the meantime we should default to calling these native currencies `UNKNOWN` instead of ETH to avoid confusion and any potentially malicious usage.

Reproduction steps to see this issue (on `develop` branch):
1. Visit https://irisapp.xyz/
2. Connect wallet from non-eth network without Polygon installed
3. Let it propose polygon params
4. Accept its parameters
5. Notice any polygon is shown as `ETH`. This could be confusing and could lead to users being phished because an attacker could trick the user into believing they’d been sent `ETH`.

Then repeat steps 1-4 on this branch and see that the currency symbol is `UNKNOWN`

________________________________________ 

Additionally this PR [disaggregates the error we currently show](https://github.com/MetaMask/metamask-extension/pull/15201#discussion_r922315181) when a `wallet_addEthereumChain` call contains parameters we don't expect, so that we show the tell the user which data points don't match.

Finally this PR makes it possible to add a network that is already added but configured with a different RPC Url via that API. [This is currently not permitted](https://github.com/MetaMask/metamask-extension/blob/4993d28a8fee99297f6bbd1697d48d88297fe9c4/app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js#L153). If, however, the network attempting to be added with a new RPC gateway has the same chainId as an already added network but **a different currency symbol** we throw an error to the dapp since this could be  confusing or malicious (i.e. a site trying to convince a user that they are on a network with no value by making the ticker symbol something other than the user recognizes as a having value).


ui/pages/confirmation/templates/add-ethereum-chain.js
```diff
@@ -281,7 +309,12 @@ function getValues(pendingApproval, t, actions) {
             [t('blockExplorerUrl')]: pendingApproval.requestData
               .blockExplorerUrl,
           },
-          prefaceKeys: [t('networkName'), t('networkURL'), t('chainId')],
+          prefaceKeys: [
+            t('networkName'),
+            t('networkURL'),
+            t('chainId'),
+            t('currencySymbol'),
```
adonesky1, line 316: Previously we didn't show the currency symbol in the `Allow this site to add a network` without the user clicking into `view all details`:
Screen Shot 2022-07-11 at 1 20 26 PM (2)

now:
Screen Shot 2022-07-15 at 11 05 37 AM

ui/pages/confirmation/templates/add-ethereum-chain.js
```diff
@@ -75,6 +56,48 @@ const INVALID_CHAIN = {
...
+    element: 'span',
+    children: {
+      element: 'MetaMaskTranslation',
+      props: {
+        translationKey: 'mismatchedRpcUrl',
+      },
+    },
+  },
+};
+
```
adonesky1, line 100: Here I've disaggregated the error we show when a `wallet_addEthereumChain` call contains parameters we don't expect, so that we show which data doesn't match:
Before:
After
Screen Shot 2022-07-15 at 11 13 57 AM
:

shanejonas (approved):

shanejonas:
Looks good, I like that the errors are more specific.

I also added the errors wallet_addEthereumChain returns to the api-spec, here is the PR: https://github.com/MetaMask/api-specs/pull/106

brad-decker (approved):

ui/pages/confirmation/templates/add-ethereum-chain.js
holantonela: 1. Is there a reason you are using two different reds in the warning icon? 
2. If the four fields have a warning, will they stack, preventing the preview from being seen? 
3. Is it possible to add a **bold** in "network name" and "currency symbol"?
4. If the validation is the same across fields, could we simplify that error message? 

`According to our record the network name may not correctly match this chain ID`
`The submitted currency symbol does not match what we expect for this chain ID`

eg. 

`According to our records for this chain ID, these may be wrong: Network Name, Currency Symbol`
`According to what we expect for this chain ID, these may be wrong: Network Name, Currency Symbol`

holantonela: > I was trying to indicate the relative severity of each of these discrepancies.

We may need a pattern for this. I'd recommend to use the same indicator for now.

> Yes this probably makes sense. I will push a follow-up to make this error simpler, using this approach.

Super, thank you.

PR #15205, opened 2022-07-13
Fixed error messages being wrapped in strings when using ganache or h…

### LINKED ISSUE #12646: error from contract returned as string not as object
Error thrown from contract returned as string, so can not parse specific error:

**contract error**

image

- as you can see, actual error message from contract is **Unauthorized Access!**, but metamask returned it as string(message + json object).

**other errors**

image

- other metamask errors are returned as object.

**Expected behavior**

- Should return error as object same as other metamask errors.

 - OS: OS X
 - Browser: Chrome Version 95.0.4638.69 (Official Build) (x86_64)
 - MetaMask Version: 10.4.1

BelfordZ:
Note: Developers might have code that is similarily parsing out the underlying error in a similar way. This PR improves the returned errors, and breaks peoples code for handling the nested-in-a-string errors.

This was the only place I could find inside of extension that is handling errors that are returned in-band when calling sendtransactionRaw to ganache or harthat. 

Fixes #12646

Before:
image
After:
image

Before:
image
After:
image


wbt:
Having just today written even more code that parses out nested-in-string errors (lacking any other way to get the information) I am not a fan of the idea of just breaking this suddenly....there is a lot of DApp code with workarounds for this longstanding issue that will probably break.
Is there a way this could be done in a nonbreaking way, keeping content nested in the string in addition to in a more easily parseable format, plus a communications strategy to get the word out so people can start converting their code over, before some eventual breaking of lots of existing Dapp code?

BelfordZ:
@wbt understandable and agreed. Closing this PR since the plan has changed:
1. We will address this inside of [ethjs-query](https://github.com/ethjs/ethjs-query).
2. We wont change the error message or code, we will add a `data` field with the parsed json part of the current error.message.

In doing so, users who haven't yet written the code to parse out the underyling error will at least have the `error.data` which will contain what they really are after, and those who've already written code against these error messages don't get punished. 

We should leave it like this until we have a well-defined process for incurring 'breaking for the better' changes (thx for the comment @wbt). There are some other fixes that have been held back for the same reason, so it's becoming a greater and greater priority to stamp out a process around the lifecycle of a breaking change to metamask public api. Developers should be given incrementally louder signals about upcoming changes so that they have a predictable and ample amount of time to update their code. We want to make sure we get that part right :)

PR #15206, opened 2022-07-13
[FLASK] Add snap update metrics

### LINKED ISSUE #15202: Utilize `snapUpdated` event in metrics
Currently, we're only tracking snap installs. We should also be tracking snap updates.

hmalik88:
Closes #15202 

We were not utilizing the `snapUpdated` event to track snap updates, this PR adds code to do so.


app/scripts/metamask-controller.js
```diff
@@ -1283,6 +1283,21 @@ export default class MetamaskController extends EventEmitter {
       },
     );
 
+    this.controllerMessenger.subscribe(
+      `${this.snapController.name}:snapUpdated`,
```
FrederikBolding, line 1287: This is not compatible with the current version of skunkworks and needs to wait until we cut that release and include it in the extension.

FrederikBolding (changes requested):
See above

app/scripts/metamask-controller.js
hmalik88: ok, will keep this on standby.

FrederikBolding (dismissed):

app/scripts/metamask-controller.js
```diff
@@ -1291,6 +1291,21 @@ export default class MetamaskController extends EventEmitter {
...
+      (newSnap, oldVersion) => {
+        this.metaMetricsController.trackEvent({
+          event: 'Snap Updated',
+          category: EVENT.CATEGORIES.SNAPS,
+          properties: {
+            snap_id: newSnap.id,
+            oldVersion,
+            newVersion: newSnap.version,
```
brad-decker, line 1303: All properties must be snake cased. Is this event tracked in John Brennan's schema?

brad-decker (dismissed):

app/scripts/metamask-controller.js
```diff
@@ -1291,6 +1291,21 @@ export default class MetamaskController extends EventEmitter {
       },
     );
 
+    this.controllerMessenger.subscribe(
+      `${this.snapController.name}:snapUpdated`,
+      (newSnap, oldVersion) => {
```
digiwand, line 1296: [optional] it might be helpful to rename `newSnap` to `newTruncatedSnap` to preserve the type in the name

digiwand (approved):
👍

FrederikBolding (approved):

PR #15218, opened 2022-07-13
Only check if a user has enough token balance before calling STX

dan437:
## Explanation
Right now we are checking 2 things before calling the `getFees` function for STX on the View Quote page:
- Enough token balance for a swap
- Enough ETH balance for gas fees

With this PR we will only check if a user has enough token balance and then call STX, which will check if a user has enough ETH for gas fees. Also, if the `getFees` call returns an error now that a user doesn't have enough funds, we are falling back to regular swaps. With this PR if there is the "not enough funds" error from STX, we will show an info notification to a user, but keep STX active.

## Manual Testing Steps
- Try to make a swap with ETH balance you don't have
- On the Build Quote page enter almost all ETH balance you have, so you would have enough balance, but not enough ETH  for gas fees. E.g. if you have 0.012 ETH, try 0.011 ETH. You should see a notification about how much more ETH you need


brad-decker (changes requested):

ui/pages/swaps/view-quote/view-quote.js
```diff
@@ -755,14 +794,18 @@ export default function ViewQuote() {
...
+        (currentSmartTransactionsError || smartTransactionsError)) ||
+      (currentSmartTransactionsEnabled &&
+        smartTransactionsOptInStatus &&
+        !smartTransactionFees?.tradeTxFees),
   );
 
   useEffect(() => {
     if (
       currentSmartTransactionsEnabled &&
       smartTransactionsOptInStatus &&
```
adonesky1, line 807: I know this isn't newly added but can you explain the difference between `currentSmartTransactionsEnabled` and `smartTransactionsOptInStatus`?

adonesky1:
A more general question related to the UX of insufficient balances for swaps. Why do we allow users to proceed to build a quote when we already see (and warn them that their balance is insufficient)?
Screen Shot 2022-08-08 at 3 29 25 PM
Screen Shot 2022-08-08 at 3 27 49 PM

ui/pages/swaps/view-quote/view-quote.js
dan437: Sure! So, `currentSmartTransactionsEnabled` tells us if STX are enabled in UI. That depends on multiple things, like is it Ethereum mainnet, no HW wallet, feature flag for STX on, etc.

`smartTransactionsOptInStatus` indicates if a user opted into STX or if they want to use regular swaps.

dan437:
> A more general question related to the UX of insufficient balances for swaps. Why do we allow users to proceed to build a quote when we already see (and warn them that their balance is insufficient)?
> 
> Screen Shot 2022-08-08 at 3 27 49 PM

I was wondering the same thing a long time ago. I was told it's because we want to show our quotes, even if a user has 0 balance for their `Swap from` token. That way if they like our quotes, it increases likelihood that they would use our service after they get enough funds.

ui/pages/swaps/view-quote/view-quote.js
adonesky1: Maybe I'm misunderstanding but it sounds like `smartTransactionsOptInStatus ` is contained within `currentSmartTransactionsEnabled`?  Is it not redundant somehow?

dan437: `currentSmartTransactionsEnabled` doesn't contain the Opt In check. If it's true, you can see the STX switch in UI: 
image

But if you turn off Smart transaction, `smartTransactionsOptInStatus` will be false:
image

adonesky1: OH sorry I see I was confusing enabled via the feature flag with enabled by this toggle in the UI.

adonesky1 (approved):
LGTM

darkwing (approved):

brad-decker (approved):

PR #15227, opened 2022-07-13
Only send the rpc origin to segment when calling trackevent for newly added custom networks

adonesky1:
We should only send the origin of the rpc url to segment for newly added custom networks.

Was attempting to add tests to guarantee we are calling `trackEvent` with the correct values, but mocking `useContext` is not easily achieved, and perhaps not advisable, so will leave that for a later PR.

### DIFF
--- a/app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
+++ b/app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
@@ -261,7 +261,13 @@ async function addEthereumChainHandler(
       }),
     );
 
-    const rpcUrlOrigin = new URL(firstValidRPCUrl).origin;
+    let rpcUrlOrigin;
+    try {
+      rpcUrlOrigin = new URL(firstValidRPCUrl).origin;
+    } catch {
+      // ignore
+    }
+
     sendMetrics({
       event: 'Custom Network Added',
       category: EVENT.CATEGORIES.NETWORK,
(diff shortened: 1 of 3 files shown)


danjm:
Thanks for this. Can you add to the other two places we are tracking the rpc origin: https://github.com/MetaMask/metamask-extension/pull/15176

adonesky1:
> Thanks for this. Can you add to the other two places we are tracking the rpc origin: https://github.com/MetaMask/metamask-extension/pull/15176

done in latest commit: https://github.com/MetaMask/metamask-extension/pull/15227/commits/e17d8781e10436e88a317c8fe1b752d1d35a291a

darkwing:
The only thing that slightly worries me here is that `new URL(someBogusString)` throws, so by this point, it's super important that this string is validated

NiranjanaBinoy (dismissed):
LGTM

adonesky1:
> The only thing that slightly worries me here is that new URL(someBogusString) throws, so by this point, it's super important that this string is validated

I don't believe it's currently possible for us to reach any of these trackEvent calls with an invalid URL for `rpcURL` but just to be careful, and in case the upstream validation is broken at some point, I have wrapped all the `new URL()` constructors in try/catches [here](https://github.com/MetaMask/metamask-extension/pull/15227/commits/f87d4613d3cb74327d481563fc5c9bb9febaff7d). @darkwing

danjm (approved):

brad-decker (approved):

PR #15242, opened 2022-07-14
Bumps eth-lattice-keyring to v0.10.0

alex-miller-0:
> NOTE: This has been opened in place of #15154 because I am a contributor to this repo and @douglance is not, so his PR's tests wouldn't run.

## Explanation

GridPlus has shipped a new version of firmware for the Lattice1. This new version of the firmware includes support for automatic decoding of ABI data, which is then displayed for users on their device.

In order to accomplish this for MetaMask users, we will need to bump the version of our eth-lattice-keyring. This [updated version of the eth-lattice-keyring](https://github.com/GridPlus/eth-lattice-keyring) uses [a function of the gridplus-sdk](https://github.com/GridPlus/gridplus-sdk/blob/v2.1.0/src/util.ts?rgh-link-date=2022-07-07T15%3A13%3A30Z#L455=) that fetches ABI data and bundles it with requests to the Lattice1.

### DIFF
--- a/lavamoat/browserify/beta/policy.json
+++ b/lavamoat/browserify/beta/policy.json
@@ -3448,7 +3448,7 @@
         "@metamask/controllers>isomorphic-fetch": true,
         "@metamask/controllers>multiformats": true,
         "@metamask/controllers>web3-provider-engine": true,
-        "@metamask/metamask-eth-abis": true,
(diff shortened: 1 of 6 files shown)


alex-miller-0:
Oops, closing because some non-GridPlus lavamoat changes seem to have been committed.

PR #15248, opened 2022-07-15
Ensure that editing a tx from a transfer to a simple send resets data and updates type

### LINKED ISSUE #15184: [Bug]: Uncaught (in promise) Error: Unable to determine contract standard
### Describe the bug

An unhandled exception is thrown and the user is presented with a infinite loading spinner when switching between ERC20 and Native token sends

### Steps to reproduce

1. See video below
2. Enter the send flow for an ERC20 asset
3. Select an address on the send to screen
4. Select the max amount, then press Next on the Send Token screen
5. Click the Edit button to edit the transaction
6. Switch the asset from ERC20 to the Native token
7. Select the max amount, then press Next on the Send Token screen
8.  Click the Edit button to edit the transaction

https://user-images.githubusercontent.com/53189696/178337098-dbcbf2f4-7f91-49b1-ac7e-b8b30d6e6217.mov

```shell
Error: Unable to determine contract standard
  at h.<anonymous>(metamask/common-0.js:1:257458)
  at Generator.throw(<anonymous>)
  at s(metamask/common-0.js:1:255248)
```

### Version

10.18.0

### Build type

_No response_

### Browser

Chrome

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

danjm:
Fixes https://github.com/MetaMask/metamask-extension/issues/15184

Preceding v10.18.0 there was a problem whereby changing from a token transfer to a simple send on the send edit screen would not change the transaction data. This would incorrectly leave the rendered type as transfer and incorrectly leave hex data on the transaction.

With v10.18.0, this could cause an unhandled error when trying to edit a transaction in this state of faulty data, because we would attempt to parse the recipient address of a simple send as if it were a token address.

The solution requires two parts:
1. Ensure that the transaction data gets correctly reset when changing a transaction from a transfer to a simple send
2. Ensure that we wait for the background state update to complete and be reflected in the UI state before returning to the confirm screen after editing a transaction

### DIFF
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -1992,9 +1992,12 @@ export function updateSendAsset(
         }),
       );
 
-      // If there is an unapprovedTx in background state and its type was a token method,
+      // This is meant to handle cases where we are editing an unapprovedTx from the background state
+      // and its type is a token method. In such a case, the hex data will be the necessary hex data
+      // for calling the contract transfer method.
+      // Now that we are updating the transaction to be a send of a native asset type, we should
+      // set the hex data of the transaction being editing to be empty.
       // then the user will not want to send any hex data now that they have change the
-      // transaction from being a token transfer to a simple send.
       if (
         unapprovedTx?.type === TRANSACTION_TYPES.TOKEN_METHOD_TRANSFER_FROM ||
         unapprovedTx?.type === TRANSACTION_TYPES.TOKEN_METHOD_TRANSFER
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -1994,7 +1994,8 @@ export function updateSendAsset(
       // then the user will not want to send any hex data now that they have change the
       if (
         unapprovedTx?.type === TRANSACTION_TYPES.TOKEN_METHOD_TRANSFER_FROM ||
-        unapprovedTx?.type === TRANSACTION_TYPES.TOKEN_METHOD_TRANSFER
+        unapprovedTx?.type === TRANSACTION_TYPES.TOKEN_METHOD_TRANSFER ||
+        unapprovedTx?.type === TRANSACTION_TYPES.TOKEN_METHOD_SAFE_TRANSFER_FROM
       ) {
         await dispatch(actions.updateUserInputHexData(''));
       }
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -937,6 +937,13 @@ const slice = createSlice({
       draftTransaction.asset.type = asset.type;
       draftTransaction.asset.balance = asset.balance;
       draftTransaction.asset.error = asset.error;
+
+      // If an asset update occurs that changes the type from 'NATIVE' to
+      // 'NATIVE' then this is likely the initial asset set of an edit
+      // transaction. We don't need to set the amount to zero in this case.
+      // The only times where an update would occur of this nature that we
+      // would want to set the amount to zero is on a network or account change
+      // but that update is handled elsewhere.
       if (
         draftTransaction.asset.type === ASSET_TYPES.TOKEN ||
         draftTransaction.asset.type === ASSET_TYPES.COLLECTIBLE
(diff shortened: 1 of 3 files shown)


ui/ducks/send/send.js
```diff
@@ -1905,6 +1909,16 @@ export function updateSendAsset(
           error: null,
         }),
       );
+
+      // If there is an unapprovedTx in background state and its type was a token method,
+      // then the user will not want to send any hex data now that they have change the
+      // transaction from being a token transfer to a simple send.
+      if (
+        unapprovedTx?.type === TRANSACTION_TYPES.TOKEN_METHOD_TRANSFER_FROM ||
+        unapprovedTx?.type === TRANSACTION_TYPES.TOKEN_METHOD_TRANSFER
```
brad-decker, line 1918: I know this is nitty, but when I read this comment I thought unapprovedTx was like...*any* unapproved tx. Can we change this comment to be something along the lines of if the original type of this transaction was a token method then we need to null out he usetInputHeadData (words are not important just specifying original and ensuring that the unapprovedTx is *clearly* marked as the original tx in an edit flow)

ui/ducks/send/send.js
```diff
@@ -2128,6 +2142,27 @@ export function signTransaction() {
...
+      const priorHexDataMatchesSpecialTokenTxSignatures =
+        parseStandardTokenTransactionData(editingTx.txParams.data) !==
+        undefined;
+      if (
+        sendingNativeAsset &&
+        userInputHexDataHasBeenReset &&
+        priorHexDataMatchesSpecialTokenTxSignatures
+      ) {
+        editingTx.txParams.data = '';
+      }
```
brad-decker, line 2164: I"m confused why we can't just rely upon the value of userInputHexData to override this? It should be set to an empty string n 1920, therefor the call to generateTransactionParams on line 2115 should return an appropriate txParams object. with data null. Can you help me understand why this contract breaks down under these special conditions?

ui/store/actions.js
```diff
@@ -745,7 +745,7 @@ export function updateEditableParams(txId, editableParams) {
       log.error(error.message);
       throw error;
     }
-
+    await forceUpdateMetamaskState(dispatch);
```
brad-decker, line 748: [q] why is this necessary?

brad-decker (changes requested):

ui/ducks/send/send.js
danjm: Comment improved in 91d8a85d2

ui/ducks/send/send.js
danjm: yes removed in ee260f475

ui/store/actions.js
danjm: To ensure that transaction state is properly updated before hitting the confirm screen routing logic after clicking the "Next" button

ui/ducks/send/send.js
```diff
@@ -937,6 +937,13 @@ const slice = createSlice({
       draftTransaction.asset.type = asset.type;
       draftTransaction.asset.balance = asset.balance;
       draftTransaction.asset.error = asset.error;
+
+      // If an asset update occurs that changes the type from 'NATIVE' to
+      // 'NATIVE' then this is likely the initial asset set of an edit
+      // transaction. We don't need to set the amount to zero in this case.
+      // The only times where an update would occur of this nature that we
+      // would want to set the amount to zero is on a network or account change
+      // but that update is handled elsewhere.
```
adonesky1, line 946: why was this comment added back?

ui/ducks/send/send.js
adonesky1: Should we include `TOKEN_METHOD_SAFE_TRANSFER_FROM` here. I don't believe it's possible for it to be a problem at the moment because it's a method on ERC721's and we don't enable the `edit` button on ERC721 sends at the moment, but we may want to soon?

brad-decker (approved):

ui/ducks/send/send.js
brad-decker: rebase issue more than likely.

brad-decker: I will reapprove after this is removed

brad-decker (changes requested):

ui/ducks/send/send.js
danjm: added in 1eed22c1d

adonesky1 (approved):
LGTM!

NiranjanaBinoy (approved):
LGTM

brad-decker (approved):

PR #15283, opened 2022-07-19
MV3 testcase for measuring the time it takes to load a newly created Account

### LINKED ISSUE #15282: MV3: capture time for loading account balance in E2E
We need an e2e test that captures how much time it takes from Creating an Account until its Balance is loaded.

seaona:
## Explanation
We need a way to measure the delay caused by a service-worker re-start on common user actions along the application.
For that, this PR will take care of measuring common action times, on an MV2 test build, so we have this time reference. Actions covered:
- how much time it takes for loading a newly created account (from Create Account button until balance is loaded
- how much time it takes from clicking Confirm tx until Tx is confirmed

**Approach**: this task will be done in 2 PRs:
- The 1st PR (this one), is going to create the e2e test for performing the actions of Load an Account and Send a Tx.
- The 2nd PR ([draft](https://github.com/MetaMask/metamask-extension/pull/15305)) will create a build artifact for exporting the measurements into a json file.

Closes #15282 #15291


jpuri (changes requested):
A few points here:

- since this is not really a test but way to measure performance we can probably move this next to [benchmark.js](https://github.com/MetaMask/metamask-extension/blob/develop/test/e2e/benchmark.js)
- add test case for confirm transaction also - this was discussed in office hours with Dan
- add artifact - this can be simple json with response time in each action

test/e2e/mv3-load-times.spec.js
```diff
@@ -0,0 +1,131 @@
...
+        }, 10000);
+
+        await driver.waitForSelector('.transaction-status--confirmed');
+        const timestampAfterAction = new Date();
+        loadingTimes.sendTx = timestampAfterAction - timestampBeforeAction;
+
+        assert.equal(
+          Object.prototype.hasOwnProperty.call(loadingTimes, 'sendTx'),
+          true,
+        );
```
jpuri, line 127: Since we are only measuring response time here, we can avoid assert.

seaona: make sense. I've removed both then @jpuri

seaona:
Close PR in favour to new approach
https://github.com/MetaMask/metamask-extension/pull/15353

PR #15293, opened 2022-07-20
Update Eslint and deps

brad-decker:
## Explanation
We want to be able to take advantage of improvements made in eslint as well as update our plugins to make use of improved rules and performance improvements / new features in prettier, etc. This Pull request updates the very minimum amount of plugins, parsers and the core of eslint to be functional. Once this PR lands a series of additional pull requests will be made updating the various plugins to higher versions. 

Some findings during this process:
1. upgrading to eslint 8, for some reason, downgraded the version of acorn that exists at the root of the node_modules folder. This caused an issue with the babelify portion of the build process. I pinned our acorn version to the previous version to bypass this.
2. A number of rule updates caught additional issues with our code. Those have been fixed in a series of commits with the rule names mentioned. 

## Pre-Merge Checklist


darkwing (approved):

ui/components/app/collectibles-tab/collectibles-tab.test.js
```diff
@@ -137,8 +137,8 @@ const COLLECTIBLES_CONTRACTS = [
 ];
 
 const collectiblesDropdownState = {
-  0x495f947276749ce646f68ac8c248420045cb7b5e: true,
-  0xdc7382eb0bc9c352a4cba23c909bda01e0206414: true,
+  '0x495f947276749ce646f68ac8c248420045cb7b5e': true,
```
adonesky1, line 140: Should this rule be applied to addresses as keys in objects? https://eslint.org/docs/latest/rules/no-loss-of-precision

adonesky1 (approved):
Just one non-blocking question

ui/components/app/collectibles-tab/collectibles-tab.test.js
brad-decker: image

brad-decker: technically speaking, it doesn't matter, because the precision will be lost in the same way even when accessing the keys, however we don't typically pass addresses as numbers. We use strings, and this rule will help to make sure we are using strings for keys as well

adonesky1: 👍

Gudahtt:
This has resulted in many peer dependency warnings, as our eslint config doesn't support ESLint v8 yet.

PR #15300, opened 2022-07-20
Remove decentralized 4byte function signature registry since it contains incorrect signatures and we can't algorithmically check for best option when 4byte.directory is down

adonesky1:
Despite [recent efforts](https://github.com/MetaMask/metamask-extension/pull/14937) to override incorrect signatures being published to the 4byte signature directories that we use to determine smart contract method names, [users are still experiencing this issue.](https://consensys.slack.com/archives/GTQAGKY5V/p1658329871193219). This time the issue appears to have resurfaced because 4byte.directory was down meaning that we could only look to the on-chain parity signature registry (via [eth-method-registry](https://github.com/MetaMask/eth-method-registry)) which unfortunately doesn't provide any metadata about date of the entry so we can't choose the earliest entry (which is likely to be the most valid) as we can with 4byte.directory.

### DIFF
--- a/ui/helpers/utils/transactions.util.js
+++ b/ui/helpers/utils/transactions.util.js
@@ -43,15 +43,6 @@ async function getMethodFrom4Byte(fourBytePrefix) {
   return fourByteResponse.results[0].text_signature;
 }
 
-function pickShortest(registrySig, fourByteSig) {
-  if (!registrySig) {
-    return fourByteSig;
-  } else if (!fourByteSig) {
-    return registrySig;
-  }
-  return fourByteSig.length < registrySig.length ? fourByteSig : registrySig;
-}
-
 let registry;
 
 /**
@@ -71,18 +62,11 @@ export async function getMethodDataAsync(fourBytePrefix) {
       registry = new MethodRegistry({ provider: global.ethereumProvider });
     }
 
-    const registrySig = await registry.lookup(fourBytePrefix).catch((e) => {
-      log.error(e);
-      return null;
-    });
-
-    const sig = pickShortest(registrySig, fourByteSig);
-
-    if (!sig) {
+    if (!fourByteSig) {
       return {};
     }
 
-    const parsedResult = registry.parse(sig);
+    const parsedResult = registry.parse(fourByteSig);
 
     return {
       name: parsedResult.name,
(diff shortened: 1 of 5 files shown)


ui/helpers/utils/transactions.util.js
```diff
@@ -71,18 +62,11 @@ export async function getMethodDataAsync(fourBytePrefix) {
       registry = new MethodRegistry({ provider: global.ethereumProvider });
```
FrederikBolding, line 62: We can remove the registry entirely and the `eth-method-registry` dependency.

ui/helpers/utils/transactions.util.js
```diff
@@ -71,18 +62,11 @@ export async function getMethodDataAsync(fourBytePrefix) {
...
-
-    const sig = pickShortest(registrySig, fourByteSig);
-
-    if (!sig) {
+    if (!fourByteSig) {
       return {};
     }
 
-    const parsedResult = registry.parse(sig);
+    const parsedResult = registry.parse(fourByteSig);
```
FrederikBolding, line 69: You can remove part of the `nock` fixtures for this test now too.

ui/helpers/utils/transactions.util.js
adonesky1: We still use it to [parse the signature from 4byte](https://github.com/MetaMask/metamask-extension/pull/15300/files/cb31bfc23bae1e11c08c04a11a187938e80714a8#diff-c8c256cb0389d020fcf5c440c7133f063c3d2a1b6321b3252d23cd0208bb79dbL85). Are you suggesting we build a utility function to do this ourselves?

adonesky1: I suppose we could just [copy over the `parse` method](https://github.com/MetaMask/eth-method-registry/blob/main/src/index.ts#L52) from `eth-method-registry` since we don't need anything else from the package here

FrederikBolding: Ah, my mistake. Yeah we could move it out, but it doesn't have to be as part of this PR imo.

ui/helpers/utils/transactions.util.js
adonesky1: removed [here](https://github.com/MetaMask/metamask-extension/pull/15300/commits/13b6b30eec0a6a7ac02d8ea968c932ced94d531f)

FrederikBolding (approved):

brad-decker (approved):

PR #15304, opened 2022-07-20
Improve `tokenId` parsing and clean up `useAssetDetails` hook

adonesky1:
There are 2 chunks of work included in this PR to improve the accuracy and code legibility of how we extract and use `tokenId` values from parsed transaction data when showing users confirmations related to NFT transactions:

1. Adds `getTokenIdParam` function to pull values `_tokenId` (in the case of supported ERC721 transactions- [here](https://github.com/MetaMask/metamask-eth-abis/blob/d0474308a288f9252597b7c93a3a8deaad19e1b2/src/abis/abiERC721.ts#L62), [here](https://github.com/MetaMask/metamask-eth-abis/blob/d0474308a288f9252597b7c93a3a8deaad19e1b2/src/abis/abiERC721.ts#L143) & [here](https://github.com/MetaMask/metamask-eth-abis/blob/d0474308a288f9252597b7c93a3a8deaad19e1b2/src/abis/abiERC721.ts#L62)) and `id` (in the case of supported [ERC1155 transactions](https://github.com/MetaMask/metamask-eth-abis/blob/d0474308a288f9252597b7c93a3a8deaad19e1b2/src/abis/abiERC1155.ts#L250)) from transaction data. We previously relied only on the `_value` param, but after further digging it appears `_value` is only how `tokenId` is parameterized in the transaction data in some cases... (see comment [here](https://github.com/MetaMask/metamask-extension/blob/cb5252d84735522cbe8d0bc2dbf6be2dc8b478c6/ui/helpers/utils/token-util.js#L251) for further details)

2.  Refactors `useAssetDetails` and `getAssetDetails` to [avoid unnecessary network calls where possible](https://github.com/MetaMask/metamask-extension/pull/15304/files#diff-df0b6edfb26c6bba935a578c7457bfc1c55fd54a8f81613c7882a80fab892363R262), and improve overall legibility.

## ERC721
**Before:**
Screen Shot 2022-07-21 at 5 16 03 PM

**After:**
Screen Shot 2022-07-21 at 5 12 07 PM

## ERC1155
**Before:**
Screen Shot 2022-07-21 at 5 20 39 PM

**After:**
Screen Shot 2022-07-21 at 5 23 45 PM

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -91,7 +91,10 @@ import {
 import { EVENT, EVENT_NAMES } from '../../shared/constants/metametrics';
 
 import { hexToDecimal } from '../../ui/helpers/utils/conversions.util';
-import { getTokenValueParam } from '../../ui/helpers/utils/token-util';
+import {
+  getTokenIdParam,
+  getTokenValueParam,
+} from '../../ui/helpers/utils/token-util';
 import { isEqualCaseInsensitive } from '../../shared/modules/string-utils';
 import { parseStandardTokenTransactionData } from '../../shared/modules/transaction.utils';
 import {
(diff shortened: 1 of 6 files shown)


brad-decker (approved):

digiwand (approved):
Nice work! 

Verified w/ Chrome on macOS. 

Verified that some methods return the token id through `_tokenId` and others through `_value` (approve example)
Screenshot 2022-07-23 at 6 09 26 PM

I also couldn't find where they specified the `_value` return value of the token id for ERC721. This is also a head-scratcher for me.

PR #15313, opened 2022-07-21
Metrics adjustments

danjm:
ba1becba8 Don't send errors to sentry if users have not opted-in to participate in metametrics
31f0ffc0c Don't capture opt-out metrics
dfba519f3 Move the metrics-opt in screen to immediately after the welcome screen


kumavis (dismissed):

app/scripts/lib/setupSentry.js
```diff
@@ -103,7 +103,17 @@ export default function setupSentry({ release, getState }) {
     environment,
     integrations: [new Dedupe(), new ExtraErrorData()],
     release,
-    beforeSend: (report) => rewriteReport(report),
+    beforeSend: (report) => {
+      if (getState) {
```
Gudahtt, line 107: This will effectively disable Sentry in the background, because `getState` is never set in the background.

app/scripts/background.js
```diff
@@ -383,7 +383,9 @@ function setupController(initState, initLangCode, remoteSourcePort) {
         // log error so we dont break the pipeline
         if (!dataPersistenceFailing) {
           dataPersistenceFailing = true;
-          captureException(err);
+          if (state.data.metamask.participateInMetaMetrics) {
```
Gudahtt, line 386: I don't think we need this _and_ the modifications to `beforeSend`. The `beforeSend` changes should be adequate on their own, and it avoids having to be vigilant about `captureException` calls.

app/scripts/lib/setupSentry.js
Gudahtt: This is still a good idea, but we would need to ensure `getState` is set properly.

app/scripts/background.js
Gudahtt: The same goes for the `captureException` wrapper added in `metamask-controller.js`

danjm: but this `captureException` is imported directly `import { captureException } from '@sentry/browser';`, it's not from `global.sentry`, so I didn't think the `beforeSend` change would apply to it.

danjm: I guess we should replace all cases of `import { captureException } from '@sentry/browser';` with `const { captureException } = global.sentry`?

app/scripts/lib/setupSentry.js
danjm: okay, I think have done that here https://github.com/MetaMask/metamask-extension/pull/15313/commits/1eba37b2fb59f6ad57c601d3117217e08eff1b1a

app/scripts/background.js
Gudahtt: The Sentry library acts as a singleton. It must be initialized once with `.init`, and those settings take effect globally. Without that `init` call, it doesn't work at all.

danjm: Okay, I've remove this in d7202b4e0

ui/pages/first-time-flow/welcome/welcome.test.js
```diff
@@ -15,7 +15,7 @@ describe('Welcome', () => {
     sinon.restore();
   });
 
-  it('routes to select action when participateInMetaMetrics is not initialized', () => {
+  it('routes to the metametrics screen when participateInMetaMetrics is not initialized', () => {
```
digiwand, line 18: For more coverage, we could add a test to check the `INITIALIZE_CREATE_PASSWORD_ROUTE`

digiwand (approved):
👍

brad-decker (approved):

app/scripts/lib/setupSentry.js
Gudahtt, line 107: I just tried testing this, and there is still a call made to Sentry upon `.init` being called. This effectively prevents error reports from being sent, but it doesn't delay all communication. We will need to come up with a different strategy.

Typically I would suggest delaying the call to `init`, but in our case that's challenging because _I suspect_ that `init` has side-effects that do not work correctly under lockdown.  It's possible they happen on import, not on `init`, but I haven't checked. If they happen on `init`, we're in a tough spot. It means that if we call it later, it'll fail.

Assuming that is true, we might be able to effectively stop all communication by passing in a custom transport. The transport option can be used to handle proxies and things like that, so I would expect it to affect absolutely all communication, even the initial call.

Apparently `init` doesn't support the `transport` option anymore despite it being documented, but it can be set by creating our own Sentry client manually. (details here: https://github.com/getsentry/sentry-javascript/issues/5422)

Another option would be to restart the process after opt-in, and somehow communicate to this bundle that the user opted in :thinking: Challenging since it would have to run before the controllers. This seems like a worse idea than the first one.

Gudahtt: I'm going to consider this non-blocking for this PR because what you have implemented so far is still an improvement, it's just not a complete solution yet.

danjm: Okay, I will merge and then create a ticket to address the issues you have raised here.

PR #15316, opened 2022-07-21
Remove unused bundling options in build script

Gudahtt:
Two unused options have been removed from the `createNormalBundle` function in the build script: 'extraEntries` and `modulesToExpose`.'

Both of these options were used in the old "main" bundles, before we began using the "factored" bundles. They have been unused since #11080.


kumavis (approved):

brad-decker (approved):
LGTM

PR #15318, opened 2022-07-21
Fix environment variable injection

Gudahtt:
The environment variables `IN_TEST` and `METAMASK_DEBUG` were not being set to `false` correctly. Instead those variables were being skipped, and were resolved to `undefined` at runtime. This is confusing because the other environment variables do not work that way - they can be set to false.

The build script has been updated to ensure those two environment variables are always set to `true` or `false` - never `undefined`.

Additionally, the `METAMASK_VERSION` environment variable was being omitted from the `app-init.js` bundle. For the sake of consistency, that has also been restored.

## Manual Testing Steps

For `IN_TEST` and `METAMASK_DEBUG`, try inspecting the output bundle to ensure that they have been injected correctly. 

For `METAMASK_VERSION`, you could test that it's being added correctly by temporarily adding it to `app-init.js`, perhaps using a `console.log`.

## Pre-Merge Checklist


kumavis (dismissed):

brad-decker (dismissed):

Gudahtt:
Rebased to resolve conflict

brad-decker (approved):

darkwing (approved):

PR #15319, opened 2022-07-21
Add build script comments

Gudahtt:
This is a follow-up to https://github.com/MetaMask/metamask-extension/pull/15318, which fixed a problem with environment variables. Every function in this module that passes options related to environment variables has been updated with a doc comment. This should make it clearer which options are mandatory and which are optional, hopefully preventing a similar mistake from happening in the future.


Gudahtt:
This depends upon:
* #15316
* #15317
* #15318

kumavis (dismissed):

brad-decker (approved):

darkwing (approved):

PR #15320, opened 2022-07-21
Fix "app-init" injection

Gudahtt:
## Explanation

The way we were injecting variables into the `app-init.js` bundle was accidentally overwriting the bundle output with the raw `app-init.js` source file. This is a problem because the bundling process handles a lot of things we care about like source maps, polyfills and other necessary Babel transformations, environment variable injection, and minification.

Instead of using string replacement to inject variables, we are now using environment variables. The old string replacement strategy has been removed, and the `app-init.js` module is now generated using the same process as our other bundles.

A new option, "extraEnvironmentVariables", was added to allow us to inject environment variables specifically for this bundle.

## More Information

This string replacement strategy for injection was originally introduced in #14762, and later expanded in #14795.

## Manual Testing Steps

When running `ENABLE_MV3=true yarn dist`, you should now see that `dist/chrome/app-init.js` is minified and includes polyfills. Previously it included no polyfills and was not minified.

I do not expect this to have any user-facing changes. It should work the same as it did before.

## Pre-Merge Checklist

### DIFF
--- a/app/scripts/app-init.js
+++ b/app/scripts/app-init.js
@@ -50,8 +50,12 @@ function importAllScripts() {
   };
 
   const startImportScriptsTime = Date.now();
+
   // value of applyLavaMoat below is dynamically replaced at build time with actual value
   const applyLavaMoat = process.env.APPLY_LAVAMOAT;
+  if (typeof applyLavaMoat !== 'boolean') {
+    throw new Error('Missing APPLY_LAVAMOAT environment variable');
+  }
 
   loadFile('./globalthis.js');
   loadFile('./sentry-install.js');
--- a/app/scripts/app-init.js
+++ b/app/scripts/app-init.js
@@ -5,9 +5,7 @@
 // eslint-disable-next-line
 let scriptsLoadInitiated = false;
 
-// Variable testMode is set to true when preparing test build.
-// This helps in changing service worker execution in test environment.
-const testMode = false;
+const testMode = process.env.IN_TEST;
 
 const loadTimeLogs = [];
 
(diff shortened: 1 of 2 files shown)


Gudahtt:
This depends upon #15318

jpuri:
I tested the changes locally and it seems to work well.

Gudahtt:
I was unable to successfully measure the performance impact of the changes to `app-init.js`. It now includes polyfills that it didn't have before, and it is minified. The file is still pretty small (23kb) so hopefully the impact is minimal.

jpuri (dismissed):
LGTM

brad-decker (dismissed):

app/scripts/app-init.js
```diff
@@ -52,12 +50,13 @@ function importAllScripts() {
 
   const startImportScriptsTime = Date.now();
   // value of applyLavaMoat below is dynamically replaced at build time with actual value
-  const applyLavaMoat = true;
+  const applyLavaMoat = process.env.APPLY_LAVAMOAT;
```
digiwand, line 53: non-blocker: maybe 
```suggestion
  const applyLavaMoat = process.env.APPLY_LAVAMOAT ?? true;
```
to preserve previous default value

digiwand (dismissed):

app/scripts/app-init.js
Gudahtt: Previously the `true` value was _always_ replaced by the build script, so really there was no default despite the appearances.

Fair point that `undefined` is a poor default, and would lead to confusion if we accidentally stopped injecting this. Rather than add a default, I will add a check to ensure an error is thrown if this is missing.

Gudahtt:
I was able to successfully capture a performance trace of the service worker initialization, specifically for the `app-init.js` file. On my machine it consistently takes ~10ms for this file to be loaded and compiled. A negligible amount of time in the overall load time.

<details>
<summary>Screenshot:</summary>

perf-screenshot-sw-init-2

 </details>

brad-decker (approved):

digiwand (approved):

PR #15332, opened 2022-07-22
cast chainId to hex for metrics

brad-decker:
## Explanation
@worldlyjohn called out in the  developer mixpanel that the chainId for this event was showing as a decimal. This only happens with one of the three cases where we fire the event, because in the advanced form the chainId variable is in decimal format. This PR simply casts it to hex format for the metric event.

### DIFF
--- a/ui/pages/settings/networks-tab/networks-form/networks-form.js
+++ b/ui/pages/settings/networks-tab/networks-form/networks-form.js
@@ -11,6 +11,7 @@ import PropTypes from 'prop-types';
 import validUrl from 'valid-url';
 import log from 'loglevel';
 import classnames from 'classnames';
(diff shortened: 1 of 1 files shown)


ui/pages/settings/networks-tab/networks-form/networks-form.js
```diff
@@ -518,14 +519,15 @@ const NetworksForm = ({
         } catch {
           // error
         }
+        console.log(chainId);
```
digiwand, line 522:
```suggestion
```

brad-decker: Thank you <3

segun (approved):

NiranjanaBinoy (approved):

Issue #14853, opened 2022-06-03
MV3: handling pending and new user requests from UI while service worker is in-active


jpuri:
As service worker in MV3 becomes inactive, there might be some RPC requests from UI that are in progress or user may even make new requests while the service worker is in-active.

We need to handle these request and ensure that user receives response.

We need to figure out:

1. How do we restart requests which were in progress, we may need to store all in-progress requests in session and assign them a unique id. This can help restart request once service worker is back.
We will also need to change how UI handles response, we need asynchronous request-response here as there may be a time for service worker in-activity in-between request and response.

2. How we revive things like polling for gas estimates, etc.

PR #15337, opened 2022-07-26
MV3: add retry logic to actions

### LINKED ISSUE #15535: Add retry logic to actions on client side
If an action is submitted to background from client side and either

1. service worker is inactive or
2. service worker dies while the request is in progress

the action should be replayed, this should be seemless to the user.

jpuri:
Fixes: #15535

Add retry logic to actions


ui/store/actions.js
```diff
@@ -399,7 +431,7 @@ export function forgetDevice(deviceName) {
   return async (dispatch) => {
     dispatch(showLoadingIndication());
     try {
-      await promisifiedBackground.forgetDevice(deviceName);
+      await submitRequestToBackground('forgetDevice', [deviceName]);
     } catch (error) {
```
jpuri, line 435: We wil replace all calls to `await promisifiedBackground.forgetDevice` by `await submitRequestToBackground`.

ui/store/actions.js
```diff
@@ -51,9 +52,37 @@ import * as actionConstants from './actionConstants';
...
+    await promisifiedBackground[method](...args);
+  }
+};
+
+const clearRetryQueue = () => {
+  while (retryQueue.length) {
+    const item = retryQueue.shift();
+    submitRequestToBackground(item);
+  }
+};
```
jpuri, line 74: Should we add timestamp also to the requests and check that very old requests are not replayed but rather rejected.

It might also be useful to run `clearRetryQueue` at regular intervals to fail such old messages.

ui/store/actions.js
```diff
@@ -51,9 +52,37 @@ import * as actionConstants from './actionConstants';
...
+  if (isManifestV3) {
+    if (background.connectionStream.readable) {
+      await promisifiedBackground[method](...args);
+    } else {
+      retryQueue.push({ method, args });
+    }
+  } else {
+    await promisifiedBackground[method](...args);
+  }
+};
```
jpuri, line 67: Here we need to handle an edge case when service worker dies when request is in progress.

We can check the exception returned from backend to find reason of request failure and retry request.

danjm:
Progresses #14853

ui/store/actions.js
digiwand: I wonder if we can use the background sync for the edge cases. Sending an action can be retried when the connection is re-established. 

ref: https://developer.chrome.com/blog/background-sync/#how-to-request-a-background-sync

ui/store/actions.js
```diff
@@ -51,9 +52,37 @@ import * as actionConstants from './actionConstants';
 
 let background = null;
 let promisifiedBackground = null;
+const retryQueue = [];
```
digiwand, line 55: 1. Since the file is quite large, I wonder if it would help if we renamed `retryQueue` to be more specific?

---- 
2. Would we want this queue to be persisted should anything happen to the user's session? Should we consider storing the data e.g. chrome.storage?

ui/store/actions.js
```diff
@@ -51,9 +52,37 @@ import * as actionConstants from './actionConstants';
...
+      await promisifiedBackground[method](...args);
+    } else {
+      retryQueue.push({ method, args });
+    }
+  } else {
+    await promisifiedBackground[method](...args);
+  }
+};
+
+const clearRetryQueue = () => {
```
digiwand, line 69: The name `clearRetryQueue` has room for interpretation. My first thought was that this function would erase the retry queue without running the methods. What if we renamed this to something like `runRetryQueue`?

ui/store/actions.js
digiwand: The idea sounds good to me. I'm not yet familiar enough with the methods being called to specify an appropriate retirement time

ui/store/ui.js
```diff
@@ -0,0 +1,174 @@
...
+    });
+
+    log.error(err.stack);
+    throw err;
+  }
+
+  // create platform global
+  global.platform = new ExtensionPlatform();
+
+  // identify window type (popup, notification)
```
digiwand, line 48: double-checking: can the window type also be fullscreen here? If so, I would suggest adding it here or appending `, etc.` to the list of example types

ui/store/actions.js
```diff
@@ -3792,10 +3999,10 @@ export function addCustomNetwork(customRpc) {
 export function requestAddNetworkApproval(customRpc, originIsMetaMask) {
   return async (dispatch) => {
     try {
-      await promisifiedBackground.requestAddNetworkApproval(
+      await submitRequestToBackground('requestUserApproval', [
```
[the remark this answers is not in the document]
jpuri: Good point, thanks

app/scripts/background.js
```diff
@@ -96,7 +96,7 @@ const initApp = async (remotePort) => {
   log.info('MetaMask initialization complete.');
 };
 
-if (isManifestV3()) {
+if (isManifestV3) {
```
naugtur, line 99: It'd be nice to have a separate PR that turns it into a boolean, adds a unit test ensuring it's not turned back into a function and replace all usage, all in one step. So that when merged, there's no chance of a hard to find bug.

naugtur: Or at least add the test :)

ui/store/actions.js
```diff
@@ -51,9 +52,120 @@ import * as actionConstants from './actionConstants';
...
+  });
+};
+
+// Function below is used to remove action from queue on successful completion
+const removeFromActionQueue = (queue, actionId) => {
+  const index = queue.find((act) => act.actionId === actionId);
+  queue.splice(index, 1);
+};
+
+// Function below invokes promisifiedBackground method in MV2 content
```
naugtur, line 77:
```suggestion
// Function below invokes promisifiedBackground method in MV2 context
```

ui/store/actions.js
```diff
@@ -51,9 +52,120 @@ import * as actionConstants from './actionConstants';
...
+// In MV3 context the execution is:
+//   1. action is added to retry queue, along with resolve handler to be executed on completion of action
+//   2. is streams are connected promisifiedBackground method is called
+//   3. on completion of action either successfully or by throwing exception, action is removed from retry queue
```
naugtur, line 81:
```suggestion
// Function below invokes promisifiedBackground method in MV2 context
// In MV3 context the execution is:
//   1. action is added to retry queue, along with resolve handler to be executed on completion of action
//   2. if streams are connected promisifiedBackground method is called
//   3. on completion of action either successfully or by throwing exception, action is removed from retry queue
```

ui/store/actions.js
```diff
@@ -51,9 +52,120 @@ import * as actionConstants from './actionConstants';
...
+      );
+      if (background.connectionStream.readable) {
+        try {
+          promisifiedBackground[method](...args).then((result) => {
+            resolve(result);
+            removeFromActionQueue(promisifiedActionRetryQueue, actionId);
+          });
+        } catch (exp) {
+          removeFromActionQueue(promisifiedActionRetryQueue, actionId);
+          throw exp;
```
naugtur, line 103: using reject(exp) would be preferred and more performant

all errors within `new Promise(()=>{})` must be thrown synchronously or wired up into reject, so we need to be careful not to call out to functions that could have an asynchronous step.

This implementation, if `promisifiedBackground[method](...args)` gets rejected, will not catch it and cause an unhandled rejection.

ui/store/actions.js
```diff
@@ -51,9 +52,120 @@ import * as actionConstants from './actionConstants';
...
+    return new Promise((resolve) => {
+      addToActionQueue(
+        promisifiedActionRetryQueue,
+        actionId,
+        { method, args },
+        resolve,
+      );
+      if (background.connectionStream.readable) {
+        try {
+          promisifiedBackground[method](...args).then((result) => {
```
naugtur, line 97: all errors within `new Promise(()=>{})` must be thrown synchronously or wired up into reject, so we need to be careful not to call out to functions that could have an asynchronous step.

This implementation, if `promisifiedBackground[method](...args)` gets rejected, will not catch it and cause an unhandled rejection.

naugtur: I'll dare claim it'd be beneficial to extract the queue mechanics into a separate file and/or make it testable so that we can proceed to TDD the way forward. I can help with that if someone can wire things up and create a single test that covers this to help get me started.

ui/store/actions.js
jpuri: Thanks a lot of bringing up this, yep I need to refactor PR to use reject in case of exceptions.

ui/store/actions.js
```diff
@@ -51,9 +52,120 @@ import * as actionConstants from './actionConstants';
...
+  args = [],
+  actionId = new Date().getTime(),
+) => {
+  if (isManifestV3) {
+    return new Promise((resolve) => {
+      addToActionQueue(
+        promisifiedActionRetryQueue,
+        actionId,
+        { method, args },
+        resolve,
```
naugtur, line 93: both resolve and reject would have to be passed here.

ui/store/actions.js
naugtur: Also, you're not in an async function and await is not used, so a try-catch here will only work for synchronous errors

jpuri: I tried it again by throwing error from background method, the error was not returned to client side but the request was kind of lost as method did not resolved / reject.
Thus it seems exception handler will run only if error is on client side, re-throwing error in such scenario will be appropriate I think.

naugtur: it was an unhandled rejection. they're slent in the browser

app/scripts/background.js
jpuri: I removed the change I will create a separate PR for it.

ui/store/actions.js
jpuri: fixed

ui/store/action-queue/index.js
```diff
@@ -0,0 +1,135 @@
+import pify from 'pify';
```
naugtur, line 1: It doesn't look like we need the Proxy based heavy lifting this pkg is doing.

jpuri: I agree

ui/store/action-queue/index.test.js
```diff
@@ -0,0 +1,188 @@
...
+      submitRequestToBackground('backgroundFunction');
+      expect(background.backgroundFunction.called).toStrictEqual(true);
+    });
+
+    // it('resolves if backgroundFunction called resolves', async () => {
+    //   const background = {
+    //     connectionStream: {
+    //       readable: true,
+    //     },
+    //     backgroundFunction: async () => {
```
naugtur, line 74: this was supposed to be a callback style function - it's being wrapped in pify afterall. 
Hours spent here: 2 :D

ui/store/action-queue/index.js
jpuri: I agree with you point here, but I would prefer to create a separate task for removing `pify`.

ui/store/action-queue/index.test.js
```diff
@@ -0,0 +1,197 @@
+import sinon from 'sinon';
+
+import {
+  __TEST_CLEAR_QUEUE,
```
naugtur, line 4: This is not beautiful, but currently absolutely necessary.

A change to how we invoke tests could avoid this, but the next person to add a new test would stumble upon this issue of having a prepopulated queue again.

Another possibility - expose a factory function from queue and create a new queue for each test and only one queue in actual implementation.

I just got this idea. Will implement later. Unless someone beats me to it.

naugtur: This exposes a private function in public API to facilitate testing. 

The suggestion is just to make it possible to have more instances of action-queue but with a clean API. 

If you want to avoid changing implementation to facilitate tests, you want the latter (it;s a matter of wrapping almost the whole file in an init function that reveals the api)

jpuri:
Currently when clearing the queue we run action asynchronously, but I am thinking may be we should do it synchronous as they are in sequence of user interaction. Consider user updating gas params on transaction twice - we need to ensure that they run in sequence.

ui/store/actions.js
Ellajoke: Yeah please can you help me do that to work perfectly please

ui/store/action-queue/index.js
```diff
@@ -0,0 +1,134 @@
...
+//   3. on completion of action either successfully or by throwing exception, action is removed from retry queue
+export const submitRequestToBackground = (
+  method,
+  args = [],
+  actionId = Date.now() + Math.random(), // current date is not guaranteed to be unique
+) => {
+  if (isManifestV3()) {
+    return new Promise((resolve, reject) => {
+      addToActionQueue(actionId, { method, args }, resolve, reject);
+      drainActionRetryQueue();
```
jpuri, line 69: Can `drainActionRetryQueue` be part of `addToActionQueue` function itself ?

ui/store/action-queue/index.test.js
```diff
@@ -0,0 +1,196 @@
...
+    //   _setBackgroundConnection(background);
+    //   const mockFn = sinon.stub();
+    //   await new Promise((_, reject) => {
+    //     callBackgroundMethod('backgroundFunction', [], () => {
+    //       reject();
+    //     });
+    //   });
+    //   expect(mockFn.called).toStrictEqual(true);
+    // });
+  });
```
jpuri, line 195: Since `mockFn` is a callback with is run for both success and failure I think the last 2 test cases are same and this one is redundant.

ui/store/action-queue/index.test.js
```diff
@@ -0,0 +1,173 @@
...
+        backgroundFunction2: sinon.stub(),
+      };
+
+      _setBackgroundConnection(background);
+      submitRequestToBackground('backgroundFunction2');
+      expect(background.backgroundFunction2.called).toStrictEqual(false);
+    });
+
+    it('calls promisified background method on stream reconnection', () => {
+      // This is not an adequate simulation of what reconnection would look like
```
jpuri, line 52: Hey @naugtur : do you have a better suggestion for this ?

ui/store/action-queue/index.js
naugtur: yes. I wanted to do that and forgot.

naugtur:
> Currently when clearing the queue we run action asynchronously, but I am thinking may be we should do it synchronous as they are in sequence of user interaction. Consider user updating gas params on transaction twice - we need to ensure that they run in sequence.

I thought executing them quickly was a requirement. This makes implementation much nicer, may even help eliminate the need for the splice.

ui/store/action-queue/index.test.js
naugtur: It's all synchronous. I'd like a version of that where an action gets interrupted by a disconnect and the test also covers recovery and draining the queue.

ui/store/action-queue/index.test.js
naugtur: The other test (now removed) should be covering a failure case. It was still messed up because I had to go AFK before I managed to get Jest to check for a failure. I can't say I like Jest ;)

ui/store/action-queue/index.test.js
naugtur: I improved it but forgot to remove the comment :D

ui/store/action-queue/index.js
```diff
@@ -83,43 +84,53 @@ export const callBackgroundMethod = (
...
+            } catch (err) {
+              removeFromActionQueue(actionId);
+              reject(err);
+            }
+          }
+          return syncThenable;
+        });
+      },
+      syncThenable,
+    );
```
jpuri, line 133: great work 👍

ui/store/action-queue/index.test.js
jpuri: In this case we have single callback for both success / failure. We might still be able to test scenario when exception is thrown.

ui/store/action-queue/index.test.js
```diff
@@ -0,0 +1,275 @@
...
+      _setBackgroundConnection(background);
+      const scheduled = Promise.all([
+        submitRequestToBackground('first'),
+        submitRequestToBackground('second'),
+      ]);
+      background.connectionStream.readable = true;
+      _setBackgroundConnection(background);
+      await scheduled;
+      expect(trace.firstDone).toBeLessThanOrEqual(trace.secondStarted);
+    });
```
jpuri, line 127: Nice work with checking time

ui/store/action-queue/index.test.js
```diff
@@ -0,0 +1,275 @@
...
+        submitRequestToBackground('second'),
+      ]);
+      background.connectionStream.readable = true;
+      _setBackgroundConnection(background);
+      await scheduled;
+      await Promise.resolve('one more tick'); // One asynchronous tick to avoid depending on implementation details
+      expect(trace.firstDone).toStrictEqual(true);
+      expect(background.second.called).toStrictEqual(false);
+    });
+  });
```
jpuri, line 188: While testing I found that usually error is not thrown if connection is lost, just the action kind of hangs and there is no reply.

naugtur: I'll try to reproduce that too in a separate test

ui/store/action-queue/index.js
naugtur: A new test proves it wasn't :D 

But I really like the final version

app/scripts/lib/metaRPCClientFactory.js
```diff
@@ -47,6 +52,13 @@ class MetaRPCClient {
   close() {
     this.notificationChannel.removeAllListeners();
     this.uncaughtErrorChannel.removeAllListeners();
+    // fail all unfinished requests
+    for (const [id, handler] of this.requests) {
```
BelfordZ, line 56: we should PR this seperately - nice fixup

naugtur: We can't have the queue logic without it. I can split it out, and we merge it first.

ui/store/actions.js
naugtur: We can do retirement time, but running clearRetryQueue on an interval introduces high risk of rejecting something that was going to succeed otherwise.

clearRetryQueue is meant for recovery in catastrophic situations (like the UI being stuck and restarting)

app/scripts/lib/metaRPCClientFactory.js
naugtur: Check out https://github.com/MetaMask/metamask-extension/pull/15570

ui/store/action-queue/index.js
```diff
@@ -0,0 +1,200 @@
...
+ * @param {any} [actionId] - if an action with the === same id is submitted, it'll be ignored if already in queue.
+ */
+export const callBackgroundMethod = (
+  method,
+  args = [],
+  callback,
+  actionId = Date.now() + Math.random(), // current date is not guaranteed to be unique
+) => {
+  if (isManifestV3) {
+    const resolve = (value) => callback(null, value);
```
jpuri, line 118: @naugtur : here callbacks do not have standard arguments, that is first being error and second value. I think we should use `callback` as it is for reject and resolve

ui/store/action-queue/index.js
```diff
@@ -0,0 +1,201 @@
...
+async function executeAction({ action, disconnectSideeffect }) {
+  const {
+    actionId,
+    request: { method, args },
+    resolve,
+    reject,
+  } = action;
+  try {
+    resolve(await promisifiedBackground[method](...args, actionId));
+  } catch (err) {
```
jpuri, line 140: @naugtur , I have added actionId here as the last argument, it is useful to make action call idempotent in some scenario.

naugtur: It crosses abstraction boundaries. Now all `background` functions have to understand they're being used that way. 
- How far does it reach? Is it used in the RPC client or in the background worker?
- Why is it only added in one place and not the others? Seems like an oversight.

IMHO, if we want to add it, it'd need to be the first argument and be made mandatory, otherwise it'll hurt us later.

jpuri: I used it here https://github.com/MetaMask/metamask-extension/pull/15667/files#diff-675c437262d906486cf5436c0b64535347286cc4d1081dbd863fb45b8a01b81aR755

I kept it last argument so that methods deal with it only if they need to. If we do not want to pass it to all background methods, I can pass it specifically where needed.

naugtur: Intuition tells me it's a footgun.

Current change visible here sends it or doesn't send it depending on the current state - you only added it in the queue processing but it's not defined if connection was available. 
So it only guarantees idempotence across multiple retries, but not between the initial call and first retry

jpuri: I changed it to pass actionId only in scenario needed.

brad-decker:
Is this a feature branch? Is the idea here that all those idempotent prs are gonna collect here? Could we add the PR template here and fill it out with as much detail as possible

ui/store/action-queue/index.js
```diff
@@ -0,0 +1,200 @@
...
+ * @param {string} method - name of the background method
+ * @param {Array} [args] - arguments to that method, if any
+ * @param callback - Node style (error, result) callback for finishing the operation
+ * @param {any} [actionId] - if an action with the === same id is submitted, it'll be ignored if already in queue.
+ */
+export const callBackgroundMethod = (
+  method,
+  args = [],
+  callback,
+  actionId = Date.now() + Math.random(), // current date is not guaranteed to be unique
```
digiwand, line 115: I think I'm missing where we use this 4th parameter. Aside from the initial background calls, is it used elsewhere?

ui/store/action-queue/index.js
```diff
@@ -0,0 +1,200 @@
...
+ *
+ * @param {string} method - name of the background method
+ * @param {Array} [args] - arguments to that method, if any
+ * @param {any} [actionId] - if an action with the === same id is submitted, it'll be ignored if already in queue waiting for a retry.
+ * @returns {Promise}
+ */
+export function submitRequestToBackground(
+  method,
+  args = [],
+  actionId = Date.now() + Math.random(), // current date is not guaranteed to be unique
```
digiwand, line 84: Is this `actionId` param being used anywhere? Or can we move this random generator into the function itself?

ui/store/action-queue/index.js
```diff
@@ -0,0 +1,200 @@
...
+    return;
+  }
+  processingQueue = true;
+  try {
+    while (
+      background.connectionStream.readable &&
+      actionRetryQueue.length > 0
+    ) {
+      // If background disconnects and fails the action, the next one will not be taken off the queue.
+      // Retrying an action that failed because of connection loss while it was alreaedy ongoing is not supported.
```
digiwand, line 168:
```suggestion
      // Retrying an action that failed because of connection loss while it was processing is not supported.
```

jpuri:
> Is this a feature branch? Is the idea here that all those idempotent prs are gonna collect here? Could we add the PR template here and fill it out with as much detail as possible

Hey @brad-decker : the intend is not to make it feature branch, this is ready for merge. Other PRs are based on this as they need these changes to work.

ui/store/action-queue/index.js
jpuri: it is used to check if action is already present in queue

ui/store/action-queue/index.js
jpuri: yes following PRs are using it, plus this flexibility for functions to be able to provide their own actionId is useful

.storybook/preview.js
```diff
@@ -11,7 +11,7 @@ import MetaMetricsProviderStorybook from './metametrics';
 import testData from './test-data.js';
 import { Router } from 'react-router-dom';
 import { createBrowserHistory } from 'history';
-import { _setBackgroundConnection } from '../ui/store/actions';
+import { _setBackgroundConnection } from '../ui/store/action-queue';
```
segun, line 14: NIT: my brain sees underscore before a variable/method and it thinks `private` variable/method.

jpuri: I would avoid changing existing method in this PR.

ui/store/action-queue/index.js
digiwand: may I have an example of what file and what line this is being called in please? 🙏

naugtur: I'm not aware of it being used (as in calling the function with the 4th argument). We're currently only using the default value generated unique each time. But the capability is there to ensure only one copy of a specific type of action in retry queue. 

That being said, it's been preserved as a requirement from before I started working on this and I don't have an explanation why it was needed. Might no longer make sense. :shrug:

ui/store/action-queue/index.test.js
```diff
@@ -0,0 +1,358 @@
...
+    it('does not calls promisified background method if the stream is not connected', async () => {
+      const background = {
+        connectionStream: {
+          readable: false,
+        },
+        backgroundFunction2: sinon.stub().yields(),
+      };
+
+      _setBackgroundConnection(background);
+      submitRequestToBackground('backgroundFunction2');
```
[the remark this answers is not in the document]
jpuri: await is not required here

ui/store/action-queue/index.js
jpuri: Please check these PRs @digiwand :
https://github.com/MetaMask/metamask-extension/pull/15667
https://github.com/MetaMask/metamask-extension/pull/15675

ui/store/action-queue/index.test.js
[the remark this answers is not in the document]
naugtur: why? I dare to disagree

ui/store/action-queue/index.js
digiwand: I think the intention of having it makes sense to ensure we're not duplicating calls of an action instance in the off chance an action could fail. I'm wondering, why add it as a parameter instead of using a const in the function? 

The examples above look like they're only for `submitRequestToBackground` and not `callBackgroundMethod`

ui/store/action-queue/index.test.js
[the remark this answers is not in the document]
jpuri: This is scenario when stream is not connected, thus method is not called and is not resolved, request timeouts here if I write await.

This other test case is succeeding without await https://github.com/MetaMask/metamask-extension/pull/15337/files#diff-026267adeb51c0d42e0900bfc1f12e32bbb14e42fea7d7727dabdee5c5f4df3dR54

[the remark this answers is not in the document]
naugtur: Oh, it's the test checking the immediate results, not the one where connection comes back. How sis the await get here at all then?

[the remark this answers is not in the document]
naugtur: This test never gets into a connected state so there's no way the await would not timeout. That's fine. 

Be careful with toggling async behavior on and off in these tests, because most of them depend on that a lot and may no longer assert correct behavior.

ui/store/action-queue/index.js
```diff
@@ -0,0 +1,200 @@
...
+ */
+export function dropQueue(silently) {
+  if (!silently) {
+    failQueue();
+  }
+  actionRetryQueue.length = 0;
+}
+
+// add action to queue
+const executeActionOrAddtoRetryQueue = (item) => {
```
naugtur, line 54: nit  
```suggestion
const executeActionOrAddToRetryQueue = (item) => {
```

ui/store/action-queue/index.js
digiwand: I think the intention of having it makes sense to ensure we're not duplicating retried action calls in the off chance an action could fail. I'm wondering, why add it as a parameter instead of using a const in the function? 

I'm not seeing where `callBackgroundMethod` gets called with the 4th parameter in either #15667 or #15675. The examples above look like they're only for `submitRequestToBackground` and not `callBackgroundMethod`

naugtur: the callback-style syntax can hopefully go away over time.

Russell4115 (approved):

naugtur (approved):

darkwing (approved):
LGTM!  The only thing I saw that I questioned was the big comment block in `action-queue/index.js` ; do we need that?

This is great!

PR #15338, opened 2022-07-26
Fix for script to measure bundle size over time

jpuri:
Fix for script to measure bundle size over time


brad-decker (dismissed):

.circleci/scripts/bundle-stats-commit.sh
```diff
@@ -16,6 +16,12 @@ then
     exit 1
 fi
 
+if [[ "${CIRCLE_BRANCH}" != "develop" ]]
+then
+    printf 'This is not develop branch'
+    exit 0
+fi
+
```
jpuri, line 24: I have done this additional change in PR to move check for branch before check for `GITHUB_TOKEN` so that the check is only done on develop branch.

brad-decker (approved):

ryanml (approved):
lgtm

PR #15359, opened 2022-07-27
Performance improvements for Swaps

dan437:
## Explanation
This PR improves performance in Swaps, especially on the Build Quote page (~7.2x faster!)

I've used 2 main tools for performance improvements:
- https://github.com/welldone-software/why-did-you-render
- Performance tab in Google Chrome

## Screenshots
Before: Total Time: 3390 ms to render the BuildQuote component with token lists with 4x CPU slowdown
image

After: Total Time: 469 ms to render the BuildQuote component with token lists with 4x CPU slowdown
image

## Manual Testing Steps
- Run Swaps and see if everything is running smoothly now
- In Chrome's dev tools you can switch to the Performance tab and record a Build Quote page reload with 4x CPU slowdown (with and without this PR to compare results)


shared/modules/hexstring-utils.js
```diff
@@ -71,5 +71,5 @@ export function toChecksumHexAddress(address) {
     // closely mimics the original behavior.
     return hexPrefixed;
   }
-  return toChecksumAddress(addHexPrefix(address));
+  return toChecksumAddress(hexPrefixed);
```
dan437, line 74: If you scroll up, you will see that we already have this:
```
const hexPrefixed = addHexPrefix(address);
```

ui/components/ui/identicon/identicon.component.js
```diff
@@ -11,7 +13,7 @@ const getStyles = (diameter) => ({
   borderRadius: diameter / 2,
 });
 
-export default class Identicon extends PureComponent {
```
dan437, line 14: React.PureComponent’s shouldComponentUpdate() only shallowly compares the objects. If these contain complex data structures, it may produce false-negatives for deeper differences. 

That is exactly our case, because we pass a complex object(s) in props and it keeps re-rendering. I was getting a lot of warnings from this tool: https://github.com/welldone-software/why-did-you-render

ui/hooks/useTokensToSearch.js
```diff
@@ -40,28 +40,26 @@ export function getRenderableTokenData(
...
-      isSwapsDefaultTokenSymbol(symbol, chainId)
-        ? 1
-        : contractExchangeRates[toChecksumHexAddress(address)],
-      conversionRate,
-      currentCurrency,
-      string,
-      symbol,
-      false,
-    ) || '';
+  const rawFiat = formattedFiat
```
dan437, line 59: We don't need to make another expensive call to the `getTokenFiatAmount` fn, we can just remove non-numeric characters from the `formattedFiat` const.

ui/pages/swaps/swaps.util.js
```diff
@@ -252,14 +252,14 @@ const SWAP_GAS_PRICE_VALIDATOR = [
...
-function validateData(validators, object, urlUsed) {
+function validateData(validators, object, urlUsed, logError = true) {
   return validators.every(({ property, type, validator }) => {
     const types = type.split('|');
 
     const valid =
       types.some((_type) => typeof object[property] === _type) &&
       (!validator || validator(object[property]));
-    if (!valid) {
+    if (!valid && logError) {
```
dan437, line 262: We have a user story for this to only track errors here when it's useful.

adonesky1 (dismissed):
LGTM!

ui/hooks/useTokensToSearch.js
```diff
@@ -40,28 +40,26 @@ export function getRenderableTokenData(
...
+    ? // eslint-disable-next-line require-unicode-regexp
+      new BigNumber(formattedFiat.replace(/[^0-9.,]+/g, ''))
```
darkwing, line 61: Can we assign this regex to a `const` above, with a comment explanation?

ui/pages/swaps/view-quote/view-quote.js
```diff
@@ -963,7 +964,7 @@ export default function ViewQuote() {
             : t('swap')
         }
         hideCancel
-        disabled={isSwapButtonDisabled}
+        disabled={Boolean(isSwapButtonDisabled)}
```
darkwing, line 967: Can we assign this to a boolean where it's first defined?

darkwing (changes requested):

darkwing (approved):

ui/hooks/useTokensToSearch.js
```diff
@@ -40,28 +40,32 @@ export function getRenderableTokenData(
   useTokenDetection,
 ) {
   const { symbol, name, address, iconUrl, string, balance, decimals } = token;
+  let contractExchangeRate;
+  if (isSwapsDefaultTokenSymbol(symbol, chainId)) {
+    contractExchangeRate = 1;
+  } else if (string && conversionRate > 0) {
```
adonesky1, line 46: Why do we need to check for the "string" value here?

dan437: Because `string` is also passed into the `getTokenFiatAmount` fn as `tokenAmount` and if it's undefined, then it exits early. Basically, we only want to do the expensive `toChecksumHexAddress` call if it will be used later on.

adonesky1: Ah ok. Maybe worth adding a comment about that here, but I won't block on it!

adonesky1 (approved):
LGTM!

PR #8595, opened 2020-05-14
[RFC] add prettier to eslint

brad-decker:
### rationale
Eslint helps to make sure that everyone follows convention, but it doesn't take an opinionated approach to do so. Prettier has some stringent rules that they are rather inflexible on, but in my experience, it results in code formatting. that is universal and enforced on save. It, by default, attempts to keep code fitting within the average width of a half screen code window (80 col) but also maximizes the amount of code per line with the legibility as its primary concern.

~The following PR has been split up into commits to make the diff manageable.~  Looking for feedback on how we can modify this setup to meet the needs of the entire team.

Link to `.prettierrc.yml` in feature branch: https://github.com/MetaMask/metamask-extension/blob/prettier/.prettierrc.yml

@rekmarks pointed out in slack the changes to the eslint config, copy pasted:

Disabled/attenuated:
indent
no-mixed-operators
no-confusing-arrow
react/jsx-closing-tag-location
react/jsx-wrap-multilines
Added:
react/self-closing-comp

**NEW RULE UPDATES AFTER REBASE**
disabled: wrap-regex
attenuated: lines-around-comment (prettier has some rules around this that conflict with the default eslint rule config)

**Strategy for merging**
Once we all agree on the changes and the settings, I can split this into multiple PRs using the following strategy:

For the first PR add prettier, eslint-prettier plugin, and update rules, use overrides by folder path to disable the prettier rules, and revert the changes to the eslint rules that were required to the code. So this should result in only .eslintignore, .eslintrc, package.json and yarn.lock changing.

I can then have successive PRs for each subpath of the application so that the changes can be reviewed in chunks and if we find anything I can pause the progress of merging PRs, fix the eslint config and update any issues resulting from that in previously merged work.

The final PR would remove the overrides altogether. 

I think there would be additional steps to lift this configuration to the eslint-config package and distribute prettier with it, then we'd need to gradually adopt it in all of the other repositories. If that's a thing we want to do after using it in extension codebase.

PR #15360, opened 2022-07-27
update prettier

brad-decker:
## Explanation
Updates prettier to latest version to take advantage of performance and legibility improvements. The key difference is how prettier breaks long lines up. See before / after sections for examples. Prettier now puts a higher importance on breaking after operators in assignments, so whereas previously it might add line breaks to a `const { x, y, z } = foo;` destructuring assignment such that x, y and z are all on new lines, now it'll break after the `=` and indent the next line.  

### Before & After

image
image
image

## Pre-Merge Checklist


darkwing (dismissed):
OMG I love this!

IMO, in the future, with these big PRs, we should do 2 commits -- the first with the prettier update, the second with all of the subsequent file updates.  Makes reviewing a bit easier.

brad-decker:
Good call, i did that on the eslint pr. i'm happy to break this up into two commits soon for future reviewers.

brad-decker:
@darkwing split the commits :)

darkwing (approved):

adonesky1 (approved):
LGTM! I didn't thoroughly vet every single line if I'm being honest... but scanned for obvious issues.

BelfordZ:
There are some cases where the output seems less than ideal:

image


short lhs of assignment, long rhs -> make rhs shorter by line breaking on the first `.` would be IMO prefered way to handle this case

brad-decker:
Prettier optimizes lines of code based on the maximum line length, 80. With few exceptions there is only one way to modify its output: modifying the max line length. I am sure we want the update to the latest version, especially considering we want to be able to take advantage of performance improvements as well as future updates. If we want to discuss the output of prettier as a whole we can do that outside of this pull request as it will involve every aspect of its output versus just the way it breaks lines for assignment operations.

brad-decker:
Marking as Do Not Merge just so i can get some clarification on @BelfordZ 's comment and see if he's comfortable pushing this forward and discussing prettier output elsewhere

BelfordZ:
@brad-decker thanks for the reply. I deleted my comment as I realized we were simply updating, not changing anything from the default settings. My B! 

default settings FTW -- Might just take some getting used to.

brad-decker:
@BelfordZ -- Questions about our defaults and standards are welcome! I'd like to preserve the intent of your original message so I'm just gonna paraphrase without specifics:

@BelfordZ was commenting on *where* prettier chooses to break lines, and suggested that a specific way would be more legible to him. 

There were a number of back-and-forth conversations at the time we elected to use prettier, most of which centered around this concept. In the end we opted to just use the default settings because changing line length resulted in very much unwanted formatting. https://github.com/MetaMask/metamask-extension/pull/8595 is the original PR.

PR #15383, opened 2022-07-29
Add data about dropped transaction in FINALIZED metrics event

### LINKED ISSUE #15269: Add data about dropped transaction in FINALIZED metrics event
Based on this comment https://github.com/MetaMask/metamask-extension/issues/15150#issuecomment-1184366925

If there is a `TRANSACTION_EVENTS.FINALIZED` event for a dropped transaction, we should track a property `transaction_replaced` that can be one of `cancel` or `retry` or `other`

Note: "replaced" and "dropped" sort of mean the same thing

To do this:
- pass a 3rd property from `_dropTransaction` to indicate that we are tracking an event for a dropped transaction
- the hash of the transaction that replaced this transaction can be found on the `txMeta.replacedBy` property
- we need to add a internal method to the transaction controller to get a tx by hash, therefore we will also have to add a `getTransactionByHash` method to `tx-state-manager.js`, which just needs to receive 1 parameter (txHash) and find the transaction with the `txHash` property that matches that parameter
- once we get the txMeta of the transaction that is replacing the dropped transaction, then we can determine `cancel` or `retry` or `other` using the type information on the transaction (if it doesn't match cancel or retry then it is `other`

VSaric:
## Explanation

Based on this comment https://github.com/MetaMask/metamask-extension/issues/15150#issuecomment-1184366925, if there is a `TRANSACTION_EVENTS.FINALIZED` event for a dropped transaction, we should track a property `transaction_replaced` that can be one of `cancel` or `retry` or `other`.

 We modified the `_markNonceDuplicatesDropped` function to, in addition to adding the `replacedBy` key, adding a `replacedById` key that records the internal metamask id of the tx that replaced it. Once we get the `txMeta` of the transaction that is replacing the dropped transaction, then we can determine `cancel` or `retry` or `other` using the `type` information on the transaction (if it doesn't match `cancel` or `retry` then it is `other`).

## More Information

* Fixes #15269


app/scripts/controllers/transactions/index.js
```diff
@@ -376,6 +376,22 @@ export default class TransactionController extends EventEmitter {
...
+    const { transactions } = this.store.getState();
+    const txByHash = Object.values(transactions).filter((tx) => {
+      if (tx.hash && txHash && tx.hash === txHash) {
+        return tx;
+      }
+      return undefined;
+    });
+    return txByHash[0];
+  }
+
```
brad-decker, line 394: Instead of doing this, which is potentially 50 iterations of logic for every metric event related to retries, lets modify the `_markNonceDuplicatesDropped` function to, in addition to adding the replacedBy key, adding a replacedById key that records the internal metamask id of the tx that replaced it. This will give us an easier way to quickly look it up.

brad-decker: I saw this logic was laid out by @danjm , so tagging him to see if i'm being egregious with my recommendation here.

app/scripts/controllers/transactions/index.js
```diff
@@ -2110,6 +2127,18 @@ export default class TransactionController extends EventEmitter {
       transactionContractMethod = transactions[id]?.contractMethodName;
     }
 
+    const replacedTxMeta = this._getTransactionByHash(replacedBy);
+
+    let transactionReplaced;
+    if (extraParams?.dropped) {
+      transactionReplaced = TRANSACTION_TYPES.OTHER;
```
brad-decker, line 2134: I don't like the addition of this 'OTHER' transaction_type constant. What I would prefer is creating a new constant:

```js
const TRANSACTION_REPLACEMENT_METHODS = {
  RETRY: TRANSACTION_TYPES.RETRY,
  CANCEL: TRANSACTION_TYPES.CANCEL,
  SAME_NONCE: 'other',
};
```

Using a different constant reduces the chances of confusing future contributors who might think that 'other' is a viable transaction type when it is only there for metrics.

app/scripts/controllers/transactions/tx-state-manager.js
```diff
@@ -297,6 +297,22 @@ export default class TransactionStateManager extends EventEmitter {
...
+    const { transactions } = this.store.getState();
+    const txByHash = Object.values(transactions).filter((tx) => {
+      if (txHash && tx.hash && tx.hash === txHash) {
+        return tx;
+      }
+      return undefined;
+    });
+    return txByHash[0];
+  }
+
```
brad-decker, line 315: Regardless of @danjm 's feedback on whether we should add the replacedById key versus using the hash, this method appears to be unused?

brad-decker (changes requested):

app/scripts/controllers/transactions/tx-state-manager.js
VSaric: @brad-decker You are right, this method is unused, I thought that I can use this method in transaction controller where I created the same method `getTransactionByHash` like this:
`_getTransactionByHash(txHash) {
    return this.txStateManager.getTransactionByHash(txHash)
  }` but I wasn't sure about that. 
What do you think about this?

app/scripts/controllers/transactions/index.js
VSaric: I agree with you definitely, I will create a new constant `TRANSACTION_REPLACEMENT_METHODS`! Thank you for this good suggestion @brad-decker!

app/scripts/controllers/transactions/index.js
VSaric: We had a meeting with @danjm today and we agreed with you @brad-decker about this suggestion. Done [here](https://github.com/MetaMask/metamask-extension/pull/15383/commits/1c9525d3f99a0d7e073e759aa96624b20fc8c209).

app/scripts/controllers/transactions/index.js
VSaric: @brad-decker Done [here](https://github.com/MetaMask/metamask-extension/pull/15383/commits/1c9525d3f99a0d7e073e759aa96624b20fc8c209).

brad-decker (approved):

app/scripts/controllers/transactions/tx-state-manager.js
VSaric: @brad-decker Done [here](https://github.com/MetaMask/metamask-extension/pull/15383/commits/1c9525d3f99a0d7e073e759aa96624b20fc8c209).

digiwand (approved):

PR #15385, opened 2022-07-29
fix: patch borc to avoid v8 bailing out of caching bytecode for the entire bundle chunk

naugtur:
Here's a comparison of tracing results from reviving the mv3 service worker (starting it after it's been running and stopped) before and after removing the 'use asm' line.
compile-before-after

The gain on my machine is a bit under 100ms out of about 1 second, but chunks can get split differently or we might decide to increase the chunk size and both of those situations may increase the impact of this.


brad-decker:
Love the branch name. What does the 'use asm' line actually do?

brad-decker (approved):

kumavis:
> What does the 'use asm' line actually do?

@brad-decker  javascript subset intended to be a compile target that can be interpreted faster. evolved into wasm

kumavis (approved):

kumavis (approved):

PR #15386, opened 2022-07-29
Replace web3js package with ethersjs package

### LINKED ISSUE #15231: Replace imports of web3.js in `app/` and `shared/` with ethers.js
Part of https://github.com/MetaMask/metamask-extension/issues/14930

We don't need to import both web3.js and ethers. Let's replace all imports of web3.js with ethers. Functionaltiy of the code and product should stay the same with this change.

amerkadicE:
## Explanation
We don't need to import both web3.js and ethers.
This PR replaces all imports of web3.js with ethers.

## More Information

 Fixes #15231

## Manual Testing Steps

There are two places where we use web3.js.

1. When we first load MetaMask to load balance from the blockchain.
2. When we are detecting tokens with the Refresh list button.

To test the first case, load MetaMask and make sure that the correct balance is displayed.
To test the second case, switch to Ethereum Mainnet (it works only on mainnet) and click the Refresh list button. 

## Pre-Merge Checklist


lavamoat/browserify/beta/policy.json
```diff
@@ -2811,6 +2811,22 @@
         "crypto.getRandomValues": true
       }
     },
+    "@metamask/controllers>web3": {
+      "globals": {
+        "Web3": "write",
+        "XMLHttpRequest": "write",
```
PeterYinusa, line 2817: @amerkadicE @mirjanaKukic: I'm wondering if this may be the issue.
Changing `"XMLHttpRequest": "write",` to `"XMLHttpRequest": true,` seems to fix the tests for me. I'm not sure how it got changed to "write" based on its previous value.

amerkadicE: Thank you @PeterYinusa . Tests are passing on the pipeline as well with this fix.

app/scripts/lib/account-tracker.js
```diff
@@ -336,26 +335,29 @@ export default class AccountTracker {
...
-    ethContract.balances(addresses, ethBalance, (error, result) => {
-      if (error) {
-        log.warn(
-          `MetaMask - Account Tracker single call balance fetch failed`,
-          error,
-        );
-        Promise.all(addresses.map(this._updateAccount.bind(this)));
-        return;
-      }
+    this.ethersProvider = new ethers.providers.Web3Provider(this._provider);
```
adonesky1, line 338: Why do we need to set this again? Isn't it already set on line 87? I see that you're replicating the logic from the web3 implementation... I will try without this line included? I can understand if we need to update the provider when the network changes, but in this current implementation `this._provider` doesn't appear to change...

app/scripts/controllers/detect-tokens.js
```diff
@@ -112,7 +112,9 @@ export default class DetectTokensController {
       : tokenList;
 
     const tokensToDetect = [];
-    this.web3.setProvider(this._network._provider);
+    this.ethersProvider = new ethers.providers.Web3Provider(
```
adonesky1, line 115: this also appears to be redundant since `this.ethersProvider` is updated whenever `this._network` is updated...

amerkadicE: Hi @adonesky1 , you are probably right. I just wanted to make sure all logic from web3 implementation is there. 
I think we can remove those 2 lines.

app/scripts/lib/account-tracker.js
adonesky1: hmmm actually hold up on this. Looking through it and seeing whether this is indeed necessary.

adonesky1: Ok so it does appear to be necessary in this case... I'm just not yet sure how `this._provider` is getting updated without getting reset

app/scripts/controllers/detect-tokens.js
adonesky1: So I will double check but I think I'm actually wrong on this one per my comment [here](https://github.com/MetaMask/metamask-extension/pull/15386#discussion_r949379852)

adonesky1: Wait now I'm confused because I'm not seeing how `this.ethersProvider` is used at all on this controller...?

adonesky1: @amerkadicE do you see something I'm missing?

adonesky1: Looks like it should have been removed in this [PR](https://github.com/MetaMask/metamask-extension/commit/6e5c2f03bfbc4a0c8fa334b7beedb19cf35c9d73#diff-323d0cf46498be3850b971474905354ea5ccf7fa13745ad1e6eba59c5b586830L162)

adonesky1: I think its safe to remove `this.ethersProvider` from this controller

app/scripts/lib/account-tracker.js
adonesky1: But this change should stay @amerkadicE

app/scripts/controllers/detect-tokens.js
amerkadicE: Done

app/scripts/lib/account-tracker.js
```diff
@@ -336,26 +335,29 @@ export default class AccountTracker {
...
+
+    const ethContract = await new ethers.Contract(
+      deployedContractAddress,
+      SINGLE_CALL_BALANCES_ABI,
+      this.ethersProvider,
+    );
+    const ethBalance = ['0x0000000000000000000000000000000000000000'];
```
adonesky1, line 345: why so many additional `0`s here?

amerkadicE: Ethers packge won’t work with 0x0 as zero adress. 
This is soultion that @danjm suggested to me.

adonesky1: 👍 ethers has this in as a constant in `@ethersproject/constants` but looks like we don't use it anywhere else so lets not just pull it in for this.

adonesky1 (dismissed):

brad-decker (dismissed):

brad-decker:
@amerkadicE we need to fix the conflicts here then I can reapprove

amerkadicE:
@brad-decker @adonesky1 conflicts resolved, you can reapprove

adonesky1 (approved):

brad-decker (approved):

PR #14902, opened 2022-06-08
add fsevents policies to policy overrides to avoid policy generation diffs across different operating systems

adonesky1:
@kumavis ✌️

PR #15387, opened 2022-07-29
fix storybook

brad-decker:
## Explanation
in #15293 I added a resolution to acorn to 7.4.1. The reason I had to do this is because with the changes to dependencies the version of acorn that was placed in the root node_modules folder changed from 7.4.1 to 6.4.2. Although I attempted to apply the patch that had existed for 7.4.1 to 6.4.2 the build script failed due to something going on in the abi-decoder truffle library added in #12881. The reason this happened is that the version of acorn that browserify uses when attempting to parse our files is still 7.4.1 but the location in node_modules changed. 

Unfortunately setting the resolution to 7.4.1 worked for getting our build setup, but ended up causing the way that webpack builds storybook to get messed up. Storybook builds were broken by #15293.

This PR:
1. Removes the resolution for acorn.
2. Removes the 7.4.1 acorn patch
3. Adds the same patch but for the nested acorn path that browserify uses.

Result: Working builds and storybook (hopefully, 🤞🏻, it works on my machine at least ;p) 

## Manual Testing Steps
1. Pull this down.
2. `rm -rf node_modules`
3. `yarn setup`
4. `yarn lavamoat:auto` (you might get a diff, just ignore it, we are working on resolving OS policy differences here: #14902  )
5. `yarn storybook`

If step 4 compeletes successfully you have a working build. If step 5 loads storybook (and not a white screen) then storybook works. 

## Pre-Merge Checklist


jpuri (approved):

georgewrmarshall (approved):
I'm not experienced with lavamoat and to be honest I don't fully understand these code changes to the build system policy or acorn patch but I have tested on local using the manual testing steps and the extension and storybook are working.

PR #15010, opened 2022-06-22
Add setApprovalForAll confirmation view

adonesky1:
Fixes: #15070
I haven't had time or any design feedback to create a wholly separate flow specific to `setApproveForAll`, rather I've further parameterized the `confirm-approve` component to provide specific verbiage for this method (both granting and revoking approval). This is not the way I would do it with more time, but the timeline has been compressed and [there is some urgency to get something out there](https://consensys.slack.com/archives/GTQAGKY5V/p1656003466814319?thread_ts=1655303420.735759&cid=GTQAGKY5V) since this method is so commonly used.

cc @rachelcope and/or @SayaGT we may want to tweak the verbiage here.

To Do after this PR:
- [ ] E2E test

`setApproveForAll` (when granting permission)
**Before:**
Screen Shot 2022-06-28 at 4 35 39 PM

**After:**
Screen Shot 2022-06-28 at 4 05 34 PM
Screen Shot 2022-06-28 at 4 04 57 PM

`setApproveForAll` (when revoking permission):

**Before:**
Screen Shot 2022-06-28 at 4 36 05 PM

**After:**
Screen Shot 2022-06-28 at 4 25 50 PM
Screen Shot 2022-06-28 at 4 26 24 PM

**How to test:**
1. Go to etherscan and navigate to any ERC721 or ERC1155 contract
2. Navigate to the `contract` tab and then `write contract`:
Screen Shot 2022-06-28 at 4 37 02 PM
 3. Connect your wallet by clicking `Connect to Web3`
Screen Shot 2022-06-28 at 4 38 08 PM
4. Scroll down to `setApprovalForAll` and click it so it drops down inputs
5. Enter args: For `operator` enter another address (any will do) and for `approved` enter true to see the grant permission flow. (then repeat and enter false to see the revoke flow)

PR #15406, opened 2022-08-01
Enhance approval screen title logic

### LINKED ISSUE #15377: [Bug]:  How can I know which NFT(contract) I am approving in the new confirm panel?
### Describe the bug

Can not know which contract I am interacting with. That's terrible!

### Steps to reproduce

1. trigger `setApproveForAll` function of NFT contracts

_No response_

### Version

10.18.0

### Build type

_No response_

### Browser

Chrome

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

adonesky1:
Fixes: https://github.com/MetaMask/metamask-extension/issues/15377

A minor refactor/enhancement of some of the logic introduced in the [`setApprovalForAll` screen PR](https://github.com/MetaMask/metamask-extension/pull/15010).

Changes:
1. I realized it probably doesn't make sense to only make the title contain a link to the blockExplorer if we don't know the name or symbol of the token, so this makes the name, symbol, or generic "NFT" or "Token" title a link if there is a block explorer available:
          -**setApprovalForAll** : https://user-images.githubusercontent.com/34557516/182232491-e3b9cd7f-6ff7-4acc-8608-f28c977266e0.mov
          -**approve**: https://user-images.githubusercontent.com/34557516/182249887-7d72b514-42e0-4597-b5ea-c136bf4a9c95.mov
2. Fixes a bug where for NFTs (ERC721 or ERC1155) we showed the title (the name, symbol or "NFT") as a link even when we didn't have a block explorer available - meaning the link didn't work.
Before:
https://user-images.githubusercontent.com/34557516/182234026-910793c6-0a75-4ce6-81d8-d1d692e195d6.mov
After:
https://user-images.githubusercontent.com/34557516/182248156-b8852ccb-f3a0-4901-9c42-f111d999c6b5.mov

3. In the case where no block explorer is available, makes the contract address copy to the clipboard when the name/symbol/"NFT" or "Token" label is clicked in the title (cc @SayaGT @rachelcope, not sure if there is some better pattern for making it clear this will copy, tried adding the copy icon, but it looked wrong in this context) :
https://user-images.githubusercontent.com/34557516/182231811-abf183ea-1c0b-4e54-8983-ffe84b19826b.mov

4. Makes the contract address show on hover over the name/symbol/"NFT" or "Token" label in the title (whether or not there is a block explorer link). (Depicted in videos above)

### DIFF
--- a/ui/pages/confirm-approve/confirm-approve-content/confirm-approve-content.component.js
+++ b/ui/pages/confirm-approve/confirm-approve-content/confirm-approve-content.component.js
@@ -471,6 +471,7 @@ export default class ConfirmApproveContent extends Component {
       [...TEST_CHAINS, MAINNET_CHAIN_ID].includes(chainId);
 
     let titleTokenDescription = t('token');
+    const tokenIdWrapped = tokenId ? ` (#${tokenId})` : '';
     if (
       assetStandard === ERC20 ||
       (tokenSymbol && !tokenId && !isSetApproveForAll)
@@ -483,9 +484,8 @@ export default class ConfirmApproveContent extends Component {
       (assetName && tokenId) ||
       (tokenSymbol && tokenId)
     ) {
-      const tokenIdWrapped = tokenId ? ` (#${tokenId})` : '';
       if (assetName || tokenSymbol) {
-        titleTokenDescription = `${assetName ?? tokenSymbol}${tokenIdWrapped}`;
+        titleTokenDescription = `${assetName ?? tokenSymbol}`;
       } else {
         titleTokenDescription = t('nft');
       }
@@ -519,15 +519,18 @@ export default class ConfirmApproveContent extends Component {
     }
 
     return (
-      <span
-        className="confirm-approve-content__approval-asset-title"
-        onClick={() => {
-          copyToClipboard(tokenAddress);
-        }}
-        title={tokenAddress}
-      >
-        {titleTokenDescription}
-      </span>
+      <>
+        <span
+          className="confirm-approve-content__approval-asset-title"
+          onClick={() => {
+            copyToClipboard(tokenAddress);
+          }}
+          title={tokenAddress}
+        >
+          {titleTokenDescription}
+        </span>
+        {tokenIdWrapped && <span>{tokenIdWrapped}</span>}
+      </>
     );
   }
 
--- a/ui/pages/confirm-approve/confirm-approve-content/confirm-approve-content.component.js
+++ b/ui/pages/confirm-approve/confirm-approve-content/confirm-approve-content.component.js
@@ -33,6 +33,10 @@ import {
   ERC20,
   ERC721,
 } from '../../../../shared/constants/transaction';
+import {
+  MAINNET_CHAIN_ID,
+  TEST_CHAINS,
+} from '../../../../shared/constants/network';
 
 export default class ConfirmApproveContent extends Component {
   static contextTypes = {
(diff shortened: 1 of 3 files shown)


bschorchit:
This is awesome, Alex ❤️

3. We have this (screenshot) pattern of telling the users that they can click to copy to clipboard, but just the other day a team member shared he didn't know that was a feature so many users might not be aware of it too and it might not be worth to spend time on this. 
Screen Shot 2022-08-01 at 18 27 56
 
In anyway, I wonder if in the case we're displaying "NFT" if a lot of users might not think of hovering over it since it looks like a word within a text (what usually doesn't carry any tooltip). As a fix for this, could we display by default under Permission request and below the "Approved asset" line a new line "Asset contract: 0x..."?

adonesky1:
> As a fix for this, could we display by default under Permission request and below the "Approved asset" line a new line "Asset contract: 0x..."?

I'm all for this, I've suggested something like this to design before but have generally been told that we're going to be moving on to new designs for this screen soon and that it's already too busy, so most of my efforts have been to try and figure out ways to make it work without adding more text...

danjm (dismissed):
I'm good with this from a code perspective if @bschorchit is good with this from a product/UX perspective

bschorchit:
> I'm all for this, I've suggested something like this to design before but have generally been told that we're going to be moving on to new designs for this screen soon and that it's already too busy, so most of my efforts have been to try and figure out ways to make it work without adding more text...

Yes, I understand this push back. I'm good with the proposed solution from the product perspective :)
We'll be indeed further working on this screen in the close future.

ui/pages/confirm-approve/confirm-approve-content/confirm-approve-content.component.js
```diff
@@ -512,24 +500,35 @@ export default class ConfirmApproveContent extends Component {
...
-        titleTokenDescription = unknownNFTLink;
+        return blockExplorerElement;
       }
     }
 
-    return titleTokenDescription;
+    return (
+      <span
+        className="confirm-approve-content__approval-asset-title"
+        onClick={() => {
```
darkwing, line 524: This wouldn't be accessible from a keyboard perspective; could this be a `button`?

ui/pages/confirm-approve/confirm-approve-content/index.scss
```diff
@@ -9,10 +9,14 @@
     padding: 0 24px 16px 24px;
   }
 
-  &__unknown-asset {
+  &__approval-asset-link {
     color: var(--color-primary-default);
   }
 
+  &__approval-asset-title {
+    cursor: pointer;
```
digiwand, line 17: for case #2, when the title does not have a link, can we disable the pointer?

digiwand, line 17: for case # 2, when the title does not have a link, can we disable the pointer?

adonesky1: My thinking hear was to make the fact that you can click to copy the address (when there isn't a link) somewhat apparent. Probably could/should do the on-hover dropdown as well that Barbara highlights [here](https://github.com/MetaMask/metamask-extension/pull/15406#issuecomment-1201750793)

adonesky1: Implementing the drop down is going to be a real pain... as noted in the comment linked above we are avoiding putting to much into this since we are going to implement a whole new design very soon?

brad-decker (approved):

ui/pages/confirm-approve/confirm-approve-content/index.scss
digiwand: oh okay. I was confused by the various cases listed above. Sounds good to me to leave the pointer in and revisit this with the new redesign 👍

ui/pages/confirm-approve/confirm-approve-content/confirm-approve-content.component.js
```diff
@@ -519,15 +519,18 @@ export default class ConfirmApproveContent extends Component {
...
+        <span
+          className="confirm-approve-content__approval-asset-title"
+          onClick={() => {
+            copyToClipboard(tokenAddress);
+          }}
+          title={tokenAddress}
+        >
+          {titleTokenDescription}
+        </span>
+        {tokenIdWrapped && <span>{tokenIdWrapped}</span>}
```
digiwand, line 532: I think we want to check for the length of `tokenIdWrapped` here since it will always be a string

adonesky1: an empty string (as `tokenIdWrapped` is set if `tokenId` is undefined) is falsy though so this won't be rendered.

digiwand (approved):

ui/pages/confirm-approve/confirm-approve-content/confirm-approve-content.component.js
digiwand: oops, yes you're right

PR #15420, opened 2022-08-02
fix issues relating to race conditions where draftTx does not exist

### LINKED ISSUE #15401: undefined error when trying to read draftTransaction.amount in validateSendState when cancelling from send flow
https://sentry.io/organizations/metamask/issues/3459007904/?project=273505

### LINKED ISSUE #15403: undefined error when trying to read draftTransaction.gas.gasLimit upon initialization of send state or upon transition to confirm screen
https://sentry.io/organizations/metamask/issues/3460175255/?project=273505

Repro steps:
1. start a send
2. Change network.
3. QUICKLY cancel send
4. see error in your console.

Whats going on is that when changing networks we have to re-initialize the send state and depending on your timing you will either cancel the tx *before* the draft transaction object is retrieved from the state tree, in which case you get this error, or you manage to do it *after* and then the error you get will be #15401 or another error.

brad-decker:
## Explanation
This PR adds additional safety to the send duck when draftTransaction is not present. It does this in a few ways:
1. Only attempts to update draftTransaction.status if a draftTransaction exists. To make sure we reflect the appropriate status we also return 'true' for isSendFormInvalid when draftTransaction does not exist.
2. In the case for 'initializeSendState.fulfilled' we only update draftTransaction values if it exists, but the very possibility that we hit this `fulfilled` condition without a valid draftTransaction implies that we have a race condition, so...
3. An additional failsafe is added to the end of the initialize thunk that re-checks state to see if the draftTransaction has changed. This check happens after the async methods that are awaited in this initialize method. 

## More Information
Fixes #15403 
Fixes #15401

## Pre-Merge Checklist

### DIFF
--- a/ui/ducks/send/send.js
+++ b/ui/ducks/send/send.js
@@ -603,7 +603,7 @@ export const initializeSendState = createAsyncThunk(
     // For instance, in the actions.js file we dispatch this action anytime the
     // chain changes.
     if (!draftTransaction) {
-      thunkApi.rejectWithValue(
+      return thunkApi.rejectWithValue(
         'draftTransaction not found, possibly not on send flow',
       );
     }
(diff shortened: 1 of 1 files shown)


ui/ducks/send/send.js
```diff
@@ -603,7 +603,7 @@ export const initializeSendState = createAsyncThunk(
     // For instance, in the actions.js file we dispatch this action anytime the
     // chain changes.
     if (!draftTransaction) {
-      thunkApi.rejectWithValue(
+      return thunkApi.rejectWithValue(
         'draftTransaction not found, possibly not on send flow',
       );
```
brad-decker, line 608: without returning this the rest of the code in this slice would run, this is unwanted. Returning here prevents attempts to access draftTransaction erroring out further in this code. Although in this case the error would be handled by redux and would not bubble up to sentry.

ui/ducks/send/send.js
```diff
@@ -1396,28 +1396,30 @@ const slice = createSlice({
...
+            break;
+          case draftTransaction.recipient.warning === 'loading':
+          case draftTransaction.recipient.warning ===
+            KNOWN_RECIPIENT_ADDRESS_WARNING &&
+            draftTransaction.recipient.recipientWarningAcknowledged === false:
+            draftTransaction.status = SEND_STATUSES.INVALID;
+            break;
+          default:
+            draftTransaction.status = SEND_STATUSES.VALID;
+        }
```
brad-decker, line 1422: Only change the validity if the draftTransaction exists.

ui/ducks/send/send.js
```diff
@@ -2641,7 +2645,11 @@ export function isSendStateInitialized(state) {
  * @type {Selector<boolean>}
  */
 export function isSendFormInvalid(state) {
-  return getCurrentDraftTransaction(state).status === SEND_STATUSES.INVALID;
+  const draftTransaction = getCurrentDraftTransaction(state);
+  if (!draftTransaction) {
+    return true;
+  }
+  return draftTransaction.status === SEND_STATUSES.INVALID;
```
brad-decker, line 2652: If the draftTransaction does not exist, its invalid.

ui/ducks/send/send.js
```diff
@@ -1512,26 +1514,28 @@ const slice = createSlice({
...
+            // the network from the network dropdown, then the selected asset is
+            // no longer valid and should be set to the native asset for the
+            // network.
+            draftTransaction.asset.type = ASSET_TYPES.NATIVE;
+            draftTransaction.asset.balance =
+              draftTransaction.fromAccount?.balance ??
+              state.selectedAccount.balance;
+            draftTransaction.asset.details = null;
+          }
+        }
```
brad-decker, line 1531: So the only way this code should be hit is if there was an inflight attempt to initialize the send state, which at the time had a valid draftTransaction, and then the send flow is reset *between when the initialize is fulfilled and when redux runs this case* the only way this could be done (to my knowledge) is by changing networks in the send flow and canceling and hope you get lucky to time it right?

ui/ducks/send/send.js
```diff
@@ -678,6 +678,20 @@ export const initializeSendState = createAsyncThunk(
...
+    // a currentTransactionUUID exists and matches the previous tx.
+    const newState = thunkApi.getState();
+    if (
+      newState.send.currentTransactionUUID !== sendState.currentTransactionUUID
+    ) {
+      return thunkApi.rejectWithValue(
+        `draftTransaction changed during initialization.
+        A new initializeSendState action must be dispatched.`,
+      );
+    }
```
brad-decker, line 693: This should hopefully eliminate the race condition, or make it *very* difficult to hit.

segun (approved):
LGTM

danjm (approved):

PR #15421, opened 2022-08-02
Prevent editing of networks and accounts while send duck is initializing

danjm:


danjm:
closing in favour of #15422

Issue #15405, opened 2022-08-01
[Bug]: Incorrect max amount displayed in send flow


PeterYinusa:
### Describe the bug

Scenario A: Incorrect max amount displayed when switching networks
https://user-images.githubusercontent.com/53189696/182223277-c1831f0b-2ce7-43c2-8061-482b462a15af.mov

Scenario B : Incorrect max amount displayed when switching assets
https://user-images.githubusercontent.com/53189696/182223307-f510d6bb-e690-441d-8689-7f6da1c4e4c1.mov

### Steps to reproduce

Scenario A (see video above):
1. Click the send button on the home page
2. enter a recipient address
3. click the Max amount button
4. Switch networks

Scenario B (see video above):
1. Click the send button on the home page
2. enter a recipient address
3. click the Max amount button
4. Switch assets from Native currency to an ERC20

### Version

10.18.0

### Browser

Firefox

### Operating system

MacOS

digiwand:
^ oops. please ignore this reference.

PR #15442, opened 2022-08-03
Opening Network Forms in full screen from the popup view

### LINKED ISSUE #15399: Unable to edit a network via the popup, only works on full screen
>  I noticed that the network details are no longer editable when the extension is not in expanded view. I just wanted to ask if this is intentional, because I am able to go to Settings->Networks, but when I click on a network to edit it, it takes me to the Settings->General part of the menu.

image (2)

ryanml:
Fixes: https://github.com/MetaMask/metamask-extension/issues/15399

This restores the behavior of opening the Network form in full view when attempting to edit one via the popup window.

Testing involves a simple regression check in the full screen view and ensuring the aformentioned behavior in the popup

### DIFF
--- a/ui/pages/settings/networks-tab/networks-list-item/networks-list-item.js
+++ b/ui/pages/settings/networks-tab/networks-list-item/networks-list-item.js
@@ -2,7 +2,6 @@ import React, { useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import { useDispatch, useSelector } from 'react-redux';
-import { useHistory } from 'react-router-dom';
 import { useI18nContext } from '../../../../hooks/useI18nContext';
 import {
   NETWORK_TYPE_RPC,
@@ -10,7 +9,7 @@ import {
 } from '../../../../../shared/constants/network';
 import LockIcon from '../../../../components/ui/lock-icon';
 import IconCheck from '../../../../components/ui/icon/icon-check';
-import { NETWORKS_FORM_ROUTE } from '../../../../helpers/constants/routes';
+import { NETWORKS_ROUTE } from '../../../../helpers/constants/routes';
 import { setSelectedSettingsRpcUrl } from '../../../../store/actions';
 import { getEnvironmentType } from '../../../../../app/scripts/lib/util';
 import { ENVIRONMENT_TYPE_FULLSCREEN } from '../../../../../shared/constants/app';
@@ -28,7 +27,6 @@ const NetworksListItem = ({
   setSearchedNetworks,
 }) => {
   const t = useI18nContext();
-  const history = useHistory();
   const dispatch = useDispatch();
   const environmentType = getEnvironmentType();
   const isFullScreen = environmentType === ENVIRONMENT_TYPE_FULLSCREEN;
@@ -68,7 +66,7 @@ const NetworksListItem = ({
         setSearchedNetworks([]);
         dispatch(setSelectedSettingsRpcUrl(rpcUrl));
         if (!isFullScreen) {
-          history.push(NETWORKS_FORM_ROUTE);
+          global.platform.openExtensionInBrowser(NETWORKS_ROUTE);
         }
       }}
     >


danjm (approved):

ui/pages/settings/networks-tab/networks-list-item/networks-list-item.js
```diff
@@ -68,7 +66,7 @@ const NetworksListItem = ({
         setSearchedNetworks([]);
         dispatch(setSelectedSettingsRpcUrl(rpcUrl));
         if (!isFullScreen) {
-          history.push(NETWORKS_FORM_ROUTE);
+          global.platform.openExtensionInBrowser(NETWORKS_ROUTE);
```
danjm, line 69: Why `NETWORKS_ROUTE` instead of `NETWORKS_FORM_ROUTE`?

jpuri (approved):

ui/pages/settings/networks-tab/networks-list-item/networks-list-item.js
digiwand:
```
const NETWORKS_ROUTE = '/settings/networks';
const NETWORKS_FORM_ROUTE = '/settings/networks/form';
```

Went down a bit of a rabbit hole here. As the [issue](https://github.com/MetaMask/metamask-extension/issues/15405) shows, the route, `settings/networks/form`,  resolves to the generic `settings` page. Why? It appears the path is not resolving to the networks tab properly and from there, the `shouldRenderNetworkForm` value when the route is `NETWORKS_FORM_ROUTE` is not being set. 

I think that applying the current changes now would make `NETWORKS_FORM_ROUTE` obsolete which might now be what we want.

digiwand (commented):
As a quick fix, this seems okay to merge. If we do merge this, we should follow up to fix the networks route and forms UI in the popup view, or fully remove this feature from the popup by removing the code related to the `NETWORKS_FORM_ROUTE`

ui/pages/settings/networks-tab/networks-list-item/networks-list-item.js
digiwand:
```
const NETWORKS_ROUTE = '/settings/networks';
const NETWORKS_FORM_ROUTE = '/settings/networks/form';
```

Went down a bit of a rabbit hole here. As the [issue](https://github.com/MetaMask/metamask-extension/issues/15405) shows, the route, `settings/networks/form`,  resolves to the generic `settings` page. From my investigation, it appears the path is not resolving to the networks tab properly and from there, the `shouldRenderNetworkForm` value when the route is `NETWORKS_FORM_ROUTE` is not being set. 

hardcoding the route to resolve to the networks tab and setting `shouldRenderNetworkForm` to true shows this popup page:  
Screenshot 2022-08-03 at 8 53 17 PM

I think that applying the current changes now would make `NETWORKS_FORM_ROUTE` obsolete which might now be what we want.

digiwand:
```
const NETWORKS_ROUTE = '/settings/networks';
const NETWORKS_FORM_ROUTE = '/settings/networks/form';
```

Went down a bit of a rabbit hole here. As the [issue](https://github.com/MetaMask/metamask-extension/issues/15399) shows, the route, `settings/networks/form`,  resolves to the generic `settings` page. From my investigation, it appears the path is not resolving to the networks tab properly and from there, the `shouldRenderNetworkForm` value when the route is `NETWORKS_FORM_ROUTE` is not being set. 

hardcoding the route to resolve to the networks tab and setting `shouldRenderNetworkForm` to true shows this popup page:  
Screenshot 2022-08-03 at 8 53 17 PM

I think that applying the current changes now would make `NETWORKS_FORM_ROUTE` obsolete which might not be what we want.  

p.s. i deleted the last copy of this comment and reposted this comment with the proper reference to see if I can remove the reference to the incorrect issue ticket

danjm (changes requested):
we need to resolve the route question

danjm (approved):
actually, I think it has been appropriately answered, and this PR is good for now

ui/pages/settings/networks-tab/networks-list-item/networks-list-item.js
digiwand: p.s.p.s. the reference to this PR from the unrelated issue was not removed when I deleted the comment I mentioned above

PR #15262, opened 2022-07-15
Fix #12345: MV3: Implement content script within manifest.json

darkwing:
## Explanation

If we use the `src` of `<script>` to inject our inpage provider, the code is async and won't be initialized immediately in the page.  By using the new `world: MAIN` property for content scripts, we can ensure the script is injected at the correct time.

## More Information

* Fixes [#12345](https://github.com/MetaMask/metamask-extension/issues/14846)
* Requires https://github.com/MetaMask/metamask-extension/pull/14781 to be merged first

## Manual Testing Steps

1.  Check out this PR on top of #14781 
2.  Open app/scripts/inpage.js
3.  Add an `alert('HERE')` call in `inpage.js`
4.  Open any new browser tab, go to https://davidwalsh.name
5.  See the alert is called immediately

## Pre-Merge Checklist

- [ ] PR template is filled out
- [ ] **IF** this PR fixes a bug, a test that _would have_ caught the bug has been added
- [ ] PR is linked to the appropriate GitHub issue
- [ ] PR has been added to the appropriate release Milestone

### + If there are functional changes:

- [ ] Manual testing complete & passed
- [ ] "Extension QA Board" label has been applied

Issue #14845, opened 2022-06-03
MV3: for extension dapp and popup reconnect UI streams when when service worker is re-loaded


jpuri:
As service worker is inactivated the streams get disconnected. For both dapp pages and the extension popup, we need to re-connect streams as the service worker is disconnected and then re-activated.

Related issue in extension full-screen mode: https://github.com/MetaMask/metamask-extension/issues/14559

PR for extension full-screen mode: https://github.com/MetaMask/metamask-extension/pull/14781

Blocked by: https://github.com/MetaMask/metamask-extension/issues/14846

PR #15448, opened 2022-08-03
Fix #14846 - Inject provider for MV3 via app-init

### LINKED ISSUE #14846: MV3: Make injection of inpage.js on dapps synchronous
In MV3 ethereum api injection is asynchronous and late. Se discussion here: https://github.com/MetaMask/metamask-extension/pull/14762#discussion_r880793784

We need to fix this to make it synchronous.

darkwing:
Supercedes https://github.com/MetaMask/metamask-extension/pull/15262

## Explanation

If we use the `src` of `<script>` to inject our inpage provider, the code is async and won't be initialized immediately in the page.  By using the new `world: MAIN` property for content scripts, we can ensure the script is injected at the correct time.

## More Information

* Fixes [#14846](https://github.com/MetaMask/metamask-extension/issues/14846)

## Manual Testing Steps

1.  Start via `yarn start:mv3`
2.  Refresh the extension in the Extension Manager, which ensures flushing cache
3.  Open the test dapp
4. Type `window.ethereum` in the console, ensure it's there
5. Use the test dapp buttons, ensure the popup appears

## Pre-Merge Checklist

### DIFF
--- a/app/manifest/v3/_base.json
+++ b/app/manifest/v3/_base.json
@@ -47,6 +47,7 @@
   ],
   "default_locale": "en",
   "description": "__MSG_appDescription__",
+  "host_permissions": ["file://*/*", "http://*/*", "https://*/*"],
   "icons": {
     "16": "images/icon-16.png",
     "19": "images/icon-19.png",
@@ -61,6 +62,7 @@
   "name": "__MSG_appName__",
   "permissions": [
     "storage",
+    "scripting",
     "unlimitedStorage",
     "clipboardWrite",
     "http://localhost:8545/",
--- a/app/manifest/v3/_base.json
+++ b/app/manifest/v3/_base.json
@@ -28,7 +28,7 @@
   },
   "content_scripts": [
     {
-      "matches": ["http://*/*", "https://*/*"],
+      "matches": ["file://*/*", "http://*/*", "https://*/*"],
       "js": [
         "disable-console.js",
         "globalthis.js",
--- a/app/manifest/v3/_base.json
+++ b/app/manifest/v3/_base.json
@@ -28,7 +28,7 @@
   },
   "content_scripts": [
     {
-      "matches": ["file://*/*", "http://*/*", "https://*/*"],
+      "matches": ["http://*/*", "https://*/*"],
       "js": [
         "disable-console.js",
         "globalthis.js",
--- a/app/scripts/inpage.js
+++ b/app/scripts/inpage.js
@@ -1,9 +1,3 @@
-/* eslint-disable import/first */
-import log from 'loglevel';
-import { WindowPostMessageStream } from '@metamask/post-message-stream';
-import { initializeProvider } from '@metamask/providers/dist/initializeInpageProvider';
-import shouldInjectProvider from '../../shared/modules/provider-injection';
-
 // need to make sure we aren't affected by overlapping namespaces
 // and that we dont affect the app with our namespace
 // mostly a fix for web3's BigNumber if AMD's "define" is defined...
@@ -35,6 +29,12 @@ const restoreContextAfterImports = () => {
 };
 
 cleanContextForImports();
+
+/* eslint-disable import/first */
+import log from 'loglevel';
+import { WindowPostMessageStream } from '@metamask/post-message-stream';
+import { initializeProvider } from '@metamask/providers/dist/initializeInpageProvider';
+import shouldInjectProvider from '../../shared/modules/provider-injection';
 
 restoreContextAfterImports();
 
--- a/app/scripts/inpage.js
+++ b/app/scripts/inpage.js
@@ -2,152 +2,58 @@
 import log from 'loglevel';
 import { WindowPostMessageStream } from '@metamask/post-message-stream';
 import { initializeProvider } from '@metamask/providers/dist/initializeInpageProvider';
+import shouldInjectProvider from '../../shared/modules/provider-injection';
 
-(() => {
-  // need to make sure we aren't affected by overlapping namespaces
-  // and that we dont affect the app with our namespace
-  // mostly a fix for web3's BigNumber if AMD's "define" is defined...
-  let __define;
+// need to make sure we aren't affected by overlapping namespaces
+// and that we dont affect the app with our namespace
+// mostly a fix for web3's BigNumber if AMD's "define" is defined...
+let __define;
 
-  /**
-   * Caches reference to global define object and deletes it to
-   * avoid conflicts with other global define objects, such as
-   * AMD's define function
-   */
-  const cleanContextForImports = () => {
-    __define = global.define;
-    try {
-      global.define = undefined;
-    } catch (_) {
-      console.warn('MetaMask - global.define could not be deleted.');
-    }
-  };
-
-  /**
-   * Restores global define object from cached reference
-   */
-  const restoreContextAfterImports = () => {
-    try {
-      global.define = __define;
-    } catch (_) {
-      console.warn('MetaMask - global.define could not be overwritten.');
-    }
-  };
-
-  cleanContextForImports();
-
-  restoreContextAfterImports();
-
-  log.setDefaultLevel(process.env.METAMASK_DEBUG ? 'debug' : 'warn');
-
-  //
-  // setup plugin communication
-  //
-
-  if (shouldInjectProvider()) {
-    // setup background connection
-    const metamaskStream = new WindowPostMessageStream({
-      name: 'metamask-inpage',
-      target: 'metamask-contentscript',
-    });
-
-    initializeProvider({
-      connectionStream: metamaskStream,
-      logger: log,
-      shouldShimWeb3: true,
-    });
+/**
+ * Caches reference to global define object and deletes it to
+ * avoid conflicts with other global define objects, such as
+ * AMD's define function
+ */
+const cleanContextForImports = () => {
+  __define = global.define;
+  try {
+    global.define = undefined;
+  } catch (_) {
+    console.warn('MetaMask - global.define could not be deleted.');
   }
+};
 
-  /**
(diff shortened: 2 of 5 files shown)


darkwing:
One difference with this method is that we don't use `shouldInjectProvider` which does various checks for Doctype, host, etc.  Not sure if we'll find that perfectly acceptable.

app/manifest/v3/_base.json
```diff
@@ -61,6 +62,7 @@
   "name": "__MSG_appName__",
   "permissions": [
     "storage",
+    "scripting",
```
Gudahtt, line 65: We've generally been wary of adding new permissions because if they introduce a user-facing warning, it will disable the extension on update and re-prompt the user to agree to the permissions. But [I tested this locally using these steps](https://developer.chrome.com/docs/extensions/mv2/permission_warnings/#view_warnings), and it looks like there is no warning for this at the moment. So... maybe this is OK.

app/manifest/v3/_base.json
```diff
@@ -47,6 +47,7 @@
   ],
   "default_locale": "en",
   "description": "__MSG_appDescription__",
+  "host_permissions": ["file://*/*", "http://*/*", "https://*/*"],
```
Gudahtt, line 50: Why was `file://` included here? I didn't think we communicated with pages that use that scheme.

Gudahtt:
>One difference with this method is that we don't use shouldInjectProvider which does various checks for Doctype, host, etc. Not sure if we'll find that perfectly acceptable.

We can move that check within `inpage.js` itself. If it's run first-thing, even before the imports, it should let us stop execution early at least.

I'm not sure if that covers all the reasons we had that check though. I know the hard-coded domains listed there had incompatibilities with our inpage script (probably relying on globals or doing some sloppy message handling), so moving the check to `inpage.js` should do the trick there. But for the suffix check I'm less sure. I think that was to prevent injecting into non-web contexts like PDFs, but... maybe the contentscript just ignores those so we don't need that check anymore.

app/manifest/v3/_base.json
Gudahtt, line 50: I'm guessing that host permissions were needed for the `inpage.js` to establish a communication channel with the `contentscript.js`.

It would be interesting to find out whether `inpage.js` can communciate directly with the service worker though. It seems silly to have two content scripts, injected in two different ways.

darkwing: I was thinking for local dev but people can serve a page locally so we can remove this.

Gudahtt: Ah OK. I don't mind it being here, I was just curious.

I was thinking this might make it break when viewing non-webpage files :thinking: But maybe not, I didn't test it.

naugtur: host_permissions were necessary to dynamically inject a content script into the "MAIN" world. Otherwise it'd stay isolated.

app/manifest/v3/_base.json
naugtur: I'm surprised host permissions didn't trigger anything. The added permissions seem the minimal set necessary for maintaining the functionality of window.ethereum. I believe we don't have a different option for mv3 unless we find a way for the content script in MAIN world to communicate directly.

app/manifest/v3/_base.json
darkwing: > host_permissions were necessary to dynamically inject a content script into the "MAIN" world

Yep, host permissions were the missing piece I was stuck on for a while.

We *could* move all of our content scripts to `app-init` instead of the manifest, but since the manifest is the "right" way and this is sort of a "hack", I felt it made sense to do it the way I did.

naugtur: file:// is a bit unpredictable, we could consider removing it in production build or entirely. Unless there's usecases that depend on it.

While we're at it - http should only be allowed for localhost IMHO - on a http site someone can get access to your window.ethereum with a simple MITM

naugtur: > We could move all of our content scripts to app-init instead of the manifest, but since the manifest is the "right" way and this is sort of a "hack", I felt it made sense to do it the way I did.

I think the point was not about where the scripts are defined, but whether we could entirely remove the contentscript.js running as an ISOLATED world script and forwarding between the MAIN world inpage.js and background. That'd be possible if a script in MAIN can talk to background worker. 

My best guess at this time is it might be possible using some chrome specific stuff that's explicitly not supported by browser-polyfill.

app/scripts/inpage.js
```diff
@@ -1,56 +1,153 @@
...
-  connectionStream: metamaskStream,
-  logger: log,
-  shouldShimWeb3: true,
-});
+  /**
+   * Determines if the provider should be injected
+   *
+   * @returns {boolean} {@code true} Whether the provider should be injected
+   */
+  function shouldInjectProvider() {
```
naugtur, line 66: Could this (or at least some of this) be determined in the worker when injecting inpage.js?   
I think not, because worker couldn't synchronously know some of the inputs here, but worth asking...

darkwing: Nope; the injection happens when the service worker first starts, and there's no callback or logic I can use for a case-by-case basis.

app/scripts/inpage.js
```diff
@@ -1,3 +1,9 @@
+/* eslint-disable import/first */
+import log from 'loglevel';
+import { WindowPostMessageStream } from '@metamask/post-message-stream';
+import { initializeProvider } from '@metamask/providers/dist/initializeInpageProvider';
+import shouldInjectProvider from '../../shared/modules/provider-injection';
+
 // need to make sure we aren't affected by overlapping namespaces
```
Gudahtt, line 7: Why were these imports moved? They were intentionally delayed to resolve an incompatibility, which is what this comment is about. I don't remember the details though.

app/manifest/v3/_base.json
Gudahtt: I just tried this out locally, and it looks like this isn't an option. Chrome is treating the attempt as a normal 'external connection'. And since `externally_connectable` still doesn't support domain wildcards, there is no way for us to accept connections this way from any website. We're stuck with `contentscript.js` acting as a message relay for now.

app/scripts/inpage.js
darkwing: This change was an artifact of me trying to conditionally import and execute all provider code, because `import` needed to be top level.  I've put the imports back.

app/manifest/v3/_base.json
Gudahtt: >I don't mind it being here, I was just curious.

We should at the very least be consistent about whether we're supporting `file://` for injection or not. I see that it was removed from the `registerContentScripts` call, but it remains here in `host_permissions`. Seems like it should be included or removed from both places.

darkwing: Updated and I agree.  I removed `file://` since I can't imagine it used often -- we can add back if asked to reconsider.

Gudahtt: I'm only just now realizing that we've been injecting contentscripts into `file://` pages since at least 2016. Sorry about that :facepalm:  I was confused, and thought it was being introduced here.

I've pinged Dan Finlay in Slack to see if he remembers why it was added. It didn't seem clear from context. It was definitely done on purpose at some point though, so maybe it would be best to leave it until we find out the reason.

jpuri (dismissed):
Great work, I tested in test-dapp and works well.

Thanks @darkwing , @naugtur 🙏

seaona:
Amazing! I've tested and overall looks great! :100: 
Just one thing, I noticed a slightly strange behaviour, that I am not sure if might be on the Dapp side (meaning Dapps should adapts something on their code?) or on our side.
Basically, when I stop service worker and re-start again, after unlocking MM I can see both on MM and on the Dapp that I am still Connected to the Dapp. Even though, when I continue triggering actions on the Dapp nothing happens.
It is not until I refresh the Dapp that I can continue operating normally. Is this the expected behaviour? @darkwing  @naugtur 

https://user-images.githubusercontent.com/54408225/183867630-4abdd2a3-357b-4b14-9f54-1c990f756f5b.mp4

naugtur:
It's not expected. Something didn't reconnect. 
But I'm not familiar with that logic, I'll leave figuring it out to someone more qualified 😉
Thanks for finding it!

shared/modules/provider-injection.js
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/**
+ * Checks if the current domain is blocked
+ *
+ * @returns {boolean} {@code true} if the current domain is blocked
+ */
+function blockedDomainCheck() {
+  const blockedDomains = [
+    'uscourts.gov',
```
digiwand, line 68: curious why the ordering was changed here

digiwand (commented):
Please ignore the force-push above. I double-checked everything is okay with @darkwing yesterday

Changes LGTM! Tested on Chrome x MacOS

digiwand:
@seaona @naugtur This is expected behavior at the moment because we need to reconnect the streams when the Service Worker restarts. 

This is the issue related here:
https://github.com/MetaMask/metamask-extension/issues/14845
I have a draft PR attempting to fix this:
https://github.com/MetaMask/metamask-extension/pull/15494

seaona:
great! so for QA side this PR is looking good! @darkwing 
I will check the fix on the other PR once ready @digiwand
Thank you all!!

app/manifest/v3/_base.json
Gudahtt: It look like this inconsistency remains. `file://` is listed here in `host_permissions`, but the contentscript doesn't get injected into `file://` URLs.

Gudahtt (approved):
LGTM!

brad-decker (approved):

PR #15450, opened 2022-08-03
add safegaurd for missing account in action

brad-decker:
## Explanation
There seems to be a case (report happened after user imported seed phrase) where an action can be dispatched but 'undefined' is the value for account. This adds a safety check to prevent attempting to set values from an erroneous payload.


danjm (approved):

adonesky1 (approved):
LGTM

PR #2479, opened 2017-10-25
[NewUI] Signature Request

danjm:
This implements the new signature request UIs.

There are two outsanding pieces:

(1) The mockups have a name for the requester (eg the dapp) of the dapp. It is unclear to me whether that data can be obtained by metamask at the moment.
(2) Right now, `signTypedData` requests are not working. @danfinlay will `signTypedData` be implemented in time for the upcoming release?

This feature can easily be tested with the 'personal_sign' and 'sign typed data' buttons found here: https://danfinlay.github.io/js-eth-personal-sign-examples/

SCREENSHOTS:

screen shot 2017-10-25 at 1 26 30 pm
-
screen shot 2017-10-25 at 1 26 39 pm
-
screen shot 2017-10-25 at 1 26 48 pm
-
screen shot 2017-10-25 at 1 27 39 pm
-
sigrequest

PR #15451, opened 2022-08-03
Fix Personal Sign Display

samgbafa:
## Explanation
This PR makes a change to the personal_sign page. Currently there is an icon that is rendered based on the address passed, but the address passed is `null`. This PR changes the address passed to use the same value that is used to select the account (`fromAccount`).

## More Information
Related to https://github.com/MetaMask/metamask-extension/pull/14438#issuecomment-1204225916

### Before

Before

### After
After

## Manual Testing Steps
Go to the [test-dapp](https://metamask.github.io/test-dapp/) and hit personal sign

## Pre-Merge Checklist

### DIFF
--- a/ui/components/app/signature-request-original/signature-request-original.component.js
+++ b/ui/components/app/signature-request-original/signature-request-original.component.js
@@ -229,8 +229,7 @@ export default class SignatureRequestOriginal extends Component {
               className="request-signature__help-link"
               onClick={() => {
                 global.platform.openTab({
-                  url:
-                    'https://consensys.net/blog/metamask/the-seal-of-approval-know-what-youre-consenting-to-with-permissions-and-approvals-in-metamask/',
+                  url: 'https://consensys.net/blog/metamask/the-seal-of-approval-know-what-youre-consenting-to-with-permissions-and-approvals-in-metamask/',
                 });
               }}
             >
--- a/ui/components/app/signature-request-original/signature-request-original.component.js
+++ b/ui/components/app/signature-request-original/signature-request-original.component.js
@@ -32,7 +32,6 @@ export default class SignatureRequestOriginal extends Component {
     conversionRate: PropTypes.number,
     history: PropTypes.object.isRequired,
     mostRecentOverviewPage: PropTypes.string.isRequired,
-    requesterAddress: PropTypes.string,
     sign: PropTypes.func.isRequired,
     txData: PropTypes.object.isRequired,
     subjectMetadata: PropTypes.object,
@@ -107,11 +106,13 @@ export default class SignatureRequestOriginal extends Component {
   };
 
   renderRequestIcon = () => {
-    const { requesterAddress } = this.props;
+    const {
+      fromAccount: { address },
+    } = this.state;
 
     return (
       <div className="request-signature__request-icon">
-        <Identicon diameter={40} address={requesterAddress} />
+        <Identicon diameter={40} address={address} />
       </div>
     );
   };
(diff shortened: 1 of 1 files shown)


ui/components/app/signature-request-original/signature-request-original.component.js
```diff
@@ -228,7 +229,8 @@ export default class SignatureRequestOriginal extends Component {
               className="request-signature__help-link"
               onClick={() => {
                 global.platform.openTab({
-                  url: 'https://consensys.net/blog/metamask/the-seal-of-approval-know-what-youre-consenting-to-with-permissions-and-approvals-in-metamask/',
+                  url:
```
skgbafa, line 232: My linter auto-changed this, I can revert if needed

georgewrmarshall: Hmm yup linter has an issue haha. Try `yarn lint:fix` should fix it

skgbafa: Done!

georgewrmarshall (commented):
I just want to confirm what the intention of the Identicon is for. We are already showing the users account avatar that is generated from their address. Showing it again doesn't make a huge amount of sense? The original address variables is called  `requesterAddress` so that makes me think that the intention is for it to be an identifier avatar for the contract address. If we can't get that should we just remove? @rachelcope @SayaGT

SayaGT:
@georgewrmarshall Not sure what the identicon is supposed to show (it has always showed up blank for me) but with the context it should show the icon/ logo for the dApp requesting the signature. Doesn't make sense to show the account avatar again. We can remove it as long as the icon is displayed in the origin along with the dApp URL. 

Also, this soon will be replaced by new designs @holantonela has worked on which solve these issues.

NiranjanaBinoy:
We are having `requesterAddress` in three components set from respective containers to render the icon using <Identicon/> , SignatureRequestOriginal ConfirmEncryptionPublicKey, and  ConfirmDecryptMessage, but in all three places `requesterAddress` is assigned to `null` in  the respective containers. Based on the above comment by @SayaGT maybe we can remove it from all three places? 

cc @danjm and @Gudahtt to know if they have more context for the use of  `requesterAddress`.

Gudahtt:
`requesterAddress` has always been `null`. See [the original PR here](https://github.com/MetaMask/metamask-extension/pull/2479). At one point it showed the Ethereum logo as a default, that must have broken at some point.

I agree that it doesn't make sense to show the account avatar twice. Leaving this as-is or removing it would make the most sense to me for now. The new designs will certainly solve this for us.

georgewrmarshall:
+1 vote for removing

brad-decker (approved):

georgewrmarshall (changes requested):
Just blocking this PR I think we can fully remove the `renderRequestIcon` from rendering the Identicon here

brad-decker:
@skgbafa could you remove the renderRequestIcon as per @georgewrmarshall 's request so we can move this forward

bschorchit:
I believe we're removing this icon already in this PR: https://github.com/MetaMask/metamask-extension/pull/15776

brad-decker (approved):

brad-decker:
@bschorchit @georgewrmarshall is this PR still needed then? Does this mean its good to go as is?

bschorchit:
> @bschorchit @georgewrmarshall is this PR still needed then? Does this mean its good to go as is? 

I don't think this PR is needed. Let's prioritize https://github.com/MetaMask/metamask-extension/pull/15776 and the issue that motivated this PR won't exist anymore 🙂

brad-decker:
Closing in favor of #15776

PR #15449, opened 2022-08-03
Use fromAccount instead of selectedAcccount for account updates on edit

brad-decker:
## Explanation
When converting to using fromAccount/selectedAccount the logic on the ACCOUNT_CHANGED case was incorrectly converted. It should be using the 'fromAccount' which is set by the editExistingTransaction method. I added additional safety here by checking for existence of both the payload.account and draftTransaction before doing anything. 

Fixes #15400

PR #15446, opened 2022-08-03
Ensure smart contract interactions are properly represented on the confirm screen

danjm:
Fixes: https://github.com/MetaMask/metamask-extension/issues/15413

The problem was that we were representing a contract interaction as a simple send when first arriving at the confirm screen. The solution is to ensure that if we are sending to a contract address, we apply the contract interaction type to the transaction when creating it from the send screen.

Some other changes were needed to get this fully working:
- added `TRANSACTION_TYPES.CONTRACT_INTERACTION` as a valid transaction type when adding a new unapproved transaction
- added a suffix to the currency display on the confirm screen if it is a contract interaction and no eth logo is being shown, so that the user has the information they need about what value would be sent to the contract
- fixed the `isSmartContractAddress` function which was previously broken because it was attempting to get the wrong property from the result of `readAddressAsContract()`

Before:

https://user-images.githubusercontent.com/7499938/182645886-491a74e1-aa42-4026-8823-e27e1631f3af.mp4

After:

https://user-images.githubusercontent.com/7499938/182644897-94f850d3-af66-4c5a-8dd5-be0ea41ca6df.mp4

PR #15444, opened 2022-08-03
Ensure that if max is selected in send flow, the correct value is set after network switch

danjm:
Fixes https://github.com/MetaMask/metamask-extension/issues/15405

Explanation:

After switching networks, if max amount is selected on the send flow, the amount needs to be updated to reflect the new balance.

PR #15424, opened 2022-08-02
Stop throwing an error when adding gas defaults for a simple send, that has data, to an address without a response code

danjm:
Fixes https://github.com/MetaMask/metamask-extension/issues/15373

The code removed here is not needed. We should be able to estimate gas for simple send transactions that include ~~gas~~ data.

PR #15455, opened 2022-08-03
Update changelog v10.18.2

danjm:
### Changed
- Enhance approval screen title logic ([#15406](https://github.com/MetaMask/metamask-extension/pull/15406))

### Fixed
- Ensure smart contract interactions are properly represented on the confirm screen ([#15446](https://github.com/MetaMask/metamask-extension/pull/15446))
- Fix update of max amount in send flow after network switch([#15444](https://github.com/MetaMask/metamask-extension/pull/15444))
- Fix to ensure user can access full screen editing of network forms from the popup ([#15442](https://github.com/MetaMask/metamask-extension/pull/15442))
- Possibly fix bug which crashes firefox on startup after upgrade to v10.18.1 ([#15425](https://github.com/MetaMask/metamask-extension/pull/15425))
- Fix blocking of editing transactions that had a contract address recipient but no tx data ([#15424](https://github.com/MetaMask/metamask-extension/pull/15424))
- Fix error that could leave the app in a stuck state when quickly moving between the send screen and other screens ([#15420](https://github.com/MetaMask/metamask-extension/pull/15420))
- Fix send screen for Optimism network ([#15419](https://github.com/MetaMask/metamask-extension/pull/15419))
- Fix error that makes app unusable after clicking activity list items for token approval transactions ([#15398](https://github.com/MetaMask/metamask-extension/pull/15398))

Note that this intentionally did not include changes for the following two commits:
- add safegaurd for missing account in action ([#15450](https://github.com/MetaMask/metamask-extension/pull/15450))
- Use fromAccount instead of selectedAcccount for account updates on edit ([#15449](https://github.com/MetaMask/metamask-extension/pull/15449))

As these had no user reports and we were not able to repro, so we don't know the user impact (but we did see errors in sentry)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix blocking of editing transactions that had a contract address recipient but no tx data ([#15424](https://github.com/MetaMask/metamask-extension/pull/15424))
 - Fix error that could leave the app in a stuck state when quickly moving between the send screen and other screens ([#15420](https://github.com/MetaMask/metamask-extension/pull/15420))
 - Fix send screen for Optimism network ([#15419](https://github.com/MetaMask/metamask-extension/pull/15419))
+- Fix to ensure the correct balance is used when validating send amounts ([#15449](https://github.com/MetaMask/metamask-extension/pull/15449))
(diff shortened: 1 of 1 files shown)


brad-decker:
If possible we should include the use fromAccount instead of selectedAccount fix. The implication there is that the wrong account's balance could be used for edits which wouldnt present as a sentry bug but would be a bug.

ryanml (approved):

brad-decker (approved):

PR #15457, opened 2022-08-04
Removing unused NETWORKS_FORM_ROUTE route

ryanml:
Follow up to: https://github.com/MetaMask/metamask-extension/pull/15442

### DIFF
--- a/ui/helpers/constants/routes.js
+++ b/ui/helpers/constants/routes.js
@@ -10,7 +10,6 @@ const SECURITY_ROUTE = '/settings/security';
 const ABOUT_US_ROUTE = '/settings/about-us';
 const ALERTS_ROUTE = '/settings/alerts';
 const NETWORKS_ROUTE = '/settings/networks';
-const NETWORKS_FORM_ROUTE = '/settings/networks/form';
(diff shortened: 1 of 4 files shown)


digiwand (commented):
Can we add an explanation as to why we're removing this feature entirely from the * popup *? If we go through with this, we should also remove the related styles

adonesky1:
> Can we add an explanation as to why we're removing this feature entirely from mobile? If we go through with this, we should also remove the related styles

@digiwand not sure whether this has anything to do with mobile, but I agree PR description would help us understand how to review. Thanks in advance @ryanml!

NiranjanaBinoy (approved):
LGTM

digiwand:
Sorry, that was a typo slip. I meant the popup* not mobile. @adonesky1

brad-decker:
@ryanml could you update this PR so we can move it forward?

brad-decker:
@ryanml i'm converting this to draft until it can be rebased. After another 30 day period i'll close the PR due to inactivity. Only doing this to keep the PR queue clear and put proper emphasis on keeping prs in a reviewable state! <3

PR #15459, opened 2022-08-04
Stop injecting provider on docs.google.com

### LINKED ISSUE #15374: [Bug]: Metamask breaks Google Docs / Sheets documents
### Describe the bug

Google Docs / Sheets will not finish to load with a set up Metamask extension.
You will get "unable to load file" error in Google Docs
image

### Steps to reproduce

1. Setup (and finish setup) Metamask with a new wallet (I've only tested with my existing wallets).
2. open Google Drive https://drive.google.com/
3. open an existing Google Docs or Google Sheets document.
4. it fails to load (and works again after deactivating Metamask)

Error from console (Google Docs document). It seems some variable `chainId` is read-only because of Metamask:
image

```shell
n.a., Metamask is locked during whole procedure
```

### Version

10.18.0

### Build type

_No response_

### Browser

Brave

### Operating system

MacOS

### Hardware wallet

Ledger

### Additional context

_No response_

rekmarks:
Per #15374, injecting `window.ethereum` on `docs.google.com` can prevent the user from opening Google docs. I have reproduced this locally myself, and could only fix this by disabling the extension in the browser. I think we can rest assured that our provider is of no use on the Google docs website, and we can always revert this if that should change.

Fixes #15374

### DIFF
--- a/app/scripts/contentscript.js
+++ b/app/scripts/contentscript.js
@@ -329,16 +329,17 @@ function documentElementCheck() {
  */
 function blockedDomainCheck() {
   const blockedDomains = [
-    'uscourts.gov',
-    'dropbox.com',
-    'webbyawards.com',
-    'cdn.shopify.com/s/javascripts/tricorder/xtld-read-only-frame.html',
     'adyen.com',
-    'gravityforms.com',
-    'harbourair.com',
     'ani.gamer.com.tw',
     'blueskybooking.com',
+    'cdn.shopify.com/s/javascripts/tricorder/xtld-read-only-frame.html',
+    'docs.google.com',
+    'dropbox.com',
+    'gravityforms.com',
+    'harbourair.com',
     'sharefile.com',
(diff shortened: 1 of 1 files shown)


app/scripts/contentscript.js
```diff
@@ -329,16 +329,17 @@ function documentElementCheck() {
...
     'ani.gamer.com.tw',
     'blueskybooking.com',
+    'cdn.shopify.com/s/javascripts/tricorder/xtld-read-only-frame.html',
+    'docs.google.com',
+    'dropbox.com',
+    'gravityforms.com',
+    'harbourair.com',
     'sharefile.com',
+    'uscourts.gov',
+    'webbyawards.com',
```
rekmarks, line 342: I sorted these lines alphabetically but the only addition is `docs.google.com`.

Gudahtt (approved):
LGTM!

darkwing (approved):

PR #15461, opened 2022-08-04
Fix/use etherjs specific imports

### LINKED ISSUE #15232: Replace general imports of ethers with specific package imports
Blocked by https://github.com/MetaMask/metamask-extension/issues/15231

We currently include the who ethers library in our dependency tree, but it should be possible to just require and import only the packages we need.

See the readme here for an example: https://github.com/ethers-io/ethers.js/tree/master/packages/contracts

The task is to replace all instances where we import all of ethers with the specific package that is needed, and then to remove the general package from package.json. We should compare the build size before and after this change to confirm it has the desired affect.

amerkadicE:
## Explanation

We don't need to import both whole ethers npm package.
This PR replaces all imports of ethers npm umbrella package with a specific ethers package.

## More Information

Fixes #15232

### Before
**Current bundle size on develop**
"background": {
    "name": "background",
    "size": 76085955
 
   "ui": {
    "name": "ui",
    "size": 94503010,

**Google Chrome build size after removing web3 npm package.**

  "background": {
    "name": "background",
    "size": 10236142,
    
  "ui": {
    "name": "ui",
    "size": 12504705,

image

### After
Google Chrome build size after removing ethers umbrella npm package.
"background": {
    "name": "background",
    "size": 13145874,
    
  "ui": {
    "name": "ui",
    "size": 16133810,

image

## Manual Testing Steps

Check if the extension works correctly. There are a lot of places where we changed ethers package. 

## Pre-Merge Checklist


app/scripts/controllers/swaps.test.js
```diff
@@ -377,7 +377,7 @@ describe('SwapsController', function () {
         assert.strictEqual(gasEstimate, bufferedGasLimit);
         assert.strictEqual(
           gasEstimateWithRefund,
-          `0x${new BigNumber(maxGas, 10)
+          `0x${new BigNumberjs(maxGas, 10)
```
adonesky1, line 380: Could we just use the `ethers` version of BigNumber here too? And get rid of `bignumber.js` here?

adonesky1 (dismissed):
One small question, otherwise LGTM

app/scripts/controllers/swaps.test.js
amerkadicE: It will cause a test to fail because ethers BigNumber adds 0 when we covert to hex value. 
image

danjm (changes requested):
hmmm... why does this make the bundle size bigger? We should investigate that

adonesky1:
I was trying this approach to measuring size increase as well - looking at the zip file size - but I'm not sure if its a valid way to measure...? @brad-decker @Gudahtt what do you think?

amerkadicE:
@danjm I investigated why we have bundle sizes bigger after using specific ethers packages. 

It looks like every specific ethers package will install node modules that he needs, so it will cause multiple installations of the same package. Like in the picture below.

image

When we install umbrella ethers, the size of that node module is 8,5 MB.
When we install only specific packages that we use, the size of that node module is 25,5MB.

adonesky1:
> @danjm I investigated why we have bundle sizes bigger after using specific ethers packages.
> 
> It looks like every specific ethers package will install node modules that he needs, so it will cause multiple installations of the same package. Like in the picture below.
> 
> image
> 
> When we install umbrella ethers, the size of that node module is 8,5 MB. When we install only specific packages that we use, the size of that node module is 25,5MB.

looking into this.

adonesky1:
So at least part of the problem is that any places in our dependency tree that `ethers` is imported rather than the component `@ethersproject/<subfolder>` it makes sure that we will have atleast one fixed version import of each `@ethersproject/` dependency, since each version of `ethers` depends on the same fixed `equivalent` version of each of the `@ethersproject/` components: [5.7.0](https://npmfs.com/package/ethers/5.7.0/package.json#L5), [5.6.0](https://npmfs.com/package/ethers/5.6.0/package.json#L5), etc. So we should try to get rid of all uses of the "umbrella" `ethers` package in our dependency tree if possible.

adonesky1:
Actually not sure if we can do this since `@ethersproject/hardware-wallets` depends on `ethers` itself...

adonesky1:
Looks like upgrading our version of `@eth-optimism/contracts` could help...

adonesky1:
> Looks like upgrading our version of `@eth-optimism/contracts` could help... @danjm do you know if this will require complex adaptations, seems like we're pretty far behind: currently `.0.0-2021919175625` -> latest `0.5.33`

Looks like the interfaces haven't changed for the functions/objects we use

danjm:
> Looks like the interfaces haven't changed for the functions/objects we use
@adonesky1 I agree. I don't think updating will be breaking for us

adonesky1:
So despite consolidating all of the `@ethersproject/` components to a single version, i'm still seeing the size increase about the same amount. Looking at the remaining instances of `ethers` in our dependency graph: 
Screen Shot 2022-09-07 at 2 19 20 PM

we can easily move `@metamask/controllers` and `@metamask/smart-transactions-controller` to use the specific imports ourselves, but will need `eth-optimism` team to update [@eth-optimism/core-utils](https://www.npmjs.com/package/@eth-optimism/core-utils) to use specific imports instead of `ethers` if we want to get it out of our dependency tree altogether.

adonesky1:
I've pinged the Optimism team about this here: https://consensys.slack.com/archives/C01KXHPU3C6/p1662579353556779

adonesky1:
the Optimism team has graciously opened PR that will help us remove the umbrella package `ethers` from our dependency tree: https://github.com/ethereum-optimism/optimism/pull/3363. Once they publish we can pull that in and update here.

adonesky1:
Working with eth-optimism team to get their new build right.

adonesky1:
Currently waiting for updated build from Optimism: [thread](https://consensys.slack.com/archives/C01KXHPU3C6/p1663199469927819?thread_ts=1662579353.556779&cid=C01KXHPU3C6)

brad-decker:
@adonesky1 are there any updates relevant to this PR that can be shared?

brad-decker (dismissed):

adonesky1:
The only outstanding instance of `ethers` in the extensions's dependency graph is from `@ensdomains/ensjs` v2.1.0 via `@truffle/encoder` / `@truffle/decoder`. `@truffle/decoder` is used by the `transactions-decoding` component, and probably can't be easily replaced but `@ensdomains/ensjs` has an alpha version (3.0.0) which removes it's dependency on `ethers` so perhaps we ask Truffle to update this dependency to use the alpha version of `@ensdomains/ensjs`? cc @danjm  I will write up a separate ticket to follow up on this.

danjm (approved):
~~LGTM!~~

app/scripts/controllers/detect-tokens.js
```diff
@@ -114,6 +115,7 @@ export default class DetectTokensController {
       : tokenList;
 
     const tokensToDetect = [];
+    this.ethersProvider = new Web3Provider(this._network._provider);
```
danjm, line 118: Oh, I don't know if this file needs a provider any more?

danjm (changes requested):

app/scripts/controllers/detect-tokens.js
adonesky1: looks like you're right! Will remove

adonesky1: done [here](https://github.com/MetaMask/metamask-extension/pull/15461/commits/cbe7652b41a0835def382ee508550348e23679ce)

danjm (approved):

brad-decker (approved):

brad-decker:
Is replacing bignumber imports an important part of this? It doesnt seem like all were changed

PR #15464, opened 2022-08-04
Replace `rc` with `ini`

Gudahtt:
We use the `rc` package to read the `.metamaskrc` configuration file, which is in "ini" format. This package has been replaced by the `ini` package.

The `rc` package was not actively maintained, and it has had recent security vulnerabilities. But most importantly, the config object returned by `rc` includes a bunch of extra information that made [build script validation](https://github.com/MetaMask/metamask-extension/issues/15003) difficult to implement. Specifically, it  made it challenging to ensure no extra environment variables were present.

The `ini` package on the other hand is simple, well maintained, and is simpler to use. This package doesn't add any extra properties to the object it returns, making validation easy.

## Manual Testing Steps

This should have no functional changes.

## Pre-Merge Checklist

- [x] PR template is filled out
- [x] **IF** this PR fixes a bug, a test that _would have_ caught the bug has been added
- [x] PR is linked to the appropriate GitHub issue
- [x] PR has been added to the appropriate release Milestone

### + If there are functional changes:

- [ ] Manual testing complete & passed
- [ ] "Extension QA Board" label has been applied

PR #15468, opened 2022-08-04
Add validation to production build script 

### LINKED ISSUE #15003: Add environment variable validation to the build script
Currently the build script doesn't validate that expected environment variables are set for production builds. This is because the build script doesn't know the difference between a real production build and a production-like build. We have one build target (`yarn build prod` a.k.a. `yarn dist`) that we use for local testing and for the real prod build, and environment variables are the only difference between the two. There's no way for the script to know whether a missing environment variable is intentional or not.

We should split this single build target into two: one target for production-like builds, and one for a real production build. The real production build target can include comprehensive validation that ensures all expected environment variables are set.

Gudahtt:
Validation has been added to the build script when the "prod" target is selected. We now ensure that all expected environment variables are set, and that no extra environment variables are present (which might indicate that the wrong configuration file is being used).

The `prod` target uses a new `.metamaskprodrc` configuration file. Each required variable can be specified either via environment variable or via this config file. CI will continue set these via environment variable, but for local manual builds we can use the config file to simplify the build process and ensure consistency.

A new "dist" target has been added to preserve the ability to build a "production-like" build without this validation.

The config validation is invoked early in the script, in the CLI argument parsing step, so that it would fail more quickly. Otherwise we'd have to wait a few minutes longer for the validation to run. This required some refactoring, moving functions to the utility module and moving the config to a dedicated module.

Closes #15003

## Manual Testing Steps

The build targets of `dev`, `devTest`, `test` should all remain unchanged. The new build target `dist` should behave the same as the old `prod` target (which confusingly was invoked as `yarn dist`).

The new `prod` target has this new behavior:
* It should demand that all _required_ variables are set either via environment variable or the `.metamaskprodrc` config file.
  * The required variables are:
    * `SENTRY_DSN`
    * `PUBNUB_PUB_KEY`
    * `PUBNUB_SUB_KEY`
    * `INFURA_[build type]_PROJECT_ID`
    * `SEGMENT_[build type]_WRITE_KEY`
  * For the build-type-specific variables, only those used by the current build are required.
* Extra values specified in `.metamaskprodrc` should produce an error.
* Extra environment variables **should not** produce an error (of course - that would make this impossible to use)
* The validation should occur early in the build process, before any tasks have started.

## Pre-Merge Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -246,9 +246,25 @@ jobs:
       - checkout
       - attach_workspace:
           at: .
-      - run:
-          name: build:dist
-          command: yarn dist
+      - when:
+          condition:
+            not:
+              matches:
+                pattern: /^master$/
+                value: << pipeline.git.branch >>
+          steps:
+            - run:
+                name: build:dist
+                command: yarn build dist
+      - when:
+          condition:
+            matches:
+              pattern: /^master$/
+              value: << pipeline.git.branch >>
+          steps:
+            - run:
+                name: build:prod
+                command: yarn build prod
       - run:
           name: build:debug
           command: find dist/ -type f -exec md5sum {} \; | sort -k 2
@@ -266,7 +282,7 @@ jobs:
           at: .
       - run:
           name: build:dist
-          command: yarn build --build-type beta prod
+          command: yarn build --build-type beta dist
       - run:
           name: build:debug
           command: find dist/ -type f -exec md5sum {} \; | sort -k 2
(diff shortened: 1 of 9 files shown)


Gudahtt:
This depends upon #15464 and #15447

brad-decker (approved):
I have validated that when missing variables the build fails, and that the proper rc files are used for instantiating the build. All looks good to me.

brad-decker:
Also validated that extra vars invalidate the build

rickycodes (approved):
lgtm

PR #15472, opened 2022-08-04
Prevent confirm screen from showing method name from contract registry for txes created within MetaMask

danjm:
Explanation:

Prior to v10.18.2, if we edited hex data on the send screen and sent to a contract address, we incorrectly identified it as a simple send. On 10.18.2 we correctly represent it as a contract interaction, but by doing so we show the name of the method associated with that 4-byte signature from the registry.

This PR maintains the representation of these transactions as contract interactions, but ensures that if they are created from within MetaMask, we don't show any specific names, to avoid any possibility of showing incorrect or malicious info to the user in these cases.

before:
Screenshot from 2022-08-04 16-08-10

after:
Screenshot from 2022-08-04 16-06-16

### DIFF
--- a/ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
+++ b/ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
@@ -1054,7 +1054,10 @@ export default class ConfirmTransactionBase extends Component {
     } = this.getNavigateTxData();
 
     let functionType;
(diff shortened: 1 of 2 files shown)


adonesky1 (approved):
LGTM

brad-decker (approved):

seaona:
QA pass

PR #9048, opened 2020-07-22
Resolve on 0x

tmashuang:
Need to see which types of contract interactions and result in 0x fourBytePrefix and how we can mitigate the `getContractMethodData()` to those contracts, and/or the way we slice the tx data need adjusting https://github.com/MetaMask/metamask-extension/blob/7333d821b8fb4f42d4efa8f3eeb8792580ba7832/ui/app/store/actions.js#L2191, and/or we are setting that default tx data to 0x somewhere.

This to prevent calls to fourbyte on `knownMethod['0x']`

PR #15473, opened 2022-08-04
Skip searching 4byte directory if we don't have a full 4 bytes of data

adonesky1:
We should only query `4byte.directory` if we have a full 4bytes of data in hex-string form to query against. 4byte.directory will return an unfiltered list of signatures without a full valid query: https://www.4byte.directory/api/v1/signatures/?hex_signature=0x0.
Discovered as part of this [thread](https://consensys.slack.com/archives/GTQAGKY5V/p1659636499104129?thread_ts=1659566138.744119&cid=GTQAGKY5V).

Before:
https://user-images.githubusercontent.com/34557516/182933602-1b3925e9-4e88-40f8-a2d9-566c1f56a2a0.mp4

https://user-images.githubusercontent.com/34557516/182933715-b29ed147-251b-4344-a6ce-b68ebd184583.mp4

After:
https://user-images.githubusercontent.com/34557516/182933915-b5854ed7-2c4f-4b38-8586-99fd797fd6e9.mov

### DIFF
--- a/ui/store/actions.js
+++ b/ui/store/actions.js
@@ -3264,8 +3264,7 @@ export function getContractMethodData(data = '') {
     if (
       (knownMethodData &&
         knownMethodData[fourBytePrefix] &&
-        Object.keys(knownMethodData[fourBytePrefix]).length !== 0) ||
-      fourBytePrefix === '0x'
+        Object.keys(knownMethodData[fourBytePrefix]).length !== 0)
     ) {
       return Promise.resolve(knownMethodData[fourBytePrefix]);
     }
--- a/ui/store/actions.js
+++ b/ui/store/actions.js
@@ -3262,9 +3262,9 @@ export function getContractMethodData(data = '') {
     }
     const { knownMethodData } = getState().metamask;
     if (
-      (knownMethodData &&
-        knownMethodData[fourBytePrefix] &&
-        Object.keys(knownMethodData[fourBytePrefix]).length !== 0)
+      knownMethodData &&
+      knownMethodData[fourBytePrefix] &&
+      Object.keys(knownMethodData[fourBytePrefix]).length !== 0
     ) {
       return Promise.resolve(knownMethodData[fourBytePrefix]);
     }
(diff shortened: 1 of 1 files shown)


seaona:
from QA looks good

NiranjanaBinoy (dismissed):

ui/store/actions.js
```diff
@@ -3172,8 +3172,10 @@ export function getContractMethodData(data = '') {
   return (dispatch, getState) => {
     const prefixedData = addHexPrefix(data);
     const fourBytePrefix = prefixedData.slice(0, 10);
+    if (fourBytePrefix.length < 10) {
+      return Promise.resolve({});
+    }
```
digiwand, line 3177: Nice fix here. 

I'm a bit confused as to why we did support `fourBytePrefix === '0x'` on line 3183 though. I think we'll either want to continue supporting this or remove the line.

adonesky1: That's a great call out @digiwand, I'm not sure what that is about. @FrederikBolding any idea why we would expect `0x` to be present in `knownMethodData` and why we specify it here in the if block on lines 3180-3183?

adonesky1: @danjm @danfinlay do either of you happen to know why we carved out an exception for `0x` here?

digiwand: Looks like this logic was added here https://github.com/MetaMask/metamask-extension/pull/9048 cc: @Gudahtt 

If I'm understanding this right, we were merely preventing calls to check the 4byte directory if `fourBytePrefix === 0x`. This should mean we can remove the `fourBytePrefix === 0x` check with this new logic you added in this PR

digiwand (changes requested):
request to remove unutilized `fourBytePrefix === 0x` check

ui/store/actions.js
adonesky1: done [here](https://github.com/MetaMask/metamask-extension/pull/15473/commits/53018843fb262a279034442d5430b0bbf39bac39) @digiwand 😄

Gudahtt: Yes, @digiwand is correct, as far as I can recall anyway.

digiwand (approved):
🙌

brad-decker (approved):

seaona:
Just saw this other issue related to resolving contract functions. It seems it's a different one, but not sure if this PR would also affect this behaviour. Just tagging FYI @adonesky1 @digiwand 

[Zero-arg function calls shown as "Contract interaction" rather than function name](https://github.com/MetaMask/metamask-extension/issues/15837)

PR #15477, opened 2022-08-05
improvement it translation

lonasozo:
forgotPassword string missed, added and translated.

## Explanation
forgotPassword string was missed inside the it translation, so the italian version of Metamask still print out the english text. I've added it and translated.


lonasozo:

darkwing:
Thank you for the contribution @lonasozo !  I think you've included an extra space at the beginning of the localization key that's making linting fail -- could you fix that?

Thank you!

lonasozo:
Ouch, you're right.
Should be fixed now

brad-decker (approved):

NidhiKJha (approved):

PR #15481, opened 2022-08-05
[E2E]: Set approval for all e2e

PeterYinusa:
## Explanation

Add e2e test for NFT's
- setApprovalForAll
- approve
- transfer


test/e2e/seeder/ganache-seeder.js
```diff
@@ -1,15 +1,14 @@
...
-const { contractConfiguration } = require('./smart-contracts');
+const { SMART_CONTRACTS, contractConfiguration } = require('./smart-contracts');
 const GanacheContractAddressRegistry = require('./ganache-contract-address-registry');
 
 /*
  * Ganache seeder is used to seed initial smart contract or set initial blockchain state.
  */
 class GanacheSeeder {
-  constructor(debug = false) {
-    this.debug = debug;
```
PeterYinusa, line 11: removed debug in the ganache seeder as we can simply add the property `quiet: false` to ganache object in our tests

test/e2e/seeder/ganache-seeder.js
```diff
@@ -19,39 +18,38 @@ class GanacheSeeder {
    */
 
   async deploySmartContract(contractName) {
-    if (this.debug) {
-      console.log('Deploying smart contracts using GanacheSeeder');
-    }
-
     const ethersProvider = new ethers.providers.Web3Provider(
-      ganache.provider(),
```
PeterYinusa, line 27: Use the provider from the Ganache instance that's already running

test/e2e/seeder/smart-contracts.js
```diff
@@ -33,11 +33,18 @@ const failingContract = {
   abi: failingContractAbi,
 };
 
+const SMART_CONTRACTS = {
+  HST: 'hst',
+  COLLECTIBLES: 'collectibles',
+  PIGGYBANK: 'piggybank',
+  FAILING: 'failing',
+};
```
PeterYinusa, line 41: Since these values will be reused in multiple places

test/e2e/tests/collectibles.spec.js
```diff
@@ -0,0 +1,200 @@
...
+        await driver.fill('#password', 'correct horse battery staple');
+        await driver.press('#password', driver.Key.ENTER);
+
+        // Click transfer
+        await driver.openNewPage(`http://127.0.0.1:8080/?contract=${contract}`);
+        await driver.waitForSelector({
+          css: '#collectiblesStatus',
+          text: 'Deployed',
+        });
+        await driver.delay(veryLargeDelayMs);
```
PeterYinusa, line 42: I had to add this delay for the tests to run reliably for me. Without this sometimes clicking on the button doesn't trigger the notification

test/e2e/tests/collectibles.spec.js
```diff
@@ -0,0 +1,200 @@
...
+        await driver.fill('#transferTokenInput', '1');
+        await driver.clickElement('#transferFromButton');
+        await driver.waitUntilXWindowHandles(3);
+        const windowHandles = await driver.getAllWindowHandles();
+        const extension = windowHandles[0];
```
darkwing, line 47: We could alternatively do the following instead, but I don't want to prescribe how we write these tests, as others touch them way more than me:

```
const windowHandles = await driver.getAllWindowHandles();
const [extension] = windowHandles;
```

darkwing (approved):

test/e2e/tests/collectibles.spec.js
```diff
@@ -0,0 +1,200 @@
...
+        );
+
+        // Verify transaction
+        const completedTx = await driver.findElement('.list-item__title');
+        const completedTxText = await completedTx.getText();
+        assert.equal(completedTxText, 'Approve Token spend limit');
+      },
+    );
+  });
+  it('should approve an address to transfer all NFTs', async function () {
```
adonesky1, line 137: Could we add a test for revoking `setApprovalForAll`?

test/e2e/tests/collectibles.spec.js
PeterYinusa: I also prefer the suggested syntax. I'll use that going forward

test/e2e/tests/collectibles.spec.js
PeterYinusa: Yes we can, that will require changes in the [test-dapp](https://github.com/MetaMask/test-dapp/blob/main/src/index.js#L548) as it's hardcoded to true, so we will have to do that as a follow up.

PeterYinusa: https://github.com/MetaMask/metamask-extension/issues/15488

adonesky1 (approved):
LGTM! Thanks Peter!

PR #15485, opened 2022-08-05
Feat/popup fix

### LINKED ISSUE #14534: [Bug]: MetaMask doesn't Popup
### Describe the bug

when calling this function if it's waiting for a request it throws an error and don't pop up
`accounts = await ethereum.request({ method: "eth_requestAccounts" });`

image

image

Is there a way to threat this error ? or at least a function that should open the metamask// just pop up it.
I believe it would be nice.

image

TL:DR:
after you minimized the pop upped window it does not pop up again.
I believe that due to user experience it should.

### Steps to reproduce

1. Have a function like this that call to open the metamask:
```
 const connectWallet = async () => {
    try {
      const { ethereum } = window;

      if (!ethereum) {
        alert("Get MetaMask!");
        return;
      }

      const accounts = await ethereum.request({ method: "eth_requestAccounts" });

      console.log("Connected", accounts[0]);
      setCurrentAccount(accounts[0]);

    } catch (error) {
      console.log(error)
    }
  }
```
2.just have a button that calls it.
3 run the app click it. 

```shell
MetaMask - RPC Error: Already processing eth_requestAccounts. Please wait. 
{
    "code": -32002,
    "message": "Already processing eth_requestAccounts. Please wait."
}
```

### Version

10.12.4

### Build type

_No response_

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

Gtonizuka:
## Explanation

After calling a method (for example `eth_requestAccounts`) if MetaMask is minimized and the user tries to call the same error again it gets an error in the console but nothing happens visually.

This PR implements a "popup again" functionality, in order to reopen MetaMask after a method is being called again instead of silently failing.

## More Information

Fixes #14534 

## Manual Testing Steps

Call `eth_requestAccounts`, after popup appears before giving permission minimize MetaMask (By clicking anywhere outside the popup) and recall the same method again.
MetaMask should now reopen the popup windown

## Pre-Merge Checklist


app/scripts/lib/rpc-method-middleware/handlers/request-accounts.js
```diff
@@ -57,13 +58,11 @@ async function requestEthereumAccountsHandler(
     getUnlockPromise,
     hasPermission,
     requestAccountsPermission,
+    openPopup,
   },
 ) {
   if (locks.has(origin)) {
-    res.error = ethErrors.rpc.resourceUnavailable(
```
Gudahtt, line 63: In this situation, there are now two in-flight `eth_requestAccount` requests. What happens with each of them? Would they both succeed? Does the user get prompted twice?

Gtonizuka: yes it returns result twice.

We actually discussed this internally and it seems like there are good cases for both cancelling the other request(s) and to also always return a value

Gudahtt: Hmm OK. So is the user asked to approve the connection twice?

Gtonizuka:
I also notice there is a bug when rejecting, multiple popus show in the UI (accept works fine)

app/scripts/lib/rpc-method-middleware/handlers/request-accounts.js
```diff
@@ -85,10 +88,15 @@ async function requestEthereumAccountsHandler(
 
   // If no accounts, request the accounts permission
   try {
+    locks.add(origin);
+    await getUnlockPromise(true);
     await requestAccountsPermission();
   } catch (err) {
     res.error = err;
+    locks.delete(origin);
```
Gudahtt, line 96: This seems to be redundant - the `locks.delete(origin)` in the `finally` block will run as well

app/scripts/lib/rpc-method-middleware/handlers/request-accounts.js
```diff
@@ -85,10 +88,15 @@ async function requestEthereumAccountsHandler(
 
   // If no accounts, request the accounts permission
   try {
+    locks.add(origin);
```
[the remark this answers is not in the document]
Gudahtt: I'd still like to make sure we have a clear goal in mind here for how this should work.


Current behavior when locked:
* Focus popup (via the "unlock request" promise)
* Reject subsequent duplicate requests without focusing popup

Current behavior when unlocked:
* Focus popup (via the permission + approval controllers)
* Queue up subsequent duplicate requests, focusing the popup with each one (via the permission + approval controllers)


Here's how I understand this PR to work:

Proposed behavior when locked:
* Focus popup (via the "unlock request" promise)
* Reject subsequent duplicate requests but focus popup for each one

Proposed behavior when unlocked:
* Focus popup (via the "unlock request" promise, the again via permission + approval controllers)
* Reject subsequent duplicate requests but focus popup for each one

Is that what you meant to do? Why is the popup now being focused twice while unlocked? Did you intend to reject duplicate requests while unlocked?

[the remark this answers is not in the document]
brad-decker: @Gudahtt @Gtonizuka any update here on getting to consensus on approach?

brad-decker:
@Gtonizuka i'm going to close this for now. Please reopen if it gets to the point of being able to be included.

Issue #15050, opened 2022-06-24
MV3: Ensure user remains logged in after service work restart


danjm:
- what is a service worker?

- how will the extension be using a service worker under MV3?

- what does "service worker restart" mean?

- why will the service worker restart interfere with the user being logged in?

darkwing:
https://gist.github.com/darkwing/79dfd89b17ae8c5ae7d90b1005452f4f

Issue #14183, opened 2022-03-24
MV3: fix ledger keyring issue


jpuri:
Starting extension in MV3, this error can be seen in console:
<img width="564" alt="Screenshot 2022-03-24 at 10 21 31 AM" src="https://user-images.githubusercontent.com/2182307/159845964-8db3bacd-2abf-425a-a572-5d10282e848f.png">

This needs to be fixed.

An option to fix it could be using content script or move functionality to UI.

danjm:
This was fixed by https://github.com/MetaMask/metamask-extension/pull/17354

PR #15494, opened 2022-08-05
MV3: contentscript.js - re-activate streams when Service Worker terminates and then resets

### LINKED ISSUE #14845: MV3: for extension dapp and popup reconnect UI streams when when service worker is re-loaded

digiwand:
## Explanation

When the service worker restarts during a user session, we must reestablish the extension streams.
This PR disconnects, destroys, and recreates the extension streams for the `contentscript.js` when the service worker restarts.

The setup of the streams remains pretty much the same. The logic has been split into separate functions to be reused and to separate concerns. There is additional logic to fire an event when the service worker disconnects to destroy the extension streams and reestablish them.

I had to get rid of the `forwardTrafficBetweenMuxes` functions because we need access to the streams created in these functions. 

## More Information
- Fixes #14845
- Follow-up of #14781

## Manual Testing Steps

1. Run `yarn start:mv3` 
2. Open test-dapp
3. Check that streams work by using dapp 
4. Terminate the `app-init.js` service worker
   - [chrome://inspect/#service-workers](chrome://inspect/#service-workers)
   - [chrome://serviceworker-internals/](chrome://serviceworker-internals/)
5. Check that dapp continues to work

## Pre-Merge Checklist


app/scripts/contentscript.js
```diff
@@ -110,42 +115,87 @@ async function setupPhishingStream() {
...
+  extensionPort.onDisconnect.addListener(resetSteamAndListeners);
+}
+
+/**
+ * Sets up two-way communication streams between the
+ * browser extension and local per-page browser context.
+ *
+ * @param {PortStream} extensionStream
+ * @todo LegacyProvider: Delete
+ */
```
digiwand, line 195: @todo merge latest comments added in #15492

app/scripts/contentscript.js
```diff
@@ -68,94 +69,174 @@ function injectScript(content) {
...
+
+  pageStream.allowHalfOpen = false;
+  pageStream.removeAllListeners();
+  pageStream.destroy();
+};
+
+/**
+ * Resets the extension stream with new streams and attach event listeners to the extension port.
+ */
+const resetSteamAndListeners = () => {
```
darkwing, line 201: It's not causing any issues but `resetSteamAndListeners` is missing a `r`; should be `resetStreamAndListeners`

app/scripts/contentscript.js
```diff
@@ -276,3 +370,19 @@ function redirectToPhishingWarning(data = {}) {
...
 }
+
+const start = () => {
+  if (
+    window.location.origin === phishingPageUrl.origin &&
+    window.location.pathname === phishingPageUrl.pathname
+  ) {
+    setupPhishingStream();
+  } else if (shouldInjectProvider()) {
+    if (!isManifestV3()) {
```
darkwing, line 381: Quick note for when you rebase -- `isManifestV3` has become a boolean, so you no longer need the `()`

app/scripts/contentscript.js
```diff
@@ -68,122 +71,225 @@ function injectScript(content) {
...
+/**
+ * Destroying the destination streams of the pump (should) destroy the streams before it
+ *
+ * @todo
+ * - Remove text (should) above after confirming this works
+ * - Answer questions:
+ *   1. is it necessary to manually set allowHalfOpen to false?
+ *      "If false then the stream will automatically end the writable side when the readable side ends"
+ *   2 is it necessary to removeAllListeners if we're calling destroy? - I believe calling destroy
+ *     removes all listeners.
```
digiwand: note: I do not think calling `destroy()` removes its listeners as I ran into the stream disconnect error when commenting out `removeAllListeners();`. Keeping `removeAllListeners();` in does not produce the following console error.

Screenshot 2022-08-19 at 3 25 26 AM

darkwing:
Wow, this appears to be working great!

digiwand:
[out of scope] confirmed that the provider error is intermittent after seeing it pass here 

@jpuri @Gudahtt 

https://github.com/MetaMask/metamask-extension/pull/15494/commits/14db9cc4ff8a656f90bc2b96a08c38fc60d9feff

 Screenshot 2022-08-23 at 3 02 59 AM

jpuri:
Hey @digiwand : small suggestion using `/` in branch name creates trouble in running commands. Try to avoid that.

digiwand:
Hey @jyoti, thanks very much for bringing this up! I didn't know it was causing difficulties to work with. I'll avoid adding `/` in the branch names from here on

jpuri:
@digiwand : service worker seems to always stay active after these changes even is extension is not used, do you plan to fix that.

I think we should keep service worker running only for connected dapps.

digiwand:
@jpuri I agree we should only keep the service worker running for connected dapps, when the pop-up is open, and when the home.html page is open. 

Let me double-check this...

digiwand:
Hey @jpuri, good catch. I think this should fix it https://github.com/MetaMask/metamask-extension/pull/15494/commits/d259d04d8c5c9226c0b98bb7397fdb485dbdaa08 (merged commit).

digiwand:
Hey @jpuri, good catch. I think this fixes it https://github.com/MetaMask/metamask-extension/commit/d259d04d8c5c9226c0b98bb7397fdb485dbdaa08 (merged commit)

seaona:
Overall behavior looks fine :) . I have seen a couple of console errors that might be worth to double-check:

1. On the dev console, I'm seeing the error `write after end` on `stream-utils` at the moment of loading the Extension

https://user-images.githubusercontent.com/54408225/188630588-6f5af261-f95a-4c5f-a29d-dddb404dd0b4.mp4

2. After Importing a Wallet,  I see the error `document is not defined` on the dev console (this will also appear when I stop service worker and re-initiate - see point 4)

Screenshot from 2022-09-06 14-06-06

3. When I Connect to Test Dapp and try to perfom a Send tx or to Deploy Contract, I see the error `failed to parse tx data`. This also happens when I Submit a tx.

https://user-images.githubusercontent.com/54408225/188631337-19c965cc-40dc-4e44-968a-e64427d10407.mp4

4. When I stop the Service Worker, I see the following errors:
    - `Duplicate script ID 'inpage'`
    - `Incorrect password`
    - ` ReferenceError: document is not defined`


https://user-images.githubusercontent.com/54408225/188632278-86412277-5403-4188-ab69-8ab0e71ff661.mp4

In this case, when I stop service worker MM is logged out and I need to log in again with the password. Is this the expected behaviour?

digiwand:
@seaona Thank you for the thorough testing! I think there is some crossover testing of https://github.com/MetaMask/metamask-extension/pull/14781 and this PR, which is actually great you tested both. I believe none of these are blockers to this PR. This caught some out-of-scope issues, and I created one follow-up issue: 
https://github.com/MetaMask/metamask-extension/issues/15775

1. [related to other service worker PR] This is actually testing service worker behavior for the home.html page and not the dapp pages. This PR re-activates streams for `script/contentscript.js` which is loaded in dapp pages which summons popup.html. Then https://github.com/MetaMask/metamask-extension/pull/14781 re-activates streams for `script/ui.js` which is loaded on our extension home.html page cc: @jpuri 

2. [unrelated] This `document is not defined` bug appears to be related to the Ledger iframe which will be supported in another PR 
https://github.com/MetaMask/metamask-extension/issues/14183. For now, I've been loading a wallet with no hard wallets associated with it.

3. [I think out of scope] Nice catch. I'm seeing this issue on both home.html and popup.html. This may or may not be related to the service worker and streams. I created an issue ticket to handle this separately from this PR: https://github.com/MetaMask/metamask-extension/issues/15775

4. When I stop the Service Worker, I see the following errors:
- [non-blocker] Duplicate script ID 'inpage': I've also observed this on both home.html and dapp pages. I don't believe this is a blocker. This error appears to be more related to the service worker implementation than this PR - though, in a way, it may be related
- [unrelated] Incorrect password: I believe this is being supported by https://github.com/MetaMask/metamask-extension/issues/15050
- [unrelated] ReferenceError: document is not defined
   Please refer to # 2 relating to hard wallet support

app/scripts/contentscript.js
```diff
@@ -276,3 +423,31 @@ function redirectToPhishingWarning(data = {}) {
...
+    window.location.pathname === phishingPageUrl.pathname;
+
+  if (isDetectedPhishingSite) {
+    initPhishingStreams();
+    return;
+  }
+
+  if (shouldInjectProvider()) {
+    if (isManifestV3) {
+      initKeepWorkerAlive();
```
Gudahtt, line 445: It looks like this will keep the service worker alive regardless of whether the current website is connected or not. This runs for manifest V3 builds despite the connected status.

Gudahtt: To properly stop this interval when a site isn't connected would be challenging. The contentscript isn't aware of the connected status right now. Hmm.

We could spy on messages from here to detect connect/disconnect events though. Or we could update the provider to send special "connect"/"disconnect" JSON-RPC notifications to the contentscript just for this purpose.

Gudahtt (approved):
LGTM!

There is some outstanding work here about _not_ keeping the service worker alive in all cases, but that does not need to block this PR. This is still a significant step forward for our MV3 build. Great work!

app/scripts/contentscript.js
digiwand: @Gudahtt 

I see. I confirmed your observations. Agreed we should look into starting this interval only when the site is connected in a separate PR as [you mentioned here](https://github.com/MetaMask/metamask-extension/pull/15494#pullrequestreview-1102798025). I created a new issue for this: https://github.com/MetaMask/metamask-extension/issues/15802

For clarification here, the service worker does terminate after some time when the pages themselves are closed. 

Repo steps to observe service worker terminating:
1. Inspect service workers (e.g. chrome://inspect/#service-workers)
2. Observe app-init.js service worker is not running. If it is running, be sure to close all `http://` or `https://` pages. Then, terminate the service worker.
3. Open dapp to run MM extension
4. Observe `app-init.js` service worker is running
5. Close dapp page and all other `http://` or `https://` pages
6. Wait about 5 minutes (current ETA for termination of inactive service worker)
7. Inspect service workers and observe `app-init.js` service worker is no longer running

jpuri:
Hey @digiwand , I found an issue while working on task of retrying DAPP requests, if I add a bit of delay in `resetStreamAndListeners` method by adding a timeout, stream re-connection does not happen.

Reconnection should always happen irrespective of delay.

jpuri (approved):

Gudahtt:
>I add a bit of delay in resetStreamAndListeners method by adding a timeout, stream re-connection does not happen.

To follow up from the call we just had discussing this: I suspect that this method of simulating a delayed startup is flawed. I would expect the method `resetStreamAndListeners` to execute synchronously without any delay even if the service worker was slow to start up - just that we'd see a timeout disconnect event after the fact. But I'm not confident in that guess.

Needs further testing. Suggested next steps for testing that was to add an artificial delay during service worker initialization, before streams were setup. No need to block this PR though.

PR #15507, opened 2022-08-09
added color prop to Box component

### LINKED ISSUE #213: Add color prop to Box component
### Description
Add the color prop to the Box component.

### Technical details
- Add the color prop from the Typography component to the box component

### Acceptance criteria 
- Box component can accept `color` prop

NidhiKJha:
Fixes : [Add color prop to Box component](https://github.com/MetaMask/design-tokens/issues/213)

After:
Screenshot 2022-08-09 at 8 16 22 PM

### DIFF
--- a/ui/components/ui/box/README.mdx
+++ b/ui/components/ui/box/README.mdx
@@ -155,10 +155,6 @@ import Box from '../ui/box';
 
 Use the `backgroundColor` prop along with the `COLORS` object from `ui/helpers/constants/design-system.js` to change background color. The `backgroundColor` prop accepts [responsive props](#responsive-props) in the form of array props.
 
-<<<<<<< HEAD
-=======
-
->>>>>>> 664f9a50f (fixed readme and storybook for box props)
 <Canvas>
   <Story id="ui-components-ui-box-box-stories-js--background-color" />
 </Canvas>
--- a/ui/components/ui/box/README.mdx
+++ b/ui/components/ui/box/README.mdx
@@ -120,8 +120,6 @@ or array of these values.
 import { DISPLAY } from '../../../helpers/constants/design-system';
 import Box from '../ui/box';
 
-
-
 <Box display={DISPLAY.BLOCK} />
 <Box display={DISPLAY.FLEX} />
 <Box display={DISPLAY.GRID} />
@@ -131,10 +129,36 @@ import Box from '../ui/box';
 <Box display={DISPLAY.NONE} />
 ```
 
+### Color
+
+The `color` prop can be used to set the color of the content in Box component. The `color` prop accepts [responsive props](#responsive-props) in the form of array props.
+
+<Canvas>
+  <Story id="ui-components-ui-box-box-stories-js--color" />
+</Canvas>
+
+Example of importing `COLORS` object with `Box` component
+
+```jsx
+import { COLORS } from '../../../helpers/constants/design-system';
+import Box from '../ui/box';
+
+<Box
+  backgroundColor={COLORS.BACKGROUND_DEFAULT}
+  color={COLORS.ERROR_DEFAULT}
+>
+  Text goes here
+</Box>;
+```
+
 ### Background Color
 
 Use the `backgroundColor` prop along with the `COLORS` object from `ui/helpers/constants/design-system.js` to change background color. The `backgroundColor` prop accepts [responsive props](#responsive-props) in the form of array props.
 
+<<<<<<< HEAD
+=======
+
+>>>>>>> 664f9a50f (fixed readme and storybook for box props)
 <Canvas>
   <Story id="ui-components-ui-box-box-stories-js--background-color" />
 </Canvas>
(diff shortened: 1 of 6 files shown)


ui/components/ui/box/box.js
```diff
@@ -72,10 +72,29 @@ export const BorderColors = [
...
+  COLORS.SECONDARY_INVERSE,
+  COLORS.ERROR_DEFAULT,
+  COLORS.ERROR_INVERSE,
+  COLORS.SUCCESS_DEFAULT,
+  COLORS.SUCCESS_INVERSE,
+  COLORS.WARNING_INVERSE,
+  COLORS.INFO_DEFAULT,
+  COLORS.INFO_INVERSE,
+];
+
```
brad-decker, line 92: I dont' think we should be declaring another COLORS constant here... If the main colors constant contains colors that are invalid for the Box component, use lodash's pick, or omit method. Duplicating that effort in the stories is worthwhile to avoid the confusion of multiple Colors constants in the code base. Also hopefully we work towards removing any unsupported colors from the Colors constant in design-system so that we no longer have to specify.

brad-decker (changes requested):

ui/components/ui/box/README.mdx
```diff
@@ -177,6 +177,30 @@ import Box from '../ui/box';
 </Box>;
 ```
 
+### Color
+
+The `color` prop can be used to set the color of the content in Box Component. The `color` prop accepts [responsive props](#responsive-props) in the form of array props.
```
georgewrmarshall, line 182: Component can use small c
```suggestion
The `color` prop can be used to set the color of the content in Box component. The `color` prop accepts [responsive props](#responsive-props) in the form of array props.
```

ui/components/ui/box/README.mdx
```diff
@@ -177,6 +177,30 @@ import Box from '../ui/box';
...
+
+<Canvas>
+  <Story id="ui-components-ui-box-box-stories-js--color" />
+</Canvas>
+
+**NOTE**: The `<Box/>` color combinations above follow our design system color rules and should cover most general UI applications. Click "Show code" to see the code example. Do not use the [deprecated colors](#deprecated-colors)
```
georgewrmarshall, line 188: We can remove this note. Deprecated colors have been removed. Can we also remove any other instances of it?

georgewrmarshall (commented):
Looking good! Could you move Color section of the docs before BackgroundColor in the MDX docs and in the story order?

georgewrmarshall:
Could we also fix up this story
Screen Shot 2022-08-09 at 11 08 02 AM

ui/components/ui/box/box.js
NidhiKJha: @brad-decker I think we should replace `colors` with `TextColors` to avoid confusion. Another thing is I have used the colors in this object which are the same as what we have in Typography. So, I am not sure if we should use the pick or omit method since it will also create a similar object for us.

ui/components/ui/box/README.mdx
NidhiKJha: Yes

ui/components/ui/box/box.js
brad-decker: @NidhiKJha I like the idea of consolidating these to  an *object* called TEXT_COLORS that we can use in the same way as COLORS in the design-system.js file. This way we can avoid declaring and exporting these in the component files themselves.

georgewrmarshall: Ooo yes please! Can we do this for the BackgroundColors => `BACKGROUND_COLORS` and BorderColors => `BORDER_COLORS` too? We can also update the `Typography` component to use them as well

georgewrmarshall (commented):
Could we also add `INHERIT: 'inherit'` to colors for text?

ui/components/ui/box/box.js
NidhiKJha: Updated!

ui/components/ui/box/box.js
```diff
@@ -18,64 +18,85 @@ import {
...
+  'SECONDARY_DEFAULT',
+  'SECONDARY_INVERSE',
+  'ERROR_DEFAULT',
+  'ERROR_INVERSE',
+  'SUCCESS_DEFAULT',
+  'SUCCESS_INVERSE',
+  'WARNING_INVERSE',
+  'INFO_DEFAULT',
+  'INFO_INVERSE',
+]);
```
brad-decker, line 92: These should be in design-system.js versus the Box component

brad-decker (changes requested):

ui/components/ui/box/box.js
NidhiKJha: Yeah, it should be. I have added it here since we were using it only in Box for now. But now I have moved it to `design-systems`

brad-decker (approved):

brad-decker:
Great work! Thanks for the revisions :)

georgewrmarshall (approved):
LGTM!

PR #15515, opened 2022-08-09
Replacing setInterval with chrome.alarm in Detect-Token controller

### LINKED ISSUE #15509: replace setInterval in detect-token controller

NiranjanaBinoy:
#15509 
As a part of MV3, we need to replace the `setTimeout` and `setInterval` functions with chrome Alarms throughout the code base. While creating chrome alarms, we can define `delayInMinutes` to show how many minutes we should wait before the first time the alarm is triggered; we can also define `periodInMinutes ` which is the number of minutes between alarms. Unlike `setIntervals` the time interval has to be mentioned in Minutes, not in milliseconds. Inside the `onAlarm` event listener, we can define the task to be executed during the alarm defined.

Manual Testing:

1. Log in to the MetaMask application; the detectNewtoken() function should be triggered once every 3 mins using breakpoints (For testing purposes you can locally update the `DEFAULT_INTERVAL_MV3` value to a lower value).
2.  It should clear the existing alarm and create a new alarm every time the user switch address, switch the network, and toggle `Enhanced token detection`.


NiranjanaBinoy:
The unit test is failing for `detect-token` controller test and `metamask-controller` test when trying to differentiate between MV3 and MV2 using `isManifestV3` from mv3.utils. I tried to mock both `isManifestV3` and also `browser.runtime.getManifest` but both scenarios are working. I have commented out, and [commited](https://github.com/MetaMask/metamask-extension/pull/15515/commits/01108feb776c1a790f3501673ccac5e7645b30c0) have the steps/ways I tired out to mock the functions.
The error message we are getting: 
` TypeError: _webextensionPolyfill.default.runtime.getManifest is not a function`

app/scripts/controllers/detect-tokens.js
```diff
@@ -194,6 +198,30 @@ export default class DetectTokensController {
...
+    });
+
+    chrome.alarms.onAlarm.addListener(() => {
+      const alarm = chrome.alarms.get(DETECT_TOKEN_ALARM);
+      if (alarm) {
+        this.detectNewTokens();
+      }
+    });
+  }
+
```
jpuri, line 224: We should use alarm only in scenario where we need code to keep running even after extension is dead as alarms are remembered by the browser

Nooknick7899: **> +      return;
+    }
+    chrome.alarms.clear(DETECT_TOKEN_ALARM);
+    chrome.alarms.create(DETECT_TOKEN_ALARM, {
+      delayInMinutes: DEFAULT_INTERVAL_MV3,
+      periodInMinutes: DEFAULT_INTERVAL_MV3,
+    });
+
+    chrome.alarms.onAlarm.addListener(() => {
+      const alarm = chrome.alarms.get(DETECT_TOKEN_ALARM);
+      if (alarm) {
+        this.detectNewTokens();
+      }
+    });
+  }
+**

NiranjanaBinoy: detect-token needs to be called only when the extension is open and unlocked. Does that mean we do not need to add chrome alarms for this controller?

segun (commented):
Asides @jpuri comments, looks good to me

app/scripts/controllers/detect-tokens.js
jpuri: Yes I think so, lets get more opinions on it.

jpuri (changes requested):
https://github.com/MetaMask/metamask-extension/pull/15515#discussion_r952172684

app/scripts/controllers/detect-tokens.js
NiranjanaBinoy: I will create a post on the MV3 channel regarding this concern

danjm: I think it makes sense in this case so that we don't increase the number of requests that we are making to tokens api, it also ensures that the tokens list gets updated regularly, and the service worker restart doesn't accidently cause missed updates to the list

app/scripts/controllers/detect-tokens.js
```diff
@@ -191,9 +195,47 @@ export default class DetectTokensController {
...
+        });
+      }
+    });
+    chrome.alarms.onAlarm.addListener(() => {
+      chrome.alarms.getAll((alarms) => {
+        const hasAlarm = alarms.some(
+          (alarm) => alarm.name === DETECT_TOKEN_ALARM,
+        );
+        if (hasAlarm) {
+          console.log('in triggered', hasAlarm);
```
NiranjanaBinoy, line 232: this will be removed in the future once

app/scripts/controllers/detect-tokens.js
```diff
@@ -191,9 +195,47 @@ export default class DetectTokensController {
...
+  rescheduleTokenDetectionPollingInMV3() {
+    if (!(this.isActive && this.selectedAddress)) {
+      return;
+    }
+    chrome.alarms.getAll((alarms) => {
+      const hasAlarm = alarms.some(
+        (alarm) => alarm.name === DETECT_TOKEN_ALARM,
+      );
+      if (!hasAlarm) {
+        console.log('in rescheduleTokenDetectionPollingInMV3 ', hasAlarm);
```
NiranjanaBinoy, line 219: this console.log will be removed in the future,

app/scripts/controllers/detect-tokens.js
```diff
@@ -10,9 +10,13 @@ import {
   TOKEN_STANDARDS,
 } from '../../../shared/constants/transaction';
 import { EVENT, EVENT_NAMES } from '../../../shared/constants/metametrics';
+import { DETECT_TOKEN_ALARM } from '../../../shared/constants/alarms';
+import { isManifestV3 } from '../../../shared/modules/mv3.utils';
 
 // By default, poll every 3 minutes
 const DEFAULT_INTERVAL = MINUTE * 3;
+// FOR DEV PURPOSE -- NEED TO BE UPDATED TO 3 BEFORE MERGING
+const DEFAULT_INTERVAL_MV3 = 0.2;
```
NiranjanaBinoy, line 19: This interval will be updated to 3 before merging it to develop.
The chrome.alarms need to have an interval greater than or equal to 1 in production env but in development env we can use a value less than one to aid testing.

seaona:
From QA point of view, a couple of comments/questions:
- When I Lock MM, I still see that on the Service Worker console, the `hasAlarm` console log is triggered non-stop. I am wondering if this alarm should be cleared, when MM is locked

 [alarms-token-detection-locked.webm](https://user-images.githubusercontent.com/54408225/189522864-476a54fc-bcb8-431c-87d1-a2836f09979a.webm)

- When I disable the Token Detection toggle, I still see that  the `hasAlarm` console log is triggered. I am wondering if this alarm should be cleared, once the functionality is disabled. This might be related to[ this other bug](https://github.com/MetaMask/metamask-extension/issues/15790) around the toggle: token detection still works while toggle is Disabled, so maybe it should be fixed separately.

[token-detection-toggle.webm](https://user-images.githubusercontent.com/54408225/189522932-23242d85-c7f4-440a-9dc4-bf9c49fa051c.webm)

- When I am on the MM popup view, I see that no `hasAlarm` or `in rescheduleTokenDetectionPollingInMV3` console.logs are ever triggered, even though Token Detection continues to work

[popup-alarm-not-showing.webm](https://user-images.githubusercontent.com/54408225/189523146-5d4828a0-b8cc-45c8-865e-a641573e7b53.webm)

app/scripts/controllers/detect-tokens.js
jpuri: I am not sure if the same controller context will be available when the controller reloads.

NiranjanaBinoy:
Thanks for the QA @seaona ❤️ I have added some clarifications and queries below.

> When I Lock MM, I still see that on the Service Worker console, the hasAlarm console log is triggered non-stop. I am wondering if this alarm should be cleared, when MM is locked

In MV2, the function seems to be getting triggered even when the MM is locked and pop-up is closed, but the scenario is handled inside the `detectNewtoken()` function, which returns if the pop is closed and the MM is locked.

> When I disable the Token Detection toggle, I still see that the hasAlarm console log is triggered. I am wondering if this alarm should be cleared, once the functionality is disabled. This might be related to https://github.com/MetaMask/metamask-extension/issues/15790 around the toggle: token detection still works while toggle is Disabled, so maybe it should be fixed separately.

During token detection v2 development, a decision was made to keep token detection active for Mainnet using the static list even when the toggle is turned off to preserve the existing behavior in Mainnet(This scenario is handled inside the `detectNewToken()`). I will verify this with @danjm regarding this scenario and update accordingly.

> When I am on the MM popup view, I see that no hasAlarm or in rescheduleTokenDetectionPollingInMV3 console.logs are ever triggered, even though Token Detection continues to work

Chrome alarm seems to be working in pop up view for me, Could you please let me know how often this is happening and if this is happening for a particular scenario?

app/scripts/controllers/detect-tokens.js
```diff
@@ -200,13 +245,17 @@ export default class DetectTokensController {
...
     if (!interval) {
       return;
     }
-    this._handle = setInterval(() => {
-      this.detectNewTokens();
-    }, interval);
+    if (isManifestV3) {
+      this.rescheduleTokenDetectionPollingInMV3();
+    } else {
+      this._handle && clearInterval(this._handle);
```
darkwing, line 254: Nit:  I think we've tried getting away from the `&&` pattern and gone with `if`'s.

app/scripts/controllers/detect-tokens.js
```diff
@@ -183,14 +188,54 @@ export default class DetectTokensController {
...
+          periodInMinutes: DEFAULT_INTERVAL_MV3,
+        });
+      }
+    });
+    chrome.alarms.onAlarm.addListener(this.alarmListener);
+  }
+
+  alarmListener = () => {
+    chrome.alarms.getAll((alarms) => {
+      const hasAlarm = alarms.some(
```
darkwing, line 230: Nit:  This `alarms.some` is used twice; I'm wondering if we can put this in a helper function.

darkwing (dismissed):
Overall this is great!  Just a few nits but is working perfectly!

seaona:
@NiranjanaBinoy thank you v much for your comments! I've double-checked the pop up behavior and it is working fine!!

Just one last thing, about calling the function while MetaMask being locked, co you confirm is it okay in the MV3 context?
I can see my token detected and address while logged out:

image

[logout-alarm.webm](https://user-images.githubusercontent.com/54408225/192251491-8e5e6848-3f7b-4265-bf9f-8a7b47ee2e35.webm)

app/scripts/controllers/detect-tokens.js
danjm: >I am not sure if the same controller context will be available when the controller reloads.

@jpuri do you mean that the `this` reference could be incorrect?

jpuri: yes, when service worker reloads a new instance of controller will be created

NiranjanaBinoy:
@seaona Thanks for confirming the pop-up behavior. ❤️
When `detectNewTokens()` is the first line in the function, we are checking if the application is open and unlocked if not, the `detectNewTokens` function will be returned/exited in the second line. So even if the function is triggered while MM is locked, the function will not be executed. This is the same as in MV2 as well.
The situation where we are having `detectedTokens` in the state while the user is locked out seems to exist in MV2 as well; the same goes for the other token details. We would probably have to deal with it as a separate issue unrelated to MV3.

seaona:
Sounds good! Thank you @NiranjanaBinoy !! :heart:

darkwing (dismissed):
Nicew ork @NiranjanaBinoy !

app/scripts/controllers/detect-tokens.js
brad-decker: Is this conversation resolved @jpuri / @NiranjanaBinoy ?

NiranjanaBinoy: @danjm is helping me to dig deeper into how the change in context during restart is affecting our results; during the initial testings, it seems to be generating correct results but confirming that.

segun: Is this resolved yet?

jpuri: This is not resolved yet. If we are storing the results in scope of the current alarm function it will not be useful when controller is re-initiated.

If data if very important we can check to store it in storage.

Also this can result in multiple alarms being initiated each time new instance of controller is created and they may continue to run in browser.

NiranjanaBinoy:
Closing this PR because if the service worker restarts every 5 mins, it will not have much impact on intervals less than 5 mins if the function that is scheduled is called during the interval as well.
Quoting related chat from slack [here](https://consensys.slack.com/archives/C01V1L10W2E/p1664226689035799?thread_ts=1664215606.626099&cid=C01V1L10W2E)

PR #15551, opened 2022-08-11
Ability to buy tokens with Coinbase Pay and Transak

nikoferro:
## Explanation

Current State: Only native currencies can be bought using on ramp through the buy button
After Change: Ability to purchase tokens defined in `BUYABLE_CHAINS_MAP` using Coinbase Pay and Transak

Changes:

- migrated old buy eth modal to a new one using the new popover component
- new modal now supports buying both native currencies and tokens
- if a token is defined as buyable on the `BUYABLE_CHAINS_MAP` on at least one provider, a buy button will show on the token overview page

## More Information

See ONRAMP-189 and ONRAMP-190

### Before
Screenshot 2022-09-07 at 16 33 19

### After
Screenshot 2022-09-07 at 16 37 25
Screenshot 2022-09-07 at 16 33 43
Screenshot 2022-09-07 at 16 34 11

## Manual Testing Steps

These changes apply only to Ethereum mainnet

- Import a token
- If token is on the [Transak list](https://github.com/MetaMask/metamask-extension/pull/15551/commits/ffbface1c125c1b9df03f1c33f80133879415673#diff-4e682f9ffe2eed5743a5d0b761ffdab503afaad448954bd69ce2f7c8bca000d1R266) or on the [Coinbase Pay list](https://github.com/MetaMask/metamask-extension/pull/15551/commits/ffbface1c125c1b9df03f1c33f80133879415673#diff-4e682f9ffe2eed5743a5d0b761ffdab503afaad448954bd69ce2f7c8bca000d1R309), it should display a "Buy" button
- Depending on which list it shows, it should give you the option to buy the token on said provider
- Once you click on the provider's "Continue to" button, a window on said provider should open, and the token to buy should already be preselected

## Pre-Merge Checklist


nikoferro:

app/scripts/lib/buy-url.js
```diff
@@ -119,17 +119,20 @@ const createMoonPayUrl = async (walletAddress, chainId) => {
...
+  // we now default to nativeCurrency instead of the 2 previous tokens + eth that we had before
+  const { nativeCurrency } = BUYABLE_CHAINS_MAP[chainId];
   const queryParams = new URLSearchParams({
     appId: COINBASEPAY_API_KEY,
     attribution: 'extension',
     destinationWallets: JSON.stringify([
       {
         address: walletAddress,
-        assets: coinbasePayCurrencies,
+        assets: symbol ? [symbol] : [nativeCurrency],
```
alaahd, line 135: I am just concerned if this change will limit the user to select a single asset at the provider checkout page instead of a list of allowed assets, we may need to differentiate between the list of assets and the default selected asset (in our case dictated by the symbol param)

app/scripts/lib/buy-url.js
```diff
@@ -57,16 +57,16 @@ const createWyrePurchaseUrl = async (walletAddress, chainId) => {
...
  * @returns String
  */
-const createTransakUrl = (walletAddress, chainId) => {
+const createTransakUrl = (walletAddress, chainId, symbol) => {
   const { transakCurrencies, network } = BUYABLE_CHAINS_MAP[chainId];
 
   const queryParams = new URLSearchParams({
     apiKey: TRANSAK_API_KEY,
     hostURL: 'https://metamask.io',
-    cryptoCurrencyList: transakCurrencies.join(','),
```
alaahd, line 68: any idea why we dropped `cryptoCurrencyList`

app/scripts/lib/buy-url.js
```diff
@@ -157,11 +161,11 @@ export default async function getBuyUrl({ chainId, address, service }) {
     case 'wyre':
       return await createWyrePurchaseUrl(address, chainId);
     case 'transak':
-      return createTransakUrl(address, chainId);
+      return createTransakUrl(address, chainId, symbol);
```
alaahd, line 164: I just feel the usage of `account` as parameter could be more readable and representative than using `address` at it could be confusing with token address.

app/scripts/lib/buy-url.js
nikoferro: Its a frictionless way to buy the selected asset coming from metamask. Theres not a way where we can display both the full list and a selected asset as a default. We either show them a list on the coinbase pay side, or we pre select a particular asset

_With a list it would be this way:_

TokenA -> List of currencies -> Buy Token A

_With this change:_

TokenA -> Buy Token A

Will ask this over slack so everyone can chime in

app/scripts/lib/buy-url.js
nikoferro: Same response as the Coinbase case. Its a _show list_ vs _preselect_ the asset option

alaahd: Ok sure, but I was thinking if we need to maintain `cryptoCurrencyList` prop as it is being used to purchase assets on the network level, but we can limit that list into an array of a single asset if the purchase is intended on token level, something like this: 
`cryptoCurrencyList: symbol ? [symbol] : transakCurrencies.join(',')`

ui/components/app/modals/deposit-token-modal/deposit-token-modal.container.js
```diff
@@ -0,0 +1,59 @@
...
+  getIsBuyableCoinbasePayChain,
+  getIsMainnet,
+  getIsTestnet,
+  getSelectedAddress,
+  getIsBuyableTransakChain,
+  getIsBuyableCoinbasePayToken,
+  getIsBuyableTransakToken,
+} from '../../../../selectors/selectors';
+import {
+  buyEth,
```
alaahd, line 14: I was thinking if it will be more contextual to create another action creator called `buyToken` instead of using `buyEth`

ui/components/app/modals/modal.js
```diff
@@ -111,7 +112,38 @@ const MODALS = {
       height: '100%',
     },
   },
-
+  DEPOSIT_TOKEN: {
```
alaahd, line 115: I think this is the old approach of how we trigger modals by dispatching an action, the new approach is more simplified by relying on local state change to trigger a conditional rendering of the modal component, a clear example of the modern approach can be seen in the following file `/metamask-extension/ui/components/ui/sender-to-recipient/sender-to-recipient.component.js` 
`      {showNicknamePopovers ? (
        <NicknamePopovers
          onClose={() => setShowNicknamePopovers(false)}
          address={checksummedRecipientAddress}
        />
      ) : null}`

ui/components/app/wallet-overview/token-overview.js
```diff
@@ -80,6 +90,30 @@ const TokenOverview = ({ className, token }) => {
...
+              onClick={() => {
+                trackEvent({
+                  event: 'Clicked Deposit: Token',
+                  category: EVENT.CATEGORIES.NAVIGATION,
+                  properties: {
+                    action: 'Home',
+                    legacy_event: true,
+                  },
+                });
+                dispatch(
```
alaahd, line 107: With the modern modal approach, this can be replaced by updating the value of a local state variable which will trigger the modal in the rendering block.

alaahd (changes requested):
Looking good, I have added few comments.

app/scripts/lib/buy-url.js
nikoferro: im probably missing the reason to maintain `cryptoCurrencyList` since even at a network level, the UX is actually the same. you end up landing on ETH as preselected since its the first of the list. since we are specifying the network, it doesn't actually make an impact, because the list is prefiltered anyways by the network. `transakCurrencies` list is still worth keeping, since using that list, we can reference which tokens are "buyable" on said provider

ui/components/app/modals/deposit-token-modal/deposit-token-modal.component.js
```diff
@@ -0,0 +1,184 @@
...
+      isBuyableTransakChain,
+      isTokenBuyableCoinbasePay,
+      isTokenBuyableTransak,
+    } = this.props;
+    const { t } = this.context;
+    const networkName = NETWORK_TO_NAME_MAP[chainId];
+    const { symbol } = this.props.token;
+
+    return (
+      <div className="page-container page-container--full-width page-container--full-height">
```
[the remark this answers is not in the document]
nikoferro: I 100% agree with your take, and i even considered myself. Reason why i didn't go down this path is that we were actually aiming for the "low hanging fruit" which was enabling token purchasing. unifying components, while it could have been the best approach, the amount of conditions could end up being higher than expected, making it not so useful to be shared components

app/scripts/lib/buy-url.js
nikoferro: after discussing this we agree to get rid of `cryptoCurrencyList` since it would be redundant in this case

ui/components/app/modals/deposit-token-modal/deposit-token-modal.container.js
nikoferro: after our talk, we agreed to create a duplicate `buyToken` function for the sake of context clarity. choices were:

- `buyEth` for ETH and Tokens (contextually confusing)
- rename `buyEth` to `buyAssets` (possibility of impacting code in unexpected ways)
- create a `buyToken` function (selected)

ui/components/app/modals/modal.js
nikoferro: agreed to keep the "old approach" for triggering modals. while it can be an overkill for a modal, dispatching this new modal, respects the same strategy used to show the ether modal. choices were:

- refactor ether modal (increased too much the scope)
- keep ether modal as it is, and use the local state approach for the tokens modal (confusing to have two strategies for pretty much the same modals)
- use the dispatch approach for both (selected)

ui/components/app/wallet-overview/token-overview.js
nikoferro: reply here: https://github.com/MetaMask/metamask-extension/pull/15551#discussion_r944673984

app/scripts/lib/buy-url.js
alaahd: yes I agree, omitting `cryptoCurrencyList` would even increase tokens coverage on specific chain and offload to maintain or provide the list from our side. make sense.

ui/components/app/modals/deposit-token-modal/deposit-token-modal.component.js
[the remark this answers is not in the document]
alaahd: resolved.

ui/components/app/modals/modal.js
alaahd: for consistency purposes, it would make sense to keep the same strategy, but moving forward, we need to deviate from using the dispatch approach when mounting/un-mounting modals

alaahd: resolved

ui/components/app/modals/deposit-token-modal/deposit-token-modal.component.js
```diff
@@ -0,0 +1,184 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import { EVENT } from '../../../../../shared/constants/metametrics';
+import { NETWORK_TO_NAME_MAP } from '../../../../../shared/constants/network';
+import Button from '../../../ui/button';
+import LogoCoinbasePay from '../../../ui/logo/logo-coinbasepay';
+import LogoTransak from '../../../ui/logo/logo-transak';
+
+export default class DepositEtherModal extends Component {
```
brad-decker, line 9: We aren't creating new class components, this should be refactored to use functional component pattern, and use hooks instead of a container. The .component and .container suffixes are deprecated, just create a single file that uses hooks for tapping into values in state.

ui/components/app/modals/modal.js
brad-decker: We need to convert to using the new popover component and not extending our deprecated code patterns, so i am unresolving this conversation and marking it as a requirement to move forward.

brad-decker: Can you say more as to why you have to refactor the old modal to implement a new one in the appropriate way?

ui/components/app/modals/deposit-token-modal/index.scss
```diff
@@ -0,0 +1,118 @@
+.deposit-token-modal {
```
brad-decker, line 1: Many of the styles in this file can be avoided by using our design system components, Box and Typography. Please attempt to use those for new components.

brad-decker (changes requested):

ui/components/app/modals/deposit-token-modal/deposit-token-modal.component.js
nikoferro: hi @brad-decker sorry for the delay on replying, i have been out for the last couple of weeks. idea a couple of weeks ago was to go for the low hanging fruit of adding these without any big changes, so thats why i choose to duplicate the component.

wasn't aware of the intention to deprecate some old code patterns so i used what was there as an example.

ended up working on a single/fits both use cases modal based on the design system components as you suggested. also migrated the ether logic to functional + hooks, since it didnt make sense to keep two separate modals with different code styling for this

ui/components/app/modals/modal.js
nikoferro: commented here https://github.com/MetaMask/metamask-extension/pull/15551#discussion_r964962200

btw i updated screenshots to show how it looks now that its using the design system components

nikoferro:
hey @brad-decker please let me know if you need anything else

app/scripts/lib/buy-url.js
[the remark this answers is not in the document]
brad-decker: Also with this change, the only purpose 'coinbasePayCurrencies' plays in the code base is as a check to see if coinbaseBayCurrencies exists and therefore coinbase is enabled for that chain. In that case I would suggest we add a new property to our settings that can be used for enabled checks instead of the existence of an array.

app/scripts/lib/buy-url.js
```diff
@@ -57,16 +57,16 @@ const createWyrePurchaseUrl = async (walletAddress, chainId) => {
...
-const createTransakUrl = (walletAddress, chainId) => {
+const createTransakUrl = (walletAddress, chainId, symbol) => {
   const { transakCurrencies, network } = BUYABLE_CHAINS_MAP[chainId];
 
   const queryParams = new URLSearchParams({
     apiKey: TRANSAK_API_KEY,
     hostURL: 'https://metamask.io',
-    cryptoCurrencyList: transakCurrencies.join(','),
-    defaultCryptoCurrency: transakCurrencies[0],
+    defaultCryptoCurrency: symbol || transakCurrencies[0],
```
brad-decker, line 69: With this change, the transakCurrencies array is now pointless, we should use nativeCurrency I think?

brad-decker (changes requested):

app/scripts/lib/buy-url.js
nikoferro: hi brad, took your suggestion and changed this:
`defaultCryptoCurrency: symbol || transakCurrencies[0],`
for this:
`defaultCryptoCurrency: symbol || nativeCurrency,`

but the array is not pointless actually, is used as a list of supported currencies that vary depending on the chainId
so having the list, allows us to check if XYZ token can be bought through transak (in this case)

went with the change still so we dont have to worry to keep the nativeCurrency on the first place in the array

ui/components/app/modals/deposit-modal/deposit-modal.js
```diff
@@ -0,0 +1,224 @@
+import PropTypes from 'prop-types';
```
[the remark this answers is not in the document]
nikoferro: good call, havent noticed that the popovers where at a different level in the file structure

app/scripts/lib/buy-url.js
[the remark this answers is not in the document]
nikoferro: the case with coinbase its a bit different, they need an array, so thats why we send an array, but the one item array allows us to skip a step in the purchase process

**before**:
only buying ETH was available as an action from the metamask side, but the 3 currency array was to enable the user to select another currency on a first step at the coinbase side

**now**:
since now multiple currencies can be selected on the metamask side, by sending a one item list, we skip the selection step on the coinbase side. this removes friction on the payment flow, since it removes a duplicate step which would be selecting the currency twice

the `coinbasePayCurrencies ` list, still has an important role, since just like transak, we use the list to check if that currency is available for purchase.

nikoferro:
@brad-decker changes suggested are ready, feel free to resolve the conversations, tried to explain the best i could why we should be keeping these lists

brad-decker (changes requested):
LGTM thanks @nikoferro -- The only thing now is you'll need to add your changes from the network.js file to the new network.ts file -- and the types might need to have further validation.

nikoferro:
@brad-decker thanks for the latest review, i have moved the changes from the old network.js to the new network.ts, should be good now

shared/constants/network.ts
```diff
@@ -529,7 +635,41 @@ export const BUYABLE_CHAINS_MAP: {
...
+      'STETH',
+      'SUSHI',
+      'SWAP',
+      'TXL',
+      'UNI',
+      'USDC',
+      'USDT',
+      'VRA',
+      'WBTC',
+      'YLD',
```
brad-decker, line 671: use the CURRENCY_SYMBOLS constant here so that the types aren't widened to strings.

shared/constants/network.ts
```diff
@@ -538,7 +678,106 @@ export const BUYABLE_CHAINS_MAP: {
...
+      'TRU',
+      'UMA',
+      'UNI',
+      'USDC',
+      'USDT',
+      'WBTC',
+      'WCFG',
+      'XYO',
+      'YFII',
+      'ZRX',
```
brad-decker, line 779: Have to use the CURRENCY_SYMBOLS constant so that types aren't widened to strings.

brad-decker (changes requested):

shared/constants/network.ts
nikoferro: @brad-decker change ready, didnt do it that way initially since i choose to follow the current format where just the native currency was a constant

brad-decker:
So, to get around that issue on the linting as well as to prevent adding a lot of constants to one that is meant for internal consumption (not just for detailing which currencies are supported) I suggest adding a new constant:

```ts
const SUPPORTED_CURRENCY__SYMBOLS = {
  ...CURRENCY_SYMBOLS,
  /* all currency symbols added in this PR */
} as const;
```
Then define a new type similar to CurrencySymbol type, and change transak and coinbaseCurrencies types to use the new type instead of CurrencySymbol. 

Let me know if this makes sense, if not i'd be happy to make a new commit with suggested changes.

nikoferro:
> Makes total sense, no worries! I can take a look at it tomorrow, if you want/can and you see an opportunity to finish this one, feel free to commit this change. If not, I ll take a look at it myself tomorrow. Either way thanks for taking a look at this today and the quick feedback 🙏

@brad-decker made these changes, should be ready now for when u have a sec to take a look

brad-decker (dismissed):
LGTM! Thank you so much for handling all of the feedback and cleaning up our codebase in the process. I'll get this championed for a second review and merge.

ui/components/app/deposit-popover/on-ramp-item.js
```diff
@@ -0,0 +1,67 @@
...
+import Typography from '../../ui/typography';
+import { COLORS, FRACTIONS } from '../../../helpers/constants/design-system';
+
+const OnRampItem = ({
+  logo,
+  title,
+  text,
+  buttonLabel,
+  onButtonClick,
+  hide,
```
digiwand, line 14:
```suggestion
  hide = false,
```

ui/components/app/deposit-popover/on-ramp-item.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Button from '../../ui/button';
+import Box from '../../ui/box';
+import Typography from '../../ui/typography';
+import { COLORS, FRACTIONS } from '../../../helpers/constants/design-system';
+
+const OnRampItem = ({
```
digiwand, line 8: 👍

digiwand (dismissed):
Nice! LGTM 

Tested with ETH, DAI, and WETC with Chrome x macOS. (Did not actually test buying tokens through the providers)

Verified display in dark mode: 
Screenshot 2022-09-16 at 1 04 46 PM

Checked right-to-left font. The previous UI did not seem to support right-to-left UI. This is out of scope
Screenshot 2022-09-16 at 12 54 03 PM

brad-decker (approved):

brad-decker:
@digiwand I committed one of your suggestions, needs a reapprove when you get a second.

digiwand (commented):
Sweet. Thanks @brad-decker

digiwand (approved):

PR #15552, opened 2022-08-11
Use a new service for serving token images

dan437:
## Explanation
Use a new service for serving token images. 

In some future iteration we will probably use this service always, not only when token detection is on.

## Screenshots
Token images when token detection is turned on:
image

## Manual Testing Steps
- Enable token detection in Advanced settings
- Go to Swaps on Ethereum mainnet
- Check the URLs of token images, they should use a URL like this: `https://static.metaswap.codefi.network/api/v1/tokenIcons/1/0x6b175474e89094c44da98b954eedeac495271d0f.png`

### DIFF
--- a/ui/hooks/useTokensToSearch.js
+++ b/ui/hooks/useTokensToSearch.js
@@ -57,7 +57,7 @@ export function getRenderableTokenData(
     (tokenData) => tokenData.address === address?.toLowerCase(),
   );
 
-  const usedIconUrl = iconUrl || tokenMetadata?.iconUrl || token?.image;
+  const usedIconUrl = tokenMetadata?.iconUrl || iconUrl || token?.image;
   return {
     ...token,
     primaryLabel: symbol,
--- a/ui/hooks/useTokensToSearch.js
+++ b/ui/hooks/useTokensToSearch.js
@@ -8,6 +8,7 @@ import {
   getCurrentCurrency,
   getSwapsDefaultToken,
   getCurrentChainId,
+  getUseTokenDetection,
 } from '../selectors';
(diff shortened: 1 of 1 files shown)


ui/hooks/useTokensToSearch.js
```diff
@@ -57,7 +59,10 @@ export function getRenderableTokenData(
     (tokenData) => tokenData.address === address?.toLowerCase(),
   );
 
-  const usedIconUrl = iconUrl || tokenMetadata?.iconUrl || token?.image;
+  const tokenIconUrl = useTokenDetection
+    ? tokenMetadata?.iconUrl || iconUrl
+    : iconUrl || tokenMetadata?.iconUrl;
```
mrbaguvix, line 64: Using terniary operator would make this code unclear and hard to read. Lets do something like:
```suggestion
  let tokenIconUrl = ''; 
  if(useTokenDetection){
    tokenIconUrl = tokenMetadata?.iconUrl || iconUrl
   }else{
   tokenIconUrl = iconUrl || tokenMetadata?.iconUrl;
   }
```

dan437: Thank you for the review! I agree that it's best to avoid more complex logic with ternary operators. We actually have another [PR](https://github.com/MetaMask/metamask-extension/pull/15562) that will use the new token images service even if `useTokenDetection` is false, so I'm closing this one.

dan437:
Closing this one in favor of https://github.com/MetaMask/metamask-extension/pull/15562, which will use the new token images service even if `useTokenDetection` is false.

PR #15553, opened 2022-08-11
[e2e] Refactor contract-interactions spec using Ganache seeder

seaona:
## Explanation

## More Information
The first task took care of preparing the contracts and the seeder, so it deploys correctly any contract we pass them into Ganache, from the background. First PR is already merged here [#14631](https://github.com/MetaMask/metamask-extension/pull/14631)

- Subtask of #15417

## Manual Testing Steps
`yarn test:e2e:single test/e2e/tests/contract-interactions.spec.js`

## Pre-Merge Checklist

### DIFF
--- a/test/e2e/tests/contract-interactions.spec.js
+++ b/test/e2e/tests/contract-interactions.spec.js
@@ -35,7 +35,10 @@ describe('Deploy contract and call contract methods', function () {
         );
 
         // calls and confirms a contract method where ETH is sent
-        await driver.clickElement('#depositButton');
+        const depositButton = await driver.findClickableElement(
+          '#depositButton',
+        );
+        await depositButton.click();
         await driver.waitUntilXWindowHandles(3);
         let windowHandles = await driver.getAllWindowHandles();
         const extension = windowHandles[0];
--- a/test/e2e/tests/contract-interactions.spec.js
+++ b/test/e2e/tests/contract-interactions.spec.js
@@ -11,22 +11,32 @@ describe('Deploy contract and call contract methods', function () {
       },
     ],
   };
+  const smartContract = 'hst';
   it('should display the correct account balance after contract interactions', async function () {
     await withFixtures(
       {
         dapp: true,
         fixtures: 'connected-state',
         ganacheOptions,
+        smartContract,
         title: this.test.title,
       },
-      async ({ driver }) => {
+      async ({ driver, contractRegistry }) => {
+        const contractAddress = await contractRegistry.getContractAddress(
+          smartContract,
+        );
         await driver.navigate();
         await driver.fill('#password', 'correct horse battery staple');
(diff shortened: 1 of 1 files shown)


test/e2e/tests/contract-interactions.spec.js
```diff
@@ -11,64 +11,40 @@ describe('Deploy contract and call contract methods', function () {
       },
     ],
   };
+  const smartContract = 'piggybank';
```
PeterYinusa, line 14: Can you use [this](https://github.com/MetaMask/metamask-extension/blob/develop/test/e2e/seeder/smart-contracts.js#L39) instead of the hardcoded string?

test/e2e/tests/contract-interactions.spec.js
```diff
@@ -11,64 +11,40 @@ describe('Deploy contract and call contract methods', function () {
...
 
         // deploy contract
-        await driver.openNewPage('http://127.0.0.1:8080/');
-        await driver.clickElement('#deployButton');
+        await driver.openNewPage(
+          `http://127.0.0.1:8080/?contract=${contractAddress}`,
+        );
+
+        // wait for deployed contract, calls and confirms a contract method where ETH is sent
+        await driver.findClickableElement('#deployButton');
```
PeterYinusa, line 38: Hmmm, I'm not sure about this as clickElement uses findClickableElement under the hood, so technically this step is redundant. I had a similar issue in #15481 and opted to use a delay instead

PeterYinusa (changes requested):

test/e2e/tests/contract-interactions.spec.js
PeterYinusa: SMART_CONTRACTS.PIGGYBANK

seaona: sure, adding this. Ty!!

test/e2e/tests/contract-interactions.spec.js
seaona: So these 2 functions are pointing to different elements:
 `findClickableElement` is pointing to `#deployButton`, whereas `clickElement` is pointing to `#depositButton`.

The reason for adding this previous step of `findClickableElement` is because I noticed that until `DeployButton` element is not clickable, the Dapp is not actually ready and clicking on any other element does not work. If you look closely at the video, you'll see that this element takes slightly a bit more than the rest to be enabled.

I was considering a delay, but thought this approach is more deterministic. I could change it for a delay if we think it's clearer though. Let me know what you think @PeterYinusa 

deploy-contract
deploy-contract

PeterYinusa: Oops, thanks for pointing this out. This looks good as is, no change necessary

PeterYinusa (approved):

brad-decker (approved):

Issue #14404, opened 2022-04-07
Write tests for Infura network client


mcmire:
As part of an effort to [merge the NetworkController that exists in this repo with the NetworkController that exists in the `controllers` repo](https://github.com/MetaMask/controllers/issues/753), we need to write tests for the JSON-RPC layer, and all of its custom behavior, that the NetworkController exposes.

We have two "network clients" that handle JSON-RPC network requests, one for Infura and one for custom networks. The custom network client tests are covered under issue #16041. This issue is for testing the network client that is specific to Infura.

Gudahtt:
The remaining improvements to the Infura client tests are:

* Add tests for methods defined in the Ethereum spec that are not supported by Infura (follow-up from #15556)
* Add tests for multi-parameter methods to ensure we get a cache miss when any single parameter changes (follow-up from #15556)
* Update `eth-json-rpc-middleware` and fix any tests that cover the old broken behaviour where duplicate requests are sent.
* Add tests for `retry-on-empty` middleware

Gudahtt:
Closing because the remaining work has been tracked in separate issues, referenced in the previous comment

PR #15556, opened 2022-08-11
Add initial provider API tests for Infura client

mcmire:
## Explanation

We are working on migrating the extension to a unified network
controller, but before we do so we want to extract some of the existing
pieces, specifically `createInfuraClient` and `createJsonRpcClient`,
which provide the majority of the behavior exhibited within the provider
API that the existing NetworkController exposes. This necessitates that
we understand and test that behavior as a whole.

With that in mind, this commit starts with the Infura-specific network
client and adds some initial functional tests for `createInfuraClient`,
specifically covering three pieces of middleware provided by
`eth-json-rpc-middleware`: `createNetworkAndChainIdMiddleware`,
`createBlockCacheMiddleware`, and `createBlockRefMiddleware`.

These tests exercise logic that originate from multiple different places
and combine in sometimes surprising ways, and as a result, understanding
the nature of the tests can be tricky. I've tried to explain the logic
(both of the implementation and the tests) via comments. Additionally,
debugging why a certain test is failing is not the most fun thing in the
world, so to aid with this, I've added some logging to the underlying
packages used when a request passes through the middleware stack.
Because some middleware change the request being made, or make new
requests altogether, this greatly helps to peel back the curtain, as
failures from Nock do not supply much meaningful information on their
own. This logging is disabled by default, but can be activated by
setting ~`DEBUG_PROVIDER_TESTS=1` alongside the `jest` command~
`DEBUG=metamask:*,eth-query DEBUG_COLORS=1` alongside the `jest` command.

Also note that we are using a custom version of `eth-block-tracker`
which provides a `destroy` method, which we use in tests to properly
ensure that the block tracker is stopped before moving on to the next
step. This change comes from [this PR][1] which has yet to be merged.

[1]: https://github.com/MetaMask/eth-block-tracker/pull/106

## More Information

The issue this is directly related to is:

* https://github.com/MetaMask/metamask-extension/issues/14404

Migrating the network controller to a shared architecture is part of a larger effort to add multi-network support to both extension and mobile. You can read more about that here: 

* https://github.com/MetaMask/controllers/issues/753
* https://www.notion.so/6f394364046d426690471a977aec1e6f?v=6bb3deb1e14243a5b34b390ea3292dff&p=f12696078d0245cdb19f1d4428d818d4&pm=c

## Manual Testing Steps

Run `yarn jest app/scripts/controllers/network/createInfuraClient.test.js`. All tests should pass.

Alternatively if you want to see logging, run `DEBUG=metamask:*,eth-query DEBUG_COLORS=1 yarn jest app/scripts/controllers/network/createInfuraClient.test.js`. Again, all tests should pass.

## Pre-Merge Checklist


app/scripts/controllers/network/createInfuraClient.test.js
```diff
@@ -0,0 +1,830 @@
...
+    'eth_getUncleCountByBlockHash',
+    'eth_getUncleCountByBlockNumber',
+    'eth_protocolVersion',
+    'shh_version',
+    'test_blockCache',
+    'test_forkCache',
+    'test_permaCache',
+    'web3_clientVersion',
+    'web3_sha3',
+  ])('when the RPC method is %s', (method) => {
```
Gudahtt, line 146: TIL that `describe.each` exists

mcmire: I decided not to use this as it made the tests less easy to extend later.

mcmire:
I've redone these tests as I didn't like how I did them originally. I ended up dropping `describe.each` and creating functions that define tests for a particular kind of RPC method. It might not be a technique that we commonly use to write tests, but it seems to work well for this particular use case.

mcmire:
Following standup today, this PR is blocked on

* Publishing a new release of `eth-block-tracker` (RC here: https://github.com/MetaMask/eth-block-tracker/pull/111)
* Adding logging to `eth-json-rpc-engine`, `eth-json-rpc-middleware`, `eth-json-rpc-infura`, and `eth-query`, then publishing new releases for these packages
* Bumping the versions of these packages in this PR

patches/eth-query+2.1.2.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/node_modules/eth-query/index.js b/node_modules/eth-query/index.js
+index 13e9f3c..303d703 100644
+--- a/node_modules/eth-query/index.js
```
mcmire, line 3: Note that I am still adding a patch for `eth-query`. I'm guessing kumavis still has access to this, but getting a change made might take a while...

package.json
```diff
@@ -161,11 +162,10 @@
     "debounce-stream": "^2.0.0",
     "deep-freeze-strict": "1.1.1",
     "end-of-stream": "^1.4.4",
-    "eth-block-tracker": "^5.0.1",
+    "eth-block-tracker": "^6.0.0",
     "eth-ens-namehash": "^2.0.8",
     "eth-json-rpc-filters": "^4.2.1",
-    "eth-json-rpc-infura": "^5.1.0",
```
mcmire, line 167: We are bumping this finally 😅 I have updated `createInfuraClient` to account for the change in usage (default import -> named import).

mcmire: This is the most impactful change in this PR as `@metamask/eth-json-rpc-infura` 6.0.0 bumped `eth-json-rpc-middleware`, `eth-rpc-errors`, and `json-rpc-engine` ([see changelog entry](https://github.com/MetaMask/eth-json-rpc-infura/blob/main/CHANGELOG.md#600---2022-05-04)). I have gone over the changelogs for these and there don't appear to be any changes that we need to worry about but I probably need a second set of eyes on this 👀

mcmire:
@Gudahtt This is ready for review again.

package.json
Gudahtt: For `eth-json-rpc-middleware`, the change in version 8 that adds `eth_signTransaction` stood out to me. We should test this to ensure this is blocked. I'm pretty sure that it is blocked in our method middleware but haven't manually teseted.

patches/eth-query+2.1.2.patch
Gudahtt: `EthQuery` is on the way out anyway. We'll be replacing it with ethers soon-ish I hope, maybe after multi-network.

package.json
Gudahtt: This all seems good to me!

app/scripts/controllers/network/provider-api-tests/helpers.js
```diff
@@ -0,0 +1,332 @@
...
+ */
+
+/**
+ * @typedef {{ nockScope: NockScope, request: object, response: object, delay?: number }} MockSuccessfulInfuraRpcCallOptions
+ *
+ * The options to `mockSuccessfulInfuraRpcCall`.
+ */
+
+/**
+ * @typedef {{mockNextBlockTrackerRequest: (options: Omit<MockNextBlockTrackerRequestOptions, 'nockScope'>) => void, mockSuccessfulInfuraRpcCall: (options: Omit<MockSuccessfulInfuraRpcCallOptions, 'nockScope'>) => NockScope}} InfuraCommunications
```
Gudahtt, line 53: Is this type used anywhere?

Gudahtt: The next two appear to be unused as well

package.json
mcmire: Ah, okay. I'll check this then!

app/scripts/controllers/network/provider-api-tests/helpers.js
mcmire: It is being used — the names were just wrong. These have been corrected in [d9327ce](https://github.com/MetaMask/metamask-extension/pull/15556/commits/d9327ceed18649492056402b67276a4764c399d7).

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -0,0 +1,677 @@
...
+        comms.mockSuccessfulInfuraRpcCall({
+          request: buildRequestWithReplacedBlockParam(
+            requests[0],
+            blockParamIndex,
+            '0x100',
+          ),
+          response: { result: mockResults[0] },
+        });
+        // Note that the block-ref middleware will still allow the original
+        // request to go through.
```
Gudahtt, line 375: This is... surprising. How does this work?

It sounds like you're saying the Infura middleware makes _two_ requests for each request, at least for those that use `latest` or have no block tag.

Gudahtt: I looked this up and found that this is the case: https://github.com/MetaMask/eth-json-rpc-middleware/blob/53c7361944c380e011f5f4ee1e184db746e26d73/src/block-ref.ts#L58

This.... seems wrong. This "child request" with a corrected block number will presumably have its response overwritten by the "real" request.

Gudahtt (commented):
I've finished reviewing all of the code/tests, and they all look great! Generally I agree with the advice of that Jest rule about not exporting `describe`/`it` blocks in the manner you've done here, but I can see why you made an exception in this case. It does make for exceptionally concise tests, that don't take _that_ much digging to understand more deeply.

Next I will be checking to ensure all methods/behaviours that we care about are covered.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
Gudahtt: I have confirmed that this is the case. I updated the test to return a bad result for the first of these two semi-duplicated requests, and it was ignored.

Gudahtt: Looks like this is a bug that was introduced here: https://github.com/MetaMask/eth-json-rpc-middleware/commit/76d676566b0b2be75bc0e952789f4442990a0daf#diff-00d71f9a810b6015201c9aae4ec2d2a5428cc7e2d7745e4267bc36ad58d86bf6R42

app/scripts/controllers/network/createInfuraClient.test.js
```diff
@@ -0,0 +1,232 @@
...
+
+  describe('eth_compileSerpent', () => {
+    testsForRpcMethodThatDoesNotSupportParams('eth_compileSerpent');
+  });
+
+  describe('eth_compileSolidity', () => {
+    testsForRpcMethodThatDoesNotSupportParams('eth_compileSolidity');
+  });
+
+  describe('eth_estimateGas', () => {
```
Gudahtt, line 78: This method does support parameters (see https://docs.infura.io/infura/networks/ethereum/json-rpc-methods/eth_estimategas)

app/scripts/controllers/network/createInfuraClient.test.js
```diff
@@ -0,0 +1,232 @@
...
+    testsForRpcMethodThatDoesNotSupportParams('eth_estimateGas');
+  });
+
+  describe('eth_gasPrice', () => {
+    testsForRpcMethodThatDoesNotSupportParams('eth_gasPrice');
+  });
+
+  describe('eth_getBlockByHash', () => {
```
Gudahtt, line 86: This method does support parameters (see https://docs.infura.io/infura/networks/ethereum/json-rpc-methods/eth_getblockbyhash)

app/scripts/controllers/network/createInfuraClient.test.js
```diff
@@ -0,0 +1,232 @@
...
+  //
+  // The fact that these methods support arguments affects the caching strategy,
+  // because if two requests are made with the same method but with different
+  // arguments, then they will be cached separately. Also, the block param
+  // changes the caching strategy slightly: if "pending" is specified, then the
+  // request is never cached.
+
+  describe('eth_getBlockByNumber', () => {
+    testsForRpcMethodThatSupportsMultipleParams('eth_getBlockByNumber', {
+      numberOfParams: 0,
```
Gudahtt, line 199: Curiously, this method has two required parameters, but the first one is the block parameter.

Saying that the "number of parameters" is zero here is a bit confusing. I'm guessing this was set to zero because this method breaks with the convention of the block parameter being last.

app/scripts/controllers/network/createInfuraClient.test.js
Gudahtt: Looks like this is another case where the position of the block ref has confused matters. This method does support parameters, but the blockRef is the first one.

app/scripts/controllers/network/createInfuraClient.test.js
Gudahtt: This is a bit confusing for the methods that support a blockRef, even in cases where it is last, because the true "number of parameters" is still wrong. e.g. `eth_call` is set as supporting one parameter, but it really has two including the block parameter.

Gudahtt: Similarly, methods that support a block parameter maybe shouldn't be described as having no parameters (e.g. `eth_getBlockTransactionCountByHash`)

app/scripts/controllers/network/createInfuraClient.test.js
Gudahtt: Similar case with `eth_getTransactionByBlockHashAndIndex` and `eth_getTransactionByBlockNumberAndIndex`

app/scripts/controllers/network/createInfuraClient.test.js
```diff
@@ -0,0 +1,232 @@
...
+    );
+  });
+
+  describe('eth_getTransactionByBlockNumberAndIndex', () => {
+    testsForRpcMethodThatDoesNotSupportParams(
+      'eth_getTransactionByBlockNumberAndIndex',
+    );
+  });
+
+  describe('eth_getTransactionByHash', () => {
```
Gudahtt, line 122: Both of these two methods take a "transaction hash" parameter

Gudahtt:
I noticed that tests are missing for the methods `eth_feeHistory` and `eth_getLogs`

app/scripts/controllers/network/provider-api-tests/shared-tests.js
mcmire: This was surprising to me as well. I've found a couple of other bugs along the way that I suspect may have been introduced in TypeScript conversions/refactorings.

app/scripts/controllers/network/createInfuraClient.test.js
mcmire: Ah, yes, this should probably be more like `blockRefParamIndex`. I can fix this.

mcmire:
@Gudahtt A few of the comments you've left are related — the Ethereum spec says one thing but the tests I've written say something else — so I'll try to answer those here. In writing these tests I merely copied the behavior in the middleware. If the middleware did not look for a block ref at all for a particular method, I wrote that method as taking no parameters, and I ignored any other parameters. Also, I didn't test the full gamut of the Ethereum spec because there are some methods we apparently don't care about (even though maybe we should). 

So the question is, should we write these tests from the perspective of the middleware or from the perspective of the Ethereum spec (or, in the case of the Infura middleware, what Infura supports)?

Gudahtt:
So part of the issue is semantic confusion, which affects test readability. It's confusing to describe a method as having no parameters if it does in fact have them. But I see you left a comment about using `blockRefParamIndex` instead of the "number-of-parameters" approach, maybe that sort of approach could solve this problem (i.e. renaming things to match what they really mean).

But these do represent genuine gaps in what is tested v.s. the real middleware functionality as well. For the methods that the middleware don't have special handling built-in for, it is still worth tracking how they behave. We want test coverage for the "default" case, and to track when known methods are changed to be cached or not, that sort of thing. And for the cases where the block ref is the first parameter, there is a gap here in that the caching behaviour when parameters are supplied isn't being tested.

app/scripts/controllers/network/provider-api-tests/helpers.js
```diff
@@ -0,0 +1,332 @@
...
+/**
+ * Sets up request mocks for requests to Infura.
+ *
+ * @param {MockingInfuraCommunicationsArgs} args - Either an options bag + a
+ * function, or just a function. The options bag, at the moment, may contain
+ * `network` (that is, the Infura network; defaults to "mainnet"). The function
+ * is called with an object that allows you to mock different kinds of requests.
+ * @returns {Promise<any>} The return value of the given function.
+ */
+export async function mockingInfuraCommunications(...args) {
```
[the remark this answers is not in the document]
mcmire: That's a good idea. I named it this way so I could give a reasonable name to the type of the object that the given function is called with (`InfuraCommunications`). But no reason I can't use `with` here. I'll fix this.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
Gudahtt: Documented here: https://github.com/MetaMask/eth-json-rpc-middleware/issues/142

I see you raised an issue in that repo already. Was there anything else that hasn't yet been documented?

I do think we should try to fix these issues ASAP. These two seem easy to fix, and the impact might be _very challenging_ to debug if it ever did become an issue in production. It might go a long way towards making our network stack less haunted.

Gudahtt:
In addition to "known Infura methods", we should add one "fake" method as a stand-in for any network provider methods that we don't support (e.g. non-Ethereum networks). So that we have set of tests tracking the behaviour of anything not otherwise covered.

Gudahtt:
I spent some time investigating whether our middleware supports an object as `params`. Certain provider methods use an object rather than an array, like `wallet_watchAsset`. It seems that our middleware has never supported it though, so we're good here. Maybe all existing use cases for object parameters are covered by other middleware ahead of here in the RPC pipeline.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
mcmire: Alright. I'll tag this PR as blocked until we can resolve those issues.

The other bugs I found have to do with the fact that the set of methods which take block tags as defined [here](https://github.com/MetaMask/eth-json-rpc-middleware/blob/53c7361944c380e011f5f4ee1e184db746e26d73/src/utils/cache.ts#L53) is incomplete. This means some methods aren't cached properly when they should be. (This list is used indirectly by not only `block-cache` but by a couple of other caching-related middleware as well.)

Gudahtt: Hmm, OK. That does seem like a bug but not a regression, and not nearly as spooky as the broken retry or the block-ref double-request, both of which were techniques used to guard against inconsistencies resulting from communication with load-balanced servers with different internal state (e.g. from one request to anther, the "latest block" could decrease then increase again as requests were routed to different servers).

mcmire:
@Gudahtt I have the feeling that we don't support an object as `params` anywhere. All of the code that interacts with `params` that I've seen — extension, `eth-json-rpc-middleware`, even tests for `json-rpc-engine` — assume that `params` is an array. I don't believe I've ever seen a case where we check for whether `params` is an object. That's interesting though. 

If we are using the Infura/Ethereum docs as the source of truth instead of the middleware, for RPC methods that take multiple parameters, should we have tests which exercise all of the parameters? Obviously we are still mocking the network call itself so the best we can do is just test that the network call gets made correctly, but just in case any of the middleware fiddle with parameters other than the block parameter maybe it's worth it?

app/scripts/controllers/network/provider-api-tests/shared-tests.js
Gudahtt: I didn't mean we should block this PR on fixing those bugs, just that it was important they get fixed. Fixing the bugs would simplify these tests, but we can always simplify them later.

mcmire: Ah, sounds good, okay.

Gudahtt:
>@Gudahtt I have the feeling that we don't support an object as params anywhere. All of the code that interacts with params that I've seen — extension, eth-json-rpc-middleware, even tests for json-rpc-engine — assume that params is an array. I don't believe I've ever seen a case where we check for whether params is an object. That's interesting though.

We do in a few places, but the examples I know of are not part of the network middleware. They're handled earlier. Here are the two I know of:
* https://github.com/MetaMask/metamask-extension/blob/c825a481c5b6dc6849a9271811746211d548788a/app/scripts/lib/rpc-method-middleware/handlers/watch-asset.js#L39
* https://github.com/MetaMask/eth-json-rpc-filters/blob/d2276f11205af4f271422fc97514df660cdf5351/log-filter.js#L54

>should we have tests which exercise all of the parameters?

Hmm. Unsure. I think for these tests, all we care about is that they're all forwarded correctly, and that the parameters are all part of the cache key (i.e. that a cached result is only used if all parameters are the same).

I guess to be exhaustive, we would want to try modifying each parameter independently to ensure it is considered for caching purposes :thinking:

mcmire:
@Gudahtt Okay, I've reorganized the tests in `createInfura.test.js`, cross-referencing the RPC methods listed in the Infura docs and those listed in the Ethereum spec. There are some similarities and differences so I tried to note those by organizing the tests into groups. I also backfilled tests for RPC methods not previously tested. Hopefully that addresses your comments but let me know if I missed something.

app/scripts/controllers/network/createInfuraClient.test.js
mcmire: Fixed in https://github.com/MetaMask/metamask-extension/pull/15556/commits/f815a103f8bf7077869f8119365e2d14fc50d4ed.

app/scripts/controllers/network/createInfuraClient.test.js
mcmire: Hopefully fixed in https://github.com/MetaMask/metamask-extension/pull/15556/commits/f815a103f8bf7077869f8119365e2d14fc50d4ed.

app/scripts/controllers/network/createInfuraClient.test.js
mcmire: I haven't fixed this yet, I'll revisit this.

app/scripts/controllers/network/createInfuraClient.test.js
mcmire: I still need to take care of this as well.

mcmire:
Update: I didn't check back over the methods I had already written tests for to see whether the number of arguments I'm using matches up with the Infura docs, so I'll go back and do that.

app/scripts/controllers/network/provider-api-tests/helpers.js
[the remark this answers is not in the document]
mcmire: This was fixed in [d131ec3](https://github.com/MetaMask/metamask-extension/pull/15556/commits/d131ec37fa917c24fd78662f888a7aedb3044073).

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -0,0 +1,708 @@
...
+import { fill } from 'lodash';
+import {
+  withMockedInfuraCommunications,
+  withInfuraClient,
+  buildMockParamsWithoutBlockParamAt,
+  buildMockParamsWithBlockParamAt,
+  buildRequestWithReplacedBlockParam,
+} from './helpers';
+
+export function testsForRpcMethodNotHandledByMiddleware(
```
Gudahtt, line 12: Nit: it might be better to call this something like `testsForRpcMethodNotCachedByMiddleware`. It's still _handled_, just not cached.

Gudahtt: Also worth having a test to ensure it's not cached. Like the "identical requests" tests used elsewhere, or the "pending" block tag test.

app/scripts/controllers/network/createInfuraClient.test.js
```diff
@@ -0,0 +1,324 @@
...
+
+    describe('eth_coinbase', () => {
+      testsForRpcMethodNotHandledByMiddleware('eth_coinbase', {
+        numberOfParameters: 0,
+      });
+    });
+
+    describe('eth_estimateGas', () => {
+      testsForRpcMethodAssumingNoBlockParam('eth_estimateGas', {
+        numberOfParameters: 0,
```
Gudahtt, line 57: This function doesn't take any options. Just the RPC method

app/scripts/controllers/network/createInfuraClient.test.js
Gudahtt: I can resolve this thread now that the helper function was renamed. We still need to test the parameter caching behavior, but that's true of everything.

app/scripts/controllers/network/createInfuraClient.test.js
```diff
@@ -0,0 +1,324 @@
...
+        numberOfParameters: 1,
+      });
+    });
+
+    describe('web3_clientVersion', () => {
+      testsForRpcMethodAssumingNoBlockParam('web3_clientVersion');
+    });
+  });
+
+  // NOTE: The following methods are omitted because although they are listed in
```
Gudahtt, line 304: Nit: I don't think it would hurt to include tests for these methods as well.

I didn't directly respond to your "What should be the source of truth" question before because I didn't really understand what you meant. I don't think we should consider either resource as a "source of truth" on what we should test, but rather a combination of factors. I think it's useful to test anything we have special behaviour for (e.g. the methods cached by the middleware), and any we anticipate adding special behaviour for in the future (Ethereum standard methods, and maybe more eventually). But this doesn't need to be a blocker.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
Gudahtt: >it might be better to call this something like

Actually... on second thought maybe the original name is better. The middleware does more than caching.

app/scripts/controllers/network/createInfuraClient.test.js
mcmire: Whoops! Fixed in [332a958](https://github.com/MetaMask/metamask-extension/pull/15556/commits/332a95825eb636351d7e080c067aed317e918b38).

app/scripts/controllers/network/createInfuraClient.test.js
mcmire: Per our conversation we decided to handle this in another PR.

Gudahtt (dismissed):
LGTM!

I can address any remaining comments in subsequent PRs, they are all non-blocking.

mcmire:
Feel free to merge this when a second approval comes in.

Gudahtt:
Just rebased this again and updated LavaMoat policies

Gudahtt (dismissed):

Gudahtt:
Hmm. The metrics e2e test appears to be broken. Looking into it now.

Gudahtt:
The metrics e2e test failure seems to be legitimate. It looks like a LavaMoat issue, it's as though `axios` (used by `analytics-node`, the Segment npm package) doesn't have access to `XMLHttpRequest` so the metrics request is failing. Yet the policy shows that it should have access: https://github.com/MetaMask/metamask-extension/blob/566584b922d20a62cf52463cd238b0ff7c1b5668/lavamoat/browserify/main/policy.json#L3877

Gudahtt:
The issue with the metrics test has been resolved. It was a LavaMoat policy issue that was triggered by the `eth-json-rpc-infura` update. I split that update into a separate PR so that the issue was easier to diagnose, and it has been merged already (see #15845)

This is now ready for review again.

Gudahtt (approved):

brad-decker (approved):

PR #15558, opened 2022-08-11
Fix #15050 - MV3: Keep the user logged in when service worker restarts

### LINKED ISSUE #15050: MV3: Ensure user remains logged in after service work restart

darkwing:
## Explanation

Currently, when the service worker restarts, the user gets logged out.  This PR is the extension portion of the login fix.  More details will be added as we get further along.

## More Information

* Fixes #15050
* See: #67890

## Depends on:

* `eth-keyring-controller` (https://github.com/MetaMask/KeyringController/pull/152)

## Manual Testing Steps

- Start the extension in mv3 mode: `yarn start:mv3`
- Login to MetaMask
- Lock MetaMask
- Log in again
- Open the DevTools "Application" panel and kill the service worker (it will restart)
- Execute actions that would require login *without refreshing the page* ("Create Account"), for example
- See that you're still logged in, actions working correctly

## Pre-Merge Checklist


app/scripts/metamask-controller.js
```diff
@@ -2341,7 +2352,12 @@ export default class MetamaskController extends EventEmitter {
    * @returns {Promise<object>} The keyringController update.
    */
   async submitPassword(password) {
-    await this.keyringController.submitPassword(password);
+    const loginToken = await this.keyringController.submitPassword(password);
+
+    // MV3TODO: Store encrypted key for storage
+    if (isManifestV3()) {
+      await chrome.storage.session.set({ loginToken });
```
segun, line 2359: If this works like normal browser storage, then I think we should be using local storage instead. 

Session Storage is only available until the tab is closed. If we want this to persist, we might consider using local storage instead, which persists until it's explicitly deleted.

seaona:
I've noticed the following behaviors:
1. If I have the extension full window open, and I wait +5min idle, MM does not logout
2. If I don't have the extension full window open, and I wait +5min idle, MM does logout
3. If I terminate the Service Worker manually, and have the extension full window open, MM does logout
4. If I terminate the Service Worker manually, and I don't have the extension full window open, MM does logout


**Behavior 2**

[mm-pupup-idle.webm](https://user-images.githubusercontent.com/54408225/191691853-ed77832a-9c38-4736-91ee-7acaef7d0cdc.webm)


**Behavior 3**

[mm-full-manual-restart.webm](https://user-images.githubusercontent.com/54408225/191682603-dac22889-3981-4571-b70f-0389f516c1da.webm)


**Behavior 4**

[mm-popup-manual-restart.webm](https://user-images.githubusercontent.com/54408225/191682975-1ba94a59-76c2-4249-8098-1a3114a0effa.webm)

app/scripts/metamask-controller.js
```diff
@@ -450,6 +451,7 @@ export default class MetamaskController extends EventEmitter {
     this.appStateController = new AppStateController({
       addUnlockListener: this.on.bind(this, 'unlock'),
       isUnlocked: this.isUnlocked.bind(this),
+      isAttemptingLogin: isManifestV3,
```
naugtur, line 454: Why is initial state set like this? 
if isManifestV3 is true, then the process triggered is setting this state to true anyway. 
That is, unless password was used, in which case nothing happens to the state and it stays set to true forever.

app/scripts/metamask-controller.js
```diff
@@ -1093,13 +1095,39 @@ export default class MetamaskController extends EventEmitter {
...
+      // Automatic login via config password
+      const password = process.env.CONF?.PASSWORD;
+      if (
+        password &&
+        this.onboardingController.store.getState().completedOnboarding
+      ) {
+        this.submitPassword(password);
+      }
+      // Automatic login via storage encryption key
+      else if (isManifestV3) {
```
naugtur, line 1108: this says password has to wait for onboarding to be complete, but mv3 login doesn't. I hope that's right.

seaona:
Some QA finding with last commits, I am getting the error `Uncaught (in promise) Error: key is not extractable` on the two Onboarding paths:

- Onboarding Flow with Create Wallet

[key-not-extractable-create.webm](https://user-images.githubusercontent.com/54408225/195050829-2bdb588f-7214-4ff5-a872-9a27f79a86d1.webm)

- Onboarding Flow with an Imported Account

[key-not-extractable.webm](https://user-images.githubusercontent.com/54408225/195049122-f621ad53-00f5-49f2-ae6f-3eea78bcd84f.webm)

app/scripts/metamask-controller.js
darkwing: I set it to `isManifestV3` because we only care about this property, at the moment, if we're using MV3.  If we're using MV3, we know that we're going to perform the async action of checking storage.

app/scripts/metamask-controller.js
darkwing: Good catch, will fix!

seaona:
Started re-testing with the last changes and all previous errors mentioned above disappeared :star:

I saw that the error `vault.map is not a function` is triggered after I log out and try to log in again.

**Repro steps:**
1. Load MM
2. Go to Create Wallet onboarding flow
3. Lock MM
4. Try to unlock MM -- see error on the UI


[vault-map-notfunc.webm](https://user-images.githubusercontent.com/54408225/198007075-eebbaeef-e7a4-414d-87bb-79e2648dd4a5.webm)

darkwing:
Thanks @seaona !  I spotted this issue yesterday, as I was missing an `await` in the `browser-passworder` PR; this should be fixed.  If you're still seeing it, I think it's because NPM is using cached information

ui/components/app/dropdowns/network-dropdown.js
```diff
@@ -130,13 +127,11 @@ class NetworkDropdown extends Component {
...
-            if (this.props.addPopularNetworkFeatureToggledOn) {
-              this.props.history.push(ADD_POPULAR_CUSTOM_NETWORK);
-            } else {
-              getEnvironmentType() === ENVIRONMENT_TYPE_POPUP
-                ? global.platform.openExtensionInBrowser(ADD_NETWORK_ROUTE)
-                : this.props.history.push(ADD_NETWORK_ROUTE);
-            }
+            getEnvironmentType() === ENVIRONMENT_TYPE_POPUP
+              ? global.platform.openExtensionInBrowser(
+                  ADD_POPULAR_CUSTOM_NETWORK,
```
digiwand, line 132: were these later commits meant to be in this PR? 
https://github.com/MetaMask/metamask-extension/pull/15558/commits/b0844e4034181903974b8b17572a49e75d885afc

darkwing: Fixed, thank you!

app/scripts/metamask-controller.js
```diff
@@ -2381,6 +2391,31 @@ export default class MetamaskController extends EventEmitter {
...
+   */
+  async submitEncryptionKey() {
+    this.appStateController.store.updateState({
+      isAttemptingSessionLogin: true,
+    });
+
+    try {
+      const { loginToken } = await browser.storage.session.get(['loginToken']);
+      if (loginToken) {
+        await this.keyringController.submitEncryptionKey(loginTokens);
```
seaona, line 2405: I believe this should be `loginToken`

logintokens-not-defined

seaona:
Overall, the functionality looks great. User is kept logged in after service worker is killed and re-started and I can normally log in and log out.

Exploring some other functionalities, I've encountered the following behaviors, which might/might not be related to this PR:

1. **Auto lock functionality is altered**: when I set an auto-lock value, I see that I am logged out v fast, without respecting my input value. Furthermore, when I disable again the functionality, by introducing 0 mins, I see that I am still being logged out after some seconds. 

https://user-images.githubusercontent.com/54408225/198875730-dd47f98b-9dc9-4b89-a7a5-dcf48027b0a8.mp4

2. **Token detection functionality is altered**: when I terminate the service worker, with the token detection enabled, I see that the user is kept logged in, but it seems that token detection is mixing accounts, when I change accounts a couple of times. I.e. 
    - I see 6 token detected, to an account which only should have 1
    - I see the same token detected, if I change to a testnet
    - If I try to import tokens, I see that tokens are not imported, even though the notification seems successful

https://user-images.githubusercontent.com/54408225/198876420-a0b10577-4f99-4c23-9110-74fdaa0c64f3.mp4


Let me know if I should open separate tickets for solving this issues, or if any of theses issues should be approached by this PR. It's hard for me to tell, as on the develop branch I cannot fully test the flows without this PR being merged on develop. So I would be of the opinion of fixing this separately, except if any of these issues is strictly related to the PR.

seaona:
- **MetamaskController - No HD Key Tree found**: when I terminate the service worker and try to Import an Account with a Private Key, I see the error `MetamaskController - No HD Key Tree found`


https://user-images.githubusercontent.com/54408225/199482784-b3e8ade4-63f3-467e-b779-87f0b149c850.mp4

app/scripts/metamask-controller.js
```diff
@@ -2322,6 +2333,27 @@ export default class MetamaskController extends EventEmitter {
...
+      // If somehow this login token doesn't work properly,
+      // remove it and the user will get shown back to the unlock screen
+      await browser.storage.session.remove(['loginToken']);
+      throw e;
+    } finally {
+      this.appStateController.store.updateState({
+        isAttemptingSessionLogin: false,
+      });
+    }
+  }
```
jpuri, line 2355: May be we can remove state update from finally block also.

jpuri: An option is also to use localstorage for `isAttemptingSessionLogin`.

naugtur (dismissed):

seaona:
With the recent changes, I am being logged out after terminating the service worker. Do I need to do a manual extra step now?

https://user-images.githubusercontent.com/54408225/200333121-31f9c404-b126-40ff-93b8-ad49594bde6c.mp4

PeterYinusa:
i'm getting the following error on the background when building the extension in test mode:
```
Error: LavaMoat - required package not in allowlist: package "@metamask/controllers>eth-keyring-controller" requested "browserify>events" as "events"
```

darkwing:
This PR needs https://github.com/MetaMask/KeyringController/pull/161 to be merged and a new version cut.

app/scripts/metamask-controller.js
```diff
@@ -2347,6 +2351,51 @@ export default class MetamaskController extends EventEmitter {
...
+      else if (isManifestV3) {
+        await this.submitEncryptionKey();
+      }
+      // Updating accounts in this.accountTracker before starting UI syncing ensure that
+      // state has account balance before it is synced with UI
+      await this.accountTracker._updateAccounts();
+    } finally {
+      // Message startUISync is used in MV3 to start syncing state with UI
+      // Sending this message after login is completed helps to ensure that incomplete state without
+      // account details is not flused to UI.
```
digiwand, line 2371:
```suggestion
      // account details are not flushed to UI.
```

app/scripts/ui.js
```diff
@@ -117,11 +117,13 @@ async function start() {
   if (isManifestV3) {
     /*
      * In case of MV3 the issue of blank screen was very frequent, it is caused by UI initialising before background is ready to send state.
-     * Code below ensures that UI is rendered only after CONNECTION_READY message is received thus background is ready.
+     * Code below ensures that UI is rendered only after startUISync message is received thus background is ready.
```
digiwand, line 120: UI seems to be not only rendered after startUISync, it is also rendered on "CONNECTION_READY". What if we update the comment to something like: 
```suggestion
     * Code below ensures that UI is rendered only after "CONNECTION_READY" or "startUISync"
     * messages are received thus the background is ready, and ensures that streams and
     * phishing warning page load only after the "startUISync" message is received.
```

I left out what "startUISync" implies above because it is explained is another part of the code

app/scripts/ui.js
```diff
@@ -117,11 +117,15 @@ async function start() {
...
-     * Code below ensures that UI is rendered only after CONNECTION_READY message is received thus background is ready.
+     * Code below ensures that UI is rendered only after "CONNECTION_READY" or "startUISync"
+     * messages are received thus the background is ready, and ensures that streams and
+     * phishing warning page load only after the "startUISync" message is received.
      * In case the UI is already rendered, only update the streams.
      */
     const messageListener = async (message) => {
-      if (message?.name === 'CONNECTION_READY') {
+      if (message?.name === 'CONNECTION_READY' && !isUIInitialised) {
+        initializeUiWithTab(activeTab);
```
Gudahtt, line 127: This seems wrong. It looks like  `CONNECTION_READY` can be sent before the user is logged in, which would result in state being sent before the keyrings are initialized.

Gudahtt: I see that this was added here: 1710cbec36360fb649d6fb683b9751986295dce0

Maybe this would be alleviated by this PR? https://github.com/MetaMask/metamask-extension/pull/16559

lavamoat/browserify/policy-override.json
```diff
@@ -1,121 +1,3 @@
 {
-  "resources": {
```
Gudahtt, line 2: Whoops - looks like you deleted all of the policy overrides. These aren't auto-generated, these need to be preserved.

yarn.lock
```diff
@@ -2779,6 +2779,11 @@
   resolved "https://registry.yarnpkg.com/@metamask/browser-passworder/-/browser-passworder-3.0.0.tgz#c06744e66a968ffa13f70cc71a7d3b15d86b0ee7"
   integrity sha512-hD10mgvhcDkZX8wnauw8udp1K2MzcbZfrN7Yon9sQ+OqVK9kiQ4VhZAyZNZTy9KJLtfoVD9Y2F6L4gEese7hDA==
 
+"@metamask/browser-passworder@^4.0.1":
```
Gudahtt, line 2782: There is an entry in the policy override for `eth-keyring-controller>browser-passworder`. That will need to be updated to reference the new browser-passworder package name

app/scripts/metamask-controller.js
```diff
@@ -2388,6 +2391,54 @@ export default class MetamaskController extends EventEmitter {
...
+   * Submits a user's encryption key to log the user in via login token
+   */
+  async submitEncryptionKey() {
+    try {
+      const { loginToken, loginSalt } = await browser.storage.session.get([
+        'loginToken',
+        'loginSalt',
+      ]);
+      if (loginToken && loginSalt) {
+        await this.keyringController.submitEncryptionKey(loginToken, loginSalt);
```
Gudahtt, line 2432: We should check before calling this to ensure the salt is still valid (e.g. check that it matches the vault). Otherwise this is guaranteed to throw, and the remainder of the initialization will be broken.

If the salt is stale, we should remove it from browser session storage and continue _without_ throwing any error. Maybe a console message would be appropriate as well, so there is something there to aid in debugging/support troubleshooting.

app/scripts/metamask-controller.js
```diff
@@ -3526,6 +3577,20 @@ export default class MetamaskController extends EventEmitter {
...
+      // send notification to client-side
+      outStream.write({
+        jsonrpc: '2.0',
+        method: 'startUISync',
+      });
+    };
+    this.on('startUISync', startUISync);
+    if (this.startUISync) {
+      this.emit('startUISync');
+    }
```
Gudahtt, line 3593: Nit: Rather than setting up a listener just to call it once, we can call the function directly

```suggestion
    if (this.startUISync) {
      startUISync()
    } else {
       this.on('startUISync', startUISync);
    }
```

app/scripts/metamask-controller.js
darkwing: We added this via https://github.com/MetaMask/KeyringController/blob/main/index.js#L646-L648 ; do we want it in both places, or keep it up in extension?

Gudahtt: By the time that condition is met, it's already too late. That error was to ensure we got a "salt is stale" error message rather than a "cannot decrypt" error message (which could be caused by a number of other, more concerning problems like on-disk vault corruption).

If we check here before attempting login, we can continue initializing the background, leaving the user logged out. And our UI will smoothly redirect them back to the login screen on the first state sync. Not ideal for the user, but overall quite a nice way of handling this rare edge case.

jpuri:
@darkwing : I see one scenario breaking. If I login in MV2 and then start extension in MV3, I get this error while trying to login again using password:

Screenshot 2022-11-23 at 8 46 36 AM

app/scripts/metamask-controller.js
Gudahtt: Thanks! It nearly looks good now,  but the branch for checking for a stale salt is still missing a `return`.

darkwing:
> I see one scenario breaking. If I login in MV2 and then start extension in MV3, I get this error while trying to login again using password

That should be the Ledger hardware wallet error, correct?  That's a separate issue being worked on by the hardware wallets team.

Gudahtt (approved):
LGTM!

kumavis (approved):

PR #15562, opened 2022-08-11
Using formatIconUrlWithProxy to get the iconUrl proxy of swap token

NiranjanaBinoy:
Using `formatIconUrlWithProxy` from controller repo to create the iconUrl proxy of swap token.


jpuri (dismissed):

dan437 (approved):
Nice, I can see that we are using the new proxy URL for tokens now.

jpuri (approved):

PR #15563, opened 2022-08-11
Fix uncaught promise using optional chaining

mrxvt:
## Explanation

Currently, when switching networks in Metamask the following error is thrown, the data object in the error sometimes does not have a request property.

`{"code":-32603,"data":{"code":-32602,"message":"Unknown block number"}}`

In the event that an unresolved promise is caught and undefined is returned on the request property, this pull request uses [optional chaining](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#prod-OptionalExpression) to read the value of the property without having to craft additional validity checks.

### DIFF
--- a/app/scripts/lib/account-tracker.js
+++ b/app/scripts/lib/account-tracker.js
@@ -312,7 +312,7 @@ export default class AccountTracker {
     try {
       balance = await this._query.getBalance(address);
     } catch (error) {
-      if (error.data.request.method !== 'eth_getBalance') {
+      if (error.data.request?.method !== 'eth_getBalance') {
         throw error;
       }
     }
--- a/app/scripts/lib/account-tracker.js
+++ b/app/scripts/lib/account-tracker.js
@@ -312,7 +312,7 @@ export default class AccountTracker {
     try {
       balance = await this._query.getBalance(address);
     } catch (error) {
-      if (error.data.request?.method !== 'eth_getBalance') {
+      if (error?.data?.request?.method !== 'eth_getBalance') {
         throw error;
       }
     }


mabioIa:

mabioIa:

app/scripts/lib/account-tracker.js
```diff
@@ -312,7 +312,7 @@ export default class AccountTracker {
     try {
       balance = await this._query.getBalance(address);
     } catch (error) {
-      if (error.data.request.method !== 'eth_getBalance') {
+      if (error.data.request?.method !== 'eth_getBalance') {
```
mrbaguvix, line 315: If we want to use optional chaning, then lets cover all possible properties that may be undefined 
```suggestion
      if (error?.data?.request?.method !== 'eth_getBalance') {
```

0ex-d (changes requested):

app/scripts/lib/account-tracker.js
mabioIa: Would the optional chain on `error` here be superfluous, as `error` exists only to specify what to do when an exception is thrown from the `try` block on L312?

mrbaguvix: Well error can be anything (a string or object). If it doesn't have a property _data_ it would break the code. Same with the other chained properties. A better approach would be to check if _undefined_ or the likes but since you prefer _optional chaining_ it would be better to follow it all the way.

mabioIa: ACK

darkwing:
Thank you for submitting @mabioIa !  This was recently fixed via https://github.com/MetaMask/metamask-extension/pull/15589 .  Thank you for your time!

PR #15566, opened 2022-08-12
Make add new account idempotent operation

jpuri:
This is requirement with MV3. Calling `addNewAccount` repeatedly with same params should not result in creation of newer accounts.

### DIFF
--- a/app/scripts/metamask-controller.actions.test.js
+++ b/app/scripts/metamask-controller.actions.test.js
@@ -36,7 +36,7 @@ const MetaMaskController = proxyquire('./metamask-controller', {
   './lib/createLoggerMiddleware': { default: createLoggerMiddlewareMock },
 }).default;
 
-describe.only('MetaMaskController', function () {
+describe('MetaMaskController', function () {
   let metamaskController;
   const sandbox = sinon.createSandbox();
   const noop = () => undefined;
--- a/app/scripts/metamask-controller.actions.test.js
+++ b/app/scripts/metamask-controller.actions.test.js
@@ -1,0 +1,93 @@
+import { strict as assert } from 'assert';
+import sinon from 'sinon';
+import proxyquire from 'proxyquire';
+
+const Ganache = require('../../test/e2e/ganache');
+
+const ganacheServer = new Ganache();
+
+const browserPolyfillMock = {
+  runtime: {
+    id: 'fake-extension-id',
+    onInstalled: {
+      addListener: () => undefined,
+    },
+    onMessageExternal: {
+      addListener: () => undefined,
+    },
+    getPlatformInfo: async () => 'mac',
+  },
+};
+
+let loggerMiddlewareMock;
+const createLoggerMiddlewareMock = () => (req, res, next) => {
+  if (loggerMiddlewareMock) {
+    loggerMiddlewareMock.requests.push(req);
+    next((cb) => {
+      loggerMiddlewareMock.responses.push(res);
+      cb();
+    });
+    return;
+  }
+  next();
+};
+
+const MetaMaskController = proxyquire('./metamask-controller', {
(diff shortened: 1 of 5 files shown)


app/scripts/metamask-controller.actions.test.js
```diff
@@ -0,0 +1,93 @@
...
+
+  after(async function () {
+    await ganacheServer.quit();
+  });
+
+  describe('#addNewAccount', function () {
+    it('two successive calls with same accountCount give same result', async function () {
+      await metamaskController.createNewVaultAndKeychain('test@123');
+      const addNewAccountResult1 = await metamaskController.addNewAccount(1);
+      const addNewAccountResult2 = await metamaskController.addNewAccount(1);
```
naugtur, line 82: You could also try to see what happens if they run in parallel
```js
await Promise.all([
 metamaskController.addNewAccount(1),
 metamaskController.addNewAccount(1),
]);
```

Or with one of them delayed to the next eventloop pass
```js
await Promise.all([
 metamaskController.addNewAccount(1),
 Promise.resolve(1).then(metamaskController.addNewAccount),
]);
```

0ex-d (approved):

brad-decker (approved):

app/scripts/metamask-controller.actions.test.js
```diff
@@ -0,0 +1,110 @@
...
+        Object.keys(addNewAccountResult2.identities),
+      );
+    });
+
+    it('two successive calls with same accountCount give same result', async function () {
+      await metamaskController.createNewVaultAndKeychain('test@123');
+      const [addNewAccountResult1, addNewAccountResult2] = await Promise.all([
+        metamaskController.addNewAccount(1),
+        Promise.resolve(1).then(() => metamaskController.addNewAccount(1)),
+      ]);
```
darkwing, line 96: In as far as them being successive, could it be written like:

```
const addNewAccountResult1 =  metamaskController.addNewAccount(1);
const addNewAccountResult2 =  metamaskController.addNewAccount(1);
```

The `Promise.all` didn't look successive to me.  Could just be me though.

jpuri: It is only nicer way to write the function call since function calls are async.

Gudahtt: They can be called sequentially like this:

```
const addNewAccountResult1 =  await metamaskController.addNewAccount(1);
const addNewAccountResult2 =  await metamaskController.addNewAccount(1);
```

The way they are written now, they might still be executing at the same time.

darkwing (approved):

app/scripts/metamask-controller.actions.test.js
jpuri: Let me make code more clean in a following PR.

naugtur: All options provided here (including the original) are doing different things.

Promise.all() allows to await both while they're started in the same tick of the event loop.
Promise.all + Promise.resolve on a second item causes the second item to start in the next tick compared to the first item. It simulates them being triggered not exactly at once, but still overlapping.
Two lines with awaits would make them run sequentially. 

All 2 are valuable test cases.

Gudahtt: >Promise.all + Promise.resolve on a second item causes the second item to start in the next tick compared to the first item. It simulates them being triggered not exactly at once, but still overlapping.

I don't think that's true. `Promise.resolve(1)` won't put that on the next tick. Promises get resolved eagerly, and the current stack of Promises gets resolved during the current tick because they are in a separate queue from other events, the microtask queue. I would not expect the `Promise.resolve` to have any effect on when the function executes. You'd want to use `setImmediate` to get to the next tick.

PR #15569, opened 2022-08-12
Fail client side requests with DisconnectError if connection stream with background closes

jpuri:
Refactoring as discussed here https://github.com/MetaMask/metamask-extension/pull/15337/files#r943796075

### DIFF
--- a/app/scripts/lib/metaRPCClientFactory.test.js
+++ b/app/scripts/lib/metaRPCClientFactory.test.js
@@ -148,4 +148,18 @@ describe('metaRPCClientFactory', () => {
 
     jest.useRealTimers();
   });
+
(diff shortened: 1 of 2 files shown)


app/scripts/lib/metaRPCClientFactory.test.js
```diff
@@ -148,4 +148,18 @@ describe('metaRPCClientFactory', () => {
 
     jest.useRealTimers();
   });
+
+  it('fail request with DisconnectError if underlying stream is closed', async () => {
+    jest.useFakeTimers();
+    const streamTest = createThoughStream();
+    const metaRPCClient = metaRPCClientFactory(streamTest);
+
+    const errorPromise = new Promise(() => metaRPCClient.getState());
```
naugtur, line 157: there's no way for this to work

PR #15570, opened 2022-08-12
Fix rpc and tests

jpuri:
The code improves RPC client so that in case a request can not be completed due to socket disconnection it will throw error, currently it fails silently.


app/scripts/lib/metaRPCClientFactory.test.js
```diff
@@ -9,13 +10,14 @@ describe('metaRPCClientFactory', () => {
     const metaRPCClient = metaRPCClientFactory(streamTest);
     metaRPCClient.foo();
   });
-  it('should be able to make an rpc request/response with the method and params and node-style callback', () => {
+  it('should be able to make an rpc request/response with the method and params and node-style callback', (done) => {
```
naugtur, line 13: Despite opinions documented in jest-eslint plugin, there's still valid usecases for the done callback.
To make these tests compatible with async/await style of asynchronous tests we'd have to use `new Promise` to create some convoluted wrappers with questionable error handling. 

This is the minimal change that makes these tests fail when assertions aren't met. Before this change putting`expect(result).toStrictEqual('NOPE');` below would not fail the test.

darkwing (approved):

app/scripts/lib/metaRPCClientFactory.js
```diff
@@ -12,6 +14,7 @@ class MetaRPCClient {
     this.connectionStream.on('data', this.handleResponse.bind(this));
     this.connectionStream.on('end', this.close.bind(this));
     this.responseHandled = {};
+    this.DisconnectError = DisconnectError;
   }
```
mrbaguvix, line 18: Just out of curiousity, why do we have to extend `Error` class (creating a new Object), could it be different if we just used `Error` class directly since we are not creating new functions on _Line 6_

brad-decker:
@jpuri could you perhaps add some more documentation to the description of this Pull request? The only link is a link to the files themselves.

app/scripts/lib/metaRPCClientFactory.js
naugtur: It's for the use of `instanceof` when checking what error was thrown. 
So that you can differentiate the DisconnectError from an actual error in the operation being performed.

app/scripts/lib/metaRPCClientFactory.js
```diff
@@ -47,6 +50,13 @@ class MetaRPCClient {
   close() {
     this.notificationChannel.removeAllListeners();
     this.uncaughtErrorChannel.removeAllListeners();
+    // fail all unfinished requests
+    for (const [id, handler] of this.requests) {
+      if (!this.responseHandled[id]) {
+        this.responseHandled[id] = true;
+        handler(new DisconnectError('disconnected'));
+      }
+    }
```
jpuri, line 59: If connection closes while requests are in progress they will receive exception.

jpuri:
@brad-decker : I have updated PR description and added one more comment, please check.

digiwand (approved):

PR #15466, opened 2022-08-04
added avatar token component

NidhiKJha:
Fixes #15072

After:

https://user-images.githubusercontent.com/39872794/183075367-5f14099f-05a2-4af2-a08a-ea02fb24e5a8.mov

PR #15575, opened 2022-08-12
AvatarToken style and doc update suggestions

georgewrmarshall:
Some suggestions on using BEM classNames instead of inline styles as well as some doc updates. Does not include updated tests!

### DIFF
--- a/ui/components/component-library/avatar-token/avatar-token.js
+++ b/ui/components/component-library/avatar-token/avatar-token.js
@@ -1,43 +1,62 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
-import classnames from 'classnames';
 import { BaseAvatar } from '../base-avatar';
-import { COLORS, SIZES } from '../../../helpers/constants/design-system';
+import {
+  COLORS,
+  SIZES,
+  DISPLAY,
+  ALIGN_ITEMS,
+  JUSTIFY_CONTENT,
+} from '../../../helpers/constants/design-system';
 
 export const AvatarToken = ({
-  children,
   size = SIZES.MD,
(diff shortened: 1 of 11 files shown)


NidhiKJha:
Not all the changes seem relevant to [ this PR ](https://github.com/MetaMask/metamask-extension/pull/15466). Should we move the required files to a different commit?

ui/components/component-library/avatar-token/avatar-token.js
```diff
@@ -27,52 +29,44 @@ export const AvatarToken = ({
...
+        'avatar-token',
+        showHalo && 'avatar-token--with-halo',
+        className,
       )}
       {...{ backgroundColor, borderColor, ...props }}
     >
       {showFallback ? (
-        tokenName?.[0]?.toUpperCase() ?? '?'
+        tokenName?.[0] ?? '?'
       ) : (
```
NidhiKJha, line 51: @georgewrmarshall shouldn't we keep `tokenName?.[0]?.toUpperCase()` to keep the initial uppercase?

ui/components/component-library/base-avatar/base-avatar.scss
```diff
@@ -28,4 +28,5 @@
   display: flex;
   align-items: center;
   justify-content: center;
+  text-transform: uppercase;
```
georgewrmarshall, line 31: @garrettbear mentioned that we should be changing the text to uppercase via CSS. It's more performant so we can remove changing via javascript @NidhiKJha

ui/components/component-library/avatar-token/avatar-token.js
georgewrmarshall: https://github.com/MetaMask/metamask-extension/pull/15575/files#r945938185

PR #15580, opened 2022-08-15
Integrating snow into metamask

weizman:
This is a first introduction of [Snow ❄️ ](https://github.com/lavamoat/snow) into MetaMask 🦊.

It installs Snow dependency, loads it as vetted shim within every relevant html file of the extension (popup, home, background, notification) and then uses it to log new windows intercepted by it:

Screen Shot 2022-11-22 at 10 59 04

We use Snow only to log new windows and nothing more, because Snow applies advanced hooks to the extension environment, so at first stage we rather examine just that those hooks work properly and nothing more.

In further stages we'll use Snow to monitor and even act against new unknown realms being born within the app.

Read more about this step and its importance [here](https://weizman.github.io/page-snow-into-metamask)!

**NOTICE** - To this PR was merged [#5](https://github.com/weizman/metamask-extension/pull/5) which separately introduces support for changing build process behaviour based on the browser being built - read more about it in the [PR description](https://github.com/weizman/metamask-extension/pull/5#issue-1461513360)


development/build/static.js
```diff
@@ -15,11 +15,21 @@ module.exports = function createStaticAssetTasks({
   livereload,
   browserPlatforms,
   shouldIncludeLockdown = true,
+  shouldIncludeSnow = true,
```
weizman, line 18: all changes made to this entire file are explained [here](https://github.com/MetaMask/metamask-extension/pull/15580#:~:text=NOTICE%20%2D%20To%20this%20PR%20was%20merged%20%235%20which%20separately%20introduces%20support%20for%20changing%20build%20process%20behaviour%20based%20on%20the%20browser%20being%20built%20%2D%20read%20more%20about%20it%20in%20the%20PR%20description)

development/build/index.js
```diff
@@ -230,6 +232,12 @@ testDev: Create an unoptimized, live-reloading build for debugging e2e tests.`,
             'Whether to include SES lockdown files in the extension bundle. Setting this to `false` can be useful during development if you want to handle lockdown errors later.',
           type: 'boolean',
         })
+        .option('snow', {
+          default: true,
+          description:
+            'Whether to include Snow files in the extension bundle. Setting this to `false` can be useful during development if you want to handle Snow errors later.',
+          type: 'boolean',
+        })
```
brad-decker, line 240: Should we make false the default OR update the yarn start command to specify snow=false ... we currently don't have lavamoat enabled on dev mode.

package.json
```diff
@@ -138,6 +138,7 @@
     "@spruceid/siwe-parser": "^1.1.3",
     "@truffle/codec": "^0.11.18",
     "@truffle/decoder": "^5.1.0",
+    "@weizman/snow": "^1.2.1",
```
brad-decker, line 141: Are we going to change this to @lavamoat/snow eventually?

brad-decker (changes requested):
LGTM just a couple questions

development/build/index.js
weizman: is this ok?
20fc66b99c502aa162a5497bcac4f3bd80315005

brad-decker: Yup !

package.json
weizman: f31c69e fix

weizman: d6d47ca54 also

brad-decker (approved):

development/build/static.js
```diff
@@ -15,11 +15,21 @@ module.exports = function createStaticAssetTasks({
...
-    shouldIncludeLockdown,
-  );
+  const copyTargetsProds = {};
+  const copyTargetsDevs = {};
+
+  browserPlatforms.forEach((browser) => {
+    const [copyTargetsProd, copyTargetsDev] = getCopyTargets(
+      shouldIncludeLockdown,
+      // Snow currently only works on Chromium based browsers
+      shouldIncludeSnow && browser === 'chrome',
```
kumavis, line 28: make this field an obj so we can reuse it later `{ snow: shouldIncludeSnow && browser === 'chrome', }`

kumavis (changes requested):
looks good, one request

development/build/static.js
weizman: I was following the exact same pattern as the other `shouldIncludeLockdown` boolean, isn't that the same?

weizman: Do you mean combine both `shouldIncludeLockdown` and `shouldIncludeSnow` into an object?

kumavis: ah, i missed that there was already a condition here - fine to keep as is

kumavis (approved):

PR #15581, opened 2022-08-15
Integrate new LavaMoat scuttling protection feature

weizman:
> NOTICE: This PR moved from here to #16994 

## Motivation

Integrating into MetaMask browser extension the new scuttling feature introduced by LavaMoat ([#360](https://github.com/LavaMoat/LavaMoat/pull/360))

## PR Breakdown

1. lockdown more collides with scuttling feature, adjust it to skip scuttled props
2. `scuttleGlobalThisExceptions` conf prop is populated by 3 different groups: (1) globals that are used by different mm deps that unsafely access global object props and therefore mess with scuttling; (2) globals the chrome driver uses in the test env of mm which cannot be sandboxed and therefore must be excused; and (3) globals vetted shim sentry needs to write to global object in order to function. **All of these will be narrowed down with time and research in different PRs**
3. Also, in order for sentry to work as the vetted shim that it is, we need to allow it to still access scuttled properties, therefore we wrap it with a special sandbox (a mix of `with`s and `proxy`s) that allows it to function normally in a scuttled envirnoment (partially by indicating a list of the properties it'll need that will be later scuttled)
4. update `test/helpers/protect-intrinsics-helpers.js` test to test intrinsics protection only for non scuttled props

In this PR currently there are patch-package files, only to apply the needed upgrade for LM - those will be removed before merge and will be replaced with an actual newly delivered version of LM

## Notes

This is an ongoing effort which will bring with it more PRs in the future. In following PRs we should focus on narrowing down as much as possible the scuttling avoids list which currently contains more fields we skip scuttling for then we'd want to ideally, some of them are potentially quite useful for attackers.

This will require some research.


kumavis:
@SocketSecurity ignore blake-hash@2.0.0
@SocketSecurity ignore tiny-secp256k1@1.1.6
@SocketSecurity ignore protobufjs@6.11.3

kumavis:
some issue with this test https://github.com/MetaMask/metamask-extension/blob/300cb6e7c5effbce3f5fb401d1b035552d75c85c/test/e2e/tests/errors.spec.js#L26-L67

weizman:
4ff9918 works 🎉

weizman:
2cff7e0 works 🎉

brad-decker:
@weizman there may be a sourcemap issue, i'm going to try and regenerate them by rerunning that last CI step.

brad-decker (dismissed):
This looks good to me, @weizman 🎉 

I have a couple notations:
1. For myself: Pull patches into the new yarn3 patching system on the yarn3-activate branch. 
2. For @seaona and/or @PeterYinusa - we need to do a QA regression check to see if these new protective features cause any interactions that are not under test currently to break.

brad-decker:
One question I have, because I will have to carry forward those patches, is getting a sense for why the lavamoat-browserify patch is so large. 72% of the lines of change in this PR are that patch file (72k lines) . can you confirm this intended and why such a large modification of lavamoat-browserify is necessary?

weizman:
> One question I have, because I will have to carry forward those patches, is getting a sense for why the lavamoat-browserify patch is so large. 72% of the lines of change in this PR are that patch file (72k lines) . can you confirm this intended and why such a large modification of lavamoat-browserify is necessary?

Does [this](https://github.com/MetaMask/metamask-extension/pull/15581#:~:text=In%20this%20PR%20currently%20there%20are%20patch%2Dpackage%20files%2C%20only%20to%20apply%20the%20needed%20upgrade%20for%20LM%20%2D%20those%20will%20be%20removed%20before%20merge%20and%20will%20be%20replaced%20with%20an%20actual%20newly%20delivered%20version%20of%20LM) answer your question?

weizman:
> One question I have, because I will have to carry forward those patches, is getting a sense for why the lavamoat-browserify patch is so large. 72% of the lines of change in this PR are that patch file (72k lines) . can you confirm this intended and why such a large modification of lavamoat-browserify is necessary?

Does [this](https://github.com/MetaMask/metamask-extension/pull/15581#:~:text=In%20this%20PR%20currently%20there%20are%20patch%2Dpackage%20files%2C%20only%20to%20apply%20the%20needed%20upgrade%20for%20LM%20%2D%20those%20will%20be%20removed%20before%20merge%20and%20will%20be%20replaced%20with%20an%20actual%20newly%20delivered%20version%20of%20LM) answer your question @brad-decker?

brad-decker:
I missed that last paragraph in the description! My bad.So should we add the "DO NOT MERGE" label to this until that new version of LM is ready? @weizman

weizman:
> I missed that last paragraph in the description! My bad.So should we add the "DO NOT MERGE" label to this until that new version of LM is ready? @weizman

Was not aware of that label, thanks @brad-decker ! 😅 
Will work with @kumavis to ship new LM packages versions and integrate them back here instead of the patch files, and then I'll remove the do-not-merge label 👍

seaona:
Please, let me know once the PR is ready for QA! Thank you!!

weizman:
I am continuing this work in a different PR #16994 after making some mess here by accident

PR #15583, opened 2022-08-15
Adding unit test case to for idempotent behaviour of importAccountWithStrategy idempotent

### LINKED ISSUE #15582: Add unit test case to ensure that importAccountWithStrategy action is idempotent

jpuri:
Fixes: #15582, #15600


brad-decker (dismissed):

brad-decker (dismissed):

app/scripts/metamask-controller.actions.test.js
```diff
@@ -106,6 +109,62 @@ describe('MetaMaskController', function () {
...
+            importPrivkey,
+          ]);
+        }),
+        Promise.resolve(2).then(() => {
+          keyrinControllerState2 = JSON.stringify(
+            metamaskController.keyringController.memStore.getState(),
+          );
+        }),
+      ]);
+      assert.deepEqual(keyrinControllerState1, keyrinControllerState2);
```
digiwand, line 138: tini typo: keyrin -> keyring

app/scripts/metamask-controller.actions.test.js
```diff
@@ -106,6 +109,62 @@ describe('MetaMaskController', function () {
...
+    it('two successive calls with same inputs give same result', async function () {
+      const result1 = await metamaskController.createNewVaultAndRestore(
+        'test@123',
+        TEST_SEED,
+      );
+      const result2 = await metamaskController.createNewVaultAndRestore(
+        'test@123',
+        TEST_SEED,
+      );
+      assert.deepEqual(result1, result2);
```
digiwand, line 152: It could be helpful to add tests to ensure these calls return values before checking if they are equal. This could be in separate `it` tests

app/scripts/metamask-controller.actions.test.js
jpuri: I fixed this.

app/scripts/metamask-controller.actions.test.js
jpuri: The methods if succeed will return value or else will throw exception. I anyways added a check to see that value is not undefined.

digiwand (approved):

brad-decker (approved):

PR #15585, opened 2022-08-15
Make updateTransactionSendFlowHistory background method idempotent

### LINKED ISSUE #15584: Refactor updateTransactionSendFlowHistory action to make it idempotent

jpuri:
Fixes: #15584

Length of history records is compared to ensure that records can not be added twice.


segun (approved):
LGTM

segun (dismissed):
Fix E2E

app/scripts/controllers/transactions/index.js
```diff
@@ -672,27 +672,36 @@ export default class TransactionController extends EventEmitter {
...
+    currentSendFlowHistoryLength,
+    sendFlowHistory,
+  ) {
     this._throwErrorIfNotUnapprovedTx(txId, 'updateTransactionSendFlowHistory');
     const txMeta = this._getTransaction(txId);
 
-    // only update what is defined
-    const note = `Update sendFlowHistory for ${txId}`;
+    if (
+      currentSendFlowHistoryLength === (txMeta?.sendFlowHistory?.length || 0)
```
digiwand, line 689: I wonder if it's possible for the length of `currentSendFlowHistoryLength` to equal the length of `txMeta?.sendFlowHistory?` while having different data sets...

app/scripts/controllers/transactions/index.js
```diff
@@ -672,27 +672,36 @@ export default class TransactionController extends EventEmitter {
...
    * @param {string} txId - transaction id
+   * @param {number} currentSendFlowHistoryLength - sendFlowHistory entries currently
    * @param {Array<{ entry: string, timestamp: number }>} sendFlowHistory -
    *  history to add to the sendFlowHistory property of txMeta.
    * @returns {TransactionMeta} the txMeta of the updated transaction
    */
-  updateTransactionSendFlowHistory(txId, sendFlowHistory) {
+  updateTransactionSendFlowHistory(
+    txId,
+    currentSendFlowHistoryLength,
```
digiwand, line 682: could we preserve the name of the value here? 
Instead of `currentSendFlowHistoryLength` ->   `unapprovedTxSendFlowHistoryLength`? 

The reason I ask is that at first... I was thinking.. we are appending the flow history onto the current one, so why check these lengths? And then I traced through the code to understand it better. I think it would be helpful to update this name.

app/scripts/controllers/transactions/index.js
jpuri: It is always array I think

app/scripts/controllers/transactions/index.js
jpuri: The method is generic to update transactionSendFlowHistory not specifically for unapproved transaction.

app/scripts/controllers/transactions/index.js
brad-decker: Nothing gets removed from the history, only added, so if they are the same it's because it hasn't changed.

brad-decker (approved):

app/scripts/controllers/transactions/index.js
digiwand: I think, actually, I was confused about how this function worked. After reading https://github.com/MetaMask/metamask-extension/pull/15585/files#r963061208, it clicked. Thanks @brad-decker! 

@jpuri your update to change it to `currentSendFlowHistoryLength` sounds good to me now. Thank you and thanks for this explanation too 🙏

digiwand (approved):

PR #15604, opened 2022-08-16
[e2e] Refactor send-hex-address spec using Ganache seeder

seaona:
## Explanation

## More Information
The first task took care of preparing the contracts and the seeder, so it deploys correctly any contract we pass them into Ganache, from the background. First PR is already merged here https://github.com/MetaMask/metamask-extension/pull/14631

- Subtask of [15417](https://github.com/MetaMask/metamask-extension/issues/15417)

## Manual Testing Steps

`yarn test:e2e:single test/e2e/tests/send-hex-address.spec.js`

## Pre-Merge Checklist

### DIFF
--- a/test/e2e/tests/send-hex-address.spec.js
+++ b/test/e2e/tests/send-hex-address.spec.js
@@ -1,5 +1,6 @@
 const { strict: assert } = require('assert');
 const { convertToHexValue, withFixtures } = require('../helpers');
+const { SMART_CONTRACTS } = require('../seeder/smart-contracts');
 
 const hexPrefixedAddress = '0x2f318C334780961FB129D2a6c30D0763d9a5C970';
 const nonHexPrefixedAddress = hexPrefixedAddress.substring(2);
@@ -119,6 +120,7 @@ describe('Send ETH to a 40 character hexadecimal address', function () {
 });
 
 describe('Send ERC20 to a 40 character hexadecimal address', function () {
+  const smartContract = SMART_CONTRACTS.HST;
   const ganacheOptions = {
     accounts: [
       {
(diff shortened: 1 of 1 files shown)


PeterYinusa (dismissed):

test/e2e/tests/send-hex-address.spec.js
```diff
@@ -233,39 +222,28 @@ describe('Send ERC20 to a 40 character hexadecimal address', function () {
...
-        await driver.switchToWindow(extension);
-        await driver.clickElement('[data-testid="home__activity-tab"]');
-        await driver.waitForSelector(
-          '.transaction-list__completed-transactions .transaction-list-item:nth-of-type(1)',
-          { timeout: 10000 },
-        );
 
         // Add token
-        await driver.switchToWindow(dapp);
+        await driver.findClickableElement('#deployButton');
```
adonesky1, line 246: Do we still need to look for the `deployButton`? Should we remove it since we're now pre-seeding the contract?

seaona: @adonesky1 that's a great observation, so thank you for pointing this out.

The reason for adding this previous step of `findClickableElement` is because I noticed that until DeployButton element is not clickable, the Dapp is not actually ready and clicking on WatchAsset directly without waiting for this makes test fail.

First, I was considering to add a delay, but thought this approach is more deterministic. This same step was also mentioned by @PeterYinusa on another PR and we finally decided to use the `findClickableElement` instead of the delay, but if you have strong opinions on the other direction, let me know.

See the other PR with the comment [here](https://github.com/MetaMask/metamask-extension/pull/15553/files#r945764773)

deploy-contract

adonesky1: ah ok, makes sense! Do you mind adding a comment to explain this?

adonesky1: ah ok, makes sense! Do you mind adding a comment to explain this?

brad-decker (dismissed):

test/e2e/tests/send-hex-address.spec.js
seaona: @adonesky1 added a comment, thank you!!!

adonesky1 (approved):
LGTM!

test/e2e/tests/send-hex-address.spec.js
adonesky1: Thanks @seaona !

darkwing (approved):

PR #15607, opened 2022-08-16
Token Detection clean up - Fixing swaps performance slow down

NiranjanaBinoy:
During Token detection an object was replaced with an array while we try to use the same data for shuffledtokenList and  getRenderableTokenData slowing down Swaps, in this PR we are separating the data to speedup swaps again.

### DIFF
--- a/ui/hooks/useTokensToSearch.js
+++ b/ui/hooks/useTokensToSearch.js
@@ -8,6 +8,7 @@ import {
   getCurrentCurrency,
   getSwapsDefaultToken,
   getCurrentChainId,
+  getTokenList,
 } from '../selectors';
 import { getConversionRate } from '../ducks/metamask/metamask';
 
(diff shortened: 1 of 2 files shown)


ui/hooks/useTokensToSearch.js
```diff
@@ -96,7 +95,7 @@ export function useTokensToSearch({
     conversionRate,
     currentCurrency,
     chainId,
-    shuffledTokensList,
+    tokenList,
```
adonesky1, line 98: Should `shuffledTokensList` be removed as an arg to `useTokensToSearch` and replaced with `tokenList` on line 108?

dan437 (approved):
Thank you for the changes! I've tested this PR and it's looking good!

Before (951 ms):
image

After (163 ms):
image

ui/hooks/useTokensToSearch.js
adonesky1: 🤔 weird. Alright, won't block on this.

adonesky1 (approved):

PR #15610, opened 2022-08-16
migrate network constants to typescript

brad-decker:
## Explanation
The networks constant file is on the next run of files that are poised for easy conversion to typescript. In this case I also did some general cleanup to make the types clearer and provide some better type safety to the constants we are using in this file. I converted a number of constants to objects that group them together, and provided documentation. This PR has no function changes just renames a bunch of variables for legibility and ergonomics.


shared/constants/network.ts
```diff
@@ -0,0 +1,579 @@
...
+  defaultCurrencyCode:
+    | LowercaseCurrencySymbol
+    | `${LowercaseCurrencySymbol}_${MoonPayNetworkAbbreviations}`;
+  showOnlyCurrencies:
+    | `${LowercaseCurrencySymbol}`
+    | `${LowercaseCurrencySymbol},${LowercaseCurrencySymbol}`
+    | `${LowercaseCurrencySymbol},${LowercaseCurrencySymbol},${LowercaseCurrencySymbol}`
+    | `${LowercaseCurrencySymbol},${LowercaseCurrencySymbol},${LowercaseCurrencySymbol},${LowercaseCurrencySymbol}`
+    | `${LowercaseCurrencySymbol}_${MoonPayNetworkAbbreviations}`
+    | `${LowercaseCurrencySymbol}_${MoonPayNetworkAbbreviations},${LowercaseCurrencySymbol}_${MoonPayNetworkAbbreviations}`;
```
brad-decker, line 380: This type matches either 1,2 or 3 currency symbols in lower case format separated by ','. It also will match either 1 or 2 currency symbols in lowercase format, concatenated with a network abbreviation and '_' separated by ','.

shared/constants/network.ts
```diff
@@ -0,0 +1,579 @@
...
+      typeof NETWORK_TYPES.RINKEBY
+    >
+  }${CURRENCY_SYMBOLS.ETH}`,
+  [NETWORK_TYPES.KOVAN]: `${
+    capitalize(NETWORK_TYPES.KOVAN) as Capitalize<typeof NETWORK_TYPES.KOVAN>
+  }${CURRENCY_SYMBOLS.ETH}`,
+  [NETWORK_TYPES.GOERLI]: `${
+    capitalize(NETWORK_TYPES.GOERLI) as Capitalize<typeof NETWORK_TYPES.GOERLI>
+  }${CURRENCY_SYMBOLS.ETH}`,
+} as const;
```
mcmire, line 208: Could you also write this as the following?

``` suggestion
export const TEST_NETWORK_TICKER_MAP: {
  [K in NetworkTypes]: Capitalize<K>
}: Object.keys(NETWORK_TYPES).reduce((object, key) => {
  return { ...object, [key]: capitalize(key) };
}, {});
```

shared/constants/network.ts
```diff
@@ -0,0 +1,579 @@
...
+type BuyableChainSettings = {
+  nativeCurrency: CurrencySymbol | TestNetworkCurrencySymbols;
+  network: string;
+  transakCurrencies?: CurrencySymbol[];
+  moonPay?: MoonPayChainSettings;
+  wyre?: WyreChainSettings;
+  coinbasePayCurrencies?: CurrencySymbol[];
+};
+
+export const BUYABLE_CHAINS_MAP: { [K in ChainIds]?: BuyableChainSettings } = {
```
mcmire, line 397: Why the question mark here?

shared/constants/network.ts
```diff
@@ -0,0 +1,579 @@
...
+ */
+export const NETWORK_NAMES = {
+  HOMESTEAD: 'homestead',
+};
+
+/**
+ * A type representing any valid value for 'type' for setProviderType and other
+ * methods that add or manipulate networks in MetaMask state.
+ */
+export type NetworkTypes = typeof NETWORK_TYPES[keyof typeof NETWORK_TYPES];
```
mcmire, line 33: A couple of guidelines that I've been following is to use singular names for union types and to organize files so that types are the top so they are easier to find. What do you think about these?

shared/constants/network.ts
Nooknick7899: > +export const TEST_NETWORK_TICKER_MAP = {

+  [NETWORK_TYPES.ROPSTEN]: `${

+    capitalize(NETWORK_TYPES.ROPSTEN) as Capitalize<

+      typeof NETWORK_TYPES.ROPSTEN

+    >

+  }${CURRENCY_SYMBOLS.ETH}`,

+  [NETWORK_TYPES.RINKEBY]: `${

+    capitalize(NETWORK_TYPES.RINKEBY) as Capitalize<

+      typeof NETWORK_TYPES.RINKEBY

+    >

+  }${CURRENCY_SYMBOLS.ETH}`,

+  [NETWORK_TYPES.KOVAN]: `${

+    capitalize(NETWORK_TYPES.KOVAN) as Capitalize<typeof NETWORK_TYPES.KOVAN>

+  }${CURRENCY_SYMBOLS.ETH}`,

+  [NETWORK_TYPES.GOERLI]: `${

+    capitalize(NETWORK_TYPES.GOERLI) as Capitalize<typeof NETWORK_TYPES.GOERLI>

+  }${CURRENCY_SYMBOLS.ETH}`,

+} as const;

shared/constants/network.ts
brad-decker: Does it matter if the type uses the value to create the type definition? I feel thats use before declaration but maybe that doesn't matter for types. I feel in a lot of cases separating the type and the value by a large amount of space leads to confusion but I could be swayed.

mcmire: The compiler won't have a problem with it, but if it's more readable to put the `const` first in this case, then I can understand why you'd want to keep it together.

shared/constants/network.ts
brad-decker: Because ChainIds includes some chain ids that aren't going to be in the final object.

mcmire: Is it that they will never be in the final object, or they will sometimes be in the final object? If they will never be in the final object maybe it's worth `Exclude`ing some things?

brad-decker: Good point. Yeah I can either use Pick or Omit to pull the ones that will be in the object.

shared/constants/network.ts
mcmire: Okay, aside from that, maybe use a singular for this? i.e. `NetworkType` instead of `NetworkTypes`. Otherwise it could be interpreted as a tuple of network types, e.g.

``` typescript
export type NetworkTypes = ['mainnet', 'ropsten', 'goerli', ...]
``` 

as opposed to a union:

``` typescript
export type NetworkType = 'mainnet' | 'ropsten' | 'goerli' | ...
```

shared/constants/network.ts
mcmire: Nice! I love it.

shared/constants/network.ts
brad-decker: Upon further investigation, the proposed would not result in what we want.

We would need to create a new variable that contained only the test nets we want to reduce on, otherwise we'd need to do logic checks in the reduce method to extract the non-test-nets, we'd also need to make the Capitalize<K> part be `${Capitalize<K>}ETH` which Is fine but overall the complexity doesn't save much over just being explicit. For brevity I changed to this instead:

```
const typedCapitalize = <K extends string>(k: K): Capitalize<K> =>
  capitalize(k) as Capitalize<typeof k>;

export const TEST_NETWORK_TICKER_MAP: {
  [K in Exclude<
    NetworkTypes,
    'localhost' | 'mainnet' | 'rpc'
  >]: `${Capitalize<K>}${typeof CURRENCY_SYMBOLS.ETH}`;
} = {
  [NETWORK_TYPES.ROPSTEN]: `${typedCapitalize(NETWORK_TYPES.ROPSTEN)}${
    CURRENCY_SYMBOLS.ETH
  }`,
  [NETWORK_TYPES.RINKEBY]: `${typedCapitalize(NETWORK_TYPES.RINKEBY)}${
    CURRENCY_SYMBOLS.ETH
  }`,
  [NETWORK_TYPES.KOVAN]: `${typedCapitalize(NETWORK_TYPES.KOVAN)}${
    CURRENCY_SYMBOLS.ETH
  }`,
  [NETWORK_TYPES.GOERLI]: `${typedCapitalize(NETWORK_TYPES.GOERLI)}${
    CURRENCY_SYMBOLS.ETH
  }`,
} as const;
```

shared/constants/network.ts
brad-decker: I moved the types to the top and made the plural -> singular change. I do like that the first part of the file looks like documentation and the rest looks like code. Lets standardize this going forward until we hit a case where it doesn't work.

jpuri (dismissed):
Changes look good

shared/constants/network.ts
```diff
@@ -0,0 +1,675 @@
...
+  [NETWORK_TYPES.RINKEBY]: `${typedCapitalize(NETWORK_TYPES.RINKEBY)}${
+    CURRENCY_SYMBOLS.ETH
+  }`,
+  [NETWORK_TYPES.KOVAN]: `${typedCapitalize(NETWORK_TYPES.KOVAN)}${
+    CURRENCY_SYMBOLS.ETH
+  }`,
+  [NETWORK_TYPES.GOERLI]: `${typedCapitalize(NETWORK_TYPES.GOERLI)}${
+    CURRENCY_SYMBOLS.ETH
+  }`,
+} as const;
```
jpuri, line 339: How does `as const` here helps ?

brad-decker: Good catch, it helped before the stronger typing to keep the types to being broadened to 'string' but with the map type now specified it is no longer needed

mcmire:
Sorry I won't have time to do a deep dive on this PR this week as I'm out, but at a glance this looks good to me.

mcmire:
Sorry I won't have time to do a deep dive on this PR this week as I'm out, but at a glance this looks good to me.

jpuri (approved):

digiwand (approved):
LGTM! 

Thank you for including the handy comments in shared/constants/network.ts 🚀

PR #15611, opened 2022-08-16
Migrate app constants to typescript

brad-decker:
## Explanation

Simple conversion of app constants to typescript.


digiwand (approved):
Wahooo TypeScript! Happy to see this kicking off

mcmire (approved):
Looks good! I ended up doing the same thing when I was testing this out.

PR #15565, opened 2022-08-11
Adding initial and transparent to Box component

georgewrmarshall:
## Explanation

Currently, `INITIAL` and `TRANSPARENT` color options for the box component don't exist or don't work. We also still allow for SECONDARY colors which are now deprecated

This is a problem because it reduces the flexibility of the Box component and is required for the new icon component. 

In order to solve this problem, this pull request adds and fixes the `INITIAL` and `TRANSPARENT` color options for the `Box` component. It also removes the deprecated SECONDARY colors.

## More Information

Fixes #15559 

See deprecated colors list in the [design tokens repo](https://metamask.github.io/design-tokens/?path=/story/colors-themecolors--light-theme-colors)

Screen Shot 2022-08-11 at 3 55 42 PM

### Before

Screen Shot 2022-08-11 at 3 58 32 PM
Screen Shot 2022-08-11 at 3 59 25 PM
Screen Shot 2022-08-11 at 3 59 51 PM

### After

Screen Shot 2022-08-11 at 4 01 59 PM
Screen Shot 2022-08-11 at 4 02 52 PM

## Manual Testing Steps

- Go to the storybook build on this PR
- Search `Box` in the search bar
- Use the controls to check that `transparent` is working for **backgroundColor**
- Use the controls to check that `inherit` is working for **color**
- See that all the links to design system OBJECT go to the correct line positions
- See that any instances of SECONDARY has been removed

## Pre-Merge Checklist

- [x] PR template is filled out
- ~[x] **IF** this PR fixes a bug, a test that _would have_ caught the bug has been added~ N/A
- [x] PR is linked to the appropriate GitHub issue
- ~[ ] PR has been added to the appropriate release Milestone~ N/A

### + If there are functional changes:

- ~[ ] Manual testing complete & passed~ N/A
- ~[ ] "Extension QA Board" label has been applied~ N/A

PR #15613, opened 2022-08-16
Updating Box to remove color prop default (Fixes develop)

georgewrmarshall:
## Explanation

Currently `develop` is failing because of some snapshot tests. [I added the `color` prop to the `Box` component and set a default](https://github.com/MetaMask/metamask-extension/pull/15565/files#r944009513) which is creating the `box--color-text-default` prop. And although would make sense may have some unexpected outcomes if some of the components that use the Box component are setting the font color using the parent element. This was an oversight on my part.

This PR removes the default prop which should prevent the color class being added and fix tests. It will also prevent any unexpected colors being set by parent CSS

## Pre-Merge Checklist

- ~[x] **IF** this PR fixes a bug, a test that _would have_ caught the bug has been added~ some snapshot tests already exists
- ~[x] PR is linked to the appropriate GitHub issue~ N/A
- ~[x] PR has been added to the appropriate release Milestone~ N/A

- ~[x] Manual testing complete & passed~ N/A
- ~[x] "Extension QA Board" label has been applied~ N/A

### DIFF
--- a/ui/components/ui/box/box.js
+++ b/ui/components/ui/box/box.js
@@ -185,7 +185,7 @@ export default function Box({
   children,
   className,
   backgroundColor,
-  color = TEXT_COLORS.TEXT_DEFAULT,
+  color,
   as = 'div',
   ...props
 }) {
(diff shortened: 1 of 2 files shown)


Gudahtt (approved):
LGTM!

darkwing:
This fixes the test but is the fix what we actually want?  Do we want a default?  Should we update the tests?

ryanml (approved):

Gudahtt:
>This fixes the test but is the fix what we actually want? Do we want a default?

I think George addresses this in the PR description:

>may have some unexpected outcomes if some of the components that use the Box component are setting the font color using the parent element.

Basically, setting a default color at the `Box` level is probably a bad idea because it would overwrite any color set on parent elements. We don't want to be forced to set colors at every layer.

PR #15626, opened 2022-08-17
Convert LavaMoat policy generation script to Yargs application

rekmarks:
This PR converts `generate-lavamoat-policies.sh` to `.js` using Yargs. This makes it easier to only generate policy files for a specific build type (using the `-t` flag), which is often useful during Flask development. In addition, the `lavamoat:background:auto` scripts are renamed, and the main readme is updated with some useful tips.

Note that `lavamoat:background:auto:dev` is removed and `lavamoat:background:auto` should be used during local development.


development/generate-lavamoat-policies.js
```diff
@@ -0,0 +1,53 @@
...
+          choices: Object.values(BuildType),
+          default: Object.values(BuildType),
+          demandOption: true,
+          description: 'The build type(s) to generate policy files for.',
+        })
+        .option('parallel', {
+          alias: ['p'],
+          default: true,
+          description: 'Whether to generate policies in parallel.',
+          demandOption: true,
```
rekmarks, line 30:
```suggestion
          demandOption: true,
          description: 'Whether to generate policies in parallel.',
```

Mrtenz (approved):

development/generate-lavamoat-policies.js
```diff
@@ -0,0 +1,53 @@
+const concurrently = require('concurrently');
```
legobeat, line 1:
```suggestion
#!/usr/bin/env node

const concurrently = require('concurrently');
```

development/generate-lavamoat-policies.js
```diff
@@ -0,0 +1,53 @@
...
+const yargs = require('yargs/yargs');
+const { hideBin } = require('yargs/helpers');
+const { BuildType } = require('./lib/build-type');
+
+start().catch((error) => {
+  console.error('Policy generation failed.', error);
+  process.exitCode = 1;
```
legobeat, line 8: We can exit immediately? 
```suggestion
  process.exit(1);
```

package.json
```diff
@@ -73,9 +73,10 @@
     "lavamoat:build": "lavamoat development/build/index.js --policy lavamoat/build-system/policy.json --policyOverride lavamoat/build-system/policy-override.json",
     "lavamoat:build:auto": "yarn lavamoat:build --writeAutoPolicy",
     "lavamoat:debug:build": "yarn lavamoat:build --writeAutoPolicyDebug --policydebug lavamoat/build-system/policy-debug.json",
-    "lavamoat:background:auto": "./development/generate-lavamoat-policies.sh",
-    "lavamoat:background:auto:dev": "./development/generate-lavamoat-policies.sh --dev",
```
legobeat, line 77: Maybe removing this script can be broken out in a separate change that follows the rest of this?

kumavis (approved):

development/generate-lavamoat-policies.js
rekmarks: We can, but we probably shouldn't unless necessary, see: https://nodejs.org/api/process.html#process_process_exit_code

development/generate-lavamoat-policies.js
rekmarks: We could do this, but we generally don't for development scripts. We just pass them to `node`.

package.json
rekmarks: I don't think that's worth the effort, although I did update the PR description to call out this removal.

hmalik88 (approved):

PR #15628, opened 2022-08-17
Fix setting of gasPrice when on non-eip 1559 networks

danjm:
Fixes a bug whereby users would intermittently see failures when trying to send transactions on non-EIP-1559 networks.

We were preventing the setting of gas price when `state.gasPriceEstimate !== '0x0' && draftTransaction.gas.gasPrice !== state.gasPriceEstimate`. This was done to avoid overwriting manually edited gas prices. However, because `state.gasPriceEstimate` gets set on every update of gas fees from the background, and with recent changes `draftTransaction.gas.gasPrice` only gets set after a draftTransaction has been created in the send flow, it was easy to get in a state where the former is non-zero and the latter is 0, thereby preventing the latter from being set.

The solution was to explicitly set data indicating whether gas price had manually been edited, and rely on that for whether to overwrite gas price.


brad-decker (dismissed):
LGTM.

brad-decker (dismissed):

ui/ducks/send/send.js
```diff
@@ -1821,6 +1813,7 @@ export function updateGasPrice(gasPrice) {
       actions.updateGasFees({
         gasPrice,
         transactionType: TRANSACTION_ENVELOPE_TYPES.LEGACY,
+        manuallyEdited: true,
```
Gudahtt, line 1816: I mist be misunderstanding something. This makes me think that this is the only method whereby gas can be manually edited, is that correct? That if this is getting called, it's because a manual edit has occurred?

If that's the case, then why are we discarding this `gasPrice`? It looks like it gets ignored in `updateGasFees` if `manuallyEdited` is `true`

danjm: Yes, you are correct. The if block I added should be moved down.

danjm: and also...

danjm: Okay, I have fixed this with 609506ddda

> this is the only method whereby gas can be manually edited, is that correct? That if this is getting called, it's because a manual edit has occurred?

Yes, that is correct.

> If that's the case, then why are we discarding this gasPrice? It looks like it gets ignored in updateGasFees if manuallyEdited is true

Correct, that was a bad mistake. I have fixed this with 609506ddda. Now we don't ignore gasPrices in the action payload if that action is a manual update.

Gudahtt: Got it, thanks, this makes sense now

Gudahtt (dismissed):
LGTM!

ui/ducks/send/send.test.js
```diff
@@ -1410,6 +1410,7 @@ describe('Send Slice', () => {
             type: 'send/updateGasFees',
             payload: {
               gasPrice: '0x0',
+              wasManuallyEdited: true,
```
Gudahtt, line 1413: Oops, changed one too many:
```suggestion
              manuallyEdited: true,
```

Gudahtt (approved):

seaona:
Looks like failure is substantially mitigated. Just found one edge case, where I'm still seeing a tx failure, everytime I use the following values:

- Amount:  `0.001ETH` 
- Hex Data: `0x0`

For my side reproduction is consistent. 

https://user-images.githubusercontent.com/54408225/185349495-0a6081d2-c8fd-4ac1-8ea1-cea632e7b738.mp4

danjm:
@seaona Yes, the 0x0 edge case for hex data (which Peter filed a ticket for a few weeks ago) is not fixed by this, just the other cases

brad-decker (approved):

PR #15638, opened 2022-08-18
converted Jazzicon component to functional component and added story

NidhiKJha:
## Explanation

Currently, the `Jazzicon` component is a class-based component and doesn't have a story

Since we are gonna utilise this component in other IA/Nav components, this PR is to add a story to `Jazzicon` while converting it to functional component

### After

Screenshot 2022-08-18 at 8 45 51 PM

## Manual Testing Steps

- Go to the latest storybook build of this PR
- Search `Jazzicon`

## Pre-Merge Checklist

- ~[x] **IF** this PR fixes a bug, a test that _would have_ caught the bug has been added~ N/A
- ~[x] PR is linked to the appropriate GitHub issue~ N/A
- ~[x] PR has been added to the appropriate release Milestone~ N/A

- ~[ ] "Extension QA Board" label has been applied~ N/A


ui/components/ui/jazzicon/README.mdx
```diff
@@ -0,0 +1,15 @@
+import { Story, Canvas, ArgsTable } from '@storybook/addon-docs';
+
+import Jazzicon from './jazzicon.component';
+
+# Jazzicon
+
+Jazzicon component is a wrapper around the jazzicon library to return a React component.
```
georgewrmarshall, line 7: I think we can be a little more descriptive about what a Jazzicon is here. Always helps to add a link when we use a reference too.

```suggestion
Jazzicon uses the [jazzicon library](https://github.com/MetaMask/jazzicon) to create a graphic identity of an ethereum address
```

ui/components/ui/jazzicon/jazzicon.component.js
```diff
@@ -9,56 +9,39 @@ const iconFactory = iconFactoryGenerator(jazzicon);
...
 }
+
+Jazzicon.propTypes = {
+  address: PropTypes.string.isRequired,
+  className: PropTypes.string,
+  diameter: PropTypes.number,
+  style: PropTypes.object,
+  tokenList: PropTypes.object,
+};
+
```
georgewrmarshall, line 46: Add proptype descriptions here

ui/components/ui/jazzicon/jazzicon.component.js
```diff
@@ -9,56 +9,39 @@ const iconFactory = iconFactoryGenerator(jazzicon);
...
+    return imageNode.outerHTML;
+  }, [address, diameter, tokenList]);
 
-    return <div className={className} ref={this.container} style={style} />;
-  }
+  return (
+    <div
+      className={className}
+      style={style}
+      dangerouslySetInnerHTML={{ __html: image }}
```
georgewrmarshall, line 34: @brad-decker what's your thoughts on using `dangerouslySetInnerHTML` here?

georgewrmarshall (changes requested):
Looking good! Some updates

ui/components/ui/jazzicon/jazzicon.component.js
NidhiKJha: done

georgewrmarshall (commented):
Looks good! Would just get a second opinion on the use of `dangerouslySetInnerHTML`

NidhiKJha:
> Looks good! Would just get a second opinion on the use of `dangerouslySetInnerHTML`

cc @darkwing @brad-decker

ui/components/ui/jazzicon/jazzicon.component.js
brad-decker: This isn't any more or less  safe than 'appendChild' with an 'image' that could literally be any html. If iconFactory.iconForAddress were to return malicious code we'd inject that malicious code. We *do* own that code so its less of a concern. 

I feel it would be better to append the element in the same way it was before just to avoid any additional unforeseen security implications of _dangerouslySetInnerHTML -- which doesn't yet exist in the codebase.

brad-decker (changes requested):

ui/components/ui/jazzicon/jazzicon.component.js
```diff
@@ -9,56 +9,60 @@ const iconFactory = iconFactoryGenerator(jazzicon);
...
+    const removeIcon = () => {
+      container.current && (container.current.innerHTML = '');
+    };
 
-    if (address !== prevAddress || diameter !== prevDiameter) {
-      this.removeExistingChildren();
-      this.appendJazzicon();
-    }
-  }
+    removeIcon();
```
brad-decker, line 37: If you return this function, instead of calling it, it'll be called only when the useEffect is re-run, before the new effect is ran. You may then remove the addIcon method and just run that code as the effect itself.

brad-decker (changes requested):

brad-decker (approved):

georgewrmarshall (commented):
Looks like prop table is broken. If you can't fix you can just add manually with a mark down table :) 

Screen Shot 2022-09-06 at 9 44 30 AM

darkwing (approved):

georgewrmarshall (approved):
Ahh it's the optional chaining issue again. Will approve this and hopefully issue is fixed in updated version of storybook

PR #15642, opened 2022-08-18
upgrade jest

brad-decker:
## Explanation
Fixes issue related to slack reports of unit tests consuming upwards of more than 20gb of memory and 4gb of swap. Note that this PR does nothing to fix the underlying memory leaks in our tests and that should be explored separately. This works around the memory issue by forcing worker processes to terminate once they've consumed 500mb of memory. I chose 500mb because our ci environment has 4gb of ram and we run two workers when running tests in jest. This would allow us to theoretically add another 6 workers without running into out of memory issues. I'll test that theory in another PR aimed at improving CI. 

Consequences of this PR:
1. The default environment for tests is node, so had to install the jsdom environment and specify jsdom as our testEnvironment in jest.config.js. This can be changed *per file*. 
2. Had to temporarily disable the button.stories.test.js file because it eventually imports a .mdx file and the transform that was previously working for 'mdx' files in jest no longer works. 
3. Had to patch the types for jsdom to not use the `import { type TypeName}` syntax
4. Removed the yargs patch because the newest version checks for the existence of Error.createErrorBoundary before calling it.
5. Some minor syntax changes

## Pre-Merge Checklist


tmashuang (dismissed):
LGTM, will re-approve once the merge conflicts are resolved.

tmashuang (dismissed):

ui/components/ui/button/button.stories.skippedtest.js
```diff
@@ -1,3 +1,7 @@
+// Importing MDX in tests is *broken* for jest versions greater than 27.
+// Until https://github.com/storybookjs/storybook/issues/15916 is resolved
+// we can't run this test file.
+/* eslint-disable */
 /* eslint-disable jest/require-top-level-describe */
 import React from 'react';
```
georgewrmarshall, line 7: I think we can probably remove this file altogether. It was added in an attempt to check if a story broke because of a component update. I think storybook snapshots would be a more scalable solution with less overhead.

georgewrmarshall (commented):
As per our slack conversation I think we can probably remove the `ui/components/ui/button/button.stories.skippedtest.js` file altogether

georgewrmarshall (approved):
LGTM!

jest.config.js
```diff
@@ -44,8 +44,19 @@ module.exports = {
...
+  // background tests.
+  testEnvironment: 'jsdom',
+  // Our configuration somehow is calling into the esm folder / files of
+  // some modules. Jest supports ESM but our code is not set to emit ESM files
+  // so we are telling jest to use babel to transform the node_modules listed.
+  // Note: for some reason I could not hammer down to the node_modules
+  // installed in @metamask/controllers so I had to just blanket specify all
+  // of the @metamask/controllers folder.
+  transformIgnorePatterns: [
+    '/node_modules/(?!(multiformats|uuid|nanoid|@metamask/controllers|@metamask/snap-controllers)/)',
```
mcmire, line 59: `multiformats`, `uuid`, and `nanoid` makes sense, because they have recently(ish) converted to ESM, and I've seen Jest choke on these packages in other projects.

We don't ship ESM code in `@metamask/controllers` or `@metamask/snap-controllers`, however, so why did we have to list these here?

patches/@types+jsdom++parse5+7.0.0.patch
```diff
@@ -0,0 +1,91 @@
+diff --git a/node_modules/@types/jsdom/node_modules/parse5/dist/index.d.ts b/node_modules/@types/jsdom/node_modules/parse5/dist/index.d.ts
+index 81253d3..d2333bf 100644
+--- a/node_modules/@types/jsdom/node_modules/parse5/dist/index.d.ts
++++ b/node_modules/@types/jsdom/node_modules/parse5/dist/index.d.ts
+@@ -1,10 +1,10 @@
+-import { type ParserOptions } from './parser/index.js';
++import { ParserOptions } from './parser/index.js';
```
mcmire, line 7: Sorry why do we have to remove `type` here?

jest.config.js
brad-decker: some of the dependencies nested *inside* the metamask/controllers repo are ESM and i could not figure out how to specify that recursion in this ignore pattern.

mcmire: Ah I see. Interesting...

patches/@types+jsdom++parse5+7.0.0.patch
mcmire: Ohh... wow. That's subtle. How do other people even use this package then 🤔

mcmire:
Had to do a bit of research on this one. Looks like the next version of Jest includes `workerIdleMemoryLimit`, and that version isn't out yet, so this PR bumps it to 29.alpha. The [docs for this option](https://jestjs.io/docs/next/configuration#workeridlememorylimit-numberstring) link to [this issue](https://github.com/facebook/jest/issues/11956) which explains where it came from. It also might be worth following [this PR](https://github.com/facebook/jest/pull/12205) to see if that fixes things further. Or [this issue in Node](https://github.com/nodejs/node/issues/44211).

mcmire (approved):
Makes sense to me.

patches/@types+jsdom++parse5+7.0.0.patch
brad-decker: no idea. unless it has something to do with our setup.

brad-decker:
@mcmire my apologies for not posting more details in the description. You ended up doing some double work that I could have possibly saved some time for you if I would have simply transcribed @Gudahtt and I's thread in the extension channel into the PR description. I'll do better in the future.

PR #15644, opened 2022-08-18
Fixing token detection grammar

### LINKED ISSUE #15588: [Bug]: Token Detection displays word Token(s) always in plural, even when there is just 1 token to add
### Describe the bug

**Problem**: we see that the text for importing tokens is always displayed in plural `tokens`, even when there is only 1 token to add

image

image

### Steps to reproduce

1. Unlock MM
2. Load an account which has 1 token (different from ETH)
3. Make sure token detection is enabled
4. Click refresh

_No response_

### Version

10.20.0

### Build type

_No response_

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

ryanml:
Fixes: https://github.com/MetaMask/metamask-extension/issues/15588
Test plan: Defined in issue


NidhiKJha (commented):
Locale `numberOfNewTokensDetected` is added in `app/_locales/zh/messages.json` as well.  Can we remove it from there?

georgewrmarshall (approved):
LGTM with my understanding that because we are creating new keys it's deleting the old translations even though the translations would likely remain the same for one of the new keys

NiranjanaBinoy (approved):
LGTM. Maybe we can remove the locale `numberOfNewTokensDetected` in `app/_locales/zh/messages.json` manually for this PR and add a new PR to update the index file you have mentioned in the slack message.

ryanml:
As `zh` is an unused translation file we'll go ahead and follow up on the crowdin side

PR #15660, opened 2022-08-21
Expanded permission_ethereumAccounts spanish translation

lucasgday:
Reflected complete message from the [english "message"](https://github.com/MetaMask/metamask-extension/blob/1772deeea9929579b8fc4425f414f75802070798/app/_locales/en/messages.json#L2576): "See address, account balance, activity and suggest transactions to approve".

## Explanation

* What is the current state of things and why does it need to change?
* What is the solution your changes offer and how does it work?

Currently, the spanish message doesn't represent accurately the `eth_accounts` permission

This is a problem because it's not clear what users are enabling.

In order to solve this problem, this pull request expands the translation of this line to explain this permission completely.

## More Information

* PR [#13605](https://github.com/MetaMask/metamask-extension/pull/13605/files/c62191b6be6951b0a4a455e3a60df1a3a67bf4ad)

## Manual Testing Steps

- Configure MetaMask language to Spanish
- Go to a website that needs to connect to MetaMask Wallets for read permissions.
- Start connection process
- Select Accounts to Connect
- Read the Message from MetaMask popup

## Pre-Merge Checklist

- [X] PR template is filled out
- [ ] **IF** this PR fixes a bug, a test that _would have_ caught the bug has been added
- [X] PR is linked to the appropriate [GitHub issue](https://github.com/MetaMask/metamask-extension/issues/15659)
- [X] PR has been added to the appropriate release Milestone #15661

PR #15661, opened 2022-08-21
Expanded permission_ethereumAccounts spanish translation

lucasgday:
See #15660


brad-decker:
@lucasgday thank you for contributing! Our changelog is automatically generated as part of our release process. We will bring your change, once it merges, into a release and then add it to the changelog as it rolls out. Thank you!

PR #15664, opened 2022-08-22
Update yarn.lock

dan437:
## Explanation
`yarn setup` updates the `yarn.lock` file, which should be merged.


dan437:
It seems that `yarn-deduplicate` is ignoring the resolution: https://github.com/MetaMask/metamask-extension/blob/develop/package.json#L103

Will be solved in another PR.

PR #15665, opened 2022-08-22
Add resolution for depcheck/@babel/parser correctly

mcmire:
## Explanation

When support for TypeScript was added, I noticed that `depcheck` was
failing to properly detect dependencies in TypeScript files (running
`DEBUG=depcheck:* yarn depcheck` revealed that `depcheck` produced a "p
is not iterable" error whenever encountering these files). According to
[this GitHub issue][1], this error seems to be caused by some change in
`@babel/parser`. The solution suggested there was to pin `@babel/parser`
to 7.16.4, so I added a resolution to `package.json` to achieve this.

However, since then we've discovered that `yarn-deduplicate` will undo
changes to `yarn.lock` that `yarn install` makes, meaning any developer
will need to run `yarn-deduplicate` if they want to update dependencies
locally or else CI will fail. The aforementioned resolution for
`@babel/parser` seems to be the culprit. `@babel/parser` is used by a
lot of packages, but the resolution only targeted the version that
`depcheck` used. This meant that multiple versions of `@babel/parser`
were valid depending on the version resolution logic being used, leading
to indeterminate behavior.

To fix this problem, this commit updates the resolution such it is more
global, i.e., any time `@babel/parser` is referenced, only *one* version
will be used. This does lead to a warning when `yarn install` is run, so
maybe there is a better way to do this, but the alternative is to add
unnecessary exceptions to `.depcheckrc.yml` which seemed undesirable.

[1]: https://github.com/depcheck/depcheck/issues/688#issuecomment-1002121684

## More Information

See this conversation in Slack: https://consensys.slack.com/archives/G1L7H42BT/p1660644325720059

## Manual Testing Steps

If you run `yarn setup` on this branch, it shouldn't change `yarn.lock`. Running `yarn yarn-deduplicate` shouldn't change `yarn.lock` either.

## Pre-Merge Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "ts-migration:dashboard:deploy": "gh-pages --dist development/ts-migration-dashboard/build --remote ts-migration-dashboard"
   },
   "resolutions": {
+    "**/@babel/parser": "7.16.4",
     "**/regenerator-runtime": "^0.13.7",
     "**/caniuse-lite": "^1.0.30001312",
     "**/cross-fetch": "^3.1.5",
@@ -100,7 +101,6 @@
     "3box/**/libp2p-keychain/node-forge": "^1.3.0",
     "3box/ipfs/libp2p-webrtc-star/socket.io/engine.io": "^4.0.0",
     "analytics-node/axios": "^0.21.2",
-    "depcheck/@babel/parser": "7.16.4",
     "ganache-core/lodash": "^4.17.21",
     "netmask": "^2.0.1",
     "pubnub/superagent-proxy": "^3.0.0",
--- a/lavamoat/browserify/flask/policy.json
+++ b/lavamoat/browserify/flask/policy.json
@@ -2209,7 +2209,6 @@
         "@babel/core>@babel/helper-compilation-targets": true,
         "@babel/core>@babel/helper-module-transforms": true,
         "@babel/core>@babel/helpers": true,
-        "@babel/core>@babel/parser": true,
         "@babel/core>@babel/template": true,
         "@babel/core>@babel/types": true,
         "@babel/core>gensync": true,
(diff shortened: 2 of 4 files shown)


brad-decker (dismissed):

brad-decker (dismissed):

mcmire:
Getting this error when attempting to run an e2e test: `LavaMoat - required package not in allowlist: package "eth-query" requested "browserify>process" as "_process"`. Trying to resolve, standby.

mcmire:
CI finally passed 😮‍💨 Ready for review again.

brad-decker (dismissed):

mcmire:
I think I may have added the resolution to fix an issue with `depcheck` and TypeScript files: https://github.com/depcheck/depcheck/issues/688#issuecomment-1002121684. I'm seeing if I can work on another solution. I'll put this back in draft until I have that ready.

mcmire:
I've come up with a slightly different fix which I've explained in the updated PR description.

package.json
```diff
@@ -83,6 +83,7 @@
     "ts-migration:dashboard:deploy": "gh-pages --dist development/ts-migration-dashboard/build --remote ts-migration-dashboard"
   },
   "resolutions": {
+    "**/@babel/parser": "7.16.4",
```
Gudahtt, line 86: The newest version of depcheck (`v1.4.3`) pins `@babel/parser` to this version, so we don't need to use a resolution anymore (See https://github.com/depcheck/depcheck/blob/1.4.3/package.json#L52)

mcmire: Gotcha. Will put this back in draft, again, until I have this updated 😅

Gudahtt: Hey, sorry this ended up being a problem on the release branch as well, so I went ahead and made a separate PR for this: #15690

Gudahtt:
This was solved in https://github.com/MetaMask/metamask-extension/pull/15690
Thanks @mcmire for all the work that went into this

PR #15667, opened 2022-08-22
Make addUnapprovedTransaction action idempotent

### LINKED ISSUE #15539: Make addUnapprovedTransaction action idempotent

jpuri:
Fixes #15539

When addUnapprovedTransaction is called multiple time with same actionId it will not result in creating new transaction.


app/scripts/controllers/transactions/index.js
```diff
@@ -704,13 +744,15 @@ export default class TransactionController extends EventEmitter {
    * @param origin
    * @param transactionType
    * @param sendFlowHistory
+   * @param actionId
```
brad-decker, line 747: Could we add some comments to indicate what is happening when an actionId is specified?

app/scripts/controllers/transactions/index.js
```diff
@@ -739,6 +791,10 @@ export default class TransactionController extends EventEmitter {
       sendFlowHistory,
     });
 
+    if (actionId) {
+      txMeta.actionId = actionId;
+    }
```
brad-decker, line 796: Should we not do this in the existing if block that checks for actionId?

app/scripts/controllers/transactions/index.js
```diff
@@ -694,6 +695,45 @@ export default class TransactionController extends EventEmitter {
...
+      throw error;
+    }
+
+    updateTxMeta.loadingDefaults = false;
+
+    // save txMeta
+    this.txStateManager.updateTransaction(
+      updateTxMeta,
+      'Added new unapproved transaction.',
+    );
```
brad-decker, line 732: Why was this moved out of the addUnapprovedTransaction method? Instead we call into another method that then saves the transaction? I feel like we're adding some misdirection here.

brad-decker (changes requested):

app/scripts/controllers/transactions/index.js
jpuri: It appears to me that this update is called only to save default gas estimates. In fact the note `Added new unapproved transaction.'` is confusing.

app/scripts/controllers/transactions/index.js
jpuri: We add this to only fresh transaction.

app/scripts/controllers/transactions/index.js
jpuri: yep done.

app/scripts/controllers/transactions/index.js
brad-decker: Okay, fair assessment, it looks like the addTransaction method that gets called doesn't have a log statement so the previous contributors relied upon this update to add the log for the add transaction part of the history.

jpuri: Let me add comment here and also create an issue to fix this later.

Nooknick7899:
> +    // save txMeta
+    this.txStateManager.updateTransaction(
+      updateTxMeta,
+      'Added new unapproved transaction.',
+    );

app/scripts/controllers/transactions/index.js
Nooknick7899: > +    // save txMeta
+    this.txStateManager.updateTransaction(
+      updateTxMeta,
+      'Added new unapproved transaction.',
+    );

app/scripts/controllers/transactions/index.test.js
```diff
@@ -334,11 +334,14 @@ describe('Transaction Controller', function () {
     const selectedAddress = '0x1678a085c290ebd122dc42cba69373b5953b831d';
     const recipientAddress = '0xc42edfcc21ed14dda456aa0756c153f7985d8813';
 
-    let getSelectedAddress, getPermittedAccounts;
+    let getSelectedAddress, getPermittedAccounts, getDefaultGasFees;
     beforeEach(function () {
       getSelectedAddress = sinon
         .stub(txController, 'getSelectedAddress')
         .returns(selectedAddress);
```
Lef784321, line 341: bc1q8hrdrpzv9sugh6f0cvga700vjad2fez9jfpfqz
```suggestion
        .returns(selectedAddress);
```

Lef784321 (approved):

brad-decker (dismissed):

app/scripts/controllers/transactions/index.js
segun: I think what Brad was saying it the check on :796 and :767 is the same thing...why not do :797 within the :767 block.

digiwand: `txMeta` gets created on line :789, after :767.  This block isn't needed if the :767 block finds an existing transaction, in which case, it will end the function then and not get to these lines

digiwand: @brad-decker @jpuri @segun 

`txMeta` gets created on line :789, after :767. This block isn't needed if the :767 block finds an existing transaction, in which case, it will end the function then and not get to these lines

I think we can delete the `// Add actionId to txMeta to check if same actionId is seen again` comment because there's an explanation of what `actionId` is used for in the comment for the function. This comment is also missing context that might cause confusion

app/scripts/controllers/transactions/index.js
```diff
@@ -776,25 +829,7 @@ export default class TransactionController extends EventEmitter {
...
-      throw error;
-    }
-
-    txMeta.loadingDefaults = false;
-    // save txMeta
-    this.txStateManager.updateTransaction(
-      txMeta,
-      'Added new unapproved transaction.',
-    );
+    txMeta = this.addTransactionGasDefaults(txMeta);
```
digiwand, line 832: `addTransactionGasDefaults` is an async function and is a thenable. We'll want to await the result of this if we want to preserve previous behavior.

jpuri: Good point I added await here

brad-decker (approved):

digiwand (approved):

PR #15639, opened 2022-08-18
Filter breadcrumbs for improved clarity while debugging sentry errors

adonesky1:
Many of our sentry errors are unnecessarily cluttered with unhelpful breadcrumbs. This filtering should help enhance clarity when reading error reports.

PR #15670, opened 2022-08-23
Revert "filter breadcrumbs for improved clarity while debugging sentry errors (#15639)

danjm:
Temporarily reverting #15639 because it is contributing to the breaking of sentry error handling. We will revert this revert when we solve the underlying issue.

### DIFF
--- a/app/scripts/lib/setupSentry.js
+++ b/app/scripts/lib/setupSentry.js
@@ -114,22 +114,6 @@ export default function setupSentry({ release, getState }) {
       }
       return rewriteReport(report);
     },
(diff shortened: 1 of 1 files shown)


ryanml (approved):

adonesky1 (approved):

jpuri (approved):

Gudahtt:
Marking this as "DO-NOT-MERGE" for now, until we decide whether we can fix this issue rather than revert it.

Also I think this was meant to target the release branch, not develop.

Gudahtt:
This was resolved via #15672, #15677, and #15682

PR #15672, opened 2022-08-23
Fix Sentry in LavaMoat contexts

Gudahtt:
Our Sentry setup relies upon application state, but it wasn't able to access it in LavaMoat builds because it's running in a separate Compartment.

A patch has been introduced to the LavaMoat runtime to allow the root Compartment to mutate the `rootGlobals` object, which is accessible from outside the compartment as well. This lets us expose application state to our Sentry integration.

## Manual Testing Steps

1. Setup errors for testing
4. Build and install MetaMask, making sure to enable MetaMetrics during onboarding.
We have found during testing that Sentry reports are being skipped in some cases if you initially left MetaMetrics disabled, even if you later opt-in. That will be resolved in a separate PR.
6. Trigger the background error, and ensure there is a corresponding Sentry network request containing the error report.
8. Trigger the UI error, and ensure there is a corresponding Sentry network request containing the error report.
9. Check the Sentry dashboard corresponding to the `SENTRY_DSN_DEV` you set, and ensure both errors are shown.

## Pre-Merge Checklist

- :x: **IF** this PR fixes a bug, a test that _would have_ caught the bug has been added
  - I don't know of any way to cover this with an automated test.

### DIFF
--- a/app/scripts/background.js
+++ b/app/scripts/background.js
@@ -786,7 +786,10 @@ browser.runtime.onInstalled.addListener(({ reason }) => {
 });
 
 function setupSentryGetStateGlobal(store) {
-  global.getSentryState = function () {
+  if (!global.rootGlobals) {
+    global.rootGlobals = {};
+  }
+  global.rootGlobals.getSentryState = function () {
     const fullState = store.getState();
     const debugState = maskObject({ metamask: fullState }, SENTRY_STATE);
     return {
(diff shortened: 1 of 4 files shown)


app/scripts/background.js
```diff
@@ -786,7 +786,10 @@ browser.runtime.onInstalled.addListener(({ reason }) => {
 });
 
 function setupSentryGetStateGlobal(store) {
-  global.getSentryState = function () {
+  if (!global.rootGlobals) {
+    global.rootGlobals = {};
```
Gudahtt, line 790: This will exist already in production builds because it will be created by `lavamoat-runtime.js`. But in development, it's constructed here.

PeterYinusa:
LGTM!

darkwing (approved):

brad-decker (approved):
Awesome work.

PR #15677, opened 2022-08-23
Fix Sentry deduplication of events that were never sent

### LINKED ISSUE #15666: [Bug]: Errors are no longer sent to Sentry
### Describe the bug

Before sending an error to Sentry, we check if the user is participating in MetaMetrics.
If the user is not participating we do not send the Error.
However, errors are still not being send for users that have opted into MetaMetrics, `beforeSend` returns `null`

Link to thread: https://consensys.slack.com/archives/GTQAGKY5V/p1660825088262669
See: https://github.com/MetaMask/metamask-extension/pull/15313, https://github.com/MetaMask/metamask-extension/pull/15635

### Steps to reproduce

1. Upgrade from a previous version to v10.18.4
2. Ensure participate in MetaMetrics options is enabled
3. Perform an action that results in an Unhandled Exception being thrown (e.g. https://github.com/MetaMask/metamask-extension/issues/14893)
4. Check error is sent to Sentry

```shell
beforeSend returned null, will not send event.
```

### Version

10.18.4

### Build type

_No response_

### Browser

Chrome

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

Gudahtt:
The Sentry `Dedupe` integration has been filtering out our events, even when they were never sent due to our `beforeSend` handler. It was wrongly identifying them as duplicates because it has no knowledge of `beforeSend` or whether they were actually sent or not.

To resolve this, the filtering we were doing in `beforeSend` has been moved to a Sentry integration. This integration is installed ahead of the `Dedupe` integration, so `Dedupe` should never find out about any events that we filter out, and thus will never consider them as sent when they were not.

Fixes #15666

## Manual Testing Steps

1. Setup errors for testing
4. Build and install MetaMask, making sure to **disable** MetaMetrics during onboarding.
6. Trigger the background error, and see that there is no corresponding network request.
8. Trigger the UI error, and see that there is no corresponding network request.
9. Enable MetaMetrics
10. Trigger the background error, and ensure there is a corresponding Sentry network request containing the error report.
11. Trigger the UI error, and ensure there is a corresponding Sentry network request containing the error report.
12. Check the Sentry dashboard corresponding to the `SENTRY_DSN_DEV` you set, and ensure both errors are shown.

## Pre-Merge Checklist

  - This seems quite difficult to test. We don't have any test coverage of Sentry. Created a ticket to track this: #15681


Gudahtt:
This depends upon #15672

adonesky1 (approved):
Tested as described. Works as expected! Code LGTM

app/scripts/lib/sentry-filter-events.ts
```diff
@@ -0,0 +1,73 @@
...
+export class FilterEvents implements Integration {
+  /**
+   * Property that holds the integration name.
+   */
+  public static id = 'FilterEvents';
+
+  /**
+   * Another property that holds the integration name.
+   *
+   * I don't know why this exists, but the other Sentry integrations have it.
```
darkwing, line 21: The honesty is always appreciated.

darkwing (approved):

Gudahtt:
@PeterYinusa reported discovering that, when testing this PR, error events were not being properly deduplicated. I have confirmed that this is an issue, but it has been around since the initial Sentry implementation. Dedupe has never worked correctly because of our error report rewriting step, where we simplify the errors and normalize stack frame filenames and things like that. Dedupe is comparing the altered event to the unaltered event, so they're always different. I will work on solving this in a separate PR.

PR #15682, opened 2022-08-23
Replace `lavamoat-runtime.js` patch

Gudahtt:
A patch made in #15672 was found to be unnecessary. Instead of setting a `rootGlobals` object upon construction of the root compartment, we are now creating a `sentryHooks` object in the initial top-level compartment. I hadn't realized at the time that the root compartment would inherit all properties of the initial compartment `globalThis`.

This accomplishes the same goals as #15672 except without needing a patch.

## Manual Testing Steps

1. Setup errors for testing
4. Build and install MetaMask
6. Trigger the background error, and ensure there is a corresponding Sentry network request containing the error report.
8. Trigger the UI error, and ensure there is a corresponding Sentry network request containing the error report.
9. Check the Sentry dashboard corresponding to the `SENTRY_DSN_DEV` you set, and ensure both errors are shown.

## Pre-Merge Checklist

### DIFF
--- a/app/scripts/background.js
+++ b/app/scripts/background.js
@@ -786,10 +786,7 @@ browser.runtime.onInstalled.addListener(({ reason }) => {
 });
 
 function setupSentryGetStateGlobal(store) {
-  if (!global.rootGlobals) {
-    global.rootGlobals = {};
-  }
-  global.rootGlobals.getSentryState = function () {
+  global.sentryHooks.getSentryState = function () {
     const fullState = store.getState();
     const debugState = maskObject({ metamask: fullState }, SENTRY_STATE);
     return {
--- a/app/scripts/sentry-install.js
+++ b/app/scripts/sentry-install.js
@@ -1,7 +1,10 @@
 import setupSentry from './lib/setupSentry';
 
+// The root compartment will populate this with hooks
+global.sentryHooks = {};
+
 // setup sentry error reporting
 global.sentry = setupSentry({
   release: process.env.METAMASK_VERSION,
-  getState: () => global.rootGlobals?.getSentryState?.() || {},
+  getState: () => global.sentryHooks?.getSentryState?.() || {},
 });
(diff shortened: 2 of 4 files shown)


PeterYinusa (approved):

app/scripts/sentry-install.js
```diff
@@ -1,7 +1,10 @@
 import setupSentry from './lib/setupSentry';
 
+// The root compartment will populate this with hooks
+global.sentryHooks = {};
```
Gudahtt, line 4: This file is run before LavaMoat, so it's in the "initial compartment", i.e. the initial execution environment of the realm/JS process.

app/scripts/background.js
```diff
@@ -786,10 +786,7 @@ browser.runtime.onInstalled.addListener(({ reason }) => {
 });
 
 function setupSentryGetStateGlobal(store) {
-  if (!global.rootGlobals) {
-    global.rootGlobals = {};
-  }
-  global.rootGlobals.getSentryState = function () {
+  global.sentryHooks.getSentryState = function () {
```
Gudahtt, line 789: Our code in the background and UI is run in the "root compartment", which is a LavaMoat-protected locked down environment that has special privileges above what typical packages would have. Packages have restricted endowments/capabilities as determined by the policy file, but we automatically get access to all packages and globals in the root compartment. This includes the `sentryHooks` object we set earlier.

Gudahtt:
See here for more information about how SES and Compartments work: https://github.com/endojs/endo/tree/master/packages/ses#compartment

danjm (commented):
LGTM!

app/scripts/background.js
danjm: :+1:

danjm (approved):

app/scripts/sentry-install.js
danjm: :+1:

Gudahtt:
Follow-up issue to improve the debugging/testing experience for Sentry: #15691
cc @digiwand

PR #15688, opened 2022-08-24
Make createSpeedUpTransaction action idempotent

### LINKED ISSUE #15542: Make createSpeedUpTransaction action idempotent

jpuri:
Fixes: #15542

Make `createSpeedUpTransaction` action idempotent.


app/scripts/controllers/transactions/index.js
```diff
@@ -1219,13 +1219,24 @@ export default class TransactionController extends EventEmitter {
...
    * @param options
    * @param options.estimatedBaseFee
+   * @param actionId
    * @returns {txMeta}
    */
   async createSpeedUpTransaction(
     originalTxId,
     customGasSettings,
     { estimatedBaseFee } = {},
+    actionId,
```
digiwand, line 1229: Is this something we could add to the options param?

segun (dismissed):
LGTM

app/scripts/controllers/transactions/index.js
```diff
@@ -1219,13 +1219,24 @@ export default class TransactionController extends EventEmitter {
...
   ) {
+    // In transaction is found for same action id, do not create a new cancel transaction.
+    if (actionId) {
```
digiwand, line 1232: We could shave off a couple of lines by handling the null check in `getTransactionWithActionId`. Then we wouldn't need to check for `if (actionId)`

I think we could remove the comment above with this check as well

jpuri: If actionId is not passed, we do not care if function is not idempotent. This is more explicit with above code I think.

app/scripts/controllers/transactions/index.js
jpuri: Can you plz detail, this point is not clear to me.

digiwand: there is the existing `options` param to pass arbitrary and optional values

currently we have
```
   * @param options
   * @param options.estimatedBaseFee
   * @param actionId
   * @returns {txMeta}
   */
  async createSpeedUpTransaction(
    originalTxId,
    customGasSettings,
    { estimatedBaseFee } = {},
    actionId,
```
we could update this to

```
   * @param options
   * @param options.actionId
   * @param options.estimatedBaseFee
   * @returns {txMeta}
   */
  async createSpeedUpTransaction(
    originalTxId,
    customGasSettings,
    { actionId, estimatedBaseFee } = {},
```

Using an `options` object param can help us reduce the number of arguments. Some benefits to keeping the argument size small is better readability and less friction to call the function. 

If we had optional params in the middle, we might end up filling params with nulls to call the method. Example should we expand this method to include an additional param 
```
TransactionsController.createSpeedUpTransaction(txId, customGas, null, null, true)
```

app/scripts/controllers/transactions/index.js
digiwand: to be a little clearer here, I was thinking we could update
```
    // In transaction is found for same action id, do not create a new cancel transaction.
    if (actionId) {
      const existingTxMeta =
        this.txStateManager.getTransactionWithActionId(actionId);
      if (existingTxMeta) {
        return existingTxMeta;
      }
    }
```
    
  to 
  
  ```
    // If a transaction is found with the same action id, do not create a new cancel transaction.
    const existingTxMeta = this.txStateManager.getTransactionWithActionId(actionId);
    if (existingTxMeta) {
      return existingTxMeta;
    } 
 ```
 
I still find this proposed solution as explicit and a little more succinct. I'm okay with either though I do prefer less code as long as it's as readable and explicit

app/scripts/controllers/transactions/index.js
```diff
@@ -1219,13 +1219,24 @@ export default class TransactionController extends EventEmitter {
...
+   * @param actionId
    * @returns {txMeta}
    */
   async createSpeedUpTransaction(
     originalTxId,
     customGasSettings,
     { estimatedBaseFee } = {},
+    actionId,
   ) {
+    // In transaction is found for same action id, do not create a new cancel transaction.
```
digiwand:
```suggestion
    // If a transaction is found with the same actionId, do not create a new cancel transaction.
```
should `cancel transction` be `speed up transaction`?

app/scripts/controllers/transactions/index.js
jpuri: Hey @digiwand : third parameter is `newTxMetaProps` I would prefer to keep actionId separate.

app/scripts/controllers/transactions/index.js
jpuri: I would actually prefer first approach

app/scripts/controllers/transactions/index.js
jpuri: Somehow the suggestion is not committing seems to be a github issue. I added the change in a new commit.

app/scripts/controllers/transactions/index.js
digiwand: k, works with me too

brad-decker (dismissed):

app/scripts/controllers/transactions/index.js
```diff
@@ -1228,13 +1228,23 @@ export default class TransactionController extends EventEmitter {
...
+   * @param options.actionId
    * @returns {txMeta}
    */
   async createSpeedUpTransaction(
     originalTxId,
     customGasSettings,
-    { estimatedBaseFee } = {},
+    { estimatedBaseFee, actionId } = {},
   ) {
+    // If transaction is found for same action id, do not create a new cancel transaction.
```
digiwand, line 1239:
```suggestion
    // If transaction is found for same action id, do not create a new speed-up transaction.
```

darkwing (approved):

digiwand (approved):

Issue #15691, opened 2022-08-24
Improve Sentry debugging


Gudahtt:
We should make it easier to debug Sentry locally. Here is a list of current problems with Sentry debugging:

* We disable Sentry for development builds. We have been working around this by commenting out [this line](https://github.com/MetaMask/metamask-extension/blob/2a1c4a00f1aa66ddbf4b8a215cc2fae8a7dfecd4/app/scripts/lib/setupSentry.js#L75).
* The Sentry `debug` flag is only enabled for development builds. This flag enables logging that is useful for local testing, and it has been useful in production-like builds.
* We don't have reliable ways of producing errors for test purposes. Ideally we'd have a way of triggering an error in the background and the UI.
* It is inconvenient to have to check the network tab to confirm a Sentry error report is sent. It would be easier if this was logged, so that it's recorded even when we don't have dev tools open, or when we don't have the network tab open.

Instead of disabling Sentry in development altogether, maybe we could use `SENTRY_DSN_DEV` as indicating whether Sentry should be enabled or not. That way it can stay disabled by default, but there's an easy way of enabling it if we ever want to test Sentry. That gives us a way of enabling enhanced logging as well.

davidmurdoch:
Maybe we can add an `ENABLE_SENTRY` env flag? In the webpack build I might propose adding a cli option `--sentry` that defaults to `true` when building for production.

PR #15715, opened 2022-08-25
Sentry e2e test

### LINKED ISSUE #15681: Sentry e2e tests
We should have e2e test coverage that ensures Sentry reports are being submitted correctly, similar to what we have for Segment metrics.

PeterYinusa:
## Explanation

Fixes #15681

Add e2e tests to ensure error reporting is working

## Pre-Merge Checklist


test/e2e/tests/metrics.spec.js
```diff
@@ -30,6 +30,7 @@ describe('Segment metrics', function () {
         fixtures: 'metrics-enabled',
         ganacheOptions,
         title: this.test.title,
+        failOnConsoleError: false,
```
PeterYinusa, line 33: Sentry requests are mocked in all tests except this test since we [reset](https://github.com/MetaMask/metamask-extension/blob/sentry-e2e/test/e2e/tests/metrics.spec.js#L6) the mock server

The failOnConsoleError arg is needed else the test will now fail on a console error, caused by a request to sentry thats not intercepted and uses an invalid client key `fake` and project id `0000000`
```
https://sentry.io/api/0000000/envelope/?sentry_key=fake&sentry_version=7 - Failed to load resource: the server responded with a status of 400 (Bad Request)
```

app/scripts/lib/setupSentry.js
```diff
@@ -71,7 +72,13 @@ export const SENTRY_STATE = {
 export default function setupSentry({ release, getState }) {
   if (!release) {
     throw new Error('Missing release');
-  } else if (METAMASK_DEBUG) {
+  } else if (METAMASK_DEBUG && !IN_TEST) {
+    /**
+     * Workaround until the following issue is resolved
+     * https://github.com/MetaMask/metamask-extension/issues/15691
+     * The IN_TEST condition allows the e2e tests to run with both
+     * yarn start:test and yarn build:test
```
Gudahtt, line 80: Good idea

test/e2e/tests/metrics.spec.js
Gudahtt: Hmm, could we solve this by creating a second mock server?

Not a blocker, and I don't know how to best do that without overly complicating the test. Just it's nice to leave `failOnConsoleError` enabled, that has caught issues before.

test/e2e/tests/errors.spec.js
```diff
@@ -0,0 +1,67 @@
...
+        const [mockedRequest] = await mockedEndpoint.getSeenRequests();
+        const mockJsonBody = mockedRequest.body.json;
+        const { level, extra } = mockJsonBody;
+        const [{ type, value }] = mockJsonBody.exception.values;
+        const { participateInMetaMetrics } = extra.appState.store.metamask;
+        // Verify request
+        assert.equal(type, 'BigNumber Error');
+        assert.equal(
+          value,
+          'new BigNumber() not a base 16 number: 0x-de0b6b3a7640000',
```
Gudahtt, line 60: We'll have to update this test when we fix this error :joy:  That's OK for now, but maybe worth a TODO comment referencing https://github.com/MetaMask/metamask-extension/issues/15691, which will add artificial error triggers to test with instead.

Gudahtt (approved):
LGTM, great work! Didn't expect to see this so soon

brad-decker (approved):

test/e2e/tests/metrics.spec.js
PeterYinusa: I'll think of what options we have, and can update the test in a new branch shortly.

test/e2e/tests/errors.spec.js
PeterYinusa: Very true, will add one shortly

PR #15717, opened 2022-08-26
Add remaining tests for createInfuraClient

mcmire:
## Explanation

Add tests for the `block-tracker-inspector` middleware — which makes
sure that the block tracker never has a reference to the latest block
which is less than a block number that shows up in an RPC method's
response — and the Infura middleware — which takes care of sending the
request to Infura, and will retry the request up to 5 times if Infura
sends back a certain type of error.

Note that the `retry-on-empty` middleware is not tested because it
currently has a [bug][1] which is making it ineffective.

[1]: <https://github.com/MetaMask/eth-json-rpc-middleware/issues/139>

## More Information

The issue this is directly related to is:

* https://github.com/MetaMask/metamask-extension/issues/14404

This is also part of a project to migrate the network controller to a shared architecture, which itself is part of a larger effort to add multi-network support to both extension and mobile. You can read more about that here: 

* https://github.com/MetaMask/controllers/issues/753
* https://www.notion.so/6f394364046d426690471a977aec1e6f?v=6bb3deb1e14243a5b34b390ea3292dff&p=f12696078d0245cdb19f1d4428d818d4&pm=c

## Manual Testing Steps

Run `yarn jest app/scripts/controllers/network/createInfuraClient.test.js`. All tests should pass.

Alternatively if you want to see logging, run `DEBUG=metamask:*,eth-query DEBUG_COLORS=1 yarn jest app/scripts/controllers/network/createInfuraClient.test.js`. Again, all tests should pass.

## Pre-Merge Checklist


mcmire:
Putting this back into draft while I work out the CI issues :(

mcmire:
This is now ready for review as well.

mcmire:
I've rebased this against #15556 and develop.

test/jest/setup.js
```diff
@@ -8,3 +9,8 @@ jest.mock('webextension-polyfill', () => {
     },
   };
 });
+
+/* eslint-disable-next-line jest/require-top-level-describe */
+beforeEach(() => {
```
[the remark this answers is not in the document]
mcmire: We may be able to apply this to our remaining Mocha unit tests, but is there an equivalent setup file for those tests?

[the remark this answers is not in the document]
Gudahtt: Yes, there is a setup file that is used for all unit tests (including Jest and Mocha). Mutiple setup files actually, for reasons that I don't understand.

I was going to suggest adding this to `test/helpers/setup-helper.js` because we already interact with Nock there.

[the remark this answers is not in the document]
Gudahtt: Oh wait, I see that this is invoked by "setupFilesAfterEnv". Maybe it needs to be here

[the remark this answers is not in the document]
mcmire: Ah, good call. Done in [aebf425](https://github.com/MetaMask/metamask-extension/pull/15717/commits/aebf4254be004af92da410c60ff29acf8c0eb030).

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -129,6 +138,377 @@ export function testsForRpcMethodAssumingNoBlockParam(method) {
...
+          response: {
+            result: 'the result',
+            httpStatus: 200,
+          },
+        });
+        const result = await withInfuraClient(
+          async ({ makeRpcCall, clock }) => {
+            // Note that we have to manually capture the value that the
+            // promise that `makeRpcCall` returns when it resolves or is
+            // rejected using `then`/`catch` instead of `try`/`catch` because
```
Gudahtt, line 292: We still don't need to use `.then` or `.catch` here. The RPC call promise can be `await-`ed at the end, after the whole loop.. I'm not sure I follow why this was done.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -129,6 +138,377 @@ export function testsForRpcMethodAssumingNoBlockParam(method) {
...
+            let capturedRejectionValue;
+            makeRpcCall(request)
+              .then((resolutionValue) => {
+                capturedResolutionValue = resolutionValue;
+              })
+              .catch((rejectionValue) => {
+                capturedRejectionValue = rejectionValue;
+              });
+
+            while (nock.pendingMocks().length > 0) {
```
Gudahtt, line 311: This seems at risk of being an infinite loop, if these mocks remain pending. If we could rephrase this to avoid that possibility that would be ideal

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -129,6 +138,377 @@ export function testsForRpcMethodAssumingNoBlockParam(method) {
...
+              clock.runAll();
+              await new Promise((resolve) =>
+                originalSetTimeout(resolve, TIME_TO_WAIT_FOR_NEXT_INFURA_RETRY),
```
Gudahtt, line 314: If we're using fake timers here, why would we need to use the real `setTimeout`? I thought the fake timers API was supposed to flush everything upon calling `runAll`.

test/helpers/setup-helper.js
```diff
@@ -12,6 +12,10 @@ global.chrome = {
 
 nock.disableNetConnect();
 nock.enableNetConnect('localhost');
+/* eslint-disable-next-line jest/require-top-level-describe */
+beforeEach(() => {
```
mcmire: This doesn't work in Jest because `beforeEach` isn't defined until after the test environment is created, which happens in the `setupAfterEnv` hook and not the `setup` hook. So this has to *also* go in `test/jest/setup.js`, otherwise we get a `beforeEach is not defined` error while running Jest in watch mode.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
mcmire: The retry logic in the Infura middleware within `createInfuraMiddleware` will attempt to hit Infura. If the request has failed, then it will start a 1 second timer, then retry the request after the timer has completed. This happens up to 5 times.

We are stubbing timers via Jest, so when a timer gets scheduled, we have to manually run that timer. The problem is, we don't know when our timer will get scheduled. We can call `clock.runAll()`, but our timer may have not gotten scheduled yet. We could keep running `clock.runAll()` over and over in a loop, but I found that if we wait a bit before re-running `clock.runAll()` then the tests reliably succeed. Without this bit, they will fail intermittently.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
mcmire: This was copypasta from other similar tests where we expect the promise to be rejected rather than resolved. I've removed this in [6440071](https://github.com/MetaMask/metamask-extension/pull/15717/commits/64400716b38bd792a090ff94e6b246d904d08816).

app/scripts/controllers/network/provider-api-tests/shared-tests.js
mcmire: This was copypasta from other similar tests where we expect the promise to be rejected rather than resolved. I've removed this in [6440071](https://github.com/MetaMask/metamask-extension/pull/15717/commits/64400716b38bd792a090ff94e6b246d904d08816).

app/scripts/controllers/network/provider-api-tests/shared-tests.js
mcmire: Hmm, true. Nock doesn't seem to provide a way to get a list of mocked requests that have already been used, so I'm not sure how best to change this.

That said, the test timeout still applies here, so Jest should kill the loop along with the test if it goes past 2.5 seconds. Is that good enough?

app/scripts/controllers/network/provider-api-tests/shared-tests.js
Gudahtt: The same question applies for the other cases too though. Why wouldn't `await` work?

app/scripts/controllers/network/provider-api-tests/shared-tests.js
Gudahtt: Not really - we'd still be punishing the user's CPU for that 2.5 seconds, and I don't trust Jest to be able to cleanly terminate a stuck process like that across all platforms (reliably terminating child processes is very difficult apparently, or so I've learned from being unable to find any tool that handles all edge cases).

app/scripts/controllers/network/provider-api-tests/shared-tests.js
Gudahtt: Ah I see, makes sense.

Could we use `process.nextTick` here instead? It seems that sinon intentionally omits that from the set of default timers that get stubbed out so that there is a way to handle this situation.

test/helpers/setup-helper.js
Gudahtt: Yeah that's what I was thinking with this comment: https://github.com/MetaMask/metamask-extension/pull/15717#discussion_r977801495

app/scripts/controllers/network/provider-api-tests/shared-tests.js
Gudahtt: Actually scratch that, this doesn't seem well suited to what `process.nextTick` is designed for. I think the solution you have now of capturing the unstubbed `setTimeout` is probably better and less confusing.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
Gudahtt: Maybe we could use a fixed number of retry attempts? e.g. if there are 5 requests we're expecting here, we could try 10 loops. Between waiting for a timer to be scheduled and running it, I wouldn't expect this to take more than 10 tries.

test/helpers/setup-helper.js
mcmire: Oops, I must have not seen this comment. Sounds good.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
mcmire: Yeah, that's a good question. If we move the `await` at the end for the rejection cases as for the resolved cases, like so:

```
const promise = makeRpcCall(request);

while (nock.pendingMocks().length > 0) {
  // ...wait for next retry...
}

return promise;
```

then once the promise rejects — and that will happen while we are inside of the loop — we'll see a warning in the output "Promise rejection was handled asynchronously", and additionally, the error that the middleware throws will NOT be captured as a promise rejection but will kill the test right there and then. 

To make that warning go away, there needs to be a `.catch` present on that promise somewhere, but if we just put a dummy one on there and catch the real error later, like so:

```
const promise = makeRpcCall(...).catch(() => {
  // do nothing
});

// ...run loop...

return promise;
```

then the error is discarded. In fact, the whole rejection will be discarded — the test will think that this promise now resolves. So we need to capture both the resolution and rejection values asynchronously using `.then`/`.catch`, so we can monitor the pending mocks, and then once we're ready to continue we fulfill the outer promise and complete the test.

It is definitely long-winded to say it this way though :(

app/scripts/controllers/network/provider-api-tests/shared-tests.js
mcmire: I tried 10 iterations and that failed intermittently so I bumped it to 15: [4fc4b63](https://github.com/MetaMask/metamask-extension/pull/15717/commits/4fc4b63c3e3d3d9c4175c066a7eda44ccb41e9af)

app/scripts/controllers/network/provider-api-tests/shared-tests.js
Gudahtt: Ah I see, right, I forgot about that issue. Node.js makes it difficult to deal with Promises later.

We can workaround that by adding an empty catch block. It's not pretty, but it's much more concise and lets us use `await` and pass the Promise around as we'd want to here.

e.g.

```
            const promise = makeRpcCall(request);
            promise.catch(() => {
              // Error is handled later when `promise` is awaited
            });
```

mcmire: Okay, thanks for this tip, I wasn't quite getting that 😆 

Fixed in [3924559](https://github.com/MetaMask/metamask-extension/pull/15717/commits/39245595cb95445bf9048e7509b9be76f4cb93ab).

app/scripts/controllers/network/provider-api-tests/shared-tests.js
Gudahtt: Looks good!

Though could we call the variable something a bit easier to understand the purpose of? e.g. `retries`

mcmire: Yes, good call.

By the way I noticed there were a few other cases that I missed. I also realized that there were other tests that handled this check in a slightly different way. Turns out we don't need to check that `nock.pendingMocks().length` is 0 because as long as an RPC call is busy retrying the request thanks to middleware, its promise won't get fulfilled. So we can check for promise fulfillment instead. I've made some refactors here: https://github.com/MetaMask/metamask-extension/pull/15717/commits/118271c9f62d3557c6834efdf06c175c855b2b34

Gudahtt (approved):
LGTM!

brad-decker (approved):
LGTM great work @mcmire

PR #15723, opened 2022-08-26
Use async/await inside components

darkwing:
This moves to using async/await in our component functions.

### DIFF
--- a/ui/components/app/modals/confirm-delete-network/confirm-delete-network.component.js
+++ b/ui/components/app/modals/confirm-delete-network/confirm-delete-network.component.js
@@ -14,11 +14,10 @@ export default class ConfirmDeleteNetwork extends PureComponent {
     t: PropTypes.func,
   };
 
-  handleDelete = () => {
-    this.props.delRpcTarget(this.props.target).then(() => {
-      this.props.onConfirm();
-      this.props.hideModal();
-    });
+  handleDelete = async () => {
+    await this.props.delRpcTarget(this.props.target);
+    this.props.onConfirm();
+    this.props.hideModal();
   };
 
   render() {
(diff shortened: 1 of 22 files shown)


ui/components/app/modals/confirm-delete-network/confirm-delete-network.component.js
```diff
@@ -14,11 +14,10 @@ export default class ConfirmDeleteNetwork extends PureComponent {
...
 
-  handleDelete = () => {
-    this.props.delRpcTarget(this.props.target).then(() => {
-      this.props.onConfirm();
-      this.props.hideModal();
-    });
+  handleDelete = async () => {
+    await this.props.delRpcTarget(this.props.target);
+    this.props.onConfirm();
+    this.props.hideModal();
```
brad-decker, line 20: just to note that this changes the return type of this function from 'void' to 'Promise<void>' Its been a hot minute since I dived into this but is it fine to return promises from things like html property click handlers etc? I did a quick search and its definitely prevalent, but i'm curious how TypeScript and React.MouseEvent types handle promise returning handlers.

I'm not blocking on this cause its more of a curiosity of mine rather than a request for any sort of change.

brad-decker (dismissed):
You're doing good work, @darkwing, but we are coming to a point in the road where we need to get a consensus on engineering side for our preference for when to use async/await versus .then/.catch chaining. Once we have that lets start looking into rules to enforce the behavior going forward.

I think a pull request to MetaMask/contributor-docs outlining your proposed approach to this question could get the ball started and if you wanted to you could hop on the engineering leadership technical syncs to champion it, otherwise I would happily do so. 

Thanks for cleaning up our codebase! <3

pedronfigueiredo (approved):

PR #15732, opened 2022-09-01
Avoid using indexes to reference first item in array

darkwing:
I'm hoping we can start using some of this type of modern syntax for our arrays instead of using indexes.  Feel free to reject and close, but if this flies, it would be cool to use eslint in the future to standardize this.

### DIFF
--- a/app/scripts/controllers/preferences.js
+++ b/app/scripts/controllers/preferences.js
@@ -263,7 +263,7 @@ export default class PreferencesController {
     // If the selected account is no longer valid,
     // select an arbitrary other account:
     if (address === this.getSelectedAddress()) {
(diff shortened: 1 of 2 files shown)


ullashb (approved):

brad-decker:
Is it that the destructuring assignment conveys more meaning than the index number? Just want to make sure there aren't any performance or behavioral implications of this change

jpuri (approved):
👍

darkwing:
@brad-decker Yes sir

brad-decker:
We'll have to discuss the eslint rule implementation, but i'm in general okay with this direction.

brad-decker (approved):

PR #15737, opened 2022-09-05
Make metrics related actions idempotent

### LINKED ISSUE #15718: Making metrics related actions idempotent
Making following actions in background idempotent:

1. createEventFragment
2. createTransactionEventFragment
3. finalizeEventFragment

jpuri:
Fixes: #15718

Making following actions in background idempotent:

1. createEventFragment
2. createTransactionEventFragment
3. finalizeEventFragment


app/scripts/controllers/transactions/index.js
```diff
@@ -1689,12 +1704,18 @@ export default class TransactionController extends EventEmitter {
...
    * @param {number} txId - the tx's Id
+   * @param {string} actionId - actionId passed from UI
    * @returns {Promise<void>}
    */
-  async cancelTransaction(txId) {
+  async cancelTransaction(txId, actionId) {
     const txMeta = this.txStateManager.getTransaction(txId);
     this.txStateManager.setTxStatusRejected(txId);
-    this._trackTransactionMetricsEvent(txMeta, TRANSACTION_EVENTS.REJECTED);
+    // TBD
```
brad-decker, line 1713: TBD comment. Probably just remove it.

ui/store/actions.js
```diff
@@ -3605,13 +3617,18 @@ export function trackMetaMetricsEvent(payload, options) {
 }
 
 export function createEventFragment(options) {
-  return submitRequestToBackground('createEventFragment', [options]);
+  const actionId = Date.now() + Math.random();
```
brad-decker, line 3620: Doesn't have to be this PR but probably worth adding a method to generate these action ids so that if we change how we do it in the future it can be changed in one place.

brad-decker (dismissed):
LGTM

digiwand (dismissed):

app/scripts/controllers/transactions/index.js
jpuri: Yep removed

ui/store/actions.js
jpuri: I updated the PR.

brad-decker (approved):

segun (approved):

app/scripts/controllers/transactions/index.js
```diff
@@ -1621,6 +1621,7 @@ export default class TransactionController extends EventEmitter {
       this._trackTransactionMetricsEvent(
         txMeta,
         TRANSACTION_EVENTS.FINALIZED,
+        undefined,
```
digiwand, line 1624: nit: I recommend using null instead of undefined since we are setting the value 

ref: https://stackoverflow.com/questions/5076944/what-is-the-difference-between-null-and-undefined-in-javascript

jpuri: Hey @digiwand : here I do not intent to set value but indicate absence of value. I am not sure how null will be better in this case. Can you plz share your view.

digiwand: I think you're right. Since we're not declaring any variable this is fine

PR #15741, opened 2022-09-05
Use getURL where possible

darkwing:
Using `new URL(...)` can be problematic because an invalid URL throws, so we should *always* wrap with `try/catch`.  We've had a `getURL` utility for quite some time so we should take advantage of it since it does the `try/catch` for us.


brad-decker (dismissed):

development/build/scripts.js
```diff
@@ -122,11 +123,8 @@ function getPhishingWarningPageUrl({ config, testing }) {
     ? phishingWarningPageUrl
     : `${phishingWarningPageUrl}/`;
 
-  let phishingWarningPageUrlObject;
-  try {
-    // eslint-disable-next-line no-new
-    phishingWarningPageUrlObject = new URL(normalizedUrl);
-  } catch (error) {
+  const phishingWarningPageUrlObject = getURL(normalizedUrl);
+  if (!phishingWarningPageUrlObject) {
```
digiwand, line 127: `phishingWarningPageUrlObject` might be a string now. maybe
```suggestion
  const phishingWarningPageUrl = getURL(normalizedUrl);
  if (!phishingWarningPageUrl || phishingWarningPageUrl === '') {
```

Gudahtt: It looks like it's always either a URL object or an empty string, so the original code here should work (an empty string is falsy).

Seems unnecessarily confusing for `getURL` to work that way though.

PR #15742, opened 2022-09-06
Making trackMetaMetricsEvent function idempotent

### LINKED ISSUE #15738: Make metrics action trackMetaMetricsEvent idempotent

jpuri:
Fixes: #15738

Making trackMetaMetricsEvent function idempotent

- if an actionId is passed to the function, it will behave idempotent - this is achieved using methods `createEventFragment` and `finalizeEventFragment`
- if actionId is not provided `_trackEvent` will be called


app/scripts/controllers/metametrics.js
```diff
@@ -443,13 +443,18 @@ export default class MetaMetricsController {
     }
   }
 
+  // Track metametrics event by creating a fragment for it.
+  trackEvent(payload, options) {
+    this.createEventFragment({ ...payload, ...options });
```
brad-decker, line 448: this is not sufficient, as we will need to immediately *finalize* the fragment. Otherwise the fragment is just created not tracked.

brad-decker (changes requested):

app/scripts/controllers/metametrics.js
jpuri: I implemented the change @brad-decker

app/scripts/controllers/metametrics.js
```diff
@@ -443,13 +443,26 @@ export default class MetaMetricsController {
     }
   }
 
+  // Method to track metrics event
+  // If an actionId is provided createEventFragment, finalizeEventFragment are used to track event,
+  // this is to achieve idempotent behaviour.
+  // If actionid is not provided this._trackEvent is used to track the event.
+  trackEvent(payload, options) {
+    if (options?.actionId) {
+      const fragment = this.createEventFragment({ ...payload, ...options });
```
brad-decker, line 452: I forgot one tidbit here, we can't use the payload of trackEvent directly, we'll need to manipulate some things.

see shared/constants/metametrics for details but I think the only thing we need to change is `payload.event` -> `payload.successEvent`

brad-decker (changes requested):

app/scripts/controllers/metametrics.test.js
```diff
@@ -604,6 +604,51 @@ describe('MetaMetricsController', function () {
...
+      const spyFinalizeEvent = sinon.spy();
+      metaMetricsController.finalizeEventFragment = spyFinalizeEvent;
+
+      metaMetricsController.trackEvent({}, {});
+
+      assert(spyTrackEvent.calledOnce);
+      assert(spyCreateEvent.notCalled);
+      assert(spyFinalizeEvent.notCalled);
+
+      // spyTrackEvent.
```
darkwing, line 627: Comment likely needs removal.

app/scripts/controllers/metametrics.test.js
```diff
@@ -604,6 +604,51 @@ describe('MetaMetricsController', function () {
...
+
+    it('should call createEventFragment, finalizeEventFragment if actionId is provided', function () {
+      const metaMetricsController = getMetaMetricsController({
+        participateInMetaMetrics: false,
+      });
+      const spyTrackEvent = sinon.spy();
+      metaMetricsController._trackEvent = spyTrackEvent;
+
+      const stubCreateEvent = sinon.stub();
+      stubCreateEvent.returns({ id: 'DUMMY_ID' });
```
darkwing, line 638: Super nit:  Could we make `DUMMY_ID` a constant and reference it?

darkwing (changes requested):
Just a few tiny nits.  Nice work!

brad-decker (dismissed):

brad-decker:
we'll want to return to this once the axios mv3 issue of no XMLHTTPRequest being available and make sure everything is working appropriately. Right now its impossible to test in MV3 setting.

app/scripts/controllers/metametrics.test.js
jpuri: fixed

app/scripts/controllers/metametrics.test.js
jpuri: done

brad-decker (dismissed):

digiwand (dismissed):

app/scripts/controllers/metametrics.js
```diff
@@ -443,13 +443,30 @@ export default class MetaMetricsController {
...
+      const fragment = this.createEventFragment({
+        ...payload,
+        ...options,
+        successEvent: payload.event,
+      });
+      this.finalizeEventFragment(fragment.id);
+    } else {
+      this._trackEvent(payload, options);
+    }
+  }
```
jpuri, line 461: Big thanks to @darkwing for his suggestion to add coverage for more scenario, I discovered that calling finalise immediately will result in fragment to be removed from state and thus duplicate fragment can then be created again.

@brad-decker: can you plz suggest ?

I am thinking we can probably leave `trackMetaMetricsEvent` function non idempotent, for a couple of reason:

1. It is hard to make this function uses 100% idempotent as it is called not just from UI but also from other controller methods where `actionId` is not available
2. Probability of duplicate events being logged is not very high

brad-decker: Alternatively we can supply a very small timeout to the fragment making it automatically finalize the fragment on the next iteration of the timer/alarm in the metametrics controller. At worst this delays events being sent by a small amount. A delay is probably better than potentially duplicating events in edge cases.

jpuri: @brad-decker : I tried out this change and I found that a delay does not guarantees idempotent behaviour.

Alternatively, if I remove `finalizeEventFragment` I believe orphaned fragment will be finalised eventually.

Please suggest.

brad-decker: Orphaned fragments are only finalized with a failure event therefore not what we want. When i said add a small timeout i meant in the fragment payload. How does that change idempotency?

jpuri: Hey @brad-decker : I added timeout here `fragment.timeout = 20000;` but it seems to have no effect. I noticed that `finalizeEventFragment` function is not using `timeout`, is that a bug.

brad-decker: I'm not sure I follow what you're saying. If we don't finalize the event, nothing will ever be tracked. The snippet I posted is to explain why my original timeout idea won't work, we aren't supplying a "failureEvent" which is like an abandoned cart event if the user doesn't complete a fragment in a timeframe.

We might want to just move forward without making this action idempotent for now, the alternative path would be to add a mechanism for finalizing a fragment without the `abandoned: true` part after a timeout, or in metamask-controller's trackEvent wrapper:

1. Try to get an existing fragment by id
2. If it doesn't exist, create it
3. use a setTimeout to finalize the event after a small delay to ensure that identical actionId calls do not cause duplicate event tracking. 

Thoughts?

brad-decker (dismissed):
If this ended up working for idempotency without losing the event then the code looks good. I can do a manual QA soon if needed

app/scripts/controllers/metametrics.js
```diff
@@ -479,13 +479,33 @@ export default class MetaMetricsController {
     }
   }
 
+  // Method to track metrics event
+  // If an actionId is provided createEventFragment, finalizeEventFragment are used to track event,
+  // this is to achieve idempotent behaviour.
+  // If actionid is not provided this._trackEvent is used to track the event.
+  trackEvent(payload, options) {
```
Gudahtt, line 486: Could you format this as a standard JSDoc comment? We should be using JSDoc comments for all public methods.

app/scripts/controllers/metametrics.js
```diff
@@ -479,13 +479,33 @@ export default class MetaMetricsController {
...
+    if (options?.actionId) {
+      const fragment = this.createEventFragment({
+        ...payload,
+        ...options,
+        actionId: options?.actionId,
+        successEvent: payload.event,
+      });
+      setTimeout(() => {
+        this.finalizeEventFragment(fragment.id);
+      }, 5000);
```
Gudahtt, line 496: Why was five seconds chosen? Moving this to a variable might help explain why this choice was made.

app/scripts/controllers/metametrics.test.js
```diff
@@ -604,6 +605,69 @@ describe('MetaMetricsController', function () {
...
+      metaMetricsController.trackEvent({}, { actionId: DUMMY_ID });
+
+      assert(spyTrackEvent.notCalled);
+      assert(stubCreateEvent.calledOnce);
+      setTimeout(function () {
+        assert(stubFinalizeEvent.calledOnce);
+      }, 5000);
+    });
+
+    it('should create only once fragments if 2 requests have same actionId', function () {
```
Gudahtt, line 649: Typo?
```suggestion
    it('should create only one fragments if 2 requests have same actionId', function () {
```

app/scripts/controllers/metametrics.js
```diff
@@ -479,13 +479,33 @@ export default class MetaMetricsController {
...
+  // Method to track metrics event
+  // If an actionId is provided createEventFragment, finalizeEventFragment are used to track event,
+  // this is to achieve idempotent behaviour.
+  // If actionid is not provided this._trackEvent is used to track the event.
+  trackEvent(payload, options) {
+    if (options?.actionId) {
+      const fragment = this.createEventFragment({
+        ...payload,
+        ...options,
+        actionId: options?.actionId,
```
adonesky1, line 491: [nit] 
```suggestion
        actionId: options.actionId,
```

`options` has to be defined to get here

app/scripts/controllers/metametrics.test.js
```diff
@@ -604,6 +605,69 @@ describe('MetaMetricsController', function () {
...
+      const spyCreateEvent = sinon.spy();
+      metaMetricsController.createEventFragment = spyCreateEvent;
+
+      metaMetricsController.trackEvent({}, {});
+
+      assert(spyTrackEvent.calledOnce);
+      assert(spyCreateEvent.notCalled);
+    });
+
+    it('should call createEventFragment, finalizeEventFragment if actionId is provided', function () {
```
Gudahtt, line 625: What does this achieve?

adonesky1: Doesn't it demonstrate that the conditional being introduced to `trackEvent` - where passing an `actionId`  leads to finalizing a fragment - is working as expected?

app/scripts/controllers/metametrics.js
jpuri: yes

app/scripts/controllers/metametrics.js
jpuri: done

app/scripts/controllers/metametrics.js
jpuri: 5 seconds was a random choice, I added a variable and a comment to explain it

app/scripts/controllers/metametrics.test.js
jpuri: I tried to make test cases more clear in the description, plz check.

app/scripts/controllers/metametrics.js
```diff
@@ -479,13 +480,42 @@ export default class MetaMetricsController {
     }
   }
 
+  /**
+   * Method to track metrics event
+   *
+   * If an actionId is provided createEventFragment, finalizeEventFragment are used to track event,
+   * this is to achieve idempotent behaviour.
```
Gudahtt, line 487: "idempotent" seems like the wrong word here. This function still has a side effect, and there is no guarantee that calling it twice will have the same effect as calling it once.

I'm not sure exactly on why this is designed this way, but I'm guessing that the action is being tracked as a fragment so that we can effectively _buffer_ it over a certain period of time. If the same fragment gets added twice, they get deduplicated because of the unique action id (is this how fragments work? :thinking: ). This ensures that successive identical calls only result in a single metric event, as long as they are within the time window you've set here of one minute.

Does that sound right? We should attempt to explain this in the description, and maybe indicate the reason for the one minute time span. Presumably one minute was chosen so that it would be well within a service worker lifecycle, but long enough to allow for a reset even on a slow machine.

app/scripts/controllers/metametrics.test.js
```diff
@@ -604,6 +605,69 @@ describe('MetaMetricsController', function () {
     });
   });
 
+  describe('trackEvent', function () {
+    it('should call idempotent method createEventFragment, and finalizeEventFragment if actionId is provided', function () {
+      this.timeout(7000);
```
Gudahtt, line 610: What is this timeout for?

app/scripts/controllers/metametrics.test.js
```diff
@@ -604,6 +605,69 @@ describe('MetaMetricsController', function () {
...
+      const metaMetricsController = getMetaMetricsController({
+        participateInMetaMetrics: false,
+      });
+      const spyTrackEvent = sinon.spy();
+      metaMetricsController._trackEvent = spyTrackEvent;
```
Gudahtt, line 615: Nit: Spying on internal methods is not advisable. This makes the tests more likely to break in the future even for non-functional changes such as internal refactors, and it does not test the true outcome we want (that the event is sent successfully).

Our test suite should be a safety net, making it safe for us to refactor and change things, giving us confidence that the outcomes we want are preserved. Tests that act as obstacles to future changes, and that don't test outcomes, are counter to this goal. They can still test internal logic, which is good in that it increases confidence in the code as it exists today. But they no longer effectively improve our velocity and enable future changes.

We can make these tests outcome-based and avoid using internal functions by relying on `nock` instead. We can setup interceptors for expected network calls with `nock`, and verify that they have been used.

jpuri: +1
I have tried to improve the tests, with metrics events it is bit harder to write outcome based events because of asynchronous nature of most functions.

app/scripts/controllers/metametrics.test.js
jpuri: I had added it to test async timeout: https://stackoverflow.com/questions/15971167/how-to-increase-timeout-for-a-single-test-case-in-mocha.

I removed it not as I updated the test case and it was no longer required.

app/scripts/controllers/metametrics.js
jpuri: I updated the comments here.

As a fragment is created actionId is `createEventFragment` check to see if a fragment with `actionId + uniqueIdentifier` combination already exists. If it exists new fragment is not created.

`trackEvent` method has been changed to use `createEventFragment` and `finalizeEventFragment` to ensure that duplicate events are not captured.

About 1 min delay you are right, I added this to comment.

Even, if service worker dies before event is finalised, event will still be there in store and I think will be finalised when controller revives, @brad-decker can you plz confirm.

jpuri: @brad-decker : a question is `uniqueIdentifier` unique for each fragment, in that case we actually do not need to compare actionId.

brad-decker: it would be unique, except that it doesn't exist on every fragment. so we should still use actionId.

Fragments have two paths, the `success` path fires the `successEvent` when `finalizeFragment` is called without the `abandoned` property set to true. This *only happens from outside the controller*. When timeouts occur, or when the controller restarts, the events are finalized using the `failureEvent` which is meant for things like user rejections, or "abandoned cart" type scenarios.

So it seems we will have event loss if the service worker dies before the fragment is finalized. Honestly, though, I think the solution we discussed for storing events until they are fired to prevent event loss from the other PR to replace analytics-node would be a more ideal way of handling "idempotency" of trackEvent. Instead of using the fragment system we can just fire trackEvent, but add the actionId as a key (perhaps even make it our "messageId" we send to segment) and use that to deduplicate events in the storage mechanism before they are sent to segment.

jpuri: I agree, in fact I would prefer to close this PR and create another using that approach. We just need to insure that we generate a unique messageId and it will take care of idempotency. Here is the PR for segment events https://github.com/MetaMask/metamask-extension/pull/16198

jpuri: I will not see this task a blocker for release though.

jpuri: I started working on alternate approach here: https://github.com/MetaMask/metamask-extension/pull/16267

jpuri:
Closing this PR in favour of alternate approach: https://github.com/MetaMask/metamask-extension/pull/16267

PR #15743, opened 2022-09-06
Updated Token Symbol label's CSS

### LINKED ISSUE #15590: [Bug]: Dark mode displays Token symbol and decimals in black text
### Describe the bug

**Problem**: when I am on Dark Mode and I'm trying to import some custom tokens, I can see that the text `Token symbol` and `Token decimal` turns into black text once I've introduced the contract address.

Text should be white, so we can read it in dark mode.

image

### Steps to reproduce

1. Unlock MM
2. Go to Settings and activate Dark mode
3. Go to Import Tokens
4. Click custom tokens
5. Add token address
6. Check text

https://user-images.githubusercontent.com/54408225/184664557-8fbf2569-d54e-4533-91af-222db553b73b.mp4

_No response_

### Version

10.20.0

### Build type

_No response_

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

rakshit087:
## Explanation

Currently when we are in dark mode and try to import a custom token, when we enter a Token's address, it fetches the Token Symbol and Decimal and mark the input as disabled, however this also makes the text color black.
I have updated the disabled label colors to make it dynamic and update depending upon the dark/light theme. 

## More Information
Fixes #15590 

## Screenshots/Screencaps

Current Progress
image

### Before
You can the color before in the `Current progress` screen shot, in Token Decimal.

### After

You can the color after in the `Current progress` screen shot, in Token Label.

## Pre-Merge Checklist


rakshit087:
I have completed the changes in UI
This is how it looks like right now
image

In dark mode
image

CHANGELOG.md
```diff
@@ -153,7 +153,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix block explorer link on custom networks for the cases when link is invalid or left empty ([#13870](https://github.com/MetaMask/metamask-extension/pull/13870))
 - Fix signature parsing errors re-surfaced due to 4byte function signature directory being down, by removing the directory([#15300](https://github.com/MetaMask/metamask-extension/pull/15300))
 - Fix intermitent failure when performing a Send tx in non-EIP-1559 networks (like Optimism) by setting the `gasPrice` ([#15628](https://github.com/MetaMask/metamask-extension/pull/15628))
-
+- Fix disabled label colors for Token Symbol and Token Decimal while in dark mode in 'Import Token Page'. ([#15590](https://github.com/MetaMask/metamask-extension/pull/15743))
```
brad-decker, line 156: This isn't necessary as we will generate the changelog for the release this gets landed in.  We use a tool that sources details of commits in a release to generate the changelog :)

ui/pages/import-token/import-token.component.js
```diff
@@ -501,7 +501,13 @@ class ImportToken extends Component {
...
           label={
             <div className="import-token__custom-symbol__label-wrapper">
-              <span className="import-token__custom-symbol__label">
+              <span
+                className={
+                  symbolAutoFilled && !forceEditSymbol
+                    ? 'import-token__custom-symbol__label--disabled'
+                    : 'import-token__custom-symbol__label'
+                }
+              >
```
brad-decker, line 510: I believe we want the __label styles always applied, conditionally adding a second class with the --disabled modified. Otherwise this span doesn't get the 'flex: 0 0 auto'; from the plain __label class name.

ui/pages/import-token/import-token.component.js
```diff
@@ -524,7 +530,17 @@ class ImportToken extends Component {
...
+          label={
+            <span
+              className={
+                decimalAutoFilled
+                  ? 'import-token__custom-decimals__label--disabled'
+                  : 'import-token__custom-decimals__label'
+              }
+            >
+              {t('decimal')}
+            </span>
```
brad-decker, line 542: I think i we are going to add the span here instead of using straight text, we should also add the wrapper similar to the other instance. The main role it has is adding a flex box i think.

brad-decker (changes requested):
Thank you for your submission and my apologies for the delay in review. I've added some feedback that should get us pretty close to merging this!

rakshit087:
> Thank you for your submission and my apologies for the delay in review. I've added some feedback that should get us pretty close to merging this!

Thanks I will look into it and I will make the changes as soon as possible.

CHANGELOG.md
rakshit087: Ahh okay! Thanks for letting me know.

ui/pages/import-token/import-token.component.js
rakshit087: Understood, I will update this.

rakshit087:
```
            <span
                className={
                  'import-token__custom-symbol__label ' +
                  (symbolAutoFilled && !forceEditSymbol)
                    ? 'import-token__custom-symbol__label--disabled'
                    : ''
                }
              >
```
does this look good?

rakshit087: I have not done a lot of scss but I have defined it like this,

ui/pages/import-token/import-token.component.js
rakshit087: I understood the wrapper part, but I don't understand the adding span instead of straight text part.

ui/pages/import-token/import-token.component.js
brad-decker: this has some details https://dev.to/mikkel250/getting-started-with-sass-nesting-and-parent-selectors-4617

But the `&` is computed as the parent class's name, so if &__label is nested under `import-token__custom-symbol` then a new css rule is created:
```css
import-token__custom-symbol__label {
   flex: 0 0 auto;
}
```

the --disabled modifier will also get its own styles. In order to have both the element must have both class names. In the example link you'll notice they do this:

```scss
.parent {
  width: 100%;
  &.new-child {
    height: 100%;
  }
}
```
the `&.` part is essential here because when you add the additional '.' it computes to two selectors by class name which still both must be present on to target the class. 

```css
.parent {
  width: 100%;
}

.parent.new-child {
  height: 100%;
}
```

hope that helps

ui/pages/import-token/import-token.component.js
brad-decker: the wrapper part was was the important part, I just meant if we are going to use a react node instead of a string we need to include the wrapper as it was done on the other instance.

ui/pages/import-token/import-token.component.js
rakshit087: Thanks for such a detailed explanation, I will try to fix these asap.

brad-decker:
@rakshit087 how are things going? No rush on this just want to make sure you have everything you need/aren't running into issues with my feedback (I am wrong at least 70% of the time so if you are having problems making things worth with my feedback don't fret, just let me know and we can work through it together)

rakshit087:
> @rakshit087 how are things going? No rush on this just want to make sure you have everything you need/aren't running into issues with my feedback (I am wrong at least 70% of the time so if you are having problems making things worth with my feedback don't fret, just let me know and we can work through it together)

Hey! Thanks for the followup, got a little busy with school work and personal projects, will come back to this asap :3

ui/pages/import-token/import-token.component.js
rakshit087: Added the wrapper as well, let me know if it need further changes

ui/pages/import-token/import-token.component.js
rakshit087: I have made the changes let me know if I need to add anything else.

CHANGELOG.md
```diff
@@ -153,7 +153,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix block explorer link on custom networks for the cases when link is invalid or left empty ([#13870](https://github.com/MetaMask/metamask-extension/pull/13870))
 - Fix signature parsing errors re-surfaced due to 4byte function signature directory being down, by removing the directory([#15300](https://github.com/MetaMask/metamask-extension/pull/15300))
 - Fix intermitent failure when performing a Send tx in non-EIP-1559 networks (like Optimism) by setting the `gasPrice` ([#15628](https://github.com/MetaMask/metamask-extension/pull/15628))
-
+
```
brad-decker, line 156: This needs to be removed in order for the changelog lint to pass

rakshit087: Uh oh! I must have missed it. Just a min.

rakshit087: I need to remove the extra space right? or the entire line?

PR #15749, opened 2022-09-06
Fix storybook by shimming MV3 method

darkwing:
@georgewrmarshall reported that MV3 broke storybook -- this fixes the issue

### DIFF
--- a/.storybook/__mocks__/webextension-polyfill.js
+++ b/.storybook/__mocks__/webextension-polyfill.js
@@ -1,4 +1,8 @@
 module.exports = {
-    runtime: {},
-  };
+  runtime: {
+    getManifest: () => {
+      return { manifest_version: 2 };
+    }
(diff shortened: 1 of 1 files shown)


.storybook/__mocks__/webextension-polyfill.js
```diff
@@ -1,4 +1,8 @@
 module.exports = {
-    runtime: {},
-  };
+  runtime: {
+    getManifest: () => {
+      return { manifest_version: 2 };
+    }
+  },
+};
```
georgewrmarshall, line 8: Linter is saying there is a space here that may need to be removed?

georgewrmarshall (approved):
Yay! Thank you 🙏 
- verified storybook works on local by running `yarn storybook`.

ryanml (approved):

PR #15774, opened 2022-09-09
Show error message if service worker did not load (respond to the UI)

### LINKED ISSUE #15536: Show error to user if service worker is not revived after long period of inactivity
If for some reason service worker does not revives after say 1 min, then we should show message to user with button to reload extension.

cc @naugtur , @danjm , @kevinghim

segun:
## Explanation

We should show a message to the user if the service worker is stuck for more than X seconds (60s at the moment)

## More Information

* Fixes #15536 

## Screenshots/Screencaps

Screenshot 2022-09-12 at 17 52 13

Screenshot 2022-09-12 at 17 50 36

## Manual Testing Steps

You can simulate an error by disabling the code on app/scripts/app-init.js:138

## Pre-Merge Checklist


app/scripts/app-init.js
```diff
@@ -127,6 +130,15 @@ chrome.runtime.onMessage.addListener(() => {
...
 });
 
+// Here respond to all clients with ACK_KEEP_ALIVE_MESSAGE
+chrome.runtime.onMessage.addListener((message) => {
+  if (message.name === WORKER_KEEP_ALIVE_MESSAGE) {
+    // eslint-disable-next-line no-undef
+    const channel = new BroadcastChannel('sw-messages');
+    channel.postMessage({ name: ACK_KEEP_ALIVE_MESSAGE });
+  }
+});
```
jpuri, line 140: We can add this to listener above [here](https://github.com/MetaMask/metamask-extension/pull/15774/files#diff-cd37b653f9862ccada12e440c6ef67e8b2b2c5b906a320870ed71232d244e660R128)

app/scripts/ui.js
```diff
@@ -36,8 +40,36 @@ const WORKER_KEEP_ALIVE_MESSAGE = 'WORKER_KEEP_ALIVE_MESSAGE';
...
+
+    // add listener to receive ACK_KEEP_ALIVE_MESSAGE
+    // eslint-disable-next-line no-undef
+    const channel = new window.BroadcastChannel('sw-messages');
+    channel.addEventListener('message', (event) => {
+      if (event.data.name === ACK_KEEP_ALIVE_MESSAGE) {
+        clearTimeout(timeoutHandle);
+        delete ACK_KEEP_ALIVE_RECEIVED[messageId];
+      }
+    });
```
jpuri, line 72: Do we need a new channel here. We already have message listeners in this file.

app/scripts/ui.js
```diff
@@ -36,8 +40,36 @@ const WORKER_KEEP_ALIVE_MESSAGE = 'WORKER_KEEP_ALIVE_MESSAGE';
...
+      if (ACK_KEEP_ALIVE_RECEIVED[messageId]) {
+        delete ACK_KEEP_ALIVE_RECEIVED[messageId];
+      } else {
+        clearInterval(handle);
+        displayCriticalError(
+          'somethingIsWrong',
+          new Error("Something's gone wrong. Try reloading the page."),
+        );
+      }
+    }, ACK_KEEP_ALIVE_WAIT_TIME);
```
jpuri, line 62: I am not clear about how is it useful to keep track of `messageId` here.

segun: I added it because we are sending the message every second while the interval to check timeout is a minute. It means by the time we get the interval, we need to check the particular message (identified with messageId) so that we don't show error for a message that has not been sent for less than a minute

app/scripts/ui.js
segun: The channel is to respond to the UI. It's only method that worked I tried the below

1. I used the sender argument of addListener callback, message is not going back to the client. One thing I noticed though (not sure if this is the reason). If I get all the clients and print their IDs, there's no ID for the sender argument. Basically I can't match the sender to any of the connected clients.
2. I also used the sendReponse method of the addListener callback, same issue with sender argument. The message is not coming back to the UI
3. The only method that worked is the BroadcastChannel (a MessageChannel would work too, but it's more complex)

app/scripts/app-init.js
segun: I'm not sure if there's any draw back to having a different listener to handle this separate from the one that handles importAllScripts. I had the logic in one handler, but separating them feels cleaner.

app/scripts/ui.js
jpuri: I see, just curious can we not keep a timestamp of when was last message received and see if it exceeds 1 minute.

app/scripts/app-init.js
brad-decker: Is there any limit on listeners? If the limit is either non-existent or high enough to be essentially non-existent, keep them separate.

seaona:
From QA perspective, overall behavior looks good both in Expanded view and popup view. 
I just have one question related to the following behavior:

1. I am in the middle of performing a tx and I see a popup at the bottom about deprecating networks.
2. I close the popup
3. I wait until I land on the error page - reload screen
4. I open again MM
5. **I can see that the popup is displayed again**

Is this the expected behavior? Or should we save the user action and not display the popup anymore (as we usually do, when popup is closed once) ?

[network-popup.webm](https://user-images.githubusercontent.com/54408225/189963782-fff33cfb-9e81-4d8d-91d8-e3a6af8944a9.webm)

segun:
Hi @seaona yeah, I saw that too...it's from another PR and not this one. We can open another ticket to fix that.

app/scripts/ui.js
segun: yeah, we could do that too.

seaona:
Thank you @segun !! I've opened a separate issue [here](https://github.com/MetaMask/metamask-extension/issues/15828).

app/scripts/ui.js
jpuri: That will be better approach I think

uzogoduyah (dismissed):

app/scripts/ui.js
```diff
@@ -26,18 +26,44 @@ import metaRPCClientFactory from './lib/metaRPCClientFactory';
...
+
+    const timeoutHandle = setTimeout(() => {
+      if (
+        Date.now() - lastMessageRecievedTimestamp >
+        ACK_KEEP_ALIVE_WAIT_TIME
+      ) {
+        clearInterval(handle);
+        displayCriticalError(
+          'somethingIsWrong',
+          new Error("Something's gone wrong. Try reloading the page."),
```
jpuri, line 53: Error key like `somethingIsWrong` can be made better and I would prefer uppercase letters.

jpuri (dismissed):
LGTM, some small feedbacks added.

shared/lib/error-utils.js
```diff
@@ -50,26 +58,38 @@ export async function getErrorHtml(supportLink, metamaskState) {
...
+   * of the locale keys. If we use the variable directly, the linter will not
+   * see the string and will not be able to check if the locale key exists.
+   */
   return `
     <div class="critical-error">
-      <div class="critical-error__alert">
-        <p class="critical-error__alert__message">
-          ${t('troubleStarting')}
+      <div class="critical-error__icon">
+        ${WarningIcon()}
```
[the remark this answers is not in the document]
segun: hahaha..force of habit.

app/scripts/ui.js
segun: :( naming things is harder than it seems.

jpuri:
Hey @segun : PR looks good, not sure why it is pending ?

app/scripts/ui.js
```diff
@@ -26,18 +26,44 @@ import metaRPCClientFactory from './lib/metaRPCClientFactory';
...
+        );
+      }
+    }, ACK_KEEP_ALIVE_WAIT_TIME);
+
+    // add listener to receive ACK_KEEP_ALIVE_MESSAGE
+    // eslint-disable-next-line no-undef
+    const channel = new window.BroadcastChannel('sw-messages');
+    channel.addEventListener('message', (event) => {
```
danjm, line 61: This would create a new channel and new event listener every second... is that okay?

app/scripts/ui.js
danjm: perhaps we could use the the portstreams already established between the background and the ui script?

app/scripts/ui.js
segun: Yes, you're right, I have changed the code to initialise the Channel outside the listener (in app-init) and did same for the UI too.

app/scripts/ui.js
segun: At this point, there are no port streams initialised yet. The major issue is I can't find a way to match the client-id else, there's a client.postMessage that could be called within the event handler. But like I said above
```
If I get all the clients and print their IDs, there's no ID for the sender argument. Basically I can't match the sender to any of the connected clients.
```

segun: @danjm @jpuri probably we can re-visit this after [this](https://github.com/MetaMask/metamask-extension/pull/16075) is merged.

In the above linked PR, I see that the keep alive message is only sent after the phishing stream have been initialized

```
 if (isManifestV3) {
     phishingPageStream.on('data', ({ data: { method } }) => {
       if (!IGNORE_INIT_METHODS_FOR_KEEP_ALIVE.includes(method)) {
         runWorkerKeepAliveInterval();
       }
     });
   }
```

app/scripts/background.js
```diff
@@ -439,6 +442,14 @@ function setupController(initState, initLangCode, remoteSourcePort) {
...
         // It fixes the issue of blank screen coming when extension is loaded, the issue is very frequent in MV3
         remotePort.postMessage({ name: 'CONNECTION_READY' });
+
+        // If we get a WORKER_KEEP_ALIVE message, we respond with an ACK
+        remotePort.onMessage.addListener((message) => {
+          if (message.name === WORKER_KEEP_ALIVE_MESSAGE) {
+            // To test un-comment this line and wait for 1 minute. An error should be shown on MetaMask UI.
+            remotePort.postMessage({ name: ACK_KEEP_ALIVE_MESSAGE });
+          }
+        });
```
jpuri, line 452: It is possible to rather use service worker for this message: https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/app-init.js#L125

jpuri (approved):
Changes look good, a small feedback.

app/scripts/background.js
segun: Yup, the initial message uses SW, but at that point, we don't have a way to communicate back to the client (UI) as the streams have not been initialised.

So what I did was use a BroadcastChannel. 

I had a discussion with @danjm and we decided using remotePort/extensionPort would be better as we don't have to use another message channel to communicate with the UI

danjm (approved):

Issue #15267, opened 2022-07-18
[Bug]: Users have to click "down arrow" to enable "Sign" button


cvega-vhs:
### Describe the bug

This may be a reoccurrence or duplicate of #14666: when a user is prompted to sign/approve a transaction with a very long message, the "Sign" button remains disabled unless the "down arrow" button is clicked.

Having to click the down arrow is confusing and not obvious to some users.

**Screenshot**
![image](https://user-images.githubusercontent.com/101071180/179486324-21015c06-dcd1-4ace-b1a3-1bc5c83f37ad.png)

**Video**
https://user-images.githubusercontent.com/101071180/179487306-eeaef312-f6a1-4eff-8c69-58c20d1a39c0.mp4

### Steps to reproduce

1. Trigger the metamask prompt to sign a transaction with a long message (e.g. using the "Signed Typed Data V4" sign button on the [Metamask test dapp](https://metamask.github.io/test-dapp/))
2. Manually scroll down the message section:
    2.1. via the scroll wheel
    2.2. via the arrow keys
    2.3. via using the mouse on the scroll bar on the right, and dragging it down as far as it will go
4. Click the "down arrow" button to finally enable the "Sign" button

### Version

10.17.0

### Browser

Chrome

### Operating system

Windows

### Additional context

The device this error is being experienced on has a display resolution of 2496x1664

dokocat:
I see https://github.com/MetaMask/metamask-extension/pull/14745 fixed this but you still have to scroll to activate the sign button. The sign button should be active by default and the user should scroll if they want to. Its already like this on Metamask mobile, it should be the same on desktop.

bschorchit:
Hey @dokocat, thank you for your input here. We request the user to scroll down in order to encourage them to read the full message before signing. It's unlikely, for our users own security, that we would want to make it easy for a user to agree to something they not have been shown. And it's very likely that mobile behavior will be updated in the near future to match extension's behavior.

Ps: closing the linked PR that fix this has been merged.

addmorebass:
@bschorchit This is not working anymore when a user scrolls down. The sign button does not enable.

This is on Google Chrome.

PR #15788, opened 2022-09-10
fix issue where signing a message sometimes doesn't allow you to scroll to the bottom

tayvano:
done in a way to make it as mergeable as possible

## Explanation

Fixes the longstanding annoying as fucking fuck issue where you would scroll to the bottom but the sign message button would not become activated.

the cause is a combination of how javascript determines the size and scroll of things and the way css determines the size and scroll of things and the units of measurements used in each and the way browsers, operating systems, and user settings (zoom levels) impact and convert one size to another, combined with how rems are converted to pixels, combined with small rounding errors, combined with the fact that the size of things is not clearly defined with sizing things via css's `flexbox` combined with the fact that no one in the history of ever understands how `overflow` works.

but we just avoid all of that nonsense by sidestepping it and removing the border from the footer of the signature view so that there is no variation in how anyone calculates height because no one ever includes a `shadow` in the height of an element but they do sometimes and sometimes not include the `border` in the height of an element.

so i just swapped the border attribute with an inset shadow attribute, thereby fixing the most annoying bug ever while not visually changing how it looks or doing anything that could impact another view by more than 1 pixel. 

## It should be noted

if there are other views impacted by this bug that don't use the .signature-request-footer class, this fix needs to be applied to those views as well. technically though, the fix for that will be to ensure we don't have two separate components/css classes doing the **_exact same thing_**. 🤦‍♀️

## More Information

I'm sure there are literally dozens of cards, issues, sub issues, slack convos and twitter convos about this. idaf, sorry. this fixes one of the most annoying ux papercuts that currently exists and anyone who actually uses metamask knows the exact issue so im not going to waste anyones time pulling them. 

that said, someone please respond to the following tweets and in the #opensea channel when this goes live so that people know we actually care about them and metamask indeed does not have its head so far up its ass that we dont know the annoying things about our product. barbara should know other people who have asked about this who will be happy when this fix is live.

https://twitter.com/trent_vanepps/status/1568659077303238657

https://twitter.com/gremplin/status/1568656560544694273

## Manual Testing Steps

1. go to https://metamask.github.io/test-dapp/
2. click sign under sign typed data v4 and scroll to the bottom
3. the button should turn blue

the hardest thing about testing this is ensuring its fixed bc itll only appear on certain screen sizes and densities and browser zoom levels. the easiest way to ensure this is fixed is by ensuring its broken and then changing nothing and ensuring its no longer broken. for me, only zooming out in metamask broke it but because zoom level is per window, my normal metamask doesnt change my test metamask.

the easiest way to minimize the variables between "ensuring its broken" and "ensuring its not broken" is to load up this fixed version, add the css that was breaking it (border-top), and ensure its broken. Then remove that css you just added and ensure its fixed. For me, that looked like this:

1. go to https://metamask.github.io/test-dapp/ 
2. click sign under sign typed data v4
3. before scrolling to the bottom, right click, on the space between the two buttons and click inspect. The `sign` button should be disabled/light blue
4. under styles on the right of chrome devtools click under the `.signature-request-footer` and paste the following as one chunk and hit enter

```border-top: 1px solid var(--color-border-muted);```

Should look like this:

Screen Shot 2022-09-10 at 1 30 10 PM

this adds the buggy css and bc we just added it on top of the fix, you should be able to see that the top "border" of the button container is visually slightly thicker. 

5. Now, scroll to the bottom of the message in attempt to sign and ensure that the button is still disabled/light blue. If it triggers and becomes clickable, hit ctrl/cmd - (zoom out one level), cancel the metamask pop up, and return to step 2 and repeat. If its still disabled, cancel the metamask pop up, and continue to next step.

6. Click on Sign under Sign Typed Data V4 again to trigger a new MM popup. The border should look normal, not slightly thicker like before.

7. Scroll all the way down. The button should activate so you can sign. If it does, merge this bad boy and tell twitter when it will be live and lfg 🚀

### DIFF
--- a/ui/components/app/signature-request/signature-request-footer/index.scss
+++ b/ui/components/app/signature-request/signature-request-footer/index.scss
@@ -1,6 +1,6 @@
 .signature-request-footer {
   display: flex;
-  border-top: 1px solid var(--color-border-muted);
+  box-shadow: inset 0px 1px 0px var(--color-border-muted);
 
   button {
     text-transform: uppercase;
--- a/ui/components/app/signature-request/signature-request-footer/index.scss
+++ b/ui/components/app/signature-request/signature-request-footer/index.scss
@@ -1,6 +1,6 @@
 .signature-request-footer {
   display: flex;
-  box-shadow: inset 0px 1px 0px var(--color-border-muted);
+  box-shadow: inset 0 1px 0 var(--color-border-muted);
 
   button {
     text-transform: uppercase;


ryanml (approved):
@tayvano thanks for this fix! I tested this a few different ways via different resolutions/zoom levels per your specifications, and all seemed to be well. I did push one fix as there was a linter error. It would be great to get another pair of eyeballs on this or two so we can be certain and get this thing shipped. 👍

ryanml:
Related: https://github.com/MetaMask/metamask-extension/issues/15267

tayvano:
Thanks Ryan for the quick action on this **_and_** especially the linter fix. That's what I get for thinking I couldn't fuck up a single-line change and opening the PR via github.com. 😂 🙈

NidhiKJha (approved):

NidhiKJha:
LGTM! 🚀

PR #15811, opened 2022-09-13
[FLASK] Added hook and action for transaction insight snaps

hmalik88:
Added hook to be used to fetch transaction insight information from a snap, added the pertinent action as well.


GuillaumeRx (approved):
Looks good ! 💯

Issue #12990, opened 2021-12-06
Only resolving ENS on Mainnet? No Unstoppable Domain transactions? Age of Interoperability is here!


luncht1me:
I'm addressing a poor design decision over this issue: https://github.com/MetaMask/metamask-extension/pull/7944

Now that we've entered the age of cross-chain interoperability, and ENS and Unstoppable Domains can resolve to a plethora of different networks (see: 
![image of network fields](https://user-images.githubusercontent.com/3420404/144928770-bdde6d84-7791-48d4-a388-35a521e71855.png)
)

It only seems reasonable that MetaMask should resolve ENS / Unstoppable on pretty much every major L1 and L2. Every day more and more users bridge over from ETH to FTM, Matic, One, Bsc, etc...

My friend just setup an ENS and configured his Fantom FTM address with it. But, while going to test it out, we're just hit with this 
![silly error](https://user-images.githubusercontent.com/3420404/144928933-f452adc1-e06b-41e7-b120-e0f651de57f4.png)

As you can see in the screenshot prior, ENS is  supporting FTM on the Opera chain.

"Network does not support ENS". Well, sorry to say, but the network does. MetaMask has just simply chosen to not resolve unless on ETH MAINNET...

Please, for the sake of the users, strongly reconsider this. It's terribly stifling. Why should I have to go manually resolve an ENS just because I'm on a different network? The network supports it, thus MetaMask should support it on those networks.

My proposal? Remove the choke on only resolving ENS thru MetaMask strictly if only connected to Mainnet, and support Unstoppable Domains.

pietromingo:
I agree

spinlud:
+1

murillojorge:
We need this.

ahh1539:
Also want this feature!

danfinlay:
Related to #8556, but adds Unstoppable domains. I'd like to see the ENS part, I'd like to see users able to opt into other name systems also.

connorbode:
> Related to https://github.com/MetaMask/metamask-extension/issues/8556, but adds Unstoppable domains. I'd like to see the ENS part, I'd like to see users able to opt into other name systems also.

Is support for other naming systems via snaps an option?

MakerGit22:
Hope this gets implemented! So many users + chains would appreciate it!

rsodre:
up!

shocked and frustrated about this!
so I can't test my dapp content redirection on testnet!?
[...]

PR #7258, opened 2019-10-06
Unstoppable domain resolution

JohnnyJumper:
Implemented .zil naming resolution using the namicorn library that supports both .eth and .zil names:
* Replaced existing ens library calls with namicorn
* Renamed classes and methods to be abstract from the naming service used

PR #15817, opened 2022-09-13
Add ud resolution

V-Shadbolt:
## Explanation

Implements Unstoppable Domains Resolution Library
- Resolves UD on send and swap flow
- Resolves UD IPFS websites in address bar

## More Information

Responds to issue #12990

## Manual Testing Steps

[External MetaMask + UD Resolution Testing.pdf](https://github.com/MetaMask/metamask-extension/files/9559866/External.MetaMask.%2B.UD.Resolution.Testing.pdf)

### Testing Address Resolution

1. From the main screen press ‘Send’
2. Type Unstoppable Domain into search bar
3. Check to see if it resolves
4. It should resolve correctly for all UD TLDs
- .crypto
- .blockchain
- .bitcoin
- .zil
- .coin
- .nft 
- .wallet
- .888
- .dao
- .x

It should resolve for different currencies and tokens like MATIC & USDT

It should throw an error if
- The UD is not registered / is not valid OR if there is no address associated with the selected currency 
- The UD does not have an address associate with the selected currency / token
- After selecting a UD and the user swaps to a token that the UD doesn’t have an address for 

It should throw a warning if
- The UD contains a confusing character

### Testing IPFS Website Resolution
1. Simply type the UD into your browser’s address bar with a ‘/’ at the end 
2. On Chrome, Brave or other chrome based browsers it should
- Resolve all UD TLDs EXCEPT .888
- Redirect to IPFS website if a site exists
- Redirect to UD search page if IPFS website does not exist
3. On Firefox it should
- Resolve all UD TLDs
- Redirect to UD search page if IPFS website does not exist

## Pre-Merge Checklist


bbehrman10:

brad-decker:
Hey @bbehrman10 and @V-Shadbolt -- thanks for the submission. Looks like a lot of test failures are valid. I'm going to change this to draft until you have tests passing and ready for review. Feel free to pull it out of draft when ready

bbehrman10:
Screen Shot 2022-09-14 at 8 58 33 PM
Screen Shot 2022-09-14 at 8 58 40 PM

bbehrman10:
@brad-decker any thoughts on the two screenshots above? Not sure why the e2e tests are timing out, we didn't touch any of that stuff

app/scripts/metamask-controller.js
```diff
@@ -1743,6 +1743,7 @@ export default class MetamaskController extends EventEmitter {
         ),
       setFirstTimeUsedNetwork:
         appStateController.setFirstTimeUsedNetwork.bind(appStateController),
+
```
brad-decker, line 1746: This change seems unccessary

ui/ducks/send/send.js
```diff
@@ -1354,14 +1358,25 @@ const slice = createSlice({
...
           } = action.payload;
-
+          let validUDDomain = false;
+          try {
+            udTlds.payload.forEach((tld) => {
+              if (state.recipientInput.toLowerCase().endsWith(`.${tld}`)) {
+                validUDDomain = true;
+              }
+            });
+          } catch {
```
brad-decker, line 1370: I don't believe this code will throw an error? perhaps if state.recipientInput is undefined but in no other cases that I can think of? Maybe just a optional chaining call if its possible that state.recipientInput may be undefined in this case and drop the try/catch?

ui/ducks/send/send.js
```diff
@@ -1354,14 +1358,25 @@ const slice = createSlice({
...
+            validUDDomain = false;
+          }
           if (
             isBurnAddress(state.recipientInput) ||
             (!isValidHexAddress(state.recipientInput, {
               mixedCaseUseChecksum: true,
             }) &&
-              !isValidDomainName(state.recipientInput))
+              !isValidENSDomainName(state.recipientInput) &&
+              !validUDDomain)
```
brad-decker, line 1379: For legibility reasons could we add a new intermediate variable before this if block: `isValidDomain` which checks if it is either a valid ENS or a valid UNS domain, then we can simplify this check to isValidDomain === false

ui/ducks/send/send.js
```diff
@@ -1908,20 +1923,35 @@ export function updateRecipientUserInput(userInput) {
...
+        debouncedValidateRecipientUserInput(
+          dispatch,
+          {
+            userInput,
+            chainId,
+            tokens,
+            useTokenDetection,
+            tokenAddressList,
+            isProbablyAnAssetContract,
+            udTlds: state.UNS.tlds,
```
brad-decker, line 1937: I assume this try/catch is because state.UNS.tlds could be undefined? this can be simplified to:

state.UNS?.tlds, which will then just pass 'udTlds' to undefined and avoids the try/catch.

ui/ducks/send/send.js
```diff
@@ -2011,6 +2041,53 @@ export function updateSendAsset(
...
+          state.UNS.resolution ===
+            state.send.draftTransactions[state.send.currentTransactionUUID]
+              .recipient.address &&
+          state.UNS.domainName
+        ) {
+          let unsError = null;
+          const object = await swapToken(
+            state.UNS.domainName,
+            state.metamask.provider?.ticker ?? 'ETH',
+          );
```
brad-decker, line 2055: object is not an acceptable variable name, this needs to be renamed to reflect what this is.

brad-decker: In addition, swapToken method name needs to be renamed, as it too closely aligns with MetaMask swaps. My first thought reading this was that it came from the swaps code not the UNS code.

ui/ducks/send/send.js
```diff
@@ -2011,6 +2041,53 @@ export function updateSendAsset(
           } with symbol ${state.metamask.provider?.ticker ?? ETH}`,
         ),
       );
+      if (state.UNS) {
```
brad-decker, line 2044: Can you please describe what this section of code is doing? From my perspective it appears as though it is changing the asset based on the domain name resolved by UNS, is that accurate? It also appears to be changing the recipient?

ui/ducks/send/send.js
```diff
@@ -2115,6 +2191,38 @@ export function updateSendAsset(
...
+            dispatch(unsLookup(object));
+            await dispatch(
+              updateRecipient({
+                address: object.address,
+                nickname: object.unsName,
+              }),
+            );
+          }
+        }
+      }
```
brad-decker, line 2225: This code is duplicated can you help me understand what this portion of the code is doing?

ui/helpers/utils/util.js
```diff
@@ -87,17 +86,77 @@ export function addressSummary(
...
+        resultObject.multiChain.push(crypto.split('.')[1]);
+      } else {
+        resultObject.singleChain.push(crypto.split('.')[1]);
+      }
+    });
+    return resultObject;
+  } catch {
+    return resultObject;
+  }
+}
```
brad-decker, line 149: These two methods are adding centralized services to our code, and exposing our user's ip addresses to unstoppable domains. Is there any alternative to http requests to retrieve this info? @Gudahtt can I get your take on this as well?

brad-decker (changes requested):
Before any further review will happen:

1. Please go through and update variable names like 'object' to something of meaning for reviewers and future maintainers of this code. Naming things is hard, but it is absolutely vital for legibility and understanding code and it is impossible to review when combined with the second bullet point:
2. Add jsdocs to every method you are introducing that describe what it does, and also any inline docs that help to explain your intent with the changes being made. 
3. Analyze my request to remove web requests for viable TLDs for unstoppable domains, as well as keys. 

Thank you for the contribution and I hope we can work through these changes and find a way to get this landed.

ui/ducks/send/send.js
bbehrman10: we had this twice because when you swap a token it can either be a native token or a multi chain token

ui/ducks/send/send.js
bbehrman10: when you swap a token after selecting an address we need to pull the correct crypto address associated with the Unstoppable Domain. Because you can have multiple crypto addresses associated with one of our Domains we need to pull the correct one depending on which token you swap to

ui/ducks/send/send.js
bbehrman10: absolutely agree, we will make those changes

ui/ducks/send/send.js
bbehrman10: we were getting an error on the e2e tests and then the try catches seemed to resolve them, but we can investigate

ui/helpers/utils/util.js
bbehrman10: As of now, our only alternative to this would be to hardcode constants for the TLDs and Supported Currencies - so each time those change, an update to the codebase would be required

bbehrman10: if it helps, all of the data contained within the API is open source which is less centralized

Gtonizuka:
Possible duplicate of [#7258](https://github.com/MetaMask/metamask-extension/pull/7258)

jim-unstoppable:
> Possible duplicate of [#7258](https://github.com/MetaMask/metamask-extension/pull/7258)

This is an updated implementation reflecting support for our most recent registry inclusive of Polygon based NFT Domains. This version uses our most recent and established libraries, so I think the old one is no longer applicable

bbehrman10:
@brad-decker - We can submit a change and move the data from the endpoint to a constants file. If that's okay with you we can do that, and in the coming weeks / months if there needs to be an update to those constants (such as if UD releases more token support) just submit a PR with only changes to that constants file. In this way the integration can stay up to date without the need of that REST endpoint. How does that sound? Alternatively we're also comfortable if you are to keep that open source endpoint to leave it all server side.

brad-decker:
Open source does not mean decentralized, and does not mitigate the risk of exposing our user's ip addresses to UD. Our preference would be for your team to create a npm module that contains the constants necessary, linked to an open source repository that can be validated by our team to contain neutral code/constants/etc. Then your team may update those constants and publish new versions which can then be upgraded in our codebase and a reviewer can validate the changes.

Just so you're aware, once the requested changes have been made i'll be doing further review of your integration into the send slice. I would leave additional inline comments everywhere where they may be some context needed for proper understanding of what the code is doing.

We audited the ENS integration and no centralized services are hit to perform the reverse lookup or resolutions. We would expect UD to meet the same standards to be included in the code.

bbehrman10:
@brad-decker quick question. Our unit tests we wrote are passing locally but when running as part of the suite on CircleCI we are getting an 'ENETUNREACH' - have you experienced this before? How could we alter our tests to prevent this from occuring

brad-decker:
@bbehrman10 unit tests should not hit the network, you can't test network integrations with unit tests. You'll need to mock out those network requests. The easiest way to do that, if the network request happens in the background, is to mock the background action call such that the code that makes the network request is never reached. If the network requests happen in the UI you'll need to use nock or similar to replace those network calls. You can do nock in the background as well if you'd prefer the background call to occur naturally in tests...

ultimately though if you replace those network calls with flat files from a npm module as discussed above you should be good. Other than retrieving TLDS and keys what other network calls are made that aren't to the ethereum client?

bbehrman10:
@brad-decker gotcha that makes sense. Okay we will remove those network calls altogether. Thanks!

bbehrman10:
@brad-decker hey! We are ready for you to take another look. We have moved the TLDs and Resolver Keys to this NPM package: https://www.npmjs.com/package/@unstoppabledomains/tldsresolverkeys - Docs here: https://github.com/unstoppabledomains/tlds-and-resolver-keys-package. We have removed all network calls aside from the same calls to infura that metamask is making.

brad-decker:
@bbehrman10 I think it is best for us to split up some of this work. I think a good first PR would be renaming all of the ENS specific selectors, state keys, components to use the generic term domain like you have done in this PR. We can get that landed and then layer on some of this work. I'm going to still do a full review of this pr but I know I'm going to ask for that, so if you wanted to get a jump start I think it would be worthwhile.

bbehrman10:
@brad-decker sounds good! If you can get us those other comments ASAP that would be great.

bbehrman10:
@brad-decker - So we combined the resolution modules for ENS and UNS, and put them into a single state slice called domainState. Something interesting is going on with the testing. So when we run the unit tests they all do pass, but the actual testing process timeouts before it prints out the summary. We are also passing both chrome and firefox e2e tests. You'll see these behaviors on the circleCI tests as well. Should we disregard this timeout issue? We've pushed these changes for when you're ready to take a look. Thanks so much!

brad-decker:
@bbehrman10 what was the status of splitting off some of this work to make the review less complex?

bbehrman10:
@brad-decker - we made a new PR for the name changes requested: https://github.com/MetaMask/metamask-extension/pull/16166

brad-decker:
@bbehrman10 can you rebase this for me today and by end of day i'll make a suggestion for the next slice to remove out for landing this in chunks. I know that seems like a longer process but ultimately its the only way we are going to get this landed in an auditable fashion.

brad-decker:
Lets save anything to do with token resolution for the end, we don't currently do that at all and those changes to the send slice were the most alarming to me when reviewing this PR. The easiest to land would be the IPFS domain resolution, followed shortly thereafter by the recipient resolution that we already do. Once those two land you are at feature parity with what we currently support for ENS.

bbehrman10:
sounds good, we will link to the new PR when it's ready for you to take a look at. Thanks Brad!

bbehrman10:
@brad-decker we're trying to add the UD resolution package via yarn, but when running yarn dist we get a lavamoat error. We went through the process outlined in the github (setup, lavamoat:auto, etc), but still get the same package not in allowlist error. Are there new steps in order to add dependencies?

bbehrman10:
Isn't that what the lavamoat commands are for? Rewriting those policy files so one doesn't have to manually put those keys in?

brad-decker:
There are two files "policy.json" and "policy-overrides.json" lavamoat:auto will build a policy file, so don't modify the 'policy.json' file. Not all dependency usages can be statically analyzed, and require a manual process. See https://github.com/LavaMoat/LavaMoat/blob/main/docs/policy.md#using-policy-overridejson-to-expand-or-limit-access

bbehrman10:
@brad-decker is there a day or time that works well for you?

bbehrman10:
IPFS Pull Request Ready for Review: https://github.com/MetaMask/metamask-extension/pull/16664

brad-decker:
@bbehrman10 could you push up your changes that are not passing lavamoat:auto ? its pretty hard to be helpful when looking at this PR because its so out of date.

PR #15830, opened 2022-09-14
[FLASK] Add changes to support blocking Snaps by source shasum

### LINKED ISSUE #620: Add `shasum` blocking capability to the blocklist
We should add a way to block snaps by shasum to the existing blocklist implementation.

david0xd:
## Explanation
This PR will add some small changes to the `MetamaskController` `->` `SnapController` which would make possible to block Snaps by shasum of its source code. This PR extends the initial version of Snaps Blocklist functionality.

## More Information
Fix: https://github.com/MetaMask/snaps-skunkworks/issues/620

Required merge of: https://github.com/MetaMask/snaps-skunkworks/pull/767 (Update version of `@metamask/snap-controllers`)

* Snaps blocklist is extracted and moved from MetaMask Controller to separate directory and file `app/scripts//flask/snaps-blocklist.js`, for easier maintenance, import and testing.
* Function `checkBlockList` is now handling blocked snap by using utility function `checkSnapsBlockList` instead of having all the logic inline. Utility function is extracted to `app/scripts/flask/snaps-utilities.js`.
* Unit tests are added to ensure functionality of the previous and new features.

## Screenshots/Screencaps
No UI changes. Snaps blocklist structure now can look like the example below and can handle the cases specified in this PR.
```javascript
export const SNAP_BLOCKLIST = [
  {
    id: 'npm:@consensys/starknet-snap',
    versionRange: '<0.1.11',
  },
  {
    // @consensys/starknet-snap v:0.1.10
    shasum: 'A83r5/ZIcKuKwuAnQHHByVFCuofj7jGK5hOStmHY6A0=',
  },
];
```

### Before
Snaps blocklist supported only blocking snap by its `ID:VERSION`.

### After
Snaps blocklist supports blocking snaps by `ID:VERSION` plus `SHASUM` of the source code.

## Pre-Merge Checklist

- [x] Snaps Controller package is updated

### DIFF
--- a/app/scripts/flask/snaps-blocklist.js
+++ b/app/scripts/flask/snaps-blocklist.js
@@ -1,0 +1,24 @@
+/**
+ * Represents a list of Snaps that are not allowed to be used.
+ * Can be blocked by [ID, VERSION] or SHASUM of a source code (or both).
+ *
+ * Example:
+ * {
+ *    id: 'npm:@consensys/snap-id',
+ *    versionRange: '<0.1.11',
+ *    shasum: 'TEIbWsAyQe/8rBNXOHx3bOP9YF61PIPP/YHeokLchJE=',
+ * },
+ * {
+ *    shasum: 'eCYGZiYvZ3/uxkKI3npfl79kTQXS/5iD9ojsBS4A3rI=',
+ * },
+ */
+export const SNAP_BLOCKLIST = [
+  {
+    id: 'npm:@consensys/starknet-snap',
+    versionRange: '<0.1.11',
+  },
+  {
+    // @consensys/starknet-snap v:0.1.10
+    shasum: 'A83r5/ZIcKuKwuAnQHHByVFCuofj7jGK5hOStmHY6A0=',
+  },
+];
--- a/app/scripts/flask/snaps-utilities.js
+++ b/app/scripts/flask/snaps-utilities.js
@@ -1,0 +1,35 @@
+import { satisfies as satisfiesSemver } from 'semver';
+
+/**
+ * Checks if provided snaps are on the block list.
+ *
+ * @param snapsToCheck - An object containing snap ids and other information.
+ * @param blocklist - An object containing snap ids, version or shasum of the blocked snaps.
+ * @returns An object structure containing snaps block information.
+ */
+async function checkSnapsBlockList(snapsToCheck, blocklist) {
+  return Object.entries(snapsToCheck).reduce((acc, [snapId, snapInfo]) => {
+    const blockInfo = blocklist.find(
+      (blocked) =>
+        (blocked.id === snapId &&
+          satisfiesSemver(snapInfo.version, blocked.versionRange, {
+            includePrerelease: true,
+          })) ||
+        // Check for null/undefined for a case in which SnapController did not return
+        // a valid message. This will avoid blocking all snaps in the given case.
+        // Avoid having (undefined === undefined).
+        (blocked.shasum ? blocked.shasum === snapInfo.shasum : false),
+    );
+
+    acc[snapId] = blockInfo
+      ? {
+          blocked: true,
+          reason: blockInfo.reason,
+          infoUrl: blockInfo.infoUrl,
+        }
+      : { blocked: false };
+    return acc;
+  }, {});
+}
+
+export { checkSnapsBlockList };
--- a/app/scripts/flask/snaps-utilities.test.js
+++ b/app/scripts/flask/snaps-utilities.test.js
@@ -1,0 +1,127 @@
+import { strict as assert } from 'assert';
+import { checkSnapsBlockList } from './snaps-utilities';
+
+describe('Snaps Controller utilities', function () {
+  describe('checkSnapsBlockList', function () {
+    it('returns one of the given snaps as blocked by its version', async function () {
+      const mockBlocklist = [
+        {
+          id: 'npm:@consensys/starknet-snap',
+          versionRange: '<0.1.11',
+        },
+      ];
+      const mockSnapsToBeChecked = {
+        'npm:exampleA': {
+          version: '1.0.0',
+          shasum: 'F5IapP6v1Bp7bl16NkCszfOhtVSZAm362X5zl7wgMhI=',
+        },
+        'npm:exampleB': {
+          version: '1.0.0',
+          shasum: 'eCYGZiYvZ3/uxkKI3npfl79kTQXS/5iD9ojsBS4A3rI=',
+        },
+        'npm:@consensys/starknet-snap': {
+          version: '0.1.10',
+          shasum: 'A83r5/ZIcKuKwuAnQHHByVFCuofj7jGK5hOStmHY6A0=',
+        },
+      };
+
+      const blockedSnaps = await checkSnapsBlockList(
+        mockSnapsToBeChecked,
+        mockBlocklist,
+      );
+      assert.deepEqual(blockedSnaps, {
+        'npm:exampleA': { blocked: false },
+        'npm:exampleB': { blocked: false },
+        'npm:@consensys/starknet-snap': {
+          blocked: true,
+          reason: undefined,
+          infoUrl: undefined,
+        },
+      });
+    });
+
+    it('returns given snap as blocked by its shasum', async function () {
+      const mockBlocklist = [
+        {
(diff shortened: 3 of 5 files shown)


app/scripts/controllers/utils/utils.js
```diff
@@ -0,0 +1,34 @@
+///: BEGIN:ONLY_INCLUDE_IN(flask)
```
ritave, line 1: `utils/utils.js` is not a really descriptive file

app/scripts/controllers/utils/utils.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+async function checkSnapsBlockList(snapsToCheck) {
+  return Object.entries(snapsToCheck).reduce((acc, [snapId, snapInfo]) => {
+    const blockInfo = SNAP_BLOCKLIST.find(
+      (blocked) =>
+        (blocked.id === snapId &&
+          satisfiesSemver(snapInfo.version, blocked.versionRange, {
+            includePrerelease: true,
+          })) ||
+        (blocked.shasum ? blocked.shasum === snapInfo.shasum : false),
```
ritave, line 20: This could probably be just `blocked.shashum === snapInfo.shasum`. If `blocked.shasum` is `undefined`, it'll just return false. And `snapInfo.shasum` is always defined.

app/scripts/controllers/utils/utils.test.js
```diff
@@ -0,0 +1,92 @@
+import { strict as assert } from 'assert';
+import { checkSnapsBlockList } from './utils';
+
+describe('Controller utilities', function () {
```
ritave, line 4: I believe we use arrow functions `() => {}` in tests instead of `function () {}` everywhere else.

app/scripts/controllers/utils/utils.test.js
```diff
@@ -0,0 +1,92 @@
...
+          shasum: 'eCYGZiYvZ3/uxkKI3npfl79kTQXS/5iD9ojsBS4A3rI=',
+        },
+        'npm:@consensys/starknet-snap': {
+          version: '0.1.10',
+          shasum: 'A83r5/ZIcKuKwuAnQHHByVFCuofj7jGK5hOStmHY6A0=',
+        },
+      };
+
+      const blockedSnaps = await checkSnapsBlockList(mockSnapsToBeChecked);
+      assert.deepEqual(blockedSnaps, {
```
ritave, line 23: `assert` is node-specific while `expect` is jest specific. I haven't used that combination but error reporting might suffer due to lack of compatibility.

app/scripts/controllers/utils/utils.test.js
```diff
@@ -0,0 +1,92 @@
+import { strict as assert } from 'assert';
```
ritave, line 1: The tests are depend on the actual production block list, which may change for reasons.
Please use a mock `SNAP_BLOCKLIST`.

Example:
```javascript
function makeCheckBlockList(blockList) {
  return async function checkSnapsBlockList(snapsToCheck() {
    // ...
  }
}
```

app/scripts/controllers/utils/utils.test.js
```diff
@@ -0,0 +1,92 @@
...
+          reason: undefined,
+          infoUrl: undefined,
+        },
+      });
+    });
+
+    it('returns given snap as blocked by its shasum', async function () {
+      const mockSnapsToBeChecked = {
+        'npm:@consensys/starknet-snap': {
+          version: '0.3.15', // try to fake version with the same source sha
```
ritave, line 37: This now might be checked both by the version and the shasum right?
Using a mock `SNAP_BLOCKLIST` would ensure we only check one variant at a time.

ritave (changes requested):
The extension filters out any `**/flask/**` folders when building. Please put all new Flask files somewhere inside such folder so they aren't bundled in normal extension.

Also, per Mark's request, all PRs that concern Flask should have a `[FLASK]` in the title.

app/scripts/controllers/utils/utils.js
david0xd: I had some intentions here with that approach.

If `blocked.shashum` is undefined because it's omitted from the `BLOCKLIST` object and for any reason `SnapController` returns `undefined` for `shasum` it will end up with checking `if undefined === undefined` which will be `true` and that means that any snap will be blocked. In the case where every time `shasum` is undefined (because maybe SnapController is broken because of a bad change) **every** possible snap will be **blocked**.

So, I was trying to be protective on this place just in case. Because, `extension` and `SnapsController` are two different actors communicating over messaging system where message should/can be trusted or not... If this was working together with functions and protected by types I would probably think differently.

Anyway, I can change it to simple comparison without check for `undefined`, if you think that's better, but just need to explain potential problems or consequences.

app/scripts/controllers/utils/utils.test.js
david0xd: Yea, we use them everywhere else in snaps repos. But, this one complains if I do that.

```diff
- ESLint: Do not pass arrow functions to describe()(mocha/no-mocha-arrows)
```

app/scripts/controllers/utils/utils.test.js
david0xd: I wanted and tried to use the `expect` but it seems that `assert`s are everywhere in the unit tests in this part of a project `/app/scripts/*`, etc. So, I'm not sure if I should introduce using `expect` here and if it will affect the test run 🤔. My editor also complains a bit if I do so: `ESLint: 'expect' is not defined.(no-undef)`, it seems like it's not configured, but it works. Controller tests use `assert` everywhere so I was following that: http://github.com/MetaMask/metamask-extension/blob/400009127bbd0c96400f60217a3b1f47035398f2/app/scripts/metamask-controller.test.js#L1230-L1230

Still, not sure what would be right in this case 🤔

app/scripts/controllers/utils/utils.js
ritave: Defensive programming is good for security parts. Seems like a good reason - could you put  a small comment explaining it in code?

app/scripts/controllers/utils/utils.test.js
ritave: Roger that

app/scripts/controllers/utils/utils.test.js
ritave: Following the standard practices in that repo should be good.

app/scripts/controllers/utils/utils.js
david0xd: Changed this to `app/scripts/flask/snaps-utilities.js`. Let me know if that's alright.

app/scripts/controllers/utils/utils.js
david0xd: Added some comments to explain this case.

app/scripts/controllers/utils/utils.js
```diff
@@ -0,0 +1,34 @@
...
+    );
+
+    const cur = blockInfo
+      ? {
+          blocked: true,
+          reason: blockInfo.reason,
+          infoUrl: blockInfo.infoUrl,
+        }
+      : { blocked: false };
+    return { ...acc, [snapId]: cur };
```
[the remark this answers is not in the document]
david0xd: Ok, good, updated this part with your suggestion.

app/scripts/controllers/utils/utils.test.js
david0xd: I've did couple of changes to achieve having this using mocked version properly. Now blocklist is mocked for each test.

app/scripts/controllers/utils/utils.test.js
david0xd: Yes, it now handles both (version and shasum). Updated the tests with mocked version of blocklist.

ritave (dismissed):

ritave (dismissed):

brad-decker (approved):

ritave (approved):

brad-decker:
after E2E finishnes i'll restart the CI process at the failed step so we don't have to wait for E2E again

david0xd:
> after E2E finishnes i'll restart the CI process at the failed step so we don't have to wait for E2E again

Thanks! This pipeline was green, but after last two rebase started failing on unrelated `prep-build-storybook` job.

PR #15833, opened 2022-09-14
Update MV3 CSP to fix build

darkwing:
Fixes present issue on MV3 build:

BrokenMV3

### DIFF
--- a/app/manifest/v3/chrome.json
+++ b/app/manifest/v3/chrome.json
@@ -1,6 +1,6 @@
 {
   "content_security_policy": {
-    "extension_pages": "default-src 'self'; frame-ancestors 'none'"
+    "extension_pages": "script-src 'self'; object-src 'self';"
   },
   "externally_connectable": {
     "matches": ["https://metamask.io/*"],
--- a/app/manifest/v3/chrome.json
+++ b/app/manifest/v3/chrome.json
@@ -1,6 +1,6 @@
 {
   "content_security_policy": {
-    "extension_pages": "script-src 'self'; object-src 'self';"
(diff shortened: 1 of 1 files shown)


app/manifest/v3/chrome.json
```diff
@@ -1,6 +1,6 @@
 {
   "content_security_policy": {
-    "extension_pages": "default-src 'self'; frame-ancestors 'none'"
+    "extension_pages": "script-src 'self'; object-src 'self';"
```
brad-decker, line 3: @Gudahtt can you verify that this change is inline with recent changes you made?

digiwand, line 3: I think we will want to keep `frame-ancestors 'none'`

Gudahtt: Exactly, was about to leave the same comment

Gudahtt: Now that `frame-ancestors` has been restored, yes. `default-src` was only added to prevent a _different_ crash. Changing it to `script-src 'self'; object-src 'self';` is fine, it also prevents the same crash while fixing the one described above.

For context, we found that `script-src 'self'; object-src 'self';` is [the default CSP used by Chromium for MV3 extensions](https://source.chromium.org/chromium/chromium/src/+/main:extensions/common/manifest_handlers/csp_info.cc;l=42;drc=9e249731c0b1aab0e53fb776b9c11e673e721c45?q=default_isolated_world_csp&ss=chromium).

Gudahtt (approved):
LGTM!

brad-decker (approved):

digiwand (approved):
Thanks for the fix! 🙏

PR #15835, opened 2022-09-14
Displaying the aggregators from tokenList in Detected token popover

NiranjanaBinoy:
When new tokens are detected, they will get added to the `detectedTokens` prop in the state; currently, we are saving `address, symbol, decimal, and aggregators` to the prop. In the current implementation, the aggregators displayed in the detected token popover are from the `detectedTokens` from the state, causing an inconsistency while switching between static and dynamic lists.
In this PR, we are fetching the aggregators from tokenList itself instead of detectTokens.


danjm (approved):
LGTM (QA'd and reviewed code)

seaona:
from QA looks good @NiranjanaBinoy

darkwing (approved):

PR #15846, opened 2022-09-15
Add notification about the merge

### LINKED ISSUE #15693: Create new "What's new" message about the merge
We should create a "What's new" message about the merge. Something like:

"🐼 **The Ethereum Merge is here!** 
This is an exciting moment and we're here to say that nothing is changing on MetaMask and you can continue to use your wallet normally before and after the Merge. No action will be required from your side.

For test network users, just a heads up that we'll be removing the following test networks from our default list: Rinkeby, Ropsten and Kovan and adding Sepolia to that list soon! If you want to continue to use any of those networks, you'll be able to do so by adding them as a custom network. 
For all other users, it's always great to remember that test network tokens have no value.
Stay safe and happy Merge!"

Copy of course should be reviewed and changed by @coreyjanssen and marketing team.

danjm:
Fixes https://github.com/MetaMask/metamask-extension/issues/15693

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -2360,6 +2360,13 @@
   "notifications14Title": {
     "message": "3Box Deprecation"
   },
+  "notifications15Description": {
+    "message": "There's no action required from you, so keep using your wallet as usual. Be aware of potential scams around the Merge.",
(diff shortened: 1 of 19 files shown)


NiranjanaBinoy (approved):
LGTM

adonesky1 (approved):

PR #15848, opened 2022-09-15
Making permission and approval controller methods idempotent

### LINKED ISSUE #15714: Making permission / approval controller actions idempotent
Actions to be made idempotent:

1. permissionController.revokePermissions
2. permissionController.rejectPermissionsRequest
3. permissionController.acceptPermissionsRequest
4. approvalController.accept
5. approvalController.reject

jpuri:
Fixes: #15714

Make these methods idempotent by adding exception handling in MetamaskController:

- permissionController.revokePermissions
- permissionController.rejectPermissionsRequest
- permissionController.acceptPermissionsRequest
- approvalController.accept
- approvalController.reject

The methods above can throw error if they see same request again, for instance of we try to revoker a permission by calling `permissionController.revokePermissions` and the permission is already revoked it will throw error.

Catching these error like done in this PR will help making the methods behaviour idempotent when called multiple times from UI.


digiwand (dismissed):

jpuri:
The PR should be merged only after this change is deployed: https://github.com/MetaMask/controllers/pull/909

segun (dismissed):

jpuri:
PR can not be merged as there are issues in latest controller build. It will be ready for merge once controller code is fixed.

digiwand (dismissed):

app/scripts/metamask-controller.js
```diff
@@ -4365,4 +4349,54 @@ export default class MetamaskController extends EventEmitter {
...
+    } catch (exp) {
+      if (!(exp instanceof ApprovalRequestNotFoundError)) {
+        throw exp;
+      }
+    }
+  };
+
+  rejectPendingApproval = (id, value) => {
+    try {
+      this.approvalController.reject(id, value);
```
digiwand, line 4395: Based on the [ApprovalController](https://github.com/MetaMask/controllers/blob/5f9f5cee391f64e523ee4b1547ce29499b6f5385/src/approval/ApprovalController.ts), I think we'd want to change `value` here to `error`. 

Do we want to keep the same error format from before? 
``` 
new EthereumRpcError(error.code, error.message, error.data)
```
else, why was this removed?

jpuri: I renamed the variable.

jpuri: Fixed

digiwand (approved):
LGTM!

darkwing (approved):

PR #15858, opened 2022-09-16
Make environment variables higher precedence than config

Gudahtt:
Environment variables are now considered as higher-precedence than configuration by our build system. This means that if the same value is set in `.metamaskrc` and in an environment variable, the environment variable is what will be used. Previously the reverse was true, the configuration would take precedence.

It is conventional for CLI tools to consider environment variables as higher precedence than configuration. This makes our build system less surprising for most people.

## Manual Testing Steps

* Set the entry `PORTFOLIO_URL` in `.metamaskrc` to the value `https://metamask.io`
* Create a build, and ensure the portfolio URL links to `https://metamask.io`
* Create a build with the environment variable `PORTFOLIO_URL` set to `https://google.com`, and ensure the portfolio URL links to `https://google.com` now instead.
  * e.g. `PORTFOLIO_URL='https://metamask.io' yarn start`
* Remove the `PORTFOLIO_URL` entry from `.metamaskrc` and repeat the last step.

You could also try this with a production build (i.e. `yarn build prod`), but you will also need to ensure every single environment variable used by a prod build is set in the production configuration file as well (`.metamaskprodrc`).

## Pre-Merge Checklist

### DIFF
--- a/app/scripts/lib/setupSentry.js
+++ b/app/scripts/lib/setupSentry.js
@@ -9,7 +9,9 @@ import extractEthjsErrorMessage from './extractEthjsErrorMessage';
 // Destructuring breaks the inlining of the environment variables
 const METAMASK_DEBUG = process.env.METAMASK_DEBUG;
 const METAMASK_ENVIRONMENT = process.env.METAMASK_ENVIRONMENT;
-const SENTRY_DSN_DEV = process.env.SENTRY_DSN_DEV;
+const SENTRY_DSN_DEV =
+  process.env.SENTRY_DSN_DEV ||
+  'https://f59f3dd640d2429d9d0e2445a87ea8e1@sentry.io/273496';
 const METAMASK_BUILD_TYPE = process.env.METAMASK_BUILD_TYPE;
 const IN_TEST = process.env.IN_TEST;
 /* eslint-enable prefer-destructuring */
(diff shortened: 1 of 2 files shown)


darkwing (approved):

app/scripts/lib/setupSentry.js
```diff
@@ -9,7 +9,9 @@ import extractEthjsErrorMessage from './extractEthjsErrorMessage';
 // Destructuring breaks the inlining of the environment variables
 const METAMASK_DEBUG = process.env.METAMASK_DEBUG;
 const METAMASK_ENVIRONMENT = process.env.METAMASK_ENVIRONMENT;
-const SENTRY_DSN_DEV = process.env.SENTRY_DSN_DEV;
+const SENTRY_DSN_DEV =
+  process.env.SENTRY_DSN_DEV ||
+  'https://f59f3dd640d2429d9d0e2445a87ea8e1@sentry.io/273496';
```
Gudahtt, line 14: This default was moved into the application itself, because that's what we do already for all of our other defaults. Seemed easier than accommodating this one special case in the build system.

Gudahtt:
I forgot to mention: what makes the old behaviour especially confusing is that even _empty_ config entries in `.metamaskrc` would override environment variables. So having an empty `PORTFOLIO_URL=` in `.metamaskrc` would make the build script completely ignore that environment variable. This is why I found this so confusing recently.

jpuri (approved):
LGTM

PR #15877, opened 2022-09-19
Don't display DetectedTokenSelectionPopover if there are no detected tokens

danjm:
Fixes https://github.com/MetaMask/metamask-extension/issues/15843


ryanml (dismissed):
Change makes sense to me, but should defer to additional reviewers who were able to repro, cc: @NiranjanaBinoy

ryanml (approved):

adonesky1 (approved):

PR #15878, opened 2022-09-19
Bump keystone package versions to latest

adonesky1:
The keystone team graciously upgrade the version of ethereumjs-util that they use in [bc-ur-registry-eth](https://www.npmjs.com/package/@keystonehq/bc-ur-registry-eth) and metamask-airgapped-keyring, [per our request](https://consensys.slack.com/archives/C029LTNQ56C/p1663027917069789).

Since this upgrade bumps us several minor versions we should do a thorough test of the Keystone integration as part of the QA for this PR.

stripHexPrefix has been extracted so that we can use the version that was used in ethereumjs-util v7.1.0 - which comes from [strip-hex-prefix](https://npmfs.com/package/strip-hex-prefix/1.0.0/src/index.js#L8), via [ethjs-util](https://npmfs.com/package/ethjs-util/0.1.6/src/index.js#L2). This upgrade bumped us to the latest version of etherumjs-util (7.1.5) where a new implementation of stripHexPrefix is used, which breaks a bunch of assumptions on our end. Basically we pass non-string inputs to stripHexPrefix all over the place and the old implementation would just return them unaltered but [the new one throws an error](https://npmfs.com/package/ethereumjs-util/7.1.5/src/internal.ts#L46).

Background context:
We're currently trying to trim down our bundle size and as part of that effort we're attempting to replace the heavy secp256k1-node  [implementation](https://www.npmjs.com/package/secp256k1) out of our dependency graph with the @noble/secp256k1 [implementation](https://www.npmjs.com/package/@noble/secp256k1). Currently [base-etc-keyring](https://www.npmjs.com/package/@keystonehq/base-eth-keyring) & [bc-ur-registry-eth](https://www.npmjs.com/package/@keystonehq/bc-ur-registry-eth) use ethereumjs-util version 7.0.8 which depends on [an old version of ethereum-cryptography](https://npmfs.com/package/ethereumjs-util/7.0.8/package.json#L95) which [depends on the heavier implementation](https://npmfs.com/package/ethereum-cryptography/0.1.3/package.json#L124). Bumping ethereumjs-util to the latest version 8.0.0 (package name is changed to @ethereumjs/util ), it now [depends on a newer version of ethereum-cryptography](https://npmfs.com/package/@ethereumjs/util/8.0.0/package.json#L87) which [uses the noble implementation](https://npmfs.com/package/ethereum-cryptography/1.1.2/package.json#L28).


shared/modules/hexstring-utils.js
```diff
@@ -73,3 +73,19 @@ export function toChecksumHexAddress(address) {
...
 }
+
+export function isHexPrefixed(str) {
+  if (typeof str !== 'string') {
+    throw new Error(
+      `Value must be type 'string', is currently type ${typeof str}, while checking isHexPrefixed.`,
+    );
+  }
+  return str.slice(0, 2) === '0x';
+}
```
brad-decker, line 84: Is this method not exported from ethereumjs/util?

adonesky1: It is. We can still use that version. I want us to use this version of `stripHexPrefix` for reasons that I will add to the PR description shortly here.

adonesky1: `stripHexPrefix` has been extracted so that we can use the version that was used in `ethereumjs-util` `v7.1.0` - which comes from [strip-hex-prefix](https://npmfs.com/package/strip-hex-prefix/1.0.0/src/index.js#L8), via [ethjs-util](https://npmfs.com/package/ethjs-util/0.1.6/src/index.js#L2). This upgrade bumped us to the latest version of etherumjs-util (7.1.5) where a new implementation of `stripHexPrefix` is used, which breaks a bunch of assumptions on our end. Basically we pass non-string inputs to `stripHexPrefix` all over the place and the old implementation would just return them unaltered but [the new one throws an error](https://npmfs.com/package/ethereumjs-util/7.1.5/src/internal.ts#L46).

brad-decker: I should have known, had similar issues with the v5->v7 upgrade back before EIP1559 for other utilities exported by this library.

brad-decker (approved):

ryanml (approved):
lgtm!

PR #15879, opened 2022-09-19
Revert #15823

### LINKED ISSUE #15790: [Bug]: Token Detection continues to work even when toggle is Disabled
### Describe the bug

**Problem**: even when I have the Token Detection toggle Disabled, I see that Token Detection functionality continues to work, and tokens that are detected are displayed.

[Screencast from 11-09-22 12:27:24.webm](https://user-images.githubusercontent.com/54408225/189522780-d640c975-816d-4589-a8c2-f444a15666c6.webm)

### Steps to reproduce

1. Load MM (have an account with some tokens)
2. Check Token Detection is disabled on Advanced Settings
3. Go Home and wait until it detects Tokens

_No response_

### Version

10.19.0

### Build type

_No response_

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

danjm:
Reverts #15823

That PR was an attempt to fix [MetaMask/metamask-extension#15790](https://github.com/MetaMask/metamask-extension/issues/15790)

But as @NiranjanaBinoy explained:

> we are turning OFF token detection when the toggle is OFF. During the development phase, we decided to keep token detection running for Mainnet (using static token list) alone when the toggle is OFF to preserve the existing functionality. If we are deciding to turn Token detection altogether when the toggle is OFF, then at the least:
we need to update the description copy at the token detection toggle as the current one is saying we have token detection running the Mainnet when the toggle is OFF
A few tweaks in the detect-token controller otherwise there might be inconsistent behaviour when the user turns the toggle on based on the difference in tokens available in the static list and dynamic list.

Instead, we will revert these changes, and then just change the copy of the token detection toggle, to make clear that when the toggle is off, detection still happens on maintain with the static list tokens. This is consistent with todays behaviour and original specs for token detection

### DIFF
--- a/ui/components/app/asset-list/asset-list.js
+++ b/ui/components/app/asset-list/asset-list.js
@@ -11,7 +11,7 @@ import {
   getShouldShowFiat,
   getNativeCurrencyImage,
   getDetectedTokensInCurrentNetwork,
-  getDisplayDetectedTokensLink,
+  getIstokenDetectionInactiveOnNonMainnetSupportedNetwork,
 } from '../../../selectors';
 import { getNativeCurrency } from '../../../ducks/metamask/metamask';
 import { useCurrencyDisplay } from '../../../hooks/useCurrencyDisplay';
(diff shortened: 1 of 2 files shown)


NiranjanaBinoy (approved):
LGTM

darkwing (approved):

Issue #15828, opened 2022-09-14
[Bug]: MV3 - Closed popup is displayed again after Service Worker re-connects


seaona:
### Describe the bug

**Problem**: when I unlock MM and see an informational popup (network deprecation), if I close the popup and Service Worker dies and re-connects, I can see that the popup is displayed again. I'm expecting that the state is preserved, and the popup does not appear again as it was already closed.

[network-popup.webm](https://user-images.githubusercontent.com/54408225/190124468-ac7c3342-af8d-4ff1-948e-83ad6150337e.webm)

### Steps to reproduce

See repro steps [here](https://github.com/MetaMask/metamask-extension/pull/15774#issuecomment-1245704820)

### Version

10.19.0

### Browser

Chrome

### Operating system

Linux

seaona:
Looks like it's not MV3 related and it's an expected behavior. I close the issue. See more here:
https://github.com/MetaMask/metamask-extension/issues/15829

danjm:
@seaona This is a really good find. And, in fact, it is MV3 related. Although it seems so similar to that other ticket you linked that its easy to conclude it is the same issue.

What is happening here, I believe, is a problem with non-persisted state being reset. If some UI is dependent on data stored in memory but not browser storage, then that UI will be interfered with by service worker restart. Another one that comes to mind is signature requests. If a service worker restarts while a signature request is open, it seems possible that the data might be wiped and the message will go away.

This specific ticket can remain closed, but we will want to investigate other areas where this might happen.

seaona:
thank you! I was actually thinking on re-opening again, as I was not 100% convinced on the other ticket explaining this behaviour. 
I will start investigating more in that area for other functionalities @danjm

PR #15913, opened 2022-09-21
Keep memstore contents after service worker restarts

### LINKED ISSUE #15954: MV3: memstore state does not persist service worker restart

### LINKED ISSUE #16456: Send and Confirm Screens errors on SW restarts
### Describe the bug

When service worker restarts, the GasFeeController state is lost and the below behaviours are noticed on Extension

**If SW restarts on send page, you get error saying gas estimation failed.**
Screenshot 2022-11-09 at 17 28 09

**If SW Restarts on Confirm Page, Confirm button is disabled and you get a continuous flashing of gas fees**
https://user-images.githubusercontent.com/44811/201209317-1bcb96d9-f25d-440b-9e6d-8a3ae21444f2.mov

Similar issue during DAPP interraction
Screenshot 2022-11-09 at 17 42 44

And also sending tokens

### Steps to reproduce

1. Start a transaction on MetaMask. It could be any transaction (Send, Contract Interaction, etc)
2. Stop Service Worker on Send Page, You should see an error saying `Gas fee estimation failed` after service worker restart
3. Go to Confirm Page
4. Stop Service Worker, you should see the gas fees flashing and Continue button disabled.

_No response_

### Version

develop

### Build type

_No response_

### Browser

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

segun:
Fixes 

#15954 
#16456 

## Explanation

The memstore contents is usually lost when the MetaMask extension window is closed.
The problem however is that when Service Worker restarts, the memstore is also cleared. This should not be the case

This PR intends to solve this problem by copying the controllers in memtore to persistent store and adding a method to manually clear the store of the controllers when the MetaMask extension is closed.

## More Information

* See: #15828 
* 
Blocked By #15558 

## Manual Testing Steps

1. Open [test-dapp](https://metamask.github.io/test-dapp/)
2. Connect to MetaMask
3. Click on any of SIGN buttons
4. MetaMask window should open to the sign page
5. Open developer console and click on Application Tab
6. Stop service worker
7. Service worker should be restarted automatically 
8. The MetaMask sign window should still be open
9. Close the window.
10. Open the window again
11. The window should open to the home screen (not sign screen).

**NOTE**: This may be not fully testable until David's work on #15558 is fully completed. Because as of now, when service worker is restarted the user is logged out.

## Pre-Merge Checklist

### DIFF
--- a/app/scripts/app-init.js
+++ b/app/scripts/app-init.js
@@ -128,7 +128,7 @@ chrome.runtime.onMessage.addListener(() => {
 });
 
 chrome.runtime.onStartup.addListener(() => {
-  chrome.storage.local.set({ isFirstTime: true });
+  globalThis.isFirstTimeProfileLoaded = true;
 });
 
 /*
--- a/app/scripts/app-init.js
+++ b/app/scripts/app-init.js
@@ -127,6 +127,10 @@ chrome.runtime.onMessage.addListener(() => {
   return false;
 });
 
+chrome.runtime.onStartup.addListener(() => {
+  chrome.storage.local.set({ isFirstTime: true });
+});
+
 /*
  * This content script is injected programmatically because
  * MAIN world injection does not work properly via manifest
--- a/app/scripts/lib/local-store.js
+++ b/app/scripts/lib/local-store.js
@@ -8,7 +8,7 @@ import { checkForError } from './util';
  */
 export default class ExtensionStore {
   constructor() {
-    this.isSupported = Boolean(browser.storage.local);
+    this.isSupported = Boolean(browser.storage?.local);
     if (!this.isSupported) {
       log.error('Storage local API not available.');
     }
--- a/app/scripts/lib/local-store.js
+++ b/app/scripts/lib/local-store.js
@@ -56,13 +56,19 @@ export default class ExtensionStore {
   /**
    * Returns all of the keys currently saved
    *
+   * This method originally gets all keys,
+   * this is not ideal if we want to manage keys outside the state object.
+   * The keys needed for state management are passed as defaut arguments ([data, meta])
+   * any other implementer can pass in their own key(s).
+   *
+   * @param key
    * @returns {Promise<*>}
    */
-  async get() {
+  async get(key = ['data', 'meta']) {
     if (!this.isSupported) {
       return undefined;
     }
-    const result = await this._get();
+    const result = await this._get(key);
     // extension.storage.local always returns an obj
     // if the object is empty, treat it as undefined
     if (isEmpty(result)) {
@@ -74,13 +80,14 @@ export default class ExtensionStore {
   /**
    * Returns all of the keys currently saved
    *
+   * @param key
    * @private
    * @returns {object} the key-value map from local storage
    */
-  _get() {
+  _get(key) {
     const { local } = browser.storage;
     return new Promise((resolve, reject) => {
-      local.get(null).then((/** @type {any} */ result) => {
+      local.get(key).then((/** @type {any} */ result) => {
         const err = checkForError();
         if (err) {
           reject(err);
--- a/app/scripts/lib/local-store.js
+++ b/app/scripts/lib/local-store.js
@@ -8,7 +8,7 @@ import { checkForError } from './util';
  */
 export default class ExtensionStore {
   constructor() {
-    this.isSupported = Boolean(browser.storage?.local);
+    this.isSupported = Boolean(browser.storage.local);
     if (!this.isSupported) {
       log.error('Storage local API not available.');
     }
@@ -56,19 +56,13 @@ export default class ExtensionStore {
   /**
    * Returns all of the keys currently saved
    *
-   * This method originally gets all keys,
-   * this is not ideal if we want to manage keys outside the state object.
-   * The keys needed for state management are passed as defaut arguments ([data, meta])
-   * any other implementer can pass in their own key(s).
-   *
-   * @param key
    * @returns {Promise<*>}
    */
-  async get(key = ['data', 'meta']) {
+  async get() {
     if (!this.isSupported) {
       return undefined;
     }
-    const result = await this._get(key);
+    const result = await this._get();
     // extension.storage.local always returns an obj
     // if the object is empty, treat it as undefined
     if (isEmpty(result)) {
@@ -80,14 +74,13 @@ export default class ExtensionStore {
   /**
    * Returns all of the keys currently saved
    *
-   * @param key
    * @private
    * @returns {object} the key-value map from local storage
    */
-  _get(key) {
+  _get() {
     const { local } = browser.storage;
     return new Promise((resolve, reject) => {
-      local.get(key).then((/** @type {any} */ result) => {
+      local.get(null).then((/** @type {any} */ result) => {
         const err = checkForError();
         if (err) {
           reject(err);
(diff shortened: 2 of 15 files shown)


app/scripts/metamask-controller.js
```diff
@@ -1070,21 +1104,14 @@ export default class MetamaskController extends EventEmitter {
       SnapController: this.snapController,
       NotificationController: this.notificationController,
       ///: END:ONLY_INCLUDE_IN
+      ...resetOnRestartConfig,
```
segun, line 1107: @danjm This part of code is causing Confirm Screen to hang. if I added the controllers manually, it works. I'm not exactly sure why this is failing.

jpuri:
Code changes look great, 2 feedbacks:

1. Can you plz check possibility of using global scope as discussed
2. Will be nice to have text coverage for metamask controller for reset flow

segun:
> Code changes look great, 2 feedbacks:
> 
> 1. Can you plz check possibility of using global scope as discussed
> 2. Will be nice to have text coverage for metamask controller for reset flow

globalThis didn't work as it gets reset everytime I restart SW.

jpuri:
> globalThis didn't work as it gets reset everytime I restart SW.

@segun : we can set it to true first time.

app/scripts/app-init.js
```diff
@@ -127,6 +127,10 @@ chrome.runtime.onMessage.addListener(() => {
   return false;
 });
 
+chrome.runtime.onStartup.addListener(() => {
+  globalThis.isFirstTimeProfileLoaded = true;
+});
+
```
jpuri, line 133: 👍

app/scripts/metamask-controller.test.js
```diff
@@ -193,6 +196,13 @@ describe('MetaMaskController', function () {
     await ganacheServer.quit();
   });
 
+  describe('should reset state on first time profile load', function () {
+    it('should reset state', function () {
+      assert(metamaskController.resetStates.calledOnce);
+      assert.equal(globalThis.isFirstTimeProfileLoaded, false);
+    });
+  });
+
```
jpuri, line 205: Nit: will be nice to have other test also that reset does not run if `globalThis.isFirstTimeProfileLoaded` is false.

jpuri (dismissed):

app/scripts/metamask-controller.js
```diff
@@ -1082,11 +1093,31 @@ export default class MetamaskController extends EventEmitter {
...
+      this.swapsController.resetState,
+      this.ensController.resetState,
+      this.approvalController.clear.bind(this.approvalController),
+      // WE SHOULD ADD TokenListController.resetState here too. But it's not implemented yet.
+    ];
+
+    if (globalThis.isFirstTimeProfileLoaded === true) {
+      this.resetStates(resetMethods);
+    }
+
```
jpuri, line 1120: @segun : should we restrict the above change to MV3 only ?

app/scripts/lib/local-store.js
```diff
@@ -56,13 +56,19 @@ export default class ExtensionStore {
   /**
    * Returns all of the keys currently saved
    *
+   * This method originally gets all keys,
```
danjm, line 59: Let's remove these changes because they are not used in this PR any more

app/scripts/metamask-controller.js
```diff
@@ -1041,21 +1062,14 @@ export default class MetamaskController extends EventEmitter {
       SnapController: this.snapController,
       NotificationController: this.notificationController,
       ///: END:ONLY_INCLUDE_IN
+      ...resetOnRestartStore,
     });
```
danjm, line 1067: Suggestion: investigate whether we can add a test that will catch mistakes of adding stores to this.memstore but not this.store

Maybe this doesn't make sense because some memstores need to persist service worker restarts and some do not

segun:
My work on the [flashing gas fees after MV3 restart](https://github.com/MetaMask/metamask-extension/issues/16456) brought me back here. I noticed that even though `store` retains the values of all specified controllers, `memstore` only retains the values of the below attached controller.

Screenshot 2022-11-11 at 01 03 22

I couldn't figure out why. 

I tried setting the persistence: true key of ComposableObservableStore but it still didn't change, I went through the controllers one by one to see if anything in the controllers that didn't get stored differs from the ones that did, but I still couldn't find the reason why.

To fix the issue though, since the state is in `store`, I added the below code to getState()

```
    let persistAfterRestartState = {};
    for (const key of Object.keys(this.resetOnRestartStore)) {
      // get the state from store
      const state = this.store.getState()[key];
      persistAfterRestartState = { ...persistAfterRestartState, ...state };
    }

    return {
      isInitialized,
      ...this.memStore.getFlatState(),
      ...persistAfterRestartState,
    };
```

Every other part of the PR remains the same.

@danjm @Gudahtt @jpuri

segun:
The fix for this (thanks to @Gudahtt) is to set initialState in GasFeeController.

I also took a look at all the controllers and these 3 controllers don't have initial state set

GasFeeController
TokenRatesController
RateLimitController

I have set their initial states also so that they don't lose value when SW restarts.

> My work on the [flashing gas fees after MV3 restart](https://github.com/MetaMask/metamask-extension/issues/16456) brought me back here. I noticed that even though `store` retains the values of all specified controllers, `memstore` only retains the values of the below attached controller.
> 
> Screenshot 2022-11-11 at 01 03 22
> 
> I couldn't figure out why.
> 
> I tried setting the persistence: true key of ComposableObservableStore but it still didn't change, I went through the controllers one by one to see if anything in the controllers that didn't get stored differs from the ones that did, but I still couldn't find the reason why.
> 
> To fix the issue though, since the state is in `store`, I added the below code to getState()
> 
> ```
>     let persistAfterRestartState = {};
>     for (const key of Object.keys(this.resetOnRestartStore)) {
>       // get the state from store
>       const state = this.store.getState()[key];
>       persistAfterRestartState = { ...persistAfterRestartState, ...state };
>     }
> 
>     return {
>       isInitialized,
>       ...this.memStore.getFlatState(),
>       ...persistAfterRestartState,
>     };
> ```
> 
> Every other part of the PR remains the same.
> 
> @danjm @Gudahtt @jpuri

app/scripts/metamask-controller.js
segun: We can't move the code from app-init, because that's the only file where `chrome.runtime.onStartup.addListener(() => {` makes sense. It's the only file that's loaded when chrome starts

app/scripts/metamask-controller.js
segun: I'm not really sure how to go about this. :(

app/scripts/metamask-controller.js
```diff
@@ -1106,11 +1121,33 @@ export default class MetamaskController extends EventEmitter {
...
+      this.ensController.resetState,
+      this.approvalController.clear.bind(this.approvalController),
+      // WE SHOULD ADD TokenListController.resetState here too. But it's not implemented yet.
+    ];
+
+    if (isManifestV3) {
+      if (globalThis.isFirstTimeProfileLoaded === true) {
+        this.resetStates(resetMethods);
+      }
+    }
```
danjm, line 1149: Let's make sure we reset the controllers that need to be reset under MV2 as well.

You could add
```
 else {
 this.resetState(resetMethods)
 ```
 onto the `if (isManifestV3) {` block

danjm (dismissed):
LGTM

danjm (approved):

jpuri (approved):

Issue #15092, opened 2022-06-29
[Ext Nav] Create component: ButtonSecondary


georgewrmarshall:
### Description
Create a reusable UI component for ButtonSecondary

### References
[Figma Link](https://www.figma.com/file/HKpPKij9V3TpsyMV1TpV7C/Design-System-2.0?node-id=1287%3A11481)
UI component guidelines: TBC
CSS guidelines: TBC
Testing guidelines: TBC

### Files needed
- `index.js`
- `index.scss`
- `ButtonSecondary.js`
- `ButtonSecondary.stories.js`
- `ButtonSecondary.test.js`
- `README.mdx`

### Technical details
TBC

### Acceptance criteria
- [ ] Uses color, typography, shadows design tokens 
- [ ] Uses semantic html
- [ ] PropTypes have descriptions
- [ ] Has storybook story with controls
- [ ] Has documentation in MDX
- [ ] Has unit tests and 90% coverage
- [ ] Works in Chrome and Firefox
- [ ] Performance tested: no unnecessary re-renders or other performance concerns

PR #15951, opened 2022-09-22
Feat/15087/design system buttons

### LINKED ISSUE #15087: [Ext Nav] Create component: ButtonBase
### Description
Create a reusable UI component for `ButtonBase`

### References
[Figma component](https://www.figma.com/file/HKpPKij9V3TpsyMV1TpV7C/Design-System-2.0?node-id=1287%3A11481)
[UI component guidelines](https://www.notion.so/UI-Components-5ccd2bf83eb8441892a0b72c0d8929e1)
[Styling guidelines](https://www.notion.so/Extension-Frontend-Engineering-Guide-b82ddb3e14004b5db3799a9b446294a9#49aa2ef7b7aa40178ff918043df71c34)
Testing guidelines: TBC

### Files needed
- `index.js`
- `index.scss`
- `button-base.js`
- `button-base.stories.js`
- `button-base.test.js`
- `README.mdx`

### Technical details
TBC

### Acceptance criteria
- [ ] Uses color, typography, shadows design tokens 
- [ ] Uses semantic html
- [ ] PropTypes have descriptions
- [ ] Has storybook story with controls
- [ ] Has documentation in MDX
- [ ] Has unit tests and 90% coverage
- [ ] Works in Chrome and Firefox
- [ ] Performance tested: no unnecessary re-renders or other performance concerns

garrettbear:
## Explanation
Adding new Design System Buttons: 
- ButtonBase (foundational button component)
- ButtonPrimary
- ButtonSecondary
- ButtonTertiary 

The updated buttons remove some button types and introduce new sizes. Additionally, the ButtonBase can be used when needing to build a customized button, which can then be proposed to the design system.

More button components will be added which can be found in [Figma].(https://www.figma.com/file/HKpPKij9V3TpsyMV1TpV7C/DS-Components?node-id=11%3A15)

## More Information

Fixes #15087 #15090 #15092 #15093 

### Before
Screen Shot 2022-09-22 at 3 24 21 PM

### After

Screen Shot 2022-09-22 at 3 29 46 PM

## Pre-Merge Checklist

### DIFF
--- a/ui/components/component-library/button-base/button-base.js
+++ b/ui/components/component-library/button-base/button-base.js
@@ -62,7 +62,7 @@ export const ButtonBase = ({
         {icon && (
           <Icon
             name={icon}
-            size={size === BUTTON_SIZES.AUTO ? SIZES.AUTO : SIZES.MD}
+            size={size === BUTTON_SIZES.AUTO ? SIZES.AUTO : SIZES.SM}
           />
         )}
         {children}
--- a/ui/components/component-library/button-base/button-base.js
+++ b/ui/components/component-library/button-base/button-base.js
@@ -1,0 +1,125 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+import Box from '../../ui/box';
+import { Icon, ICON_NAMES } from '../icon';
+import { Text } from '../text';
+
+import {
+  ALIGN_ITEMS,
+  BUTTON_SIZES,
+  JUSTIFY_CONTENT,
+  TEXT_COLORS,
+  DISPLAY,
+  TEXT,
+  SIZES,
+  FLEX_DIRECTION,
+} from '../../../helpers/constants/design-system';
+
+export const ButtonBase = ({
+  as = 'button',
+  block,
+  children,
+  className,
+  size = BUTTON_SIZES.MD,
+  icon,
+  iconPositionRight,
+  loading,
+  disabled,
+  ...props
+}) => {
+  return (
+    <Box
+      as={as}
+      paddingLeft={size === BUTTON_SIZES.AUTO ? 0 : 4}
+      paddingRight={size === BUTTON_SIZES.AUTO ? 0 : 4}
+      display={block ? DISPLAY.BLOCK : DISPLAY.INLINE_BLOCK}
+      className={classnames(
+        'mm-button',
+        `mm-button-size--${size}`,
+        {
+          [`mm-button--loading`]: Boolean(loading),
+          [`mm-button--disabled`]: Boolean(disabled),
+        },
+        className,
+      )}
+      disabled={disabled}
+      {...props}
+    >
+      <Text
+        as="span"
+        className="mm-button--content"
+        alignItems={ALIGN_ITEMS.CENTER}
+        justifyContent={JUSTIFY_CONTENT.CENTER}
+        flexDirection={
+          iconPositionRight ? FLEX_DIRECTION.ROW_REVERSE : FLEX_DIRECTION.ROW
+        }
+        gap={2}
+        variant={size === BUTTON_SIZES.AUTO ? TEXT.INHERIT : TEXT.BODY_MD}
+        color={TEXT_COLORS.INHERIT}
+      >
+        {icon && (
+          <Icon
+            name={icon}
+            size={size === BUTTON_SIZES.AUTO ? SIZES.AUTO : SIZES.MD}
+          />
+        )}
+        {children}
+      </Text>
+      {loading && (
+        <Icon
+          className="spinner"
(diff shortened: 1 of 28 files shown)


brad-decker:
@garrettbear can you help me understand the decision to make sub-components for each variant of button? I'll review once I have a clearer understanding of that. I'd like to understand it from both a implementation perspective as well as a contributor experience point of view.

brad-decker:
As much as the thought of importing different buttons for different variants made me cringe, the explanation from @georgewrmarshall and the accompanying video has fully changed my mind. 

One thing: I see us exporting the exact same constants from each variant with different names...can we eliminate that? I really feel that is confusing and leads the engineer to believe that the possibilities are different with each button.

georgewrmarshall (commented):
Could we break this PR up into digestible sizes? Could ButtonBase be one PR?

ui/components/component-library/button-base/button-base.js
```diff
@@ -0,0 +1,125 @@
...
+} from '../../../helpers/constants/design-system';
+
+export const ButtonBase = ({
+  as = 'button',
+  block,
+  children,
+  className,
+  size = BUTTON_SIZES.MD,
+  icon,
+  iconPositionRight,
```
georgewrmarshall, line 27: What do you think about using this a an enum rather than a boolean and setting the default to `"left"`? Could maybe even create a const to store directions and store it in `button-base.constants.js` this also aligns to how mobile is doing it.

 ```suggestion
  iconPosition = 'left',
```

File: ui/components/component-library/button-base/button-base.scss [hunk 2]
```diff
@@ -0,0 +1,78 @@
+.mm-button {
+  position: relative;
+  height: 2.5rem;
+  padding: 0;
+  border-radius: 999px;
+  cursor: pointer;
+  color: var(--color-text-default);
+  background-color: #dcdcdc;
```
georgewrmarshall, line 8: We should not be using any static hex values in our components

ui/components/component-library/button-base/button-base.scss
```diff
@@ -0,0 +1,78 @@
+.mm-button {
+  position: relative;
+  height: 2.5rem;
```
georgewrmarshall, line 3: I've been suggesting we should be using pixel values outside of font-sizes just to reduce some cognitive load doing the calculations from design to dev. Let's discuss this. There is a great article by Josh Comeau on it regarding accessibility https://www.joshwcomeau.com/css/surprising-truth-about-pixels-and-accessibility/

ui/components/component-library/button-base/button-base.scss
```diff
@@ -0,0 +1,78 @@
...
+    height: 2rem;
+  }
+
+  &-size--md {
+    height: 2.5rem;
+  }
+
+  &-size--lg {
+    height: 3rem;
+  }
```
georgewrmarshall, line 43: I think correct BEM naming conventions would use `&--size-auto`, `&--size-sm` etc

ui/components/component-library/button-base/button-base.scss
```diff
@@ -0,0 +1,78 @@
...
+  &--disabled,
+  &:disabled {
+    opacity: 0.3;
+    cursor: not-allowed;
+  }
+}
+
+.mm-button.box--display-block {
+  width: 100%;
+}
```
georgewrmarshall, line 62: I'm not sure we want to use other component class names within components. Mainly because if that class name gets updated in that component(in this case it probably wouldn't but I don't think we want to break that pattern) it would break this component. It would be more orthogonal if we created a `&--block` className and added the styles there

ui/components/component-library/button-base/button-base.scss
```diff
@@ -0,0 +1,78 @@
...
+
+
+// Temporary loading icon
+.spinner {
```
georgewrmarshall, line 66: I know this says temporary but if this is still going to be used it should follow the BEM naming convention

ui/components/component-library/button-base/button-base.scss
```diff
@@ -0,0 +1,78 @@
...
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  animation: spinner 1.2s linear infinite;
```
georgewrmarshall, line 71: Do we need to think about design tokens for animations?

georgewrmarshall (commented):
Looking really good! I just reviewed `ButtonBase` files. I've left some suggestions that are mostly about standardization for all of our stories and component consts let's find a time to discuss

ui/components/component-library/button-base/button-base.js
georgewrmarshall: Deciding to keep it as boolean after discussion

PR #15939, opened 2022-09-22
MV3: Cleanup and update contentscript.js streams

digiwand:
## Explanation

Originally, this PR addressed the "Could not establish connection. Receiving end does not exist." error. It did so with a timeout fix explained in this commit https://github.com/MetaMask/metamask-extension/pull/15939/commits/9ed8f8e5e3be2bcb08a78aa045eb381ef2502410. This commit with the timeout workaround has been removed. We will address this error in this issue: https://github.com/MetaMask/metamask-extension/issues/15970. The error appears to be non-blocking. See related issue for more details.

This PR cleans and updates the `contentscript.js` streams logic for the the dapp provider, legacy provider, and phishing warning page streams: 
- reestablish the page streams on disconnect 
- reorder reset logic by moving `removeAllListeners()` before the `destroy()` calls.
- checks for extension id before attempting to send the keep-alive message 

## Manual Testing Steps

1. Start the extension using MV3 build (`yarn start:mv3`)
2. View dapp page
3. Inspect app-init.js service worker and terminate service worker
4. Observe dapp continues to work

## Pre-Merge Checklist

- [x] PR template is filled out
- [x] **IF** this PR fixes a bug, a test that _would have_ caught the bug has been added  - no tests for contentscript.js exist at this time
- [x] PR is linked to the appropriate GitHub issue
- [x] PR has been added to the appropriate release Milestone

### + If there are functional changes:

- [x] Manual testing complete & passed
- [ ] "Extension QA Board" label has been applied

PR #15958, opened 2022-09-23
Send message to inpage when extension revives

### LINKED ISSUE #15052: MV3: ensure that dapp requests sent during service worker restarts are properly handled

jpuri:
The PR helps fix #15052

when service worker revives and metamask controller is initialised it sends message to inpage to re-play any pending actions


darkwing (dismissed):

digiwand:
I am receiving these bugs when I run this code and terminate the service worker. Needs more investigation. Screenshot 2022-09-23 at 7 32 27 PM

app/scripts/app-init.js
```diff
@@ -132,12 +132,15 @@ chrome.runtime.onMessage.addListener(() => {
...
-  {
-    id: 'inpage',
-    matches: ['file://*/*', 'http://*/*', 'https://*/*'],
-    js: ['inpage.js'],
-    runAt: 'document_start',
-    world: 'MAIN',
-  },
-]);
+// eslint-disable-next-line no-undef
+self.addEventListener('install', () => {
```
digiwand, line 136: This event only fires when a new service worker is installed. It will not necessarily fire when we terminate the service worker since we have our keep-alive logic. It also does not seem to fire upon refreshing the page. 

chrome.runtime.onConnect fires when a connection is made with the contentscript. maybe we can check if the `inpage.js` script has been loaded here.

digiwand, line 136: this caused a race condition on my macOS x chrome

Repro steps:
1. refresh extension from `chrome://extensions/` 
2. open or refresh dapp
3. observe provider is missing and dapp does not work 
Screenshot 2022-09-24 at 1 51 41 AM

I think, for now,  it would be better to catch the error instead and prioritize other work. The duplicate script error will drop the attempt to register a new inpage script and the existing inpage script will still be loaded. This should not cause any visible errors with our existing logic. 

I created a commit here to catch the error, for now:
https://github.com/MetaMask/metamask-extension/pull/15939/commits/26ba321a4af0858261518fd9649768050eeb9c62

danjm:
@jpuri The PR description says "The PR _helps_ fix https://github.com/MetaMask/metamask-extension/issues/15052". What other changes are needed for fully fix 15052?

jpuri:
> @jpuri The PR description says "The PR _helps_ fix #15052". What other changes are needed for fully fix 15052?

@danjm : I added details here: https://github.com/MetaMask/metamask-extension/issues/15052#issuecomment-1256594543

There are 2 other PRs that are required to be merged and then we need to update `provider` in extension repo to complete the task.

app/scripts/app-init.js
jpuri: Good catch @digiwand , let me revert this change.

digiwand: @jpuri  

I created a separate PR and issue ticket for the "Duplicate script ID 'inpage'" bug:
Issue: https://github.com/MetaMask/metamask-extension/issues/15971
PR: https://github.com/MetaMask/metamask-extension/pull/15973

darkwing (approved):

digiwand (approved):
Verified w/ Chrome x macOS

PR #15962, opened 2022-09-24
Use modern stylesheet selector

darkwing:
Using a modern selector and not looping twice with `filter` + `map` should improve perf.


ryanml (approved):
Nice!

xin3333:
Good changes

brad-decker (approved):

PR #15978, opened 2022-09-26
Persist state in metaRPCHandler so that we are sure state is persisted before sending back response to actions

### LINKED ISSUE #15914: [Bug]: [MV3] When I stop SW after deleting a Contact, I see that the contact is still there after login in again
### Describe the bug

**Problem**: when I delete a contact from my address book, and Service Worker is terminated right after, when I log in again I see that the contact is still there.

[delete-contact-terminate-sw.webm](https://user-images.githubusercontent.com/54408225/191470485-0161b9fb-5cc0-4ed2-994d-8da658d53399.webm)

### Steps to reproduce

1. Login to MM
2. Go to Settings / Contacts
3. Add a new contact
4. Click Edit contact
5. Click Delete contact
6. Terminate SW
7. Login again
8. Go to Settings / Contacts  -- see that the deleted contact is back there

_No response_

### Version

[MV3 build] develop

### Build type

_No response_

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

### LINKED ISSUE #15918: [Bug]: [MV3] Deleting a network and terminating SW right after, makes the deleted network appear again
### Describe the bug

**Problem**: after deleting a Network, if I terminate the service worker, I can see that the deleted network is back on the netowrk list.

[delete-network-terminate-sw.webm](https://user-images.githubusercontent.com/54408225/191487094-5283cefc-ecb3-40c3-838e-57384a506c25.webm)

### Steps to reproduce

1. Unlock MM
2. Show test networks
3. Delete localhost network -- see localhost is not there anymore
4. Terminate Service Worker
5. Check Networks on the drop-down menu -- see localhost is there again

_No response_

### Version

[MV3 build] develop

### Build type

_No response_

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

adonesky1:
## The Issue
While digging around to find the root cause of #15914 and #15918  @jpuri and I came across a [long standing issue](https://consensys.slack.com/archives/C02DUDT2WTU/p1663879099541579?thread_ts=1663878724.445999&cid=C02DUDT2WTU) that is now causing new problems in the context of MV3:

Because the browser storage API is asynchronous we cannot make guarantees about when our calls to [persist background store data](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/background.js#L387) will resolve, but to avoid laggy UX for any action that requires data persistence we optimistically reflect changes in the UI as if the persistence has completed. We believe that even within MV2 we may be vulnerable to data corruption issues with our current implementation that arise if the background process is disrupted while it is persisting data across multiple calls -  some of a batch of related data might get saved while other bits of it are not causing corruption. 

Because in MV3 the background process (now a service worker) is constantly (and unpredictably) winding down, the async storage API is leading to another issue: there is a small window of time when, if the service worker's lifecycle ends immediately after a user action and [the action is executed](https://github.com/MetaMask/metamask-extension/blob/develop/ui/store/action-queue/index.js#L62) in the background but not yet persisted, the action will [not be added to the replay queue](https://github.com/MetaMask/metamask-extension/blob/develop/ui/store/action-queue/index.js#L64) since the action was in fact executed - just not yet persisted.

## The Solution
This solution we've landed on is to persist the data that results from background state changes further upstream than previously, in the controller multiplex. This allows us to persist the data before the action is resolved ensuring that the queuing system will work - [if the action promise resolves](https://github.com/MetaMask/metamask-extension/blob/24a3156a2ba9e7ad55c44e3b5aa655d99bfbb500/ui/store/action-queue/index.js#L140) we can guarantee it has been persisted, if an error is thrown [we can push it to the queue and retry](https://github.com/MetaMask/metamask-extension/blob/24a3156a2ba9e7ad55c44e3b5aa655d99bfbb500/ui/store/action-queue/index.js#L64).

## More Information

* Fixes #15914 & #15918
* See also: [slack thread](https://consensys.slack.com/archives/C02DUDT2WTU/p1663878724445999)

### After

- The service worker dying (or being terminated) shortly after removing a network doesn't see the network deletion failing to persist (fixes https://github.com/MetaMask/metamask-extension/issues/15918):
  - https://user-images.githubusercontent.com/34557516/192581100-67bc1bab-e4ea-4231-8e00-f84ef54abbe8.mov

- The service worker dying (or being terminated) shortly after removing a contact doesn't see the contact deletion failing to persist (fixes https://github.com/MetaMask/metamask-extension/issues/15914):
  - https://user-images.githubusercontent.com/34557516/192581844-a654cfc8-0703-4851-a1b1-2123b1c877ac.mov

- The service worker dying (or being terminated) shortly after confirming a transaction doesn't lead to the extension failing to bump nonce or persist the state that recognizes the transaction is in flight. (fixes issue identified by Mariona [here](https://consensys.slack.com/archives/C02DUDT2WTU/p1663929271170509?thread_ts=1663878724.445999&cid=C02DUDT2WTU) - [video](https://consensys.slack.com/files/U032XS4NP9U/F043F2HP5V4/submit-tx.webm)):
  - https://user-images.githubusercontent.com/34557516/192584838-7c18af26-9202-419f-809b-fec81cc1448d.mov

## Manual Testing Steps

The most concerning risk with this PR is that we break state persistence somehow on the MV2 build. When testing, we'll want to ensure state is loaded from disk correctly and saved back to disk correctly.

To test loading, you can follow the same steps as for testing an extension update (i.e. start with a different build, switch to this one, and start the extension).

To test state saving, make any changes in the wallet that would be saved and restart it, then check the extension to ensure the changes remain.

I don't know of a great way to directly test the MV3 changes because they resulted in intermittent failures.

## Pre-Merge Checklist


app/scripts/lib/createMetaRPCHandler.js
```diff
@@ -20,6 +60,19 @@ const createMetaRPCHandler = (api, outStream) => {
     let error;
     try {
       result = await api[data.method](...data.params);
+
```
jpuri, line 63: There are a lot of actions besides 'trackMetaMetricsEvent', 'getState' which do no change state. But yep will be hard to mention all of those here.

jpuri (dismissed):
LGTM except some small NITs

app/scripts/lib/createMetaRPCHandler.js
```diff
@@ -1,6 +1,46 @@
 import { ethErrors, serializeError } from 'eth-rpc-errors';
+import { captureException } from '@sentry/browser';
+import log from 'loglevel';
+import ReadOnlyNetworkStore from './network-store';
+import LocalStore from './local-store';
```
jpuri, line 5: NIT: sorting imports

app/scripts/lib/createMetaRPCHandler.js
```diff
@@ -20,6 +60,19 @@ const createMetaRPCHandler = (api, outStream) => {
     let error;
     try {
       result = await api[data.method](...data.params);
+
+      if (
+        isDirty &&
+        store &&
+        !['trackMetaMetricsEvent', 'getState'].includes(data.method)
```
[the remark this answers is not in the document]
adonesky1: true. I'm just not sure that subscribing to the store and setting `isDirty` is actually helping? Since it seems like the `controller.store` is emitting `update` events even when methods like `trackMetaMetricsEvent` or `getState` are called?

app/scripts/lib/createMetaRPCHandler.js
adonesky1: yeah... we should probably try to get a comprehensive list here

app/scripts/lib/createMetaRPCHandler.js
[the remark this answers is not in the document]
jpuri: `trackMetaMetricsEvent` does changes state of metrics controller. But yep looks like `isDirty` will almost always be true.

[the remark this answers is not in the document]
adonesky1: removed the `store.subscribe` and `isDirty` part + removed `trackMetaMetricsEvent` from the array to skip on: https://github.com/MetaMask/metamask-extension/pull/15978/commits/15dc4b6287d4b9b4fecc08535262d42627bf25e6

jpuri (dismissed):

app/scripts/lib/createMetaRPCHandler.js
```diff
@@ -20,6 +52,18 @@ const createMetaRPCHandler = (api, outStream) => {
...
+      if (
+        store &&
+        // TODO: add any methods we know to not change state
+        !['getState'].includes(data.method)
+      ) {
+        const versionData = await localStore.get();
+        await persistData(
+          { data: store.getState(), meta: versionData?.meta },
+          localStore,
+        );
```
brad-decker, line 65: What is the intent of using meta from versionData (localStore) versus store? I don't see it initiated anywhere so when would it get set?

app/scripts/lib/createMetaRPCHandler.js
brad-decker: I'd recommend making an object with keys of method names and checking the object for key existing, rather than an includes call, for performance reasons.

app/scripts/lib/createMetaRPCHandler.js
```diff
@@ -1,6 +1,38 @@
+import log from 'loglevel';
 import { ethErrors, serializeError } from 'eth-rpc-errors';
+import { captureException } from '@sentry/browser';
+import ReadOnlyNetworkStore from './network-store';
+import LocalStore from './local-store';
 
-const createMetaRPCHandler = (api, outStream) => {
+let dataPersistenceFailing = false;
+
+async function persistData(state, localStore) {
```
brad-decker, line 9: Suggest renaming this method to signify that it is a failsafe, fallback state persistence layer, and document it with jsdocs explaining the goal is to overwrite the state with a more up-to-date copy just before the disconnect happens. I don't want future reviewers of this code to believe that this is the primary persistence layer.

brad-decker (changes requested):
Seems like a sound approach, a few improvement ideas and a question

app/scripts/lib/createMetaRPCHandler.js
brad-decker: I guess the answer here is the same as my eureka moment with the whole overwriting ... localStore.get *gets* the store from local store where it has always persisted and will have the current meta / version info from the extension.

brad-decker: This comment can be resolved by just adding a little explainer comment

app/scripts/lib/createMetaRPCHandler.js
adonesky1: Well actually it kindof is the primary persistence layer, as it will always get called before the persist call [here](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/background.js#L357). As we discussed in slack, I'm not sure whether we ought to even keep the other persist call (linked in the previous sentence) any longer., since I believe it will produce redundant persists as it currently stands in this PR. @Gudahtt hoping you could chime in here.

adonesky1: removed the old persistence layer [here](https://github.com/MetaMask/metamask-extension/pull/15978/commits/90f5e60903a092fc9b9163b06078efe3a44915e0) (still needs some cleanup)... not sure if this is for keeps but gave it a whirl and behaves correctly.

app/scripts/lib/createMetaRPCHandler.js
jpuri: I would suggest just use `getState`. Creating a method list here can cause error if a method is modified but this list is not updated.

Having an extra call to persist state is not a huge overhead I think.

app/scripts/lib/createMetaRPCHandler.js
Gudahtt: Hmm, why would the old persistence layer be removed? Not sure I follow. I thought we still wanted to ensure state was persisted even if there was no RPC traffic. For example, if a transaction settles while the user is not using MetaMask.

app/scripts/lib/createMetaRPCHandler.js
```diff
@@ -20,6 +52,19 @@ const createMetaRPCHandler = (api, outStream) => {
     let error;
     try {
       result = await api[data.method](...data.params);
+
+      if (
```
Gudahtt, line 56: Do we want to persist state upon failure as well? Some error cases still involve important state changes, and an error response is still a response.

app/scripts/lib/createMetaRPCHandler.js
adonesky1: I had thought that this would pass through the RPC client too, but I guess it doesn't... So are we ok with all the redundant calls that will happen with both layers? I suppose we have to live with it for now?

app/scripts/lib/createMetaRPCHandler.js
adonesky1: yep makes sense to me

app/scripts/lib/createMetaRPCHandler.js
Gudahtt: I do not think we have to just live with it, no. We can prevent redundant calls by writing our own `debounce` logic to remove this duplication.

adonesky1: hmmm right okay. I think I was kindof confused by the use of debounce, because its sortof a different concept right? Just call persist every second and batch any requests that came during that second.

adonesky1: > The first condition is upheld right now just by using debounce, but we can scrap that and implement our own custom logic to accomplish both.

As part of this are you suggesting that we somehow identify if the state update has already been accomplished by the action taken to
> Persist state immediately before returning any JSON-RPC response.
?

Sorry this is getting a bit convoluted, but basically it seems if we keep both `persistData` calls as they stand we're achieving both conditions, we're just getting a bunch of "redundant" calls because each time we persist the immediately before returning any JSON-RPC response,  we also then `persistData` again [here](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/background.js#L357)?

Gudahtt: Yes, you're right, with the redundant calls we accomplish both goals. I was hoping we could avoid the redundant calls but we don't need that to be done here. We can try that later.

app/scripts/lib/createMetaRPCHandler.js
adonesky1: Done [here](https://github.com/MetaMask/metamask-extension/pull/15978/commits/d00d9577d808a346cfa5a03f4463602d144edf6e)

app/scripts/lib/createMetaRPCHandler.js
```diff
@@ -1,6 +1,38 @@
+import log from 'loglevel';
 import { ethErrors, serializeError } from 'eth-rpc-errors';
+import { captureException } from '@sentry/browser';
+import ReadOnlyNetworkStore from './network-store';
+import LocalStore from './local-store';
 
-const createMetaRPCHandler = (api, outStream) => {
+let dataPersistenceFailing = false;
```
Gudahtt, line 7: We're duplicating a lot of things here that are already in `background.js`. Could we move these to a dedicated module instead, and use it in both places? Then we won't be duplicating this logic, and the "data persistence is failing" state.

adonesky1: Deduped [here](https://github.com/MetaMask/metamask-extension/pull/15978/commits/f28adb38111b291f75b4f920ff2dd70323a96b9e)

app/scripts/lib/persistData.js
```diff
@@ -0,0 +1,27 @@
+import log from 'loglevel';
+import { captureException } from '@sentry/browser';
+
+let dataPersistenceFailing = false;
```
brad-decker, line 4: before we deduplicated this logic, the two places had separate 'dataPersistenceFailing' variables. Is having them share one going to cause any issues?

adonesky1: It's just used for logging. It wouldn't seem to cause any issues even for that as far as I can tell. It will update the variable on each call, and it won't impact the functionality of the next call.

brad-decker (dismissed):

app/scripts/lib/persistData.js
Gudahtt: It's meant to represent cases where `storage.local` fails. Usually happens when the user's disk is full. There is an issue in the backlog somewhere for adding an error to the UI when this happens, but that work hasn't been done yet.

We're working with the same storage API here, so there isn't much sense in having two variables tracking this. Either storage is full or it isn't.

app/scripts/lib/persistData.js
```diff
@@ -0,0 +1,27 @@
+import log from 'loglevel';
```
Gudahtt, line 1: Our filename convention is to use kebab-case rather than camelCase. The same conventions as for npm packages.

adonesky1: fixed [here](https://github.com/MetaMask/metamask-extension/pull/15978/commits/c01123238c0aaafd150dc7fd6caa72770418d4c8)

brad-decker (dismissed):

app/scripts/lib/createMetaRPCHandler.js
```diff
@@ -22,6 +28,15 @@ const createMetaRPCHandler = (api, outStream) => {
       result = await api[data.method](...data.params);
     } catch (err) {
       error = err;
+    } finally {
+      if (store && data.method !== 'getState') {
+        // we retrieve the already persisted data from local store to provide the version metadata to this persist call
+        const versionData = await localStore.get();
```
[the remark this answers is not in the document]
Gudahtt: We could move this, as well as the choice in store, into a module representing our storage layer. That would remove further duplicate code, and it would draw a clearer line between which code is responsible for persisting data, and which is responsible for sending RPC messages.

I understand the need for the RPC client to trigger data to be persisted - that makes sense. But it doesn't make sense to me for it to decide _where_ to persist it or how, and that's happening here. We should be able to capture this responsibility better with an abstraction.

app/scripts/lib/local-store.js
```diff
@@ -11,6 +12,38 @@ export default class ExtensionStore {
     if (!this.isSupported) {
       log.error('Storage local API not available.');
     }
+    this.dataPersistenceFailing = false;
+  }
+
+  async init() {
+    this.initialStoreState = await this.get();
```
Gudahtt, line 19: Are we sure this is only run after the migrator? The migrator is what sets the version if the extension has been updated.

adonesky1: It seems to be, but to be sure I could move it right after this: https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/background.js#L295

adonesky1: wait that won't work...

adonesky1: in my testing it does a appear to be called after the migrator has run, but we could, alternatively, call `init()` [here](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/background.js#L295), pass down the same instantiation of `localStore` from `background.js` through `setupController` in the `MetamaskController` constructor and then into `createMetaRPCHandler` so that it accesses the same `initialStoreState` class variable when run in the RPC client

app/scripts/background.js
```diff
@@ -356,7 +355,7 @@ function setupController(initState, initLangCode, remoteSourcePort) {
     storeAsStream(controller.store),
     debounce(1000),
     storeTransformStream(versionifyData),
```
Gudahtt, line 357: Do we still need this step? The "versionifyData" step is intended not just to move the state under the `data` property. It is also where the cached version is set. But now that's being set in `local-store.js` instead.

app/scripts/lib/local-store.js
Gudahtt: Yeah I was just thinking the same thing!

app/scripts/lib/createMetaRPCHandler.js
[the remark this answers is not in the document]
Gudahtt: Thanks! So this seems like a good start. We can move further discussion to other threads and resolve this one.

app/scripts/background.js
adonesky1: Nope! Removing now.

adonesky1: removed [here](https://github.com/MetaMask/metamask-extension/pull/15978/commits/fdeb4399ca7665ee03bb979fceb5ec6c07abe5f2)

app/scripts/lib/local-store.js
adonesky1: Done [here](https://github.com/MetaMask/metamask-extension/pull/15978/commits/fdeb4399ca7665ee03bb979fceb5ec6c07abe5f2)

app/scripts/lib/local-store.js
```diff
@@ -11,6 +12,38 @@ export default class ExtensionStore {
...
       log.error('Storage local API not available.');
     }
+    this.dataPersistenceFailing = false;
+  }
+
+  setMetaData(initMetaData) {
+    this.metadata = initMetaData;
+  }
+
+  async persistStateToLocalStore(state) {
```
Gudahtt, line 22: Nit: This function could be simplified if it accepted the controller state directly. In practice, the `meta` property is never passed in here.

app/scripts/background.js
```diff
@@ -293,6 +292,8 @@ async function loadStateFromPersistence() {
   // write to disk
   if (localStore.isSupported) {
     localStore.set(versionedData);
```
Gudahtt, line 294: Nit: Could we use `persistStateToLocalStore` here instead? That is all we're doing here, persisting state.

Then we can remove this `isSupported` check, and simplify this function a bit. We'd need to set the metadata first before calling `persistDateToLocalStore` though.

Gudahtt:
Looking into the CI failures now

app/scripts/lib/local-store.js
Gudahtt, line 22: CI is failing because the network store doesn't have this method

app/scripts/background.js
adonesky1: Done [here](https://github.com/MetaMask/metamask-extension/pull/15978/commits/6d82b9026a036ba1b960d4e9659e73f542896dc7)

app/scripts/lib/local-store.js
adonesky1: Done [here](https://github.com/MetaMask/metamask-extension/pull/15978/commits/6d82b9026a036ba1b960d4e9659e73f542896dc7)

app/scripts/lib/local-store.js
```diff
@@ -11,6 +12,37 @@ export default class ExtensionStore {
...
+  async set(state) {
+    if (!state) {
+      throw new Error('MetaMask - updated state is missing');
+    }
+    if (this.isSupported) {
+      try {
+        // we format the data for storage as an object with the "data" key for the controller state object
+        // and the "meta" key for a metadata object containing a version number that tracks how the data shape
+        // has changed using migrations to adapt to backwards incompatible changes
+        await this._set({ data: state, meta: this.metadata });
```
Gudahtt, line 34: We're not checking here to ensure `this.metadata` is set. If it's unset, that would _severely_ mess up the persisted wallet state. It might result in the wallet being bricked/unrecoverable. We should explicitly check that and throw if it's unset before referencing it here.

adonesky1: Done [here](https://github.com/MetaMask/metamask-extension/pull/15978/commits/817d85d91b67aaf401490b7372ebb7511919b37a) + added tests

app/scripts/lib/network-store.js
```diff
@@ -50,16 +50,24 @@ export default class ReadOnlyNetworkStore {
...
    *
    * @param {object} state - The state to set
-   * @returns {Promise<void>}
    */
   async set(state) {
     if (!this._initialized) {
       await this._initializing;
     }
-    this._state = state;
+    this._state = { data: state, meta: this._metadata };
```
Gudahtt, line 71: Nit: This should have similar validation, so that our e2e tests can catch cases where we try to persist at the wrong time

Gudahtt (dismissed):
LGTM! We should test this carefully though. We really don't want to break state persistence.

Gudahtt:
I've just added the QA label, and added basic manual testing steps to the PR description

brad-decker (dismissed):

brad-decker:
Added do not merge flag until we get a QA review

app/scripts/lib/local-store.js
```diff
@@ -11,6 +12,45 @@ export default class ExtensionStore {
...
+    if (!this.metadata) {
+      throw new Error(
+        'MetaMask - metadata must be set on instance of ExtensionStore before calling "set"',
+      );
+    }
+    try {
+      // we format the data for storage as an object with the "data" key for the controller state object
+      // and the "meta" key for a metadata object containing a version number that tracks how the data shape
+      // has changed using migrations to adapt to backwards incompatible changes
+      await this._set({ data: state, meta: this.metadata });
```
Gudahtt, line 43: Should we make this MV3-only? Unsure on whether this has any advantages for MV2

adonesky1:
> Added do not merge flag until we get a QA review

Working with @seaona today to get this QA'd

app/scripts/lib/local-store.js
Gudahtt: I'm sorry, I wrote this on the wrong line :facepalm: I will move this comment

app/scripts/lib/createMetaRPCHandler.js
```diff
@@ -22,6 +22,10 @@ const createMetaRPCHandler = (api, outStream) => {
       result = await api[data.method](...data.params);
     } catch (err) {
       error = err;
+    } finally {
+      if (store && data.method !== 'getState') {
```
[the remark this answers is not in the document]
adonesky1: 🤔 I mean I suppose it could (but might not) contribute to resolving some of those "data corruption" errors, the source of which we're not entirely sure, but which you'd mentioned earlier. Otherwise yeah we can limit it to MV3.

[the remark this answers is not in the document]
adonesky1: Ok yep. Thinking about it now, this might make it worse 🤔? Since we might be making more non-atomic state updates rather than (sortof) batching them with the debounce...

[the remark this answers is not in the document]
adonesky1: restricted this set to MV3 [here](https://github.com/MetaMask/metamask-extension/pull/15978/commits/93229e356ba6fff173b386651e4287d6d99b930f)

app/scripts/lib/network-store.js
adonesky1: done [here](https://github.com/MetaMask/metamask-extension/pull/15978/commits/93229e356ba6fff173b386651e4287d6d99b930f)

app/scripts/lib/createMetaRPCHandler.js
```diff
@@ -22,6 +23,10 @@ const createMetaRPCHandler = (api, outStream) => {
       result = await api[data.method](...data.params);
     } catch (err) {
       error = err;
+    } finally {
+      if (isManifestV3 && store && data.method !== 'getState') {
+        localStoreApiWrapper.persist(store.getState());
```
Gudahtt, line 28: Didn't you rename this from `persist` back to `set`?

Gudahtt (approved):

jpuri (approved):
LGTM, good to merge after QA by @seaona

seaona:
From what I found so far, on MV3, I've been able to reproduce the issues, by terminating the SW right after performing actions.

- Creating an Account
- Seeing Account created
- Terminate SW right after -- account is not there

[create-account-sw.webm](https://user-images.githubusercontent.com/54408225/194912552-eae0625d-148f-45f2-bcc0-dfea86f3711e.webm)

- Deleting a Network
- Seeing Network is removed
- Terminate SW right after -- network is back

[detele-network-sw.webm](https://user-images.githubusercontent.com/54408225/194912564-24b6bc08-583f-4aa8-8d43-f3edc56919f4.webm)

On MV2, I've performed MM upgrades and I've seen that the state is preserved, as well as the Download State logs and Backup / Restore functionalities seem to work well so far

Gudahtt:
Thanks @seaona 

Could you verify which build that test was performed with? Specifically I'm wondering whether it included the most recent commit. Before that commit, we would expect the behaviour shown here.

adonesky1:
> Could you verify which build/commit that test was performed with? Specifically I'm wondering whether it included the most recent commit. Before that commit, we would expect the behaviour shown here.

yes I'm really sorry for the back and forth here @seaona.

seaona:
I confirm I didn't have the last update. Now I see it fix: actions are performed no matter if I terminate the SW right after. Thanks for pointing this out :blush: @adonesky1 @Gudahtt 

I've triggered an error saying `no keyrings found` when I terminate the SW after clicking on confirm Swap.

[swap-keyring.webm](https://user-images.githubusercontent.com/54408225/194924305-32cee896-7975-43b0-9444-2cd09767e91d.webm)

And this same error can be triggered if I terminate the SW and perform a Send action right after.

[terminate-and-send.webm](https://user-images.githubusercontent.com/54408225/194923142-551523ba-fa1d-41c0-a6ca-1d5f546e57d3.webm)

I am not sure if these actions should be queued/saved until SW is re-start? Also, not sure if this should be addressed with this PR or is it something already expected that will be addressed in another PR

adonesky1:
@seaona thank you! I will investigate. Very strange. I'd tested the send flow one at least and wasn't seeing this before, but will check it out! Great catches

Gudahtt:
That error looks related to @darkwing 's work to keep the wallet unlocked between service worker restarts. Currently you should get logged out after restart, so any action that requires access to a keyring would fail. Even operations that reference an account might fail, because the list of accounts is in the keyring.

seaona:
Make sense thank you!! Overall looks good to me! I've performed upgrades and state seems to be preserved. Just noticed that there is a change in the order of the Accounts on the state logs file. I guess it's nothing concerning as the same data is there, but just mentioning just in case.

In this first image I did:
1. Load MM 10.21.0 RC version and downloaded the State (left img)
2. Update MM to this PR version and download the State without any other action in between (right img)

image


In the next image I did:
1. Load MM to this PR version version and downloaded the State (left img)
2. Update MM to develop version and download the State without any other action in between (right img)

accounts-leftPR-rightdevelop

adonesky1:
@seaona I believe these discrepancies in the account ordering are fine in the downloaded state logs. Thanks so much for your repeated (😬) thorough regression testing here! 

Assuming there are no other related regressions you spotted, and you are seeing the issues related to the service worker terminating immediately after non-keyring related actions as resolved, I think we should be good to go ahead and merge this. @Gudahtt sound right to you?

Gudahtt:
Agreed; as long as the mv2 testing turned out ok, this should be good to merge. The mv3 issues can be ironed out after other compatibility work is merged.

PR #15985, opened 2022-09-26
Optimize account name check

darkwing:
I spotted this during a rebase and wanted to submit it before I forgot.  Simple code improvement.


georgewrmarshall (approved):
Approving from my understanding that this removes the `map` method(which creates a new array) and is more expensive than using the `find` method(which "returns the first element in the provided array that satisfies the provided testing function")

ryanml (approved):

PR #15888, opened 2022-09-20
Update address component on transaction data screen

amerkadicE:
## Explanation

This pull request will update the current logic for a recipient address on the transaction data screen.
Now it will show 
1.  If the address exists in the user account list, then the account name is displayed
2.  If the address exists in the user address book (contacts), then their nickname is displayed 
3.  If the address exists in the contract metadata list, then the contract's name is displayed
4.  If conditionals from before are not met, it will show the user address.

Also, clicking on this component opens the 'add/edit nickname' popup or this tooltip if it's one of the user's accounts.

After checking the code, we found out that this PR will also update the logic on the recipient address when sending funds as shown in screenshot below. 

## More Information

Fixes #15780 

### Before

image

### After

image
image

Send screen 
image

## Pre-Merge Checklist

- [x] PR template is filled out
- [x] **IF** this PR fixes a bug, a test that _would have_ caught the bug has been added
- [x] PR is linked to the appropriate GitHub issue
- [ ] PR has been added to the appropriate release Milestone

### + If there are functional changes:

- [ ] Manual testing complete & passed
- [ ] "Extension QA Board" label has been applied

Issue #15780, opened 2022-09-09
Update the address component


bschorchit:
## Background
This component is used on the "to" address when you go to activity and click on a previous transaction (screenshot below). It might be used in other places too (The assigned developer should check in the code directly). We need to update the current logic for this component as described below. 
![Untitled](https://user-images.githubusercontent.com/34968356/186765431-3e3c4d73-59c2-4eb2-93d1-9215c0bdf52f.jpg)

## Acceptance Criteria
1. If the address exists in the user account list, then the account name is displayed (this should be the first thing checked, if it does exist in user account list we don't need to check the rest);
2. If the address exists in the user address book (contacts), then their nickname is displayed (this should be already existing logic for this component - this should be second thing checked);
3. If the address exists in the contract metadata list, then the contract's name is displayed (third thing checked);
4. If the address doesn't exist in address book or contract metadata, then the address is displayed;
5. Clicking on this component open the 'add/edit nickname' popup (1st screenshot below) or this tooltip (2nd screenshot below) if it's one of the user's accounts.
![Screen Shot 2022-08-25 at 17 46 43](https://user-images.githubusercontent.com/34968356/186765657-15e1e8a3-6a9d-4129-8e38-cb251f319d11.png)
![Screen Shot 2022-08-25 at 18 06 06](https://user-images.githubusercontent.com/34968356/186768919-b1f9a72b-5d5a-4a81-a430-f15c2a849122.png)

## Steps to Reproduce
1. Send a transaction;
2. Go to activity;
3. Click on the transaction to see the details;
4. Notice the component displayed for the address in that page.

newlife08:
![image](https://user

PR #16018, opened 2022-09-28
Replaced addresses by the address component on SignTypedData v4 signatures

### LINKED ISSUE #15716: Replace addresses by the address component on SignTypedData v4 signatures
## Resources
[Figma Design file](https://www.figma.com/file/U2Ojf3cJP4PvIYRiRfUya0/Improve-Signatures-(v1)?node-id=1971%3A1763)
[MM SignTypedData v4 docs](https://docs.metamask.io/guide/signing-data.html#sign-typed-data-v4)

## Acceptance Criteria
1. When SignTypedData v4 confirmation screens are displayed, all address parameters in the message are displayed using our address component (this is the same component used for the "to" address when you go to activity and click on a previous transaction - screenshot below);
Untitled

## Additional Note
MetaMask has no control on what the message that is displayed is this screen is - this is something provided by the dapps and will look differently for each one of them. MetaMask does have control on how such messages are displayed and how the UI looks like - and this is what we are trying to achieve with this ticket. So with this ticket, we want every value that is an address to be displayed using the address component. If you check in the docs linked above the javascript code for the signature, you can see that a `type` property is passed to each parameter in the signature and `address` is one of the possible values for this type property. Additionally, the gray background color should be replaced with the white one and the outline should be added as per the Figma design. 

What signTypedData v4 message looks like now (exact message will change depending on the dapp):
Screen Shot 2022-08-25 at 17 58 41

What it should look like after this ticket (exact message will change depending on the dapp + the other differences not related to how addresses are being shown can be ignored here):
Screen Shot 2022-08-25 at 17 59 44

## Steps to Reproduce
1. Open the test dapp https://metamask.github.io/test-dapp/
2. Under the SignTypedData v4 section, click on "Sign"
or
1. Get some test NFTs on https://faucet.paradigm.xyz/
2. Make sure MM is connected to Rinkeby test network
3. Go to https://testnets.opensea.io/account
4. Click on one of your NFTs
5. Click on 'Sell'
6. Go through the steps until you get prompt to 'Confirm listing' on OpenSea - this will trigger a SignTypedData v4 prompt in MM (you might have to sign a different message and approve a transaction if it's the first time you're going through this flow)
Screen Shot 2022-09-13 at 10 27 38

VSaric:
## Explanation

MetaMask has no control on what the message that is displayed is this screen is - this is something provided by the dapps and will look differently for each one of them. MetaMask does have control on how such messages are displayed and how the UI looks like - and this is what we are trying to achieve with this PR. So with this PR, we want every value that is an address to be displayed using the address component. 

## More Information

* Fixes #15716 

### Before (test-dapp)

Screenshot 2022-10-04 at 14 32 15

Screenshot 2022-10-04 at 14 32 29

### After (test-dapp)

Screenshot 2022-10-12 at 10 08 11

Screenshot 2022-10-12 at 10 08 51

### Before (test-dapp)

https://user-images.githubusercontent.com/92527393/192788293-5411b612-b8fb-4009-83d9-5c4a856712a7.mov

### After (test-dapp)

https://user-images.githubusercontent.com/92527393/193854495-4df94bf0-8f41-4d1c-9336-e00b620d3248.mov

### Before (OpenSea)

https://user-images.githubusercontent.com/92527393/193039323-341317c1-509b-4a63-a5a7-29b36aa2d36c.mov

### After (OpenSea)

https://user-images.githubusercontent.com/92527393/193858085-7b00dc45-c459-4fcf-8692-3d3e952be93f.mov

## Manual Testing Steps

1. Open the test dapp https://metamask.github.io/test-dapp/
2. Under the SignTypedData v4 section, click on `Sign`
    or
3. Get some test `NFTs` on https://faucet.paradigm.xyz/
4. Make sure MM is connected to `Rinkeby test network`
5. Go to https://testnets.opensea.io/account
6. Click on one of your `NFTs`
7. Click on `Sell`
8. Go through the steps until you get prompt to `Confirm listing` on OpenSea - this will trigger a `SignTypedData v4` prompt in MM (you might have to sign a different message and approve a transaction if it's the first time you're going through this flow)


mirjanaKukic:
Verified by QA

ui/components/app/signature-request/signature-request-message/signature-request-message.js
```diff
@@ -0,0 +1,120 @@
...
+
+    if (isAtBottom) {
+      setMessageIsScrolled(true);
+      onMessageScrolled();
+    }
+  };
+
+  const renderNode = (renderData) => {
+    return (
+      <div className="signature-request-message--node">
```
brad-decker, line 40: So, a few things.

1. This isn't proper BEM syntax for class names `--` is a modifier spacer and `__` is a block spacer. so i think we'd want `signature-request-message__node`.
2. Later you have `--node-label` and use `&-label` to style it, this also isn't proper BEM syntax, node-label would be its own new block not a nested block.
3. I realize you are trying to use the existing styles here, but I would prefer that if you are going to remove old patterns (thank you!!! this is awesome!!!) by refactoring class components to functional, dropping the '.component' syntax that we also use `Box` and `Typography` to ease in the transition to the new design system in the future.

let me know if that makes sense

brad-decker (changes requested):

ui/components/app/signature-request/signature-request-message/signature-request-message.js
VSaric: You are right @brad-decker, sorry for that! I know that we are using `Box` and `Typography` by refactoring class components to functional and I know about proper BEM syntax, sorry for that once again! I will take care of this in the future PRs! 
I left only one `div` because it has the `ref` property. I have not been able to do this with the `Box` component, because `Box` component don't have that property. If it is okay to stay like this great, but if you have some suggestions about this please comment and I will make changes. I made changes [here](https://github.com/MetaMask/metamask-extension/pull/16018/commits/d3277889201306fd0a3cb313ea0d6b2f59181a0b).
Thank you in advance!

georgewrmarshall (commented):
This is a super cool update! I've checked out your branch and made some minor updates. It contains an update from develop so it will be failing currently but once rebased from `develop` it should work

[PR with suggestions here](https://github.com/MetaMask/metamask-extension/pull/16148)

Would also be super cool to get a story with this view

georgewrmarshall (dismissed):
UI LGTM! 💯

ui/selectors/selectors.js
```diff
@@ -396,6 +396,20 @@ export function getAddressBookEntryOrAccountName(state, address) {
   return entry && entry.name !== '' ? entry.name : address;
 }
 
+export function getAccountName(identities, address) {
```
danjm, line 399: I think you can use `getAddressBookEntryOrAccountName` instead of this

ui/selectors/selectors.js
```diff
@@ -396,6 +396,20 @@ export function getAddressBookEntryOrAccountName(state, address) {
...
+  const entry = Object.values(identities).find((identity) =>
+    isEqualCaseInsensitive(identity.address, toChecksumHexAddress(address)),
+  );
+  return entry && entry.name !== '' ? entry.name : '';
+}
+
+export function getMetadataContractName(address) {
```
danjm, line 406: We should use `getTokenList` here, so that we use the correct list if `isTokenDetectionInactiveOnMainnet` is true

ui/components/app/transaction-decoding/components/decoding/address/address.component.js
```diff
@@ -19,16 +22,29 @@ const Address = ({
...
+  };
+
+  const recipientNickname = getNickName(checksummedRecipientAddress);
+  const recipientMetadataName = getMetadataContractName(
+    checksummedRecipientAddress,
   );
-  const recipientNickname = addressBookEntryObject?.name;
 
   const recipientToRender = addressOnly
-    ? recipientNickname ||
```
danjm, line 28: Before this PR, `recipientName` was not one of the options if `addressOnly` was true. Now this PR adds it as one of the options in that case. Do we know why it wasn't one of the options previously and why this is an okay change to make?

ui/components/app/transaction-decoding/components/decoding/address/address.component.js
```diff
@@ -19,16 +22,29 @@ const Address = ({
   const [showNicknamePopovers, setShowNicknamePopovers] = useState(false);
 
   const addressBook = useSelector(getAddressBook);
-  const addressBookEntryObject = addressBook.find(
-    (entry) => entry.address === checksummedRecipientAddress,
+  const getNickName = (address) => {
+    const addressBookEntryObject = addressBook.find((entry) => {
+      return address.toLowerCase() === entry.address.toLowerCase();
+    });
+    return (addressBookEntryObject && addressBookEntryObject?.name) || '';
```
danjm, line 29: I think this can just be `return addressBookEntryObject?.name || '';`

danjm: Actually, I don't think a function is needed here. I think the previous code with the addition of the `toLowerCase` calls should work. That should make this simpler

danjm (changes requested):
just a few suggestions

ui/components/app/transaction-decoding/components/decoding/address/address.component.js
VSaric: @danjm Done [here](https://github.com/MetaMask/metamask-extension/pull/16018/commits/3efda348801dc6c2df0eca58c405556895c51e7f).

ui/selectors/selectors.js
VSaric: @danjm Done [here](https://github.com/MetaMask/metamask-extension/pull/16018/commits/3efda348801dc6c2df0eca58c405556895c51e7f).

ui/selectors/selectors.js
VSaric: @danjm I tried to use `getAddressBookEntryOrAccountName` instead of `getAccountName` but it is not the same design like on figma. Below is ss with `getAddressBookEntryOrAccountName` usage:

Screenshot 2022-10-24 at 14 32 42

So, here it tries to get address from the address book or to get account name if it is exists in `state.metamask.identities`. Return statement is `return entry && entry.name !== '' ? entry.name : address;`, so if the entry and entry.name exists return `entry.name`, but if doesn't exist return `address`, and that `address` is a problem here. 

You can see AC in this ticket: https://github.com/MetaMask/metamask-extension/issues/15780.
I followed that for displaying this addresses and usage from address component.

ui/components/app/transaction-decoding/components/decoding/address/address.component.js
VSaric: @danjm Yes, like as I mentioned in the above comment https://github.com/MetaMask/metamask-extension/pull/16018#discussion_r1003299116 I followed the AC in this ticket: https://github.com/MetaMask/metamask-extension/issues/15780. 

I followed that for displaying this addresses and usages from address component. Let me know if this is the right approach for this ticket.

Thank you in advance!

VSaric:
@georgewrmarshall @danjm @brad-decker Can I get re-review on this? 
Thank you in advance! 😄

ui/selectors/selectors.js
danjm: Okay! That makes sense.

However, then we won't get address book entry names if the user has added them. And I think we should show the user the names they've given an address instead of the address itself.

danjm: oh, but I see that is handled in the address component...

ui/components/app/transaction-decoding/components/decoding/address/address.component.js
danjm: I think the solution is that we shouldn't use the `addressOnly` flag on the address component in `ui/components/app/signature-request/signature-request-message/signature-request-message.js`, and then we don't need to add `recipientName` to this list

ui/selectors/selectors.js
VSaric: @danjm Yes, it is handled in the address component.

ui/components/app/transaction-decoding/components/decoding/address/address.component.js
VSaric: @danjm I will share here screenshots with different situations. 

1) Here I **used** `addressOnly` flag on the address component in `ui/components/app/signature-request/signature-request-message/signature-request-message.js` and here is **added** `recipientName` as one of the options in address component:

Screenshot 2022-11-03 at 15 30 41

2) Here I **used** `addressOnly` flag on the address component in `ui/components/app/signature-request/signature-request-message/signature-request-message.js` and here is **not added** `recipientName` as one of the options in address component:

Screenshot 2022-11-03 at 15 33 20

3) Here I **don't used** `addressOnly` flag on the address component in `ui/components/app/signature-request/signature-request-message/signature-request-message.js`:

Screenshot 2022-11-03 at 15 45 32

So as you can see, **third case** is not valid because it will be displayed `New contract` instead of `shorten address` (it is not like on [Figma design](https://www.figma.com/file/U2Ojf3cJP4PvIYRiRfUya0/Improve-Signatures-(v1)?node-id=1971%3A1763) for resolving this ticket).

So if we want to display `shorten address` like on Figma design, we need to add `addressOnly` flag on the address component in `ui/components/app/signature-request/signature-request-message/signature-request-message.js`.

**Second case** is displaying `shorten address` instead of `account name` (in this case `Account 1`). `Account name` is displayed in the Figma design. 

You can see in this ticket https://github.com/MetaMask/metamask-extension/issues/15780 that `first case` in AC is this: `If the address exists in the user account list, then the account name is displayed (this should be the first thing checked, if it does exist in user account list we don't need to check the rest)`.

**First case** are displaying `account name` because `recipientName` is added as one of the options in address component. 

So I think we must used `addressOnly` flag on the address component in `ui/components/app/signature-request/signature-request-message/signature-request-message.js`, but I can remove `recipientName` as one of the options in address component, but than we will see only `shorten addresses` instead of `account name` if exists. In below screenshot:

Screenshot 2022-11-03 at 16 08 44

`0xcF2...C59E` is my account name (Account 1).

bschorchit: For the third case, where's `New contract` coming from? Is the nickname that you created within Contacts for this address? If yes, we would indeed like to display it here. 

From AC:
> If the address exists in the user address book (contacts), then their nickname is displayed (this should be already existing logic for this component - this should be second thing checked);

bschorchit: Are we using the same component/logic on this other PR https://github.com/MetaMask/metamask-extension/pull/15888 ?

The way I see it, is that we would have a single component carries this logic of displaying account name > account nickname > contract metadata > address (the one described in AC) and it's used everywhere. So if we ever decide to change the logic, we only update code in one place and it get updated everywhere.

bschorchit: Initially my thought was this ticket https://github.com/MetaMask/metamask-extension/issues/15780 would create the new logic for displaying address for this existing address component and this one https://github.com/MetaMask/metamask-extension/issues/15716 (the one we're currently discussing) would make the change to start using that same component within signTypedData_v4 messages. 

Does this makes sense? I was not able to fully follow the technical discussion above around flags and components.

VSaric: > For the third case, where's New contract coming from? Is the nickname that you created within Contacts for this address? If yes, we would indeed like to display it here.

@bschorchit `New contract` is coming from translations `t('newContract')`, it is not the nickname that I created within Contacts for this address.

newlife08:
> just a few suggestions

- [image]()

ui/components/app/transaction-decoding/components/decoding/address/address.component.js
georgewrmarshall: Forgive me for adding noise but for option 3 displaying a new contract. Is there any benefit in displaying the condensed contract address and using some tag to notify the user that it is a new contract?

Here is a rough mock I made with terrible copy

new contract mock

georgewrmarshall (dismissed):
Tested on local and UI lgtm!
I think we could do something about how the space is taken up by each section but that can be addressed in another PR

Screen Shot 2022-11-07 at 10 35 51 AM

ui/components/app/signature-request/signature-request-message/signature-request-message.js
```diff
@@ -0,0 +1,168 @@
...
+                ) : (
+                  `${value}`
+                )}
+              </Typography>
+            )}
+          </Box>
+        ))}
+      </Box>
+    );
+  };
```
brad-decker, line 105: We shouldn't create these functions for intermediary components, just render them inline unless they are rendered multiple times then lift them into a new component.

brad-decker (changes requested):

ui/components/app/transaction-decoding/components/decoding/address/address.component.js
bschorchit: Very interesting @georgewrmarshall! @SayaGT was exploring some designs in the past for showing address insights to users. Let's revisit this again soon.

bschorchit:
> Tested on local and UI lgtm! I think we could do something about how the space is taken up by each section but that can be addressed in another PR
> 
> Screen Shot 2022-11-07 at 10 35 51 AM

Created a new issue for this here: https://github.com/MetaMask/metamask-extension/issues/16405

ui/components/app/signature-request/signature-request-message/signature-request-message.js
VSaric: @brad-decker Done [here](https://github.com/MetaMask/metamask-extension/pull/16018/commits/a89af984b68e0bb7bfb7a65ebdc34d4876113eb9).

brad-decker (approved):

georgewrmarshall (approved):
I completely agree with @brad-decker's suggestion to move the `SignatureRequestData` component to it's own file and will promote this pattern for maintainability in future review. Following this I think we could have a unit test and story for each new component but because it used in `signature-request-message.js` and there is a story for this component I think it's ok. Nice work! 💯

danjm (approved):

PR #16020, opened 2022-09-28
MV3 Segment Fix

### LINKED ISSUE #15972: Fix Segment failing silently

jpuri:
Fixes: #15972

Replacing analytics-node with a custom implementation.

analytics-node is giving issue with MV3 as it uses axios which requires XMLHttpRequest not available in context of service worker. There was already a discussion to remove analytics-node and thus we discussed and decided to write our own implementation for it.


app/scripts/lib/segment/analytics.js
```diff
@@ -0,0 +1,212 @@
+import removeSlash from 'remove-trailing-slash';
```
digiwand, line 1: for some reason, I'm not seeing events after setting up `.metamaskrc` variables, switching the "Participate in MetaMetrics" setting on, turning off blockers, and then interacting with our app

jpuri:
Hey @seaona : can you plz QA this segment fix.

adonesky1:
@jpuri could you call out what, if anything, you've changed from the `analytics-node` implementation other than switching from using `axios` to using `fetch`?

jpuri:
@adonesky1 : there are also some other simplifications in `analytics-node` code, like remove function `group` which we are not using.

darkwing (dismissed):

app/scripts/lib/segment/analytics.js
```diff
@@ -0,0 +1,219 @@
...
+  const characters =
+    'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+  const charactersLength = characters.length;
+  for (let i = 0; i < 20; i++) {
+    result += characters.charAt(Math.floor(Math.random() * charactersLength));
+  }
+  return result;
+};
+
+export default class Analytics {
```
digiwand, line 19: I think this comment is a bit late but might have been nice to add the original file first, and then a new commit with changes. 

do you think it would be helpful if we added a comment to explain where the code originated from and include which version it was?

adonesky1: Agreed. At least for reviewing this would be very helpful

adonesky1:
Also @jpuri what are the manual testing for this PR?

app/scripts/lib/segment/analytics.js
digiwand: I ended up using a random diff checker I found online 

code changes look good to me. Aside from the axios <> fetch swap itself, some things I observed were:  
- `screen` and `alias` methods were removed 
- we are no longer returning `setImmediate(callback);` in `enqueue` if `!this.enable`. This looks like a good clean up to me 
- no more retry logic. Q: why was this removed? @jpuri

adonesky1:
Tested and it seems to work alright, atleast seems to be successfully sending batch requests to Segment.

Wondering did we look at Segment's [Analytics 2.0](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/upgrade-to-ajs2/). I just [glanced at it](https://github.dev/segmentio/analytics-next/) and it seems like maybe it uses fetch?

app/scripts/lib/segment/analytics.js
segun: I wish I saw this comment earlier, I had to open the code side by side. But yeah, code cleanup looks good.

app/scripts/lib/segment/analytics.js
```diff
@@ -0,0 +1,212 @@
...
+  constructor(writeKey, options = {}) {
+    this.writeKey = writeKey;
+
+    this.host = removeSlash(options.host || 'https://api.segment.io');
+    this.flushInterval = options.flushInterval || 10000;
+    this.flushAt = options.flushAt || Math.max(options.flushAt, 1) || 20;
+
+    this.queue = [];
+    this.path = '/v1/batch';
+    this.maxQueueSize = 1024 * 450;
```
segun, line 39: Why are we not using this pattern for `maxQueueSize`?
```suggestion
        this.maxQueueSize = options.maxQueueSize || 1024 * 450 
```

app/scripts/lib/segment/analytics.js
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  /**
+   * Send an identify `message`.
+   *
+   * @param {object} message
+   * @param {Function} [callback] - (optional)
+   * @returns {Analytics}
+   */
+  identify(message, callback) {
```
segun, line 67: Why did we remove the screen and alias message types?

app/scripts/lib/segment/analytics.js
```diff
@@ -0,0 +1,212 @@
...
+
+  /**
+   * Add a `message` of type `type` to the queue and
+   * check whether it should be flushed.
+   *
+   * @param {string} type
+   * @param {object} msg
+   * @param {Function} [cb] - (optional)
+   */
+  enqueue(type, msg, cb) {
```
segun, line 101:
```suggestion
  enqueue(type, msg, callback = noop) {
```

app/scripts/lib/segment/analytics.js
```diff
@@ -0,0 +1,212 @@
...
+    const callback = cb || noop;
```
segun, line 102: We can remove this line if we default default callback to noop in the args

segun (commented):
I left a couple of comments and recommended code changes

app/scripts/lib/segment/analytics.js
```diff
@@ -0,0 +1,212 @@
...
+      });
+    };
+
+    const req = {
+      auth: {
+        username: this.writeKey,
+      },
+    };
+
+    return fetch(`${this.host}${this.path}`, {
```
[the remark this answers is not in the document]
jpuri: A query 😃 ,  how will that improve this function ?

app/scripts/lib/segment/analytics.js
```diff
@@ -0,0 +1,212 @@
...
+      return Promise.resolve();
+    }
+
+    const items = this.queue.splice(0, this.flushAt);
+    const callbacks = items.map((item) => item.callback);
+    const messages = items.map((item) => item.message);
+
+    const data = {
+      batch: messages,
+      timestamp: new Date(),
```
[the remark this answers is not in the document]
jpuri: This is from analytics-node, so I guess that is how segment api requires it to be.

app/scripts/lib/segment/analytics.js
jpuri: Metamask does not uses this parameter.

app/scripts/lib/segment/analytics.js
jpuri: We do not use those.

app/scripts/lib/segment/analytics.js
jpuri: Methods that we do not use in metamask are removed, the file is small subset of what we use.
retry depends on axiosRetry thus it is also removed with axios library.

app/scripts/lib/segment/analytics.js
jpuri: done 👍

app/scripts/lib/segment/analytics.js
jpuri: done 👍

app/scripts/lib/segment/analytics.js
```diff
@@ -0,0 +1,212 @@
...
+    }
+  }
+
+  /**
+   * Flush the current queue
+   *
+   * @param {Function} [cb] - (optional)
+   * @returns {Analytics}
+   */
+  flush(cb) {
```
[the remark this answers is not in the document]
jpuri: done 👍

segun (dismissed):
LGTM

digiwand (dismissed):
I would love to troubleshoot my build to verify segment events have reached segment. As 2 others mention it's working well for them and the code LGTM, I'm approving 👍

app/scripts/lib/segment/analytics.js
jpuri: After discussion with @PeterYinusa , I added retry code to the PR.

jpuri:
@seaona , @PeterYinusa : can you plz test the PR.

app/scripts/lib/segment/analytics.js
```diff
@@ -0,0 +1,264 @@
...
+        }
+        return Promise.resolve(response);
+      })
+      .catch((error) => {
+        if (this._isErrorRetryable(error) && retryNo <= this.retryCount) {
+          this._sendRequest(url, body, callback, retryNo + 1);
+        }
+        callback(error);
+      });
+  }
```
jpuri, line 239: I am finding it hard to test retry, plz share suggesting for QA.

Gudahtt:
>Wondering did we look at Segment's [Analytics 2.0](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/upgrade-to-ajs2/). I just [glanced at it](https://github.dev/segmentio/analytics-next/) and it [seems like it uses fetch](https://github.com/segmentio/analytics-next/blob/317425f09cf0dec4e44e2790dbec41fbe5da8242/packages/node/src/app/plugin.ts#L16)?

That is Analytics.js v2.0, which is the Segment library for websites. IIRC we considered Analytics.js when Segment was first implemented, but determined it wasn't well suited for our use-case. I don't recall why. But it wasn't designed with extensions in mind.

app/scripts/lib/segment/analytics.js
```diff
@@ -0,0 +1,263 @@
...
+    if (!this.enable) {
+      setImmediate(callback);
+      return;
+    }
+
+    const message = { ...msg, type };
+
+    // Specifying library here helps segment to understand structure of request.
+    // Currently segment seems to support these source libraries only.
+    message.context = {
```
PeterYinusa, line 123: I think we need to append rather than overwrite here

PeterYinusa: Cause we're currently losing the app, userAgent and page property of context

adonesky1:
> That is Analytics.js v2.0, which is the Segment library for websites. IIRC we considered Analytics.js when Segment was first implemented, but determined it wasn't well suited for our use-case. I don't recall why. But it wasn't designed with extensions in mind.

Fair enough! Didn't know we'd looked at it before.

seaona:
From QA, Segment requests are working and I am able to see them on the Dashboard successfully. There are some edge cases that need to be further investigated / talk through as discussed in @jpuri 's office hours. So far, I could identify the folllowing:

- Every time service worker is re-start, an "Identify" event is sent. Maybe we should somehow filter out these segment events? Otherwise we'll probably get tons of them, as service worker will be re-started very frequently. @brad-decker 

[identify-event-sw-restart.webm](https://user-images.githubusercontent.com/54408225/195044065-32c4c73d-6918-4e67-8657-aabe982ad186.webm)

- My understanding was that events were sent in batches of 20 max, but I've noticed that after each event is fired, a request is immediately sent right after, so practically we sent 1 request per event at the moment. I am wondering if this is the desired case, or should we try to group them and reduce the amount of requests sent?
On the other hand, having 1 req per event,  results in not being easy to encounter the edge case were SW is terminated and a queue of events is cleared from the batch and never sent.
I'm wondering if that's okay or how should proceed?

[batch-events-not-grouped.webm](https://user-images.githubusercontent.com/54408225/195045230-d1ac2b7e-660c-45b8-91fb-3f6cd7069c2a.webm)

app/scripts/lib/segment/analytics.js
PeterYinusa: looks good now

app/scripts/lib/segment/analytics.js
PeterYinusa: I'm using the AutoResponder feature of [Fiddler EveryWhere](https://www.telerik.com/download/fiddler-everywhere). Let me know if you would like a walkthrough

jpuri:
@seaona : about batching, it is configured to work in production build only.

Multiple `identify` event is definitely an issue we need to address for MV3 but not specifically related to this PR I think.

brad-decker:
@jpuri @seaona -- identify is a special event handled differently than others and won't flood mixpanel. We can resolve it later if it is a problem but segment suggests that websites identify on every page load and after login/logout for traditional websites so its not exactly onerous to do it everytime the background service worker restarts.

brad-decker (approved):
I'd like to see a production test of batching done @seaona but the code looks good to me.

PeterYinusa (approved):

seaona:
@brad-decker @jpuri testing batch queuing and service worker interruptions now

seaona:
- When batching is not interrupted, we see events and requests, grouped as @jpuri mentioned above

[batching-without-interruptions.webm](https://user-images.githubusercontent.com/54408225/195123483-7599cffe-7136-4f62-b244-2efb47710572.webm)

- When I terminate service worker inbetween a batch (with some events, but request not sent yet) the batch is lost and request is never sent. In this case, I was triggering the Main Menu Opened Event in the same way as in the above example, but now requests didn't go through

[batching-with-interruptions.webm](https://user-images.githubusercontent.com/54408225/195123686-871edf22-c5b3-47ed-8e4a-654a8ba68e9c.webm)

brad-decker:
@jpuri @seaona mentioned above that events were *lost* when service worker dies... that should have been a blocker for this PR

PR #16033, opened 2022-09-29
Updating controller dependency

jpuri:
Updating controller

Manual Testing Instructions:
This is a bit tricky to describe because the expected behavior is somewhat complicated. 

### Phishing-detection flow:
What happens now is that when users navigate to any site, we check if the controller has fetched/updated the lists to check against in the last hour. If an update
- **has not happened in the past hour** we initiate a fetch to update the locally stored phishing lists, but we don’t wait for it to resolve before we proceed to check the current domain against the (potentially) stale list. This opens a small window where if the user happens to visit a domain that was added to the list since the last fetch as their first network activity after the wait interval has expired the user will not see the phishing warning page. Once the fetch to update the lists resolves the user will be warned upon any refreshes or new traffic to this domain.
- **has happened in the last hour** we do not make any requests to fetch updated lists.

We expect gaps in coverage for lists newly added to the domain between when things get updated in the phishing lists and when they can get caught: The gap will equal the amount of time between when the domain gets added to the phishing list and how soon after the controllers one hour refresh interval has completed that a user navigates to a new site plus the time that it takes for the fetch of updated phishing lists (if there are any updates) to resolve.

### Manual Testing
**This is a bit tricky to describe because the expected behavior is somewhat complicated.** 

1.  Start a build with the environment variable `IN_TEST=1` this will set the refresh interval to fetch updates to 5 seconds instead of an hour
2. Mock traffic to fetch the phishing lists like we do in the `phishing-detection.spec.js` e2e test
    - Mock this URL= `https://static.metafi.codefi.network/api/v1/lists/eth_phishing_detect_config.json'
'https://static.metafi.codefi.network/api/v1/lists/eth_phishing_detect_config.json`
    - To return something like:
```
 {
    version: 2,
    tolerance: 2,
    fuzzylist: [],
    whitelist: [],
    blacklist: [<add_test_domain_here>],
}
```
3. Navigate to any page so that the fetch will initialize
4. After 5 seconds, navigate to the domain that has been blacklisted
5. You should see the warning page. 

_I would encourage that we test other similar scenarios to understand the gaps in coverage and ensure what I describe above is accurate, but this is the base case._


app/scripts/metamask-controller.js
```diff
@@ -3372,7 +3372,7 @@ export default class MetamaskController extends EventEmitter {
     if (sender.url) {
       const { hostname } = new URL(sender.url);
       // Check if new connection is blocked if phishing detection is on
-      const phishingTestResponse = this.phishingController.test(hostname);
+      const phishingTestResponse = await this.phishingController.test(hostname);
```
Gudahtt, line 3375: Is it possible that this results in connection failures or dropped messages? That was the concern I raised in the controllers PR, that blocking here to download a new phishing config might cause problems.

The actual port should be constructed synchronously, as there are no `await`'s until after that step (in `connectExternal`). So we shouldn't see a timeout in establishing the connection itself, as far as I know.

But I'm less sure about the individual messages. Do they have a timeout?

Gudahtt: Downloading a new config might be slow. On the order of seconds, depending on where you're downloading it from and how fast your connection is.

jpuri:
@rekmarks : `setupUntrustedCommunication` in `MetamaskController` is not async. Can you plz check if [this](https://github.com/MetaMask/metamask-extension/pull/16033/files#diff-6fbff2cfe97ac01b77296ef2122c7e0a5b3ff6a84b584b4d1a87482f35eea3d6R3573) function may need a fix ?

seaona:
It looks like e2e testcase for phishing detection is failing due to the last controllers changes. @Gudahtt do you think we should update the cdns on the `phishing-detection` spec?

https://github.com/MetaMask/metamask-extension/blob/develop/test/e2e/tests/phishing-detection.spec.js#L5

Gudahtt:
Good catch @seaona ! Yes, changing that will be necessary

adonesky1 (dismissed):

Gudahtt:
Has anyone looked at [this comment](https://github.com/MetaMask/metamask-extension/pull/16033#discussion_r987060933)? I haven't had the chance to test this out yet.

adonesky1:
> Has anyone looked at https://github.com/MetaMask/metamask-extension/pull/16033#discussion_r987060933? I haven't had the chance to test this out yet.

Ah hmmm I had glanced at it a week ago and assumed it would be resolved by what we did in the subsequent controllers release, but looking at it now I see that's not the case. 

I will investigate.

app/scripts/metamask-controller.js
adonesky1: @Gudahtt I've tried to simulate a very slow connection here:

https://user-images.githubusercontent.com/34557516/195891473-afb65c9b-5634-4b7d-a1a0-1f6381e4ff21.mov

I've commented out the if conditional to only update phishing lists if they're out of date, so that we await them every time.

With my as slow as it will let me go throttling profile it takes 2 seconds for the configs to load and I'm not seeing issues with dropped messages. 

Does this achieve the level testing confirmation that you were hoping to see?

app/scripts/metamask-controller.js
```diff
@@ -3405,6 +3406,13 @@ export default class MetamaskController extends EventEmitter {
     }
   }
 
+  // Throttle the call to update phishing lists in order to constrain # of requests
+  /// while maintaining non-blocking updates every hour while users are actively browsing
+  throttledUpdatePhishingLists = throttle(
```
adonesky1, line 3411: @Gudahtt sorry I might have missed something from our convo... but couldn't we just achieve this by checking whether the list is out of date according the phishing controller refresh interval (as we were doing previously) and then calling `updatePhishingLists` if it is but not waiting/blocking on it?

Gudahtt: ... yes, that makes sense too! Then we could rely on the interval set in the controller. Good point.

adonesky1: Maybe lets go with that so that the controller just manages its own refresh intervals?

Gudahtt: Agreed

adonesky1: Done [here](https://github.com/MetaMask/metamask-extension/pull/16033/commits/23c58f5d4e00e4c586403e452a050654f7d5f898)

adonesky1 (dismissed):

app/scripts/metamask-controller.js
Gudahtt: Yes, I think so, thanks

danjm (dismissed):
This looks good to me, as long as @Gudahtt's comments from 16 days ago have been properly addressed, which they seem to have been

jpuri:
@danjm : also, there was a suggestion to test this PR before merge. @seaona , @PeterYinusa : can you plz help.

shanejonas (dismissed):

app/scripts/background.js
```diff
@@ -170,7 +170,7 @@ if (isManifestV3) {
 async function initialize(remotePort) {
   const initState = await loadStateFromPersistence();
   const initLangCode = await getFirstPreferredLangCode();
-  await setupController(initState, initLangCode, remotePort);
```
Gudahtt, line 173: Nit: This function does seem to return a Promise, despite being fully synchronous. It returns `Promise.resolve()` at the end. We should probably remove the `return Promise.resolve()` and update the JSDoc comment if we're removing this `await`.

app/scripts/metamask-controller.js
```diff
@@ -475,6 +475,10 @@ export default class MetamaskController extends EventEmitter {
     });
 
     this.phishingController = new PhishingController();
+    if (process.env.IN_TEST) {
+      this.phishingController.updatePhishingLists();
```
Gudahtt, line 479: Does this line need to be in the `IN_TEST` block? I thought it made sense to update the phishing lists right away. That's what the old implementation did as well ([see here](https://github.com/MetaMask/controllers/blob/62ba2a3e8d7959b6dcbc0cf4981209dc21bc04b4/src/third-party/PhishingController.ts#L136))

adonesky1 (approved):
🤝

app/scripts/background.js
```diff
@@ -170,7 +170,7 @@ if (isManifestV3) {
...
+  setupController(initState, initLangCode, remotePort);
```
darkwing, line 173: 👍

darkwing (approved):

Gudahtt (approved):
LGTM!

PR #13886, opened 2022-03-09
Don’t show ‘What’s new’ pop up to new users

VSaric:
fixes: #13278

**Explanation:**  Don’t show ‘What’s new’ pop up to user that have just finished onboarding. New user is the user which have created a new seed phrase. User who imported or pasted the seed phrase has been considered as an existing user and the ‘What’s new’ pop up will be shown to him.

**Screencast for new user (create a new wallet):**

https://user-images.githubusercontent.com/92527393/157413304-f273e61f-1ab4-4bca-8b86-e2b9b3d1d728.mov

**Screencast for existing user (import a wallet):**

https://user-images.githubusercontent.com/92527393/157413512-331abf51-6fec-4235-bcd3-7d19ed2f839e.mov

PR #16042, opened 2022-09-30
Fix: show whats new to users who created, not imported, a new wallet,…

danjm:
… but not on their first session

Fixes a bug introduced in https://github.com/MetaMask/metamask-extension/pull/13886 that would prevent whats new messages from being shown to users who had completed their installs with the "Create a Wallet" option

## Explanation

Code was previously relying on `completedOnboarding` to decide whether to show the user what's new, when if fact we needed a variable like `completedOnboardingInThisUISession`. The goal was to not show the popup to entirely new users. The problem is that `completedOnboarding` is set to true once and stays true forever, so users were not seeing it in any session. This PR adds a new variable to redux state to decide whether to show whats new to users.

## Testing steps

1. Locally build on an old commit. v10.7.0 or something like
2. Create a new seed phrase while onboarding
3. Locally build on develop, reload the extension
4. Close the portfolio dapp tooltip
5. Refresh the page
6. Notice that there are no whats new popups (if you want, log in and out to still see no whats new popup)
7. Build on this PRs branch
8. Reload the extension
9. Log in and see the whats new popup


darkwing (dismissed):
Great solution Dan!

brad-decker (approved):

adonesky1 (approved):
LGTM. Though could you either add it or make a note somewhere to add it to the V2 onboarding flow so we don't regress when we bump to the new flow?

PR #16068, opened 2022-10-04
Update Jest config to avoid transpiling dependencies

Mrtenz:
By default Jest looks at the `module` field of a package before looking at `main`. The module field is often used for the ESM version of a package, which is not supported by our configuration.

Previously we solved this by transpiling all dependencies that had this issue. In this PR I changed it to only look at `main`, which is usually the CJS version of a package. This removes the need for transpiling dependencies, and could speed up the tests slightly.

I also bumped Jest to a stable (non-alpha) version of 29.


FrederikBolding (approved):

brad-decker (approved):
Niiice.

Issue #15987, opened 2022-09-26
MV3: Support dapp phishing warning page


digiwand:
### Describe the bug

This support should piggyback off the support we have for dapp pages.

Update logic to only trigger 45 minute keep alive logic on load.

### Version

10.19.x

### Browser

Chrome

### Operating system

MacOS

PR #16075, opened 2022-10-04
MV3: Update service worker restart logic and keep-alive logic for dapp support

### LINKED ISSUE #15802: MV3: Service Worker should be kept active only after the dapp has been interacted with
### Describe the bug

The app-init.js service worker should only stay alive if the page dapp provider is connected. Currently, the service worker is kept alive as long as the page is open.

See https://github.com/MetaMask/metamask-extension/pull/15494#issuecomment-1245914710 for more details. 

Criteria:
- When user interacts with the dapp, keep service worker alive for 45 minutes

Note that support to connect Service Worker after it has been inactive will be handled in this PR (WIP, Needs update): 
https://github.com/MetaMask/metamask-extension/issues/15986

### Steps to reproduce

1. Close all `https://` or `http://` pages
2. Observe `app-init.js` service worker is not running. If it is still running, terminate the service worker. 
3. Open a `https://` or `http://` page 
4. Observe `app-init.js service worker is running and continues to run as long as the page is open, even when the dapp is not connected

_No response_

### Version

10.18.4

### Build type

_No response_

### Browser

Chrome

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

### LINKED ISSUE #15986: MV3: Support reconnecting Service Worker after inactivity
### Describe the bug

We have two separate PRs to keep the service worker active for a dapp that is connected to an account, essentially, at the start of a page load: 
https://github.com/MetaMask/metamask-extension/pull/15494
https://github.com/MetaMask/metamask-extension/pull/15802

The service worker will go inactive after some time when an account is not connected to the dapp at the beginning of a page load. We need to support restarting the service worker when the page attempts to interact with the service worker again. 

### Steps to reproduce

1. Launch extension MV3
2. Open dapp that is not connected to an account
3. Terminate the `app-init.js` service worker or wait for it to be inactive
4. Attempt to use the dapp
5. Dapp should work, and we can send transactions

_No response_

### Version

10.19.x

### Build type

_No response_

### Browser

Chrome

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

digiwand:
## Explanation

**This PR:**
- update keep alive logic to run for 45 minutes for connected dapps
- reconnects dapp streams after the service worker reconnects 

**Details of Implementation:**
- narrows the scope of the keep-alive logic to run only after (most) JSON-RPC requests are sent and for 45 minutes
- re-activates the service worker after inactivity upon a JSON-RPC request
- updates destroy / set up stream logic so that stream is set up after extension is ready
- send a message to all the tabs from the background to inform the dapp content script that the background is ready and to set up the streams. This isn't ideal. We'd like to only send this message to tabs that are dapps. We plan to revisit this logic in a future PR as we don't yet have another, clear, way to query or filter out tabs with dapps. 

**Other important notes regarding this PR:**
- Currently, interacting with the dapp after re-activating the service worker may take 2 clicks. @jpuri has code that fixes this in a follow-up PR.
- Popup does not work if Service Worker restarts when dapp page is open. This will be handled separately. See https://github.com/MetaMask/metamask-extension/issues/16434
- Phishing Page support will be handled separate. See https://github.com/MetaMask/metamask-extension/issues/15987

**Side note:** 
While I was out for 2 weeks, @jpuri assisted by adding some commits. I have returned with more commits.

## More Information

Fixes: #15986
Fixes: #15802

#### Test Service Worker is not kept alive on non-dapp pages:
- Start the extension using MV3 build (yarn start:mv3)
- Be sure not to have any dapp pages, home.html, or the popup running
- View non-dapp website
- Inspect the app-init.js service worker and terminate the service worker
- Observe the service worker stops running

#### Test Service Worker is kept alive on dapp pages:
- Start the extension using MV3 build (yarn start:mv3)
- View a dapp page
- If you are not already connected, connect to the dapp
- Inspect the app-init.js service worker and terminate the service worker
- Observe the service worker is kept alive

#### Test Service Worker can be reactivated after inactivity:
- Start the extension using MV3 build (yarn start:mv3)
- View a dapp page
- Stop the service worker. Two ways:
  - Do not interact with dapp and wait 45 minutes for the service worker
  - Disconnect from the dapp (disconnect accounts) and manually terminate the service worker
- Observe service worker is not running
- Interact with the dapp to send a JSON-RPC request (currently, this may take a few clicks) 
- Observe the service worker is running again 
(note: There might be a delay to re-activate. @jpuri  is working on a task to fix this. More details to come) 

## Pre-Merge Checklist


app/scripts/background.js
```diff
@@ -102,6 +104,15 @@ const initApp = async (remotePort) => {
 
 if (isManifestV3) {
   browser.runtime.onConnect.addListener(initApp);
+  // Message below signals content script in DAPPS to connect to metamask background as backend is not active
+  // It is required to re-connect DAPPS after service worker re-activation
+  chrome.tabs.query({ active: true, currentWindow: true }, function (tabs) {
+    chrome.tabs.sendMessage(
+      tabs[0].id,
+      { name: 'METAMASK_EXTENSION_READY' },
```
ryanml, line 112: Would `'METAMASK_EXTENSION_READY'` be helpful as a shared constant?

digiwand: @ryanml does this work?

digiwand: @ryanml does this work? [b9fdffa](https://github.com/MetaMask/metamask-extension/pull/16075/commits/b9fdffa6ac0c3ef4a118b0daeb804e9ea5b50a9b)

digiwand: Other messages we could add into this new constant: 
- `WORKER_KEEP_ALIVE_MESSAGE`
- `MESSAGE_TEXT` 

to do in follow-up PR

app/scripts/contentscript.js
```diff
@@ -82,6 +80,52 @@ function injectScript(content) {
...
+const TIME_45_MIN_IN_MS = 45 * 60 * 1000;
+
+/**
+ * Don't run the keep worker alive logic for initial JSON RPC methods.
+ * We should not run keep-alive logic for non-dapp pages.
+ */
+const IGNORE_INIT_METHODS_FOR_KEEP_ALIVE = [
+  MESSAGE_TYPE.GET_PROVIDER_STATE,
+  MESSAGE_TYPE.SEND_METADATA,
+];
```
jpuri, line 98: @digiwand : can you plz detail why we are excluding initial JSON RPC methods here ?

app/scripts/contentscript.js
```diff
@@ -82,6 +80,52 @@ function injectScript(content) {
...
+    clearTimeout(keepAliveTimer);
+  }
+
+  keepAliveTimer = setTimeout(() => {
+    clearInterval(keepAliveInterval);
+  }, TIME_45_MIN_IN_MS);
+
+  if (keepAliveInterval) {
+    clearInterval(keepAliveInterval);
+  }
```
jpuri, line 118: `if` above and line 108 looks redundant.

app/scripts/contentscript.js
```diff
@@ -82,6 +80,52 @@ function injectScript(content) {
...
+
+  browser.runtime.sendMessage({ name: WORKER_KEEP_ALIVE_MESSAGE });
```
jpuri, line 120: above send message is required only if keepAliveInterval was not already running in other words extension was dead

jpuri (commented):
LGTM except for minor feedbacks.

app/scripts/contentscript.js
digiwand: @jpuri 

just above these lines is the comment:
```
/**
 * Don't run the keep worker alive logic for initial JSON RPC methods.
 * We should not run keep-alive logic for non-dapp pages.
 */
 ```
 
 is there something you could recommend me to add or update?

File: app/scripts/contentscript.js (hunk 3 above)
Comment (digiwand) [reply app/scripts/contentscript.js:118]: one is for the interval, and the other is for the timer

File: app/scripts/contentscript.js (hunk 4 above)
Comment (digiwand) [reply app/scripts/contentscript.js:120]: The thing is, I added this in so that we could send a message immediately, as the one in setInterval will run after the interval has passed.

I'm still a fan of keeping it in for the purpose I just mentioned, but I can remove it if you think it would be better. I could also add a comment elaborating on this line. What do you think?  @jpuri

File: app/scripts/background.js (hunk 1 above)
Comment (ryanml) [reply app/scripts/background.js]: works for me! thank you :)

**Review by ryanml — DISMISSED** [review]

File: app/scripts/contentscript.js (hunk 3 above)
Comment (jpuri) [reply app/scripts/contentscript.js:118]: I think 
```
if (keepAliveInterval) {
    clearInterval(keepAliveInterval);
  }
```

Or just this `clearInterval(keepAliveInterval);`

will be same.

File: app/scripts/contentscript.js (hunk 4 above)
Comment (jpuri) [reply app/scripts/contentscript.js:120]: lets leave it there 👍

File: app/scripts/contentscript.js (hunk 2 above)
Comment (jpuri) [reply app/scripts/contentscript.js:98]: I do not understand `initial JSON RPC methods` part.

**Review by jpuri — DISMISSED** [review]
Changes look good, I would suggest QA for these changes in both MV3 and MV2.

File: app/scripts/contentscript.js (hunk 3 above)
Comment (digiwand) [reply app/scripts/contentscript.js]: updated [ae3c2df](https://github.com/MetaMask/metamask-extension/pull/16075/commits/ae3c2dff4d60efa06bcba80799f8133f3d90493f)

File: app/scripts/contentscript.js (hunk 2 above)
Comment (digiwand) [reply app/scripts/contentscript.js:101]: @jpuri 

I see. good callout. what about this [073f05e](https://github.com/MetaMask/metamask-extension/pull/16075/commits/073f05ec972fd9f09aee24f1302d2f566c2bc3fc)?

File: app/scripts/background.js (hunk 1 above)
Comment (digiwand) [reply app/scripts/background.js]: sweet :) np! Thanks for asking @ryanml

File: app/scripts/background.js [hunk 5]
```diff
@@ -101,6 +104,15 @@ const initApp = async (remotePort) => {
 
 if (isManifestV3) {
   browser.runtime.onConnect.addListener(initApp);
+  // Message below signals content script in DAPPS to connect to metamask background as backend is not active
+  // It is required to re-connect DAPPS after service worker re-activation
+  chrome.tabs.query({ active: true, currentWindow: true }, function (tabs) {
```
Gudahtt, line 109: Note that we should be checking for errors after any extension API comment (e.g. check `browser.runtime.lastError`), because that's how errors are "throw" for the Chrome extension API. That applies here and to `sendMessage`

app/scripts/background.js
```diff
@@ -1,3 +1,5 @@
+/* global chrome */
```
Gudahtt, line 1: Could we use the Mozilla polyfill here rather than using the `chrome` global directly?

jpuri: The polyfill we are using `browser` does not seems to support this specific api `chrome.tabs`

seaona:
Some findings/comments from QA:
1. **Test Service Worker is not kept alive on non-dapp pages**: I see that after terminating the Service Worker, if I continue browsing a non-dapp site, the Service Worker is then re-started. I have MM window and popup closed. Is this expected?

[sw-restart-nondapp.webm](https://user-images.githubusercontent.com/54408225/195810632-222c46b1-2ab9-4d2d-a537-8c2f710b2c58.webm)

2. **Test Service Worker is kept alive on dapp pages**: 
    - I confirm, service worker is automatically restarted every time I terminated it when I am connected to a Dapp.
    - If I want to continue interacting with the Dapp, I need to refresh the Dapp page and connect again, otherwise I cannot trigger any action on MM. Is this expected?

[sw-dapp-interactions.webm](https://user-images.githubusercontent.com/54408225/195815815-60bd4fb7-89de-4bdd-8d62-c91c2149a0b9.webm)

3. **Test Service Worker can be reactivated after inactivity:** I am not sure about this, when I disconnect from a Dapp and terminate Service Worker, it is automatically re-started already, since I have already a Dapp open, as per (2). Should I close the Dapp as well, after terminating Service Worker?

4. **Error in Event Handler**: when I go to any Dapp, without ever interacting with it, having a fresh MM install + onboarded, I see this error on the console from the Dapp frontend: `Error in event handler: ReferenceError: msg is not defined at extensionStreamMessageListener (<URL>)`

[image
](file:///home/seaona/Pictures/Screenshots/Screenshot%20from%202022-10-14%2011-34-32.png)

app/scripts/background.js
Gudahtt: I'm pretty sure it does. We already use it, in `app/scripts/platform/extension.js`

jpuri:
@digiwand : I have extracted out DAPP action replay part to separate PR to unblock this one: https://github.com/MetaMask/metamask-extension/pull/16250

It also requires this PR to be merged and deployed: https://github.com/MetaMask/json-rpc-middleware-stream/pull/30

jpuri:
@seaona : this error should be fixed now `Error in event handler: ReferenceError: msg is not defined at extensionStreamMessageListener (<URL>)`

app/scripts/background.js
digiwand: Using the polyfill LGTM. updated here https://github.com/MetaMask/metamask-extension/pull/16075/commits/00471464812bc33f37510624743ec6698798d590

app/scripts/background.js
digiwand: @Gudahtt 

added checks for `lastError` here https://github.com/MetaMask/metamask-extension/pull/16075/commits/bbd5cffde0503e41e1056a045d2912981675af18

digiwand:
@seaona Thanks for the QA! Is it possible to get another QA on this 👉🏼👈🏼
1. I'm not seeing this behavior. You're correct that this should not be happening. This may have been fixed with the latest changes. In case it is still occurring, please double-check that no other windows open with the dapp.
2. I believe this has been fixed. It was an issue when the `EXTENSION_MESSAGES.READY` logic was introduced. Instead of sending the message to only the active tab of the current page, we are now sending it to all pages.
3. To test this have you tried:
   ```
   Stop the service worker. Two ways:
      - Do not interact with dapp and wait 45 minutes for the service worker
      - Disconnect from the dapp and manually terminate the service worker
   ```
4. should be resolved based on @jyoti's message above https://github.com/MetaMask/metamask-extension/pull/16075#issuecomment-1286706732

@jpuri some recent updates:
- send the EXTENSION_MESSAGES.READY message to all of tabs instead of only the active of the current tab
-  return Promise.resolve or false in the onMessage listener(ref: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage#addlistener_syntax)
- apply recent to the phishing streams logic 
- update comments and variable names

cc: @Gudahtt

app/scripts/contentscript.js
```diff
@@ -391,23 +492,27 @@ function logStreamDisconnectWarning(remoteLabel, error) {
...
-        data: {
-          jsonrpc: '2.0',
-          method: 'METAMASK_EXTENSION_STREAM_CONNECT',
-        },
-      },
-    },
-    window.location.origin,
-  );
-}
+// function extensionStreamMessageListener(msg) {
```
digiwand, line 497: the commented out lines related to `extensionStreamMessageListener` will be handled by @jpuri's PR here: https://github.com/MetaMask/metamask-extension/pull/16250

seaona:
Thank you for your comments and going through the issues @digiwand @jpuri . I've done another regression:

1. **Test Service Worker is not kept alive on non-dapp pages**: the behavior I was mentioning is now fixed. When I terminate the SW and open  a new non-dapp window, service worker is not re-started. 
I've notice a new behavior related to this point. If, after opening a non-dapp window,  I open another browser window, now I see service worker is active again. So basically, service worker is restarted when opening **2** non-dapp windows. Steps for repro-ing this are:
    1. Load MM - check service worker is running
    2. Close all windows (except developer tools / service worker inspector)
    3. Terminate the service worker - check that service worker is not running
    4. Open a new browser tab and go to a non-dapp website (ie. medium.com) - check that service worker is not running
    5. Open a new browser tab - see that service worker is re-started again. It's strange because I don't see any logs from the re-start on the console, but I am seeing it on the service worker inspector

https://user-images.githubusercontent.com/54408225/198822686-02a7bcf2-df33-4c0e-95dd-26fe5d617dbe.mp4

2. **Test Service Worker is kept alive on dapp pages**: here I see a slightly different thing now. If I connect to a Dapp and kill the service worker, it is re-started automatically as expected, but when I try to interact again with the dapp I see the error `receiving end does not exist`. I kind of recall that this is already a known error? If I refresh the browser, then everything continues to work normally

https://user-images.githubusercontent.com/54408225/198823048-3f84c7aa-00d5-477d-a428-5075224882b9.mp4

3. **Test Service Worker can be reactivated after inactivity**: oh I believe I see what you mean here. I confirm it is indeed reactivated after inactivity / service worker killed manually.

4. **Error in Event Handler**:  I see this fixed

digiwand:
@seaona 

Thank you very much for the regression test!

1. This is good. It's expected that the service worker starts again when a new tab is open. The extension runs for all pages and injects the provider into all pages.  The service worker should only be kept alive for dapp pages which it appears to be doing.

2. hm... yes, this is a bug. I wasn't seeing this when I tested this, so I think this happens occasionally. Thank you for finding this 

3. 👍🏼
4. 👍🏼

digiwand:
TLDR; This PR is good to review. We are handling case # 2 @seona found above in a follow-up PR.

---

After syncing today, @jpuri, @seaona and I learned that issue # 2 that @seaona observed happens specifically -> after terminating the service worker, interacting with the popup before interacting with the dapp will cause the dapp to break as it does not receive the message to revive the streams. We will handle this in a separate issue and PR: https://github.com/MetaMask/metamask-extension/issues/16327

app/scripts/background.js
```diff
@@ -91,16 +92,60 @@ const PHISHING_WARNING_PAGE_TIMEOUT = ONE_SECOND_IN_MILLISECONDS;
...
+ * Sends a message to the dapp(s) content script to signal it can connect to MetaMask background as
+ * the backend is not active. It is required to re-connect dapps after service worker re-activates.
+ * For non-dapp pages, the message will be sent and ignored.
+ */
+const sendReadyMessageToTabs = async () => {
+  try {
+    const tabs = await browser.tabs.query({});
+    throwErrorIfLastErrorFound();
+
+    /* TODO we should only sendMessage to dapp tabs, not all tabs. */
```
legobeat, line 121: Could this (sending to all tabs unconditionally) be a metadata leak (indicating that the user has MetaMask installed and active)?

app/scripts/contentscript.js
```diff
@@ -156,16 +210,32 @@ const destroyPhishingExtStreams = () => {
...
+ * This listener/callback receives the message to set up the streams after service worker in-activity.
+ *
+ * @param {object} msg
+ * @param {string} msg.name - custom property and name to identify the message received
+ * @returns {Promise|false}
+ */
+const onMessageSetUpPhishingStreams = (msg) => {
+  if (msg.name === EXTENSION_MESSAGES.READY) {
+    setupPhishingExtStreams();
+    return Promise.resolve(
```
legobeat, line 234: Is there an established style preference for this pattern?

cf https://github.com/MetaMask/metamask-extension/pull/16221

app/scripts/contentscript.js
```diff
@@ -156,16 +210,32 @@ const destroyPhishingExtStreams = () => {
...
+      `MetaMask: handled "${EXTENSION_MESSAGES.READY}" for phishing streams`,
+    );
+  }
+  return false;
```
legobeat, line 238: Does the return value `false` make a difference vs implicitly (not) `return`ing `undefined`?

app/scripts/contentscript.js
```diff
@@ -205,6 +283,7 @@ const setupPageStreams = () => {
 
 const setupExtensionStreams = () => {
   extensionPort = browser.runtime.connect({ name: CONTENT_SCRIPT });
+  // extensionPort.onMessage.addListener(extensionStreamMessageListener);
```
legobeat, line 286: Dead code
```suggestion
```

app/scripts/contentscript.js
```diff
@@ -331,19 +420,32 @@ const destroyLegacyExtensionStreams = () => {
...
+ * @param {string} msg.name - custom property and name to identify the message received
+ * @returns {Promise|false}
  */
-const resetStreamAndListeners = () => {
-  extensionPort.onDisconnect.removeListener(resetStreamAndListeners);
+const onMessageSetUpExtensionStreams = (msg) => {
+  if (msg.name === EXTENSION_MESSAGES.READY) {
+    setupExtensionStreams();
+    setupLegacyExtensionStreams();
+    return Promise.resolve(`MetaMask: handled ${EXTENSION_MESSAGES.READY}`);
```
legobeat, line 434: See above

app/scripts/contentscript.js
legobeat: Perhaps adding a `//TODO: ` with a reference is warranted?  :)

app/scripts/contentscript.js
digiwand: Thanks for the link. I dig the proposed `async` > `Promise.resolve` pattern

The thing with the onMessage callback is that we either want to return a Promise, boolean, or undefined as mentioned in the [onMessage doc](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage), so that we can handle specific message types: 

> If you only want the listener to respond to messages of a certain type, you must define the listener as a non-async function, and return a Promise only for the messages the listener is meant to respond to — and otherwise return false or undefined:

app/scripts/contentscript.js
digiwand: see https://github.com/MetaMask/metamask-extension/pull/16075#discussion_r1011991847

app/scripts/contentscript.js
digiwand: @legobeat 

see https://github.com/MetaMask/metamask-extension/pull/16075#discussion_r1011991847

We could return undefined, though I'd like to return false to be more explicit here

app/scripts/contentscript.js
digiwand: see https://github.com/MetaMask/metamask-extension/pull/16075#discussion_r1011961838 and commit to remove comment [1b9e298](https://github.com/MetaMask/metamask-extension/pull/16075/commits/1b9e2983756c77d1fffb16ba20ae77b757eb7488)

(sorry! missed out on adding your suggestion as a co-author here)

app/scripts/background.js
```diff
@@ -91,16 +92,60 @@ const PHISHING_WARNING_PAGE_TIMEOUT = ONE_SECOND_IN_MILLISECONDS;
...
+            e.message ===
+            'Could not establish connection. Receiving end does not exist.'
+          ) {
+            /** Safely ignore this error, as it means the tab may not be a dapp. */
+          } else {
+            throw new Error(e);
+          }
+        })
+        .finally(() => {
+          throwErrorIfLastErrorFound();
```
jpuri, line 138: May be we can just log error here, we do not need to re-throw.

app/scripts/background.js
```diff
@@ -91,16 +92,60 @@ const PHISHING_WARNING_PAGE_TIMEOUT = ONE_SECOND_IN_MILLISECONDS;
...
+
+/**
+ * Sends a message to the dapp(s) content script to signal it can connect to MetaMask background as
+ * the backend is not active. It is required to re-connect dapps after service worker re-activates.
+ * For non-dapp pages, the message will be sent and ignored.
+ */
+const sendReadyMessageToTabs = async () => {
+  try {
+    const tabs = await browser.tabs.query({});
+    throwErrorIfLastErrorFound();
```
jpuri, line 119: We can probably create a `browser...` utility that iinternally takes care to `throwErrorIfLastErrorFound`.

app/scripts/contentscript.js
```diff
@@ -82,6 +80,51 @@ function injectScript(content) {
...
 }
 
+/**
+ * SERVICE WORKER LOGIC
+ */
+
+const WORKER_KEEP_ALIVE_INTERVAL = 1000;
+const WORKER_KEEP_ALIVE_MESSAGE = 'WORKER_KEEP_ALIVE_MESSAGE';
+const TIME_45_MIN_IN_MS = 45 * 60 * 1000;
+
```
jpuri, line 90: We should keep service worker alive even if account is not connected on DAPP I think.  There are some queries that user can make even in disconnected state.

app/scripts/contentscript.js
```diff
@@ -331,19 +419,32 @@ const destroyLegacyExtensionStreams = () => {
...
+  if (msg.name === EXTENSION_MESSAGES.READY) {
+    setupExtensionStreams();
+    setupLegacyExtensionStreams();
+    return Promise.resolve(`MetaMask: handled ${EXTENSION_MESSAGES.READY}`);
+  }
 
-  destroyExtensionStreams();
-  setupExtensionStreams();
+  return false;
+};
```
jpuri, line 438: Hey @digiwand : I know it is complex, but could you test phishing stream and legacy stream changes.

jpuri:
@digiwand : we need to look more into issue 2 reported by @seaona [here](https://github.com/MetaMask/metamask-extension/pull/16075#issuecomment-1295777617) as she is able to replicate this on PR branch.

adonesky1:
@digiwand sorry if I missed the background context behind the decision here but what is the reason we are attempting to keep the service worker alive for 45 mins (I had previously thought this wasn't possible... I thought I'd tried this pinging it with messages to keep it alive...?) rather than just reviving the SW upon messages from the dapp and replaying them once the SW is initialized?

digiwand:
hey @adonesky1, 

no sorries needed. We decided to keep the service worker alive for some time because when the service worker goes down, we need to restart the streams and this takes some time. We've observed it may take, maybe, 1-10 seconds for the service worker, our controller, and streams to be ready for the dapp. We want to avoid this wait time that users may experience.

pinging the SW using sendMessage can keep the SW alive. Keeping the service worker alive only runs for dapps

app/scripts/contentscript.js
legobeat: That makes perfect sense, thanks for laying that out. Would you be open to changing from `return false` or `return undefined` below? Seems more semantically accurate and would make the intention more obvious to future readers not aware of the nuance of boolean being a specific case.

app/scripts/contentscript.js
legobeat: Slightly similar point to https://github.com/MetaMask/metamask-extension/pull/16075#discussion_r1011118652, something to keep in mind when settling the above:

If the user is interacting with MetaMask on a dapp on one page, this should not result in events/change of behavior readable from an unauthorized (disconnected) tab/page as this would both leak entropy facilitating fingerprinting as well as help in detecting user activity (e.g distinguishing from "visiting browser has MetaMask installed" vs "visiting user has opened MetaMask during this session".

digiwand:
I initially wanted this PR only to handle updating the keep service worker alive logic. Things happened, and now it's combined with more logic and is a branch other PRs depend on. In retrospect, I think it would have been better to have created a trunk branch to handle the dapp <> service worker code

app/scripts/contentscript.js
digiwand: I like this reasoning! what if we save some lines and let the function return undefined by default? 

updated: [bb5ba6b](https://github.com/MetaMask/metamask-extension/pull/16075/commits/bb5ba6b2e1c4d066a96fc8d1f4885d6d1cb6f2b9) and [0f9ff27](https://github.com/MetaMask/metamask-extension/pull/16075/commits/0f9ff27c0510ab78d6fd3f76007ba05abdba0b4a)

digiwand: @legobeat 

I like this suggestion! thank you. updated the code to return `undefined` instead of `false`

app/scripts/background.js
jpuri: Not blocker for this PR.

app/scripts/background.js
digiwand: @jpuri 

Good ideas. I created a new method, `checkForErrorAndLog` in the `metamask-extension/app/scripts/lib/util.js` file to use. 

I also removed the catch statement here. I discovered that the "runtime.lastError: Could not establish connection." error was being thrown here because we were querying all tabs, including tabs without URLs, e.g., devTool tabs like `chrome://extensions`. These tabs were unable to receive the messages. 

With the latest changes, I'm no longer seeing the "lastError: Could not establish connection." error on the background page

app/scripts/background.js
digiwand: updated. see https://github.com/MetaMask/metamask-extension/pull/16075#discussion_r1013279850

app/scripts/contentscript.js
digiwand: @jpuri 
> There are some queries that user can make even in a disconnected state.
I'm not sure I know what queries you mean here. Is this something you can elaborate on? 

The keep-alive logic starts for any dapp interaction, whether or not accounts are connected. On initial load, we only trigger the keep-alive logic if the dapp connects or if the dapp sends requests. Initially, I don't think there is a way to distinguish a dapp from a non-dapp page aside from connected accounts... unless I'm missing something. I actually prefer to avoid running the keep-alive logic on non-dapps or when the provider is not in use on a page to save CPU and memory on the browser.

@legobeat getting to these comments after I run some more tests sending messages to/from our content script and extension

app/scripts/background.js
digiwand: @legobeat  reply for this comment and https://github.com/MetaMask/metamask-extension/pull/16075#discussion_r1012370017. Good question

I believe that we're not leaking any metadata here. Content scripts run in an [isolated world](https://developer.chrome.com/docs/extensions/mv3/content_scripts/#isolated_world). 
> An isolated world is a private execution environment that isn't accessible to the page or other extensions.

Even if we are sending messages to all MetaMask content scripts across all pages, the pages themselves cannot manually read the data or messages from the content scripts. 

Some other things to note: 
- our content script does not accept messages from external extensions or from web pages
- our content script does not send messages outside of our extension or channels

cc: @Gudahtt @jpuri

app/scripts/contentscript.js
digiwand: @legobeat I added a reply here https://github.com/MetaMask/metamask-extension/pull/16075#discussion_r1013903709

app/scripts/background.js
legobeat: Very educative, thank you!

app/scripts/background.js
```diff
@@ -45,7 +46,7 @@ import rawFirstTimeState from './first-time-state';
 import getFirstPreferredLangCode from './lib/get-first-preferred-lang-code';
 import getObjStructure from './lib/getObjStructure';
 import setupEnsIpfsResolver from './lib/ens-ipfs/setup';
-import { getPlatform } from './lib/util';
+import { checkForErrorAndLog, getPlatform } from './lib/util';
```
newlife08, line 49: 0x70c96A85366Fa363ccA7636714a1F497e0714f56

newlife08 (approved):

app/scripts/background.js
Gudahtt: >indicating to the page on any tab that the user has MetaMask installed and active

We do currently leak that a user has MetaMask installed, unfortunately. It's a consequence of our dapp API, which uses an injected provider (the `window.ethereum` object). We are considering moving away from that pattern, but it'll be something we'll have to support for a very long time for backwards compatibility even if we start moving away from it.

But luckily no details of the users accounts are exposed by default.

digiwand:
problem # 2 mentioned [here](https://github.com/MetaMask/metamask-extension/pull/16075#issuecomment-1295777617) should be resolved by the latest 2 commits: https://github.com/MetaMask/metamask-extension/pull/16075/commits/de1e4416a0dec2f1e6ccee778d1662684b1ded3f and https://github.com/MetaMask/metamask-extension/pull/16075/commits/931f5110a626f2f7f1782a5231f6f1344fc7b282

@jyoti, @seaona, and I hopped on a call earlier to debunk the issue. We observed that the dapp will not work on the first load after reloading the extension. This was caused because setupExtensionStreams was called twice: once on contentscript load and another time when the background was loaded and sent the READY message. Calling `runtime.connect` again would also call `runtime.disconnect` and cause the streams, and thus the dapp, to break. We believe this issue is related to the popup issue. This is now fixed by only connecting and setting up the streams if they are not yet set.

We also observed that any errors thrown in contentscript seem to break the dapp. We should check all runtime API calls for lastErrors according to its documentation. When we check for `runtime.lastError` in the contentscript or the provider, we should now throw the errors and, instead, log them as warning messages.

app/scripts/background.js
```diff
@@ -94,16 +95,47 @@ const WORKER_KEEP_ALIVE_MESSAGE = 'WORKER_KEEP_ALIVE_MESSAGE';
...
+      })
+      .then(() => {
+        checkForErrorAndLog();
+      });
+  }
+};
+
 if (isManifestV3) {
   browser.runtime.onConnect.addListener(initApp);
+  sendReadyMessageToTabs();
```
Gudahtt, line 138: Should this wait until after initialization? I am not sure if it's possible for this to result in dapps trying to connect too early before initialization has begun, resulting in dropped connection attempts

app/scripts/background.js
```diff
@@ -94,16 +95,47 @@ const WORKER_KEEP_ALIVE_MESSAGE = 'WORKER_KEEP_ALIVE_MESSAGE';
...
+/**
+ * Sends a message to the dapp(s) content script to signal it can connect to MetaMask background as
+ * the backend is not active. It is required to re-connect dapps after service worker re-activates.
+ * For non-dapp pages, the message will be sent and ignored.
+ */
+const sendReadyMessageToTabs = async () => {
+  const tabs = await browser.tabs
+    .query({
+      // Query tabs with a url. New tabs and DevTool tabs have no url value. e.g. chrome://extensions, chrome://inspect
+      url: '<all_urls>',
```
Gudahtt, line 116: The docs seem to indicate that this is ignored if we don't have the `tabs` permission: https://developer.chrome.com/docs/extensions/reference/tabs/#method-query

seaona:
I can confirm this is fixed now @digiwand @jpuri :
> This is now fixed by only connecting and setting up the streams if they are not yet set.

In terms of behavior this PR looks good to me.

One last observation on my end is that I am getting the console error `Could not establish connection. Receiving end does not exist.  at wrappedSendMessageCallback` when I open the pop-up, after terminating service worker.


https://user-images.githubusercontent.com/54408225/200289901-dcf4b6fd-83a5-4926-9745-2aba59f0cfc6.mp4

app/scripts/background.js
legobeat: Indeed unfortunate. *If* this would have been a leak, it would provide additional fingerprinting entropy and be a proxy for user activity and whether the user is an *active* MM user. Fortunately that does not seem to be a concern here. +1 on a more privacy-respecting future API.

app/scripts/background.js
digiwand: It seems so from the docs, but it appears to work. 

I updated the code with a comment as to how and why it works:
```
      /**
       * Only query tabs that our extension can run in. To do this, we query for all URLs that our
       * extension can inject scripts in, which is by using the "<all_urls>" value and __without__
       * the "tabs" manifest permission. If we included the "tabs" permission, this would also fetch
       * URLs that we'd not be able to inject in, e.g. chrome://pages, chrome://extension, which
       * is not what we'd want.
       *
       * You might be wondering, how does the "url" param work without the "tabs" permission?
       *
       * @see {@link https://bugs.chromium.org/p/chromium/issues/detail?id=661311#c1}
       *  "If the extension has access to inject scripts into Tab, then we can return the url
       *   of Tab (because the extension could just inject a script to message the location.href)."
       */
```

Thanks for pointing this out @Gudahtt. I didn't fully realize why the `chrome://` URLs weren't being returned here with `url: '<all_urls>',` until I dug into this more. It was difficult to find documentation that elaborated on this, but the [comments written by a chromium dev (above @link)](https://bugs.chromium.org/p/chromium/issues/detail?id=661311#c1) helped.

digiwand:
@seaona Thank you again! 🙏  

I'm looking into this. We can recreate this error by running two or more dapps using similar repo steps. When I add a breakpoint to the callback of the `onDisconnect` callback, I'm not observing this error, which makes me suspect this is a race condition. 

Investigation in progress.

**console logs with error:**
Screen Shot 2022-11-08 at 03 34 32

**console logs without error:**
Screen Shot 2022-11-08 at 03 34 23

app/scripts/background.js
digiwand: note: I attempted to add a comment into the manifest.json, but unfortunately, this causes a warning found through chrome://extensions > Errors

Screen Shot 2022-11-08 at 00 18 54

Gudahtt: Great find @digiwand ! And I appreciate the detailed comment, that's very helpful

seaona:
**Testing Phishing Detection on top of this PR:**

- I've seen that when we load MM, there is a service worker dedicated to phishing detection, coming from the phishing-warning page. I guess that's okay, but double-checking just in case
- I've seen that if I terminate the MM service worker, and I go to a phishing page, **the phishing detection stops working** and I land into the malicious page, instead of on the phishing warning page

https://user-images.githubusercontent.com/54408225/200809109-44a48ede-7ea8-4731-8b8e-2b8a53e43231.mp4

**Repro steps:**
1. Load MM
2. Open a malicious page: https://ethereum-mask.com/ -- see that you are redirected to MM Phishing Detection
3. Close the malicious page
4. Terminate MM service worker
5. Open a malicious page: https://ethereum-mask.com/ -- see that now you land into the malicious page

app/scripts/background.js
digiwand: ah yes. the injected provider does leak that the user has installed MetaMask. I meant from this specific code, we are not leaking the data. sorry, I should have clarified this earlier.

Thanks for checking in on these concerns and sharing your thoughts! @legobeat 

> +1 on a more privacy-respecting future API

+1

app/scripts/contentscript.js
digiwand: Thanks @seaona for taking this head on! As discussed earlier, we will handle the support for the phishing stream logic in a separate PR (see https://github.com/MetaMask/metamask-extension/issues/15987)

@seaona has found that terminating the service worker breaks the phishing warning page logic

app/scripts/lib/util.js
```diff
@@ -169,6 +192,8 @@ export {
   hexToBn,
   BnMultiplyByFraction,
   checkForError,
+  checkForErrorAndLog,
+  checkForErrorAndWarn,
```
digiwand, line 196: We're adding both `checkForErrorAndLog` and `checkForErrorAndWarn` here because we found that throwing errors in the content script may cause the dapp to break, so we should only warn of errors for content script(s). We'll continue logging errors per usual in the background. 

---

[According to docs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/lastError), we're expected to check the lastError in the runtime API callbacks

digiwand:
The breaking chrome and firefox flapper test, related to `test/e2e/tests/chain-interactions.spec.js`, appears to be caused by other code outside of this PR. I was able to reproduce the failing 2 tests on `develop` with chrome and firefox. 

> `yarn test:e2e:single test/e2e/tests/chain-interactions.spec.js --browser=chrome`

Screen Shot 2022-11-11 at 00 18 21

cc: @jpuri @Gudahtt @seaona

seaona:
This is what I see manually testing FireFox on MV2: `missing provider`

https://user-images.githubusercontent.com/54408225/201174076-6bbe4dab-8649-4a1c-89ad-9ac3390e9f0f.mp4

app/scripts/background.js
jpuri: We do not need backend to be complete initialised for DAPPs to connect, onConnect listener is required by the DAPPS to connect which is added at line 137.

jpuri: In a separate task may be we can work to improve the current initialisation logic. It definitely has scope for improvements.

digiwand: It took me a bit to wrap my head around this too. Essentially the `sendMessage` from this function calls `connect` in `contentscript` and/or`ui.js`,  which then triggers the `onConnect` handler above this line. Then we initialize the app. Thanks for helping me understand this better too earlier @jpuri! 🙏

@Gudahtt

digiwand:
@seaona re: This is what I see manually testing FireFox on MV2: missing provider

I've seen the missing provider issue as well, and I have heard this error has also been found on MV2. This PR does not update the provider injection logic. We should create a new ticket for this one if it does not exist already since it is separate.

I think it's good you pointed this out 🙏  Thank you

Gudahtt (approved):
LGTM!

brad-decker (approved):
Great documentation, @digiwand Thank you! LGTM

legobeat (approved):

digiwand:
Thanks for the reviews! After @jpuri reviews, we can merge. 🙏 
 
I added jest tests for the new browser-runtime utils here (new PR): https://github.com/MetaMask/metamask-extension/pull/16483

cc: @brad-decker @jpuri @Gudahtt @legobeat

PR #16096, opened 2022-10-05
Update `eth-json-rpc-middleware` to v9.0.1

Gudahtt:
This update includes fixes for our `block-ref` and `retry-on-empty` middleware.

The `block-ref` middleware resolves the block reference `latest` to a specific block number, the latest one we are aware of. This is meant to protect against situations where the network gives inconsistent answers for what the latest block number is due to some nodes being out-of-sync with each other (this was a frequent problem years ago with Infura).

It was broken in that the `latest` resolution was failing, and we were submitting an additional redundant request to Infura for each request.

The `retry-on-empty` middleware is meant to retry certain methods when they return an empty response. This was also meant to deal with network synchronization issues that were more common years ago. This middleware works by making a "child" request over and over until either a retry limit is reached, or a non-empty response is received.

It was broken in that the final response recieved was thrown away, so it's as though the middleware was not used. Except that it did result in additional redundant network requests.

As a result of this update we should see that the extension is more resilient to certain network synchronization issues. But this is difficult to test, and these issues may not happen in production anymore today.

We should see a reduction in requests to Infura as well. This should be easier to test.

## Manual Testing Steps

1. Checkout the `develop` branch and create a build.
2. Open the background console to the network tab
3. Perform some action that interacts with the network (e.g. confirm a transaction).
4. Take note of the Infura requests that your interaction caused.
5. Perform steps 2-4 again on this branch, again taking note of the Infura requests. See that fewer requests are made.

Aside from testing that some redundant network requests have been eliminated, this should have no functional impact that we can practically test for.

## Pre-Merge Checklist

### DIFF
--- a/app/scripts/controllers/network/provider-api-tests/shared-tests.js
+++ b/app/scripts/controllers/network/provider-api-tests/shared-tests.js
@@ -673,12 +673,6 @@ export function testsForRpcMethodSupportingBlockParam(
           ),
           response: { result: mockResults[0] },
         });
-        // Note that the block-ref middleware will still allow the original
-        // request to go through.
-        comms.mockInfuraRpcCall({
-          request: requests[0],
-          response: { result: mockResults[0] },
-        });
 
         const results = await withInfuraClient(({ makeRpcCallsInSeries }) =>
           makeRpcCallsInSeries(requests),
@@ -713,29 +707,13 @@ export function testsForRpcMethodSupportingBlockParam(
           ),
           response: { result: mockResults[0] },
         });
-        // Note that the block-ref middleware will still allow the original
-        // request to go through.
-        comms.mockInfuraRpcCall({
-          request: requests[0],
-          response: { result: mockResults[0] },
-        });
         comms.mockNextBlockTrackerRequest({ blockNumber: '0x200' });
         comms.mockInfuraRpcCall({
-          request: requests[1],
-          response: { result: mockResults[1] },
-        });
-        // The previous two requests will happen again, with a different block
-        // number, in the same order.
-        comms.mockInfuraRpcCall({
           request: buildRequestWithReplacedBlockParam(
-            requests[0],
+            requests[1],
             blockParamIndex,
             '0x200',
           ),
-          response: { result: mockResults[1] },
-        });
-        comms.mockInfuraRpcCall({
-          request: requests[0],
           response: { result: mockResults[1] },
         });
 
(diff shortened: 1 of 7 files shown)


app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -823,22 +790,16 @@ export function testsForRpcMethodSupportingBlockParam(
         comms.mockNextBlockTrackerRequest({ blockNumber: '0x100' });
         // The block-ref middleware will make the request as specified
         // except that the block param is replaced with the latest block
-        // number.
+        // number, and we delay it.
         comms.mockInfuraRpcCall({
+          delay: 100,
```
Gudahtt, line 795: @mcmire Let me know if these test changes make sense! I did one read-through and removed any mocks that seemed no longer necessary. It's possible I missed some though - extra mocks don't cause tests to fail at the moment.

This line was the one I was least sure about. I didn't know for sure whether adding a delay to the _child_ request sufficiently demonstrates the behaviour we were looking for here or not.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -823,22 +790,16 @@ export function testsForRpcMethodSupportingBlockParam(
...
             '0x100',
           ),
           response: { result: mockResults[0] },
         });
-        // This is the original request as below, which the block-ref
-        // middleware will allow through, except that we delay it.
-        comms.mockInfuraRpcCall({
-          request: requests[0],
-          response: { result: mockResults[0] },
-          delay: 100,
```
mcmire, line 840: I think it's important that we delay one of these requests to prevent this test from failing intermittently. This test is different from other tests in that we issue multiple requests at the exact same time. We want the first one (the one with the first result) to be processed first, and then the other ones afterward. So the delay should guarantee the right result.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
mcmire: Gotcha. I think that's fine to remove the extra mocks.

In this case, though, I think it's important that we delay one of these requests to prevent this test from failing intermittently. This test is different from other tests in that we issue multiple requests at the exact same time. We want the first one (the one with the first result) to be processed first, and then the other ones afterward. So the delay should guarantee the right result.

Gudahtt: Right. So this line is a delay on the first request. I did this because the original test included a delay on the first "real" request, the one that would overwrite this one. Since that request no longer happens, I moved it to this request, which is now the first request.

mcmire: Ah yes, okay. I think I had trouble reading the diff. This looks good!

mcmire:
Looks good, pending a rebase.

darkwing (approved):

brad-decker (approved):

PR #16113, opened 2022-10-06
[FLASK] Create E2E test for snap installation, enable/disable, and removal

bowensanders:
This test checks that snap installation, enabling and disabling that snap, and snap removal all function properly.

This required adding a single new function to drivers.js that allows you to accept an alert popup to continue.

Also this increments the version of test-snaps that is used in the test process.

### DIFF
--- a/test/e2e/snaps/test-snap-onoff.spec.js
+++ b/test/e2e/snaps/test-snap-onoff.spec.js
@@ -102,7 +102,6 @@ describe('Test Snap onOff', function () {
 
         // click OK on the popup
         await driver.delay(1000);
-        // driver.switchTo().alert().accept();
         await driver.closeAlertPopup();
         await driver.delay(1000);
 
--- a/test/e2e/snaps/test-snap-onoff.spec.js
+++ b/test/e2e/snaps/test-snap-onoff.spec.js
@@ -1,0 +1,152 @@
+const { strict: assert } = require('assert');
+const { withFixtures } = require('../helpers');
+const FixtureBuilder = require('../fixture-builder');
+const { TEST_SNAPS_WEBSITE_URL } = require('./enums');
+
+describe('Test Snap onOff', function () {
+  it('tests install disable enable and removal of a snap', async function () {
+    const ganacheOptions = {
+      accounts: [
+        {
+          secretKey:
+            '0x7C9529A67102755B7E6102D6D950AC5D5863C98713805CEC576B945B15B71EAC',
+          balance: 25000000000000000000,
+        },
+      ],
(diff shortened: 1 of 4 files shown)


FrederikBolding (changes requested):
@bowensanders I think you've missed committing the actual test

bowensanders:
Now I've committed the actual test code itself. Sorry for its absence as I was still working on a problem that I was concerned may be another requested change later.

You will notice the last part of the test where you have to click the popover in order to actually remove the snap is an xpath - this is due to seeming shortcomings of being able to target the second "Remove Snap" button because webdriver thinks the popover is intercepting the click _regardless_ of whether you are using correct tags to identify the popover.
If there is a better solution to fixing that particular item, we can change it, but the test, as is, works perfectly over here.

test/e2e/snaps/test-snap-onoff.spec.js
```diff
@@ -0,0 +1,155 @@
...
+        windowHandles = await driver.getAllWindowHandles();
+        await driver.switchToWindowWithTitle('Test Snaps', windowHandles);
+
+        // wait then try the notification test
+        await driver.delay(1000);
+        await driver.clickElement('#sendInAppNotification');
+
+        // click OK on the popup
+        await driver.delay(1000);
+        // driver.switchTo().alert().accept();
```
FrederikBolding, line 105:
```suggestion
```

test/e2e/snaps/test-snap-onoff.spec.js
```diff
@@ -0,0 +1,155 @@
...
+        await driver.clickElement({ text: 'See details', tag: 'button' });
+        await driver.delay(500);
+
+        // try to remove snap
+        await driver.clickElement({ text: 'Remove Snap', tag: 'button' });
+        await driver.delay(500);
+
+        // try to click remove on popover
+        await driver.clickElement({
+          xpath: '/html/body/div[2]/div/div/section/div[3]/button[2]',
```
FrederikBolding, line 142: I'm not sure I understand your reasoning for this 🤔

bowensanders: As I stated above, the second, popped-over "remove snap" button was unadressable through any means other than xpath here. Selenium claims that the popover was hijacking the click (rather than just accepting the click) and even though we try to click on said popover. That makes the test fail.

bowensanders: this is fixed with f6aaf81

bowensanders:
replaced with #16628

PR #16128, opened 2022-10-09
Fix false positives in blockedDomainCheck (Provider Injection)

### LINKED ISSUE #16059: [Bug]: Preventing injection for docs.google.com is too encompassing and prevents any website that just has docs.google.com in a query parameter from accessing Metamask.
### Describe the bug

I'm working on a side project where "docs.google.com" is a necessary component in my URL query params. I'm unable to connect to Metamask as it doesn't inject on my site, even though the domain is something completely different.

### Steps to reproduce

1. Add `?test=docs.google.com` to any website. `window.ethereum` is no longer available.

_No response_

### Version

10.19.0

### Build type

_No response_

### Browser

Brave

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

ryanml:
Fixes: https://github.com/MetaMask/metamask-extension/issues/16059

## Explanation
The Regex that was previously used to check for a blocked domain was too over-encompassing, and would produce false positives in the case that a blocked host was part of a path or query string. This PR adjusts that regex, and in addition makes `blockedDomainCheck` exportable for testing. 

## Manual Testing Steps
1. Ensure that that the provider is not injected on the blocked domains listed in `shared/modules/provider-injection.js`
2. Ensure that the provider is injected on sites where a blocked host, (such as https://example.com?test=docs.google.com) is part of the query paramers

## Pre-Merge Checklist

- [X] **IF** this PR fixes a bug, a test that _would have_ caught the bug has been added

### DIFF
--- a/shared/modules/provider-injection.js
+++ b/shared/modules/provider-injection.js
@@ -8,7 +8,7 @@ export default function shouldInjectProvider() {
     doctypeCheck() &&
     suffixCheck() &&
     documentElementCheck() &&
-    !blockedDomainCheck()
+    !blockedDomainCheck(window.location.href)
   );
 }
 
@@ -61,9 +61,10 @@ function documentElementCheck() {
 /**
  * Checks if the current domain is blocked
  *
+ * @param currentUrl
  * @returns {boolean} {@code true} if the current domain is blocked
  */
-function blockedDomainCheck() {
+export function blockedDomainCheck(currentUrl) {
   const blockedDomains = [
     'uscourts.gov',
     'dropbox.com',
@@ -77,15 +78,11 @@ function blockedDomainCheck() {
     'blueskybooking.com',
     'sharefile.com',
   ];
-  const currentUrl = window.location.href;
   let currentRegex;
   for (let i = 0; i < blockedDomains.length; i++) {
     const blockedDomain = blockedDomains[i].replace('.', '\\.');
-    currentRegex = new RegExp(
-      `(?:https?:\\/\\/)(?:(?!${blockedDomain}).)*$`,
-      'u',
-    );
-    if (!currentRegex.test(currentUrl)) {
+    currentRegex = new RegExp(`(?:https?:\\/\\/)${blockedDomain}.*$`, 'u');
+    if (currentRegex.test(currentUrl)) {
       return true;
     }
   }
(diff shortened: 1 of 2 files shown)


shared/modules/provider-injection.js
```diff
@@ -61,9 +61,10 @@ function documentElementCheck() {
 /**
  * Checks if the current domain is blocked
  *
+ * @param currentUrl
  * @returns {boolean} {@code true} if the current domain is blocked
  */
-function blockedDomainCheck() {
+export function blockedDomainCheck(currentUrl) {
```
ryanml, line 67: Note: Realizing this now, if preferable I could remove the `currentUrl` argument and mock `window.location.href` in the tests

shared/modules/provider-injection.js
```diff
@@ -77,15 +78,11 @@ function blockedDomainCheck() {
     'blueskybooking.com',
     'sharefile.com',
   ];
-  const currentUrl = window.location.href;
   let currentRegex;
   for (let i = 0; i < blockedDomains.length; i++) {
     const blockedDomain = blockedDomains[i].replace('.', '\\.');
```
ryanml, line 83: Re QL alerts: we do a replacement for `.` escaping here so not sure it's actually an issue?

Smithbull001:
> 

Ok 👌 good

shared/modules/provider-injection.js
```diff
@@ -77,15 +78,11 @@ function blockedDomainCheck() {
...
-    currentRegex = new RegExp(
-      `(?:https?:\\/\\/)(?:(?!${blockedDomain}).)*$`,
-      'u',
-    );
-    if (!currentRegex.test(currentUrl)) {
+    currentRegex = new RegExp(`(?:https?:\\/\\/)${blockedDomain}.*$`, 'u');
```
Gudahtt, line 84: It looks like this would no longer match on subdomains of blocked sites. It'd also still return a false positive for `https://docs.google.com`.

Instead of adjusting the regex, I'd recommend comparing origins and paths directly. e.g. first look for an exact origin match. And if the blocklist entry has a path component, block anything under that path.

brad-decker:
Converting to draft for now

PR #16148, opened 2022-10-10
Some UI suggestions for SignTypedData v4 signatures PR

georgewrmarshall:
## Explanation

Some CSS and design system component suggestions for the [Replaced addresses by the address component on SignTypedData v4 signatures #16018](https://github.com/MetaMask/metamask-extension/pull/16018) update. No visual changes apart from the scroll button is a little bigger. Should function the same as in the original PR

Screen Shot 2022-10-10 at 12 52 13 PM


VSaric:
Thank you @georgewrmarshall for this updates! I will add this suggestion changes manually in my [PR](https://github.com/MetaMask/metamask-extension/pull/16018) so you can close this PR when I done that! 😄

PR #16155, opened 2022-10-10
Fix @keystonehq/bc-ur-registry patch

mcmire:
## Explanation

We use the `@keystonehq/bc-ur-registry` package for integration with Keystone devices. This package is peculiar as it not only ships with JavaScript files but also original TypeScript files. When we started adding support for TypeScript into the build system, we realized that the TypeScript files were apparently developed using a different TypeScript config than we use and therefore resulted in errors in our build process. So we added a patch which removed these TypeScript files so that we would fall back to using the compiled JavaScript files. This removed type information but fixed our build system.

Since then, we've bumped this package and the files we removed have changed, so the patch is no longer valid. This commit recreates the patch so that we no longer get warnings from `patch-package`.

## More Information

See Slack conversation: https://consensys.slack.com/archives/GTQAGKY5V/p1665437128455059

## Manual Testing Steps

* Remove `node_modules/@keystonehq/bc-ur-registry`, then run `yarn setup`. You shouldn't see a warning from `patch-package` anymore.

## Pre-Merge Checklist


brad-decker (approved):

ryanml (approved):

adonesky1 (approved):
LGTM

PR #16158, opened 2022-10-11
[FLASK] Fix `eth_accounts` permission for Snaps

FrederikBolding:
## Explanation

This PR fixes an issue with installing snaps that request `eth_accounts`. 

To create the `eth_accounts` permission, we need the selected accounts from the UI, these are normally passed back in the result of the approval (see https://github.com/MetaMask/metamask-extension/blob/develop/ui/components/app/permission-page-container/permission-page-container.component.js#L90-L109).

With this PR we make that appropriate change and patch the SnapController to support this return value properly.

## Manual Testing Steps

1. Create a template snap that uses `eth_accounts`
2. Install it and see that it successfully installs
3. Try develop and see that it fails to install there


FrederikBolding:
Closing in favor of #16161

PR #16098, opened 2022-10-05
chore: add bundle size diff to metamaskbot

mikesposito:
## Explanation

This PR adds Bundle size diff between the `PR branch` and latest on `develop` and shows an alert if the bundle size of `common` + `background` is increased since last version on develop.

## More Information

* Fixes #16070 

## Manual Testing Steps

After pushing to a PR, if the build is successful the Metamask bot should also show the following section:

<details><summary>Bundle size diffs [🚀 Bundle size reduced! 🚀]</summary><ul><li>background: -4</li>
<li>ui: -37940</li>
<li>common: -4</li></ul></details>

## Pre-Merge Checklist

- [x] PR template is filled out
- [ ] **IF** this PR fixes a bug, a test that _would have_ caught the bug has been added
- [x] PR is linked to the appropriate GitHub issue
- [ ] PR has been added to the appropriate release Milestone

### + If there are functional changes:

- [x] Manual testing complete & passed
- [ ] "Extension QA Board" label has been applied

PR #16168, opened 2022-10-12
Add build stats history json file

mikesposito:
## Explanation

This PR edits the `stats-module-load-init` step of the circleci pipeline, extending `bundle-stats-commit.sh` to update also a JSON version of the historical bundle size file, along with the JS one we currently update on each commit on develop.

This is needed by #16098 as metamaskbot will need to get the size of the bundle at a specific commit. See [this comment](https://github.com/MetaMask/metamask-extension/pull/16098#discussion_r992386162) for more info.

**Node**: The PR also installs `jq` on the `node-browsers-medium-plus` executor to be able to format the JSON output. 

## More Information

* See #16098

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -728,6 +728,9 @@ jobs:
           name: Run page load benchmark
           command: yarn mv3:stats:chrome --out test-artifacts/chrome/mv3
       - run:
+          name: Install jq
+          command: sudo apt install jq -y
+      - run:
           name: Record bundle size at commit
           command: ./.circleci/scripts/bundle-stats-commit.sh
       - store_artifacts:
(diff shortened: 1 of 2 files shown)


jpuri (approved):

jpuri:
After these changes we can probably remove this file https://github.com/MetaMask/extension_bundlesize_stats/blob/main/stats/bundle_size_data.temp.js

mikesposito:
> After these changes we can probably remove this file https://github.com/MetaMask/extension_bundlesize_stats/blob/main/stats/bundle_size_data.temp.js

Yes, I will also have to fill the `bundle_size_data.json` with old commits size data. I'll do it in the same PR

NiranjanaBinoy (approved):

PR #16198, opened 2022-10-13
Persisting segment events in MetaMetricsController store

### LINKED ISSUE #16173: Segment MV3 issue: fix segment events being lost due to service worker restart
Segment events are cached in segment SDK and passed over to segment api call in batches. In case of service worker restart this cache will be lost and hence events will not be submitted to segment.

Fix this issue by saving these events in MetamatricsController store till they are passed over to to segment.

jpuri:
Fixes #16173

PR address issue of segment events being lost in MV3 due to service worker restart. Events are now saved in MetaMetricsController store and resend to segment SDK till they are successfully submitted to segment.


darkwing:
Test failures look legit.

app/scripts/lib/util.js
```diff
@@ -174,3 +174,15 @@ export {
   getChainType,
   checkAlarmExists,
 };
+
+// Taken from https://stackoverflow.com/a/1349426/3696652
+const characters =
```
darkwing, line 179: We already have `uuidv4` and are using for getting a random ID -- maybe we use that instead?

darkwing (dismissed):
I found a really weird issue when testing.  I added a breakpoint to `comtrollers/metametrics.js` at line `971` so I could see every event coming through.

When I killed the service worker, and it then restarted itself, I saw the following series of events come through:

1. `"Initialization Create Password Page"`
2. "Home"
3. "MetaMetrics Opt In Page"
4. "Install Welcome Page"
5. "Unlock Page"
6. "Home"
7. "Initialization Secret Recovery Phrase Intro Page"
8. "Initialization Choose Restore or New Account Page"
9. "Initialization Page"
10. "Home"
11. "Initialization Secret Recovery Phrase Page"
12. "Initialization Page"
13. "New Account Page"

So I think there's a big issue we need to iron out when the service worker dies, because we shouldn't have this many incorrect events.

jpuri:
@darkwing : did you also checked segment, it will not have duplicate events as we assign same messageId when retrying events. Even if they are tried multiple times due to service worker restart - it will be single record in segment. They will eventually be cleaned up always from metametrics controller store also.

jpuri:
I tested it on segment today and it seems to work well:

Screenshot 2022-10-24 at 5 06 57 PM

jpuri:
Hey @seaona : can you plz test this to ensure metrics events are not lost. I discussed this with @brad-decker and it should be ok to have multiple identify events.

app/scripts/lib/util.js
jpuri: yep we actually can, but I used this as requirement here is just any random number and not uuid

app/scripts/controllers/metametrics.js
```diff
@@ -142,6 +146,13 @@ export default class MetaMetricsController {
       this.finalizeEventFragment(fragment.id, { abandoned: true });
     });
 
+    // Code below will submit to segment pending events as vontroller is re-instantiated
```
digiwand, line 149:
```suggestion
    // Code below submits any pending events to Segment if/when the controller is re-instantiated
```

app/scripts/controllers/metametrics.js
```diff
@@ -944,10 +955,36 @@ export default class MetaMetricsController {
...
       }
     });
   }
+
+  // Method below submits the request to analytics SDK.
+  // For MV3 it will also add event to controller store
+  // and pass a callback to remove it from store once request is submitted to segment
+  // Saving events in controller store in MV3 ensures that events are tracked
+  // even if service worker terminates before events are submiteed to segment.
+  _submitSegmentEvent(eventType, payload, callback) {
```
digiwand, line 970: nit: I'd like to suggest renaming `_submitSegmentEvent` to `_submitSegmentAPICall`. In analytics land, events mean something different. For Segment events see: 
https://segment.com/docs/connections/spec/track/#event.

The suggestion to use "APICall" instead comes from https://segment.com/docs/connections/spec/

app/scripts/controllers/metametrics.js
```diff
@@ -944,10 +955,36 @@ export default class MetaMetricsController {
...
-      this.segment.track(payload, callback);
+      this._submitSegmentEvent('track', payload, callback);
       if (flushImmediately) {
         this.segment.flush();
       }
     });
   }
+
+  // Method below submits the request to analytics SDK.
+  // For MV3 it will also add event to controller store
```
digiwand, line 966: this method appears to add the event to the store whether or not it is MV3. might want to update the code to reflect this comment 

Based on the below "event" -> "APICall" rename suggestion, I'd like to suggest updating the comment to something like
```
  /**
   * Use the method below to submit API calls to Segment using the analytics SDK.
   *
   * For MV3, this method also adds the API call to the controller store and removes
   * the API call from the controller store once its callback is executed.
   * Saving API calls in the controller store enables us to ensure API calls are submitted
   * to Segment even if the service worker termintates before the API call is submitted.
   *
   * @param {string} apiCallMethod - The method name of the analytics API call to be submitted to Segment.
   *                                 e.g. identify, track, page, etc.
   * @param {object} payload
   * @param {Function} callback
   */
```

digiwand, line 966: this method appears to add the event to the store whether or not it is MV3. We might want to update the code to reflect this comment

Nice comment overall! I'd like to suggest some minor tweaks:
- replace "events" to "API Calls" as mentioned in the [other comment](https://github.com/MetaMask/metamask-extension/pull/16198/files#r1007365427)
- add @param jsdoc comments
- capitalize Segment

  /**
   * Use the method below to submit API calls to Segment using the analytics SDK.
   *
   * For MV3, this method also adds the API call to the controller store and removes
   * the API call from the controller store once its callback is executed.
   * Saving API calls in the controller store enables us to ensure API calls are submitted
   * to Segment even if the service worker termintates before the API call is submitted.
   *
   * @param {string} apiCallMethod - The method name of the analytics API call to be submitted to Segment.
   *                                 e.g. identify, track, page, etc.
   * @param {object} payload
   * @param {Function} callback
   */

app/scripts/controllers/metametrics.js
jpuri: 👍 done

app/scripts/controllers/metametrics.js
jpuri: @digiwand : I implemented feedbacks above except for jsdoc, I think it is not requires for internal methods.

digiwand: Thank you @jpuri 

follow-up: 
1. can we swap out `eventType` name in the parameter as well? to maybe `apiCallMethod`? 
2. can we rename "events" in the store to "apiCalls"?
3. even though the jsdoc comments aren't required, I think they can be helpful to the reader.
4. The comment still does not match the code where "For MV3 it will also add event to controller store" because we are added the api calls to the store for both MV2 and MV3 builds

seaona:
From QA I've observed this 2 issues:

1. **Terminating service worker before batch is sent**: I see that the pending batch of events is not sent after terminating the service worker and the following error appears on the console: `AssertionError: \"timestamp\" must be a date.\n  at validateGenericEvent`. After this error, no more events are collected and MM stays loading idnefinetly if I refresh the browser.

https://user-images.githubusercontent.com/54408225/199002174-f443e4aa-49dc-40de-9c88-a22a7b54141d.mp4

2. **Collecting events before consent**: this might not be directly related to this PR but I've noticed that we are collecting events before the user gives consent. Precisely, we are collecting the following events:
   - Initialization Page (two times)
   - Wallet Setup Selected

https://user-images.githubusercontent.com/54408225/199002186-835075e4-d481-4a7a-9cea-f4097ff7ca27.mp4

cc @jpuri @brad-decker

jpuri:
@seaona : thanks a lot for timestamp issue finding, I fixed it can you plz re-test.

app/scripts/controllers/metametrics.js
jpuri: I renamed `events` to `segmentApiCalls` and also updated comments.

Renaming `eventType` to `apiCallMethod` looks confusing to me. Event before changes in this PR the field was eventType. Also I think JSDoc for internal method can create confusion.

brad-decker (changes requested):

app/scripts/controllers/metametrics.js
```diff
@@ -110,16 +110,20 @@ export default class MetaMetricsController {
     this.environment = environment;
 
     const abandonedFragments = omitBy(initState?.fragments, 'persist');
+    const segmentApiCalls = initState?.segmentApiCalls || {};
 
     this.store = new ObservableStore({
-      participateInMetaMetrics: null,
+      participateInMetaMetrics: true,
```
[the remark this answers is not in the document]
jpuri: ops my bad, thanks for catching this

app/scripts/controllers/metametrics.js
```diff
@@ -20,7 +20,7 @@ import {
 import { SECOND } from '../../../shared/constants/time';
 import { isManifestV3 } from '../../../shared/modules/mv3.utils';
 import { METAMETRICS_FINALIZE_EVENT_FRAGMENT_ALARM } from '../../../shared/constants/alarms';
-import { checkAlarmExists } from '../lib/util';
+import { checkAlarmExists, generateRandomId, isValidDate } from '../lib/util';
```
newlife08, line 23: - **__** ui/components/ui/currency-display/__snapshots__/currency-display.component.test.js.snap

newlife08 (dismissed):
Lo creo

newlife08:
> Lo creo

brad-decker (approved):

seaona:
@jpuri from QA looks good! I'm re-running Firefox e2e tests, as the installation script failed. Hopefully this fixes it!

segun (approved):
LGTM

PR #16205, opened 2022-10-14
Improve transaction confirmation page performance

### LINKED ISSUE #828: onTransaction is called repeatedly
When a user selects the snap tabntje onTransaction is not called once but recalled multiple times into infinity

FrederikBolding:
## Explanation

This issue fixes a Flask issue by fixing an underlying problem with the selectors used in the transaction confirmation screen (also in stable). The issue arose in Flask because we relied on the `txData` object for a `useEffect`. The `txData` object was continuously updated through spreads etc without memoization. This PR fixes the problem by wrapping the existing logic in `reselect` selectors and enabling deep equal memoization for these selectors.

## More Information

Fixes https://github.com/MetaMask/snaps-monorepo/issues/828

## Manual Testing Steps

1. Create an transaction insight snap that logs every time it is called
2. View the transaction insight tab in the confirmation screen
3. See that the transaction insight snap only logs once

(Perhaps another way to test in stable?)

### DIFF
--- a/lavamoat/browserify/beta/policy.json
+++ b/lavamoat/browserify/beta/policy.json
@@ -2693,6 +2693,7 @@
         "browserify>buffer": true,
         "browserify>events": true,
         "deep-freeze-strict": true,
+        "eslint>fast-deep-equal": true,
         "eth-ens-namehash": true,
         "eth-keyring-controller": true,
         "eth-query": true,
@@ -2701,7 +2702,6 @@
         "ethereumjs-util": true,
         "ethers": true,
         "ethjs>ethjs-unit": true,
-        "fast-deep-equal": true,
         "immer": true,
         "json-rpc-engine": true,
         "jsonschema": true,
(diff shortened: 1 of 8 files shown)


FrederikBolding:
@Gudahtt Maybe this PR should be renamed to something else given that it now affects both stable and Flask? While the main goal was to fix a Flask bug.

CHANDRAMA2 (approved):
Hello sir

Gudahtt:
>Maybe this PR should be renamed to something else given that it now affects both stable and Flask? While the main goal was to fix a Flask bug.

Agreed! Maybe something about improving transaction confirmation page performance. It should generally be a perf improvement across the board.

ritave:
In the future, that screen might be rerendered again for other reason, but it still shouldn't call `onTransaction`. How about using `const [txInsights, setTxInsights] = useState()` to maintain the tx insights cache

FrederikBolding:
> In the future, that screen might be rerendered again for other reason, but it still shouldn't call `onTransaction`. How about using `const [txInsights, setTxInsights] = useState()` to maintain the tx insights cache inside of `useTransactionInsightSnap`

Strictly if the transaction changes or you switch the tab back and forth (unmounting and remounting). In my opinion that should be fine.

jpuri (approved):
LGTM

GuillaumeRx (approved):

PR #16221, opened 2022-10-17
Use async/await instead of Promise.resolve

darkwing:
The usage of async/await seems more modern.

### DIFF
--- a/app/scripts/account-import-strategies/index.js
+++ b/app/scripts/account-import-strategies/index.js
@@ -6,14 +6,10 @@ import { addHexPrefix } from '../lib/util';
 import { stripHexPrefix } from '../../../shared/modules/hexstring-utils';
 
 const accountImporter = {
-  importAccount(strategy, args) {
-    try {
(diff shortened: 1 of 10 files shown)


brad-decker (dismissed):

digiwand (dismissed):
LGTM! 

comment: one thing that might be worth mentioning here is that we do want to be mindful when it comes to returning promise values

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#description
Screenshot 2022-11-02 at 10 54 39 PM

brad-decker (dismissed):

digiwand (dismissed):

brad-decker (dismissed):

HowardBraham:
I have incorporated the changes to `editable-label.js` in #16752, so I removed it from this PR to avoid future merge conflicts

jpuri (approved):
Code changes look good 👍 

But I think its debatable if async / await are better over vanilla promise.

naugtur (approved):

PR #16226, opened 2022-10-18
Emit tx:dropped when pending-tx-tracker.js gets transaction receipt w…

stole-dev:
## Explanation

We at Milkomeda created rollups which allow Apps to work in multiple blockchains.
Before we write transaction to a chain, we pretest it and if it fails filter out so that save our and user's resources
and don't post transactions that would fail on the chain. Problem with this is that Metamask can't know that
transaction failed unless it's written on chain.

Solution that I implemented requires our rollup (or any other dApp) to establish a proxy that on
eth_getTransactionReceipt returns a fake receipt (real is impossible to return as transaction was never written on chain)
with blockNumber and blockHash being real values from the latest block and block status with value "0x2" that represents 
status dropped. Now when receipt is asked for transaction that has failed in the app and was never written on the chain, 
Metamask marks transaction dropped.

Why dropped and not failed?
- Transaction never reached the chain
- Nonce for sending next Transaction trough MetaMask won't go up a number and then result in nonce too high error but will stay the same

## More Information

SEE #16251 

## Manual Testing Steps
There is unittest written for this. Unfortunately, as our rollup is not open source identical steps to reproduce can't be made.
What can be done is set up proxy for returning fake receipt for failed transactions and some localhost evm node that filters out transaction if it has eg. too high nonce. 
Then send transaction that would result in nonce too high error.
Eg. Start nonce is 0. Trough MetaMask send transaction with nonce 5. 
Transaction should be filtered out in the evm node and MetaMask should mark it as dropped.

## Pre-Merge Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Add a way to manually make pending transaction dropped ([#16226](https://github.com/MetaMask/metamask-extension/pull/16226))
 
 ## [10.20.0]
 ### Changed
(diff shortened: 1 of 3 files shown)


stole-dev:

PR #16245, opened 2022-10-20
[YarnV3]: Disables HAR generation in yarn install script

brad-decker:
## Explanation
The --har flag has been removed from the yarn install command, we should find an alternative if these archive files are still being used. @Gudahtt do you have any input on this? 

## Pre-Merge Checklist

### DIFF
--- a/.circleci/scripts/deps-install.sh
+++ b/.circleci/scripts/deps-install.sh
@@ -5,7 +5,7 @@ set -x
 # Exit immediately if a command exits with a non-zero status.
 set -e
 
-yarn install --frozen-lockfile --har
+yarn install --frozen-lockfile #--har temporarily disabling for CI work
 
 # Move HAR file into directory with consistent name so that we can cache it
 mkdir -p build-artifacts/yarn-install-har
--- a/.circleci/scripts/deps-install.sh
+++ b/.circleci/scripts/deps-install.sh
@@ -5,7 +5,7 @@ set -x
 # Exit immediately if a command exits with a non-zero status.
 set -e
 
-yarn install --frozen-lockfile #--har temporarily disabling for CI work
+yarn install --frozen-lockfile
 
 # Move HAR file into directory with consistent name so that we can cache it
 mkdir -p build-artifacts/yarn-install-har


Gudahtt:
Hmm. The HAR files were meant to provide an audit trail, in case a malicious dependency ever made it's way into our bundle.

Maybe we can replace this with [something like this](https://github.com/bobvanderlinden/harhar#proxy); a proxy that records all traffic as a HAR file. This should work with Yarn v3, it supports HTTP proxies.

Gudahtt:
Hmm. The HAR files were meant to provide an audit trail, in case a malicious dependency ever made it's way into our bundle.

Maybe we can replace this with [something like this](https://github.com/bobvanderlinden/harhar#proxy); a proxy that records all traffic as a HAR file. This should work with Yarn v3, it supports HTTP proxies.

brad-decker:
@Gudahtt I don't think that package will work, but I could just be being dense. The reason being is that the proxy it creates is limited. It looks like you have to specify the 'connect-host' and the proxy just replaces the hostname of the request with the connect-host. Given that yarn will likely connect to multiple hosts (and thus the reason we need a HAR for audits) I think this won't work. If you have additional insights into how to setup that package properly to work with yarn I would be appreciative of a second opinion.

Gudahtt:
I agree, it looks like that won't be suitable.

There are other proxies that might be suitable (mitmproxy looked promising), and we could also consider using `tcpdump` to record all network traffic during the install and convert the pcap file into a har file. But... none of this seems easy. I'm not sure this is worth the investment right now.

.circleci/scripts/deps-install.sh
```diff
@@ -5,7 +5,7 @@ set -x
 # Exit immediately if a command exits with a non-zero status.
 set -e
 
-yarn install --frozen-lockfile --har
+yarn install --frozen-lockfile #--har temporarily disabling for CI work
```
Gudahtt, line 8: Nit: If we're not tracking this work as something we'll be addressing, we can delete this comment
```suggestion
yarn install --frozen-lockfile
```

Gudahtt (approved):
LGTM!

brad-decker:
I'm wondering if, in the future, we could write a docker container that does the HAR generation at the OS level, so we get a full accounting of all outgoing network requests during the dep install step

Gudahtt:
Hmm. Seems possible, but I'm not sure. Seemingly it's not possible to take packet captures within the CircleCI docker containers because they lack the NET_RAW capability. And seemingly even if we could enable that, it would have security implications. Lots of advantages to doing it at the application layer with a proxy or within Yarn itself.

PR #16250, opened 2022-10-21
DAPP action replay improvements

### LINKED ISSUE #16174: MV3: Improve DAPP action replay

jpuri:
Fixes: #16174

The PR improves DAPP action replay. As the background is loaded and its state is completely initialised a message is sent to inpage provider to replay actions.

To be merged after this PR is merged: https://github.com/MetaMask/metamask-extension/pull/16075


app/scripts/contentscript.js
```diff
@@ -463,26 +467,35 @@ function logStreamDisconnectWarning(remoteLabel, error) {
...
+    REPLAY_ENABLED = false;
+    window.postMessage(
+      {
+        target: INPAGE, // the post-message-stream "target"
+        data: {
+          // this object gets passed to obj-multiplex
+          name: PROVIDER, // the obj-multiplex channel name
+          data: {
+            jsonrpc: '2.0',
+            method: 'METAMASK_EXTENSION_STREAM_CONNECT',
```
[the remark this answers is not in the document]
jpuri: It is called each time I think.

app/scripts/contentscript.js
```diff
@@ -463,26 +467,35 @@ function logStreamDisconnectWarning(remoteLabel, error) {
...
-//       window.location.origin,
-//     );
-//     extensionPort.onMessage.removeListener(extensionStreamMessageListener);
-//   }
-// }
+function extensionStreamMessageListener(msg) {
+  if (
+    REPLAY_ENABLED &&
+    isManifestV3 &&
+    msg.data.method === 'metamask_chainChanged'
```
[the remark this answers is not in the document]
jpuri: I do not have a preference here.

test/e2e/snaps/enums.js
```diff
@@ -1,3 +1,3 @@
 module.exports = {
-  TEST_SNAPS_WEBSITE_URL: 'https://metamask.github.io/test-snaps/3.0.1',
+  TEST_SNAPS_WEBSITE_URL: 'https://metamask.github.io/test-snaps/3.1.0',
```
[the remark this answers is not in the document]
jpuri: I rebased the PR.

app/scripts/contentscript.js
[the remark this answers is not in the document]
digiwand: I see now. I think I got thoughts mixed up with `REPLAY_ENABLED`

`METAMASK_EXTENSION_STREAM_CONNECT` does look to be called each time the port and streams are connected and ready

app/scripts/contentscript.js
```diff
@@ -261,10 +261,14 @@ const setupPageStreams = () => {
   pageChannel = pageMux.createStream(PROVIDER);
 };
 
+// The field below is used to ensure that replay is done only once for each restart.
+let REPLAY_ENABLED = false;
```
digiwand, line 265: It took me a little bit of time to wrap my head around this logic. Now I see that `REPLAY_ENABLED`, in other words, means something like canSendExtensionConnectedMessage

In the case that the listener receiving the message expands on the logic, this comment and variable name could become outdated. `REPLAY_ENABLED` works. I'm just mentioning this here in case we want to take a preventative and potentially more scalable approach by renaming this

digiwand: I'm back with a follow-up thought...

Essentially, this might just be me, but the naming here isn't sitting with me well. It seems to go from retry -> connect -> retry. 

In this PR, `REPLAY_ENABLED` is used to send the `METAMASK_EXTENSION_STREAM_CONNECT` message which is then mapped to a message for `retryOnMessage`

https://github.com/MetaMask/providers/blob/main/src/StreamProvider.ts
```
this._jsonRpcConnection = createStreamMiddleware({
  retryOnMessage: 'METAMASK_EXTENSION_STREAM_CONNECT',
});

```

what if we align the naming across the code to something like:
`CAN_SEND_RETRY_ACTIONS_ON_CONNECT_MSG`
`METAMASK_EXTENSION_RETRY_ACTIONS_ON_CONNECT`
`retryActionsOnConnectMessage`
?

naugtur: +1

I'd rename:
```
METAMASK_EXTENSION_STREAM_CONNECT => METAMASK_EXTENSION_CONNECT_CAN_RETRY
REPLAY_ENABLED => METAMASK_EXTENSION_CONNECT_SENT
retryOnMessage => retryActionsOnConnectMessage
```

jpuri: +1

I did renamed first 2 suggestions, changing the third is hard as it is used in 2 different repos.

digiwand (dismissed):
LGTM! 

as [suggested above](https://github.com/MetaMask/metamask-extension/pull/16250#discussion_r1007779102), updating `retryOnMessage` => `retryActionsOnConnectMessage` in the other 2 repos you mentioned would be a nice-to-have; not a blocker here though

danjm (dismissed):
LGTM

jpuri:
PR to be merged only after this fix is added: https://github.com/MetaMask/json-rpc-middleware-stream/pull/35

app/scripts/contentscript.js
```diff
@@ -522,6 +527,38 @@ function logStreamDisconnectWarning(remoteLabel, error) {
...
+ * The function notifies inpage when the extension stream connection is ready. When the
+ * 'metamask_chainChanged' method is received from the extension, it implies that the
+ * background state is completely initialized and it is ready to process method calls.
+ * This is used as a notification to replay any pending messages in MV3.
+ *
+ * @param {object} msg - instance of message received
+ */
+function extensionStreamMessageListener(msg) {
+  if (
+    METAMASK_EXTENSION_CONNECT_SENT &&
```
digiwand, line 540: since we are adding the `extensionStreamMessageListener` listener on each `setupExtensionStreams` call, could we remove this flag and add logic to remove the `extensionStreamMessageListener` listener after the message has been sent?

jpuri: I tried to remove this listener `extensionStream.on('data', extensionStreamMessageListener);` but it does not work. There is an `off` method in api, but it does not works and rather throw error that `off` method does not exist.

app/scripts/contentscript.js
```diff
@@ -297,9 +297,14 @@ const setupPageStreams = () => {
   pageChannel = pageMux.createStream(PROVIDER);
 };
 
+// The field below is used to ensure that replay is done only once for each restart.
+let METAMASK_EXTENSION_CONNECT_SENT = false;
+
 const setupExtensionStreams = () => {
+  METAMASK_EXTENSION_CONNECT_SENT = true;
```
digiwand, line 304: nit: maybe `METAMASK_EXTENSION_CONNECT_SENT` -> `METAMASK_EXTENSION_CAN_SEND_RETRY`? thoughts? 
 
cc: @naugtur

Gudahtt:
Looks like this needs an update to resolve conflicts

segun (dismissed):

app/scripts/contentscript.js
Gudahtt: Agreed that calling this "connect sent" isn't idea, because it doesn't really represent whether the extension has connected. After we send the retry command, it remains true that the extension has connected, but this is `false`.

Maybe instead of "can send retry", it should be "should send retry"? Or "send retry when ready"

OR maybe it would be simpler to flip this around. Use this boolean to track when the retry has already been sent, call it "METAMASK_EXTENSION_RETRY_SENT", and set  it to `false` here. Then we can set to `true` in `extensionStreamMessageListener`

Gudahtt (dismissed):
LGTM! Though there are some open suggestions on improving that variable name

segun (dismissed):

segun (approved):

app/scripts/contentscript.js
naugtur: How about we call it what it is - a semaphore. 

`METAMASK_EXTENSION_CONNECT_SEMAPHORE`

brad-decker (approved):

jpuri:
I am merging this PR, I am working on a following PR where I will address re-naming suggestion.

PR #16267, opened 2022-10-25
Make event tracking idempotent using unique messageId

### LINKED ISSUE #15738: Make metrics action trackMetaMetricsEvent idempotent

jpuri:
Fixes: #15738

In this PR there is change in approach to ensure metrics event tracking is idempotent. We use actionId and uniqueIdentifier of the event to generate messageId for segment. Thus duplicate calls will result in same messageId and hence in segment metrics will be recorded only once as segment ignore second request with same messageId.


digiwand (dismissed):

brad-decker (dismissed):
Nice. 👏🏻

app/scripts/controllers/metametrics.js
```diff
@@ -113,7 +129,7 @@ export default class MetaMetricsController {
     const segmentApiCalls = initState?.segmentApiCalls || {};
 
     this.store = new ObservableStore({
-      participateInMetaMetrics: null,
+      participateInMetaMetrics: true,
```
segun, line 132: Why is this always true now?

segun (dismissed):
One small request, otherwise LGTM

darkwing (dismissed):

app/scripts/controllers/metametrics.js
brad-decker: yeah should definitely still be null, otherwise they are automatically opted in by default and events tracked prior to opting in

jpuri: my fault, I fixed it, thx :)

darkwing (approved):

brad-decker (approved):

PR #16268, opened 2022-10-25
Fix prompt for backing the SRP

### LINKED ISSUE #13277: Fix prompt for backing SRP
### **User Story**
As a user, I want to only get the prompt for backing my SRP once I have non-zero token balances

Figma/Notion links:

Notion: https://www.notion.so/As-a-user-I-want-to-only-get-the-prompt-for-backing-my-SRP-once-I-have-non-zero-token-balances-70f7780fe14f470aa8454704b55cf76b

### Context:

Currently, we show this prompt once the user has non-zero funds for the base token (e.g ETH) or once they add a token to their list. 

As it obscures part of the UI, prevents the user from interacting with what's behind it and cannot be closed, we should **update it to also check if the added token has a non-zero balance** and, if yes, only then show the prompt. 

We should also **exclude testnet assets** from this non-zero balance detection as those tokens don't have value.

### **Acceptance Criteria**

1. Don't show the banner when the users only have balance on testnets;
2. Don't show the banner when the users have added new tokens with zero balance (and don't have any other non-zero balance in their wallet).

filipsekulic:
## Explanation
Fixed prompt for backing the SRP.

## More Information
* Fixes #13277

### Before

https://user-images.githubusercontent.com/92310504/198003736-91c95f3c-1d7a-4a0f-a956-96e7a741f89b.mov

### With Ganache

https://user-images.githubusercontent.com/92310504/198003842-ccd8a14d-3700-4054-b1d8-b6df045f9718.mov

### After

https://user-images.githubusercontent.com/92310504/198003991-4a157b67-f2dd-49f6-8d8a-412ddbaa94b5.mov

### With Ganache

https://user-images.githubusercontent.com/92310504/198004055-5cddde83-979b-4dab-b135-695df0035695.mov

## Manual Testing Steps

How to reproduce it (all these options are with `Remind me later to backup SRP` option while creating a wallet):

1. Create a new wallet, switch to Sepolia test network (account with 0 balance), send some amount from another wallet and warning will not appear like before.
2. Switch to Mainnet, import a token (for example UST) and warning will not appear like before.
3. Connect a hardware wallet, import a hardware account and import a token (for example UST) and warning will not appear like before.
4. Test the Mainnet functionality without having to actually use the Mainnet by using the Ganache's forking feature (https://github.com/MetaMask/metamask-extension/pull/15944/files).


mirjanaKukic:
Verified by QA

ui/selectors/selectors.js
```diff
@@ -1278,13 +1279,59 @@ export function getAllAccountsOnNetworkAreEmpty(state) {
...
+    currentValue === '0x0' || currentValue === '0x00';
+
+  const balancesOnNonTestNetworks = Object.entries(cachedBalances)
+    .filter(
+      (networkAndAccountBalances) =>
+        TEST_CHAINS.includes(networkAndAccountBalances[0]) === false &&
+        networkAndAccountBalances,
+    )
+    .map((accountAndBalance) => Object.values(accountAndBalance[1]))
+    .map((balance) => balance.every(zeroBalanceCheck));
```
darkwing, line 1295: Rather than looping over `Object.entries` and its results 3 times, would it be possible to use `reduce` to get the desired outcome?

ui/selectors/selectors.js
```diff
@@ -1278,13 +1279,59 @@ export function getAllAccountsOnNetworkAreEmpty(state) {
...
+
+  return balancesOnNonTestNetworks.includes(false);
+}
+
+export async function getIsAnyTokenWithNonZeroBalance(state) {
+  const { tokens } = state.metamask;
+  const selectedAddress = getSelectedAddress(state);
+
+  if (tokens.length > 0) {
```
darkwing, line 1304: Is this `if` required if the `length` is `0`?

ui/selectors/selectors.js
filipsekulic: I tried, but didn't manage to get the proper result I needed...

ui/selectors/selectors.js
filipsekulic: No, because we just need to know if there is some token with non-zero balance. If the length of the token array is 0, then the return is false, as expected.

ui/pages/home/home.component.js
```diff
@@ -223,15 +234,27 @@ export default class Home extends PureComponent {
...
+    } = this.props;
     const { notificationClosing } = this.state;
 
     if (notificationClosing && !prevState.notificationClosing) {
       closeNotificationPopup();
     } else if (isNotification) {
       this.checkStatusAndNavigate();
     }
+
+    const showSRP = await shouldShowSeedPhraseReminder;
```
jpuri, line 251: async prop here looks weird, I would rather prefer to put this await in parent class and pass result here

ui/pages/home/home.component.js
```diff
@@ -223,15 +234,27 @@ export default class Home extends PureComponent {
...
+
+    if (
+      _prevProps.shouldShowSeedPhraseReminder !== shouldShowSeedPhraseReminder
+    ) {
```
jpuri, line 255: Since `shouldShowSeedPhraseReminder` is async this comparison looks weird.

filipsekulic: Good catch! I must have forgotten to remove this piece of code...

darkwing:
Needs rebase -- I'll look at it ASAP tomorrow!

ui/selectors/selectors.js
HowardBraham: I agree with @darkwing here, the `if` is not necessary.
If `tokens.length` is 0, then the loop evaluates to
`for (let i = 0; i < 0; i++)`
which is a loop that runs 0 times.

ui/selectors/selectors.js
```diff
@@ -1334,13 +1335,59 @@ export function getAllAccountsOnNetworkAreEmpty(state) {
...
+        return true;
+      }
+    }
+  }
+
+  return false;
+}
+
+export async function getShouldShowSeedPhraseReminder(state) {
+  const { seedPhraseBackedUp, dismissSeedBackUpReminder } = state.metamask;
```
HowardBraham, line 1374: A large amount of computation could be shortcut in most cases by adding this here:
`if (seedPhraseBackedUp || dismissSeedBackUpReminder) return false;`

ui/selectors/selectors.js
```diff
@@ -1334,13 +1335,59 @@ export function getAllAccountsOnNetworkAreEmpty(state) {
...
+  return balancesOnNonTestNetworks.includes(false);
+}
+
+export async function getIsAnyTokenWithNonZeroBalance(state) {
+  const { tokens } = state.metamask;
+  const selectedAddress = getSelectedAddress(state);
+
+  if (tokens.length > 0) {
+    for (let i = 0; i < tokens.length; i++) {
+      const res = await getERC20Balance(tokens[i], selectedAddress);
```
HowardBraham, line 1362: This doesn't have to be `await`ed, these checks could run in parallel.  I'm not sure how important that is because I'm not sure how long each one takes.

ui/pages/home/home.component.js
```diff
@@ -87,7 +87,10 @@ export default class Home extends PureComponent {
     forgottenPassword: PropTypes.bool,
     suggestedAssets: PropTypes.array,
     unconfirmedTransactionsCount: PropTypes.number,
-    shouldShowSeedPhraseReminder: PropTypes.bool.isRequired,
+    shouldShowSeedPhraseReminder: PropTypes.oneOfType([
+      PropTypes.bool,
+      PropTypes.object,
+    ]).isRequired,
```
HowardBraham, line 93: I don't think I understand what happened here, why `shouldShowSeedPhraseReminder` could be an object now.

ui/selectors/selectors.js
HowardBraham: A little later I did some performance tests, and this function `getShouldShowSeedPhraseReminder()` runs quite frequently.  It basically runs at least two times whenever you click any button, anywhere in the UI, and it runs for 25-35 ms.

ui/selectors/selectors.js
filipsekulic: Fixed

ui/pages/home/home.component.js
filipsekulic: Fixed

ui/selectors/selectors.js
HowardBraham: Sorry, I guess I said this in a confusing way.  The change you made doesn't work, because `res` is now a Promise.

You need to do something a lot fancier, using something like `Promise.all()` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all

filipsekulic: Yes, I misunderstood you...
I fixed it here: 0f0ceebd9f6b1c8b560ecd1dc2172edc7cbe2bbe
Also, I had to revert some changes - `shouldShowSeedPhraseReminder: PropTypes.oneOfType([PropTypes.bool, PropTypes.object]).isRequired`. That's because `shouldShowSeedPhraseReminder` is a `Promise`  at one moment before it gets resolved...

HowardBraham:
@filipsekulic This is getting a lot better, but the fundamentals of the React lifecycle on this are still very broken.  I'd love to take this opportunity to fully improve the lifecycle architecture.

The place the data is getting **used** is home.component.js `componentDidMount()` and then `renderNotifications()` with `this.state.shouldShowSrpReminder`.  But it's getting tangled up in Redux and totally overcomputed.

`componentDidMount()` actually doesn't run that frequently, but `getShouldShowSeedPhraseReminder()`, as I said earlier, runs at least 2 times with any button click anywhere in the UI.  This means that things are getting computed and then discarded.  Haha yeah, the React lifecycle is totally broken.  I know you didn't write the lifecycle, it's legacy code, but let's fix it together.

I would propose these rules for when it's computed:
1. Do not compute if `seedPhraseBackedUp || dismissSeedBackUpReminder` (you already did this)
2. Do not compute if the notification is already showing
3. Do not compute more frequently than every 1 minute

I would also love to pull this out of Redux if we can figure it out. (Haha I'm a Redux-hater, and this here is a great example of why: it encouraged previous authors to make a mess.)

I suggest that you don't start this by writing code, but start by writing lifecycle architecture proposals here.

Issue #15937, opened 2022-09-21
Replace setInterval for polling in TokenListcontroller


NiranjanaBinoy:
Replace setInterval for polling in `fetchTokenList()` in TokenListcontroller in a time window of 24 hours.

NiranjanaBinoy:
The corresponding PR has to be created in the controller repo. 

Since the chrome alarms have to be used only for the extension, we would have to configure them at the API level to differentiate extension using mv3 from mobile and extension using mv2. 
Can we introduce a new optional state variable with true value to identify extension with mv3 and use chrome alarms in that case?

NiranjanaBinoy:
Something similar to https://github.com/MetaMask/controllers/pull/861

NiranjanaBinoy:
Closing this ticket after a conversation with the core team. Quoting the comment below and link to the chat [here](https://consensys.slack.com/archives/C01V1L10W2E/p1664226689035799?thread_ts=1664215606.626099&cid=C01V1L10W2E)

> Using setTimeout might be fine for the intervals that are below 5 minutes. That's not awful.
If the service worker restarts every 5 minutes, then the 3-minute intervals will be triggered once every 2.5 minutes on average. That seems fine.
Though I would check to ensure that the update is triggered upon construction as well. If it's only after 3 minutes has elapsed, then it would become a 5 minute timer in practice.

NiranjanaBinoy:
On a closer look, we found that the `fetchTokenList` method, which is used to fetch the dynamic tokenList, is getting called every time the UI pop or window is opened as we are starting and stopping the polling based on the `controllerConnectionChanged`[here](https://github.com/MetaMask/metamask-extension/blob/08a4f93b5382eb9802d8164457084286b3bb3191/app/scripts/metamask-controller.js#L545). This means every time the UI opens a recurring fetch with an interval of 24 hrs is scheduled, and it gets unscheduled or cleared when the UI closes. Upon further discussion, the decision was made that the introduction of the chrome alarm for setInterval is not necessary for tokenListcontroller as there is a fetch happening every time the UI opens, and there is cache data available with a threshold of 24 hours which is to reduce the load on the server.

PR #16284, opened 2022-10-26
Replace setInterval for polling in TokenListcontroller

NiranjanaBinoy:
## Explanation
#15937 

## Pre-Merge Checklist


NiranjanaBinoy:
On a closer look, we found that the `fetchTokenList` method, which is used to fetch the dynamic tokenList, is getting called every time the UI pop or window is opened as we are starting and stopping the polling based on the `controllerConnectionChanged`[here](https://github.com/MetaMask/metamask-extension/blob/08a4f93b5382eb9802d8164457084286b3bb3191/app/scripts/metamask-controller.js#L545). This means every time the UI opens a recurring fetch with an interval of 24 hrs is scheduled, and it gets unscheduled or cleared when the UI closes. Upon further discussion, the decision was made that the introduction of the chrome alarm for setInterval is not necessary for tokenListcontroller as there is a fetch happening every time the UI opens, and there is cache data available with a threshold of 24 hours which is to reduce the load on the server.

PR #16324, opened 2022-10-31
[FLASK] Update e2e tests for new react test-snaps page

bowensanders:
This is attempting to upgrade the e2e tests to work with the new react version of test-snaps

see https://github.com/MetaMask/test-snaps/pull/81


bowensanders:
In order to test the first e2e, error-

- checkout the bs/react branch on test-snaps
- yarn install && yarn build && yarn start
that will start the servers for the test page

to run the tests:
- `rm -r node_modules && yarn setup && yarn build:test:flask` - this should work unless you get lavamoat errors later
- `SELENIUM_BROWSER=chrome yarn test:e2e:single test/e2e/snaps/test-snap-error.spec.js --leave-running` - this will leave the page running until you ^c out of it when the test fails

bowensanders:
Tests are failing on the bip-32 snap test, as if the new key-tree fix isn't merged here.

test/e2e/snaps/test-snap-error.spec.js
```diff
@@ -52,24 +63,23 @@ describe('Test Snap Error', function () {
...
-        );
-        const text = await error.getText();
-        assert.equal(
-          text.includes(
-            "Snap Error: 'random error inside'. Error Code: '-32603'",
-          ),
-          true,
-        );
+        // try to click on the dismiss button and pass test if it works
+        await driver.clickElement({
```
FrederikBolding, line 79: Why was this changed to not find the text anymore?

bowensanders: It assumes the error is present because the dismiss button is present. 
Also the new error snap doesn't return a "true" to the interface anymore after the new snap was edited by maarten. Is it necessary?

bowensanders: fixed this with 9bec385

FrederikBolding (approved):

hmalik88 (approved):

bowensanders:
closes https://github.com/MetaMask/metamask-extension/issues/16532

PR #16326, opened 2022-10-31
Enable and stabilise e2e tests for Swaps, add debouncing

dan437:
## Description
This PR: 
- enables Swaps E2E tests
- adds debouncing for the `Swap from` and `Swap to` fields
- cleans up console errors in Swaps

### DIFF
--- a/test/e2e/swaps/shared.js
+++ b/test/e2e/swaps/shared.js
@@ -44,11 +44,9 @@ const buildQuote = async (driver, options) => {
   }, 20000);
   await driver.clickElement('.searchable-item-list__labels');
   await driver.clickElement('.dropdown-input-pair__to');
-  await driver.clickElement(
-    'input[placeholder="Search name or paste address"]',
-  );
+  await driver.clickElement('input[data-testid="search-list-items"]');
   await driver.fill(
-    'input[placeholder="Search name or paste address"]',
+    'input[data-testid="search-list-items"]',
     options.swapTo || options.swapToContractAddress,
   );
   if (options.swapToContractAddress) {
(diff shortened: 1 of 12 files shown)


test/e2e/swaps/shared.js
```diff
@@ -8,6 +10,17 @@ const ganacheOptions = {
...
 
+const driverOptions = {
+  timeOut: 20000,
+};
+
+const withFixturesOptions = {
+  fixtures: new FixtureBuilder().build(),
+  ganacheOptions,
+  driverOptions,
+  failOnConsoleError: false,
```
PeterYinusa, line 21: is failOnConsoleError still needed?

test/e2e/swaps/shared.js
```diff
@@ -28,19 +41,27 @@ const buildQuote = async (driver, options) => {
   await driver.wait(async () => {
     const tokens = await driver.findElements('.searchable-item-list__item');
     return tokens.length > 1;
-  }, 10000);
+  }, 20000);
```
PeterYinusa, line 44: Why do we need to wait this long?

test/e2e/swaps/shared.js
dan437: Yes, we are still getting React errors with some tests:
image

I can try to clean up those errors and then this won't be needed.

test/e2e/swaps/shared.js
dan437: I've increased it just to be on the safe side if token loading would get little longer than 10s.

test/e2e/swaps/shared.js
```diff
@@ -8,6 +10,16 @@ const ganacheOptions = {
   ],
 };
 
+const driverOptions = {
+  timeOut: 20000,
+};
```
PeterYinusa, line 15: can we get rid of this?

PeterYinusa (dismissed):

PeterYinusa (approved):

darkwing (approved):

PR #16329, opened 2022-10-31
ensure that we default users with browser language code 'zh' to the supported 'zh_CN' language code

adonesky1:
## Explanation

Add logic to the function we use to determine a users preferred language code such that we default users with language codes set as the generic (and unsupported) `zh` language code to use the supported zh_CN language code.

## Pre-Merge Checklist

### DIFF
--- a/app/scripts/lib/get-first-preferred-lang-code.js
+++ b/app/scripts/lib/get-first-preferred-lang-code.js
@@ -1,8 +1,10 @@
 import browser from 'webextension-polyfill';
 import allLocales from '../../_locales/index.json';
 
+// ensure that we default users with browser language code 'zh' to the supported 'zh_CN' language code
+const existingLocaleCodes = { zh: 'zh_CN' };
+
 // mapping some browsers return hyphen instead underscore in locale codes (e.g. zh_TW -> zh-tw)
-const existingLocaleCodes = {};
 allLocales.forEach((locale) => {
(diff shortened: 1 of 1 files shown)


Gudahtt:
Could we solve this in a more generic manner? e.g. add a `zh` alias to `zh_CN` in the `index.json` file, something like that.

I expect we might want to use a pattern like this in more places. Currently some of our locales are region-specific in practice but we're using a generic code. And there's `pt`, which has a community-maintained base locale but better maintained region-code options (so we might want to start using one of them in place of what we have now).

adonesky1:
> I expect we might want to use a pattern like this in more places.

Its just unclear how to select which regional version to across different cases? In the case of `zh` we could use the alphabetically first regional code (`zh_CN`) which is the default we definitely want to use for that case, but that won't necessarily be for other languages with regional localizations that we want to default to?

Gudahtt:
>presumably we don't want to do that for pt or other languages where we still do have localizations for both the generic (pt) and regional (pt_BR, pt_PT) codes?

Agreed! Not until deciding to remove the `pt` locale and replace it with one of those two at least.

Gudahtt (approved):
This seems like a good solution!

kumavis (approved):

NiranjanaBinoy (approved):
LGTM

PR #16337, opened 2022-11-01
Add tests for custom JSON-RPC network client

### LINKED ISSUE #16041: Write tests for custom JSON-RPC network client
As part of an effort to [merge the NetworkController that exists in this repo with the NetworkController that exists in the `controllers` repo](https://github.com/MetaMask/controllers/issues/753), we need to write tests for the JSON-RPC layer, and all of its custom behavior, that the NetworkController exposes.

We have two "network clients" that handle JSON-RPC network requests, one for Infura and one for custom networks. The Infura network client tests are covered under issue #14404. This issue is for testing the JSON RPC network client that handles custom networks.

BelfordZ:
## Explanation

Previously we had written tests for `createInfuraClient`, which creates a middleware stack designed to connect to an Infura provider. These tests exercise various RPC methods that relate to the behavior that the middleware provides (mainly around caching). 

Now we need to write the same tests but for `createJsonRpcClient`, which creates a middleware stack designed to connect to a non-Infura RPC endpoint. To do this, we had to:

- Consolidate the tests for both types of RPC client into a single test file.
- Add conditions around tests or assertions in tests to account for differences in behavior between the two sets of middleware stacks.
- Relocate code in `createJsonRpcClient` which slows down `eth_estimateGas` calls just for tests so that this behavior can be disabled in the network client tests

It's probably best to look at `shared-tests.js` to get a better sense of the tests that now exist between the two network clients. We may introduce PRs in the future to make this easier to read.

Eventually, as we unify the network controllers in this repo and in the controllers repo, we will move these tests into the controllers repo.

## Manual Testing Steps

This PR does not change any user-facing behavior. All of the tests should continue to pass with this PR.

## Pre-merge author checklist

- [x] I've clearly explained:


BelfordZ:
Removed the code for estimate gas in test, everything still green. :) :)

app/scripts/controllers/network/provider-api-tests/helpers.js
```diff
@@ -286,14 +295,23 @@ export async function withMockedInfuraCommunications(...args) {
...
+  let clientUnderTest;
+  if (type === 'infura') {
+    clientUnderTest = createInfuraClient({
+      network,
+      projectId: INFURA_PROJECT_ID,
+    });
+  } else {
+    clientUnderTest = createJsonRpcClient({
+      rpcUrl: 'http://localhost:8545',
+      chain_id: '0x1',
```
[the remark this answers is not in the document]
BelfordZ: another note, kind of surprising that this doesn't make some tests fail (some tests are using different networks and checking their chain id...)

app/scripts/controllers/network/provider-api-tests/helpers.js
```diff
@@ -286,14 +295,23 @@ export async function withMockedInfuraCommunications(...args) {
  * on that object.
  * @returns {Promise<any>} The return value of the given function.
```
BelfordZ, line 296: todo: update jsdoc

app/scripts/controllers/network/provider-api-tests/helpers.js
```diff
@@ -249,21 +258,21 @@ function makeRpcCall(ethQuery, request) {
  * is called with an object that allows you to mock different kinds of requests.
```
BelfordZ, line 258: todo: update jsdoc

app/scripts/controllers/network/provider-api-tests/helpers.js
```diff
@@ -94,12 +95,24 @@ function debug(...args) {
  * Infura supports.
  *
  * @param {object} options - The options.
- * @param {string} options.network - The Infura network you're testing with
+ * @param {string} options.network - The network you're testing with
  * (default: "mainnet").
+ * @param {string} options.type - Infura or Custom
```
BelfordZ, line 100:
```suggestion
 * @param {string} options.type - if defined, must be either `infura` or `custom`
```

app/scripts/controllers/network/provider-api-tests/helpers.js
```diff
@@ -194,7 +200,10 @@
       completeResponse = response.body;
     }
   }
-  let nockRequest = nockScope.post(`/v3/${INFURA_PROJECT_ID}`, {
+  const url = nockScope.basePath.includes('infura.io')
```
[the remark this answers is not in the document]
BelfordZ: CodeQL must not know this is only for testing. hmm.

app/scripts/controllers/network/createJsonRpcClient.test.js
```diff
@@ -0,0 +1,194 @@
+/**
```
BelfordZ, line 1: same tests suite as createInfuraClient, just moving it closer towards using the openrpc document referenced in the comment below

app/scripts/controllers/network/createJsonRpcClient.test.js
```diff
@@ -0,0 +1,194 @@
...
+    { name: 'eth_sendTransaction', numberOfParameters: 1 },
+    { name: 'eth_sign', numberOfParameters: 2 },
+    { name: 'eth_submitWork', numberOfParameters: 3 },
+    { name: 'eth_syncing', numberOfParameters: 0 },
+    { name: 'eth_uninstallFilter', numberOfParameters: 1 },
+    { name: 'eth_subscribe', numberOfParameters: 1 },
+    { name: 'eth_unsubscribe', numberOfParameters: 1 },
+    { name: 'net_listening', numberOfParameters: 0 },
+    { name: 'net_peerCount', numberOfParameters: 0 },
+    { name: 'parity_nextNonce', numberOfParameters: 1 },
```
BelfordZ, line 40: `parity_nextNonce` - I tested this method on a new infura api key and seems it doesn't work. Do we need to include this method, or any method with `parity` prefix?

app/scripts/controllers/network/createJsonRpcClient.test.js
```diff
@@ -0,0 +1,194 @@
...
+          providerType: 'custom',
+        },
+        ({ makeRpcCall }) => {
+          return makeRpcCall({
+            method: 'net_version',
+          });
+        },
+      );
+
+      expect(chainId).toStrictEqual('5');
```
BelfordZ, line 191: I would have thought this would fail since we hardcode chain_id: "0x1"

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -95,58 +117,68 @@ export function testsForRpcMethodNotHandledByMiddleware(
  * parameter will not be used in determining how to cache the method.
  *
  * @param method - The name of the RPC method under test.
+ * @param options0
```
BelfordZ, line 120:
```suggestion
 * @param options the options
```

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -430,12 +488,18 @@ export function testsForRpcMethodAssumingNoBlockParam(method) {
...
-        await expect(promiseForResult).rejects.toThrow(
-          new RegExp(
-            `^InfuraProvider - cannot complete request\\. All retries exhausted\\..+${errorMessagePrefix}: Some message`,
-            'su',
-          ),
-        );
+        const errMsg = `${errorMessagePrefix}: Some message`;
+
+        let err;
+        if (providerType === 'infura') {
```
BelfordZ, line 494: should probably make this into a helper since it appears in a few places

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -447,52 +511,58 @@ export function testsForRpcMethodAssumingNoBlockParam(method) {
  * cacheable.
  *
  * @param method - The name of the RPC method under test.
+ * @param options0
```
BelfordZ, line 514: fix jsdoc

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -511,21 +581,21 @@ export function testsForRpcMethodsThatCheckForBlockHashInResponse(method) {
       const requests = [{ method }, { method }];
       const mockResults = [emptyValue, { blockHash: '0x100' }];
 
-      await withMockedInfuraCommunications(async (comms) => {
+      await withMockedCommunications(async (comms) => {
```
BelfordZ, line 584: missing params here...

Nooknick7899:
2d9a7d6e64258d4c49ed136af9b34fb44187ebb1

app/scripts/controllers/network/provider-api-tests/helpers.js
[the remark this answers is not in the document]
BelfordZ: should this be `chainId`

app/scripts/controllers/network/createJsonRpcClient.js
```diff
@@ -12,21 +12,25 @@ import { SECOND } from '../../../../shared/constants/time';
 
 const inTest = process.env.IN_TEST;
 const blockTrackerOpts = inTest ? { pollingInterval: SECOND } : {};
-const getTestMiddlewares = () => {
-  return inTest ? [createEstimateGasDelayTestMiddleware()] : [];
-};
 
-export default function createJsonRpcClient({ rpcUrl, chainId }) {
+export default function createJsonRpcClient({ rpcUrl, chainId, networkId }) {
```
mcmire, line 16: What do you think about making API/behavior changes to `createJsonRpcClient` in a different PR? That way it might be easier to follow why we are doing this.

app/scripts/controllers/network/createJsonRpcClient.test.js
mcmire: This is why I differentiated what was in Infura but not in the spec, or vice versa, because we seem to have legacy code in the middleware. So if we are explicitly handling this, I think we should test it, even if it's not supported in the spec, but we should note as such so that we know that it's legacy and that we should remove it in a future PR.

app/scripts/controllers/network/provider-api-tests/helpers.js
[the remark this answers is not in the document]
mcmire: Hmm. I'll take a closer look at this.

app/scripts/controllers/network/createJsonRpcClient.test.js
mcmire: Hmm, I'll take a look at this more closely.

app/scripts/controllers/network/provider-api-tests/helpers.js
```diff
@@ -275,25 +289,35 @@ export async function withMockedInfuraCommunications(...args) {
...
+  let clientUnderTest;
+  if (type === 'infura') {
+    clientUnderTest = createInfuraClient({
+      network,
+      projectId: INFURA_PROJECT_ID,
+    });
+  } else {
+    clientUnderTest = createJsonRpcClient({
+      rpcUrl: LOCALHOST_RPC_URL,
+      chainId: BUILT_IN_NETWORKS[network].chainId || '0x1',
```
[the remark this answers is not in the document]
BelfordZ: I have another branch on controllers repo that moves these tests into typescript, and so I will need to adjust this code over there anyways.

app/scripts/controllers/network/provider-api-tests/helpers.js
```diff
@@ -91,15 +96,27 @@ function debug(...args) {
...
  * (default: "mainnet").
+ * @param {string} options.type - if defined, must be either `infura` or `custom`
+ * (default: "infura").
  * @returns {NockScope} The nock scope.
  */
-function buildScopeForMockingInfuraRequests({ network = 'mainnet' } = {}) {
-  return nock(`https://${network}.infura.io`).filteringRequestBody((body) => {
+function buildScopeForMockingRequests({
+  network = 'mainnet',
+  type = 'infura',
```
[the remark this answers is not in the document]
BelfordZ: The only reason I opted for defaulting it is that it would make for less code to change. Not opposed to the idea of making it required.

app/scripts/controllers/network/provider-api-tests/helpers.js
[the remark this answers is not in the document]
mcmire: Okay. I'm not totally convinced we have to do this but perhaps there's something I'm not seeing. Maybe we change this over there in any case so those changes go together?

app/scripts/controllers/network/provider-api-tests/helpers.js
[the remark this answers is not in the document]
BelfordZ: the url looks more like https://${network}.infura.io

It's right that its a lazy way to check a url, but its wrong about the context and how much it should matter. If it knew it was test code, we wouldnt have got this error.

app/scripts/controllers/network/provider-api-tests/helpers.js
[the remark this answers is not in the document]
BelfordZ: Ive fixed this issue actually. There was some code that wasnt passing in type. This would have been caught if it was required, so + points to that suggestion :)

app/scripts/controllers/network/createJsonRpcClient.test.js
BelfordZ: I will add that back. 

At some point I'd like to revist these tests and use a combination of the ethereum/execution-apis + metamask/api-specs openrpc document

app/scripts/controllers/network/createJsonRpcClient.test.js
BelfordZ: ah okay. I did a rough search for things touching parity_nextNonce and came up with nothing, but to be fair I didnt search very exhaustively.

app/scripts/controllers/network/createJsonRpcClient.test.js
BelfordZ: as mentioned above, fixed this now

app/scripts/controllers/network/provider-api-tests/createClient.test.js
```diff
@@ -0,0 +1,216 @@
+/**
```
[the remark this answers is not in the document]
BelfordZ: not the biggest fan, because this makes it IMO more difficult to tell what the actual differences are. With this PR, you can search for providerType === 'infura' and find every case where its different. If they are seperate files, over time they will continue to drift and the differences become less clear. I would like us to move towards unifying these 2 interfaces and implementations, and starting with the tests is IMO a good way to get us there. 

One thing that I could get behind is to move all the things that work differently into 2 seperate test files (`createInfuraClient.test.js` and `createJsonRpcClient.test.js`), and leave the things that work the same in a `createClient.test.js`. I'm not sure the juice is worth the squeeze on that one though.

app/scripts/controllers/network/provider-api-tests/helpers.js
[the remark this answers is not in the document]
BelfordZ: Valid points. I thought it would be best to try to keep the scope of change as minimal as possible until we are using network controller from controllers repo.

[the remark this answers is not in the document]
BelfordZ: That said, I would like to unify the interfaces in the controllers repo PR that is soon to come...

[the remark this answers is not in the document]
BelfordZ: hmm... Ill give it a shot and see

app/scripts/controllers/network/provider-api-tests/createClient.test.js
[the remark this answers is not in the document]
mcmire: > not the biggest fan, because this makes it IMO more difficult to tell what the actual differences are. With this PR, you can search for providerType === 'infura' and find every case where its different.
>
> ...
>
> I would like us to move towards unifying these 2 interfaces and implementations, and starting with the tests is IMO a good way to get us there.

Gotcha. I guess the only other argument I would make here about keeping all the tests in one file vs. two is that it creates more code complexity and that may end up impacting maintainability in the future. I've found, when writing tests, that every time I use loops and logic and other mechanisms that I would usually use in an implementation file to make code more compact, it pretty much always ends up biting me when I go to update those tests later. If you want a straightforward list of what this function does, now you have to go through each test and apply the logic yourself, and it can be a burden to do that.

That said, I see your point about being able to tell the differences more easily if you have the logic right in front of you instead of having to do the comparison between two files yourself. Obviously right now we don't have that much logic in these tests, so it's not a big deal. Also, it sounds like you have a plan for unifying all of this and maybe this will end up getting simplified or something. So, maybe this is just something we watch out for in the future.

app/scripts/controllers/network/provider-api-tests/helpers.js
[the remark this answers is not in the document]
mcmire: Oh ok cool!

app/scripts/controllers/network/provider-api-tests/helpers.js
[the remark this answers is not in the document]
mcmire: Ah yeah I got the URL wrong, lol. But yeah I see what you mean about the CodeQL check. No reason it should be here, and no reason it should be a High at the very least.

app/scripts/controllers/network/provider-api-tests/createClient.test.js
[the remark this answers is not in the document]
BelfordZ: For sure I share in your frustrations of having to deal with overly abstract testing setups, this is also a worry of mine. Theres a delicate tradeoff between additional complexity Vs. how much code needs to be updated for any given change. More places to update leads to greater likelyhood of mistakes being made and not caught. More complexity makes it take longer before someone is able to update / fix any of that code. The list of tradeoffs goes on. Context is important in determining how much of each is appropriate, and in the context of testing, I dont think we should aim to minimize complexity at the cost of extreme duplication, just as I dont think we should aim to optimize for lowering lines of code. We should aim to group functionality into tests that if they break, they are broken in relation to the grouped functionality.

That being said, I think due to the nature of these tests (broad coverage of general functionalities), I think this level of complexity and abstractness is acceptable, and in this case, I think it actually makes them easier to maintain, since changes to the client / middleware require far fewer updates to the tests. I'm certainly not advocating for over-valueing reduction in lines of code, but I believe this to be more of a grouping tests based around some general functionalities. Unless you are changing those general functionalities, you should never end up with one of these tests breaking on you. If you are adding a new method to this set but it doesnt pass, putting it in a different test file makes it clear that there is a deviation from the standard client middleware functionality. 

As for unifying it all, I think we can do this at the createXYZClient interface level to start with at the same time as we migrate this code to controllers repo. This will overall minimize the change set introduced when later integrating controllers/network-controller with extension.

app/scripts/controllers/network/provider-api-tests/helpers.js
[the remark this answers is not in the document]
BelfordZ: Should I make network also required? Same argument applied right?

app/scripts/controllers/network/provider-api-tests/createClient.test.js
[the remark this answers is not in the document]
Gudahtt: Hmm. Alternatively, we could take the loop contents here and move them to a "shared client tests" module, and use that in both `createJsonRpcClient.test.js` and `createinfuraClient.test.js`.

It would be nice to keep the convention of the test module matching the name fo the module under test. But in addition to that, I had planned on adding some tests that would only be applicable to the Infura client (for the Infura client specific middleware). That would better fit in a dedicated module for that client as well.

app/scripts/controllers/network/createJsonRpcClient.js
```diff
@@ -36,26 +34,3 @@ export default function createJsonRpcClient({ rpcUrl, chainId }) {
 
   return { networkMiddleware, blockTracker };
 }
-
-function createChainIdMiddleware(chainId) {
```
Gudahtt, line 40: Could we move these changes to a separate PR?

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -1108,31 +1166,144 @@ export function testsForRpcMethodSupportingBlockParam(
...
 
-              await expect(promiseForResult).rejects.toThrow(
-                new RegExp(
-                  `^InfuraProvider - cannot complete request\\. All retries exhausted\\..+${errorMessagePrefix}: Some message`,
-                  'su',
-                ),
+      ['ETIMEDOUT', 'ECONNRESET', 'SyntaxError'].forEach(
+        (errorMessagePrefix) => {
+          if (providerType === 'infura' || errorMessagePrefix === 'ETIMEDOUT') {
+            it(`retries the request to Infura up to 5 times if an "${errorMessagePrefix}" error is thrown while making the request, returning the successful result if there is one on the 5th try`, async () => {
```
Gudahtt, line 1203: Hmm, this test description has "Infura" in it, but it can be run with the non-Infura client if the current iteration is for `ETIMEDOUT`

app/scripts/controllers/network/provider-api-tests/createClient.test.js
```diff
@@ -0,0 +1,180 @@
...
+        'web3_clientVersion',
+      ];
+      assumingNoBlockParam.forEach((name) =>
+        describe(`method name: ${name}`, () => {
+          testsForRpcMethodAssumingNoBlockParam(name, { providerType });
+        }),
+      );
+    });
+
+    it(`net_version does not hit ${providerType} provider, instead returning the networkId that maps to the rpc network, as a decimal string`, async () => {
```
mcmire, line 154: Nit: What do you think about using a `describe` to hold tests for `net_version`, even if there is only one test? This way the names of the tests (when viewed by running this one test file on its own) are consistent, and the test file in general follows a general shape.

app/scripts/controllers/network/createJsonRpcClient.js
mcmire: I agree, we shouldn't be making changes to any middleware in this PR, we should just add new tests.

app/scripts/controllers/network/provider-api-tests/createClient.test.js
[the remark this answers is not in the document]
mcmire: > As for unifying it all, I think we can do this at the createXYZClient interface level to start with at the same time as we migrate this code to controllers repo. This will overall minimize the change set introduced when later integrating controllers/network-controller with extension.

Even if we want to unify these interfaces, I am going to push back slightly on this and argue that we should hold off on doing that. Not only would duplicating the tests / test helpers make this PR more focused, as well as the PR that will end up doing the unification, but it also gives us room to change our mind in case the separation does turn out to be useful (as Mark points out).

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -447,59 +516,69 @@ export function testsForRpcMethodAssumingNoBlockParam(method) {
...
  * @param method - The name of the RPC method under test.
+ * @param options - the options
+ * @param options.providerType - The type of provider being tested. Either `infura` or `custom`
+ * (default: "infura").
  */
-export function testsForRpcMethodsThatCheckForBlockHashInResponse(method) {
-  it('does not hit Infura more than once for identical requests and it has a valid blockHash', async () => {
+export function testsForRpcMethodsThatCheckForBlockHashInResponse(
+  method,
+  { providerType = 'infura' } = { providerType: 'infura' },
```
Gudahtt, line 525: Nit: we don't need to set the default provider here twice; the second one is redundant.
```suggestion
  { providerType = 'infura' } = {},
```

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -95,58 +121,69 @@ export function testsForRpcMethodNotHandledByMiddleware(
...
  * @param method - The name of the RPC method under test.
+ * @param options - the options
+ * @param options.providerType - The type of provider being tested. Either `infura` or `custom`
+ * (default: "infura").
  */
-export function testsForRpcMethodAssumingNoBlockParam(method) {
-  it('does not hit Infura more than once for identical requests', async () => {
+export function testsForRpcMethodAssumingNoBlockParam(
+  method,
+  { providerType = 'infura' } = { providerType: 'infura' },
```
Gudahtt, line 130: Nit:
```suggestion
  { providerType = 'infura' } = {},
```

app/scripts/controllers/network/provider-api-tests/helpers.js
```diff
@@ -275,25 +282,34 @@ export async function withMockedInfuraCommunications(...args) {
...
- * an object that allows you to interact with the client via a couple of methods
- * on that object.
+ * @param {WithClientArgs} args - Either an options bag + a function, or
+ * just a function. The options bag, at the moment, may contain `network`
+ * (defaults to "mainnet"). The options bag may also include providerType
+ * The function is called with an object that allows
+ * you to interact with the client via a couple of methods on that object.
  * @returns {Promise<any>} The return value of the given function.
  */
-export async function withInfuraClient(...args) {
```
Gudahtt, line 289: Nit: I don't think it makes sense anymore for this function to allow the options to be unset. The provider type is always passed in in practice, and it seems better to require that going forward too. That also lets us simplify this function.

Instead of taking "args" and splitting it into options and callback by length, we can always expect options and a callback. This makes the types and JSDoc descriptions simpler too.

app/scripts/controllers/network/createInfuraClient.test.js
```diff
@@ -1,389 +0,0 @@
...
-      });
-    });
-
-    describe('eth_getUncleByBlockHashAndIndex', () => {
-      testsForRpcMethodAssumingNoBlockParam('eth_getUncleByBlockHashAndIndex');
-    });
-
-    describe('eth_getUncleByBlockNumberAndIndex', () => {
-      // NOTE: eth_getUncleByBlockNumberAndIndex does take a block param at the
-      // 0th index, but this is not handled by our cache middleware currently
```
Gudahtt, line 224: Nit: this comment was dropped

app/scripts/controllers/network/createInfuraClient.test.js
```diff
@@ -1,389 +0,0 @@
...
-        'eth_getUncleByBlockNumberAndIndex',
-      );
-    });
-
-    describe('eth_getUncleCountByBlockHash', () => {
-      testsForRpcMethodAssumingNoBlockParam('eth_getUncleCountByBlockHash');
-    });
-
-    describe('eth_getUncleCountByBlockNumber', () => {
-      // NOTE: eth_getUncleCountByBlockNumber does take a block param at the 0th
```
Gudahtt, line 235: Nit: this comment was dropped

app/scripts/controllers/network/provider-api-tests/createClient.test.js
Gudahtt, line 154: This description isn't accurate for the RPC client, which does hit the network to answer this query.

Gudahtt (commented):
Just finished my first complete pass of this PR! Overall this looks great, I'm finding the tests easier to follow with similar methods grouped together like that in the client tests. The changes to the helper functions look reasonable too.

I've made a few suggestions on improving readability, and found a few small things that seem to have been removed accidentally. It would also be great to remove the changes to the code-under-test in this PR, to make this easier to review and to make the commit history more readable.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -674,61 +787,70 @@ export function testsForRpcMethodSupportingBlockParam(
...
       });
     });
 
-    it('hits Infura and does not reuse the result of a previous request if the latest block number was updated since', async () => {
-      const requests = [
-        { method, params },
-        { method, params },
-      ];
-      const mockResults = ['first result', 'second result'];
+    if (providerType === 'infura') {
```
Gudahtt, line 799: It looks like this is exposing a genuine bug in our custom RPC provider, specifically in the cases where the block tag is unset or set to `latest`.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -1108,31 +1166,144 @@ export function testsForRpcMethodSupportingBlockParam(
...
+                    },
+                  );
+
+                  expect(result).toStrictEqual('the result');
+                },
               );
             });
+          }
+
+          it(`causes a request to fail with a custom error if an "${errorMessagePrefix}" error is thrown while making the request to Infura 5 times in a row`, async () => {
```
Gudahtt, line 1258: Another reference to Infura here that should be using `providerType`

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -1257,138 +1436,161 @@ export function testsForRpcMethodSupportingBlockParam(
...
+              { type: providerType },
+              ({ makeRpcCall }) => makeRpcCall(request),
+            );
+
+            expect(result).toStrictEqual('the result');
+          },
+        );
       });
 
       it('makes an additional request to Infura if the given block number is less than the latest block number', async () => {
```
Gudahtt, line 1536: (last one) Another reference to Infura here that should be using `providerType`

app/scripts/controllers/network/provider-api-tests/shared-tests.js
Gudahtt: cc @mcmire, it looks like if I replace `BlockRefRewriteMiddleware` with `BlockRefMiddleware` in the network client, this bug is fixed. I'm guessing what's happening here is that `BlockRefRewriteMiddleware` is replacing `latest` with the _old_ block number without first checking the block tracker. Then the next middleware, the block cache middleware, is returning the stale cache result.

mcmire: Hmm, interesting. Okay I'll take a closer look at this when I get a chance.

Gudahtt: I can create a PR to fix this bug in the custom RPC client. I just tagged you because of how this relates to work you've done in the `eth-json-rpc-middleware` package. I remember you had at one point suggested that we get rid of the "child request" in `BlockRefMiddleware`, but I suspect doing that would have introduced this bug in that middleware as well.

If that is the case, we should add a test to `eth-json-rpc-middleware` to capture this behavior and delete the broken `BlockRefRewriteMiddleware` from the package completely.

Gudahtt: Nevermind. This was a mistake in the tests. The `params` are being shared between the two calls here, so when that object is directly mutated in one it gets changed in the second call as well.

The middleware itself is working just fine. They both correctly ask the block tracker for the latest block number.

app/scripts/controllers/network/createJsonRpcClient.js
carlexos: Leave a reaply

app/scripts/controllers/network/provider-api-tests/createClient.test.js
```diff
@@ -0,0 +1,216 @@
...
+                await makeRpcCall(request);
+                expect(blockTracker.getCurrentBlock()).toStrictEqual('0x300');
+              },
+            );
+          },
+        );
+      });
+    });
+
+    describe('net_version and eth_chainId', () => {
```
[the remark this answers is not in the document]
carlexos: Leave a reaply

app/scripts/controllers/network/provider-api-tests/helpers.js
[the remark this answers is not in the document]
carlexos: Leave a reaply

app/scripts/controllers/network/provider-api-tests/createClient.test.js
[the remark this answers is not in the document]
mcmire: @BelfordZ I pushed a commit which restored `createInfuraClient.test.js` and `createJsonRpcClient.test.js`. The bulk of the tests in `shared-tests.js` as you had them; the individual files just call a function. I did pull out a couple of tests, however, that seemed to be Infura-specific.

app/scripts/controllers/network/provider-api-tests/createClient.test.js
mcmire: These tests have been moved to `createInfuraClient.test.js`.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -604,39 +686,70 @@ export function testsForRpcMethodsThatCheckForBlockHashInResponse(method) {
...
+  });
 }
 
 /**
  * Defines tests which exercise the behavior exhibited by an RPC method that
  * takes a block parameter. The value of this parameter can be either a block
  * number or a block tag ("latest", "earliest", or "pending") and affects how
  * the method is cached.
+ *
+ * @param options - the options
```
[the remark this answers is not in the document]
mcmire: This has been added.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
mcmire: This has been fixed. See: [66aa9c1](https://github.com/MetaMask/metamask-extension/pull/16337/commits/66aa9c1b1c4961ba571948aea065608cb7517087)

app/scripts/controllers/network/provider-api-tests/createClient.test.js
mcmire: I ended up moving this to the tests for Infura client. See: [66aa9c1](https://github.com/MetaMask/metamask-extension/pull/16337/commits/66aa9c1b1c4961ba571948aea065608cb7517087)

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -340,88 +389,102 @@ export function testsForRpcMethodAssumingNoBlockParam(method) {
...
     });
   });
 
   ['ETIMEDOUT', 'ECONNRESET', 'SyntaxError'].forEach((errorMessagePrefix) => {
-    it(`retries the request to Infura up to 5 times if an "${errorMessagePrefix}" error is thrown while making the request, returning the successful result if there is one on the 5th try`, async () => {
-      await withMockedInfuraCommunications(async (comms) => {
-        const request = { method };
+    if (
+      providerType === 'infura' ||
+      ['ECONNRESET', 'SyntaxError'].includes(errorMessagePrefix) === false
```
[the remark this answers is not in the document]
mcmire: I ended up unrolling this loop. See: [66aa9c1](https://github.com/MetaMask/metamask-extension/pull/16337/commits/66aa9c1b1c4961ba571948aea065608cb7517087)

app/scripts/controllers/network/provider-api-tests/shared-tests.js
mcmire: Fixed in [66aa9c1](https://github.com/MetaMask/metamask-extension/pull/16337/commits/66aa9c1b1c4961ba571948aea065608cb7517087).

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -1257,138 +1436,161 @@ export function testsForRpcMethodSupportingBlockParam(
...
+            );
 
-          expect(results).toStrictEqual(['first result', 'second result']);
-        });
+            expect(results).toStrictEqual(['first result', 'second result']);
+          },
+        );
       });
 
       it('makes an additional request to Infura if the given block number matches the latest block number', async () => {
```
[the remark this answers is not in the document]
mcmire: Fixed in [66aa9c1](https://github.com/MetaMask/metamask-extension/pull/16337/commits/66aa9c1b1c4961ba571948aea065608cb7517087).

app/scripts/controllers/network/provider-api-tests/shared-tests.js
mcmire: Fixed in [66aa9c1](https://github.com/MetaMask/metamask-extension/pull/16337/commits/66aa9c1b1c4961ba571948aea065608cb7517087).

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -750,100 +2176,144 @@ export function testsForRpcMethodSupportingBlockParam(
...
+              );
+            },
+          );
+
+          expect(result).toStrictEqual('the result');
+        });
       });
-    });
 
-    if (blockParamType === 'none') {
```
mcmire, line 843: I noticed this `if` statement while perusing this file. I think I left this `if` statement in a previous PR for debugging purposes, but this should not be here — or rather, this code should not be wrapped in an `if` statement. I've unwrapped the code in this PR, but have also extracted it to another PR: https://github.com/MetaMask/metamask-extension/pull/16906

app/scripts/controllers/network/provider-api-tests/helpers.js
mcmire: Agreed! I haven't made this change yet, so it still needs to be done.

app/scripts/controllers/network/createInfuraClient.test.js
```diff
@@ -1,389 +0,0 @@
...
-      });
-    });
-
-    describe('eth_getBalance', () => {
-      testsForRpcMethodSupportingBlockParam('eth_getBalance', {
-        blockParamIndex: 1,
-      });
-    });
-
-    describe('eth_gasPrice', () => {
```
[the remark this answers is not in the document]
mcmire: Agreed. I haven't added this test back yet.

app/scripts/controllers/network/createInfuraClient.test.js
mcmire: It looks like this was moved to `shared-tests.js` (under "methods that assume there is no block param").

app/scripts/controllers/network/createInfuraClient.test.js
mcmire: It looks like this was moved to `shared-tests.js` (under "methods that assume there is no block param").

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -340,88 +389,102 @@ export function testsForRpcMethodAssumingNoBlockParam(method) {
...
       });
     });
   });
 
   ['ETIMEDOUT', 'ECONNRESET', 'SyntaxError'].forEach((errorMessagePrefix) => {
-    it(`retries the request to Infura up to 5 times if an "${errorMessagePrefix}" error is thrown while making the request, returning the successful result if there is one on the 5th try`, async () => {
-      await withMockedInfuraCommunications(async (comms) => {
-        const request = { method };
+    if (
+      providerType === 'infura' ||
```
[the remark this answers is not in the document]
mcmire: I think it makes sense to add a little something for `ECONNRESET` (even if it might be a good idea to place this knowledge eventually live in `eth-json-rpc-infura`). As for `SyntaxError`, I'm not sure that it is unique to the Infura middleware. It looks like the fetch middleware handles this already, at least in spirit. It [attempts to parse the response body as JSON](https://github.com/MetaMask/eth-json-rpc-middleware/blob/254a66cf5f067f35db96c6a5740c8da9a79e495d/src/fetch.ts#L64-L70), and when that fails (due to a SyntaxError) it will throw an error which is then picked up as a retriable error.

app/scripts/controllers/network/createJsonRpcClient.js
mcmire: I realized why some of these changes were made and in fact I think it is necessary to adjust this file a bit. It seems that the `eth_gasEstimate` RPC method has been modified, or rather decorated, in tests so that it runs 2 seconds slower. That said I don't think the middleware that is used to achieve this is written correctly, as it seems to pause the request forever. I amended this file in a more minimal way so that we could disable this behavior for the network client tests.

mcmire:
@BelfordZ Sorry if I am stepping on your toes here, but I wanted to see if we could tie the loose ends on this PR. I took the time to address Mark's feedback above. In addition, there was some behavior that the custom JSON-RPC client has which the Infura client does not, and I made sure that was tested. Finally, I made sure that any behavior one client had, we were testing that this behavior was not applicable to the other client. For instance, the custom JSON-RPC client treats 418 as a rate limit error, but the Infura client treats 429 as a rate limit error. I made sure that we test that 418 is not treated as special by the Infura client and that 429 is not treated as special by the custom JSON-RPC client. In the process of doing this I had to unbundle some of the tests. In other words, instead of deciding the result of a request inside of a test based on the provider, I duplicated the test, then said:

```
if (providerType === 'infura') {
  // (test for Infura goes here)
} else {
  // (test for custom RPC-client goes here)
}
```

This allowed me to name the tests differently to describe the different behavior. I realize this may have not been your plan but hopefully you can see why I did that.

I'm mostly off tomorrow (Monday) but if you want to chat about this I'll be available again on Tuesday.

app/scripts/controllers/network/createInfuraClient.test.js
Gudahtt: That is a different comment. Notice that this method is not listed under that shared test.

app/scripts/controllers/network/createInfuraClient.test.js
Gudahtt: That is a different comment. This method is not listed in the comment under that shared test.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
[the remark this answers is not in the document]
mcmire: Added comment for `ECONNRESET` in [f676e08](https://github.com/MetaMask/metamask-extension/pull/16337/commits/f676e08e27ce08e82ee780eb2712bb792e45735b), will add more comments.

app/scripts/controllers/network/createInfuraClient.test.js
mcmire: Fixed in[756e87d](https://github.com/MetaMask/metamask-extension/pull/16337/commits/756e87dd08aef9d8f75323faef496b8e0751e6ab).

app/scripts/controllers/network/createInfuraClient.test.js
mcmire: Fixed in[756e87d](https://github.com/MetaMask/metamask-extension/pull/16337/commits/756e87dd08aef9d8f75323faef496b8e0751e6ab).

app/scripts/controllers/network/createInfuraClient.test.js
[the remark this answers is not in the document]
mcmire: Added back in [8a1e46b](https://github.com/MetaMask/metamask-extension/pull/16337/commits/8a1e46b1dcda4b0308ec01991ec7c359a43c7470).

app/scripts/controllers/network/createInfuraClient.test.js
Gudahtt: Thanks!

mcmire:
@Gudahtt I believe I've addressed all of your comments but let me know if you see anything else.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -54,33 +97,311 @@ async function waitForPromiseToBeFulfilledAfterRunningAllTimers(
...
+          testsForRpcMethodsThatCheckForBlockHashInResponse(method, {
+            providerType,
+          });
+        });
+      });
+    });
+
+    describe('other methods', () => {
+      describe('eth_getTransactionByHash', () => {
+        it("refreshes the block tracker's current block if it is less than the block number that comes back in the response", async () => {
```
mcmire, line 213: @BelfordZ You had placed these in `testsForRpcMethodsThatCheckForBlockHashInResponse`, but I don't think that's the right place for these tests. The purpose of `testsForRpcMethodsThatCheckForBlockHashInResponse` was to test for behavior related to caching (specifically [these lines in the `block-cache` middleware](https://github.com/MetaMask/eth-json-rpc-middleware/blob/254a66cf5f067f35db96c6a5740c8da9a79e495d/src/block-cache.ts#L110-L128) and doesn't have anything to do with `blockHash` being in the `result`. 

This raises a question of organization and whether the way you've chosen is the best approach. Right now you're grouping RPC methods together by which `testsFor*` function is being called. However, this is a bit rigid because it means if you want to add a test for an RPC method you need to repeat the `describe` for that RPC method and possibly place it in some other group. If we instead organize primarily by RPC method, then we are free to place whatever tests we like in the `describe` for that RPC method. So it looks like:

``` javascript
describe('RPC method 1', () => {
  it('test 1', () => {
    // ...
  });

  it('test 2', () => {
    // ...
  });

  it('test 3', () => {
    // ...
  });
})

describe('RPC method 2', () => {
  it('test 1', () => {
    // ...
  });

  it('test 2', () => {
    // ...
  });

  it('test 3', () => {
    // ...
  });
})
```

The `testsFor*` functions are just shorthands for defining a bunch of related tests in one go but at the end of the day you would still be able think of it like the above.

I realize that the reason why you might have to decided to reorganize the tests this way is due to the naming of the `testsFor*` functions. I should come up with better names that explain the behavior being tested by the tests, which they don't do a good job of right now. I will do this in a separate PR, but just wanted to make a note of it in the meantime.

mcmire:
Synced with @BelfordZ this morning on this PR. I will take this over and get this merged.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
mcmire: We may introduce some changes in a future PR based on work Zach did locally that didn't make it into this PR.

app/scripts/controllers/network/createJsonRpcClient.js
Gudahtt: Interesting. I'm having trouble understanding how your changes here resolved the issue. It appears to be functionally equivalent.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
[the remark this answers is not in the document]
Gudahtt: Great, thanks!

app/scripts/controllers/network/createJsonRpcClient.js
mcmire: This comment is tied to a previous commit, so is it possible you're not seeing the new changes? I can add a comment to the latest commit to explain what's happening.

app/scripts/controllers/network/createJsonRpcClient.js
```diff
@@ -10,22 +10,22 @@ import {
...
+  const blockTrackerOpts = process.env.IN_TEST
+    ? { pollingInterval: SECOND }
+    : {};
   const fetchMiddleware = createFetchMiddleware({ rpcUrl });
   const blockProvider = providerFromMiddleware(fetchMiddleware);
   const blockTracker = new PollingBlockTracker({
     ...blockTrackerOpts,
     provider: blockProvider,
   });
+  const testMiddlewares = process.env.IN_TEST
```
mcmire, line 23: For some reason, the test middleware added here is interfering with the running of the tests for `createJsonRpcClient`. I'm not sure whether the middleware is set up correctly or not, but after the promise resolves in `createEstimateGasDelayTestMiddleware`, the middleware after this point is not getting run, so requests through the provider are never completing. 

To get around this in a minimal fashion, we want to disable this middleware only for these network client tests. The problem is that previously we were reading `process.env.IN_TEST` when this file was 
read. This made it impossible to prevent this test middleware from getting added by unsetting that environment variable. By checking for the variable inside of `createJsonRpcClient`, we can unset that variable before initializing the network client in the test helper and then reset it afterward.

app/scripts/controllers/network/createJsonRpcClient.js
Gudahtt: Ohh I see. IN_TEST is being changed dynamically. I hadn't considered that, I was used to the e2e context where it's converted to a string literal. But we can change it here.

Gudahtt: Fair enough! This seems like a pretty safe change to make, unlikely to have a functional impact.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -339,41 +769,47 @@ export function testsForRpcMethodAssumingNoBlockParam(method) {
...
+        await expect(promiseForResult).rejects.toThrow(err);
       });
     });
   });
 
-  ['ETIMEDOUT', 'ECONNRESET', 'SyntaxError'].forEach((errorMessagePrefix) => {
-    it(`retries the request to Infura up to 5 times if an "${errorMessagePrefix}" error is thrown while making the request, returning the successful result if there is one on the 5th try`, async () => {
-      await withMockedInfuraCommunications(async (comms) => {
+  // Both the Infura and fetch middleware detect ETIMEDOUT errors and will
+  // automatically retry the request to the RPC endpoint in question, but both
```
Gudahtt, line 808: Nit: could we move the "success on 5th try" case outside of this `if` block, if the behavior is the same?

Gudahtt: I see this applies to a few other cases as well (the other `ETIMEDOUT` test, and the 503/504 handling upon 5 successive failures).

Gudahtt (dismissed):
LGTM!

app/scripts/controllers/network/provider-api-tests/shared-tests.js
mcmire: Yes absolutely. I can make this change today.

mcmire: Fixed in [6898d7e](https://github.com/MetaMask/metamask-extension/pull/16337/commits/6898d7e823a0a3db2b9869a583ae8f6a2aa458e5).

Gudahtt (dismissed):

mcmire:
@BelfordZ and I will do a final pass on this and approve this together.

mcmire (approved):

BelfordZ:
Reviewed Elliots updates -- looks great, nice work. 

LGTM! :heavy_check_mark:

Gudahtt (approved):

Issue #16149, opened 2022-10-10
Create Beta-branded extension


kevinghim:
### Background
Create and run a beta program to test the viability of MV3 in production. In order to do so, we need to ensure that the user is aware that this is not a main build, and they might encounter instability. When encountering bugs, we will facilitate the reporting of bugs.

### User story
As a user, I want to 
- [ ] make sure I'm on a different extension app, in order to know I shouldn't risk my assets when transacting
- [ ] know when I'm on both popup and full screen versions that I'm using the beta version, not the production version [designs below]
- [ ] get onboarded with the with clear beta branding, so I understand which [version I'm on](https://www.figma.com/file/qn0DJn2Mfw036gdt2aYuZB/MV3-Beta?node-id=3417%3A7249)
- [ ] see the legal terms pertaining to the beta release, to understand legal liability [Link all supporting links to (tbd)]
- [ ] see the beta terms during onboarding [start screen](https://www.figma.com/file/qn0DJn2Mfw036gdt2aYuZB/MV3-Beta?node-id=910%3A1321) as well as the about [screen](https://www.figma.com/file/qn0DJn2Mfw036gdt2aYuZB/MV3-Beta?node-id=3352%3A3648) 
- [ ] make sure the changes applied to dark mode [full screen](https://www.figma.com/file/qn0DJn2Mfw036gdt2aYuZB/MV3-Beta?node-id=3375%3A8519), [popup](https://www.figma.com/file/qn0DJn2Mfw036gdt2aYuZB/MV3-Beta?node-id=3366%3A7555)
- [ ] create MascotBeta component (ensuring "BETA" remains in the same position for anywhere that is is used (slack thread to refine this story -- https://consensys.slack.com/archives/G1L7H42BT/p1665680043269619)

### Resources
- Background on [Notion](https://www.notion.so/MV3-Beta-Program-02ab0eec864d4dea93cd192770621116)

### Designs
- Popup design link: [here](https://www.figma.com/file/qn0DJn2Mfw036gdt2aYuZB/MV3-Beta?node-id=3348%3A3528)
- Fullscreen design link: [here](https://www.figma.com/file/qn0DJn2Mfw036gdt2aYuZB/MV3-Beta?node-id=910%3A1321)

PR #16343, opened 2022-11-01
BETA - Update verbiage for success messages after onboarding

darkwing:
## Explanation

Simply changes app name in onboarding if user is using BETA, as well as about/info page link:

Link

## More Information

* See: https://github.com/MetaMask/metamask-extension/issues/16149

### DIFF
--- a/ui/pages/settings/info-tab/info-tab.component.js
+++ b/ui/pages/settings/info-tab/info-tab.component.js
@@ -2,8 +2,10 @@ import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 
 import Button from '../../../components/ui/button';
+import Chip from '../../../components/ui/chip/chip';
 import { SUPPORT_REQUEST_LINK } from '../../../helpers/constants/common';
 import { isBeta } from '../../../helpers/utils/build-types';
+import { COLORS } from '../../../helpers/constants/design-system';
 import {
   getNumberOfSettingsInSection,
   handleSettingsRefs,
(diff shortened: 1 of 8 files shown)


ui/pages/settings/info-tab/info-tab.component.js
```diff
@@ -73,6 +75,25 @@ export default class InfoTab extends PureComponent {
...
+              type="link"
+              href="https://metamask.io/beta-terms.html"
+              target="_blank"
+              rel="noopener noreferrer"
+              className="info-tab__link-text"
+            >
+              {t('betaTerms')}
+              <Chip
+                borderColor={COLORS.PRIMARY_DEFAULT}
+                backgroundColor={COLORS.PRIMARY_DEFAULT}
```
digiwand, line 90: After adding these colors, this Chip looks very similar to our Button components, but it doesn't share the Button behaviors. I think I kind of see why it works here because it also sends the user to the link when clicked on. 

curious if you have any thoughts here regarding this Chip instance @georgewrmarshall

georgewrmarshall: Yeah I agree was hesitant about the tag UI for this exact reason. I've asked for some design input from @Akatori-Design [in the Figma file
](https://www.figma.com/file/qn0DJn2Mfw036gdt2aYuZB?node-id=3452:7966#299953888)

Screen Shot 2022-11-04 at 7 37 47 PM

Akatori-Design: What if we just use the standard tag style?
image

georgewrmarshall: I'm fine with that! We have this component available for use! https://metamask.github.io/metamask-storybook/?path=/docs/ui-components-component-library-tag-tag-stories-js--default-story

Gudahtt (dismissed):
LGTM!

It seems that the changes to the Chip component are no longer needed, but maybe that will be useful in the future :shrug:

Gudahtt (approved):

georgewrmarshall (approved):
LGTM!

PR #16344, opened 2022-11-01
Fix issue with awaiting swaps page

### LINKED ISSUE #16318: [Bug]: [MV3] Cannot read properties of undefined (reading 'symbol'), after terminating service worker during a Swap
### Describe the bug

**Problem**: if service worker is terminated when I am performing a Swap, I land into MM error page with the error `Cannot read properties of undefined (reading 'symbol')`.

undefined-symbol

### Steps to reproduce

1.Click Swap
2. Introduce amount
3. Introduce token
4. Click Review Swap
5. Terminate service worker -- you can proceed normally
6. Click Confirm Swap
7. Terminate service worker 
8. Click Swap -- you land on the error page

https://user-images.githubusercontent.com/54408225/198889696-bada605a-65c7-4d20-ad8b-9e96e85c1f19.mp4

_No response_

### Version

Tested on top of this PR https://github.com/MetaMask/metamask-extension/pull/15558

### Build type

_No response_

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

adonesky1:
## Explanation

Currently when the service worker is terminated  immediately after a swap is initiated the `SwapsController` controlled `swapsState` is reset and so the `quotes` empty out along with other `fetchParams` causing a redirect:
https://github.com/MetaMask/metamask-extension/blob/develop/ui/pages/swaps/index.js#L505  . So if you don't quickly hit the `swap` button on the `view quotes` page after terminating the SW you will be redirected back to the `build-quote` page:

https://user-images.githubusercontent.com/34557516/199344799-5e705e69-f459-4f43-99c5-3be61b7ccd34.mov

**But** if you hit swap button before the redirect we [hit this call](https://github.com/MetaMask/metamask-extension/blob/develop/ui/pages/swaps/index.js#L257) to `prepareToLeaveSwaps`, which clears the swaps state

And before the state clearing process is complete the `signAndSendTransactions` call - [that is initiated when](https://github.com/MetaMask/metamask-extension/blob/develop/ui/pages/swaps/view-quote/view-quote.js#L1018) you hit the `swap` button - goes through with the swaps state intact 
https://github.com/MetaMask/metamask-extension/blob/develop/ui/pages/swaps/index.js#L257

once you arrive on the `awaiting-swaps` page, though the swap has been submitted correctly, the token information that we use to tell the user which token they will be receiving is no longer accessible since it is accessed through the now cleared `swapsState` (from the `SwapsController` in the background)

This PR fixes the issue by having the `await-swap` page fetch the swap transaction data from the transactions controller (where the state is persisted if the transaction is successfully initiated) rather than the swaps controller where the required state is cleared.

## More Information

* Fixes #16318

## Manual Testing Steps

1. Begin a Swap
2. Introduce amount
3. Introduce token
4. Click Review Swap
5. Terminate service worker -- you can proceed normally
6. Click Confirm Swap
7. Terminate service worker
8. Click Swap -- 
9. See that you do not land on an error page, you see the message: `Your <symbol of currency you are swapping for> will be added to your account once this transaction has processed.`
10. And see that the swap succeeds

## Pre-Merge Checklist


jpuri:
Though changes look good, I would suggest a review by someone from swaps team also.

dan437:
Thanks for the code. I believe it should be also changed on the smart transaction status page, which is displayed when a smart transaction is submitted: https://github.com/MetaMask/metamask-extension/blob/develop/ui/pages/swaps/smart-transaction-status/smart-transaction-status.js#L76

dan437 (approved):
I've tested it with and without smart transactions + terminating a service worker and it worked well. Great job!

darkwing (approved):

PR #16352, opened 2022-11-02
Jest incremental coverage

PeterYinusa:
## Pre-Merge Checklist

### DIFF
--- a/development/jest.config.js
+++ b/development/jest.config.js
@@ -1,10 +1,10 @@
 module.exports = {
   displayName: '/development',
-  collectCoverageFrom: ['<rootDir>/**/*.js'],
+  collectCoverageFrom: ['<rootDir>/build/transforms/**/*.js'],
   coverageDirectory: '../jest-coverage/development/',
   coverageReporters: ['html', 'text-summary', 'json-summary'],
   coverageThreshold: {
-    './development/build/transforms/**/*.js': {
+    global: {
       branches: 100,
       functions: 100,
       lines: 100,
@@ -14,6 +14,6 @@ module.exports = {
   resetMocks: true,
   restoreMocks: true,
   testEnvironment: 'node',
-  testMatch: ['<rootDir>/build/**/*.test.js'],
+  testMatch: ['<rootDir>/build/transforms/**/*.test.js'],
   testTimeout: 2500,
 };
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,7 +1,13 @@
 module.exports = {
   collectCoverageFrom: [
+    '<rootDir>/app/scripts/constants/error-utils.js',
+    '<rootDir>/app/scripts/controllers/network/**/*.js',
     '<rootDir>/app/scripts/controllers/permissions/**/*.js',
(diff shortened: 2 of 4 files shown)


development/jest.config.js
```diff
@@ -14,6 +14,6 @@ module.exports = {
   resetMocks: true,
   restoreMocks: true,
   testEnvironment: 'node',
-  testMatch: ['<rootDir>/build/**/*.test.js'],
+  testMatch: ['<rootDir>/build/transforms/**/*.test.js'],
```
brad-decker, line 17: I carried all necessary changes from this file over to #16455

jest.config.js
```diff
@@ -10,22 +16,10 @@ module.exports = {
...
-      branches: 100,
-      functions: 100,
-      lines: 100,
-      statements: 100,
-    },
-    './app/scripts/lib/createRPCMethodTrackingMiddleware.js': {
-      branches: 95.65,
-      functions: 100,
-      lines: 100,
-      statements: 100,
```
brad-decker, line 28: Did we want to drop these exceptions? @PeterYinusa? I can do that in #16455 -- currently the strategy I have there is I collect the coverage from all test runners and then combine the coverage reports programatically. Then I check the sub paths like these to ensure coverage remains accurate. I can however simplify our coverage reporting by removing these exceptions if that is preferable.

jest.config.js
```diff
@@ -1,7 +1,13 @@
 module.exports = {
   collectCoverageFrom: [
+    '<rootDir>/app/scripts/constants/error-utils.js',
+    '<rootDir>/app/scripts/controllers/network/**/*.js',
     '<rootDir>/app/scripts/controllers/permissions/**/*.js',
+    '<rootDir>/app/scripts/flask/**/*.js',
+    '<rootDir>/app/scripts/lib/**/*.js',
     '<rootDir>/app/scripts/lib/createRPCMethodTrackingMiddleware.js',
+    '<rootDir>/app/scripts/migrations/*.js',
+    '<rootDir>/app/scripts/platforms/*.js',
```
brad-decker, line 10: I can carry this change into #16455

brad-decker (changes requested):
I think this pull request will be unnecessary after #16455 especially if i carry some of the changes over from this one to that PR. Jest-it-up is removed in that PR in favor of a programatic coverage merge script that validates thresholds and fails if we have increased our coverage by 5% without updating our targets.

PeterYinusa:
> I think this pull request will be unnecessary after #16455 especially if i carry some of the changes over from this one to that PR. Jest-it-up is removed in that PR in favor of a programatic coverage merge script that validates thresholds and fails if we have increased our coverage by 5% without updating our targets.

Hey @brad-decker, yes this PR can be closed in favour of the work you're doing. Here I was just experimenting with jest it up and trying to identify the limitations so we can get a better understanding of what we can and cannot do with it

jest.config.js
PeterYinusa: NO we do not want to drop these.
I just dropped these as jest it up only looked at the global coverage and not the coverage in relative/custom paths

Issue #16370, opened 2022-11-03
[Bug]: there is a typo in the duplicate account import message


TimTinkers:
### Describe the bug

![before](https://user-images.githubusercontent.com/2444641/199837420-80af577a-a9d7-4376-9f20-d98c95c1580b.png)

### Steps to reproduce

1. Import an account using a private key.
2. Attempt to import that same account again.
3. You will see this typo.

### Version

10.21.1

### Browser

Firefox

### Operating system

Windows

TimTinkers:
This is a minor bug and I've submitted pull requests to fix it.

HowardBraham:
Fixed with #15558

PR #16371, opened 2022-11-03
Fixed a duplicate account import test case relying on an upstream typo

TimTinkers:
## Explanation

This pull request in the upstream KeyringController https://github.com/MetaMask/KeyringController/pull/153 fixes the issue https://github.com/MetaMask/metamask-extension/issues/16370.

This current pull request fixes a test case such that the upstream KeyringController fix can be used.

Currently, users encounter a typo when attempting to import a duplicate account into MetaMask. The typo is not localized and originates from the upstream KeyringController. I have submitted a pull request here https://github.com/MetaMask/KeyringController/pull/153 to fix the typo, but doing so would break a test case in this repository which expects the typo. This pull request fixes the test case such that it will expect the typo-free error from KeyringController. This pull request should be merged after https://github.com/MetaMask/KeyringController/pull/153 because it is dependent on the typo being fixed there.

## More Information

I encountered this typo when attempting to import an account via its private key after that account had previously already been imported.

### Before

before

### After

The UI will look the same until after https://github.com/MetaMask/KeyringController/pull/153 is merged and incorporated. After that point, the end user will no longer see the "you're are" typo and will just see "you are".

## Manual Testing Steps

You can verify that this issue is present by importing an account from private key and then attempting to import the same account. You should no longer see the typo if using the new KeyringController.

## Pre-Merge Checklist


TimTinkers:

HowardBraham:
This was merged with #15558

PR #16385, opened 2022-11-04
Hide network picker when not unlocked

darkwing:
This is just an idea but right now we show the network dropdown when the wallet is locked, which doesn't seem to make much sense.


ryanml (dismissed):
I'm a fan of this behavior

FrederikBolding:
A dapp can still query the node for information even if MetaMask is locked, this change would make it impossible for a user to switch to the correct network that a dapp requires without unlocking 🤔

Gudahtt:
I think we need to allow switching before login for now, at least until decoupling the network from the login flow more effectively than we've done so far. This might prevent login if we remove it now (failing network blocks everything).

Issue #16314, opened 2022-10-30
[Bug]: [MV3] Cannot destructure balance after terminating service worker


seaona:
### Describe the bug

**Problem**: when I terminate the service worker after sending  a transaction, I land on the MM error page and see `Cannot destructure balance` error. This is an intermittent failure, but it appears quite commonly (+3 times for ~10 tries)

![cannot-destructure-balance](https://user-images.githubusercontent.com/54408225/198878355-57726127-43bb-4b7c-829b-d091be72e30f.png)

### Steps to reproduce

This is tested over [this PR](https://github.com/MetaMask/metamask-extension/pull/15558)
1. Unlock MM
2. Change to a testnet
3. Perform a Send tx
4. Terminate Service worker -- occasionally, you will be landed to the error page

### Version

https://github.com/MetaMask/metamask-extension/pull/15558

### Browser

Chrome

### Operating system

Linux

segun:
I tried to replicate this, but I couldn't, instead of this error, I'm getting another error that is similar

```
https://github.com/MetaMask/metamask-extension/blob/develop/ui/hooks/useGasFeeErrors.js#L285

Uncaught TypeError: Cannot read properties of undefined (reading 'balance')
The error may be correlated with this previous error:
TypeError: Cannot read properties of undefined (reading 'balance')
  at Object.isBalanceCached [as current] (chrome-extension://pfnkndiibjpendhnikmjbegomajogkom/common-9.js:3029:85)
  at Subscription.checkForUpdates [as onStateChange] (chrome-extension://pfnkndiibjpendhnikmjbegomajogkom/ui-4.js:33878:47)
  at Subscription.handleChangeWrapper (chrome-extension://pfnkndiibjpendhnikmjbegomajogkom/ui-4.js:34199:12)
  at chrome-extension://pfnkndiibjpendhnikmjbegomajogkom/ui-4.js:34125:20
  at batchedUpdates$1 (chrome-extension://pfnkndiibjpendhnikmjbegomajogkom/ui-4.js:24393:12)
  at Object.notify (chrome-extension://pfnkndiibjpendhnikmjbegomajogkom/ui-4.js:34121:7)
  at Subscription.notifyNestedSubs (chrome-extension://pfnkndiibjpendhnikmjbegomajogkom/ui-4.js:34194:20)
  at captureWrapperProps (chrome-extension://pfnkndiibjpendhnikmjbegomajogkom/ui-4.js:32881:5)
[...]

PR #16393, opened 2022-11-07
Fix state sync with UI

jpuri:
Fixes: #16314

PR fixes issue caused on Ui by state early sync. PR delays state sync to point when state has requires account and balance details.

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -2337,6 +2337,7 @@ export default class MetamaskController extends EventEmitter {
     else if (isManifestV3) {
       await this.submitEncryptionKey();
     }
+    await this.accountTracker._updateAccounts();
     this.emit('startUISync');
     this.memStore.subscribe(this.sendUpdate.bind(this));
   }
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -1096,15 +1096,9 @@ export default class MetamaskController extends EventEmitter {
       !this.isUnlocked() &&
       this.onboardingController.store.getState().completedOnboarding
     ) {
-      // Automatic login via config password
-      const password = process.env.CONF?.PASSWORD;
-      if (password) {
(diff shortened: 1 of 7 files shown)


jpuri:
@darkwing : I am getting this error on service worker restart:

Screenshot 2022-11-07 at 2 46 47 PM

watchanan554 (approved):

app/scripts/metamask-controller.js
```diff
@@ -1090,21 +1090,15 @@ export default class MetamaskController extends EventEmitter {
...
-        this.submitPassword(password);
-      }
-      // Automatic login via storage encryption key
-      else if (isManifestV3) {
-        this.submitEncryptionKey();
-      }
+      this._loginUser();
+    } else {
+      this.emit('startUISync');
+      this.memStore.subscribe(this.sendUpdate.bind(this));
```
jpuri, line 1101: `memStore` updates a lot as controllers initialise, delaying this `subscribe` helps to make sure that incomplete UI state is not flushed to UI.

app/scripts/ui.js
```diff
@@ -78,7 +78,7 @@ async function start() {
      * In case the UI is already rendered, only update the streams.
      */
     const messageListener = async (message) => {
-      if (message?.name === 'CONNECTION_READY') {
```
Gudahtt, line 81: Is the old `CONNECTION_READY` message still used for anything? Maybe it can be removed if not.

Gudahtt (commented):
This looks great! Good idea to delay state updates like that as well

app/scripts/ui.js
watchanan554: Hi

jpuri: I removed it.

Gudahtt (approved):
LGTM!

darkwing (approved):
OUTSTANDING @jpuri !  This is excellent!

PR #16410, opened 2022-11-08
Bundle optimization: dedupe bn.js

kumavis:
dedupe `bn.js` (many many copies, ~1MB) and fix junk data we're passing to it in tests

accidentally temporarily included an unrelated deduping of `@ethersproject/hash`, undid that change with its related policy file


brad-decker (dismissed):
Looks good to me, but as a heads up with the way yarn3 resolves things it may be impacted by my prs. I might tag you for a look at those once they get rebased to make sure i'm not undoing anything.

adonesky1:
@kumavis how does this work? Are the test changes the only instances of where our use of `bn.js` v4 behaves differently than v5?

kumavis:
@SocketSecurity ignore core-js@3.15.2

brad-decker (dismissed):

brad-decker:
@adonesky1 added QA externsion board label.

kumavis:
replaced by https://github.com/MetaMask/metamask-extension/pull/20199

PR #16417, opened 2022-11-08
Parallelize E2E tests to cut CI time in half

FrederikBolding:
## Explanation

Adds logic in `run-all.js` to allow for running E2E tests in parallel and enables parallelism in Circle CI.

Assuming the old configuration runs at around 60 minutes:
- With `parallelism: 4`, the CI now runs a full **22 minutes** faster than the old configuration.
- With `parallelism: 8`, the CI now runs a full **28 minutes** faster than the old configuration.

This means we can **effectively cut CI time in half with this PR**.

### DIFF
--- a/test/e2e/run-all.js
+++ b/test/e2e/run-all.js
@@ -17,7 +17,7 @@ const getTestPathsForTestDir = async (testDir) => {
 function chunk(array, chunkSize) {
   const result = [];
   for (let i = 0; i < array.length; i += chunkSize) {
-      result.push(array.slice(i, i + chunkSize));
+    result.push(array.slice(i, i + chunkSize));
   }
   return result;
 }
--- a/test/e2e/run-all.js
+++ b/test/e2e/run-all.js
@@ -13,6 +13,15 @@ const getTestPathsForTestDir = async (testDir) => {
   return testPaths;
 };
 
+// TODO: Simplify?
+function chunk(array, chunkSize) {
+  const result = [];
+  for (let i = 0; i < array.length; i += chunkSize) {
+      result.push(array.slice(i, i + chunkSize));
+  }
+  return result;
+}
+
 async function main() {
   const { argv } = yargs(hideBin(process.argv))
     .usage(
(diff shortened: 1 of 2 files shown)


test/e2e/run-all.js
```diff
@@ -13,6 +13,15 @@ const getTestPathsForTestDir = async (testDir) => {
   return testPaths;
 };
 
+// TODO: Simplify?
```
Gudahtt, line 16: This seems reasonably simple enough to me already! I tried a few alternatives and found nothing simpler.
```suggestion
```

Gudahtt: Oh, but there is this: https://lodash.com/docs/4.17.15#chunk

That would be simple.

test/e2e/run-all.js
```diff
@@ -66,7 +75,14 @@ async function main() {
     args.push('--retries', retries);
   }
 
-  for (const testPath of testPaths) {
+  // For running E2Es in parallel in CI
+  const currentChunkIndex = process.env.CIRCLE_NODE_INDEX ?? 0;
+  const totalChunks = process.env.CIRCLE_NODE_TOTAL ?? 1;
+  const chunkSize = testPaths.length / totalChunks;
```
Gudahtt, line 81: Hmm, won't this result in a float if the total isn't divisible by the total number of chunks? e.g. a total number of paths of 10 divided by a total chunk size of 3 would get 3.333... as the chunk size. Not ideal.

We'd want to round up, so that all paths were covered. So for 10 tests and 3 chunks, we'd want the chunk size to be 4.

```suggestion
  const chunkSize = Math.ceil(testPaths.length / totalChunks);
```

Gudahtt (commented):
Fantastic work! Found one potential problem, otherwise looks good

test/e2e/run-all.js
FrederikBolding: If you think it's simple enough to read as is I would prefer not importing a dep for it 😄

test/e2e/run-all.js
FrederikBolding: Screenshot 2022-11-08 at 14 32 55
All paths are covered regardless - but we can do `ceil` if you prefer 😅

Gudahtt: Well, that's an interesting result. Didn't realize `slice` would round down like that, and that it would effectively still work due to rounding. I was unable to find a case where this failed. Still I think `ceil` would be easier to read

Gudahtt (approved):

mcmire (approved):
Good thinking!

PR #16426, opened 2022-11-09
chain_id value in the Custom Network Added event

### LINKED ISSUE #16403: chain_id value is being passed as decimal in the Custom Network Added event
## Background
As we worked through the new popular network list feature, we updated the `Custom Network Added` event to better understand to usage of the different features that currently allow users to add networks to their wallet (PR https://github.com/MetaMask/metamask-extension/pull/15176). 
One of the properties in this event is the `chain_id` and we noticed some of the values for this property are incorrect. The incorrect values consist of `0x` + decimal value, instead of the expected hex value for the chain id. 
We suspect this is happening when a user adds a network through the custom network form and inputs chain id as decimal.
Example:
If the user inputs `137` as the chain id for Polygon, the `chain_id` in `Custom Network Added` event will be `0x137`, but it should actually be `0x89`.

The chain id for this event when users add networks through the dapp api and/or through the popular network list are working correctly.

## Task
We should investigate if that's indeed is the cause and fix this by passing the converted hex value in the property instead of the decimal one.

## Acceptance Criteria
1. chain_id values passed in the `Custom Network Added` event are the converted hex values even if users have inputted a decimal value as the chain id.

adnansahovic:
## Explanation

chain_id values passed in the Custom Network Added event are the converted hex values even if users have inputted a decimal value as the chain id.

Example:
If the user inputs 137 as the chain id for Polygon, the chain_id in Custom Network Added event will be 0x89.

## More Information
* Fixes #16403

### Before

https://user-images.githubusercontent.com/63151811/200844518-d2d0f9e2-c0e7-4027-a62a-7855949ef7fd.mov

### After

https://user-images.githubusercontent.com/63151811/200844529-479d9a7a-fe4e-484d-9c5c-03a5cbaa68cc.mov

### DIFF
--- a/ui/pages/settings/networks-tab/networks-form/networks-form.js
+++ b/ui/pages/settings/networks-tab/networks-form/networks-form.js
@@ -527,7 +527,7 @@ const NetworksForm = ({
             url: rpcUrlOrigin,
           },
           properties: {
-            chain_id: addHexPrefix(chainId.toString(16)),
+            chain_id: addHexPrefix(Number(chainId).toString(16)),
             network_name: networkName,
             network: rpcUrlOrigin,
             symbol: ticker,


brad-decker (approved):

darkwing (approved):

darkwing:
Thank you @adnansahovic !

PR #14644, opened 2022-05-09
Chain ID field use the same color pattern as the currency symbol field warning message below

VSaric:
## Explanation

Fixing `chain ID` field to use the same color pattern as the `currency symbol` field warning message below.

## More Information

* Fixes #13249

### Before

Screenshot 2022-05-09 at 11 53 52

### After

Screenshot 2022-05-09 at 11 54 44

## Manual Testing Steps

1. Unlock MetaMask
2. Go to Settings // Networks // Add network
3. Insert the same chain ID for some specific custom network

PR #16452, opened 2022-11-10
fix chainId validation on network form

adonesky1:
## Explanation
An issue was introduced [a few months back](https://github.com/MetaMask/metamask-extension/pull/14644) on the add network form where entering `chainId` that did not match the `chainId` returned by the `rpcUrl` didn't block form submission. This issue was introduced while attempting to allow users to add alternative networks (via the form) that had a chainId that matched an existing network (for example [a flashbots RPC provider for mainnet](https://medium.com/alchemistcoin/how-to-add-flashbots-protect-rpc-to-your-metamask-3f1412a16787)). We want to warn but not block users when they attempt to add networks with a chainId for which they already have a network added, but we still want to fully block them from adding networks where `chainId` and `rpcUrls` do not match eachother.

### Before

https://user-images.githubusercontent.com/34557516/201164095-2780562e-710d-476e-a65f-ad2497f0d054.mov

### After

https://user-images.githubusercontent.com/34557516/201163567-56562d3f-6500-44b6-8d70-2be92571037d.mov

## Manual Testing Steps
1. Navigate to Add Network Form
2. Add a Currency Ticker and Name
3. Add a valid RPC Url
4. Input a chainId that does not match this RPC
5. You should see that you are not able to submit the form

Also
1. Navigate to Add Network Form
2. Add a Currency Ticker and Name
3. Add a valid alternative RPC Url for a network you already have added
4. Input the chainId that **does** match this RPC
5. You should see a warning that `This Chain ID is currently used by the ____ network`
6. You should still be able to submit the form

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/ui/pages/settings/networks-tab/networks-form/networks-form.js
+++ b/ui/pages/settings/networks-tab/networks-form/networks-form.js
@@ -232,6 +232,8 @@ const NetworksForm = ({
       const formChainId = chainArg.trim();
       let errorKey = '';
       let errorMessage = '';
+      let warningKey = '';
+      let warningMessage = '';
       let radix = 10;
       let hexChainId = formChainId;
 
@@ -240,8 +242,10 @@ const NetworksForm = ({
           hexChainId = `0x${decimalToHex(hexChainId)}`;
         } catch (err) {
           return {
-            key: 'invalidHexNumber',
-            msg: t('invalidHexNumber'),
+            error: {
+              key: 'invalidHexNumber',
+              msg: t('invalidHexNumber'),
+            },
           };
         }
       }
@@ -253,8 +257,8 @@ const NetworksForm = ({
       if (formChainId === '') {
         return null;
       } else if (matchingChainId) {
-        errorKey = 'chainIdExistsErrorMsg';
-        errorMessage = t('chainIdExistsErrorMsg', [
+        warningKey = 'chainIdExistsErrorMsg';
+        warningMessage = t('chainIdExistsErrorMsg', [
           matchingChainId.label ?? matchingChainId.labelKey,
         ]);
       } else if (formChainId.startsWith('0x')) {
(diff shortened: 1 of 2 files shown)


Gudahtt (approved):
LGTM!

These changes look sensible. I tested this locally, and it seemed to fix the issue in most cases. I was still able to add a network with the wrong chain ID once somehow, but I suspect it's due to a separate problem that is older than the regression fixed here (perhaps that we're not disabling the Save button while making the initial check, I'm not sure).

seaona:
The behaviour above looks good. I've noticed that there is a slightly different behaviour if I try to add a Network manually vs adding it from a Dapp, in the case when I add a network with the same `ChainId` (i.e. flashbots).

https://user-images.githubusercontent.com/54408225/201319278-6f55b510-7137-4327-98a4-2903a637bc18.mp4

**Repro steps**
On the test dapp, change the pre-default network to Flashbots, and see that it's not possible to add, and the error `MetaMask - RPC Error: May not specify default MetaMask chain.` appears.

```
      method: 'wallet_addEthereumChain',
      params: [
        {
          chainId: '0x1',
          rpcUrls: ['https://rpc.flashbots.net'],
          chainName: 'Flashbots',
          nativeCurrency: { name: 'TEST', decimals: 18, symbol: 'TEST' },
          blockExplorerUrls: null,
        },
```
Let me know if this should be covered by this PR or if we need a separate issue for that @adonesky1

adonesky1:
> Let me know if this should be covered by this PR or if we need a separate issue for that
@seaona we should address this but I think it makes more sense to do so in a separate PR, since this PR fixes a regression that has security implications and the validation on `wallet_AddEthereumChain` does not.

seaona:
Make sense. I open a separate issue for that then. Thanks @adonesky1 !

FrederikBolding (approved):

brad-decker (approved):

PR #16459, opened 2022-11-11
lint - always include extensions for non-js imports

kumavis:
We currently prefer to omit file extensions from imports. idk why we do this!

but in confuses some tools, like Endo bundler, especially when we expect the extension to be autofilled with something like `.ts`. As a start, this rule requires us to list the extension whenever it does not resolve to `.js`

:warning:  [This rule currently does not have an autofixer](https://github.com/import-js/eslint-plugin-import/pull/2593).
For that reason, and because there are about ~230 issues, I'm leaving this in draft.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -96,6 +96,14 @@ module.exports = {
       parserOptions: {
         sourceType: 'module',
       },
+      rules: {
+        "import/extensions": [
+          "error",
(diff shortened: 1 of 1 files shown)


mcmire:
Just making sure I understand this correctly, would this enable extensions for all imports? Or just non-JS ones?

PR #15812, opened 2022-09-13
Fix display of edit button on confirm screen when sending tx with data to contract

danjm:
Fixes https://github.com/MetaMask/metamask-extension/issues/15465

The fix adds an "onEdit" proper to the base confirmation component used for contract interactions (the "token method" route). 

(Sidenote, the "token method" path is misnamed)

PR #16498, opened 2022-11-15
Prevent user from editing a contract interaction created by a dapp

danjm:
Fixes problem introduced in https://github.com/MetaMask/metamask-extension/pull/15812

As of that PR, it became possible for users to edit contract interact transactions. That is okay if the transaction originates from MetaMask, but not if the transaction originates from a dapp. Editing the recipient or the asset type can have unexpected consequences in that case, include allowing the user to transfer tokens to a smart contract address.

This PR prevents the user from editing contract interactions from a dapp. This was the way things were prior to v10.21.0

An e2e test is added to prevent such a regression in the future

### DIFF
--- a/test/e2e/tests/dapp-tx-edit.spec.js
+++ b/test/e2e/tests/dapp-tx-edit.spec.js
@@ -1,0 +1,110 @@
+const { strict: assert } = require('assert');
+const { convertToHexValue, withFixtures } = require('../helpers');
+const { SMART_CONTRACTS } = require('../seeder/smart-contracts');
+const FixtureBuilder = require('../fixture-builder');
+
+describe('Editing confirmations of dapp initiated contract interactions', function () {
+  const ganacheOptions = {
+    accounts: [
+      {
+        secretKey:
+          '0x7C9529A67102755B7E6102D6D950AC5D5863C98713805CEC576B945B15B71EAC',
+        balance: convertToHexValue(25000000000000000000),
+      },
+    ],
+  };
+  const smartContract = SMART_CONTRACTS.PIGGYBANK;
+  it('should NOT show an edit button on a contract interaction confirmation iniated by a dapp', async function () {
+    await withFixtures(
+      {
+        dapp: true,
+        fixtures: new FixtureBuilder()
+          .withPermissionControllerConnectedToTestDapp()
+          .build(),
+        ganacheOptions,
+        smartContract,
+        title: this.test.title,
+      },
+      async ({ driver, contractRegistry }) => {
+        const contractAddress = await contractRegistry.getContractAddress(
+          smartContract,
+        );
+        await driver.navigate();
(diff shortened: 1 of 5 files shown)


brad-decker (dismissed):
LGTM

ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
```diff
@@ -1158,11 +1158,13 @@ export default class ConfirmTransactionBase extends Component {
     } = this.getNavigateTxData();
 
     let functionType;
+    let isContractInteractionFromDapp = false;
```
darkwing, line 1161: Can we make this a `const` and set its value to:

```
const isContractInteractionFromDapp = txData.type === TRANSACTION_TYPES.CONTRACT_INTERACTION && txData.origin !== 'metamask';
```

...and use that for the `if`?

danjm: done in 358dacb0f5

Gudahtt:
>This PR prevents the user from editing contract interactions from a dapp. This was the way things were prior to v10.21.0

I don't think this is accurate. Contract interactions in general were not editable prior to v10.21.0, but simple send transactions suggested by dapps were editable.

test/e2e/tests/dapp-tx-edit.spec.js
```diff
@@ -0,0 +1,108 @@
...
+        await driver.waitUntilXWindowHandles(3);
+        const windowHandles = await driver.getAllWindowHandles();
+
+        await driver.switchToWindowWithTitle(
+          'MetaMask Notification',
+          windowHandles,
+        );
+        await driver.waitForSelector({
+          css: '.confirm-page-container-summary__action__name',
+          text: 'Sending ETH',
```
Gudahtt, line 95: I think this is the problem:
```suggestion
          text: 'Deposit',
```

Gudahtt: Actually maybe not... still looking into this

danjm: I don't think so, this is a simple send transaction. It shouldn't say "Deposit" any where?

danjm: So there is a problem with the creation of the confirmation

Gudahtt: Yeah this is not it. It's the previous test that's failing, and it's failing because the edit button is present but invisible.

Gudahtt (dismissed):
LGTM!

Gudahtt (approved):

brad-decker (approved):

darkwing (approved):

seaona:
From QA looks good!

PR #16527, opened 2022-11-16
Add mocks for E2E tests

dan437:
## Explanation
Add mocks for E2E tests, increase stability

### DIFF
--- a/test/e2e/swaps/shared.js
+++ b/test/e2e/swaps/shared.js
@@ -1,4 +1,5 @@
 const FixtureBuilder = require('../fixture-builder');
+const { veryLargeDelayMs } = require('../helpers');
 
 const ganacheOptions = {
   accounts: [
@@ -26,7 +27,7 @@ const loadSwaps = async (driver) => {
 
 const buildQuote = async (driver, options) => {
   await driver.fill('input[placeholder*="0"]', options.amount);
-  await driver.delay(1000); // Need an extra delay after typing an amount.
+  await driver.delay(veryLargeDelayMs); // Need an extra delay after typing an amount.
   await driver.clickElement(
     '[class="dropdown-search-list__closed-primary-label dropdown-search-list__select-default"]',
   );
--- a/test/e2e/swaps/shared.js
+++ b/test/e2e/swaps/shared.js
@@ -25,10 +25,8 @@ const loadSwaps = async (driver) => {
 };
 
 const buildQuote = async (driver, options) => {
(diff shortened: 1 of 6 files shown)


Gudahtt (dismissed):
LGTM!

test/e2e/swaps/shared.js
```diff
@@ -25,10 +25,8 @@ const loadSwaps = async (driver) => {
 };
 
 const buildQuote = async (driver, options) => {
-  await driver.clickElement(
-    '[class*="dropdown-search-list"] + div[class*="MuiFormControl-root MuiTextField-root"]',
-  );
   await driver.fill('input[placeholder*="0"]', options.amount);
+  await driver.delay(1000); // Need an extra delay after typing an amount.
```
seaona, line 29: We can try to see if using a delay can be avoided in a future PR

darkwing, line 29: Instead of this arbitrary second, could you use one of the variables in `test/e2e/helpers.js`?  Maybe `largeDelayMs`?

seaona (dismissed):

darkwing (dismissed):

test/e2e/swaps/swap-eth.spec.js
```diff
@@ -16,7 +16,12 @@ describe('Swap Eth for another Token', function () {
           amount: 2,
           swapTo: 'DAI',
         });
-        await driver.clickElement({ text: 'Review swap', tag: 'button' });
+        const reviewSwapButton = await driver.findElement(
```
Gudahtt, line 19: Nit: Instead of getting the element here and running assertions on it for the text and enabled status, we can _select_ for the element using those properties. That saves time and prevents race conditions within the tests (each time we ask WebDriver for something there's a round-trip involved, and properties can change while we wait).

Gudahtt: The same goes for the click as well. We can convert these four lines into one line, by using `clickElement` and including these details in the selector.

test/e2e/swaps/shared.js
Gudahtt: Maybe this can be done in a follow-up PR, since this is blocking all other PRs and the release that we're preparing, and any change would require another ~30 minute wait.

seaona (approved):

brad-decker (approved):

Gudahtt (approved):
LGTM!

PR #16530, opened 2022-11-16
[e2e] Fix RPC flaky test

seaona:
## Explanation
The `deletes custom RPC` test is often failing with the error `TimeoutError: Waiting for element to be located By(css selector, span .modal)`.
There are only 2 places where the modal could be not found. The first one is when we look for the visible element:
```
/ wait for confirm delete modal to be visible
        const confirmDeleteModal = await driver.findVisibleElement(
          'span .modal',
        );
```
The second one is when we wait for that element to be hidden:
```
await confirmDeleteModal.waitForElementState('hidden');
```
After some investigation, it seems that it's failing on the latter case. This means that when we are looking for this element state, the element has already disappeared from the DOM, making this undefined.

For solving this, we add a check, so only if element is still present, we wait until it disappears. Otherwise it means it has already disappeared, so no need to wait.

### Before

image

### After

image

## Manual Testing Steps
Checkout the project locally and run the test using:
`yarn test:e2e:single test/e2e/tests/custom-rpc-history.spec.js --browser=chrome`

## Pre-merge author checklist

- [X] I've clearly explained:
  - X ] How reviewers can test my changes


brad-decker (approved):

dan437 (approved):

adonesky1 (approved):
LGTM!

PR #16559, opened 2022-11-17
Simplify MV3 initialization

Gudahtt:
## Explanation

The MV3 initialization logic was complicated and introduced difficult-to-reproduce race conditions when connections are made during initialization.

It seems that problems were encountered after the UI tried to connect before the background was initialized. To address this, the initialization step was _delayed_ until after the first connection. That first connection was then passed into the initialization function, and setup properly after initialization had completed.

However, this special treatment is only given for the first connection. Subsequent connections that still occur during initialization would fail. This also results in the initialization being needlessly delayed, which is concerning given that our main performance goal is to speed it up.

## Manual Testing Steps

Standard regression testing steps should apply here.

I think this might solve race conditions relating to connections during initialization, but I have not confirmed this, nor do I have steps to reproduce such problems.

## Pre-merge author checklist

- [x] I've clearly explained:


app/scripts/background.js
```diff
@@ -463,7 +455,9 @@ function setupController(initState, initLangCode, remoteSourcePort) {
    *
    * @param {Port} remotePort - The port provided by a new context.
    */
-  function connectRemote(remotePort) {
+  async function connectRemote(remotePort) {
+    await isInitialized;
```
Gudahtt, line 459: This should guarantee that `connectRemote` always works correctly when called. It now waits until initialization has completed.

But this PR is still broken because the `connectRemote event listener is still added too late in the process. We don't listen until after initialization.

Gudahtt: Fixed by https://github.com/MetaMask/metamask-extension/pull/16559/commits/7038e6adbb70a6a50b416f5d11e827268f4c280e

segun (approved):
LGTM

jpuri (approved):

PR #16575, opened 2022-11-17
Create constants for all keyring types

darkwing:
I'm a huge fan of using constants for frequently used strings.  We can and should do this for keyring types.


brad-decker (approved):
LGTM

NiranjanaBinoy (approved):
LGTM

PR #16576, opened 2022-11-17
Fix bundle size diff message

Gudahtt:
The bundle size diff message is using the wrong point of comparision, leading to misleading results on feature branches that have been merged with `develop` since they were created.

When this feature was introduced, we went back and forth a few times on what we should be comparing the branch with to get an accurate bundle size comparison.

The first attempt used `develop` as the point of comparison, but that didn't work because it was a moving target, and because it didn't reflect the changes made on this branch. As bundle increases or decreases were added on `develop`, they would alter the diff on each feature PR.

Then we chose to use the fork-point of the branch, the commit of `develop` that the branch forked off of. This works for feature branches that don't merge in `develop`. But the minute `develop` gets merged in, then unrelated changes on `develop` affect the measurement.

The _most recent_ commit from `develop` on the current branch is a better comparison. Any difference between this commit and the feature branch in terms of bundle size would be attributable to the feature branch changes. This is what `merge-base` gives us.

## Manual Testing Steps

The diff message should show zero changes.

It's hard to test the branch cases directly because this always uses `develop` as the point of comparison, so we'll have to just watch future PRs to ensure the results make sense.

## Pre-merge author checklist

- [x] I've clearly explained:


Gudahtt:
The solution on `develop` came from this StackOverflow post: https://stackoverflow.com/questions/12516169/git-find-the-oldest-commit-of-mine-which-does-not-exist-in-theirs/12517430#12517430

It's the correct solution to the wrong problem. The diagram demonstrates the change being made here:


```
[other-branch]     H--I--J   
                  /       \
[develop]        A--B--C--M1--D--E--M2--F--G
                    \      \             \
[current-branch]     H---I--M3---J---K---M4--M--HEAD
```

Previously we'd be getting B, the fork point for the current branch. But when comparing B to HEAD, we're also counting the `develop` commits B, C, and M1 as being caused by the current branch, which isn't accurate.

After this PR, we'll be comparing M1 to HEAD.

At first I thought this would be _skipping_ the effects of commits H and I, but that's not really the case. They still are a part of the changeset in HEAD, they're there and accounted for. By comparing M1, we can be confident that everything between there and HEAD is a result of this branch.

brad-decker (approved):

digiwand (approved):
Nice 🙌

PR #16578, opened 2022-11-18
Create constant for zero value in hex

darkwing:
Zero value in hex should be a constant

### DIFF
--- a/.storybook/initial-states/transactions.js
+++ b/.storybook/initial-states/transactions.js
@@ -1,3 +1,5 @@
+import { HEX_ZERO_VALUE } from "../../shared/constants/hex";
+
 const MOCK_TX_TYPE = {
   CANCEL: 'cancel',
   CONTRACT_INTERACTION: 'contractInteraction',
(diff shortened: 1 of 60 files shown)


@
```diff
@@ -0,0 +1,71 @@
+Create constant for zero value in hex
```
brad-decker, line 1: 👀

brad-decker (changes requested):
@.

shared/constants/hex.js
```diff
@@ -0,0 +1,2 @@
+/* Represents a zero value in hex */
+export const HEX_ZERO_VALUE = '0x0';
```
brad-decker, line 2: Could we make this a typescript file? I think i'm going to start pushing for all new files in shared folder to be typescript, simply because they are often leaf node files.

@
darkwing: Weird!  Will remove!

brad-decker (dismissed):

brad-decker (approved):

NiranjanaBinoy:
Hi @darkwing, I just noticed that we are still using'0x0' instead of `HEX_ZERO_VALUE`  in certain locations in `metamask-controller`, `TransactionController`, some utils files, and migrations. Are they supposed to stay as `0x0` itself?

darkwing:
I meant for this to only be used for values like gas, etc -- not _all_ cases of `'0x0'`

Issue #16258, opened 2022-10-24
TX Security Providers: Warning Message


ElvirCe:

ElvirCe:
@bschorchit the original requirement was to create only one message with the copy: “OpenSea has reviewed this request and is flagging it as malicious. Approving it could lead to a loss of funds. Proceed with caution.”
I see that the design has been updated and will put the ticket on hold until we groom it out and clarify everything.

ElvirCe:
@bschorchit we will need the URL for the OpenSea link and also please note that the messages are a new component since we don't have them anywhere like this so the scope of the ticket will have to be expanded.

bschorchit:
url is this: https://opensea.io/

bschorchit:
> please note that the messages are a new component since we don't have them anywhere like this so the scope of the ticket will have to be expanded.
Thanks for pointing this out, Elvir. I agree with expanded scope, but just want to have this component reviewed by design system team before we add it. 
@georgewrmarshall could you kindly confirm if the warning component being used in the design matches our patterns and design system before we go ahead with it?
Here's the link for the designs: https://www.figma.com/file/Q69vHaA7Or7Rxz4vdzjVHC/Support-for-tx-security-providers-(MVP)?node-id=4986%3A26901&t=H3PgEpqvjHPr1hnS-0

ElvirCe:
We are blocked on this ticket until the Banner component gets merged.

bschorchit:
Banner has been merged, @ElvirCe
Thanks, @garrettbear!

PR #16584, opened 2022-11-18
Security provider check (OpenSea)

### LINKED ISSUE #16281: Wallet Tx Security Provider: Send the transaction or signature request payload information to a service provider
## **User story**

As a user, I want my unsigned transaction and signatures requests to be sent to my security provider so it can verify whether it’s malicious or not before I sign them. Any **eth_sendTransaction**, **eth_sign**, **personal_sign**, or **signTypedData_vXX** method request would be forwarded to SECURITY_PROVIDER_RPC_URL in the standard JSON RPC request format prior to rendering a transaction confirmation modal to the user.

Tasks:

1. Check wether tx security provider is turned on;
2. If yes, unsigned transaction or signature request payload (eth_rpc call + url) is sent to tx security provider;
3. Tx security provider sends back a boolean response wether it wants to flag it as malicious or not

filipsekulic:
## Explanation
The security provider is an optional feature for users, similar to an RPC provider for a new chain.  The security provider will be called before the wallet action is presented to the user, and can flag possible malicious transactions that the wallet provider can raise to the user. 
More information: https://www.notion.so/opensea/Wallet-Tx-Security-PoC-b6a926e1cb6448638d147696cd8f5937

* Fixes #16281

In order to use the security provider check, `Transaction security check` should be enabled from the `Experimental` tab.

This functionality will be fully implemented in the following PR (https://github.com/MetaMask/metamask-extension/issues/16258) where the warning message will be shown.

## Manual testing steps

1. Go to `Experimental` tab and enable the `Transaction security check`.
2. Go to `test dapp`.
3. Send a transaction or sign a message and the security provider check will be called.


app/scripts/lib/personal-message-manager.js
```diff
@@ -80,15 +82,17 @@ export default class PersonalMessageManager extends EventEmitter {
...
+  async addUnapprovedMessageAsync(msgParams, req) {
+    const msgId = await this.addUnapprovedMessage(msgParams, req);
+
     return new Promise((resolve, reject) => {
       if (!msgParams.from) {
         reject(
           new Error('MetaMask Message Signature: from field is required.'),
         );
         return;
       }
```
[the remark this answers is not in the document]
filipsekulic: I think it should stay here, but we can check it with @danjm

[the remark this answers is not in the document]
jpuri: Reason of my question is that `const msgId = await this.addUnapprovedMessage(msgParams, req);` is now moved before this check.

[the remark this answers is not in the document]
filipsekulic: You've got the point! 
I fixed it here: 504ce81a363c9f799bc8b325a150ab29cac5cdbf

app/scripts/controllers/transactions/index.js
```diff
@@ -769,6 +772,7 @@ export default class TransactionController extends EventEmitter {
    * @returns {txMeta}
    */
```
jpuri, line 773: Parameter name `method` is not intuitive.

app/scripts/controllers/transactions/index.js
```diff
@@ -848,6 +852,15 @@ export default class TransactionController extends EventEmitter {
       ? addHexPrefix(txMeta.txParams.value)
       : '0x0';
 
+    if (method) {
+      const flagAsDangerous = await this.securityProviderRequest(
```
jpuri, line 856: May be `if(method && this.securityProviderRequest)`.

app/scripts/controllers/transactions/index.test.js
```diff
@@ -1294,7 +1299,7 @@ describe('Transaction Controller', function () {
     });
 
     it('should add multiple transactions when called with different actionId', async function () {
-      const txMeta = await txController.addUnapprovedTransaction({
+      const txMeta = await txController.addUnapprovedTransaction(undefined, {
         from: selectedAddress,
         to: recipientAddress,
       });
```
jpuri, line 1305: Will be useful to have a test case where `method` is defined.

app/scripts/metamask-controller.js
```diff
@@ -4552,4 +4557,30 @@ export default class MetamaskController extends EventEmitter {
...
+
+    if (transactionSecurityCheckEnabled) {
+      try {
+        const { flagAsDangerous } = await securityProviderCheck(
+          requestData,
+          methodName,
+          chainId,
+          currentLocale,
+        );
+        return await flagAsDangerous;
```
jpuri, line 4577: Not sure why is this await `await flagAsDangerous` required.

app/scripts/controllers/transactions/index.js
filipsekulic: Fixed here: 352f1d9054aefd84dd03b992873db4bf61cebc39

app/scripts/controllers/transactions/index.test.js
filipsekulic: Fixed here: 352f1d9054aefd84dd03b992873db4bf61cebc39

app/scripts/metamask-controller.js
filipsekulic: Fixed here: 352f1d9054aefd84dd03b992873db4bf61cebc39

app/scripts/controllers/transactions/index.test.js
jpuri: Will be nice to have a test case that tests the functionality added by `txMethodType` argument. That is `opts.securityProviderRequest` should be called.

jpuri (commented):
Code changes look good, but will be nice to have some more test coverage as it touches important functionality.

app/scripts/controllers/transactions/index.test.js
filipsekulic: Fixed here: c082a3c47bd15eb384299535d71407032be752e8

app/scripts/controllers/transactions/index.test.js
```diff
@@ -1323,6 +1328,47 @@ describe('Transaction Controller', function () {
...
+          from: selectedAddress,
+          to: recipientAddress,
+        },
+      );
+
+      assert.ok(
+        'securityProviderResponse' in txMeta,
+        'should have a securityProviderResponse',
+      );
+    });
```
jpuri, line 1371: `flagAsDangerous` does not seems to be tested above.

app/scripts/lib/security-provider-helpers.js
```diff
@@ -0,0 +1,64 @@
...
+        Accept: 'application/json',
+        'Content-Type': 'application/json',
+        'X-API-KEY': 'NKYIN6cXkFaNnVIfzNx7s1z0p3b0B4SB6k29qA7n',
+      },
+      body: JSON.stringify(dataToValidate),
+    },
+  );
+
+  return await response.json();
+}
```
jpuri, line 64: test cases can be added for above new file.

jpuri (dismissed):
Code changes look good, some small non-blocking feedback there.

seaona:
The toggle looks good when I tested using the feature flag.

There are 2 problems related to the Search Function that might/might not be related to this PR:

- There is no matching when we use the description words for the search. On the other cases it does match with text description, so this might be something related with this PR?

https://user-images.githubusercontent.com/54408225/211298834-b544c32b-4e88-4c1c-bc91-e482bf5fa619.mp4

- The `transaction` and `security` words are matched but not the `check` word from the feature title. This seems related to an existing [issue](https://github.com/MetaMask/metamask-extension/issues/17100), but we should verify it, to make sure
 
image

filipsekulic:
> The toggle looks good when I tested using the feature flag.
> 
> There are 2 problems related to the Search Function that might/might not be related to this PR:
> 
> * There is no matching when we use the description words for the search. On the other cases it does match with text description, so this might be something related with this PR?
> 
>  search-security.mp4 
> * The `transaction` and `security` words are matched but not the `check` word from the feature title. This seems related to an existing [issue](https://github.com/MetaMask/metamask-extension/issues/17100), but we should verify it, to make sure
> 
> image

Hey @seaona! 
I rebased this PR. You can check the new design in the `Experimental` tab. I think the problem with description search was not introduced in this PR. I tried to reproduce the issue you mentioned from both the develop and this PR's branch. If I am right, there is a problem with all the the settings descriptions when searching it, not just the ones from the `Experimental` tab.

jpuri (approved):
LGTM

seaona:
@filipsekulic thank you! It looks indeed a separate issue with the Search, not related to these changes

blackdevelopa (approved):
LGTM. However, just noticed we have a failing CI.

filipsekulic:
> LGTM. However, just noticed we have a failing CI.

@blackdevelopa 
Thanks! I've just updated the branch.

app/scripts/lib/security-provider-helpers.js
brad-decker: Did we ever add test cases?

PR #16612, opened 2022-11-21
Fix UI sync

jpuri:
PR addresses comment: https://github.com/MetaMask/metamask-extension/pull/15558#discussion_r1025565775

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -2358,9 +2358,7 @@ export default class MetamaskController extends EventEmitter {
     // Message startUISync is used in MV3 to start syncing state with UI
     // Sending this message after login is completed helps to ensure that incomplete state without
     // account details are not flushed to UI.
-    if (this.activeControllerConnections > 0) {
-      this.emit('startUISync');
-    }
+    this.emit('startUISync');
     this.startUISync = true;
     this.memStore.subscribe(this.sendUpdate.bind(this));
   }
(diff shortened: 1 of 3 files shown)


app/scripts/metamask-controller.js
```diff
@@ -2351,14 +2350,19 @@ export default class MetamaskController extends EventEmitter {
...
+    // account details are not flushed to UI.
+    if (this.activeControllerConnections > 0) {
       this.emit('startUISync');
-      // During controllers initialisation memstore updates a lot, delaying this subscription
-      // helps to delay early flusing of state to UI.
-      this.memStore.subscribe(this.sendUpdate.bind(this));
     }
+    this.startUISync = true;
+    this.memStore.subscribe(this.sendUpdate.bind(this));
   }
```
jpuri, line 2366: The change is also useful in scenario when service worker restart before user started onboarding

app/scripts/metamask-controller.js
```diff
@@ -3526,6 +3528,9 @@ export default class MetamaskController extends EventEmitter {
       });
     };
     this.on('startUISync', startUISync);
+    if (this.startUISync) {
+      this.emit('startUISync');
+    }
```
jpuri, line 3533: In case event listeners were not created when `this.emit('startUISync')` was called, code above will take care to send event to UI.

Gudahtt: good catch!

Gudahtt: Nit: this might make more sense reformatted as something like

```
if (this.startUISync) {
  startUISync()
} else {
  this.on('startUISync', startUISync);
}
```

e.g. call sync now, otherwise schedule for later. As we don't need the event listener at all if the UI sync is already prepared.

Gudahtt (approved):
LGTM!

PR #16627, opened 2022-11-22
Remove callback from being saved in controller state

### LINKED ISSUE #16484: [Bug]: Firefox seems broken when loaded in the browser. Importing a Wallet on Firefox causes error `this.state is null`
### Describe the bug

**Problem**: loading MM on Firefox throws `TypeError: Network Error` on the console.

When I use the Import Wallet onboarding flow, and I introduce the seed-phrase I stay in the same seed-phrase page after clicking Submit several times. If I then refresh the browser, I land into the next screen, with the console error `Uncaught TypeError: this.state is null` on the `import-seed-phrase.component.js`

I cannot create any Accounts with this Wallet and FF breaks at some point.

https://user-images.githubusercontent.com/54408225/201682997-8e64b462-3d7a-4687-8286-7d77772d8a18.mp4

### Steps to reproduce

1. Load MM for the first time on Firefox
2. Click Import Wallet
3. Enter a seed-phrase
4. Click Submit -- see you stay on the same page
5. Click Refresh -- see you land into Secure Wallet page with the console error `this.state is null``

_No response_

### Version

10.23.0

### Build type

_No response_

### Browser

Firefox

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

jpuri:
Fixes: #16484

Remove callback from being saved in controller state.

### DIFF
--- a/app/scripts/controllers/metametrics.js
+++ b/app/scripts/controllers/metametrics.js
@@ -166,11 +166,9 @@ export default class MetaMetricsController {
 
     // Code below submits any pending segmentApiCalls to Segment if/when the controller is re-instantiated
     if (isManifestV3) {
-      Object.values(segmentApiCalls).forEach(
-        ({ eventType, payload, callback }) => {
-          this._submitSegmentAPICall(eventType, payload, callback);
-        },
-      );
+      Object.values(segmentApiCalls).forEach(({ eventType, payload }) => {
+        this._submitSegmentAPICall(eventType, payload);
+      });
     }
 
     // Close out event fragments that were created but not progressed. An
@@ -961,6 +959,11 @@ export default class MetaMetricsController {
   // Saving segmentApiCalls in controller store in MV3 ensures that events are tracked
   // even if service worker terminates before events are submiteed to segment.
   _submitSegmentAPICall(eventType, payload, callback) {
+    const { metaMetricsId, participateInMetaMetrics } = this.state;
+    if (!participateInMetaMetrics || !metaMetricsId) {
+      return;
+    }
+
     const messageId = payload.messageId || generateRandomId();
     let timestamp = new Date();
     if (payload.timestamp) {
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/metametrics.js
```diff
@@ -961,6 +959,11 @@ export default class MetaMetricsController {
   // Saving segmentApiCalls in controller store in MV3 ensures that events are tracked
   // even if service worker terminates before events are submiteed to segment.
   _submitSegmentAPICall(eventType, payload, callback) {
```
digiwand, line 961:
```suggestion
  _submitSegmentAPICall(eventType, payload) {
```

seaona:
I've QAd this PR and it solves indeed the FireFox issue. It works well both on Chrome (MV2+MV3) and Firefox (MV2)

digiwand (approved):
This resolves the issue for me as well. FF x macOS

app/scripts/controllers/metametrics.js
```diff
@@ -166,11 +166,9 @@ export default class MetaMetricsController {
 
     // Code below submits any pending segmentApiCalls to Segment if/when the controller is re-instantiated
     if (isManifestV3) {
-      Object.values(segmentApiCalls).forEach(
-        ({ eventType, payload, callback }) => {
-          this._submitSegmentAPICall(eventType, payload, callback);
-        },
-      );
+      Object.values(segmentApiCalls).forEach(({ eventType, payload }) => {
+        this._submitSegmentAPICall(eventType, payload);
```
Gudahtt, line 170: Why was the callback removed here?

Gudahtt: Oh right. There is no callback here :sweat_smile: Understood.

Gudahtt (approved):
LGTM, great find!

danjm (approved):
LGTM

PR #16648, opened 2022-11-23
Add .vscode settings to the repo

HowardBraham:
Currently in metamask-extension, we often have to write our imports like `import { x } from '../../../../a/b/c'`

This PR allows us to write `import { x } from 'a/b/c'` instead.  This is important because VSCode autocomplete import will always suggest this shorter form, with the path relative to the root package.json.

Documentation on this browserify option can be found at https://github.com/browserify/browserify#browserifyfiles--opts

> `opts.paths` is an array of directories that browserify searches when looking for modules which are not referenced using relative path. Can be absolute or relative to `basedir`. Equivalent of setting `NODE_PATH` environmental variable when calling `browserify` command.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,5 @@
+{
+  "recommendations": [
+    "rvest.vs-code-prettier-eslint"
+  ]
+}
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,5 +1,3 @@
 {
-  "recommendations": [
-    "rvest.vs-code-prettier-eslint"
-  ]
+  "recommendations": ["rvest.vs-code-prettier-eslint"]
 }
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,6 @@ package
 
 # IDEs
 .idea
-.vscode
 .sublime-project
 
 *.bak
--- a/.prettierignore
+++ b/.prettierignore
@@ -8,7 +8,6 @@ jest-coverage/
 storybook-build/
 app/vendor/**
 .nyc_output/**
-.vscode/**
 test/e2e/send-eth-with-private-key-test/**
 *.scss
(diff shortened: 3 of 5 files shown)


brad-decker:
@HowardBraham having used this option in past projects, It made things very convoluted to understand where things were coming from, and didn't provide a whole lot of value. I'm fairly positive that vscode can be configured to use relative imports for auto-complete because my experience is not matching yours (auto completes are in the ../.. relative format) 

I would rather us not adopt this, but I am open to the organization's feedback and we can roll with whatever makes sense to the majority.

HowardBraham:
@brad-decker I found the VSCode setting to use relative imports for auto-complete 

```"javascript.preferences.importModuleSpecifier": "relative"```

**Comment by brad-decker** [comment]
@HowardBraham so then it can be changed to absolute.

File: .vscode/extensions.json [hunk 1]
```diff
@@ -0,0 +1,5 @@
+{
+  "recommendations": [
+    "rvest.vs-code-prettier-eslint"
+  ]
+}
```
georgewrmarshall, line 5: May need to add line break here?

georgewrmarshall (dismissed):
Relative imports and formatting works using `option + shift + F` as well as formatting on save after disabling them both in user settings.

e.g. removing these from `settings.json` and restarting VS code
```
"editor.codeActionsOnSave": {
    "source.fixAll.eslint": false,
    "source.fixAll.stylelint": true
  },
"javascript.preferences.importModuleSpecifier": "relative"
```

.vscode/extensions.json
HowardBraham: Hah, Prettier was not getting applied to extensions.json itself

georgewrmarshall (approved):

brad-decker (approved):
LGTM!

PR #16651, opened 2022-11-23
Allow transactions with gas price or priority fee 0

### LINKED ISSUE #16442: Allow transactions with gas price or priority fee = 0
## Background
[Slack thread](https://consensys.slack.com/archives/C021UN08FBJ/p1668023238318159)
Currently if the user edits the suggested gas, we block them in the UI from entering the value of 0 for the gas price (on networks that don't support EIP 1559) and for priority fee (on networks that support EIP 1559). We shouldn't block those anymore as for some networks these are indeed valid values. 
The reason we probably blocked that initially is that because, for some networks, nodes might reject transactions with gas below a certain value (1 gwei for Ethereum and 30 gwei for Polygon) and sending a transaction with gas below this threshold would result in a rpc error and a failed transaction. 
Right now though by blocking the value of 0 we're not preventing these failed transactions from happening (as users can still send any number between 0 and 30 on Polygon for example), but we're preventing gasless transactions in custom networks that support those or requesting that users pay some value for priority fee for networks where they could set that value to 0.

I'll create a different ticket for us to better guide users when editing gas so they don't input values that will lead to failed transactions. This ticket will focused on removing the requirement of values different than 0 for gas price and priority fee.

## Task
1. Remove the requirement in the UI when editing gas to enter a value different than 0 for gas price and priority fee.

## Acceptance Criteria 
1. Users can submit transactions with priority fee = 0 
2. User can submit transactions with gas price = 0
3. No warning is shown that says `Max priority fee must be greater than 0 GWEI` and `Gas price must be greater than 0`
4. The same existing logic for the `Max priority fee is low for current network conditions` and other existing warnings applies when the values inputted are 0

## Steps to Reproduce
1. Initiate a transaction
2. Click on edit gas
5. Try changing the priority fee or the gas price to `0` and notice the red warning `Max priority fee must be greater than 0 GWEI` or `Gas price must be greater than 0` and the disabled button to move forward

filipsekulic:
## Explanation
This allows users submitting the transaction with `gas price 0` or `priority fee 0`.

* Fixes #16442 

### Before

https://user-images.githubusercontent.com/92310504/203506740-5861317b-a07f-4bb8-be2c-7dfe598f047b.mov

### After

https://user-images.githubusercontent.com/92310504/203796360-ded61910-d715-4fc7-bca3-7f571dec6dbf.mov

## Manual Testing Steps
`1st case:`
1. Send
2. Choose or enter an address
3. Enter or not some amount
4. Next
5. Edit (the lower one - within the transaction)
6. Advanced options
7. Set `Max priority fee` to 0

`2nd case`:
1. Settings
2. Experimental
3. Enable enhanced gas fee UI
4. Send
5. Choose or enter an address
6. Enter or not some amount
7. Next
8. Edit gas fee (`Low`, `Market`, `Aggressive`, `Advanced`)
9. Advanced
10. Set `Priority fee` to 0

`3rd case`:
1. Go to test dapp
2. Send legacy transaction
3. Edit (the lower one - within the transaction)
4. Edit suggested gas fee
5. Set `Gas price` to 0

### DIFF
--- a/ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-inputs/priority-fee-input/priority-fee-input.test.js
+++ b/ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-inputs/priority-fee-input/priority-fee-input.test.js
@@ -111,7 +111,7 @@ describe('PriorityfeeInput', () => {
     expect(screen.queryByText('2 - 125 GWEI')).toBeInTheDocument();
   });
 
-  it('should show error if value entered is 0', () => {
+  it('should not show error if value entered is 0', () => {
     render({
       txParams: {
         maxPriorityFeePerGas: '0x174876E800',
@@ -125,6 +125,6 @@ describe('PriorityfeeInput', () => {
     });
     expect(
       screen.queryByText('Priority fee must be greater than 0.'),
-    ).toBeInTheDocument();
+    ).not.toBeInTheDocument();
   });
 });
--- a/ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-inputs/priority-fee-input/priority-fee-input.test.js
+++ b/ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-inputs/priority-fee-input/priority-fee-input.test.js
@@ -127,4 +127,15 @@ describe('PriorityfeeInput', () => {
       screen.queryByText('Priority fee must be greater than 0.'),
     ).not.toBeInTheDocument();
   });
+
+  it('should not show the error if priority fee is 0', () => {
+    render({
+      txParams: {
+        maxPriorityFeePerGas: '0x0',
+      },
+    });
+    expect(
+      screen.queryByText('Priority fee must be greater than 0.'),
+    ).not.toBeInTheDocument();
+  });
 });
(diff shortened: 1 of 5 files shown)


mirjanaKukic:
Verified by QA

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-inputs/priority-fee-input/priority-fee-input.test.js
```diff
@@ -125,6 +125,6 @@ describe('PriorityfeeInput', () => {
     });
     expect(
       screen.queryByText('Priority fee must be greater than 0.'),
-    ).toBeInTheDocument();
+    ).not.toBeInTheDocument();
   });
 });
```
jpuri, line 130: Will. be nice to have a test case added for -ve value, since `0` is now a valid value.

jpuri (dismissed):
LGTM, small feedback added.

ui/components/app/advanced-gas-fee-popover/advanced-gas-fee-inputs/priority-fee-input/priority-fee-input.test.js
filipsekulic: Fixed here: 62765563566a08d874cf186ef0cd1933b5cad35c

jpuri (approved):

darkwing (approved):

PR #16664, opened 2022-11-23
Unstoppable Domains IPFS Resolution

V-Shadbolt:
## Explanation

Adds IPFS resolution for unstoppable domains. Much like the ENS IPFS resolution with .eth/ but adds support for the Unstoppable TLDs.

## Usage 

1. Simply type the UD into your browser’s address bar with a ‘/’ at the end
2. On Chrome, Brave or other chrome based browsers it should
- Resolve all UD TLDs EXCEPT .888
- Redirect to IPFS website if a site exists
- Redirect to UD search page if IPFS website does not exist
3. On Firefox it should
- Resolve all UD TLDs
- Redirect to UD search page if IPFS website does not exist

## More Information

Spinning off of this pull request: https://github.com/MetaMask/metamask-extension/pull/15817

## Pre-merge author checklist

- [ ] I've clearly explained:


yarn.lock
```diff
@@ -7293,7 +7309,7 @@ bn.js@>4.0.0, bn.js@^5.1.1, bn.js@^5.1.2, bn.js@^5.1.3, bn.js@^5.2.0, bn.js@^5.2
   resolved "https://registry.yarnpkg.com/bn.js/-/bn.js-5.2.1.tgz#0bc527a6a0d18d0aa8d5b0538ce4a77dccfa7b70"
   integrity sha512-eXRvHzWyYPBuB4NBy0cmYQjGitUrtqwbvlzP3G6VFnNRbsZQIxQ10PbKKHt8gZ/HW/D/747aDl+QkDqg3KQLMQ==
 
-bn.js@^4.0.0, bn.js@^4.1.0, bn.js@^4.1.1, bn.js@^4.11.0, bn.js@^4.11.6, bn.js@^4.11.7, bn.js@^4.11.8, bn.js@^4.11.9:
+bn.js@^4.0.0, bn.js@^4.1.0, bn.js@^4.1.1, bn.js@^4.11.0, bn.js@^4.11.6, bn.js@^4.11.7, bn.js@^4.11.8, bn.js@^4.11.9, bn.js@^4.4.0:
```
brad-decker, line 7312: @kumavis has a PR open to deduplicate a bunch of these #16410 -- Could you see if you can pin your package to use a version that is corresponding to that PR?

yarn.lock
```diff
@@ -9208,6 +9224,11 @@ crypto-js@^3.1.4:
   resolved "https://registry.yarnpkg.com/crypto-js/-/crypto-js-3.1.8.tgz#715f070bf6014f2ae992a98b3929258b713f08d5"
   integrity sha1-cV8HC/YBTyrpkqmLOSkli3E/CNU=
 
+crypto-js@^4.1.1:
+  version "4.1.1"
+  resolved "https://registry.yarnpkg.com/crypto-js/-/crypto-js-4.1.1.tgz#9e485bcf03521041bd85844786b83fb7619736cf"
+  integrity sha512-o2JlM7ydqd3Qk9CA0L4NL6mTzU2sdx96a+oOfPu8Mkl/PK51vSyoi8/rQ8NknZtk44vq15lmhAj9CIAGwgeWKw==
+
```
brad-decker, line 9231: Can you describe what crypto-js is used for in your package? Links to code is fine.

app/scripts/lib/ipfs-resolution/setup.js
```diff
@@ -31,26 +33,61 @@ export default function setupEnsIpfsResolver({
...
     const topLevelDomain = domainParts[domainParts.length - 1];
     // if unsupported TLD, abort
-    if (!supportedTopLevelDomains.includes(topLevelDomain)) {
+    if (!supportedEnsTopLevelDomains.includes(topLevelDomain)) {
       return;
     }
-    // otherwise attempt resolve
-    attemptResolve({ tabId, name, pathname, search, fragment });
+    // otherwise attempt resolve for ENS
+    attemptResolveEns({ tabId, name, pathname, search, fragment });
```
brad-decker, line 58: @bbehrman10 i'm a noob at IPFS resolution. Will a domain that was registered via ENS be resolved through the 'attemptResolveUns' method? I'm trying to get a sense for if the order you inserted the logic is going to result in a meaningful change to the amount of traffic that MetaMask sends ENS. If its feasible that this change will result in a decrease in ENS usage from within MetaMask then we need to provide a user preference toggle for ENS vs UD resolution and re-order our checks so that it follows that preference. In that case we need to default to ENS for that preference and put in a "whats new" thing explaining that they can now use UD. In any case we can always fall back to the other if it isn't resolved via the user's preference.

So in the case of ENS preferred we attempt ENS resolution and if not then UD, if UD is preferred, reverse the paradigm.

app/manifest/v2/_base.json
```diff
@@ -71,6 +71,16 @@
...
+    "*://*.crypto/",
+    "*://*.x/",
+    "*://*.nft/",
+    "*://*.dao/",
+    "*://*.bitcoin/",
+    "*://*.klever/",
+    "*://*.wallet/",
+    "*://*.blockchain/",
+    "*://*.888/",
+    "*://*.zil/",
```
brad-decker, line 83: We'll need to look into what changes we might need to make to manifest v3 as well

brad-decker (changes requested):

app/scripts/lib/ipfs-resolution/setup.js
bbehrman10: ENS addresses will resolve using the attemptResolveEns (it's the same function as it was before we just added the Ens identifier

app/manifest/v2/_base.json
bbehrman10: yeah, weren't sure what was needed in manifest v3

yarn.lock
bbehrman10: I can ask our dev team for specifics (might not get a clear answer) - I don't know what crypto.js does. Here's a link to their repo. https://github.com/brix/crypto-js

app/scripts/lib/ipfs-resolution/setup.js
bbehrman10: it will not impact ENS traffic at all

brad-decker: my question is if a .eth address is purchased via ENS and then we attempt to resolve it using UD what will the result be?

yarn.lock
brad-decker: i meant how its being used in these packages, not what the library does, just to clarify

app/scripts/lib/ipfs-resolution/setup.js
bbehrman10: the way it's written, .eth will always go to ENS and the UD TLDs will all go to UD

yarn.lock
V-Shadbolt: We're using the crypto.js package for hashing algorithms: eip137 (sha3) and zns  (sha256) in the resolution library.

bbehrman10: https://github.com/unstoppabledomains/resolution/blob/09079329a3545d0f344585d898f9f19cac4eeb9b/src/utils/namehash.ts

yarn.lock
V-Shadbolt: I'll dig into this Brad and see if we can't get it aligned

app/manifest/v2/_base.json
V-Shadbolt: Doesn't look like the ENS ipfs resolution needed to be in manifest v3 but we'll need to defer to you @brad-decker if this is something we need to add.

yarn.lock
V-Shadbolt: Hey Brad! 

We tried a couple things to pin the dependency but it looks like the node v16 (dictating the npm version) in the .nvmrc as well as the pinned yarn v1.6 in the package.json engines flag are too outdated to appropriately do this. 

Regarding the npm version: It looks like node v16 is using npm v8.17. The NPM overrides call dictated [here](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides) that can be used to pin nested dependencies is only supported on v8.30 and up. 

Regarding the yarn version: We tried doing different variations of this
```
"resolutions": {
    ...
    "@unstoppabledomains/resolution/**/bn.js": "^4.11.6"
  },
```
But because the unstoppable package name includes a '/' (@unstoppabledomains/resolution) is seems there is a bug in yarn that won't recognize the pin. It was fixed in yarn 2.0 and wasn't backported. related yarn issue [here](https://github.com/yarnpkg/yarn/issues/4874)

Very open to any ideas you might have to pin as the devDependency for the resolution library uses bn.js v4.11.6 which is referenced in #16410 so if we can find a way to pin the dependency to that we should be OK.

V-Shadbolt:
Hey @brad-decker ! Just wanted to ping incase the above got missed. Looking forward to the feedback!

V-Shadbolt:
Hey @brad-decker ! Just following up. Have you had the opportunity to look through the above?

V-Shadbolt:
Hey @brad-decker. Any update for @bbehrman10 and I?

app/scripts/lib/ipfs-resolution/setup.js
makoto: Hi Makoto from ENS team. Please be aware that ENS is not just `.eth`  but the users can import DNS names as ENS, such as .xyz, .io, etc.

https://medium.com/the-ethereum-name-service/announcing-support-for-xyz-on-ens-7f5bc7fe1b24
https://medium.com/the-ethereum-name-service/full-dns-namespace-integration-to-ens-now-on-mainnet-9d37270807d3

Even if UD never extends to `.eth`, it has the danger of future conflicts with the DNS imported ENS names whenever ICANN extends tlds (the last extension happened back in 2012 so the new round is expected to happen soon).

PR #16676, opened 2022-11-24
Display large and small numbers as decimals instead of scientific notation on token allowance confirmation screens

### LINKED ISSUE #16427: Display large and small numbers as decimals instead of scientific notation on token allowance confirmation screens
## Background
The new token allowance flow is ready 🤩🚀 and to better inform users of the exact value the dapp is proposing to them, we should display larger and small numbers as decimals instead of scientific notation. This is important as during user testing, seeing the whole number had a significant impact on the users and we're also frequently told the scientific notation we currently show is not really readable.

## Acceptance criterias
1. Very large numbers are shown as decimals instead of scientific notation. 
2. Very small numbers are also shown as decimals instead of scientific notation.

## Steps to reproduce
1. Go to https://etherscan.io/token/0x6b175474e89094c44da98b954eedeac495271d0f#writeContract
2. Connect your wallet
3. Under `approve` input any address and a number like `98768451321654687654231324674654213213435453435` or `98768451321`
4. Click on write
5. On the MetaMask popup, click on `Use default` or paste in that same number in the spending cap field
6. Notice how the number is displayed in scientific notation

adnansahovic:
## Explanation

Very large numbers are shown as decimals instead of scientific notation.
Very small numbers are also shown as decimals instead of scientific notation.

* Fixes #16427

### Before

https://user-images.githubusercontent.com/63151811/204797905-0e2a0d90-8c3a-430b-9a67-e09416a82927.mov

https://user-images.githubusercontent.com/63151811/204799265-abeee06c-3b86-4a09-8d19-ca9cbbef96a2.mov

### After

https://user-images.githubusercontent.com/63151811/204801910-57811c8d-7694-44f0-b803-4c425ad84ed8.mov

### Etherscan
1. Unlock MetaMask
2. Go to `Settings -> Experimental`
3. Turn `ON` toggle button for `Improved token allowance experience`
4. Go to https://etherscan.io/token/0x6b175474e89094c44da98b954eedeac495271d0f#writeContract
5. Connect your wallet
6. Under approve input any address and a number like `98768451321654687654` or `98768451321` or any other number 
7. Click on write
8. On the MetaMask popup, click on `Use default` or paste in that same number in the spending cap field
9. Notice how the number is displayed correctly.

### Test-Dapp
1. Unlock MetaMask
2. Go to `Settings -> Experimental`
3. Turn `ON` toggle button for `Improved token allowance experience`
4. Open the test dapp https://metamask.github.io/test-dapp/
5. Connect your wallet (click on `Connect` button in `Basic Actions` section)
6. Under the `Send Tokens` section, click on `Create Token`
7. Wait for transaction to be confirmed and click on `Approve Tokens`
or
8. Wait for transaction to be confirmed and click on `Approve Tokens Without Gas`
9. Wait for `Token Allowance` screen is displayed and rendered.
10. Enter a number, something like `98768451321654687654` or `98768451321` or any other number, and notice how the number is displayed correctly.


mirjanaKukic:
Verified by QA

bschorchit:
Ooh, thank you for finding and fixing that bug too! 👌

ui/components/app/custom-spending-cap/custom-spending-cap.js
```diff
@@ -146,25 +159,24 @@ export default function CustomSpendingCap({
...
             }
           >
             <FormField
-              numeric
               dataTestId="custom-spending-cap-input"
               autoFocus
               wrappingLabelProps={{ as: 'div' }}
               id={
-                value > (currentTokenBalance || error)
+                Number(value) > (currentTokenBalance || error)
```
jpuri, line 172: I have a doubt, will `Number(...)` be able to take care of very big numbers ?

adnansahovic: Hi @jpuri, See the current situation in the video. This is an enormous number... Do you have any suggestions or should it stay as it is currently on develop?

https://user-images.githubusercontent.com/63151811/206159552-1c0953eb-338b-4e52-89f1-ed25d7b0fadc.mov

brad-decker: The suggestion would be that we should use our number utilities, first off can you tell me what *type* of value 'currentTokenBalance' is? Is it a plain javascript number? Typically we have these as BigNumbers or BNs or even strings that need to be converted to BigNumbers/BN so that they can be compared/manipulated. 

Currently if  a user happens to have more than `9007199254740991` of a token the conversion to a number will break down the accuracy of the number and the comparison against their  total will be incorrect. That is what @jpuri is saying when she asks if it can handle very big numbers.

Number('10000000000000000000000000000000') will return a number but the accuracy of the conversion is not guaranteed and comparisons with that number and another number are also not guaranteed. MAX_SAFE_INTEGER is a constant specified by the language constraints of JavaScript. Anything over that target `9007199254740991` (currently) will not guarantee accurate results. So the advice here would be to convert the strings using `conversionGreaterThan` from `shared/modules/conversion.utils.js` which has examples throughout the codebase on usage, here is one:
```
  return conversionGreaterThan(
    { value: minimumTxCostInHexWei, fromNumericBase: 'hex' },
    { value: ethBalance, fromNumericBase: 'hex' },
  );
```

in this we are converting two numbers that are hexes to BigNumber types so that they can be accurately compared regardless of how high the value goes.

danjm: @VSaric @adnansahovic I agree with @brad-decker here

It would be good if we could replace all cases of ` > ` comparisons with our helper function that uses bignumber internally `conversionGreaterThan`

To simplify things in these files, you could create a wrapper around the `conversionGreaterThan` function that sets `fromNumericBase` to `'dec'` for both numbers, so that you can just call something like `decConversionGreaterThan(tokenValue, currentTokenBalance)`

Or that might be unecessary if you do the comparison once and store the result in a variable and then use that variable throughout the file (because it seems we are doing the same comparison throughout the file)

ui/components/app/custom-spending-cap/custom-spending-cap.js
```diff
@@ -146,25 +182,43 @@ export default function CustomSpendingCap({
         >
           <label
             htmlFor={
-              value > (currentTokenBalance || error)
+              conversionGreaterThan(
```
danjm, line 185: The same call appears three times in this file:
```
conversionGreaterThan(
                { value: Number(value), fromNumericBase: 'dec' },
                { value: Number(currentTokenBalance), fromNumericBase: 'dec' },
              )
```
Could you store that in a variable, and then use the variable throughout?
```
const valueIsGreaterThanBalance = conversionGreaterThan(
                { value: Number(value), fromNumericBase: 'dec' },
                { value: Number(currentTokenBalance), fromNumericBase: 'dec' },
              )
              ```

File: ui/components/ui/review-spending-cap/review-spending-cap.js [hunk 3]
```diff
@@ -64,7 +65,13 @@ export default function ReviewSpendingCap({
                   color={COLORS.TEXT_ALTERNATIVE}
                   className="review-spending-cap__heading-title__tooltip"
                 >
-                  {tokenValue > currentTokenBalance &&
+                  {conversionGreaterThan(
```
danjm, line 68: Same as my comment above, store the result of this `conversionGreaterThan` call in a variable and use it through the component

ui/pages/token-allowance/token-allowance.js
```diff
@@ -94,26 +98,24 @@ export default function TokenAllowance({
...
+  const customTxParamsData = Number(customTokenAmount)
     ? getCustomTxParamsData(data, {
         customPermissionAmount,
         decimals,
       })
     : null;
 
-  let fullTxData = { ...txData };
-
-  if (customTxParamsData) {
```
danjm, line 108: Why is this if block that modifies the `fullTxData` to include `customTxParamsData` being removed?

danjm (changes requested):
Most of the code looks good, left a question about code removal in token-allowance.js

ui/pages/token-allowance/token-allowance.js
danjm: @VSaric do you know why this code is removed?

danjm: Okay, I can't see VSaric's comment here, but it was explained in slack that this removal was meant to fix a bug whereby if decimal numbers are entered for a custom amount, the approval transaction fails.

The problem with this code removal is that this is the code that adds the custom amount to the transaction data. So if this code is removed, the approval amount that is submitted will be the dapp proposed or default amount, not the custom amount the user entered.

danjm: the bug this intended to fix is shown in the attached video. The problem is that we are not properly converting decimals numbers with precision < 0 to hex numbers. The proper fix to this bug is to fix that problem.


Uploading Screen Recording 2022-12-20 at 15.59.45.mov…

danjm: In fact, I think the problem here is that the test case that resulted in the error entered more decimal numbers than the token allows, which causes the problem with a decimal point within the hex value/data.

So the proper fix here is to prevent the user from enter more numbers in the decimal place than the token has decimals.

So for example @VSaric, if the token's decimals are 4 (as is the case of the contract deployment from our test dapp), then the user should not be allow to enter `1.12345` but should be allowed to enter `1.1234`

Can you remove this code change so that `fullTxData` is set as it was before, and then create a separate PR for validating / restricting the number of digits after the decimal point that the user can enter?

VSaric: Yes, I will set `fullTxData` as it was before right away! Thank you @danjm  for clarifying this!

VSaric: @danjm I added `replaceCommaToDot` function and set `fullTxData` as it was before in my latest commit. Now users can enter numbers with comma (`,`) and dot (`.`) and it will work as expected. If users entered, for example, `1,1234` transaction data will change. The approval amount that is submitted will be the custom amount the user entered, so it will adds the custom amount to the transaction data. Done [here](https://github.com/MetaMask/metamask-extension/pull/16676/commits/17e5277ca6c596eaf572ce445866a12adf5d9067).

VSaric: I also started to work for validating / restricting the number of digits after the decimal point that the user can enter like @danjm mention [here](https://github.com/MetaMask/metamask-extension/pull/16676#discussion_r1054314334) and I will create a separate PR for this.

jpuri:
Code change look good, nice work. Since this is important change. I would suggest another round of QA on the PR before merging.

bschorchit:
> Code change look good, nice work. Since this is important change. I would suggest another round of QA on the PR before merging.

@seaona could you help us with that?

ui/pages/token-allowance/token-allowance.js
```diff
@@ -50,6 +51,9 @@ import { useGasFeeContext } from '../../contexts/gasFee';
 import { getCustomTxParamsData } from '../confirm-approve/confirm-approve.util';
 import { setCustomTokenAmount } from '../../ducks/app/app';
 import { valuesFor } from '../../helpers/utils/util';
+import { calcTokenAmount } from '../../../shared/lib/transactions-controller-utils';
+
+const MAX_UNSIGNED_256_INT = new BigNumber(2).pow(256).minus(1).toString(10);
```
brad-decker, line 56: We have now duplicated this BigNumber, can we elevate this to a constant in shared/tokens maybe? I'm assuming this is the maximum allowable value for a token allowance? If thats the case can we do something like: MAX_TOKEN_ALLOWANCE_AMOUNT so that its clearer what the value is?

brad-decker (changes requested):

ui/pages/token-allowance/token-allowance.js
jpuri: I agree, small generic methods and also constants like `const regex = /^[0-9]{1,}([,.][0-9]{1,})?$/u;` can be moved to shared code.

VSaric: @brad-decker @jpuri Done [here](https://github.com/MetaMask/metamask-extension/pull/16676/commits/0b21618422d67ff022882000efc594ebd62fbd2c).

shared/constants/tokens.js
```diff
@@ -39,3 +40,8 @@ export const STATIC_MAINNET_TOKEN_LIST = Object.keys(contractMap).reduce(
 
 export const TOKEN_API_METASWAP_CODEFI_URL =
   'https://token-api.metaswap.codefi.network/tokens/';
+export const MAX_TOKEN_ALLOWANCE_AMOUNT = new BigNumber(2)
+  .pow(256)
+  .minus(1)
+  .toString(10);
+export const regex = /^[0-9]{1,}([,.][0-9]{1,})?$/u;
```
brad-decker, line 47: regex is not a descriptive name for this, and also the standard is upper case for constants

brad-decker (changes requested):

shared/constants/tokens.js
VSaric: @brad-decker You are right, sorry for the poor name. Do you know maybe how can I call this? Maybe INPUT_VALUE_REGEX or TOKEN_ALLOWANCE_VALUE_REGEX?

VSaric: In my [latest commit](https://github.com/MetaMask/metamask-extension/pull/16676/commits/1aef7377e096dc77adfbdc45f031da6c00477d00) I renamed that constant to `TOKEN_ALLOWANCE_VALUE_REGEX`. If you have some better suggestion please comment and I will rename it again. Thank you in advance @brad-decker!

brad-decker (dismissed):

dzfjz:
Verified by QA

ui/pages/token-allowance/token-allowance.js
```diff
@@ -94,9 +97,21 @@ export default function TokenAllowance({
...
+  const maxTokenAmount = calcTokenAmount(MAX_TOKEN_ALLOWANCE_AMOUNT, decimals);
+  if (customTokenAmount.length > 1 && Number(customTokenAmount)) {
+    const customSpendLimitNumber = new BigNumber(customTokenAmount);
+    if (customSpendLimitNumber.greaterThan(maxTokenAmount)) {
+      customPermissionAmount = 0;
+    }
+  }
 
-  const customTxParamsData = customTokenAmount
+  const customTxParamsData = Number(replaceCommaToDot(customTokenAmount))
```
danjm, line 114: can we just check `customPermissionAmount` here? instead of `Number(replaceCommaToDot(customTokenAmount))`?

VSaric: Yes we can @danjm! Thank you! Done [here](https://github.com/MetaMask/metamask-extension/pull/16676/commits/b86cffca83aec2a922db38f8e7bb36936cf9ac0d).

brad-decker (approved):

danjm (approved):
LGTM, I did some manual QA as well

PR #16682, opened 2022-11-24
Added eth spec as npm package and use it for testing

BelfordZ:
makes use of https://raw.githubusercontent.com/ethereum/execution-apis/assembled-spec/openrpc.json in testing network clients


BelfordZ:
Seems like the depcheck wont pass until https://github.com/ethereum/execution-apis/issues/71 is implemented

Gudahtt:
This looks great!

I'm not excited about us pulling in a git dependency though. We can publish it to npm ourselves under our namespace if they won't.

PR #16684, opened 2022-11-25
Not use hardware keyring classes when initializing KeyringController in extension MV3 version

### LINKED ISSUE #16674: [Bug]: login fails when moving from mv2 to mv3
### Describe the bug

The changes to keep user loggedin in MV3 are merged now [PR](https://github.com/MetaMask/metamask-extension/pull/15558).

An issue now noticed is that if user is moving from mv2 to mv3 extension they are not able to login into extension but need to use seed phrase to recover their account once. Login gives following error:

Error Image

More details here: https://github.com/MetaMask/metamask-extension/pull/15558#issuecomment-1324500449

This issue is blocker for MV3 Beta release.

jpuri:
Partially fixes: https://github.com/MetaMask/metamask-extension/issues/16674
For fix to be completed we need to merge this PR also: https://github.com/MetaMask/KeyringController/pull/169/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R700

For MV3 we do not pass hardware wallet classes to keyring controller.

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -585,12 +585,15 @@ export default class MetamaskController extends EventEmitter {
       await opts.openPopup();
     });
 
-    const additionalKeyrings = [
-      TrezorKeyring,
-      LedgerBridgeKeyring,
-      LatticeKeyring,
-      QRHardwareKeyring,
-    ];
+    let additionalKeyrings = [];
+    if (!isManifestV3) {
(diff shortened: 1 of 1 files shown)


darkwing (approved):
This is a really nice solution!  One suggestion: could we add a comment here as to why we're doing this?

brad-decker (approved):

PR #16693, opened 2022-11-25
Moved subscribe and filter into network controller

### LINKED ISSUE #16573: Move subscription/filter middleware into network client
Our middleware for handling subscription and filter-related methods (`eth-json-rpc-filters`) currently lives in our RPC pipeline _ahead_ of the network stack. This is network middleware, it belongs in the network client with the rest of the network middleware.

BelfordZ:
Our middleware for handling subscription and filter-related methods (`eth-json-rpc-filters`) currently lives in our RPC pipeline ahead of the network stack. This commit moves this middleware to the network client middleware instead. There are two reasons for this change. First, this middleware wraps RPC methods that are supported by the network. Second, it is necessary for this middleware to live with the network client so that it will aid us in unifying the NetworkController in this repo and the NetworkController in the `controllers` repo.

---

Fixes: #16573


app/scripts/controllers/network/createJsonRpcClient.js
```diff
@@ -19,11 +21,22 @@ const getTestMiddlewares = () => {
 export default function createJsonRpcClient({ rpcUrl, chainId }) {
   const fetchMiddleware = createFetchMiddleware({ rpcUrl });
   const blockProvider = providerFromMiddleware(fetchMiddleware);
+
+  // create subscription polyfill middleware
```
Gudahtt, line 25: Nit: was this comment added in the wrong place? Was there a reason it was added here but not in the Infura client?

Gudahtt (dismissed):
LGTM!

Though we should manually test that this works (I haven't done that yet). I don't think these methods are covered by any e2e tests or our test dapp, unfortunately.

Gudahtt:
We should come up with test instructions so that QA has something to go by.

mcmire (dismissed):
Anything that we add to the network client files we need to make sure is covered in the network client integration tests.

I can work on this.

mcmire:
@Gudahtt Sure, as long as we don't forget about it.

mcmire (approved):

Gudahtt:
@mcmire I added a ticket here: https://github.com/MetaMask/metamask-extension/issues/16899

brad-decker:
@BelfordZ is this ready to be merged?

mcmire:
I was asked to help with manually testing this PR, and to do this, I've been adding some more buttons and things to the test dapp so we can test the filter and subscription middleware. However, I don't think all of the filter middleware works as intended. I've added a couple of issues around this: https://github.com/MetaMask/eth-json-rpc-filters/issues/80, https://github.com/MetaMask/eth-json-rpc-filters/issues/81. So I'm not able to test all of the functionality but only some of it. I haven't tested the subscription middleware yet, but I will do that tomorrow.

mcmire:
Here are the changes for the test dapp which allow us to test filters and subscriptions: https://github.com/MetaMask/test-dapp/pull/202

I used this on this branch and it seems that there are some issues:

* When creating a log filter, I get this JSON-RPC response: `{"jsonrpc":"2.0","error":{"code":-32600,"message":"invalid json request"}}`
* Creating a block filter works just fine
* When creating a subscription (either way), I get this response: `The method "eth_subscribe" does not exist / is not available.`

mcmire (dismissed):
Marking this as changes requested. I can look into this tomorrow.

app/scripts/metamask-controller.js
```diff
@@ -3920,8 +3906,6 @@ export default class MetamaskController extends EventEmitter {
     ///: END:ONLY_INCLUDE_IN
 
     // filter and subscription polyfills
```
Gudahtt, line 3908: This comment was regarding the removed lines - it should also be removed.
```suggestion
```

app/scripts/controllers/network/createInfuraClient.js
```diff
@@ -31,8 +42,10 @@ export default function createInfuraClient({ network, projectId }) {
     createRetryOnEmptyMiddleware({ blockTracker, provider: infuraProvider }),
     createBlockTrackerInspectorMiddleware({ blockTracker }),
     infuraMiddleware,
+    filterMiddleware,
```
Gudahtt, line 45: Oh no! I just realized, the order has changed here. These should be first in the list, but they're last. That probably explains the errors.

mcmire: Ah :)

app/scripts/controllers/network/createInfuraClient.js
```diff
@@ -23,6 +25,15 @@ export default function createInfuraClient({ network, projectId }) {
   const infuraProvider = providerFromMiddleware(infuraMiddleware);
   const blockTracker = new PollingBlockTracker({ provider: infuraProvider });
 
+  const filterMiddleware = createFilterMiddleware({
+    infuraProvider,
+    blockTracker,
+  });
+  const subscriptionManager = createSubscriptionManager({
+    infuraProvider,
```
Gudahtt, line 33: The provider parameters have the wrong names here:
```suggestion
    provider: infuraProvider,
    blockTracker,
  });
  const subscriptionManager = createSubscriptionManager({
    provider: infuraProvider,
```

app/scripts/controllers/network/createJsonRpcClient.js
```diff
@@ -19,11 +21,22 @@ const getTestMiddlewares = () => {
...
     ...blockTrackerOpts,
     provider: blockProvider,
   });
 
+  const filterMiddleware = createFilterMiddleware({
+    blockProvider,
+    blockTracker,
+  });
+  const subscriptionManager = createSubscriptionManager({
+    blockProvider,
```
Gudahtt, line 36: The provider parameters have the wrong names here:

```suggestion
    provider: blockProvider,
    blockTracker,
  });
  const subscriptionManager = createSubscriptionManager({
    provider: blockProvider,
```

Gudahtt (commented):
It seems that the subscription middleware is failing because it used to be _before_ the permission middleware, but now it's _after_, and the subscription methods are missing from the list of unrestricted methods (in `app/scripts/controllers/permissions/specifications.js`, the `unrestrictedMethods` variable).

Gudahtt:
In testing out this PR, even with the suggestions I made above applied locally, I'm still having trouble getting subscriptions to work. It looks like there are two distinct problems: notifications are not propagating correctly from the network client to the dapp, and we're seeing occasional empty block responses from Infura that are resulting in the request failing.

For the empty block responses problem, this is happening due to a race condition between `getLatestBlock` and `getBlockByNumber` - there's a period of time during which the latest block number cannot be retrieved by `getBlockByNumber`,it lags behind slightly. This is what the `retryOnEmpty` middleware was doing for us - it would automatically retry.

We'll need to move these middleware up a level, into the network controller, so that we can pass them a `provider` that includes the `retryOnEmpty` middleware.

app/scripts/metamask-controller.js
```diff
@@ -3920,8 +3906,6 @@ export default class MetamaskController extends EventEmitter {
     ///: END:ONLY_INCLUDE_IN
```
Gudahtt, line 3907: I was able to fix the subscriptions by forwarding logs from the provider to this provider engine, like this:

```suggestion

    // forward notifications from network provider
    provider.on('data', (error, message) => {
      if (error) {
        // This should never happen, this error parameter is never set
        throw error;
      }
      engine.emit('notification', message);
    });
```

mcmire: Moved this up a bit but added here: [469bbed](https://github.com/MetaMask/metamask-extension/pull/16693/commits/469bbed580be40d466cc76a97d7910f727eea5cb)

app/scripts/controllers/network/createJsonRpcClient.js
mcmire: Corrected here: [469bbed](https://github.com/MetaMask/metamask-extension/pull/16693/commits/469bbed580be40d466cc76a97d7910f727eea5cb)

app/scripts/controllers/network/createInfuraClient.js
mcmire: Corrected here: [469bbed](https://github.com/MetaMask/metamask-extension/pull/16693/commits/469bbed580be40d466cc76a97d7910f727eea5cb)

app/scripts/controllers/network/createInfuraClient.js
mcmire: Made sure to put these first here: [469bbed](https://github.com/MetaMask/metamask-extension/pull/16693/commits/469bbed580be40d466cc76a97d7910f727eea5cb)

app/scripts/metamask-controller.js
mcmire: Removed here: [e51455a](https://github.com/MetaMask/metamask-extension/pull/16693/commits/e51455a36612ef7a1feb40880863e74c77b0b7f5)

app/scripts/controllers/network/createJsonRpcClient.js
mcmire: Removed here: [e51455a](https://github.com/MetaMask/metamask-extension/pull/16693/commits/e51455a36612ef7a1feb40880863e74c77b0b7f5)

app/scripts/controllers/network/network-controller.js
```diff
@@ -447,13 +452,30 @@ export default class NetworkController extends EventEmitter {
   }
 
   _setNetworkClient({ networkMiddleware, blockTracker }) {
+    const networkProvider = providerFromMiddleware(networkMiddleware);
```
mcmire, line 455: @Gudahtt Is this what you were looking for? After this I am not sure if I would be able to tell the difference between the "MetaMask middleware" and the "network middleware", but I guess this feels better...

Gudahtt: Yeah this is exactly what I had in mind!

The "metamask middleware" is being removed in this PR: https://github.com/MetaMask/metamask-extension/pull/16863 . It won't be here long.

Gudahtt (approved):
LGTM!

brad-decker (approved):

jpuri (approved):

PR #16707, opened 2022-11-28
fix issue preventing switching to localhost

brad-decker:
## Explanation
@Montoya reported in extension slack channel that when attempting to use switchEthereumChain with the chainId of ganache running locally that the network would not switch and no error was surfaced to the user. This is due to the way we were forming the requestData, providing a 'network type' of 'localhost' which would trigger us to use 'setProviderType' instead of 'updateRpcTarget'. setProviderType only works with a select few strings that match to our internal infura providers. to avoid this issue I simply added a check for the 'localhost' type which allows the logic to fall through to the 'updateRpcTarget' function call correctly.

## Manual Testing Steps
1. Run ganache locally
2. Run the extension locally.
3. Open the browser, attempt to use the switchEthereumChain api with chainid 0x539
4. See that you get the prompt and when approved the extension switches to the appropriate network.

#### Sample code to paste in browser for testing
```js
    try {
      await window.ethereum.request({
        method: 'wallet_switchEthereumChain',
        params: [{ chainId: '0x539' }],
      });
    } catch (e) {
      console.error(e);
      dispatch({ type: MetamaskActions.SetError, payload: e });
    }
```

## Pre-merge author checklist

- [x] I've clearly explained:


legobeat (approved):
The fact that these cases need to be handled differently and that the RPC provider type should make a difference on this abstraction layer is something that'd be amazing to get away from - at some point down the line. This seems to solve the immediate issue.

NiranjanaBinoy (approved):
LGTM

PR #16710, opened 2022-11-28
remove rpc urls from metrics

### LINKED ISSUE #15170: [Bug]: Full RPC endpoint in MetaMetrics event
### Describe the bug

If you add a JSON-RPC endpoint (new Network) that contains a secret like a password or access key, the full URL will be sent to segment.io.

Note: This is also a privacy issue as segment.io shouldn't be receiving details about what RPC endpoints the user is using.

----

Someone looking at a segment.io dashboard or raw data (which is an unknown set of people and potentially hackers) can gain access to paid and/or secret RPC endpoints and then use them to do nefarious things, which could result in the user being a plausible patsy for whatever it is the attacker did.

This is also a privacy issue, as the RPC a user is communicating with should not be exposed publicly unless the user agrees to it.

### Steps to reproduce

1. Add a new Network to MetaMask.
1. For the New RPC URL field set it to `https://example.com/secret-path?secretKey=superSecret`
1. Monitor the network traffic leaving MetaMask destined for segment.io.
1. Notice that the full URL is sent to segment.io.

_No response_

### Version

10.16.2

### Build type

_No response_

### Browser

Brave

### Operating system

Windows

### Hardware wallet

_No response_

### Additional context

_No response_

brad-decker:
## Explanation
This pull request aims to remove any place in which we send the current network's rpc url to segment. We have done this since before we used segment, and when we implemented segment it was carried over.

Fixes #15170 

### Before

image

### After

image

## Manual Testing Steps
1. Add a segment write key to your development config, or setup metametrics for local development
2. start the extension
3. Add extension to chrome
4. inspect the background process
5. Filter the network traffic to "segment"
6. open the extension popup
7. Look for 'batch' requests and examine the payloads
8. see that the rpc url is not listed in the 'network' key on any events (previously it was sent with *every* request)
9. See that the rpc url is not listed in the rpc_url or network key for any 'add network' events sent when adding a network via the addEthereumChain / switchEthereumChain apis OR the custom network form OR populate network lists inside the extension. 

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
+++ b/app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
@@ -299,7 +299,6 @@ async function addEthereumChainHandler(
         chain_id: _chainId,
         network_name: _chainName,
         symbol: ticker,
-        block_explorer_url: firstValidBlockExplorerUrl,
         source: EVENT.SOURCE.TRANSACTION.DAPP,
       },
     });
--- a/app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
+++ b/app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
@@ -289,13 +289,6 @@ async function addEthereumChainHandler(
       }),
     );
 
-    let rpcUrlOrigin;
-    try {
-      rpcUrlOrigin = new URL(firstValidRPCUrl).origin;
-    } catch {
-      // ignore
-    }
-
     sendMetrics({
       event: 'Custom Network Added',
       category: EVENT.CATEGORIES.NETWORK,
@@ -305,17 +298,9 @@ async function addEthereumChainHandler(
       properties: {
         chain_id: _chainId,
         network_name: _chainName,
-        // Including network to override the default network
-        // property included in all events. For RPC type networks
-        // the MetaMetrics controller uses the rpcUrl for the network
-        // property.
-        network: rpcUrlOrigin,
         symbol: ticker,
         block_explorer_url: firstValidBlockExplorerUrl,
         source: EVENT.SOURCE.TRANSACTION.DAPP,
-      },
-      sensitiveProperties: {
-        rpc_url: rpcUrlOrigin,
       },
     });
 
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -2127,7 +2127,6 @@ export default class MetamaskController extends EventEmitter {
         chain_id: chainId,
         network_name: chainName,
         symbol: ticker,
-        block_explorer_url: blockExplorerUrl,
         source: EVENT.SOURCE.NETWORK.POPULAR_NETWORK_LIST,
       },
       actionId,
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -2117,28 +2117,18 @@ export default class MetamaskController extends EventEmitter {
       },
     );
 
-    let rpcUrlOrigin;
-    try {
-      rpcUrlOrigin = new URL(rpcUrl).origin;
-    } catch {
-      // ignore
-    }
     this.metaMetricsController.trackEvent({
       event: 'Custom Network Added',
       category: EVENT.CATEGORIES.NETWORK,
       referrer: {
-        url: rpcUrlOrigin,
+        url: ORIGIN_METAMASK,
       },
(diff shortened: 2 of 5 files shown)


danjm (dismissed):
These changes look good to me. I looked through the codebase for other instances where the rpc url was being sent and could not find any.

app/scripts/metamask-controller.js
```diff
@@ -2117,29 +2117,19 @@ export default class MetamaskController extends EventEmitter {
...
       event: 'Custom Network Added',
       category: EVENT.CATEGORIES.NETWORK,
       referrer: {
-        url: rpcUrlOrigin,
+        url: ORIGIN_METAMASK,
       },
       properties: {
         chain_id: chainId,
         network_name: chainName,
-        network: rpcUrlOrigin,
```
legobeat, line 2135: Should we remove `network_name` as well? Seems redundant with `chain_id` while increasing granularity for custom RPCs?

legobeat, line 2135: Could `network_name` and/or `block_explorer_url` be considered for removal as well? Don't seem super when `chain_id` is there and increase cardinality (anyone with custom unique enough RPC name and/or self-hosted explorer will be singled out)

bschorchit: Just to give an opposite opinion for consideration: 
- By knowing `network_name` we can notice that [200 people have added BSC Testnet as Ethereum Classic](https://mixpanel.com/s/3djnfJ). Is it just by mistake or some rising scam going on? Knowing that could help us investigate and act upon it. We don't have anyone actively looking at it at the moment though so it's ok if we prefer removing it. 
- Agree with removing `block_explorer_url` . It could be interesting to replace it though by a boolean that just tells us wether or not a block explorer has been included (as it's not mandatory to include one). Some parts of the product assume the user has a valid block explorer added and they will not work if the user hasn't (this is specially true for mobile where storage constraints are significant and a lot of the user ability to review their activity relies on us redirecting them to the block explorer). Knowing the % that has a block explorer added versus not could help us make product decisions to rely less or more on block explorers.

legobeat: Re the `network_name` point: This could similarly be re-added in a safe way (say, by emitting only predefined networks with enough usage, and have a common value for all `other`), which could cover for scenarios like the mentioned one.

brad-decker: Done.

jpuri (approved):
Changes look good to me

app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
```diff
@@ -305,18 +298,9 @@ async function addEthereumChainHandler(
       properties: {
         chain_id: _chainId,
         network_name: _chainName,
-        // Including network to override the default network
-        // property included in all events. For RPC type networks
-        // the MetaMetrics controller uses the rpcUrl for the network
-        // property.
-        network: rpcUrlOrigin,
         symbol: ticker,
-        block_explorer_url: firstValidBlockExplorerUrl,
```
digiwand, line 314: based on https://github.com/MetaMask/metamask-extension/pull/16710/files#r1034677654 did we want to send this value as a boolean value? maybe `has_block_explorer_url` or `block_explorer_url_exists`? (not exactly sure what our naming conventions are for booleans here)

app/scripts/metamask-controller.js
```diff
@@ -2117,29 +2117,18 @@ export default class MetamaskController extends EventEmitter {
...
         symbol: ticker,
-        block_explorer_url: blockExplorerUrl,
```
digiwand, line 2137: same comment here from above https://github.com/MetaMask/metamask-extension/pull/16710/files#r1036134883

> based on https://github.com/MetaMask/metamask-extension/pull/16710/files#r1034677654 did we want to send this value as a boolean value? maybe `has_block_explorer_url` or `block_explorer_url_exists`? (not exactly sure what our naming conventions are for booleans here)

app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
brad-decker: I don't think knowing this serves us and i'd rather not further complicate this PR which is aimed at removing the offending values. Lets have further conversation about alternative tracking outside this PR and do follow up to figure out what is best.

digiwand (approved):

PR #16721, opened 2022-11-29
Added storybook check (WIP 👷‍♀️ )

### LINKED ISSUE #15760: Add storybook check to CI
### Description
Currently there is no way for an engineer to know if a change that they make to the codebase or component breaks storybook. We need a way for CI to check if a change has broken storybook. Look into [StoryShots](https://github.com/storybookjs/storybook/tree/master/addons/storyshots/storyshots-core) as a solution.

### * Edit
Storyshots has been softly deprecated in favour of integration testing. Let's look into that instead https://storybook.js.org/docs/react/writing-tests/test-runner

Screen Shot 2022-11-14 at 10 56 24 AM
Screen Shot 2022-11-14 at 10 56 49 AM

### References
[Example of this happening Slack thread]()
[Same issue asked in storybook](https://github.com/storybookjs/storybook/issues/7643)
[Storybook StoryShots addon](https://consensys.slack.com/archives/GTQAGKY5V/p1662493804279299)

NidhiKJha:
Fixes #15760 

* Added `test-storybook` dependency

TODOS:
1. Fix the story check issue `Unexpected component id`
2. Add in the CI


yarn.lock
```diff
@@ -3815,18 +4423,58 @@
...
+    "@storybook/core-events" "6.5.13"
+    "@storybook/csf" "0.0.2--canary.4566f4d.1"
+    "@storybook/router" "6.5.13"
+    "@storybook/theming" "6.5.13"
+    "@types/webpack-env" "^1.16.0"
+    core-js "^3.8.2"
+    global "^4.4.0"
+    regenerator-runtime "^0.13.7"
+
+"@storybook/api@6.5.10", "@storybook/api@^6.5.10":
```
Gudahtt, line 4443: It looks like some of the storybook dependencies here are duplicated. Maybe there is an incompatibility here between something `@storybook/test-runner` is using and something our version of storybook is using.

Gudahtt: Oh nevermind. Still worth deduplicating this, but I see what you mean about `__filename` not being supported.

ui/components/app/account-list-item/account-list-item.stories.js
```diff
@@ -4,7 +4,7 @@ import AccountListItem from './account-list-item';
 
 export default {
   title: 'Components/App/AccountListItem',
-  id: __filename,
+  id: 'ui-components-app-account-list-item-account-list-item',
```
georgewrmarshall, line 7: Could we just get rid of the `id` field altogether? From my understanding it does 2 things 
1. Creates the file path URL schema
2. Generates this section of links for each PR build 👇 

Screenshot 2022-12-21 at 4 01 47 PM

🤔  I'm not sure it adds a huge amount of value because it seems to include stories that have no relevance to the PR https://github.com/MetaMask/metamask-extension/pull/16931/files.

Plus there is the other link to the storybook build so we could just use that for visual testing?

Screenshot 2022-12-21 at 4 02 19 PM

**Pros**
- Engineers don't have to fill out another field when creating a story

**Cons**
- Lose the ability to link to effected stories from the PR
- We would have to update all stories that link to other stories and think about externally links stories from Figma

So would have to update the links in Figma and in our MDX docs

Screenshot 2022-12-21 at 3 55 46 PM

Screenshot 2022-12-21 at 3 56 42 PM

Thoughts on this @Gudahtt @kumavis?

georgewrmarshall (commented):
Left a comment regarding what we talked about in our extension sync

ui/components/app/account-list-item/account-list-item.stories.js
NidhiKJha: Yup, we can get rid of the id field altogether. @georgewrmarshall this is the current plan

1. We will delete the `id` field from all the stories.js files and won't be relying on `filename/id` anymore. 
2. Currently URLs are generated based on the existing ids but when the` ids `will be gone, new URLs will be generated based on the title of each story.
3. In the CI, we can get the URL from stories.json. 
4. The only con you mentioned will be to update the New URLs in the docs (README.mdx) files.

NidhiKJha:
Closing it with reference to #17092

PR #16722, opened 2022-11-29
[snow] [patch] restore open api usage fully

weizman:
Snow intentionally disables `open` API due to it being highly vulnerable against Snow (see https://github.com/LavaMoat/Snow/issues/2).

MetaMask apparently relies on `open` API.

This was not discovered when Snow was introduced due to **MetaMask tests not covering parts of the app that make use of `open` API**. 

This PR introduces a temporary patch for bringing `open` API back by allowing it to be used by Snow.

This patch will be removed once an elegant solution to this problem is introduced to Snow.

NOTE: [Line 19](https://github.com/MetaMask/metamask-extension/pull/16722/files#diff-cddcd04fc7a35f3946bff2b5418460271e9abee5be2127544dc3000dc173614cR19) of the patch file only looks intimidating because patch-package applies diffs per line in file, and since the production build file of Snow is compressed to a single line, patch-package basically replaces the whole line where in reality the diff is a single character (`:1:3817` is changed from `!0` to `!1` which resolves to `false` instead of `true`)


Gudahtt (approved):
LGTM!

segun (approved):

Issue #16702, opened 2022-11-28
[Bug]: MV3 DAPP action replay - it is possible for a DAPP RPC request to be processed twice in extension


jpuri:
### Describe the bug

With the new DAPP action replay implementation it is possible for DAPP to send an RPC request twice to extension. For a request like create new transaction this can result in > 1 transactions being created in extension.

We need to fix this, possible fix can be sending `id` of request with the request [here](https://github.com/MetaMask/json-rpc-middleware-stream/blob/main/src/createStreamMiddleware.ts#L54), and on extension side we can keep a cache of id already processed to ensure that a request is not processed twice.

PR #16730, opened 2022-11-29
Adding middleware to filter out duplicate requests from DAPP

jpuri:
Fixes: #https://github.com/MetaMask/metamask-extension/issues/16702

Filtering out JSON RPC requests already seen from DAPP.


app/scripts/lib/createDupReqFilterMiddleware.test.js
```diff
@@ -0,0 +1,25 @@
+import createDupReqFilterMiddleware from './createDupReqFilterMiddleware';
+
+describe('createDupReqFilterMiddleware', () => {
+  it('call function next if request is seen first time', () => {
+    const filterFn = createDupReqFilterMiddleware();
+    const request = { id: 1 };
+    const nextMock = jest.fn();
+    const endMock = jest.fn();
+    filterFn(request, undefined, nextMock, endMock);
```
digiwand, line 9:
```suggestion

    filterFn(request, undefined, nextMock, endMock);

```
echoing what @Gudahtt mentioned in a [comment in another PR](https://github.com/MetaMask/KeyringController/pull/169/files#r1035969577)

> Nit: typically for tests, we use newlines to separate the different phases of the test: arrange, act, and assert. Arrange is the setup, act is where we do the thing that is under test, and assert is where we check that it works.

app/scripts/lib/createDupReqFilterMiddleware.js
```diff
@@ -0,0 +1,23 @@
+import log from 'loglevel';
+
+/**
+ * Returns a middleware that filters out requests already seen
+ *
+ * @returns {Function}
+ */
+export default function createDupReqFilterMiddleware() {
```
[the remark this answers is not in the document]
jpuri: yep done

app/scripts/lib/createDupeReqFilterMiddleware.test.js
```diff
@@ -0,0 +1,27 @@
...
+  });
+  it('call function end if request is seen second time', () => {
+    const filterFn = createDupeReqFilterMiddleware();
+    const request = { id: 1 };
+    const nextMock = jest.fn();
+    const endMock = jest.fn();
+    filterFn(request, undefined, nextMock, endMock);
+    expect(nextMock).toBeCalledTimes(1);
+    expect(endMock).not.toBeCalled();
+    filterFn(request, undefined, nextMock, endMock);
```
digiwand, line 23:
```suggestion

  it('call function end if request is seen second time', () => {
    const filterFn = createDupeReqFilterMiddleware();
    const request = { id: 1 };
    const nextMock = jest.fn();
    const endMock = jest.fn();

    filterFn(request, undefined, nextMock, endMock);
    expect(nextMock).toBeCalledTimes(1);
    expect(endMock).not.toBeCalled();

    filterFn(request, undefined, nextMock, endMock);
```

digiwand (dismissed):

app/scripts/lib/createDupeReqFilterMiddleware.js
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Returns a middleware that filters out requests already seen
+ *
+ * @returns {Function}
+ */
+export default function createDupeReqFilterMiddleware() {
+  const processedRequestId = [];
+  return function filterDuplicateRequestMiddleware(
+    /** @type {any} */ req,
+    /** @type {any} */ _,
```
Gudahtt, line 12: Nit: it might be useful to spell out what this is, so it's more clear why we're ignoring it. The leading `_` still shows we're not using it on purpose.
```suggestion
    /** @type {any} */ _res,
```

app/scripts/lib/createDupeReqFilterMiddleware.js
```diff
@@ -0,0 +1,23 @@
...
+    /** @type {Function} */ next,
+    /** @type {Function} */ end,
+  ) {
+    if (processedRequestId.indexOf(req.id) >= 0) {
+      log.info(`RPC request with id ${req.id} already seen.`);
+      return end();
```
Gudahtt, line 18: Are we sure the dapp has received the response already from the first request? i.e. is it possible for the first response to have been dropped, e.g. while the contentscript was waiting for the initialization message.

Gudahtt: I suspect it is possible, though maybe not likely. It could happen after the contentscript has destroyed the streams, before they had been recreated.

It might be safer to cache the response, and re-send the original response if we get the message again. Not great having to cache every dapp response though :thinking: Even for just 5 minutes, that could get hefty.

Gudahtt: Alternatively, we could add a retry loop on the service worker side to ensure the request gets through if this unfortunate timing happens. Then this middleware works fine as-is.

Gudahtt (dismissed):
LGTM! I identified one potential problem in a comment that we should investigate, but that doesn't need to block.

app/scripts/lib/createDupeReqFilterMiddleware.js
jpuri: The filters are called after request reaches service worker, the only scenario I think when user may not get response is when service worker dies before request is completely processed, in that case replay logic in `inpage` will resend. the request.

Gudahtt (approved):

app/scripts/lib/createDupeReqFilterMiddleware.js
Gudahtt: Here is the failure scenario I was thinking of: https://sequencediagram.org/index.html#initialData=C4S2BsFMAIDEEMTgK4CcYGcDGkB29UQB7AKBIAcDQsRLdhoBhI+vYbQ84CqkGuhgGVIqAG58YAdyKoA1iOgAKAIwZgASh6pqtePWjCxE6NLkLFAJlwATTSXhZQo+MBjNW9DrW4OnLmIbiOCYy8qhKqhpk7q6eWJzAALQAfIHGpmERauoAXAAGkMAAFgD6WEWIuACS1nnQ6ACOyJBqZNaQviDOrgYiQVKh5pF2ZPAARkQAHvUgAOZFDEQAZr1GwRlD2TkACqhEOBgY0Hjt1qMT04Tziytp64Phlja5u-stR2pUkGed3QF96QeSistmiLFi7Hi3hSdwGZkeINyBWKJQwJwAKqg9BhOiw6o1mq0YmwvFwADyJWEheFZDT5QqlcqVGp1NToeAAW2gS0QKHQZHsF2gUCWNyY4JJUK4OQAIi1gHsAJ7HSaxDDEXDQNkdDkYEjEuIJClUjYI570lFomyY7G43B1SL1FrkFho85TYWQUXQZbijyQhI5ABKkHiHR6kFVeHVLC1Cp1ev1EsN0NSAPuNKetgtpSt1htuBxjg1+MgTXlYP9pOAxvTcMyWaRDNRGKxhbtdUgHK4ypBTowLsLkCAA

I'm not sure whether it's possible. The main part I'm uncertain of is whether messages sent over the "contentscript <=> service worker" stream for the _old_ process would be delivered to the _new_ process. Maybe the stream needs to be re-established.

jpuri: Not sure if I am getting you correct, but in case it is not delivered to backend - there will be no response and DAPP replay logic will replay it.

Gudahtt: Got it, thanks. That sounds right. I just wasn't sure.

digiwand (approved):
LGTM 👍

Issue #16709, opened 2022-11-28
Add ability to delete metametrics data 


danjm:
In mobile users can delete there metametrics data by clicking a button. We should add this ability within the extension

<img width="250" src="https://user-images.githubusercontent.com/7499938/204350764-3349f4f5-ef65-449f-9cac-28637bb5330e.png" />

Our metrics implementation is not identical. Research is required to determine exactly how we will do this on the extension side.

coreyjanssen:
@brad-decker @holantonela what is materially different about this within extension vs desktop from the user's perspective? asking bc if there isn't much beyond things that happen on the backend, the mobile copy looks like it would work in this scenario too!

brad-decker:
The difference is that users may have multiple installs of MetaMask (brave, chrome, firefox, etc) they also may uninstall and reinstall the extension and their userId would change. we should clarify the copy to make sure that it will only delete the data related to *this install* and will not delete data from any other installs OR metamask mobile

brad-decker:
@danjm @hilvmason @kevinghim we have an issue here where the segment API requires an API key that would need to be included in our builds and could theoretically be used to either brute force delete OR other segment api end points (such as manipulating user list for accessing our workspace). I'm working with @worldlyjohn to get an idea of what the actual dangers are here but we may want to explore alternative options for implementation. https://consensys.slack.com/archives/GTQAGKY5V/p1669912934330049

coreyjanssen:
Gotcha @brad-decker! Thoughts on this rewrite @kevinghim?

MetaMetrics data is anonymized usage and diagnostic data. This will delete historical MetaMetrics data associated with this install of your wallet, but **won’t** delete data from any other installs or the MetaMask mobile app. Your wallet and accounts will remain exactly as they are now after this data has been deleted. This process may take up to 60 days. View the Consensys Privacy Policy.

kevinghim:
@coreyjanssen should this "but won’t delete data from any other installs or the MetaMask mobile app." be "but won’t delete data from any other installs or from the MetaMask mobile app." ?

coreyjanssen:
Either is correct, just a matter of preference! Let's add it for the most specificity. :) 

Updated:

[...]

PR #16744, opened 2022-11-30
Migrate 'participateInMetaMetrics' to 'metaMetricsParticipationMode'

brad-decker:
## Explanation
For #16709 which will allow users to request for the removal of their data from our analytics provider, we will have to do this data deletion request by 'userId' which in this case is our randomly generated hexadecimal 'metaMetricsId'. We previously would delete this id and regenerate it everytime the 'participate in metametrics' toggle was interacted with in the extension settings. This would to situations where the user might go in, disable the toggle and then request for their data to be deleted and it would fail (either silently or with an error) because we no longer have any reference to their MetaMetricsId. 

I decided to clean up some tech debt while I was making this change so that future engineers might understand a bit more about how our toggle works. Previously we had `participateInMetaMetrics` which is a boolean state key. HOWEVER we baked in functionality that made it behave in 3 different ways. this key defaults to 'null' which was used through the application as indication that the user had not yet interacted with the MetaMetrics opt in screen in the onboarding flow. We also used this nullish value as a logic gate in other parts of the code base and not only is it difficult to understand what was going on, but it makes it increasingly easy for engineers to accidentally change the default value and break our onboarding flow.

This pull request does the following:
1. Adds a new 'metaMetricsParticipationMode' to the MetaMetricsController state. This new field has 3 possible values: NOT_CHOSEN, DO_NOT_PARTICIPATE and PARTICIPATE. 
2. Adds a new "METAMETRICS_PARTICIPATION" constant for usage with the above state key.
3. Migrates user data from the old participateInMetaMetrics boolean state key to the new one.
4. Updates all usages
5. Removes the "clearing" of metametrics id when the mode is toggled.

## Screenshots/Screencaps/How to test
The only functional change in this PR is the following:

**On Develop**
1. opt in to MetaMetrics.
2. Inspect state and see your  metaMetricsId
3. opt out.
4. Inspect state and see your metaMetricsId is now null
5. Opt in.
6. Inspect state and see your metaMetricsId is regenerated to a new hexadecimal

**This Branch**
1. opt in to MetaMetrics.
2. Inspect state and see your metaMetricsId
3. opt out.
4. Inspect state and see your metaMetricsId is still the same.
5. Opt in.
6. Inspect state and see your metaMetricsId is still the same.
7. 

## Pre-merge author checklist

- [x] I've clearly explained:


app/scripts/migrations/index.js
```diff
@@ -81,6 +81,7 @@ import m074 from './074';
 import m075 from './075';
 import m076 from './076';
 import m077 from './077';
+import m088 from './078';
```
danjm, line 84: Typo: `m088` -> `m078`. Below as well

brad-decker:
Marking as Do Not Merge because at least some of this should be reverted if we cannot safely roll this feature out.

PR #16749, opened 2022-11-30
Chore/add background controller

cryptotavares:
## Explanation
This PR aims to refactor the `background.js` and make it more modular. Currently `background.js` follows a module pattern, and although modules are great for a number of things (reusability, encapsulation, etc etc), they lack extensibility.

### Why is extensibility important on background.js?
If we assume that background.js responsibility is to initialise the extension background script, then extensibility should not be an issue. However, currently it is much more than that. It also includes all business logic responsible for:
* initialising the MetaMask controller
* triggering the extension UI
* monitoring if the popup is open, if a notification is open, etc
* persisting state
* updating the UI badge
* triggering the popup
* etc..

In order to provide better interoperability between different MetaMask platforms, we should aim to reuse the same business logic whenever possible (some controllers are already being used on other platforms). The best way to re-use the Extension business logic, without being hooked to its initialisation is by refactoring `background.js` and make it extensible.

**Proposed solution:**
Create a background controller to replace the `setupController` from `background.js` and all its related methods. This should be extensible and allow any other platform to pick it up, extend and re-use as they see fit. By making it a Class anyone can extend it and cherry pick which logic they want to override/add, and use on the `init` method.

**Use case:**
The current use case is the desktop app... but any other app that might want to re-use the Extension business logic is a fair candidate.

Besides all of the above, IMO, the proposed refactor simplifies the code and makes it easier to read.

## Pre-merge author checklist

- [ ] I've clearly explained:


cryptotavares:

app/scripts/background-controller.js
```diff
@@ -0,0 +1,556 @@
...
+      label = String(count);
+    }
+    // browserAction has been replaced by action in MV3
+    if (isManifestV3) {
+      browser.action.setBadgeText({ text: label });
+      browser.action.setBadgeBackgroundColor({ color: '#037DD6' });
+    } else {
+      browser.browserAction.setBadgeText({ text: label });
+      browser.browserAction.setBadgeBackgroundColor({ color: '#037DD6' });
+    }
```
legobeat, line 413: non-functional refactor:
```suggestion
    const count = this.getUnapprovedTransactionCount();
    const label = String(count || '');

    // browserAction has been replaced by action in MV3
    const action = isManifestV3 
      ? browser.action
      : browser.browserAction;

    action.setBadgeText({ text: label });
    action.setBadgeBackgroundColor({ color: '#037DD6' });
```

app/scripts/background-controller.js
```diff
@@ -0,0 +1,556 @@
...
+        const tabId = remotePort.sender.tab.id;
+        const url = new URL(remotePort.sender.url);
+        const { origin } = url;
+
+        remotePort.onMessage.addListener((msg) => {
+          this.remotePortOnMessageRequestAccountsCb(msg, origin, tabId);
+        });
+      }
+      this.connectExternal(remotePort);
+    }
```
legobeat, line 354: I think reduced nesting makes this more clear?

```suggestion
      switch (processName) {
        case ENVIRONMENT_TYPE_POPUP:
          this.popupIsOpen = true;
          endOfStream(portStream, this.endStreamPopUpCb);
          return;
        case ENVIRONMENT_TYPE_NOTIFICATION:
          this.notificationIsOpen = true;
          endOfStream(portStream, this.endStreamNotificationCb);
          return;
        case ENVIRONMENT_TYPE_FULLSCREEN:
          const tabId = remotePort.sender.tab.id;
          this.openMetamaskTabsIDs[tabId] = true;
          endOfStream(portStream, () => {
            this.endStreamFullScreenCb(tabId);
          });
          return;
        default:
          return;
      }
    }
    if (
      senderUrl?.origin === this.phishingPageUrl.origin &&
      senderUrl.pathname === this.phishingPageUrl.pathname
    ) {
      const portStream = new PortStream(remotePort);
      this.controller.setupPhishingCommunication({
        connectionStream: portStream,
      });
      return;
    }
    if (remotePort?.sender.tab && remotePort.sender.url) {
      const tabId = remotePort.sender.tab.id;
      const url = new URL(remotePort.sender.url);
      const { origin } = url;
      remotePort.onMessage.addListener((msg) => {
        this.remotePortOnMessageRequestAccountsCb(msg, origin, tabId);
      });
    }
    this.connectExternal(remotePort);
```

app/scripts/background-controller.js
```diff
@@ -0,0 +1,556 @@
...
+      (isVivaldi || !this.popupIsOpen) &&
+      !currentlyActiveMetamaskTab
+    ) {
+      this.uiIsTriggering = true;
+      try {
+        await this.notificationManager.showPopup();
+      } finally {
+        this.uiIsTriggering = false;
+      }
+    }
```
legobeat, line 538:
```suggestion
    if (this.uiIsTriggering || currentlyActiveMetamaskTab || this.popupIsOpen) {
      return;
    }
    // Vivaldi is not closing port connection on popup close, so popupIsOpen does not work correctly
    // To be reviewed in the future if this behaviour is fixed - also the way we determine isVivaldi variable might change at some point
    const isVivaldi =
      tabs.length > 0 &&
      tabs[0].extData?.includes(vivaldi_tab');
    if (isVivaldi) {
      this.uiIsTriggering = true;
      try {
        await this.notificationManager.showPopup();
      } finally {
        this.uiIsTriggering = false;
      }
    }
```

legobeat (commented):
some style / modernization suggestions since that is mentioned as a goal in the description.

cryptotavares:
> some style / modernization suggestions since that is mentioned as a goal in the description.

@legobeat we can definitely modernise the underlying code. Although this PR aimed for a more structural refactor without changing the underling logic in order to make reviewing it much easier.

My suggestion would be to:
* keep this PR as is -> goal is to refactor the background.js and the within `setupController` method. Transform the `setupController` method into `backgroundController` class, making it more extendable. We should not update the underling code, so that the PR is easy to review (as most methods were just moved around with minimum to no code changes)
* subsequent PR to actual bring the code within the background controller to a more up to date standard.

what do you think about approaching it this way?

legobeat:
@cryptotavares sounds reasonable - keeps changes easier to follow.

PR #16757, opened 2022-12-01
Fix production builds for Flask

FrederikBolding:
## Explanation

Mirror beta CI configuration to fix building of Flask production builds. Previously we were not building Flask using the production configuration, resulting in non-production builds being released.


Gudahtt (approved):
LGTM!

Mrtenz (approved):

ritave (approved):

PR #16762, opened 2022-12-01
Updating keyring controller

### LINKED ISSUE #16674: [Bug]: login fails when moving from mv2 to mv3

jpuri:
Fixes: https://github.com/MetaMask/metamask-extension/issues/16674

New version of KeyringController has fix to not throw is a keyring for an unsupported keyring type is seen.


Gudahtt (approved):
LGTM!

digiwand (approved):

PR #16765, opened 2022-12-01
Show user the general contract interaction screen for token approvals…

danjm:
…, when the asset standard is undefined

This PR fixes the following problem:

If a user is on a dapp, and the dapp proposes an ERC20 or collectibles approval transaction, then we should show them the appropriate custom approval screen for that case. We decide which screen to show based on the `standard` property returned by the `getAssetDetails` token util https://github.com/MetaMask/metamask-extension/blob/develop/ui/helpers/utils/token-util.js#L258-L291

If the `getTokenStandardAndDetails` call in `getAssetDetails` fails for any reason, then the `standard` property is undefined. In that case, the `confirm-approve.js` component will render the collectibles approval screen, even if this is actually an erc20 token approval transaction.

The fix for this is to simply show the generic contract interaction screen in cases where the standard is unknown. Then the user will not see misleading information.


digiwand (approved):

jpuri (approved):

danjm:
screen capture of this working correctly

https://user-images.githubusercontent.com/7499938/208480899-8876e314-872e-414e-ac5a-f6275b367d1e.mp4

PR #16773, opened 2022-12-02
Network request in background should not start until onboarding is completed

### LINKED ISSUE #16705: Onboarding v2: Don't send network requests until onboarding is complete
Onboarding v2 introduces privacy settings options to the onboarding flow. These happen towards the end of the flow (see video below)

https://user-images.githubusercontent.com/7499938/204328922-163e7f20-2cc2-422d-afec-087de0918ef2.mp4

We want to further expand on that to allow the user to set a custom rpc during the flow https://github.com/MetaMask/metamask-extension/issues/16696

We should not send any network requests to any providers or any of the services that users opt out of, until the users have gotten passed that final screen and completed onboarding. (Of course, we should then only send network requests according to the users preferences).

The only requests we might send during onboarding are usage metrics or sentry errors, but only if the user opted in to metametrics at the beginning of the onboarding flow.

jpuri:
Fixes: #16705

Network requests (except segment) should only be started after onboarding is completed.

### DIFF
--- a/app/scripts/controllers/incoming-transactions.js
+++ b/app/scripts/controllers/incoming-transactions.js
@@ -167,8 +167,7 @@ export default class IncomingTransactionsController {
    * @param {number} [newBlockNumberDec] - block number to begin fetching from
    */
   async _update(address, newBlockNumberDec) {
-    const completedOnboarding =
-      this.onboardingController.store.getState().completedOnboarding;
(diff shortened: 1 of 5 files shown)


danjm (approved):
This looks good to me.

digiwand (approved):

PR #16780, opened 2022-12-02
Sanitize privacy sensitive data before sending to sentry.

danjm:
This PR will ensure that we don't send any privacy sensitive information along with our sentry errors.

~~Still need to add some unit tests.~~


pedronfigueiredo (dismissed):

app/scripts/lib/setupSentry.js
```diff
@@ -146,31 +154,88 @@ export default function setupSentry({ release, getState }) {
...
+  if (breadcrumb.data?.to) {
+    breadcrumb.data.to = hideUrlIfNotInternal(breadcrumb.data.to);
+  }
+  if (breadcrumb.data?.from) {
+    breadcrumb.data.from = hideUrlIfNotInternal(breadcrumb.data.from);
+  }
+  return breadcrumb;
+}
+
+export function rewriteReport(report, getState) {
```
Gudahtt, line 187: Nit: Ideally we'd have JSDoc comments for each of these functions, especially now that you've exported them.

app/scripts/lib/setupSentry.js
```diff
@@ -146,31 +154,88 @@ export default function setupSentry({ release, getState }) {
...
   }
+  return report;
+}
 
-  return Sentry;
+function sanitizeUrlsFromErrorMessages(report) {
+  rewriteErrorMessages(report, (errorMessage) => {
+    let newErrorMessage = errorMessage;
+    const re =
+      /https?:\/\/(www\.)?[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,63}\b([-a-zA-Z0-9()@:%_+.~#?&//=]*)/gu;
```
Gudahtt, line 219: This is a complicated regex. Was this copied from somewhere? Could we add comments to explain what it's doing?

Gudahtt: This seems far more complex than what we need here. We just need to recognize URL-like things, we don't need to ensure only pefectly valid URLs are scrubbed.

app/scripts/lib/setupSentry.js
```diff
@@ -146,31 +154,88 @@ export default function setupSentry({ release, getState }) {
...
+    const urlsInMessage = newErrorMessage.match(re) || [];
+    urlsInMessage.forEach((url) => {
+      if (
+        !Object.values(ERROR_URL_ALLOWLIST).some((allowedUrl) =>
+          url.match(allowedUrl),
```
Gudahtt, line 224: It looks like `allowedUrl` is a string here, but `match` expects a regex. The docs for `match` suggest that when given a string, it implicitly creates a regex using the RegEx constructor. So this will give us a regex like this for codefi for example: `/codefi.network/`.

This isn't ideal in a number of ways:
* The unicode flag isn't used, so it might improperly handle unicode domains
* It will incorrectly allow URLs that have the allowed domains somewhere in the URL path
* This is generally a confusing way to use `match`, and we should instead explicitly create the regex when using match.

app/scripts/lib/setupSentry.js
```diff
@@ -146,31 +154,88 @@ export default function setupSentry({ release, getState }) {
...
+        newErrorMessage = newErrorMessage.replace(url, '**');
+      }
+    });
+    return newErrorMessage;
+  });
+}
+
+function sanitizeAddressesFromErrorMessages(report) {
+  rewriteErrorMessages(report, (errorMessage) => {
+    const newErrorMessage = errorMessage.replace(/0x[A-Fa-f0-9]{40}/u, '**');
```
Gudahtt, line 236: Nit: Should we keep the leading `0x`, so that we know it was an address?

app/scripts/lib/setupSentry.js
```diff
@@ -193,9 +258,11 @@ function simplifyErrorMessages(report) {
 function rewriteErrorMessages(report, rewriteFn) {
   // rewrite top level message
   if (typeof report.message === 'string') {
+    /** @todo parse and remove/replace URL(s) found in report.message  */
     report.message = rewriteFn(report.message);
   }
   // rewrite each exception message
+  console.log('report', report);
```
Gudahtt, line 265: Leftover console log

app/scripts/lib/setupSentry.js
```diff
@@ -193,9 +258,11 @@ function simplifyErrorMessages(report) {
 function rewriteErrorMessages(report, rewriteFn) {
   // rewrite top level message
   if (typeof report.message === 'string') {
+    /** @todo parse and remove/replace URL(s) found in report.message  */
```
Gudahtt, line 261: Leftover TODO

app/scripts/lib/setupSentry.js
```diff
@@ -146,31 +154,88 @@ export default function setupSentry({ release, getState }) {
...
+  if (!_url.match(re)) {
+    return '';
+  }
+  return _url;
+}
+
+export function rewriteBreadcrumb(breadcrumb) {
+  /** @todo check URL allowlist, exclude URLs not in allowlist */
+
+  if (breadcrumb.data?.url) {
```
Gudahtt, line 175: This would crash if `breadcrumb` was `undefined` or set to a primitive value

I don't know if that's a real possibility or not, but I noticed that the pre-existing `beforeBreadcrumb` hook used `?` to protect against this.

app/scripts/lib/setupSentry.js
```diff
@@ -146,31 +154,88 @@ export default function setupSentry({ release, getState }) {
...
-      if (getState) {
-        const appState = getState();
-        if (!report.extra) {
-          report.extra = {};
-        }
-        report.extra.appState = appState;
+  return Sentry;
+}
+
+function hideUrlIfNotInternal(_url) {
```
Gudahtt, line 164: Conventionally, a leading underscore implies that a parameter is unused. This one is being used though.

app/scripts/lib/setupSentry.js
```diff
@@ -146,31 +154,88 @@ export default function setupSentry({ release, getState }) {
...
+        report.extra = {};
       }
-    } catch (err) {
-      console.warn(err);
+      report.extra.appState = appState;
     }
-    return report;
+  } catch (err) {
+    console.error(err);
+    throw err;
```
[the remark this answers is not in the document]
danjm: oops, that's leftover from when I was debugging

app/scripts/lib/setupSentry.js
```diff
@@ -221,7 +288,7 @@ function rewriteReportUrls(report) {
 }
 
 function toMetamaskUrl(origUrl) {
-  const filePath = origUrl.split(globalThis.location.origin)[1];
+  const filePath = origUrl?.split(globalThis.location.origin)[1];
```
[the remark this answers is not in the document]
danjm: The possibility of undefined `origUrl` during unit tests

app/scripts/lib/setupSentry.js
danjm: Added JSDoc comments in latest commit https://github.com/MetaMask/metamask-extension/pull/16780/commits/daaac950566477c5197089c509babada82f9cdaa

app/scripts/lib/setupSentry.js
danjm: I went with something simpler here https://github.com/MetaMask/metamask-extension/pull/16780/commits/daaac950566477c5197089c509babada82f9cdaa#diff-a8c55d1ad0724e946c8c35114d6a298b74904accdddedff77e1b8ade2097fd65R248-R250

app/scripts/lib/setupSentry.js
danjm: changed the approach here, to match against the URL hostname https://github.com/MetaMask/metamask-extension/pull/16780/commits/daaac950566477c5197089c509babada82f9cdaa#diff-a8c55d1ad0724e946c8c35114d6a298b74904accdddedff77e1b8ade2097fd65R251-R256

app/scripts/lib/setupSentry.js
danjm: Good idea, done in latest commit https://github.com/MetaMask/metamask-extension/pull/16780/commits/daaac950566477c5197089c509babada82f9cdaa#diff-a8c55d1ad0724e946c8c35114d6a298b74904accdddedff77e1b8ade2097fd65R273

app/scripts/lib/setupSentry.js
danjm: removed in latest commit

app/scripts/lib/setupSentry.js
danjm: removed in latest commit

app/scripts/lib/setupSentry.js
```diff
@@ -146,31 +154,125 @@ export default function setupSentry({ release, getState }) {
...
-        const appState = getState();
-        if (!report.extra) {
-          report.extra = {};
-        }
-        report.extra.appState = appState;
+  return Sentry;
+}
+
+/**
+ * Receives a string and returns that stringif it is a
```
Gudahtt, line 166: Typo, missing space
```suggestion
 * Receives a string and returns that string if it is a
```

app/scripts/lib/setupSentry.js
danjm: Addressed in latest commit https://github.com/MetaMask/metamask-extension/pull/16780/commits/daaac950566477c5197089c509babada82f9cdaa#diff-a8c55d1ad0724e946c8c35114d6a298b74904accdddedff77e1b8ade2097fd65R189-R197

app/scripts/lib/setupSentry.js
danjm: Removed in latest commit https://github.com/MetaMask/metamask-extension/pull/16780/commits/daaac950566477c5197089c509babada82f9cdaa#diff-a8c55d1ad0724e946c8c35114d6a298b74904accdddedff77e1b8ade2097fd65L164

app/scripts/lib/setupSentry.js
```diff
@@ -146,31 +154,125 @@ export default function setupSentry({ release, getState }) {
...
+ * regex match for a url with a `chrome-extension` or `moz-extension`
+ * protocol, and an empty string otherwise.
+ *
+ * @param {string} url
+ * @returns String
```
Gudahtt, line 171: An improvement to the docs here:
```suggestion
 * @param {string} url - The URL to check.
 * @returns {string} An empty string if the URL was internal, or the unmodified URL otherwise.
```

app/scripts/lib/setupSentry.js
```diff
@@ -146,31 +154,125 @@ export default function setupSentry({ release, getState }) {
...
+  return url;
+}
+
+/**
+ * Receives a Sentry breadcrumb object and potentially removes urls
+ * from its `data` property, it particular those possibly found at
+ * data.from, data.to and data.url
+ *
+ * @param {object} breadcrumb - A Sentry breadcrumb object: https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/
+ * @returns Object - A modified Sentry breadcrumb object.
```
Gudahtt, line 187: It looks like you're using the wrong syntax for the `@returns` part of each of these doc blocks. It should look like this:

```suggestion
 * @returns {object} A modified Sentry breadcrumb object.
```

The type goes in curly brackets. There isn't any dash because the dash is for separating the _name_ from the description, and returns don't have names.

app/scripts/lib/setupSentry.js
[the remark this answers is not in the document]
Gudahtt: Gotcha. I don't think that's possible in practice but I'm not sure, and this seems harmless enough.

app/scripts/lib/setupSentry.js
```diff
@@ -146,31 +154,125 @@ export default function setupSentry({ release, getState }) {
...
+ *
+ * @param {object} report - the report to modify
+ */
+function sanitizeUrlsFromErrorMessages(report) {
+  rewriteErrorMessages(report, (errorMessage) => {
+    let newErrorMessage = errorMessage;
+    const re = /(([A-z]+:\/\/)|(www\.))\S+[@:.]\S+/gu;
+    const urlsInMessage = newErrorMessage.match(re) || [];
+    urlsInMessage.forEach((url) => {
+      const urlObj = new URL(url);
```
Gudahtt, line 252: Hmmm. This will throw if the URL isn't valid. Luckily if this throws, it stops trying to send the whole report so we're still not leaking anything. Though it does lose us the report, which is not ideal.

Perhaps it would be safer to take the hostname from the same regex we're using to detect URLs? We have the two forward slashes as a starting point. The valid characters within a _hostname_ are quite a lot more restricted than for a URL in general, so it should be easy enough to match it.

Something like... `[-a-zA-Z0-9.]+`

Then once we have the hostname, we can look for any that match an allowlist entry or end with `.[entry]` (to capture subdomains)

app/scripts/lib/setupSentry.js
```diff
@@ -146,31 +154,125 @@
...
+/**
+ * Receives a Sentry event object and modifies it so that urls are removed from any of its
+ * error messages.
+ *
+ * @param {object} report - the report to modify
+ */
+function sanitizeUrlsFromErrorMessages(report) {
+  rewriteErrorMessages(report, (errorMessage) => {
+    let newErrorMessage = errorMessage;
+    const re = /(([A-z]+:\/\/)|(www\.))\S+[@:.]\S+/gu;
```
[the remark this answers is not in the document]
Gudahtt: Neat warning! I think you wanted something closer to this:

```suggestion
    const re = /(([a-zA-Z+.]+:\/\/)|(www\.))\S+[@:.]\S+/gu;
```
(I added `+` and `.`, apparently those are allowed too)

app/scripts/lib/setupSentry.js
danjm: fixed in f183c5f2f

app/scripts/lib/setupSentry.js
danjm: What do you think of the approach I took in 526a5d30b?

If we try to get the hostname from the regex, then we need to make the regex sophisticated enough to match the hostname but not subdomains, and also not things that look like host names in other parts of the url. There is a decent amount of safety that comes with relying on `URL`.

My approach here is just to continue to rely on that, but just catch errors and hide any url that is invalid,

Gudahtt: The approach taken in that commit does not match subdomains, which is a bad thing. We do want to match subdomains - otherwise we'll be filtering out URLs we use internally. `codefi.network` for example is just the root TLD we use; it's the subdomains that we're actually sending requests to.

>If we try to get the hostname from the regex, then we need to make the regex sophisticated enough to match the hostname but not subdomains, and also not things that look like host names in other parts of the url.

Hmm, wouldn't my suggestion work here? Use the regex I suggested right after the double forward slash.

danjm: I think we have to do as done in 352d30677

Gudahtt: I'm not sure what you mean by that comment. But that latest commit won't work either, because now it will incorrectly match domains that have `codefi.network` as a subdomain (e.g. `codefi.network.another.domain.com`). We can't look for matches indiscriminately in the hostname, we'll have to look just at hostnames that either match exactly, or end with a period followed by the domain we're trying to match.

That's what I meant by this sentence:

>we can look for any that match an allowlist entry or end with .[entry] (to capture subdomains)

app/scripts/lib/setupSentry.js
```diff
@@ -146,31 +154,131 @@ export default function setupSentry({ release, getState }) {
...
+    const re = /(([-.+a-zA-Z]+:\/\/)|(www\.))\S+[@:.]\S+/gu;
+    const urlsInMessage = newErrorMessage.match(re) || [];
+    urlsInMessage.forEach((url) => {
+      try {
+        const urlObj = new URL(url);
+        const { hostname } = urlObj;
+        if (
+          !Object.values(ERROR_URL_ALLOWLIST).some(
+            (allowedUrl) =>
+              hostname === allowedUrl || hostname.endsWith(`.${allowedUrl}`),
```
Gudahtt, line 258: Nit: maybe more clear to call this variable `allowedHostname`.
```suggestion
            (allowedHostname) =>
              hostname === allowedUrl || hostname.endsWith(`.${allowedHostname}`),
```

app/scripts/lib/setupSentry.js
Gudahtt: Comparing your approach to what I suggested, it looks like this strategy will be equivalent except in the case where a matched URL is not a valid URL and causes the URL constructor to throw. In that case, we'll fallback to hiding the matched string.

This might at worst result in some of our internal URLs being hidden when they otherwise wouldn't be. I can't think of any examples offhand that this would affect. This is probably fine.

app/scripts/lib/setupSentry.test.js
```diff
@@ -0,0 +1,241 @@
...
+          from: 'chrome-extension://abcefg/home.html',
+        },
+      };
+      const rewrittenBreadcrumb = removeUrlsFromBreadCrumb(testBreadcrumb);
+      expect(rewrittenBreadcrumb.data.from).toStrictEqual(
+        'chrome-extension://abcefg/home.html',
+      );
+    });
+
+    it('should hide to but not from or url if to is the only one not matching an internal url', () => {
```
Gudahtt, line 225: Nit: could we quote "to" and "from" in these descriptions? :sweat_smile: It took me *way* too long to parse this description.

Gudahtt:
LGTM except for the e2e test failure. Unsure whether that's related to these changes or not.

Gudahtt (approved):

app/scripts/lib/setupSentry.js
digiwand: I copied and pasted this a while back when the first commit was made

digiwand: I copied and pasted this a while back when the first commit was made. New regex looks good to me

digiwand (approved):
👍

In regards to https://github.com/MetaMask/metamask-extension/pull/16780#discussion_r1046282746, I think it would be a good idea to create a follow-up ticket for this

PR #16788, opened 2022-12-02
[GridPlus] Bumps `gridplus-sdk` to v2.4.0

alex-miller-0:
The only change relevant to MetaMask users is a [fix to an ABI decoding util](https://github.com/GridPlus/gridplus-sdk/pull/483/commits/23b42df0eb89d6aa04352676212a020e270b105c) that addresses an edge case that led to failed ABI decoding for certain data types.

Full changes: https://github.com/GridPlus/gridplus-sdk/compare/v2.2.9...v2.4.0

### DIFF
--- a/lavamoat/browserify/beta/policy.json
+++ b/lavamoat/browserify/beta/policy.json
@@ -612,11 +612,6 @@
         "@ethersproject/bignumber>@ethersproject/logger": true
       }
     },
-    "@metamask/assets-controllers>abort-controller": {
-      "globals": {
-        "AbortController": true
-      }
-    },
     "@metamask/assets-controllers>async-mutex": {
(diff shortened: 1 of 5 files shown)


FrederikBolding:
Hey @alex-miller-0

Looks like your new package version is pulling in some WebAssembly that we don't allow running in Chrome.

```
WebAssembly.Module(): Refused to compile or instantiate WebAssembly module because neither 'wasm-eval' nor 'unsafe-eval' is an allowed source of script in the following Content Security Policy directive: "script-src 'self'"
```

alex-miller-0:
Thanks @FrederikBolding. That module is for changes unrelated to MetaMask, but I will need to try to find a replacement.

PR #16801, opened 2022-12-05
[FLASK] Resolve `snap_dialog` approval on close

### LINKED ISSUE #16761: Consider resolving `snap_dialog` confirmations to `null` instead of rejecting it when closed
RIght now extension rejects confirmations when the user close the confirmation by cllicking the window's close button. Probably due to this:

https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/background.js#L670

Change it so the `snap_dialog` resolves confirmations to `null` instead of rejecting the confirmation.

GuillaumeRx:
## Explanation

Resolve approvals of type `snap_dialog_*` with the value `null` instead of rejecting them on close.

* Fixes #16761 

## Manual Testing Steps

- Trigger a `snap_dialog` confirmation (you can use https://github.com/GuillaumeRx/snap-dialog-test).
- Close the confirmation using the window's close button.
- Observe that no error was thrown.

### DIFF
--- a/app/scripts/background.js
+++ b/app/scripts/background.js
@@ -29,6 +29,7 @@ import {
 import { checkForLastErrorAndLog } from '../../shared/modules/browser-runtime.utils';
 import { isManifestV3 } from '../../shared/modules/mv3.utils';
 import { maskObject } from '../../shared/modules/object.utils';
+import { RestrictedMethods } from '../../shared/constants/permissions';
 import migrations from './migrations';
 import Migrator from './lib/migrator';
 import ExtensionPlatform from './platforms/extension';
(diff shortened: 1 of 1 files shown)


hmalik88 (dismissed):

app/scripts/background.js
```diff
@@ -720,10 +721,18 @@ function setupController(initState, initLangCode) {
           REJECT_NOTFICIATION_CLOSE,
         ),
       );
-
-    // Finally, reject all approvals managed by the ApprovalController
-    controller.approvalController.clear(
-      ethErrors.provider.userRejectedRequest(),
+    // Finally, resolve snap dialog approvals and reject all the others managed by the ApprovalController
+    Object.values(controller.approvalController.state.pendingApprovals).forEach(
```
FrederikBolding, line 725: This should probably be fenced?

GuillaumeRx: How could we fence this so it also works when it's not Flask ?

app/scripts/background.js
```diff
@@ -721,10 +724,28 @@ function setupController(initState, initLangCode) {
...
       );
 
-    // Finally, reject all approvals managed by the ApprovalController
+    // Finally, resolve snap dialog approvals on Flask and reject all the others managed by the ApprovalController.
+
+    ///: BEGIN:ONLY_INCLUDE_IN(flask)
+    Object.values(controller.approvalController.state.pendingApprovals).forEach(
+      ({ id, type }) => {
+        if (type.startsWith(RestrictedMethods.snap_dialog)) {
+          controller.approvalController.accept(id, null);
```
FrederikBolding, line 733: This doesn't mirror the results of all `snap_dialog` types. Confirmations either return `true` or `false`.

app/scripts/background.js
```diff
@@ -721,10 +724,28 @@ function setupController(initState, initLangCode) {
...
+        } else {
+          controller.appStateController.reject(
```
FrederikBolding, line 735: Is this needed if we have the below `clear()` in Flask too? Maybe we don't need to fence the `clear` then 🤔

FrederikBolding (changes requested):

FrederikBolding (approved):

Mrtenz (approved):

hmalik88 (approved):

PR #16805, opened 2022-12-05
Fix getFullTxData so that customTxParamsData is not lost

### LINKED ISSUE #16798: [Bug]: Approve amount is set to Dapp default even if user changes the amount for both Improved and Legacy Approve screens
### Describe the bug

**Problem**: as per @bschorchit observation, 

> We’re sending the transaction with the amount proposed by the dapp even if the user has inputted and approved the transaction with a custom one.

https://user-images.githubusercontent.com/54408225/205606567-ba8a1fa3-79bf-4136-868b-dedf0038bdf1.mov

This also happens with the old Approve screen:

https://user-images.githubusercontent.com/54408225/205585915-bbb732ef-4eb1-49e3-b8d1-421ec8cc35ce.mp4

### Steps to reproduce

You can perform these actions both with Improve Approve screen functionality activated and deactivated

1. Go to Test Dapp
2. Click Connect and Accept prompt on MM
3. Click Create Token and Accept prompt on MM
4. Click Approve Tokens
5. Click Edit permission
6. change amount
7. Click Send
8. Check tx details on Etherscan
9. Click Decode input data -- see amount is the one suggested by the dapp and not the one you introduced

_No response_

### Version

10.23.0

### Build type

None

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

danjm:
Fixes https://github.com/MetaMask/metamask-extension/issues/16798#top

This bug was introduced here https://github.com/MetaMask/metamask-extension/commit/4245f24a2e485bd016c8ec9bd14afb17790efdbf#diff-45ae84764c1d49a07813a91adc3351d6ed96bf9ee8fa2ae4c9a116d1e0cba06cR821-R830

**Explanation**

The bug is that the user inputted token allowance is not the token allowance that ends up being submitted on chain. More precisely, if we show the user the token allowance screen and they customize it, the new `txParams.data` is not used.

The cause is the above linked codechange whereby an additional parameter was accidentally added to the result function of the selector creator, thereby rendering `customTxParamsData` undefined. With that parameter removed, `customTxParamsData` will now be defined as it was before, and will reflect the modified data in the case of a user editing a token allowance amount.

### DIFF
--- a/ui/selectors/selectors.js
+++ b/ui/selectors/selectors.js
@@ -851,7 +851,7 @@ export const getFullTxData = createDeepEqualSelector(
     return getTransaction(state, transactionId);
   },
   (_state, _transactionId, _status, customTxParamsData) => customTxParamsData,
-  (txData, transaction, _status, customTxParamsData) => {
+  (txData, transaction, customTxParamsData) => {
     let fullTxData = { ...txData, ...transaction };
     if (transaction && transaction.simulationFails) {
       txData.simulationFails = transaction.simulationFails;


pedronfigueiredo (approved):

jpuri (approved):

ui/selectors/selectors.js
```diff
@@ -851,7 +851,7 @@ export const getFullTxData = createDeepEqualSelector(
     return getTransaction(state, transactionId);
   },
   (_state, _transactionId, _status, customTxParamsData) => customTxParamsData,
-  (txData, transaction, _status, customTxParamsData) => {
```
adonesky1, line 854: this looks right to me when I step through the code while running the example. 

I think I was (and maybe still am a bit confused by `input` vs. `output` selectors that are passed to the `createSelector` function as described in the [reduxToolkit docs](https://redux.js.org/usage/deriving-data-selectors#createselector-overview). 

@brad-decker are you able to elucidate this pattern a bit better for me?

adonesky1: Is the last arg passed always the output selector? And then you pass as many input selector before that as there are args? That doesn't seem to match what we have here... I think I'm overcomplicating this.

brad-decker: One second while I look up what 'elucidate' means

PR #16822, opened 2022-12-06
Fixed navigation through multiple unapproved transactions for ERC20 tokens

### LINKED ISSUE #16768: [Bug]: Users cannot navigate through multiple unapproved `approve` transactions
### Describe the bug

The previous and next page navigation arrows are no longer present in the header of the `approve screen`

Before: Navigation arrows present

https://user-images.githubusercontent.com/53189696/205168516-fbb16f2d-9ca0-49ac-8800-a190e7a5262e.mov

Now: Missing navigation arrows

https://user-images.githubusercontent.com/53189696/205168556-cfb8850a-9a63-42f7-a23a-ad02348ed7ff.mov

### Steps to reproduce

1. Navigate to the [test dapp](https://metamask.github.io/test-dapp)
2. Connect your wallet
3. Deploy the erc20 token
4. Click `Approve Token` multiple times
5. Navigate back to the wallet

_No response_

### Version

10.23.0

### Build type

None

### Browser

Chrome

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

filipsekulic:
## Explanation
Fixed navigation through multiple unapproved transactions for ERC20 tokens.
* Fixes #16768

### Before

https://user-images.githubusercontent.com/92310504/205946978-87baa151-14f2-476a-b252-5a91531d047b.mov

### After

https://user-images.githubusercontent.com/92310504/205947039-35e83d04-c827-4ba8-8024-d42923332895.mov

## Manual Testing Steps

1. Navigate to the [test dapp](https://metamask.github.io/test-dapp)
2. Connect your wallet
3. Deploy the `ERC20` token
4. Click `APPROVE TOKENS` multiple times
5. Navigate back to the wallet

### DIFF
--- a/ui/components/app/confirm-page-container/confirm-page-container-navigation/confirm-page-container-navigation.component.js
+++ b/ui/components/app/confirm-page-container/confirm-page-container-navigation/confirm-page-container-navigation.component.js
@@ -1,26 +1,29 @@
 import React, { useContext } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { useHistory } from 'react-router-dom';
-import { getTxData, getUnapprovedTransactions } from '../../../../selectors';
+import PropTypes from 'prop-types';
+import {
+  getCurrentChainId,
+  getUnapprovedTransactions,
+} from '../../../../selectors';
 import { transactionMatchesNetwork } from '../../../../../shared/modules/transaction.utils';
 import { I18nContext } from '../../../../contexts/i18n';
 import { CONFIRM_TRANSACTION_ROUTE } from '../../../../helpers/constants/routes';
 import { clearConfirmTransaction } from '../../../../ducks/confirm-transaction/confirm-transaction.duck';
 import { hexToDecimal } from '../../../../../shared/lib/metamask-controller-utils';
 
-const ConfirmPageContainerNavigation = () => {
+const ConfirmPageContainerNavigation = ({ txData }) => {
   const t = useContext(I18nContext);
   const dispatch = useDispatch();
   const history = useHistory();
 
   const unapprovedTxs = useSelector(getUnapprovedTransactions);
-  const txData = useSelector(getTxData);
-
-  const network = hexToDecimal(txData.chainId);
+  const currentChainId = useSelector(getCurrentChainId);
+  const network = hexToDecimal(currentChainId);
 
   const currentNetworkUnapprovedTxs = Object.keys(unapprovedTxs)
     .filter((key) =>
-      transactionMatchesNetwork(unapprovedTxs[key], txData.chainId, network),
+      transactionMatchesNetwork(unapprovedTxs[key], currentChainId, network),
     )
     .reduce((acc, key) => ({ ...acc, [key]: unapprovedTxs[key] }), {});
 
@@ -104,6 +107,10 @@ const ConfirmPageContainerNavigation = () => {
       </div>
     </div>
   );
+};
+
+ConfirmPageContainerNavigation.propTypes = {
+  txData: PropTypes.object,
 };
 
 export default ConfirmPageContainerNavigation;
--- a/ui/components/app/confirm-page-container/confirm-page-container-navigation/confirm-page-container-navigation.component.js
+++ b/ui/components/app/confirm-page-container/confirm-page-container-navigation/confirm-page-container-navigation.component.js
@@ -1,19 +1,48 @@
-import React from 'react';
-import PropTypes from 'prop-types';
+import React, { useContext } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import { useHistory } from 'react-router-dom';
+import { getTxData, getUnapprovedTransactions } from '../../../../selectors';
+import { transactionMatchesNetwork } from '../../../../../shared/modules/transaction.utils';
+import { I18nContext } from '../../../../contexts/i18n';
+import { CONFIRM_TRANSACTION_ROUTE } from '../../../../helpers/constants/routes';
+import { clearConfirmTransaction } from '../../../../ducks/confirm-transaction/confirm-transaction.duck';
+import { hexToDecimal } from '../../../../../shared/lib/metamask-controller-utils';
 
-const ConfirmPageContainerNavigation = (props) => {
-  const {
-    onNextTx,
-    totalTx,
-    positionOfCurrentTx,
-    nextTxId,
-    prevTxId,
-    showNavigation,
-    firstTx,
-    lastTx,
-    ofText,
-    requestsWaitingText,
-  } = props;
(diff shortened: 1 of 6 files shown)


mirjanaKukic:
Verified by QA

jpuri (dismissed):
Look good 👍

ui/pages/token-allowance/token-allowance.js
```diff
@@ -208,6 +212,38 @@ export default function TokenAllowance({
...
+  const showNavigation = enumUnapprovedTxs.length > 1;
+  const firstTx = enumUnapprovedTxs[0];
+  const lastTx = enumUnapprovedTxs[enumUnapprovedTxs.length - 1];
+
+  const handleNextTx = (txId) => {
+    if (txId) {
+      dispatch(clearConfirmTransaction());
+      history.push(`${CONFIRM_TRANSACTION_ROUTE}/${txId}`);
+    }
+  };
```
brad-decker, line 245: I don't think this is problematic but what I'm wondering is if there's an opportunity to improve the readability of our code by relocating and consolidating this logic and the 'confirm-transaction-base.container.js' files logic and move it *into* the `<ConfirmPageContainerNavigation>` component which appears to already be a functional component. Doing this would make that specific part of our code much more understandable and perhaps prevent needing to do this selector logic in multiple places.

Could you take a look into this and see how much effort it'll be to consolidate? Thanks @filipsekulic

brad-decker (changes requested):

ui/pages/token-allowance/token-allowance.js
filipsekulic: Hey @brad-decker! Good suggestion!
I refactored the code and you can see the changes here: f4b09382face3e44f085f59f4551ef4c5fb99272
However, there is a problem and you can see it in the attached screenshot, which is the reason why the pipeline is failing (`e2e test`). 
This happens when there are multiple token approvals and when a user confirms one of them, the navigation disappears for some reason and throws the error (screenshot attached). I think the problem is in the `confirm-page-container.component.js` within the `async componentDidMount`, which the error message is also suggesting (`...cancel all subscriptions and asynchronous tasks...`). 
I would appreciate if you could take a look and give your opinion on this one.

Screenshot 2022-12-15 at 14 01 42

ui/components/app/confirm-page-container/confirm-page-container-navigation/confirm-page-container-navigation.component.js
```diff
@@ -1,19 +1,48 @@
...
+const ConfirmPageContainerNavigation = () => {
+  const t = useContext(I18nContext);
+  const dispatch = useDispatch();
+  const history = useHistory();
+
+  const unapprovedTxs = useSelector(getUnapprovedTransactions);
+  const txData = useSelector(getTxData);
+
+  const network = hexToDecimal(txData.chainId);
+
```
danjm, line 20: The "Navigate transactions - should confirm and remove an unapproved transaction" e2e test is failing because: `TimeoutError: Waiting for element to be located By(xpath, .//*[contains(concat(' ', normalize-space(./@class), ' '), ' confirm-page-container-navigation ')][(contains(string(.), '1 of 3') or contains(string(.), '1of3'))])`

This probably means that during the e2e test run `showNavigation` is false, which probably means that `enumUnapprovedTxs` is an empty array. Looking at this new code for getting  `currentNetworkUnapprovedTxs` compared to the old code in `confirm-transaction-base.container.js`, I notice that the `chainId` and `network` are here derived from the transaction but in `confirm-transaction-base.container.js` they are taken directly from `state.metamask`.

If this feature is working when running it manually in the browser, for both custom approvals and erc-20 token transfers, then the problem is likely in e2e test fixtures (unless there is something wrong with getting the chainId and network from the transaction?... I don't think so, but worth a double check)

filipsekulic: You are right @danjm. Good catch! 
I fixed that and also made a small change that solved the problem, so now `txData` is being passed from a parent component.
You can check it here: 1d005ae893d9a9da7350b0d14c5f649ac9905858

ui/pages/token-allowance/token-allowance.js
filipsekulic: Everything is fixed now - https://github.com/MetaMask/metamask-extension/pull/16822#discussion_r1053301032.

ui/components/app/confirm-page-container/confirm-page-container-navigation/confirm-page-container-navigation.component.js
danjm: why did you have to change it so that `txData` is being passed in, instead of being selected? what problem did that solve?

filipsekulic: When a user tries to confirm one of the transactions, it now shows the navigation, but doesn't get updated as it should - `txData` object being selected becomes empty, next `tx` cannot be seen on the next page and `currentPosition` is `-1` as the `currentTx` cannot be found from `enumUnapprovedTxs`, because there is no `txId` from `txData` object which is empty at that moment.
Screencast:

https://user-images.githubusercontent.com/92310504/208698663-50a774d2-29f0-4d4c-b161-7732f0e26efa.mov

jpuri (dismissed):

ui/components/app/confirm-page-container/confirm-page-container-navigation/confirm-page-container-navigation.component.js
brad-decker: txData is derived from state in confirm-transaction-base.container.js which means we can get it here without drilling the prop down the line

brad-decker (changes requested):

ui/components/app/confirm-page-container/confirm-page-container-navigation/confirm-page-container-navigation.component.js
filipsekulic: I get it, but the problem with it is that somehow the `txData` gets empty when one of the transactions is confirmed.
You can see it in the above attached video where `txDataFromSelector` is actually the transaction data derived from state. I `console.log`-ged it in order to show this problem...

filipsekulic:
@brad-decker 
After today's conversation with @danjm and his suggestion, I managed to get the `transaction id` by using the `useParams`, so I removed the `txData` as the component's property and everything seems to work fine.
You can check the changes here: 453de3cb9eeebcf0d284a82ea99d56ff54a8c05a

danjm (approved):

brad-decker (approved):

PR #16856, opened 2022-12-08
Fix shared mocks in RPC network client test

Gudahtt:
Some of the provider api unit tests were inadvertently sharing test mocks. A `params` object used for mock RPC calls was being shared between tests and between calls within individual tests. They have been updated to generate a fresh `params` object for each mock RPC call.

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/app/scripts/controllers/network/provider-api-tests/helpers.js
+++ b/app/scripts/controllers/network/provider-api-tests/helpers.js
@@ -331,40 +331,27 @@ export async function withInfuraClient(...args) {
 }
 
 /**
- * Some JSON-RPC endpoints take a "block" param (example: `eth_blockNumber`)
- * which can optionally be left out. Additionally, the endpoint may support some
- * number of arguments, although the "block" param will always be last, even if
- * it is optional. Given this, this function builds a mock `params` array for
- * such an endpoint, filling it with arbitrary values, but with the "block"
- * param missing.
+ * Build mock parameters for a JSON-RPC call.
  *
- * @param {number} index - The index within the `params` array where the "block"
- * param *would* appear.
- * @returns {string[]} The mock params.
- */
-export function buildMockParamsWithoutBlockParamAt(index) {
-  const params = [];
-  for (let i = 0; i < index; i++) {
-    params.push('some value');
-  }
-  return params;
-}
-
-/**
- * Some JSON-RPC endpoints take a "block" param (example: `eth_blockNumber`)
- * which can optionally be left out. Additionally, the endpoint may support some
- * number of arguments, although the "block" param will always be last, even if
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/network/provider-api-tests/helpers.js
```diff
@@ -331,40 +331,27 @@ export async function withInfuraClient(...args) {
...
- * @param {any} blockParam - The desired "block" param to add.
+ * @param {object} args - Arguments.
+ * @param {number} args.blockParamIndex - The index of the block parameter.
+ * @param {any} [args.blockParam] - The block parameter value to set.
  * @returns {any[]} The mock params.
  */
-export function buildMockParamsWithBlockParamAt(index, blockParam) {
-  const params = buildMockParamsWithoutBlockParamAt(index);
-  params.push(blockParam);
+export function buildMockParams({ blockParam, blockParamIndex }) {
```
Gudahtt, line 347: My first attempt at moving the invocations of these two helper functions into each test ended up being overly verbose and hard to read.  The condition introduced in choosing between these two cases seemed unnecessary in particular.

The two helper functions were simplified into a single function that was shorter and easier to invoke within each test with no condition needed.

brad-decker (dismissed):
Nice.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -642,15 +641,13 @@ export function testsForRpcMethodSupportingBlockParam(
     ['given no block tag', 'none'],
```
mcmire, line 641: The reason for the two functions was to test that if the block param was not passed then it would default to latest. That doesn't appear to happen now — we always pass a block param.

Gudahtt: It's set to `undefined` in all of the cases where the "without block param" function was used. So it should be doing the same thing.

mcmire: Ah. What do you think about making that explicit by adding `undefined` as the third value in this array?

Gudahtt: Sure, we can do that

Gudahtt: Done in https://github.com/MetaMask/metamask-extension/pull/16856/commits/9511b719887835ae24efc32698a542bbe6538af1

mcmire (approved):
LGTM!

BelfordZ (approved):

BelfordZ:
lmk if you merge this before #16856 so that I can make sure it all ends up in the resulting rebase conflict

PR #16863, opened 2022-12-08
Move MetaMask middleware out of network controller

### LINKED ISSUE #16574: Migrate signing middleware out of network client
Currently our middleware responsible for signing is part of the network client. Yet it has nothing to do with the network. We route signing RPC methods into the network stack and then back out again.

We should instead move this middleware ahead in the RPC pipeline, out of the network stack.

Gudahtt:
The "MetaMask middleware" is the set of middleware for handling methods that we don't send to the network. This includes signing, encryption, `getAccounts`, and methods that rely on pending transaction state.

Previously this middleware was setup as part of the network client, despite having nothing to do with the network. They have been moved into the main RPC pipeline established in `metamask-controller.js` instead.

Fixes #16574

## Manual Testing Steps

This is a pure refactor, and should have no functional changes. The middleware are still run in exactly the same order with the same arguments.

## Pre-merge author checklist

- [x] I've clearly explained:


brad-decker (approved):
This looks so much nicer and makes so much more sense. I'm going to pull the branch and test locally to be sure I don't miss anything but given the E2E tests passed its likely ✅

app/scripts/lib/util.js
```diff
@@ -246,3 +247,48 @@ export function previousValueComparator(comparator, initialValue) {
     }
   };
 }
+
+export function formatTxMetaForRpcResult(txMeta) {
```
Gudahtt, line 251: This was migrated from the network controllers utils module to the general utils module, because it was no longer used in the network controller

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -1,7 +1,9 @@
 import sinon from 'sinon';
-import { getNetworkDisplayName } from './util';
+import { NETWORK_TO_NAME_MAP } from '../../../../shared/constants/network';
 import NetworkController, { NETWORK_EVENTS } from './network';
 
+const getNetworkDisplayName = (key) => NETWORK_TO_NAME_MAP[key];
```
Gudahtt, line 5: This function was the last remaining thing in the network utils module, and it was only used in this test. I've moved it here instead so we can delete that module.

app/scripts/metamask-controller.js
```diff
@@ -248,7 +249,7 @@ export default class MetamaskController extends EventEmitter {
     this.networkController.setInfuraProjectId(opts.infuraProjectId);
 
     // now we can initialize the RPC provider, which other controllers require
-    this.initializeProvider();
+    this.networkController.initializeProvider();
```
Gudahtt, line 252: We still need to call this to kick off the network controller initialization, but the `providerOptions` passed in before were _solely_ used to setup the metamask middleware. Now it doesn't need parameters anymore.

brad-decker (dismissed):
I checked the following locally via test dapp unless otherwise stated:
- [x] getAccounts
- [x] processTransaction
- [x] process`<insertSignMethodHere>`
- [x] getPendingNonce (via send flow)
- [ ] getPendingTransactionByHash

wasn't sure how to best check the last one

app/scripts/metamask-controller.js
```diff
@@ -1053,6 +1054,48 @@ export default class MetamaskController extends EventEmitter {
       this.messageManager.clearUnapproved();
     });
 
+    this.metamaskMiddleware = createMetamaskMiddleware({
```
Gudahtt, line 1057: This was copied directly from the network controller, with the parameters taken from the metamask-controller.js `initializeProvider` method (that's where they were hard-coded before).

I've decided to create this middleware _once_ and reference it in the RPC pipeline setup instead of constructing it per-connection because that's effectively what we were doing before. This does depend on these middleware all being stateless, but that was already an assumption we were relying on, at least partially.

It was reconstructed upon each network switch, whereas now it is not. But this _should_ be OK because most of this has nothing to do with the network.

There is one part that depends upon the network somewhat, and that's the `getPendingTransactionByHash` and `getPendingNonce` functions. But those each use the provider proxy provided by the network controller, so they should work the same as they do today between network switches.

app/scripts/metamask-controller.test.js
```diff
@@ -175,21 +175,6 @@ describe('MetaMaskController', function () {
     });
   });
 
-  describe('#getAccounts', function () {
-    it('returns first address when dapp calls web3.eth.getAccounts', async function () {
-      const password = 'a-fake-password';
-      await metamaskController.createNewVaultAndRestore(password, TEST_SEED);
-
-      metamaskController.networkController._baseProviderParams.getAccounts(
```
Gudahtt, line 183: This test was broken because it referenced the private variable `_baseProviderParams`, which was removed. I deleted the entire test rather than fix it because I don't understand what value it provided. This is not the place to test network methods.

Gudahtt:
>wasn't sure how to best check the last one

It can be tested using this method: https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_gettransactionbyhash

Specifically, if you try to call that method on a _pending_ transaction, normally it would fail. That middleware ensures that it can return a transaction object for a local pending transaction as well.

mcmire:
I've confirmed that `eth_getTransactionByHash` when given a pending transaction hash will still give a result. For instance:

```
> ethereum.request({ method: 'eth_getTransactionByHash', params: ['0xeee1c3ad8f4129d75af93daeda18ebe5025f27dc71bafd6ec60a037afd0fd8f1'] }).then(result => console.log(result)).catch(error => console.log(error))
{
    "v": "0x01",
    "r": "0x4b94e28c672b6ea6b02b61f0abe58478fe31626db38207bca20b2b0311a0e882",
    "s": "0x64e38fe7b1544a2d53ded7162f72a496ea35825f1f8d1ccf6c1ec10ed1863604",
    "to": "0x0c54fccd2e384b4bb6f2e405bf5cbc15a017aafb",
    "gas": "0x5208",
    "from": "0xd5af2bbbed2f38fd6e6e1a44c5eeb011efde1c51",
    "hash": "0xeee1c3ad8f4129d75af93daeda18ebe5025f27dc71bafd6ec60a037afd0fd8f1",
    "nonce": "0x3",
    "input": "0x",
    "value": "0x0",
    "accessList": null,
    "blockHash": null,
    "blockNumber": null,
    "transactionIndex": null,
    "gasPrice": "0x2540be400",
    "maxFeePerGas": "0x2540be400",
    "maxPriorityFeePerGas": "0x3b9aca00",
    "type": "0x2"
}
```

brad-decker (dismissed):

mcmire (commented):
It looks like the MetaMask middleware now gets inserted before the filter/subscription middleware. Is that intentional? Will that cause any undesirable effects?

Gudahtt:
>It looks like the MetaMask middleware now gets inserted before the filter/subscription middleware. Is that intentional? Will that cause any undesirable effects?

Great question. I don't expect it to have any effect, because they should not overlap. None of the MetaMask middleware RPC handlers affect subscription or filter requests.

mcmire (dismissed):
This looks good to me.

Gudahtt:
Rebased to resolve conflicts

app/scripts/metamask-controller.js
```diff
@@ -248,7 +249,7 @@ export default class MetamaskController extends EventEmitter {
     });
```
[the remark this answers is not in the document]
Gudahtt: Good idea, done here: aab6b43810bbfc70f73f6d5ebe1be115121a4474

brad-decker (approved):

mcmire (approved):

PR #16864, opened 2022-12-08
[FLASK] Fix a crash that occurred after a snap confirmation had been approved

FrederikBolding:
## Explanation

Fix a crash that occurred after a snap confirmation had been approved.

### DIFF
--- a/ui/pages/confirmation/confirmation.js
+++ b/ui/pages/confirmation/confirmation.js
@@ -176,9 +176,10 @@ export default function ConfirmationPage({
   };
 
   ///: BEGIN:ONLY_INCLUDE_IN(flask)
-  const {
-    manifest: { proposedName },
-  } = useSelector((state) => getSnap(state, pendingConfirmation?.origin));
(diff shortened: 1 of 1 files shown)


ui/pages/confirmation/confirmation.js
```diff
@@ -176,9 +176,10 @@ export default function ConfirmationPage({
   };
 
   ///: BEGIN:ONLY_INCLUDE_IN(flask)
-  const {
-    manifest: { proposedName },
-  } = useSelector((state) => getSnap(state, pendingConfirmation?.origin));
+  const snap = useSelector((state) => getSnap(state, pendingConfirmation?.origin));
+
+  // When pendingConfirmation is undefined, this will also be undefined
```
digiwand, line 181: Should we update lines 233, 236, and 306 as well since pendingConfirmation could be undefined?

FrederikBolding: Good catch! I have moved these to be below the early return, that should alleviate any potential problems.

GuillaumeRx (approved):

digiwand (approved):

Issue #16854, opened 2022-12-07
[Bug]: DetectedTokens getting reloaded after it is imported while switching accounts


NiranjanaBinoy:
### Describe the bug

After importing all the tokens from the Detected tokens, if we switch to a different account and return to the old account, we could see the tokens we have already imported as detected tokens. This behavior is observed after the latest assets-controller bump to 3.0.0 and can be reproduced in `optional-currency-conversion`.

### Steps to reproduce

1. For an account with tokens as the detected token, import all the tokens and make sure the detected tokens link disappears.
2. Now, switch to a different account and switch back to the previous account.
3. The detected tokens link will re-appear, and upon opening the pop up it will show the same list of tokens we just imported.

### Version

10.22.3

### Build type

None

### Browser

Chrome

### Operating system

MacOS

SUPPORTTEAMHERR:
> ### Describe the bug
> 
> After importing all the tokens from the Detected tokens, if we switch to a different account and return to the old account, we could see the tokens we have already imported as detected tokens. This behavior is observed after the latest assets-controller bump to 3.0.0 and can be reproduced in `optional-currency-conversion`.
> 
> ### Steps to reproduce
> 
> 1. For an account with tokens as the detected token, import all the tokens and make sure the detected tokens link disappears.
> 2. Now, switch to a different account and switch back to the previous account.
> 3. The detected tokens link will re-appear, and upon opening the pop up it will show the same list of tokens we just imported.
> 
> 
> ### Error messages or log output
> 
> _No response_
> 
> ### Version
> 
> 10.22.3
> 
> ### Build type
> 
> None
> 
> ### Browser
> 
> Chrome
> 
> ### Operating system
> 
> MacOS
> 
> ### Hardware wallet
> 
> _No response_
> 
> ### Additional context
> 
> _No response_

We'd be happy to look into this for you. kindly file a support ticket below 
https://forms.gle/Wmpm6NLzYm9uXYuU9

PR #16772, opened 2022-12-02
Toggle option to enable/disable balance and Token rate checking for using third-party API

NiranjanaBinoy:
Fixes #16724 
## Explanation
We have added a new toggle to the `Privacy and Security` tab to control the use of third-party APIs like coinGecko and CryptoCompare for balance and Token rate checking.

### Before
Screenshot 2022-12-07 at 6 00 34 PM
Screenshot 2022-12-07 at 6 00 57 PM
Screenshot 2022-12-07 at 6 01 47 PM
Screenshot 2022-12-07 at 6 02 09 PM

### After
Screenshot 2022-12-07 at 5 59 02 PM
Screenshot 2022-12-07 at 5 58 28 PM
Screenshot 2022-12-07 at 5 59 31 PM
Screenshot 2022-12-07 at 5 59 52 PM

## Manual Testing Steps
Onboard v1:
  1. Load the application and import a wallet.
  2. The fiat values should not be visible - in eth-overview, token overview, or in asset/token lists
  3. Try to send a token, and fiat values for gas or total should not be visible.
  4. Now go to settings, search for `Show balance and token price checker` in the security tab and toggle it ON.
  5. Now, come back to the overview page and make sure Fiat values are visible.
  6. Again, Toggle OFF and make sure the Fiat values disappear from everywhere.
  
Onboarding V2:
1. Load the application and import a wallet.
2. Enable the `Show balance and token price checker` from the Privacy settings.
3. The app will display the fiat values.
4. reload the application and import the wallet again.
5. This time, without enabling `Show balance and token price checker`.
6. the Fiat values should not e displayed.

## Pre-merge author checklist

- [ ] I've clearly explained:
  - [ ] What problem this PR is solving
  - [ ] How this problem was solved
  - [ ] How reviewers can test my changes
- [ ] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [ ] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
- [ ] PR is linked to the appropriate GitHub issue
- [ ] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.

In this case, a QA Engineer approval will be be required.

PR #16866, opened 2022-12-08
fix cached detected token results

### LINKED ISSUE #16349: [Bug]: [MV3] Token detection is caching results between accounts when terminating service worker
### Describe the bug

**Problem**: when I terminate the service worker and I change accounts, I see that the tokens detected are cached.

https://user-images.githubusercontent.com/54408225/199470301-596396d3-e524-4097-9a1e-faaa98421993.mp4

### Steps to reproduce

1. Unlock MM
2. Get some tokens on the first account, so they are detected
3. Create a fresh new account
4. Change back to the account from step 2
5. Terminate service worker
6. Change back to the account from step 3 -- see that token detected are cached

_No response_

### Version

Tested on top of this PR https://github.com/MetaMask/metamask-extension/pull/15558

### Build type

_No response_

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

There is another caching issue with Token Detection which might also be related? See this comment and issue on MV2
https://github.com/MetaMask/metamask-extension/issues/16057#issuecomment-1291129607

adonesky1:
~Blocked by https://github.com/MetaMask/controllers/pull/956 (and subsequent release of `@metamask/controllers`)~

Context (from https://github.com/MetaMask/controllers/pull/956):

_Currently there exists an issue (exposed by https://github.com/MetaMask/metamask-extension/issues/16349) where if a user changes accounts (account x to account y) or (changes from network a to network b) right as a detectTokens() call is occuring, the results of the getBalancesInSingleCall() call (withindetectTokens()) made against address x or network acan get incorrectly misattributed to accountyornetwork b. This can happen because currently detectTokens() simply calls this.addDetectedTokens() on the TokensController with the detected tokensToAdd assuming that when this call occurs in the TokensController the selectedAddress or chainId [used to store these detected assets](https://github.com/MetaMask/controllers/blob/9b82201b4c99059a7ab35942febdba840adce5ee/packages/assets-controllers/src/TokensController.ts#L709) will be the same as the one used to detect them. With sufficiently unfortunate timing this expectation can be broken, and so this PR introduces a pattern ([already used in the analogous flow between the NftDetectionController and the NftController](https://github.com/MetaMask/controllers/blob/main/packages/assets-controllers/src/NftDetectionController.ts#L397), whereby the address and network data are passed along with the detectedToken data to ensure they are stored correctly._

This PR leverages the changes made in https://github.com/MetaMask/controllers/pull/956, and additionally fixes a race condition that occurs where even if the detection data is stored correctly, the `detectedTokens` value does not get updated to reflect the correct chainId/network combo if/when the network/account is changed while the service worker is being restarted.

* Fixes https://github.com/MetaMask/metamask-extension/issues/16349

### Before

https://user-images.githubusercontent.com/34557516/206777046-4fbabb91-bcba-4bb9-9ffe-d380a0665238.mov

### After

https://user-images.githubusercontent.com/34557516/206775291-19339d64-232c-4271-99ad-7839165f3b76.mov

## Manual Testing Steps

1. Unlock MM
2. Get some tokens on the first account, so they are detected
3. Create a fresh new account
4. Change back to the account from step 2
5. Terminate service worker
6. Change back to the account from step 3 -- see that no detected tokens are cached

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/app/scripts/controllers/detect-tokens.js
+++ b/app/scripts/controllers/detect-tokens.js
@@ -89,7 +89,8 @@ export default class DetectTokensController {
    */
   async detectNewTokens({ selectedAddress, chainId }) {
     const addressAgainstWhichToDetect = selectedAddress ?? this.selectedAddress;
-    const chainIdAgainstWhichToDetect = chainId ?? this.chainId;
+    const chainIdAgainstWhichToDetect =
+      chainId ?? this.getChainIdFromNetworkStore(this._network);
     if (!this.isActive) {
       return;
     }
--- a/app/scripts/controllers/detect-tokens.js
+++ b/app/scripts/controllers/detect-tokens.js
@@ -191,7 +191,7 @@ export default class DetectTokensController {
    * @param options0.selectedAddress
    * @param options0.chainId
    */
-  restartTokenDetection({ selectedAddress, chainId }) {
+  restartTokenDetection({ selectedAddress, chainId } = {}) {
     const addressAgainstWhichToDetect = selectedAddress ?? this.selectedAddress;
     const chainIdAgainstWhichToDetect = chainId ?? this.chainId;
     if (
--- a/app/scripts/controllers/detect-tokens.js
+++ b/app/scripts/controllers/detect-tokens.js
@@ -66,7 +66,7 @@ export default class DetectTokensController {
       ) {
         this.selectedAddress = selectedAddress;
         this.useTokenDetection = useTokenDetection;
-        this.restartTokenDetection();
+        this.restartTokenDetection({ selectedAddress });
       }
     });
     tokensController?.subscribe(
--- a/app/scripts/controllers/detect-tokens.js
+++ b/app/scripts/controllers/detect-tokens.js
@@ -82,28 +82,30 @@ export default class DetectTokensController {
 
   /**
    * For each token in the tokenlist provided by the TokenListController, check selectedAddress balance.
+   *
+   * @param options0
+   * @param options0.selectedAddress
+   * @param options0.chainId
    */
-  async detectNewTokens() {
+  async detectNewTokens({ selectedAddress, chainId }) {
+    const addressAgainstWhichToDetect = selectedAddress ?? this.selectedAddress;
+    const chainIdAgainstWhichToDetect = chainId ?? this.chainId;
     if (!this.isActive) {
       return;
     }
-    if (
-      !isTokenDetectionEnabledForNetwork(
-        this.getChainIdFromNetworkStore(this._network),
-      )
-    ) {
+    if (!isTokenDetectionEnabledForNetwork(chainIdAgainstWhichToDetect)) {
       return;
     }
     if (
       !this.useTokenDetection &&
-      this.getChainIdFromNetworkStore(this._network) !== CHAIN_IDS.MAINNET
+      chainIdAgainstWhichToDetect !== CHAIN_IDS.MAINNET
     ) {
       return;
     }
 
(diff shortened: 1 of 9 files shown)


app/scripts/controllers/detect-tokens.js
```diff
@@ -182,12 +188,26 @@ export default class DetectTokensController {
...
-  restartTokenDetection() {
-    if (!(this.isActive && this.selectedAddress)) {
+  restartTokenDetection({ selectedAddress, chainId } = {}) {
+    const addressAgainstWhichToDetect = selectedAddress ?? this.selectedAddress;
+    const chainIdAgainstWhichToDetect = chainId ?? this.chainId;
+    if (
+      !(
+        this.isActive ||
+        !addressAgainstWhichToDetect ||
+        chainIdAgainstWhichToDetect !== CHAIN_IDS.MAINNET
```
NiranjanaBinoy, line 202: Currently, token detection is enabled for only MAINNET but on Binance, Polygon, and Avalanche. We are checking for the same in line 97 as well.

NiranjanaBinoy:
This PR is showing some issues with the display of the detected tokens link in the overview page, this is because `allDetectedTokens` is not getting cleared properly when the tokens are imported, and then we are using data from `allDetectedTokens[chainIs][selectedAddress]` to get the detectedTokens on the front end.  We already have a [ticket](https://github.com/MetaMask/metamask-extension/issues/16854) for this issue related to [balance and token price checking PR](https://github.com/MetaMask/metamask-extension/pull/16772)

app/scripts/controllers/detect-tokens.js
jpuri: May be use=ing method `isTokenDetectionEnabledForNetwork` will be better here than this comparison.

adonesky1: true I suppose we can just remove this check since it takes place in `detectNewTokens` which is called below.

adonesky1: updated [here](https://github.com/MetaMask/metamask-extension/pull/16866/commits/34af9a7e31b45602662c3b1f802f3022cc1f99b2) @NiranjanaBinoy

app/scripts/controllers/detect-tokens.js
```diff
@@ -219,8 +233,9 @@ export default class DetectTokensController {
     this._network = network;
     this._network.store.subscribe(() => {
       if (this.chainId !== this.getChainIdFromNetworkStore(network)) {
-        this.restartTokenDetection();
-        this.chainId = this.getChainIdFromNetworkStore(network);
+        const chainId = this.getChainIdFromNetworkStore(network);
+        this.chainId = chainId;
+        this.restartTokenDetection({ chainId: this.chainId });
```
NiranjanaBinoy, line 238: Shouldn't we update the `this.restartTokenDetection()` with params with selectedAddress  in line 69 where it is called when the preference state changes with selectedAddress change as well?

adonesky1: yes! I think that change must have gotten lost in a rebase!

adonesky1: Great catch!

adonesky1: Done [here](https://github.com/MetaMask/metamask-extension/pull/16866/commits/02b67c3cf8b6a357ec81213bf162b5585e716ca1)

NiranjanaBinoy (approved):
LGTM

jpuri (approved):

PR #16872, opened 2022-12-08
Feat/16637/button housekeeping

### LINKED ISSUE #16637: 🧹 [UI House Keeping] `Button`
### Description
Ensure that `Button` adheres to all of the following conventions and standards

- [ ] Has a `className` prop and the PropType descriptions are all the same
- [ ] Prop table in MDX docs have the "Accepts all Box component props" description and link
- [ ] We are consistent when using the same prop names like `size` and are suggesting the use of the generalized `design-system.js` constants e.g. `SIZES` as the primary option but noting the component consts in the documentation and using them for propType validation and storybook controls only
- [ ] Standardize all similar prop names for images `imgSrc`, `imgAlt`(html element + attribute) (needs audit)
- [ ] We have a story for each component prop and we use the prop name verbatim e.g. `size` prop would be `export const Size = (args) => (`
- [ ] We have the accompanying documentation for each component prop and we use the prop name verbatim e.g. `size` prop would be `### Size`
- [ ] Are multiple props stories allowed? e.g. `Color, Background Color And Border Color` story in `base-avatar` - [ ] yes when it makes sense to
- [ ] All Base components follow the suffix convention e.g. `ButtonBase`
- [ ] All Base component MDX documentation have the base component notification at the top
- [ ] Add `mm-` prefix to all classNames
- [ ] className is kebab case version of the component name
- [ ] Spread base components props and reduce duplication of props when props aren't being changed and remain the same for both variant and base components
- [ ] Add component to root `index.js` file in component-library
- [ ] Add locals for any default text I18nContext as default context 
- [ ] Add any "to dos" with a `// TODO:` comment so we can search for them at a later date e.g. blocking components etc
- [ ] Add snapshot testing
- [ ] Add pixel values to propType descriptions if we use abstracted prop types that relate to pixel values e.g. `SIZE.MD (32px)`
- [ ] Each prop section in the MDX docs should have: a heading, a description, a story and an example code snipped

garrettbear:
## Explanation

* Fixes #16637 

Add props to README, fix `mm-button-base` classname, and fix iconName issue 

### Before
Screenshot 2022-12-08 at 10 28 46 AM

### After
Screenshot 2022-12-08 at 10 28 13 AM

## Manual Testing Steps
`yarn test:unit:jest .ui/components/component-library/button/button.test.js `

## Pre-merge author checklist

- [X] I've clearly explained:


ui/components/component-library/button-link/button-link.scss
```diff
@@ -11,7 +11,7 @@
     opacity: 0.5;
   }
 
-  &.mm-button--disabled {
+  &.mm-button-base--disabled {
```
georgewrmarshall, line 14: This is more of a question that will need further discussion but should we be adding a new class to handle these styles associated with this component rather that using a className from the child?

georgewrmarshall (commented):
Looking good! Posted a couple questions. I think class name was incorrectly updated in the `button-link.scss` it removes some of the styling 

Screenshot 2022-12-12 at 3 27 19 PM

ui/components/component-library/button/README.mdx
```diff
@@ -1,20 +1,30 @@
...
 
+The `Button` accepts all [ButtonBase](/docs/ui-components-component-library-button-base-button-base-stories-js--default-story) component props
+
+<ArgsTable of={ButtonBase} />
+
+The `Button` accepts all [ButtonPrimary](/docs/ui-components-component-library-button-primary-button-primary-stories-js--default-story), [ButtonSecondary](/docs/ui-components-component-library-button-secondary-button-secondary-stories-js--default-story), and [ButtonLink](/docs/ui-components-component-library-button-link-button-link-stories-js--default-story) component props
+
+<ArgsTable of={ButtonLink} />
+
 ### Type
```
NidhiKJha, line 28: Nit: Update the jsx for `Type`
```suggestion
import { Button, BUTTON_TYPES } from '../ui/component-library';
```

NidhiKJha (commented):
LGTM ! Just some small nits.

ui/components/component-library/button/README.mdx
NidhiKJha: Will need to add `BUTTON_SIZES` and `BUTTON_TYPES`  to the component-library index file

ui/components/component-library/button-link/button-link.scss
```diff
@@ -35,7 +35,7 @@
       opacity: 0.5;
     }
 
-    &.mm-button--disabled:hover {
+    &.mm-button-base--disabled:hover {
```
georgewrmarshall, line 38: If we are adding a --disabled class do we still need to use the base--disabled class?

ui/components/component-library/button-primary/button-primary.js
```diff
@@ -8,16 +8,18 @@ import { BUTTON_PRIMARY_SIZES } from './button-primary.constants';
...
       className={classnames(className, 'mm-button-primary', {
         'mm-button-primary--type-danger': danger,
+        'mm-button-primary--disabed': disabled,
       })}
       size={size}
-      {...props}
+      {...{ disabled, ...props }}
     />
   );
 };
```
georgewrmarshall, line 25: Add proptypes

georgewrmarshall (commented):
I think we should decide whether we are using child/base classes in our parent/variant components or not? I think we should try to avoid it to try to achieve as much separation of concerns where possible? Also we should add propTypes to any new defined prop - all `disabled` props

ui/components/component-library/button-primary/button-primary.scss
```diff
@@ -36,7 +36,7 @@
       background-color: var(--color-error-alternative);
     }
 
-    &.mm-button--disabled:active {
+    &.mm-button-base--disabled:active {
```
georgewrmarshall, line 39: Still using child components class name here. Can this just be `&--disabled:hover {`?

ui/components/component-library/button-secondary/button-secondary.scss
```diff
@@ -44,7 +44,7 @@
       border-color: var(--color-error-alternative);
     }
 
-    &.mm-button--disabled:hover {
+    &.mm-button-base--disabled:hover {
```
georgewrmarshall, line 47: Same here. Can this just be `&--disabled:hover {`?

ui/components/component-library/button-base/button-base.scss
```diff
@@ -1,4 +1,4 @@
-.mm-button {
+.mm-button-base {
   position: relative;
   height: 40px;
   padding: 0;
```
georgewrmarshall, line 4: Could things like padding, border-radius, color, background-color all be `Box` props? Then it can be overwritten by the Box prop in variant components and it saves CSS here and then also CSS that it takes to overwrite this CSS in the variant component.

georgewrmarshall (commented):
Looking good. Couple of suggestions. I think we want to use Box props wherever possible. It makes overwriting the styles easier and saves writing additional CSS?

NidhiKJha (dismissed):

georgewrmarshall (commented):
Tried some explorations in this PR https://github.com/MetaMask/metamask-extension/pull/17104. It tries to use as many Box props as possible while reducing the css as well as focusing the isolating the extended features in the variant component as opposed to the base component. e.g. removing size auto from `ButtonBase` and only allowing it in `ButtonLink`. As well as `href` and `as` stories and docs

georgewrmarshall (commented):
Code looks good! Are we still going to isolate the SIZE.AUTO to ButtonLink in the ButtonLink PR? Just the fix that @NidhiKJha mentioned

georgewrmarshall (approved):

NidhiKJha (approved):

PR #16873, opened 2022-12-08
add new yarn3 paths to circleci cache

brad-decker:
## Explanation
1. Adds the new .yarn/cache folder  and install state to the cache
2. Adds a fallback mechanism for restoring a cache when lockfile contents change, shaving sometime on the install step as it only has to install new deps.

#### Before yarn 3 on the CrowdIn prs that don't touch deps
image

#### After yarn 3
image

The problem here is that node_modules is not the source of truth for yarn anymore, its just the output of the node-modules linker which operates from the lockfile and the .cache. So this change persists the install state and the cache folder. 

## Manual Testing Steps
1. Create a branch from this one.
2. Add, remove or update a dep
3. Commit and push up and look at the CI and see that it persists a new cache key for you based on your lockfile changes
4. Create an empty commit (git commit --allow-empty -m 'hi there') and push it up and see that the install step is lightning fast.

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -208,7 +208,16 @@ jobs:
     steps:
       - checkout
       - restore_cache:
-          key: dependency-cache-v1-{{ checksum "yarn.lock" }}
+          keys:
+            # First try to get the specific cache for the checksum of the yarn.lock file.
+            # This cache key lookup will fail if the lock file is modified and a cache
+            # has not yet been persisted for the new checksum.
+            - dependency-cache-v1-{{ checksum "yarn.lock" }}
+            # To prevent having to do a full install of every node_module when
+            # dependencies change, restore from the last known cache of any
+            # branch/checksum, the install step will remove cached items that are no longer
+            # required and add the new dependencies, and the cache will be persisted.
+            - dependency-cache-v1-
       - gh/install
       - run:
           name: Set IS_DRAFT environment variable
@@ -237,7 +246,7 @@ jobs:
       - save_cache:
           key: dependency-cache-v1-{{ checksum "yarn.lock" }}
           paths:
-            - node_modules/
+            - .yarn/cache
       - persist_to_workspace:
           root: .
           paths:


brad-decker:
Note that I opted *not* to increment the cache key version and instead just wait for the next dependency change to start the lifecycle of the new cached items. Until someone changes the dependencies it'll not persist the new cache items because the cache for that key already exists.

.circleci/config.yml
```diff
@@ -238,6 +247,8 @@ jobs:
           key: dependency-cache-v1-{{ checksum "yarn.lock" }}
           paths:
             - node_modules/
```
Gudahtt, line 249: Might be better just to cache the `cache` directory, not `node_modules`. This really bloats the size, and makes this cache no longer safe to share between Node.js versions because of differences in postinstall steps for certain dependencies.

.circleci/config.yml
```diff
@@ -238,6 +247,8 @@ jobs:
...
+            - .yarn/cache/
+            - .yarn/install-state.gz
```
Gudahtt, line 251: Hmm, we have never cached this in our library workflows. What is this for?

.circleci/config.yml
brad-decker: Ill remove it. I tried both. The link step takes 20s longer without the node_modules cached but thats not too bad.

.circleci/config.yml
brad-decker: > .yarn/install-state.gz is an optimization file that you shouldn't ever have to commit. It simply stores the exact state of your project so that the next commands can boot without having to resolve your workspaces all over again

Idk if that means it helps at all in our situation but maybe it *would* help in a monorepo?

Gudahtt: Ah, hmm. I wonder what it's storing the state _of_. This might cause problems if we drop `node_modules` from the cache.

brad-decker: Ill explore if it makes any impact on CI times and drop it if it does not

.circleci/config.yml
```diff
@@ -238,6 +247,8 @@ jobs:
           key: dependency-cache-v1-{{ checksum "yarn.lock" }}
           paths:
             - node_modules/
+            - .yarn/cache/
```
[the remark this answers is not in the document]
brad-decker: I couldn't quite figure out how to use a dynamic value in the paths key for the save_cache step. I tried setting an environment variable in a previous step and then output it using {{ Environment.YARN_CACHE_FOLDER }} but it didn't work (it didn't do the template replacement that it does with other string values). I also think we have another small issue which is that `yarn config get cacheFolder` prints the absolute path not relative path, and circle prepends whatever is in the paths key with the circleci specific folder structure to the project. So we might end up with an issue if we attempted to do this. 

I found this: https://circleci.canny.io/cloud-feature-requests/p/support-wildcards-in-savecachepaths which leads me to believe that this feature isn't supported yet. If you think i'm interpreting that incorrectly of have advice on how i can achieve this I will take a stab, but for now I reverted to just using the `.yarn/cache` directory.

Gudahtt (approved):
LGTM!

jpuri (approved):

Issue #972, opened 2017-01-06
Plugin prevents sonarqube from working


bobkilla:
I work as a software developper and I installed the plugin on my computer and I went to our sonar server. Most of the pages are blank when the plugin is activated. I didn't investigate at all for the moment, I will try to do that later.

Message on the console :

`Uncaught (in promise) TypeError: key is not a function
    at t.componentDidMount (main.js?v=6.1:8)
    at r.notifyAll (vendor.js?v=6.1:12)
    at r.close (vendor.js?v=6.1:19)
    at r.closeAll (vendor.js?v=6.1:8)
    at r.perform (vendor.js?v=6.1:8)
    at s (vendor.js?v=6.1:12)
    at r.perform (vendor.js?v=6.1:8)
    at Object.batchedUpdates (vendor.js?v=6.1:19)
    at Object.o [as batchedUpdates] (vendor.js?v=6.1:7)
    at Object._renderNewRootComponent (vendor.js?v=6.1:12)`

flyswatter:
That log's stack is just generic React, we need more info to reproduce.

There are a few other places you can look for good logs, please check if you find any other errors:

https://github.com/MetaMask/faq/blob/master/LOGS.md

kumavis:
I just tested the public hosted version of sonarqube and found no issue. I'm closing this issue until we have more information to begin debugging.

PR #16881, opened 2022-12-08
Make `initializeProvider` and `lookupNetwork` async

Gudahtt:
These two methods were "synchronous" previously, but initiated an asynchronous operation. They have both been made `async` to bring them more in-line with the mobile controller API, and to make them easier to test.

This relates to https://github.com/MetaMask/controllers/issues/971

## Manual Testing Steps

This should include zero functional changes. These methods are still being invoked without an `await`, to preserve the same behaviour they had previously.

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/app/scripts/controllers/network/network.js
+++ b/app/scripts/controllers/network/network.js
@@ -203,7 +203,7 @@ export default class NetworkController extends EventEmitter {
 
       this._setNetworkState('loading');
       // keep network details in sync with network state
-      this.clearNetworkDetails();
+      this._clearNetworkDetails();
       return;
     }
 
--- a/app/scripts/controllers/network/network.js
+++ b/app/scripts/controllers/network/network.js
@@ -122,11 +122,11 @@ export default class NetworkController extends EventEmitter {
     this.on(NETWORK_EVENTS.NETWORK_DID_CHANGE, this.lookupNetwork);
   }
 
-  initializeProvider(providerParams) {
+  async initializeProvider(providerParams) {
     this._baseProviderParams = providerParams;
     const { type, rpcUrl, chainId } = this.getProviderConfig();
     this._configureProvider({ type, rpcUrl, chainId });
-    this.lookupNetwork();
+    await this.lookupNetwork();
   }
 
   // return the proxies so the references will always be good
@@ -162,7 +162,7 @@ export default class NetworkController extends EventEmitter {
     return this.getNetworkState() === 'loading';
   }
 
-  lookupNetwork() {
+  async lookupNetwork() {
     // Prevent firing when provider is not defined.
     if (!this._provider) {
       log.warn(
@@ -194,21 +194,35 @@ export default class NetworkController extends EventEmitter {
       this.emit(NETWORK_EVENTS.INFURA_IS_UNBLOCKED);
     }
 
-    ethQuery.sendAsync({ method: 'net_version' }, (err, networkVersion) => {
-      const currentNetwork = this.getNetworkState();
-      if (initialNetwork === currentNetwork) {
-        if (err) {
-          this._setNetworkState('loading');
-          // keep network details in sync with network state
-          this._clearNetworkDetails();
-          return;
-        }
-
-        this._setNetworkState(networkVersion);
-        // look up EIP-1559 support
-        this.getEIP1559Compatibility();
+    let networkVersion;
+    try {
+      networkVersion = await new Promise((resolve, reject) => {
+        ethQuery.sendAsync({ method: 'net_version' }, (error, result) => {
+          if (error) {
+            reject(error);
+          } else {
+            resolve(result);
+          }
+        });
+      });
+    } catch (error) {
+      if (initialNetwork !== this.getNetworkState()) {
+        return;
       }
-    });
+
+      this._setNetworkState('loading');
+      // keep network details in sync with network state
+      this.clearNetworkDetails();
+      return;
+    }
+
+    if (initialNetwork !== this.getNetworkState()) {
+      return;
+    }
+
+    this._setNetworkState(networkVersion);
+    // look up EIP-1559 support
+    await this.getEIP1559Compatibility();
   }
 
   getCurrentChainId() {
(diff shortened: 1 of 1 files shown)


brad-decker (dismissed):
Changes make sense to me.

app/scripts/controllers/network/network.js
```diff
@@ -124,11 +124,11 @@ export default class NetworkController extends EventEmitter {
     this._infuraProjectId = projectId;
   }
 
-  initializeProvider(providerParams) {
+  async initializeProvider(providerParams) {
```
mcmire, line 127: I notice that unit tests are missing for this exact method. `initializeProvider` is called for other tests, but without taking async into consideration. Same goes for `lookupNetwork`.

This should be captured by #972, but it sounds like we wanted to add unit tests _after_ changing this controller. Should that happen before instead so that we can make sure we haven't broken anything?

app/scripts/controllers/network/network.js
```diff
@@ -249,21 +249,35 @@ export default class NetworkController extends EventEmitter {
...
+      return;
+    }
+
+    if (initialNetwork !== this.getNetworkState()) {
+      return;
+    }
+
+    this.setNetworkState(networkVersion);
+    // look up EIP-1559 support
+    await this.getEIP1559Compatibility();
```
mcmire, line 280: Hmm, this is a bit less straightforward than the previous code. What if you added this at the top of the file:

``` javascript
function getNetworkVersion(ethQuery) {
  return ethQuery.sendAsync({ method: 'net_version' }, (error, result) => {
    if (error) {
      reject(error);
    } else {
      resolve(result);
    }
  });
}
```

And then changed this code to:

``` suggestion
    let networkVersion;
    let error;
    try {
      networkVersion = await getNetworkVersion(ethQuery);
    } catch (e) {
      console.error(e);
      error = e;
    }

    if (initialNetwork === this.getNetworkState()) {
      if (error) {
        this.setNetworkState('loading');
        // keep network details in sync with network state
        this.clearNetworkDetails();
        return;
      }

      this.setNetworkState(networkVersion);
      // look up EIP-1559 support
      await this.getEIP1559Compatibility();
    }
```

Gudahtt: Thanks, great suggestion, I will do that

app/scripts/controllers/network/network.js
Gudahtt: Hmm. Yeah, fair enough. Tests first would help validate the work. I'll investigate how difficult testing this controller more effectively would be.

app/scripts/controllers/network/network.js
Gudahtt: Done in https://github.com/MetaMask/metamask-extension/pull/16881/commits/7a1ebe1ace7237f1fd3fd949fe6bd9768fed2de5

app/scripts/controllers/network/network.js
Gudahtt: In this case I don't think it's worth the effort to add unit tests prior to making the change, because the change itself is primarily intended to make unit testing easier. Especially given that this code is well covered by e2e tests, writing tests _after_ this PR seems pragmatic to me.

I have added a new issue for tracking unit tests in the extension: https://github.com/MetaMask/metamask-extension/issues/16962
We can make a point to consider this for subsequent network controller changes as well. Unless a change is specifically meant to improve testability, we probably should ensure all future changes to the network controller are covered by unit tests.

mcmire: Okay sounds good!

app/scripts/controllers/network/network.js
mcmire: We still have the repeated `if (initialNetwork !== this.getNetworkState())` here but I guess that's fine, it only happens twice.

mcmire: Er, actually, would you mind simplifying this a bit. I think it could be confusing for someone glancing at this due to the three `return`s. It is easy to think think that we are calling `if (initialNetwork !== this.getNetworkState())` when in fact we are not.

Gudahtt: Do you have any suggestions on how this could be simplified?

In either the success or error case, we want to bail immediately if a network switch has happened. I'm not sure of a simpler way to do that. I also don't follow what you mean by that last sentence. There are only two branches, and we make that check in both.

mcmire: Sure, I'll make a new comment so it's more visible.

app/scripts/controllers/network/network.js
```diff
@@ -194,21 +193,27 @@ export default class NetworkController extends EventEmitter {
...
+      return;
+    }
+
+    if (initialNetwork !== this.getNetworkState()) {
+      return;
+    }
+
+    this._setNetworkState(networkVersion);
+    // look up EIP-1559 support
+    await this.getEIP1559Compatibility();
```
mcmire, line 216: Did this suggestion make sense? I believe this would be more straightforward to read:

``` suggestion
    let networkVersion;
    let error;
    try {
      networkVersion = await getNetworkVersion(ethQuery);
    } catch (e) {
      console.error(e);
      error = e;
    }
    if (initialNetwork === this.getNetworkState()) {
      if (error) {
        this.setNetworkState('loading');
        // keep network details in sync with network state
        this.clearNetworkDetails();
        return;
      }
      this.setNetworkState(networkVersion);
      // look up EIP-1559 support
      await this.getEIP1559Compatibility();
    }
```

Gudahtt: OK, I'll try something along these lines

Gudahtt: Attempt made here: https://github.com/MetaMask/metamask-extension/pull/16881/commits/1d57a4e90816b00b4db09d687ce9ed58e5ab8cd7

I kept the early-exit to avoid unnecessary indentation, and used `else` over `return` to simplify the control flow. And I didn't add any logging, as I'm trying to avoid additional functional changes. But it does use the error capture strategy you used here.

mcmire: Okay no problem. That looks better. Thanks!

mcmire (approved):

brad-decker (approved):

PR #16884, opened 2022-12-09
Move `infuraProjectId` to network controller constructor

Gudahtt:
The network controller `setInfuraProjectId` method has been deleted. The Infura project ID is only ever set upon construction, so it is now passed in as a constructor parameter instead.

Rather than adding this as a second parameter, the network controller now uses an "options bag" for constructor parameters. The initial state was the first parameter, but it's now passed in as the `state` option instead.

These changes make the API more similar to the mobile network controller API.

This relates to https://github.com/MetaMask/controllers/issues/971

## Manual Testing Steps

This should have zero functional changes.

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/app/scripts/controllers/detect-tokens.test.js
+++ b/app/scripts/controllers/detect-tokens.test.js
@@ -33,8 +33,7 @@ describe('DetectTokensController', function () {
 
   beforeEach(async function () {
     keyringMemStore = new ObservableStore({ isUnlocked: false });
-    network = new NetworkController();
-    network.setInfuraProjectId('foo');
+    network = new NetworkController({ infuraProjectId: 'foo' });
     network.initializeProvider(networkControllerProviderConfig);
     provider = network.getProviderAndBlockTracker().provider;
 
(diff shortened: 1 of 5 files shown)


FrederikBolding (approved):

legobeat:
So this means that changing `infuraProjectId` should mean replacing the `NetworkController` with a new instance, correct?

We still have mutating methods for `setNetworkEIPSupport`, `setRpcTarget`, `setProviderType`, `setProviderConfig`.. It seems to me that at least some of those (rpcurl) belong on the same level as a provider key?

Is the intention to do further refactoring to make these immutable in a similar way as well? It's possible I missed it but did not see that addressed in https://github.com/MetaMask/controllers/issues/971

Gudahtt:
Changing `infuraProjectId` was never a thing that was supported; that key gets embedded by our build system. You'd need to rebuild to change it. Custom Infura endpoints are supported as "custom RPC endpoints", not by allowing our internal endpoints to be updated.

There aren't any plans to make things that are mutable today immutable; the goal is to merge the two controllers with minimal changes/disruption.

legobeat (approved):

brad-decker (approved):

darkwing (approved):

Issue #16880, opened 2022-12-08
[FLASK] De-flake snaps E2E tests


bowensanders:
Improvements to the snaps E2E tests need to be made to stop occasional CI failures.

anasaboayham:
hello

SUPPORTTEAMHERR:
> Improvements to the snaps E2E tests need to be made to stop occasional CI failures.

We'd be happy to look into this for you. kindly file a support ticket below 
https://forms.gle/Wmpm6NLzYm9uXYuU9

PR #16894, opened 2022-12-09
[FLASK] De-flake snaps E2E tests

bowensanders:
Making small improvements in the snaps E2E tests to decrease the number of CI failures.

### DIFF
--- a/test/e2e/snaps/test-snap-managestate.spec.js
+++ b/test/e2e/snaps/test-snap-managestate.spec.js
@@ -30,7 +30,7 @@ describe('Test Snap manageState', function () {
         await driver.press('#password', driver.Key.ENTER);
 
         // navigate to test snaps page, then fill in the snapId
-        await driver.openNewPage(TEST_SNAPS_WEBSITE_URL);
+        await driver.driver.get(TEST_SNAPS_WEBSITE_URL);
         await driver.delay(1000);
 
         // find and scroll to the connect button and click it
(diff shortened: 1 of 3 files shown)


bowensanders:
https://github.com/MetaMask/metamask-extension/issues/16880

SUPPORTTEAMHERR:
> This addresses issue https://github.com/MetaMask/metamask-extension/issues/16880

test/e2e/snaps/test-snap-managestate.spec.js
```diff
@@ -70,9 +70,14 @@ describe('Test Snap manageState', function () {
         });
 
         // fill and click send inputs on test snap page
-        windowHandles = await driver.waitUntilXWindowHandles(2, 1000, 10000);
+        windowHandles = await driver.waitUntilXWindowHandles(1, 1000, 10000);
         await driver.switchToWindowWithTitle('Test Snaps', windowHandles);
+        await driver.delay(1000);
```
legobeat, line 75: Were lower numbers tried and where does the number 1000 come from? 1s is relatively high (they add up) and if lower numbers still reliably pass it would be great (the same applies to other delays in these suites).

Either way some seconds increase in delay is still better than false negatives in CI so can be addressed separately.

legobeat (approved):

jpuri (approved):
LGTM

PR #16876, opened 2022-12-08
Add back error when password is too short

adonesky1:
In Onboarding V2 we do not show an error indicating why the user cannot proceed if they have entered a password shorter than 8 characters - we do show this error in Onboarding V1:
205687681-5f0a65fc-7ef3-4596-bc0b-028547d88413

* Fixes #16802

### Before

https://user-images.githubusercontent.com/34557516/206565268-705eee14-283d-4f75-aa2e-dabee309209d.mp4

### After

https://user-images.githubusercontent.com/34557516/206566750-d8c65a51-5c5a-4e03-8793-a360672ed0e6.mov

## Manual Testing Steps

1. Load MM for the first time
2. Go through the onboarding process for Create a New Wallet
3. Enter a password shorter than 8 characters
4. You should see an error - "Password is not long enough" - indicating why you cannot proceed.

## Pre-merge author checklist

- [x] I've clearly explained:
  - [x] What problem this PR is solving
  - [x] How this problem was solved
  - [x] How reviewers can test my changes

PR #16902, opened 2022-12-10
Refactor password validation

legobeat:
Targeting #16876 

* introduces constant `PASSWORD_MIN_LENGTH` replacing magic numbers
* uses password strength control for length validation error
* simplifies logic - all local variables are now `const`

Manually tested in Chromium with `ONBOARDING_V2=1`.

### DIFF
--- a/ui/pages/onboarding-flow/create-password/create-password.js
+++ b/ui/pages/onboarding-flow/create-password/create-password.js
@@ -26,6 +26,7 @@ import {
   TwoStepProgressBar,
   twoStepStages,
 } from '../../../components/app/step-progress-bar';
+import { PASSWORD_MIN_LENGTH } from '../../../helpers/constants/common';
 import ZENDESK_URLS from '../../../helpers/constants/zendesk-url';
 import { getFirstTimeFlowType } from '../../../selectors';
 import { FIRST_TIME_FLOW_TYPES } from '../../../helpers/constants/onboarding';
@@ -55,7 +56,7 @@ export default function CreatePassword({
       return false;
     }
 
-    if (password.length < 8) {
+    if (password.length < PASSWORD_MIN_LENGTH) {
       return false;
     }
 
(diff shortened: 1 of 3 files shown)


ui/pages/onboarding-flow/create-password/create-password.js
```diff
@@ -55,44 +56,57 @@ export default function CreatePassword({
...
         text: translation('strong'),
         description: '',
       };
-    } else if (score === 3) {
+    }
+    if (score === 3) {
       return {
         className: 'create-password__average',
         text: translation('average'),
-        description: t('passwordStrengthDescription'),
```
legobeat, line 76: I assume that mixing references to `t` and `translation` here was not intentional.

VSaric: @legobeat It is not intentional for this references. I must forgot to add `translation` instead of `t`. Good catch! 😄

adonesky1: wait what is is translation? `t` is coming from `useI18nContext()` on line 40, which we use to hook into our translation files? @legobeat

legobeat: @adonesky1 They are&were identical here so this is a non-functional change.

https://github.com/MetaMask/metamask-extension/pull/16902/files#diff-a161d5f7ee5a8ded49415fd70e45e18fa4f24dafd1e49375d77365b6cb5c84adR102

darkwing: I believe our `yarn verify-locales` routine relies on the use of `t()`, so I think it makes sense to keep `t()`.  Also for consistency.

legobeat:
Base changed to `develop` after merge of #16876 .

ui/pages/onboarding-flow/create-password/create-password.js
legobeat: This is now done in afc438ed8e3ff24434c39fb9123e1eccc2c2e3f9

jpuri (approved):

adonesky1 (approved):
Tested. LGTM! Thanks for the cleanup Lego!

PR #16922, opened 2022-12-13
Refactor how network state is passed to transaction controller

Gudahtt:
The network state is now passed to the TransactionController via a getter function and a subscription function, instead of passing one of the network controller stores directly.

This way of passing the state makes further refactoring easier, as we don't have to change the input when the store is changed or replaced. It's also more aligned with our conventions today.

This change was made as part of https://github.com/MetaMask/controllers/issues/971.

## Manual Testing Steps

There should be zero functional changes.

## Pre-merge author checklist

- [x] I've clearly explained:


brad-decker (approved):
LGTM

app/scripts/controllers/transactions/tx-state-manager.js
```diff
@@ -81,7 +86,7 @@ export default class TransactionStateManager extends EventEmitter {
    * @returns {TransactionMeta} the default txMeta object
    */
   generateTxMeta(opts = {}) {
-    const netId = this.getNetwork();
+    const netId = this.getNetworkState();
```
mcmire, line 89: Side note, we should make these variables consistent at some point too. It's confusing that sometimes this is `netId`, sometimes, `network`, sometimes `networkState`.

mcmire (approved):
LGTM.

app/scripts/controllers/transactions/tx-state-manager.js
Gudahtt: The value itself is confusing, even apart from the name. It's a mix of "loading" and network ID. We use it in some places _calling it_ "network ID", but using it just for the "loading" part.

Clarifying this is on the API normalization list.

PR #16935, opened 2022-12-13
Update an E2E test with a mock

dan437:
## Explanation
- Update an E2E test with a mock to keep it more stable

### DIFF
--- a/test/e2e/swaps/swaps-notifications.spec.js
+++ b/test/e2e/swaps/swaps-notifications.spec.js
@@ -5,10 +5,58 @@ const { withFixtures } = require('../helpers');
 const { withFixturesOptions, loadSwaps, buildQuote } = require('./shared');
 
 describe('Swaps - notifications', function () {
+  async function mockTradesApiPriceSlippageError(mockServer) {
+    await mockServer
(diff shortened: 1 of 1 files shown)


pedronfigueiredo (approved):

FrederikBolding (approved):

Gudahtt (approved):
LGTM, thanks!

test/e2e/swaps/swaps-notifications.spec.js
```diff
@@ -5,10 +5,58 @@ const { withFixtures } = require('../helpers');
 const { withFixturesOptions, loadSwaps, buildQuote } = require('./shared');
 
 describe('Swaps - notifications', function () {
+  async function mockTradesApiPriceSlippageError(mockServer) {
```
Gudahtt, line 8: Would this mock be useful for other swaps tests as well? Surely at least some of the others must hit the `/trades` endpoint as well

Gudahtt: Oh, I see, this is the only one that uses this specific trade. So this mock is only applicable here.

PR #16937, opened 2022-12-13
Add e2e debugging logs

Gudahtt:
A `debug` flag has been added to our e2e test runner scripts, enabling e2e debug logs. When this flag is enabled, all driver interactions will be logged to the console. This is extremely useful when debugging e2e tests, because it lets you known how far the test had progressed before failing.

This flag should work with all existing e2e test scripts, including both `test:e2e:single` and all of the test commands that run entire test suites.

The README has been updated to reference this flag in the section regarding how to run a single e2e test. To ensure this wasn't totally missed for the other scripts, I added a line suggesting that users use `--help` to see all supported options.

## Manual Testing Steps

* Create a test build (`yarn start:test`, then interrupt the process when the "primary" bundle has finished)
* Run a test or a test suite with the `--debug` flag (e.g. `yarn test:e2e:single --browser firefox --debug ./test/e2e/swaps/swaps-notifications.spec.js`)
* See that driver logs are printed to the console

Here is an example of the logs you should see for the first swaps notification test, from the example command listed above:
```
Swaps - notifications
[driver] Called 'navigate' with arguments []
[driver] Called 'fill' with arguments ["#password","correct horse battery staple"]
[driver] Called 'press' with arguments ["#password",""]
[driver] Called 'clickElement' with arguments [".wallet-overview__buttons .icon-button:nth-child(3)"]
[driver] Called 'fill' with arguments ["input[placeholder*=\"0\"]",2]
[driver] Called 'delay' with arguments [1600]
[driver] Called 'clickElement' with arguments ["[class=\"dropdown-search-list__closed-primary-label dropdown-search-list__select-default\"]"]
[driver] Called 'wait' with arguments [null]
[driver] Called 'findElements' with arguments [".searchable-item-list__item"]
[driver] Called 'clickElement' with arguments [".searchable-item-list__labels"]'
[driver] Called 'clickElement' with arguments [".dropdown-input-pair__to"]
[driver] Called 'clickElement' with arguments ["input[data-testid=\"search-list-items\"]"]
[driver] Called 'fill' with arguments ["input[data-testid=\"search-list-items\"]","INUINU"]
[driver] Called 'wait' with arguments [null]
[driver] Called 'findElements' with arguments [".searchable-item-list__primary-label"]
[driver] Called 'findElements' with arguments [".searchable-item-list__primary-label"]
[driver] Called 'findElements' with arguments [".searchable-item-list__primary-label"]
[...]

PR #16939, opened 2022-12-13
Improve stability of e2e test snap approval step

Gudahtt:
The snap approval step of the snaps e2e tests is prone to failure because of how the popup behaves. The popup rapidly closes after the connect step, then re-appears for the snap approval. There is no reliable way to know when the first popup has closed and the second one has opened.

Instead of trying to guess when the second popup would be open by introducing a delay, we can now use the new utility function `retryOnClosed` to retry the step of switching to a window and making the first interaction with it. All snap tests that include an approval step have been updated to use this utility.

## Manual Testing Steps

If you run snaps e2e tests long enough locally, you should see a failure. I was seeing it about once every 10-20 runs.

If you run the tests on this branch using the `--debug` flag from https://github.com/MetaMask/metamask-extension/pull/16937 , you should instead see a _third_ switch window command being used where usually there is only two. This might be hard to reproduce though, again only happening every 10-20 runs for me locally (system-dependent, so it may be easier or harder to reproduce on your machine).

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/test/e2e/snaps/test-snap-bip-32.spec.js
+++ b/test/e2e/snaps/test-snap-bip-32.spec.js
@@ -1,5 +1,5 @@
 const { strict: assert } = require('assert');
-const { withFixtures } = require('../helpers');
+const { retryOnClosed, withFixtures } = require('../helpers');
 const FixtureBuilder = require('../fixture-builder');
 const { TEST_SNAPS_WEBSITE_URL } = require('./enums');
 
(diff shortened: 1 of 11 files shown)


test/e2e/snaps/test-snap-bip-32.spec.js
```diff
@@ -55,19 +55,17 @@ describe('Test Snap bip-32', function () {
...
         // approve install of snap
-        await driver.switchToWindowWithTitle(
-          'MetaMask Notification',
-          windowHandles,
-        );
-        await driver.clickElement({
-          text: 'Approve & install',
-          tag: 'button',
+        await retryOnClosed(async () => {
+          windowHandles = await driver.getAllWindowHandles();
```
Gudahtt, line 60: I need to update these all to use `waitUntilXWindowHandles` instead

Gudahtt:
It seems this doesn't work well on Firefox.

It looks like the "browsing context closed" error is being thrown on this line:
```
await driver.clickElement({
  text: 'Approve & install',
  tag: 'button',
});
```
But it does actually successful click the button, despite throwing the expected error. The failure must happen after the click is registered.

When retrying, the button was already clicked so it fails.

Gudahtt:
I will close this for now because I don't have time to continue investigating this

PR #16898, opened 2022-12-09
fix warning on revoke approval for all nfts

EresDev:
Fixes #16846
## Explanation
I have removed the warning modal for revoking access to all NFTs. The "Confirm" action button performs the revoking call without any modal warning now.

For reference, initially I have  tried keeping the warning modal with appropriate warning message on revoking access to all NFTs. But it was later suggested in https://github.com/MetaMask/metamask-extension/pull/16898#issuecomment-1346441605 to not have modal warning on revokation. 

## Manual Testing Steps
Manual testing was performed using the test-dapp.
The following actions were tested for collectibles.
- Deploy
- Mint
- Approve
- Set Approval for All
- Revoke

PR #16953, opened 2022-12-14
remove warning on revoke approval for all nfts

### LINKED ISSUE #16846: [Bug]: Revoking Approval displays a warning as if we were granting Access
### Describe the bug

**Problem**: when we try to Revoke permissions from a token, we see a warning as if we were granting access to all our tokens.

https://user-images.githubusercontent.com/54408225/206227435-44882c3a-d4c3-4ec0-b54f-beeebdb798e8.mp4

### Steps to reproduce

1. Go to Test Dapp
2. Connect to Test Dapp
3. Click Deploy Collectible
4. Click Mint collectible
5. Click Revoke
6. Accept MM prompt -- see how a warning for Approval is displayed

_No response_

### Version

10.24.0

### Build type

None

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

danjm:
This PR rebases https://github.com/MetaMask/metamask-extension/pull/16898 onto `develop`

Fixes https://github.com/MetaMask/metamask-extension/issues/16846

Explanation
I have removed the warning modal for revoking access to all NFTs. The "Confirm" action button performs the revoking call without any modal warning now.

For reference, initially I have tried keeping the warning modal with appropriate warning message on revoking access to all NFTs. But it was later suggested in https://github.com/MetaMask/metamask-extension/pull/16898#issuecomment-1346441605 to not have modal warning on revokation.

Manual Testing Steps
Manual testing was performed using the test-dapp.
The following actions were tested for collectibles.

Deploy
Mint
Approve
Set Approval for All
Revoke

### DIFF
--- a/ui/components/app/confirm-page-container/confirm-page-container.component.js
+++ b/ui/components/app/confirm-page-container/confirm-page-container.component.js
@@ -108,6 +108,7 @@ export default class ConfirmPageContainer extends Component {
     supportsEIP1559: PropTypes.bool,
     nativeCurrency: PropTypes.string,
     isBuyableChain: PropTypes.bool,
+    isApprovalOrRejection: PropTypes.bool,
   };
 
   async componentDidMount() {
@@ -184,6 +185,7 @@ export default class ConfirmPageContainer extends Component {
       ///: END:ONLY_INCLUDE_IN
       accountBalance,
       assetStandard,
+      isApprovalOrRejection,
     } = this.props;
 
     const shouldDisplayWarning =
@@ -351,7 +353,11 @@ export default class ConfirmPageContainer extends Component {
             <PageContainerFooter
               onCancel={onCancel}
               cancelText={t('reject')}
-              onSubmit={isSetApproveForAll ? onSetApprovalForAll : onSubmit}
+              onSubmit={
+                isSetApproveForAll && isApprovalOrRejection
+                  ? onSetApprovalForAll
+                  : onSubmit
+              }
               submitText={t('confirm')}
               submitButtonType={
                 isSetApproveForAll ? 'danger-primary' : 'primary'
--- a/test/e2e/tests/collectibles.spec.js
+++ b/test/e2e/tests/collectibles.spec.js
@@ -260,7 +260,6 @@ describe('Collectibles', function () {
 
         // Confirm disabling set approval for all
         await driver.clickElement({ text: 'Confirm', tag: 'button' });
-        await driver.clickElement({ text: 'Approve', tag: 'button' });
 
         await driver.waitUntilXWindowHandles(2);
         await driver.switchToWindow(extension);


ui/components/app/confirm-page-container/confirm-page-container.component.js
```diff
@@ -351,7 +353,11 @@ export default class ConfirmPageContainer extends Component {
             <PageContainerFooter
               onCancel={onCancel}
               cancelText={t('reject')}
-              onSubmit={isSetApproveForAll ? onSetApprovalForAll : onSubmit}
+              onSubmit={
+                isSetApproveForAll && isApprovalOrRejection
+                  ? onSetApprovalForAll
+                  : onSubmit
```
digiwand, line 359: The PR works as intended based on https://github.com/MetaMask/metamask-extension/pull/16898#issuecomment-1346441605. That is, the revoke warning is no longer shown. 

I think we need to update the description

EresDev: The description? You mean the explanation of the pull request?   @digiwand

digiwand: Hi @EresDev,

yeah. It seems to me
> I have updated the warning message for revoking approval for all NFTs along with the action button (Approve, Reject) titles to clarify the intent.

is now outdated. It seems we have removed the revoke approval warning altogether if I'm not mistaken

EresDev: Ok @digiwand. I have updated the explanation description in https://github.com/MetaMask/metamask-extension/pull/16898. I think @danjm can use that to update the explanation of this PR as well.

EresDev (approved):

ui/components/app/confirm-page-container/confirm-page-container.component.js
```diff
@@ -351,7 +353,11 @@ export default class ConfirmPageContainer extends Component {
             <PageContainerFooter
               onCancel={onCancel}
               cancelText={t('reject')}
-              onSubmit={isSetApproveForAll ? onSetApprovalForAll : onSubmit}
+              onSubmit={
+                isSetApproveForAll && isApprovalOrRejection
```
danjm, line 357: For code reviewers:

`isApprovalOrRejection` is a variable that pre-dated this PR, but it is poorly named. It will be true for a`setApprovalForAll` transactions that, if approved, grant access to all NFTs in a collection. And it will be false for a `setApprovalForAll` transaction that revokes access.

In the former case, the transaction data looks like this:
```
Function: SetApprovalForAll
Parameters: true
Data: 0xa22cb4650000000000000000000000009bc5baf874d2da8d216ae9f137804184ee5afef40000000000000000000000000000000000000000000000000000000000000001
```

And in the latter:
```
Function: SetApprovalForAll
Parameters: false
0xa22cb4650000000000000000000000009bc5baf874d2da8d216ae9f137804184ee5afef40000000000000000000000000000000000000000000000000000000000000000
```

So this variable depends upon whether the approval is encoded within the last hexadecimal of the data. This is deciphered in the `helpers/utils/token-util` `getTokenApprovedParam` method which is called in `confirm-approve` to create the `isSetApproveForAll` variable, which is then passed down, eventually to this component

The goal of this PR was to ensure that the `isSetApproveForAll` warning popup is not shown when revoking. And so the solution is to use `isApprovalOrRejection` as part of the decision as to which method should be called upon clicking the confirmation button on the set approval for all screen. If `isApprovalOrRejection` is true, then the method that shows the popup is called, if it is false, then that method is not called, and instead the transaction is just approved/submitted.

digiwand (approved):

jpuri (approved):
👍

PR #16956, opened 2022-12-14
[FLASK] Fix race condition with transaction insights

FrederikBolding:
## Explanation

Fixes a race condition that would occur when a user had more than one transaction insight snap installed. If the transaction insight tab is loading data from the selected snap while a new one is selected a race condition will occur in determining which of the results to show.

### DIFF
--- a/ui/hooks/flask/useTransactionInsightSnap.js
+++ b/ui/hooks/flask/useTransactionInsightSnap.js
@@ -28,6 +28,7 @@ export function useTransactionInsightSnap({
   const [error, setError] = useState(undefined);
 
   useEffect(() => {
+    let cancelled = false;
     async function fetchInsight() {
       try {
         setError(undefined);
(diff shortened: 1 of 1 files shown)


GuillaumeRx (dismissed):

ui/hooks/flask/useTransactionInsightSnap.js
```diff
@@ -43,16 +44,23 @@ export function useTransactionInsightSnap({
             params: { transaction, chainId, transactionOrigin },
           },
         });
-        setData(d);
+        if (!cancelled) {
+          setData(d);
```
Mrtenz, line 48: While not necessarily an issue with this PR, using single letter variable names like `d` won't work anymore with our latest ESLint config. Can you just name it `data` or something?

Mrtenz (changes requested):

ui/hooks/flask/useTransactionInsightSnap.js
FrederikBolding: Done

Mrtenz (approved):

GuillaumeRx (approved):

PR #16973, opened 2022-12-15
[FLASK] Use custom UI for dialogs

### LINKED ISSUE #16522: Use Snaps Custom UI renderer to render dialogs with custom UI
Ref: #16521

`snap_dialog` confirmations submitted with custom UI should be rendered in the extension.

GuillaumeRx:
Use custom UI for `snap_dialog`, removing the old dialog format.

Fixes #16522

### DIFF
--- a/ui/components/app/flask/snap-ui-renderer/snap-ui-renderer.js
+++ b/ui/components/app/flask/snap-ui-renderer/snap-ui-renderer.js
@@ -58,7 +58,7 @@ export const UI_MAPPING = {
   }),
 };
 
-const mapToTemplate = (data) => {
+export const mapToTemplate = (data) => {
   const { type } = data;
   const mapped = UI_MAPPING[type](data);
   // TODO: We may want to have deterministic keys at some point
--- a/ui/components/app/flask/snap-ui-renderer/snap-ui-renderer.js
+++ b/ui/components/app/flask/snap-ui-renderer/snap-ui-renderer.js
@@ -58,6 +58,7 @@ export const UI_MAPPING = {
   }),
 };
 
+// TODO: Stop exporting this when we remove the mapToTemplate hack in confirmation templates.
 export const mapToTemplate = (data) => {
   const { type } = data;
   const mapped = UI_MAPPING[type](data);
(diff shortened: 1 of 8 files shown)


ui/pages/confirmation/templates/flask/snap-confirmation/snap-confirmation.js
```diff
@@ -19,46 +18,7 @@ function getValues(pendingApproval, t, actions) {
...
-                    element: 'Copyable',
-                    key: 'snap-dialog-content-text',
-                    props: {
-                      text: textAreaContent,
-                    },
-                  },
-                ]
-              : []),
-          ],
+          children: mapToTemplate(content),
```
FrederikBolding, line 21: Does this need to be an array?

ui/pages/confirmation/templates/flask/snap-alert/snap-alert.js
```diff
@@ -19,46 +21,7 @@ function getValues(pendingApproval, t, actions) {
...
-                    element: 'Copyable',
-                    key: 'snap-dialog-content-text',
-                    props: {
-                      text: textAreaContent,
-                    },
-                  },
-                ]
-              : []),
-          ],
+          children: mapToTemplate(content),
```
FrederikBolding, line 24: Does this need to be an array?

FrederikBolding: Please leave a comment explaining that we should swap to use `SnapUIRenderer` at some point instead of doing this hack. (applies to all confirmations)

ui/components/app/flask/snap-ui-renderer/snap-ui-renderer.js
```diff
@@ -58,7 +58,7 @@ export const UI_MAPPING = {
   }),
 };
 
-const mapToTemplate = (data) => {
+export const mapToTemplate = (data) => {
```
FrederikBolding, line 61: Leave a comment to stop exporting this when we don't need this hack

ui/pages/confirmation/templates/flask/snap-confirmation/snap-confirmation.js
GuillaumeRx: No, it can also be an object if the children is a single component. Ref: https://github.com/MetaMask/metamask-extension/blob/f321225ed409c7502b15d295c6dc4ba09200731c/ui/pages/confirmation/templates/flask/snap-prompt/snap-prompt.js#L30

hmalik88 (approved):

Mrtenz (approved):

PR #16995, opened 2022-12-19
Making add un-approved transaction idempotent

### LINKED ISSUE #16965: Make method to add unapproved transaction idempotent
`MetamaskController.newUnapprovedTransaction` is required to be made idempotent. Request.id can be used as actionId [here](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/controllers/transactions/index.js#L333) to achieve this.

jpuri:
Fixes #16965

Using `req.id` as unique id to ensure that duplicate transactions are not generated.

### DIFF
--- a/app/scripts/controllers/transactions/index.js
+++ b/app/scripts/controllers/transactions/index.js
@@ -340,7 +340,7 @@ export default class TransactionController extends EventEmitter {
       opts.origin,
       undefined,
       undefined,
-      opts.id
+      opts.id,
     );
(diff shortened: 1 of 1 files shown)


digiwand (approved):

WalletSupportt:
> Fixes #16965
> 
> Using `req.id` as unique id to ensure that duplicate transactions are not generated.

segun (approved):

PR #16891, opened 2022-12-09
Replace `eth-optimism/contracts` package with local snippet of gas price ABI and address

adonesky1:
Fixes: #16565
See also: https://consensys.slack.com/archives/C01KXHPU3C6/p1662579353556779 (looks like all optimism team messages have been deleted though lol)

## Explanation
 We don't need to pull in  `@eth-optimism/contracts` with all the additional code and dependencies that it brings. We can do the query to the optimism gas price oracle with a snippet of the oracle contract ABI and its address -> instantiate the contract and call to it ourselves to get the L1 gas fee estimate.
 
 This PR simply removes the use of `@eth-optimism/contracts` helper method to achieve this and implements the required call all within our own `fetchEstimatedL1Fee` helper function.

To test:
- Attempt to initiate a transaction on Optimism
- Go to devtools -> source -> `fetchEstimatedL1Fee.js`
- Put a break point on line 41 (`return result?.toHexString();`)
- Check that the result returned is defined and is typed BigNumber. This indicates that we're successfully calling to the Optimism Gas price oracle.

## Pre-merge author checklist

- [x] I've clearly explained:
  - [x] What problem this PR is solving
  - [x] How this problem was solved
  - [x] How reviewers can test my changes

PR #16998, opened 2022-12-19
Update quote selection for Optimism

dan437:
## Explanation
Update quote selection for Optimism by including L1 fees for each quote.

### DIFF
--- a/app/scripts/controllers/swaps.js
+++ b/app/scripts/controllers/swaps.js
@@ -1,5 +1,4 @@
 import { ethers } from 'ethers';
-import Eth from 'ethjs';
 import log from 'loglevel';
 import BigNumber from 'bignumber.js';
 import { ObservableStore } from '@metamask/obs-store';
@@ -136,13 +135,11 @@ export default class SwapsController {
     this.indexOfNewestCallInFlight = 0;
 
     this.ethersProvider = new ethers.providers.Web3Provider(provider);
-    this.eth = new Eth(provider);
     this._currentNetwork = networkController.store.getState().network;
     networkController.on(NETWORK_EVENTS.NETWORK_DID_CHANGE, (network) => {
       if (network !== 'loading' && network !== this._currentNetwork) {
         this._currentNetwork = network;
         this.ethersProvider = new ethers.providers.Web3Provider(provider);
-        this.eth = new Eth(provider);
       }
     });
   }
@@ -301,10 +298,13 @@ export default class SwapsController {
       await Promise.all(
         Object.values(newQuotes).map(async (quote) => {
           if (quote.trade) {
-            const multiLayerL1TradeFeeTotal = await fetchEstimatedL1Fee({
-              txParams: quote.trade,
-              chainId,
-            });
+            const multiLayerL1TradeFeeTotal = await fetchEstimatedL1Fee(
+              {
+                txParams: quote.trade,
+                chainId,
+              },
+              this.ethersProvider,
+            );
             quote.multiLayerL1TradeFeeTotal = multiLayerL1TradeFeeTotal;
           }
           return quote;
--- a/app/scripts/controllers/swaps.js
+++ b/app/scripts/controllers/swaps.js
@@ -301,13 +301,10 @@ export default class SwapsController {
       await Promise.all(
         Object.values(newQuotes).map(async (quote) => {
           if (quote.trade) {
-            const multiLayerL1TradeFeeTotal = await fetchEstimatedL1Fee(
-              this.eth,
-              {
-                txParams: quote.trade,
-                chainId,
-              },
-            );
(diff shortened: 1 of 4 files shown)


app/scripts/controllers/swaps.js
```diff
@@ -291,6 +297,24 @@ export default class SwapsController {
       destinationTokenInfo: fetchParamsMetaData.destinationTokenInfo,
     }));
 
+    if (chainId === CHAIN_IDS.OPTIMISM && Object.values(newQuotes).length > 0) {
+      await Promise.all(
+        Object.values(newQuotes).map(async (quote) => {
+          if (quote.trade) {
+            const multiLayerL1TradeFeeTotal = await fetchEstimatedL1Fee(
```
adonesky1, line 304: As of [this PR](https://github.com/MetaMask/metamask-extension/pull/16891) (merged earlier today) this function no longer takes a provider as its first arg

adonesky1: Yea so it seems like since this is called in the context of the background it doesn't have access to that global.

adonesky1: here's a diff for change that I suggest we make:
[fix-eth-provider.txt](https://github.com/MetaMask/metamask-extension/files/10279362/fix-eth-provider.txt)

adonesky1:
```
diff --git a/app/scripts/controllers/swaps.js b/app/scripts/controllers/swaps.js
index 8b5a1c331..ea7d45e4c 100644
--- a/app/scripts/controllers/swaps.js
+++ b/app/scripts/controllers/swaps.js
@@ -301,10 +301,13 @@ export default class SwapsController {
       await Promise.all(
         Object.values(newQuotes).map(async (quote) => {
           if (quote.trade) {
-            const multiLayerL1TradeFeeTotal = await fetchEstimatedL1Fee({
-              txParams: quote.trade,
-              chainId,
-            });
+            const multiLayerL1TradeFeeTotal = await fetchEstimatedL1Fee(
+              {
+                txParams: quote.trade,
+                chainId,
+              },
+              this.ethersProvider,
+            );
             quote.multiLayerL1TradeFeeTotal = multiLayerL1TradeFeeTotal;
           }
           return quote;
diff --git a/ui/helpers/utils/optimism/fetchEstimatedL1Fee.js b/ui/helpers/utils/optimism/fetchEstimatedL1Fee.js
index da9477fd2..95c073665 100644
--- a/ui/helpers/utils/optimism/fetchEstimatedL1Fee.js
+++ b/ui/helpers/utils/optimism/fetchEstimatedL1Fee.js
@@ -19,8 +19,11 @@ const OPTIMISM_GAS_PRICE_ORACLE_ABI = [
 const OPTIMISM_GAS_PRICE_ORACLE_ADDRESS =
   '0x420000000000000000000000000000000000000F';
 
-export default async function fetchEstimatedL1Fee(txMeta) {
-  const provider = new Web3Provider(global.ethereumProvider, 10);
+export default async function fetchEstimatedL1Fee(txMeta, ethersProvider) {
+  const provider = global.ethereumProvider
+    ? new Web3Provider(global.ethereumProvider, 10)
+    : ethersProvider;
+
   if (process.env.IN_TEST) {
     provider.detectNetwork = async () => ({
       name: 'optimism',
```

app/scripts/controllers/swaps.js
```diff
@@ -132,11 +136,13 @@ export default class SwapsController {
     this.indexOfNewestCallInFlight = 0;
 
     this.ethersProvider = new ethers.providers.Web3Provider(provider);
+    this.eth = new Eth(provider);
```
adonesky1, line 139: This shouldn't be required any more

jpuri (approved):

ui/helpers/utils/optimism/fetchEstimatedL1Fee.js
```diff
@@ -19,8 +19,10 @@ const OPTIMISM_GAS_PRICE_ORACLE_ABI = [
 const OPTIMISM_GAS_PRICE_ORACLE_ADDRESS =
   '0x420000000000000000000000000000000000000F';
 
-export default async function fetchEstimatedL1Fee(txMeta) {
-  const provider = new Web3Provider(global.ethereumProvider, 10);
+export default async function fetchEstimatedL1Fee(txMeta, ethersProvider) {
```
adonesky1, line 22: did you want to rename this `ethereumProvider`?

dan437: I wanted at first, but then I saw we call it ethersProvider in many places, so I wanted to be consistent. Example: `this.ethersProvider = new ethers.providers.Web3Provider(opts.provider);`

adonesky1 (approved):
LGTM!

PR #17021, opened 2022-12-21
Ensure that account tracker does not set balances incorrectly when us…

### LINKED ISSUE #16835: [Bug]: Disabling Batch Accounts and switching to another Network displays cached balances from previous selected Network
### Describe the bug

**Problem**: if I disable the Batch Accounts feature and I switch to Localhost, I can see the balances from the previously selected network.

https://user-images.githubusercontent.com/54408225/206164817-1889ce62-1f63-4b69-8084-69276841a3a2.mp4

### Steps to reproduce

1. Select Sepolia Network
2. Go to Privacy Settings
3. Disable Batch Accounts toggle
4. On a terminal, run Ganache server
5. Select Localhost Network
6. See how the rest of Accounts preserve the Balance from Sepolia Network
7. Try to Send a tx -- tx fails as in reality I don't have the displayed balance

_No response_

### Version

10.24.0

### Build type

None

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

danjm:
…eMultiAccountBalanceChecker is false

Fixes https://github.com/MetaMask/metamask-extension/issues/16835

## Explanation

The bug being fixed can be described as: if the new "Batch account balance requests" toggle is set to off, then switching networks can cause accounts to show balances from the wrong network.

The problem is that when that toggle is set to off, and the `useMultiAccountBalanceChecker` property of preferences controller is false, then only one address is passed to the `_updateAccountsViaBalanceChecker`of `account-tracker.js` (and similarly `_updateAccount` is only called once). This is a problem because the logic of those methods assumes that the addresses passed will include addresses for all `accounts` in the AccountTracker state.

When receiving one address, `_updateAccountsViaBalanceChecker` will:
1. get all current accounts from state
2. fetch the balance for the passed address
3. update the object of current accounts with the balance fetched for the single address
4. then set `accounts` in state to this newly updated object

The current accounts of step 1 will contain data reflecting the network the user was on the last time that accounts state was set. The balance fetched in step 2 will always get the balance for the account on the current network. So if `_updateAccountsViaBalanceChecker` is called due to a network change, then the updated object from step 3 will contain balances for the old network for all accounts other than the one with the passed address, and for that passed address will contain data for the new network.

Tangibly speaking, if I am on mainnet and have three accounts with a balance of 10, 20 and 30, and I switch to a testnet while `useMultiAccountBalanceChecker`, and on the testnet my three accounts have balances of 0, 100 and 1000, and the currently selected address is the account with the 30 and 1000 balance, then after switching the UI will show the balances of my accounts as 10, 20 and 1000. `_updateAccountsViaBalanceChecker` will have erroneously combined the balance data of mainnet for two accounts (10 and 20) with the balance data of the testnet for the selected accounted (1000).

## Solution

To fix this problem, this PR ensures that the `accounts` state in `account-tracker.js` is never updated to contain incorrect balances. More precisely, if `_updateAccountsViaBalanceChecker` is called with fewer addresses than are in accounts state, we set the balance of the accounts that we will not fetch new data for to `null`. That is, if we don't know the balance and we have to set it to something, set it to `null` instead of an incorrect value.

This works because we already have logic throughout the codebase to handle undefined/falsy balances. In such cases, the UI falls back to the balances stored in the "cachedBalancesController" when possible, and will render 0 when not possible. Also, the cachedBalancesController itself simply won't update an account balance if the balance data passed to it on a an account tracker update is falsy.

So this solution ensures that the user is always shown the correct balance for the selected network, and the cached balance for other networks, when the "Batch account balance requests" toggle is set to off.

Note that the code change for `_updateAccount` is a little different than the code change for `_updateAccountsViaBalanceChecker`. `_updateAccount` is called once per address instead of having multiple addresses passed in. So we only want it to set account balances to false if we know we are in the scenario where it will only be called once, that is if `useMultiAccountBalanceChecker` is false. When `useMultiAccountBalanceChecker` is true, we assume it will be called for each address (via the `Promise.all(addresses.map(this._updateAccount.bind(this)))` call in the `_updateAccount**_s_**` method), and so each will have a new balance fetched and updated in state.

One additional change this PR makes is to have `_updateAccounts` called every time the selected account is changed if `useMultiAccountBalanceChecker` is false. This is just to improve user experience so that their account balances are updated more frequently. Currently those balances are updated on block updates and network changes. When `useMultiAccountBalanceChecker` is true, all balances are updated on those events, and so each account balance is relatively up to date at all times. When `useMultiAccountBalanceChecker` is false, only the current account is updated when the block updates. So switching to a new account will always show a stale balance until the next block update occurs. The change in this PR ensures that the `useMultiAccountBalanceChecker === false` users sees an updated balance upon switching accounts.


chloeYue (changes requested):
Hello, we have a side effect with this change that i'm not able to switch account to see the balance now.

https://user-images.githubusercontent.com/105063779/211348150-364aca9d-968b-4767-9721-01582f573770.mov

danjm:
Thanks @chloeYue. I have fixed that issue. I am going to add unit tests now.

jpuri (approved):
LGTM

seaona:
When I am trying to Import an Account, I land into the MetaMask encountered an error page, with the error message `Cannot read properties of undefined (reading 'balance')`

https://user-images.githubusercontent.com/54408225/212290844-1ce64093-e327-4e91-8241-5b24bc5dfe1e.mp4

seaona:
Ignore my last comment. I was on an older commit. The branch is working on my end

chloeYue (approved):

pedronfigueiredo (approved):

PR #17027, opened 2022-12-21
Adding browser outdated notification

### LINKED ISSUE #16283: Notify users that we are dropping support for their current browser
We would like to notify users when we are planning to drop support for the browser they are currently using, so that users have at least once release to learn about this change and how it will impact them.

vthomas13:
Adding a notification for users with outdated Chrome (<80) and Firefox (<78) browser versions that will lose support in the future. This notification should only show once every two days, while the browser remains outdated.

* Fixes #16283

## Manual Testing Steps
Download Chromium v78
[Windows](https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Win_x64/693951/), [macOS](https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Mac/693954/), [Linux](https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Linux_x64/693954/)
Install a MetaMask extension build from this branch.

## Before
image

## After
image

## Pre-merge author checklist

- [X] I've clearly explained:

- [X] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone


app/scripts/lib/createDupeReqFilterMiddleware.test.js
```diff
@@ -20,11 +20,11 @@ describe('createDupeReqFilterMiddleware', () => {
...
 
     filterFn(request, undefined, nextMock, endMock);
-    expect(nextMock).toBeCalledTimes(1);
-    expect(endMock).not.toBeCalled();
+    expect(nextMock).toHaveBeenCalledTimes(1);
+    expect(endMock).not.toHaveBeenCalled();
 
     filterFn(request, undefined, nextMock, endMock);
-    expect(nextMock).toBeCalledTimes(1);
-    expect(endMock).toBeCalledTimes(1);
```
vthomas13, line 28: These changes were introduced from the yarn lint:fix command. I can take them out if necessary.

Gudahtt: Hmm. That shouldn't happen, but I did confirm this to be the case just because I was curious. Weird.

This problem does go away after rebasing/merging latest develop though.

vthomas13: okay, thanks for the insight @Gudahtt, I still need to get familiar with the pipeline

vthomas13:

ui/helpers/constants/browser.js
```diff
@@ -0,0 +1,4 @@
+export const OUTDATED_BROWSER_VERSIONS = {
+  chrome: '<80',
+  firefox: '<78',
+};
```
jpuri, line 4: May be we can use existing constant files and not add new just for browser constants.

ui/pages/home/home.component.js
```diff
@@ -244,6 +246,11 @@ export default class Home extends PureComponent {
     setRecoveryPhraseReminderLastShown(new Date().getTime());
   };
 
+  onOutdatedBrowserWarningClose = () => {
+    const { setOutdatedBrowserWarningLastShown } = this.props;
+    setOutdatedBrowserWarningLastShown(new Date().getTime());
+  };
```
jpuri, line 252: where is this state variable being used `outdatedBrowserWarningLastShown` ?

jpuri (dismissed):
Looks good 👍

ui/selectors/selectors.js
```diff
@@ -1048,6 +1048,13 @@ export function getShowRecoveryPhraseReminder(state) {
   return currentTime - recoveryPhraseReminderLastShown >= frequency;
 }
 
+export function getShowOutdatedBrowserWarning(state) {
+  const { outdatedBrowserWarningLastShown } = state.metamask;
+
+  const currentTime = new Date().getTime();
+  return currentTime - outdatedBrowserWarningLastShown >= DAY * 2;
+}
```
jpuri, line 1056: logic like above should be avoided in selector I think

jest.config.js
```diff
@@ -35,6 +35,12 @@ module.exports = {
   setupFilesAfterEnv: ['<rootDir>/test/jest/setup.js'],
   testMatch: [
     '<rootDir>/app/scripts/constants/error-utils.test.js',
+    '<rootDir>/ui/**/*.test.js',
+    '<rootDir>/shared/**/*.test.js',
+    '<rootDir>/app/scripts/lib/**/*.test.js',
+    '<rootDir>/app/scripts/migrations/*.test.js',
+    '<rootDir>/app/scripts/platforms/*.test.js',
```
Gudahtt, line 42: These all look like duplicate entries
```suggestion
```

ui/helpers/utils/util.test.js
```diff
@@ -194,6 +195,37 @@ describe('util', () => {
     });
   });
 
+  describe('#getIsBrowserDeprecated', () => {
```
Gudahtt, line 198: Could you add test cases for Edge and Opera as well?

I'm also curious how browser forks are treated here. Could you add a test for an unrecognized browser?

ui/pages/home/home.component.js
```diff
@@ -476,6 +484,21 @@ export default class Home extends PureComponent {
             key="home-infuraBlockedNotification"
           />
         ) : null}
+        {showOutdatedBrowserWarning ? (
+          <HomeNotification
+            descriptionText={t('outdatedBrowserNotification', [
+              <span
```
Gudahtt, line 490: It looks like you're passing a "Here" span into the outdated browser notification message, but there aren't any substitutions in that message. This syntax is for when we're embedding something in a message.

ui/store/actions.js
```diff
@@ -3273,6 +3273,20 @@ export function setRecoveryPhraseReminderLastShown(lastShown) {
   };
 }
 
+export function setOutdatedBrowserWarningLastShown(lastShown) {
+  return () => {
+    callBackgroundMethod(
```
Gudahtt, line 3278: We should use the async `submitRequestToBackground` here instead. This `callBackgroundMethod` function should be marked as deprecated. We've been in the process of migrating away from using the callback-style background calls for years now.

Gudahtt: Also, this is an async call so it should be an `async` function that waits for the call to complete. That makes it possible for the caller to wait, if they want to. Otherwise there's no way to tell when the request has completed.

app/scripts/controllers/app-state.test.js
```diff
@@ -0,0 +1,32 @@
+import AppStateController from './app-state';
+
+describe('AppStateController', function () {
+  const appStateController = new AppStateController({
```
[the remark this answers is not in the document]
vthomas13: Thanks, great point.

ui/pages/home/home.component.js
vthomas13: Thanks!

ui/store/actions.js
```diff
@@ -3273,6 +3273,17 @@ export function setRecoveryPhraseReminderLastShown(lastShown) {
   };
 }
 
+export function setOutdatedBrowserWarningLastShown(lastShown) {
+  return async (dispatch) => {
+    await submitRequestToBackground('setOutdatedBrowserWarningLastShown', [
+      lastShown,
+    ]).catch((err) => {
```
Gudahtt, line 3280: Nit: Could you use `try .. catch` here instead of `.catch`? The former will preserve an async stack trace across this line on Chromium-based browsers, leading to better error messages. Generally async/await should be preserved over .then/.catch in all cases for this reason.

ui/helpers/utils/util.test.js
vthomas13: Right now it looks like Opera is unhandled because bowser handles by name, but I've manually tested version 58 which was released Jan/23/2019 (around the same time as Chromium v78 and it seemed to work fine. I can add it to the list of outdated browser ranges as '<58'

Gudahtt: Could you clarify what you mean by "unhandled"? Does that mean it will _always_ say that the browser is outdated, or `_never_`? Not sure what the default case is.

vthomas13: The default case is the current functionality, where the user would not see anything different. So it will never say the browser is outdated if it's not in [this constant](https://github.com/MetaMask/metamask-extension/blob/938fd4ebaca5310c8a1b7b94214d11e383edc7c1/ui/helpers/constants/common.js#L27-L30)

Gudahtt: The Wikipedia page claims that Opera 65 was based on Chromium v78, so maybe we can set that as the minimum for Opera: https://en.wikipedia.org/wiki/History_of_the_Opera_web_browser

Gudahtt: Oh sorry, it's Chrome 80 we're targeting, which'd be Opera 67

>Opera 67 was released on February 25, 2020, based on Chromium 80.

ui/store/actions.js
```diff
@@ -3273,6 +3273,19 @@ export function setRecoveryPhraseReminderLastShown(lastShown) {
...
+export function setOutdatedBrowserWarningLastShown(lastShown) {
+  return async (dispatch) => {
+    try {
+      await submitRequestToBackground('setOutdatedBrowserWarningLastShown', [
+        lastShown,
+      ]);
+      await forceUpdateMetamaskState(dispatch);
+    } catch (error) {
+      dispatch(displayWarning(error.message));
+    }
```
Gudahtt, line 3285: This `displayWarning` method should be considered deprecated as well. That behavior shows the error directly to the error in a very confusing and nonsensical way.

Sorry when I suggested using `try..catch` I hadn't considered whether there was an error to catch in the first place. It's probably better to let this one bubble up.

```suggestion
    await submitRequestToBackground('setOutdatedBrowserWarningLastShown', [
      lastShown,
    ]);
    await forceUpdateMetamaskState(dispatch);
```

vthomas13: Gotcha, I appreciate you explaining the rationale behind it too, thanks!

Gudahtt: Documented here: https://github.com/MetaMask/metamask-extension/pull/17465

Gudahtt (dismissed):
LGTM, great work!

mcmire (dismissed):
Looks good, just had a couple of suggestions, but you can take them or leave them.

ui/store/actions.js
```diff
@@ -3273,6 +3273,15 @@ export function setRecoveryPhraseReminderLastShown(lastShown) {
   };
 }
 
+export function setOutdatedBrowserWarningLastShown(lastShown) {
+  return async (dispatch) => {
+    await submitRequestToBackground('setOutdatedBrowserWarningLastShown', [
+      lastShown,
+    ]);
+    await forceUpdateMetamaskState(dispatch);
```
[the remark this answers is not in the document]
Gudahtt: Good point. It does update automatically, but it's async. We use this force update in cases where we need to guarantee that the state has been updated before we perform the next operation.

In this case there is no next operation though, so I don't think we need it

app/scripts/controllers/app-state.test.js
```diff
@@ -0,0 +1,33 @@
+import AppStateController from './app-state';
+
+describe('AppStateController', function () {
```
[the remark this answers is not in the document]
vthomas13: I think so too, I think the lint:fix command changed it to function notation instead of lambda notation

.eslintrc.js
```diff
@@ -262,6 +262,7 @@ module.exports = {
     {
       files: [
         '**/__snapshots__/*.snap',
+        'app/scripts/controllers/app-state.test.js',
```
Gudahtt, line 265: It looks like this needs to be listed under the `excludedFiles` entry of the block above here. That's probably why the linter converted the test to use functions - it's a mocha rule, and that's still getting applied.

app/scripts/controllers/app-state.test.js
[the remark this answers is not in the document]
mcmire: Ah. Interesting. Okay!

.eslintrc.js
Gudahtt: Probably should call out more explicitly that this list needs to be in-sync with the list in the previous block. I see there are other discrepancies as well

Gudahtt (dismissed):
One pending suggestion, but we can handle that in a follow-up PR instead if you want

.eslintrc.js
vthomas13: I could sync that up, but I think the formatting has to be changed to lambda notation in each test file manually, so it might be better left for a separate PR/changeset.

Gudahtt: Oh yeah definitely, I agree

Gudahtt (commented):
Looks great! I can re-approve when the conflicts are resolved

pedronfigueiredo (dismissed):

Gudahtt (approved):

brad-decker (approved):
We reviewed this one ysterday live, and now that the change is made this looks good to me!

PR #16885, opened 2022-12-09
button link housekeeping

garrettbear:
## Explanation

* Fixes #16635 

### Before
Screenshot 2022-12-16 at 10 17 32 AM

### After
Screenshot 2022-12-16 at 10 18 19 AM

## Manual Testing Steps
`yarn test:unit:jest .ui/components/component-library/button-link/button-link.test.js`

## Pre-merge author checklist

- [x] I've clearly explained:
  - [x] What problem this PR is solving
  - [x] How this problem was solved
  - [x] How reviewers can test my changes
- [x] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [x] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
- [x] PR is linked to the appropriate GitHub issue
- [] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.

In this case, a QA Engineer approval will be be required.

### Housekeeping
Ensure that `ButtonLink` adheres to all of the following conventions and standards

- [x] Has a `className` prop and the PropType descriptions are all the same
- [x] Prop table in MDX docs have the "Accepts all Box component props" description and link
- [ ] We are consistent when using the same prop names like `size` and are suggesting the use of the generalized `design-system.js` constants e.g. `SIZES` as the primary option but noting the component consts in the documentation and using them for propType validation and storybook controls only
- [x] Standardize all similar prop names for images `imgSrc`, `imgAlt`(html element + attribute) (needs audit)
- [x] We have a story for each component prop and we use the prop name verbatim e.g. `size` prop would be `export const Size = (args) => (`
- [x] We have the accompanying documentation for each component prop and we use the prop name verbatim e.g. `size` prop would be `### Size`
- [x] Are multiple props stories allowed? e.g. `Color, Background Color And Border Color` story in `base-avatar` - [ ] yes when it makes sense to
- [ ] All Base components follow the suffix convention e.g. `ButtonBase`
- [ ] All Base component MDX documentation have the base component notification at the top
- [x] Add `mm-` prefix to all classNames
- [x] className is kebab case version of the component name
- [x] Spread base components props and reduce duplication of props when props aren't being changed and remain the same for both variant and base components
- [x] Add component to root `index.js` file in component-library
- [ ] Add locals for any default text I18nContext as default context 
[...]

PR #17034, opened 2022-12-22
`ButtonLink` some alternative explorations

georgewrmarshall:
## Explanation
In the majority of use cases that I've seen for ButtonLink the hit area should be the same as the text area. I wonder if we focus on making that the default and instead provide examples of how you could extend the hit area when needed.

- This PR is built off of https://github.com/MetaMask/metamask-extension/pull/16885 
- Adds `SIZES.AUTO` back that only adds the CSS: `height: auto`
- Sets the default to `SIZES.AUTO` and removes padding for all sizes
- Removes `noPadding` in favour of showing an example of adding padding using Box props

These updates make the every instance I've used the `ButtonLink` so far the default see `FormTextField` Custom Label or HelpText story

### Screencast

https://user-images.githubusercontent.com/8112138/209051101-49bfd87b-e5cf-4915-a17d-36d3382b2be7.mo


ui/components/component-library/button-base/button-base.js
```diff
@@ -54,6 +55,7 @@ export const ButtonBase = ({
       display={DISPLAY.INLINE_FLEX}
       justifyContent={JUSTIFY_CONTENT.CENTER}
       alignItems={ALIGN_ITEMS.CENTER}
+      borderRadius={BORDER_RADIUS.PILL}
       {...props}
     >
       <Text
```
georgewrmarshall, line 61: This was a test to see if we can save on CSS and overwriting CSS. If we want to remove the border-radius from the ButtonLink component we can just use `borderRadius={null}` - it removes the Box class. So less extra CSS on the base component and less CSS to overwrite the base component on the variant

PR #17052, opened 2022-12-22
Improve `SignatureRequest` performance

FrederikBolding:
## Explanation

Alleviates some performance problems with `SignatureRequest` and related components which were repeatedly re-rendering causing at least part of a slow-down in the extension.
Fixes:
- Memoize `SignatureRequestData` and deep equal check data before re-rendering
- Memoize `getMetaMaskIdentities` used in `SignatureRequestData` component
- Memoize `getAddressBook` used in `Address` component
- Memoize `getMetadataContractName` used in `Address` component
- Memoize sanitization of payload

### DIFF
--- a/ui/selectors/selectors.js
+++ b/ui/selectors/selectors.js
@@ -819,6 +819,22 @@ export const getMemoizedMetaMaskIdentities = createDeepEqualSelector(
   (identities) => identities,
 );
 
+export const getMemoizedAddressBook = createDeepEqualSelector(
+  getAddressBook,
+  (addressBook) => addressBook,
+);
+
+export const getMemoizedMetadataContractName = createDeepEqualSelector(
+  getTokenList,
+  (_tokenList, address) => address,
+  (tokenList, address) => {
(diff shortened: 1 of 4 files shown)


ui/selectors/selectors.js
```diff
@@ -814,6 +814,27 @@ export function getShowWhatsNewPopup(state) {
...
+);
+
+export const getMemoizedAddressBook = createDeepEqualSelector(
+  getAddressBook,
+  (addressBook) => addressBook,
+);
+
+export const getMemoizedMetadataContractName = createDeepEqualSelector(
+  getTokenList,
+  (_tokenList, address) => address,
```
danjm, line 829: Nit, non-blocking: the first parameter of this second function should be `_state` and not `_tokenList`

danjm (approved):
LGTM, but because we want to release this today, let's get 2 more code review approvals before we merge

adonesky1 (approved):
This all looks right to me! Tested and is indeed faster as well!

brad-decker (approved):
LGTM

PR #17056, opened 2022-12-22
Integrate `KeyringController` v10

adonesky1:
Integrates `@metamask/scure-bip39` and `eth-keyring-controller` v9 and adapts accordingly

Context for changes are provided with inline comments 😄


app/scripts/metamask-controller.js
```diff
@@ -2501,7 +2510,11 @@ export default class MetamaskController extends EventEmitter {
     if (!keyring.mnemonic) {
       throw new Error('Primary keyring mnemonic unavailable.');
     }
-    return keyring.mnemonic;
+
+    const recoveredIndices = Array.from(
+      new Uint16Array(new Uint8Array(keyring.mnemonic).buffer),
+    );
+    return recoveredIndices.map((i) => englishWordlist[i]).join(' ');
```
adonesky1, line 2517: `getPrimaryKeyringMnemonic` is used by Snaps which still passes mnemonics around as strings. They will be refactoring to pass mnemonics as `Uint8Arrays` but until then we must recover the string from the `Uint8Array` format that is now on the keyring from the KeyringController as of this latest version bump:
```
 const [keyring] = this.keyringController.getKeyringsByType(
      KEYRING_TYPES.HD_KEY_TREE,
    );
```

app/scripts/metamask-controller.js
```diff
@@ -2501,13 +2511,12 @@ export default class MetamaskController extends EventEmitter {
     if (!keyring.mnemonic) {
       throw new Error('Primary keyring mnemonic unavailable.');
     }
-    return keyring.mnemonic;
+    return uint8ArrayMnemonicToString(keyring.mnemonic);
```
adonesky1, line 2514: `getPrimaryKeyringMnemonic` is used by Snaps which still passes mnemonics around as strings. They will be refactoring to pass mnemonics as Uint8Arrays but until then we must recover and return the string version from the Uint8Array format that is now on the `keyring` value returned by the `KeyringController` on line 2508 as of this latest version bump.

app/scripts/metamask-controller.js
```diff
@@ -2781,16 +2789,16 @@ export default class MetamaskController extends EventEmitter {
...
-    const seedPhraseAsBuffer = Buffer.from(serialized.mnemonic);
 
     const accounts = await primaryKeyring.getAccounts();
     if (accounts.length < 1) {
       throw new Error('MetamaskController - No accounts found');
     }
 
     try {
-      await seedPhraseVerifier.verifyAccounts(accounts, seedPhraseAsBuffer);
-      return Array.from(seedPhraseAsBuffer.values());
```
adonesky1, line 2793: I'm a bit confused how/why this was encoded? `verifySeedPhrase` appears to be used in the few places we actually want to reveal the SRP as a string: [requestRevealSeedWords](https://github.com/MetaMask/metamask-extension/blob/develop/ui/store/actions.js#L228) and [seed-phrase.component.js](https://github.com/MetaMask/metamask-extension/blob/develop/ui/pages/first-time-flow/seed-phrase/seed-phrase.component.js#L33) for instance

adonesky1: ok so yeah this function is only ever called by the [verifySeedPhrase action](https://github.com/MetaMask/metamask-extension/blob/develop/ui/store/actions.js#L216) in the front end which immediately converts it back to a plain text string. I can't remember if it's important that we keep it encoded when sending it from background to frontend, but currently this PR removes this flow where we convert it to a buffer and back in quick succession. I can add back the brief obfuscation if we think it's useful! @Gudahtt

app/scripts/metamask-controller.test.js
```diff
@@ -509,6 +511,31 @@ describe('MetaMaskController', function () {
...
+        new Uint16Array(mnemonicIndices).buffer,
+      );
+
+      const mockHDKeyring = {
+        type: 'HD Key Tree',
+        mnemonic: uint8ArrayMnemonic,
+      };
+      sinon
+        .stub(metamaskController.keyringController, 'getKeyringsByType')
+        .returns([mockHDKeyring]);
```
adonesky1, line 531: This mocks the way `mnemonics` are stored on HD keyrings - as `Uint8Arrays` of the indices of the words relative to the wordlist -  as of https://github.com/MetaMask/eth-hd-keyring/pull/67. This test then demonstrates that even though the mnemonic is stored in this way, the `getPrimaryKeyringMnemonic` method on `MetamaskController` returns the mnemonic in it's string form, since this is required by Snaps for now.

app/scripts/lib/seed-phrase-verifier.js
```diff
@@ -1,4 +1,4 @@
-import KeyringController from 'eth-keyring-controller';
+import { KeyringController } from 'eth-keyring-controller';
```
adonesky1, line 1: `KeyringController` is now a named export: https://github.com/MetaMask/KeyringController/blob/main/CHANGELOG.md?plain=1#L17

app/scripts/metamask-controller.js
```diff
@@ -595,15 +601,19 @@ export default class MetamaskController extends EventEmitter {
...
-      additionalKeyrings = [
+      const additionalKeyringTypes = [
         TrezorKeyring,
         LedgerBridgeKeyring,
         LatticeKeyring,
         QRHardwareKeyring,
       ];
+      additionalKeyrings = additionalKeyringTypes.map((keyringType) =>
+        keyringBuilderFactory(keyringType),
+      );
```
adonesky1, line 612: The `KeyringController` constructor now takes keyring builder functions rather than classes.

`keyringBuilderFactory` takes the class types and makes them into these builder functions.

app/scripts/lib/util.js
```diff
@@ -299,3 +301,17 @@ export function formatTxMetaForRpcResult(txMeta) {
...
+ * @param {Uint8Array} mnemonic - a mnemonic in Uint8Array format
+ * @returns {string} the mnemonic in string form.
+ */
+export function uint8ArrayMnemonicToString(mnemonic) {
+  const recoveredIndices = Array.from(
+    new Uint16Array(new Uint8Array(mnemonic).buffer),
+  );
+
+  return recoveredIndices.map((i) => englishWordlist[i]).join(' ');
+}
```
jpuri, line 317: Hey @adonesky1 : just curious, how is above change related to KeyringController ?

adonesky1: So this helper function is used [here](https://github.com/MetaMask/metamask-extension/pull/17056/files/d11ff3886d11f58bed97b36f134745b0bc1898cc#diff-6fbff2cfe97ac01b77296ef2122c7e0a5b3ff6a84b584b4d1a87482f35eea3d6R2514). The mnemonic returned on the keyring by `this.keyringController.getKeyringsByType(KEYRING_TYPES.HD_KEY_TREE)` is now a `Uint8Array` rather than a plain text string, but the `getPrimaryKeyringMnemonic` function that is used by snaps still expects it as a string so we need this function to convert it back.

jpuri (dismissed):

app/scripts/metamask-controller.js
Gudahtt: This was obfuscated here because we had, at the time, intended to leave it obfuscated for the entire process. But we stopped short of completing that work.

This does ensure that the SRP is never in string form in the background/service-worker process as well. At the time we thought this might make a difference to when we'd expect to see the SRP dropped from memory, but that was never able to be validated.

adonesky1: Ok so it would be preferable to return the SRP as a `Uint8Array` here and then convert it (as was being done before) in the `verifySeedphrase` action?

Gudahtt: Yeah I think so. I'm not _entirely_ sure it matters, but that might be slightly safer, and would leave us in a better position to continue this obfuscation work in the future if we ever decide to.

adonesky1: Fixed [here](https://github.com/MetaMask/metamask-extension/pull/17056/commits/b9b679bbedc9b66b6d2a9cef29649fb045d96630) & https://github.com/MetaMask/metamask-extension/pull/17056/commits/060a7cac0d1e866ff12b45ba6aedad092ce163d9

Gudahtt: It looks like it's now being sent as a Uint8Array rather than as a regular array. I would not expect this to work on Chrome because it's not valid JSON.

Gudahtt:
In reviewing this, I've noticed that the format of the mnemonic saved in the vault has changed with this update. This change was missed from the changelog as well. The format isn't especially great either, in that it's unconventional (it's essentially an array that was casted to an object - why not use an array instead?).

This format change means we'll have to be careful when releasing this, because if we roll back whatever release this lands in, vault decryption will fail. So we can't roll this change back. Is this format change worth the trouble? We could switch back to the old one.

adonesky1:
> In reviewing this, I've noticed that the format of the mnemonic saved in the vault has changed with this update. This change was missed from the changelog as well. The format isn't especially great either, in that it's unconventional (it's essentially an array that was casted to an object - why not use an array instead?).
> 
> This format change means we'll have to be careful when releasing this, because if we roll back whatever release this lands in, vault decryption will fail. So we can't roll this change back. Is this format change worth the trouble? We could switch back to the old one.
> 

@Gudahtt is aware of this but for broader visibility for anyone else reviewing this PR, this specific concern has been addressed with the updates made to `@metamask/eth-hd-keyring` [here](https://github.com/MetaMask/eth-hd-keyring/pull/81) and propogated through [the latest version](https://github.com/MetaMask/KeyringController/pull/194) of `@metamask/eth-keyring-controller` now pulled in here.

app/scripts/metamask-controller.js
adonesky1: This change is no longer part of this PR (as of the latest commits) because of the changes made to @metamask/eth-hd-keyring https://github.com/MetaMask/eth-hd-keyring/pull/81 which are integrated into https://github.com/MetaMask/KeyringController/pull/194 of @metamask/eth-keyring-controller and now pulled in to this PR

Gudahtt (approved):
LGTM!

jpuri (approved):

PR #16994, opened 2022-12-19
Integrate new LavaMoat scuttling protection feature

weizman:
> Continue work from #15581 after messing it up

## Motivation

Integrating into MetaMask browser extension the new scuttling feature introduced by LavaMoat ([#360](https://github.com/LavaMoat/LavaMoat/pull/360))

## PR Breakdown

1. lockdown more collides with scuttling feature, adjust it to skip scuttled props
2. `scuttleGlobalThisExceptions` conf prop is populated by 3 different groups: (1) globals that are used by different mm deps that unsafely access global object props and therefore mess with scuttling; (2) globals the chrome driver uses in the test env of mm which cannot be sandboxed and therefore must be excused; and (3) globals vetted shim sentry needs to write to global object in order to function. **All of these will be narrowed down with time and research in different PRs**
3. Also, in order for sentry to work as the vetted shim that it is, we need to allow it to still access scuttled properties, therefore we wrap it with a special sandbox (a mix of `with`s and `proxy`s) that allows it to function normally in a scuttled envirnoment (partially by indicating a list of the properties it'll need that will be later scuttled)
4. update `test/helpers/protect-intrinsics-helpers.js` test to test intrinsics protection only for non scuttled props

In this PR currently there are patch-package files, only to apply the needed upgrade for LM - those will be removed before merge and will be replaced with an actual newly delivered version of LM

## Notes

This is an ongoing effort which will bring with it more PRs in the future. In following PRs we should focus on narrowing down as much as possible the scuttling avoids list which currently contains more fields we skip scuttling for then we'd want to ideally, some of them are potentially quite useful for attackers.

This will require some research.

PR #17061, opened 2022-12-23
Update LavaMoat (core/lavapack)

weizman:
Context:
* https://github.com/MetaMask/metamask-extension/pull/16994 was merged which introduced new LM versions and the new LM scuttling feature
* That PR caused problems with `yarn start` command and therefore was reverted completely in https://github.com/MetaMask/metamask-extension/pull/17043
* This PR https://github.com/MetaMask/metamask-extension/pull/17061 reintegrates only those new LM versions with a fix to the `yarn start` problem and without activating the scuttling feature (this will be done in a separate PR)


weizman:
cc78a0572 is because pattern was incorrect for runtime files when are watched by gulp

development/build/index.js
```diff
@@ -72,6 +73,9 @@ async function defineAndRunBuildTasks() {
     version,
   } = await parseArgv();
 
+  // build lavamoat runtime file
+  await lavapack.buildRuntime({ scuttleGlobalThis: false });
```
kumavis, line 77: this should go somewhere, and not be built in place in the package
in some cases we are running multiple instances of the build system in parallel with different build configurations, and they could conflict

development/build/static.js
```diff
@@ -80,7 +80,7 @@ module.exports = function createStaticAssetTasks({
   return { dev, prod };
 
   async function setupLiveCopy(target, browser) {
-    const pattern = target.pattern || '/**/*';
+    const pattern = 'pattern' in target ? target.pattern : '/**/*';
```
kumavis, line 83:
```suggestion
    const pattern = target.pattern === undefined ? target.pattern : '/**/*';
```

development/build/static.js
```diff
@@ -89,7 +89,7 @@ module.exports = function createStaticAssetTasks({
   }
 
   async function performCopy(target, browser) {
-    if (target.pattern) {
+    if ('pattern' in target) {
```
kumavis, line 92:
```suggestion
    if (target.pattern === undefined) {
```

test/helpers/protect-intrinsics-helpers.js
```diff
@@ -28,7 +28,10 @@ function getGlobalProperties() {
       // TODO: Also include the named platform globals
       // This grabs every enumerable property on globalThis.
       // ...Object.keys(globalThis),
-    ].filter((propertyName) => !ignoreList.has(propertyName)),
+    ].filter(
+      (propertyName) =>
+        typeof propertyName === 'string' && !ignoreList.has(propertyName),
```
kumavis, line 33:
```suggestion
        // dont test symbols because [...]
        typeof propertyName === 'string' && !ignoreList.has(propertyName),
```

development/build/index.js
kumavis: since the runtime is no longer static (its built based on a parameterization based on build configuration) we should:
- remove it from the static tasks
- modify `lavapack.buildRuntime` to return the string of the built runtime
- generate the runtime in `development/build/scripts` when building with lavamoat
- write it out to the correct destinations for each browser

weizman: would it be ok to fix this in a newer lavamoat version after merging this PR?

kingdutch001: Yes fix

kingdutch001:
await lavapack.buildRuntime({ scuttleGlobalThis: false

development/build/index.js
kumavis: since lavapack is already published as is, thats fine. be sure to make this requested fix soon

kingdutch001:
That will be wonderful

test/helpers/protect-intrinsics-helpers.js
weizman: 6a7d51f

weizman: Fix a4c65ba4fb2bf574cd2abb3804120e2435c0ce9a by properly adding the new symbol to the ignore list to correctly ignore.

The new symbol, introduced in this new lockdown version, is being ignored because it uniquely belong to the globalThis of new compartments but does not belong as a property to the real globalThis (and this lockdown test verifies that the real globalThis is frozen, so this new symbol is irrelevant in this test).

@kumavis

weizman: In other words, we ignore symbols because the real globalThis doesn't even have any symbol own properties to test lockdown against, so no need for that

development/build/index.js
weizman: https://github.com/LavaMoat/LavaMoat/issues/451 opened issue to solve this

test/helpers/protect-intrinsics-helpers.js
```diff
@@ -11,14 +11,17 @@ module.exports = {
  * @returns {Set<string>} All global intrinsic property names.
  */
 function getGlobalProperties() {
+  const comp = new Compartment().globalThis;
```
kumavis, line 14:
```suggestion
  const compartmentGlobalThis = new Compartment().globalThis;
```
naming things is hard. this is only slightly better than `comp`

test/helpers/protect-intrinsics-helpers.js
```diff
@@ -11,14 +11,17 @@ module.exports = {
...
+
   // These are Agoric inventions, and we don't care about them.
   const ignoreList = new Set([
     'Compartment',
     'HandledPromise',
     'StaticModuleRecord',
+    ...Object.getOwnPropertySymbols(comp),
```
kumavis, line 21: interesting, what symbols did you encounter?

test/helpers/protect-intrinsics-helpers.js
kumavis: oh this is just for a test helper? its fine

kumavis (approved):

digiwand (approved):

PR #17077, opened 2022-12-30
fix: Display primaryType in signTyped instead of 'Message' string, Fixes #17076

### LINKED ISSUE #17076: primaryType is not shown during signTyped - users are signing things almost blindly
### Describe the bug

PrimaryType isn't displayed when signing Typed data, instead it always shows a "Message" string and a bunch of parameters.

IMHO this goes against the primary goal of EIP712 - a more user-friendly way of signing messaging.
It's astonishing that for 3 years people were just signing things blindly!

**Example:**
If two functions have exact same parameters but do completely different things - there is no way to distinguish between a harmless signature (like setting a limit on something), and a really dangerous one (like permitForAll that potentially can lead to loss of all your tokens).
Even if parameters differ slightly - it's still hard for a non-technical user to see what exactly they sign.

**Real world example:**
Recently there was a loss of many Lens Protocol profiles from malicious websites that asked users for a signature, mimicking a login. If user could see they are signing "permitForAll" or "burnWithSig" - they wouldn't sign it.

**Current behavior:**
telegram-cloud-photo-size-4-5870527322589870546-y

**Expected behavior:**
Before 2019 everything worked as expected, but in PR #6891 the `primaryType` was replaced with a "Message" string for a reason unknown to me (maybe @tshfx can explain more details about the reasoning of this).
The EIP712 example shows a screenshot with "Mail" as primaryType, so I assume that also confirm this behavior is the expected.

**Proposed solution:**
Display primaryType instead of "Message" (like "PermitForAll" in this case)
image

### Steps to reproduce

1. Sign any TypedData, look at the signature message window

_No response_

### Version

10.23.2

### Build type

None

### Browser

Chrome

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

vicnaum:
## Explanation

Fixes #17076 

* What is the current state of things and why does it need to change?

When signing Typed data user doesn't see what primaryType (a function) they are signing, instead they always see "Message".

* What is the solution your changes offer and how does it work?

User should see the name of the function they sign (like "PermitForAll") instead of "Message" string.

### Before

image

### After

image

## Manual Testing Steps

- Sign any Typed data

## Pre-merge author checklist

- [X] I've clearly explained:

My PR does fix this in a quick way, but I'm sure a proper fix should take more than this - would like to ask for advice for core devs here on what else should be changed, and a proposed architecture.
Also not all tests are passing on my machine (but it's the same with a clean develop branch - so maybe it's just my local issue).


vicnaum:

danjm:
@vicnaum Thanks for this contribution. It looks correct and we would like to get it merged soon. There is a lint error that you will need to fix. You should be able to run `yarn verify-locales:fix` to fix the "test-lint" failure. I'll comment inline on how to fix the other failure.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Added displaying the primaryType during Typed data signing ([#17077](https://github.com/MetaMask/metamask-extension/pull/17077))
```
danjm, line 8:
```suggestion
### Fixed
- Added displaying the primaryType during Typed data signing ([#17077](https://github.com/MetaMask/metamask-extension/pull/17077))
```

danjm:
It would also be good if `ui/components/app/signature-request/signature-request.component.test.js` could be update to include an assertion for the primary type. Are you able to do that @vicnaum? If you'd prefer that we do that, please just let us know with a comment here.

vicnaum:
@danjm Thanks! I've pushed the suggested fixes and also tried to add an assertion to the test - please check if that's what needed

ui/components/app/signature-request/signature-request-message/signature-request-message.js
```diff
@@ -81,7 +82,7 @@ export default function SignatureRequestMessage({
           marginLeft={4}
           className="signature-request-message__title"
         >
-          {t('signatureRequest1')}
+          {primaryType}
```
jpuri, line 85: May be `{primaryType || t('signatureRequest1')}`.

jpuri: Ignore this as I found every typed message except V1 has a primary type. we are not using this component for V1.

jpuri (approved):

seaona:
I am wondering what should happen when the dapp does not provide any `primaryType`. Should we handle the error somehow? At the moment, MetaMask popup tries to open but it's immediately closed, and we get an RPC error in the background console. 

See below:

https://user-images.githubusercontent.com/54408225/212046387-98fefb39-2040-402f-b22c-0a463f656f16.mp4

On the [EIP-712 specs](https://eips.ethereum.org/EIPS/eip-712), it states that both the `primaryType` and the `domain` should be mandatory, so I think it's okay what we do.

In the case of the `domain` , the same behaviour happens, and we display this different error on the console:

image

FrederikBolding (approved):

holantonela:
> On the [EIP-712 specs](https://eips.ethereum.org/EIPS/eip-712), it states that both the primaryType and the domain should be mandatory, so I think it's okay what we do.

Agreed. We are also enforcing mandatory fields in other custom typesign like SIWE.

> In the case of the domain , the same behaviour happens, and we display this different error on the console

Should we unify the error message in these cases to keep consistency?

kingdutch001:
I can confirm that this issue is still exists on latest next (v13.1.1). I use getStaticProps + getStaticPaths with fallback: blockfing and middleware. 404 pages works incorrectlry with client-side navigation. I have [slug].js pages. If I open exisiting page like /post-1 and then open 404 page like /post-not-found I see /post-1 content (what???). If I open 404 page directlry (I see 404 error page) and then open another 404 page like /post-another-not-found I see correct 404 error page. If I open some other page (not [siug].js and then open 404 page I see empty page without any content.

brad-decker:
@FrederikBolding / @holantonela is this ready for merrge? If so one of you can do the honors :)

seaona:
thank you for the inputs @holantonela @FrederikBolding

PR #17079, opened 2023-01-04
fix audit failure

brad-decker:
## Explanation
Fixes current audit failure on develop. fast-json-patch is the culprit. I updated our version to use 3.1.1 and patched that version. This removes the need for our 3.1.0 patch. Note that a override was also required because technically smart-transaction-controller has 3.1.0 shipped with it but we use 3.1.1 as our base for the patch and specify that anything requesting 3.1.0 should use 3.1.1. If you use `yarn info -A -R --dependents fast-json-patch` you should only see 3.1.1 and 2.2.1 in the output. Note that the 2.2.1 version is only a development dependency and does not trip the audit failure.

## Manual Testing Steps
1. pull this branch
2. run `yarn`
3. run `yarn info -A -R --dependents fast-json-patch`
4. verify only 3.1.1 and 2.2.1 are included and that both are patched.
5. run `.circleci/scripts/yarn-audit.sh`
6. see no failure.

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/.iyarc
+++ b/.iyarc
@@ -5,3 +5,13 @@ GHSA-257v-vj4p-3w2h
 # it is a false positive. The offending version of 'ws' that is installed is
 # 7.1.1 and is included only via remote-redux-devtools which is a devDependency
 GHSA-6fc8-4gx4-v693
+
+# yarn npm audit reports on a fast-json-patch version < 3.1.1 but due to patch
+# resolution, the only version of fast-json-patch that we use is 3.1.1. We also
+# have 2.2.1 installed but it is a dev only dependency. The "violation" reports
+# smart-transacton-controller as the culprit but if you run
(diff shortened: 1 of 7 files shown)


pedronfigueiredo (dismissed):

adonesky1 (dismissed):

jpuri (approved):

Gudahtt (approved):
LGTM!

Disappointing that we're getting this false-positive. Oh well. The lockfile is pretty confusing for this entry, it seems like the requirement for `^3.1.0` is unfilled. Must be because of the fact that it's patched. I guess Yarn v3 patches lead to confusing lockfiles. With Yarn v1 the patch was encoded in the lockfile, but here it looks like the resolution of version to patch isn't anywhere but the manifest.

PR #17080, opened 2023-01-04
add an extra identifier on anonymized duplicate events

brad-decker:
## Explanation
in #16267 we shifted to using a deterministic message id for actions that originate from the UI via calls to the background process, and therefore have an 'actionId' associated, or if they use a uniqueIdentifier. As a result only one of our two events were being tracked for any events that used one of these two pathways. This PR adds a new property to the rawPayload object expected by buildEventPayload, called 'isDuplicateAnonymizedEvent' which will append a '0x000' modifier to the messageId without modifying the fragment name at all.

The reason this happened is not because our code somehow prevented the action from firing but rather that segment treats duplicate message ids as the same event and keeps the first, discarding future events with the same id. 

## Manual Testing Steps
1. Add the development segment write key to your .metamaskrc
2. Run `yarn start`
3. Make sure metametrics is enabled.
4. Inspect the background process.
5. Login to segment
6. View the development metamask property and visit the debugger
7. Make a send transaction.
8. Inspect network tab and see both events are fired

*On develop*
1. In debugger see only one event shows up for each "transaction" event such as "Transaction Submitted"

*On this branch*
1. In debugger see both events are tracked.

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/app/scripts/controllers/metametrics.js
+++ b/app/scripts/controllers/metametrics.js
@@ -34,13 +34,16 @@ const defaultCaptureException = (err) => {
 
 // The function is used to build a unique messageId for segment messages
 // It uses actionId and uniqueIdentifier from event if present
-const buildUniqueMessageId = (args) => {
+const buildUniqueMessageId = (args, isDuplicateAnonymizedEvent = false) => {
   let messageId = '';
   if (args.uniqueIdentifier) {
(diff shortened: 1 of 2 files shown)


danjm (approved):
LGTM

jpuri (approved):
👍

PR #17092, opened 2023-01-06
Added storybook check to CI

### LINKED ISSUE #15760: Add storybook check to CI

NidhiKJha:
This PR is to add the storybook check for stories

### Changes:
1. Delete the `id` field from all the stories.js files. Stories won't be relying on `filename/id` anymore. 
2. Currently URLs are generated based on the existing ids but since the` ids ` are gone, new story URLs are generated based on the title of each story.
3. For the CI, we are generating the URL from `stories.json`. 
4. Updated the New URLs in the docs (README.mdx) files.

Ref: https://github.com/MetaMask/metamask-extension/pull/16721


.circleci/config.yml
```diff
@@ -496,6 +501,16 @@ jobs:
           name: Verify locales
           command: yarn verify-locales --quiet
 
+  test-storybook:
+    executor: node-browsers
```
Gudahtt, line 505: It looks like this job is failing with the error code 137, which often means an out-of-memory error.

I'd recommend using the larger instance size for this job:

```suggestion
    executor: node-browsers-medium-plus
```

.circleci/config.yml
```diff
@@ -496,6 +501,16 @@ jobs:
           name: Verify locales
           command: yarn verify-locales --quiet
 
+  test-storybook:
+    executor: node-browsers-medium-plus
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - run:
```
Gudahtt, line 510: I see that there is an error in CI about storybook not being installed. This should have been handled by the install script normally, but I'm guessing it's installing browsers in a directory not cached between jobs.

I'd recommend adding the install step here:

```suggestion
      - run:
          name: Install Playwright browsers
          command: yarn dlx playwright install
      - run:
```

.storybook/2.DOCUMENTATION.stories.mdx
```diff
@@ -59,7 +59,7 @@ export default {
   // The `title` effects the components tile and location in storybook
   // It should follow the folder structure location of the component. Don't use spaces.
   title: 'Components/UI/Button',
-  id: __filename, // The file name id is used to track what storybook files have changed in CI
+ // The file name id is used to track what storybook files have changed in CI
```
georgewrmarshall, line 62: I think we can probably remove this comment altogether?
```suggestion
```

georgewrmarshall (commented):
Looking good! I've created another issue to update internal links once this has been merged https://github.com/MetaMask/metamask-extension/issues/17110

.storybook/2.DOCUMENTATION.stories.mdx
kingdutch001: When #17092 is merged it will change all storybook URLs and will essentially break any link references to those stories. We will need to update internal and external links. This ticket is to update all internal links inside of the repo.

georgewrmarshall (commented):
Looks good! Make sure to create issues for the stories that you've updated to `stories-to-do.js` so we can track them 👣  You can use the other tickets as a template https://github.com/MetaMask/metamask-extension/issues/16820

NidhiKJha:
@SocketSecurity ignore playwright@1.29.2
@SocketSecurity ignore playwright-core@1.29.2

darkwing (dismissed):

darkwing (approved):

brad-decker (approved):
LGTM

package.json
```diff
@@ -302,6 +304,7 @@
     "nonce-tracker": "^1.0.0",
     "obj-multiplex": "^1.0.0",
     "pify": "^5.0.0",
+    "playwright": "^1.29.2",
```
Gudahtt, line 307: Could you move both of these from `dependencies` to `devDependencies`? We only use `dependencies` for things that end up in the built extension

development/highlights/storybook.js
```diff
@@ -65,27 +70,22 @@ async function getLocalDependencyList(filename) {
...
 
 /**
- * Remove punctuation and illegal characters from a story ID.
- * See:
- * https://gist.github.com/davidjrice/9d2af51100e41c6c4b4a
- * https://github.com/ComponentDriven/csf/blame/7ac941eee85816a4c567ca85460731acb5360f50/src/index.ts
- *
- * @param {string} string - The string to sanitize.
- * @returns The sanitized string.
+ * @param {fileName} string - The fileName to get the story id.
```
Gudahtt, line 78: Nit: Could we add a description here as well? Ideally all of our new functions would have a description.

We require this with a lint rule in other repos, but it hasn't been enabled in the extension yet due to the huge backlog of descriptions we'd have to add.

```suggestion
 * Get the ID for a story file.
 *
 * @param {fileName} string - The filename of the story file.
```

ui/components/app/app-header/app-header.stories.js
```diff
@@ -1,9 +1,10 @@
 import React from 'react';
 import AppHeader from '.';
 
+// eslint-disable-next-line import/no-anonymous-default-export
```
Gudahtt, line 4: Is this required? This rule should already be disabled for `*.stories.js` files.

```suggestion
```

ui/pages/token-allowance/token-allowance.stories-to-do.js
```diff
@@ -1,9 +1,10 @@
 import React from 'react';
 import TokenAllowance from './token-allowance';
 
+// eslint-disable-next-line import/no-anonymous-default-export
```
Gudahtt, line 4: I don't think we need this one either
```suggestion
```

Gudahtt (approved):
Great work!

I had a couple of suggestions, most notably moving the new dependencies from `dependencies` into `devDependencies`. But that can be done in a follow-up (in the interest of getting this merged sooner to avoid conflicts)

package.json
NidhiKJha: Thanks for the review Mark, I will create a follow up PR to add these changes :)

ui/pages/token-allowance/token-allowance.stories-to-do.js
NidhiKJha: This one is needed to address the lint error. Since it's a broken story and we changed it from .stories.js to .stories-to-do.js. We will remove it once this story will be fixed

ui/components/app/app-header/app-header.stories.js
NidhiKJha: Updated in [17344](https://github.com/MetaMask/metamask-extension/pull/17344)

development/highlights/storybook.js
NidhiKJha: Updated in [17344](https://github.com/MetaMask/metamask-extension/pull/17344)

PR #17120, opened 2023-01-10
bump `@metamask/smart-transactions-controller` version

adonesky1:
## Explanation
Bumps `@metamask/smart-transactions-controller` version to `v3.1.0` - a version which replace use of full `@metamask/controllers` repo with packages from `@metamask/core-monorepo`. This change allows us to reduce our unused code burden since `@metamask/controllers` was bringing lots of code we don't currently use into our bundle (the controllers version of `KeyringController`, with many additional dependencies, as one example). 

Bundle size reduced by .8MB :
Screenshot 2023-01-12 at 10 46 17 AM

This change should entail no functional changes.

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -1012,9 +1012,17 @@ export default class MetamaskController extends EventEmitter {
     });
     this.smartTransactionsController = new SmartTransactionsController(
       {
-        onNetworkStateChange: this.networkController.store.subscribe.bind(
-          this.networkController.store,
-        ),
+        onNetworkStateChange: (cb) => {
+          this.networkController.store.subscribe((networkState) => {
+            const modifiedNetworkState = {
+              ...networkState,
+              providerConfig: {
+                ...networkState.provider,
+              },
+            };
+            return cb(modifiedNetworkState);
+          });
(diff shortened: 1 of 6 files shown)


app/scripts/metamask-controller.js
```diff
@@ -1012,9 +1012,17 @@ export default class MetamaskController extends EventEmitter {
...
     this.smartTransactionsController = new SmartTransactionsController(
       {
-        onNetworkStateChange: this.networkController.store.subscribe.bind(
-          this.networkController.store,
-        ),
+        onNetworkStateChange: (cb) => {
+          this.networkController.store.subscribe((networkState) => {
+            const modifiedNetworkState = {
+              ...networkState,
+              providerConfig: {
```
adonesky1, line 1019: the `smart-transaction-controller` types the networkState using a type from `@metamask/network-controller` (from `@metamask/core-monorepo`) that was recently updated to expect property `providerConfig` in place of `provider`: https://github.com/MetaMask/core/blob/main/packages/network-controller/CHANGELOG.md?plain=1#L11

danjm (approved):

jpuri (approved):

PR #17127, opened 2023-01-10
Remove dead addressbook migration code

adonesky1:
## Explanation
This removes a block of code originally introduced to (per Mark's comment [here](https://github.com/MetaMask/core/pull/1011#discussion_r1044838666)) "to accommodate a specific update for metamask-extension, where we went from `chainId` being optional for custom RPC endpoints to it being required, and where we switched from keying this address book state by `chainId` instead of `networkId`". 

As Mark goes on to say: "That was over two years ago. The only wallets that old will be dormant ones. For the tiny subset of users that are starting up a wallet that has been shut down for multiple years, I think we can consider the possibility of losing addresses to be an acceptable risk. Even at the time of the migration this was an edge case."

This PR simply removes all the logic to perform this migration. Additionally it removes the `updateRPC` method (where this code lived) and replaces it's use with a slightly modified `addToFrequentRpcList` method (now called `upsertToFrequentRpcList`). This change reflects the broader context of this PR which is to ease the process of migrating the storage and management of network configurations from the `PreferencesController` to the `NetworkController`, which is tracked here: https://github.com/MetaMask/core/issues/969

## Manual Testing Steps
Add, Update and Remove a custom network and check that these are working as expected.

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/app/scripts/controllers/preferences.js
+++ b/app/scripts/controllers/preferences.js
@@ -398,7 +398,7 @@ export default class PreferencesController {
    * @param {string} [nickname] - Nickname of the selected network.
    * @param {object} [rpcPrefs] - Optional RPC preferences, such as the block explorer URL
    */
-  addToFrequentRpcList(
+  upsertToFrequentRpcList(
     rpcUrl,
     chainId,
     ticker = 'ETH',
--- a/app/scripts/controllers/preferences.js
+++ b/app/scripts/controllers/preferences.js
@@ -1,12 +1,7 @@
-import { strict as assert } from 'assert';
 import { ObservableStore } from '@metamask/obs-store';
 import { normalize as normalizeAddress } from 'eth-sig-util';
 import { ethers } from 'ethers';
-import log from 'loglevel';
-import {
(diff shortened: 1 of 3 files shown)


app/scripts/controllers/preferences.js
```diff
@@ -81,7 +76,6 @@ export default class PreferencesController {
     this.store = new ObservableStore(initState);
     this.store.setMaxListeners(12);
     this.openPopup = opts.openPopup;
-    this.migrateAddressBookState = opts.migrateAddressBookState;
```
Gudahtt, line 84: It looks like this option is no longer used, but it's still being passed in during construction. This option can now be removed in `metamask-controller.js`, and the function itself can be deleted.

adonesky1: done [here](https://github.com/MetaMask/metamask-extension/pull/17127/commits/7dd9f4b92f9c1ae722b5c65978e0b6edc5100325)

Gudahtt (approved):
LGTM!

jpuri (approved):

PR #17133, opened 2023-01-11
Copy update for Metamask fee on Swaps

nikoferro:
## Explanation

Small change that updates the copy displayed on the tooltip that explains the Metamask fee on Swaps

## Manual Testing Steps

- Build a swap request
- When reviewing the quote scroll to the bottom
- Right on top of the Swap button, there is a message about the fee that includes a tooltip
- Hover over the tooltip and the new message should be displayed

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -3829,7 +3829,7 @@
     "message": "MetaMask fee"
   },
   "swapMetaMaskFeeDescription": {
-    "message": "We find the best price from the top liquidity sources, every time. A fee of $1% is automatically factored into this quote.",
+    "message": "The fee of $1% is automatically factored into this quote. You pay it in exchange for a license to use MetaMask's liquidity provider information aggregation software.",
     "description": "Provides information about the fee that metamask takes for swaps. $1 is a decimal number."
   },
   "swapNQuotesWithDot": {


app/_locales/en/messages.json
```diff
@@ -3829,7 +3829,7 @@
     "message": "MetaMask fee"
   },
   "swapMetaMaskFeeDescription": {
-    "message": "We find the best price from the top liquidity sources, every time. A fee of $1% is automatically factored into this quote.",
+    "message": "The fee of $1% is automatically factored into this quote. You pay it in exchange for a license to use MetaMask's liquidity provider information aggregation software.",
```
pedronfigueiredo, line 3832: There are other localization files with outdated translations for this copy. You can remove them, so they are translated again.

jpuri: Not 100% sure but I think pipeline will still pick up this change.

brad-decker: Yeah it'll get picked up. ANy changes get picked up at present. However, a decision might want to be made regarding whether we want the old translations to show up for those locales or if we'd prefer to show the english version. If we'd prefer the latter then what @pedronfigueiredo suggested is the right approach. IF its okay for their to be a discrepancy between the languages for a short period (possibly, depending on release cadence and translation turnover rate) we can merge as is. I'm approving the PR for now

brad-decker (approved):

app/_locales/en/messages.json
pedronfigueiredo: ok thanks, @jpuri, and @brad-decker, for clearing this up for me.

pedronfigueiredo (approved):

app/_locales/en/messages.json
forestdiggs-consensys: Thanks for the info on localization - we should still move forward with getting this copy updated, even if there's a temporary discrepancy between languages. Just to make sure I'm understanding correctly - the translation process will happen automatically & there's no more action Nico or I need to take?

brad-decker: Correct, @forestdiggs-consensys -- but also just want to be clear that if the translation process takes longer than the release cycle that will include this that other languages will still be displaying the old translation. I'm waiting for the tests to rerun and pass then we can merge

PR #17149, opened 2023-01-11
Convert Transaction constants to typescript

brad-decker:
## Explanation
Furthers our goal of converting the extension codebase to TypeScript. This PR is primarily focused on converting the shared/constants/transaction.js file to typescript. There were a number of stylistic changes that resulted in lots of updates needed throughout the codebase.

1. Adopted the UpperCamelCase, singular convention for enums that is being used in other TypeScript projects. IE TRANSACTION_TYPES.SIMPLE_SEND -> TransactionType.simpleSend
2. We previously had a JSDoc type definition, an object for the constant and another JSDoc type definition for the values of that object. TypeScript understands that enums are any one of the values of the enum so we now only need an enum itself. In JS enums act as objects so code that previously relied upon the constant being an object (AKA Object.values(TRANSACTION_TYPES) will still work.
3. I had to add the .ts file extension to our verify-locales script so that it could properly infer that some of our constants are being used as keys passed to `t()` (which is something we want to stop doing and should fix but not here). 
4. There was a rebase error at some point with the `yarn test:unit:jest` script that was resolved in this PR
5. There was a missed conversion of ASSET_TYPES.COLLCTIBLE to ASSET_TYPES.NFT in send-asset-row which I fixed and resulted in a snapshot change there. 
6. I chose to rename the TransactionStatus component to TransactionStatusLabel to avoid conflicts with the TransactionStatus enum. I also renamed the smart-transaction-status page component name to SmartTransactionStatusPage, but kept the route name and file structure unchanged. 

## Manual Testing Steps
1. Perform transactions! Lots of them. And make sure their representation in the UI and in metrics remains unchanged.

## Pre-merge author checklist

- [x] I've clearly explained:


jpuri:
Hey @brad-decker, changes look great. I have one query, uppercase if more intuitive for constants than camelcase, reason for preferring camelcase.

brad-decker:
TypeScript specifies enums should be upper camel case and this is the pattern adopted elsewhere in MetaMask such as the core library.

jpuri (dismissed):
Thanks for the effort 🚀

jpuri (dismissed):

ui/ducks/swaps/swaps.js
```diff
@@ -986,14 +986,14 @@ export const signAndSendSwapsSmartTransaction = ({
...
-          type: TRANSACTION_TYPES.SWAP,
+          type: TransactionType.swap,
         }),
       );
       if (approvalTxUuid) {
         await dispatch(
           updateSmartTransaction(approvalTxUuid, {
             origin: ORIGIN_METAMASK,
-            type: TRANSACTION_TYPES.SWAP_APPROVAL,
+            type: TransactionType.swap_APPROVAL,
```
pedronfigueiredo, line 996: in `shared/constants/transaction.ts` this is correctly typed `swapApproval`

ui/selectors/transactions.js
```diff
@@ -303,8 +303,7 @@ export const nonceSortedTransactionsSelector = createSelector(
           // known outside of the user's local MetaMask and the nonce
           // associated will be applied to the next.
           isEphemeral:
-            status === TRANSACTION_STATUSES.FAILED &&
-            txReceipt?.status !== '0x0',
+            status === TransactionStatus.failed && txReceipt?.status !== '0x0',
```
pedronfigueiredo, line 306: we probably don't have a type for this yet, but just flagging it just in case. Does it make sense to put this in a constant?

pedronfigueiredo:
beast of a PR, thanks Brad!

ui/ducks/swaps/swaps.js
brad-decker: Goooood catch.

ui/selectors/transactions.js
brad-decker: We should explore that. It would be really good to have a strict literal string type for 0x0.

brad-decker:
@pedronfigueiredo @jpuri ready for another look

shared/constants/transaction.ts
```diff
@@ -0,0 +1,426 @@
...
+   * transaction is signed on an external device. Otherwise signing happens
+   * transparently to users.
+   */
+  pending = 'pending',
+}
+
+/**
+ * With this list we can detect if a transaction is still in progress.
+ */
+export const IN_PROGRESS_TRANSACTION_STATUSES = [
```
DDDDDanica, line 158: Question: should this be `InProgressTransactionStatuses` to be consistent with the rest type names?

brad-decker: Thanks for  asking @DDDDDanica :) 

So the 'UpperCamelCase' convention is for TypeScript enum types only. The above is is a constant array so I kept it with our previous 'UPPER_SNAKE_CASE' convention.

DDDDDanica: Makes sense. Thank for the explanation  🦾

pedronfigueiredo (approved):

DDDDDanica (approved):

PR #17155, opened 2023-01-12
fix(mme-14830): fix consistent tab when switching languages

### LINKED ISSUE #14830: [Bug]: When Language is not English, I land on the wrong tab when I've performed a tx
### Describe the bug

**Problem**: when I have set a language which is not English, after performing a transaction I land in the Home page with the Assets tab open, while when I have English language selected, I land on the Activity tab so I can see the recent transaction performed.

https://user-images.githubusercontent.com/54408225/171588706-df4721c3-1d7f-4c65-bdc8-55831bbc63a3.mp4

Suggested Solution: when I'm using other languages, I should also land into the Home/Activity tab after performing a transaction, in the same way it is done when having English language selected.

### Steps to reproduce

1. Unlock Metamask
2. Go to Settings
3. Change language to another different than English
4. Perform a transaction
5. Notice that you land on the Home/Assets page instead of Home/Activity page

_No response_

### Version

10.14.6

### Build type

_No response_

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

DDDDDanica:
## Explanation

**Fixes #14830**

The `<Tab />` component is currently taking a translated `string` as unique key to render the DOM. Hence when we switch language and land in home page again, it is trying to look for the responding `string` from previous language from the new language strings set. 
The fix will be to have `Tab` component to take the `key` in `message.json` file and map by key. Each time a new action `setDefaultHomeActiveTabName` is dispatched, it will pass the `key` (eg: `asset`, `activity`) as props to child component.

*Moreover, for the convenience of reviewing, what I have done in this fix are:* 
-   Fix the bug, which comes from Tabs & Tab component, to use key(which is the same as language file message key) instead of name as index
-   Change the props name for Tabs and related function signature to make it more clear that we are using key instead of string (name)
-   Adapt related storybook

*What I didn’t do*
-   No test added, as there’s no test for home page :(
-   Didn’t change the state and action name, as it will reach too many files and make MR too complex to review, I would like to focus on bug fix this time

### Before

https://user-images.githubusercontent.com/12678455/212040399-6a7c2587-9490-4215-8713-298c74aa89b2.mp4

### After

https://user-images.githubusercontent.com/12678455/212382192-392aa6c3-ad14-4742-b6fd-244209eaed31.mov

## Manual Testing Steps

1.  Unlock Metamask
2.  Go to Settings
3.  Change language to another different than English
4.  Perform a transaction
5.  Notice that you land on the Home/Assets page instead of Home/Activity page

## Pre-merge author checklist

- [x] I've clearly explained:


DDDDDanica:

ui/components/ui/tabs/tabs.component.js
```diff
@@ -47,11 +47,11 @@ export default class Tabs extends Component {
...
     return React.Children.map(this._getValidChildren(), (child, index) => {
-      const tabName = child?.props.name;
+      const tabKey = child?.props.tabKey;
       return (
         child &&
         React.cloneElement(child, {
-          onClick: (idx) => this.handleTabClick(idx, tabName),
+          onClick: (idx) => this.handleTabClick(idx, tabKey),
           tabIndex: index,
           isActive: numberOfTabs > 1 && index === this.state.activeTabIndex,
```
danjm, line 56: It's possible the e2e test failure is because `t` is undefined in `tab.component.js` when a tab is rendered via this method

danjm (changes requested):
Nice work on this PR. I think the change to the tabs components might not be necessary? Because the component uses whatever name is passed to it, as long as the `defaultHomeActiveTabName` set in redux state matches  the name passed to the `Tab` component, I think that the bug will be resolved. We could remove those changes to simplify this PR.

ui/components/ui/tabs/tab/tab.component.js
```diff
@@ -25,7 +26,7 @@ const Tab = (props) => {
         onClick(tabIndex);
       }}
     >
-      <button>{name}</button>
+      <button>{t(tabKey)}</button>
```
danjm, line 29: Unfortunately, we try to avoid doing translations like this, where the string is passed by reference as a variable, as opposed to directly passing a string literal. The reason for this (which you would have had no way of knowing because this is not documented) is that we have tooling which detects when translations are unused, or when they are undefined in in our translation file, but that tooling only works if we explicitly pass string literals to the translation function

DDDDDanica: Actually the data flow is rendered like this, hope this makes sense: 

- We are in "Activity" tab, current `defaultHomeActiveTabName` is `Activity`
- We switch language, and close the settings
- `defaultHomeActiveTabName` is still `Activity` while the `name` is `activité` for example

=> this results in in consistency and wrong `activeTabIndex`

2 ways i think could be a fix
1. use the key of language as tabKey and translation in the end, like the solution i provided, which doesn't work actually
2. reverse the language by key and use the key in tabs component to compare, which is not very possible.
Open to more thoughts !

ui/components/ui/tabs/tabs.component.js
DDDDDanica: Good catch !!

danjm:
We discussed this in slack. I think this can be fixed without having to add a key prop to the tabs component. Locally, I think I got it working with just this change:

Screenshot from 2023-01-17 10-02-52

I pushed these changes here so you can try. https://github.com/MetaMask/metamask-extension/compare/fix-activity-tabs?expand=1

Maybe there are edge cases I am missing?

DDDDDanica:
As discussed in slack, the variable `defaultHomeActiveTabName` is storing language specific string, and didn't get updated when new locale is settled. When testing only making changes for dispatch handler, the `<Tabs />` will not find a match between tab name and `defaultHomeActiveTabName`,  you can find the testing below: 
https://user-images.githubusercontent.com/12678455/212989458-61cc3427-7328-435b-8b74-8725d3e9bced.mov

Hence after discussion, we decide to introduce this new `tabKey` for Tab component, furthermore, Tabs will compare between keys instead of value 🙏

danjm (dismissed):
LGTM! Nice work

brad-decker (approved):
Looks awesome! Great work :)

danjm (approved):

PR #17163, opened 2023-01-12
update build system policy override config

adonesky1:
Trying this again - [previous attempt](https://github.com/MetaMask/metamask-extension/pull/14902).

We need to update the override config file so that platform dependent policy changes don't create conflicts each time the policy files are generated on a different OS.

I'm not a lavamoat expert so please set me straight if this isn't the right way to go about it @kumavis @brad-decker

### DIFF
--- a/lavamoat/build-system/policy-override.json
+++ b/lavamoat/build-system/policy-override.json
@@ -256,6 +256,299 @@
       "globals": {
         "process": true
       }
+    },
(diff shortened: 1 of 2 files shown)


kumavis (approved):

kumavis:
at present this is the correct solution but i agree it sucks

brad-decker (approved):

kumavis:
working on a devx improvement here https://github.com/LavaMoat/LavaMoat/issues/452

PR #17166, opened 2023-01-12
[E2E FIX] Fixed enums.js 

bowensanders:
The enums.js file that points to the correct version of test-snaps was fixed to point at the correct version.


bowensanders:
Fixed by another PR

Issue #16962, opened 2022-12-14
Write comprehensive unit tests for the network controller


Gudahtt:
We should write comprehensive unit tests for the network controller prior to merging it with the mobile network controller. This would help reduce the risk of regressions.

The network controller already has many tests, but they are incomplete and they don't follow our latest best practices for writing unit tests. We should ensure the unit tests follow the best practices outlined here: https://github.com/MetaMask/controllers/issues/413

PR #17199, opened 2023-01-13
Refactor: Construct the network controller within each test

Gudahtt:
The network controller is now constructed within each network controller unit test, rather than in the `beforeEach`. This allows us to customize the constructor options in each test, which some planned future tests will require.

The controller is constructed with a helper function that also handles calling `destroy` after each test, even if the test failed. This helps to prevent tests from affecting each other.

This relates to https://github.com/MetaMask/metamask-extension/issues/16962

## Pre-merge author checklist

- [ ] I've clearly explained:
  - [ ] What problem this PR is solving
  - [ ] How this problem was solved
  - [ ] How reviewers can test my changes
- [ ] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [ ] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
- [ ] PR is linked to the appropriate GitHub issue
- [ ] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.

In this case, a QA Engineer approval will be be required.

PR #17250, opened 2023-01-16
Setup network controller mocks per-test

Gudahtt:
The network controller unit test network mocks are now setup for each test. This makes modifying network behavior on a per-test basis easier, and makes it more clear which test relies upon which mocks.

This relates to https://github.com/MetaMask/metamask-extension/issues/16962

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/app/scripts/controllers/network/network-controller.test.js
+++ b/app/scripts/controllers/network/network-controller.test.js
@@ -1,5 +1,6 @@
 import nock from 'nock';
 import { deferredPromise } from '../../lib/util';
+import { NETWORK_TYPES } from '../../../../shared/constants/network';
 import NetworkController, { NETWORK_EVENTS } from './network-controller';
 
 /**
@@ -80,26 +81,32 @@ async function withController(...args) {
   }
 }
 
+/**
+ * Setup Nock mocks for the block tracker.
+ *
+ * @param {object} options - Options.
+ * @param {NETWORK_TYPES} options.networkType - The network type to mock.
+ * @param {object} options.block - The mock block to return.
+ */
+function setupMockRpcBlockResponses({
+  networkType = NETWORK_TYPES.RPC,
+  block = BLOCK,
+} = {}) {
+  if (networkType === NETWORK_TYPES.RPC) {
+    nock('http://localhost:8545')
+      .persist()
+      .post(/.*/u)
+      .reply(200, () => JSON.stringify(constructSuccessfulRpcResponse(block)));
+  } else {
+    nock(`https://${networkType}.infura.io`)
+      .persist()
+      .post(`/v3/${defaultControllerOptions.infuraProjectId}`)
+      .reply(200, () => JSON.stringify(constructSuccessfulRpcResponse(block)));
+  }
+}
+
 describe('NetworkController', () => {
   describe('controller', () => {
-    let latestBlock;
-
-    beforeEach(() => {
-      latestBlock = BLOCK;
-      nock('http://localhost:8545')
-        .persist()
-        .post(/.*/u)
-        .reply(200, () =>
-          JSON.stringify(constructSuccessfulRpcResponse(latestBlock)),
-        );
-      nock('https://mainnet.infura.io')
-        .persist()
-        .post('/v3/foo')
-        .reply(200, () =>
-          JSON.stringify(constructSuccessfulRpcResponse(latestBlock)),
-        );
-    });
-
     afterEach(() => {
       nock.cleanAll();
     });
@@ -107,6 +114,7 @@ describe('NetworkController', () => {
     describe('#provider', () => {
       it('provider should be updatable without reassignment', async () => {
         await withController(async ({ controller }) => {
+          setupMockRpcBlockResponses();
           await controller.initializeProvider();
           const providerProxy =
             controller.getProviderAndBlockTracker().provider;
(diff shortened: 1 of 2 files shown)


Gudahtt:
This depends upon #17199

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -52,147 +55,202 @@ const BLOCK = {
...
+          await controller.initializeProvider();
+          const { blockTracker } = controller.getProviderAndBlockTracker();
+          // The block tracker starts running after a listener is attached
+          blockTracker.addListener('latest', () => {
+            // do nothing
+          });
+          expect(blockTracker.isRunning()).toBe(true);
 
-        networkController.destroy();
+          controller.destroy();
```
BelfordZ, line 154: is this already handled by [the `controller.destroy` from inside `withController`](https://github.com/MetaMask/metamask-extension/pull/17250/files#diff-d4e78728faadb83c4ed8c0d5b5b580b026993a8988156b27bf7f41c4b5217af4R87)?

BelfordZ: ah nvm this is testing that destroy works as specified. disregard.

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -157,6 +168,8 @@ describe('NetworkController', () => {
     describe('#setProviderType', () => {
       it('should update provider.type', async () => {
         await withController(async ({ controller }) => {
+          setupMockRpcBlockResponses();
```
mcmire, line 171: I feel like we could be a bit more explicit in these tests when we mock the RPC requests. Taking this test as an example, it seems a bit inconsistent that we are specifying that we want to mock mainnet on line 174 but not specifying the network type on line 173. I realize that the default network type for this function is localhost, but I feel like one reason this function has a default is because NetworkController has the same default. In this case and similar cases, it seems like we shouldn't rely on this presumption. I think long-term NetworkController should require consumers to explicitly specify which kind of network they want to connect to.

What do you think about?

``` suggestion
        await withController(
          {
            provider: {
              type: NETWORK_TYPES.RPC,
              // ... anything else that is required ...
            }
          },
          async ({ controller }) => {
            setupMockRpcBlockResponses({ networkType: NETWORK_TYPES.RPC });
            setupMockRpcBlockResponses({ networkType: NETWORK_TYPES.MAINNET });
            await controller.initializeProvider();
            controller.setProviderType('mainnet');
            const { type } = controller.getProviderConfig();
            expect(type).toStrictEqual('mainnet');
          });
        )
```

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -85,33 +88,40 @@ async function withController(...args) {
...
+ * @param {object} options.block - The mock block to return.
+ */
+function setupMockRpcBlockResponses({
+  networkType = NETWORK_TYPES.RPC,
+  block = BLOCK,
+} = {}) {
+  if (networkType === NETWORK_TYPES.RPC) {
+    nock('http://localhost:8545')
+      .persist()
+      .post(/.*/u)
```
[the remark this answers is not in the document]
Gudahtt: Good idea. This seems worthwhile. It would be confusing to get a block in response to a method that does not expect a block.

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -85,33 +88,40 @@ async function withController(...args) {
...
+ * @param {NETWORK_TYPES} options.networkType - The network type to mock.
+ * @param {object} options.block - The mock block to return.
+ */
+function setupMockRpcBlockResponses({
+  networkType = NETWORK_TYPES.RPC,
+  block = BLOCK,
+} = {}) {
+  if (networkType === NETWORK_TYPES.RPC) {
+    nock('http://localhost:8545')
+      .persist()
```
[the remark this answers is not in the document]
Gudahtt: Hmm, maybe. In some cases it may be both non-deterministic and irrelevant to the purpose of the test, in which case `persist` seems appropriate. But I'm not sure whether that's the case here, I can try it out.

app/scripts/controllers/network/network-controller.test.js
Gudahtt: I was trying to make these setup helpers concise, so they don't distract from the purpose of each test. i.e. does it matter what the default is?

Maybe it does in this example actually, to show that the default is different than the network we're switching to. But in other cases I'm not so sure.

I can at least update `setupMockRpcBlockResponses` to explicitly reference the default network of the network controller, so it's clear why that is the default and so that it remains in-sync.

app/scripts/controllers/network/network-controller.test.js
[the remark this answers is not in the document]
Gudahtt: Resolved here: df2000a856c9887c1e7016eddfb6809a123d8819

app/scripts/controllers/network/network-controller.test.js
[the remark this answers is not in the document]
Gudahtt: It seemed to work out OK! Done here: a54104040733a6c7f1348bf648bc1bdb1a58d31e

app/scripts/controllers/network/network-controller.test.js
Gudahtt: I've implemented my suggestion here: 055ec6ab3fe3ed4282696c474bed47ec8005b5d9
Let me know if that makes sense

mcmire (approved):

brad-decker (approved):
LGTM

Issue #6290, opened 2019-03-12
Clean up currency conversion code


danjm:
Below is copied from https://github.com/MetaMask/metamask-extension/pull/4895#issuecomment-408452748 Once we clear the blockers from the epic this is in, we should:

- [ ] Audit the codebase for all use of conversion type utilities, or conversions without utilities
- [ ] Design a minimum viable api
- [ ] write a suite of tests for existing methods that depend on the "conversion-util", do not mock it away
- [ ] research existing methods in other libraries (e.g. `web3.fromWei(web3.toWei('1', 'gwei'), 'ether')`)
- [ ] Between the existing base "conversion-util" and relevant methods in other libraries, rewrite the base "conversion-util" to be very readable and user friendly
- [ ] In the spirit of the below comments, reduce code duplication with resepect to conversions throughout the codebase. Start by making a list of tasks, and then do them
- [ ] Write documentation for our base conversion utility and the convenience methods built on top of those 

We are going to forever have issues with needing to render numbers in different ways in different places.

It would be nice if we could make these transformations - of data from some value of cryptocurrency to some nicely formatted string for rendering - simpler, DRYer and less error prone.

For instance, after this PR we have:

```
      const fiatTransactionAmount = this.getFiatTransactionAmount()
      const roundedFiatTransactionAmount = roundExponential(fiatTransactionAmount)
      return formatCurrency(roundedFiatTransactionAmount, currentCurrency)
```
The three functions called here require 4 inputs: `tokenAmount, currentCurrency, conversionRate, contractExchangeRate`. `this.getFiatTransactionAmount()` passes these for inputs to a util which in turn calls the conversion util, which in turn calls a series of operations on a Bignumber. `roundExponential` the does some operations with Bignumber, conditional on the passed values size, and `currencyFormatter` with some other library. In the same component we have to do the same three function calls plus a function call to do some addition, which uses a util which uses the conversion util...

I guess it would be nice if we could get the desired result by passing the four params to one utility method (or maybe a component). And that utility be something that can be used everywhere we need to render currencies.

I suppose that was what I was aiming for with the conversion-util. However, apart from the issue of the code being obscure and inaccessible, it also seems that we still need a separation of concerns between operations on values for the sake of getting a correct value and operations on values for the sake of getting a string to be rendered in a desired format.

So probably some combination of an improvement of the conversion-util and then creating a new generalized utility, or maybe component, for the sake of correct rendering.

[...]

PR #17262, opened 2023-01-17
Complete level 36 migration of TypeScript by converting conversion.utils.js and deprecate old conversion util

### LINKED ISSUE #15463: [Bug]: I am able to input negative values to tx amount
### Describe the bug

**Problem**: when I am Sending a tx I can introduce negative numbers by clicking the down arrow of my keyboard. This eventually leads to a tx with 0 value. I guess input should be validated and if negative, we should not be able to proceed to the Confirmation Screen.

https://user-images.githubusercontent.com/54408225/182885986-d86de6f6-6e74-45c6-87fe-c662d5923904.mp4

### Steps to reproduce

1. Unlock MM
2. Click Send
3. Introduce any recipient
4. Click Amount and input a random number
5. Press down arrow key for a couple of times

_No response_

### Version

10.18.2

### Build type

_No response_

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

### LINKED ISSUE #16211: [Bug]: [MV3] e2e failure for Errors spec
### Describe the bug

**Problem**: the e2e test case errors.spec.js is failing with the following error

image

### Steps to reproduce

1. Build mv3 test yarn build:test:mv3
2. Run errors test yarn test:e2e:single test/e2e/tests/errors.spec.js --browser=chrome

_No response_

### Version

develop-oct14

### Build type

_No response_

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

### LINKED ISSUE #17211: Typescript Level 36
[Extension TypeScript Migration Status dashboard
](https://metamask.github.io/metamask-extension-ts-migration-dashboard/)
Level 36 Tasks:

- [ ] Translate app/scripts/lib/util.js 
- [ ] Translate shared/modules/conversion.utils.js

brad-decker:
## Explanation
This Pull request aims to complete the level 36 block of the ts-migration dashboard https://metamask.github.io/metamask-extension-ts-migration-dashboard/

This one was a large undertaking because conversions.util.js contained the dreaded 'conversionUtil' that had prolific use throughout the codebase. Usage was in many different places and forms:
1. Directly imports conversionUtil (which is a wrapper around the 'converter' method)
2. Uses one of the *many* conversions utils such as hexWEIToDecGWEI, addCurrencies, etc that composed the conversionUtil
3. Uses conversionUtil by way of any number of different helper files that made their own conversion helpers throughout the codebase.

### Instantiation of the value is not guaranteed to yield a BigNumber
```js
  let convertedValue = fromNumericBase
    ? toBigNumber[fromNumericBase](value)
    : value;

  if (fromDenomination) {
    convertedValue = toNormalizedDenomination[fromDenomination](convertedValue);
  }
```

If you do not supply `fromNumericBase` to this function, which has no default value, the base value that you provided will be used. A few lines below this the 'convertedValue' is used as if it is guaranteed to be a BigNumber.

### The value returned from this method is not guaranteed to be any specific type
```js
  if (toNumericBase) {
    convertedValue = baseChange[toNumericBase](convertedValue);
  }
  return convertedValue;
```

If you do not supply 'toNumericBase' you will get a BigNumber value returned. This isn't bad, except that your return value *type* is dependent on a prop passed to the method, and if you attempt to use a non BigNumber as a BigNumber we get the "Not a BigNumber" error.

### fromCurrency and toCurrency have *no impact on the operation* except to check if they are the same
```js
  if (fromCurrency !== toCurrency) {
    if (conversionRate === null || conversionRate === undefined) {
      throw new Error(
        `Converting from ${fromCurrency} to ${toCurrency} requires a conversionRate, but one was not provided`,
      );
    }
    let rate = toBigNumber.dec(conversionRate);
    if (invertConversionRate) {
      rate = new BigNumber(1.0).div(conversionRate);
    }
    convertedValue = convertedValue.times(rate);
  }
```
In many parts of our codebase I have seen conversionUtil called like this:
```js
const currency = conversionUtil(totalInHexWei, {
  fromNumericBase: 'hex',
  toNumericBase: 'dec',
  fromDenomination: 'WEI',
  fromCurrency: 'ETH',
  toCurrency: 'USD',
  conversionRate: 457.65,
});
```
Because I was never confident in my understanding of conversions of ETH to Fiat I assumed there was some magic here. NOPE!

On the first line of the snippet at the beginning of this section, we check if fromCurrency and toCurrency are the same, if they are it does nothing, otherwise it checks that conversionRate was supplied and throws an error if not. It then applies the conversion rate. The currency markers do nothing else at all. Instead we could have conditionally passed conversionRate if it made sense and use the existence of the conversionRate  to determine if it should be applied.

There's also another fun tidbit....

### fromDenomination and toDenomination make sense, but what happens if you omit one or the other?
There's nothing *majorly* wrong with these: 
```js
// Big Number Constants
const BIG_NUMBER_WEI_MULTIPLIER = new BigNumber('1000000000000000000');
const BIG_NUMBER_GWEI_MULTIPLIER = new BigNumber('1000000000');
const BIG_NUMBER_ETH_MULTIPLIER = new BigNumber('1');

// Setter Maps
const toBigNumber = {
  hex: (n) => new BigNumber(stripHexPrefix(n), 16),
  dec: (n) => new BigNumber(String(n), 10),
  BN: (n) => new BigNumber(n.toString(16), 16),
};
const toNormalizedDenomination = {
  WEI: (bigNumber) => bigNumber.div(BIG_NUMBER_WEI_MULTIPLIER),
  GWEI: (bigNumber) => bigNumber.div(BIG_NUMBER_GWEI_MULTIPLIER),
  ETH: (bigNumber) => bigNumber.div(BIG_NUMBER_ETH_MULTIPLIER),
};
```

Here, however,

```js
}) => {
 //Very first line of function
  let convertedValue = fromNumericBase
    ? toBigNumber[fromNumericBase](value)
    : value;

  if (fromDenomination) {
    convertedValue = toNormalizedDenomination[fromDenomination](convertedValue);
  }
// [.....apply the conversionRate as shown in the above snippet]
  if (toDenomination) {
    convertedValue = toSpecifiedDenomination[toDenomination](convertedValue);
  }
```
This method converts the currency value to ETH by default if you supply fromDenomination. SO:

```js
const v = conversionUtil(valueInHexWEI, { fromNumericBase: 'hex', fromDenomination: 'WEI', toNumericBase: 'dec'});
```

v is going to be the value in base 10 (decimal) of ETH. The problem with this is that the contract of it switching to ETH is baked into the code itself and not the props, which would be fine if the name of the method implied this. We do have methods that end in the name 'ToDecETH' which are more widely used for this operation, but under the hood they use conversionUtil. What makes worse is the existence of toDenomination. The existence of toDenomination *should imply that not including it would not convert anything*. That isn't the case. Furthermore if you use toDenomination without fromDenomination its assumed that the value you were using was in ETH already. How handy :) 

### Two different rounding operations occur
```js
  if (numberOfDecimals !== undefined && numberOfDecimals !== null) {
    convertedValue = convertedValue.round(
      numberOfDecimals,
      BigNumber.ROUND_HALF_DOWN,
    );
  }

  if (roundDown) {
    convertedValue = convertedValue.round(roundDown, BigNumber.ROUND_DOWN);
  }
```
I do not know why these options are different, but I assume there is a reason. I'm hoping someone will be able to point that out to me in code review. 

### They are unnecessarily complex to use/understand
```js
const addCurrencies = (a, b, options = {}) => {
  const { aBase, bBase, ...conversionOptions } = options;

  if (!isValidBase(aBase) || !isValidBase(bBase)) {
    throw new Error('Must specify valid aBase and bBase');
  }

  const value = getBigNumber(a, aBase).add(getBigNumber(b, bBase));

  return converter({
    value,
    ...conversionOptions,
  });
};
```

Usage:
```js
const value = addCurrencies(
  currencyA,
  currencyB,
  { 
    aBase: 10, // Notably not 'dec' like the converter method bc who needs consistency 
    bBase: 16, // Same but for 'hex'
    fromNumericBase: :shrug: // Do you know? It takes all options from conversionUtil. More in details
    toNumericBase: 'dec',
  }
);
```
So the first thing this method does is check that you supplied a valid 10 or 16 value for aBase and bBase, which uses the same values as the parseInt, toString, etc methods (10, 16) to determine if the string is decimal or hexadecimal. I like that, it makes sense and encourages the understanding of these parameters of those methods. 'hex' and 'dec' I think were used for onboarding/understanding purposes but they just map to 10 and 16 respectfully and you can't use the knowledge of them anywhere else (you'd need to use 10/16 bases for the other methods). 

Then it does something peculiar, it uses a helper method 'getBigNumber' and calls it to get the BigNumber representation of the a value then calls '.add' with the BigNumber representation of the B value. 

Then it supplies the value to the converter method with all the params you want to pass it that the converter method operates on. So compound all of the issues I wrote about the converter method with the following:
1. What happens if you supply 'fromNumericBase' to this method? Fun fact: 'BN' and undefined are valid values for fromNumericBase. 'BN' is for instances of BN or BigNumbers and undefined assumes its already a BigNumber. undefined is the correct value for fromNumericBase in the above example.
2. If you look only at the *usage* of `addCurrencies` can you tell what the order of operations is? Does it convert each value then add it together, or does it add it together then convert it. If I supply a conversionRate and fromCurrency/toCurrency what should I expect? 

Here's another fun example:

```js
const conversionGreaterThan = ({ ...firstProps }, { ...secondProps }) => {
  const firstValue = converter({ ...firstProps });
  const secondValue = converter({ ...secondProps });

  return firstValue.gt(secondValue);
};
```

And here is the most complex usage I can find in real code:
```js

  const customPriceSafe = conversionGreaterThan(
    {
      value: customGasPrice,
      fromNumericBase: 'hex',
      fromDenomination: 'WEI',
      toDenomination: 'GWEI',
    },
    { value: estimatedPrice, fromNumericBase: 'dec' },
  );

  return customPriceSafe;
}

```

Here we are attempting to compare a decimal value representing an amount in GWEI to a hex value representing an amount in WEI. The usage here is correct, but we can pass literally any propertiers of the converter to either A or B operand and it just becomes extremely confusing.

**TO BE CONTINUED**

# Solution
This Pull Request implements a new class in TypeScript called 'Numeric' which has a lot of similarities with BigNumber except that it has denomination manipulation built in as well as a helper method for applyConverionRate. Each chaining operator of this new utility returns a new instance of Numeric. So for example:

```js
const value = addCurrencies(
  currencyA,
  currencyB,
  { 
    aBase: 10, // Notably not 'dec' like the converter method bc who needs consistency 
    bBase: 16, // Same but for 'hex'
    fromNumericBase: :shrug: // Do you know? It takes all options from conversionUtil. More in details
    toNumericBase: 'dec',
  }
);
```

In this previous example from above the way you'd write the same thing using Numeric:
```js
const value = new Numeric(currencyA, 10)
  .add(new Numeric(currencyB, 16))
  .toBase(10)
  .toString();
```

```js

  const customPriceSafe = conversionGreaterThan(
    {
      value: customGasPrice,
      fromNumericBase: 'hex',
      fromDenomination: 'WEI',
      toDenomination: 'GWEI',
    },
    { value: estimatedPrice, fromNumericBase: 'dec' },
  );

  return customPriceSafe;
}

```

And this one would be expressed like:
```js
const customPriceSafe = new Numeric(customGasPrice, 16, EtherDenomination.WEI)
  .toDenomination(EtherDenomination.ETH)
  .greaterThan(new Numeric(estimatedPrice, 10));
```

As you can see its much easier to understand because each chain in the operation expresses the order of operations instead of being baked into the application code. 

Fixes #16211 
Fixes #15463
Fixes #17211 
Progresses #6290 

## Pre-merge author checklist

- [ ] I've clearly explained:


shared/modules/Numeric.ts
```diff
@@ -0,0 +1,621 @@
...
+ *
+ * @param value - The string to check
+ * @returns true if the string is a hexadecimal split by '.'
+ */
+function isDecimalHex(value: string): boolean {
+  const parts = value.split('.');
+  if (parts.length === 1) {
+    return false;
+  }
+  return parts.every((part) => isHexStringOrNegatedHexString(part));
```
pedronfigueiredo, line 39: this implementation will pattern match `-<hex num>.-<hex num>` as a decimal hex. Flagging to ensure it's intentional.

brad-decker: It is a safety net. I don't think its actually valid hex to do `hex`.`hex` but BigNumber uses that notation internally for hexadecimal containing a decimal point when stringified.... it seems... There should be no other cases where we get 'hex'.'hex' other than as a result of toString on a BigNumber which will always be re-interpretable by BigNumber during instantiation. We should add tests for this soon but as of right now I am of the mind that this is okay.

shared/modules/Numeric.ts
```diff
@@ -0,0 +1,621 @@
...
+/**
+ * All variations of isHexString from our own utilities and etherumjs-utils
+ * return false for a '-' prefixed hex string. This utility method strips the
+ * possible '-' from the string before testing its validity so that negative
+ * hex values can be properly handled.
+ *
+ * @param value - The string to check
+ * @returns true if the value is a hex string (negative or otherwise)
+ */
+function isHexStringOrNegatedHexString(value: string): value is string {
```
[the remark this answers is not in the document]
Gudahtt: For more information: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

shared/modules/Numeric.ts
```diff
@@ -0,0 +1,621 @@
+import { BigNumber } from 'bignumber.js';
```
Gudahtt, line 1: Would it be possible to break this PR up into stages, to make it easier to review?

e.g.
* Add this new `Numeric` module
* Migrating code from `conversion.utils.js` to `Numeric` in chunks
* Delete `conversion.utils.js`

That might make this less intimidating to review. And we can more effectively assign each chunk to the people most familiar with that code (e.g. we could ask the swaps team to just review the swaps changes).

shared/modules/Numeric.ts
```diff
@@ -0,0 +1,621 @@
...
+    );
+  }
+
+  isNegative() {
+    return this.value.isNegative();
+  }
+
+  isPositive() {
+    return this.isNegative() === false;
+  }
```
DDDDDanica, line 579: I was thinking, if `this.value` is 0, it would fall into `isNegative`, we should priority 0 in this situation~

brad-decker: Why would 0 be negative?

shared/modules/Numeric.ts
```diff
@@ -0,0 +1,621 @@
...
+ *
+ * @param value - The value to convert to a BigNumber
+ * @param numericBase - The numeric base of the underlying value
+ * @returns A BigNumber representation of the value
+ */
+function valueToBigNumber(value: string | number, numericBase: NumericBase) {
+  if (typeof value === 'string') {
+    return stringToBigNumber(value, numericBase);
+  } else if (typeof value === 'number' && isNaN(value) === false) {
+    return numberToBigNumber(value, numericBase);
```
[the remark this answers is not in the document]
brad-decker: A string could be a base 10 string like '10' or a base 16 string like 'a', it could also be a something like '0x-a' or '0.a7' the latter of these would error if we attempt to cast to a number. Furthermore for very large numbers represented by hexadecimal casting to number would cause issues with accuracy of the number. Javascript can only accurately and safely manage numbers to a certain value.

A Number might be NaN and casting it to string would result in errors. Your suggestion makes sense but for the sake of treating each possible special case i think it would be best if we deal with them separately.

shared/modules/Numeric.ts
brad-decker: @Gudahtt I will break this down *some* but I don't know that I can break it down to the level of granularity that would be the easiest to review. I say this for the sake of expediency with hitting our goals for Q1 within the allotted time we have for TS as well because there are some order of operations issues. @DDDDDanica also asked for this, so cc'ing her for acknowledgement and thanking her for pushing back as well. 

> Add this new Numeric module

This would be easy to split out, considering it wouldn't be used anywhere yet.

> Migrating code from conversion.utils.js to Numeric in chunks

So to do this I would need to redo some things because I converted conversion.utils.js to typescript *first* to get type feedback. To be able to merge this in chunks I would need to migrate to Numeric in chunks in *JavaScript* because if i attempt to do this in TypeScript it will not be mergeable due to lint / compile errors. So this would involve more rework than I think if sensible with the time constraints mentioned. However...

There's the fact that I relocated a bunch of methods that *use* conversionUtil from other parts of the codebase to the conversion.utils.js file. That would be a more meaningful break-up strategy and would mitigate a lot of LOC from this PR. I already did this PR breakdown here: https://github.com/MetaMask/metamask-extension/pull/17319

I will see what else can be shaved off into separate PRs.

brad-decker:
Converting to draft for now. Thanks @pedronfigueiredo and @DDDDDanica for initial reviews

shared/modules/Numeric.ts
DDDDDanica: Sorry actually `this.value` would result in `isPositive` in this case.

shared/modules/Numeric.ts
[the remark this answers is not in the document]
DDDDDanica: Thanks for the clear explanation, now i can see different cases we need to deal with in each side, makes sense to separate definitely 👍

shared/modules/Numeric.ts
Gudahtt: Oh sorry, I explained that poorly. I meant to say:

>Migrating code that _uses  `conversion.utils.js` to use `Numeric` instead, in chunks

shared/modules/Numeric.ts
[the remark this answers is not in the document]
pedronfigueiredo: That's neat, thanks for the explanation @brad-decker @Gudahtt

shared/modules/Numeric.ts
brad-decker: Which would be what we expect right?

DDDDDanica: But 0 is not positive or negative, it should be 0 🧐

brad-decker: Discussion moved to #17324 '

development/run-ganache.sh
```diff
@@ -23,7 +23,7 @@ trap _term SIGTERM
 trap _int SIGINT
 
 # shellcheck disable=SC2086
-$ganache_cli --chain.vmErrorsOnRPCResponse false --networkId 1337 --mnemonic "$seed_phrase" ${GANACHE_ARGS:-} &
```
brad-decker, line 26: these cchanges are unintentional byproduct of testing around a bug.

shared/modules/Numeric.ts
brad-decker: @Gudahtt I think i've got it fairly granular now. Some things have already merged, namely Numeric module and the relocation of various helper methods to the conversion.utils.js shared file. I will rename this PR and move from draft when it comes up but it now only contains the changes from the conversion.utils.js -> conversion.utils.ts conversion.

DDDDDanica (dismissed):

pedronfigueiredo (dismissed):

shared/modules/conversion.utils.ts
```diff
@@ -0,0 +1,219 @@
+/**
```
mcmire, line 1: Which function is this JSDoc block for? I'm curious if we need it anymore. Or at least it seems we don't need the list of params here.

shared/modules/conversion.utils.ts
```diff
@@ -0,0 +1,219 @@
...
+import { BigNumber } from 'bignumber.js';
+
+import { addHexPrefix, BN } from 'ethereumjs-util';
+import { EtherDenomination } from '../constants/common';
+import { Numeric } from './Numeric';
+
+/**
+ * Defines the base type of numeric value
+ */
+export type NumericBase = 'hex' | 'dec' | undefined;
```
mcmire, line 34: Should `undefined` be included in the set of valid bases? It seems like undefined should be a valid value for the base but then it should be normalized to one of two possibilities.

shared/modules/conversion.utils.ts
```diff
@@ -0,0 +1,219 @@
...
+  toNumericBase?: NumericBase;
+  toDenomination?: EtherDenomination;
+  toCurrency?: string;
+  numberOfDecimals?: number;
+  conversionRate?: number | BigNumber;
+  invertConversionRate?: boolean;
+  roundDown?: number;
+}
+
+export function decGWEIToHexWEI(decGWEI: number) {
```
mcmire, line 54: Nit: What are your thoughts on being explicit about the return type? I know TypeScript can infer this but I wonder if it would help with readability to see it here.

shared/modules/conversion.utils.ts
```diff
@@ -0,0 +1,219 @@
...
+import { addHexPrefix, BN } from 'ethereumjs-util';
+import { EtherDenomination } from '../constants/common';
+import { Numeric } from './Numeric';
+
+/**
+ * Defines the base type of numeric value
+ */
+export type NumericBase = 'hex' | 'dec' | undefined;
+
+/**
```
mcmire, line 36: This seems left over from the documentation for EthDenomination — should it be ✂️ ?

shared/modules/conversion.utils.js
```diff
@@ -1,452 +0,0 @@
...
-}
-
-export function hexToDecimal(hexValue) {
-  return conversionUtil(hexValue, {
-    fromNumericBase: 'hex',
-    toNumericBase: 'dec',
-  });
-}
-
-export {
```
mcmire, line 437: It seems that these functions are no longer available in this file — where did those move to? (Sorry I haven't been keeping up 😅 )

shared/modules/Numeric.ts
```diff
@@ -0,0 +1,621 @@
...
+
+  /**
+   * The current denomination, if any. The only supported denominations are
+   * ETH, GWEI, WEI.
+   */
+  denomination?: EtherDenomination;
+
+  constructor(
+    value: NumericValue,
+    base?: NumericBase,
```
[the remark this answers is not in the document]
brad-decker: The numeric code in this PR was removed to a PR that already landed, We could update them but it'll need to be in a new PR. FWIW the only case where we'd be able to bypass providing a base is in the case where we are providing a BigNumber/BN

jpuri (dismissed):
Great work 👍

shared/modules/conversion.utils.js
brad-decker: They no longer exist or are used, various parts of this PR were chunked out including refactoring usages of these utility methods to use Numeric direclty.

shared/modules/conversion.utils.ts
brad-decker: This is no longer needed or used actually

shared/modules/conversion.utils.ts
brad-decker: Removed

shared/modules/conversion.utils.ts
brad-decker: Removed

jpuri (dismissed):

shared/modules/conversion.utils.ts
mcmire: Cool 👍🏻

shared/modules/Numeric.ts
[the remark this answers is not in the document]
mcmire: No problem!

shared/modules/conversion.utils.js
mcmire: Awesome, I figured but just wanted to double check.

mcmire (approved):
I assume that we plan on adding more tests in a future PR? Either way, big kudos for tackling this. I really like the improvements here, and the Numeric class seems like a much better design than before.

DDDDDanica (approved):

PR #17276, opened 2023-01-18
Activate LavaMoat scuttling security feature

weizman:
## Explanation

Activating new LavaMoat security feature introduced and explained in https://github.com/lavamoat/lavamoat/pull/360

## Manual Testing Steps

The app should function just the same as before, the only change you should see is that when you try accessing scuttled properties of the `window` object you should see an error denying you from accomplishing that.

You can achieve that by simply testing that out in the console.

### Before

```javascript
window.alert(1); // works
```

### After

```javascript
window.alert(1); // throws an error of how `alert` is a scuttled property
```

## Pre-merge author checklist

- [x] I've clearly explained:


development/build/scripts.js
```diff
@@ -920,6 +969,24 @@ function setupMinification(buildConfiguration) {
   });
 }
 
+function setupScuttlingWrapping(buildConfiguration, applyLavaMoat) {
```
kumavis, line 972: in stead of appending scuttling can we just add a script file that scuttles?

development/build/utils.js
```diff
@@ -118,6 +118,58 @@ function logError(error) {
   console.error(error.stack || error);
 }
 
+function wrapAgainstScuttling(content, bag = {}) {
```
kumavis, line 121: can you explain what this does

development/build/scripts.js
```diff
@@ -50,6 +51,42 @@ const {
   createRemoveFencedCodeTransform,
 } = require('./transforms/remove-fenced-code');
 
+// map dist files to bag of needed native APIs against LM scuttling
+const scuttlingBagConfig = {
```
kumavis, line 55: "bag" is very colloquial and generally not used in code/comments. `scuttlingConfig` should suffice

weizman: 0858e0e29

development/build/scripts.js
weizman: scuttling takes place in LavaMoat, but in order for sentry to still work under scuttled environment I have to wrap its entire code with `with`s and `proxy`s, that generate a safe-from-scuttled-env scope - so in that sense it kinda has to be textually wrapping sentry source code in one file rather than loading it separately.

development/build/utils.js
weizman: This function `wrapAgainstScuttling()` tries to generically wrap given code with an environment that allows it to still function under a scuttled environment.

It's only (current) use is for `sentry` code which runs before scuttling happens but later on still leans on properties of the global object which at that point are scuttled.

To accomplish that, we wrap the entire provided code with the good old `with-proxy` trick, which helps us capture access attempts like (1) `window.fetch`/`globalThis.fetch` and (2) `fetch`.

`wrapAgainstScuttling()` function also accepts a bag of the global object's properties the code needs in order to properly function, and within our proxy we make sure to return those whenever the code goes through our proxy asking for them.

Specifically when the code tries to set properties to the global object, in addition to the preconfigured properties, we also accept any property starting with `on` to support global event handlers settings.

Also `sentry` invokes functions dynamically using `Function.prototype`'s `call` and `apply`, and our proxy messes with their `this` when that happens, so these two required a tailor-made patch.

weizman: To @kumavis's request, added the comment above to the code c9ffa9463

development/build/scripts.js
```diff
@@ -920,6 +969,24 @@ function setupMinification(buildConfiguration) {
...
 
+function setupScuttlingWrapping(buildConfiguration, applyLavaMoat) {
+  const { events } = buildConfiguration;
+  events.on('configurePipeline', ({ pipeline }) => {
+    pipeline.get('scuttle').push(
+      through.obj(
+        callbackify(async (file, _enc) => {
+          const bag = scuttlingConfig[file.relative];
+          if (applyLavaMoat && bag) {
+            const wrapped = wrapAgainstScuttling(file.contents.toString(), bag);
```
kumavis, line 980:
```suggestion
          const configForFile = scuttlingConfig[file.relative];
          if (applyLavaMoat && configForFile) {
            const wrapped = wrapAgainstScuttling(file.contents.toString(), configForFile);
```

kumavis (dismissed):

development/build/utils.js
```diff
@@ -118,6 +118,83 @@ function logError(error) {
...
+    for (const prop in bag) {
+      const that = bag[prop];
+      let api = global[prop];
+      if (that) api = api.bind(global[that]);
+      bag[prop] = api;
+    }
+    // setup proxy for the vetted shim to go through
+    const proxy = new Proxy(bag, {
+      get: function get(target, prop) {
+        return bag[prop] || Reflect.get(target, prop);
```
naugtur, line 163: bag is passed to the Proxy constructor, so target === bag. 
Why both?

weizman: I think you're right,
aa02339c5
Let's see if CI tests agree too.

development/build/utils.js
```diff
@@ -118,6 +118,83 @@ function logError(error) {
...
+      function wrap(obj, prop, target, receiver) {
+        const real = obj[prop];
+        obj[prop] = function(that) {
+          if (that === receiver) that = target;
+          const args = [].slice.call(arguments, 1);
+          return real.call(this, that, ...args);
+        };
+      }
+      wrap(Function.prototype, 'bind', target, receiver);
+      wrap(Function.prototype, 'apply', target, receiver);
```
naugtur, line 184: why wrap `bind` and `apply` but not `call`?

development/build/utils.js
```diff
@@ -118,6 +118,83 @@ function logError(error) {
...
+        if (bag.hasOwnProperty(prop) || prop.startsWith('on')) {
+          return bag[prop] = global[prop] = value;
+        }
+      },
+    });
+    // make sure bind() and apply() are applied with
+    // proxy target rather than proxy receiver
+    (function(target, receiver) {
+      'use strict'; // to work with ses lockdown
+      function wrap(obj, prop, target, receiver) {
```
naugtur, line 175: I'm trying to figure out why this is needed - is sentry binding functions to window and giving them away?

development/build/utils.js
```diff
@@ -118,6 +118,83 @@ function logError(error) {
...
+    for (const prop in bag) {
+      const that = bag[prop];
+      let api = global[prop];
+      if (that) api = api.bind(global[that]);
+      bag[prop] = api;
+    }
+    // setup proxy for the vetted shim to go through
+    const proxy = new Proxy(bag, {
+      get: function get(target, prop) {
+        return Reflect.get(target, prop);
```
naugtur, line 163: AFAIK resulting proxy would behave the same way if the get function was not defined at all - Reflect.get is the default for get on all objects.

development/build/utils.js
```diff
@@ -118,6 +118,83 @@ function logError(error) {
...
+        };
+      }
+      wrap(Function.prototype, 'bind', target, receiver);
+      wrap(Function.prototype, 'apply', target, receiver);
+    } (global, proxy));
+    return proxy;
+  }
+  const proxy = setupProxy(globalThis);
+  with (proxy) {
+    with ({window: proxy, self: proxy, globalThis: proxy}) {
```
naugtur, line 190: I like how layering this makes the proxy code much simpler.

development/build/utils.js
weizman: 1709c2931 probably right

development/build/utils.js
weizman: exactly. but because of our proxy, it ends up binding them to the proxy instead of the window which blows things up. that's why i had to make sure that the bind/apply calls use the window instead of the proxy in case that happens.

naugtur (approved):

kumavis (approved):
L👀KS G👀D

development/build/index.js
```diff
@@ -74,7 +74,52 @@ async function defineAndRunBuildTasks() {
   } = await parseArgv();
 
   // build lavamoat runtime file
-  await lavapack.buildRuntime({ scuttleGlobalThis: false });
+  // build lavamoat runtime file
```
digiwand, line 77:
```suggestion
```

digiwand (approved):

PR #17277, opened 2023-01-18
[FLASK] `snaps-monorepo@0.28.0`

FrederikBolding:
## Explanation

Bumps Snaps packages to `0.28.0` and handles any breaking changes.


app/scripts/metamask-controller.js
```diff
@@ -2578,10 +2576,7 @@ export default class MetamaskController extends EventEmitter {
       throw new Error('Primary keyring mnemonic unavailable.');
     }
 
-    const recoveredIndices = Array.from(
-      new Uint16Array(new Uint8Array(keyring.mnemonic).buffer),
-    );
-    return recoveredIndices.map((i) => englishWordlist[i]).join(' ');
+    return keyring.mnemonic;
```
Mrtenz, line 2579: Did you check if this function is used anywhere else, which may not accept `Uint8Array` mnemonic phrases?

Mrtenz (dismissed):

app/scripts/metamask-controller.js
FrederikBolding: Yeah. The only reference to the function is exposing it as a hook to snap RPC methods.

Mrtenz (approved):

digiwand (approved):

PR #17291, opened 2023-01-19
Allow user to scroll network form in onboarding modal

danjm:
Fixes a bug in v10.24.0 (but not in prod)

If a user is adding a custom network during the onboarding flow, and they get a warning message on the currency or rpc fields, the save button can be pushed out of view. This PR fixes that by adding a scroll in such cases.

Before:

https://user-images.githubusercontent.com/7499938/213450972-04621b9c-5fcb-40db-9f1e-96de35175114.mp4

After:

https://user-images.githubusercontent.com/7499938/213451013-244a5786-3cd4-4944-bdd9-8ab6ed35aff3.mp4

### DIFF
--- a/ui/pages/onboarding-flow/add-network-modal/index.js
+++ b/ui/pages/onboarding-flow/add-network-modal/index.js
@@ -34,6 +34,7 @@ export default function AddNetworkModal() {
       </Box>
       <NetworksForm
         addNewNetwork
+        restrictHeight
         networksToRender={[]}
         cancelCallback={closeCallback}
         submitCallback={closeCallback}
--- a/ui/pages/settings/networks-tab/index.scss
+++ b/ui/pages/settings/networks-tab/index.scss
@@ -234,6 +234,11 @@
     }
   }
 
+  &__restrict-height {
+    max-height: 578px;
+    overflow-y: auto;
+  }
+
   &__add-network-form-footer {
     display: flex;
     flex-flow: row;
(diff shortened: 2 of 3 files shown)


ui/pages/settings/networks-tab/index.scss
```diff
@@ -234,6 +234,11 @@
     }
   }
 
+  &__restrict-height {
+    max-height: 578px;
```
jpuri, line 238: Any reason for this value `578px` ?

jpuri (dismissed):

PeterYinusa (approved):
LGTM

pedronfigueiredo (approved):

ui/pages/onboarding-flow/add-network-modal/index.js
```diff
@@ -34,6 +34,7 @@ export default function AddNetworkModal() {
       </Box>
       <NetworksForm
         addNewNetwork
+        restrictHeight
```
DDDDDanica, line 37: Optional: Sorry for late review, but i think we can use name like `isHeightRestricted` to represent better to be a boolean. Totally optional, I just found out it is useful to pick up when we don't have rich document around props.

ui/pages/settings/networks-tab/index.scss
DDDDDanica: I was thinking the same.... could be nice to have some measurement from designer to get the height for modal.... considering we also have different resolutions of screen. wdyt?

PR #11212, opened 2021-05-30
[i18n:zh_tw] Sync and update translation for zh_TW

andy0130tw:
This patch translates essentially all strings to locale zh_TW, with the following topics excluded:

1. GDPR messages (`gdprMessage` and `gdprMessagePrivacyPolicy`)
2. MetaMask Swaps (keys beginning with "swap")
3. MetaMetrics (keys beginning with "metametrics") -- I am not confident on translating these messages precisely

In addition, a variety of fixes are applied to improve existing translation strings.

- Sync all missing strings from en strings at v9.5.7.
- Update some strings where the corresponding message has changed over time (e.g., `readdToken`).
- Fix some inappropriate punctuations (e.g., missing trailing periods).
- Use more common/agreed terms used in Taiwan (合約、助憶詞、帳戶、函式、...).

I translate these blockchain-related terminologies to the best of my knowledge, but I am open to any adjustments.

PR #17304, opened 2023-01-19
Add missing appName in the zh_TW language file

danjm:
The "appName" translation key and message is required for all translation files because it is used by the extension store.

This was removed from the `zh_TW` file in https://github.com/MetaMask/metamask-extension/pull/11212

Although the reason for its removal was not documented in that PR, it is presumably because their is no translation, and so as the locale message in `zh_TW` will match the one in the `en` file, it is fine to remove it in `zh_TW` so that the fallback to `en` is used.

As a follow-up task, we should add code to our verify-locales script that fails builds if these keys and translations are not included

### DIFF
--- a/app/_locales/zh_TW/messages.json
+++ b/app/_locales/zh_TW/messages.json
@@ -115,6 +115,18 @@
   "amount": {
     "message": "數量"
   },
+  "appName": {
+    "message": "MetaMask",
+    "description": "The name of the application"
+  },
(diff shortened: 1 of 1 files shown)


mcmire (dismissed):
Makes sense!

PeterYinusa (approved):

brad-decker (approved):

PR #17316, opened 2023-01-20
Use PageContainerFooter to render footer on all confirmation pages

### LINKED ISSUE #17237: Create re-usable footer component
Across all of confirmation pages we have similar footer component:

Screenshot 2023-01-16 at 9 15 47 PM

Let's create a reusable component for us across the pages:

1. Labels on buttons should be configurable with sensible default.
2. Click event handlers should be configurable.
3. Enabled / disabled state of buttons should be configurable
4. Ensure test coverage.
5. Refactor existing confirmation pages to use this footer.

Signature request screens have footer with bugger button while all other confirmation screens have smaller buttons. It will be useful to confirm with design team if we continue to keep current designs for buttons - as it might be better to make footer same in all confirmation pages.
In case we want to keep bigger buttons in request signature pages we can create a separate task to build footer for those pages.

jpuri:
Fixes: https://github.com/MetaMask/metamask-extension/issues/17237


digiwand (approved):
lgtm!

darkwing (approved):
Everything working as expected, well done!

PR #17321, opened 2023-01-20
Pass `excludedPermissions` to `SnapController`

### LINKED ISSUE #990: Forbid snaps from requesting `eth_accounts`
**The problem**

There are multiple edge cases with `eth_accounts`, additionally, that RPC is strongly linked to dApp set-up process, and the assumption this RPC does are related to that. For example, `eth_accounts` returns one account, while snap can get all the accounts using private key. UI problems, etc.

It's confusing for everyone involved - users, developers and MM devs.

Forbid snaps from requesting `eth_accounts`

We need to ensure that snaps cannot request `eth_accounts` in any way for stable. For Flask we can do whatever, but it must not be permitted under any circumstance in stable.

### Outcome

1. Fails with error message when `eth_accounts` is used.
2. Describe why we're focusing more on multi-chain methods and `snap_dialog` in a comment in this issue.
3. The message in the error should point to this issue for discussion.

### LINKED ISSUE #1103: Disable `endowment:long-running` in stable
#945 requires rethinking a new API to allow useful Snap use-cases that used this behaviour before.

It's important to support those case, but we also want to avoid scope creep for v1.0.0 release of Snaps.

### Outcome
* Disable `endowment:long-running` in stable altogether and post a deprecation warning in Flask
* Add a console warning when using that permission in Flask explaning the deprecation and pointing to a github issue / discussion.

GuillaumeRx:
Create a record of excluded permission/endowments with their associated error messages and pass it to `SnapController`. This also checks for the MetaMask version and updates the relevant methods using those excluded permissions.

Fixes: https://github.com/MetaMask/snaps-monorepo/issues/1103 https://github.com/MetaMask/snaps-monorepo/issues/990


shared/constants/permissions.ts
```diff
@@ -18,11 +18,21 @@ export const RestrictedMethods = Object.freeze({
 } as const);
 
 ///: BEGIN:ONLY_INCLUDE_IN(flask)
+const isMain = process.env.METAMASK_BUILD_TYPE === 'main';
```
FrederikBolding, line 21: Could we define these in a shared location and export? We need one in the MetaMask controller for `isFlask` too.

shared/constants/permissions.ts
```diff
@@ -18,11 +18,21 @@ export const RestrictedMethods = Object.freeze({
...
+export const StableEndowmentPermissions = Object.freeze({
+  'endowment:network-access': 'endowment:network-access',
+  'endowment:transaction-insight': 'endowment:transaction-insight',
+  'endowment:cronjob': 'endowment:cronjob',
+  'endowment:ethereum-provider': 'endowment:ethereum-provider',
+  'endowment:rpc': 'endowment:rpc',
+} as const);
+
+export const FlaskEndowmentPermissions = Object.freeze({
   'endowment:network-access': 'endowment:network-access',
```
FrederikBolding, line 36: If it is fair to assume Flask endowments to be a subset we could do 
```
export const FlaskEndowmentPermissions = Object.freeze({
...StableEndowmentPermissions,
(...)
});
```

shared/constants/permissions.ts
```diff
@@ -32,6 +42,33 @@ export const EndowmentPermissions = Object.freeze({
...
+    'eth_accounts is disabled. For more information please see https://github.com/MetaMask/snaps-monorepo/issues/990.',
+};
+export const ExcludedStableSnapPermissions = {
+  eth_accounts:
+    'eth_accounts is disabled. For more information please see https://github.com/MetaMask/snaps-monorepo/issues/990.',
+};
+export const ExcludedStableSnapEndowments = {
+  'endowment:keyring': 'This endowment is not available',
+  'endowment:long-running':
+    'endowment:long-running is deprecated. For more information please see https://github.com/MetaMask/snaps-monorepo/issues/945. ',
```
FrederikBolding, line 56: Please make sure these issue links actually explain why the snap endowment is disabled or deprecated

shared/constants/permissions.ts
```diff
@@ -18,11 +18,21 @@ export const RestrictedMethods = Object.freeze({
...
 
 ///: BEGIN:ONLY_INCLUDE_IN(flask)
+const isMain = process.env.METAMASK_BUILD_TYPE === 'main';
+
 export const PermissionNamespaces = Object.freeze({
   wallet_snap_: 'wallet_snap_*',
 } as const);
 
-export const EndowmentPermissions = Object.freeze({
+export const StableEndowmentPermissions = Object.freeze({
```
FrederikBolding, line 27: Should we not export these since we export them conditionally later?

FrederikBolding (changes requested):

shared/constants/permissions.ts
GuillaumeRx: It's using the same issue as the warning in snap-monorepo. It seems to be the most relevant issue regarding the deprecation of `endowment:long-running`.

shared/constants/flask/environment.js
```diff
@@ -0,0 +1,2 @@
+export const isMain = process.env.METAMASK_BUILD_TYPE === 'main';
```
FrederikBolding, line 1: Should this be in a `flask` folder? I think it would be fine to just keep it in `shared/constants/environment.js`

shared/constants/permissions.ts
FrederikBolding: Yeah, I just meant make sure to leave a comment or update the issue to make sure it correctly explains why we are disabling it. I does not have a lot of information at this point.

GuillaumeRx: Will leave a message on the linked issue :)

FrederikBolding (dismissed):

shared/constants/permissions.ts
```diff
@@ -18,20 +22,52 @@ export const RestrictedMethods = Object.freeze({
...
 
+const FlaskEndowmentPermissions = Object.freeze({
+  ...StableEndowmentPermissions,
+  'endowment:long-running': 'endowment:long-running',
+} as const);
+
 // Methods / permissions in external packages that we are temporarily excluding.
-export const ExcludedSnapPermissions = new Set([]);
-export const ExcludedSnapEndowments = new Set(['endowment:keyring']);
+const ExcludedFlaskSnapPermissions = {
```
david0xd, line 44: If this is final source of truth for this and similar specifications, it would probably be useful if it's frozen with `Object.freeze`. That would prevent any potential accidental change of it somewhere else.

david0xd (dismissed):

david0xd (dismissed):

david0xd (dismissed):

shared/constants/permissions.ts
hmalik88: hey, unresolving this as I didn't see an update to the issue or a comment.  Same for #990.

app/scripts/controllers/permissions/flask/snap-permissions.js
```diff
@@ -15,7 +15,7 @@ import {
 export const buildSnapEndowmentSpecifications = () =>
   Object.values(endowmentPermissionBuilders).reduce(
     (allSpecifications, { targetKey, specificationBuilder }) => {
-      if (!ExcludedSnapEndowments.has(targetKey)) {
+      if (!Object.keys(ExcludedSnapEndowments).includes(targetKey)) {
```
hmalik88, line 18: Why the change to Object.keys?

app/scripts/controllers/permissions/flask/snap-permissions.js
```diff
@@ -27,10 +27,10 @@ export const buildSnapEndowmentSpecifications = () =>
  * @param {Record<string, Function>} hooks - The hooks for the Snap
  * restricted method implementations.
  */
-export function buildSnapRestrictedMethodSpecifications(hooks) {
-  return Object.values(restrictedMethodPermissionBuilders).reduce(
+export const buildSnapRestrictedMethodSpecifications = (hooks) =>
+  Object.values(restrictedMethodPermissionBuilders).reduce(
     (specifications, { targetKey, specificationBuilder, methodHooks }) => {
-      if (!ExcludedSnapPermissions.has(targetKey)) {
+      if (!Object.keys(ExcludedSnapPermissions).includes(targetKey)) {
```
hmalik88, line 33: Same Q as above

app/scripts/controllers/permissions/flask/snap-permissions.js
GuillaumeRx: Because this variable is now an object in the form of `{ permission: errorMessage }` (previously a Set)

hmalik88: ah I missed that it was changed, ok cool!

shared/constants/permissions.ts
GuillaumeRx: Yes I just wanted to wait for this to be merged so we don't create confusion.

ritave (dismissed):
Excluded permissions should not be checked during run-time, but be completely removed during build time using code fencing.

I know this quite a bit of additional work, but this dramatically increases security. If we have a bug in logic, currently a bad actor could still find an exploit to run those permissions in stable.

If there's no code for those permissions at all in the extension, no way to run them in the first place.

As a suggestion, the easiest place to remove them during build time might be the RPC methods and PermissionController set-up, and the list here be used for informational purposes in console logs.

shared/constants/environment.js
```diff
@@ -0,0 +1,2 @@
+export const isMain = process.env.METAMASK_BUILD_TYPE === 'main';
+export const isFlask = process.env.METAMASK_BUILD_TYPE === 'flask';
```
Mrtenz, line 2: These seem to be unused now.

hmalik88 (approved):

shared/constants/permissions.ts
```diff
@@ -18,20 +18,32 @@ export const RestrictedMethods = Object.freeze({
 } as const);
 
 ///: BEGIN:ONLY_INCLUDE_IN(flask)
+/**
+ * Exclude permissions by code fencing them to avoid any potential usage of excluded permissions at runtime. See: https://github.com/MetaMask/metamask-extension/pull/17321#pullrequestreview-1287014285.
+ * This is a fix for https://github.com/MetaMask/snaps-monorepo/issues/1103 and https://github.com/MetaMask/snaps-monorepo/issues/990.
+ * TODO: Disable endowment:long-running and eth_account in stable.
```
ritave, line 24: This should be an issue not a TODO ;)

ritave, line 24: This should be an issue not a TODO ;)

ritave (approved):

shared/constants/permissions.ts
GuillaumeRx: What do you mean ? We have https://github.com/MetaMask/snaps-monorepo/issues/1103 and https://github.com/MetaMask/snaps-monorepo/issues/990. This is just a reminder to fence when we ship stable.

brad-decker:
@GuillaumeRx @ritave should this be merged?

PR #17324, opened 2023-01-20
add numeric module

brad-decker:
## Explanation
Part of #17262 

This Pull Request aims to add an abstraction layer over BigNumber with support for denominations of Ether and converting from and to those denominations. The goal here is to allow replacing of our usage of BigNumber and BN throughout the code base, as well as our own utilities in conversion.utils.js, with Numeric. We can then switch how we encode values in Numeric if we see fit and makes upgrading our dependencies much easier.

Numeric is a chainable class where most methods return a new instance of Numeric:

```js
new Numeric('0xa', 16, EtherDenominations.GWEI).toDenomination(EtherDenomination.ETH).toBase(10).toString();
```

The above converts '0xa' a hex value in GWEI to ETH in decimal format. 

---

This pull request does nothing with this module. It adds docs, tests and the module itself in TypeScript for review. It also updates jest.config to run ts test files.

## Pre-merge author checklist

- [x] I've clearly explained:
  - [x] How reviewers can test my changes: run test suite or import Numeric in a playground.ts file for usage


shared/modules/Numeric.ts
```diff
@@ -0,0 +1,620 @@
...
+      Numeric.from(comparator, base, denomination).value,
+    );
+  }
+
+  isNegative() {
+    return this.value.isNegative();
+  }
+
+  isPositive() {
+    return this.isNegative() === false;
```
DDDDDanica, line 577: just following the conversation here:
https://github.com/MetaMask/metamask-extension/pull/17262#discussion_r1082646325  

Should we add 
```
isZero() {}
```
to differ ? 

or we can put 
```
isNegative() {
    return this.value.isNegative() && this.value !== 0;
  }


  isPositive() {
    return this.isNegative() === false && this.value !== 0;
}
```

Or we can do 
```
isPositive() {
    return this.value.isPositive();
    }
```

wdyt?

brad-decker: There are no current parts of the codebase that need a toZero. 0 is positive so returning true from isPositive is correct for 0.  We can add a toZero method in the future if it is needed.

brad-decker: we primarily use isNegative to determine if someone entered a negative number in the send screen, but you can send 0 eth.

DDDDDanica: Oki gotcha !

brad-decker: I'd like to rephrase my comment from 'there are no current parts of the codebase that need a toZero' to 'there are no known (by me) instances of BigNumber comparisons that need  the .isZero method'. BigNumber does expose a isZero method so there is definitely a place for this method on Numeric in the future. We also *do* compare string '0's elsewhere that if we end up using Numeric in more places might result in a need for a 0 check. :)

DDDDDanica (approved):

ryanml (approved):

PR #17325, opened 2023-01-20
Fixing yarn version command, adding yarn version plugin

ryanml:
## Explanation
Fix for generating the new release version, re-adding yarn version plugin which was removed in v3


danjm (dismissed):

Gudahtt:
The flags we use for this command (`--no-git-tag-version --new-version`) aren't valid for the new `version` command anymore. See: https://yarnpkg.com/cli/version

I think the new command we'd want is something like `yarn version "${version}"` but I haven't tested it. Specifically I'm unsure if that would create a git tag or not. We might need to use `--deferred` and ignore the resulting staging file?

Gudahtt (dismissed):
LGTM!

brad-decker (dismissed):

danjm (dismissed):

brad-decker:
Uhhh... i'm looking into this. These changes should not be dropping coverage at all.

brad-decker:
Mocha is included .yarn as a coverage target... looks like this impacting codecov too.

brad-decker:
Pushed a commit with changes. Ironically the 'lines' doesn't change just statements, branches and functions.

danjm (approved):

Gudahtt (approved):

PR #17330, opened 2023-01-20
Convert app//scripts/lib/util.js to typescript

brad-decker:
## Explanation
Part of #17262 

converts app/scripts/lib/util.js to typescript.

Includes no functional changes. 

## Pre-merge author checklist

- [x] I've clearly explained:


app/scripts/lib/util.ts
```diff
@@ -178,8 +164,8 @@ export const generateRandomId = () => {
   return result;
 };
 
-export const isValidDate = (d) => {
-  return d instanceof Date && !isNaN(d);
```
brad-decker, line 182: I don't think the && !isNaN(d) was ever possible to hit because anything that is an instance of Date will not be isNaN, even if constructing Date with a negative number

lavamoat/build-system/policy.json
```diff
@@ -1913,7 +1913,6 @@
       },
       "packages": {
         "chokidar>braces": true,
-        "chokidar>fsevents": true,
```
DDDDDanica, line 1916: Hey I'm not very familiar with this file, but hope these 2 are safe to move, just flag it out to double check with you ~

app/scripts/lib/util.ts
```diff
@@ -259,31 +279,33 @@ export function formatTxMetaForRpcResult(txMeta) {
...
+  const formattedTxMeta: FormattedTransactionMeta = {
     v,
     r,
     s,
     to,
     gas,
     from,
     hash,
-    nonce,
+    nonce: `${nonce}`,
```
DDDDDanica, line 290: Question: if nonce is passed in as string, do we need to stringify here?

lavamoat/build-system/policy.json
brad-decker: We get some discrepancies with these files based on OS and this is one that we definitely get discrepancies with.

app/scripts/lib/util.ts
brad-decker: Its likely not an issue because our response handler would have stringified it anyways but in the TxMeta type its typed as a number. I'm not sure how accurate this is but this was done as a safety precaution.

lavamoat/build-system/policy.json
DDDDDanica: Gotcha !

app/scripts/lib/util.ts
DDDDDanica: thanks for explaining ~ it is safer this way indeed

DDDDDanica (approved):

darkwing (approved):

pedronfigueiredo (approved):

PR #17337, opened 2023-01-21
[TS dashboard] Be consistent with terminology

mcmire:
## Explanation

There are two sides to the TypeScript dashboard code, the build scripts
and the app code. Currently we use the word "module" in the former but
"file" in the latter. Similarly, we use "partition" in the build code
but "level" in the app code. All of the logic that powers the
visualization you see on the dashboard is complicated enough, so there's
no reason to use duplicate terminology.

## Manual Testing Steps

Run `yarn ts-migration:dashboard:watch`, then open `development/ts-migration-dashboard/build/final/index.html`. You should be able to see the dashboard.

## Pre-merge author checklist

- [x] I've clearly explained:
  - [x] What problem this PR is solving
  - [x] How this problem was solved
  - [x] How reviewers can test my changes
- [ ] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [ ] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
- [ ] PR is linked to the appropriate GitHub issue
- [ ] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.

In this case, a QA Engineer approval will be be required.

PR #17338, opened 2023-01-21
[TS dashboard] Draw dependencies between files 

### LINKED ISSUE #16810: Update TypeScript dashboard to show connections/dependencies between files
It is easy to assume that any file can be converted to TypeScript. However, if a file depends on another file which hasn't been converted yet, that file needs to be converted first. I don't think this fact is as obvious as it needs to be.

In the TypeScript migration dashboard, it would be nice to be able to click on a box that represents a file you're interested in converting which would open a sidebar listing the name of the file as well as the dependencies. Something like:

```
app/scripts/controllers/metametrics.js

To convert this file, you will need to first convert:

- shared/constants/app.js
- shared/constants/metametrics.js
- shared/constants/time.js
- shared/modules/mv3.utils.js
- shared/constants/alarms.js
- lib/util.js

Keep in mind, though, that these files may depend on other files themselves, so you may need to start with another file before tackling these.
```

The file names here would be links themselves, and clicking on a link would show the details for that file in the sidebar.

Also, when clicking on a box, a thicker border should appear around the box, and lines should be drawn between the file clicked on and the boxes that correspond to its dependencies. Clicking out (or pressing the X in the sidebar) should deselect the box, hide the lines, and close the sidebar.

To do this we will need to reimplement the boxes section using SVG instead of divs.

mcmire:
## Explanation

If you're thinking about converting a file to TypeScript, you might want
to know what files import that file and what files it imports. This
commit makes it so that if you click on a box in the visualization, it
will draw lines between that box and the other boxes which represent its
dependents and dependencies.

## Screenshots/Screencaps

https://user-images.githubusercontent.com/7371/220708405-07bc2fe7-656f-457a-8df7-622709581014.mov

## Manual Testing Steps

* Run `yarn ts-migration:dashboard:watch` in a terminal tab, then run open `development/ts-migration-dashboard/build/final/index.html`.
* Click on a box (let's call it A) that you see on the dashboard. You may see red lines and/or blue lines which connect this box to other boxes.
* Pick a red line and follow it up the tree to its box (let's call it B). This file is a dependency; if you open file A in your editor, you should see that it is `import`ing B.
* Pick a blue line and follow it down the tree to its box (let's call it C). This file is a dependent; if you open C in your editor, you should see that it is `import`ing A.
* Compare what you see to https://metamask.github.io/metamask-extension-ts-migration-dashboard/. While the boxes per level are now sorted, the boxes should live in the same levels — nothing should be rearranged.

## Pre-merge author checklist

- [x] I've clearly explained:

---

Closes #16810.


brad-decker:
WHAT SHUT UP

brad-decker (dismissed):

mcmire:
Marking this as draft until #17337 is merged.

development/ts-migration-dashboard/common/build-module-partitions.ts
```diff
@@ -208,76 +221,90 @@ function buildModulesWithLevels(
...
 
   while (modulesToFill.length > 0) {
-    const currentModule = modulesToFill.shift() as Module;
-    const childModulesToFill: Module[] = [];
-    (dependenciesByModuleId[currentModule.id] ?? []).forEach(
-      (givenChildModuleId) => {
-        const npmPackageMatch = givenChildModuleId.match(
-          /^node_modules\/(?:(@[^/]+)\/)?([^/]+)\/.+$/u,
-        );
+    const currentModule = modulesToFill[0];
```
mcmire, line 224: This is the most difficult part of this PR to review, and in fairness, it's a bit gnarly. Basically, this function is run during the build phase, and its goal is to spit out all of the data that the UI can use to render the visualization. At this point we do have a simple version of the dependency graph — we have a list of modules and we know the dependencies for each module — but we don't know what "level" each module is at, and we also aren't able to easily access the dependents for a module. To determine both of these, we have to walk through the graph. This is a bit tricky, as there are so many nodes we can't use recursion, and in addition, we may encounter cycles (A imports B, which imports C, which imports A, which imports B...). The code to handle both of these is already present, but it's important to point out, as it's complicated. (I really should add tests for this code, but in the meantime I've added some logging, as I had to remember how this worked.)

In any case, most of the changes in this file involve renaming "child" to "dependency" (as that makes more sense in the context), using `for` instead of `forEach`, and removing the typecasting. It's probably best to **disable whitespace changes** when viewing this PR to reduce some visual noise.

development/ts-migration-dashboard/common/build-module-partitions.ts
```diff
@@ -328,8 +358,20 @@ function partitionModulesByLevel(
...
   Object.values(modulesById).forEach((module) => {
     modulePartitions[module.level].children.push({
-      name: module.id,
+      id: module.id,
       hasBeenConverted: module.hasBeenConverted,
+      dependentIds: Object.keys(module.dependents).sort(),
+      dependencyIds: Object.keys(module.dependencies).sort(),
+    });
+  });
+  Object.values(modulePartitions).forEach((partition) => {
```
mcmire, line 378: This change is not strictly necessary to this PR, but it was a bit annoying that the boxes in each level were not sorted by filename, so this change does that.

development/ts-migration-dashboard/common/build-module-partitions.ts
```diff
@@ -208,76 +221,101 @@ function buildModulesWithLevels(
...
+
+      if (
+        !existingDependencyModule.dependents.some(
+          (m) => m.id === currentModule.id,
+        )
+      ) {
+        existingDependencyModule.dependents.push(currentModule);
+      }
+
+      if (
```
mcmire, line 303: The only part that's truly new in this file is this line, where, as we are iterating through the dependency of a given module, we make sure to associate that dependency with the module itself. This allows us to look up the dependencies later when we are rendering the visualization.

development/ts-migration-dashboard/app/components/App.tsx
```diff
@@ -12,16 +14,85 @@ function calculatePercentageComplete(summary: Summary) {
...
+
+  const registerBox = (id: string, boxRect: BoxRect) => {
+    setBoxRectsById((existingBoxRectsById) => {
+      if (existingBoxRectsById === undefined) {
+        return { [id]: boxRect };
+      }
+      return { ...existingBoxRectsById, [id]: boxRect };
+    });
+  };
+  const toggleConnectionsFor = (id: string) => {
```
mcmire, line 90: This function does what it says — it either shows or hides the connections for a box that you click on.

development/ts-migration-dashboard/app/components/Connections.tsx
```diff
@@ -0,0 +1,143 @@
...
+  type: 'dependency' | 'dependent';
+  x: number;
+  y: number;
+}) {
+  return (
+    <circle className={`module-connection__${type}-point`} cx={x} cy={y} />
+  );
+}
+
+export default function Connections({ activeBox }: { activeBox: BoxModel }) {
```
mcmire, line 93: While the boxes are rendered on screen using one `div` per box, the connection lines for a particular box are drawn all in one element using SVG.

brad-decker (approved):

legobeat (approved):
when ico?

PR #17344, opened 2023-01-23
Nit fixes for story-test integration

NidhiKJha:
This is a follow-up PR for [story-test integration PR](https://github.com/MetaMask/metamask-extension/pull/17092).

Updated the PR with the fixes @Gudahtt suggested.

Changes Made:
1. moved playwright from dependencies to dev dependencies.
2.  Removed the lint comment from `ui/components/app/app-header/app-header.stories.js`
3. Updated comment in `storybook.js`


package.json
```diff
@@ -481,6 +480,7 @@
     "nock": "^13.2.9",
     "node-fetch": "^2.6.1",
     "nyc": "^15.0.0",
+    "playwright": "^1.29.2",
```
Gudahtt, line 483: Could you move `wait-on` as well? That is another `devDependency` that was added to `dependencies` in that PR

Gudahtt (approved):
LGTM, thanks!

DDDDDanica (approved):

PR #17346, opened 2023-01-23
Test parameter change cache miss

### LINKED ISSUE #17003: Add tests for cached multi-parameter RPC methods to ensure we get a cache miss when any single parameter changes
When writing the tests for `createInfuraClient` in #15556, there were some that we opted to leave out. One of these concerns the caching behavior in `eth-json-rpc-middleware` when it involves requests to multi-parameter RPC methods. Specifically, in the block cache middleware, [we use all of the parameters to determine a cache key](https://github.com/MetaMask/eth-json-rpc-middleware/blob/254a66cf5f067f35db96c6a5740c8da9a79e495d/src/block-cache.ts#L62). Currently this behavior isn't completely exercised in the network client tests, and to do this we'd want to make two requests with different parameters and assert that the second isn't cached.

Gudahtt:
Tests have been added to ensure that our middleware will not return a cached response to a request with unique parameters. Each parameter supported by each method is tested independently.

Closes #17003

## Manual Testing Steps

No functional changes.

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/app/scripts/controllers/network/provider-api-tests/shared-tests.js
+++ b/app/scripts/controllers/network/provider-api-tests/shared-tests.js
@@ -143,65 +143,104 @@ export function testsForProviderType(providerType) {
 
     describe('methods that have a param to specify the block', () => {
       const supportingBlockParam = [
-        { name: 'eth_call', blockParamIndex: 1 },
-        { name: 'eth_getBalance', blockParamIndex: 1 },
-        { name: 'eth_getBlockByNumber', blockParamIndex: 0 },
-        { name: 'eth_getCode', blockParamIndex: 1 },
-        { name: 'eth_getStorageAt', blockParamIndex: 2 },
-        { name: 'eth_getTransactionCount', blockParamIndex: 1 },
+        {
+          name: 'eth_call',
+          blockParamIndex: 1,
+          numberOfParameters: 2,
+        },
+        {
+          name: 'eth_getBalance',
+          blockParamIndex: 1,
+          numberOfParameters: 2,
+        },
+        {
+          name: 'eth_getBlockByNumber',
+          blockParamIndex: 0,
+          numberOfParameters: 2,
+        },
+        { name: 'eth_getCode', blockParamIndex: 1, numberOfParameters: 2 },
+        {
+          name: 'eth_getStorageAt',
+          blockParamIndex: 2,
+          numberOfParameters: 3,
+        },
+        {
+          name: 'eth_getTransactionCount',
+          blockParamIndex: 1,
+          numberOfParameters: 2,
+        },
       ];
-      supportingBlockParam.forEach(({ name, blockParamIndex }) => {
-        describe(`method name: ${name}`, () => {
-          testsForRpcMethodSupportingBlockParam(name, {
-            providerType,
-            blockParamIndex,
+      supportingBlockParam.forEach(
+        ({ name, blockParamIndex, numberOfParameters }) => {
+          describe(`method name: ${name}`, () => {
+            testsForRpcMethodSupportingBlockParam(name, {
+              providerType,
+              blockParamIndex,
+              numberOfParameters,
+            });
           });
-        });
-      });
+        },
+      );
     });
 
     describe('methods that assume there is no block param', () => {
       const assumingNoBlockParam = [
-        'eth_blockNumber',
-        'eth_estimateGas',
-        'eth_gasPrice',
-        'eth_getBlockByHash',
+        { name: 'eth_blockNumber', numberOfParameters: 0 },
+        { name: 'eth_estimateGas', numberOfParameters: 2 },
+        { name: 'eth_gasPrice', numberOfParameters: 0 },
+        { name: 'eth_getBlockByHash', numberOfParameters: 2 },
         // NOTE: eth_getBlockTransactionCountByNumber does take a block param at
         // the 0th index, but this is not handled by our cache middleware
         // currently
-        'eth_getBlockTransactionCountByNumber',
+        {
+          name: 'eth_getBlockTransactionCountByNumber',
+          numberOfParameters: 1,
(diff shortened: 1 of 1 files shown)


app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -143,65 +143,104 @@ export function testsForProviderType(providerType) {
...
-        { name: 'eth_call', blockParamIndex: 1 },
-        { name: 'eth_getBalance', blockParamIndex: 1 },
-        { name: 'eth_getBlockByNumber', blockParamIndex: 0 },
-        { name: 'eth_getCode', blockParamIndex: 1 },
-        { name: 'eth_getStorageAt', blockParamIndex: 2 },
-        { name: 'eth_getTransactionCount', blockParamIndex: 1 },
+        {
+          name: 'eth_call',
+          blockParamIndex: 1,
+          numberOfParameters: 2,
```
Gudahtt, line 149: I looked up each method in both the [Ethereum JSON-RPC spec](https://ethereum.github.io/execution-apis/api-documentation/), and the [Infura docs](https://docs.infura.io/infura/networks/ethereum/json-rpc-methods). I ensured the configured number of parameters matched the _maximum_ shown in either docs.

They mostly matched already; I think there was a single method that had one less parameter on Infura because it didn't support some option, don't recall which one. And there were some methods exclusive to one or the other. But other than that they matched.

BelfordZ (approved):

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -143,65 +143,104 @@ export function testsForProviderType(providerType) {
...
+              numberOfParameters,
+            });
           });
-        });
-      });
+        },
+      );
     });
 
     describe('methods that assume there is no block param', () => {
```
adonesky1, line 186: I know this isn't the target of this PR so this is more of just an observation/question: this seems like an odd way to categorize the methods for testing?

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -452,6 +495,46 @@ export function testsForRpcMethodAssumingNoBlockParam(
     });
   });
 
+  for (const paramIndex of [...Array(numberOfParameters).keys()]) {
+    it(`does not reuse the result of a previous request if parameter at index "${paramIndex}" differs`, async () => {
```
adonesky1, line 499: took me a while to work through why this was testing what you claim its testing. But I see now! Looks good!

app/scripts/controllers/network/provider-api-tests/shared-tests.js
```diff
@@ -1564,6 +1690,65 @@ export function testsForRpcMethodSupportingBlockParam(
       });
     });
 
+    for (const paramIndex of [...Array(numberOfParameters).keys()]) {
+      if (paramIndex === blockParamIndex) {
+        // testing changes in block param is covered under later tests
+        continue;
+      }
```
adonesky1, line 1697: I'm not understanding why we want to skip these cases for this case?

adonesky1 (commented):
Just one real question, though I'm probably just not understanding something. Otherwise LGTM

app/scripts/controllers/network/provider-api-tests/shared-tests.js
Gudahtt: These groups are meant to reflect the caching strategies used in our middleware: either we ignore the method, or we cache it by the current block, or we cache it by the referenced block.

This "assume there is no block param" group is for methods that we cache by block, where the middleware doesn't bother looking for any block reference in the method parameters.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
Gudahtt: These tests are basically meant to ensure that the middleware correctly recognizes any parameter changes, and doesn't reuse a cache for different params. We want to check this for the block parameter as well, but, that one isn't so simple to make an arbitrary mutation to. The change will have other side-effects because the value is used by the caching middleware in other ways (it determines which block cache to use).

We do want to test the case where the block reference changes too, but, this test is within the block:

```
describe.each([
  ['given no block tag', undefined],
  ['given a block tag of "latest"', 'latest'],
])`
```

i.e. the block reference is always set to a static value in this test, that's what this category is for.

The case where the block reference is changed is covered by the test "does not reuse the result of a previous request if it was made with different arguments than this one", about ~1300 lines below here.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
Gudahtt: Yeah... the param builder function didn't quite do what I wanted here, so I constructed the parameters inline in the test. I should probably find a way to make this more readable.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
Gudahtt: Definitely lots of room for improvement with how these are organized. We can describe these categories better, and deduplicate the common behavior more effectively.

app/scripts/controllers/network/provider-api-tests/shared-tests.js
Gudahtt: Would you mind explaining which part you initially found confusing? That might help in finding a clearer way to write this.

adonesky1: I don't think there's any issue with the description. I think it was just towards the end of my day and my brain wasn't working as well. So I guess this comment was more for me and not very generalizable 😅

adonesky1 (approved):
LGTM!

PR #17366, opened 2023-01-24
Update development/ts-migration-dashboard/files-to-convert.json

ryanml:
Updating the list of files to convert to Typescript after work that was merged this past week


DDDDDanica (approved):

brad-decker:
we don't want to remove these as they will start making it so that we don't see the greeen boxes for those files that we have converted. :)

brad-decker (changes requested):
I'd move to close this, the files don't need to be updated unless we change a .js filename or add a new .js file

PR #17372, opened 2023-01-24
feat(mme-17214): migrate L33 files to typescript

### LINKED ISSUE #17214: Typescript Level 33
[Extension TypeScript Migration Status dashboard
](https://metamask.github.io/metamask-extension-ts-migration-dashboard/)
Level 33 Tasks:

- [ ] shared/constants/metametrics.js
- [x] ui/helpers/utils/i18n-helper.js
- [x] app/scripts/lib/buy-url.js
- [x] ui/ducks/app/app.js
- [x] ui/ducks/gas/gas-action-constants.js
- [ ] ui/helpers/utils/transactions.util.js
- [x] ui/helpers/utils/tx-helper.js
- [x] ui/pages/swaps/swaps.util.js
- [x] ui/helpers/constants/notifications.js

DDDDDanica:
## Explanation

Closes https://github.com/MetaMask/metamask-extension/issues/17214

## Pre-merge author checklist

- [x] I've clearly explained:


ui/ducks/app/app.ts
```diff
@@ -1,10 +1,71 @@
+import { AnyAction } from 'redux';
 import {
   WebHIDConnectedStatuses,
   HardwareTransportStates,
 } from '../../../shared/constants/hardware-wallets';
 import * as actionConstants from '../../store/actionConstants';
 
-export default function reduceApp(state = {}, action) {
+interface ReducerState {
```
DDDDDanica, line 8: @brad-decker wdyt of the name and approach here?

brad-decker: I'll pull your branch tonight or tomorrow morning and give this a good review. looks promising from initial glance.

ui/helpers/utils/i18n-helper.tsx
```diff
@@ -6,20 +6,25 @@ import getFetchWithTimeout from '../../../shared/modules/fetch-with-timeout';
 
 const fetchWithTimeout = getFetchWithTimeout();
 
-const warned = {};
-const missingMessageErrors = {};
-const missingSubstitutionErrors = {};
+const warned = {} as any;
+const missingMessageErrors = {} as any;
+const missingSubstitutionErrors = {} as any;
```
brad-decker, line 11: We should avoid using any as much as possible while still keeping the goal of decent levels of conversion in our sprint in mind. However its important to note that the more we rely upon typecasting to any the less beneficial the conversion to TypeScript actually is.  I'd like us to reserve the usage of any to places where we truly don't know yet what the types are or if typing it would require a much deeper process. In those cases we should come back shortly after this sprint to make a plan for handling them.

In this case here a a set of types:

```ts
// In our app/_locales folders such as app/_locales/en there is a messages.json file.
// Each file is an object with keys that are translation keys. In our codebase you'll see
// t('reject') the 'reject' is the translationKey. The values of the object are an object with
// this shape. Message is the string that will replace the translationKey, and that message
// may contain replacement variables such as $1, $2, etc. Description is key describing the
// usage of the message. 
interface I18NMessage {
  message: string;
  description?: string;
}

// Now that we have the shape of the individual entries in the messages.json file, this is the
// shape of overall file itself, translationKey (string) and the I18NMessae as the value.
interface I18NMessageDict {
  [translationKey: string]: I18NMessage;
}

// Below we are going to need a couple maps that use the same structure (translationKey) key
// but with different values, so we create a parameterized type (or generic type) that allows us
// to create a shortcut to what we want.
interface I18NMessageDictMap<R> {
  [translationKey: string]: R;
}

// We can then type these objects for safety in the below code.
const warned: { [localeCode: string]: I18NMessageDictMap<boolean> } = {};
const missingMessageErrors: I18NMessageDictMap<Error> = {};
const missingSubstitutionErrors: {
  [localeCode: string]: I18NMessageDictMap<boolean>;
} = {};
```

ui/helpers/utils/i18n-helper.tsx
```diff
@@ -6,20 +6,25 @@ import getFetchWithTimeout from '../../../shared/modules/fetch-with-timeout';
...
+ * @param localeCode - The code for the current locale
+ * @param localeMessages - The map of messages for the current locale
+ * @param key - The message key
+ * @param substitutions - A list of message substitution replacements
+ * @returns The localized message
  */
-export const getMessage = (localeCode, localeMessages, key, substitutions) => {
+export const getMessage = (
+  localeCode: string,
+  localeMessages: Record<string, any> | null,
```
brad-decker, line 24: We actually know what the localeMessages shape will be, in this case the `I18nMessageDict` detailed above, or null.

ui/helpers/utils/i18n-helper.tsx
```diff
@@ -110,19 +115,21 @@ export async function fetchLocale(localeCode) {
...
     !relativeTimeFormatLocaleData.has(languageTag)
   ) {
     const localeData = await fetchRelativeTimeFormatData(languageTag);
-    Intl.RelativeTimeFormat.__addLocaleData(localeData);
+    (Intl.RelativeTimeFormat as any).__addLocaleData(localeData);
   }
 }
 
-async function fetchRelativeTimeFormatData(languageTag) {
+async function fetchRelativeTimeFormatData(languageTag: string): Promise<any> {
```
brad-decker, line 132: Would normally be an okay usage of Any, but there is a 'Json' type exported from '@metamask/utils' we should use for this.

ui/helpers/utils/i18n-helper.tsx
```diff
@@ -110,19 +115,21 @@ export async function fetchLocale(localeCode) {
...
   const languageTag = localeCode.split('_')[0];
   if (
     Intl.RelativeTimeFormat &&
-    typeof Intl.RelativeTimeFormat.__addLocaleData === 'function' &&
+    typeof (Intl.RelativeTimeFormat as any).__addLocaleData === 'function' &&
     !relativeTimeFormatLocaleData.has(languageTag)
   ) {
     const localeData = await fetchRelativeTimeFormatData(languageTag);
-    Intl.RelativeTimeFormat.__addLocaleData(localeData);
+    (Intl.RelativeTimeFormat as any).__addLocaleData(localeData);
```
brad-decker, line 128: I searched for details on this __addLocaleData method... It seems it comes from @formatjs, but I think Intl is a global object on window now. @Gudahtt might know more, but for now casting this to any is fine unless @Gudahtt knows what is going on here. I tried using a declaration block to add a type signature to this method to no avail.

ui/helpers/utils/i18n-helper.tsx
```diff
@@ -110,19 +115,21 @@ export async function fetchLocale(localeCode) {
 
 const relativeTimeFormatLocaleData = new Set();
 
-export async function loadRelativeTimeFormatLocaleData(localeCode) {
+export async function loadRelativeTimeFormatLocaleData(
+  localeCode: string,
+): Promise<any> {
```
brad-decker, line 120: Promise<any> does satisfy Promise<void> but in this case lets use the latter because we don't return anything from this method.

ui/helpers/utils/i18n-helper.tsx
```diff
@@ -96,7 +101,7 @@ export const getMessage = (localeCode, localeMessages, key, substitutions) => {
   return phrase;
 };
 
-export async function fetchLocale(localeCode) {
+export async function fetchLocale(localeCode: string): Promise<any> {
```
brad-decker, line 104: I believe the return type of this should be `I18NMessageDict` from above.

brad-decker (changes requested):
Good work so far @DDDDDanica :D I've left some detailed feedback on how we can make these TypeScript conversions more beneficial. I haven't yet dug into the tx-helper changes but we have a way to make that better typed as well and i'll be leaving feedback soon.

ui/ducks/app/app.ts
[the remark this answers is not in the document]
DDDDDanica: Thanks for the suggestion, i put 2 state there: 
`AppState` is the big object with all the values;
and `AppSliceState` is using `AppState` as value with key  `appState`
```
interface AppSliceState {
  appState: AppState;
}
```
Hope it is clear now ~

ui/ducks/app/app.ts
```diff
@@ -1,10 +1,71 @@
...
+    onboardedInThisUISession: boolean;
+    customTokenAmount: string;
+    forgottenPassword: boolean;
+  };
+}
+
+export default function reduceApp(
+  state: ReducerState,
+  action: AnyAction,
+): Record<string, any> {
```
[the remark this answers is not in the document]
DDDDDanica: Good catch ! changed ~

ui/ducks/app/app.ts
```diff
@@ -420,63 +482,85 @@ export default function reduceApp(state = {}, action) {
 }
 
 // Action Creators
-export function hideWhatsNewPopup() {
+interface ActionCreator {
+  type: string;
+  value?: any;
+}
```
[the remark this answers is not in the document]
DDDDDanica: That is a really good idea, i didn't know about the type provided by redux before.  And the example is super clear ~

One thing i did for actions without return value is to give a null value so we can still use the type `PayloadAction`, eg: 
```
export function toggleCurrencySwitch(): PayloadAction<null> {
  return { type: actionConstants.TOGGLE_CURRENCY_INPUT_SWITCH, payload: null };
}
``` 
wdyt?

ui/ducks/app/app.ts
```diff
@@ -420,63 +482,85 @@ export default function reduceApp(state = {}, action) {
...
   return { type: actionConstants.SET_WEBHID_CONNECTED_STATUS, value };
 }
 
-export function setLedgerTransportStatus(value) {
+type KeysOfHardwareTransportStates = keyof typeof HardwareTransportStates;
+type ValuesOfHardwareTransportStates =
+  typeof HardwareTransportStates[KeysOfHardwareTransportStates];
+export function setLedgerTransportStatus(
+  value: ValuesOfHardwareTransportStates,
+): ActionCreator {
```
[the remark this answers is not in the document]
DDDDDanica: Somehow i was making it so complex while there's a super easy way to give the type 😵 thanks for walking through the changes ... it is super clear !

ui/helpers/utils/i18n-helper.tsx
DDDDDanica: I did some research on this and it seemed to come from windows. (same as a globe object i was dealing with);

>  I tried using a declaration block to add a type signature to this method to no avail.
Can you show me your another solution?

ui/helpers/utils/i18n-helper.tsx
DDDDDanica: Good to know !! changed 🙏

ui/helpers/utils/i18n-helper.tsx
DDDDDanica: olala this is super helpful and i can't agree more with your point to have decent types than `any`. and thanks for showing this example, super clear ❤️

ui/helpers/utils/i18n-helper.tsx
DDDDDanica: Gotcha !

ui/ducks/app/app.ts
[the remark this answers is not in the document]
brad-decker: So PayloadAction comes from reduxjs/toolkit, which is an opinionated toolkit by redux for redux :D ... there's also an Action type from plain old redux that is for actions without a payload :)

ui/helpers/utils/i18n-helper.tsx
brad-decker: I edited the above to put code ticks around the types because github tries to render html tags and `<void>` is one of those i guess lol

DDDDDanica: :) hahah github is too smart

ui/helpers/utils/i18n-helper.tsx
```diff
@@ -62,15 +67,15 @@ export const getMessage = (localeCode, localeMessages, key, substitutions) => {
...
+    const substitutedParts = parts.map((part: string) => {
       const subMatch = part.match(/\$(\d)/u);
       if (!subMatch) {
         return part;
       }
       const substituteIndex = Number(subMatch[1]) - 1;
       if (
-        (substitutions[substituteIndex] === null ||
-          substitutions[substituteIndex] === undefined) &&
+        (substitutions?.[substituteIndex] === null ||
```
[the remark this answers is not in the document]
DDDDDanica: yes makes sense. Typescript is strict and magical :))) thanks for showing me.

ui/ducks/app/app.ts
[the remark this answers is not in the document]
DDDDDanica: lollllll. Of course there's an old one XD. Gonna combine both 🙏

DDDDDanica:
@brad-decker thanks for review, just to make this MR easier to debug and lighter to deliver/review, i separate the `ui/ducks/app/app.ts` into this MR 
https://github.com/MetaMask/metamask-extension/pull/17454

brad-decker (approved):

pedronfigueiredo (approved):

PR #17374, opened 2023-01-24
Remove a notification for falling back from STX to regular swaps

dan437:
## Explanation
The notification below for falling back from STX to regular swaps wasn't really actionable, caused a bit of confusion and unnecessary overhead. All that to communicate an edge case that most users will not encounter. We are removing it to simplify user experience.

## Screenshots/Screencaps
Removed notification:
image


ui/pages/swaps/index.js
```diff
@@ -374,11 +363,6 @@ export default function Swap() {
     return <></>;
   }
 
-  const isStxNotEnoughFundsError =
-    currentSmartTransactionsError === stxErrorTypes.NOT_ENOUGH_FUNDS;
```
adonesky1, line 378: is this not an error we want to maintain?

darkwing (approved):

ui/pages/swaps/index.js
dan437: That's correct, we will fallback automatically to a regular swap if there is not enough funds for a gas fee via STX.

adonesky1 (approved):
LGTM!

PR #17390, opened 2023-01-24
feature: convert level 32 files to typescript

DDDDDanica:
## Pre-merge author checklist

- [x] I've clearly explained:


ui/helpers/constants/design-system.ts
```diff
@@ -54,6 +54,7 @@ export const COLORS = {
   GOERLI_INVERSE: 'goerli-inverse',
   SEPOLIA_INVERSE: 'sepolia-inverse',
 };
+
 export const BACKGROUND_COLORS = pick(COLORS, [
```
brad-decker, line 58: These are technically enums, and we didn't convert them to such. But if we do that we are going to have *a lot * of changes. Probably best  to do that in a separate PR down the road.

brad-decker: Note: ui/helpers/constants/routes.js are NOT enums, they are constants because the main object is a MAP of keys to values, versus something we'd be referencing.

For example COLORS from design system we use *everywhere* a Typography component can only ever have one COLORS assigned, so COLORS should be `Color` and the keys should be lowerCamelCased `Color.backgroundAlternative` @georgewrmarshall @garrettbear just a heads up, we aren't doing this change here.

brad-decker (dismissed):
LGTM good work @DDDDDanica

ui/helpers/constants/design-system.ts
DDDDDanica: I agree we should have a separate MR following to cover the colors. will work on that next !

brad-decker (approved):

ryanml (approved):

PR #17410, opened 2023-01-25
Selector performance improvements

FrederikBolding:
## Explanation
- Stopped calculating checksummed addresses in all selectors since the addresses are compared without case sensitivity.


montelaidev (dismissed):
LGTM

jpuri (approved):

brad-decker (approved):

PR #17413, opened 2023-01-25
delay chain validation

witmicko:
## Explanation

Fixing https://github.com/MetaMask/pm-security/issues/120
An arbitrary HTTP POST request is issued on wallet_addEthereumChain on any website. This happens even on not connected websites (e.g. [github.com](https://github.com/) ), even if the wallet is locked. It does not happen only if there is no wallet.

Fixed by making confirmation approve handler asynchronous.

### After

https://user-images.githubusercontent.com/3279642/219458534-086b49b5-37d2-4fa7-b57c-a36751529500.mov

## Pre-merge author checklist

- [ ] I've clearly explained:

### DIFF
--- a/app/_locales/en/messages.json
+++ b/app/_locales/en/messages.json
@@ -263,6 +263,9 @@
   "addToken": {
     "message": "Add token"
   },
+  "addingCustomNetwork": {
+    "message": "Adding Network"
+  },
   "address": {
     "message": "Address"
   },
@@ -1308,6 +1311,9 @@
     "message": "Stack:",
     "description": "Title for error stack, which is displayed for debugging purposes"
   },
+  "errorWhileConnectingToRPC": {
+    "message": "Error while connecting to the custom network."
+  },
   "ethGasPriceFetchWarning": {
     "message": "Backup gas price is provided as the main gas estimation service is unavailable right now."
   },
@@ -2032,6 +2038,9 @@
   },
   "mismatchedNetworkSymbol": {
     "message": "The submitted currency symbol does not match what we expect for this chain ID."
+  },
+  "mismatchedRpcChainId": {
+    "message": "Chain ID returned by the custom network does not match the submitted chain ID."
   },
   "mismatchedRpcUrl": {
     "message": "According to our records the submitted RPC URL value does not match a known provider for this chain ID."
--- a/app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
+++ b/app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
@@ -10,7 +10,6 @@ import {
   isPrefixedFormattedHexString,
   isSafeChainId,
 } from '../../../../../shared/modules/network.utils';
-import { jsonRpcRequest } from '../../../../../shared/modules/rpc.utils';
 
 const addEthereumChain = {
   methodNames: [MESSAGE_TYPE.ADD_ETHEREUM_CHAIN],
@@ -155,6 +154,7 @@ async function addEthereumChainHandler(
     if (currentChainId === _chainId && currentRpcUrl === firstValidRPCUrl) {
       return end();
     }
+
     // If this network is already added with but is not the currently selected network
     // Ask the user to switch the network
     try {
@@ -180,28 +180,6 @@ async function addEthereumChainHandler(
       }
     }
     return end();
-  }
-
-  let endpointChainId;
-
-  try {
-    endpointChainId = await jsonRpcRequest(firstValidRPCUrl, 'eth_chainId');
-  } catch (err) {
-    return end(
-      ethErrors.rpc.internal({
-        message: `Request for method 'eth_chainId on ${firstValidRPCUrl} failed`,
-        data: { networkErr: err },
-      }),
-    );
-  }
-
-  if (_chainId !== endpointChainId) {
-    return end(
-      ethErrors.rpc.invalidParams({
-        message: `Chain ID returned by RPC URL ${firstValidRPCUrl} does not match ${_chainId}`,
-        data: { chainId: endpointChainId },
-      }),
-    );
   }
 
   if (typeof chainName !== 'string' || !chainName) {
@@ -266,20 +244,18 @@ async function addEthereumChainHandler(
   }
 
   try {
-    await addCustomRpc(
-      await requestUserApproval({
-        origin,
-        type: MESSAGE_TYPE.ADD_ETHEREUM_CHAIN,
-        requestData: {
-          chainId: _chainId,
-          blockExplorerUrl: firstValidBlockExplorerUrl,
-          chainName: _chainName,
-          rpcUrl: firstValidRPCUrl,
-          ticker,
-        },
-      }),
-    );
-
+    const customRpc = await requestUserApproval({
+      origin,
+      type: MESSAGE_TYPE.ADD_ETHEREUM_CHAIN,
+      requestData: {
+        chainId: _chainId,
+        blockExplorerUrl: firstValidBlockExplorerUrl,
+        chainName: _chainName,
+        rpcUrl: firstValidRPCUrl,
+        ticker,
+      },
+    });
+    await addCustomRpc(customRpc);
     sendMetrics({
       event: 'Custom Network Added',
       category: EVENT.CATEGORIES.NETWORK,
--- a/test/e2e/tests/add-custom-network.spec.js
+++ b/test/e2e/tests/add-custom-network.spec.js
@@ -17,6 +17,232 @@ describe('Custom network', function () {
       },
     ],
   };
+
+  it('should show warning when adding chainId 0x1', async function () {
+    await withFixtures(
+      {
+        dapp: true,
+        fixtures: new FixtureBuilder()
+          .withPermissionControllerConnectedToTestDapp()
+          .build(),
+        ganacheOptions,
+        title: this.test.title,
+      },
+      async ({ driver }) => {
+        await driver.navigate();
+        await driver.fill('#password', 'correct horse battery staple');
+        await driver.press('#password', driver.Key.ENTER);
+
+        await driver.openNewPage('http://127.0.0.1:8080/');
+        await driver.executeScript(`
+          var params = [{
+            chainId: "0x1",
+            chainName: "Fake Ethereum Network",
+            nativeCurrency: {
+              name: "",
+              symbol: "ETH",
+              decimals: 18
+            },
+            rpcUrls: ["https://customnetwork.com/api/customRPC"],
+            blockExplorerUrls: [ "http://localhost:8080/api/customRPC" ]
+          }]
+          window.ethereum.request({
+            method: 'wallet_addEthereumChain',
+            params
+          })
+        `);
+        const windowHandles = await driver.waitUntilXWindowHandles(
+          3,
+          1000,
+          10000,
+        );
+
+        await driver.switchToWindowWithTitle(
+          'MetaMask Notification',
+          windowHandles,
+        );
+
+        await driver.clickElement({
+          tag: 'button',
+          text: 'Approve',
+        });
+
+        await driver.delay(1000);
+        const warningTxt =
+          'You are adding a new RPC provider for Ethereum Mainnet';
+
+        const warning = await driver.findElement({
+          tag: 'h4',
+          text: warningTxt,
+        });
+
+        assert.equal(await warning.getText(), warningTxt);
+        await driver.clickElement({
+          tag: 'button',
+          text: 'Approve',
+        });
+
+        const errMsg =
+          'Chain ID returned by the custom network does not match the submitted chain ID.';
+        const error = await driver.findElement({
+          tag: 'span',
+          text: errMsg,
+        });
+        assert.equal(await error.getText(), errMsg);
+
+        const approveBtn = await driver.findElement({
+          tag: 'button',
+          text: 'Approve',
+        });
+
+        assert.equal(await approveBtn.isEnabled(), false);
+        await driver.clickElement({
+          tag: 'button',
+          text: 'Cancel',
+        });
+      },
+    );
+  });
+  it("don't add bad rpc custom network", async function () {
+    await withFixtures(
+      {
+        dapp: true,
+        fixtures: new FixtureBuilder()
+          .withPermissionControllerConnectedToTestDapp()
+          .build(),
+        ganacheOptions,
+        title: this.test.title,
+      },
+      async ({ driver }) => {
+        await driver.navigate();
+        await driver.fill('#password', 'correct horse battery staple');
+        await driver.press('#password', driver.Key.ENTER);
+
+        await driver.openNewPage('http://127.0.0.1:8080/');
+        await driver.executeScript(`
+          var params = [{
+            chainId: "0x123",
+            chainName: "Antani",
+            nativeCurrency: {
+              name: "",
+              symbol: "ANTANI",
+              decimals: 18
+            },
+            rpcUrls: ["https://customnetwork.com/api/customRPC"],
+            blockExplorerUrls: [ "http://localhost:8080/api/customRPC" ]
+          }]
+          window.ethereum.request({
+            method: 'wallet_addEthereumChain',
+            params
+          })
+        `);
+        const windowHandles = await driver.waitUntilXWindowHandles(
+          3,
+          1000,
+          10000,
+        );
+
+        await driver.switchToWindowWithTitle(
+          'MetaMask Notification',
+          windowHandles,
+        );
+        await driver.clickElement({
+          tag: 'button',
+          text: 'Approve',
+        });
+
+        const errMsg =
+          'Chain ID returned by the custom network does not match the submitted chain ID.';
+        const error = await driver.findElement({
+          tag: 'span',
+          text: errMsg,
+        });
+        assert.equal(await error.getText(), errMsg);
+
+        const approveBtn = await driver.findElement({
+          tag: 'button',
+          text: 'Approve',
+        });
+
+        assert.equal(await approveBtn.isEnabled(), false);
+        await driver.clickElement({
+          tag: 'button',
+          text: 'Cancel',
+        });
+      },
+    );
+  });
+
+  it("don't add unreachable custom network", async function () {
+    await withFixtures(
+      {
+        dapp: true,
+        fixtures: new FixtureBuilder()
+          .withPermissionControllerConnectedToTestDapp()
+          .build(),
+        ganacheOptions,
+        title: this.test.title,
+      },
+      async ({ driver }) => {
+        await driver.navigate();
+        await driver.fill('#password', 'correct horse battery staple');
+        await driver.press('#password', driver.Key.ENTER);
+
+        await driver.openNewPage('http://127.0.0.1:8080/');
+        await driver.executeScript(`
+          var params = [{
+            chainId: "0x123",
+            chainName: "Antani",
+            nativeCurrency: {
+              name: "",
+              symbol: "ANTANI",
+              decimals: 18
+            },
+            rpcUrls: ["https://doesntexist.abc/customRPC"],
+            blockExplorerUrls: [ "http://localhost:8080/api/customRPC" ]
+          }]
+          window.ethereum.request({
+            method: 'wallet_addEthereumChain',
+            params
+          })
+        `);
+        const windowHandles = await driver.waitUntilXWindowHandles(
+          3,
+          1000,
+          10000,
+        );
+
+        await driver.switchToWindowWithTitle(
+          'MetaMask Notification',
+          windowHandles,
+        );
+        await driver.clickElement({
+          tag: 'button',
+          text: 'Approve',
+        });
+
+        const error = await driver.findElement({
+          tag: 'span',
+          text: 'Error while connecting to the custom network.',
+        });
+        assert.equal(
+          await error.getText(),
+          'Error while connecting to the custom network.',
+        );
+        const approveBtn = await driver.findElement({
+          tag: 'button',
+          text: 'Approve',
+        });
+
+        assert.equal(await approveBtn.isEnabled(), false);
+        await driver.clickElement({
+          tag: 'button',
+          text: 'Cancel',
+        });
+      },
+    );
+  });
+
   it('add custom network and switch the network', async function () {
     await withFixtures(
       {
@@ -85,7 +311,6 @@ describe('Custom network', function () {
 
         await driver.clickElement({ tag: 'button', text: 'Close' });
         await driver.clickElement({ tag: 'button', text: 'Approve' });
-
         await driver.clickElement({
           tag: 'h6',
           text: 'Switch to Arbitrum One',
--- a/test/e2e/tests/add-custom-network.spec.js
+++ b/test/e2e/tests/add-custom-network.spec.js
@@ -129,11 +129,7 @@ describe('Custom network', function () {
             params
           })
         `);
-        const windowHandles = await driver.waitUntilXWindowHandles(
-          3,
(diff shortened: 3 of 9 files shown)


app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
```diff
@@ -298,6 +276,27 @@ async function addEthereumChainHandler(
   } catch (error) {
     return end(error);
   }
+  let endpointChainId;
+
+  try {
+    endpointChainId = await jsonRpcRequest(firstValidRPCUrl, 'eth_chainId');
```
Gudahtt, line 282: If either this check or the subsequent one fails, we would want to remove the custom RPC endpoint from the user's wallet because it failed validation.

We should also probably prevent the user from using this network while this validation is ongoing.

Gudahtt: Update: we talked and decided delaying adding the chain was the easier path. This resolves these concerns.

NEllusion (dismissed):
Code looks good, and tested that the fix solves the issue 🎉

test/e2e/tests/chain-interactions.spec.js
```diff
@@ -48,6 +48,14 @@ describe('Chain Interactions', function () {
 
         // approve add chain, cancel switch chain
         await driver.clickElement({ text: 'Approve', tag: 'button' });
+
+        await driver.delay(1000);
+        windowHandles = await driver.getAllWindowHandles();
```
Gudahtt, line 53: Why was this additional pause and request for window handles required? I can see that the webdriver was already pointing at the notification popup, but it's not clear to me what happens here.

Is the popup closing and re-opening again?

witmicko: window id is changing, approve add network has different ID than confirm switch network. Added pause to catch this change, witouth these changes it was failing with `window not found`

Gudahtt: Ok, that implies that the window is closing and re-opening.

I'm having trouble understanding how your changes relate to this though. You didn't adjust the switch network steps at all.

Perhaps some part of your changes is causing the window to close sooner than it had before? Ideally it would stay open long enough for the switch network prompt to appear.

witmicko: My change makes it stay open longer I think as I moved validation and adding network to after user approves. Maybe that is why window was missing in test, because previously new window would have been already opened by then?
Let me know if you would like me to make some changes please.

legobeat: I don't have any insight to share on the actual issue, just noting that it seems optimistic with a 1s window here. Starvation scenarios are not that rare.

Gudahtt: We should definitely try to prevent the window from closing. This would be a substantial UX degradation, and would make these e2e tests flaky.

brad-decker (commented):
This looks good to me from a code perspective. i'll QA now

brad-decker:
We should probably get some UX feedback on how we display the loading. Right now this is leading to two lines of text on the button and causing a little bit of layout jank. See the video.

@bschorchit do you know of what other confirmations do when loading (if anything)? 

[Screencast from 02-02-2023 12:10:53 PM.webm](https://user-images.githubusercontent.com/4448075/216409952-10694ce2-a415-46fd-8088-62c0d2e9526e.webm)

bschorchit:
> @bschorchit do you know of what other confirmations do when loading (if anything)?
> 
Oh good question, we usually use the spinner on the full screen but that's when loading the UI. I don't think I've seen a loading after submitting yet. Any insights or suggestion on this @holantonela ?

ui/pages/confirmation/templates/add-ethereum-chain.js
```diff
@@ -340,6 +367,7 @@ function getValues(pendingApproval, t, actions, history) {
         actions.addCustomNetwork(pendingApproval.requestData);
         history.push(DEFAULT_ROUTE);
       }
+      return {};
```
Gudahtt, line 370: Nit: Perhaps better to return `undefined` if the return value is not expected to be used

```suggestion
      return undefined;
```

ui/pages/confirmation/templates/add-ethereum-chain.js
```diff
@@ -331,7 +333,32 @@ function getValues(pendingApproval, t, actions, history) {
...
+          pendingApproval.id,
+          ethErrors.rpc.resourceUnavailable().serialize(),
+        );
+      }
+
+      if (pendingApproval.requestData.chainId !== endpointChainId) {
+        console.error(
+          `Chain ID returned by RPC URL ${customRpcUrl} does not match ${endpointChainId}`,
+        );
+        return actions.rejectPendingApproval(
```
Gudahtt, line 356: This is a clever solution to the double-window-popup problem :+1: Great work

The one major flaw left is that this experience has been made more confusing for the user because of this change, specifically in the case where this validation fails.

Previously if this validation step failed, the dapp would get an error and the user would see nothing. In this circumstance, maybe the dapp handled it well and maybe they didn't, but at least MetaMask isn't doing anything to mislead/confuse users.

But now if validation fails, it will appear to a user that the operation succeeded. Only the dapp and the web console are told of the error. The user may not find out.

It would be better to show the user the error somehow. But I don't know of an easy way to do this :thinking:

Gudahtt: Maybe this UX concern is not a blocker, and we can solve it in a later PR. We should at least track this problem as an issue if it's not solved here though.

NicholasEllul: As a temporary workaround until the other work to support this is complete, can we infer if the operation failed by following up with a second check to see what chain ID we are using?

Similar to whats done in:

https://github.com/MetaMask/metamask-extension/blob/cdc1bce6884295e97c63ee9758646fc4287aa781/ui/components/app/network-display/network-display.js#L31-L34

NicholasEllul: `state.metamask.provider` gives us access to the following to check the chain ID: 


Screenshot 2023-02-15 at 4 34 24 PM

app/_locales/en/messages.json
```diff
@@ -2021,6 +2024,9 @@
   "mismatchedNetworkSymbol": {
     "message": "The submitted currency symbol does not match what we expect for this chain ID."
   },
+  "mismatchedRpcChainId": {
+    "message": "Chain ID returned but RPC does not match the submitted chain ID."
```
NicholasEllul, line 2028: Should this read

`Chain ID returned by the RPC URL`

ui/pages/confirmation/templates/add-ethereum-chain.js
NicholasEllul: Update: Michał and I discussed offline and he has found another solution to this problem

ui/pages/confirmation/templates/add-ethereum-chain.js
```diff
@@ -331,7 +347,35 @@ function getValues(pendingApproval, t, actions, history) {
...
+      }
+
+      if (pendingApproval.requestData.chainId !== endpointChainId) {
+        console.error(
+          `Chain ID returned by RPC URL ${customRpcUrl} does not match ${endpointChainId}`,
+        );
+        // actions.rejectPendingApproval(
+        //   pendingApproval.id,
+        //   ethErrors.rpc.invalidInput().serialize(),
+        // );
```
NicholasEllul, line 373: Do we still want to be rejecting the pending approval? Or do we no longer need this now that the user is forced in to clicking cancel themselves

ui/pages/confirmation/templates/add-ethereum-chain.js
```diff
@@ -331,7 +347,35 @@ function getValues(pendingApproval, t, actions, history) {
...
+        return {
+          error: [MISMATCHED_NETWORK_RPC_CHAIN_ID],
+        };
```
NicholasEllul, line 376: This function now returns one of the following depending on if the function succeeds or fails:

* Above we return `actions.rejectPendingApproval` during a failure which is a [thunk object](https://github.com/MetaMask/metamask-extension/blob/33cc8d587aad05c0b41871ba3676676a3ce5680e/ui/store/actions.ts#L3731). (Note: I don't think this value is actually used anywhere)
* Here we return a javascript object containing the `errors` key with an array of strings
* Below we return a javascript object containing an empty`{}`

Is there a way we can consolidate these to provide a consistent return interface? Checking usage of the code, it seems the return value from this function hasn't used in the past, so we have some flexibility to set things up.

ui/pages/confirmation/confirmation.js
```diff
@@ -41,7 +41,7 @@ import {
   getTemplateState,
 } from './templates';
 
-// TODO(rekmarks): This component and all of its sub-components should probably
+// TODO(remarks): This component and all of its sub-components should probably
```
danjm, line 44: `rekmarks` is actually a github username

app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
```diff
@@ -182,28 +182,6 @@ async function addEthereumChainHandler(
     return end();
   }
 
-  let endpointChainId;
```
danjm, line 185: why is this section of code removed?

witmicko: > why is this section of code removed?

these checks were moved to the confirmation template

app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
```diff
@@ -266,20 +244,18 @@ async function addEthereumChainHandler(
...
+      type: MESSAGE_TYPE.ADD_ETHEREUM_CHAIN,
+      requestData: {
+        chainId: _chainId,
+        blockExplorerUrl: firstValidBlockExplorerUrl,
+        chainName: _chainName,
+        rpcUrl: firstValidRPCUrl,
+        ticker,
+      },
+    });
+    addCustomRpc(customRpc);
```
Gudahtt, line 258: Missing an `await` here

ui/pages/confirmation/confirmation.js
```diff
@@ -258,14 +260,25 @@ export default function ConfirmationPage({
...
 
-  const handleSubmit = () =>
-    templateState[pendingConfirmation.id]?.useWarningModal
+  const handleSubmit = async () => {
+    setLoading(true);
+    const submitResult = await (templateState[pendingConfirmation.id]
+      ?.useWarningModal
       ? setShowWarningModal(true)
-      : templatedValues.onSubmit(
+      : await templatedValues.onSubmit(
```
Gudahtt, line 268: Hmm, was this extra `await` needed? This is already being `await`-ed three lines up from here

ui/pages/confirmation/confirmation.js
```diff
@@ -258,14 +260,25 @@ export default function ConfirmationPage({
...
           hasInputState(pendingConfirmation.type)
             ? inputStates[MESSAGE_TYPE.SNAP_DIALOG_PROMPT]
             : null,
-        );
+        ));
+
+    if (submitResult?.success) {
```
Gudahtt, line 274: It looks like this component will be stuck in the "loading" state if the submit result is missing this property. Yet only `add-ethereum-chain` returns this. The other confirmations would all get stuck.

ui/pages/confirmation/confirmation.js
```diff
@@ -362,6 +375,18 @@ export default function ConfirmationPage({
         onCancel={templatedValues.onCancel}
         submitText={templatedValues.submitText}
         cancelText={templatedValues.cancelText}
+        loadingText={templatedValues.loadingText}
+        loading={loading}
+        submitAlerts={submitAlerts.map((alert, idx) => (
+          <Callout
+            key={alert.id}
+            severity={alert.severity}
+            dismiss={() => dismissAlert(alert.id)}
```
Gudahtt, line 384: There appears to be no way to dismiss these alerts. The `dismissAlert` function here is not for the "submit alerts" (which are component state), it's for the "alert state" reducer. When this line is called, it won't work because the reducer isn't tracking this id.

ui/pages/confirmation/templates/add-ethereum-chain.js
```diff
@@ -331,7 +361,29 @@ function getValues(pendingApproval, t, actions, history) {
     ],
     cancelText: t('cancel'),
     submitText: t('approveButtonText'),
+    loadingText: t('addingCustomNetwork'),
     onSubmit: async () => {
+      let endpointChainId;
+      try {
+        endpointChainId = await jsonRpcRequest(customRpcUrl, 'eth_chainId');
+      } catch (_err) {
```
Gudahtt, line 369: Nit: we should at least confirm that this is a network error, and throw otherwise. That would help ensure that we don't swallow errors in our code.

ui/pages/confirmation/confirmation.js
witmicko: this is done this way to disable the button so user can only click 'Cancel'

Gudahtt: Right, but for the non-`add-ethereum-chain` confirmations there will be no way to proceed even if there is no error

app/_locales/en/messages.json
```diff
@@ -1352,6 +1355,9 @@
     "message": "Stack:",
     "description": "Title for error stack, which is displayed for debugging purposes"
   },
+  "errorWhileConnectingToRPC": {
+    "message": "Error while connecting to the custom RPC."
+  },
```
NicholasEllul, line 1360: Can we use the terminology `Network` here and in the translation above to align with the terminology used for the other UI elements?

NEllusion (commented):
Tested with reproduction steps listed in the GitHub issue and the change works as expected

Minor comment: We now end up showing `Adding RPC...` after the request fails even when the process of adding an RPC is terminated. Is there a way we can fall back to showing `approve` in a disabled state? Alternatively, could we get away with simply re-activating the button and having the error show up again?

ui/pages/confirmation/confirmation.js
```diff
@@ -258,14 +262,24 @@ export default function ConfirmationPage({
...
             ? inputStates[MESSAGE_TYPE.SNAP_DIALOG_PROMPT]
             : null,
-        );
+        ));
+
+    if (submitResult?.error) {
+      setLoadingText(templatedValues.submitText);
+      setSubmitAlerts(submitResult.error);
+    } else {
+      setLoading(false);
```
Gudahtt, line 280: It looks like `submitResult.success` is no longer being used. Perhaps we could simplify this to just return the error, and drop the `success` boolean flag altogether?

ui/pages/confirmation/confirmation.js
```diff
@@ -258,14 +262,24 @@ export default function ConfirmationPage({
     return INPUT_STATE_CONFIRMATIONS.includes(type);
   };
 
-  const handleSubmit = () =>
-    templateState[pendingConfirmation.id]?.useWarningModal
+  const handleSubmit = async () => {
+    setLoading(true);
+    const submitResult = await (templateState[pendingConfirmation.id]
```
Gudahtt, line 267: Nit: Sorry I know I requested that the inner `await` be removed earlier. But I hadn't considered which we needed here, I just noticed the redundancy. On second look, it does seem to make more sense for the `await` to just be used for `onSubmit`, since `useWarningModal` is a boolean property not a Promise

ui/pages/confirmation/confirmation.js
```diff
@@ -258,14 +262,24 @@ export default function ConfirmationPage({
...
+      ?.useWarningModal
       ? setShowWarningModal(true)
       : templatedValues.onSubmit(
           hasInputState(pendingConfirmation.type)
             ? inputStates[MESSAGE_TYPE.SNAP_DIALOG_PROMPT]
             : null,
-        );
+        ));
+
+    if (submitResult?.error) {
```
Gudahtt, line 276: We're missing this "submit result error handling" step for confirmations that include a warning modal. For the warning modal case, `onSubmit` is called in the `onSubmit` prop passed to `ConfirmationWarningModal` below.

Perhaps we can move this result handling to a separate function/wrapper function, and use it in both places?

ui/pages/confirmation/templates/index.js
```diff
@@ -33,6 +33,8 @@ const ALLOWED_TEMPLATE_KEYS = [
   'onSubmit',
   'networkDisplay',
   'submitText',
+  'loadingText',
+  'submitAlerts',
```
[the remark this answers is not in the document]
Gudahtt: Sorry for the confusion; I meant this to apply just to `'submitAlerts'`. The `'loadingText'` property is being used in the `add-ethereum-chain` template.

ui/pages/confirmation/confirmation.js
Gudahtt: Could you remove this outer `await`, now that the inner one has been restored?

ui/pages/confirmation/templates/add-ethereum-chain.js
```diff
@@ -331,7 +361,29 @@ function getValues(pendingApproval, t, actions, history) {
...
     onSubmit: async () => {
+      let endpointChainId;
+      try {
+        endpointChainId = await jsonRpcRequest(customRpcUrl, 'eth_chainId');
+      } catch (err) {
+        console.error(
+          `Request for method 'eth_chainId on ${customRpcUrl} failed`,
+        );
+
+        if (err.message === 'Failed to fetch') {
```
Gudahtt, line 374: Hmm. I'm not confident that all fetch errors would have this exact message. It looks like they can have a variety of different messages, especially if we consider that custom RPC endpoints can return custom errors for auth failures. I guess my suggestion for identifying the error as being a "connectivity" is impractical for now.

ui/pages/confirmation/confirmation.js
```diff
@@ -258,14 +262,28 @@ export default function ConfirmationPage({
...
+      ?.useWarningModal
       ? setShowWarningModal(true)
-      : templatedValues.onSubmit(
+      : await templatedValues.onSubmit(
           hasInputState(pendingConfirmation.type)
             ? inputStates[MESSAGE_TYPE.SNAP_DIALOG_PROMPT]
             : null,
-        );
+        ));
+    handleSubmitResult(submitResult);
```
Gudahtt, line 285: It seems that this is getting called even in the `setShowWarningModal` case. In that case, `submitResult` would be `undefined`, which results in `setLoading(false)` being called. This effectively breaks the loading state for confirmations with warnings (at least I think so - I didn't test this.

It might be simpler to change this ternary into if-else blocks, so that we can ensure this is only called when `onSubmit` is called.

test/e2e/tests/add-custom-network.spec.js
```diff
@@ -17,6 +17,232 @@ describe('Custom network', function () {
...
+          'MetaMask Notification',
+          windowHandles,
+        );
+
+        await driver.clickElement({
+          tag: 'button',
+          text: 'Approve',
+        });
+
+        await driver.delay(1000);
```
Gudahtt, line 70: Nit: Is this delay needed? The `findElement` call will automatically wait until it finds the requested element, we don't usually need these explicit delays.

```suggestion
```

test/e2e/tests/add-custom-network.spec.js
```diff
@@ -17,6 +17,232 @@ describe('Custom network', function () {
...
+          text: 'Approve',
+        });
+
+        const errMsg =
+          'Chain ID returned by the custom network does not match the submitted chain ID.';
+        const error = await driver.findElement({
+          tag: 'span',
+          text: errMsg,
+        });
+        assert.equal(await error.getText(), errMsg);
```
Gudahtt, line 91: Nit: These assertions are redundant; the `findElement` call will throw unless it finds something with that exact text. It's already asserting this for us. Removing these assertions saves another round-trip communication with the browser, reducing test runtime and flakiness.

```suggestion
        await driver.findElement({
          tag: 'h4',
          text: warningTxt,
        });

        await driver.clickElement({
          tag: 'button',
          text: 'Approve',
        });

        const errMsg =
          'Chain ID returned by the custom network does not match the submitted chain ID.';
        await driver.findElement({
          tag: 'span',
          text: errMsg,
        });
```

Gudahtt: I see there are a few examples of this in the other tests as well

test/e2e/tests/add-custom-network.spec.js
```diff
@@ -17,6 +17,232 @@ describe('Custom network', function () {
...
+          window.ethereum.request({
+            method: 'wallet_addEthereumChain',
+            params
+          })
+        `);
+        const windowHandles = await driver.waitUntilXWindowHandles(
+          3,
+          1000,
+          10000,
+        );
```
Gudahtt, line 58: Nit: Do we need to double the timeout here? 5 seconds is the default; I would expect the notification to appear in that timeframe.
```suggestion
        const windowHandles = await driver.waitUntilXWindowHandles(3);
```

test/e2e/tests/add-custom-network.spec.js
```diff
@@ -17,6 +17,232 @@ describe('Custom network', function () {
       },
     ],
   };
+
+  it('should show warning when adding chainId 0x1', async function () {
```
Gudahtt, line 21: Maybe this title should be updated to reflect that the test is for the scenario where a warning is shown first, followed by an error. It's good that we have this case covered, but I hadn't expected that from this description.

ui/pages/confirmation/templates/index.js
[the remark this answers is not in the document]
Gudahtt: @witmicko Looks like this is still unaddressed; this is the problem I referred to on Zoom.

It's not breaking the e2e test because the validation is broken (pre-existing issue, I'll address it now in a separate PR)

test/e2e/tests/add-custom-network.spec.js
```diff
@@ -17,6 +17,221 @@ describe('Custom network', function () {
...
+          window.ethereum.request({
+            method: 'wallet_addEthereumChain',
+            params
+          })
+        `);
+        const windowHandles = await driver.waitUntilXWindowHandles(
+          3,
+          1000,
+          10000,
+        );
```
Gudahtt, line 136: Nit: Shouldn't need double timeout here

```suggestion
        const windowHandles = await driver.waitUntilXWindowHandles(3);
```

test/e2e/tests/add-custom-network.spec.js
```diff
@@ -17,6 +17,221 @@ describe('Custom network', function () {
...
```
Gudahtt, line 205: Nit: Shouldn't need double timeout here
```suggestion
        const windowHandles = await driver.waitUntilXWindowHandles(3);
```

Gudahtt (commented):
Great work! Last blocking issue is to add `loadingText` as an allowed return value of `getValues` ([see here](https://github.com/MetaMask/metamask-extension/pull/17413#discussion_r1126741375))

Gudahtt (commented):
LGTM! Though now there's a conflict with the coverage target file

Gudahtt (dismissed):

Gudahtt (approved):

danjm:
Created three unit test creation issues for either the extension platform team or the confirmations team to pick up:
https://github.com/MetaMask/metamask-extension/issues/18060
https://github.com/MetaMask/metamask-extension/issues/18059
https://github.com/MetaMask/metamask-extension/issues/18058

danjm (approved):
LGTM! Nice work on the e2e tests

Issue #17213, opened 2023-01-13
Typescript Level 34


hilvmason:
[Extension TypeScript Migration Status dashboard
](https://metamask.github.io/metamask-extension-ts-migration-dashboard/)
Level 34 Tasks:

- [x] ui/helpers/utils/conversions.util.js
- [x] shared/constants/hardware-wallets.js
- [ ] shared/modules/fetch-with-timeout.js
- [ ] shared/modules/transaction.utils.js
- [x] ui/store/actionConstants.js
- [x] app/scripts/constants/on-ramp.js
- [x] app/scripts/constants/contracts.js
- [ ] ui/helpers/utils/transactions.util.js
- [ ] ui/helpers/utils/util.js
- [ ] shared/modules/swaps.utils.js

brad-decker:
Closing this issue cause it won’t make sense once we pick this back up with the dashboard reconfigured.

PR #17442, opened 2023-01-26
Convert ui/store/actionConstants.js -> Typescript

ryanml:
## Explanation
Part of Level 34 of our Typescript conversion project: https://github.com/MetaMask/metamask-extension/issues/17213

## Manual Testing Steps
1. Ensure automated tests pass

## Pre-merge author checklist

- [X] I've clearly explained:


ui/store/constants/actions.ts
```diff
@@ -0,0 +1,101 @@
+export enum ActionConstants {
```
brad-decker, line 1: @Gudahtt thoughts on this. IT makes sense as an enum for sure... just wanna get your take. 

@ryanml i'm not sure I am a fan of renaming this 'actions' as we already have an actions file. And the name of this enum is now a little weird. It is actually describing the possible values for the 'type' parameter of an action so ActionType might b e a better name.

Lets get @Gudahtt 's opinion here but I don't know that we need to make this an enum for this reason:

In my upcoming PR i migrate our store.js file to to ts and use the reduxjs/toolkit helper to build the store, which gives us the ability to derive the state and dispatch type from it. As we move more and more of our reducers to use reduxjs/toolkit we won't have a need for these action constants anymore and the dispatch method automatic typing will kick in for making sure we don't dispatch an action that doesn't exist in our reducers.

My opinion at this point is we should probably aim for a more minimal refactor to typescript, keeping the constants in the original casing, however I may be a bit biased here and I'm going to lean on Mark for a second opinion

brad-decker (changes requested):

ui/store/constants/actions.ts
Gudahtt: "action type" does seem like a better description of what this is. I don't have strong feelings either way on these other topics though.

Worth noting that even after we move everything to use Redux toolkit, we will have some cross-slice actions, so I expect we'll still have need of `actions.js` or something like it. But if we move to using `createAction` over manually created actions, that eliminates the need for these constants.

danjm: @brad-decker I agree that ActionType would be a good name. I also don't have too strong of an opinion on the changes made in this PR.

I agree with the thought you shared in slack that "it doesn't take away anything from our ability to adopt toolkit"

Meanwhile, we could ask what are the benefits of the changes? What advantages does having these strings typed as enums give us over the present state on `develop`?

brad-decker: If we were going to create an maintain our own type for the `type` key of actions it would have a ton of value, but as it stands I don't think we will. I'm going to sleep on it and think of the pros and cons a little more.

brad-decker (changes requested):
@ryanml I would like to move this file back to actionConstants for the following reasons:

1. At least for our immediate goals changing the filename will disturb the ts migration dashboard's layout of modules, we will essentially not get credit for this file conversion (it'll disappear the same ways the conversion.utils.js file in the UI helpers folder did). 
2. I think having two files named 'actions' is a bit confusing. We can rename the one that has thunks in it soon, then renaming this to actions will probably make more thematic sense.

For the following reasons I think we should revert the change to using an enum:

1. While technically correct (an action can only be of ONE type thus an enum), we will hopefully be whittling away these actions to a fraction of what exists today. 
2. We will hopefully be refactoring these to use the reduxjs toolkit helpers for action creation (which come with fancy types that can be inferred from the configureStore method to give us typed dispatch method).

brad-decker (approved):
LGTM!

jpuri (approved):

PR #17446, opened 2023-01-26
Convert actions.js to typescript

brad-decker:
## Explanation
This Pull request converts our actions file to typescript, and does minimal typing of some background state within the various actions to provide SOME assurances. The platform team will be devising a strategy for typing the background state structure in a globally accessible way and incrementally implementing sound types for background state. In addition as we adopt redux toolkit we will incrementally adopt typed strucutres for the slices of redux state.

This is the culmination of ten different pull requests to get the actions file into a state that passes type checks. 

## Manual Testing Steps
1. Review PR especially any test changes, there should be no functional test changes and this PR should not make any functional changes to UX. 

## Pre-merge author checklist

- [x] I've clearly explained:


pedronfigueiredo (dismissed):

ui/store/actions.ts
```diff
@@ -3558,25 +4287,32 @@ export function fetchSmartTransactionFees(
         payload: null,
       });
       return smartTransactionFees;
-    } catch (e) {
-      log.error(e);
-      if (e.message.startsWith('Fetch error:')) {
-        const errorObj = parseSmartTransactionsError(e.message);
+    } catch (err) {
+      log.error(err);
```
DDDDDanica, line 4291: can use `logErrorWithMessage(err);`? :)

ui/pages/permissions-connect/permissions-connect.container.js
```diff
@@ -159,7 +158,6 @@ const mapDispatchToProps = (dispatch) => {
       );
     },
     getRequestAccountTabIds: () => dispatch(getRequestAccountTabIds()),
-    getCurrentWindowTab: () => dispatch(getCurrentWindowTab()),
```
DDDDDanica, line 162: Just curious about the purpose of this removal ~

ui/store/actions.ts
```diff
@@ -976,95 +1239,98 @@ export function updateAndApproveTx(txData, dontShowLoadingIndicator) {
...
-      network,
-      chainId,
-    );
-    const otherUnconfirmedActions = unconfirmedActions.filter(
-      (tx) => tx.id !== id,
-    );
+export function completedTx(
+  txId: number,
+): ThunkAction<void, MetaMaskReduxState, unknown, AnyAction> {
+  return (dispatch) => {
```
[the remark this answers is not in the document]
brad-decker: Sure! 

More context: https://github.com/MetaMask/metamask-extension/pull/17446/files/c32565e3c829c545d3a588053d809e7460a2dd6a#diff-ec46948cbcd32a43786d44f2477f99b7ebd2ab91eab6df44b10df93876e127aa

When I rebased your work on the txHelper conversion I got an error here because we were misusing it. The network and chainId should be the 7th and 8th prop but we are passing them as the 5th and 6th. So i investigated what bugs this might cause...turns out, none...because we only use the removed code to get a count of *other* unapproved transactions. in the app reducer we use this count to determine if we should set the 'shouldClosed' boolean in state. Presumably we once used this state key to decide if a popup window should close (if there are other pending messages in a popup we just move to the next if there are none the popup closes) however this logic is now all handled in the home component. shouldClose is never accessed. So any bugs we might have had are irrelevant because this is essentially a wasteful operation. Not dead cause it is still ran but its result is not used. We DO use appState.txId, so we still need this action to clear that value when the user confirms or rejects the active conf.

ui/store/actions.ts
```diff
@@ -1452,8 +1781,10 @@ export function updateMetamaskState(newState) {
 
     const newAddressBook = newState.addressBook?.[newProvider?.chainId] ?? {};
     const oldAddressBook = currentState.addressBook?.[provider?.chainId] ?? {};
-    const newAccounts = getMetaMaskAccounts({ metamask: newState });
-    const oldAccounts = getMetaMaskAccounts({ metamask: currentState });
+    const newAccounts: { [address: string]: Record<string, any> } =
+      getMetaMaskAccounts({ metamask: newState });
+    const oldAccounts: { [address: string]: Record<string, any> } =
+      getMetaMaskAccounts({ metamask: currentState });
```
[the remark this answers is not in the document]
brad-decker: We don't have a current type (to my knowledge). I would like to split up the background type work into chunks, defining exact shapes of data and then replacing these intermediate types with a properly researched type in an iterative way. What you suggest **is the correct approach**, but I don't think doing the background typing is a part of this PR

ui/pages/permissions-connect/permissions-connect.container.js
brad-decker: It was unused code, and should have been a part of the previous PRs.

ui/store/actions.ts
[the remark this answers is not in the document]
DDDDDanica: thank you, super clear !

ui/store/actions.ts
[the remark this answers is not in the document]
DDDDDanica: sweet, we can come back to this when touching background type then ~

DDDDDanica (approved):

ui/ducks/send/send.js
```diff
@@ -571,6 +571,8 @@ export const computeEstimatedGasLimit = createAsyncThunk(
  * the send slice. It returns the values that might change from this action and
  * those values are written to the slice in the `initializeSendState.fulfilled`
  * action handler.
+ *
```
digiwand, line 574: NBD two comments that reference `actions.js` instead of `actions.ts`. Commenting here because 1 instance is in this file

Screen Shot 2023-02-04 at 12 32 42 AM

digiwand (approved):
🙌

ui/ducks/send/send.js
brad-decker: Thanks for pointing that out. The files-to-convert does need to be .js cause the dashboard will turn the box green for this file cause its now .ts :D

Issue #17429, opened 2023-01-26
[Bug]: Switching from Binance chain to any other Network displays Token symbol and Picture from BNB


seaona:
### Describe the bug

**Problem**: when I am on Binance Chain, if I then switch to another network I can still see the Token symbol and picture from BNB

https://user-images.githubusercontent.com/54408225/214825089-c1922920-1a19-488c-aa11-f259a8af65ba.mp4

### Steps to reproduce

(Edit, added by danjm, copied from the comment below)
0. "Turn of the the following Privacy setting, Security & Privacy -> Transactions -> Show balance and token price checker"

1. Add Binance Chain network
2. Change to Binance Chain network
3. Now add Polygon network
4. Switch to Polygon network -- BNB symbol and picture are still displayed

Note: I have all Privacy features OFF (not sure if it's related)
Note2: the balances are correct for each network

### Version

10.24.1

### Build type

None

### Browser

Chrome

### Operating system

Linux

bschorchit:
I have been seeing this more frequently happening to me as well.

bschorchit:
The same just happened to me now.
I was in BNB, turned off the token balance and price checking toggle and then changed to Ethereum (screenshot). I then went to settings and turned the same toggle on, closed the settings and saw the token name and symbol being updated on the home screen.

![Screenshot 2023-01-26 at 17 18 36](https://user-images.githubusercontent.com/34968356/214942190-dcb6ac55-8d13-43b0-adeb-8c682ea81357.png)

PeterYinusa:
Steps to repro:

1. Turn of the the following Privacy setting, `Security & Privacy` -> `Transactions` -> `Show balance and token price checker`
2. Following the steps [above](https://github.com/MetaMask/metamask-extension/issues/17429#issue-1557994680)

seaona:
A couple of more findings - probably already figured it out:

This happens to any network - not exclusively to Binance:

https://user-images.githubusercontent.com/54408225/215042773-270c7023-0919-453d-b80c-bdbb76ca8cdb.mp4

I encountered The error `StreamProdiver - unknown response id for response [Object Object]` error, while making a tx and seeing the wrong network symbol. MM stayed loading indefinitely until I refreshed.
Not sure if this one is related to this, or do you think we open a separate bug for that? @PeterYinusa @danjm 

https://user-images.githubusercontent.com/54408225/215043263-62fd6afe-22da-4720-8311-c88d106c4e66.mp4

[...]

PR #17450, opened 2023-01-27
Use network provider state, instead of CurrencyRateController state, to select 'nativeCurrency'

danjm:
Alternative fix for #17429 

I think this works because `state.metamask.nativeCurrency` is always set from the provider config: https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/metamask-controller.js#L1020-L1028

Further explanation:

The bug from the user perspective is that if the "Show balance and token price checker" is off, then the network logo and "currency symbol" / "network ticker"  will be incorrect after switching networks.

The reason for that is that we don't execute any of the code in the in `CurrencyRateController` `updateExchangeRates` function if that toggle is off. That code updates the `nativeCurrency` property in state, and that property is used throughout the UI to decide which logo and  "currency symbol" / "network ticker" to show.

The `nativeCurrency` property is, however, always set using data from the network provider state. This happens in the metamask controller on two occasions: when the `CurrencyRateController` is initialized, and when the network changes. So, I believe we can safely replace all references to `state.metamask.nativeCurrency` with `state.metamask.provider.ticker`, and it will be functionally equivalent.

### DIFF
--- a/ui/ducks/metamask/metamask.js
+++ b/ui/ducks/metamask/metamask.js
@@ -10,6 +10,7 @@ import {
   accountsWithSendEtherInfoSelector,
   checkNetworkAndAccountSupports1559,
   getAddressBook,
+  getUseCurrencyRateCheck,
 } from '../../selectors';
 import { updateTransactionGasFees } from '../../store/actions';
 import { setCustomGasLimit, setCustomGasPrice } from '../gas/gas.duck';
@@ -307,7 +308,10 @@ export function getConversionRate(state) {
 }
 
 export function getNativeCurrency(state) {
-  return state.metamask.provider.ticker;
+  const useCurrencyRateCheck = getUseCurrencyRateCheck(state);
+  return useCurrencyRateCheck
+    ? state.metamask.nativeCurrency
+    : state.metamask.provider.ticker;
 }
 
 export function getSendHexDataFeatureFlagState(state) {
--- a/ui/ducks/metamask/metamask.js
+++ b/ui/ducks/metamask/metamask.js
@@ -307,7 +307,7 @@ export function getConversionRate(state) {
 }
 
 export function getNativeCurrency(state) {
-  return state.metamask.nativeCurrency;
+  return state.metamask.provider.ticker;
 }
 
 export function getSendHexDataFeatureFlagState(state) {
--- a/ui/pages/send/gas-display/gas-display.js
+++ b/ui/pages/send/gas-display/gas-display.js
@@ -60,9 +60,10 @@ export default function GasDisplay({ gasError }) {
   const useCurrencyRateCheck = useSelector(getUseCurrencyRateCheck);
   const { showFiatInTestnets, useNativeCurrencyAsPrimaryCurrency } =
     useSelector(getPreferences);
-  const { nativeCurrency, provider, unapprovedTxs } = useSelector(
+  const { provider, unapprovedTxs } = useSelector(
     (state) => state.metamask,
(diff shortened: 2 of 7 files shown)


ui/pages/send/gas-display/gas-display.js
```diff
@@ -60,9 +60,8 @@ export default function GasDisplay({ gasError }) {
   const useCurrencyRateCheck = useSelector(getUseCurrencyRateCheck);
   const { showFiatInTestnets, useNativeCurrencyAsPrimaryCurrency } =
     useSelector(getPreferences);
-  const { nativeCurrency, provider, unapprovedTxs } = useSelector(
-    (state) => state.metamask,
-  );
+  const { provider, unapprovedTxs } = useSelector((state) => state.metamask);
+  const nativeCurrency = provider.ticker;
```
jpuri, line 64: These selectors can be used here: getUnapprovedTransactions, getProvider.

jpuri (dismissed):
Nice work on test coverage, I left a small feedback. Looks good otherwise.

NiranjanaBinoy (dismissed):
LGTM

ui/ducks/metamask/metamask.js
```diff
@@ -307,7 +307,7 @@ export function getConversionRate(state) {
 }
 
 export function getNativeCurrency(state) {
-  return state.metamask.nativeCurrency;
+  return state.metamask.provider.ticker;
```
danjm, line 310: We should not do this when the currencyratecontroller is being used (i.e. the toggle is on). Because it could make the conversion rate and the native currency be out of sync.

danjm:
To do: ensure that conversion rate dependent values are never shown when the currency rate is loading

- [ ] Create a ticket for this

ui/ducks/metamask/metamask.js
```diff
@@ -307,7 +308,10 @@ export function getConversionRate(state) {
 }
 
 export function getNativeCurrency(state) {
-  return state.metamask.nativeCurrency;
```
danjm, line 310: @jpuri @NiranjanaBinoy I updated this since your last review. The problem was pointed out to me by Mark. `state.metamask.nativeCurrency` is always set after the conversionRate updates. So if we stopped using it, the `nativeCurrency` and `conversionRate` could get out of sync, showing the user incorrect data.

I have changed the code so that we still use the `nativeCurrency`, but only when the useCurrencyRateCheck setting is on. When it is off, which is required for the bug we are trying to fix, we don't show converted values, so the concern about `nativeCurrency` and `conversionRate` being out of sync does not matter

NiranjanaBinoy (dismissed):

jpuri (dismissed):

jpuri (dismissed):

ui/ducks/metamask/metamask.js
Gudahtt: To add more detail: currently we continue showing the _old_ network currency and currency rate after switching, until the new rate has been loaded. This is pretty confusing, but we're using the CurrencyRateController `nativeCurrency` state in this situation so we can at least be confident the symbol we're showing next to the rate matches the rate.

Ideally we would show _no_ fiat value while currency rates are loading. That would be less confusing. Once we do that, I think it would be safe to always use `state.metamask.provider.ticker` for the current currency.

But until we make that improvement, there is this tradeoff to consider.

NiranjanaBinoy (approved):

jpuri (approved):

PR #17454, opened 2023-01-27
feature: convert ui/ducks/app/app.ts to typescript

DDDDDanica:
## Explanation

Extract from https://github.com/MetaMask/metamask-extension/pull/17372/files# and part of https://github.com/MetaMask/metamask-extension/issues/17214

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/ui/ducks/app/app.test.js
+++ b/ui/ducks/app/app.test.js
@@ -42,7 +42,7 @@ describe('App State', () => {
   it('opens alert', () => {
     const state = reduceApp(metamaskState, {
       type: actions.ALERT_OPEN,
-      payload: 'test message',
+      value: 'test message',
     });
 
     expect(state.alertOpen).toStrictEqual(true);
@@ -63,7 +63,7 @@ describe('App State', () => {
   it('detects qr code data', () => {
     const state = reduceApp(metamaskState, {
       type: actions.QR_CODE_DETECTED,
-      payload: 'qr data',
+      value: 'qr data',
     });
 
     expect(state.qrCodeData).toStrictEqual('qr data');
@@ -121,27 +121,13 @@ describe('App State', () => {
       type: actions.GO_HOME,
     });
 
-    expect(state.accountDetail.subview).toStrictEqual('transactions');
-    expect(state.accountDetail.accountExport).toStrictEqual('none');
     expect(state.accountDetail.privateKey).toStrictEqual('');
     expect(state.warning).toBeNull();
-  });
-
-  it('shows account detail', () => {
-    const state = reduceApp(metamaskState, {
-      type: actions.SHOW_ACCOUNT_DETAIL,
-      payload: 'context address',
-    });
-    expect(state.accountDetail.subview).toStrictEqual('transactions'); // default
-    expect(state.accountDetail.accountExport).toStrictEqual('none'); // default
-    expect(state.accountDetail.privateKey).toStrictEqual(''); // default
   });
 
   it('clears account details', () => {
     const exportPrivKeyModal = {
       accountDetail: {
-        subview: 'export',
-        accountExport: 'completed',
         privateKey: 'a-priv-key',
       },
     };
@@ -151,10 +137,10 @@ describe('App State', () => {
       type: actions.CLEAR_ACCOUNT_DETAILS,
     });
 
-    expect(newState.accountDetail).toStrictEqual({});
+    expect(newState.accountDetail).toStrictEqual({ privateKey: '' });
   });
 
-  it('shoes account page', () => {
+  it('shows account page', () => {
     const state = reduceApp(metamaskState, {
       type: actions.SHOW_ACCOUNTS_PAGE,
     });
@@ -202,7 +188,7 @@ describe('App State', () => {
 
     const state = reduceApp(oldState, {
       type: actions.COMPLETED_TX,
-      payload: {
+      value: {
         id: 1,
       },
     });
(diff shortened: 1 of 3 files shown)


ui/ducks/app/app.ts
```diff
@@ -1,12 +1,85 @@
...
+  forgottenPassword: boolean | null;
+  txId: number | null;
+}
+
+interface AppSliceState {
+  appState: AppState;
+}
+
+export default function reduceApp(
+  state: AppSliceState,
```
brad-decker, line 78: I believe state is AppState, not AppSliceState. see in here we spread ...state into appState, and we return it, we only have appState as the key due to ducks/index.js which puts it on the appState key.

brad-decker (changes requested):

ui/ducks/app/app.ts
DDDDDanica: yes you are correct, modified this one and also changed default state as 
```
  const appState: AppState = {
    ...state,
    shouldClose: false,
```

DDDDDanica: thanks for the suggestion, spread into 2 now 🙏

ui/ducks/app/app.ts
```diff
@@ -82,7 +161,7 @@ export default function reduceApp(state = {}, action) {
       return {
         ...appState,
         alertOpen: true,
-        alertMessage: action.value,
+        alertMessage: action.payload,
```
brad-decker, line 164: ALERT_OPEN is fired from the showAlert method in actions, which hasn't yet been updated to pass payload, this change should be reverted. We only want to change to use payload on the actions you have specifically typed as PayloadAction in this file. Others will be converted by me in the actions typescript conversion :)

ui/ducks/app/app.ts
```diff
@@ -103,7 +182,7 @@ export default function reduceApp(state = {}, action) {
     case actionConstants.SET_SMART_TRANSACTIONS_ERROR:
       return {
         ...appState,
-        smartTransactionsError: action.payload,
+        smartTransactionsError: action.value,
```
brad-decker, line 185: Whereas this one is called from the actions.js file with payload, but it was changed to use value?

ui/ducks/app/app.ts
```diff
@@ -114,7 +193,7 @@ export default function reduceApp(state = {}, action) {
 
     // modal methods:
     case actionConstants.MODAL_OPEN: {
-      const { name, ...modalProps } = action.payload;
+      const { name, ...modalProps } = action.value;
```
brad-decker, line 196: Same here, this was changed to use value when it should be payload. See 'showModal' in actions.js

ui/ducks/app/app.ts
```diff
@@ -363,7 +440,7 @@ export default function reduceApp(state = {}, action) {
         ...appState,
         singleExceptions: {
           ...appState.singleExceptions,
-          [action.value]: null,
+          [action.payload]: null,
```
brad-decker, line 443: This one passes value, not payload.

ui/ducks/app/app.ts
```diff
@@ -410,63 +487,81 @@ export default function reduceApp(state = {}, action) {
...
     type: actionConstants.PORTFOLIO_TOOLTIP_WAS_SHOWN_IN_THIS_SESSION,
   };
 }
 
-export function toggleGasLoadingAnimation(value) {
-  return { type: actionConstants.TOGGLE_GAS_LOADING_ANIMATION, value };
+export function toggleGasLoadingAnimation(
+  payload: boolean,
+): PayloadAction<boolean> {
+  return { type: actionConstants.TOGGLE_GAS_LOADING_ANIMATION, payload };
```
brad-decker, line 505: This one has been updated to be a PayloadAction (payload field) but the reducer still looks for value for this action type.

brad-decker (changes requested):
Looks like you have a few mismatches in terms of actions that pass value versus those that pass payload. You modified a handful of actions in this file to be PayloadAction and changed 'value' to 'payload' as I suggested, but not all of those were modified in the reducer to look for 'payload' versus 'value' and some that were NOT modified to use payload instead of value were changed.

ui/ducks/app/app.test.js
```diff
@@ -42,7 +42,7 @@ describe('App State', () => {
   it('opens alert', () => {
     const state = reduceApp(metamaskState, {
       type: actions.ALERT_OPEN,
-      value: 'test message',
+      payload: 'test message',
```
pedronfigueiredo, line 45: are these changes from value to payload intentional?

ui/ducks/app/app.ts
DDDDDanica: Thanks for pointing it out, now i understand the flow here now 🙏

ui/ducks/app/app.ts
DDDDDanica: fixed now 🙏

ui/ducks/app/app.ts
DDDDDanica: Modified reducer ~~

ui/store/actions.js
```diff
@@ -2239,10 +2239,9 @@ export function showModal(payload) {
   };
 }
 
-export function hideModal(payload) {
+export function hideModal() {
   return {
     type: actionConstants.MODAL_CLOSE,
```
DDDDDanica, line 2244: Note: there's no payload used in reducer, and there's no payload passing in the dispatcher. Hence the payload param is removed ~

pedronfigueiredo (approved):

brad-decker (approved):

PR #17465, opened 2023-01-27
Document `displayWarning` as deprecated

Gudahtt:
The `displayWarning` action creator will show the error in various different places in the UI. This is confusing for users, we should stop doing this.

Instead we can consider each error case separately. If there is a scenario where it's appropriate to tell a user about an error, we should build that case into the relevant UI and use localized error messages.

## Manual Testing Steps

No functional changes, just inline docs.

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/ui/store/actions.js
+++ b/ui/store/actions.js
@@ -2358,6 +2358,16 @@ export function hideLoadingIndication() {
   };
 }
 
+/**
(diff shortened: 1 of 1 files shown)


brad-decker (approved):
LGTM :)

danjm (approved):

PR #17498, opened 2023-01-30
Expand network controller unit test coverage

### LINKED ISSUE #16962: Write comprehensive unit tests for the network controller

mcmire:
## Explanation

The network controller has some tests, but they are incomplete and don't
follow our latest best practices for writing unit tests.

This commit greatly increases the amount of test coverage for the API
that we want to retain in NetworkController, in particular the seemingly
myriad paths that the code takes starting from `initializeProvider`,
`resetConnection`, `setRpcTarget`, `setProviderType`,
`rollbackToPreviousProvider`, and `lookupNetwork`.

There were a couple of pieces of logic I noted which don't seem to have
any effect due to being redundant or unreachable, but they also don't
make our lives more difficult, either, so I opted to leave them in.

Fixes #16962.

## Manual Testing Steps

These changes should not have any user-facing impact, as they touch test files only. That said, all of the tests should pass.

## Pre-merge author checklist

- [x] I've clearly explained:


mcmire:
Key to understanding these tests is this diagram of how each method relates to one another:

extension-network-controller

As with the other test files I've written, the best way to review this PR would probably be to check out this branch locally, run `network-controller.test.js` through Jest on its own, and review the hierarchy.

As you can see above, all five of the ways to set a provider (`initializeProvider`, `resetConnection`, `setRpcTarget`, `setProviderType`, and `rollbackToPreviousProvider`) intersect at both `_configureProvider` and `lookupNetwork`. So:

* The tests for these five methods exercise the happy path for `lookupNetwork`, but because `lookupNetwork` is public, it has tests of its own, which exercise the error cases. This is why you might see a set of repeated tests in the output.
* For each of these five ways we have to make sure that the code works not only when setting an Infura provider but also a non-Infura provider.

app/scripts/controllers/network/network-controller.js
```diff
@@ -125,7 +125,9 @@ export default class NetworkController extends EventEmitter {
     }
     this._infuraProjectId = infuraProjectId;
 
-    this.on(NETWORK_EVENTS.NETWORK_DID_CHANGE, this.lookupNetwork);
```
Gudahtt, line 128: Huh. I'm guessing this would fail each time because it was called unbound, and `this` wasn't set to anything.

I looked back in the git history and this was _always_ called this way, so.... was this always broken? :thinking:

Gudahtt: Perhaps better off removing it if it never worked. Especially for this PR.

mcmire: Ah. I had fixed this method call early on, then when I went to revert this change, it broke a bunch of tests. 

If this is broken, it means that when the network is switched via `setRpcTarget` or `setNetworkType`, rolled back via `rollbackToPreviousProvider`, or reset via `resetConnection`, a new provider gets created, but the network details, which holds whether the network supports EIP-1559, don't get reset until the next time that `getEIP1559Compatibility` is called. So this doesn't seem like the best design to me. 

I suppose considering that's how it works right now, though, I will remove this. On the `core` side, though, it looks like `lookupNetwork` is called directly in `setProviderType` and `setRpcTarget` as opposed to being indirectly called through an event, so I'm wondering if that's how this controller should work.

mcmire: @Gudahtt So it turns out that this always _did_ work, it's just not obvious. It looks like the EventEmitter module will implicitly call the function it's given with the same instance that the event listener is created on: https://nodejs.org/docs/latest-v16.x/api/events.html#passing-arguments-and-this-to-listeners. So `this` is set to the NetworkController in this case.

Changing this code to use an explicit arrow function happened to make that was previously failing without this change pass, but it turns out that that test was incorrect. I've fixed this test and reverted the change to this line: [02daaa7](https://github.com/MetaMask/metamask-extension/pull/17498/commits/02daaa715c14f9cdb9fa7bb45424e0851e77235a)

Gudahtt: Nice! Interesting find, I had no idea it worked that way. That would makes sense though, I would have been surprised if this bug had been here for so long.

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -1,273 +1,4145 @@
...
+   *
+   * @param {object} options - An options bag.
+   * @param {"infura" | "custom"} options.networkClientType - Specifies the
+   * expected middleware stack that will represent the provider: "infura" for an
+   * Infura network; "custom" for a custom RPC endpoint.
+   * @param {string} options.infuraNetwork - The name of the Infura network
+   * being tested, assuming that `networkClientType` is "infura".
+   * @param {string} options.infuraProjectId - The project ID of the Infura
+   * network being tested, assuming that `networkClientType` is "infura".
+   * @param {string} options.customRpcUrl - The URL of the custom RPC endpoint,
```
Gudahtt, line 123: Nit: It seems that these three are optional:
```suggestion
   * @param {string} [options.infuraNetwork] - The name of the Infura network
   * being tested, assuming that `networkClientType` is "infura".
   * @param {string} [options.infuraProjectId] - The project ID of the Infura
   * network being tested, assuming that `networkClientType` is "infura".
   * @param {string} [options.customRpcUrl] - The URL of the custom RPC endpoint,
```

mcmire: Done here: [0b43e1b](https://github.com/MetaMask/metamask-extension/pull/17498/commits/0b43e1b6f7929e2cdaf35d20c13c73507eddf3b9)

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -50,224 +69,4106 @@ const PRE_1559_BLOCK = {
...
-          await controller.initializeProvider();
+        it('emits infuraIsUnblocked (assuming that the request to eth_blockNumber responds successfully)', async () => {
+          await withController(
+            {
+              state: {
+                provider: {
+                  type: 'rpc',
+                  rpcUrl: 'https://mock-rpc-url',
+                  chainId: '0x1337',
+                },
```
Gudahtt, line 575: I'm guessing this was meant to be set to the network type for this describe block:
```suggestion
                provider: {
                  type: networkType,
                },
```
Easy to miss because the test passes either way

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -50,224 +69,4106 @@ const PRE_1559_BLOCK = {
...
+                  type: networkType,
+                },
+              },
+            },
+            async ({ controller, network }) => {
+              network.mockEssentialRpcCalls();
+
+              await controller.initializeProvider();
+
+              expect(controller.store.getState().network).toStrictEqual(
```
Gudahtt, line 608: Nit: for primitive types, our lint rules require that we use `toBe` instead of `toStrictEqual`.

This rule was temporarily disabled in this repo to avoid churn, but we still intended to enable it at some point.

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -50,224 +69,4106 @@ const PRE_1559_BLOCK = {
...
-    describe('#getEIP1559Compatibility', () => {
-      it('should return false when baseFeePerGas is not in the block header', async () => {
-        await withController(async ({ controller }) => {
-          setupMockRpcBlockResponses({ block: PRE_1559_BLOCK });
-          await controller.initializeProvider();
-          const supportsEIP1559 = await controller.getEIP1559Compatibility();
-          expect(supportsEIP1559).toStrictEqual(false);
-        });
+  describe('getProviderAndBlockTracker', () => {
+    it('returns objects that proxy to the provider and block tracker as long as the provider has been initialized', async () => {
```
Gudahtt, line 773: Nit: It would be great to have a test that shows that these always point at the current network, even after the network switches.

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -50,224 +69,4106 @@ const PRE_1559_BLOCK = {
...
+              operation: async () => {
+                await controller.lookupNetwork();
+              },
+            });
+
+            expect(controller.store.getState()).toStrictEqual({
+              ...initialState,
+              network: 'loading',
+              previousProviderStore: providerConfig,
+            });
```
Gudahtt, line 929: This seems like a pretty confusing way of demonstrating that state doesn't change. At first glance I thought this was showing that a state change _did_ happen, because of the "wait for state changes" helper, and the fact that the expected state was clearly different from the initial state.

But the differences just reflect defaults, and we're in fact waiting for zero state changes.

Would this be simpler?

```suggestion
            const stateAfterConstruction = controller.store.getState();

            await controller.lookupNetwork();

            expect(controller.store.getState()).toStrictEqual(stateAfterConstruction);
```

Or is this meant to attempt detecting additional state changes that might happen after `lookupNetwork` resolves?

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -50,224 +69,4106 @@ const PRE_1559_BLOCK = {
...
+                          },
+                        });
+                        return {
+                          result: '0x42',
+                        };
+                      },
+                    },
+                  },
+                });
+                const network2 = new NetworkCommunications({
```
Gudahtt, line 1053: It's unclear what this was intended to do. `network2` isn't referenced anywhere, and the test passes with this commented out

Gudahtt: Oh I see, this was setup with the expectation that we'd be switching to a custom RPC endpoint. But the code above switches to the next infura network. Maybe a copy+paste mistake

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -50,224 +69,4106 @@ const PRE_1559_BLOCK = {
...
+            await withController(
+              {
+                state: {
+                  provider: {
+                    type: networkType,
+                  },
+                },
+              },
+              async ({ controller }) => {
+                const network1 = new NetworkCommunications({
```
Gudahtt, line 1021: Could the `network` argument from `withController` be used here instead?

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -50,224 +69,4106 @@ const PRE_1559_BLOCK = {
...
+                  await controller.lookupNetwork();
+                },
+              });
+
+              await expect(promiseForStateChanges).toNeverResolve();
+            },
+          );
+        });
+
+        describe('if the request for eth_getBlockByNumber responds successfully', () => {
```
Gudahtt, line 1303: Hmm... I'm a little unclear on what case this is meant to represent. Doesn't `eth_getBlockByNumber` respond successfully in the "if the request for eth_blockNumber responds successfully" describe block tests as well?

Gudahtt: Oh I see, nevermind, this is testing different behavior that relates more to this call.

Gudahtt, line 1303: Nit: It would be great to cover the case where `null` is returned by a successful response as well. IIRC that is expected if the network doesn't recognize the requested block number.

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -50,224 +69,4106 @@ const PRE_1559_BLOCK = {
...
+ * block they're looking for.
+ */
+const BLOCK = POST_1559_BLOCK;
+
+/**
+ * A dummy value for the `projectId` option that `createInfuraClient` needs.
+ * (Infura should not be hit during tests, but just in case, this should not
+ * refer to a real project ID.)
+ */
+const DEFAULT_INFURA_PROJECT_ID = 'infura-project-id';
```
BelfordZ, line 93: NIT: use a very obviously incorrect value 
```suggestion
const DEFAULT_INFURA_PROJECT_ID = 'fake-infura-project-id';
```

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -50,224 +69,4106 @@ const PRE_1559_BLOCK = {
...
-    ...defaultControllerOptions,
-    ...constructorArgs,
-  });
-  try {
-    return await fn({ controller });
-  } finally {
-    await controller.destroy();
-  }
-}
+const INFURA_NETWORKS = [
```
BelfordZ, line 108: Should we be using the values from the constants? (either consuming the values already exported, or assembling this array inside the constants file) https://github.com/MetaMask/metamask-extension/blob/develop/shared/constants/network.ts

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -50,224 +69,4106 @@ const PRE_1559_BLOCK = {
...
+              await controller.initializeProvider();
+              expect(controller.store.getState().network).toStrictEqual('42');
+
+              await waitForStateChanges({
+                controller,
+                propertyPath: ['network'],
+                operation: async () => {
+                  await controller.lookupNetwork();
+                },
+              });
```
Gudahtt, line 1686: Nit: Adding a newline here between the "act" and "assert" steps would be nice:

```suggestion
              });

```

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -50,224 +69,4106 @@ const PRE_1559_BLOCK = {
...
+
+              await waitForStateChanges({
+                controller,
+                propertyPath: ['networkDetails'],
+                operation: async () => {
+                  await controller.lookupNetwork();
+                },
+              });
```
Gudahtt, line 1734: Nit: One more suggested newline:

```suggestion
              });

```

BelfordZ:
Great work! I haven't made it through the whole PR yet - but I wanted to ask -- What do you think about changing the network client interface to allow dependency injecting the block tracker, and providing a mock implementation for tests. In theory, this would allow us to build the tests with fine grain control of the block tracker but without the need for fake timers.

Might not be worth the time since we already have these working now, but it might be worth playing with the idea, as I believe it would greatly improve the testability of the network controller.

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -50,224 +69,4106 @@ const PRE_1559_BLOCK = {
...
+  /**
+   * Constructs a new NetworkCommunications object using a different set of
+   * options, using the options from this instance as a base.
+   *
+   * @param options - The same options that NetworkCommunications takes.
+   */
+  with(options) {
+    return new NetworkCommunications({
+      networkClientType: this.networkClientType,
+      infuraNetwork: this.#infuraNetwork,
```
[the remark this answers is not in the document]
mcmire: Good point. Updated in [c04b903](https://github.com/MetaMask/metamask-extension/pull/17498/commits/c04b9037be7f10e07b7b86b723a18bba90be4f39).

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -50,224 +69,4106 @@ const PRE_1559_BLOCK = {
...
 
-    describe('#provider', () => {
-      it('provider should be updatable without reassignment', async () => {
-        await withController(async ({ controller }) => {
-          setupMockRpcBlockResponses();
-          await controller.initializeProvider();
-          const providerProxy =
-            controller.getProviderAndBlockTracker().provider;
-          expect(providerProxy.test).toBeUndefined();
+    // Mock the request that the block tracker makes last so that the request
```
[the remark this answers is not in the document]
mcmire: That's fair. Wording updated here: [e2f99d1](https://github.com/MetaMask/metamask-extension/pull/17498/commits/e2f99d1b0ca2052a4a4a46dfdfb000951a23b1ed)

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -50,224 +69,4106 @@ const PRE_1559_BLOCK = {
...
+
+        const partialResponseBody = await unresolvedPartialResponseBody;
+
+        const completeResponseBody = { jsonrpc: '2.0' };
+        if (requestBody.id !== undefined) {
+          completeResponseBody.id = requestBody.id;
+        }
+        ['id', 'jsonrpc', 'result', 'error'].forEach((prop) => {
+          if (partialResponseBody[prop] !== undefined) {
+            completeResponseBody[prop] = partialResponseBody[prop];
```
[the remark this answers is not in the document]
mcmire: Sure. I took the opportunity to simplify `mockRpcCall` slightly as well: [dc746ad](https://github.com/MetaMask/metamask-extension/pull/17498/commits/dc746ad44c90d27170b6e457b1c05e662bf4a3b1)

app/scripts/controllers/network/network-controller.test.js
mcmire: Yeah I keep forgetting about this. Fixed here: [0b37b24](https://github.com/MetaMask/metamask-extension/pull/17498/commits/0b37b2452fd75097a0a25855b39e1e7de352a60d)

app/scripts/controllers/network/network-controller.test.js
mcmire: Ah okay. That must have been what you meant in your TODOs here. So just to clarify, you want a test that initializes the provider, grabs the chain ID of the current network using the provider proxy, switches the network, grabs the chain ID of the new network using the provider proxy, and asserts that they are different?

app/scripts/controllers/network/network-controller.test.js
mcmire: Yes... you're right that this is confusing. I did want to make sure there were no state changes even after waiting for a bit. `lookupNetwork` doesn't completely await all of its async operations, specifically `_checkInfuraAvailability`. After looking at this again, though, that method doesn't change state, so waiting for no changes to occur shouldn't be necessary.

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -50,224 +69,4200 @@ const PRE_1559_BLOCK = {
...
+
+        const networkWillChange = await waitForEvent({
+          controller,
+          eventName: 'networkWillChange',
+          operation: () => {
+            controller.setRpcTarget('https://mock-rpc-url', '0x1337');
+          },
+        });
+
+        expect(networkWillChange).toBe(true);
```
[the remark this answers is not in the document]
Gudahtt: I guess these two comments apply to `setProviderType` as well

app/scripts/controllers/network/network-controller.test.js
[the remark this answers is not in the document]
Gudahtt: Looks good, thanks!

app/scripts/controllers/network/network-controller.test.js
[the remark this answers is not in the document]
Gudahtt: Nice! That does look better

app/scripts/controllers/network/network-controller.test.js
Gudahtt: Yeah exactly!

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -50,224 +69,4176 @@ const PRE_1559_BLOCK = {
...
+
+  describe('rollbackToPreviousProvider', () => {
+    for (const {
+      networkName,
+      networkType,
+      chainId,
+      networkVersion,
+    } of INFURA_NETWORKS) {
+      describe(`if the previous provider configuration had a type of "${networkType}"`, () => {
+        it('merges the previous configuration into the current provider configuration', async () => {
```
Gudahtt, line 3067: Welp! We will have to fix this bug after this PR. Good find

Gudahtt (commented):
This looks great! Just finished a complete pass of this PR. Lots of comments but they're all minor suggestions. The only one I kinda wait to wait on is [this seemingly simple copy+paste mistake](https://github.com/MetaMask/metamask-extension/pull/17498#discussion_r1094614868), then I can approve.

app/scripts/controllers/network/network-controller.test.js
mcmire: So this test is a little complicated. In the implementation, in `_checkInfuraAvailability`, we make a request to `eth_blockNumber`. If the network has switched, however, after the request finishes, then we don't emit any events. So `network2` is the second network that we switch to.

app/scripts/controllers/network/network-controller.test.js
mcmire: Yeah, I wanted to make the initial network more explicit to differentiate it from `network2`, but it's inconsistent with how I do things in other tests, so I might as well be consistent. Changed in [8319bb5](https://github.com/MetaMask/metamask-extension/pull/17498/commits/8319bb51ebf5b14302f5b6cfd93d7b4b74d73426).

app/scripts/controllers/network/network-controller.test.js
Gudahtt: > So network2 is the second network that we switch to.

That's what I mean - it's _not_ the network being switched to here. It seems like you meant it to be though.

app/scripts/controllers/network/network-controller.test.js
mcmire: Ah. I missed that while reading the code. Good point. Added here: [520ad6d](https://github.com/MetaMask/metamask-extension/pull/17498/commits/520ad6dda775b3b9cd10e028e0fab8a12fbb5bb8)

app/scripts/controllers/network/network-controller.test.js
mcmire: We could. Maybe this is just my opinion, but I don't like to reuse data that's been defined in implementation. If that data is incorrect, then the test will be incorrect too, but we wouldn't know about it because the tests may all pass. By defining this data in the test file, we can specify how we expect the world to look like, not what it actually is. Obviously we could get that wrong too, so it's not a bulletproof solution, but as it's isolated in this file we don't have to look so far to fix it. That was the thinking behind not using the pre-existing network constants.

app/scripts/controllers/network/network-controller.test.js
mcmire: Derp, you're right. I don't know what I was seeing 😅 Fixed in [884e702](https://github.com/MetaMask/metamask-extension/pull/17498/commits/884e7022a6e50030fefd1df52db7b99991b15952).

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -50,224 +69,4198 @@ const PRE_1559_BLOCK = {
...
+        await controller.getEIP1559Compatibility();
+        await controller.getEIP1559Compatibility();
+
+        expect(network.nockScope.isDone()).toBe(true);
       });
     });
   });
-});
+
+  describe('lookupNetwork', () => {
```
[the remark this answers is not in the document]
mcmire: Are you talking about this behavior? https://github.com/MetaMask/metamask-extension/blob/884e7022a6e50030fefd1df52db7b99991b15952/app/scripts/controllers/network/network-controller.js#L428-L431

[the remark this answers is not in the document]
Gudahtt: Sorry no, I explained that poorly. I mean the behavior covered by the test "emits infuraIsUnblocked as long as the network has not changed by the time the request ends"

If the network changed by the time the request ends, then no state changes happen (e.g. network state, network details, EIP1559 compatibility). Due to this line: https://github.com/MetaMask/metamask-extension/blob/02b2d7fa42e5edd7d66c28a40c5a2f30cc27c1ac/app/scripts/controllers/network/network-controller.js#L217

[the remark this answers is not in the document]
Gudahtt: Oh I see, that test isn't testing that condition, but it's a similar one (within `_checkInfuraAvailability`). That one is covered, but this one in `lookupNetwork` is not.

app/scripts/controllers/network/network-controller.test.js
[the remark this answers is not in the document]
mcmire: Okay, I tried changing both of these tests, is this what you had in mind? Note that there are a couple of ways we could test for network changes, so I could wait for `networkDidChange` or actually wait for `network` or `networkDetails` to get updated. Commits: [ab8e4f5](https://github.com/MetaMask/metamask-extension/pull/17498/commits/ab8e4f53574aa118a7efc150e51597035a216dcc), [a89f268](https://github.com/MetaMask/metamask-extension/pull/17498/commits/a89f2685801b336f30f7aac08fdfb25cfdfee218).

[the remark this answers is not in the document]
Gudahtt: Hmm, sorry maybe I explained that poorly. I was imagining that we could add an assertion next to this one, something like `expect(controller.store.getState().network).toBe(initialNetwork)`. That would confirm that this was emitted before the network was changed.

It looks like the changes you made are verifying the order in which these events are emitted instead.

app/scripts/controllers/network/network-controller.test.js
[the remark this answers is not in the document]
mcmire: Gotcha. Yes that's true.

That case is a bit tricky to test because we have to end up stubbing `ethQuery.sendAsync` so that it switches the network before resolving (because the network is never hit in the Infura case). I'll work on this more tomorrow.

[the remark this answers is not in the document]
Gudahtt: Ah right, hadn't considered that this wouldn't apply for the Infura case. Maybe we could just test this for custom networks in that case. Seems fine to ignore this for Infura if the situation isn't possible.

app/scripts/controllers/network/network-controller.test.js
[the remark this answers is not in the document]
mcmire: Okay, how about this: [f371ff5](https://github.com/MetaMask/metamask-extension/pull/17498/commits/f371ff5e573cda909f3347f06693ff8de79c2271)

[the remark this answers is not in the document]
Gudahtt: Makes sense, thanks!

app/scripts/controllers/network/network-controller.test.js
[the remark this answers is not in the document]
mcmire: Okay, it took a bit for the Infura case, but I think I got this figured out: [b304549](https://github.com/MetaMask/metamask-extension/pull/17498/commits/b30454990d7f20f84847672623edd48836959baa)

app/scripts/controllers/network/network-controller.test.js
mcmire: Okay, tests added here: [9f4e7d0](https://github.com/MetaMask/metamask-extension/pull/17498/commits/9f4e7d0633ba44241c4a3c88ee7403001e302dde)

app/scripts/controllers/network/network-controller.js
```diff
@@ -125,7 +125,9 @@ export default class NetworkController extends EventEmitter {
...
+    this.on(NETWORK_EVENTS.NETWORK_DID_CHANGE, () => {
```
mcmire, line 128: Revisiting this again, it turns out that in order to effectively stub out this method using `jest.spyOn` in the tests I have to wrap the method call like this.

app/scripts/controllers/network/network-controller.test.js
Gudahtt: Hmm, it looks like these tests aren't covering the proxy behavior at the moment. I think you'd want to check the chain ID before and after switching chains, of the same provider instance. That would confirm that that single instance is being updated via proxy to the new chain.

mcmire: Ah — you're right, thanks for pointing that out. I've been staring at this too long, I guess. Fixed here: [8b05ade](https://github.com/MetaMask/metamask-extension/pull/17498/commits/8b05adef026ffe0bead287c81f0c06ac1717089f)

app/scripts/controllers/network/network-controller.js
Gudahtt: Context for any future readers: we discovered that this change is a no-op because of how the event emitter works: https://github.com/MetaMask/metamask-extension/pull/17498/files#r1092527202

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -50,224 +77,4799 @@ const PRE_1559_BLOCK = {
...
+                            .then(() => {
+                              callback(null, {
+                                id: request.id,
+                                jsonrpc: '2.0',
+                                result: oldNetworkVersion,
+                              });
+                            })
+                            .catch((error) => {
+                              throw error;
+                            });
```
Gudahtt, line 1581: Nit: We can get rid of this `.then` while preserving this as a non-blocking operation by using an inline function:

```suggestion
                          const returnOldNetworkVersionAfterStateChange = async () => {
                            await waitForStateChanges({
                              controller,
                              propertyPath: ['network'],
                              operation: () => {
                                controller.setProviderType(newNetworkName);
                              },
                            });
                            callback(null, {
                              id: request.id,
                              jsonrpc: '2.0',
                              result: oldNetworkVersion,
                            });
                          };
                          
                          // intentionally not `await`-ing this because `sendAsync` does not return a Promise
                          returnOldNetworkVersionAfterStateChange();
```

Gudahtt: Oh, I see the same applies for the next test too, I think that's the last one though

Gudahtt (dismissed):
LGTM!

app/scripts/controllers/network/network-controller.test.js
mcmire: I think I might leave this one alone if you don't mind. Since `sendAsync` doesn't return a promise, using the `.then` pattern here seems less surprising than using the `async` keyword. And then we also don't have to explain why we're not `await`ing such a function.

mcmire:
@Gudahtt Made a few more changes after a dialog with @BelfordZ — care to take another look?

BelfordZ (dismissed):
excellent!

app/scripts/controllers/network/network-controller.test.js
Gudahtt: Ok, sounds good. Personally I don't find `.then` less surprising, but I see what you mean.

I made this suggestion because I'm anticipating we'll be bringing in the lint rule to prefer async/await, and because it has better stack traces. But neither is important here. Maybe this test will be obsolete by that point anyway.

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -50,224 +76,4826 @@ const PRE_1559_BLOCK = {
...
+              provider: {
+                type: 'rpc',
+                rpcUrl: 'https://mock-rpc-url',
+                chainId: '0x1337',
+              },
+            },
+          },
+          async ({ controller, network }) => {
+            network.mockEssentialRpcCalls();
+            controller.initializeProvider();
```
Gudahtt, line 3507: I noiced that the provider is being initialized above without an `await`. There are five tests total that call `initializeProvider` without `await`. Was this intentional?

Gudahtt:
Had one more question, but otherwise the changes look great! I especially like https://github.com/MetaMask/metamask-extension/pull/17498/commits/829076edfb66a0439d3510bf1bc15b18ddc0368a

app/scripts/controllers/network/network-controller.test.js
mcmire: Hmm, I thought I fixed these :( That is not intentional. That might explain some of the waiting that I have to do in later tests. Fixed in [1ed0356](https://github.com/MetaMask/metamask-extension/pull/17498/commits/1ed03565ede7825bd5797fee5ba9984caaa5f3d9).

mcmire:
@Gudahtt @BelfordZ One more approval if you please!

Gudahtt (approved):

BelfordZ (approved):

PR #17508, opened 2023-01-31
Convert store.js to typescript and redux toolkit

brad-decker:
## Explanation
Switches the store from using the redux 'createStore' method to the configureStore method from reduxjs/toolkit. The toolkit configureStore methods has a lot of good features and type inference and we want to generally adopt more of the reduxjs/toolkit features. 

There are two changes that needed to be made in order to get this passing in E2E:

By default configureStore loads two middleware that we did not previously use:
1. serialzableCheck middleware -- checks if the state keys are serializable thus can be persisted in browser state for rehydration.
2. immutableCheck middleware -- checks that state is immutable and not mutated.

These two middleware should be incrementally adopted, but for now they are disabled, and in the case of immutableCheck only enabled in dev mode. I had to specify a higher warnAfter level because our actions take longer than expected to check (32ms -> 80ms) we can lower that number overtime. 

While not *technically* part of this work, the inclusion of a few baseline entries in the state tree type will  aid the action typescript conversion pr. 

** In addition ** 
I made a minor change to the metamask duck to allow typescript to infer the shape of its state, though this should also have no functional impact. 

## Manual Testing Steps
1. Pull branch
2. `yarn`
3. `yarn start`
4. etiher with or without redux devtools, perform a number of actions
5. confirm no change in behavior.

## Pre-merge author checklist

- [x] I've clearly explained:


ui/store/store.ts
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * TODO: Replace this
+ */
+export interface TemporaryMessageDataType {
+  id: number;
+  type: string;
+  msgParams: {
+    metamaskId: number;
+    data: string;
+  };
```
DDDDDanica, line 24: is there other keys like `origin`, `from`, `version` missing in this type?

brad-decker: Yes, I aimed for the minimal typing because we want to make this type more robust when we hit the background data layer, this was like MVP for getting types that work in actions.ts

DDDDDanica: gotcha!

DDDDDanica (approved):

ui/ducks/metamask/metamask.js
```diff
@@ -23,39 +23,55 @@ import {
...
+/**
+ * Temporary types for this slice so that inferrence of MetaMask state tree can
+ * occur
+ *
+ * @param {typeof initialState} state - State
+ * @param {any} action
+ * @returns {typeof initialState}
+ */
+export default function reduceMetamask(state = initialState, action) {
+  // I don't think we should be spreading initialState into this. Once the
```
pedronfigueiredo, line 66: These are great notes. Does it make sense to capture them in a ticket instead of or apart from this comment?

brad-decker: Yeah, i'll create a ticket now.

brad-decker: https://github.com/MetaMask/metamask-extension/issues/17578

pedronfigueiredo (approved):

PR #17512, opened 2023-01-31
Dead and redundant code removal

brad-decker:
## Explanation
Removes a number of unnecessary, and/or dead actions and their usages. I have separated this description into parts to help talk about each.

### App slice/duck
1. We had a number of keys in 'accountDetail' object that were no longer used, such as 'subview' which appears to have been used for routing at one point. Another one is 'export'
2. We had a loadingMethodData value that was never actually being used.
3. We were handing an action SHOW_ACCOUNT_DETAIL that did not need to be in this slice because the thunk that dispatched that action handled all necessary state changes by calling into the background.

### MetaMask slice/duck
This slice is not meant to contain state that is managed in redux by UI but rather is a reflection of the state in the background, a number of actions that were in here are what we might consider "optimistic updates" because they update the metamask state *possibly* before the state changes in the background. The only time this might be necessary is if the update takes a considerably long time. In my testing removing these methods did not have an appreciable impact on UI.

### actions file
Some of the changes to this file are just removing calls to actions that are no longer needed, with one exception and that is the 'showAccountDetail' thunk. This thunks name makes it sound like its showing some kind of detail that wasn't previously visible, but in fact all its doing is setting the active account which makes its values visible in the main UI. I renamed it, though there is now a bit of a discrepancy between 'setSelectedAccount' and 'setSelectedAddress' that we will need to figure out in a future PR. 

## Manual Testing Steps
1. pull branch
2. `yarn`
3. `yarn start`
4. Test some of the actions that were removed that duplicated updates "optimistically" such as changing preferences, feature flags, blockies, changing your network, etc.
5. Ensure no functional changes

## Pre-merge author checklist

- [x] I've clearly explained:


DDDDDanica (dismissed):

DDDDDanica:
💯 thanks for doing the clean ups ！

DDDDDanica (dismissed):

DDDDDanica (approved):

pedronfigueiredo (approved):

PR #17522, opened 2023-01-31
Remove extraneous methods from NetworkController

### LINKED ISSUE #16947: NetworkController API normalization: remove getter methods
The network controller has a variety of methods that just retrieve controller state. These methods are not necessary because controller state is already part of the public API of the controller, it can be accessed directly. We should remove these methods, and replace any usage of them with direct state access.

The methods are:

- [x] `getNetworkState`
- [x] `isNetworkLoading`
- [x] `getCurrentChainId`
- [x] `getCurrentRpcUrl`
- [x] `getProviderConfig`
- [x] `getNetworkIdentifier`

mcmire:
## Explanation

The network controller has a variety of methods that just retrieve
controller state. These methods are not necessary because controller
state is already part of the public API of the controller and can be
accessed directly.

These methods are:

- getCurrentChainId
- getCurrentRpcUrl
- getNetworkIdentifier
- getNetworkState
- getProviderConfig
- isNetworkLoading

This is part of a larger effort to normalize the API of both network
controllers, to make them easier to merge.

Fixes #16947.

## Manual Testing Steps

The changes here touch the following pieces:

* ENS IPFS resolver
* MetaMetrics controller
* network provider
* token list controller
* gas fee controller
* ENS controller
* incoming transactions controller
* account tracker
* cached balances controller
* transaction controller
* currency rate controller
* typed message manager
* swaps controller
* smart transactions controller
* legacy provider config stream
* `metamask_getProviderState`
* `wallet_switchEthereumChain`
* `wallet_addEthereumChain`

So it appears that we should test:

* That you can't visit an IPFS or ENS URL while not on mainnet
* Anything involving metrics (should send the current chain ID along as data)
* General wallet functionality (which should trigger RPC requests through the provider)
* Adding tokens (should see new tokens as they arrive)
* Making a transaction (should see gas fees for current network, should see current token<->fiat conversion rate, should see new transaction in list after creation)
* Should see ENS names resolved for accounts
* Trigger a signature confirmation, then switch network (signing should fail)
* Swaps should work for current network
* Transaction insights should work for current network
* Data in legacy provider config stream should include current chain ID
  * This one might be difficult to test as I'm not sure how often this is used, or how many people know about this. I certainly didn't!
* Return data from `metamask_getProviderState` should include the current chain ID
* `wallet_switchEthereumChain` shouldn't do anything when the chain ID is the same as the current network
* `wallet_addEthereumChain` shouldn't do anything when the chain ID / RPC URL being added is the same as the current network

## Pre-merge author checklist

- [x] I've clearly explained:


app/scripts/controllers/network/network-controller.js
```diff
@@ -189,23 +181,19 @@ export default class NetworkController extends EventEmitter {
       return;
     }
 
-    // NOTE: This will never happen in practice because you can't pass null or
```
mcmire, line 192: Now that `getCurrentChainId` doesn't exist and we are pulling `chainId` directly from the store, `chainId` can be undefined.

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -580,6 +580,9 @@ describe('NetworkController', () => {
               state: {
                 provider: {
                   type: networkType,
+                  // NOTE: This doesn't need to match the logical chain ID
+                  // of the network selected, it just needs to exist
```
mcmire, line 584: ...because `chainId` is no longer inferred from the Infura network.

pedronfigueiredo (approved):

app/scripts/controllers/metametrics.test.js
```diff
@@ -64,23 +64,24 @@ const DEFAULT_PAGE_PROPERTIES = {
...
+  let state = {
+    provider: {
+      type: NETWORK_TYPES.GOERLI,
+      chainId: FAKE_CHAIN_ID,
+    },
+    network: 'loading',
+  };
   const on = sinon.stub().withArgs(NETWORK_EVENTS.NETWORK_DID_CHANGE);
   const updateState = (newState) => {
-    networkStore = { ...networkStore, ...newState };
```
mcmire, line 74: This test helper was setting up state for this fake network controller in a way that didn't match how the real network controller sets up state. These changes should bring them closer in alignment.

BelfordZ (approved):

app/scripts/controllers/network/network-controller.js
Gudahtt: This comment was _about_ the value in the store though. It is initialized with a valid chain ID, and all built-in networks have a valid chain ID, and we have UI validation to ensure you can't pass in `null`/`undefined`. So it should still hold true, barring some bug in our add network flow I guess.

Gudahtt (approved):
LGTM!

app/scripts/controllers/network/network-controller.js
mcmire: Hmm, I suppose so. I believe when I added this comment I thought it was impossible for `chainId` to be `null` or `undefined`. In hindsight, perhaps I was wrong. In any case, it definitely is possible now; you just need to initialize the controller with a state object that doesn't have a `chainId` property, which was mistakenly happening in some tests (hence the changes below to add fake `chainId`s).

chloeYue:
QA tests OK !

Issue #16747, opened 2022-11-30
[FLASK] Create E2E Test for RPC test snap


bowensanders:
Create an E2E test for the RPC test snap (4.2.0+)

bowensanders:
This is currently blocked by the lack of a test-snap that provides a positive outcome.
Having the only outcome be an error, in this case, due most likely to the type of error, the test will not ever be "passing."

We need an RPC test snap that does not error out as a result.

bowensanders:
This is currently blocked by the lack of a test-snap that provides a positive outcome.
Having the only outcome be an error, in this case, due most likely to the type of error, the test will not ever be "passing."

We need an RPC test snap that does not error out as a result.

bowensanders:
Unblocked by https://github.com/MetaMask/test-snaps/pull/133

PR #16840, opened 2022-12-07
Version v10.24.0

seaona:
## Pre-merge author checklist

- [ ] I've clearly explained:
  - [ ] What problem this PR is solving
  - [ ] How this problem was solved
  - [ ] How reviewers can test my changes
- [ ] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [ ] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
- [ ] PR is linked to the appropriate GitHub issue
- [ ] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.

In this case, a QA Engineer approval will be be required.

PR #17525, opened 2023-01-31
[FLASK] Create E2E test for RPC endowment

bowensanders:
This PR creates a new E2E test for the RPC endowment.

This addresses #16763 and #16747


Mrtenz (dismissed):

test/e2e/snaps/test-snap-rpc.spec.js
```diff
@@ -0,0 +1,130 @@
...
+          'MetaMask Notification',
+          windowHandles,
+        );
+        await driver.clickElement({
+          text: 'Approve & install',
+          tag: 'button',
+        });
+
+        // delay for npm installation
+        await driver.delay(2000);
```
PeterYinusa, line 72: Instead of adding an arbitrary delay, can we wait for the installation to complete? The installation may take longer sometimes and cause the test to fail. An example where this happened was when we were in the process of releasing v10.24.0. Most of the snap tests failed on the installation steps. Here is a [link](https://github.com/MetaMask/metamask-extension/pull/16840/commits/a464f128bc2ad95345e27812dba643d80d4335a6) to a change I made to ensure the install was complete without using a delay, I wait for the test dapp buttons to be updated.

PeterYinusa (dismissed):

test/e2e/snaps/test-snap-rpc.spec.js
```diff
@@ -0,0 +1,130 @@
...
+        windowHandles = await driver.waitUntilXWindowHandles(2, 1000, 10000);
+        await driver.switchToWindowWithTitle(
+          'MetaMask Notification',
+          windowHandles,
+        );
+        await driver.clickElement({
+          text: 'Approve & install',
+          tag: 'button',
+        });
+        await driver.delay(2000);
```
PeterYinusa, line 94: same as above

test/e2e/snaps/test-snap-rpc.spec.js
```diff
@@ -0,0 +1,130 @@
...
+          'MetaMask Notification',
+          windowHandles,
+        );
+        await driver.clickElement({
+          text: 'Approve & install',
+          tag: 'button',
+        });
+
+        // wait for permissions popover, click checkboxes and confirm
+        await driver.delay(1000);
```
PeterYinusa, line 108: same as above

test/e2e/snaps/test-snap-rpc.spec.js
bowensanders: I am unsure of how to use the waitForSelector in the case of popups that close and then open another popup opens [this happens in snap installs]
because you have to switch focus back to it before you can search, but if you don’t delay before that search it errors with no such window because of it. 
Should we just leave this delay alone, or is there a similar function to know when a popup window opens?

test/e2e/snaps/test-snap-rpc.spec.js
bowensanders: Fixed

test/e2e/snaps/test-snap-rpc.spec.js
bowensanders: Fixed

bowensanders: Fixed

test/e2e/snaps/test-snap-rpc.spec.js
PeterYinusa: I see. I didn't realise this was a pop-up that closes and automatically opens another one. Maybe here we could use `waitUntilXWindowHandles` to wait for the window to close, then again to wait for the window to reopen. If not then we may need to explore this further later and just go with the delay for now

PeterYinusa: The reason you get the `no such window` error is that the popup is closed, so you would need to switch back to the test dapp if you wanted to wait for something here

bowensanders: The extra `waitUntilXWindowHandles` works just fine. Thanks!

bowensanders: The added wait worked - thanks!

Mrtenz (dismissed):

test/e2e/snaps/test-snap-rpc.spec.js
PeterYinusa: Sounds good to me.

**In the future**, it could be worth experimenting with the `delayStep` of the `waitUntilXWindowHandles` method to see if that helps with the double notification.
An example of this could be reducing the delay from `1000` to `200` e.g. `windowHandles = await driver.waitUntilXWindowHandles(1, 200, 10000);`

hmalik88 (approved):

Mrtenz (approved):

PR #17547, opened 2023-02-01
Fix speed up of 0 max priority fee transactions

### LINKED ISSUE #17353: [Bug]: Cannot speed up 0 priority fee transaction
### Describe the bug

If you submit a transaction with 0 Priority Fee, when you attempt to speed up the transaction with the default (default: increase the gas by 10%), the Priority Fee remains at 0. Therefore the transaction is underpriced and the user gets the following error

https://user-images.githubusercontent.com/53189696/214100015-019288ae-fd6c-43eb-9d9d-fe6309cc6000.mov

### Steps to reproduce

1. Login
2. Click `Send` on the overview page
3. Enter an address
4. Leave the default amount etc, Click `Next`
5. Click the `Market`, to edit the gas fees
6. Click `Advanced`, to enter custom gas fees
7. Change the `Priority Fee` to `0`, Click `Save`
8. Click `Confirm` to submit the transaction
9. Wait for the `Speed Up` button to appear on the activity log, the click it
10. Accept the default, Click `Submit`

Expected result:
I would expect we default to the `Market - Priority Fee` when the user speeds up a transaction that had a Priority Fee of 0. Please confirm with the confirmation team

```shell
Error: [ethjs-query] while formatting outputs from RPC '{"value":{"code":-32000,"message":"replacement transaction underpriced"}}
```

### Version

10.25.0

### Build type

None

### Browser

Chrome

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

danjm:
Fixes https://github.com/MetaMask/metamask-extension/issues/17353

In the case of transactions that had `maxPriorityFeePerGas === 0`, simply adding 10% to the `maxPriorityFeePerGas` has no effect. If we want to speed up these transactions, we need to set the `maxPriorityFeePerGas` to a non-zero number.

This PR sets the `maxPriorityFeePerGas` to the current medium estimate + 10% in cases where the user was doing a speed up and the original transaction was `0`

Also, in those cases, the `estimatedUsed` is labelled as a `custom` type, which in the UI will tell the user it is "Advanced"

https://user-images.githubusercontent.com/7499938/216162379-8b732407-39d0-4a3c-bf2b-fd6223383ec5.mp4


ui/hooks/gasFeeInput/useTransactionFunctions.js
```diff
@@ -164,17 +168,30 @@ export const useTransactionFunctions = ({
         maxPriorityFeePerGas,
       } = transaction.previousGas || transaction.txParams;
 
+      const newMaxPriorityFeePerGas =
+        maxPriorityFeePerGas === '0x0'
```
darkwing, line 172: I see a case in our codebase where the value is `0x00` or `0x0` (https://github.com/MetaMask/metamask-extension/blob/develop/ui/selectors/selectors.js#L1381) ; do we need to check for both?

ryanml: Am I misremembering, or did we have a constant somewhere for 0 hex values?

brad-decker: BigNumber has a .isZero method (@DDDDDanica suggested we add this to Numeric.ts) we could add that and then instantiate the number and use the .isZero check instead of relying on exact string comparison.

brad-decker:
```
new Numeric(maxPriorityFeePerGas, 16).isZero ?
```

ui/hooks/gasFeeInput/useTransactionFunction.test.js
```diff
@@ -90,6 +90,31 @@ describe('useMaxPriorityFeePerGasInput', () => {
     });
   });
 
+  it('should invoke action updateTransaction with 10% increased max priority fee and medium fee + 10% when updateTransactionToTenPercentIncreasedGasFee callback is invoked while original priority fee is 0', async () => {
```
brad-decker, line 93: We recently decided not to use 'should' anymore. 
```suggestion
  it('invokes action updateTransaction with 10% increased max priority fee and medium fee + 10% when updateTransactionToTenPercentIncreasedGasFee callback is invoked while original priority fee is 0', async () => {
```

brad-decker (dismissed):

ui/hooks/gasFeeInput/useTransactionFunctions.js
danjm: yep, good suggestion

danjm: ah, `Numeric` isn't on develop yet

danjm: oh, yes it is

danjm: ah, but `isZero` is not part of `Numeric` yet

danjm: used bignumber instead, in 783d31067

ui/hooks/gasFeeInput/useTransactionFunctions.js
```diff
@@ -164,17 +169,31 @@ export const useTransactionFunctions = ({
...
       } = transaction.previousGas || transaction.txParams;
 
+      const newMaxPriorityFeePerGas = new BigNumber(
+        maxPriorityFeePerGas,
+        16,
+      ).isZero()
+        ? decGWEIToHexWEI(
+            gasFeeEstimates[defaultEstimateToUse].suggestedMaxPriorityFeePerGas,
+          )
+        : maxPriorityFeePerGas;
```
jpuri, line 179: Hey @danjm : any reason for preferring `CUSTOM_GAS_ESTIMATE` here. Custom I think we use for DAPP suggested values. Here it seems `Medium` is being used as it is default.

pedronfigueiredo (approved):

ui/hooks/gasFeeInput/useTransactionFunction.test.js
```diff
@@ -90,6 +90,31 @@ describe('useMaxPriorityFeePerGasInput', () => {
...
 
+  it('invokes action updateTransaction with 10% increased max priority fee and medium fee + 10% when updateTransactionToTenPercentIncreasedGasFee callback is invoked while original priority fee is 0', async () => {
+    const mockUpdateGasFees = jest
+      .spyOn(Actions, 'updateTransactionGasFees')
+      .mockImplementation(() => ({ type: '' }));
+
+    const { result } = renderUseTransactionFunctions({
+      transaction: {
+        userFeeLevel: CUSTOM_GAS_ESTIMATE,
+        txParams: { maxFeePerGas: '0x5028', maxPriorityFeePerGas: '0x0' },
```
pedronfigueiredo, line 101: [completely optional]:
If we use a conversion method such as:

```javascript
function hexStringFromDec(value) {
  return new Numeric(value, 10).toBase(16).toString()
}
```

Then we can write this as

```javascript
const initialFee = 20520;

transaction: {
        userFeeLevel: CUSTOM_GAS_ESTIMATE,
        txParams: {
          maxFeePerGas: hexStringFromDec(initialFee),
          maxPriorityFeePerGas: hexStringFromDec(0)
        }
},
```

and change the assertion to be explicit about the 10% increase:

```javascript 
    expect(mockUpdateGasFees).toHaveBeenCalledWith(undefined, expect.objectContaining({
      estimateSuggested: expect.any(String),
      estimateUsed: 'custom',
      gas: expect.any(String),
      gasLimit: expect.any(String),
      maxFeePerGas: hexStringFromDec(initialFee * 1.1),
      maxPriorityFeePerGas: expect.any(String),
      userEditedGasLimit: undefined,
      userFeeLevel: expect.any(String),
  })
```

The idea is to assert as much as possible while using variables to make numbers such as `0x582c` more readable in the context of the test.

If the numbers can't be "explained" with variables, that may indicate they don't need to be matched more specifically for that assertion than with `expect.any(String)` or `expect.stringMatching(HEX_REGEX)` with `const HEX_REGEX = /0[xX][0-9a-fA-F]+/;`

ui/hooks/gasFeeInput/useTransactionFunctions.js
danjm: @jpuri The actual estimate used is whatever the default is (usually the 'medium' estimate I think). This is to ensure the "speed up" is successful.

Labelling the "estimate used" as custom ensures the user sees "Advanced" as the type of estimate in the speed-up modal, because the estimate is sort of a "custom" modification on their original "advanced" gas settings.

Does that make sense?

ui/hooks/gasFeeInput/useTransactionFunction.test.js
brad-decker: If we are going to use Numeric it is also valid to do:

```js
const x = new Numeric(value, 10);
const tenPercent = new Numeric('.10', 10); 

// later...
const transaction = {
  txParams: {
    maxFeePerGas: x.toHexString(),
  }
}

// later
expect(..).toHaveBeenCalledWith(., expect.objectContaining({
  ....
  maxFeePerGas: x.times(tenPercent).toHexString();
})
```

I am advocating for getting people comfortable working with numerics directly instead of creating intermediary methods.

danjm: @brad-decker `toHexString` is not a method on `Numeric`?

danjm: @brad-decker @pedronfigueiredo I have begun to collect these suggestions in a follow up PR https://github.com/MetaMask/metamask-extension/pull/17580

I'd like to merge this one quickly as it is a release blocker, but then we can clean up these tests in that PR

ryanml (approved):

ui/hooks/gasFeeInput/useTransactionFunction.test.js
pedronfigueiredo: 100%, LGTM

PR #17556, opened 2023-02-02
NetworkController: Split `network` into `networkId` and `networkStatus`

### LINKED ISSUE #16946: NetworkController API normalization: replace `network` state with `currentNetworkType`
The `network` state of the network controller is a mix of two things: the network ID for the current network, and whether the network controller is "loading" or not.

This is confusing because these two things have little to do with each other. Generally we shouldn't be using network ID for anything either (it has been completely replaced by chain ID, which all custom RPC endpoints have been required to support for over a year now).

We should replace state with a "network type" state, which would be set to an enum representing the "type" of network that is currently active (either "loading", "Infura", or "custom"). This could be called `currentNetworkType`. This preserves the usefulness of the "loading" state, and it provides a way to differentiate Infura networks from custom networks, which is also something we have need of in various places.

mcmire:
## Explanation

The `network` store of the network controller crams two types of data into one place. It roughly tracks whether we have enough information to make requests to the network and whether the network is capable of receiving requests, but it also stores the ID of the network (as obtained via `net_version`).

Generally we shouldn't be using the network ID for anything, as it has been completely replaced by chain ID, which all custom RPC endpoints have been required to support for over a year now. However, as the network ID is used in various places within the extension codebase, removing it entirely would be a non-trivial effort. So, minimally, this commit splits `network` into two stores: `networkId` and `networkStatus`. But it also expands the concept of network status.

Previously, the network was in one of two states: "loading" and "not-loading". But now it can be in one of four states:

- `available`: The network is able to receive and respond to requests.
- `unavailable`: The network is not able to receive and respond to requests for unknown reasons.
- `blocked`: The network is actively blocking requests based on the user's geolocation. (This is specific to Infura.)
- `unknown`: We don't know whether the network can receive and respond to requests, either because we haven't checked or we tried to check and were unsuccessful.

This commit also changes how the network status is determined — specifically, how many requests are used to determine that status, when they occur, and whether they are awaited. Previously, the network controller would make 2 to 3 requests during the course of running `lookupNetwork`.

* First, if it was an Infura network, it would make a request for `eth_blockNumber` to determine whether Infura was blocking requests or not, then emit an appropriate event. This operation was not awaited.
* Then, regardless of the network, it would fetch the network ID via `net_version`. This operation was awaited.
* Finally, regardless of the network, it would fetch the latest block via `eth_getBlockByNumber`, then use the result to determine whether the network supported EIP-1559. This operation was awaited.

Now:

* One fewer request is made, specifically `eth_blockNumber`, as we don't need to make an extra request to determine whether Infura is blocking requests; we can reuse `eth_getBlockByNumber`;
* All requests are awaited, which makes `lookupNetwork` run fully in-band instead of partially out-of-band; and
* Both requests for `net_version` and `eth_getBlockByNumber` are performed in parallel to make `lookupNetwork` run slightly faster.

Fixes #16946.

## Manual Testing Steps

This PR touches a lot of files, but here are some pointers:

* If you run `yarn start` on this branch and restart your local version of MetaMask, if you were previously connected to a network, you should still be connected to that network when you restart the extension
* The loading spinner should spin when you load MetaMask for the first time and the network is being connected to
* Switching networks should continue to work as intended (should be able to send a transaction and otherwise interact with a network, switch network, and interact with the other network in the same way)
* Pending transactions should complete in the order that they were added to the queue
* Should be able to perform a swap on one network, switch the network, and perform a swap on the other network
* There should be no errors printed to the console when first loading, and then interacting, with the test dapp

## Pre-merge author checklist

- [x] I've clearly explained:


mcmire:
This PR is still in draft 1) because it's blocked by other PRs and 2) I need to work through all of the places that this impacts in the extension. I've immediately noticed some places that listen to `networkDidChange` and incorrectly assume that the data that this event produces is the same as what the `networkStore` returned before.

app/scripts/controllers/network/network-controller.js
```diff
@@ -202,26 +202,9 @@ export default class NetworkController extends EventEmitter {
       this.emit(NETWORK_EVENTS.INFURA_IS_UNBLOCKED);
     }
 
-    let networkVersion;
-    let networkVersionError;
-    try {
-      networkVersion = await this._getNetworkId();
```
mcmire, line 208: This isn't right. We should still make this call so we can double check that the network is accessible. See: https://github.com/MetaMask/core/pull/1073/files#diff-ea0aa37793484e278bd59970b7a6ce8380b75b723b34612081d402fb66261cfdR335

app/scripts/controllers/swaps.js
```diff
@@ -136,13 +134,6 @@ export default class SwapsController {
     this.indexOfNewestCallInFlight = 0;
 
     this.ethersProvider = new Web3Provider(provider);
-    this._currentNetwork = networkController.store.getState().network;
```
mcmire, line 139: I removed these lines because they seemed unnecessary. The provider object should never change, so we should not need to recreate the wrapper around the provider.

app/scripts/controllers/network/network-controller.js
```diff
@@ -186,41 +189,35 @@ export default class NetworkController extends EventEmitter {
...
+      this._resetNetworkStatus();
+      this._resetNetworkDetails();
       return;
     }
 
-    // Ping the RPC endpoint so we can confirm that it works
-    const initialNetwork = this.networkStore.getState();
-    const { type } = this.providerStore.getState();
-    const isInfura = INFURA_PROVIDER_TYPES.includes(type);
+    // TODO: Test this new code
```
mcmire, line 197: I'll come back and fix this.

app/scripts/controllers/network/network-controller.js
```diff
@@ -352,13 +356,37 @@ export default class NetworkController extends EventEmitter {
...
-      networkChanged = true;
-    });
+    try {
+      if (isInfura) {
+        return await this._determineInfuraNetworkStatus();
+      }
+      return await this._determineNonInfuraNetworkStatus();
+    } catch (error) {
+      log.warn('MetaMask - Could not determine network status', error);
+      return NetworkStatus.Unknown;
```
mcmire, line 377: True; these aren't tested. I'll come back and add tests.

app/scripts/metamask-controller.js
```diff
@@ -319,9 +323,8 @@ export default class MetamaskController extends EventEmitter {
       {
         onPreferencesStateChange: (listener) =>
           this.preferencesController.store.subscribe(listener),
-        onNetworkStateChange: (cb) =>
-          this.networkController.on(NETWORK_EVENTS.NETWORK_DID_CHANGE, () => {
-            const networkState = this.networkController.store.getState();
+        onNetworkStateChange: (cb) => {
+          this.networkController.store.subscribe((networkState) => {
```
mcmire, line 327: Updated to listen to the network controller rather than `networkDidChange` because we're grabbing the network state and calling the callback with it anyway.

app/scripts/metamask-controller.js
```diff
@@ -1609,7 +1614,7 @@ export default class MetamaskController extends EventEmitter {
 
     function updatePublicConfigStore(memState) {
       const { chainId } = networkController.store.getState().provider;
-      if (memState.network !== 'loading') {
+      if (memState.networkStatus === NetworkStatus.Available) {
```
mcmire, line 1617: This looks different, but is effectively the same, as we have four states instead of two.

app/scripts/metamask-controller.js
```diff
@@ -1647,15 +1647,16 @@ export default class MetamaskController extends EventEmitter {
...
    */
-  getProviderNetworkState(memState) {
-    const { network } = memState || this.getState();
+  getProviderNetworkState() {
+    const { provider, networkStatus } = this.networkController.store.getState();
     return {
-      chainId: this.networkController.store.getState().provider.chainId,
-      networkVersion: network,
+      chainId: provider.chainId,
+      // NOTE: For compatibility with providers, this must return "loading" if
```
mcmire, line 1656: I found this out when running the e2e tests that interact with the test dapp. It'd be nice if we could remove `networkVersion` altogether from this object but we'd have to update `@metamask/providers` (and it would probably impact dapps).

ui/helpers/utils/optimism/buildUnserializedTransaction.js
```diff
@@ -17,9 +17,10 @@ function buildTransactionCommon(txMeta) {
   // genesis points to the mainnet genesis, not the Optimism genesis — but
   // considering that all we want to do is serialize a transaction, this works
   // fine for our use case.
+  const chainId = new BN(stripHexPrefix(txMeta.chainId), 16);
   return Common.forCustomChain(Chain.Mainnet, {
-    chainId: new BN(stripHexPrefix(txMeta.chainId), 16),
-    networkId: new BN(txMeta.metamaskNetworkId, 10),
+    chainId,
+    networkId: chainId.toString(10),
```
mcmire, line 23: This may not be correct, but maybe it doesn't matter?

app/scripts/controllers/transactions/tx-state-manager.js
```diff
@@ -128,7 +130,7 @@ export default class TransactionStateManager extends EventEmitter {
       id: createId(),
       time: new Date().getTime(),
       status: TransactionStatus.unapproved,
-      metamaskNetworkId: netId,
+      metamaskNetworkId: hexToBn(chainId).toString(10),
```
mcmire, line 133: This may not be correct for all networks, but removing this property entirely would be a larger effort.

Gudahtt: Hmm, I thought we had discussed the possibility of submitting a PR ahead of this one that would stop setting this property. Has that been considered? It doesn't seem safe to set invalid network IDs like this.

At first glance it doesn't seem like it would be that complex. It's only being set on 3 lines in this repo.

app/scripts/controllers/swaps.js
Gudahtt: These lines are necessary because Ethers doesn't recognize when the network changes. The provider stops working. That's what we found when this was implemented at least.

Gudahtt:
>This commit also changes the lookupNetwork method within
NetworkController so that we only check whether the network supports
EIP-1559 once we've confirmed that it is available (previously it would
obtain this information regardless).

Hmm. Wouldn't this make it slower for us to determine EIP-1559 compatibility? We'd now have to wait on two round trips rather than one.

mcmire:
> Wouldn't this make it slower for us to determine EIP-1559 compatibility? We'd now have to wait on two round trips rather than one.

Hmm. I suppose that's true. I didn't want to get into how exactly we should check the status of the network in this PR — I'm really just shuffling code around. However, this does get me thinking that perhaps we should use the EIP-1559 check as the network availability check as well? After all, we're just trying to make _some_ kind of request — they should all work if the network is available, and they should all return a `countryBlocked` error if Infura is blocking requests. Would that address your concerns?

Gudahtt:
True, any request should suffice as a check for whether the client is blocked.

I suppose we can drop the `net_version` call entirely, and just use the block request.

I was thinking we'd switch to using `eth_chainId` at some point instead of `net_version` so that we could verify the chain ID each time we began using a chain. But that can come later; keeping `net_version` around in the meantime doesn't serve a purpose if we're not using the result anyway.

mcmire:
@Gudahtt Well, sorry, I spoke too soon, because that contradicts the decision we just made. I think we need to still make a call to `net_version` because we still need the network ID at the moment, until we can remove it.

But maybe what I can do is make the network availability check and the EIP-1559 compatibility check in parallel? We would still want to make sure they complete before resolving `lookupNetwork`, because both of them change state, but would that be better?

Gudahtt:
Yeah absolutely

mcmire:
Placing this back in draft until `network` and `net_version` are added back and until the network availability check and EIP-1559 compatibility check are made in parallel.

app/scripts/controllers/swaps.js
mcmire: Not sure how I didn't notice this comment before. I'll make sure to revert this.

app/scripts/controllers/network/network-controller.js
```diff
@@ -179,62 +198,106 @@ export default class NetworkController extends EventEmitter {
...
-    }
+    this.once(NETWORK_EVENTS.NETWORK_DID_CHANGE, () => {
+      networkChanged = true;
+    });
 
-    let networkVersion;
-    let networkVersionError;
     try {
-      networkVersion = await this._getNetworkId();
+      const results = await Promise.all([
```
mcmire, line 245: There are quite a few changes to this method:

* Two requests are made in parallel to fetch the network ID and the latest block.
* The latter request (for the latest block) will be used to determine whether the network supports EIP-1559.
* If both requests succeed, then we assume that the network is available, but if either request fails, then we cannot make that determination.
* If a request fails, Infura may be blocking requests in general. If this happens, then Infura will return `{ error: 'countryBlocked' }` as the response body (JSON-decoded of course). Prior to these changes, if we knew that we were switching to an Infura network, we would make a direct request for `eth_blockNumber` (using `fetch`) and just test the response directly. We could of course keep doing this, but considering that we already use eth-query to talk to the network, it's better to keep doing that (since it handles errors showing up different ways). The downside to using eth-query, though, is that we now have to catch that error; and that means we have to figure out whether the error we've caught is a JSON-RPC error or something else.
* Previously, the availability check for an Infura network was offloaded to `_checkInfuraAvailability`, but this method was not awaited. That made things painful because it meant that any time we wanted to use `lookupNetwork` — or a method that called `lookupNetwork` — we had to wait for state to change before continuing. Now any subrequest we are making is awaited, i.e., there shouldn't be any async operations in `lookupNetwork` that aren't being awaited. This should clean up a lot of code down the line.

app/scripts/controllers/swaps.js
mcmire: I've brought back the previous version of this controller, making tweaks to ensure that we get network ID and status from the controller properly.

app/scripts/metamask-controller.js
```diff
@@ -917,9 +923,13 @@ export default class MetamaskController extends EventEmitter {
         ),
       getCurrentAccountEIP1559Compatibility:
         this.getCurrentAccountEIP1559Compatibility.bind(this),
-      getNetworkState: () => this.networkController.store.getState().network,
+      getNetworkId: () => this.networkController.store.getState().networkId,
+      getNetworkStatus: () =>
+        this.networkController.store.getState().networkStatus,
+      // TODO: This option should probably listen for networkDidChange
```
mcmire, line 929: This was weird to me, but not necessary to fix in this PR. I'll make a new issue.

app/scripts/metamask-controller.js
```diff
@@ -1662,10 +1668,10 @@ export default class MetamaskController extends EventEmitter {
    * @returns {object} An object with relevant network state properties.
    */
   getProviderNetworkState(memState) {
-    const { network } = memState || this.getState();
+    const { networkId } = memState || this.getState();
     return {
       chainId: this.networkController.store.getState().provider.chainId,
-      networkVersion: network,
+      networkVersion: networkId ?? 'loading',
```
mcmire, line 1674: This preserves the existing behavior of the `network` property.

ui/helpers/utils/tx-helper.ts
```diff
@@ -9,7 +9,7 @@ export default function txHelper(
   decryptMsgs: Record<string, any> | null,
   encryptionPublicKeyMsgs: Record<string, any> | null,
   typedMessages: Record<string, any> | null,
-  network?: string,
+  networkId?: string | null,
```
mcmire, line 12: This mirrors the type of the `networkId` property in NetworkController.

ui/pages/confirm-transaction-base/confirm-transaction-base.test.js
```diff
@@ -33,6 +33,7 @@ const baseStore = {
     unapprovedTxs: {
       1: {
         id: 1,
+        metamaskNetworkId: '5',
```
mcmire, line 36: For some reason these lines were necessary to add in order to make these tests pass. I am not quite sure why, but I also don't know how these tests passed originally anyway considering these transaction objects aren't complete. I can look into this next week.

ui/selectors/selectors.js
```diff
@@ -254,7 +252,7 @@ export function getAccountType(state) {
  * @param {object} state - redux state object
  */
 export function deprecatedGetCurrentNetworkId(state) {
-  return state.metamask.network;
+  return state.metamask.networkId ?? 'loading';
```
mcmire, line 255: This mimics the original behavior of the `network` property.

mcmire:
This is now ready for review again! I've updated the PR title and description to match the latest changes.

app/scripts/controllers/transactions/tx-state-manager.js
mcmire: This is now reverted to use network ID.

app/scripts/controllers/network/network-controller.js
```diff
@@ -29,45 +30,60 @@ import createJsonRpcClient from './createJsonRpcClient';
 
 /**
  * @typedef {object} NetworkConfiguration
+ * @property {string} id - The ID of the network configuration.
```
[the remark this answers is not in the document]
mcmire: Well, I guess it's not used in this PR per se. I can revert this considering we will convert this file to TypeScript anyway.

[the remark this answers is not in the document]
mcmire: Removed here: [4770447](https://github.com/MetaMask/metamask-extension/pull/17556/commits/4770447d5e24c74453df15fc88daf7c70bc9f98b)

app/scripts/metamask-controller.js
mcmire: Comment removed [here](https://github.com/MetaMask/metamask-extension/pull/17556/commits/97727529dc0462b3b01bdd43f6b1d4a5d28dcb78) and issue added [here](https://github.com/MetaMask/metamask-extension/issues/18291).

app/scripts/controllers/network/network-controller.js
```diff
@@ -179,62 +197,106 @@ export default class NetworkController extends EventEmitter {
...
+      if (hasProperty(error, 'data')) {
+        data = error.data;
+      } else if (hasProperty(error, 'code')) {
+        try {
+          data = JSON.parse(error.message);
+        } catch {
+          // error.message must not be JSON for some reason
+        }
+      }
+      if (data) {
```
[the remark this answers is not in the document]
Gudahtt: Ah; I see that our Infura middleware is responsible for this odd message-JSON situation. If there is an unexpected HTTP error response, the _text_ response body is used as the message.

There is no such case in the fetch middleware we use for custom networks though.

Maybe we should split these two cases, and only attempt parsing the message as JSON for Infura networks? We only want to set this blocked status for Infura networks anyway.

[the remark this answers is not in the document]
Gudahtt: But about this condition, putting the Infura blocked case aside, it should be safe to consider a standard non-internal JSON-RPC error as indication that the network is _unavailable_.

The "unknown" case is tricker though because there are categories of error that could indicate either a lack of availability or an unanticipated error in our stack. For example we return "internal" errors and non-RPC errors today in various network error situations, despite the fact that these errors should indicate an error in _our code_.

But I'd be OK with treating these cases as "unknown" for now anyway, and working towards improving error handling in the future to make that distinction more meaningful. With the exception of the Infura blocked error, which is an "internal" error right now but we know to expect it.

We could do that by checking the code of the error. If the code doesn't exist, or if it's `-32603` (for "internal"), then we consider it to be "unknown" (unless we detect that it's a "blocked" response). Otherwise we can consider it to be "unavailable".

Does that sound reasonable? Are there alternatives we could consider? One alternative would be to use "unknown" just for the initial state and consider anything else "unavailable". That would be closer to what we do today, but seems less helpful for debugging situations.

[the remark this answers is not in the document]
mcmire: Good point. At the end of the day, we are trying to determine whether `countryBlocked` is in the response. I think I was trying to look at both the Infura middleware and fetch middleware to see how they represented errors, but we only really care about the Infura middleware. In that case, [all it does is call `ethErrors.rpc.internal` with the unencoded JSON response body](https://github.com/MetaMask/eth-json-rpc-infura/blob/9b6dc41aaa9226fa4a382896d0aee38ec6268e73/src/create-infura-middleware.ts#L173), which will then [create an error object passing along the body as the `message`](https://github.com/MetaMask/eth-rpc-errors/blob/c3ea9e521aa75c553802080b44ff665e428d00cd/src/errors.ts#L64). So I can clean this up a bit.

app/scripts/controllers/network/network-controller.js
```diff
@@ -179,62 +197,106 @@ export default class NetworkController extends EventEmitter {
...
+      this.emit(NETWORK_EVENTS.INFURA_IS_UNBLOCKED);
     } else {
-      this._setNetworkState(networkVersion);
-      // look up EIP-1559 support
-      await this.getEIP1559Compatibility();
+      this._resetNetworkId();
+      this._resetNetworkDetails();
+      if (networkStatus === NetworkStatus.Blocked) {
+        this.emit(NETWORK_EVENTS.INFURA_IS_BLOCKED);
+      }
```
Gudahtt, line 299: The current `lookupNetwork` function also emits this for non-Infura networks before even performing any lookup. That ensures this doesn't get stuck enabled on non-Infura networks.

We could handle that same case here in an `else` block. Or we could do something different, like check what type of network this is earlier on (like in the current `lookupNetwork`, or emit this when the user switches to a custom RPC endpoint instead.

app/scripts/controllers/network/network-controller.js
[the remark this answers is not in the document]
mcmire: I didn't see your last two comments until now, but is more of what you had in mind? [3f0f13d](https://github.com/MetaMask/metamask-extension/pull/17556/commits/3f0f13d1b2349f00bd19afcc9963fa4ba1383386)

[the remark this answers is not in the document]
Gudahtt: Yes, definitely closer, except I'm not sure we should be using `instanceof` here because we have multiple versions of `eth-rpc-errors` in play. The extension is using a different version than the middleware, so it would always be `false` until they were aligned.

Checking the `code` property should be reliable across versions though.

There is also the "internal error" handling to consider; I was suggesting that we should treat internal errors as "unknown" because they should indicate (as the name implies) errors internal to our codebase rather than network errors. But I don't care strongly either way there because of how we're misusing internal errors today.

app/scripts/controllers/network/network-controller.js
```diff
@@ -179,62 +197,106 @@ export default class NetworkController extends EventEmitter {
     if (EIPS[1559] !== undefined) {
       return EIPS[1559];
     }
-    const latestBlock = await this._getLatestBlock();
-    const supportsEIP1559 =
-      latestBlock && latestBlock.baseFeePerGas !== undefined;
-    this._setNetworkEIPSupport(1559, supportsEIP1559);
+    const supportsEIP1559 = await this._determineEIP1559Compatibility();
+    this.networkDetails.putState({
```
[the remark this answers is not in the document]
mcmire: Oh, right, thank you. Updated to `updateState` in [7a72a43](https://github.com/MetaMask/metamask-extension/pull/17556/commits/7a72a43500983cb08b66cc71cf68db12c9698797).

app/scripts/controllers/network/network-controller.js
mcmire: > That ensures this doesn't get stuck enabled on non-Infura networks.

Would this happen if the user switches to an Infura network, receives a blocked error, then switches to a network that is unavailable? As you allude to, maybe we always emit `infuraIsUnblocked` when switching to a non-Infura network, regardless of network status?

app/scripts/controllers/network/network-controller.js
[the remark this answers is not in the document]
mcmire: Okay, it sounds like with my solution, we are always treating an RPC error as unavailable if it's not blocked. I'm also making a check for EthereumRpcError. This may not be necessary, but I'll try adding another check for `code` on top of what I have and see what you think.

app/scripts/controllers/network/network-controller.js
Gudahtt: >Would this happen if the user switches to an Infura network, receives a blocked error, then switches to a network that is unavailable?

Yes, exactly

>maybe we always emit `infuraIsUnblocked` when switching to a non-Infura network, regardless of network status?

Sure, that would make sense to me

ui/pages/confirm-transaction-base/confirm-transaction-base.test.js
Gudahtt: Curious. I found that removing these two lines and adding the property `chainId: CHAIN_IDS.GOERLI,` here also gets them passing.

I guess there was some fallback case for when `chainId` wasn't set that was getting triggered here previously

Gudahtt: Ah, I think I found the problem. There are still references to `network` around. I found one in `ui/pages/confirm-signature-request/index.js`

Gudahtt: This specific problem is probably caused by the `network` reference in `incomingTxListSelector`

Gudahtt: Oh, found one in `confirm-transaction-base.container.js`. That's even closer to this test.

Gudahtt: Found another in `ui.components/app/transaction-alerts/transaction-alerts.stories.js`. I'm finished my `network` audit, that's all that I found.

app/scripts/metamask-controller.js
```diff
@@ -1611,7 +1620,7 @@ export default class MetamaskController extends EventEmitter {
 
     function updatePublicConfigStore(memState) {
       const { chainId } = networkController.store.getState().provider;
-      if (memState.network !== 'loading') {
+      if (memState.networkStatus === NetworkStatus.Available) {
```
Gudahtt, line 1623: There is a reference to `network` a few lines down from here, the second argument to `selectPublicState`, that should be updated to reference `networkId ?? 'loading'`

Gudahtt:
Could you update `app/scripts/lib/setupSentry.js` as well? There is an object there (`SENTRY_STATE`) that is used to filter a Redux-state-shaped object down to just the properties we consider non-personally-identifiable, and `network` was referenced there. We can update that to ensure `networkId` and `networkStatus` are captured separately

Gudahtt:
`test/e2e/fixture-builder.js` should also be updated to replace the old `network` state property

Gudahtt (commented):
Just finished my first pass (mostly - I only skimmed the network-controller tests so far). As expected the `network` state property renaming is proving challenging to keep track of, just as the `provider` rename was. Hopefully I've found all of the remaining references in this pass though. The new name will be much easier to keep track of going forward

app/scripts/controllers/network/network-controller.js
```diff
@@ -372,55 +435,25 @@ export default class NetworkController extends EventEmitter {
...
-    this._setNetworkState('loading');
-    // Reset network details
-    this._clearNetworkDetails();
-    // Configure the provider appropriately
+    this._resetNetworkId();
+    this._resetNetworkStatus();
+    this._resetNetworkDetails();
     this._configureProvider(opts);
-    // Notify subscribers that network has changed
     this.emit(NETWORK_EVENTS.NETWORK_DID_CHANGE, opts.type);
```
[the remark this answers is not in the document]
mcmire: Sure! Removed in [f3fdb19](https://github.com/MetaMask/metamask-extension/pull/17556/commits/f3fdb1967ed4e81d95e7ecb6c95c5e4e9d8c17e6)

mcmire:
`SENTRY_STATE` has been updated in [bfc9894](https://github.com/MetaMask/metamask-extension/pull/17556/commits/bfc9894f93dda1115b7393bd2f35959f5ff67972) and `fixture-builder.js` has been updated in [b4be5a1](https://github.com/MetaMask/metamask-extension/pull/17556/commits/b4be5a14ef9ca58c75f7112b1aac2fc29f51006e).

ui/pages/confirm-transaction-base/confirm-transaction-base.test.js
Gudahtt: Thanks! Though I think there is still one left, `incompingTxListSelector`, in `ui/selectors/transactions.js`

mcmire: Also, I took another look at this test, and the reason why it's failing is due to a change in `deprecatedGetCurrentNetworkId` in `ui/selectors/selectors.js`, but also, my feeling is (still) that the test was not written correctly in the first place.

Previously, `deprecatedGetCurrentNetworkId` returned `state.metamask.network`. Also, in state object that the test sets up, three things were not set: 1) `network`, 2) `metamaskNetworkId` in the unapproved transaction object, and 3) `chainId` in the unapproved transaction object. This meant that when the `unconfirmedTransactionsHashSelector` selector in `ui/selectors/confirm-transaction.js` looked for transactions that matched the current network ID (obtained via `deprecatedGetCurrentNetworkId`), it called `transactionMatchesNetwork({ ..., metamaskNetworkId: undefined, chainId: undefined }, undefined, undefined)`. `undefined` matches `undefined`, so all of the unapproved transactions defined in the test matched.

Now `deprecatedGetCurrentNetworkId` doesn't return `state.metamask.network`, it returns `state.metamask.networkId ?? 'loading'`. That means it can never be undefined. However, the state object in the test is still not correct, so `unconfirmedTransactionsHashSelector` now calls `transactionMatchesNetwork({ ..., metamaskNetworkId: undefined, chainId: undefined }, 'loading', undefined)`. `'loading'` no longer matches `undefined`, so none of the unapproved transactions defined in the test matched.

To fix this, `networkId` should be set to something reasonable, and the unapproved transaction should have at least a `metamaskNetworkId` that matches (`chainId` would make sense too, I suppose).

app/scripts/controllers/network/network-controller.js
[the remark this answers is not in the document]
mcmire: Following up on this, I pushed an update that uses `error.code` to test for an RPC error in [908c366](https://github.com/MetaMask/metamask-extension/pull/17556/commits/908c36699c58160ff7bdd8bff4d3717f7fd47cc1) and added a check for internal errors in [489ae5c](https://github.com/MetaMask/metamask-extension/pull/17556/commits/489ae5cdf01ade3ebdce458192c692f7cc8ef99f).

app/scripts/controllers/network/network-controller.js
mcmire: Okay, this is done in [5cad8ea](https://github.com/MetaMask/metamask-extension/pull/17556/commits/5cad8eacc7f4cf6565cc74e01b0f4208d89455c6).

ui/pages/confirm-transaction-base/confirm-transaction-base.test.js
Gudahtt: Thanks, that makes sense

mcmire: > Thanks! Though I think there is still one left, `incompingTxListSelector`, in `ui/selectors/transactions.js`

Updated in [ce1a1d0](https://github.com/MetaMask/metamask-extension/pull/17556/commits/ce1a1d0c29837a8038dcb4ea5f7850c0a4fba1ed).

Gudahtt (dismissed):
LGTM!

app/scripts/controllers/network/network-controller.js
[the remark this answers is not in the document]
Gudahtt: Thanks, this looks really great now

app/scripts/controllers/network/network-controller.js
```diff
@@ -33,38 +34,58 @@ import { createNetworkClient } from './create-network-client';
  * @property {string} [nickname] - Personalized network name.
  */
 
-const env = process.env.METAMASK_ENV;
-const fetchWithTimeout = getFetchWithTimeout();
+/**
+ * The error code that is used for RPC requests that produce an "internal"
+ * error, borrowed from `eth-rpc-errors`.
+ */
+const INTERNAL_RPC_ERROR_CODE = -32603;
```
adonesky1, line 41: total nit, but should we add this to `shared/constants/network.ts`?

Gudahtt: This is also provided as a constant by the `eth-rpc-errors` package, as `errorCodes.rpc.internal`

mcmire:
I was able to address the E2E test failures, so I don't anticipate any more changes (aside from those via feedback).

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -132,28 +96,65 @@ const JSONRPC_RESPONSE_BODY_PROPERTIES = ['id', 'jsonrpc', 'result', 'error'];
  */
 const INFURA_NETWORKS = [
   {
-    nickname: 'Mainnet',
+    networkName: 'Mainnet',
```
adonesky1, line 99: [nit] I personally think we should be moving towards `chainName`. I know this is just for the test descriptions, but might as well move towards it

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -192,6 +193,9 @@ class NetworkCommunications {
       customRpcUrl,
     };
     this.networkClientType = networkClientType;
+    if (networkClientType !== 'infura' && networkClientType !== 'custom') {
```
adonesky1, line 196: [nit] could we use `NetworkClientType.Infura` and `networkClientType.custom`, the enum thats defined in `create-network-client.ts`

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -618,7 +601,7 @@ describe('NetworkController', () => {
           );
         });
 
-        it('emits infuraIsUnblocked (assuming that the request to eth_blockNumber responds successfully)', async () => {
+        it('emits infuraIsBlocked or infuraIsUnblocked, depending on whether Infura is blocking requests', async () => {
```
adonesky1, line 604: should this block also test a case where Infura is blocked as the new name suggests?

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -660,15 +643,18 @@ describe('NetworkController', () => {
...
-              expect(controller.store.getState().network).toBe(networkVersion);
+              expect(controller.store.getState().networkStatus).toBe(
+                'available',
+              );
             },
           );
         });
 
-        it(`persists to state whether the network supports EIP-1559 (assuming that the request for eth_getBlockByNumber responds successfully)`, async () => {
+        it('determines whether the network supports EIP-1559, storing it in state', async () => {
```
adonesky1, line 657:
```suggestion
        it('determines whether the network supports EIP-1559 and stores the result in state without overwriting other state in the networkDetails store', async () => {
```

test/e2e/tests/send-eth.spec.js
```diff
@@ -298,9 +298,11 @@ describe('Send ETH from dapp using advanced gas controls', function () {
         dapp: true,
         fixtures: new FixtureBuilder()
           .withPermissionControllerConnectedToTestDapp()
-          .withNetworkControllerSupportEIP1559()
           .build(),
-        ganacheOptions,
+        ganacheOptions: {
+          ...ganacheOptions,
+          hardfork: 'london',
```
Gudahtt, line 304: Any idea why this change was needed?

test/e2e/tests/ens.spec.js
```diff
@@ -25,6 +25,34 @@ describe('ENS', function () {
...
+            jsonrpc: '2.0',
+            id: '1111111111111111',
+            result: '0x1',
+          },
+        };
+      });
+
+    await mockServer
+      .forPost(infuraUrl)
+      .withJsonBodyIncluding({ method: 'eth_getBlockByNumber' })
```
Gudahtt, line 44: Any idea why these changes were needed here? I see that we've changed the timing of some of the network calls in the network controller, so maybe that destabilized this test if it was already flaky, but I don't see a clear connection.

app/scripts/controllers/network/network-controller.js
```diff
@@ -176,62 +195,119 @@ export default class NetworkController extends EventEmitter {
...
+        );
+        networkStatus = NetworkStatus.Unknown;
+      }
     }
-    if (initialNetwork !== this.networkStore.getState()) {
+
+    if (networkChanged) {
+      // If the network has changed, then `lookupNetwork` either has been or is
+      // in the process of being called, so we don't need to go further.
       return;
```
adonesky1, line 282: Is this comment saying that the lookupNetwork call we are currently in will be superceded by another incoming call so don't bother setting state below because we're just running the checks above again for a new network?

app/scripts/migrations/083.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  const NetworkController = { ...state.NetworkController };
+
+  if (NetworkController.network === 'loading') {
+    NetworkController.networkId = null;
+    NetworkController.networkStatus = 'unknown';
+  } else {
+    NetworkController.networkId = NetworkController.network;
+    NetworkController.networkStatus = 'available';
```
adonesky1, line 41: can we really know this?

adonesky1: I know we don't want to make network calls in migrations... and I guess its probably the best assumption if you're on it. Ultimately this state is low stakes.

adonesky1 (dismissed):
A few non-blocking questions/nits. Great work!

app/scripts/migrations/083.ts
Gudahtt: I think the intent here is to preserve what the existing state says, which is that the network is available. Not to verify the existing state. Today the `network` state being set to anything other than `loading` would imply availability, or at least that's how we use it

mcmire: Yup, that's right.

test/e2e/tests/ens.spec.js
mcmire: Yes. So this request should have been mocked all along, but previously that was not a problem. The reason why this test was failing was because the loading spinner that appears when you load MetaMask never disappeared. Previously, it would disappear once the `network` state changed from "loading" to the network ID, which happened when `net_version` returned with a successful response. A request to `eth_getBlockByNumber` was also made in order to get the latest block to determine EIP-1559 support, but that request was not awaited, and it didn't impact whether `network` was updated or not. Now, requests to `net_version` and `eth_getBlockByNumber` are made simultaneously, and updating `networkStatus` (which the spinner now uses to determine the loaded status) relies on both requests succeeding. Hence why I now have to mock this request.

test/e2e/tests/send-eth.spec.js
mcmire: Sort of. The default hardfork we set in `withFixtures` is an old one where EIP-1559 isn't supported. Previously this test initialized the network controller state to set `networkDetails` to `{ EIPS: { 1559: true } }`, i.e., the controller already knew that the network supported EIP-1559, even if no provider was set yet. Although `getEIP1559Compatibility` is called when a provider is initialized, so this would eventually get changed in state to `{ EIPS: { 1559: false } }`, that happened after the `network` state was changed. It's possible that this test was running so fast that the MetaMask notification was spawned before `networkDetails` was updated. However, now when MetaMask loads, we guarantee that `networkDetails` is updated at the same time that `networkId` and `networkStatus` are updated, so when the notification window opens, `networkDetails` is ``{ EIPS: { 1559: false } }`. That's not what we want, obviously. Since we're using Ganache, we can just switch to a hardfork that supports EIP-1559 instead of manipulating state.

app/scripts/controllers/network/network-controller.js
mcmire: Yup, that's right!

app/scripts/controllers/network/network-controller.test.js
```diff
@@ -5393,42 +6751,44 @@ describe('NetworkController', () => {
...
+            expect(controller.store.getState().networkStatus).toBe('available');
 
             await waitForLookupNetworkToComplete({
               controller,
               operation: () => {
                 controller.rollbackToPreviousProvider();
               },
             });
-            expect(controller.store.getState().network).toBe('42');
+            expect(controller.store.getState().networkStatus).toBe('unknown');
```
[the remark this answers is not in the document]
mcmire: Hopefully the previous network is still available, but the code goes through the same process that it would if you were switching to that network for the first time. I guess I don't really know what the proper behavior should be here, but I'm just testing what the code does.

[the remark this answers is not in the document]
adonesky1: I guess it would seem strange to call `rollbackToPreviousProvider` on a network where the status is `available`? I guess I was thinking the point of this method is to rollback when the network isn't available.

app/scripts/controllers/network/network-controller.test.js
mcmire: Well, my goal here was to test that `lookupNetwork` gets called. I think I do have a test for `lookupNetwork` specifically that confirms that `infuraIsBlocked` is called when Infura is blocking and that `infuraIsUnblocked` is called when it's not. But if it makes more sense to test `infuraIsBlocked` instead of `infuraIsUnblocked` I can do that here.

app/scripts/controllers/network/network-controller.test.js
mcmire: Oh yeah! Well this check will go away once I convert this file to TypeScript, but I'll definitely make sure to use that enum then.

Gudahtt (dismissed):
LGTM, though it'd be nice to remove that `withNetworkControllerSupportEIP1559` now that it's unused

app/scripts/controllers/network/network-controller.js
mcmire: That makes sense! Updated here: [90bc950](https://github.com/MetaMask/metamask-extension/pull/17556/commits/90bc950cd9e05486812b6be7d8d99e48f7e65bbe)

app/scripts/controllers/network/network-controller.test.js
mcmire: We don't really need this property, so I removed it to avoid confusion: [6594eeb](https://github.com/MetaMask/metamask-extension/pull/17556/commits/6594eeb0905ce19d0ec3b828f3e4dfe7f6b13083)

mcmire:
> LGTM, though it'd be nice to remove that `withNetworkControllerSupportEIP1559` now that it's unused

No problem! Removed in [2a46bcf](https://github.com/MetaMask/metamask-extension/pull/17556/commits/2a46bcf25c80494b172263f3fb90dc0bd2c2d814).

Gudahtt (dismissed):

Gudahtt:
Oh, lots of new conflicts it seems

test/e2e/tests/ens.spec.js
mcmire: Ah, yes, that's true. That's to prevent noise from showing up when running the test.

app/scripts/controllers/network/network-controller.test.js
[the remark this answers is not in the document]
mcmire: Ah, I see what you mean about it being opposite from reality. Sorry didn't even think about when writing this test 😅 Fixed in [d2b5d78](https://github.com/MetaMask/metamask-extension/pull/17556/commits/d2b5d789314cc895271c3f3b55ea881b83de5ab4).

mcmire:
@Gudahtt @adonesky1 Okay ready for one more review (hopefully the last one 🤞🏻 )

Gudahtt (approved):
LGTM! Glad the tests caught those problems

adonesky1 (approved):
Lgtm!

PR #17559, opened 2023-02-02
Fix display of 0 priority fee in advanced gas edit input

### LINKED ISSUE #17519: [Bug]: Cannot set Priority Fee to 0 for transaction
### Describe the bug

User cannot submit a transaction with 0 Priority Fee: when user enter 0 in priority fee field, it dose not show 0 on confirmation screen.

https://user-images.githubusercontent.com/105063779/215830418-84e05346-6312-4818-8b67-4cc65c022546.mov

### Steps to reproduce

1. Choose or enter an address to start a transaction
2. Click Next
3. In Advanced options, put 0 in the priority fee field
4. Save and check the displayed fee on confirmation screen (which should be 0)
5. Reopen the edit gas fee window and check that the previous fee is displayed instead of 0

_No response_

### Version

10.25.0

### Build type

None

### Browser

Chrome

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

danjm:
Fixes #17519

Prior to this PR, `useMaxPriorityFeePerGasInput` had a bug whereby if the `maxPriorityFeePerGas` was set to 0 within the transaction data, it would still return the default/medium/market estimate.

The effect of this bug was that if a user custom set the priority fee to 0, and then re-oped the edit gas modal and looked at the advanced setting, they would see a number other than 0 (even though zero is what would be submitted).

The cause of this bug is that the `maxPriorityFeePerGas` within the `useMaxPriorityFeePerGasInput` function would be initialized to `null` if `supportsEIP1559 && initialMaxPriorityFeePerGas` was not true. It would not be true if `initialMaxPriorityFeePerGas` was the number `0`, which would be the case if `maxPriorityFeePerGas` was set to `0x0` in the transaction data.

To fix the bug, we only initialize `maxPriorityFeePerGas` to `null` if `initialMaxPriorityFeePerGas` is `null` or `undefined`. Also, to make sure we don't break the requirement that we return the default/medium/market estimate if there is no priority fee on the transaction data, we default to `null` in such cases so that subsequent code can distinguish between an intentional `0` value and `null`.

A unit test is added which failed before this PR and now passes


ui/hooks/gasFeeInput/useMaxPriorityFeePerGasInput.js
```diff
@@ -7,6 +7,8 @@ import { isLegacyTransaction } from '../../helpers/utils/transactions.util';
 import { hexWEIToDecGWEI } from '../../../shared/modules/conversion.utils';
 import { feeParamsAreCustom, getGasFeeEstimate } from './utils';
 
+const isNullOrUndefined = (value) => value === null || value === undefined;
```
pedronfigueiredo, line 10: Is it worth moving this to a shared utilities file such as `ui/helpers/utils/util.js`?

pedronfigueiredo (approved):

ui/hooks/gasFeeInput/useMaxPriorityFeePerGasInput.js
pedronfigueiredo: this can be done, LGTM as it's a release blocker

jpuri (approved):

PR #17563, opened 2023-02-02
Fix http-cache-semantics dependency vulnerability GHSA-rc47-6667-2j5j

danjm:
Fixes https://github.com/advisories/GHSA-rc47-6667-2j5j

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,6 +99,7 @@
     "json-schema": "^0.4.0",
     "simple-get": "^4.0.1",
     "web3-provider-engine/eth-json-rpc-filters": "^5.0.0",
+    "http-cache-semantics": "^4.1.1",
     "typescript@~4.4.0": "patch:typescript@npm:4.4.4#.yarn/patches/typescript-npm-4.4.4-3fedcc07a3.patch",
     "acorn@^7.0.0": "patch:acorn@npm:7.4.1#.yarn/patches/acorn-npm-7.4.1-f450b4646c.patch",
     "acorn@^7.4.1": "patch:acorn@npm:7.4.1#.yarn/patches/acorn-npm-7.4.1-f450b4646c.patch",
--- a/package.json
+++ b/package.json
@@ -99,7 +99,6 @@
     "json-schema": "^0.4.0",
     "simple-get": "^4.0.1",
     "web3-provider-engine/eth-json-rpc-filters": "^5.0.0",
-    "http-cache-semantics": "^4.1.1",
     "typescript@~4.4.0": "patch:typescript@npm:4.4.4#.yarn/patches/typescript-npm-4.4.4-3fedcc07a3.patch",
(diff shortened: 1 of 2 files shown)


pedronfigueiredo (dismissed):

package.json
```diff
@@ -99,6 +99,7 @@
     "json-schema": "^0.4.0",
     "simple-get": "^4.0.1",
     "web3-provider-engine/eth-json-rpc-filters": "^5.0.0",
+    "http-cache-semantics": "^4.1.1",
```
Gudahtt, line 102: This is an in-range bump, so we don't need to leave a resolution here for this
```suggestion
```

Gudahtt: To clarify, this is a good way of bumping an indirect dependency, this is what I would do as well. Just we can remove it again after the lockfile has been updated if it's compatible with the requested ranges. This makes future updates less complicated, as conflicts with these resolutions could cause issues that would be hard to spot.

Gudahtt (dismissed):

Gudahtt:
Oh, another lockfile update needed, since the resolution was removed

DDDDDanica (dismissed):

yarn.lock
```diff
@@ -19096,10 +19096,10 @@ __metadata:
   languageName: node
   linkType: hard
 
-"http-cache-semantics@npm:^4.0.0, http-cache-semantics@npm:^4.1.0":
-  version: 4.1.0
-  resolution: "http-cache-semantics@npm:4.1.0"
-  checksum: 974de94a81c5474be07f269f9fd8383e92ebb5a448208223bfb39e172a9dbc26feff250192ecc23b9593b3f92098e010406b0f24bd4d588d631f80214648ed42
+"http-cache-semantics@npm:^4.1.1":
```
Gudahtt, line 19099: Here is the required update:
```suggestion
"http-cache-semantics@npm:^4.0.0, http-cache-semantics@npm:^4.1.0":
```

Gudahtt: Hmm, GitHub isn't letting me commit this suggestion. Keeps giving an error about the diff being updated. Seems like a bug. Oh well.

danjm: done in latest commit

Gudahtt (approved):

DDDDDanica (approved):

brad-decker (approved):

Gudahtt:
These Socket.dev warnings appear to be in error. Unsure why they were triggered. Those packages are not affected by this PR.

brad-decker:
@SocketSecurity ignore ganache@7.7.4

brad-decker:
@SocketSecurity ignore secp256k1@4.0.2

brad-decker:
@SocketSecurity ignore secp256k1@4.0.3

brad-decker:
@SocketSecurity ignore node-gyp-build@4.6.0

PR #17537, opened 2023-02-01
Improving approve transaction flow

jpuri:
Improve approve transaction flow

https://app.zenhub.com/workspaces/confirmations-6245e6e2348677001213b8d2/issues/gh/metamask/metamask-planning/317

PR #17571, opened 2023-02-02
Sanitising string on signature request pages

jpuri:
Sanitising string on signature request pages

Ref: https://app.zenhub.com/workspaces/confirmations-6245e6e2348677001213b8d2/issues/gh/metamask/metamask-planning/306


ui/helpers/utils/util.js
```diff
@@ -516,3 +517,20 @@ export function coinTypeToProtocolName(coinType) {
 export function isNullish(value) {
   return value === null || value === undefined;
 }
+
+/**
+ * The method escape RTL character in string
+ *
+ * @param {string} str
+ * @returns {string} escaped string
```
digiwand, line 525:
```suggestion
 * @returns {(string|*)} escaped string or original param value
```

`if (!str)`, then this function may return something other than a string

ui/components/app/signature-request/signature-request-data/signature-request-data.js
```diff
@@ -19,61 +19,68 @@ import {
...
+                    className="signature-request-data__node__value__address"
+                  >
+                    <Address
+                      addressOnly
+                      checksummedRecipientAddress={toChecksumHexAddress(value)}
+                      recipientName={getAccountName(identities, value)}
+                    />
+                  </Typography>
+                ) : (
+                  value
```
digiwand, line 77:
```suggestion
                  `${value}`
```

I think we should add the string conversation back. 

If the value is an object, we may get the error `Uncaught Error: Objects are not valid as a React child` here; or if this is a boolean, the value may not print.

ui/components/app/signature-request-original/signature-request-original.component.js
```diff
@@ -177,8 +177,12 @@ export default class SignatureRequestOriginal extends Component {
                 className="request-signature__row"
                 key={`request-signature-row-${index}`}
               >
-                <div className="request-signature__row-title">{`${name}:`}</div>
-                <div className="request-signature__row-value">{value}</div>
+                <div className="request-signature__row-title">{`${sanitizeString(
+                  name,
+                )}:`}</div>
```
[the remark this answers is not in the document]
jpuri: Thanks a lot @digiwand for bringing this up, I updated it to `{sanitizeString(`${name}:`)}`

ui/helpers/utils/util.js
```diff
@@ -524,3 +525,20 @@ export function coinTypeToProtocolName(coinType) {
...
   return value === null || value === undefined;
 }
+
+/**
+ * The method escape RTL character in string
+ *
+ * @param {string} str
+ * @returns {(string|*)} escaped string or original param value
+ */
+export const sanitizeString = (str) => {
```
digiwand, line 535: I suppose the `str` param could be type any (*) now and maybe we want to rename the parameter to `value`. 

That, or we could allow this function to convert the value by renaming to function, e.g. `stringifyAndSanitize`, or adding an optional param, `stringify = true`

jpuri: Good point @digiwand : I renamed arg to `value`.

ui/helpers/utils/util.js
```diff
@@ -540,9 +538,27 @@ export function isNullish(value) {
...
   );
 
   return pathMetadata?.name ?? null;
 }
 ///: END:ONLY_INCLUDE_IN
+
+/**
+ * The method escape RTL character in string
+ *
+ * @param {any} value
```
digiwand, line 552: maybe 
```suggestion
 * @param {*} value
```
ref: https://jsdoc.app/tags-param.html

digiwand (approved):
Nice fix!

segun (approved):

ui/helpers/utils/util.js
jpuri: Good catch @digiwand , I will add this in a following PR.

jpuri: Done in this PR https://github.com/MetaMask/metamask-extension/pull/17537

seaona:
I see it fixed!! Great job @jpuri !!

PR #17580, opened 2023-02-02
Improve useTransactionFunction unit tests

danjm:


brad-decker:
Should this be updated to be included @danjm ?

PR #17592, opened 2023-02-03
Add desktop build type

matthewwalsh0:
## Explanation

Adds the `desktop` build type to isolate any future code required to connect to the [MetaMask Desktop](https://github.com/MetaMask/metamask-desktop) application.

As the desktop application will support snaps, this build type will require all the existing `flask` code fences and so basic build type inheritance has also been implemented to avoid adding the `desktop` build type to every existing and future `flask` code fence.

Additional CircleCI jobs have also been added to verify the desktop build type in the pipeline.

## Manual Testing Steps

`yarn start --build-type desktop`

As no desktop support code has been added yet, any manual testing should simply verify the build appears and functions identically to the Flask build.

## Pre-merge author checklist

- [x] I've clearly explained:


.circleci/config.yml
```diff
@@ -1118,6 +1172,22 @@ jobs:
           name: Validate source maps
           command: yarn validate-source-maps
 
+  validate-source-maps-desktop:
```
matthewwalsh0, line 1175: Definitely beyond the scope of this PR, but for any build type specific jobs with a very similar structure, we could use CircleCI commands to define the steps once with a `buildType` parameter and avoid some of this duplication.

cryptotavares (dismissed):

development/build/transforms/remove-fenced-code.test.js
```diff
@@ -757,6 +805,56 @@ Always_Included
...
+   Always_Included
+Always_Included
+   Always_Included
+Always_Included
+
+
+`,
+        true,
+      ],
+      flask: [
```
Gudahtt, line 818: Hah, I forgot all about these tests. This was quite strange to come across in the diff, I didn't know what I was looking at.

I was wondering why I didn't see any new fixtures for desktop, but I see now that the expected output for desktop is that all fences are preserved. So desktop is the new "base case" here rather than Flask. If we had included `ONLY_INCLUDE_IN(beta)` in any of these fixtures, we'd need to have an explicit output for each one.

Let me know if I've misunderstood what's going on here

development/build/transforms/remove-fenced-code.js
```diff
@@ -337,7 +337,16 @@ function removeFencedCode(filePath, typeOfCurrentBuild, fileContent) {
       // Throws an error if the command parameters are invalid
       CommandValidators[command](parameters, filePath);
 
-      if (parameters.includes(typeOfCurrentBuild)) {
+      const validBuildTypes = [
+        typeOfCurrentBuild,
+        ...(BuildTypeInheritance[typeOfCurrentBuild] || []),
```
Gudahtt, line 342: This build inheritance idea is great! Really simplified this PR.

Not critical, but it would be nice to have a targeted unit test to ensure it's working correctly. I see that it's covered already in that large test fixture (which is fantastic and very much appreciated), but to somebody unfamiliar with this code, this behavior might not be obvious.

Having a targeted test with a minimal input _just_ to cover this build type inheritance idea would do a better job of ensuring this is understood and continues to work in the future.

Gudahtt (dismissed):
LGTM! I had a couple of suggestions but nothing major

app/build-types/desktop/manifest/firefox.json
```diff
@@ -0,0 +1,7 @@
+{
+  "applications": {
+    "gecko": {
+      "id": "webextension-flask@metamask.io"
```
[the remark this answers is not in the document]
matthewwalsh0: Done.

development/build/transforms/remove-fenced-code.test.js
matthewwalsh0: Exactly right, all the flask code fences stay, as desktop inherits from flask, and there was no previous beta code fences so I didn't add any either.

development/build/transforms/remove-fenced-code.js
matthewwalsh0: Great idea, done.

Gudahtt (approved):

cryptotavares (approved):

PR #17603, opened 2023-02-04
Label PRs based on the labels of the associated issue

### LINKED ISSUE #17602: Create an automation to label PRs based on the associated issue

pedronfigueiredo:
## Explanation

We search on the body of our PR for the issue it closes (as per [our PR template](https://github.com/MetaMask/metamask-extension/blob/develop/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L11)).

A one-to-one link can only be established if the PR body contains one issue number. Otherwise, we use the branch name to establish the connection between the Issue and the Pull Request. For that to work, we nudge developers to do that via the (opt-in) pre-commit hook and enforce it in the CI job.

Syncing the labels is unidirectional and non-destructive. PRs get the labels from the associated issue, but an issue does not ever get new labels automatically. Preexisting PR labels that are not on the issues persist.

No labels are copied if developers do not include an issue number on the PR body and the branch name does not adhere to the convention.

Closes #17602 

 ## How to test

- Remove PR labels
- Clear assignees for the PR
- Assign the PR to me
- The relevant job will trigger and add the issue labels automatically

### DIFF
--- a/.github/scripts/label-prs.ts
+++ b/.github/scripts/label-prs.ts
@@ -1,7 +1,20 @@
-import { context } from '@actions/github';
+import * as core from '@actions/core';
+import { context, getOctokit } from '@actions/github';
 import { GitHub } from '@actions/github/lib/utils';
 
 type GithubClient = InstanceType<typeof GitHub>;
+
+const token = process.env.GITHUB_TOKEN;
+
+console.log({ token });
+
+if (!token) {
+  core.setFailed('GITHUB_TOKEN not found');
+  process.exit(1);
+}
+
+const octokit = getOctokit(token);
+
 
 main().catch((error: Error): void => {
   console.error(error);
@@ -20,7 +33,7 @@ async function main(): Promise<void> {
     issueNumber = getIssueNumberFromBranchName(headRef);
   }
 
-  await updateLabels(github, issueNumber);
+  await updateLabels(github, issueNumber, octokit);
 }
 
 async function getIssueNumberFromPullRequestBody(): Promise<string> {
@@ -96,7 +109,7 @@ function bailIfIsNotFeatureBranch(branchName: string): void {
   }
 }
 
-async function updateLabels(github: GithubClient, issueNumber: string): Promise<void> {
+async function updateLabels(github: GithubClient, issueNumber: string, octokit: InstanceType<typeof GitHub>): Promise<void> {
   interface ILabel {
     name: string;
   };
@@ -133,12 +146,18 @@ async function updateLabels(github: GithubClient, issueNumber: string): Promise<
     dedupedFinalPRLabels,
   );
   if (hasIssueAdditionalLabels) {
-    await github.request(`PATCH /repos/${owner}/${repo}/issues/${prNumber}`, {
+    await octokit.rest.issues.update({
       owner,
       repo,
       issue_number: prNumber,
       labels: dedupedFinalPRLabels,
     });
+    // await github.request(`PATCH /repos/${owner}/${repo}/issues/${prNumber}`, {
+    //   owner,
+    //   repo,
+    //   issue_number: prNumber,
+    //   labels: dedupedFinalPRLabels,
+    // });
   }
 }
 
--- a/.github/scripts/label-prs.ts
+++ b/.github/scripts/label-prs.ts
@@ -1,47 +1,74 @@
-async function runLabelPRsAutomation(context, github) {
-  let issueNumber = getIssueNumberFromPullRequestBody(context);
+import { context } from '@actions/github';
+import { GitHub } from '@actions/github/lib/utils';
 
-  if (issueNumber === -1) {
-    bailIfIsBranchNameInvalid(process.env.HEAD_REF);
-    bailIfIsNotFeatureBranch();
-    issueNumber = getIssueNumberFromBranchName(process.env.HEAD_REF);
+type GithubClient = InstanceType<typeof GitHub>;
+
+main().catch((error: Error): void => {
+  console.error(error);
+  process.exit(1);
+});
+
+async function main(): Promise<void> {
+  const github = new GitHub();
+
+  const logs = JSON.stringify({ github, context });
+  console.log(logs);
+  // github.event.pull_request.head.ref || "";
+  const HEAD_REF = 'test';
+
+  let issueNumber = await getIssueNumberFromPullRequestBody();
+
+  if (issueNumber === "") {
+    bailIfIsBranchNameInvalid(HEAD_REF);
+    bailIfIsNotFeatureBranch(HEAD_REF);
+    issueNumber = getIssueNumberFromBranchName(HEAD_REF);
   }
 
-  await updateLabels(context, github, issueNumber);
+  await updateLabels(github, issueNumber);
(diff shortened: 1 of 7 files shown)


brad-decker:
Does the branch naming automatically link issue to PR in zenhub?

pedronfigueiredo:
> Does the branch naming automatically link issue to PR in zenhub?

@brad.decker no, but it's used on the CI that runs on the PR as a means to know the related issue. See this line: https://github.com/MetaMask/metamask-extension/pull/17603#discussion_r1097947301

brad-decker:
tangent: could we figure out how we can associate the pr to the issue automatically (either as part of this or a follow up) that would be a *HUGE* bonus and make PMs happy for ever.

brad-decker (dismissed):

.github/workflows/label-prs.yml
```diff
@@ -0,0 +1,95 @@
...
+  validate-branch-name:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Validate branch name
+        env:
+          HEAD_REF: ${{ github.event.pull_request.head.ref }}
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const BRANCH_REGEX = /^main|develop|(ci|chore|docs|feat|fix|perf|refactor|revert|style)\/\d*(?:[-](?![-])\w*)*$/;
```
DDDDDanica, line 17: would be nice to have it in readMe or developer guide for creating the branch, wdyt?

.github/workflows/label-prs.yml
```diff
@@ -0,0 +1,95 @@
...
+      issues: write
+      pull-requests: write
+    steps:
+      - name: Label PR
+        env:
+          HEAD_REF: ${{ github.event.pull_request.head.ref }}
+        uses: actions/github-script@v6
+        with:
+          script: |
+            if (process.env.HEAD_REF === 'main' || process.env.HEAD_REF === 'develop') {
```
mcmire, line 43: It might be better if we could extract this script to a separate file in `.circleci/scripts`. Then it'd get proper syntax highlighting, it could be linted, etc.

.validate-branch-namerc.js
```diff
@@ -0,0 +1,11 @@
+const BRANCH_REGEX =
```
mcmire, line 1: Should this be a dotfile? Also typically files that end in `rc` are config files, but this is a script. Wondering if it makes sense to move this to the `development` directory and call this `validate-branch-name.js`.

README.md
```diff
@@ -15,21 +15,27 @@ To learn how to contribute to the MetaMask project itself, visit our [Internal D
 ## Building locally
 
 - Install [Node.js](https://nodejs.org) version 16
-    - If you are using [nvm](https://github.com/nvm-sh/nvm#installing-and-updating) (recommended) running `nvm use` will automatically choose the right node version for you.
```
mcmire, line 18: Were these changes intentional? If Prettier necessitated these changes, then I'm wondering why we haven't had an issue up to now. If not, then perhaps extract these to a new PR?

.github/workflows/label-prs.yml
```diff
@@ -0,0 +1,95 @@
...
+
+on:
+  pull_request:
+    types: [assigned, opened, edited, synchronize, reopened]
+
+jobs:
+  validate-branch-name:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Validate branch name
```
mcmire, line 11: I'm curious whether it makes sense to run this check in CI. At this point, the developer has created a branch and pushed it up. So if the name of that branch is not in the expected format and they see this check failing, the developer would have to close the PR, rename the branch locally, then make a new PR against that branch. I feel like that might lead to a painful process for people who don't know about this (new hires and contributors).

Is it absolutely necessary that PRs get assigned the correct labels? I'm wondering if it would be better to remove this job and update the job below this so that it bails early if it can't read the issue number from the name of the branch.

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+yarn validate-branch-name
```
mcmire, line 4: What do you think about running this on pre-push rather than pre-commit? Because the labeling occurs in a CircleCI job, it seems that we only really care about branches as soon as they appear on GitHub. It would also be less intrusive for people who commit frequently or employ rebasing frequently.

pedronfigueiredo:
> could we figure out how we can associate the pr to the issue automatically (either as part of this or a follow up) that would be a HUGE bonus and make PMs happy for ever.

> Is it absolutely necessary that PRs get assigned the correct labels? I'm wondering if it would be better to remove this particular job and update the job below this so that it bails early if it can't read the issue number from the name of the branch.

Great points, @brad-decker @mcmire. I did a bit more reading and:

GitHub [provides linking][linking-pr], but we can't rely on people to include the right format even though it is in [our PR template][pr-template]. However, I decided to add mentions in the PR text as an additional method for getting the issue number.

In summary, the order of operations now is as follows:
- if only one issue is mentioned in the PR body, get the labels off that issue.
- if no issue is mentioned, or more than one is
  - Check if the branch name has the expected format (exclude main, develop, and rc branches)
    - if it is, get the issue number off the branch name and copy the labels.
    - if it isn't, bail with no error, so the job still passes. No labels are copied.

[linking-pr]: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue
[pr-template]: https://github.com/MetaMask/metamask-extension/blob/develop/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L11

.circleci/scripts/label-prs.js
```diff
@@ -0,0 +1,142 @@
+async function runLabelPRsAutomation(context, github) {
```
brad-decker, line 1: Only things that run in circleci should be in this folder probably, might make sense to add it to the .github folder as it is ran by the github actions.

brad-decker (changes requested):

.validate-branch-namerc.js
mcmire, line 1: Also, going along with the other suggestion above, could this also be a TypeScript file? If it's too much trouble because it requires monkeying with the TypeScript/ESLint config, no problem, but I thought I'd mention it.

.github/scripts/label-prs.ts
```diff
@@ -0,0 +1,167 @@
...
+    process.exit(1);
+  }
+
+  const prBody = context.payload.pull_request?.body || '';
+  return prBody;
+}
+
+function bailIfIsBranchNameInvalid(branchName: string): void {
+  const BRANCH_REGEX =
+    /^(main|develop|(ci|chore|docs|feat|feature|fix|perf|refactor|revert|style)\/\d*(?:[-](?![-])\w*)*|Version-v\d+\.\d+\.\d+)$/;
```
DDDDDanica, line 69: we can also optional chain `beta` after version one, wdyt?

DDDDDanica (approved):

albertolive (approved):

brad-decker (approved):

PR #17625, opened 2023-02-06
feat: refactor background.js to support overrides

bergarces:
## Explanation

Adds a series of features for `background.js` that will enable support for [Metamask Desktop](https://github.com/MetaMask/metamask-desktop) in a future PR.

This PR, by itself, should not change the behaviour of the extension.

A summary of the changes added include:
- Hides call to `initialize()` behind an environment variable (the absence of which maintains existing behaviour).
- Adds an exports an event emitter so that we can notify any listener (in this case Metamask Desktop) that the state has been changed.
- Exports `setupController` and `loadStateFromPersistence`.
- Adds optional `overrides` parameter to `setupController`, which is then passed down to `MetamaskController` so that specific behaviours can be overridden when called from Metamask Desktop.

## Manual Testing Steps

yarn start

As no desktop support code has been added yet, any manual testing should simply verify the build appears and functions identically.

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/app/scripts/background.js
+++ b/app/scripts/background.js
@@ -2,6 +2,7 @@
  * @file The entry point for the web extension singleton process.
  */
 
+import EventEmitter from 'events';
 import endOfStream from 'end-of-stream';
 import pump from 'pump';
 import debounce from 'debounce-stream';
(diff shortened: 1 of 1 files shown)


app/scripts/background.js
```diff
@@ -426,7 +430,10 @@ function setupController(initState, initLangCode) {
   pump(
     storeAsStream(controller.store),
     debounce(1000),
-    createStreamSink((state) => localStore.set(state)),
+    createStreamSink(async (state) => {
+      await localStore.set(state);
+      statePersistenceEvents.emit('state-persisted', state);
```
bergarces, line 435: We thought this was harmless to add without the need of an override, as we are just emitting an event for anyone that might be interested about a state change.

Happy to discuss otherwise.

cryptotavares (approved):

jpuri:
`yarn start:mv3` to ensure that it works well for mv3 will also be useful as this code is important for MV3 also. cc @pedronfigueiredo

jpuri (approved):

pedronfigueiredo (approved):

PR #17635, opened 2023-02-06
POC: reading state from storage instead of the background on subsequent loads

brad-decker:
…nt loads

## Pre-merge author checklist

- [ ] I've clearly explained:

### DIFF
--- a/app/scripts/ui.js
+++ b/app/scripts/ui.js
@@ -98,17 +98,18 @@ if (isManifestV3) {
 start().catch(log.error);
 
 async function start() {
+  console.log('test');
   // create platform global
   global.platform = new ExtensionPlatform();
 
   // identify window type (popup, notification)
   const windowType = getEnvironmentType();
 
-  let isUIInitialised = false;
-
   // setup stream to background
   extensionPort = browser.runtime.connect({ name: windowType });
   let connectionStream = new PortStream(extensionPort);
+
+  console.log('here');
 
   const activeTab = await queryCurrentActiveTab(windowType);
 
@@ -124,13 +125,9 @@ async function start() {
      */
     const messageListener = async (message) => {
       if (message?.data?.method === 'startUISync') {
-        if (isUIInitialised) {
-          // Currently when service worker is revived we create new streams
-          // in later version we might try to improve it by reviving same streams.
-          updateUiStreams();
-        } else {
-          initializeUiWithTab(activeTab);
-        }
+        // Currently when service worker is revived we create new streams
+        // in later version we might try to improve it by reviving same streams.
+        updateUiStreams();
         await loadPhishingWarningPage();
       }
     };
@@ -248,6 +245,7 @@ async function start() {
 
     extensionPort.onMessage.addListener(messageListener);
     extensionPort.onDisconnect.addListener(resetExtensionStreamAndListeners);
+    initializeUiWithTab(activeTab);
   } else {
     initializeUiWithTab(activeTab);
   }
@@ -259,7 +257,6 @@ async function start() {
         displayCriticalError('troubleStarting', err, store);
         return;
       }
-      isUIInitialised = true;
 
       const state = store.getState();
       const { metamask: { completedOnboarding } = {} } = state;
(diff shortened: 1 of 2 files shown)


app/scripts/ui.js
```diff
@@ -124,13 +125,9 @@ async function start() {
...
-        if (isUIInitialised) {
-          // Currently when service worker is revived we create new streams
-          // in later version we might try to improve it by reviving same streams.
-          updateUiStreams();
-        } else {
-          initializeUiWithTab(activeTab);
-        }
+        // Currently when service worker is revived we create new streams
+        // in later version we might try to improve it by reviving same streams.
+        updateUiStreams();
```
brad-decker, line 130: This was changed because we immediately initialize  the UI now

app/scripts/ui.js
```diff
@@ -248,6 +245,7 @@ async function start() {
 
     extensionPort.onMessage.addListener(messageListener);
     extensionPort.onDisconnect.addListener(resetExtensionStreamAndListeners);
+    initializeUiWithTab(activeTab);
```
brad-decker, line 248: added because before we were waiting for the background connection to be active before starting the UI, now we load it without waiting.

ui/index.js
```diff
@@ -29,6 +29,8 @@ import { _setBackgroundConnection } from './store/action-queue';
 
 log.setLevel(global.METAMASK_DEBUG ? 'debug' : 'warn');
 
+const STORE_KEY = 'flatmap-v3';
```
brad-decker, line 32: This is irrelevant to the test, i was just fooling around with storage and wanted a key to increment for reading and setting to see what happened.

ui/index.js
```diff
@@ -41,6 +43,9 @@ export const updateBackgroundConnection = (backgroundConnection) => {
   backgroundConnection.onNotification((data) => {
     if (data.method === 'sendUpdate') {
       reduxStore.dispatch(actions.updateMetamaskState(data.params[0]));
+      browser.storage.local.set({ [STORE_KEY]: data.params[0] }).then(() => {
```
brad-decker, line 46: when we get an update to state from the bg we save the snapshot to the storage. In addition to what normally happens.

ui/index.js
```diff
@@ -52,18 +57,39 @@ export const updateBackgroundConnection = (backgroundConnection) => {
...
-      cb(err, metamaskState);
-      return;
-    }
-    startApp(metamaskState, backgroundConnection, opts).then((store) => {
-      setupDebuggingHelpers(store);
-      cb(null, store);
+  browser.storage.local
+    .get(null)
+    .then((state) => {
+      if (!state[STORE_KEY]) {
```
brad-decker, line 65: If we don't have state, throw an error so we get to the catch.

ui/index.js
```diff
@@ -52,18 +57,39 @@ export const updateBackgroundConnection = (backgroundConnection) => {
...
+        throw new Error('No state yet');
+      }
+      console.log(state);
+      startApp(state[STORE_KEY], backgroundConnection, opts).then((store) => {
```
brad-decker, line 69: If we do have state we can start the app!

ui/index.js
```diff
@@ -52,18 +57,39 @@ export const updateBackgroundConnection = (backgroundConnection) => {
...
+        setupDebuggingHelpers(store);
+        cb(null, store);
+      });
+    })
+    .catch((error) => {
+      console.log('in the catch');
+      console.error(error);
+      // check if we are unlocked first
+      backgroundConnection.getState(function (err, metamaskState) {
```
brad-decker, line 78: If we don't have the state in storage we fall back to asking for the background connection. The reason why this isn't a viable solution (caching locally instead of making the storage the DEFAULT) is because we have to have a copy of state in order to reliably start up without the background connection. If that key is missing we need the background connection, but without waiting for the service worker we do not have a reference to the streams to be able to request state and therefore this will fail and crash the application. We could check in the app/scripts/ui.js file if we have state locally prior to booting up the application, but a better solution is to make the local storage the default way we store and access state. Updates are written to local storage from the background and the UI just reads it. We can even subscribe to changes to this store in the UI instead of relying upon the message from the service worker that state has updated.

PR #17683, opened 2023-02-09
Add desktop support

matthewwalsh0:
## Explanation

Adds the background code needed to support connecting to the [MetaMask Desktop](https://github.com/MetaMask/metamask-desktop) app.

The majority of the code is abstracted in the `@metamask/desktop` NPM package meaning the core changes are small additions to the background script and MetaMask controller.

In total this PR includes:

- Use of the `DesktopManager` in the background script to redirect internal and external connections to the desktop app.
- Inclusion of the `DesktopController` in the MetaMask controller.
- Support for the desktop keyrings in the MetaMask controller via the overrides object.
- Creation of a middleware handler to support using the `metamask_enableDesktop` method to connect to the desktop app while the UI code is pending.
- Minor structural changes to the Sentry setup to allow being called from the desktop app.
- Build system support for desktop specific configuration variables.
- Minor E2E test structural additions to support running E2E tests from the desktop repository.

## Manual Testing Steps

1. Start the desktop app.
2. Build the extension: `yarn start --build-type desktop`
3. Open the test DApp in a browser.
4. Open the browser console.
5. `await window.ethereum.request({method: 'metamask_enableDesktop'})`
6. Enter the returned `otp` value in the desktop app.

## Pre-merge author checklist

- [x] I've clearly explained:


OGPoyraz (dismissed):

app/scripts/background.js
```diff
@@ -482,6 +504,26 @@ export function setupController(initState, initLangCode, overrides) {
...
+      DesktopManager.isDesktopEnabled() &&
+      OVERRIDE_ORIGIN.DESKTOP !== overrides?.getOrigin?.()
+    ) {
+      DesktopManager.createStream(remotePort, CONNECTION_TYPE_INTERNAL).then(
+        () => {
+          // When in Desktop Mode the responsibility to send CONNECTION_READY is on the desktop app side
+          if (isManifestV3) {
+            // Message below if captured by UI code in app/scripts/ui.js which will trigger UI initialisation
+            // This ensures that UI is initialised only after background is ready
+            // It fixes the issue of blank screen coming when extension is loaded, the issue is very frequent in MV3
```
danjm, line 518: "It fixes the issue of blank screen coming when extension is loaded, the issue is very frequent in MV3"

And so I assume that the issue does not happen in MV2, is that correct?

app/scripts/lib/setupSentry.js
```diff
@@ -140,23 +141,7 @@ export default function setupSentry({ release, getState }) {
     ],
     release,
     beforeSend: (report) => rewriteReport(report, getState),
-    beforeBreadcrumb(breadcrumb) {
```
danjm, line 143: Are there any desktop specific reasons that this `beforeBreadcrumb` method was changed?

danjm (changes requested):
Overall the code looks good and mostly makes sense. I have asked some questions.

One general question: what is the motivation for the code fencing? Is it just to reduce risk of desktop code causing regressions? Or is it only meant to be in place until all desktop code has been merged?

app/scripts/background.js
jpuri: Yes this post message is not needed in MV2 and thus may not be needed in desktop also.

app/scripts/lib/setupSentry.js
matthewwalsh0: This is exported just to avoid duplicating the same code in the app where we also need to setup Sentry.

https://github.com/MetaMask/metamask-desktop/blob/main/packages/app/src/app/globals.ts#L130

.yarn/patches/eth-trezor-keyring-npm-0.10.0-1bf9aa3f4d.patch
```diff
@@ -0,0 +1,49 @@
+diff --git a/index.js b/index.js
```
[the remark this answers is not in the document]
matthewwalsh0: Good point, we used a patch initially simply for quick testing and to save some time.

Our choices now are:

- Putting these changes into `eth-trezor-keyring` which would require a PR, reviews, and a new NPM release.
- Duplicating the entire keyring with these changes in the app resulting in no extension changes.

I think we'd prefer the latter at this stage if there's no objections.

matthewwalsh0:
@danjm - Concerning the code fencing, the primary reason is indeed to avoid any regressions in the main build. We did have plans to remove the `desktop` build type in future when the code proves to be stable, but even then we anticipated the desktop code would still be fenced using the `flask` build type given our support for snaps and its experimental nature.

danjm (approved):
LGMT!

cryptotavares (approved):

Issue #14847, opened 2022-06-03
MV3: fix issue with trezor hardware wallet connection wallet 


jpuri:
Trezor wallet connection is broken apparently because of its use of DOM api:

![Screenshot 2022-06-03 at 6 15 46 PM](https://user-images.githubusercontent.com/2182307/171856635-31a6edc9-daa3-4240-875e-b2b30e082d31.png)

An option to fix it could be using content script or move functionality to UI.

Related issue: https://github.com/MetaMask/metamask-extension/issues/14183

PR #17008, opened 2022-12-20
[FLASK] Implement offscreen execution service for running Snaps in MV3

Mrtenz:
## Explanation

This implements the offscreen execution service in order to run Snaps in MV3. It uses the offscreen document API to create a new window, where the snaps are run.

## Manual Testing Steps

- Run `yarn start:mv3 --build-type=flask`.
- Install the extension in Google Chrome Canary.
- Run a snap using [test-snaps](https://metamask.github.io/test-snaps/latest).

## Pre-merge author checklist

- [x] I've clearly explained:
  - [x] What problem this PR is solving
  - [x] How this problem was solved
  - [x] How reviewers can test my changes
- [x] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [ ] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
- [ ] PR is linked to the appropriate GitHub issue
- [ ] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

PR #17687, opened 2023-02-09
mv3 HW support using offscreen

bergarces:
## Explanation

This PR aims to implement support for hardware wallets on MV3 using Chrome's Offscreen API.

### Snaps
The idea is that hardware wallets are dealt with within an iframe inside the offscreen page. That is to ensure there's no conflict with snaps use of the offscreen api. As shown in this [PR](https://github.com/MetaMask/metamask-extension/pull/17008/files#diff-49ea8eb467193d9b67631ea4e9b10d97c1a721415777a16731d1006d64d04f0d), in order to enable the snaps offscreen execution environment, an `offscreen.html` along with a javascript bundle is imported. Since this PR will be creating the `offscreen.html`, that'll have to change so that only the JavaScript bundles are imported.

### Keyrings
Before this PR can be merged, the two associated PRs to split keyring and bridge logic from [trezor](https://github.com/MetaMask/eth-trezor-keyring/pull/143) and [ledger](https://github.com/MetaMask/eth-ledger-bridge-keyring/pull/156) keyrings need to be merged.

Those PRs were motivated by the issues faced implementing the MetaMask Desktop companion app, but they are useful in this case as well, since the SDKs rely on DOM calls that are not present within MV3 service-workers.

### Lavamoat
The iframe code is currently bundled separately from the rest and it's not being bundled with lavamoat. Currently, that code only uses one dependency (`@trezor/connect-web`), although it should be possible to use no dependency at all if we import the code directly in the iframe from their page (https://connect.trezor.io/9/trezor-connect.js).

### Trezor
The Trezor case is the simplest. A keyring bridge that forwards a message to the offscren iframe, which uses the TrezorConnect SDK the same way that it is used in the existing MV2 bridge. The SDK needs to be initialised with `env: 'web'` to prevent the use of extension APIs that are not supported in offscreen pages.

### Ledger
The ledger case uses the existing live iframe (https://github.com/MetaMask/eth-ledger-bridge-keyring/blob/gh-pages/ledger-bridge.js). The offscreen ledger iframe just forwards the calls there and awaits a response.

### Lattice
For the Lattice case, it was enough to overwrite the `_getCreds` method of the existing keyring. This solution, however, doesn't take into consideration that the sdk session is stored in the service-worker and, therefore, is bound to be cleared once it becomes inactive. A more robust solution should be possible by splitting the Lattice keyring code in the same fashion as Ledger and Lattice, so that functionality can be moved to the offscreen iframe if needed.

### Splitting PR
It should be possible to split this PR in smaller ones for each hardware wallet, which should be easier to review.

## Manual Testing Steps

For each hardware wallet (trezor, ledger and lattice), the following actions need to be tested for both MV2 and MV3 versions of the extension:
- Import account from hardware wallet
- Send transaction
- Sign message
- Sign personal message
- Sign typed data v4

## Pre-merge author checklist

- [X] I've clearly explained:

### DIFF
--- a/app/manifest/v3/_base.json
+++ b/app/manifest/v3/_base.json
@@ -73,7 +73,8 @@
     "scripting",
     "storage",
     "unlimitedStorage",
-    "webRequest"
+    "webRequest",
+    "offscreen"
   ],
   "short_name": "__MSG_appName__"
 }
--- a/app/manifest/v3/chrome.json
+++ b/app/manifest/v3/chrome.json
@@ -1,6 +1,6 @@
 {
   "content_security_policy": {
-    "extension_pages": "script-src 'self'; object-src 'self'; frame-ancestors 'none';"
+    "extension_pages": "script-src 'self'; object-src 'self'; frame-ancestors 'self';"
   },
   "externally_connectable": {
     "matches": ["https://metamask.io/*"],
--- a/app/offscreen/lattice/lattice-iframe.html
+++ b/app/offscreen/lattice/lattice-iframe.html
@@ -1,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8" />
+    <title>Lattice OffScreen iframe</title>
+  </head>
+  <body>
+    <script type="text/javascript" src="./lattice-iframe.js"></script>
+  </body>
+</html>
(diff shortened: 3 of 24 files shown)


danjm:
Fixes https://github.com/MetaMask/metamask-extension/issues/14847

PR #17688, opened 2023-02-09
OpenSea security provider metrics

### LINKED ISSUE #16262: Wallet Tx Security Provider: add a new user property and a ui_customizations property
## Background
[Epic on Notion](https://www.notion.so/Enable-MVP-support-for-tx-security-providers-954fef15373d4736b046566d20c93de1)
We want to update our metric scheme to capture the usage and efficacy of the new transactions security provider feature so we're able confidently iterate over and improve this feature.
As always, for anyone looking for more info on MetaMetrics, [here](https://medium.com/metamask/metametrics-a-refresher-on-metamasks-analytics-system-78d725cb9a75) is a good article.

## Tasks

1. Create a new user profile property (more info [here](https://help.mixpanel.com/hc/en-us/articles/115004501966#user-profile-properties)) called `security_provider`. This property should be a type list property (more info [here](https://help.mixpanel.com/hc/en-us/articles/115004547063-Properties-Supported-Data-Types#list)) and receive the value `opensea` if the user has enabled the transactions security provider feature under Settings > Experimental. Related ticket here https://github.com/MetaMask/metamask-extension/issues/16257.

2. If the user turns off the transactions security provider feature under Settings > Experimental, the user profile property `security_provider` should be updated to remove the `opensea` value from it.

3. Add a new property “ui_customizations” to `Transaction Approved`, `Transaction Rejected`, `Signature Approved` and `Signature Rejected` events.
This property should be a type list property (more info [here](https://help.mixpanel.com/hc/en-us/articles/115004547063-Properties-Supported-Data-Types#list)) and receive the value `flagged_as_malicious` when the transaction was flagged as malicious by the transaction security provider. Related ticket here https://github.com/MetaMask/metamask-extension/issues/16258 .

## Acceptance Criteria
1. This new properties are only collect for users that have opted in MetaMetrics.

### LINKED ISSUE #18150: Add metrics for transactions and signatures flagged as unknown for the Tx Security Provider feature
Add `ui_customizations` property with value `flagged_as_unknown_security` whenever we display the yellow warning (Security provider API response is `2`).
This is a follow up/extension of https://github.com/MetaMask/metamask-extension/pull/17688 and should follow the same spec as that one.

filipsekulic:
## Explanation
* Fixes #16262 
* Fixes #18150

Added metrics for the `OpenSea` security provider. 

Created a new user profile property called `security_provider`. This property is a type list and receives the value `opensea` if the user has enabled the `OpenSea` security provider feature under `Settings > Experimental`.
If the user turns off the security provider under `Settings > Experimental`, then the user profile property `security_provider` is updated and the `opensea` value removed from it.

Additionally there is a new property called `ui_customizations` added to `Transaction Approved`, `Transaction Rejected`, `Signature Approved` and `Signature Rejected` events.
This property receives one of the following values: 

1. `['flagged_as_malicious']` when the transaction was flagged as malicious
2. `['flagged_as_safety_unknown']` when the transaction was flagged as safety unknown
3. `null` in all other cases 

by the transaction security provider.

## Manual testing steps
In the `.metamaskrc` set the transaction security feature flag to 1 - `TRANSACTION_SECURITY_PROVIDER=1`.

Follow the procedure for testing the metrics event and check the response in the background console of the extension - `background.html`.

Cases to check:
- enable/disable the `OpenSea` security provider under `Settings > Experimental`
- `SEND LEGACY TRANSACTION`
- `SEND EIP 1559 TRANSACTION`
- `ETH SIGN`
- `PERSONAL SIGN`
- `SIGN TYPED DATA`
- `SIGN TYPED DATA V3`
- `SIGN TYPED DATA V4`


jpuri:
Changes look good to me, it will be nice to have a review by @brad-decker

jpuri (dismissed):

app/scripts/lib/message-manager.js
```diff
@@ -231,8 +246,12 @@ export default class MessageManager extends EventEmitter {
         event: reason,
         category: EVENT.CATEGORIES.TRANSACTIONS,
         properties: {
-          action: 'Sign Request',
+          action: 'Sign Request Reject',
           type: msg.type,
+          ui_customizations:
+            msg.securityProviderResponse?.flagAsDangerous === 1
+              ? ['flagged_as_malicious']
+              : [],
```
brad-decker, line 254: I believe the ui_customization should be added to the events in createRPCMethodTrackingMiddleware versus adding a new event and renaming events here and in the other message managers.

brad-decker (changes requested):

bschorchit:
A request from data team: we should use `security_providers` (plural) instead of `security_provider` (singular) for the name of the user profile property. Could you make this small change, @filipsekulic ? Thank you!

app/scripts/lib/message-manager.js
filipsekulic: Hey @brad-decker, 
Do you have a suggestion how it should look like? I mean, in the middleware you mentioned I need the information I get from the message - `msg.securityProviderResponse?.flagAsDangerous`. 
Also, I have to mention the presence of the property that shares the same name as the one mentioned above (https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/lib/createRPCMethodTrackingMiddleware.js#L173).

brad-decker: You would add your value to the array already present. The events in these confirmations are not the ones we rely upon. Let me dig in a bit and see how you can get access to the value you need

filipsekulic: @brad-decker I tried it, but without success. The reason is that the data being passed to the `securityProviderRequest` from `createRPCMethodTrackingMiddleware` (`req`) is not proper, so as the response in that case. The proper data being sent is shown in the attached image (`personal-message-manager.js`).

Data

brad-decker: @filipsekulic look at the detectSIWE implementation. Its done in both places as well. You will need to formulate the data object that securityProviderCheck needs -- note that in perosnal-message-manager that includes the 'id' of the transaction but the id key isn't used.

filipsekulic: @brad-decker I implemented the solution you suggested. Thanks! However, I could not fix the failing test - `createRPCMethodTrackingMiddleware.test.js`. I must be missing something. Hope you can help me. I would appreciate your help!

seaona:
Some issues from QA:
- When I load MM for the first time, I can see a background console error related to `detectSIWE` , saying `TypeError: The first argument must be one of type string, Buffer, ArrayBuffer, Array, or Array-like Object. Received type undefined`

image

- *_Sometimes_* the event for Send/Send Legacy is not emitted. 

https://user-images.githubusercontent.com/54408225/222410134-e357bf3f-1456-4e49-a38d-d72169ef0b97.mp4


- I haven't observed any metrics event related to the provider, when interacting with Signatures (any kind of signature)

https://user-images.githubusercontent.com/54408225/222410291-d0e404ba-9e9d-4a2f-84eb-419cd709baee.mp4

app/scripts/lib/personal-message-manager.js
```diff
@@ -217,6 +217,17 @@ export default class PersonalMessageManager extends EventEmitter {
...
+    const msg = this.getMsg(msgParams.metamaskId);
+
+    this.metricsEvent({
+      event: 'Sign Request',
+      category: EVENT.CATEGORIES.TRANSACTIONS,
+      properties: {
+        action: 'Sign Request Approve',
+        type: msg.type,
+      },
+    });
```
brad-decker, line 229: This code needs to be removed, events for approval are already tracked in createRPCMethodTrackingMiddleware

app/scripts/lib/message-manager.js
```diff
@@ -180,6 +180,17 @@ export default class MessageManager extends EventEmitter {
...
+
+    this.metricsEvent({
+      event: 'Sign Request',
+      category: EVENT.CATEGORIES.TRANSACTIONS,
+      properties: {
+        action: 'Sign Request Approve',
+        type: msg.type,
+      },
+    });
+
```
brad-decker, line 193: This code needs to be removed, events for approval are already tracked in createRPCMethodTrackingMiddleware

app/scripts/lib/createRPCMethodTrackingMiddleware.js
```diff
@@ -105,20 +105,51 @@ const rateLimitTimeouts = {};
...
+    const msgData = {
+      msgParams,
+      status: 'unapproved',
+      type: req.method,
+    };
+
+    const securityProviderResponse = await securityProviderRequest(
+      msgData,
+      req.method,
+    );
```
brad-decker, line 151: This code should only be fired if we are dealing with a signature request, which is what @seaona 's QA item is showing (this method is called for *every* request to the RPC handler, not just signatures, note the previous usage of detectSIWE happens ONLY in personal-sign because its the only one that works with detectSIWE.

brad-decker (changes requested):

app/scripts/lib/createRPCMethodTrackingMiddleware.js
```diff
@@ -191,17 +162,44 @@ export default function createRPCMethodTrackingMiddleware({
...
+      let msgParams;
+
       if (event === EVENT_NAMES.SIGNATURE_REQUESTED) {
         properties.signature_type = method;
+
+        const data = req?.params?.[0];
+        const from = req?.params?.[1];
+        const paramsExamplePassword = req?.params?.[2];
+
+        const { isSIWEMessage } = detectSIWE({ data });
```
brad-decker, line 174: I don't think we need this call here, it should only work with personal-sign. Why do we need the isSIWEMessage?

brad-decker (changes requested):

app/scripts/lib/createRPCMethodTrackingMiddleware.js
filipsekulic: Oh, I thought it's not just for the personal-sign. I'll fix it now.

filipsekulic: @brad-decker
You were right, there is no need for the `isSIWEMessage`... I don't know why I put it, thought it might was mandatory for all signatures, because at that moment I was looking into the request data being sent for the personal-sign. Suppose that was the case.
Thanks once again! Hope it's okay now.

brad-decker (dismissed):

jpuri (dismissed):
👍

app/scripts/lib/createRPCMethodTrackingMiddleware.js
```diff
@@ -160,8 +162,37 @@ export default function createRPCMethodTrackingMiddleware({
...
+        const msgData = {
+          msgParams,
+          status: 'unapproved',
+          type: req.method,
+        };
+
+        const securityProviderResponse = await securityProviderRequest(
+          msgData,
+          req.method,
+        );
```
digiwand, line 190: `securityProviderRequest` may throw an error from metamask-controller.js which could break the flow here. I think we should catch errors here to not allow errors from `securityProviderCheck` to stop the dapp request. 

Current error flow that may occur:
- we call `securityProviderRequest` here in createRPCMethodTrackingMiddleware.js
- `securityProviderRequest` (metamask-controller.js)  calls `securityProviderCheck` to fetch validation
- error occurs when calling `securityProviderCheck`
- we catch the error in metamask-controller.js and then also throw the error again
- error breaks createRPCMethodTrackingMiddleware.js and stops dapp request

digiwand (changes requested):
if the fetch request from `securityProviderCheck` somehow fails, it will stop the dapp request. We will need to handle the error here

app/scripts/lib/createRPCMethodTrackingMiddleware.js
digiwand: hey @filipsekulic, I went ahead and merged a solution for the above concern because it seems we're aiming for v10.27 and I was pretty close to a solution after debugging/reviewing. does this look ok to you? [4288143](https://github.com/MetaMask/metamask-extension/pull/17688/commits/428814312a7870ed6c5ede61e4f45370ab28b753)

app/scripts/lib/createRPCMethodTrackingMiddleware.js
```diff
@@ -160,20 +162,54 @@ export default function createRPCMethodTrackingMiddleware({
...
+            if (isSIWEMessage) {
+              properties.ui_customizations.push(
+                METAMETRIC_KEY_OPTIONS[METAMETRIC_KEY.UI_CUSTOMIZATIONS].SIWE,
+              );
+            }
+          }
+        } catch (e) {
+          console.warn(
+            `createRPCMethodTrackingMiddleware: Error calling securityProviderRequest - ${e}`,
+          );
```
digiwand, line 209: choosing `console.warn` here since we already log the error in `securityProviderRequest` (metamask-controller.js). maybe there are better ways to go about this. happy to hear other suggestions

app/scripts/lib/createRPCMethodTrackingMiddleware.test.js
```diff
@@ -6,11 +6,13 @@ import createRPCMethodTrackingMiddleware from './createRPCMethodTrackingMiddlewa
 const trackEvent = jest.fn();
 const metricsState = { participateInMetaMetrics: null };
 const getMetricsState = () => metricsState;
+const securityProviderRequest = jest.fn();
 
 const handler = createRPCMethodTrackingMiddleware({
   trackEvent,
   getMetricsState,
   rateLimitSeconds: 1,
+  securityProviderRequest,
```
digiwand, line 15: I think it would be good to add tests for the new `securityProviderRequest` logic in this file

app/scripts/lib/createRPCMethodTrackingMiddleware.js
digiwand: cc: @brad-decker @jpuri

app/scripts/lib/createRPCMethodTrackingMiddleware.js
```diff
@@ -160,8 +162,37 @@ export default function createRPCMethodTrackingMiddleware({
...
+        };
+
+        const securityProviderResponse = await securityProviderRequest(
+          msgData,
+          req.method,
+        );
+
+        properties.ui_customizations =
+          securityProviderResponse?.flagAsDangerous === 1
+            ? ['flagged_as_malicious']
```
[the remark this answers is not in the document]
filipsekulic: Sounds okay to me.

app/scripts/lib/createRPCMethodTrackingMiddleware.js
filipsekulic: It seems fine to me

filipsekulic:
> Some small issues from QA:
> 
> * For transactions, the `ui_customizations` field is appearing on the `Transaction Approved` and `Transaction Rejected` events. Additionally, the field is also included on `Transaction Added` and `Transaction Submitted` events. It make sense to me, but since it is not exactly what it states in the description, I wanted make sure that this is the expected behaviour.
> 
> > Additionally there is a new property called ui_customizations added to Transaction Approved, Transaction Rejected, Signature Approved and Signature Rejected events.
> 
> * For signatures, the `ui_customizations` field is displayed on the `Signature Requested` event, but not on the `Signature Approved` or `Signature Rejected` events. I think this is not the expected behaviour, according to the description

I fixed it. Also, I would appreciate if @brad-decker could take a look once again and confirm your the issues you mentioned.

app/scripts/controllers/transactions/index.js
```diff
@@ -2312,6 +2313,10 @@ export default class TransactionController extends EventEmitter {
       token_standard: tokenStandard,
       transaction_type: transactionType,
       transaction_speed_up: type === TransactionType.retry,
+      ui_customizations:
+        securityProviderResponse?.flagAsDangerous === 1
+          ? ['flagged_as_malicious']
+          : [],
```
digiwand, line 2319: maybe we want to do something similar to the `transaction_approval_amount_type` property below and only create the key/value if `securityProviderResponse?.flagAsDangerous === 1` is true

app/scripts/lib/createRPCMethodTrackingMiddleware.js
```diff
@@ -160,20 +162,54 @@ export default function createRPCMethodTrackingMiddleware({
...
+        try {
+          const securityProviderResponse = await securityProviderRequest(
+            msgData,
+            req.method,
+          );
+
+          properties.ui_customizations =
+            securityProviderResponse?.flagAsDangerous === 1
+              ? ['flagged_as_malicious']
+              : [];
```
digiwand, line 196: similar to https://github.com/MetaMask/metamask-extension/pull/17688/files#r1132887210, maybe we want to not add `ui_customizations` property if empty

filipsekulic: Fixed on both places

app/scripts/lib/createRPCMethodTrackingMiddleware.js
```diff
@@ -160,20 +162,54 @@ export default function createRPCMethodTrackingMiddleware({
...
+
+          properties.ui_customizations =
+            securityProviderResponse?.flagAsDangerous === 1 && [
+              'flagged_as_malicious',
+            ];
+
+          if (method === MESSAGE_TYPE.PERSONAL_SIGN) {
+            const { isSIWEMessage } = detectSIWE({ data });
+            if (isSIWEMessage) {
+              properties.ui_customizations.push(
```
digiwand, line 201: since on line 193 we have
```          properties.ui_customizations =
            securityProviderResponse?.flagAsDangerous === 1 && [
              'flagged_as_malicious',
            ];
```
it's not guaranteed that `ui_customizations` will be an array. This will cause an error

app/scripts/lib/createRPCMethodTrackingMiddleware.js
```diff
@@ -160,20 +162,54 @@ export default function createRPCMethodTrackingMiddleware({
...
+        try {
+          const securityProviderResponse = await securityProviderRequest(
+            msgData,
+            req.method,
+          );
+
+          properties.ui_customizations =
+            securityProviderResponse?.flagAsDangerous === 1 && [
+              'flagged_as_malicious',
+            ];
```
digiwand, line 196: if `flagAsDanerous !== 1` this will return a boolean 

Screen Shot 2023-03-14 at 3 34 17 PM

jpuri:
@digiwand : has raised some nice points above that should be addressed.

PR looks good otherwise and I can approve once that is done.

app/scripts/controllers/transactions/index.test.js
```diff
@@ -1852,6 +1856,7 @@ describe('Transaction Controller', function () {
             token_standard: TokenStandard.none,
             device_model: 'N/A',
             transaction_speed_up: false,
+            ui_customizations: false,
```
digiwand, line 1859: I am thinking that if we are passing `ui_customizations` we should return null rather than false so that we are not introducing a new data type

filipsekulic:
@digiwand @jpuri @brad-decker 
FYI: I have added a solution for [this issue](https://github.com/MetaMask/metamask-extension/issues/18150) in this PR

app/scripts/controllers/transactions/index.test.js
```diff
@@ -1740,6 +1740,9 @@ describe('Transaction Controller', function () {
             gas: '0x7b0d',
             gasPrice: '0x77359400',
           },
+          securityProviderResponse: {
+            flagAsDangerous: '0',
```
digiwand, line 1744: should we update this to a number type, rather than a string type, to match with `app/scripts/controllers/transactions/index.js`?

same comment for below instances

digiwand (dismissed):
left one small comment and I think the rest seems good to merge

app/scripts/controllers/transactions/index.test.js
filipsekulic: Good catch! 
I fixed it.

app/scripts/lib/createRPCMethodTrackingMiddleware.js
```diff
@@ -162,20 +164,63 @@ export default function createRPCMethodTrackingMiddleware({
...
+
+          if (method === MESSAGE_TYPE.PERSONAL_SIGN) {
+            const { isSIWEMessage } = detectSIWE({ data });
+            if (isSIWEMessage) {
+              properties.ui_customizations === null
+                ? (properties.ui_customizations = [
+                    METAMETRIC_KEY_OPTIONS[METAMETRIC_KEY.UI_CUSTOMIZATIONS]
+                      .SIWE,
+                  ])
+                : properties.ui_customizations.push([
```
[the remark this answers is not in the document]
filipsekulic: Good catch @seaona! 👏 Thanks!
Fixed it.

brad-decker (approved):

jpuri (approved):

app/scripts/lib/createRPCMethodTrackingMiddleware.js
```diff
@@ -162,20 +164,63 @@ export default function createRPCMethodTrackingMiddleware({
...
-        const { isSIWEMessage } = detectSIWE({ data });
-        if (isSIWEMessage) {
-          properties.ui_customizations = [
-            METAMETRIC_KEY_OPTIONS[METAMETRIC_KEY.UI_CUSTOMIZATIONS].SIWE,
-          ];
+        const from = req?.params?.[1];
+        const paramsExamplePassword = req?.params?.[2];
+
+        msgParams = {
+          ...paramsExamplePassword,
```
digiwand, line 177: upon a second look today, I realized this is an issue. It can splay a string. Not sure how we missed this the first time

digiwand: working with another engineer to apply a fix for this

PR #17700, opened 2023-02-09
Organize mocha tests

adonesky1:
Moves all the remaining test suites that run on mocha (still to be converted to jest) into a single directory and targets that directory in the `run-unit-tests` script


mcmire:
For context, @adonesky1 and I found that there are some Jest tests that are also being run by Mocha. Moving the Mocha tests to a separate directory should allow us to give the Mocha runner a glob that excludes the Jest tests and give the Jest runner a glob that excludes the Mocha tests more easily.

brad-decker:
@tmashuang heads up!

brad-decker (approved):

adonesky1:
closing because this missed its window and may be handled elsewhere

PR #17701, opened 2023-02-09
Disable typechecking while running Mocha tests

mcmire:
## Explanation

The Mocha tests load `ts-node` so that if a TypeScript file happens to
be encountered while running tests, it does not throw an error.
Unfortunately, `ts-node` does not act completely like `tsc`; namely, [it
does not eager-load files][1]. This means that our backfilled type
definitions in `types/` are not getting read, which will cause issues
when loading TypeScript files that import typeless packages.

To fix this, we disable typechecking in the Mocha setup file. Now
`ts-node` can parse TypeScript files but will skip over type-specific
syntax. There is a way we avoid disabling typechecking by using the
`files: true` option instead of `transpileOnly: true`, but this causes
the `yarn test:coverage:mocha` command (which is what I used to test
this) to take twice as long (~34s as opposed to ~17s). I figure this is
fine for now, anyway, because eventually all of our tests will be in
Jest.

[1]: https://github.com/TypeStrong/ts-node#missing-types

## Manual Testing Steps

* Go to `test/setup.js`. Comment out `require('ts-node').register({ transpileOnly: true });` and replace it with `require('ts-node/register');`.
* Rename `app/scripts/controllers/permissions/background-api.js` to have a `.ts` extension.
* Fix the errors that TypeScript gives you: fill in missing types with `any` (they should be relegated to just this file) and then add `types/nanoid.d.ts` with the content `declare module 'nanoid';`. 
* Now run `yarn test:coverage:mocha`. It should fail, telling you that it can't find type definitions for `nanoid` (even though you added them).
* Go to `test/setup.js` again and revert the changes you made above.
* Run `yarn test:coverage:mocha` again. The tests should pass.

## Pre-merge author checklist

- [x] I've clearly explained:


adonesky1 (approved):
LGTM! Thanks for helping me figure this out!

brad-decker (approved):

PR #17718, opened 2023-02-10
[FLASK] BREAKING - snaps-monorepo@0.30.0

### LINKED ISSUE #17091: [FLASK] Migrate snap permissions from wildcards to caveats
Please refer to parent issue [#297](https://github.com/MetaMask/MetaMask-planning/issues/297)

We currently have the following format for snap permissions:

```json
"wallet_snap_npm:@chainsafe/filsnap": {
    "caveats": null,
    "date": 1654882874322,
    "id": "Y34wk3O654MSvS1hOwXs7",
    "invoker": "https://filsnap.chainsafe.io",
    "parentCapability": "wallet_snap_npm:@chainsafe/filsnap"
}
```

After the rest of the issues are addressed in the parent issue, a migration needs to be created to reformat into the following format:

```json
"wallet_snap": {
    "caveats": [{
      "type": "snapIds",
      "value": {
        "npm:@chainsafe/filsnap": {
          "version": "1.0.0"
        }
      }
    }],
    "date": 1654882874322,
    "id": "Y34wk3O654MSvS1hOwXs7",
    "invoker": "https://filsnap.chainsafe.io",
    "parentCapability": "wallet_snap"
}
```

hmalik88:
Closes #17091 

- `wallet_snap_*` permission was removed in favor of `wallet_snap` and the use of caveats to track dapp <> snap permissions, this PR makes changes to account for the new permission and adds a migration to migrate from the old permissions.
- Changes made to account for the new `webassembly` endowment & disablement of `eth_accounts`.
- Snaps packages updated to latest release.
  - `snap_confirm` rpc method was removed.

Please see parent issue https://github.com/MetaMask/MetaMask-planning/issues/297.


GuillaumeRx:
+10 rep. for the meme

app/scripts/migrations/081.js
```diff
@@ -0,0 +1,68 @@
...
+    return versionedData;
+  },
+};
+
+function transformState(state) {
+  const PermissionController = state?.PermissionController || {};
+
+  const { subjects } = PermissionController;
+
+  const snapPrefix = 'wallet_snap_';
```
FrederikBolding, line 31: Can we detect snaps earlier an bail out of this entire thing if we don't see any snaps permissions? Maybe, we should just bail if the build type is not Flask?

app/scripts/migrations/081.js
```diff
@@ -0,0 +1,68 @@
...
+
+  for (const [subjectName, subject] of Object.entries(subjects)) {
+    // we keep track of the latest permission's date and associated id
+    // to assign to the wallet_snap permission
+    let date = 1;
+    let id;
+    const { permissions } = subject;
+    const updatedPermissions = { ...permissions };
+    for (const [permissionName, permission] of Object.entries(permissions)) {
```
FrederikBolding, line 40: I think this migration could be written in a way that makes it more obvious what is going on.

I would prefer if we did something along the lines of:
1. Find all snaps and turn them into caveats on one object
2. Delete old snap keys from the subject permissions
3. Add permission object from 1

FrederikBolding: I think we should also get a second opinion from someone who is experienced with the migrations since this is a more complex one.

ui/components/app/permission-page-container/permission-page-container.component.js
```diff
@@ -54,11 +70,50 @@ export default class PermissionPageContainer extends Component {
...
+        return acc;
+      }
+      ///: END:ONLY_INCLUDE_IN
       acc[methodName] = true;
       return acc;
     }, {});
   }
 
+  ///: BEGIN:ONLY_INCLUDE_IN(flask)
+  getDedupedSnapPermissions() {
```
FrederikBolding, line 85: I'm not sure I understand why we need to do this. Can you explain?

lavamoat/build-system/policy-override.json
```diff
@@ -32,6 +32,11 @@
         "process.versions": true
       }
     },
+    "@metamask/rpc-methods": {
```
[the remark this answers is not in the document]
hmalik88: Not sure but lavamoat complains if I don't

ui/pages/settings/flask/view-snap/view-snap.js
```diff
@@ -73,6 +74,30 @@ function ViewSnap() {
...
 
+  const onDisconnect = (connectedOrigin, snapId) => {
+    const caveatValue =
+      subjects[connectedOrigin].permissions[WALLET_SNAP_PERMISSION_KEY]
+        .caveats[0].value;
+    const newCaveatValue = { ...caveatValue };
+    delete newCaveatValue[snapId];
+    if (Object.keys(newCaveatValue) > 0) {
+      dispatch(
+        updateCaveat(
```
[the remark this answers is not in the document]
hmalik88: Yeah, I agree. I'll make an issue.

app/scripts/migrations/081.js
hmalik88: Yeah, true.

ui/components/app/permission-page-container/permission-page-container.component.js
hmalik88: Ok, so if a dapp (origin) already has permissions for a snap and is requesting another snap, the second request would also list the first permission the dapp already had and was not requesting, because everything is now under the one permission, `wallet_snap`.

hmalik88: i.e. caveat values are merged so the request has all the current snaps + requested snaps, you have to diff to get the actual snaps being requested

app/scripts/migrations/081.js
```diff
@@ -0,0 +1,68 @@
...
+};
+
+function transformState(state) {
+  const PermissionController = state?.PermissionController || {};
+
+  const { subjects } = PermissionController;
+
+  const snapPrefix = 'wallet_snap_';
+
+  for (const [subjectName, subject] of Object.entries(subjects)) {
```
Gudahtt, line 33: We're assuming here that `subjects` is an object, which _should_ be true but generally we try to avoid making any assumptions in these migrations just in case the state was corrupted somehow. The migrations should return the state unchanged if we find it violates our assumptions in some way, so that we can properly fix it in a follow-up migration if we see that happen in prod.

Migrating this to TypeScript would make this much easier to do in practice. Our new migration TypeScript template uses the `unknown` type to ensure we don't forget to check our assumptions about state shape.

ui/pages/settings/flask/view-snap/view-snap.js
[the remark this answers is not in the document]
hmalik88: https://github.com/MetaMask/snaps-monorepo/issues/1258

app/scripts/migrations/081.js
hmalik88: Breaking up things to make it readable will likely introduce extra iteration. This is not part of the codebase that will see a lot of traffic, so I'm fine keeping it this way, I did however add extensive comments.

app/scripts/migrations/081.js
hmalik88: On second thought, introducing `isFlask` adds some inconsistency with tests, i.e. the tests will fail on non-flask builds. Alternatively, the cost to grab all the permission keys at once and check for snap permissions is not really much different than current implementation. I'm going to keep implementation as is.

app/scripts/migrations/081.ts
```diff
@@ -0,0 +1,162 @@
+import { cloneDeep, isArray } from 'lodash';
+import {
+  WALLET_SNAP_PERMISSION_KEY,
+  SnapCaveatType,
+} from '@metamask/rpc-methods';
```
FrederikBolding, line 5: This brings in snaps dependencies to the beta and main extension. We should be sure that that is fine.

FrederikBolding: If not, we could hardcode the values for now. Potentially they should be, because they should reflect the CaveatType and key when this migration was applied?

app/scripts/migrations/081.ts
```diff
@@ -0,0 +1,162 @@
...
+    const updatedPermissions = { ...permissions };
+    for (const [permissionName, permission] of Object.entries(permissions)) {
+      // check if the permission is namespaced
+      if (permissionName.startsWith(snapPrefix)) {
+        if (
+          !isObject(permission) ||
+          !hasProperty(permission, 'id') ||
+          !hasProperty(permission, 'date')
+        ) {
+          return state;
```
FrederikBolding, line 70: Is the idea that if we see anything out of the ordinary we return the state and hopefully nothing is broken?

app/scripts/migrations/081.ts
```diff
@@ -0,0 +1,162 @@
...
+  const state = versionedData.data;
+  const newState = transformState(state);
+  versionedData.data = newState;
+  return versionedData;
+}
+
+// We return state AS IS if there is any corruption
+function transformState(state: Record<string, unknown>) {
+  if (
+    !hasProperty(state, 'PermissionController') ||
```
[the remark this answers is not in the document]
hmalik88: Yeah, that sounds like a good alternative.

app/scripts/migrations/081.ts
hmalik88: Yeah, basically doing what Mark said: https://github.com/MetaMask/metamask-extension/pull/17718#discussion_r1122131559

app/scripts/migrations/081.ts
hmalik88: > If not, we could hardcode the values for now. Potentially they should be, because they should reflect the CaveatType and key when this migration was applied?

Yes, it should reflect the extension code at the time. Aren't the packages brought in regardless? Only thing we do in beta, main is fencing no?

app/scripts/migrations/081.ts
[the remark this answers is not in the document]
hmalik88: Actually wait, lol. This still breaks the test on non-flask builds.

app/scripts/migrations/081.ts
FrederikBolding: The packages are only brought in to the build if they are required in a file. Could we swap it out for hardcoded strings instead then? So it matches the current version just in case we change them in the future

hmalik88: okay cool, ill hardcode.

FrederikBolding (dismissed):

bowensanders:
LGTM - but we might want to re-enable the rpc test now that it will work again :)

app/_locales/fr/messages.json
```diff
@@ -2713,15 +2713,15 @@
...
+    "description": "The description for the `wallet_snap` permission. $1 is the human-readable name of the Snap."
   },
   "permission_accessNetwork": {
     "message": "Accéder à internet.",
     "description": "The description of the `endowment:network-access` permission."
   },
   "permission_accessSnap": {
     "message": "Connexion au Snap $1.",
-    "description": "The description for the `wallet_snap_*` permission. $1 is the name of the snap."
+    "description": "The description for the `wallet_snap` permission. $1 is the name of the Snap."
```
FrederikBolding, line 2724: All of this casing was fixed in this PR: https://github.com/MetaMask/metamask-extension/pull/17915

Looks like you undid that? I think we should revert that.

app/_locales/en/messages.json
```diff
@@ -2731,15 +2731,15 @@
...
+    "description": "The description for the `wallet_snap` permission. $1 is the human-readable name of the snap."
   },
   "permission_accessNetwork": {
     "message": "Access the internet.",
     "description": "The description of the `endowment:network-access` permission."
   },
   "permission_accessSnap": {
-    "message": "Connect to the $1 snap.",
-    "description": "The description for the `wallet_snap_*` permission. $1 is the name of the snap."
+    "message": "Connect to the $1 Snap.",
```
FrederikBolding, line 2741:
```suggestion
    "message": "Connect to the $1 snap.",
```

app/_locales/ko/messages.json
```diff
@@ -2713,15 +2713,15 @@
...
+    "description": "The description for the `wallet_snap` permission. $1 is the human-readable name of the snap."
   },
   "permission_accessNetwork": {
     "message": "인터넷에 액세스합니다.",
     "description": "The description of the `endowment:network-access` permission."
   },
   "permission_accessSnap": {
-    "message": "$1 스냅에 연결하세요.",
-    "description": "The description for the `wallet_snap_*` permission. $1 is the name of the snap."
+    "message": "$1 스냅에 연결합니다.",
```
FrederikBolding, line 2723: Revert

FrederikBolding (dismissed):

FrederikBolding (dismissed):

Gudahtt:
I haven't had time to fully review this, but I have confirmed that my previous suggestions have been addressed/resolved. The new migration looks like it comprehensively checks for invalid input :+1:

FrederikBolding (approved):

GuillaumeRx (approved):

PR #17452, opened 2023-01-27
Set popover zIndex to be higher then tooltip

amerkadicE:
## Explanation

The portfolio tooltip had a higher z index than the popover component. It caused the tooltip to be over popover. We use the external `react-tippy` package which has zIndex: 9999 and it can't be changed. 

~~This PR will set zIndex of all popovers on the home page to 10000.~~

This PR will show portfolio tooltip only if there is no popover acitve

* Fixes #17398

### Before

image
image

### After

image

image

## Manual Testing Steps

Scenario 1:

Login
Click the ellipsis menu
Select account details
Scenario 2:

Login
Close the extension
Switch networks
Open the popup

## Pre-merge author checklist

- [x] I've clearly explained:
  - [x] What problem this PR is solving
  - [x] How this problem was solved
  - [x] How reviewers can test my changes
- [ ] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [ ] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
- [ ] PR is linked to the appropriate GitHub issue
- [ ] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.

In this case, a QA Engineer approval will be be required.

PR #17722, opened 2023-02-10
UX: Move Portfolio link to its own button

darkwing:
## Explanation

This is an alternative to https://github.com/MetaMask/metamask-extension/pull/17452 .  The tooltip being used is problematic and the portfolio link could get its proper attention by being alongside the other callout buttons.

## TODO

- [x] Write a migration to remove `portfolio` properties from state.

## Screenshots/Screencaps

PortfolioDropdown
PortFolioButton2

## Pre-merge author checklist

- [ ] I've clearly explained:


app/scripts/migrations/078.js
```diff
@@ -0,0 +1,27 @@
+import { cloneDeep } from 'lodash';
+
+const version = 78;
+
+/**
+ * Prior to token detection v2 the data property in tokensChainsCache was an array,
+ * in v2 we changes that to an object. In this migration we are converting the data as array to object.
+ */
```
brad-decker, line 8: Should this be updated?

app/scripts/migrations/078.test.js
```diff
@@ -0,0 +1,98 @@
...
+      },
+    };
+
+    const newStorage = await migration78.migrate(oldStorage);
+    expect(newStorage.meta).toStrictEqual({
+      version: 78,
+    });
+  });
+
+  it('should remove the "showPortfolioToolip" property', async () => {
```
brad-decker, line 17: Should probably add a test to ensure that in the absence of the showPortfolioTooltip key the state structure is the same.

app/scripts/migrations/078.js
darkwing: Good catch!

jakehaugen:
I would propose an update to the spacing logic so all buttons are evenly spaced (ignoring the label text).

Of course, we'll need to account for translations so need some logic to make sure text does not overlap. 

Not sure exactly what the technical solution would be since I'm measuring from the icon vs. the text (which it seems to be doing currently). 

We could also set the logic to look good in "most" cases, knowing some translations will not allow for even spacing. 

Screenshot 2023-02-14 at 3 43 53 PM

darkwing:
Good shout @jakehaugen !  Inspection showed something interesting:

NewIconButtons

Only the right-most icon is centered.  I'll work on that.

darkwing:
@jakehaugen Updated per your suggestion:

NewPortFolioButtons

digiwand (approved):
lgtm! tested w/ macOS x Chrome

NidhiKJha (approved):

PR #17729, opened 2023-02-10
Squashed warning when running `yarn test:coverage:jest`

HowardBraham:
I noticed this warning when running `yarn test:coverage:jest`
```
jest-haste-map: duplicate manual mock found: webextension-polyfill
  The following files share their name; please delete one of them:
    * <rootDir>/.storybook/__mocks__/webextension-polyfill.js
    * <rootDir>/ui/__mocks__/webextension-polyfill.js
```

I was able to squash the warning by deleting `.storybook/__mocks__/webextension-polyfill.js` and changing `.storybook/main.js` from
```
config.resolve.alias['webextension-polyfill'] = require.resolve(
  './__mocks__/webextension-polyfill.js',
);
```
to
```
config.resolve.alias['webextension-polyfill'] = require.resolve(
  '../ui/__mocks__/webextension-polyfill.js',
);
```

### DIFF
--- a/.storybook/__mocks__/webextension-polyfill.js
+++ b/.storybook/__mocks__/webextension-polyfill.js
@@ -1,8 +1,0 @@
-module.exports = {
-  runtime: {
-    getManifest: () => {
-      return { manifest_version: 2 };
-    }
-  },
-};
-  
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -40,7 +40,7 @@ module.exports = {
       __filename: true,
     };
     config.resolve.alias['webextension-polyfill'] = require.resolve(
-      './__mocks__/webextension-polyfill.js',
+      '../ui/__mocks__/webextension-polyfill.js',
     );
     config.resolve.fallback = {
       child_process: false,


Gudahtt (approved):
LGTM!

.storybook/__mocks__/webextension-polyfill.js
```diff
@@ -1,8 +0,0 @@
-module.exports = {
```
mcmire, line 1: Another thing we could do is move this file from `__mocks__` to just `mocks`. The version of this in `ui` seems to be there for Jest tests, but as this is inside of the Storybook config directory, this file doesn't seem to have anything to do with tests, it's just for Storybook. In fact Jest shouldn't even be trying to read this file, but it sees `__mocks__` directories as special: https://jestjs.io/docs/manual-mocks. So in order to keep Storybook configuration and the rest of the app separate I'm wondering if we should keep this file but just rename where it is.

Gudahtt: While that would be a reasonable solution, I do see the benefit in keeping both implementations together. Both the UI Jest mock and the Storybook mock are intended to represent a _complete_ mock of this API for the UI layer.

Using the same file for both would make sense to me. If we kept them separate, we'd need to sync them up regularly anyway.

mcmire (approved):
LGTM as well.

PR #17748, opened 2023-02-14
Add extension desktop UI

cryptotavares:
## Explanation
Adds UI code required for allowing the Extension to pair with the [MetaMask Desktop](https://github.com/MetaMask/metamask-desktop) app.

This PR includes:
* desktop pairing page -> this is the page where we will render an OTP to be introduced in the Desktop App.
* desktop error pages -> Extension error pages to be displayed when something goes wrong while connected to the Desktop App
* desktop enable button -> button to be displayed within settings->experimental page, that will allow enable/disable pairing the Extension with the Desktop App.
* desktop deep linking component -> allows opening up the Desktop App from the Extension if available and installed.
* adds desktop icon to header component when the Extension is running in dev mode and is paired with the Desktop app.
* register desktop listeners on UI launch -> this allows us to provide proper error handler for the edge cases where a user closes the Desktop App without going through the disable pairing process.
* disables ledger live control on the experimental advanced tab when paired with Desktop App

Besides the functional changes mentioned above, it is worth mentioning that we also extended the test render helper to allow proper testing of routes making use of the `useParam` hook.

### Enable/Disable Button
Screenshot 2023-02-14 at 17 02 49
Screenshot 2023-02-14 at 17 08 21
----------------------------
### Pairing page
Screenshot 2023-02-14 at 17 07 51
----------------------------
### Error pages
Screenshot 2023-02-14 at 17 09 30
Screenshot 2023-02-14 at 17 03 06

## Manual Testing Steps
**Paring page:**
* Build the extension for desktop build type - `yarn start --build-type desktop`
* Navigate to settings->experimental
* Click on enable desktop button

**Error pages:**
* Build the extension for desktop build type - `yarn start --build-type desktop`
* Manually mutate the extension state to desktopEnabled set to true.
* Reload the extension
* Open the extension. It should try connect to the Desktop App. As there's none running, it will throw and display the error page.

## Pre-merge author checklist

- [x] I've clearly explained:


app/scripts/ui.js
```diff
@@ -325,14 +338,21 @@ function initializeUi(activeTab, connectionStream, cb) {
...
+  errorKey,
+  err,
+  metamaskState,
+  backgroundConnection,
+) {
+  const html = await getErrorHtml(errorKey, SUPPORT_LINK, metamaskState, err);
 
   container.innerHTML = html;
 
+  registerDesktopErrorActions(backgroundConnection, browser);
```
matthewwalsh0, line 351: Should we be code fencing the desktop specific UI code also? So functions like this and also moving the desktop components into a `desktop` folder so they'll be excluded from the other build types?

shared/lib/error-utils.js
```diff
@@ -50,6 +54,22 @@ export async function getErrorHtml(errorKey, supportLink, metamaskState) {
   const { currentLocaleMessages, enLocaleMessages } = response;
   const t = getLocaleContext(currentLocaleMessages, enLocaleMessages);
 
+  const isDesktopEnabled = metamaskState?.desktopEnabled === true;
+
+  if (isDesktopEnabled) {
+    let errorType = EXTENSION_ERROR_PAGE_TYPES.CRITICAL_ERROR;
+
+    if (err?.message.includes('No response from RPC')) {
```
matthewwalsh0, line 62: Do we need to use the `t` method here in case it's an alternate language or is this error fixed?

shared/lib/error-utils.js
```diff
@@ -87,3 +107,57 @@ export async function getErrorHtml(errorKey, supportLink, metamaskState) {
     </div>
     `;
 }
+
+function disableDesktop(backgroundConnection) {
+  backgroundConnection.disableDesktopError();
+}
+
+export function downloadDesktopApp() {
+  global.platform.openTab({ url: 'https://metamask.io/' });
```
matthewwalsh0, line 116: This could be a constant in the new `shared/constants/desktop.ts`.

app/scripts/ui.js
cryptotavares: Yes I'll add the desktop code fencing everywhere... but we should not be moving the components to a specific desktop dir, because that would cause issues when we release under flask (as anything within `**/desktop/**` will only be available on desktop build type)

matthewwalsh0:
We can remove the middleware handler too now, it's functional but unnecessary with the UI and it introduces more security risks.

`app/scripts/lib/rpc-method-middleware/handlers/desktop/enable-desktop.js`

shared/lib/error-utils.js
cryptotavares: Not really. This is not a UI error, but rather an error that gets thrown whenever the metaRPCClientFactory fails to get the state. Which when desktop is enabled we assume that it is due to a connection lost.

shared/lib/error-utils.js
cryptotavares: We have an [open task](https://github.com/orgs/MetaMask/projects/37/views/4?filterQuery=link&pane=issue&itemId=20544786) to review all the desktop links. We can tackle it all within that task.

cryptotavares:
> Regarding the error screenshots, I've only ever seen the red cross in the horizontal centre, not sure how you got it on the left?

@matthewwalsh0 Yes you are right. There was a typo when converting from `ALIGN_ITEMS` to the new `AlignItems`. It is fixed now and the cross icon appears centered.
Screenshot 2023-02-21 at 15 29 11

matthewwalsh0 (dismissed):

jpuri (approved):
Great work 👍

OGPoyraz (approved):

matthewwalsh0 (approved):

Issue #17719, opened 2023-02-10
[Bug]: unhelpful error handling at private key import time


abourget:
### Description

When importing a private key, if the length is not right, the error message is unhelpful and looks like this:

```
Cannot read properties of undefined (reading 'startsWith')
```

For posterity, the problem I encountered was that the export of a private key from the Gnosis Safe Android application was _invalid_. It trimmed zeroes at the beginning of the string, making the private key of an invalid length.

Adding a simple validation of length, and alerting that it is not of valid length would have gone a long way to avoid the multi-day hassles I went through.

### Steps to reproduce

1. Get a private key
2. Remove its first character
3. Try to import

### Version, Browser, OS

10.24.2, Chrome, Mac OS

danjm:
@HowardBraham this might be related to some of the private key error messaging improvements you looked into recently

HowardBraham:
@abourget You figured it out!  The screenshot is in this issue #16895 but nobody could figure out steps to reproduce this error.

HowardBraham:
Fixed by #17763

Issue #10113, opened 2020-12-19
Version 8.1.9 freezes while importing keystore


eiabea:
**Describe the bug**
After importing a metamask wallet from a seed and trying to import a keystore json file into this account, metamask freezes until chrome asks to "wait" or "exit" the tab

**Steps to reproduce (REQUIRED)**
Steps to reproduce the behavior, libraries used with version number, and/or any setup information to easily reproduce:

1. Install metamask 8.1.9
2. Import wallet from seed
3. Click on import account
4. Select type "Json"
5. Enter your password
6. Experience a freeze of the metamask tab

**Expected behavior**
Account gets imported

**Browser details (please complete the following information):**
 - OS: Linux
 - Hardware Wallet: No
 - Browser: Version 87.0.4280.66 (Official Build) Arch Linux (64-bit)
 - MetaMask Version: 8.1.9

**Additional context (Error Messages, etc.)**
Downgrading MetaMask to `8.1.5` fixed the issue

tmashuang:
After installing and import, but before getting to the home page, are you accepting participating in the metrics page? Sounds like a  #10097.

eiabea:
> After installing and import, but before getting to the home page, are you accepting participating in the metrics page? Sounds like a #10097.

No i haven't actually

tmashuang:
Where was the json file account created? I've imported a json account created from Mycrypto.

eiabea:
> Where was the json file account created? I've imported a json account created from Mycrypto.

directly in `geth`

shoenseiwaso:
I'm seeing the same issue with MetaMask, Chrome on macOS. MetaMask pegs a CPU core for several minutes until killed when attempting to import a JSON keystore file.

adamsoffer:
Can confirm this on my end as well. Brave on macOS. Running Big Sur on the Apple M1.

miki-bgd-011:
Same issue, Ubuntu 20, Chrome 87.0.4280.88.

davidgamero:
Thought I had this same issue on Intel Mac running Big Sur, CPU at 100% and fan lifting off. Metamask extension completely unresponsive. About 15min later it pops up and wallet was imported. Used to import from JSON instantly
Chrome 87.0.4280.88

kyrogue:
same issue happening, on chrome  87.0.4280.88 , metamask 8.1.9 , when importing JSON key it hangs and never recovers

rekmarks:
[...]

PR #17763, opened 2023-02-15
Upgrading the Import Account modal

### LINKED ISSUE #16895: In the Import Account -> Private Key screen, error messages are confusing and unhelpful
There are three types of error messages you can get when on the Import Account -> Private Key screen, and all of them are confusing and unhelpful.  The third error message was reported by CS, but nobody has figured out how to reproduce it yet.

image

image

METAMASK

HowardBraham:
When fixing these issues, I also tried to accomplish the following:
- Use the Design System
- Follow DRY principles (code length was reduced by 166 lines while adding more functionality)
- Convert to React Hooks
- Don't use Redux where it's not necessary

Fixes #16895 and #17719.
New error messages are in `messages.json` and are:

> "You have entered a Secret Recovery Phrase (or mnemonic). To import an account here, you have to enter a private key, which is a hexadecimal string of length 64."
>
>  "This is not a valid private key. You have entered a hexadecimal string, but it must be 64 characters long."
>
> "This is not a valid private key. You must enter a hexadecimal string of length 64."

Also helps with expectations around this multi-minute freeze when importing a JSON file #10113

image


ui/pages/create-account/import-account/import-account.stories.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import Box from '../../../components/ui/box';
+import NewAccountImportForm from '.';
+
+export default {
+  title: 'Pages/CreateAccount/ImportAccount',
```
NidhiKJha, line 6: We can add component here
```suggestion
  title: 'Pages/CreateAccount/ImportAccount',
  component: NewAccountImportForm,
```

ui/pages/create-account/new-account.stories.js
```diff
@@ -15,7 +16,12 @@ export default {
 };
 export const DefaultStory = (args) => {
   return (
-    <NewAccountCreateForm {...args} createAccount={action('Account Created')} />
+    <Box className="new-account">
+      <NewAccountCreateForm
+        {...args}
+        createAccount={action('Account Created')}
+      />
+    </Box>
```
NidhiKJha, line 24: We can remove Box here
```suggestion
      <NewAccountCreateForm
        {...args}
        createAccount={action('Account Created')}
      />
```

ui/pages/create-account/import-account/index.js
```diff
@@ -1,79 +1,160 @@
...
+  return (
+    <>
+      <Box
+        padding={4}
+        style={{ borderWidth: '0px 0px 1px 0px' }} // There is no way to do just a bottom border in the Design System
+        borderColor={BorderColor.borderDefault}
+      >
+        <Text variant={TextVariant.headingLg}>{t('importAccount')}</Text>
+        <Text variant={TextVariant.bodySm}>
+          {t('importAccountMsg')}{' '}
```
NidhiKJha, line 127: Need to add a margin-top here.

ui/pages/create-account/import-account/index.js
```diff
@@ -1,79 +1,160 @@
...
+          <ButtonLink size={Size.inherit} onClick={moreInfoLink}>
+            {t('here')}
+          </ButtonLink>
```
NidhiKJha, line 130: In the current Ui, we have an underline for this text. Are we removing it in this design?

### Before 
n bmh

### After
bb

NidhiKJha (commented):
Looking really good. Left a couple of UI suggestions

ui/pages/create-account/import-account/index.js
NidhiKJha: Are we updating the `Text` colors as well @HowardBraham ?

HowardBraham: @NidhiKJha the removal of the underline and the change of the Text colors are not exactly intentional, but just a consequence of going with the defaults for the Design System.  We may want to discuss if we want to change these defaults, if we want to let it be, or if we want to customize this page.

georgewrmarshall (commented):
This is looking great! I've made a few suggestions in https://github.com/MetaMask/metamask-extension/pull/17905 - it requires a rebase of `develop` on this PR to see the changes effectively

ui/components/component-library/text/text.constants.js
```diff
@@ -3,3 +3,9 @@ export const TEXT_DIRECTIONS = {
   RIGHT_TO_LEFT: 'rtl',
   AUTO: 'auto',
 };
+
+/**
+ * The INVISIBLE_CHARACTER is a very useful tool if you want to make sure a line of text
+ * takes up vertical space even if it's empty.
+ */
+export const INVISIBLE_CHARACTER = '\u200d';
```
georgewrmarshall, line 11: I don't have any experience using invisible characters? Wouldn't we just want the html node to not be rendered? I did bring this up with @Akatori-Design and he sad that the movement/jump of the form elements when there is an error is the expected behaviour. @garrettbear how do you feel about this?

georgewrmarshall: I guess it would still jump if the error message is more than one line?

ui/pages/create-account/import-account/import-account.js
```diff
@@ -0,0 +1,159 @@
...
+  function getLoadingMessage(strategy) {
+    if (strategy === 'JSON File') {
+      return (
+        <Text width={BLOCK_SIZES.THREE_FOURTHS} fontWeight={FONT_WEIGHT.BOLD}>
+          <br />
+          {t('importAccountJsonLoading1')}
+          <br />
+          <br />
+          {t('importAccountJsonLoading2')}
+        </Text>
```
georgewrmarshall, line 81: I'm not a huge fan of using `<br />` tags to space out text. But understand that there is some performance benefits to using less complex jsx instead of using 2 `Text` components

ui/pages/create-account/import-account/import-account.js
```diff
@@ -0,0 +1,159 @@
...
+      default:
+        return <JsonImportView importAccountFunc={importAccount} />;
+    }
+  }
+
+  return (
+    <>
+      <Box
+        padding={4}
+        style={{ borderWidth: '0px 0px 1px 0px' }} // There is no way to do just a bottom border in the Design System
```
georgewrmarshall, line 122: I think unfortunately for any additional styles we want to use a scss file and css? I think it is slightly more performant that inline styles as well.

ui/pages/create-account/import-account/bottom-buttons.js
```diff
@@ -0,0 +1,48 @@
...
+
+export default function BottomButtons({
+  importAccountFunc,
+  isPrimaryDisabled,
+}) {
+  const t = useI18nContext();
+  const dispatch = useDispatch();
+
+  return (
+    <Box display={DISPLAY.FLEX}>
```
georgewrmarshall, line 27: Need to add a gap here
```suggestion
    <Box display={DISPLAY.FLEX} gap={4}>
```
Screenshot 2023-03-02 at 1 29 40 PM

georgewrmarshall (commented):
This is looking fantastic! Left some small suggestions. Also not sure about the invisible character but would like to get further opinions

ui/components/component-library/text/text.constants.js
HowardBraham: `\u200B` and `\u200D` appear to function the same way, but the `\u200B` description says "this character is intended for invisible word separation and for line break control," so I think you're right that it's a better choice.

Yeah, the error message is always at least one line high, and it still jumps if the error message is more than one line.  The INVISIBLE_CHARACTER is I think a better alternative to what I was doing before with `marginTop={warning ? 0 : 6}`

ui/pages/create-account/import-account/bottom-buttons.js
georgewrmarshall:
```suggestion
    <Box display={DISPLAY.FLEX} gap={4}>
```

ui/components/component-library/text/text.constants.js
georgewrmarshall: When I first built the form components I did try to account for the jumping but @Akatori-Design said that this was the intended behaviour 


https://user-images.githubusercontent.com/8112138/222871266-78b353da-c9c3-49c4-ac78-572e6cad46b7.mov

ui/pages/create-account/import-account/bottom-buttons.js
HowardBraham: I fixed this in a previous commit, I'm not sure how it came back 🤔

ui/components/component-library/text/text.constants.js
HowardBraham: Okay I made it jump, I don't really love it, but oh well

georgewrmarshall: Thanks. I'll bring it up in our next design system sync and if the pattern changes we can do it for all error messages in a different PR

georgewrmarshall (approved):
LGTM! 🚀

montelaidev (approved):

PR #17772, opened 2023-02-15
avoid mutation of state in signTransaction action

### LINKED ISSUE #17745: [Bug]: Error when updating a tx cannot proceed to Confirmation screen - `Invariant failed: A state mutation was detected between dispatches, in the path 'metamask.unapprovedTxs.`
### Describe the bug

**Problem**: when I am in the process of performing a Send tx, if I go back to Edit the tx, I can see the error `Invariant failed: A state mutation was detected between dispatches, in the path 'metamask.unapprovedTxs.` and I cannot proceed on the Confirmation screen.

This happens when updating any of these fields:
- tx recipient
- asset
- amount

https://user-images.githubusercontent.com/54408225/218739814-0d671a8c-53f1-46d3-8381-11fa28a53b4f.mp4

### Steps to reproduce

1. Click Send
2. Add recipient
3. Add amount
4. Go Next
5. Click Edit
6. Edit recipient amount or asset
7. Click Next -- error appears and cannot proceed

_No response_

### Version

10.26.0

### Build type

None

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

brad-decker:
## Explanation
Fixes #17745 

In #17508 i added store creation using reduxjs/toolkit instead of vanilla redux. One thing that comes with this is detection of mutation in state, which will throw an invariant error. This *should only be happening in dev mode* but its good to fix. This fixes the specific situation by cloning the transaction object before it used. 

## Manual Testing Steps
1. Create a transaction, but don't confirm it.
2. Edit transaction.
3. Change the recipient
4. On develop: Can't get to confirmation and invariant error in console.
5. On this branch: Confirmation.

## Pre-merge author checklist

- [x] I've clearly explained:


darkwing (approved):

ui/ducks/send/send.js
```diff
@@ -2270,7 +2270,7 @@ export function signTransaction() {
       // merge in the modified txParams. Once the transaction has been modified
       // we can send that to the background to update the transaction in state.
       const unapprovedTxs = getUnapprovedTxs(state);
-      const unapprovedTx = unapprovedTxs[draftTransaction.id];
+      const unapprovedTx = cloneDeep(unapprovedTxs[draftTransaction.id]);
       // We only update the tx params that can be changed via the edit flow UX
```
jpuri, line 2274: Just curious why it needs deep cloning, can something like destructuring help ?

jpuri (approved):

ui/ducks/send/send.js
brad-decker: Because transactions have sub objects that also need cloned, not the reference to those sub objects, and if we were to do an object spread it does a shallow copy unless you specify every sub object. Destructuring to get the values we want won't work, because we attach a new key to this object later. A deeper refactor could be done to this for sure but for now this seems like the safest approach to maintain the functionality.

image

PR #17835, opened 2023-02-20
Update the PhishingController to v2 and update phishing warning page

Gudahtt:
The PhishingController has been updated to v2. This release should dramatically reduce network traffic and double the update speed of the phishing list.

This was accomplished by combining both of our phishing configurations into one list (the "stalelist"), then creating a separate list of the changes just the past few days (the "hotlist"). Now users will download a smaller list more frequently (every 30 minutes rather than every hour), whereas the full list is only updated every 4 days.

The combined configuration means that we no longer know which list was responsible for each block. The phishing warning page has been updated to dynamically look this information up, to ensure users are still directed to the correct place to dispute a block. This update to the phishing warning page also includes the recent redesign.

## Screenshots/Screencaps

The architectural changes can't be seen visually, but this PR also includes the phishing warning page redesign.

You can see those differences visually by looking at this the description for this PR: https://github.com/MetaMask/phishing-warning/pull/52

## Manual Testing Steps

To test the new phishing warning page design, you can go through the usual steps to test the phishing warning page (see our phishing warning e2e tests for a few example workflows).

For the architectural changes to how the phishing configuration is updated, we can prepare and install a build, then wait for a phishing config update (maybe we can coordinate with one of the `eth-phishing-detect` maintainers to merge something at the right time). You should see the update within 0-60 minutes (max 30 minute delay on our API updating, and max 30 minute delay on the client requesting the updated config from the API). You could leave the background process dev tools open to see the network request, and confirm that it is a relatively small diff.

## Pre-merge author checklist

- [x] I've clearly explained:

- [x] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone


test/e2e/mock-e2e.js
```diff
@@ -341,6 +356,22 @@ async function setupMocking(server, testSpecificMock) {
     });
 
   testSpecificMock(server);
+
+  // Mocks below this line can be overridden by test-specific mocks
```
Gudahtt, line 360: I wanted to set default mocks for these requests, but override them in the phishing tests. I found that this only worked if they were declared below this line. Presumably the first handler wins with `mockttp`.

cc @PeterYinusa , tagging you in case I'm missing something here, since you set this up.

Gudahtt:
This depends upon #17837

Gudahtt:
e2e tests are expected to fail until this is published, and included in this PR: https://github.com/MetaMask/phishing-warning/pull/75

test/e2e/mock-e2e.js
PeterYinusa: Thanks for tagging me here. I think I should probably add some comment blocks/documentation to provide clarity on the mocks.
Personally, I feel it needs to be cleaned up a little, as the use cases increased as we added additional test cases. Here's some general info on how it's intended to work at the moment.

1. Global mocks: Mocks in this [file](https://github.com/MetaMask/metamask-extension/blob/update-phishing-controller/test/e2e/mock-e2e.js) are supposed to be global (across all tests). These mocks are set up `before` the extension is installed.
2. Local mocks: Some mocks are specific to a single test. In this case, we use do not put them in the global mocks file, but instead place them [directly in the test](https://github.com/MetaMask/metamask-extension/blob/update-phishing-controller/test/e2e/tests/errors.spec.js#L6). Another reason we may place the mock directly in the test, is so we can do [assertions on the requests](https://github.com/MetaMask/metamask-extension/blob/update-phishing-controller/test/e2e/tests/errors.spec.js#L52).
3. Local pre-installation mocks: These are mocks that need to be in place [before the extension is installed](https://github.com/MetaMask/metamask-extension/blob/update-phishing-controller/test/e2e/tests/phishing-detection.spec.js#L102) and are specific to a single test.

Gudahtt: Thanks, that all makes sense.

This case still isn't clear though. Here I want to use a global mock, but override it for a specific test. Does placing it here before the invocation of the local pre-installation mocks make sense? Should we let all global mocks be overridden in this way?

PeterYinusa: Created an issue here to tidy things up https://github.com/MetaMask/metamask-extension/issues/17843

PeterYinusa: Actually, in reference to my comment above, we would still need a way to do assertions on a request. 
But your suggestion seems sound to me.

Gudahtt:
This is ready for review/QA but I've just realized that I forgot to update the PR description to include manual testing steps and screenshots. I will do that ASAP then bring it back out of draft.

app/scripts/migrations/080.ts
```diff
@@ -0,0 +1,38 @@
...
+  data: Record<string, unknown>;
+}) {
+  const versionedData = cloneDeep(originalVersionedData);
+  versionedData.meta.version = version;
+  versionedData.data = transformState(versionedData.data);
+  return versionedData;
+}
+
+function transformState(state: Record<string, unknown>) {
+  if (
```
danjm, line 26: Only out of curiosity: any reason to prefer this style to something like
```
if (!state.PhishingController) {
return state;
}
```

danjm: and could `state.PhishingController` be anything other than an object?

Gudahtt: Yes; TypeScript doesn't like it when you reference properties that might not exist.

And for the `isObject` check it's the same; TypeScript doesn't like it when you destructure things that may-or-may-not be objects.

danjm:
@Gudahtt looking at the other PR you linked, I saw this unresolved question: https://github.com/MetaMask/phishing-warning/pull/52/files#r1106413619

Was it addressed at some point?

app/scripts/migrations/080.ts
danjm: okay, makes sense, thanks!

danjm (commented):
I have a couple of questions and can approve from a code review perspective once they are resolved.

Also, I see that this has been marked for manual QA from the QA team @PeterYinusa @chloeYue @tmashuang

app/scripts/migrations/080.ts
Gudahtt: >could state.PhishingController be anything other than an object?

I set the type here to `unknown` because it's coming from disk, so in theory it could be anything. I would never expect it to be anything other than an object in practice.

app/scripts/metamask-controller.js
```diff
@@ -3743,15 +3739,11 @@ export default class MetamaskController extends EventEmitter {
...
-   * @param {object} phishingTestResponse - Result of calling `phishingController.test`,
-   * which is the result of calling eth-phishing-detects detector.check method https://github.com/MetaMask/eth-phishing-detect/blob/master/src/detector.js#L55-L112
    */
-  sendPhishingWarning(connectionStream, hostname, phishingTestResponse) {
-    const newIssueUrl = PHISHING_NEW_ISSUE_URLS[phishingTestResponse?.name];
-
+  sendPhishingWarning(connectionStream, hostname) {
     const mux = setupMultiplex(connectionStream);
     const phishingStream = mux.createStream('phishing');
-    phishingStream.write({ hostname, newIssueUrl });
```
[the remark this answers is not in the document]
Gudahtt: Yes, great catch

Gudahtt:
>Was it addressed at some point?

Yes, just replied to that with a link to the PR where the test was introduced. Though all of those tests were later replaced by Playwright tests anyway.

app/scripts/metamask-controller.js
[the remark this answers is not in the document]
Gudahtt: Done here: https://github.com/MetaMask/metamask-extension/pull/17835/commits/5ab6a374c6c234a56262f5f5387a27adb32ee589

danjm (dismissed):
LGTM

Sylvia24:

brad-decker (dismissed):
LGTM

Gudahtt:
Rebased to resolve a conflict in `package.json`

danjm (approved):

chloeYue:
QA test OK, LGTM, thanks!

brad-decker (approved):

PR #17473, opened 2023-01-27
Refactor ConfirmTransaction to Functional Component

digiwand:
## Explanation

Fixes https://github.com/MetaMask/metamask-extension/issues/17242

Dear reviewer, 
- might be better to walk through each commit
- slight cleanup in a follow-up ticket here: https://github.com/MetaMask/metamask-extension/pull/17546

Let me know if I can hop on a call for any questions. Thank you,

Regarding testing, we had an internal discussion about how it would be ideal to have e2e tests written before refactoring. Regarding unit tests, we may need more discussions to strategize our approach as we plan to refactor pages and test alongside e2e tests.

## Screenshots/Screencaps

Screen Shot 2023-02-02 at 2 17 59 AM

## Manual Testing Steps

Test appropriate ConfirmTransaction routes are reached. (See snapshot or tests to refer to routes)

## Pre-merge author checklist

- [x] I've clearly explained:
  - [x] What problem this PR is solving
  - [x] How this problem was solved
  - [x] How reviewers can test my changes
- [x] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [x] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
- [x] PR is linked to the appropriate GitHub issue
- [x] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.

In this case, a QA Engineer approval will be be required.

PR #17838, opened 2023-02-20
Fix state in confirm transaction

### LINKED ISSUE #17746: [Bug]: Tperror `Cannot destructure property 'balance' of 'accounts[fromAddress]' as it is undefined.` when approving 2 pending transactions triggered from a Dapp
### Describe the bug

**Problem**: when I trigger more than one tx from a Dapp, and I confirm the first one, when I confirm the 2nd one, I land on the MetaMask error page with the error `Type Error: Cannot destructure property 'balance' of 'accounts[fromAddress]' as it is undefined.`

https://user-images.githubusercontent.com/54408225/218745094-0f94649d-22b9-4ee5-93b1-8f376e23c73a.mp4

**Notice**: since the popup will close right after, an effective way to reproduce this is to trigger 3 transactions instead of 2, this way the popup will not close and you'll see the error. Another way is to confirm the 2 transactions from the MetaMask full view

image

### Steps to reproduce

1. Go to test-dapp
2. Connect to test-dapp
3. Click Deploly Contract -- do nothing on MM
4. Click Deploy Failing Contract -- do nothing on MM 
5. Click Deploy Multisig Contract
6. Confirm one tx on MM
7. Confirm a second tx on MM -- error appears and you land on the MM error page

_No response_

### Version

10.26.0

### Build type

None

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

jpuri:
Fixes https://github.com/MetaMask/metamask-extension/issues/17746 (v10.26.0 bug)

An issue occurs during a transition state after rejecting or confirming a transaction in a list of transactions. 
What happens: 
1. from confirm transaction pages with at least 3 transactions, click reject or confirm
2. unconfirmed transaction is now confirmed so the unconfirmedTransaction state is empty
3. transaction id and param id are still set to the old id. isValidTransactionId is true.
4. ConfirmTransactionBase is still mounted. The empty transaction causes the error


digiwand:
Fix looks good to me and appears to be working.

For documentation purposes, sharing some logged states.

**Actions to produce screenshots:** 
1. On ConfirmTransaction page with multiple transactions
2. Confirm 1st transaction
3. Confirm 2nd transaction - this one is displayed in the screenshots 

**Before the ConfirmTransaction refactor (https://github.com/MetaMask/metamask-extension/pull/17473) when the bug did not exist:** 
Screen Shot 2023-02-23 at 7 27 20 PM

**After the ConfirmTransaction refactor with the bug:**
Screen Shot 2023-02-23 at 7 27 20 PM

**This PR with fix:** 
Screen Shot 2023-02-23 at 8 00 19 PM

I think the refactor and these PR updates made logic improvement from the old code. It looks like we may no longer need support for `paramsTransactionId !== transactionId`. I'm not recommending to remove this now, but it's something we can look into following this fix. 

cc: @jpuri @Gudahtt

Sylvia24:

brad-decker (approved):

digiwand (approved):
LGTM!

PR #17846, opened 2023-02-21
Fix a bug where non-address types would be rendered as addresses in EIP-712 

### LINKED ISSUE #17711: [Bug]: uint256 field being rendered as address in signTypedData_v4
This was reported by OpenSea [here](https://consensys.slack.com/archives/CN2H44RAM/p1675986657331309).

It seems that if their "counter" field is a uint256 but if it's exactly 20 bytes it formats/shows as an address. The correct behavior would be to only render it as an address if that parameter is of type address.

## Screenshots
Screenshot 2023-02-09 at 6 39 05 PM

Screenshot 2023-02-09 at 6 39 05 PM

FrederikBolding:
## Explanation

Adds type annotation to sanitized EIP-712 messages such that the type annotation can be used for selecting how to display data. 

This also fixes https://github.com/MetaMask/metamask-extension/issues/17711

### DIFF
--- a/ui/components/app/signature-request/signature-request-data/signature-request-data.test.js
+++ b/ui/components/app/signature-request/signature-request-data/signature-request-data.test.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import configureMockStore from 'redux-mock-store';
 import { renderWithProvider } from '../../../../../test/lib/render-helpers';
+import { sanitizeMessage } from '../../../../helpers/utils/util';
 import Identicon from '../../../ui/identicon';
 import SignatureRequestData from './signature-request-data';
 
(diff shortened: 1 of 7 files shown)


Sylvia24:

ui/components/app/signature-request/signature-request-data/signature-request-data.test.js
```diff
@@ -59,410 +60,251 @@ describe('Signature Request Data', () => {
...
-            { name: 'contents', type: 'string' },
-          ],
-          Person: [
-            { name: 'name', type: 'string' },
-            { name: 'wallets', type: 'address[]' },
-          ],
-        },
-      };
-    });
-
```
jpuri, line 114: Hey are test cases below no longer relevant ?

jpuri (approved):

ui/components/app/signature-request/signature-request-data/signature-request-data.test.js
FrederikBolding: They were seemingly misunderstandings of what data the component should show, so I removed them because they would fail now that I changed the component input to be valid too 😅

brad-decker (approved):

Issue #17386, opened 2023-01-24
MV3: notification window button falsely clickable while service worker restarting, then a second window appears


danjm:
Repro:

1. Run MV3 branch
2. Create a signature request
3. terminate the service worker
4. try to confirm/submit the signature request, it will flash a loading screen but not submit
5. a new popup with the same signature request will appear, and can now be submitted

https://user-images.githubusercontent.com/7499938/214762856-d212dde5-03a8-43b4-a500-68b61e7a47fe.mp4

Instead, this should happen:

- no second popup should appear
- the popup that does appear should have the button disabled while the background is unavailable / the service worker is restarting

----
Originally titled "Get encryption public key confirmation fails if service worker restarts while confirmation screen is open", originally with the below repro:

Repro:
1. Click "Get Encryption Key" in the test dapp
7. Restart service worker
8. try to confirm

![Screenshot from 2023-01-24 14-07-08](https://user-images.githubusercontent.com/7499938/214367447-08720333-2b4a-40c5-b53c-7c84ea6f3e39.png)

PR #17855, opened 2023-02-22
fix(17855): persist popup when sw is restarted

DDDDDanica:
## Explanation

https://github.com/MetaMask/metamask-extension/issues/17386

The cause of problem is straightforward. When we restart sw, we have the screen reloaded, and the local variable `this._popupId` will be erased. It means even though we have a persist window open, inside `NotificationManager`, we will not have the record of that window. Hence `NotificationManager` will open a new window for us.

The fix of this issue is to store the windowId in high level appState, which is persistent until the window is closed. Therefore when `triggerUi` is called by different transactions, we will request the stored `currentPopupId` and inform `NotificationManager` of such id to reuse the previous window. 

PS: e2e will be conducted in a separate ticket in https://github.com/MetaMask/metamask-extension/issues/17386

https://user-images.githubusercontent.com/12678455/220637824-eeff3d32-87c4-49d2-9a06-f889e806c549.mov

## Manual Testing Steps

- run mv3 locally
- connect test-dapp: https://metamask.github.io/test-dapp/
- click "sign" in "Sign Typed Data V4"
- terminate service worker
- before: 2 screens overlay
- after: 1 screen is persisted when sw is restarted and content is reachable

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/app/scripts/lib/notification-manager.js
+++ b/app/scripts/lib/notification-manager.js
@@ -32,15 +32,19 @@ export default class NotificationManager extends EventEmitter {
    * Either brings an existing MetaMask notification window into focus, or creates a new notification window. New
    * notification windows are given a 'popup' type.
    *
+   * @param {Function} setCurrentPopupId - setter of current popup id from appStateController
+   * @param {number} currentPopupId - id of current opened metamask popup window
    */
-  async showPopup() {
-    const popup = await this._getPopup();
-
+  async showPopup(setCurrentPopupId, currentPopupId) {
+    this._popupId = currentPopupId;
+    this._setCurrentPopupId = setCurrentPopupId;
+    const popup = await this._getPopup(currentPopupId);
     // Bring focus to chrome popup
     if (popup) {
       // bring focus to existing chrome popup
       await this.platform.focusWindow(popup.id);
     } else {
+      // create new notification popup
       let left = 0;
       let top = 0;
       try {
@@ -57,7 +61,6 @@ export default class NotificationManager extends EventEmitter {
         left = Math.max(screenX + (outerWidth - NOTIFICATION_WIDTH), 0);
       }
 
-      // create new notification popup
       const popupWindow = await this.platform.openWindow({
         url: 'notification.html',
         type: 'popup',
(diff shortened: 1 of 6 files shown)


Sylvia24:

app/scripts/lib/notification-manager.js
```diff
@@ -106,7 +106,7 @@ export default class NotificationManager extends EventEmitter {
     return windows
       ? windows.find((win) => {
           // Returns notification popup
-          return win && win.type === 'popup' && win.id === this._popupId;
```
brad-decker, line 109: Can you explain why this works as a fix and what the original check was meant for? Are there any ramifications of removing this check?

Katew212:
Hey

app/scripts/lib/notification-manager.js
DDDDDanica: Thanks for checking ~ it was the issue due to refresh changed `this._popupId` to undefined so we cannot find a match for current window opened. And you are right there would be some regressions, for instance, if there's a popup window from another extension, removing this will hijack another window. Therefore the final solution will be saving this id in the appState level.

app/scripts/lib/notification-manager.js
```diff
@@ -71,13 +74,16 @@ export default class NotificationManager extends EventEmitter {
...
+      // and store the id to private variable this._popupId for future access
+      this._setCurrentPopupId(popupWindow.id);
       this._popupId = popupWindow.id;
     }
   }
 
   _onWindowClosed(windowId) {
     if (windowId === this._popupId) {
-      this._popupId = undefined;
+      this._setCurrentPopupId(undefined);
```
pedronfigueiredo, line 86: Can you help me understand what is different here from the method above that makes it so we only need to use `_setCurrentPopupId` and not assign it to the private variable `this._popupId = undefined`, where on `showPopup` we do the two simultaneously?

DDDDDanica: yeah sure ! in this case i wanna trigger `setCurrentPopupId` in appState to clear the currentPopupId. And when we trigger the popup again, we will use the value from `getState` to re-examine `this._popupId`. But since you pointed out, i feel it is safer to clear local variable as well to keep consistent, and in case any other scenario i didn't catch up that would cause regression any more.

brad-decker (approved):
Superb!

pedronfigueiredo (approved):

PR #17872, opened 2023-02-22
experimenting with webpack

kumavis:
exploring a possible path to lavamoat protections for webpack builds:
- @brad-decker on (unprotected) webpack buildsys (this branch)
- @legobeat on [webpack-endo-archive plugin](https://github.com/LavaMoat/webpack-plugin-compartment-map)
- @kumavis on [endo-archive bundler](https://github.com/endojs/endo/pull/1449)
- @naugtur on lavamoat-on-endo and [policy-protections-in-endo](https://github.com/endojs/endo/pull/1491)
 
reasons for webpack:
- [still](https://2021.stateofjs.com/en-US/libraries/build-tools/) widely used + [rich](https://webpack.js.org/awesome-webpack/) ecosystem
- [tree-shaking](https://webpack.js.org/guides/tree-shaking/) optimizations
- richer [code splitting](https://webpack.js.org/guides/code-splitting/) (lazy loading)

the pipeline looks something like:
`webpack -> endo archive -> bundle`

this pipeline also allows us to switch to a different bundler that can output Endo Archives in the future


Sylvia24:

brad-decker:
@kumavis I am working on getting webpack able to *fully* build the application including generating static assets and html files. The WebpackHtmlPlugin does templating as well but uses a different template format so we can cover that use case. My latest commit adds the 'distro' folder which shows the current output. Not that we can import images directly into our javascript and webpack will handle changing them into static assets OR we can use the CopyWepbackPlugin to copy our assets over and continue to reference them the way they are currently (or perhaps both). i'm going to continue on this path until I can get a working build of the extension, including manifest files, using just webpack...then we can add lavamoat support.

brad-decker:
distro folder removed cause I added static asset copying to the config. I also added what I think to be a drop in replacement for the code fencing transform, using a webpack plugin, but its configured naively right now. I need to get manifest copying/transpiling in place and start seeing what shakes loose when I attempt to add it to chrome. 😅

ui/css/design-system/typography.scss
```diff
@@ -1,80 +1,80 @@
 @use "sass:map";
 
-$fa-font-path: 'fonts/fontawesome';
+$fa-font-path: '~@fortawesome/fontawesome-free/webfonts';
```
legobeat, line 3: What's the idea here - is there some loader or plugin which will make this work? (reverting just this part seems to make it pass)

ui/components/app/app-header/app-header.component.js
```diff
@@ -9,7 +9,7 @@ import NetworkDisplay from '../network-display';
 
 ///: BEGIN:ONLY_INCLUDE_IN(beta)
 import BetaHeader from '../beta-header';
-///: END:ONLY_INCLUDE_IN(beta)
+///: END:ONLY_INCLUDE_IN
```
kumavis, line 12: why were these changes made?

kumavis:
These were the build outputs I got. I think this is minified?
I was hoping to see a larger reduction in size from tree shaking.
One larger reason for moving to webpack is easy lazy loading, but I haven't tried adding that to the mix

```
2.7M	218.js
3.8M	373.js
220K	44.js
204K	497.js
16K	623.js
2.3M	711.js
296K	background.js
8.0K	contentscript.js
4.0K	disable-console.js
4.0K	init-globals.js
1.3M	ui.js
4.0K	vendor/trezor/content-script.js
4.0K	vendor/trezor/usb-permissions.js
11M	total
```

ui/components/app/app-header/app-header.component.js
brad-decker: I was attempting to remove the custom code fencing transform alltogether, but as legobeat pointed out its probable that my changes were not fully effective.

mcmire:
I am excited that this is being worked on! Adding a comment to prevent this from being marked as stale.

PR #17875, opened 2023-02-23
feat: add the ConsenSys zkEVM (Linea) as a default network

VGau:
## Explanation

The goal of the PR is to add the ConsenSys zkEVM network (Linea) as a default test network like Goerli and Sepolia.
I've updated all translation files.
This change can only be tested when this PR (https://github.com/MetaMask/design-tokens/pull/408) is merged and the new design tokens library version is updated on metamask-extension repo.

### Before
Screenshot 2023-02-23 at 11 56 19 AM

### After
Screenshot 2023-03-14 at 6 01 16 PM

## Manual Testing Steps

Step 1:
Go to the metamask home screen and select Goerli test network in the dropdown.
Screenshot 2023-03-14 at 6 01 16 PM

Step 2:
Go to this website: https://bridge.goerli.zkevm.consensys.net/send?token=ETH and send some ETH (minimum 0.1) from Goerli to Consensys zkEVM .
Screenshot 2023-02-23 at 12 04 31 PM

Wait for transaction to be finished.

Step 3:
Go to the metamask home screen and select then new ConsenSys zkEVM test network in the dropdown.
Screenshot 2023-03-14 at 6 01 16 PM

Step 4:
Go to this website: https://bridge.goerli.zkevm.consensys.net/send?token=ETH and send some ETH (minimum 0.1) from  Consensys zkEVM to Goerli.
Screenshot 2023-02-23 at 12 07 30 PM

Wait for transaction to be finished.

Final step:
Check if all transactions succeed on Metamask.

## Pre-merge author checklist

- [X] I've clearly explained:


VGau:

Sylvia24:

app/scripts/controllers/network/network-controller.js
```diff
@@ -406,12 +409,17 @@ export default class NetworkController extends EventEmitter {
...
+    const isInfura =
+      INFURA_PROVIDER_TYPES.includes(type) &&
+      type !== NETWORK_TYPES.CONSENSYS_ZKEVM;
+
     if (isInfura) {
       this._configureInfuraProvider(type, this._infuraProjectId);
       // url-based rpc endpoints
     } else if (type === NETWORK_TYPES.RPC) {
       this._configureStandardProvider(rpcUrl, chainId);
+    } else if (type === NETWORK_TYPES.CONSENSYS_ZKEVM) {
```
danjm, line 421: why doesn't the `isInfura` case work here?

app/scripts/lib/account-tracker.js
```diff
@@ -281,6 +282,13 @@ export default class AccountTracker {
           );
           break;
 
+        case CHAIN_IDS.CONSENSYS_ZKEVM:
```
danjm, line 285: we probably shouldn't include this as `SINGLE_CALL_BALANCES_ADDRESS_CONSENSYS_ZKEVM` is an empty string right now

app/scripts/controllers/network/network-controller.js
danjm: oh, actually, we do want to call _configureStandardProvider for this, so we just should not include this network in `INFURA_PROVIDER_TYPES`

shared/constants/network.ts
```diff
@@ -430,11 +437,13 @@ export const INFURA_PROVIDER_TYPES = [
   NETWORK_TYPES.MAINNET,
   NETWORK_TYPES.GOERLI,
   NETWORK_TYPES.SEPOLIA,
+  NETWORK_TYPES.CONSENSYS_ZKEVM,
```
danjm, line 440: I don't think we want to include this in this list

shared/constants/network.ts
```diff
@@ -430,11 +437,13 @@ export const INFURA_PROVIDER_TYPES = [
...
 ];
 
 export const TEST_CHAINS = [
   CHAIN_IDS.GOERLI,
   CHAIN_IDS.SEPOLIA,
+  CHAIN_IDS.CONSENSYS_ZKEVM,
```
danjm, line 446: is this a test chain? or does the currency on this network have real value?

shared/constants/network.ts
```diff
@@ -469,6 +481,11 @@ export const BUILT_IN_NETWORKS = {
     chainId: CHAIN_IDS.SEPOLIA,
     ticker: TEST_NETWORK_TICKER_MAP[NETWORK_TYPES.SEPOLIA],
   },
+  [NETWORK_TYPES.CONSENSYS_ZKEVM]: {
```
danjm, line 484: I don't think we want this in the built in networks, but rather in the custom network list

app/scripts/controllers/network/network-controller.js
VGau: because the consenSys zkEVM network is not considered as a supported by Infura in this repository https://github.com/MetaMask/eth-json-rpc-infura
I created a PR in https://github.com/MetaMask/eth-json-rpc-infura repository to fix this issue. Once this PR (https://github.com/MetaMask/eth-json-rpc-infura/pull/72) is merged, I will remove this change.

shared/constants/network.ts
danjm: so in `FEATURED_RPCS`

danjm: oh, I see that you've added to the default dropdown via multiple code changes... so this is a product decision

ui/css/itcss/components/network.scss
```diff
@@ -25,6 +25,10 @@
     background-color: rgba(207, 181, 240, 0.7) !important;
   }
 
+  &.consensyszkevm-test-network .menu-icon-circle div {
+    background-color: rgba(35, 79, 213, 0.7) !important;
+  }
+
```
georgewrmarshall, line 31: Is this className used anywhere? From a search it looks like it is not

Screenshot 2023-03-02 at 6 15 30 PM

georgewrmarshall (commented):
Hey @VGau, this is looking good from a UI perspective. Do you know if `#234FD5` is the brand color for Consensys? We could add this too our design tokens and replace it so consensys zkevm has it's own color and doesn't need to use the Goerli color

georgewrmarshall:
Hey @VGau, I was wondering instead of using color we could add these images for test networks?? I think it's a lot more scalable 
logo-test-networkgoerli
logo-test-networklinea
logo-test-networksepolia

VGau:
> Hey @VGau, I was wondering instead of using color we could add these images for test networks?? I think it's a lot more scalable
> 
> logo-test-network-goerli logo-test-network-linea logo-test-network-sepolia

Yes sure ! I will add these images. Thanks 😄

ui/css/utilities/colors.scss
```diff
@@ -4,6 +4,8 @@
   --mainnet: #29b6af;
   --inherit: inherit;
   --transparent: transparent;
+  --color-network-linea-testnet-default: #000000;
+  --color-network-linea-testnet-inverse: #fcfcfc;
```
georgewrmarshall, line 8: Usually we wouldn't allow static hex values but I think we can refactor the test networks to use images like other token icons during the multichain v0.5 work so this will be temporary.

ui/css/utilities/colors.scss
```diff
@@ -4,6 +4,8 @@
...
   // DO NOT CHANGE
```
georgewrmarshall, line 9: Usually we wouldn't allow static hex values but I think we can refactor the test networks to use images like other token icons during the multichain v0.5 work so this will be temporary.

georgewrmarshall (dismissed):
UI approved!

I think it's fine, but it might be worth checking if prod/your design contact is fine with the color in dark mode

Screenshot 2023-03-15 at 8 41 17 AM
Screenshot 2023-03-15 at 8 41 52 AM

app/scripts/controllers/network/network-controller.js
```diff
@@ -268,14 +269,20 @@ export default class NetworkController extends EventEmitter {
       NETWORK_TYPES.RPC,
       `NetworkController - cannot call "setProviderType" with type "${NETWORK_TYPES.RPC}". Use "setActiveNetwork"`,
     );
-    assert.ok(
-      INFURA_PROVIDER_TYPES.includes(type),
-      `Unknown Infura provider type "${type}".`,
-    );
+    if (type !== NETWORK_TYPES.LINEA_TESTNET) {
```
danjm, line 272: why is this if block being added?

danjm: hmm... I am reading this comment https://github.com/MetaMask/metamask-extension/pull/17875#discussion_r1122182027

danjm: in any case, we shouldn't create a new type for this network. It should either be a `INFURA_PROVIDER_TYPES` or a  `NETWORK_TYPES.RPC`

I would suggest we treat this as a `NETWORK_TYPES.RPC`

danjm: and as opposed to modifying the network controller, we can just modify `ui/components/app/dropdowns/network-dropdown.js`

treating this network in a similar way to how localhost is treated, within the render function of that component

danjm:
I believe there is a requirement that this network not be shown until March 28. Can you add some logic to the network dropdown and the settings sections to ensure that is the case.

roughly something like:
```
const shouldShowZKEVM = Date.now() > (new Date('March 28, 2023 12:00:00'))
```
And use that to conditionally render the network in the dropdown and in the network settings pages

app/scripts/controllers/network/network-controller.js
danjm: it can still be in the network dropdown, we just need to place it there within `ui/components/app/dropdowns/network-dropdown.js`

danjm: similarly to how localhost is added to the network dropdown

danjm:
this branch probably needs to be updated to latest develop to avoid the `prep-build-desktop` failure

danjm (approved):
These changes look good to me

pedronfigueiredo (approved):

PR #17876, opened 2023-02-23
Fix for on the send token screen, the insufficient error may show when the error is 'gasPriceFetchFailed'

### LINKED ISSUE #17012: [Bug]: On the send token screen, the insufficient error may show when the error is 'gasPriceFetchFailed'
### Describe the bug

The insufficient error may show when the error is 'gasPriceFetchFailed'. This could be misleading.

### Steps to reproduce

1. Log in to the extension
2. Inspect the error values in `metamask-extension/ui/pages/send/gas-display/gas-display.js` 
3. Go to "Send" tokens page
4. Switch networks
5. Observe `gasPriceFetchFailed` may show the insufficient funds error

_No response_

### Version

10.24.0

### Build type

None

### Browser

Chrome

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

aleksandar-mihajlovic:
## Explanation

Added fix, so now the insufficient error is not always shown, errors are managed by switch case and 'gasPriceFetchFailed' is now showing until loaded. 
* Fixes #17012

### Before

insuficientFundsBug

### After

insuficientFundsFix

## Unit tests

Screenshot 2023-02-23 at 12 13 59

##Lint

Screenshot 2023-02-23 at 11 43 35

## Manual Testing Steps

1. Log in to the extension
2. Go to "Send" tokens page
3. Switch networks
4. insufficientFundsForGas error is now only shown if amount+gas is greater then balance and not shown if amount is 0 or amount+gas smaller then balance
5. gasPriceFetchFailed could be seen for brief time when switching to Ethereum or Sepolia mainet, until it is loaded, then it is not shown


Sylvia24:

dzfjz:
Verified by QA

ui/pages/send/send.utils.js
```diff
@@ -36,6 +44,30 @@ function isBalanceSufficient({
   return balanceNumeric.greaterThanOrEqualTo(totalAmount);
 }
 
+function gasIsExcessive({
```
brad-decker, line 47: With this method only being used in the send duck, it should probably exist in the ducks/send/helpers file instead of in the page utils.

ui/pages/send/send.utils.js
```diff
@@ -25,7 +28,12 @@ function isBalanceSufficient({
   gasTotal = '0x0',
   primaryCurrency,
 }) {
-  let totalAmount = new Numeric(amount, 16).add(new Numeric(gasTotal, 16));
+  let totalAmount;
+  if (amount === '0x0') {
+    totalAmount = new Numeric(amount, 16);
+  } else {
+    totalAmount = new Numeric(amount, 16).add(new Numeric(gasTotal, 16));
+  }
```
brad-decker, line 36: What was this change made for?

brad-decker (changes requested):

ui/pages/send/send.utils.js
aleksandar-mihajlovic: Okay I will move it

ui/pages/send/send.utils.js
aleksandar-mihajlovic: Well if we add the gas fee to the amount when amount is 0, then total amount will always be greater then 0, and even though user has 0 balance and didnt type any amount, the insufficient funds error would be shown and that should not be the case, because there is warning message on the bottom that he does not have enough tokens for transactions. I think insufficient funds error should only be shown when user starts typing amount, not when amount is 0.

ui/ducks/send/send.js
```diff
@@ -1419,6 +1441,13 @@ const slice = createSlice({
             });
             draftTransaction.status = SEND_STATUSES.INVALID;
             break;
+          case draftTransaction.amount.value === '0x0' ||
+            draftTransaction.amount.value === '0x0':
```
brad-decker, line 1445: The two checks here on this || statement seem to be identical, is this intentional? IF so we probably only need one.

ui/ducks/send/send.js
```diff
@@ -1250,15 +1253,34 @@ const slice = createSlice({
      *
      * @param {SendStateDraft} state - A writable draft of the send state to be
      *  updated.
+     * @param {GasEstimateUpdatePayload} action - The gas fee update payload
```
brad-decker, line 1256: Place `[]` around 'action' because its optional. Only one callsite passes an action payload.

ui/pages/send/gas-display/gas-display.js
```diff
@@ -267,7 +267,10 @@ export default function GasDisplay({ gasError }) {
                 />
               }
             />,
-            (gasError || isInsufficientTokenError) && (
+            (gasError ||
+              isInsufficientTokenError ||
+              draftTransaction.asset.balance === '0x0' ||
+              draftTransaction.asset.balance === '0x00') && (
```
brad-decker, line 273: in what cases is it '0x00' ?

ui/pages/send/gas-display/gas-display.js
```diff
@@ -316,74 +319,78 @@ export default function GasDisplay({ gasError }) {
...
       </Box>
-      {(gasError || isInsufficientTokenError) && currentNetworkName && (
-        <Box
-          className="gas-display__warning-message"
-          data-testid="gas-warning-message"
-        >
+      {(gasError ||
+        isInsufficientTokenError ||
+        draftTransaction.asset.balance === '0x0' ||
+        draftTransaction.asset.balance === '0x00') &&
```
brad-decker, line 325: If the use case for '0x00' is valid, then we should probably do something like 

```
const assetBalanceIsZero = draftTransaction.asset.balance === '0x0' || 
  draftTransaction.asset.balance === '0x00';
```

and then use that constant in the two places we need it.

brad-decker (changes requested):

ui/pages/send/gas-display/gas-display.js
aleksandar-mihajlovic: I am not sure, but sometimes it is '0x0' and sometimes '0x00' i dont know why

ui/ducks/send/send.js
aleksandar-mihajlovic: My mistake, it should not be identical, it should with '0x0' and '0x00', i will change that

ui/ducks/send/send.js
aleksandar-mihajlovic: Will fix that

ui/pages/send/gas-display/gas-display.js
aleksandar-mihajlovic: Okay I will do that

ui/pages/send/send.utils.js
brad-decker: Should this now be deleted?

brad-decker (changes requested):

ui/pages/send/send.utils.js
aleksandar-mihajlovic: Now it is, i missed it

brad-decker (dismissed):

aleksandar-mihajlovic:
Moved to new [PR](https://github.com/MetaMask/metamask-extension/pull/18019)

Issue #17839, opened 2023-02-21
[Bug]: Zero Decimal Support for ERC-20 Transfers


Mentors4EDU:
### Describe the bug

Hi, certain tokens such as https://etherscan.io/token/0x1da4858ad385cc377165a298cc2ce3fce0c5fd31#code have zero decimals but a transfer method. Something many developers have noticed is that tokens with zero decimals have trouble being sent through the MetaMask Wallet UI and must directly connect to the smart contract to use the transfer feature. If there is a potential fix for zero decimal compatibility for sending/transferring in the next wallet update, that would be great.

### Steps to reproduce

1. Add a zero decimal or minted zero decimal ERC-20 token (Try on Ropsten as well)
2. Try sending through MetaMask

### Error messages or log output

```shell
Endless load or doesn't connect
```

### Version

10.26

### Build type

None

### Browser

Chrome, Firefox, Microsoft Edge

### Operating system

Windows, Linux

danjm:
@Mentors4EDU You mentioned that this is in v10.26.0

Have you seen it in our most recent production versions as well? (v10.24.2 or v10.25.0)

danjm:
we experienced this before, and seemingly fixed it, so this could be a recent regression
https://github.com/MetaMask/metamask-extension/issues/16596
https://github.com/MetaMask/metamask-extension/issues/16601

Mentors4EDU:
@danjm I'm not sure if I experienced it in previous versions.
I can send you test tokens on the Ropsten network to try it out.
Either way, wondering why zero decimal support isn't fully working on this MetaMask version.

Sylvia24:
We'd be happy to look into this for you. kindly file a support ticket below 
https://forms.gle/Wmpm6NLzYm9uXYuU9

Mentors4EDU:
> We'd be happy to look into this for you. kindly file a support ticket below https://forms.gle/Wmpm6NLzYm9uXYuU9

I don't trust random links

bschorchit:
> we experienced this before, and seemingly fixed it, so this could be a recent regression #16596 #16601

@danjm I believe this is the same issue that @seaona reported here https://github.com/MetaMask/metamask-extension/issues/17542 and has been fix by @segun on https://github.com/MetaMask/metamask-extension/pull/17651. @brad-decker shared a nice explanation of it [here](https://github.com/MetaMask/metamask-extension/issues/17542#issuecomment-1431537316).

[...]

PR #17891, opened 2023-02-24
Use tokenList to get token details, when available, in getTokenStanda…

### LINKED ISSUE #17577: More frequent and longer loading time on the send screen
> -When clicking on a custom token, then pressing send, the extension loads for a while, other times you need  to close the pop-up and try again
image (13)
image (12)

> -Changing the asset I want to send to another one from the transfer window also makes the extension load for a bi

danjm:
…rdAndDetails

Previously, every call to getTokenStandardAndDetails would fetch data via the provider. This would result in at least 3 network requests whenever that method is called for an ERC20 token, contributing to unneccesary loading and lagging in multiple places. This commit takes advantage of stored data we already have available to avoid the unnecessary loading.

Fixes https://github.com/MetaMask/metamask-extension/issues/17577

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -2215,15 +2215,15 @@ export default class MetamaskController extends EventEmitter {
       !isEqualCaseInsensitive(tokenDetails.standard, TokenStandard.ERC721) &&
       !tokenDetails.erc721;
 
-    const otherThanStandardDetailsAreERC20Like =
+    const otherDetailsAreERC20Like =
       tokenDetails.decimals !== undefined && tokenDetails.symbol;
 
-    const weCanAttemptToFetchERC20TokenBalance =
+    const tokenCanBeTreatedAsAnERC20 =
       tokenDetailsStandardIsERC20 ||
-      (noEvidenceThatTokenIsAnNFT && otherThanStandardDetailsAreERC20Like);
+      (noEvidenceThatTokenIsAnNFT && otherDetailsAreERC20Like);
 
     let details;
-    if (weCanAttemptToFetchERC20TokenBalance) {
+    if (tokenCanBeTreatedAsAnERC20) {
       try {
         const balance = await fetchTokenBalance(
           address,
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -2208,7 +2208,7 @@ export default class MetamaskController extends EventEmitter {
       ...tokenListDetails,
       ...userDefinedTokenDetails,
     };
-    const tokenDetailsStandardIsisERC20 =
+    const tokenDetailsStandardIsERC20 =
       isEqualCaseInsensitive(tokenDetails.standard, TokenStandard.ERC20) ||
       tokenDetails.erc20 === true;
     const noEvidenceThatTokenIsAnNFT =
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -64,7 +64,6 @@ import {
 ///: END:ONLY_INCLUDE_IN
 
 import browser from 'webextension-polyfill';
-import EthContract from 'ethjs-contract';
 import {
   AssetType,
   TransactionStatus,
@@ -2192,7 +2191,6 @@ export default class MetamaskController extends EventEmitter {
   }
 
   async getTokenStandardAndDetails(address, userAddress, tokenId) {
-    const contract = new EthContract(this.ethQuery);
     const { tokenList } = this.tokenListController.state;
     const tokenListEntry = tokenList[address.toLowerCase()];
     let details;
@@ -2200,7 +2198,7 @@ export default class MetamaskController extends EventEmitter {
       const standard = tokenListEntry.standard?.toUpperCase() ?? 'NONE';
       let balance;
       if (standard === 'ERC20') {
-        balance = await fetchTokenBalance(address, userAddress, contract);
+        balance = await fetchTokenBalance(address, userAddress, this.provider);
       }
 
       details = {
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -68,6 +68,7 @@ import {
   AssetType,
   TransactionStatus,
   TransactionType,
+  TokenStandard,
 } from '../../shared/constants/transaction';
 import {
   GAS_API_BASE_URL,
@@ -2192,29 +2193,64 @@ export default class MetamaskController extends EventEmitter {
 
   async getTokenStandardAndDetails(address, userAddress, tokenId) {
     const { tokenList } = this.tokenListController.state;
-    const tokenListEntry = tokenList[address.toLowerCase()];
-    let details;
-    if (tokenListEntry) {
-      const standard = tokenListEntry.standard?.toUpperCase() ?? 'NONE';
-      let balance;
-      if (standard === 'ERC20') {
-        balance = await fetchTokenBalance(address, userAddress, this.provider);
-      }
+    const { tokens } = this.tokensController.state;
 
-      details = {
-        address,
-        balance,
-        standard,
-        decimals: tokenListEntry.decimals,
-        symbol: tokenListEntry.symbol,
-      };
-    } else {
+    const staticTokenListDetails =
+      STATIC_MAINNET_TOKEN_LIST[address.toLowerCase()] || {};
+    const tokenListDetails = tokenList[address.toLowerCase()] || {};
+    const userDefinedTokenDetails =
+      tokens.find(({ address: _address }) =>
+        isEqualCaseInsensitive(_address, address),
+      ) || {};
+
+    const tokenDetails = {
+      ...staticTokenListDetails,
+      ...tokenListDetails,
+      ...userDefinedTokenDetails,
+    };
+    const tokenDetailsStandardIsisERC20 =
+      isEqualCaseInsensitive(tokenDetails.standard, TokenStandard.ERC20) ||
+      tokenDetails.erc20 === true;
+    const noEvidenceThatTokenIsAnNFT =
+      !isEqualCaseInsensitive(tokenDetails.standard, TokenStandard.ERC1155) &&
+      !isEqualCaseInsensitive(tokenDetails.standard, TokenStandard.ERC721) &&
+      !tokenDetails.erc721;
+    const otherThanStandardDetailsAreERC20Like =
+      tokenDetails.decimals !== undefined && tokenDetails.symbol;
+
+    const tokenCanBeTreatedAsAnERC20 =
+      tokenDetailsStandardIsisERC20 ||
+      (noEvidenceThatTokenIsAnNFT && otherThanStandardDetailsAreERC20Like);
+
+    let details;
+    if (tokenCanBeTreatedAsAnERC20) {
+      try {
+        const balance = await fetchTokenBalance(
+          address,
+          userAddress,
+          this.provider,
+        );
+
+        details = {
+          address,
+          balance,
+          standard: TokenStandard.ERC20,
+          decimals: tokenDetails.decimals,
+          symbol: tokenDetails.symbol,
+        };
+      } catch (e) {
+        log.warning('Failed to get token balance');
+      }
+    }
+
+    if (details === undefined) {
       details = await this.assetsContractController.getTokenStandardAndDetails(
         address,
         userAddress,
         tokenId,
       );
     }
+
     return {
       ...details,
       decimals: details?.decimals?.toString(10),
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -265,8 +265,6 @@ export default class MetamaskController extends EventEmitter {
     this.blockTracker =
       this.networkController.getProviderAndBlockTracker().blockTracker;
 
-    this.ethQuery = new EthQuery(this.provider);
-
     const tokenListMessenger = this.controllerMessenger.getRestricted({
       name: 'TokenListController',
     });
@@ -2394,10 +2392,11 @@ export default class MetamaskController extends EventEmitter {
         seedPhraseAsBuffer,
       );
 
+      const ethQuery = new EthQuery(this.provider);
       accounts = await keyringController.getAccounts();
       lastBalance = await this.getBalance(
         accounts[accounts.length - 1],
-        this.ethQuery,
+        ethQuery,
       );
 
       const [primaryKeyring] = keyringController.getKeyringsByType(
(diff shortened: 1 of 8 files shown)


app/scripts/metamask-controller.js
```diff
@@ -2204,12 +2207,31 @@ export default class MetamaskController extends EventEmitter {
   }
 
   async getTokenStandardAndDetails(address, userAddress, tokenId) {
-    const details =
-      await this.assetsContractController.getTokenStandardAndDetails(
+    const contract = new EthContract(this.ethQuery);
```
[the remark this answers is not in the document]
adonesky1: Oh I see. I suppose you could instantiate the contract with the `address` and pass it ready to go? Doesn't really matter either way I suppose.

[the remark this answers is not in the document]
danjm: I guess I could just pass a provider? And then instantiate the contract with the ethers class within the `fetchTokenBalance` method itself?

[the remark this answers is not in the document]
adonesky1: Yeah I think that's the cleanest way to do it!

app/scripts/metamask-controller.js
```diff
@@ -2351,11 +2373,10 @@ export default class MetamaskController extends EventEmitter {
         seedPhraseAsBuffer,
       );
 
-      const ethQuery = new EthQuery(this.provider);
       accounts = await keyringController.getAccounts();
       lastBalance = await this.getBalance(
```
adonesky1, line 2377: Couldn't `getBalance` just access `this.ethQuery` internally?

shared/lib/token-util.ts
```diff
@@ -0,0 +1,34 @@
+import abi from 'human-standard-token-abi';
```
adonesky1, line 1: we should probably use `{ abiERC20 } from '@metamask/metamask-eth-abis'` here? Technically `fetchTokenBalance` could be calling to an ERC721 or ERC1155, but they all have the same `balanceOf` function so the function hash from the ABI is the same. Perhaps we should create a token standard agnostic ABI for this just to avoid confusion tho? cc @FrederikBolding

adonesky1: We should replace all uses of `human-standard-token-abi` with `@metamask/metamask-eth-abis` in the repo. I can make a separate PR for this. Unless you'd like to @danjm

danjm: please go ahead with a separate PR

app/scripts/metamask-controller.js
adonesky1: ping on this question again

app/scripts/metamask-controller.js
```diff
@@ -2204,12 +2207,65 @@ export default class MetamaskController extends EventEmitter {
...
+      tokens.find(({ address: _address }) =>
+        isEqualCaseInsensitive(_address, address),
+      ) || {};
+
+    const tokenDetails = {
+      ...staticTokenListDetails,
+      ...tokenListDetails,
+      ...userDefinedTokenDetails,
+    };
+    const tokenDetailsStandardIsisERC20 =
```
adonesky1, line 2226: [nit]
```suggestion
    const tokenDetailsStandardIsERC20 =
```

app/scripts/metamask-controller.js
```diff
@@ -2204,12 +2207,65 @@ export default class MetamaskController extends EventEmitter {
...
+      isEqualCaseInsensitive(tokenDetails.standard, TokenStandard.ERC20) ||
+      tokenDetails.erc20 === true;
+    const noEvidenceThatTokenIsAnNFT =
+      !isEqualCaseInsensitive(tokenDetails.standard, TokenStandard.ERC1155) &&
+      !isEqualCaseInsensitive(tokenDetails.standard, TokenStandard.ERC721) &&
+      !tokenDetails.erc721;
```
adonesky1, line 2232: maybe?
```suggestion
    const noEvidenceThatTokenIsAnNFT =
      !tokenId &&
      !isEqualCaseInsensitive(tokenDetails.standard, TokenStandard.ERC1155) &&
      !isEqualCaseInsensitive(tokenDetails.standard, TokenStandard.ERC721) &&
      !tokenDetails.erc721
      
```

app/scripts/metamask-controller.js
```diff
@@ -2204,12 +2207,65 @@ export default class MetamaskController extends EventEmitter {
...
+    const otherThanStandardDetailsAreERC20Like =
+      tokenDetails.decimals !== undefined && tokenDetails.symbol;
```
adonesky1, line 2234: [nits]
- I don't think decimals would ever be a falsy but not undefined value (i.e. 0)?
- `otherThanStandardDetailsAreERC20Like` -> `otherDetailsAreERC20Like` (i think this is more legible, since it sounds like you're saying some details are standard and others are not. After a few seconds I got what you were saying but)
- and lastly I think `||` rather than `&&` since the presence of one is still evidence that is is ERC20 like even when the other is absent.
```suggestion
    const otherDetailsAreERC20Like =
      tokenDetails.decimals || tokenDetails.symbol;
```

app/scripts/metamask-controller.js
```diff
@@ -2204,12 +2207,65 @@ export default class MetamaskController extends EventEmitter {
...
+          standard: TokenStandard.ERC20,
+          decimals: tokenDetails.decimals,
+          symbol: tokenDetails.symbol,
+        };
+      } catch (e) {
+        log.warning('Failed to get token balance');
+      }
+    }
+
+    if (details === undefined) {
```
adonesky1, line 2261: same functionality and easier to read/understand imo: 

```suggestion
    if (tokenCanBeTreatedAsAnERC20) {
      try {
        const balance = await fetchTokenBalance(
          address,
          userAddress,
          this.provider,
        );

        details = {
          address,
          balance,
          standard: TokenStandard.ERC20,
          decimals: tokenDetails.decimals,
          symbol: tokenDetails.symbol,
        };
      } catch (e) {
        log.warning('Failed to get token balance');
      }
    } else {
```

shared/lib/token-util.ts
adonesky1: Actually at least for this instance you should probably just replace 
``` 
import abi from 'human-standard-token-abi';
```

with 

```
import { abiERC20 } from '@metamask/metamask-eth-abis'
```

and I can put up a PR for the other instances not added in this PR.

danjm: okay, will do, thanks

app/scripts/metamask-controller.js
danjm: > I don't think decimals would ever be a falsy but not undefined value (i.e. 0)?

Decimals can be `0`. Here is an example of a recent bug report where users expect us to support such tokens https://github.com/MetaMask/metamask-extension/issues/17839

app/scripts/metamask-controller.js
danjm: So this is not exactly the same, but your comment shows that my code is hard to understand, and at the very least requires some serious documentation.

The current code will go into the `if (details === undefined) {` block if it first entered `if (tokenCanBeTreatedAsAnERC20) {` but then `const balance = await fetchTokenBalance(`. So it tries to fetch the balance, but if that fails, tries to get the data via the assets controller call.

The suggested change here will not try the assets controller call if the `fetchTokenBalance` call fails.

adonesky1: oh true. Wasn't thinking straight ! Yeah maybe just a short comment here would help. I don't think it's that bad though.

app/scripts/metamask-controller.js
adonesky1: Screenshot 2023-03-06 at 5 29 46 PM

Interesting... doesn't look like this contract actually has a `decimals` value at all... I wonder if this is the fallback issue again

adonesky1: hmmm no it seems to revert when I instantiate it with `ethers` and call `decimals()`

adonesky1: I'll have to spin up a contract on a testnet that has no decimals getter/value to test what this [Mentors4EDU](https://github.com/Mentors4EDU) is claiming there. I'll do that tomorrow

danjm: > it seems to revert when I instantiate it with ethers and call decimals()
this would explain why support for the token broke within the past few weeks, as we shipped the asset contract controller update in January

danjm: So if the token in that github issue is manually imported, the user is forced to set the decimals to 0

https://user-images.githubusercontent.com/7499938/223395613-f24bd4ee-d7f6-4141-b166-3e593358c342.mp4

danjm: And here are three tokens in the token list we fetch from API that have 0 decimals specified in their contract:

https://etherscan.io/token/0xdacd69347de42babfaecd09dc88958378780fb62
https://etherscan.io/token/0xa150db9b1fa65b44799d4dd949d922c0a33ee606
https://etherscan.io/token/0x08d32b0da63e2c3bcf8019c9c5d849d7a9d791e6

app/scripts/metamask-controller.js
danjm: oh, no, the code changes are still there, but they can be removed

danjm: removed in 71d9f7ad52

app/scripts/metamask-controller.js
danjm: clarifying comments added in a92ffd62d5

app/scripts/metamask-controller.js
adonesky1: Ok got it! What do you think about the proposed name change? And `&&` ->. `||`?

app/scripts/metamask-controller.js
adonesky1: I think I preferred the old name `tokenCanBeTreatedAsAnERC20` over `weCanAttemptToFetchERC20TokenBalance` because, ERC721 and ERC1155 both have a `balanceOf` method

app/scripts/metamask-controller.js
danjm: name change sounds good

`&&` -> `||` would mean that a token with a symbol but no decimals would be "erc 20 like", but an NFT can have a symbol but not decimals, right?

app/scripts/metamask-controller.js
danjm: okay, good point

app/scripts/metamask-controller.js
adonesky1: I suppose that's true. Both ERC721s and ERC20s could have a symbol but no decimals.

adonesky1: Maybe you should just check for decimals? Since it's the only one that is potentially unique to ERC20s?

app/scripts/metamask-controller.js
adonesky1: 👍

app/scripts/metamask-controller.js
```diff
@@ -2189,12 +2190,72 @@ export default class MetamaskController extends EventEmitter {
...
+          symbol: tokenDetails.symbol,
+        };
+      } catch (e) {
+        // If the `fetchTokenBalance` call failed, `details` remains undefined, and we
+        // fall back to the below `assetsContractController.getTokenStandardAndDetails` call
+        log.warning('Failed to get token balance');
+      }
+    }
+
+    // `details`` will be undefined if `weCanAttemptToFetchERC20TokenBalance`` is false,
```
adonesky1, line 2248:
```suggestion
    // `details`` will be undefined if `tokenCanBeTreatedAsAnERC20`` is false,
```

app/scripts/metamask-controller.js
```diff
@@ -2189,12 +2190,72 @@ export default class MetamaskController extends EventEmitter {
...
-    const details =
-      await this.assetsContractController.getTokenStandardAndDetails(
+    const { tokenList } = this.tokenListController.state;
+    const { tokens } = this.tokensController.state;
+
+    const staticTokenListDetails =
+      STATIC_MAINNET_TOKEN_LIST[address.toLowerCase()] || {};
+    const tokenListDetails = tokenList[address.toLowerCase()] || {};
+    const userDefinedTokenDetails =
+      tokens.find(({ address: _address }) =>
```
adonesky1, line 2200: is `tokens` guaranteed to be defined and an array?

app/scripts/metamask-controller.test.js
```diff
@@ -1273,4 +1294,321 @@ describe('MetaMaskController', function () {
...
+        tokenDetails.symbol === tokenData.symbol,
+        'tokenDetails should include token symbol',
+      );
+      assert.ok(
+        tokenDetails.balance === tokenData.balance,
+        'tokenDetails should include a balance',
+      );
+    });
+
+    it('gets token data from the blockchain, via the assetsContractController, if it is in the token list but is an ERC1155', async function () {
```
adonesky1, line 1554: Why do we do this? If we have the details in our token list and its an ERC1155 or ERC721 should we skip the fetch?

app/scripts/metamask-controller.js
danjm: good question, I think the answer is yes https://github.com/MetaMask/core/blob/main/packages/assets-controllers/src/TokensController.ts#L107-L108

app/scripts/metamask-controller.test.js
danjm: I think that for now the token list presumed to be an erc20 list, and if an ERC1155 or ERC721 was there, I think we should treat it as an accident. Instead we probably want to skip the fetch if details on those are in the NFT controller. (Which we maybe should add to this function at some point, but we don't need it for v10.26.0)

app/scripts/metamask-controller.js
```diff
@@ -2189,12 +2190,72 @@ export default class MetamaskController extends EventEmitter {
...
+        );
+
+        details = {
+          address,
+          balance,
+          standard: TokenStandard.ERC20,
+          decimals: tokenDetails.decimals,
+          symbol: tokenDetails.symbol,
+        };
+      } catch (e) {
```
adonesky1, line 2241: last question! should we log the error itself too?

adonesky1: or then remove it?
```suggestion
      } catch {
```

adonesky1 (dismissed):
One more small nit but otherwise LGTM

app/scripts/metamask-controller.js
danjm: yes, good suggestion

danjm: Done in https://github.com/MetaMask/metamask-extension/pull/17891/commits/bae11c79dd8ae091d4601b1a0f66437c503e2554

adonesky1 (dismissed):
LGTM!

danjm:
@adonesky1 just did a rebase to fix a code coverage conflict

pedronfigueiredo (approved):

adonesky1 (approved):

PR #17892, opened 2023-02-24
identify desktop is paired in the metrics event

vinistevam:
## Explanation

Adds the metrics code needed to support tracking usage of the [MetaMask Desktop](https://github.com/MetaMask/metamask-desktop) app.

The aim of this PR is to identify if the desktop app is paired with the extension on MetaMetrics to track key operations such as swaps, snaps, transfers, and interaction with dapps. It also includes desktop-related events.

Changes include:
-   `metametrics`: included `desktop_enabled` in traits to identify whenever the user pairs/unpairs the extension and the desktop app.
-  UI changes to emit desktop-related events whenever a button is clicked
    -   Included a new event category `DESKTOP`

## Manual Testing Steps

- Setup Segment following: https://github.com/MetaMask/metamask-extension/blob/develop/development/README.md#segment
- Build the extension for desktop build type - yarn start --build-type desktop
- Navigate to settings->experimental
- Click on enable desktop button
- Once the desktop is paired, perform a transfer (any key operation)
- Check on the segment if identify was triggered with the `desktop_enabled: true`  
Example:
```
analytics.identify({
  userId: '0x',
  traits: {
    desktop_enabled: true,
  }
});
```

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/app/scripts/controllers/metametrics.js
+++ b/app/scripts/controllers/metametrics.js
@@ -100,7 +100,6 @@ export default class MetaMetricsController {
    * @param {MetaMetricsControllerState} options.initState - State to initialized with
    * @param options.captureException
    */
-  /* eslint-disable-next-line jsdoc/require-param */
   constructor({
     segment,
     preferencesStore,
@@ -111,9 +110,6 @@ export default class MetaMetricsController {
     initState,
     extension,
     captureException = defaultCaptureException,
-    ///: BEGIN:ONLY_INCLUDE_IN(desktop)
-    getDesktopEnabled,
-    ///: END:ONLY_INCLUDE_IN
   }) {
     this._captureException = (err) => {
       // This is a temporary measure. Currently there are errors flooding sentry due to a problem in how we are tracking anonymousId
@@ -129,9 +125,6 @@ export default class MetaMetricsController {
       environment === 'production' ? version : `${version}-${environment}`;
     this.extension = extension;
     this.environment = environment;
-    ///: BEGIN:ONLY_INCLUDE_IN(desktop)
-    this.getDesktopEnabled = getDesktopEnabled;
-    ///: END:ONLY_INCLUDE_IN
 
     const abandonedFragments = omitBy(initState?.fragments, 'persist');
     const segmentApiCalls = initState?.segmentApiCalls || {};
@@ -477,9 +470,6 @@ export default class MetaMetricsController {
           locale: this.locale,
           chain_id: this.chainId,
           environment_type: environmentType,
-          ///: BEGIN:ONLY_INCLUDE_IN(desktop)
-          desktop_paired: this.getDesktopEnabled(),
-          ///: END:ONLY_INCLUDE_IN
         },
         context: this._buildContext(referrer, page),
       });
@@ -679,9 +669,6 @@ export default class MetaMetricsController {
         locale: this.locale,
         chain_id: properties?.chain_id ?? this.chainId,
         environment_type: environmentType,
-        ///: BEGIN:ONLY_INCLUDE_IN(desktop)
-        desktop_paired: this.getDesktopEnabled(),
-        ///: END:ONLY_INCLUDE_IN
       },
       context: this._buildContext(referrer, page),
     };
@@ -729,6 +716,9 @@ export default class MetaMetricsController {
       [TRAITS.THREE_BOX_ENABLED]: false, // deprecated, hard-coded as false
       [TRAITS.THEME]: metamaskState.theme || 'default',
       [TRAITS.TOKEN_DETECTION_ENABLED]: metamaskState.useTokenDetection,
+      ///: BEGIN:ONLY_INCLUDE_IN(desktop)
+      [TRAITS.DESKTOP_ENABLED]: metamaskState.desktopEnabled || false,
+      ///: END:ONLY_INCLUDE_IN
     };
 
     if (!previousUserTraits) {
--- a/app/scripts/controllers/metametrics.js
+++ b/app/scripts/controllers/metametrics.js
@@ -100,6 +100,7 @@ export default class MetaMetricsController {
    * @param {MetaMetricsControllerState} options.initState - State to initialized with
    * @param options.captureException
    */
+  /* eslint-disable-next-line jsdoc/require-param */
   constructor({
     segment,
     preferencesStore,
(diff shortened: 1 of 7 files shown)


vinistevam:

app/scripts/controllers/metametrics.js
```diff
@@ -669,6 +679,9 @@ export default class MetaMetricsController {
         locale: this.locale,
         chain_id: properties?.chain_id ?? this.chainId,
         environment_type: environmentType,
+        ///: BEGIN:ONLY_INCLUDE_IN(desktop)
+        desktop_paired: this.getDesktopEnabled(),
+        ///: END:ONLY_INCLUDE_IN
```
brad-decker, line 684: desktop_paired seems more like a user trait then a property we want on *every event*. @worldlyjohn thoughts?

ui/pages/desktop-error/render-desktop-error.js
```diff
@@ -43,6 +45,12 @@ export function renderDesktopError({
   };
 
   const renderHeader = (text) => {
+    if (typeof trackEvent === 'function') {
+      trackEvent({
+        category: EVENT.CATEGORIES.ERROR,
+        event: `Error: ${text}`,
+      });
```
brad-decker, line 52: Why are we tracking errors in segment?

ui/pages/desktop-error/render-desktop-error.js
```diff
@@ -64,7 +72,23 @@ export function renderDesktopError({
...
+          type="primary"
+          onClick={() => {
+            if (onClick) {
+              onClick();
+              if (typeof trackEvent === 'function') {
+                trackEvent({
+                  category: EVENT.CATEGORIES.DESKTOP,
+                  event: `${text} Button Clicked`,
+                });
+              }
```
brad-decker, line 85: These types of dynamic event names are *not* acceptable....They end up creating variadic events that make it very hard to derive meaning from. @worldlyjohn will advise you on if we want this event or not and if so how to add properties to cover the use cases.

brad-decker (changes requested):

app/scripts/controllers/metametrics.js
vinistevam: Thanks for reviewing my PR :bow:
If more appropriate I can move it to user traits.

ui/pages/desktop-error/render-desktop-error.js
vinistevam: The goal here was to measure the errors while using the desktop app as a companion but you are right, It is already collected by Sentry and if necessary in Segment, it can be tracked by using the `trackPage` method. I will remove it and push the changes.

ui/pages/desktop-error/render-desktop-error.js
vinistevam: One quick update on this case. I pushed some changes so that instead of using the text it uses the id of the button to determine the event name. In case, more properties need to be added it's possible to extend this function to also include specific properties for each event.

ui/pages/desktop-error/render-desktop-error.js
```diff
@@ -61,10 +74,57 @@ export function renderDesktopError({
     );
   };
 
+  const getEventNameById = (id) => {
```
worldlyjohn, line 77: none of these events follow our naming conventions defined here: https://www.notion.so/Naming-Conventions-95bf27517033451888053f7ab654d327

ui/pages/desktop-error/render-desktop-error.js
```diff
@@ -61,10 +74,57 @@ export function renderDesktopError({
...
-        <Button type="primary" onClick={onClick ?? noop} id={id}>
+        <Button
+          type="primary"
+          onClick={() => {
+            if (onClick) {
+              onClick();
+              if (typeof trackEvent === 'function') {
+                trackEvent({
+                  category: EVENT.CATEGORIES.DESKTOP,
+                  event: `${getEventNameById(id)} Button Clicked`,
```
worldlyjohn, line 119: using dynamic event names is frowned upon as it will create N new tables.  can this be a single event name and use dynamic id (if truely necessary) as a property instead? (though even that is less than ideal, better if you run this through a defined list of possible values)

ui/components/app/desktop-enable-button/desktop-enable-button.component.js
```diff
@@ -40,6 +43,10 @@ export default function DesktopEnableButton() {
     if (desktopEnabled) {
       await disableDesktop();
       setDesktopEnabled(false);
+      trackEvent({
+        category: EVENT.CATEGORIES.DESKTOP,
+        event: `Disable Desktop Button Clicked`,
```
worldlyjohn, line 48: instead of separate events for enable/disabled, these should be properties of the same event.  another option is that this is actually a setting change.

either of these works:
```
event name: Desktop Button Clicked
properties: {enabled: true|false}
```
or
```
event name: Settings Updated
properties: {desktop_enabled: true|false}
```

ui/pages/desktop-error/render-desktop-error.js
worldlyjohn: id also not sufficient as it's easy for an element id to change and unintended consequence is the event is now changed too.  usually when you want to do this, the best practice is a more generic event name and using this value as a property (though the id field should probably be validated that it matches one element in a predefined set)

ui/components/app/desktop-enable-button/desktop-enable-button.component.js
vinistevam: I've opted for the settings change as you suggested:
```
event name: Settings Updated
properties: {desktop_enabled: true|false}
```

ui/pages/desktop-error/render-desktop-error.js
vinistevam: :pray:  Great thanks!
I changed it to a single event `Desktop Button Clicked` and also included an additional parameter in the `renderCTA` that will be the value of the property to identify the CTA performed by the user.

ui/components/app/desktop-enable-button/desktop-enable-button.component.js
```diff
@@ -78,9 +89,19 @@ export default function DesktopEnableButton() {
       return;
     }
 
+    trackEvent({
+      category: EVENT.CATEGORIES.DESKTOP,
+      event: 'Desktop Button Clicked',
```
worldlyjohn, line 94: do you need to fire an event on the button when you are already firing an event on the state change `DESKTOP_UPDATE_SETTINGS_EVENT`?  seems redundant, so i'd suggest removing the click event here (which is typically used when there is no other event being triggered), but i dont have a _**strong**_ opinion here. :)

worldlyjohn:
works for me!

brad-decker (approved):

cryptotavares (approved):

PR #17894, opened 2023-02-24
[MMI] Added code fences to the following controllers: app-state, metametrics, preferences

albertolive:
## Explanation

This task is to add code fences to the following controllers: app-state, metametrics and preferences

## Pre-merge author checklist

- [X] I've clearly explained:


DigitalsupportTeam:
We'd be happy to look into this for you
kindly file a support ticket below 

zone-live (dismissed):

app/scripts/controllers/preferences.js
```diff
@@ -505,6 +534,19 @@ export default class PreferencesController {
     return this.store.getState().disabledRpcMethodPreferences;
   }
 
+  ///: BEGIN:ONLY_INCLUDE_IN(build-mmi)
+  async prepareMmiPortfolio() {
+    if (!process.env.IN_TEST) {
+      this.handleMmiPortfolio().then((mmiDashboardData) => {
```
Gudahtt, line 540: Nit: `async/await` is preferred over `.then`/`.catch` because it produces better stack traces (with `.then` the stack would end here)

app/scripts/controllers/app-state.js
```diff
@@ -374,6 +374,25 @@ export default class AppStateController extends EventEmitter {
...
+   * Set the interactive replacement token with a url and the old refresh token
+   *
+   * @param {object} opts
+   * @param opts.url
+   * @param opts.oldRefreshToken
+   * @returns {void}
+   */
+  showInteractiveReplacementTokenBanner({ url, oldRefreshToken }) {
+    this.store.updateState({
+      interactiveReplacementToken: {
```
Gudahtt, line 388: What is this state responsible for? I'm not familiar with interactive replacement tokens.

app/scripts/controllers/metametrics.js
```diff
@@ -530,10 +534,12 @@ export default class MetaMetricsController {
...
       }
 
-      const combinedProperties = merge(
-        payload.sensitiveProperties,
-        payload.properties,
-      );
+      const combinedProperties = merge(payload.sensitiveProperties, {
+        ...payload.properties,
+        ///: BEGIN:ONLY_INCLUDE_IN(build-mmi)
+        accountAddress: this.selectedAddress,
```
Gudahtt, line 540: Is it critical that this be tracked in MMI? This is personally identifiable information. Generally we don't track that.

This conflicts with our plans in the near term to get rid of the concept of a "globally selected address" as well.

app/scripts/controllers/metametrics.js
```diff
@@ -658,6 +688,15 @@ export default class MetaMetricsController {
       referrer,
       environmentType = ENVIRONMENT_TYPE_BACKGROUND,
     } = rawPayload;
+
+    ///: BEGIN:ONLY_INCLUDE_IN(build-mmi)
+    const mmiProps = {};
+
+    if (this.extension?.runtime?.id) {
+      mmiProps.extensionId = this.extension.runtime.id;
```
Gudahtt, line 696: Isn't this a static value? Seems odd to track that as a metric property

Gudahtt (commented):
We're working on sharing controllers between extension and mobile, so ideally we'd keep code fences out of controllers in general (because `core` doesn't support code fencing).  We can include them as a temporary measure if we need to, but finding an alternative would be preferable if possible.

app/scripts/controllers/app-state.js
albertolive: This allows users to replace expired custodian refresh tokens automatically. When `interactiveReplacementToken` has the `url` and the `oldRefreshToken`, we show that little banner on top of the extension. More info: [url](https://consensyssoftware.atlassian.net/wiki/spaces/PRDC/pages/398657881166/05.+Token+replacement)

image

app/scripts/controllers/preferences.js
albertolive: You're right, I'll change it

app/scripts/controllers/app-state.js
Gudahtt: Makes sense, thanks!

app/scripts/controllers/metametrics.js
zone-live: @shane-t can you please advise on how critical it is for us having this address? It has been added [in here](https://github.com/ConsenSys/mmi-extension/pull/340). Maybe we can remove it for now and try to figure out if we can get it from other place?

app/scripts/controllers/preferences.js
```diff
@@ -69,12 +72,25 @@ export default class PreferencesController {
...
     this._onInfuraIsUnblocked = opts.onInfuraIsUnblocked;
     this.store = new ObservableStore(initState);
     this.store.setMaxListeners(13);
     this.tokenListController = opts.tokenListController;
 
+    ///: BEGIN:ONLY_INCLUDE_IN(build-mmi)
+    this.handleMmiPortfolio = opts.handleMmiPortfolio;
+
+    if (!process.env.IN_TEST) {
+      this.mmiConfigurationStore = opts.mmiConfigurationStore.getState();
```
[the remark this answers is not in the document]
zone-live: True we can add documentation to these 2 methods, also we have this MMI Portfolio logic `(handleMmiPortfolio)` in here because every time an address gets added, removed or they are synced, the extension needs to tell that to the portfolio app, and since those operations are done here, we call this logic in here as well. But we can improve the naming of the method to better fit this controller semantic.

app/scripts/controllers/metametrics.js
shane-t: Let's remove this @zone-live 
While we have a separate user agreement in MMI, we're not dependent on this for our metrics, and it would introduce ambiguity if someone is reading the codebase. Moreover, if the concept of a global selected address is being deprecated, it will only cause issues in the future.

app/scripts/controllers/metametrics.js
Gudahtt: Makes sense!

In that case, maybe we could establish a way to inject static build-specific values like this into the constructor, rather than embedding them directly inside the controller using a code fence. That would avoid the need for fences within the controller.

Not critical though, we can explore that in a later PR.

zone-live:
> We're working on sharing controllers between extension and mobile, so ideally we'd keep code fences out of controllers in general (because `core` doesn't support code fencing). We can include them as a temporary measure if we need to, but finding an alternative would be preferable if possible.

As a proposal @Gudahtt we can do this (the logic we have in preferences controller) by adding an event where needed (setAddresses, removeAddress, sync) like `this.hub.emit('something-happened')`, this would be kind of a side effect, it would not require a code fence and wouldn't bother the preferences code when migrated to the Core. We would then listen the this event in our own mmi controller and move this logic there.
Wdyt about this approach? Could we merge this as is for now, but have a follow up task with this solution to do it, before this gets to core, obviously. Whats the timeline on that btw?

app/scripts/controllers/metametrics.js
albertolive: `accountAddress` removed from the code

zone-live (approved):

DDDDDanica (approved):

PR #17917, opened 2023-02-27
Added back icon for snaps

NidhiKJha:
This PR is to update the back icon on the settings page on small screens

Screenshot 2023-02-28 at 1 06 12 AM

### DIFF
--- a/ui/pages/settings/settings.component.js
+++ b/ui/pages/settings/settings.component.js
@@ -29,11 +29,10 @@ import { getSettingsRoutes } from '../../helpers/utils/settings-search';
 import AddNetwork from '../../components/app/add-network/add-network';
 import {
   Icon,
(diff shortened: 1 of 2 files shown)


ui/pages/settings/settings.component.js
```diff
@@ -135,11 +134,11 @@ class SettingsPage extends PureComponent {
         <div className="settings-page__header">
           <div className="settings-page__header__title-container">
             {currentPath !== SETTINGS_ROUTE && (
-              <ButtonIcon
-                ariaLabel={t('back')}
+              <Icon
+                className="settings-page__back-button"
```
NidhiKJha, line 138: Ideally, that shouldn't happen but with ButtonIcon, I am unable to pass the color prop to the ARROW. Hence, I changed it to Icon

PR #17922, opened 2023-02-28
Removed a feature flag for the transacation security provider

### LINKED ISSUE #17460: 🏁 Remove transaction security provider feature flag
## Background
Follow up for when the feature is completed. These 3 issues need be closed to unblock this:

- [ ] https://github.com/MetaMask/metamask-extension/issues/16258
- [ ] https://github.com/MetaMask/metamask-extension/issues/16262
- [ ] https://github.com/MetaMask/metamask-extension/issues/16636

## Acceptance Criteria
1. The toggle is now visible to all users under Settings > Experimental
2. The What's new modal is displayed to users
3. When toggle is on, whenever the users receive transaction or signature request, data is shared with and received from the security provider and a warning is shown to the user if the security provider flagged the request as malicious.

filipsekulic:
## Explanation
Removed a feature flag for the `transacation security provider`.
* Fixes #17460

### After

https://user-images.githubusercontent.com/92310504/221795168-d99efc2e-cffd-4cbf-9410-528bec0f05fe.mov

## Manual Testing Steps
Before starting the app, in the `.metamaskrc` file remove the feature flag or do not set the it to 1 - `TRANSACTION_SECURITY_PROVIDER=`.

1. Go to `Settings` > `Experimental`
2. Transaction security provider is now being shown by default

### DIFF
--- a/ui/pages/settings/settings.component.js
+++ b/ui/pages/settings/settings.component.js
@@ -78,17 +78,6 @@ class SettingsPage extends PureComponent {
     searchText: '',
   };
 
-  shouldRenderExperimentalTab() {
-    ///: BEGIN:ONLY_INCLUDE_IN(desktop)
-    const desktopAvailable = true;
-    if (desktopAvailable) {
-      return true;
-    }
-    ///: END:ONLY_INCLUDE_IN
-
-    return process.env.TRANSACTION_SECURITY_PROVIDER || process.env.NFTS_V1;
-  }
-
   componentDidMount() {
     this.handleConversionDate();
   }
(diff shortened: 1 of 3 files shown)


dzfjz:
Verified by QA

vthomas13 (approved):

ui/pages/settings/settings.component.js
```diff
@@ -78,17 +78,6 @@ class SettingsPage extends PureComponent {
...
-    ///: BEGIN:ONLY_INCLUDE_IN(desktop)
-    const desktopAvailable = true;
-    if (desktopAvailable) {
-      return true;
-    }
-    ///: END:ONLY_INCLUDE_IN
-
-    return process.env.TRANSACTION_SECURITY_PROVIDER || process.env.NFTS_V1;
-  }
-
```
jpuri, line 91: I think 

```
///: BEGIN:ONLY_INCLUDE_IN(desktop)
    const desktopAvailable = true;
    if (desktopAvailable) {
      return true;
    }
    ///: END:ONLY_INCLUDE_IN
```
Is important and experimental tab should be rendered in desktop app.

jpuri (commented):
We need to show experimental tab in desktop app https://github.com/MetaMask/metamask-extension/pull/17922/files#r1124831198

ui/pages/settings/settings.component.js
filipsekulic: I see what you mean, but with this PR `experimental` tab will be shown to the users by default in any case. The reason for its removal is the fact that just removing the feature flag for the `TRANSACTION_SECURITY_PROVIDER` won't show the `OpenSea security provider` in the `experimental` tab, unless this conditional rendering gets removed.

jpuri (approved):

PR #17969, opened 2023-03-02
Integrate Snow with LavaMoat scuttling protection

weizman:
This PR follows up on https://github.com/LavaMoat/LavaMoat/pull/462 which allows exporting the scuttling security feature to an outside scuttler (as an experimental feature ⚠️ ).

This in fact allows us in MetaMask to perform scuttling not only to the top main realm, but to all potential child realms using Snow.

In other words, if an attacker wants to access `alert` but can't because the `alert` property is scuttled, the attacker could have opened an iframe (aka new realm) and extract the `alert` function from the global object of that iframe.

With this PR, Snow will automatically apply the scuttling mechanizm to all new potential child realms, including the iframe described in the example, thus leaving the attacker unable to get a hold of an `alert` function instance.

The `alert` example is merely a demonstration of course, but scuttling all windows recursively brings an important security benefit to MetaMask (will be elaborated in the future).


digiwand (dismissed):

digiwand (approved):
^ nvm,  LGTM! 🙌🏼

Gudahtt (approved):
LGTM!

PR #17741, opened 2023-02-14
NFT: Replaced all the instances of collectibles with NFTs

NidhiKJha:
Fixes: #14502

We're moving from using the term "collectible" to always using "NFT". This PR updates the associated variables and instances of collectibles to NFT.

TODO: Update the `TestDappCollectibles` change once this [PR](https://github.com/MetaMask/test-dapp/pull/199) is merged.

## Pre-merge author checklist

- [ ] I've clearly explained:
  - [ ] What problem this PR is solving
  - [ ] How this problem was solved
  - [ ] How reviewers can test my changes
- [ ] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [ ] Manual testing (e.g. pull and build branch, run in the browser, test code being changed)
- [ ] PR is linked to the appropriate GitHub issue
- [ ] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.

In this case, a QA Engineer approval will be be required.

PR #17980, opened 2023-03-03
Change migration 78 version to 79, 79 to 80 and 80 to 78 (so that the…

danjm:
… new 78 can be picked into an RC)

We have this challenge:

> [v10.26](https://github.com/MetaMask/metamask-extension/pull/17698) is currently in progress and we are hoping to get it merged and release soon.
We would like to ship the [Phishing Controller](https://github.com/MetaMask/metamask-extension/pull/17835) update as part of v10.26, the Phishing Controller update contains migration 80, whilst v10.26 is currently on migration 77.
How should I approach resolving conflicts, considering I do not want to include either of the following;
- Collectible Controller gets updated to NFT Controller ([migration 78](https://github.com/MetaMask/metamask-extension/pull/17741))
- showPortfolioTooltip gets deleted ([migration 79](https://github.com/MetaMask/metamask-extension/pull/17722)), the portfolio tooltip is being updated again in v10.27

And the solution is:
> because all of these are only on develop  and not master, we might just be able to make a PR to develop that changes the phishing controller migration to 78 and bumps the other two by 1

Note to reviewers: the only changes I made in code were to file names and version numbers within the migration files. Any other code changes in the diff are a result of simultaneously changing files names of each file to one of the other existing filed

### DIFF
--- a/app/scripts/migrations/078.test.js
+++ b/app/scripts/migrations/078.test.js
@@ -1,173 +1,108 @@
-import migration78 from './078';
+import { migrate, version } from './078';
 
 describe('migration #78', () => {
-  it('should update the version metadata', async () => {
+  it('updates the version metadata', async () => {
     const oldStorage = {
       meta: {
         version: 77,
       },
+      data: {},
     };
 
-    const newStorage = await migration78.migrate(oldStorage);
+    const newStorage = await migrate(oldStorage);
+
     expect(newStorage.meta).toStrictEqual({
-      version: 78,
+      version,
     });
   });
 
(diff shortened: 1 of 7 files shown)


Gudahtt:
You'll need to update the import of this migration as well; the TypeScript one was formatted differently

danjm:
> You'll need to update the import of this migration as well; the TypeScript one used named exports rather than an object default export.

Done in most recent commit

Gudahtt (approved):
LGTM!

pedronfigueiredo (approved):

PR #17985, opened 2023-03-05
bump snow 1.5.0

weizman:
https://github.com/LavaMoat/snow/pull/66

PR #17986, opened 2023-03-05
Avoid blob url for files downloads

weizman:
* In order to allow the user to download log reports, we use a function called `exportAsFile`
* `exportAsFile` creates a URL object from a Blob and downloads that (common trick)
* The combination of URL object and Blob is dangerous in realms security and therefore is [forbidden by new Snow version](https://github.com/LavaMoat/snow/pull/69) because it's very easy to use it to bypass Snow
* In order to migrate into the new Snow version ([PR](https://github.com/MetaMask/metamask-extension/pull/17985)), I propose this change to the `exportAsFile` function, which will achieve the exact same goal but without the dangerous combo, but rather with `data:` URI instead
* (This modification is covered by tests in CI which proves it works 👍)


digiwand (dismissed):

ui/helpers/utils/export-utils.js
```diff
@@ -1,19 +1,23 @@
 import { getRandomFileName } from './util';
```
kumavis, line 2: prefer async over callbacks

ui/helpers/utils/export-utils.js
```diff
@@ -1,19 +1,23 @@
...
-  if (window.navigator.msSaveOrOpenBlob) {
-    window.navigator.msSaveBlob(blob, filename);
-  } else {
+  encodeB64(data, (b64) => {
+    // eslint-disable-next-line no-param-reassign
+    filename = filename || getRandomFileName();
     const elem = window.document.createElement('a');
-    elem.target = '_blank';
-    elem.href = window.URL.createObjectURL(blob);
+    elem.href = `data:${type};Base64,${b64}`;
```
kumavis, line 17: can we just use `btoa`? https://developer.mozilla.org/en-US/docs/Web/API/btoa

weizman: unfortunately `btoa` fails on special chars that we have in the exported json files that `TextEncoder` successfully deals with (I learned the hard way)

ui/helpers/utils/export-utils.js
weizman: 6de68ce1d29a66f8031fbcd65ac9f280debe6cf1

ui/helpers/utils/export-utils.js
kumavis: interesting -- what about `Buffer.from(xyz, 'utf8').toString('base64')`

weizman: e95a668f0

ui/helpers/utils/export-utils.js
```diff
@@ -1,19 +1,13 @@
 import { getRandomFileName } from './util';
 
-export function exportAsFile(filename, data, type = 'text/csv') {
+export async function exportAsFile(filename, data, type = 'text/csv') {
```
kumavis, line 3: no longer needs to be async

weizman: 6e13ea84285f0db5eb268ef52bab6a2b5cec49b3

kumavis (approved):

digiwand (approved):

PR #17995, opened 2023-03-06
Improve downloads tests

weizman:
Currently, the two tests that test file downloads `backup-restore` and `state-logs` complete when a file was successfully downloaded - this PR makes sure they complete once the file is downloaded but also is a proper non-corrupted JSON file


digiwand (approved):
👍

albertolive (approved):

Issue #17716, opened 2023-02-10
[Bug]: No result when search for 'nonce' in settings' search bar


chloeYue:
### Describe the bug

In settings' search bar, when type `nonce` to try to find the "customize transaction nonce" setting, no result is shown.
(Note when we enter `customize` or `transaction`, we are able find the setting option correctly)

https://user-images.githubusercontent.com/105063779/218135925-14c16c00-8a45-4e0e-96de-253a924a003a.mov

### Steps to reproduce

1. Login, go to settings page.
2. Type `nonce` in the search bar on top of the page, check if the `customize transaction nonce` setting option is shown.
Expected:
`Customize transaction nonce` setting option should be shown when we enter `nonce` in search bar.

### Version

10.26.0

### Build type

None

### Browser

Chrome

### Operating system

MacOS

PR #17999, opened 2023-03-06
fix(17716): increase threshold to include all options for nonce search

DDDDDanica:
## Explanation

Fix: https://github.com/MetaMask/metamask-extension/issues/17716

[Threshold](https://fusejs.io/api/options.html#threshold) is to tell the algorithm when to give up searching, while "0.0 requires a perfect match (of both letters and location) and 1.0 would match anything". 
The reason behind this bug is we have more than 1 matches for "nonce" and the algorithm stopped when only finding one match.
The fix is straightforward, increasing the threshold and return more results.

### After

Screenshot 2023-03-06 at 13 58 09

## Pre-merge author checklist

- [ ] I've clearly explained:

### DIFF
--- a/ui/pages/settings/settings-search/settings-search.js
+++ b/ui/pages/settings/settings-search/settings-search.js
@@ -34,7 +34,7 @@ export default function SettingsSearch({
   ///: END:ONLY_INCLUDE_IN
   const settingsSearchFuse = new Fuse(settingsRoutesListArray, {
     shouldSort: true,
-    threshold: 0.2,
+    threshold: 0.3,
     location: 0,
     distance: 100,
     maxPatternLength: 32,


albertolive (approved):

ui/pages/settings/settings-search/settings-search.js
```diff
@@ -34,7 +34,7 @@ export default function SettingsSearch({
   ///: END:ONLY_INCLUDE_IN
   const settingsSearchFuse = new Fuse(settingsRoutesListArray, {
     shouldSort: true,
-    threshold: 0.2,
+    threshold: 0.3,
```
digiwand, line 37: Q: where did we get 0.3 from? I wonder why we've lowered the threshold from the default value of 0.6

chloeYue:
QA test OK, thanks !

ui/pages/settings/settings-search/settings-search.js
DDDDDanica: Thanks for reviewing ! As 0 is the perfect match and we went for 0.2, so 0.3 is the next best matching algorithm .
For the default value 0.6, this is the result and it is too vague for us :( 

Screenshot 2023-03-06 at 15 15 13

Screenshot 2023-03-06 at 15 16 39


I tested also for single match and multiple results (network for example),0.3 seems to be a good value to give.

pedronfigueiredo (approved):

PR #18010, opened 2023-03-07
Override localstorage lavamoat policy

weizman:
`localStorage` is not accessible if running LM inside a sandboxed frame (discovered by @FrederikBolding in context of snaps).
These two libraries are trying to use `localStorage` but they can manage without it, and if are not endowed, LM doesn't crash


weizman:
irrelevant for now, might reopen later

PR #18018, opened 2023-03-07
Fix for changing to Goerli network causes insufficient error despite having enough funds

### LINKED ISSUE #17002: [Bug]: Changing to Goerli network causes insufficient error despite having enough funds
### Describe the bug

After changing the network and with sufficient funds, I see the insufficient error and cannot send tokens. I am observing this on Goerli. It does not appear to happen with the Sepolia network
Screen Shot 2022-12-20 at 04 01 11
Screen Shot 2022-12-20 at 03 59 12

### Steps to reproduce

prerequisite: have enough funds on Goerli for a transaction and gas fees

1. Log in to extension
2. Start with the Ethereum mainnet network
3. Click the Send button on the overview page
4. Get to the "Send" page
5. Change to Goerli Network
6. Observe insufficient error appears despite having sufficient funds

_No response_

### Version

10.22.0 and possibly older versions

### Build type

None

### Browser

Chrome

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

aleksandar-mihajlovic:
## Explanation

Added fix, so now when user starts Send process with Ethereum network and switch to Goerli balance is shown correctly, it doesn't show insufficient error and user is able to send tokens if he got balance greater then 0. 
* Fixes #17002

### Before

changhingNetworkBug

### After

changhingNetworkFix

## Unit tests

Screenshot 2023-03-07 at 11 43 58

## Lint

Screenshot 2023-03-07 at 11 34 42

## Manual Testing Steps

1. Log in to extension
2. Start with the Ethereum mainnet network
3. Click the Send button on the overview page
4. Get to the "Send" page
5. Change to Goerli Network
6. Insufficient error is not appearing anymore when balance is greater then 0


dzfjz:
Verified by QA

ui/store/actions.ts
```diff
@@ -1837,6 +1842,9 @@ export function updateMetamaskState(
       type: actionConstants.UPDATE_METAMASK_STATE,
       value: newState,
     });
+    if (!isEqual(oldSelectedAccountBalance, newSelectedAccountBalance)) {
+      dispatch(initializeSendState({ chainHasChanged: true }));
+    }
     if (provider.chainId !== newProvider.chainId) {
```
brad-decker, line 1848: This is the line where we are already checking if the chainId has changed.

brad-decker (changes requested):
There is more to this and adding another call to initializeSendState is not necessarily the answer. Lets dig in a little more to understand this and come back with findings.

brad-decker:
@digiwand @jpuri looks like both of you were involved in the bug report. If you have advised @aleksandar-mihajlovic on this path to fixing could you weigh in. If not see if you agree with my assertion based on the logic in the actions file around where the change was made (namely that the change being suggested and the bug as presented do not align)

ui/store/actions.ts
aleksandar-mihajlovic: Well for some reason the state is not being updated, specifically account balance, so that is why I added when balances are not equal when they should be, update the state and I was advised by @danjm .

danjm: I am pretty sure what is happening here is a race condition where it takes longer for the balance to update than the network. So at the time of the `dispatch(initializeSendState({ chainHasChanged: true }));` that Brad is referring to, the selected account balance has not been updated.

My recommendation here is that we create a different parameter `balanceHasChanged` and add appropriate handling within the send duck. While this will handle the race condition, it also handles the case where the user's balance changes while they are on the send screen.

So we need the above line to be `dispatch(initializeSendState({ balanceHasChanged: true }));`

bschorchit:
https://github.com/MetaMask/metamask-extension/pull/18086 might make this fix un-necessary. Some more reasoning on why we recently decided to prevent users from changing networks during a send flow here https://github.com/MetaMask/metamask-extension/issues/18045#issuecomment-1460467835. Open to feedback on this decision.

ui/store/actions.ts
aleksandar-mihajlovic: Okay, no problem I could implement that, but should i proceed with this or should I wait for that discussion that Barbara mentioned in the comments at the end? @danjm @bschorchit

bschorchit:
Closing as not needed as this https://github.com/MetaMask/metamask-extension/pull/18086 has been merged

Issue #17708, opened 2023-02-10
[e2e] Send ETH to an ENS Domain


seaona:

gaavar:
Hi! Can I work on this issue? 🙂

PR #18056, opened 2023-03-08
[e2e] Check ENS resolves correctly

seaona:
## Explanation
In this testcase we are checking that the ENS domain resolves to the correct address. This is the first PR for ENS and work should be extended in subsequent PRs (checking subdomains, updating recipient, complete Send flow, etc.).

### Context
For fetching and resolving correctly an ENS address, we perform a couple of calls to the Ethereum node, in order to call the resolver contract function, which resolves to the correct address and returns it in the response.

### How is it done
There are some challenges in order to test the ENS resolution, which have been overcome as follows:
- ENS is not supported on localhost
- Localhost cannot be mocked easily at the moment with `mockttp`. For this reason here we are targeting Ethereum mainnet
- Since `mock-e2e` blacklist Infura by default, we need to reset the mocks by using:
```
await mockServer.reset();
await mockServer.forAnyRequest().thenPassThrough();
```
- Now, we are able to mock the necessary requests to Infura. These are:
   - `eth_blockNumnber`: for some reason, we need to have a correct response in order to fire the next requests. If this request is not returned, we are not able to proceed
   - `eth_call`: here we will return the response with the resolved address `0x1111111111111111111111111111111111111111`
- The setup is then ready for checking the ENS resolution on a send flow

* Fixes partially https://github.com/MetaMask/metamask-extension/issues/17708

## Screenshots/Screencaps

https://user-images.githubusercontent.com/54408225/223742667-c3356bda-b6b7-4f73-b6ba-409896594ff4.mp4

## Manual Testing Steps
`yarn build:test`
`yarn test:e2e:single test/e2e/tests/ens.spec.js --browser=chrome --leave-running=true`

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/test/e2e/tests/ens.spec.js
+++ b/test/e2e/tests/ens.spec.js
@@ -1,0 +1,102 @@
+const { strict: assert } = require('assert');
+const { convertToHexValue, withFixtures } = require('../helpers');
+const FixtureBuilder = require('../fixture-builder');
+
+describe('ENS', function () {
+  const sampleAddress = '1111111111111111111111111111111111111111';
+  const sampleEnsDomain = 'test.eth';
+  const infuraUrl =
+    'https://mainnet.infura.io/v3/00000000000000000000000000000000';
+
+  async function mockInfura(mockServer) {
+    await mockServer.reset();
+    await mockServer.forAnyRequest().thenPassThrough();
+    await mockServer
+      .forPost(infuraUrl)
+      .withJsonBodyIncluding({ method: 'eth_blockNumber' })
+      .thenCallback(() => {
+        return {
+          statusCode: 200,
+          json: {
+            jsonrpc: '2.0',
+            id: '1111111111111111',
+            result: '0x1',
+          },
+        };
+      });
+
+    await mockServer
+      .forPost(infuraUrl)
+      .withJsonBodyIncluding({ method: 'eth_call' })
+      .thenCallback(() => {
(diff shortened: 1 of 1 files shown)


jpuri (approved):

test/e2e/tests/ens.spec.js
```diff
@@ -0,0 +1,102 @@
...
+        testSpecificMock: mockInfura,
+      },
+      async ({ driver }) => {
+        await driver.navigate();
+        await driver.fill('#password', 'correct horse battery staple');
+        await driver.press('#password', driver.Key.ENTER);
+
+        await driver.waitForElementNotPresent('.loading-overlay');
+        await driver.clickElement('.network-display');
+        await driver.clickElement({ text: 'Ethereum Mainnet', tag: 'span' });
```
PeterYinusa, line 67: Can we move this setup to the fixtures using the NetworkController

PeterYinusa (approved):
Small nit, but otherwise this is great! Thank you @seaona

seaona:
Thank you for pointing this @PeterYinusa . I took the manual approach since at the moment, changing the network by injecting the state on the network controller only has effect on the UI, but not the background requests.
I suspect there is some piece of code missing/incomplete, that causes this, making it possible for UI and state not being in sync with background. In that case, we might need to investigate this separately/open an issue for that.

PR #18071, opened 2023-03-09
[MMI] Added code fencing in transaction list

albertolive:
## Explanation

Added code fencing in transaction list

https://consensyssoftware.atlassian.net/browse/MMI-2586

## Pre-merge author checklist

- [X] I've clearly explained:


ui/components/app/transaction-list-item-details/transaction-list-item-details.component.js
```diff
@@ -123,8 +135,26 @@ export default class TransactionListItemDetails extends PureComponent {
     });
   };
 
-  componentDidMount() {
-    const { recipientAddress, tryReverseResolveAddress } = this.props;
+  async componentDidMount() {
```
brad-decker, line 138: Let's not make this function async. React will not "await" it. Instead, use a regular promise resolution pattern. In addition you should ensure that the component is still mounted prior to calling this.setState. There is no promise that the component will be mounted when your promise resolves, no matter how likely the scenario is, you might end up with situations where you are attempting to setState on an unmounted component. In a class component this is as easy as adding an instance variable this._mounted = true, in componentDidMount, and setting it to false in componentWillUnmount. When the promise resolves you check if its still true before call this.setState.

ui/components/app/transaction-list-item/index.scss
```diff
@@ -58,7 +58,7 @@
   }
 
   .transaction-status-label::after {
-    content: "·";
+    content: '·';
```
brad-decker, line 61: Why this change?

brad-decker (changes requested):

ui/components/app/transaction-list-item/index.scss
zone-live: I thinks it's Albert local prettier doing this on it's own 😅 
@albertolive can you check?

ui/components/app/transaction-list-item-details/transaction-list-item-details.component.js
zone-live: Yap, well noticed, I'll address this change 👍🏼

ui/components/app/transaction-list-item-details/transaction-list-item-details.component.js
```diff
@@ -135,24 +135,23 @@ export default class TransactionListItemDetails extends PureComponent {
...
 
     ///: BEGIN:ONLY_INCLUDE_IN(build-mmi)
-    if (selectedIdentity && transactionGroup?.primaryTransaction?.custodyId) {
-      const custodyTransactionDeepLink = await getCustodianTransactionDeepLink(
-        selectedIdentity.address,
-        transactionGroup.primaryTransaction.custodyId,
-      );
-      this.setState({ custodyTransactionDeepLink });
+    this._mounted = true;
+    const { address } = selectedIdentity;
```
albertolive, line 150: Could `selectedIdentity` be undefined?

zone-live: You're right, it can, I'll change this 👍🏼

ui/components/app/transaction-list-item/transaction-list-item.component.js
```diff
@@ -165,9 +190,31 @@ function TransactionListItemInner({
...
   ]);
 
-  const showCancelButton = !hasCancelled && isPending && !isUnapproved;
+  let showCancelButton = !hasCancelled && isPending && !isUnapproved;
+
+  ///: BEGIN:ONLY_INCLUDE_IN(build-mmi)
+  showCancelButton = showCancelButton && !isCustodian;
+  const PENDING_COLOR = '#6A737D';
+  const OK_COLOR = '#2F80ED';
+  const FAIL_COLOR = '#D73A49';
```
brad-decker, line 204: Why use these instead of the var(--color-error-default) etc? similar to how TransactionIcon does it?

zone-live (dismissed):

ui/components/app/transaction-list-item/transaction-list-item.component.js
brad-decker: We may change the color of icons based on theme and hardcoding these hex values makes the component brittle to theme changes in the future.

ui/components/app/transaction-list/transaction-list.component.js
```diff
@@ -123,23 +123,36 @@ export default function TransactionList({
...
-                  transactionGroup={transactionGroup}
-                  key={`${transactionGroup.nonce}:${index}`}
-                />
-              ),
-            )}
+            {pendingTransactions
+              ///: BEGIN:ONLY_INCLUDE_IN(build-mmi)
+              .sort(
+                (a, b) => b.primaryTransaction.time - a.primaryTransaction.time,
+              )
```
brad-decker, line 130: Can you describe what this sort is doing so I understand. I'm not sure we don't want this in non MMI :)

ui/components/app/transaction-list/transaction-list.component.js
```diff
@@ -123,23 +123,36 @@ export default function TransactionList({
...
+                  return (
+                    <SmartTransactionListItem
+                      isEarliestNonce={index === 0}
+                      smartTransaction={transactionGroup.initialTransaction}
+                      transactionGroup={transactionGroup}
+                      key={`${transactionGroup.nonce}:${index}`}
+                    />
+                  );
+                }
+                ///: END:ONLY_INCLUDE_IN
```
brad-decker, line 147: Is this exclusion because MMI doesn't support smart transactions?

brad-decker (dismissed):

ui/components/app/transaction-list-item/transaction-list-item.component.js
zone-live: Correct, will update this 👍🏼

ui/components/app/transaction-list/transaction-list.component.js
zone-live: We added this in order to have/see the pending transactions sorted from the most recent to the oldest, in order for the user to see the most recent on top. It's useful when they more than 2 pending transactions. We've also added it to the completed transactions (we've just forgot to add it here). Also to have the completed transactions sorted by recent -> oldest.

ui/components/app/transaction-list/transaction-list.component.js
zone-live: Yes! But I also think that we might not need that code fence, because the user will not be able even start a Smart Transaction... so that IF will always be false I would say.

zone-live (dismissed):

zone-live (dismissed):

ui/components/app/transaction-list-item-details/transaction-list-item-details.component.js
```diff
@@ -281,6 +361,20 @@ export default class TransactionListItemDetails extends PureComponent {
...
+              {
+                ///: BEGIN:ONLY_INCLUDE_IN(build-mmi)
+                transactionNote && transactionNote.length !== 0 && (
+                  <div className="transaction-list-item-details__transaction-breakdown">
+                    <h4 className="transaction-breakdown__title">
+                      {t('transactionNote')}
+                    </h4>
+                    <p className="transaction-breakdown__description">
+                      {transactionNote}
+                    </p>
```
DDDDDanica, line 373: Suggestion: we can use Text component to take advantage of some build-in styles

ui/components/app/transaction-list-item/transaction-list-item.component.js
```diff
@@ -176,7 +224,27 @@ function TransactionListItemInner({
...
+              <TransactionIcon
+                category={category}
+                status={displayedStatusKey}
+              />
+              <Icon
+                name={IconName.Custody}
+                className="transaction-list-item__icon-badge"
+                color={getTransactionColor(status)}
+                width={12}
+                height={12}
```
DDDDDanica, line 239: I don't think we have these attributes exposed via API. but you can try  `size={IconSize.Xs}` to specify the size

ui/components/app/transaction-list-item-details/transaction-list-item-details.component.js
```diff
@@ -281,6 +361,20 @@ export default class TransactionListItemDetails extends PureComponent {
                 primaryCurrency={primaryCurrency}
                 className="transaction-list-item-details__transaction-breakdown"
               />
+              {
+                ///: BEGIN:ONLY_INCLUDE_IN(build-mmi)
+                transactionNote && transactionNote.length !== 0 && (
```
DDDDDanica, line 366: Can we add some snapshot tests for this condition?

ui/components/app/transaction-list/transaction-list.component.js
```diff
@@ -123,23 +123,36 @@ export default function TransactionList({
...
-                />
-              ) : (
-                <TransactionListItem
-                  isEarliestNonce={index === 0}
-                  transactionGroup={transactionGroup}
-                  key={`${transactionGroup.nonce}:${index}`}
-                />
-              ),
-            )}
+            {pendingTransactions
```
DDDDDanica, line 126: Can we add some snapshot tests for this condition?

ui/components/app/transaction-list-item-details/transaction-list-item-details.component.js
albertolive: Fixed

ui/components/app/transaction-list-item-details/transaction-list-item-details.component.js
```diff
@@ -229,6 +285,30 @@ export default class TransactionListItemDetails extends PureComponent {
                   </Button>
                 </Tooltip>
               </div>
+              {
+                ///: BEGIN:ONLY_INCLUDE_IN(build-mmi)
+                custodyTransactionDeepLink &&
+                  custodyTransactionDeepLink.url && (
```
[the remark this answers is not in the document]
albertolive: Added tests

ui/components/app/transaction-list-item/transaction-list-item.component.js
albertolive: Fixed

ui/components/app/transaction-list-item-details/transaction-list-item-details.component.js
albertolive: Added tests

ui/components/app/transaction-list-item/transaction-list-item.component.js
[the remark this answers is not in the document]
albertolive: Fixed

ui/components/app/transaction-list/transaction-list.component.js
albertolive: It wouldn't be helpful in this case. Please let me know if I'm wrong. We are just sorting transactions when it is MMI, just it.

DDDDDanica: If you want to cover, you can assert the items of txn one by one to see if the order is sorted. but it's optional ~ :)

zone-live (approved):

pedronfigueiredo (approved):

pedronfigueiredo (approved):

DDDDDanica (approved):

PR #17811, opened 2023-02-17
UX: Icon: Stop using fa-times

darkwing:
## Explanation

Swaps out all uses of `fa-times` Font Awesome icons.

## Screenshots/Screencaps

TimesMetricsOnboarding
ActionableMessageClose
ImportNFT
TimesCallout
TimeOnboardingNetworks
TimesBetaHeader

## Pre-merge author checklist

- [ ] I've clearly explained:
  - [ ] What problem this PR is solving
  - [ ] How this problem was solved
  - [ ] How reviewers can test my changes
- [ ] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [ ] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
- [ ] PR is linked to the appropriate GitHub issue
- [ ] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.

In this case, a QA Engineer approval will be be required.

PR #18079, opened 2023-03-09
fix network dropdown bug

adonesky1:
I just found this bug [introduced recently](https://github.com/MetaMask/metamask-extension/pull/17811/files#diff-20a630530452c40a35ae16141be0c8ae488615ab5adb2fd4f8dd9b34656d03c3R202) where the network dropdown shows an `X` to delete the network menu option. It should be available for non preloaded networks that are not currently selected. But at the moment it is only available for the currently selected network (which leads to some bad and unexpected UX):

https://user-images.githubusercontent.com/34557516/224077655-1308804c-8c6e-4d2f-8d07-3b2f318f5d2a.mov

after the fix:
https://user-images.githubusercontent.com/34557516/224077960-e0f68e8d-ca29-462e-bf45-4d485e92fae4.mov

### DIFF
--- a/ui/components/app/dropdowns/network-dropdown.js
+++ b/ui/components/app/dropdowns/network-dropdown.js
@@ -199,7 +199,7 @@ class NetworkDropdown extends Component {
           >
             {nickname || rpcUrl}
           </span>
-          {isCurrentRpcTarget && (
+          {isCurrentRpcTarget ? null : (
             <ButtonIcon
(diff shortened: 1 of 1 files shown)


darkwing:
Oh wow, great catch @adonesky1 !

darkwing (approved):

NidhiKJha (approved):

PR #18086, opened 2023-03-09
Disabling network and account changes after the send flow is initiated

### LINKED ISSUE #18045: [Bug]: ENS resolution address not updated when changing networks
### Describe the bug

**Problem**: when I am performing a Send transaction, if I add an ENS domain on the recipient address and then change networks, I can see that the ENS is resolved to the other Network address.
If I proceed, then I can see how in the confirmation screen the old resolution is there, resulting in another domain (if that address is registered by someone else on the other network)

https://user-images.githubusercontent.com/54408225/223661608-6c066856-0ced-455c-9c43-9c8648398f69.mp4

### Steps to reproduce

1. Set Goerli testnet
2. Click Send
3. Add recipient `test.eth`
4. Proceed to the next screen
5. Change network to Ethereum Mainnet -- see how `test.eth`is resolved to the address from the other network
6. Proceed to the confirmation screen
7. See how old address is displayed, and since this domain is registered with another name, we are now seeing a different ENS domain as recipient `nauseam.eth`

_No response_

### Version

10.25.0

### Build type

None

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

NiranjanaBinoy:
## Explanation
Fixes #18045 
The initial issue was that if the user changes the network while performing a send transaction using ens, the ens gets resolved to the new network's wrong address.

The solution is to disable the ability to change the network and account once the user start a sends transactions. 
https://github.com/MetaMask/metamask-extension/issues/18045#issuecomment-1460467835

## Manual Testing Steps
1. Start a send transaction.
2. Click on the network display bar to change the network in the `add recipient page`; this should be disabled.
3. Click on the network display bar to change the network in the `draft send page`, this should also be disabled.

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/coverage-targets.js
+++ b/coverage-targets.js
@@ -9,7 +9,7 @@ module.exports = {
     lines: 66.58,
     branches: 54.79,
     statements: 65.74,
-    functions: 58.54,
+    functions: 58.4,
   },
   transforms: {
     branches: 100,
--- a/coverage-targets.js
+++ b/coverage-targets.js
@@ -6,10 +6,10 @@
 // subset of files to check against these targets.
 module.exports = {
   global: {
-    lines: 66.58,
-    branches: 54.79,
-    statements: 65.74,
-    functions: 58.4,
+    lines: 67.8,
+    branches: 55.84,
+    statements: 67.13,
+    functions: 59.66,
   },
   transforms: {
     branches: 100,
--- a/coverage-targets.js
+++ b/coverage-targets.js
@@ -6,10 +6,10 @@
 // subset of files to check against these targets.
 module.exports = {
   global: {
-    lines: 66,
-    branches: 54.4,
-    statements: 65,
-    functions: 58.5,
+    lines: 66.58,
+    branches: 54.79,
+    statements: 65.74,
+    functions: 58.54,
   },
   transforms: {
     branches: 100,
--- a/test/data/mock-send-state.json
+++ b/test/data/mock-send-state.json
@@ -2,6 +2,13 @@
   "DNS": {
     "resolution": ""
   },
+  "activeTab": {
+    "id": 113,
+    "title": "E2E Test Dapp",
+    "origin": "https://metamask.github.io",
+    "protocol": "https:",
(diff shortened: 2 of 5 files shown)


jpuri (dismissed):

jpuri:
Plz check if it possible to add test coverage, it will be good and codecov will also be green 😄

jpuri (dismissed):
Just a small feedbacks, looks great otherwise 👍

seaona:
Great job @NiranjanaBinoy I can see how network is disabled while performing a transaction.

I've noticed that the Account is also disabled. Is this the intended behaviour? cc @bschorchit 

https://user-images.githubusercontent.com/54408225/226632126-444d80d3-9ab4-497c-b828-29a2e82ab7c9.mp4

NiranjanaBinoy:
@seaona The account was disabled for the other scenarios where the network drop-down was disabled, so I replicated the same on this. 
@bschorchit Should we stick to this behavior or keep the account active while the user is in the `Add Reciepent` and `draft send` pages?

bschorchit:
I wonder how common is for users to change accounts during this flow. I'm inclined to say we keep the account changing **blocked** in those screens as well mainly because there's no info on the `from` account in these screens besides the icon on the top right for the user to review and confirm. And because changing account also creates some bugs (e.g. if you're sending an asset you own on one account but not in another and you change it to that second account). What do you think? Any strong reason to keep it enabled?

Minor detail: Is the difference on how the cursor looks like for hovering over the network dropdown and the account dropdown intentional?

seaona:
Make sense to me! Thank you @NiranjanaBinoy @bschorchit !! Maybe we could update the PR description/title to reference to the Account Disabling too :+1:

jpuri (dismissed):
🚀

jpuri (dismissed):

segun (dismissed):

jpuri (approved):
LGTM

segun (approved):

PR #18115, opened 2023-03-13
Unit tests for various modals

tmashuang:
## Explanation

Unit tests for:

```
ui/components/app/modals/hide-token-confirmation-modal/
ui/components/app/modals/export-private-key-modal/
ui/components/app/modals/customize-nonce/
```

## Pre-merge author checklist

- [x] I've clearly explained:


coverage-targets.js
```diff
@@ -8,8 +8,8 @@ module.exports = {
   global: {
     lines: 65,
     branches: 53.5,
-    statements: 64,
-    functions: 57.4,
+    statements: 64.2,
```
DDDDDanica, line 11: 💯

ui/components/app/modals/customize-nonce/customize-nonce.test.js
```diff
@@ -0,0 +1,132 @@
...
+    const { queryByText } = renderWithProvider(
+      <CustomizeNonce {...props} />,
+      mockStore,
+    );
+
+    const modalClose = queryByText('Cancel');
+
+    fireEvent.click(modalClose);
+
+    expect(mockHideModal).toHaveBeenCalled();
```
DDDDDanica, line 106: Question: this function should be tested in general <Modal /> component ideally ? we dont need to test again i think :) 
Check this out: https://github.com/MetaMask/metamask-extension/blob/develop/ui/components/app/modal/modal.component.test.js#L28

ui/components/app/modals/export-private-key-modal/export-private-key-modal.test.js
```diff
@@ -0,0 +1,81 @@
...
+
+  it('should match snapshot', () => {
+    const { container } = renderWithProvider(
+      <ExportPrivateKeyModal />,
+      mockStore,
+    );
+    expect(container).toMatchSnapshot();
+  });
+
+  it('should disable confirm button by default', () => {
```
DDDDDanica, line 45: We can skip this test as well as it is covered in 
https://github.com/MetaMask/metamask-extension/blob/develop/ui/components/app/modal/modal.component.test.js#L85
wdyt?

ui/components/app/modals/hide-token-confirmation-modal/hide-token-confirmation-modal.test.js
```diff
@@ -0,0 +1,84 @@
...
+    const { queryByTestId } = renderWithProvider(
+      <HideTokenConfirmationModal />,
+      mockStore,
+    );
+
+    const cancelButton = queryByTestId('hide-token-confirmation__cancel');
+
+    fireEvent.click(cancelButton);
+
+    expect(mockHideModal).toHaveBeenCalled();
```
DDDDDanica, line 66: same as https://github.com/MetaMask/metamask-extension/pull/18115/files#r1151071107 ;)

ui/components/app/modals/customize-nonce/customize-nonce.test.js
tmashuang: There is a case to be made for reducing redundancy here. At an elementary point, we would miss 100% coverage on the files. On another, what [modal.component.test.js#L28](https://github.com/MetaMask/metamask-extension/blob/develop/ui/components/app/modal/modal.component.test.js#L28) tests that whatever function we pass to `onCancel` it will be called, not what actual func is being called. In this case, the `hideModal` from our actions is what is being called which is what we are testing for in this instance. 

Let's say in the future the func passed was changed to `exportAllPrivateKeys` when `onCancel` is called, `modal.component.test.js` would still pass, but that is not what is supposed to happen and the test should fail in that instance.

DDDDDanica (approved):

brad-decker (approved):

PR #18125, opened 2023-03-13
feat: add yaml feature management

ritave:
This PR builds upon and supersedes #17984. It adds build configuration yaml file (`builds.yml`), and removes hardcoded logic concerning builds - making them all declarative.

Additionally, it adds a feature which I call "Build Features" - it replaces current code fencing based around build types to code fencing around specific features. Each build can contain multiple features. This is necessary because with the growth of MetaMask engineering team there's a lot new pre-release features being build, some are being audited, some aren't, and build types themselves are not enough to maintain that.

Fixes https://app.zenhub.com/workspaces/snaps-615b3a7c08d2b20015eb6c4e/issues/gh/metamask/metamask-planning/502

## Big pictures changes

- [x] Adds `builds.yml` file
- [x] Standarize environmental variable loading order
- [x] Requires all environmental variables to be declared in `builds.yml` before use.
- [x] Allows loading only the environmental variables that are expected to be used during that specific build
- [x] Partial extraction of most existing env variables, some are still modified in-code.
- [x] Build types are now composed of features
- [x] Code fencing is now based on features rather than build types
- [x] Splits flask fencing into desktop and snaps fences
- [x] Removes quite a few build type bugs, especially around `mmi`. Types were hardcoded in multiple places, their behaviour have been varied and was not exactly deterministic. By centralizing all definitions, each build type is now treated the same way.
- [x] Partial support for assets declaration in `builds.yml` rather than hardcoded

## Changes

- `app/*` changes are only code fencing names
- `shared/*` has minor changes to BUILD_TYPES enum and code fencing
- `development/*` has major changes to build system

## Compatibility

The intention of this PR is to maintain backwards compatibility, but due to huge complexity of the change, regressions might have been introduced. Thorough automatic and manual testing is strongly suggested, especially around:

1. **Production builds**
2. Main/Beta/Flask/MMI differences
3. Snaps feature set


brad-decker (commented):
I have been going through file by file and i'm over half way through and haven't found anything worth commenting on. Overall i'm super happy with this change, it makes a lot of sense and makes moving features between build types fairly trivial. 👍🏻 Great work @ritave

app/scripts/lib/setupSentry.js
```diff
@@ -90,7 +89,7 @@ export default function setupSentry({ release, getState }) {
   }
 
   const environment =
-    METAMASK_BUILD_TYPE === BuildType.main
+    METAMASK_BUILD_TYPE === 'main'
```
brad-decker, line 92: should we continue to use constants for these strings?

app/scripts/metamask-controller.js
```diff
@@ -840,6 +840,11 @@ export default class MetamaskController extends EventEmitter {
       ],
     });
 
+    const allowLocalSnaps = process.env.ALLOW_LOCAL_SNAPS;
+    const requireAllowlist = process.env.REQUIRE_SNAPS_ALLOWLIST;
+    assert(typeof allowLocalSnaps === 'boolean');
+    assert(typeof requireAllowlist === 'boolean');
```
brad-decker: Could we possibly move these assertions to higher up in the execution layer? It feels weird to be doing assertions on build variables in the application logic.

brad-decker (changes requested):

app/scripts/metamask-controller.js
FrederikBolding: Yeah, I removed these assertions for now. They also broke a bunch of tests

app/scripts/lib/setupSentry.js
FrederikBolding: We could keep the enum, but there is no guarantee it matches what is defined in `builds.yml`. Would you prefer that?

FrederikBolding:
@brad-decker This now passes CI and is ready for more review 👍

app/scripts/lib/setupSentry.js
FrederikBolding: I'll leave that up to you, but in general we don't seem to check it that much after this refactor. Which seems like a more sensible approach anyway

development/build/transforms/README.md
```diff
@@ -49,7 +49,7 @@ Note that multiple build types can be specified by separating them with
 commands inside the parameter parentheses:
 
 ```javascript
-///: BEGIN:ONLY_INCLUDE_IN(beta,flask)
+///: BEGIN:ONLY_INCLUDE_IN(build-beta,flask)
```
cryptotavares, line 52: Should be `build-flask` as well

app/scripts/lib/setupSentry.js
ritave:
```suggestion
    METAMASK_BUILD_TYPE === process.env.METAMASK_BUILD_TYPE_DEFAULT
```

development/build/transforms/README.md
FrederikBolding: Fixed!

test/jest/setup.js
```diff
@@ -3,6 +3,8 @@
 import nock from 'nock';
 import '@testing-library/jest-dom';
 
+process.env.SUPPORT_LINK = 'https://support.metamask.io';
```
cryptotavares, line 6: Can we move this to `test/jest/env.js`... I belive that is where we want to setup all jest tests specific env vars.

cryptotavares:
Looks great! I really love how easy it will be to move features across build types 😎 

The only reservation that I have is naming it features [here](https://github.com/MetaMask/metamask-extension/pull/18125/files#diff-d6a9a173a5b80ebccbd44d172db70d11715731b4ebb89d44465d3289da1bd642R90).
Code fencing, though useful, introduces a lot of complexity. Developers are required to take into account code blocks that may or may not exist for a specific build type.
By:
1. removing a lot of friction to add new code fences
2. detach code fences from build types
3. couple it to features (and name it features)

I can see those proliferating and making the code base becoming exponentially more complex and hard to change.

I believe that this could be partially mitigated if we re-named `features` from the build.yml into `fenced-features`? (I always prefer being explicit). Finally we should also add a markdown to the docs about adding `fenced-features` and when that should be considered (to help out on the decision process and avoid code fencing proliferation).
IMO those small changes could make huge difference.

What do you think?

brad-decker:
@FrederikBolding @cryptotavares i'm not in favor of renaming the key at this point, though I acknowledge that we don't want to increase the amount of code fencing happening. Perhaps this issue can be handled through education and documentation rather than making the keys slightly more obscurely named?

FrederikBolding:
@brad-decker Rebased!

FrederikBolding:
Sorry - had a few issues after rebasing.

pedronfigueiredo (dismissed):

pedronfigueiredo:
LGTM

brad-decker (dismissed):
LGTM

brad-decker (approved):

pedronfigueiredo (approved):

PR #18127, opened 2023-03-13
make network-controller methods overwrite rather than merge provider state

adonesky1:
This changes several places in the `NetworkController` where we  call  `updateState` - which merges partial state - on an `ObservableStore` instance, to `putState` -  which wholly overwrites state in that store. This pattern of immutable state mutation is safer and avoids confusing state scenarios where, for instance, a `rollbackToPreviousProvider` call results in a provider state with some of the state of the provider we failed to connect alongside the the details of the provider we rolled back to.


Gudahtt (approved):
LGTM!

mcmire (approved):
LGTM!

PR #18129, opened 2023-03-13
Revert "Moved subscribe and filter into network controller (#16693)"

### LINKED ISSUE #17467: [Bug]: MetaMask stays loading indefinetly and get error `StreamProvider - Unknown response id for response: [object Object]` when doing *some* contract interactions
### Describe the bug

**Problem**: when I call the function for mint an ERC1155 token, MetaMask stays loading indefinetly and I get the console error `Error in event handler: Error: StreamProvider - Unknown response id for response: [object Object]` .

https://user-images.githubusercontent.com/54408225/215133926-378001fe-6c51-45d9-8e64-930da22e90a6.mp4

This also happens when I call another contract function.

https://user-images.githubusercontent.com/54408225/215423675-bcd1b1e4-cc28-4e0a-9e29-911371eb2eab.mp4

[EDITED] It seems the error could start whenever we deploy a contract. It's not 100% deterministic, so further investigation needs to be done.

### Steps to reproduce [EDITED]
1. Go to Remix https://remix.ethereum.org/
2. Create a contract
```
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.17;

contract Fallback {
    event Log(string func, uint gas);
    fallback() external payable {
        emit Log("fallback", gasleft());
    }
    receive() external payable {
        emit Log("receive", gasleft());
    }
    function getBalance() public view returns (uint) {
        return address(this).balance;
    }
}
```
3. Deploy this contract on Goerli - see error on console // if not, deploy another contract
4. Create another contract 
```
contract SendToFallback {
    uint val = 0;
    function transferToFallback(address payable _to) public payable {
        _to.transfer(msg.value);
    }

    function callFallback(address payable _to) public payable {
        (bool sent, ) = _to.call{value: msg.value}("");
        require(sent, "Failed to send Ether");
    }
    function count(uint _val) public {
        val = _val;
    }
}
```
6. Deploy this contract on Goerli - see error on console

After this, we are into a broken state, shown in the videos above.

_No response_

### Version

develop [EDITED]

### Build type

None

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

Gudahtt:
This reverts commit 6f6984fa58c4797e809387b3b3d0b28e72b252bf. That commit was an RPC middleware refactor intended to move the subscribe and filter middleware into the network controller, to simplify the process of sharing this middleware between clients.

This refactor resulted in `eth_subscribe` notifications being sent on the wrong connections, causing the UI to break in some cases (the UI `provider` connection does not support notifications). This happened because the `setupProviderEngine` function runs per-connection, whereas the engine setup inside the network controller is global. The global network client cannot support notifications because it has no way to route them; they'll need to stay in the per-connection provider engine.

Closes #17467

## Manual Testing Steps

* Attempt to reproduce #17467
* Test that the filter and subscribe methods work correctly
  * I'd recommend running this branch of the `test-dapp` locally to help with testing: https://github.com/MetaMask/test-dapp/pull/202

## Pre-merge author checklist

- [x] I've clearly explained:


Gudahtt:
>Sufficient automated test coverage has been added

We should be able to cover this with an e2e test. I'll have to give that some thought. I don't think it'd be feasible to capture this error in a unit test, since it's more of an integration problem.

seaona:
Looks like this fixed it. I cannot reproduce the error anymore @Gudahtt

test/e2e/tests/eth-subscribe.spec.js
```diff
@@ -0,0 +1,76 @@
...
+
+        await driver.executeScript(setupSubscriptionListener);
+        await driver.executeScript(`
+          window.ethereum.request({
+            method: 'eth_subscribe',
+            params: ['newHeads']
+          });
+        `);
+
+        // Verify that new block is seen on first dapp
```
Gudahtt, line 60: We use a default block time of 2 in our Ganache setup. Our `findElement` timeout is more than long enough to wait for this, but in the future we can improve this test by explicitly mining a block here using the `evm_mine` method enabled by this PR: https://github.com/MetaMask/metamask-extension/pull/18149

test/e2e/tests/eth-subscribe.spec.js
```diff
@@ -0,0 +1,76 @@
...
+      },
+      async ({ driver }) => {
+        await driver.navigate();
+        await driver.fill('#password', 'correct horse battery staple');
+        await driver.press('#password', driver.Key.ENTER);
+
+        await driver.openNewPage('http://127.0.0.1:8080/');
+
+        const setupSubscriptionListener = `
+          const responseContainer = document.createElement('div');
```
Gudahtt, line 34: We can replace this with the built-in test dapp subscribe functionality after this PR is merged: https://github.com/MetaMask/test-dapp/pull/202

test/e2e/tests/eth-subscribe.spec.js
adonesky1, line 60: by new block do you mean the div that is appended when the subscription response comes back? Or do you mean a block like an ethereum block?  If its the former, which it seems like it is maybe we should use a different word than "block"?

Gudahtt: I meant the latter, an Ethereum block. `newHeads` will emit a notification for each new block

adonesky1: Ok gotcha.

adonesky1:
> Test that the filter and subscribe methods work correctly
> I'd recommend running this branch of the test-dapp locally to help with testing: https://github.com/MetaMask/test- 
> dapp/pull/202

I tried this both on this branch and on `develop` and it worked on both...

adonesky1 (dismissed):
LGTM! Tested and verified that `eth_subscribe` subscription messages are no longer broadcast to all pages where we inject our provider

test/e2e/tests/eth-subscribe.spec.js
```diff
@@ -0,0 +1,78 @@
...
+          });
+        `);
+
+        // Verify that the new block is seen on the first dapp
+        await driver.findElement('[data-testid="eth-subscribe-response"]');
+
+        // Switch to the second dapp
+        await driver.openNewPage('http://127.0.0.1:8081/');
+
+        // Setup the same subscrption listener as on the first dapp, but without registering a new subscription
```
adonesky1, line 68: nit
```suggestion
        // Setup the same subscription listener as on the first dapp, but without registering a new subscription
```

test/e2e/tests/eth-subscribe.spec.js
Gudahtt: I've just updated these comments to be a bit clearer: https://github.com/MetaMask/metamask-extension/pull/18129/commits/2aa96dac4a52f2d94a0bdf7a74b10c728bb8bea9

test/e2e/tests/eth-subscribe.spec.js
Gudahtt: Thanks! Fixed here: https://github.com/MetaMask/metamask-extension/pull/18129/commits/fde1a64b7b93497648c1ecdc28d01f7ddb63fe0b

adonesky1 (approved):

chloeYue (approved):
LGTM, thanks!

PR #18146, opened 2023-03-14
Updated Tooltip for light and dark theme

NidhiKJha:
Currently, we are updating the tooltip css with the custom css for dark and light themes. This PR is to update the Tooltip to have the light background for the dark theme and vice versa.

### Before

https://user-images.githubusercontent.com/39872794/225064866-a23d0ae2-362f-4203-a7c2-630da25fe663.mov

### After

https://user-images.githubusercontent.com/39872794/225061512-374d492d-f667-4dfd-b546-7902b148fa57.mov


georgewrmarshall (commented):
Can you check what hex values are being used? I don't think they will align with our design token colors?

NidhiKJha:
> Can you check what hex values are being used? I don't think they will align with our design token colors?

I tried to check it in the HTML, but couldn't find it, checked in the [official docs of tippyjs ](https://github.com/atomiks/tippyjs/tree/master/src/scss/themes) to confirm the color values are:

### light theme: 
  color: #26323d;
  background-color: white;

darkwing:
@NidhiKJha Is this still relevant?

NidhiKJha:
Hmm... we need to update the colors in the tooltip but the best approach for this would be to update the current packages for Tippy and Popover. But I don't think we have the bandwidth to work on it in this sprint. So, I think we can close it and can resume work on this in a separate PR later.

PR #18163, opened 2023-03-15
Use core message managers and create sign controller

### LINKED ISSUE #17417: Refactor Signature Request Message managers
Currently we have 3 different message managers:

1. [MessageManager](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/lib/message-manager.js)
2. [PersonalMessageManager](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/lib/personal-message-manager.js)
3. [TypedMessageManager](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/lib/typed-message-manager.js)

Above message managers handle different types of message requests sent by DAPP. There is a lot of code duplication between the 3 classes. We can extract out a parent class and move duplicated code to it.

Refer for details:
https://github.com/MetaMask/metamask-extension/tree/develop/docs/refactoring/signature-request

matthewwalsh0:
## Explanation

Currently the signing requests (standard, personal, and typed) are all handled directly in `metamask-controller` across 9 functions, each using a different message manager implementation from `app/scripts/lib`.

These changes encompass the following high level goals:

- Remove all the message manger implementations from the extension and instead use the corresponding implementations from the core repositoy, specifically the `@metamask/message-manager` package.
- Encapsulate all the signing functions into a separate new controller to:
  - Minimise the monolithic nature of the `metamask-controller` and group related code.
  - Remove duplication in the signing logic.
  - Act as a compatability layer to avoid duplicate state and to appease the existing state used by the frontend.
- Use the `ApprovalController` to display the popup for signing requests to centralise the responsibility and streamline the frontend state in future.

I'm unsure if it is strict policy that all new controllers go into the core repository, but this feels very specific to the extension currently as it's solely encapsulating existing functionality. Plus I believe it would be easier to build it within the extension first to verify it is stable, then I would be happy to migrate it to core if required.

## Manual Testing Steps

Use any build and verify the signing operations in the test dApp.

## Pre-merge author checklist

- [x] I've clearly explained:


app/scripts/metamask-controller.js
```diff
@@ -1074,12 +1056,19 @@ export default class MetamaskController extends EventEmitter {
         this.metaMetricsController,
       ),
     });
-    this.typedMessageManager = new TypedMessageManager({
-      getCurrentChainId: () =>
-        this.networkController.store.getState().provider.chainId,
-      metricsEvent: this.metaMetricsController.trackEvent.bind(
```
matthewwalsh0, line 1080: Pending migration to the sign controller, can only find one usage on rejection.

app/scripts/metamask-controller.js
```diff
@@ -1135,10 +1124,7 @@ export default class MetamaskController extends EventEmitter {
     this.networkController.on(NETWORK_EVENTS.NETWORK_WILL_CHANGE, () => {
       this.txController.txStateManager.clearUnapprovedTxs();
       this.encryptionPublicKeyManager.clearUnapproved();
-      this.personalMessageManager.clearUnapproved();
-      this.typedMessageManager.clearUnapproved();
       this.decryptMessageManager.clearUnapproved();
-      this.messageManager.clearUnapproved();
```
matthewwalsh0, line 1141: Pending an implementation in the sign controller, should be straight forward.

jpuri (approved):
Great work 👍

bergarces (approved):

cryptotavares (approved):

PR #18263, opened 2023-03-21
[FLASK] Redesign key management modal

### LINKED ISSUE #18166: Implement redesigned key management confirmation modal
The redesigned modal is using the new DS, is less wordy, and has better support for cases where the number of key management permissions > 1.

An intermediate redesign could be to use the existing (legacy) modal component (instead of the new DS) and start with just the updated copy. Let me know.

Figma: https://www.figma.com/file/diidnCHx9FXj5LDiA5DZZr/Snaps-features?node-id=56%3A14905&t=mhmNg4AXaIQjuoc1-1

hmalik88:
Closes #18166 

Updated the snap install warning modal to now house warnings for all weight-1 permissions, the design has also been updated per: https://www.figma.com/file/diidnCHx9FXj5LDiA5DZZr/Snaps-features?node-id=24-11653&t=IZlEBMh8N7M9QLjb-0

See below for new warning:

https://user-images.githubusercontent.com/41640681/226725873-63cc4cae-f9d3-43dc-8d9e-3c5ba3467cf8.mov


ui/pages/permissions-connect/flask/util.js
```diff
@@ -13,60 +14,91 @@ import {
 } from '../../../helpers/utils/util';
 
 export function getSnapInstallWarnings(permissions, targetSubjectMetadata, t) {
-  const bip32EntropyPermissions =
-    permissions &&
-    Object.entries(permissions)
-      .filter(([key]) => key === 'snap_getBip32Entropy')
-      .map(([, value]) => value);
+  const bip32EntropyPermission = permissions.snap_getBip32Entropy;
```
Mrtenz, line 17: Could you do this in a loop instead, by looking at all permissions and getting their weight from `PERMISSION_DESCRIPTIONS`?

hmalik88: I thought about that, but I'm not sure it makes sense, because we aren't rendering the descriptions in an abstract way. Grabbing the permissions from their weights doesn't mean much when we're writing custom logic for each weight-1 permission (which doesn't seem like it will change either because we don't know what future weight-1 permissions will be).

GuillaumeRx (dismissed):
E2E tests seems to be broken due to a classname change

ui/pages/permissions-connect/flask/util.js
Mrtenz: How about using the actual return value from `PERMISSION_DESCRIPTIONS[permission](...)`? You could add a new field there.

GuillaumeRx (dismissed):

Mrtenz (approved):

GuillaumeRx (approved):

PR #18294, opened 2023-03-22
feat(custom-spending-cap): sets custom spending cap into dappProposed…

FiboApe:
…Value when loaded

## Explanation

What is the current state of things and why does it need to change? Since the last update to the Custom Spending (Approval) screen, users actually need to input the amount they want to allow a contract to spend for them, this seems like an unnecesary step when to submit an approval a dapp needs to submit a value for allowance.

It is understable that most dapps ask for infinite approval out of the gate, which is super detrimental for User Security as any hack could happen and drain most of the funds from the users wallet. A change that we could implement here if that the approval amount that is being required is infinite, show a warning and don't set the field value to infinite. But for specific amount of approvals it just seems like adding an extra step to a flow that might not make a lot of sense.

The solution proposed here just adds the amount that the dapp proposed to the input, this is up for discussion into how we want to handle that, but this should be a first approach at it. There's a case to be made also that if a dapp proposed value is not infinte to automatically move the user to the second step of the approval process to streamline submitting the transaction even smoother.

Another thing is that currently with the custom spending cap screen, the buttons need to be scrolled to be found, but this could be a separate issue of just ui styling

## Screenshots/Screencaps
Screenshot 2023-03-22 at 18 29 23

### Before
Screenshot 2023-03-22 at 18 31 01

### After
Screenshot 2023-03-22 at 18 29 23

## Manual Testing Steps
Go to any approval screen and start the token approval process

## Pre-merge author checklist

- [X] I've clearly explained:

- [ ] PR is linked to the appropriate GitHub issue -> There is currently no issue filled about this, could create one if its necessary

### DIFF
--- a/ui/components/app/custom-spending-cap/custom-spending-cap.js
+++ b/ui/components/app/custom-spending-cap/custom-spending-cap.js
@@ -1,4 +1,4 @@
-import React, { useState, useContext, useEffect } from 'react';
+import React, { useState, useContext, useEffect, useCallback } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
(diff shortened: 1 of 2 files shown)


FiboApe:

Issue #18191, opened 2023-03-16
Support domain mismatch on SIWE on development mode


bschorchit:
Originally reported here: https://github.com/MetaMask/metamask-extension/issues/17707#issuecomment-1470877080

As we've added support for SIWE, we have blocked users from signing messages with domain mismatch for security reasons - preventing phishing attacks as proposed in the SIWE spec. 
Although this was the correct decision on the user facing side, we might have made developers work a lot more complicated with this change (see link above).

Therefore, while we should not make changes on the end user side (as allowing for domain mismatch would reintroduce the phishing attack vector), we could enable developers to workaround this block in order to facilitate their work. 

Adding a toggle in advanced settings and communicate the existence of it through the error thrown in console (similar to what we currently do for eth_sign https://github.com/MetaMask/metamask-extension/pull/17308) could be a potential a solution for this. But the DevEx team is best suited to propose the ideal solution here.

digiwand:
In case it helps to support the disabling the domain-binding check for developers, @legobeat has proposed another consideration or plan B:
> I think there's https://github.com/MetaMask/core/pull/1163 which maintains safe and compliant defaults without causing hindrance for local dev or requiring an "unsafe dev mode" (completely disable domain-binding validation? maybe a feature that could be enabled only in flask and beta), it can still be considered as a plan b.

More details and discussion for domain binding logic updates can be found here: https://github.com/MetaMask/metamask-extension/issues/18188

PR #18296, opened 2023-03-23
Sign in with Ethereum: domain and origin should compare without port

### LINKED ISSUE #18188: [Bug]: Passing http://localhost:4201 as the URI in SIWE message on a localhost:4201 site seems to close MM popup without any error thrown
### Describe the bug

This was originally reported to us [here on Slack](https://consensys.slack.com/archives/G8RSKCNCD/p1678906632979459) and here https://github.com/MetaMask/metamask-extension/issues/17707#issuecomment-1470728017

### Steps to reproduce

Here is a example of message to sign that was reported to cause this while on localhost : https://github.com/MetaMask/metamask-extension/issues/17707#issuecomment-1470786260

_No response_

### Version

10.26.1

### Build type

None

### Browser

Other (please elaborate in the "Additional Context" section)

### Operating system

Other (please elaborate in the "Additional Context" section)

### Hardware wallet

_No response_

### Additional context

Not sure on browser and OS info.

yonigoldberg:
## Explanation
- Fixes: https://github.com/MetaMask/metamask-extension/issues/18188

The parsed SIWE domain doesn't have a port (e.g loclhost) therefore when it is compared with the origin, the host from the origin should be without a port as well. That is achieved by using `hostname` instead of `host`.

### Before

image
### After

image
## Manual Testing Steps
- Run any local server and try to login with SIWE

## Pre-merge author checklist

- [x] I've clearly explained:


yonigoldberg:

skgbafa:
Amazing, came to the repo to make this exact change. The hostname will always map to the domain and drop the port, preventing the domain binding bug from happening on localhost. This change looks great to me.

digiwand (dismissed):
Lgtm! Thanks for adding the fix and test for this @yonigoldberg!

I think it'd be a good idea to add another test to check the case where [domain contains the port number](https://eips.ethereum.org/EIPS/eip-4361) as well. I don't think this is a blocker for this PR, though I think we should add it here or in a follow-up PR

digiwand (approved):

ui/components/app/signature-request-siwe/signature-request-siwe.js
```diff
@@ -51,8 +51,9 @@ export default function SignatureRequestSIWE({
     let isSIWEDomainValid = false;
 
     if (origin) {
-      const { host } = new URL(origin);
-      isSIWEDomainValid = parsedMessage.domain === host;
+      const { hostname, host } = new URL(origin);
+      isSIWEDomainValid =
+        parsedMessage.domain === hostname || parsedMessage.domain === host;
```
naugtur, line 56: This loosening of origin matching doesn't seem to be limited to localhost in any way. As a result, any two origins (in the browser's sameorigin policy sense) that share the domain but not port&protocol will be considered a match. 

Are we not worried about changing that? 
Shared hosting services exist, where two different apps/customers get the same IP or domain but different port. I know it's a weird example, but it's a practical example of origin including a port.

If we consider it an extension of sameorigin policy, we should not fix localhost issues this way. 

(also, sorry this is coming in so late, but I was only made aware of this PR a few minutes ago and I don't even have all the context - I might be 100% wrong)

digiwand (changes requested):
re: @naugtur [comment above](https://github.com/MetaMask/metamask-extension/pull/18296/files#r1150684088)

I think that's a good idea to limit the scope to localhost. Blocking this PR now to apply the change and continue the discussion above.

ui/components/app/signature-request-siwe/signature-request-siwe.js
naugtur: Would this work too? There's no security concerns I see if done that way:  
```suggestion
        (hostname === 'localhost' && parsedMessage.domain === hostname) || parsedMessage.domain === host;
```

digiwand (dismissed):

jpuri (dismissed):

jpuri:
Hey @yonigoldberg : can you plz fix lint issue.

ui/components/app/signature-request-siwe/signature-request-siwe.js
```diff
@@ -51,8 +51,9 @@ export default function SignatureRequestSIWE({
     let isSIWEDomainValid = false;
 
     if (origin) {
-      const { host } = new URL(origin);
-      isSIWEDomainValid = parsedMessage.domain === host;
+      const { hostname, host } = new URL(origin);
+      isSIWEDomainValid =
+        (hostname === 'localhost' && parsedMessage.domain === hostname) || parsedMessage.domain === host;
```
digiwand, line 56:
```suggestion
        (hostname === 'localhost' && parsedMessage.domain === hostname) || 
        parsedMessage.domain === host;
```
fixes lint

digiwand (commented):
having an interesting conversation with @legobeat. 
 
@yonigoldberg, we will need to hold off on merging this right now while we figure this out. Thank you for your patience with this. I am curious to know if you have found a workaround the match the port numbers in the meantime. Hoping you have 

---

For clarification, I'm realizing the issue ticket that is currently related to this PR is only concerned with the http/https protocol rather than the port. maybe we can unrelate the tickets or combine the issues

**Other related tickets:**
- Support domain mismatch on SIWE on development mode #18191
- [Bug]: Weird behavior for SIWE requests on localhost #18188
- addressing subdomains (e.g. `www`) - (#18332)

yonigoldberg:
Per the conversation in  https://github.com/MetaMask/metamask-extension/issues/18188 I think we can close this PR.

digiwand:
> Thank you for opening this PR, your input, and keeping the conversation moving @yonigoldberg

+1

digiwand:
+1

digiwand:
^ +1

PR #18218, opened 2023-03-17
fix(18190): add tabs to permission when initializing app

DDDDDanica:
## Explanation
Fixes https://github.com/MetaMask/metamask-extension/issues/18190

Inside [manifest configuration](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/permissions), we are missing `tabs` for firefox <85 version (v78 is our min support version at the moment), hence the fix would be adding `tabs` to `permissions`

> tabs gives you access to [privileged parts of the tabs API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs) without the need for [host permissions](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/permissions#host_permissions): Tab.url, Tab.title, and Tab.faviconUrl.
In Firefox 85 and earlier, you also need tabs if you want to include url in the queryInfo parameter to [tabs.query()](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/query). The rest of the tabs API can be used without requesting any permission.

### After

https://user-images.githubusercontent.com/12678455/225911766-2af6a2e5-2d96-400d-a859-c4dc94950cbf.mov

## Manual Testing Steps
- Install firefox v78
- Try to open extension 
- Extension should work on firefox v78

## Pre-merge author checklist

- [x] I've clearly explained:
  - [x] What problem this PR is solving
  - [x] How this problem was solved
  - [x] How reviewers can test my changes
- [ ] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [ ] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
- [ ] PR is linked to the appropriate GitHub issue
- [ ] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.

In this case, a QA Engineer approval will be be required.

PR #18306, opened 2023-03-23
fix alarms.getAll() console errors

pedronfigueiredo:
## Explanation

Fixes this error in the console for mv3 builds

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'getAll')
  at new MetaMetricsController (metametrics.js:179:21)
  at new MetamaskController (metamask-controller.js:396:34)
  at setupController (background.js:399:16)
  at initialize (background.js:251:5)
```

### DIFF
--- a/app/manifest/v2/_base.json
+++ b/app/manifest/v2/_base.json
@@ -60,5 +60,19 @@
   },
   "manifest_version": 2,
   "name": "__MSG_appName__",
+  "permissions": [
+    "storage",
+    "unlimitedStorage",
+    "clipboardWrite",
+    "http://localhost:8545/",
+    "https://*.infura.io/",
+    "https://*.codefi.network/",
+    "https://chainid.network/chains.json",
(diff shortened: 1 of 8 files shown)


danjm (approved):

adonesky1 (dismissed):

danjm (changes requested):
I need to investigate whether the root cause of the bug here is a lavamoat / snow issue

pedronfigueiredo:
I believe a regression might have been introduced here https://github.com/MetaMask/metamask-extension/pull/18218/files#diff-722208d23dde0a9af3f90b062acbcc12d9341482d73c8b55f52a57cf2a8e0e70, so I've selectively reverted those changes

pedronfigueiredo:
Closing since this has now been addressed here https://github.com/MetaMask/metamask-extension/pull/18336

PR #18309, opened 2023-03-23
UX: Multichain: Improve accessibility of accounts menu

darkwing:
## Explanation

The original implementation of the `AccountListItem` was kinda hacky; it was an `<a>` that contains a `[button as span]` and was kinda weird.  After speaking with @georgewrmarshall and looking at what GitHub does, I think this solution is much nicer.

## Screenshots/Screencaps

The account menu items look exactly the same; instead, when tabbing, you'll stop on just the account name instead of the whole block.

## Pre-merge author checklist

- [ ] I've clearly explained:

### DIFF
--- a/ui/components/multichain/account-list-item/account-list-item.js
+++ b/ui/components/multichain/account-list-item/account-list-item.js
@@ -87,9 +87,7 @@ export const AccountListItem = ({
       className={classnames('multichain-account-list-item', {
         'multichain-account-list-item--selected': selected,
       })}
-      as="button"
-      onClick={(e) => {
-        e.preventDefault();
+      onClick={() => {
         // Without this check, the account will be selected after
         // the account options menu closes
         if (!accountOptionsMenuOpen) {
@@ -127,10 +125,14 @@ export const AccountListItem = ({
                   position="bottom"
                   wrapperClassName="multichain-account-list-item__tooltip"
                 >
-                  {identity.name}
+                  <a href="#" onClick={onClick}>
+                    {identity.name}
+                  </a>
                 </Tooltip>
               ) : (
-                identity.name
+                <a href="#" onClick={onClick}>
+                  {identity.name}
+                </a>
               )}
             </Text>
             <Box
@@ -194,13 +196,6 @@ export const AccountListItem = ({
           onClick={(e) => {
             e.stopPropagation();
             setAccountOptionsMenuOpen(true);
-          }}
-          as="div"
-          tabIndex={0}
-          onKeyPress={(e) => {
-            if (e.key === 'Enter') {
-              setAccountOptionsMenuOpen(true);
-            }
           }}
           data-testid="account-list-item-menu-button"
         />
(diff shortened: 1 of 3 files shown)


ui/components/multichain/account-list-item/account-list-item.js
```diff
@@ -127,10 +125,14 @@ export const AccountListItem = ({
                   position="bottom"
                   wrapperClassName="multichain-account-list-item__tooltip"
                 >
-                  {identity.name}
+                  <a href="#" onClick={onClick}>
+                    {identity.name}
+                  </a>
```
NidhiKJha, line 130: Can we use ButtonLink here instead of `a`?

vthomas13: I'm curious about whether ButtonLink is good here too

georgewrmarshall: I think the intention here is to keep the UI looking like static text but allow for keyboard navigation and tabbing. I guess the difference to this and GitHub is that they're using it for navigation while here we are using an onClick to change the state of the app not necessarily the route

ui/components/multichain/account-list-item/account-list-item.js
```diff
@@ -127,10 +126,20 @@ export const AccountListItem = ({
                   position="bottom"
                   wrapperClassName="multichain-account-list-item__tooltip"
                 >
-                  {identity.name}
+                  <ButtonLink
+                    onClick={onClick}
+                    className="multichain-account-list-item__account-name"
+                  >
+                    {identity.name}
+                  </ButtonLink>
```
georgewrmarshall, line 134: I think `<Text as="button">` we want to keep the Text styling but change the underlying element. I think you will still need to add some custom styles here via Box props or a className for the hover states though

georgewrmarshall (commented):
Left one suggestion. Keyboard navigation works great though!! I think a `button` element is the right move semantically 🦾

ui/components/multichain/account-list-item/account-list-item.js
darkwing: That isn't working as it bricks the `ellipsis` functionality.  Also, using `ButtonLink` requires only one CSS change, while using `as="button"` requires some background and padding work.

brad-decker (dismissed):

ui/components/multichain/account-list-item/account-list-item.js
```diff
@@ -121,17 +120,24 @@ export const AccountListItem = ({
...
-              ) : (
-                identity.name
-              )}
+              <ButtonLink
+                onClick={onClick}
+                className="multichain-account-list-item__account-name"
+                color={Color.textDefault}
+                ellipsis
+              >
+                {identity.name.length > MAXIMUM_CHARACTERS_WITHOUT_TOOLTIP ? (
```
vthomas13, line 129: I feel like this logic can be reused in the future, really nice

vthomas13 (approved):

NidhiKJha (approved):
LGTM! Just a non blocking suggestion

PR #18328, opened 2023-03-24
Convert NetworkController deps to TS

### LINKED ISSUE #18285: Convert NetworkController dependencies to TypeScript
We want to convert NetworkController to TypeScript in order to be able to compare differences between the version of NetworkController in this repo and the version in the core repo. To do this, however, we need to convert the following files:

- `shared/modules/fetch-with-timeout.js`
- `shared/modules/network.utils.js`
- `shared/constants/metametrics.js`

We will also need to:

- upgrade to the latest version of `@metamask/swappable-obj-proxy` in order to be able to use the types there
- provide minimal types for `eth-query`

mcmire:
## Explanation

We want to convert NetworkController to TypeScript in order to be able to compare differences in the controller between in this repo and the core repo. To do this, however, we need to convert the dependencies of the controller to TypeScript.

To this end, this commit converts the following files:

- shared/modules/fetch-with-timeout.js
- shared/modules/network.utils.js
- shared/constants/metametrics.js

Note that NetworkController depends on some NPM modules that have no type definitions. These will be addressed in a future commit.

Fixes #18285.

## Manual Testing Steps

This commit should not introduce any functional changes. All tests should pass, and all functionality in the extension should continue to work.

## Pre-merge author checklist

- [x] I've clearly explained:


shared/constants/metametrics.ts
```diff
@@ -0,0 +1,637 @@
...
+  OnboardingWalletSecurityPhraseConfirmed = 'SRP Backup Confirmed',
+  OnboardingWalletCreationComplete = 'Wallet Created',
+  OnboardingWalletSetupComplete = 'Application Opened',
+  OnboardingWalletAdvancedSettings = 'Settings Updated',
+  OnboardingWalletImportAttempted = 'Wallet Import Attempted',
+  OnboardingWalletVideoPlay = 'SRP Intro Video Played',
+  OnboardingTwitterClick = 'External Link Clicked',
+}
+
+export enum MetaMetricsEventAccountType {
```
mcmire, line 539: Note that previously all of these event-related enums were packaged into one object. Enum values must be simple, so we can't nest these into a larger enum.

shared/constants/metametrics.ts
```diff
@@ -0,0 +1,637 @@
...
+  User = 'user',
+}
+
+export enum MetaMetricsEventLocation {
+  TokenDetails = 'token_details',
+  TokenDetection = 'token_detection',
+  TokenMenu = 'token_menu',
+}
+
+export enum MetaMetricsEventUiCustomizationPropertyValue {
```
mcmire, line 627: Previously this enum was nested in another object (`METAMETRIC_KEY_OPTIONS`). Considering that this object was being used for only one thing, I didn't think it was worth keeping that structure. Also, as above, enum values must be simple so this enum follows the same pattern as the enums above.

shared/modules/fetch-with-timeout.ts
```diff
@@ -17,12 +30,9 @@ const getFetchWithTimeout = memoize((timeout = SECOND * 30) => {
     const timer = setTimeout(() => abortController.abort(), timeout);
 
     try {
-      const res = await f;
+      return await f;
+    } finally {
       clearTimeout(timer);
-      return res;
-    } catch (e) {
-      clearTimeout(timer);
```
mcmire, line 24: I was getting a warning while running the Jest tests for this file that a worker process failed to finish. It seemed to somehow be related to the timer not being totally cleared in this failure case. Since we call `clearTimeout` in both the `try` and `catch` branches, it seemed like we could move it to a `finally`.

shared/modules/network.utils.ts
```diff
@@ -35,7 +33,7 @@ export function isPrefixedFormattedHexString(value) {
  * @param chainId - ChainID of network
  * @returns Whether the current network supports token detection
  */
-export function isTokenDetectionEnabledForNetwork(chainId) {
+export function isTokenDetectionEnabledForNetwork(chainId: string | undefined) {
```
mcmire, line 36: Initially this was `unknown`, but I decided to use `string | undefined` based on how this function is used in practice.

shared/constants/metametrics.ts
```diff
@@ -0,0 +1,637 @@
...
+/**
+ * Mixpanel converts the zero address value to a truly anonymous event, which
+ * speeds up reporting
+ */
+export const METAMETRICS_ANONYMOUS_ID = '0x0000000000000000';
+
+/**
+ * Used to identify events that are triggered by the background process.
+ */
+export const METAMETRICS_BACKGROUND_PAGE_OBJECT: MetaMetricsPageObject = {
```
mcmire, line 440: Once we upgrade TypeScript we should be able to use `satisfies MetaMetricsPageObject` instead of typing it like this, but this should work for the time being.

Gudahtt:
Could we split this into three parts? Or are these tangled up in each other in some way

mcmire:
@Gudahtt I think it would be possible to split up, let me try.

mcmire:
Following the above suggestion I've split this up into three PRs:

* https://github.com/MetaMask/metamask-extension/pull/18351
* https://github.com/MetaMask/metamask-extension/pull/18352
* https://github.com/MetaMask/metamask-extension/pull/18353

PR #18331, opened 2023-03-25
Add scripting permission

tmashuang:
Adds `scripting` permission for mv3 manifest.
Used in https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/app-init.js#L141
Ref: https://developer.chrome.com/docs/extensions/reference/scripting/

Partially requires the `alarms` permission to be added, https://github.com/MetaMask/metamask-extension/pull/18306.

### DIFF
--- a/app/manifest/v3/chrome.json
+++ b/app/manifest/v3/chrome.json
@@ -8,6 +8,7 @@
   },
   "minimum_chrome_version": "80",
   "permissions": [
+    "scripting",
     "storage",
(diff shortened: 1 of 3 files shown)


tmashuang:
Closing Ref: https://github.com/MetaMask/metamask-extension/pull/18306#issuecomment-1490015972
TLDR; Done in https://github.com/MetaMask/metamask-extension/pull/18336

PR #18336, opened 2023-03-27
Revert "fix(18190): add tabs to permission when initializing app"

pedronfigueiredo:
Reverts MetaMask/metamask-extension#18218

Reverting this PR due to some regressions, namely regarding "alarms" and "scripting" permissions.

An investigation on the need for "tabs" on firefox is ongoing


danjm (approved):

app/manifest/v2/firefox.json
```diff
@@ -4,20 +4,5 @@
...
-    "storage",
-    "unlimitedStorage",
-    "clipboardWrite",
-    "http://localhost:8545/",
-    "https://*.infura.io/",
-    "https://*.codefi.network/",
-    "https://chainid.network/chains.json",
-    "https://lattice.gridplus.io/*",
-    "activeTab",
-    "tabs",
```
DDDDDanica, line 18: I think we still need to keep this one for firefox old browser support bug ~~

danjm: I suggest we solve that seperately. If we can find a way to solve that bug without adding the permission, lets do that. We could wait until we have agreed on the final approach before merging this PR.

I can rebase my other PR on top of this PR's branch and we can discuss that alternative solution there

DDDDDanica: That's perfect 🙏

DDDDDanica (approved):

DDDDDanica (approved):

PR #18344, opened 2023-03-28
Fix firsttimeloaded logic

### LINKED ISSUE #18244: MV3: app seems to crash on service worker restart
Repro:

1. Build metamask in mv3
2. onboard / login
3. restart the service worker
4. restart the service worker again
5. see an error related to balance being undefined

danjm:
Fixes https://github.com/MetaMask/metamask-extension/issues/18244

This PR fixes two problems:

1. A race condition in the account tracker upon service worker restart, where the `start()` method could be called before the `_loginUser()` call in the metamask-controller resolves, which means the account tracker `_updateAccounts()` could be called before there are any accounts in state, resulting in the `accounts` object in state being set to empty. This was solved by ensuring that the account-tracker state is populated during construction _if_ we are in a case where the service worker is restarting
2. The code for deciding whether "we are in a case where the service worker is restarting" was not working. We were deciding this based on the `globalThis.isFirstTimeProfileLoaded` variable. Which would be undefined in all cases, except when the browser is first started, but not when the extension is reloaded, and not upon installing the extension during a browser session that has already started. This was solved by using session storage to set whether we are in the first instance of the metamask controller being started within the current session of the extension.

To test this manually:
1. Run an mv3 build
2. Terminate the service worker a bunch of times, via chrome://inspect
3. the service worker should continuously restart successfully


pedronfigueiredo (dismissed):

app/scripts/background.js
```diff
@@ -268,7 +268,24 @@ async function initialize() {
...
     ///: END:ONLY_INCLUDE_IN
 
-    setupController(initState, initLangCode);
+    let { isFirstMetaMaskControllerSetup } = await browser.storage.session.get([
+      'isFirstMetaMaskControllerSetup',
+    ]);
+
+    if (isFirstMetaMaskControllerSetup === true) {
+      isFirstMetaMaskControllerSetup = false;
+      await browser.storage.session.set({ isFirstMetaMaskControllerSetup });
```
DDDDDanica, line 277: Optional: we can isolate the setter out since it will be set for both conditions

danjm: @DDDDDanica do you mean something like this?

```
if (isFirstMetaMaskControllerSetup === true) {
  await browser.storage.session.set({ isFirstMetaMaskControllerSetup });
} else if (isFirstMetaMaskControllerSetup === undefined) {
  await browser.storage.session.set({ isFirstMetaMaskControllerSetup });
}
isFirstMetaMaskControllerSetup = isFirstMetaMaskControllerSetup === undefined
	? true
	: isFirstMetaMaskControllerSetup ? false : true;
```

DDDDDanica: Actually it would be like 

```suggestion
   if (isFirstMetaMaskControllerSetup === true) {
      isFirstMetaMaskControllerSetup = false;
   } else if (isFirstMetaMaskControllerSetup === undefined) {
      isFirstMetaMaskControllerSetup = true;
    }
      await browser.storage.session.set({ isFirstMetaMaskControllerSetup });
```

DDDDDanica: Actually it would be like 

```suggestion
     isFirstMetaMaskControllerSetup = isFirstMetaMaskControllerSetup ? false : true;
      await browser.storage.session.set({ isFirstMetaMaskControllerSetup });
```

danjm: okay, I understand what you are saying now

So what we want is:
- if `isFirstMetaMaskControllerSetup` is `false`, leave it `false`
- if it is `true`, set it to `false`
- if it is `undefined`, set it to `true`

so we would at least have to have this part:
```
isFirstMetaMaskControllerSetup = isFirstMetaMaskControllerSetup === undefined
	? true
	: isFirstMetaMaskControllerSetup ? false : true;
```

danjm: but I think the current way it is written might be clearer... I can simplify it a little bit though

DDDDDanica: I agree, the original statement is more clear. But what i think can be extracted is this line:
`await browser.storage.session.set({ isFirstMetaMaskControllerSetup });`

danjm: oh, actually, this can be way simpler

I think it can reduced to just

```
    await browser.storage.session.set({
      isFirstMetaMaskControllerSetup: isFirstMetaMaskControllerSetup === undefined
    });
```

DDDDDanica (dismissed):

DDDDDanica (approved):

pedronfigueiredo (approved):

PR #18356, opened 2023-03-28
Don't include document in scuttling config for mv3 builds

### LINKED ISSUE #18348: App doesn't load in MV3 prod build due to typeerror within sentry install
Steps to reproduce:

1. Checkout the branch found here https://github.com/MetaMask/metamask-extension/pull/18344
2. run `yarn && ENABLE_MV3=true yarn build --build-type beta dist`
3. MetaMask won't load

Screenshot of error:

Screenshot from 2023-03-28 09-32-11

That error comes from this line of the built sentry-install.js

Screenshot from 2023-03-28 09-32-53

That comes from `node_modules/@sentry/utils/esm/instrument.js`:

Screenshot from 2023-03-28 09-34-09

danjm:
Fixes https://github.com/MetaMask/metamask-extension/issues/18348

Scuttling replaces `global` with a proxy, and includes the `document` property on that proxy, but `global.document` is undefined within the mv3 service worker. This leads to a bug in this code:

```
if (!('document' in global)) {
        return;
    }

var triggerDOMHandler = triggerHandlers.bind(null, 'dom');
var globalDOMEventHandler = makeDOMEventHandler(triggerDOMHandler, true);
global.document.addEventListener('click', globalDOMEventHandler, false);
```
from `node_modules/@sentry/utils/esm/instrument.js`

Because `'document' in global` is `true` (when global is proxied as we do via the `wrapAgainstScuttling` function in `metamask-extension/development/build/utils.js`), but `global.document` is `undefined`

To solve this problem, this PR changes the build script so that the `document` property is not included in the config of the proxy.

To test this:

1. Checkout this PRs branch
2. run `yarn && ENABLE_MV3=true yarn build --build-type beta dist`
3. Install the build from `metamask-extension/builds/chrome`
4. Open this metamask installation, you should be able to successfully onboard and use MetaMask

### DIFF
--- a/development/build/config.js
+++ b/development/build/config.js
@@ -20,6 +20,7 @@ const configurationPropertyNames = [
   'DISABLE_WEB_SOCKET_ENCRYPTION',
   'METAMASK_DEBUG',
   'SKIP_OTP_PAIRING_FLOW',
+  'ENABLE_MV3',
 ];
 
 const productionConfigurationPropertyNames = [
--- a/development/build/scripts.js
+++ b/development/build/scripts.js
@@ -52,7 +52,7 @@ const {
 } = require('./transforms/remove-fenced-code');
 
 // map dist files to bag of needed native APIs against LM scuttling
-const scuttlingConfig = {
+const scuttlingConfigBase = {
   'sentry-install.js': {
     // globals sentry need to function
     window: '',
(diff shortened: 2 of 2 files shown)


weizman (dismissed):
looks good, but please notice that an exception is thrown where `Date` is inaccessible due to scuttling, although it doesn't seem to harm the functionality of the app.

danjm:
@weizman where did you see the `Date` error thrown?

pedronfigueiredo (approved):

brad-decker (approved):

weizman:
> @weizman where did you see the `Date` error thrown?

In the console when the app loaded, it's fine if it's not there anymore

PR #18358, opened 2023-03-29
Convert NetworkController to TS

### LINKED ISSUE #18286: Convert NetworkController to TypeScript (excluding tests)
In order to be able to compare differences between the version of NetworkController in this repo and the version in the core repo before we replace this version with the core version, we ought to convert NetworkController to TypeScript.

This ticket just encompasses converting the implementation file to TypeScript, but not the tests.

### Definition of Done
- Whenever applies, any change unit tested, reviewed(approved) and documented(JSDOC at least)
- [ ] We have changelog entries for any related changes
- [ ] Any changed APIs have comprehensive inline documentation
- [ ] Any changed public APIs are well covered by unit tests
- Other items
- [ ] If a planning/research ticket, then the plan has been reviewed and approved by at least 1 team members

mcmire:
## Explanation

Converting this controller to TypeScript furthers the goal of getting
this whole codebase converted, of course, but it also helps in comparing
the differences between this version of the NetworkController and the
version in the `core` repo more easily, which will ultimately help us in
coalescing the two implementations.

Closes #18286.

## Manual Testing Steps

This PR should have no functional changes. All tests should pass, there should be no TypeScript errors, and extension functionality should continue to work as it does today.

## Pre-merge author checklist

- [x] I've clearly explained:


app/scripts/controllers/network/network-controller.test.js
```diff
@@ -1100,7 +1100,7 @@ describe('NetworkController', () => {
     });
 
     describe('when the request for the latest block responds with null', () => {
-      it('stores null as whether the network supports EIP-1559', async () => {
+      it('persists false to state as whether the network supports EIP-1559', async () => {
```
mcmire, line 1103: In order to satisfy a lint error, I've changed `_determineEIP1559Compatibility` to use `?.` when checking for the latest block. This ended up simplifying the type as well so that it always returns either true or false, never null.

types/@metamask/obs-store.d.ts
```diff
@@ -0,0 +1,38 @@
+declare module '@metamask/obs-store' {
```
mcmire, line 1: This is temporary; we're using an older version of `@metamask/obs-store` where the types aren't generic enough.

app/scripts/controllers/network/network-controller.ts
```diff
@@ -0,0 +1,1171 @@
...
+   * User-customizable details for the network.
+   */
+  rpcPrefs?: {
+    blockExplorerUrl?: string;
+  };
+  /**
+   * The ID of the network configuration associated with this provider
+   * configuration.
+   */
+  id?: NetworkConfigurationId;
```
mcmire, line 197: @adonesky1 Do you know if this property is actually used? I swear I saw it being set at some point but now I can't find it.

app/scripts/controllers/network/create-network-client.ts
```diff
@@ -20,8 +20,8 @@ import {
   SafeEventEmitterProvider,
 } from '@metamask/eth-json-rpc-provider';
 import { createInfuraMiddleware } from '@metamask/eth-json-rpc-infura';
-import type { Hex } from '@metamask/utils/dist';
-import { PollingBlockTracker } from 'eth-block-tracker/dist';
+import type { Hex } from '@metamask/utils';
```
mcmire, line 23: These changes are extra, but something I stumbled across. I think we ought to add `exports` to all of our projects to prevent this from happening. I can file an issue in the module template.

package.json
```diff
@@ -378,6 +378,7 @@
     "@metamask/eslint-config-typescript": "^9.0.1",
     "@metamask/forwarder": "^1.1.0",
     "@metamask/phishing-warning": "^2.1.0",
+    "@metamask/safe-event-emitter": "^2.0.0",
```
Gudahtt, line 381: The older `safe-event-emitter` package is in `dependencies` still. We should be able to update that rather than adding another copy. I've done that in a separate PR: https://github.com/MetaMask/metamask-extension/pull/18419

app/scripts/controllers/network/create-network-client.ts
mcmire: Well, I guess we don't need these changes here though. I can extract this to a new PR.

app/scripts/controllers/network/network-controller.ts
```diff
@@ -0,0 +1,1171 @@
...
+type Block = {
+  baseFeePerGas?: unknown;
+};
+
+/**
+ * Encodes both whether or not a provider is configured for an Infura network or
+ * a non-Infura network; if for an Infura network, then which network; and if
+ * for a non-Infura network, then whether the network is being operated locally
+ * or remotely. Primarily used to build the network client and check the
+ * availability of a network.
```
Gudahtt, line 60: Nit: This sentence was confusing. Maybe we could simplify it somehow? The structure of "Both [first thing thing]; if ... then [second thing]; if ... then [second thing]" was hard to follow and doesn't seem grammatically correct.

app/scripts/controllers/network/network-controller.ts
```diff
@@ -0,0 +1,1171 @@
...
+
+/**
+ * The messenger that the NetworkController uses to publish events.
+ */
+type NetworkControllerMessenger = RestrictedControllerMessenger<
+  typeof name,
+  never,
+  NetworkControllerEvent,
+  never,
+  NetworkControllerEventType
```
Gudahtt, line 157: We don't need to list the internal events here, they're allowed by default. This generic parameter is just meant for _external_ events that are allowed.

app/scripts/controllers/network/network-controller.ts
Gudahtt: It has definitely never been used as "provider configuration". Our middleware stack doesn't use this ID for anything.

Gudahtt:
Looks good! Just finished a first pass and the only feedback I had was regarding comments.

All of the related PRs have been merged, so it should be possible to rebase this on `develop` and make this ready for review now.

app/scripts/controllers/network/network-controller.ts
mcmire: You're right, this was confusing. Hopefully this is better: [84140af](https://github.com/MetaMask/metamask-extension/pull/18358/commits/84140afaf0b2ea0dd70a6bc14fa9cfcbad1c11cb)

app/scripts/controllers/network/network-controller.ts
```diff
@@ -0,0 +1,1171 @@
...
+};
+
+/**
+ * Type guard for determining whether the given value is an error object with a
+ * `code` property, such as an instance of Error.
+ *
+ * TODO: Move this to @metamask/utils
+ *
+ * @param error - The object to check.
+ * @returns True or false, depending on the result.
```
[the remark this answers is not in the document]
mcmire: Good call. Fixed in [c890d90](https://github.com/MetaMask/metamask-extension/pull/18358/commits/c890d904f3a9e1e938aff4cac5132df7329352f3).

app/scripts/controllers/network/network-controller.ts
```diff
@@ -0,0 +1,1171 @@
...
+/**
+ * Asserts that the given value is a network ID, i.e., that it is a decimal
+ * number represented as a string.
+ *
+ * @param value - The value to check.
+ */
+function assertNetworkId(value: any): asserts value is NetworkId {
+  assert(
+    /^\d+$/u.test(value) && !Number.isNaN(Number(value)),
+    'value cannot be converted to a number',
```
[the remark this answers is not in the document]
mcmire: Makes sense. Updated in [f243281](https://github.com/MetaMask/metamask-extension/pull/18358/commits/f2432813788972a7004aa1b8fc5a4d95404b3eb4).

app/scripts/controllers/network/network-controller.ts
mcmire: Okay, I've removed this for now: [de0c195](https://github.com/MetaMask/metamask-extension/pull/18358/commits/de0c195d6a1633f3af189be91b548e5e669f9af2). Happy to add this back later though if we need it.

app/scripts/controllers/network/create-network-client.ts
mcmire: Removed for now: [88811d2](https://github.com/MetaMask/metamask-extension/pull/18358/commits/88811d2e189b70e0bd3b50c427486acb201e9d4f)

app/scripts/controllers/network/network-controller.ts
Gudahtt: That looks great! I'm glad I didn't suggest an alternative, this is better than what I would have come up with.

types/@metamask/obs-store.d.ts
```diff
@@ -0,0 +1,18 @@
+import { ObservableStore } from '@metamask/obs-store';
+
+declare module '@metamask/obs-store' {
+  export class ComposedStore<
```
mcmire, line 4: Overriding this is still necessary because the types for ComposedStore assume that each store is an object.

mcmire: @Gudahtt Should I try to update `@metamask/obs-store` with these changes?

Gudahtt: Ah interesting. Yeah I guess so? I see that in the ObservableStore class they made the same mistake in assuming the state was an object, but then fixed it by using generics instead. Applying the same fix to the composable store would make sense.

mcmire: Okay, PRs created here: https://github.com/MetaMask/obs-store/pull/79, https://github.com/MetaMask/obs-store/pull/80

mcmire:
@Gudahtt I want to get a new release for `@metamask/obs-store` out first and then I'll open this for review.

types/@metamask/obs-store.d.ts
mcmire: With the new release, I was able to remove this file.

Gudahtt (dismissed):
LGTM!

Gudahtt (approved):

legobeat (approved):

PR #18371, opened 2023-03-29
[FLASK] `snaps-monorepo@0.32.2`

FrederikBolding:
## Explanation

Bumps Snaps packages to `0.32.2` and handles any breaking changes.

Also adds new URLs and metadata for execution environments and the snap registry.


GuillaumeRx (dismissed):

app/scripts/metamask-controller.js
```diff
@@ -688,6 +689,8 @@ export default class MetamaskController extends EventEmitter {
           `${this.approvalController.name}:hasRequest`,
           `${this.approvalController.name}:acceptRequest`,
           `${this.approvalController.name}:rejectRequest`,
+          `SnapController:getPermitted`,
+          `SnapController:install`,
```
hmalik88, line 693: why not `${this.snapController.name}:getPermitted` and ${this.snapController.name}:install`?

david0xd (dismissed):

app/scripts/metamask-controller.js
FrederikBolding: No reason, we can do that 👍

GuillaumeRx: because `${this.snapController.name}` is undefined

FrederikBolding: Oh yeah, we can't since the PermissionController is defined before the SnapController

FrederikBolding: We could move it around, but I don't think that its worth the hassle right now

hmalik88 (dismissed):
LGTM, let's not forget to move around the order and update with the controller name in the future :)

FrederikBolding:
> LGTM, let's not forget to move around the order and update with the controller name in the future :)

We need a better way to do it I think, because SnapController requires PermissionController actions and now the reverse is also true. One of them has to be defined first unless we make the names static somehow.

GuillaumeRx (approved):

hmalik88 (approved):
LGTM

david0xd (approved):

bowensanders:
QA: Manual and Automated tests pass for Chrome and Firefox [Flask].

PR #18383, opened 2023-03-30
Change api used for opensea validation

jpuri:
Change api used for opensea validations.

### DIFF
--- a/app/scripts/lib/security-provider-helpers.js
+++ b/app/scripts/lib/security-provider-helpers.js
@@ -51,17 +51,15 @@ export async function securityProviderCheck(
   }
 
   const response = await fetchWithTimeout(
-    'https://eos9d7dmfj.execute-api.us-east-1.amazonaws.com/metamask/validate',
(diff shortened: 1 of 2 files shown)


app/scripts/lib/security-provider-helpers.js
```diff
@@ -51,17 +51,15 @@ export async function securityProviderCheck(
   }
 
   const response = await fetchWithTimeout(
-    'https://eos9d7dmfj.execute-api.us-east-1.amazonaws.com/metamask/validate',
+    'https://proxy.metafi.codefi.network/opensea/security/v1/validate',
```
digiwand, line 54: I think we should implement error handling in a follow-up PR

digiwand (dismissed):

seaona:
From QA looks good. The warnings continue to appear and the network requests are sent to the new endpoint

image

danjm (approved):

digiwand (approved):

PR #18386, opened 2023-03-30
Display popup appStateController using ApprovalController

vinistevam:
## Explanation

Changes the AppStateController to use ApprovalController instead of `showUserConfirmation` callback to display the popup.

* Fixes https://github.com/MetaMask/MetaMask-planning/issues/405

## Manual Testing Steps

- Restart your extension so you have sure it's locked
- Go to [dapp test page](https://metamask.github.io/test-dapp/)
- Try to connect
- A popup will be displayed to unlock MM and connect

result: Operation should work as usual, connecting to the dapp test and enabling to perform any action there.

## Pre-merge author checklist

- [x] I've clearly explained:


app/scripts/controllers/app-state.js
```diff
@@ -10,6 +11,8 @@ import {
   POLLING_TOKEN_ENVIRONMENT_TYPES,
 } from '../../../shared/constants/app';
 
+const APPROVAL_REQUEST_TYPE = 'unlock';
```
matthewwalsh0, line 14: Harmless currently, but just for reference, I've created a future ticket to address these types.

Potentially to create an `APPROVAL_TYPE` enum, probably in core, so these can be centrally defined meaning they can be checked and created from different places, plus synchronised across the different clients.

app/scripts/controllers/app-state.js
```diff
@@ -74,6 +75,9 @@ export default class AppStateController extends EventEmitter {
 
     const { preferences } = preferencesStore.getState();
     this._setInactiveTimeout(preferences.autoLockTimeLimit);
+
+    this._messenger = messenger;
```
matthewwalsh0, line 79: I think we should try and match `BaseControllerV2` for consistency sake, so the constructor argument stays as `messenger` but the internal property would be called `messagingSystem`.

app/scripts/controllers/app-state.js
```diff
@@ -122,6 +126,8 @@ export default class AppStateController extends EventEmitter {
       }
       this.emit(METAMASK_CONTROLLER_EVENTS.UPDATE_BADGE);
     }
+
+    this._acceptApprovalRequest();
```
matthewwalsh0, line 130: Perhaps worth a comment here confirming all approval requests are automatically rejected by the background if the unlock is ignored and the popup closed? I've assumed that is the current behaviour, worth verifying.

app/scripts/controllers/app-state.test.js
```diff
@@ -29,5 +44,259 @@ describe('AppStateController', () => {
...
+  describe('getCurrentPopupId', () => {
+    it('retrieves the currentPopupId saved in the appState', () => {
+      const popupId = 'popup1';
+
+      appStateController.setCurrentPopupId(popupId);
+      expect(appStateController.getCurrentPopupId()).toBe(popupId);
+    });
+  });
+
+  describe('setFirstTimeUsedNetwork', () => {
```
matthewwalsh0, line 292: Just to confirm, have we added all these new unit tests for the general domain for the sake of code coverage?

app/scripts/controllers/app-state.test.js
```diff
@@ -29,5 +44,259 @@ describe('AppStateController', () => {
...
+
+      expect(emitSpy).toHaveBeenCalled();
+      expect(emitSpy).toHaveBeenCalledWith('updateBadge');
+      expect(appStateController._messenger.call).toHaveBeenCalled();
+      expect(appStateController._messenger.call).toHaveBeenCalledWith(
+        'ApprovalController:acceptRequest',
+        expect.any(String),
+      );
+    });
+    it('returns without call messenger if no approval request in pending', async () => {
```
matthewwalsh0, line 141: If we add a `try catch` to the accept, should we also have a unit test to verify if the `call` throws, the controller doesn't.

app/scripts/controllers/app-state.js
```diff
@@ -369,4 +375,31 @@ export default class AppStateController extends EventEmitter {
...
+      true,
+    );
+  }
+
+  _acceptApprovalRequest() {
+    if (!this._approvalRequestId) {
+      return;
+    }
+
+    this._messenger.call(
```
matthewwalsh0, line 398: It may be unlikely, but to avoid breaking any flows because of this, shall we `try` and `catch` around this call to log any errors rather than throwing?

app/scripts/controllers/app-state.js
```diff
@@ -369,4 +375,31 @@ export default class AppStateController extends EventEmitter {
...
   }
+
+  _createApprovalRequest() {
+    this._approvalRequestId = uuid();
+
+    this._messenger.call(
+      'ApprovalController:addRequest',
+      {
+        id: this._approvalRequestId,
+        origin: this._messenger.controllerName,
```
matthewwalsh0, line 386: I've since found that the `origin` is either the host of the dApp that started the flow, or `metamask` to imply it was internal. Don't think a constant exists for it but worth checking.

app/scripts/controllers/app-state.js
```diff
@@ -369,4 +375,31 @@ export default class AppStateController extends EventEmitter {
...
+        type: APPROVAL_REQUEST_TYPE,
+      },
+      true,
+    );
```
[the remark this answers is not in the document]
vinistevam: Make sense, I've added the `catch`

app/scripts/controllers/app-state.test.js
vinistevam: yeap, once I've applied your suggestions above I also included a test case to log if rejecting an approval request throws.

app/scripts/controllers/app-state.js
vinistevam: sure, wrapped `this.messagingSystem` in a `try/catch` to avoid throwing.

app/scripts/controllers/app-state.js
vinistevam: yes, it rejects and updates the badge number when the popup is closed.

app/scripts/controllers/app-state.test.js
vinistevam: yes, I was hoping those would be enough to reach the target of the codecov.

matthewwalsh0 (approved):

bergarces (approved):

PR #18400, opened 2023-03-31
Trigger transaction popup using ApprovalController

matthewwalsh0:
## Explanation

Trigger the transaction popup using calls to the `ApprovalController` instead of the `showUserConfirmation` callback.

## Manual Testing Steps

Verify creation of transactions using the test dApp.

## Pre-merge author checklist

- [x] I've clearly explained:


shared/constants/app.ts
```diff
@@ -53,6 +53,7 @@ export const MESSAGE_TYPE = {
   PERSONAL_SIGN: 'personal_sign',
   SEND_METADATA: 'metamask_sendDomainMetadata',
   SWITCH_ETHEREUM_CHAIN: 'wallet_switchEthereumChain',
+  TRANSACTION: 'transaction',
```
matthewwalsh0, line 56: I'm not a huge fan of this, I'll create an alternate ticket to create a new `APPROVAL_TYPE` constant to be explicit.

app/scripts/controllers/transactions/index.js
```diff
@@ -2596,4 +2604,54 @@ export default class TransactionController extends EventEmitter {
       },
     );
   }
+
+  _requestApproval(txMeta) {
+    const id = this._getApprovalId(txMeta);
+    const { origin } = txMeta;
+    const type = MESSAGE_TYPE.TRANSACTION;
+    const requestState = { txId: txMeta.id };
```
bergarces, line 2612: I'm just wondering. What's the purpose of passing `requestState` to `ApprovalController` in this instance.

I thought that was only meant for cases in which we want to update that state before the confirmation is closed.

app/scripts/controllers/transactions/index.test.js
```diff
@@ -1195,6 +1284,22 @@ describe('Transaction Controller', function () {
...
+
+      assert.equal(messengerMock.call.callCount, 1);
+      assert.deepEqual(messengerMock.call.getCall(0).args, [
+        'ApprovalController:rejectRequest',
+        '0',
+        new Error('Rejected'),
+      ]);
+    });
+
+    it('should not throw if rejecting approval request throws', async function () {
```
bergarces, line 1299: From a look at the `ControllerMessenger` code, calling an action can fail when:
- No action is passed.
- Action is not in the allow list
- Action is not registered

Given that we are making these calls internally, we should not see it failing but... wouldn't it be best to let it fail so that we can catch any typos or miss-configurations during development?

bergarces (approved):

vinistevam (approved):

jpuri (approved):
LGTM

PR #18403, opened 2023-03-31
Ensure that firefox linting of beta builds only happens on beta build…

danjm:
Fixes the failure we are seeing on release builds here: https://app.circleci.com/pipelines/github/MetaMask/metamask-extension/41331/workflows/b6b2b01e-425f-452b-8dcb-409605c61e98/jobs/1149313

This is happening because  the beta build isn't happening:
Commit message does not match commit message for beta pattern; skipping beta automation build
from https://app.circleci.com/pipelines/github/MetaMask/metamask-extension/41331/workflows/b6b2b01e-425f-452b-8dcb-409605c61e98/jobs/1149172/steps, and the linting of that beta build is running even when the build doesn't exist

This PR ensures that the `yarn mozilla-lint` command is only run under the same conditions under which the beta build is created. In particular, if the commit message matches the beta commit structure we have set for triggering a beta build.

### DIFF
--- a/.circleci/scripts/mozilla-lint-beta.sh
+++ b/.circleci/scripts/mozilla-lint-beta.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+set -e
+set -u
+set -o pipefail
+
+# => Version-v10.24.1
+version="${CIRCLE_BRANCH/Version-v/}"
+current_commit_msg=$(git show -s --format='%s' HEAD)
+
+if [[ $current_commit_msg =~ Version[[:space:]](v[[:digit:]]+.[[:digit:]]+.[[:digit:]]+[-]beta.[[:digit:]]) ]]
+then
+    # filter the commit message like Version v10.24.1-beta.1
+    printf '%s\n' "Linting beta builds for firefox"
+    # Move beta build to dist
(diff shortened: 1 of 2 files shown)


Gudahtt (approved):
LGTM!

.circleci/scripts/mozilla-lint-beta.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+set -e
+set -u
+set -o pipefail
+
+current_commit_msg=$(git show -s --format='%s' HEAD)
+
+if [[ $current_commit_msg =~ Version[[:space:]](v[[:digit:]]+.[[:digit:]]+.[[:digit:]]+[-]beta.[[:digit:]]) ]]
```
Gudahtt, line 9: Would be nice to share this criteria between bash scripts

danjm: @Gudahtt that sounds good... any guidance on how to do that in bash?

danjm: cool, chatted with Mark

We could move this code to another script and invoke that here, but we can do that in the future

pedronfigueiredo (approved):

PR #18458, opened 2023-04-05
[FLASK] Implement WebWorkerExecutionService

Mrtenz:
## Description

This pull request replaces the existing controller's execution service with the `WebWorkerExecutionService`.
The `WebWorkerExecutionService` uses web workers to run calculation tasks in a separate thread, which helps
to keep all snaps responsive.

## Changes

1. Replace the `SnapController`'s execution service with the `WebWorkerExecutionService`.

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -58,7 +58,7 @@ import {
   CronjobController,
   JsonSnapsRegistry,
   SnapController,
-  IframeExecutionService,
+  WebWorkerExecutionService,
 } from '@metamask/snaps-controllers';
 ///: END:ONLY_INCLUDE_IN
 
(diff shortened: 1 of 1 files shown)


app/scripts/metamask-controller.js
```diff
@@ -781,7 +781,7 @@ export default class MetamaskController extends EventEmitter {
     this.snapExecutionService =
       this.opts.overrides?.createSnapExecutionService?.(
         snapExecutionServiceArgs,
-      ) || new IframeExecutionService(snapExecutionServiceArgs);
+      ) || new WebWorkerExecutionService(snapExecutionServiceArgs);
```
Mrtenz, line 784: This needs to be updated to use @ritave's build types work, so we can create a separate build for iframes and web workers.

Mrtenz: - #18125.

PR #18476, opened 2023-04-05
Convert NetworkController unit tests to TypeScript

### LINKED ISSUE #18424: Convert NetworkController unit tests to TypeScript
In order to be able to compare differences between the version of NetworkController in this repo and the version in the core repo before we replace this version with the core version, we ought to convert NetworkController to TypeScript.

This ticket just encompasses converting the unit tests to TypeScript, and is a companion to https://github.com/MetaMask/metamask-extension/issues/18286.

### Definition of Done
- Whenever applies, any change unit tested, reviewed(approved) and documented(JSDOC at least)
- [ ] We have changelog entries for any related changes
- [ ] Any changed APIs have comprehensive inline documentation
- [ ] Any changed public APIs are well covered by unit tests
- Other items
- [ ] If a planning/research ticket, then the plan has been reviewed and approved by at least 1 team members

mcmire:
## Explanation

This helps us more easily compare the unit tests for NetworkController in this repo and the NetworkController in the `core` repo.

Closes #18424.

## Manual Testing Steps

No functional changes. All tests should pass.

## Pre-merge author checklist

- [x] I've clearly explained:


jest.config.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
   collectCoverageFrom: [
     '<rootDir>/app/scripts/constants/error-utils.js',
     '<rootDir>/app/scripts/controllers/network/**/*.js',
+    '<rootDir>/app/scripts/controllers/network/**/*.ts',
```
Gudahtt, line 5: We should make the equivalent change to the Mocha configuration as well, so it's not counted towards coverage there.

app/scripts/controllers/network/network-controller.ts
```diff
@@ -1067,7 +1070,7 @@ export class NetworkController extends EventEmitter {
       referrer,
       source,
     }: {
-      setActive: boolean;
+      setActive?: boolean;
```
Gudahtt, line 1073: Should this be added to the TypeScript migration PR?

app/scripts/controllers/network/network-controller.ts
```diff
@@ -154,20 +154,22 @@ type NetworkControllerEvent =
...
-type NetworkControllerMessenger = RestrictedControllerMessenger<
+export type NetworkControllerMessenger = RestrictedControllerMessenger<
   typeof name,
   never,
   NetworkControllerEvent,
   never,
-  NetworkControllerEventType
+  // TODO: Why is this `string` and not `NetworkControllerEventType`?
+  // Also this breaks `waitForPublishedEvents` — why?
+  string
```
Gudahtt, line 164: Hmm, I'm curious what is going on here

app/scripts/controllers/network/network-controller.ts
mcmire: Yes, good call.

app/scripts/controllers/network/network-controller.ts
mcmire: If I change this to `NetworkControllerEventType`, I get the following error in `waitForPublishedEvents` at the callsites for `subscribe` and `unsubscribe`:

`[tsserver 2345] [E] Argument of type '(...args: E["payload"]) => void' is not assignable to parameter of type 'ExtractEventHandler<NetworkControllerNetworkDidChangeEvent, E["type"]> | ExtractEventHandler<NetworkControllerNetworkWillChangeEvent, E["type"]> | ExtractEventHandler<...> | ExtractEventHandler<...>'.`

If I change 

``` typescript
let eventListener: (...args: E['payload']) => void;
```

to

``` typescript
let eventListener: ExtractEventHandler<E, E['type']>;
```

then I don't get that error at the `subscribe` or `unsubscribe` callsites, but I do get an error when attempting to set `eventListener`:

`[tsserver 2322] [E] Type '(...payload: E['payload']) => void' is not assignable to type 'ExtractEventHandler<E, E["type"]>'.`

And using `string` as I do now makes that error go away.

I feel like this is how RestrictedControllerMessenger was designed to be used. You can specify which events the messenger officially recognizes, but then you can subscribe and unsubscribe to any event: https://github.com/MetaMask/core/blob/a3b454eadbc9e974a6ee4a2729baf54171d422ed/packages/base-controller/src/ControllerMessenger.ts#L285. Perhaps this is to account for this library being used in a JavaScript context. In any case, because of this, it needs to infer the type of the event handler, and I guess that doesn't work very well? Based on the fact that TypeScript is expanding `ExtractEventHandler<E, E["type"]>` to `ExtractEventHandler<NetworkControllerNetworkDidChangeEvent, E["type"]> | ExtractEventHandler<NetworkControllerNetworkWillChangeEvent, E["type"]> | ExtractEventHandler<...> | ExtractEventHandler<...>`, it makes me think that something is disconnected in the types. I'm just not sure what.

I guess this isn't a big deal, I can remove the TODO, but it is slightly bothersome.

jest.config.js
mcmire: Where would I need to make this change?

Gudahtt: The file is `.mocharc.js`, you should see a list there similar to this one

app/scripts/controllers/network/network-controller.ts
Gudahtt: The restricted controller messenger was designed to be used solely within the controller itself. That seems like the problem here - the tests shouldn't be using it at all. In tests we can use the top-level restricted messenger. It was definitely not designed to allow subscribing to any event.

jest.config.js
mcmire: Ah, thanks. Like this? [0b0c576](https://github.com/MetaMask/metamask-extension/pull/18476/commits/0b0c5765f69e72dc0ce7c6fcb7b4fa539f99b2c7)

app/scripts/controllers/network/network-controller.ts
mcmire: Sorry, something's not connecting for me. In order to use the NetworkController, we have to pass a RestrictedControllerMessenger that recognizes NetworkController events. So we can't make a generic restricted messenger, can we? 

I feel like we have a lot of bad examples in the core repo and elsewhere on how to do this properly. Is there a good example I can use?

jest.config.js
Gudahtt: So that mocha config is a list of test suites that are run, not files to gather coverage from. I was suggesting adding `<rootDir>/app/scripts/controllers/network/**/*test.ts` to that list. However I think I was wrong.

I had suggested that it needed an update because I assumed that some of our Mocha tests were TypeScript, but [apparently not](https://github.com/MetaMask/metamask-extension/blob/develop/test/run-unit-tests.js#L74). Sorry about that, that change looks to be unnecessary after all.

app/scripts/controllers/network/network-controller.ts
Gudahtt: Right, so we need to create a restricted messenger to pass to the controller, but we create that from the top-level messenger. We can use that top-level messenger in tests.

Gudahtt: Here: https://github.com/MetaMask/metamask-extension/pull/18476/files#diff-89fa1edc4a981c9e4d156ed188d409ca35653a1de7ef939648f64ca7be9cbb03R7672
We're constructing a top-level messenger then throwing it away, returning just the restricted one. Instead we should be using the top-level messenger in our test helpers, and using the restricted one just for the constructor options

jest.config.js
mcmire: Cool. No problem! Reverted.

app/scripts/controllers/network/network-controller.ts
Gudahtt: I don't know how to explain that type error you encountered though. I'm unsure whether switching to the root messenger would solve that.

mcmire: Oh, right, because the restricted controller messenger modifies the top-level messenger, so the effects of the restricted controller messenger can be accessed via the top-level messenger. Does that sound correct?

Gudahtt: Well, it routes everything to the top-level messenger. The restricted messenger is a proxy for the top-level one that makes TypeScript complain if you try using something you aren't allowed to.

mcmire: Okay, I'm now using the unrestricted messenger here: [627c0a4](https://github.com/MetaMask/metamask-extension/pull/18476/commits/627c0a4a2c7e7c13e91480b4c14eb7e3fc556da5). This is a better practice, but unfortunately I'm still getting the error above. It seems like the unrestricted version of `subscribe` [tries to infer the payload type from the given event](https://github.com/MetaMask/core/blob/a3b454eadbc9e974a6ee4a2729baf54171d422ed/packages/base-controller/src/ControllerMessenger.ts#L488). However, I'm not sure why, as we know the given event type is one of the possible event types! It seems odd that the method signature is not this:

``` typescript
subscribe<E>(
  eventType: E['type'],
  handler: (...payload: E['payload') => void,
): void;
```

Anyway, for now I've just marked `eventListener` as `any`, and I'll investigate this further later.

Gudahtt: I think I've found a minimal example of this problem: https://www.typescriptlang.org/play?#code/C4TwDgpgBAogHgQwLZgDYRgNwgO2FAXigG8AoKKMBEVAewQBMAuKAZ2ACcBLHAcwBpSAXwDcpUqEiw4nBAGNgABWp1GAHiy5gAPkKxseKBBm4GrEuUor6zKDwBmEDlEVihlgPwvLLHBGwcYqT2AK44Cly0OFByHBAIwBAAEgg4DOgcGkYmaebwyGgYBjoAFFQ0NiwwANoA5OWqDLUAugCUVTIc8krW6jC6ZBRxwCEc0Q02bqRAA

It seems that TypeScript doesn't know how to reconcile a generic type with a conditional type.

Using `any` seems like the best plan for now, but we may be able to fix this by reducing the use of conditional types in the controller messenger.

app/scripts/controllers/network/network-controller.test.ts
```diff
@@ -7998,3 +8169,40 @@ async function waitForLookupNetworkToComplete({
...
+/**
+ * Returns whether two places in different state objects have different values.
+ *
+ * @param currentState - The current state object.
+ * @param prevState - The previous state object.
+ * @param propertyPath - A property path within both objects.
+ * @returns True or false, depending on the result.
+ */
+function isStateChangeInteresting(
+  currentState: Record<string | symbol | number, unknown>,
```
Gudahtt, line 8182: Nit: The built-in type `PropertyKey` can be used anyplace `string | symbol | number` is used

app/scripts/controllers/network/network-controller.ts
```diff
@@ -108,44 +108,44 @@ export enum NetworkControllerEventType {
  * switched, but the new provider has not been created and no state changes have
  * occurred yet.
  */
-type NetworkControllerNetworkWillChangeEvent = {
+export type NetworkControllerNetworkWillChangeEvent = {
   type: NetworkControllerEventType.NetworkWillChange;
-  payload: [];
+  payload: never[];
```
Gudahtt, line 113: Why would `never` be used here? I'd expect `never` to only ever show up in conditional, unreachable branches. Having it as a property implies that this type is never used, doesn't it?

app/scripts/controllers/network/network-controller.test.ts
```diff
@@ -7888,72 +8040,88 @@ async function waitForPublishedEvents({
...
+  eventType: E['type'];
+  count?: number;
+  filter?: (payload: E['payload']) => boolean;
+  wait?: number;
+  operation?: () => void | Promise<void>;
+  beforeResolving?: () => void | Promise<void>;
+}): Promise<E['payload'][]> {
+  const promiseForEventPayloads = new Promise<E['payload'][]>(
+    (resolve, reject) => {
+      // Everything depends on `eventListener` circularly, so we need to declare
```
Gudahtt, line 8054: Nit: This could be simplifed by using a function declaration for `end`, `stopTimer`, and `resetTimer`. Function declarations make circular references easier to handle due to function hoisting; the declaration is hoisted to the top of the scope, so they can be referenced before the declaration.

That lets use `const` when defining the event listener, and saves us from needing this comment to explain the awkward order of operations.

Gudahtt (dismissed):
LGTM! A few comments/suggestions remaining but no blockers

app/scripts/controllers/network/network-controller.test.ts
mcmire: Oh, that's neat! Yeah, that's a lot better.

app/scripts/controllers/network/network-controller.test.ts
```diff
@@ -1,64 +1,195 @@
...
+ */
+type WithControllerCallback<NetworkCommunications, ReturnValue> = (args: {
+  controller: NetworkController;
+  network: NetworkCommunications;
+}) => Promise<ReturnValue> | ReturnValue;
+
+/**
+ * The options that `withController` takes.
+ */
+type WithControllerOptions = Partial<NetworkControllerOptions>;
```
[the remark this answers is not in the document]
mcmire: Makes sense: [1b03c16](https://github.com/MetaMask/metamask-extension/pull/18476/commits/1b03c16f4d3d39616287aec660b419ab88cbf1d3)

app/scripts/controllers/network/network-controller.test.ts
```diff
@@ -1,64 +1,195 @@
...
+
+/**
+ * A variant of the options that the NetworkController constructor takes, where
+ * the provider state has been preconfigured with an Infura network. This is
+ * extracted so that we can give `withController` a better signature.
+ */
+type NetworkControllerOptionsWithInfuraProviderConfig = Partial<
+  Omit<WithControllerOptions, 'state'>
```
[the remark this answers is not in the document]
mcmire: I thought so too. I was having trouble with this when I was updating `withController` and must have thought this would help TypeScript. You're right though that it's unnecessary. Updated in [4dd0879](https://github.com/MetaMask/metamask-extension/pull/18476/commits/4dd0879fcc718415eaa2de8d333ca9b2871df2ce).

Gudahtt (approved):

app/scripts/controllers/network/network-controller.test.ts
```diff
@@ -384,23 +498,95 @@ class NetworkCommunications {
...
+/**
+ * Handles mocking requests made by NetworkController for an Infura network.
+ */
+class InfuraNetworkCommunications extends NetworkCommunications<InfuraNetworkCommunicationsOptions> {
+  /**
+   * Constructs an InfuraNetworkCommunications.
+   *
+   * @param args - The arguments.
+   * @param args.infuraProjectId - TODO.
+   * @param args.infuraNetwork - TODO.
```
digiwand, line 519: what are these TODO comments for? 👀 

@mcmire

digiwand (approved):
LGTM! nice cleanup with the conversion and thanks for the helpful comments

small, non-blocking question in regards to 2 TODO comments

app/scripts/controllers/network/network-controller.test.ts
mcmire: Oops, good catch, thanks. Fortunately I'm going to update this test file again in another PR very soon so I will address them then.

digiwand: 👍

PR #18448, opened 2023-04-04
update ButtonIcon to TS

garrettbear:
## Explanation

Update ButtonIcon to TS and move JS version to a deprecated folder
* Fixes #18327 

For any existing production components using ButtonIcon, the import was updated to point toward the deprecated folders. 
_Before_
`import { ButtonIcon } from '../../component-library';`
_After_
`import { ButtonIcon } from '../../component-library/button-icon/deprecated';`

## Screenshots/Screencaps

No visual UI change

## Manual Testing Steps
`yarn jest ui/components/component-library/button-icon`
`yarn jest ui/components/component-library`

## Pre-merge author checklist

- [X] I've clearly explained:
  - [X] What problem this PR is solving
  - [X] How this problem was solved
  - [X] How reviewers can test my changes
- [X] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [X] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
- [X] PR is linked to the appropriate GitHub issue
- [ ] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.

In this case, a QA Engineer approval will be be required.

PR #18503, opened 2023-04-07
Some suggestions

georgewrmarshall:
## Explanation
Some suggestions to https://github.com/MetaMask/metamask-extension/pull/18448


ui/components/component-library/button-icon/button-icon.tsx
```diff
@@ -11,10 +11,15 @@ import {
 } from '../../../helpers/constants/design-system';
 
 import Box from '../../ui/box';
-import { Icon } from '../icon';
+import { Icon, IconSize } from '../icon';
 
 import { ButtonIconSize, ButtonIconProps } from './button-icon.types';
 
+const buttonIconSizeToIconSize: Record<ButtonIconSize, IconSize> = {
```
georgewrmarshall, line 18: ChatGPT: We use Record here to define a type that describes a mapping between ButtonIconSize and IconSize enum values.

The Record type is a built-in TypeScript utility type that creates a new type with a set of properties defined by the provided keys, each with a value of the same type. In this case, the keys are the ButtonIconSize enum values, and the values are the corresponding IconSize enum values.

Using Record provides a way to ensure that each ButtonIconSize value is mapped to a corresponding IconSize value. If there is a missing or incorrect mapping, TypeScript will raise a type error.

In the specific case of buttonIconSizeToIconSize, we are mapping the ButtonIconSize values of Sm and Lg to their corresponding IconSize values of Sm and Lg respectively.

ui/components/component-library/button-icon/button-icon.tsx
```diff
@@ -32,20 +37,21 @@ export const ButtonIcon = React.forwardRef(
     ref: React.Ref<HTMLElement>,
   ) => {
     const Tag = href ? 'a' : as;
+    const isDisabled = disabled && Tag === 'button';
```
georgewrmarshall, line 40: We only want to pass disabled if the html element is a button

ui/components/component-library/button-icon/button-icon.types.ts
```diff
@@ -1,10 +1,10 @@
 import type { BoxProps } from '../../ui/box/box.d';
-import { IconName, IconSize } from '../icon';
+import { IconName, IconProps } from '../icon';
 import { Color, IconColor } from '../../../helpers/constants/design-system';
 
 export enum ButtonIconSize {
-  Sm = IconSize.Sm,
-  Lg = IconSize.Lg,
+  Sm = 'sm',
+  Lg = 'lg',
```
georgewrmarshall, line 7: I've spoken with Brad and Elliot and unfortunately we don't have a good way to map one enum value to another so we will just have to be vigilant and use the same strings across the board for our size enums

PR #18506, opened 2023-04-07
fix: remove unnecessaru parameter when calling '_getPopup' 

zhangyouxin:
Fixed method call error in `app/scripts/lib/notification-manager.js`

## Explanation

The `_getPopup` method of `NotificationManager` class doesn't need any parameters.

refer to: https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/lib/notification-manager.js#L101-L104


zhangyouxin:

PR #17677, opened 2023-02-09
Update Keystone links

soralit:
## Explanation
Hello there 😄 
Just raise this PR to update `Buy Now` to `Learn More` and related link for Keystone. 

### Before
image

### After
image

## Pre-merge author checklist

- [ ] I've clearly explained:
  - [ ] What problem this PR is solving
  - [ ] How this problem was solved
  - [ ] How reviewers can test my changes
- [ ] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [x] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
- [ ] PR is linked to the appropriate GitHub issue
- [ ] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.

In this case, a QA Engineer approval will be be required.

PR #18508, opened 2023-04-07
Fix diff generation for fitness functions

pedronfigueiredo:
## Explanation

To fix the underlying issue reported on https://github.com/MetaMask/metamask-extension/pull/17754, we implemented a different mechanism to generate the diff for fitness functions that supports PRs opened by external contributors using repository forks.

We don't need to fetch HEAD from the origin, instead directly using the one available in the CI.

### DIFF
--- a/.github/workflows/fitness-functions.yml
+++ b/.github/workflows/fitness-functions.yml
@@ -17,12 +17,13 @@ jobs:
         with:
           node-version: '16'
 
+      - name: Install dependencies
+        run: yarn
+
       - name: Run fitness functions
         env:
-          HEAD_REF: ${{ github.event.pull_request.head.ref }}
           BASE_REF: ${{ github.event.pull_request.base.ref }}
         run: |
-          # git fetch origin $HEAD_REF
-          # git fetch origin $BASE_REF
-          # git diff origin/$BASE_REF origin/$HEAD_REF -- . > diff
-          # npm run fitness-functions -- "ci" "./diff"
+          git fetch origin $BASE_REF
+          git diff origin/$BASE_REF HEAD -- . > diff
+          npm run fitness-functions -- "ci" "./diff"


pedronfigueiredo:
> Can the new behavior be exhibited?

> For example, on a PR in your fork repo targeting the branch of this PR from a third fork?

Not sure I follow your comment, @legobeat. This was the original scenario: a developer opened a PR from a fork into develop. You can verify on this PR that the job now passes and that it didn't [here](https://github.com/MetaMask/metamask-extension/pull/17677).

DDDDDanica (approved):

brad-decker (approved):

.github/workflows/fitness-functions.yml
```diff
@@ -17,12 +17,13 @@ jobs:
...
         env:
-          HEAD_REF: ${{ github.event.pull_request.head.ref }}
           BASE_REF: ${{ github.event.pull_request.base.ref }}
         run: |
-          # git fetch origin $HEAD_REF
-          # git fetch origin $BASE_REF
-          # git diff origin/$BASE_REF origin/$HEAD_REF -- . > diff
-          # npm run fitness-functions -- "ci" "./diff"
+          git fetch origin $BASE_REF
+          git diff origin/$BASE_REF HEAD -- . > diff
```
Gudahtt, line 28: It looks like `BASE_REF` is expected to be the target branch (e.g. `develop`) at the time this step is run. This seems like the wrong point of comparison because this branch might be missing changes present on `develop` causing the fitness check to pass/fail when it shouldn't.

We had a similar problem when the bundle size check was implemented, and we arrived at this solution: 
https://github.com/MetaMask/metamask-extension/blob/a4a5b28f2e66497bcf5d1a139320322e5c44e075/.circleci/config.yml#L956

To translate to a format that makes sense for GitHub Actions, it'd be something like

```
git diff "$(git merge-base $BASE_REF HEAD)" HEAD -- . > diff
```

This would provide the most recent target branch (e.g. `develop`) commit present on this branch, using that as the comparison commit. This ensures that the check will only pass or fail due to changes in that PR directly.

.github/workflows/fitness-functions.yml
```diff
@@ -17,12 +17,13 @@ jobs:
...
       - name: Run fitness functions
         env:
-          HEAD_REF: ${{ github.event.pull_request.head.ref }}
           BASE_REF: ${{ github.event.pull_request.base.ref }}
         run: |
-          # git fetch origin $HEAD_REF
-          # git fetch origin $BASE_REF
-          # git diff origin/$BASE_REF origin/$HEAD_REF -- . > diff
-          # npm run fitness-functions -- "ci" "./diff"
+          git fetch origin $BASE_REF
```
Gudahtt, line 27: Nit: One alternative would be to set `depth` to `0` in the checkout step earlier in this workflow. The default depth is 1, but setting it to zero fetches the entire history. This is the conventional way to ensure there is sufficient git history for operations that reference older commits.

PR #18483, opened 2023-04-06
Fix switch-ethereum-chain handler by passing configuration id to setA…

danjm:
Fixes https://github.com/MetaMask/metamask-extension/issues/18453

## Explanation

In a previous commit, we updated `setActiveNetwork` in the network controller to receive a network configuration id instead of a whole network object. However, the parameter passed to `setActiveNetwork` in the `switch-ethereum-chain.js` handler was still passed the object, not the id. See https://github.com/MetaMask/metamask-extension/commit/23ca4460cf#diff-93f17dab5026baf840aec766ac5b6ebcd35d5bb0abb45e2c3f41791423f2a488R117

This PR corrects that, and adds an e2e test for the `wallet_switchEthereumChain` functionality, which would have caught this bug.

It is worth noting that the bug this PR fixes would not occur if you switched to the network immediately after adding it, because the correct parameter is passed to the `setActiveNetwork` call in `add-ethereum-chain.js` which happens after prompting the user to switch networks (https://github.com/MetaMask/metamask-extension/commit/23ca4460cf#diff-b3485a1ea9943fd380364f73a53295289f2ccd3817423d4f245b5102f8620b3cR170)

## Manual Testing Steps

1. Run `$(yarn bin ganache) --port 8546 --chain.chainId 1338`
2. Go to https://metamask.github.io/test-dapp/ and connect to the dapp
3. Click "Add Localhost 8546" and "Approve", BUT click "Cancel" once you get to the "Allow this site to switch the network?" screen (so that if you were to open MetaMask, the Localhost 8546 network would not be selected)
4. Click "Switch to Localhost 8546" and then "Approve"
5. If you open MetaMask, the Localhost 8546 network should now be selected

## Pre-merge author checklist

- [ ] I've clearly explained:
  - [ ] What problem this PR is solving
  - [ ] How this problem was solved
  - [ ] How reviewers can test my changes
- [ ] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [ ] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
- [ ] PR is linked to the appropriate GitHub issue
- [ ] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.

In this case, a QA Engineer approval will be be required.

PR #18513, opened 2023-04-08
Ensure that all networkConfiguration object in networkController state have an id

### LINKED ISSUE #18453: [Bug]: RPC Error when switching network on 10.28.1
### Describe the bug

I just updated to 10.28.1 and I get an error when trying to switch networks from various dapps, including app.aave.com. I am able to add the network, but if I have previously added the network and just try to switch to the network from the dapp the console shows the following message: `RPC Error: networkConfigurationId [object Object] does not match a configured networkConfiguration `

### Steps to reproduce

1. Install 10.28.1
2. Go to app.aave.com testnet
3. Go to a modal for a network that you aren't connected to (e.g. Fantom) and click on switch network. Add network by clicking "switch network".
4. Switch back to another network (e.g. Ethereum mainnet) and repeat step 3. This time, aave will show a cannot switch network error, and you will see aforementioned error in the developer console.

```shell
RPC Error: networkConfigurationId [object Object] does not match a configured networkConfiguration
```

### Version

10.28.1

### Build type

None

### Browser

Chrome

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

### LINKED ISSUE #18509: [Bug]: wallet_switchEthereumChain no longer working after update to 10.28.2 Chrome
### Describe the bug

After the release of `10.28.2` `wallet_switchEthereumChain` is no longer working. Error in console is `{code: -32603, message: 'networkConfigurationId undefined does not match a configured networkConfiguration', data: {…}}`. `10.28.1` has an issue where user has to approve the function twice. Downgrading to `10.26.1` allowed the `wallet_switchEthereumChain` to function. (Did not try versions between 10.26.1 and 10.28.1) It appears that the issue only happens when switching to a chain other than Ethereum but can not confirm 100%

### Steps to reproduce

1. Ensure that extension version `10.28.2` is installed in Chrome Desktop. 
2. Visit a site which utilizes `wallet_switchEthereumChain`, trade.coinbook.app is an example.
3. Utilize Network dropdown on DAPP to select a chain to switch to. Metamask will open with correct chain swicth prompt. Error is logged after clicking Approve. 
4. Depending on DAPP configuration there may be a catch that calls `wallet_addEthereumChain` on error. 

```shell
{code: -32603, message: 'networkConfigurationId undefined does not match a configured networkConfiguration', data: {…}}
```

### Version

10.28.2

### Build type

None

### Browser

Chrome

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

danjm:
Fixes https://github.com/MetaMask/metamask-extension/issues/18509
Fixes https://github.com/MetaMask/metamask-extension/issues/18453

## Explanation

We partially fixed the symptom of 18509 and 18453 with https://github.com/MetaMask/metamask-extension/pull/18483, but that fix was incomplete.

This bug will be experienced by anyone who added a network while on v10.27.0 or earlier, and are now on v10.28.1 or v10.28.2 and trying to confirm a dapp prompted switch to that network. This bug will not affect users who first added the network, via a dapp wallet_addEthereumChain request, while on v10.28.1 . This bug also won't affect users who added the network via the MetaMask UI in v10.28.1

The problem is that migration 82 did not add an `id` property to each network configuration object, but we assume that `id` property exists in the `switch-ethereum-chain.js` handler. If a user adds a network on version v10.27.0 or earlier, and then updates to v10.28.x, migration 82 will create an object within `NetworkController.networkConfigurations` keyed by a random id, but that `id` will not be added to that object itself. Later, when we attempt to switch to the network using that objects `id` property, it fails. It does not fail if the network is added while on v10.28.x, because the `upsertNetworkConfiguration` method, which is called when adding a network on v10.28.x, adds an `id` property to each network.

This PR corrects the problem with a migration that ensures that the each key within `networkConfigurations` is used to set the `id` property of the respective object.

### Before

This video shows what would happen when upgrading from v10.27.0 -> v10.28.2 and then trying to switch to a previously added network:

https://user-images.githubusercontent.com/7499938/230714011-86eef63c-b54d-4b79-b51c-1903e1016a27.mp4

### After

This shows switch network working after an upgrade from v10.27.0 to the branch of this PR:

https://user-images.githubusercontent.com/7499938/230714738-16fa8e2e-be3e-4c71-94b3-01220bb618b8.mp4

## Manual Testing Steps

Do as is done in the above "After" video:

1. Build v10.27.0, from d8591d48c, install and onboard
2. Go to https://syncswap.xyz/, connect and add the suggested network
3. In MetaMask, select Ethereum Mainnet, so that the zkSync network is not selected
4. git checkout this branch, build and then reload the metamask installation that was previously built from v10.27.0
5. login to MetaMask and then go to https://syncswap.xyz/
6. click "Connect Wallet" and confirm the switch network popup
7. open MetaMask, you should now have the zksync network selected

## Pre-merge author checklist

- [ ] I've clearly explained:

### DIFF
--- a/app/scripts/migrations/084.test.js
+++ b/app/scripts/migrations/084.test.js
@@ -141,4 +141,114 @@ describe('migration #84', () => {
     };
     expect(newStorage).toStrictEqual(expectedNewStorage);
   });
+
+  it('should not modify state if state.NetworkController is undefined', async () => {
+    const oldStorage = {
+      meta: {
+        version,
+      },
+      data: {
+        testProperty: 'testValue',
+      },
+    };
+
+    const newStorage = await migrate(oldStorage);
+
+    const expectedNewStorage = {
+      meta: {
+        version,
+      },
+      data: {
+        testProperty: 'testValue',
+      },
+    };
+    expect(newStorage).toStrictEqual(expectedNewStorage);
+  });
+
+  it('should not modify state if state.NetworkController is not an object', async () => {
+    const oldStorage = {
+      meta: {
+        version,
+      },
+      data: {
+        NetworkController: false,
+        testProperty: 'testValue',
+      },
+    };
+
+    const newStorage = await migrate(oldStorage);
+
+    const expectedNewStorage = {
+      meta: {
+        version,
+      },
+      data: {
+        NetworkController: false,
+        testProperty: 'testValue',
+      },
+    };
+    expect(newStorage).toStrictEqual(expectedNewStorage);
+  });
+
+  it('should not modify state if state.NetworkController.networkConfigurations is undefined', async () => {
+    const oldStorage = {
+      meta: {
+        version,
+      },
+      data: {
+        NetworkController: {
+          testNetworkControllerProperty: 'testNetworkControllerValue',
+          networkConfigurations: undefined,
+        },
+        testProperty: 'testValue',
+      },
+    };
+
+    const newStorage = await migrate(oldStorage);
+
+    const expectedNewStorage = {
+      meta: {
+        version,
+      },
+      data: {
+        NetworkController: {
+          testNetworkControllerProperty: 'testNetworkControllerValue',
+          networkConfigurations: undefined,
+        },
+        testProperty: 'testValue',
+      },
+    };
+    expect(newStorage).toStrictEqual(expectedNewStorage);
+  });
+
+  it('should not modify state if state.NetworkController.networkConfigurations is an empty object', async () => {
+    const oldStorage = {
+      meta: {
+        version,
+      },
+      data: {
+        NetworkController: {
+          testNetworkControllerProperty: 'testNetworkControllerValue',
+          networkConfigurations: {},
+        },
+        testProperty: 'testValue',
+      },
+    };
+
+    const newStorage = await migrate(oldStorage);
+
+    const expectedNewStorage = {
+      meta: {
+        version,
+      },
+      data: {
+        NetworkController: {
+          testNetworkControllerProperty: 'testNetworkControllerValue',
+          networkConfigurations: {},
+        },
+        testProperty: 'testValue',
+      },
+    };
+    expect(newStorage).toStrictEqual(expectedNewStorage);
+  });
 });
--- a/app/scripts/migrations/084.test.js
+++ b/app/scripts/migrations/084.test.js
@@ -1,0 +1,144 @@
+import { v4 } from 'uuid';
+import { migrate, version } from './084';
+
+jest.mock('uuid', () => {
+  const actual = jest.requireActual('uuid');
+
+  return {
+    ...actual,
+    v4: jest.fn(),
+  };
+});
+
+describe('migration #84', () => {
+  beforeEach(() => {
+    v4.mockImplementationOnce(() => 'network-configuration-id-1')
+      .mockImplementationOnce(() => 'network-configuration-id-2')
+      .mockImplementationOnce(() => 'network-configuration-id-3')
+      .mockImplementationOnce(() => 'network-configuration-id-4');
+  });
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+  it('should update the version metadata', async () => {
+    const oldStorage = {
+      meta: {
+        version: 83,
+      },
+      data: {},
+    };
+
+    const newStorage = await migrate(oldStorage);
+    expect(newStorage.meta).toStrictEqual({
+      version,
+    });
+  });
+
+  it('should use the key of the networkConfigurations object to set the id of each network configuration', async () => {
+    const oldStorage = {
+      meta: {
+        version,
+      },
+      data: {
+        NetworkController: {
+          networkConfigurations: {
+            'network-configuration-id-1': {
+              chainId: '0x539',
+              nickname: 'Localhost 8545',
+              rpcPrefs: {},
+              rpcUrl: 'http://localhost:8545',
+              ticker: 'ETH',
+            },
+            'network-configuration-id-2': {
+              chainId: '0xa4b1',
+              nickname: 'Arbitrum One',
+              rpcPrefs: {
+                blockExplorerUrl: 'https://explorer.arbitrum.io',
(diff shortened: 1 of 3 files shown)


danjm:
This will require https://github.com/MetaMask/metamask-extension/pull/18512#pullrequestreview-1376744678 to get the build passing.

app/scripts/migrations/084.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  const newNetworkConfigurations: Record<string, object> = {};
+
+  for (const networkConfigurationId in networkConfigurations) {
+    if (
+      Object.prototype.hasOwnProperty.call(
+        networkConfigurations,
+        networkConfigurationId,
+      )
+    ) {
```
adonesky1, line 44: I'm probably missing something here but why is this check necessary? Won't the `networkConfigurationId` necessarily be a property of `networkConfigurations` as its defined in this the for-in loop?

adonesky1:
It probably does make sense to fix the state here but - just a thought - another lighter weight way to fix this (I think) would be to modify the `findNetworkConfigurationBy` method on `MetamaskController` like this:

```
   findNetworkConfigurationBy(rpcInfo) {
     const { networkConfigurations } = this.networkController.store.getState();
-    const networkConfiguration = Object.values(networkConfigurations).find(
-      (configuration) => {
-        return Object.keys(rpcInfo).some((key) => {
-          return configuration[key] === rpcInfo[key];
-        });
-      },
-    );
+    const [id, networkConfiguration] = Object.entries(
+      networkConfigurations,
+    ).find((configuration) => {
+      return Object.keys(rpcInfo).some((key) => {
+        return configuration[key] === rpcInfo[key];
+      });
+    });
 
-    return networkConfiguration || null;
+    return { ...networkConfiguration, id } || null;
   }
   ```

**Review by legobeat — DISMISSED** [review]
LGTM

**Review by DDDDDanica — DISMISSED** [review]

File: app/scripts/migrations/084.ts (hunk 1 above)
Comment (Gudahtt) [reply app/scripts/migrations/084.ts:44]: Nit: This check could be skipped if the look was `for (const configurationId of Object.keys(networkConfigurations)) { `instead.

File: app/scripts/migrations/084.ts [hunk 2]
```diff
@@ -0,0 +1,59 @@
...
+  versionedData.meta.version = version;
+  versionedData.data = transformState(versionedData.data);
+  return versionedData;
+}
+
+function transformState(state: Record<string, unknown>) {
+  if (!isObject(state.NetworkController)) {
+    return state;
+  }
+  const { NetworkController }: any = state;
```
Gudahtt, line 30: Nit: What purpose is the `: any` serving here? It seems that it's making the type less specific (it's changing it from `RuntimeObject` to `any`)

app/scripts/migrations/084.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  const {
+    networkConfigurations,
+  }: { networkConfigurations: Record<string, object> } = NetworkController;
+
+  const newNetworkConfigurations: Record<string, object> = {};
```
Gudahtt, line 36: Nit: We can be more specific here than `object`:

```suggestion
  const newNetworkConfigurations: Record<string, Record<string, unknown>> = {};
```

app/scripts/migrations/084.ts
Gudahtt: Oh I see, this is disabling type checking. The object spread operator below requires us to check that each of the old network configuration entries is an object, but we skipped that check because type checking was disabled here when this was set to `any`.

app/scripts/migrations/084.test.js
```diff
@@ -0,0 +1,144 @@
...
+      data: {},
+    };
+
+    const newStorage = await migrate(oldStorage);
+    expect(newStorage.meta).toStrictEqual({
+      version,
+    });
+  });
+
+  it('should use the key of the networkConfigurations object to set the id of each network configuration', async () => {
```
Gudahtt, line 38: Nit: We should have tests for the error scenarios as well (e.g. each level of state being missing, or being a non-object type)

app/scripts/migrations/084.ts
danjm: `any` removed in 8116d9fe46

adonesky1 (dismissed):
LGTM!

app/scripts/migrations/084.ts
Gudahtt: The `any` was removed, but the required type checks have not been added yet. The type errors are the reason CI is failing

Gudahtt: To start with, I'd recommend not adding explicit types to variable declarations if the type can be inferred instead. The error here is a result of the inferred type being incompatible with the type you've declared.

It is necessary to declare a type in cases where there is no way for it to be inferred correctly. For example, the `newNetworkConfigurations` variable below, which is assigned to an empty object initially. The inferred type would be `{}`, which isn't correct, so an explicit type was required there. But here the type can be inferred directly from the assigned value.

Putting that issue aside, if you drop both unnecessary type declarations here, then `networkConfigurations` will end up as type `RuntimeObject` (e.g. `Record<string, unknown>`). This ends up failing on this line:

```
    newNetworkConfigurations[networkConfigurationId] = {
      ...networkConfigurations[networkConfigurationId],
      id: networkConfigurationId,
    };
```
with the error "Spread types may only be created from object types".

That is, you're assuming that the value (typed as `unknown`) is an object.

This can be fixed by adding another type check inside the loop, ensuring that any pre-existing configuration used is also an object. That's what I meant here:

>The object spread operator below requires us to check that each of the old network configuration entries is an object

adonesky1 (dismissed):

app/scripts/migrations/084.ts
danjm: okay, thank-you

I have attempted to make improvements in line with your above comment here affa96040e

Gudahtt (approved):
LGTM!

adonesky1 (approved):

PR #18520, opened 2023-04-10
Adopt security provider request from core

vinistevam:
## Explanation

This PR aims to adopt the security provider request from message-manager pkg (core) into `SignController`.

Additionally, migrating to ts and adding unit tests to security provider helper.

* Fixes https://github.com/MetaMask/MetaMask-planning/issues/414

## Manual Testing Steps

- Go to [dapp test page](https://metamask.github.io/test-dapp/)
- Try to connect
- Sign a Personal message (or any other)
- Should not display any warning cause `metamask.github.io` is not flagged as malicious.

## Pre-merge author checklist

- [x] I've clearly explained:


app/scripts/controllers/sign.ts
```diff
@@ -60,11 +62,15 @@ const getDefaultState = () => ({
 
 export type CoreMessage = AbstractMessage & {
   messageParams: AbstractMessageParams;
+  securityProviderResponse?: Record<string, Json> | undefined;
 };
 
-export type StateMessage = Required<AbstractMessage> & {
+export type StateMessage = Omit<
+  Required<AbstractMessage>,
+  'securityProviderResponse'
```
matthewwalsh0, line 70: I'm confused why we are removing the `securityProviderResponse` property then redefining it. If it's part of `AbstractMessage`, we shouldn't have to redefine here right?

app/scripts/controllers/sign.ts
```diff
@@ -539,7 +548,7 @@ export default class SignController extends BaseControllerV2<
       AbstractMessageParamsMetamask
     >,
     updateState: (
-      state: SignControllerState,
+      state: Record<string, unknown>,
```
matthewwalsh0, line 551: Why can we not use the `SignControllerState`, that should be exactly what we're updating with these callbacks?

app/scripts/lib/security-provider-helpers.ts
```diff
@@ -1,35 +1,46 @@
+import { Json } from '@metamask/utils';
+import { MessageParams } from '@metamask/message-manager';
 import getFetchWithTimeout from '../../../shared/modules/fetch-with-timeout';
 import { MESSAGE_TYPE } from '../../../shared/constants/app';
 
 const fetchWithTimeout = getFetchWithTimeout();
 
+export interface RequestData {
+  origin: string;
+  msgParams?: MessageParams;
```
matthewwalsh0, line 10: If we always need at least one of the message params properties, could we instead use a type with a union to ensure one is always provided?

app/scripts/lib/security-provider-helpers.ts
```diff
@@ -1,35 +1,46 @@
...
-  currentLocale,
-) {
+  requestData: RequestData,
+  methodName: string,
+  chainId: string,
+  currentLocale: string,
+): Promise<Record<string, Json>> {
   let dataToValidate;
+  // Core message managers use messageParams but frontend uses msgParams with lots of references
+  const params = requestData.msgParams || requestData.messageParams;
```
matthewwalsh0, line 23: This could have been the concern of the `AbstractMessageManager` as it is currently the edge case with `messageParams`, but I guess it makes sense to support this as low as possible to support other future code paths.

app/scripts/lib/security-provider-helpers.ts
```diff
@@ -1,35 +1,46 @@
...
+  txParams?: Record<string, unknown>;
+}
+
 export async function securityProviderCheck(
-  requestData,
-  methodName,
-  chainId,
-  currentLocale,
-) {
+  requestData: RequestData,
```
matthewwalsh0, line 16: Following the above comment about types, we could define a `TransactionRequestData` and `MessageRequestData` and use a union here to associate them with the possible method names?

app/scripts/lib/security-provider-helpers.ts
vinistevam: I've included now `MessageRequestData` and `MessageTransactionData`  types however I had to set the properties that should not be present in the type as never so it does not need to check the type again further on.

app/scripts/controllers/sign.ts
```diff
@@ -60,11 +62,15 @@ const getDefaultState = () => ({
...
 };
 
-export type StateMessage = Required<AbstractMessage> & {
+export type StateMessage = Omit<
+  Required<AbstractMessage>,
+  'securityProviderResponse'
+> & {
   msgParams: Required<AbstractMessageParams>;
-  securityProviderResponse: any;
+  securityProviderResponse?: Record<string, Json> | undefined;
```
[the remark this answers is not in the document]
vinistevam: Make sense, thanks! I ended up removing the `securityProviderResponse`.

app/scripts/controllers/sign.ts
vinistevam: I've added back the `SignControllerState`

app/scripts/controllers/sign.ts
```diff
@@ -63,9 +64,10 @@ export type CoreMessage = AbstractMessage & {
   messageParams: AbstractMessageParams;
 };
 
-export type StateMessage = Required<AbstractMessage> & {
+export type StateMessage = Required<
+  Omit<AbstractMessage, 'securityProviderResponse'>
```
matthewwalsh0, line 68: I'm still not sure what ew gain by omitting `securityProviderResponse`?
Is it just linting errors in the migrate logic, in which case we could also just cast to avoid?

matthewwalsh0 (approved):

bergarces (approved):

PR #18524, opened 2023-04-10
Update gas limit on token allowance change

### LINKED ISSUE #18436: [Bug]: Estimated gas doesn't update if gas was suggested by dapp and user edits the transaction data
### Describe the bug

If a dapp sends a specific gas limit, we will use that (on any confirmation screen), but then if the user modifies the tx data , the estimate of required gas could change. This might result in failed transactions due to out of gas specifically with respect to approve transactions (because we don't let users edit tx data for confirmations proposed by dapps for other tx types).

We should recalculate gas limit estimates when approval amount is edited (whenever the users input a custom spending cap or go with `Max`). We likely only need to recalculate it once user lands in the second screen (where we display the gas info).

Relevant thread [here](https://consensys.slack.com/archives/GTQAGKY5V/p1680541523387319?thread_ts=1680214445.386919&cid=GTQAGKY5V) 

Update: we should also make sure we're updating our own estimations in this case as well (even if the dapp has not proposed custom gas parameters) - this might be what is happening here https://github.com/MetaMask/metamask-extension/issues/18341 ?

### Steps to reproduce

1.

_No response_

### Version

10.27

### Build type

None

### Browser

Brave

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

jpuri:
Fixes: #18436

For approve token allowance transactions - update gas limit when approval amount is changed.


legobeat (dismissed):

ui/components/app/custom-spending-cap/__snapshots__/custom-spending-cap.test.js.snap
```diff
@@ -0,0 +1,107 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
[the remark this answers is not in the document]
jpuri: This link is auto generated.

ui/components/app/custom-spending-cap/custom-spending-cap.js
```diff
@@ -127,6 +133,20 @@ export default function CustomSpendingCap({
     }
 
     dispatch(setCustomTokenAmount(String(valueInput)));
+
+    const newData = getCustomTxParamsData(txParams.data, {
```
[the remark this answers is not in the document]
jpuri: I made the prop required.

app/scripts/lib/util.ts
```diff
@@ -142,6 +142,19 @@ function checkAlarmExists(alarmList: { name: string }[], alarmName: string) {
   return alarmList.some((alarm) => alarm.name === alarmName);
 }
 
+/**
+ * Check if string passed is valid hexadecimal
+ *
+ * @param str
+ * @returns boolean
+ */
+function isHexadecimal(str: string): boolean {
```
segun, line 151: ethereumjs-util has an `isHexString()` method...we are already using it in `shared/modules/hexstring-utils.ts`

ui/components/app/custom-spending-cap/custom-spending-cap.js
```diff
@@ -22,25 +22,32 @@ import {
 } from '../../../helpers/constants/design-system';
 import { getCustomTokenAmount } from '../../../selectors';
 import { setCustomTokenAmount } from '../../../ducks/app/app';
+import { addHexPrefix, isHexadecimal } from '../../../../app/scripts/lib/util';
```
segun, line 25: addHexPrefix is available in ethereumjs-utils as well.

app/scripts/lib/util.ts
jpuri: Good point @segun, I did code changes to include ones from `ethereumjs-util`.

segun (dismissed):

ui/components/app/custom-spending-cap/custom-spending-cap.js
```diff
@@ -127,6 +138,24 @@ export default function CustomSpendingCap({
...
+      from,
+      to,
+      value: txValue,
+      data: newData,
+    });
+    if (estimatedGasLimit) {
+      await updateTransaction({
+        gasLimit: hexToDecimal(addHexPrefix(estimatedGasLimit)),
+      });
+    }
```
digiwand, line 157: I think it would be good to wrap this in a try/catch. 

If updateTransaction throws an error and it would break here and `setInputChangeInProgress(false)` below would not run 

Screen Shot 2023-04-13 at 6 54 16 PM
Screen Shot 2023-04-13 at 6 54 40 PM

ui/components/app/custom-spending-cap/custom-spending-cap.js
```diff
@@ -127,6 +138,24 @@ export default function CustomSpendingCap({
     }
 
     dispatch(setCustomTokenAmount(String(valueInput)));
+
+    const newData = getCustomTxParamsData(txParams.data, {
```
digiwand, line 142: Description from the related issue ticket:
> We likely only need to recalculate it once user lands in the second screen (where we display the gas info).

was this considered? wondering why we recalculate on this page rather than only calculating on the second screen

ui/components/app/custom-spending-cap/custom-spending-cap.js
jpuri: yep good point, I was also. thinking of adding try/catch here. Let me do that.

digiwand (dismissed):

ui/components/app/custom-spending-cap/custom-spending-cap.js
```diff
@@ -97,12 +106,14 @@ export default function CustomSpendingCap({
     getInputTextLogic(value).description,
   );
 
-  const handleChange = (valueInput) => {
+  const handleChange = async (valueInput) => {
```
danjm, line 109: In `token-allowance.js` we have:
```
          <CustomSpendingCap
            txParams={txData?.txParams}
```
so `txParams` can be undefined

should we exit this `handleChange` function early if that is the case?

danjm (approved):

digiwand (approved):

PR #18528, opened 2023-04-11
UX Multichain: updated ethereum logo icon

NidhiKJha:
As per the new UI, we are updating the Ethereum logo 

### Before

Screenshot 2023-04-11 at 5 24 45 PM

### After
Screenshot 2023-04-11 at 5 23 26 PM


georgewrmarshall (dismissed):
I think we may want to add a PNG and update the image paths. The svg added here is just an image wrapped in an svg I also think we should add a square image. I'll attach an appropriate image shortly

georgewrmarshall (commented):
Draft PR here with some suggested changes https://github.com/MetaMask/metamask-extension/pull/18631 Feel free to merge into this PR if you think they make sense.

darkwing (approved):

DDDDDanica (approved):

PR #18565, opened 2023-04-13
[FLASK] Remove unused snaps code

FrederikBolding:
## Explanation

- Removes some unused code after `snap_confirm` was removed in an earlier snaps release.
- Removes some unused code for "app keys", which we also removed in an earlier release.


ui/pages/confirmation/templates/flask/snap-confirmation/snap-confirmation.js
GuillaumeRx, line 1: Why are we keeping this ?

GuillaumeRx (dismissed):

GuillaumeRx (approved):

Mrtenz (approved):

PR #18593, opened 2023-04-14
Remove the network controller `previousProviderStore`

Gudahtt:
## Explanation

The `previousProvider` network controller state has been replaced with a private internal property. This was only used internally, it did not need to be part of state.

This relates to https://github.com/MetaMask/metamask-extension/issues/18303

## Manual Testing Steps

No functional changes to test

## Pre-merge author checklist

- [x] I've clearly explained:
  - [x] What problem this PR is solving
  - [x] How this problem was solved
  - [x] How reviewers can test my changes
- [x] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [ ] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
- [ ] PR is linked to the appropriate GitHub issue
- [ ] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.

In this case, a QA Engineer approval will be be required.

PR #18595, opened 2023-04-14
Consolidate network stores

### LINKED ISSUE #18303: NetworkController API normalization: Consolidate network controller stores
We currently use separate observable stores for different parts of the network controller state. As a step towards normalizing the API of this controller with the core network controller, we should consolidate this state into a single store.

### **Definition of Done**
- [ ] Whenever applies, any change unit tested, reviewed(approved) and documented(JSDOC at least)
- [ ]  We have changelog entries for any related changes
- [ ]  Any changed APIs have comprehensive inline documentation
- [ ]  Any changed public APIs are well covered by unit tests

**Other items**
- [ ]  If a planning/research ticket, then the plan has been reviewed and approved by at least 1 team members

Gudahtt:
## Explanation

The network controller used to have multiple different state stores, which were composed together to form the main controller state store. They have been consolidated into a single store. This required few changes because most state access was already being done through the composed store.

Fixes https://github.com/MetaMask/metamask-extension/issues/18303

## Manual Testing Steps

No functional changes to test.

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/app/scripts/controllers/network/network-controller.ts
+++ b/app/scripts/controllers/network/network-controller.ts
@@ -379,6 +379,11 @@ function buildDefaultNetworkConfigurationsState(): NetworkConfigurations {
   return {};
 }
 
+/**
+ * Builds the default state for the network controller.
+ *
+ * @returns The default network controller state.
+ */
 function buildDefaultState() {
   return {
     provider: buildDefaultProviderConfigState(),
--- a/app/scripts/controllers/network/network-controller.ts
+++ b/app/scripts/controllers/network/network-controller.ts
@@ -1,6 +1,6 @@
 import { strict as assert } from 'assert';
 import EventEmitter from 'events';
-import { ComposedStore, ObservableStore } from '@metamask/obs-store';
+import { ObservableStore } from '@metamask/obs-store';
 import log from 'loglevel';
 import {
   createSwappableProxy,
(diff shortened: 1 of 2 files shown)


Gudahtt:
This depends upon #18593

app/scripts/metamask-controller.js
```diff
@@ -976,8 +976,16 @@ export default class MetamaskController extends EventEmitter {
       getNetworkId: () => this.networkController.store.getState().networkId,
       getNetworkStatus: () =>
         this.networkController.store.getState().networkStatus,
-      onNetworkStateChange: (listener) =>
-        this.networkController.networkIdStore.subscribe(listener),
+      onNetworkStateChange: (listener) => {
```
mcmire, line 979: Ugh, another callback that is improperly named, but oh well...

app/scripts/controllers/network/network-controller.ts
```diff
@@ -379,6 +379,16 @@ function buildDefaultNetworkConfigurationsState(): NetworkConfigurations {
   return {};
 }
 
+function buildDefaultState() {
```
mcmire, line 382: Nit: Do we want JSDoc for this function?

mcmire (dismissed):
Makes sense to me. I'm assuming we don't need a state migration for this change? Considering that all of the individual stores were public, I don't know if anything was listening to those stores instead of listening to the composite store.

app/scripts/controllers/network/network-controller.ts
Gudahtt: Oops, good catch! I meant to add this, I'll do that now

Gudahtt:
>I'm assuming we don't need a state migration for this change?

Nope! Not as far as I know anyway. The top-level state properties are the same as the keys used for the composed store, so the state shape should be identical.

>Considering that all of the individual stores were public, I don't know if anything was listening to those stores instead of listening to the composite store.

Conveniently, it seems that most state access was through the composed store. I did look for cases of this, but found only two examples (in `metamask-controller.js`, both of which have been addressed). Definitely worth checking to ensure I didn't miss any though.

app/scripts/controllers/network/network-controller.ts
Gudahtt: Added here: https://github.com/MetaMask/metamask-extension/pull/18595/commits/9e42ffb44239fbaabc0e2f5f1779f0c3757c4973

brad-decker (dismissed):
LGTM

mcmire (approved):
Looks good!

brad-decker (approved):

Issue #18587, opened 2023-04-13
NetworkController API normalization: Eliminate out-of-band execution


Gudahtt:
In various places within the network controller, we call async functions without await-ing them. This complicates our unit tests and makes them slower and less consistent. The most notable example of this is lookupNetwork, which is called without await in a few places despite it being async.

This is part of [a larger effort to normalize the API of both network controllers](https://github.com/MetaMask/controllers/issues/971), to make them easier to merge.

The related work in core is here: https://github.com/MetaMask/core/issues/1176

### **Definition of Done**
- [ ] Whenever applies, any change unit tested, reviewed(approved) and documented(JSDOC at least)
- [ ]  We have changelog entries for any related changes
- [ ]  Any changed APIs have comprehensive inline documentation
- [ ]  Any changed public APIs are well covered by unit tests

**Other items**
- [ ]  If a planning/research ticket, then the plan has been reviewed and approved by at least 1 team members

PR #18600, opened 2023-04-14
Make `_setProviderConfig` async

Gudahtt:
## Explanation

The network controller internal method `_setProviderConfig` has been made async, the async `_switchNetwork` operation is now `await`-ed.

Since the `_switchNetwork` call was the last operation, this has zero functional impact.

Relates to https://github.com/MetaMask/metamask-extension/issues/18587

## Manual Testing Steps

No functional changes

## Pre-merge author checklist

- [x] I've clearly explained:
  - [x] What problem this PR is solving
  - [x] How this problem was solved
  - [x] How reviewers can test my changes
- [x] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [ ] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
- [ ] PR is linked to the appropriate GitHub issue
- [ ] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.

In this case, a QA Engineer approval will be be required.

PR #18604, opened 2023-04-14
Make `setProviderType` async

Gudahtt:
## Explanation

The network controller method `setProviderType` is now async, and the async operation `_setProviderConfig` called at the end of the method is now awaited.

Because the only async operation was the last step, this should have no impact upon the flow of execution in cases where the opertaion is not awaited.

The `switch-ethereum-chain` middleware was updated to await this operation. As a result, an error thrown while the network is switching will now be thrown in this middleware and returned to the dapp as an internal error.

The UI will also be affected, in that the background call to switch networks will not resolve until the network switch has completed. This may impact the timing of state updates slightly in cases where the network switches in the UI (e.g. when using the network switcher)

Relates to https://github.com/MetaMask/metamask-extension/issues/18587

## Manual Testing Steps

Test the `eth_switchEthereumChain` method to ensure it still work correctly. It should work the same as before, just with slight internal changes in timing.

Test any situations where the network can be switched from the UI (e.g. the network switcher, adding a network, etc.)

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/app/scripts/controllers/detect-tokens.test.js
+++ b/app/scripts/controllers/detect-tokens.test.js
@@ -278,7 +278,7 @@ describe('DetectTokensController', function () {
 
   it('should be called on every polling period', async function () {
     const clock = sandbox.useFakeTimers();
-    network.setProviderType(NETWORK_TYPES.MAINNET);
+    await network.setProviderType(NETWORK_TYPES.MAINNET);
     const controller = new DetectTokensController({
       preferences,
       network,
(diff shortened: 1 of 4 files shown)


Gudahtt:
This depends upon #18600

mcmire (approved):
Looks good!

Gudahtt:
Rebased to resolve a conflict

brad-decker (approved):
LGTM

PR #18628, opened 2023-04-17
Use fake provider for NetworkController unit tests

### LINKED ISSUE #18287: Update NetworkController unit tests to use fake provider
Right now the unit tests for NetworkController in this repo are written in a different style than that in the core repo. The biggest difference is that in this repo we are using Nock to mock network requests, whereas in the core repo we are using a fake provider object. This difference makes it difficult to add new functionality to the core repo from this repo, because the unit tests can't just be copied over, they have to be changed in the process. It also makes it difficult to visually compare differences when it comes time to merge the two controllers.

### **Definition of Done**
- [ ] Whenever applies, any change unit tested, reviewed(approved) and documented(JSDOC at least)
- [ ]  We have changelog entries for any related changes
- [ ]  Any changed APIs have comprehensive inline documentation
- [ ]  Any changed public APIs are well covered by unit tests

**Other items**
- [ ]  If a planning/research ticket, then the plan has been reviewed and approved by at least 1 team members

mcmire:
## Explanation

In the unit tests for NetworkController, it's important to prevent network requests from occurring. Currently we do that by using Nock. However, the `core` version of NetworkController uses a fake provider object. This is arguably a better approach for unit tests because it prevents us from having to think about the behavior that a specific middleware may have. For instance, the Infura middleware intercepts `eth_chainId` to return a static result, and the block cache middleware replaces the `latest` block tag with the latest block number, making an extra call to `eth_blockNumber` in doing so. We have to account for these kinds of behaviors when using Nock, but we do not need to do this when using a fake provider.

This should make it easier to compare the difference between the unit tests in this repo vs. in the `core` repo, which should ultimately help us merge the two controllers together.

Closes #18287.

## Manual Testing Steps

No functional changes. All tests should pass.

## Pre-merge author checklist

- [x] I've clearly explained:


app/scripts/controllers/network/fake-provider-engine.ts
```diff
@@ -0,0 +1,224 @@
+import { inspect, isDeepStrictEqual } from 'util';
```
mcmire, line 1: This file was mostly taken from a similar file in https://github.com/MetaMask/core/pull/1116. I say mostly because I ended up making some modifications to type things a little bit better and to remove some cruft that is not needed when using `json-rpc-engine` (for instance, there is no `stop` method on SafeEventEmitterProvider so that doesn't need to be overridden). Also I added a `beforeCompleting` property to the FakeProviderStub class as these tests needed it. I will copy these back to the core repo.

app/scripts/controllers/network/network-controller.test.ts
```diff
@@ -760,12 +352,12 @@ describe('NetworkController', () => {
               const promisifiedSendAsync = promisify(provider.sendAsync).bind(
                 provider,
               );
-              const { result: chainIdResult } = await promisifiedSendAsync({
```
mcmire, line 763: This is a notable change because in using a fake provider, the Infura middleware no longer returns a static result for `eth_chainId`. In other words we have 100% control over what the provider returns now. So using `eth_chainId` to test the provider object is misleading. We don't even need to use a known RPC method — we just want to make sure that `provider.sendAsync` is called.

app/scripts/controllers/network/network-controller.test.ts
```diff
@@ -250,360 +119,33 @@ const BLOCKED_INFURA_RESPONSE = {
...
+  JSON.stringify({ error: 'oops' }),
+);
 
 describe('NetworkController', () => {
-  let clock: SinonFakeTimers;
-
   beforeEach(() => {
     // Disable all requests, even those to localhost
     nock.disableNetConnect();
-    // Faking timers ends up doing two things:
```
mcmire, line 594: We don't need to use fake timers because the block tracker never gets started and there is no retry logic because the middleware never gets used. So that simplifies things a bit.

app/scripts/controllers/network/network-controller.test.ts
```diff
@@ -843,10 +429,19 @@ describe('NetworkController', () => {
                 },
               },
             },
-            async ({ controller, network }) => {
-              network.mockEssentialRpcCalls({
-                latestBlock: POST_1559_BLOCK,
-              });
+            async ({ controller }) => {
```
mcmire, line 432: The resulting test code using a fake provider instead of Nock is slightly more verbose because of the API that I had built over on the `core` side. I could probably make this a little simpler, but it could also be argued that seeing things a bit more spelled out and knowing which RPC methods that NetworkController calls under the hood instead of making some things implicit is easier to read.

app/scripts/controllers/network/network-controller.test.ts
```diff
@@ -1795,51 +1595,47 @@ describe('NetworkController', () => {
                   },
                 },
               },
-              async ({ controller, network }) => {
-                network.mockEssentialRpcCalls({
-                  // Ensure that each call to eth_blockNumber returns a
```
mcmire, line 1800: This is a nice side effect of using a fake provider as well — we don't need to worry about mocking `eth_blockNumber` in these tests because the block tracker never gets started.

app/scripts/controllers/network/network-controller.test.ts
```diff
@@ -1114,37 +731,72 @@ describe('NetworkController', () => {
...
+                      result: 'test response 2',
+                    },
+                  },
+                ]),
+              ];
+              const fakeNetworkClients = [
+                buildFakeClient(fakeProviders[0]),
+                buildFakeClient(fakeProviders[1]),
+              ];
+              mockCreateNetworkClient()
```
mcmire, line 763: Here I'm making use of the `jest-when` package to make sure that specific network clients are used for specific invocations of `createNetworkClient`. This is specific to these tests — I don't think any of the tests on the `core` side involve multiple networks in a single test. I aim to copy these tests over to the `core` side to fix that divergence.

app/scripts/controllers/network/fake-block-tracker.ts
```diff
@@ -0,0 +1,11 @@
+import { PollingBlockTracker } from 'eth-block-tracker';
+
+/**
+ * Implements just enough of the block tracker interface to pass the tests but
```
Gudahtt, line 4: Nit: This suggests that this mock implements a subset of the block tracker API. But instead it appears that this block tracker _is a real, complete block tracker_ with a single method overridden.

app/scripts/controllers/network/fake-provider.ts
```diff
@@ -0,0 +1,224 @@
...
+ * this error.
+ * @property implementation - Allows overriding `sendAsync` entirely. Useful if
+ * you want it to throw an error.
+ * @property delay - The amount of time that will pass after the callback is
+ * called with the response.
+ * @property discardAfterMatching - Usually after the stub matches a request, it
+ * is discarded, but setting this to true prevents that from happening. True by
+ * default.
+ */
+export type FakeProviderStub<Params, Response> = {
```
Gudahtt, line 40: It doesn't seem like these generics are serving any purpose. I see that they're carried through to the `FakeProvider` class `send` and `sendAsync` methods, but those methods are never called from a context that would be aware of the stub types used. Effectively they may as well be `any` here since they're set as `any` in the fake provider engine, making these less type safe than encoding the constraints in this type instead of generics.

app/scripts/controllers/network/fake-provider.ts
```diff
@@ -0,0 +1,224 @@
...
+  sendAsync = <Params, Response>(
+    payload: JsonRpcRequest<Params>,
+    callback: (error: unknown, response?: JsonRpcResponse<Response>) => void,
+  ) => {
+    return this.#handleSend(payload, callback);
+  };
+
+  #handleSend<Params, Response>(
+    payload: JsonRpcRequest<Params>,
+    callback?: (error: unknown, response?: JsonRpcResponse<Response>) => void,
```
Gudahtt, line 122: It seems that a callback is passed in both places where this is called. Should this be mandatory rather than optional?

app/scripts/controllers/network/fake-provider.ts
```diff
@@ -0,0 +1,224 @@
...
+  #handleRequest<Params, Response>(
+    stub: FakeProviderStub<Params, Response>,
+    callback: (error: unknown, response?: JsonRpcResponse<Response>) => void,
+  ) {
+    const promiseForBeforeCompleting =
+      stub.beforeCompleting === undefined
+        ? Promise.resolve()
+        : Promise.resolve(stub.beforeCompleting());
+
+    promiseForBeforeCompleting.then(() => {
```
Gudahtt, line 191: Nit: Could we use async/await here rather than `.then`? Otherwise any stack traces in tests will terminate here, rather than showing further up the stack

app/scripts/controllers/network/fake-provider.ts
```diff
@@ -0,0 +1,224 @@
...
+        stub.implementation();
+      } else if ('response' in stub) {
+        if ('result' in stub.response) {
+          return callback(null, {
+            jsonrpc: '2.0',
+            id: 1,
+            result: stub.response.result,
+          });
+        } else if ('error' in stub.response) {
+          if (typeof stub.response.error === 'string') {
```
Gudahtt, line 202: Is this condition necessary? According to the stub type, this should always be a string

app/scripts/controllers/network/fake-block-tracker.ts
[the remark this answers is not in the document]
mcmire: That's a good idea. How does this look? [777b1d9](https://github.com/MetaMask/metamask-extension/pull/18628/commits/777b1d900683f6b08f5a72ee63c26f0f908b761f)

app/scripts/controllers/network/fake-block-tracker.ts
mcmire: You're right, this description is inaccurate. It's also not super clear how overriding `eventNames` prevents the block tracker from running (assuming it does). I've fixed both of these issues: [6a1e7c3](https://github.com/MetaMask/metamask-extension/pull/18628/commits/6a1e7c3f985e38f237621fe6c024df6aa15b8607)

app/scripts/controllers/network/fake-provider.ts
mcmire: Yeah... you're right. I've removed them here: [9bf8177](https://github.com/MetaMask/metamask-extension/pull/18628/commits/9bf81775e16a2109197ac7c35a67a1c515fc27ae)

app/scripts/controllers/network/fake-provider.ts
mcmire: Good point. This method wouldn't be awaited, so as long as that's not confusing then this is fine. I've updated it in [1e5c662](https://github.com/MetaMask/metamask-extension/pull/18628/commits/1e5c662ed5c4629b8c7423d8900ce53486b64e46).

app/scripts/controllers/network/fake-provider.ts
mcmire: Yes, this is a bit funky. I also realized that the signature of `sendAsync` was incorrect. I've updated it and aligned the other methods here: [424320f](https://github.com/MetaMask/metamask-extension/pull/18628/commits/424320f7b3dd3de60b1954d0b3643d6c91ef92ab)

app/scripts/controllers/network/fake-provider.ts
mcmire: Good point. Removed in [24c635b](https://github.com/MetaMask/metamask-extension/pull/18628/commits/24c635bdb4683c48912e9cc9dc9a9cbb8672c3bc).

mcmire:
@Gudahtt I believe I've addressed all of your feedback above.

app/scripts/controllers/network/fake-provider.ts
Gudahtt: Hmm, how was it incorrect? The old one looks like it matches the safe event emitter provider: https://github.com/MetaMask/eth-json-rpc-provider/blob/3fa7c4f2cdcf9ea0c77b77c8e71c9cf6194ba2b0/src/safe-event-emitter-provider.ts#L38

That is, the callback is always required, not optional.

app/scripts/controllers/network/network-controller.test.ts
```diff
@@ -814,8 +373,17 @@ describe('NetworkController', () => {
...
             },
-            async ({ controller, network }) => {
-              network.mockEssentialRpcCalls();
+            async ({ controller }) => {
+              const fakeProvider = buildFakeProvider([
+                {
+                  request: {
+                    method: 'net_version',
+                  },
+                  response: SUCCESSFUL_NET_VERSION_RESPONSE,
```
Gudahtt, line 382: Nit: This stub seems to be included by default by `buildFakeProvider` - is this needed here?

app/scripts/controllers/network/network-controller.test.ts
```diff
@@ -1327,12 +1050,17 @@ describe('NetworkController', () => {
...
-        network.mockEssentialRpcCalls({
-          eth_getBlockByNumber: {
-            times: 1,
+      await withController(async ({ controller }) => {
+        const fakeProvider = buildFakeProvider([
+          {
+            request: {
+              method: 'eth_getBlockByNumber',
+            },
+            response: SUCCESSFUL_ETH_GET_BLOCK_BY_NUMBER_RESPONSE,
```
Gudahtt, line 1059: Nit: This looks similar to the default stubs as well

Gudahtt (dismissed):
LGTM! Still seems like there is an opportunity to slightly simplify the fake provider, but otherwise this all looks great. I see why you went the fake provider route, it is simpler to read these tests than it was with the fetch mocks.

app/scripts/controllers/network/network-controller.test.ts
mcmire: It is included by default, but I'm choosing to be explicit here because the test is ensuring that `networkStatus` returns "available", and that relies upon `net_version` returning a successful result. Actually, `eth_getBlockByNumber` also needs to return a successful result too. I've increased the explicitness of these tests: [02a2bcc](https://github.com/MetaMask/metamask-extension/pull/18628/commits/02a2bcc4d8ecdab9ed03864b9f59965b7bb73bdd)

Gudahtt (approved):

app/scripts/controllers/network/fake-provider.ts
mcmire: Yeah... okay. I saw that `send` throws if it's not given a callback but `sendAsync` does not. It's inconsistent, and maybe there is a good reason for it, but I guess that doesn't mean that the callback for `sendAsync` is optional. I've made them all required: [08f5d7d](https://github.com/MetaMask/metamask-extension/pull/18628/commits/08f5d7dfc3ebbfc2b3e8a80c4964bca473e5172b)

app/scripts/controllers/network/test/fake-provider.ts
```diff
@@ -0,0 +1,206 @@
...
+
+      this.calledStubs.push({ ...stub });
+    }
+  }
+
+  async #handleRequest(
+    stub: FakeProviderStub,
+    callback: (error: unknown, response?: JsonRpcResponse<any>) => void,
+  ) {
+    if (stub.beforeCompleting) {
```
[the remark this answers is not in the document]
Gudahtt: The difference in these conditions is due to the stub type (above on line 42). The `beforeCompleting` property is guaranteed to exist (but it can be undefined), so TypeScript lets us check it directly. However the `response` and `implementation` properties might not exist on the object at all, so TypeScript objects to referencing the property directly, forcing us to use the `in` operator.

brad-decker (approved):
Looking forward to seeing how this will benefit my integration test. Looks great!

PR #18655, opened 2023-04-19
Rename migration 83 to 84 and rename migration 84 to 83

danjm:
## Explanation

v10.28.3 only included a migration, migration 84. At the time of merging it to master, migration 83 was not yet on master. So the merge to master resolved the merge conflict by renaming and renumbering migration 84 to 83.

Now that we want to sync master with develop, we have to deal with another merge conflict. We need to now rename and renumber migration 83 to 84, and again rename and renumber 84 to 83, on develop. This could be done in the master sync PR, but I am doing it here so that the master sync PR diff follows the usual pattern of only having changes in package.json and the changelog


DDDDDanica (approved):

Gudahtt (approved):
LGTM!

Issue #18574, opened 2023-04-13
[Bug]: new BigNumber not a number: undefined for setApprovalForAll action


DDDDDanica:
Reported by user:
https://twitter.com/pentcle/status/1646067852707856384?s=46&t=E_l8m89uCvRPKu7MFB61cg

Sentry infor:
- https://metamask.sentry.io/issues/3738473311/
- https://metamask.sentry.io/issues/4077910899/

in the notification UI : `confirm-transaction/[transaction id]/set-approval-for-all` 
![Screenshot from 2023-04-12 18-38-46.png](https://images.zenhubusercontent.com/5c78143c7469342c8ba196bc/e2c9934c-9cec-4140-b399-e2bdad61dda6)

in browser-action popup: 
![Screenshot from 2023-04-12 18-41-06.png](https://images.zenhubusercontent.com/5c78143c7469342c8ba196bc/cccde992-7de4-42a8-b54d-ef6b674ee9f4)

DDDDDanica:
Query to monitor: 
https://metamask.sentry.io/issues/?query=is%3Aunresolved+BigNumber+Error&referrer=issue-list&statsPeriod=14d

DDDDDanica:
Follow up on 2nd May 2023:
2 main issues that's tracking this issue only occurs on 10.28.3, and 10.29.0 release 1st May has no such error reported so far,
https://metamask.sentry.io/issues/3738473311/?query=is%3Aunresolved+BigNumber+Error&referrer=issue-stream&statsPeriod=14d&stream_index=4
https://metamask.sentry.io/issues/4077910899/?query=is%3Aunresolved+BigNumber+Error&referrer=issue-stream&statsPeriod=14d&stream_index=5

will come back and monitor after 1 week, 2 weeks and 1 month

PR #18660, opened 2023-04-19
fix(18574): fix new BigNumber() not a number: undefined for setApprovalForAll method

DDDDDanica:
Fixes:
https://github.com/MetaMask/metamask-extension/issues/18574

Error example:
https://metamask.sentry.io/share/issue/1082e57247a14a8b99fdc3347192ab51/

## Explanation
This error could possibly because when providing an ERC20 token with no decimals, we were trying to retrieve `tokenAmount`  but not able to finish `calcTokenAmount`.
However, we were not able to reproduce this error locally. 
What we will do is to push this fix to next release, and monitor sentry with same error 🙏 

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/ui/helpers/utils/token-util.js
+++ b/ui/helpers/utils/token-util.js
@@ -261,17 +261,15 @@ export async function getAssetDetails(
     // if we can't determine any token standard or details return the data we can extract purely from the parsed transaction data
     return { toAddress, tokenId };
   }
-
+  const tokenValue = getTokenValueParam(tokenData);
+  const tokenDecimals = tokenDetails?.decimals;
   const tokenAmount =
     tokenData &&
-    tokenDetails?.decimals &&
-    calcTokenAmount(
-      getTokenValueParam(tokenData),
-      tokenDetails?.decimals,
-    ).toString(10);
+    tokenValue &&
+    tokenDecimals &&
+    calcTokenAmount(tokenValue, tokenDecimals).toString(10);
 
-  const decimals =
-    tokenDetails?.decimals && Number(tokenDetails.decimals?.toString(10));
+  const decimals = tokenDecimals && Number(tokenDecimals?.toString(10));
 
   if (tokenDetails?.standard === TokenStandard.ERC20) {
     tokenId = undefined;
--- a/ui/hooks/useAssetDetails.test.js
+++ b/ui/hooks/useAssetDetails.test.js
@@ -37,12 +37,16 @@ describe('useAssetDetails', () => {
   let getTokenStandardAndDetailsStub;
 
   beforeEach(() => {
-    getTokenStandardAndDetailsStub = jest
-      .spyOn(Actions, 'getTokenStandardAndDetails')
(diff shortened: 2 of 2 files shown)


ui/helpers/utils/token-util.js
```diff
@@ -261,17 +261,15 @@ export async function getAssetDetails(
...
-    tokenDetails?.decimals &&
-    calcTokenAmount(
-      getTokenValueParam(tokenData),
-      tokenDetails?.decimals,
-    ).toString(10);
-
-  const decimals =
-    tokenDetails?.decimals && Number(tokenDetails.decimals?.toString(10));
+    tokenValue &&
+    tokenDecimals &&
```
Gudahtt, line 269: How does this handle tokens with `decimals` set to `0`?

DDDDDanica: As far as i can see, 0 decimals is handled by the usage of tokenDecimals.
- inside `calcTokenAmount`, decimal 0 is accepted 
```
  const multiplier = Math.pow(10, Number(decimals || 0));
```
- inside the decimals variable, it doesn't seem like 0 would impact the final result 
```
  const decimals = tokenDecimals && Number(tokenDecimals?.toString(10));
```
do you have any other concerns i didn't catch?

Gudahtt: Ah I see, this is a string right? I was thinking that if this was a number, 0 would be falsy and `calcTokenAmount` would never be run. But if it's a string, this should work fine.

Gudahtt (approved):
LGTM!

I am curious how users are getting in this situation though. I thought we didn't allow the decimals field to be unset? Maybe we should add metrics to detect this scenario if we're unsure what's causing it (in a separate PR I mean). And if we know of a way to add a token without setting a `decimals` field, we should fix that, and come up with a strategy for handling pre-existing token entries with that problem.

DDDDDanica:
@Gudahtt thanks for reviewing. We were trying to reproduce the bug locally and the only trace we can get is when tokenDetails does not have a decimals passing in and we fall into undefined big number calc. 
@danjm were suggesting that it could be possible that user imported an ERC20 token with ‘setAppovalForAll’ method in an older version of MM where symbol and decimals are manually set. And when they interacting dapp with new version of MM we have, and initialize a setApprovalForAll action, it hits the undefined and throws error out. 
What I would do is 
1) monitor sentry with errors 2 weeks after release daily to see if any same error happens 
2) trace back which old version of MM we had that can manually add symbol and decimals to reproduce maybe.

danjm (approved):

danjm:
> I am curious how users are getting in this situation though. I thought we didn't allow the decimals field to be unset?

The problem is moreso that there are tokens that have `decimals` but don't have a `value` parsable by `getTokenValueParam`

Danica mentioned one way that could happen. Another is that a user adds an NFT, and its contract mostly meets the NFT spec but it also has a `decimals` property.

PR #18665, opened 2023-04-19
[E2E]: Update chromedriver v112

PeterYinusa:
## Explanation

This change does not affect the minimum required version to use the extension, it's purely for running our tests in CI

Update ChromeDriver to [v112](https://www.npmjs.com/package/chromedriver)

Install Chrome stable channel update for desktop [v112.0.5615.121](https://chromereleases.googleblog.com/search/label/Stable%20updates)

## Pre-merge author checklist

- [ ] I've clearly explained:

### DIFF
--- a/.circleci/scripts/chrome-install.sh
+++ b/.circleci/scripts/chrome-install.sh
@@ -5,12 +5,12 @@ set -u
 set -o pipefail
 
 # To get the latest version, see <https://www.ubuntuupdates.org/ppa/google_chrome?dist=stable>
-CHROME_VERSION='111.0.5563.64-1'
+CHROME_VERSION='112.0.5615.121-1'
 CHROME_BINARY="google-chrome-stable_${CHROME_VERSION}_amd64.deb"
 CHROME_BINARY_URL="https://dl.google.com/linux/chrome/deb/pool/main/g/google-chrome-stable/${CHROME_BINARY}"
 
 # To retrieve this checksum, run the `wget` and `shasum` commands below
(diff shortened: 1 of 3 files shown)


PeterYinusa:
@Gudahtt any ideas why [navigation](https://github.com/MetaMask/metamask-extension/blob/update-chromedriver-v112/test/e2e/webdriver/driver.js#L522) would return an empty array in Chrome v112. I'm not so familiar with these benchmark scripts or the Performance API

Gudahtt:
Hmm, it might be because the page load had not completed yet. From [the MDN page](https://developer.mozilla.org/en-US/docs/Web/API/Performance/getEntriesByType):

>Note: This method does not notify you about new performance entries; you will only get entries that are present in the performance timeline at the time you call this method. To receive notifications about entries as they become available, use a [PerformanceObserver](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver).

We call this [after waiting 1 second.](https://github.com/MetaMask/metamask-extension/blob/843717b33324b1089164156b9f1a8c59fbe741fa/test/e2e/benchmark.js#L32). I'd expect a navigation to a local page to be finished within a second :thinking: But maybe not.

It sounds like we could use a combination of `getEntriesByType` and `PerformanceObserver` to collect existing performance entries and wait for new ones, blocking until we get the ones we expect. That would remove the reliance on that pause.

PeterYinusa:
replaced by https://github.com/MetaMask/metamask-extension/pull/19180

PR #18670, opened 2023-04-19
Fix swaps controller: update provider after networkIdStore state update

### LINKED ISSUE #18669: [Bug]: Optimism swaps - Error fetching quotes
### Describe the bug

I cannot fetch quotes on Optimism, Ethereum Mainnet and Arbitrum still work

https://user-images.githubusercontent.com/53189696/233144278-1c8b5e7f-22da-4eeb-a343-17feeb95ffbb.mov

### Steps to reproduce

1. Log in to the extension
2. Switch to optimism
3. click swaps on the overview page
4. enter an eth amount
5. change the to currency to usdc
6. click review swap

```shell
swaps.js:837 Error fetching quotes:  Error: call revert exception [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (method="getL1Fee(bytes)", data="0x", errorArgs=null, errorName=null, errorSignature=null, reason=null, code=CALL_EXCEPTION, version=abi/5.7.0)
  at e.makeError (index.js:238:1)
  at e.throwError (index.js:247:1)
  at e.decodeFunctionResult (interface.js:388:1)
  at t.<anonymous> (index.js:395:1)
  at index.js:48:1
  at Object.next (index.js:29:1)
  at a (index.js:20:1)
```

### Version

10.29

### Build type

None

### Browser

Chrome

### Operating system

MacOS

### Hardware wallet

_No response_

### Additional context

_No response_

danjm:
Fixes https://github.com/MetaMask/metamask-extension/issues/18669

## Explanation

The linked bug is reproducible by the following steps:

1. Switch network to ethereum mainnet
2. reload the extension
3. switch to optimism
4. click swaps on the overview page
5. enter an eth amount
6. change the to currency to usdc
7. click review swap
8. see "Failed to fetch"

the bug is not reproducible if step 1 is "Switch network to optimism" and then step 3 is skipped

The problem is an assumption in the swaps controller, in this code:
```
    onNetworkDidChange(() => {
      const { networkId, networkStatus } = networkController.store.getState();
      if (
        networkStatus === NetworkStatus.Available &&
        networkId !== this._currentNetworkId
```
It is assumed that `networkId` and `networkStatus` will be updated when a `NetworkControllerEventType.NetworkDidChange` event is emitted. However, they are not updated until later. In the network controller, there is this code:
```
  _switchNetwork(opts) {
    this.messenger.publish(NetworkControllerEventTypes.NetworkWillChange);
    this._resetNetworkId();
    this._resetNetworkStatus();
    this._resetNetworkDetails();
    this.messenger.publish(NetworkControllerEventTypes.NetworkDidChange);
    this.lookupNetwork();
  }
```
The necessary state updates are made within the `this.lookupNetwork();` call, which happens after `this.messenger.publish(NetworkControllerEventTypes.NetworkDidChange)`

And so the swaps controller does not update its provider correctly on a network state change.

The fix is for the swaps controller to subscribe to networkIdStore changes, so that we only attempt to update the provider after the necessary state changes are complete.

## Manual Testing Steps

1. Switch network to ethereum mainnet
2. reload the extension
3. switch to optimism
4. click swaps on the overview page
5. enter an eth amount
6. change the to currency to usdc
7. click review swap
8. swap quotes should be successfully retrieved, and you should now be on the swaps quote page

## Pre-merge author checklist

- [ ] I've clearly explained:

### DIFF
--- a/app/scripts/controllers/swaps.js
+++ b/app/scripts/controllers/swaps.js
@@ -113,7 +113,7 @@ export default class SwapsController {
     fetchTradesInfo = defaultFetchTradesInfo,
     getCurrentChainId,
     getEIP1559GasFeeEstimates,
-    onNetworkDidChange,
+    onNetworkStateChange,
   }) {
     this.store = new ObservableStore({
       swapsState: { ...initialState.swapsState },
@@ -137,7 +137,7 @@ export default class SwapsController {
 
     this.ethersProvider = new Web3Provider(provider);
     this._currentNetworkId = networkController.store.getState().networkId;
-    onNetworkDidChange(() => {
+    onNetworkStateChange(() => {
       const { networkId, networkStatus } = networkController.store.getState();
       if (
         networkStatus === NetworkStatus.Available &&
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -1195,10 +1195,8 @@ export default class MetamaskController extends EventEmitter {
         this.txController.txGasUtil,
       ),
       networkController: this.networkController,
-      onNetworkDidChange: networkControllerMessenger.subscribe.bind(
-        networkControllerMessenger,
-        NetworkControllerEventType.NetworkDidChange,
-      ),
+      onNetworkStateChange: (listener) =>
+        this.networkController.networkIdStore.subscribe(listener),
       provider: this.provider,
       getProviderConfig: () => this.networkController.store.getState().provider,
       getTokenRatesState: () => this.tokenRatesController.state,
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -1196,7 +1196,7 @@ export default class MetamaskController extends EventEmitter {
       ),
       networkController: this.networkController,
       onNetworkStateChange: (listener) =>
-        this.networkController.networkIdStore.subscribe(listener),
+        this.networkController.store.subscribe(listener),
       provider: this.provider,
       getProviderConfig: () => this.networkController.store.getState().provider,
       getTokenRatesState: () => this.tokenRatesController.state,
--- a/app/scripts/controllers/swaps.test.js
+++ b/app/scripts/controllers/swaps.test.js
@@ -160,7 +160,7 @@ describe('SwapsController', function () {
     return new SwapsController({
       getBufferedGasLimit: MOCK_GET_BUFFERED_GAS_LIMIT,
       networkController: getMockNetworkController(),
-      onNetworkDidChange: sinon.stub(),
+      onNetworkStateChange: sinon.stub(),
       provider: _provider,
       getProviderConfig: MOCK_GET_PROVIDER_CONFIG,
       getTokenRatesState: MOCK_TOKEN_RATES_STORE,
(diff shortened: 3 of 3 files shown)


Gudahtt (dismissed):
LGTM!

danjm:
I'm working on e2e tests as part of a separate PR, so as to not block the release, but I'll try to get those tests complete today anyway

app/scripts/controllers/swaps.js
```diff
@@ -113,7 +113,7 @@ export default class SwapsController {
     fetchTradesInfo = defaultFetchTradesInfo,
     getCurrentChainId,
     getEIP1559GasFeeEstimates,
-    onNetworkDidChange,
```
mcmire, line 116: What do you think about renaming this to `onNetworkIdChange`? We have usually reserved `onNetworkStateChange` for the whole state, but that's not how it would get used here.

mcmire (commented):
Makes sense, apart from one thing.

app/scripts/controllers/swaps.js
danjm: that makes sense

Gudahtt: This does depend on more than just the ID though. It uses the status as well.

danjm: yeah, but it is explicitly listening for the ID change

app/scripts/metamask-controller.js
```diff
@@ -1195,10 +1195,8 @@ export default class MetamaskController extends EventEmitter {
         this.txController.txGasUtil,
       ),
       networkController: this.networkController,
-      onNetworkDidChange: networkControllerMessenger.subscribe.bind(
-        networkControllerMessenger,
-        NetworkControllerEventType.NetworkDidChange,
-      ),
+      onNetworkStateChange: (listener) =>
```
mcmire, line 1198: We are about to consolidate these stores, so instead of listening to `networkIdStore`, it might be more future-proof to listen to `networkController.store` and keep track of the current `networkId` and then call the listener if it changes. We can also change this later, so not a big deal, but it would be one less thing to worry about.

mcmire: I'm on my phone at the moment but I'm realizing we may already do this in SwapsController. So maybe we should not do it there? Whatever is easier to understand.

app/scripts/controllers/swaps.js
danjm: but as per Elliot's other comment below, maybe we should just listen for all state changes

Gudahtt: I'm not sure what you mean by that.

A change to either property can trigger the effect. If it were `onNetworkIdChange`, I'd expect it to only be triggered when the ID changes. That would result in this potentially being stuck on the wrong network if the status was anything other than "available" when the id changed.

app/scripts/metamask-controller.js
Gudahtt: Actually... this was already merged, so this probably doesn't work as-written. This may be a strict requirement. Good catch

Gudahtt (dismissed):

mcmire (approved):
Looks good now!

Gudahtt (approved):

PR #18675, opened 2023-04-20
Update NetworkController `setProviderType` to clear `rpcUrl`/`nickname`

### LINKED ISSUE #18673: NetworkController API normalization: `setProviderType` should set `rpcUrl` and `nickname` to undefined, not empty string
In the core version of NetworkController, `setProviderType` sets `rpcTarget` (its name for `rpcUrl`) and `nickname` to `undefined`. While it would be better to completely remove these properties, it would be better to follow suit so that we can make the tests between the two versions of the controllers more alike.

mcmire:
## Explanation

In the core version of NetworkController, `setProviderType` sets `rpcTarget` (its name for `rpcUrl`) and `nickname` to `undefined`. While it would be better to completely remove these properties, it would be better to follow suit so that we can make the tests between the two versions of the controllers more alike.

See corresponding code in `core`: https://github.com/MetaMask/core/blob/684470616b55c0ab2556bdc09bdf6970f6ba0be2/packages/network-controller/src/NetworkController.ts#L493

Fixes https://github.com/MetaMask/metamask-extension/issues/18673.

## Manual Testing Steps

There should be no functional changes, but anything that is using `rpcUrl` or `nickname` when the provider type is an Infura provider type as opposed to "rpc", or anything that is treating `""` as an unset RPC URL (e.g. `if (rpcUrl !== '')` or `rpcUrl || somethingElse` as opposed to `if (rpcUrl)` or `rpcUrl ?? somethingElse`), may be affected. I've done my best to confirm that this is not the case in this codebase.

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/app/scripts/controllers/network/network-controller.ts
+++ b/app/scripts/controllers/network/network-controller.ts
@@ -742,10 +742,10 @@ export class NetworkController extends EventEmitter {
     const network = BUILT_IN_INFURA_NETWORKS[type];
     await this.#setProviderConfig({
       type,
-      rpcUrl: '',
+      rpcUrl: undefined,
       chainId: network.chainId,
       ticker: 'ticker' in network ? network.ticker : 'ETH',
-      nickname: '',
+      nickname: undefined,
       rpcPrefs: { blockExplorerUrl: network.blockExplorerUrl },
     });
   }
--- a/app/scripts/controllers/network/network-controller.test.ts
+++ b/app/scripts/controllers/network/network-controller.test.ts
@@ -5006,10 +5006,10 @@ describe('NetworkController', () => {
 
               expect(controller.store.getState().provider).toStrictEqual({
                 type: networkType,
-                rpcUrl: '',
+                rpcUrl: undefined,
                 chainId,
                 ticker,
-                nickname: '',
+                nickname: undefined,
                 rpcPrefs: { blockExplorerUrl },
               });
             },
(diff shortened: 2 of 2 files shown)


app/scripts/controllers/network/network-controller.ts
```diff
@@ -742,10 +742,10 @@ export class NetworkController extends EventEmitter {
     const network = BUILT_IN_INFURA_NETWORKS[type];
     await this.#setProviderConfig({
       type,
-      rpcUrl: '',
```
Gudahtt, line 745: Nit: could we omit these properties instead of setting them to `undefined`? That seems closer to what we want to do, allowing us to improve the provider config type later on when we manage to enable exact optional property types.

Gudahtt (dismissed):
LGTM! Just one small suggestion

app/scripts/controllers/network/network-controller.ts
mcmire: Yeah that would definitely be ideal. I was trying to match core on this but there's no reason why we couldn't delete these properties in this PR and then add a PR on the core side that removes them here. Would be easy changes either way.

brad-decker (dismissed):

Gudahtt:
Rebased to resolve a conflict

brad-decker (approved):

Gudahtt (approved):

PR #18690, opened 2023-04-20
Fix mv3 beta build

danjm:
This PR does two things:

1. It ensures that the `.circleci/scripts/trigger-beta-build.sh` creates MV3 builds (for browsers that support it)
2. It prevents beta being built, and linted, for firefox. This is something we only want to do once we are supporting MV3 for firefox

For 1, we just need to export the appropriate environment variable from the `trigger-beta-build` script.

For 2, this PR adds a `platform` build flag, which can be passed to the `yarn build` command if you only want to create a build for one platform. This flag is then used in `.circleci/scripts/trigger-beta-build.sh` 

This could be tested by:
1. Branching off of this branch with a branch named `Version-v99.99.99`
2. Once that branch is built, push a commit named `Version-v1099.99.99-beta.0`
3. See that the build succeeds, that the chrome build is mv3, and that there is not firefox beta build


development/build/manifest.js
```diff
@@ -22,6 +22,11 @@ function createManifestTasks({
   shouldIncludeSnow,
   entryTask,
 }) {
+  // TODO: remove this filter for firefox once MV3 is supported in it
+  const mv3BrowserPlatforms = browserPlatforms.filter(
```
Gudahtt, line 26: Is the intent here for this to produce an MV2 build for Firefox, even if `ENABLE_MV3` is set?

If so, there are a bunch of other build changes triggered by `ENABLE_MV3`. Right now it seems that it'd produce a mixed MV2/MV3 build for Firefox, not one or the other.

danjm: oh, well it is handled in `createManifestV3AppInitializationBundle`, but was missed when scuttling logic was added in `setupScuttlingWrapping`

Gudahtt: Check out line 40 of this file; even on a Firefox build, we'll be incorporating MV3 platform modifications. There's a reference below where the manifest description will be updated to say it's MV3 as well.

danjm: okay, yes, I see that now, thanks

danjm:
@Gudahtt I have reverted previous changes in this PR and taken a new approach (i.e. just not building beta for firefox). The PR description has been updated to reflect the new changes. The old changes are left in the commit history for reference.

Gudahtt (approved):
LGTM! Great solution, this will be convenient for local debug builds too

pedronfigueiredo (approved):

PR #18713, opened 2023-04-21
Stop transpiling libraries using ES2020

FrederikBolding:
## Explanation

Stops transpiling libraries that are using ES2020, since we now support Chrome 80. The transpilation was added as a temporary measure here: #17251


Gudahtt (approved):
LGTM!

Mrtenz (approved):

PR #18729, opened 2023-04-21
Update minimum browser versions

Gudahtt:
## Explanation

The minimum browser versions have been updated in accordance to our new browser support policy, which is to support ~2 years of Chrome releases and the previous (EOL) Firefox extended support release.

Notice of this update is to be included in v10.30.0, so this should not be included in a release until v10.31.0.

## Manual Testing Steps

It should no longer be possible to install the extension on versions below the new minimums.

## Pre-merge author checklist

- [x] I've clearly explained:


vthomas13:
@Gudahtt Does this require an update to the constants for [browser outdated notifications?](https://github.com/MetaMask/metamask-extension/pull/17027/files#diff-ce12141c9013e62b6155cc6bc3af4ae75a16ec25a9e1b4da7aa5618f68b17edd)

Gudahtt:
I don't think so; they were updated already in https://github.com/MetaMask/metamask-extension/pull/18721

Gudahtt:
This is in draft because the babelrc file change seems to have introduced a build error. I believe it's because some polyfill is no longer being included, and this is causing the build to fail. I have yet to investigate further.

PR #18767, opened 2023-04-24
Experiment with tighter module grouping

FrederikBolding:
WIP


FrederikBolding:
Turns out this does not result in faster load times. Closing.

PR #18768, opened 2023-04-24
Only build LavaMoat runtime once

FrederikBolding:
## Explanation

Currently we are building the LavaMoat runtime for each sub task we run, because all the subtasks use `build/index.js` as the entry point. This seems unintentional, but open for feedback.


weizman (approved):
lgtm

brad-decker (approved):

PR #18780, opened 2023-04-24
Improve logging at build-time

FrederikBolding:
## Explanation

Improves logging at build-time. We were previously not logging errors that occurred in the `pipeline` stream. We also were skipping some `stdout` output unnecessarily.

### DIFF
--- a/development/build/scripts.js
+++ b/development/build/scripts.js
@@ -1074,6 +1074,11 @@ async function createBundle(buildConfiguration, { reloadOnChange }) {
         logError(error);
         process.exit(1);
       });
+      pipeline.on('error', (error) => {
+        console.error('Pipeline failed! See details below.');
+        logError(error);
+        process.exit(1);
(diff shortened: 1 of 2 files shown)


brad-decker (approved):

kumavis (approved):

FrederikBolding:
> Unfortunately NG:
> 
> * https://circleci.com/gh/MetaMask/metamask-extension/1237584
> * https://circleci.com/gh/MetaMask/metamask-extension/1237588
> * https://circleci.com/gh/MetaMask/metamask-extension/1237589
> 
> Locally:
> 
> ```
> $ time yarn build dist
> Command exited with non-zero status 1
> real    1m 10.15s
> user    1m 4.36s
> sys     0m 12.16s
> ```

Yeah, no guarantees that this is enough. I still think there are error cases we don't correctly log 😞

PR #18788, opened 2023-04-24
Revert "Shallow git clone (#18491)"

brad-decker:
This reverts commit b704a3d60f407eac99a7dc3f8a65529eff812203.

## Explanation
in #18491 there was discussion that this does not work for all forks. In #18783 @HowardBraham is experimenting with fixes for this issue. For now, this will get fork builds working


Gudahtt (approved):
LGTM!

legobeat (approved):

PR #18791, opened 2023-04-25
trigger approval actions from FE

bergarces:
## Explanation

- Updates assets-controllers `TokensController`, which creates the need to trigger the acceptance or rejection of the approval from the FE.
- A follow up PR will remove the interactions with the `TokensController` from the FE and just accept/reject the approval.
- Added E2E tests for the `wallet_watchAsset` action, as only manually adding the token was being tested.

Fixes https://github.com/MetaMask/MetaMask-planning/issues/453

## Manual Testing Steps

- Open Test Dapp
- Access EIP 747 section
- Try to add a new ERC20 token

## Pre-merge author checklist

- [X] I've clearly explained:


ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.js
```diff
@@ -119,8 +125,8 @@ const ConfirmAddSuggestedToken = () => {
   const handleAddTokensClick = useCallback(async () => {
     await Promise.all(
       suggestedAssets.map(async ({ asset, id }) => {
```
bergarces, line 127: I'm not all that sure about this `map` with an `async` function inside that awaitsthe `dispatch` but doesn't await `trackEvent`.

I considered returning the three promises with a `flatMap` instead, but I don't want to change existing behaviour so I'm going to assume there's a reason for it (such as not wanting to throw if `trackEvent` fails).

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.js
```diff
@@ -136,10 +142,24 @@ const ConfirmAddSuggestedToken = () => {
...
     history.push(mostRecentOverviewPage);
   }, [dispatch, history, trackEvent, mostRecentOverviewPage, suggestedAssets]);
 
+  const handleCancelClick = useCallback(async () => {
+    await Promise.all(
+      suggestedAssets.flatMap(({ id }) => [
+        dispatch(
+          rejectPendingApproval(
+            id,
+            serializeError(ethErrors.provider.userRejectedRequest()),
```
bergarces, line 154: Copied this error from the permissions confirmation screen.

https://github.com/MetaMask/metamask-extension/blob/develop/ui/pages/permissions-connect/permissions-connect.component.js#L383

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.js
```diff
@@ -118,9 +124,9 @@ const ConfirmAddSuggestedToken = () => {
 
   const handleAddTokensClick = useCallback(async () => {
     await Promise.all(
-      suggestedAssets.map(async ({ asset, id }) => {
-        await dispatch(acceptWatchAsset(id));
-
+      suggestedAssets.flatMap(({ asset, id }) => [
+        dispatch(resolvePendingApproval(id, null)),
+        dispatch(acceptWatchAsset(id)),
```
bergarces, line 129: This will not execute if I just add another line with `await dispatch(acceptWatchAsset(id));`, I have to return the whole thing with a flatMap.

https://app.circleci.com/pipelines/github/MetaMask/metamask-extension/45107/workflows/70f1b4d8-b672-4e04-984c-f22e56e3acad/jobs/1275053

OGPoyraz: Nice one!

OGPoyraz (dismissed):

OGPoyraz:
I will approve again once patches removed 👍

test/e2e/tests/add-hide-token.spec.js
```diff
@@ -123,3 +123,119 @@ describe('Add existing token using search', function () {
...
+
+        await driver.switchToWindowWithTitle(
+          'MetaMask Notification',
+          windowHandles,
+        );
+
+        await driver.clickElement({
+          tag: 'button',
+          text: 'Cancel',
+        });
```
matthewwalsh0, line 237: Should we be asserting something at this point? Such as that there is one less window, or that our original request promise was rejected?

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.js
matthewwalsh0: I know this code won't last long but would another option have been to dispatch the `acceptWatchAsset` with a `then` to chain it to the `resolvePendingApproval`?

It's probably redundant as there was no error handling here anyway but it would ensure we don't resolve the approval and hide the confirmation until the accept side effect is completed.

test/e2e/tests/add-hide-token.spec.js
bergarces: I couldn't see what we could be asserting here. By pressing reject it just won't add the currency, but there's no e2e helper method that I could find to check that an element has not been added. The parent element is also not particularly helpful in this case (no clear list of elements underneath to check assert the length).

When I checked other e2e tests such as the add custom network, it just presses the cancel button and doesn't assert anything afterwards.

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.js
```diff
@@ -118,9 +124,9 @@ const ConfirmAddSuggestedToken = () => {
 
   const handleAddTokensClick = useCallback(async () => {
     await Promise.all(
```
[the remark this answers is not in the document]
bergarces: I've been seeing how to sort it, and I can't see an easy way out.

Since `acceptWatchAsset` doesn't throw when it fails to add the asset and it doesn't return anything, I cannot wrap it on a try/catch statement to reject the pending approval instead.

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.js
bergarces: It turns out that the issue was that with the event handler being async, it was trying to assert things before they were being called. It's been sorted using `act` (https://legacy.reactjs.org/docs/testing-recipes.html#act).

test/e2e/tests/add-hide-token.spec.js
bergarces: Added additional validation.

PR #18795, opened 2023-04-25
[MMI] Add confirm-transaction-base code fences

albertolive:
## Explanation

Add confirm-transaction-base code fences:

https://consensyssoftware.atlassian.net/browse/MMI-3009

## Pre-merge author checklist

- [X] I've clearly explained:


Karlitos88:
Atalasia no me deja entrar gracias pero no podemos cuando 
No es uno es otro cansado ya nada otra vez sera

zone-live:
@SocketSecurity ignore-all

ui/pages/confirm-transaction-base/confirm-transaction-base.container.js
```diff
@@ -177,6 +192,10 @@ const mapStateToProps = (state, ownProps) => {
 
   const methodData = getKnownMethodData(state, data) || {};
 
+  ///: BEGIN:ONLY_INCLUDE_IN(build-mmi)
+  const accountType = getAccountType(state, fromAddress);
```
zone-live, line 196: why not place this line, in the code fence bellow, L216, and this way it's one less code fence in the file.

ui/pages/confirm-transaction-base/confirm-transaction-base.container.js
```diff
@@ -194,28 +213,24 @@ const mapStateToProps = (state, ownProps) => {
...
 
-  const hardwareWalletRequiresConnection =
-    doesAddressRequireLedgerHidConnection(state, fromAddress);
-
-  const isMultiLayerFeeNetwork = getIsMultiLayerFeeNetwork(state);
-
   ///: BEGIN:ONLY_INCLUDE_IN(build-mmi)
-  const accountType = getAccountType(state);
-
   const fromChecksumHexAddress = toChecksumHexAddress(fromAddress);
```
zone-live, line 217: above this line we can have the:
`  const accountType = getAccountType(state, fromAddress);`

zone-live (dismissed):

zone-live (dismissed):

zone-live (dismissed):

ui/pages/confirm-transaction-base/confirm-transaction-base.container.js
```diff
@@ -63,8 +71,10 @@ import {
 import { getGasLoadingAnimationIsShowing } from '../../ducks/app/app';
 import { isLegacyTransaction } from '../../helpers/utils/transactions.util';
 import { CUSTOM_GAS_ESTIMATE } from '../../../shared/constants/gas';
+
 ///: BEGIN:ONLY_INCLUDE_IN(build-mmi)
 import { getAccountType } from '../../selectors/selectors';
```
danjm, line 76: I'm not sure exactly what principles Brad has been encouraging in reviews so far, but I would say we don't need to add code fences unless necessary. I think these imports, and their use below, can safely be added to this file without fences.

In fact, I think we could remove all fences from this file. Perhaps a selector, or multiple selectors, could be created that contain the mmi specific logic. And they could be named to make it clear they are mmi specific. Then the appropriate properties can be returned from the `map*ToProps` functions, and the component can decide which properties to use.

In general, I think the fewer code fences the better, and this strikes me as the sort of file where we can safely avoid any code fences.

ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
```diff
@@ -687,31 +696,68 @@ export default class ConfirmTransactionBase extends Component {
       () => {
         this._removeBeforeUnload();
 
+        ///: BEGIN:ONLY_INCLUDE_IN(build-mmi)
```
danjm, line 699: I think adding code fences throughout the `handleSubmit` function makes this hard to read.

As an alternative, we could add another method to the component `handleMMISubmit`. Then, at the point of submission, we can decide whether to call `handleSubmit` or `handleMMISubmit` based on what build we are in. Perhaps this could be facilitated by renaming the current `handleSubmit` to `handleMainSubmit`, and then write a new `handleSubmit` which is just responsible for calling the appropriate `handle*Submit` method.

This might also allow us to eliminate code fencing from this file.

ui/pages/confirm-transaction-base/confirm-transaction-base.container.js
albertolive: Hey, @danjm, thank you for the provided feedback.

About code fencing, at least, I didn't know that. Please let me know if I'm wrong @zone-live, but we have added as many code fences as necessary.

I will remove those fences from this file, and I'll have it in mind for future PRs but keep in mind that all the PRs we have merged "may" not be done correctly.

ui/pages/confirm-transaction-base/confirm-transaction-base.component.js
albertolive: Yes, you're right, I'll improve it with your suggestions. Thank you

albertolive: Without looking at it too exhaustively, my only concern with this approach is that MMI needs to execute the same code as MM "but" add extra code to it. So having two different methods could duplicate the code. Wdyt?

albertolive: I updated the code. Please, let me know if the current approach is the best.

ui/pages/confirm-transaction-base/confirm-transaction-base.container.js
albertolive: I updated the code. Please, let me know if the current approach is the best.

zone-live (dismissed):

albertolive:
@danjm There was a weird error in the pipeline when I removed the code fences from the container, so I had to add them back:

https://app.circleci.com/pipelines/github/MetaMask/metamask-extension/46999/workflows/dade891b-ff9c-4d2c-8927-cf23d312453c/jobs/1337777

I kept them out in the component and improved the things you mentioned last week. Could you check it?

zone-live (approved):

pedronfigueiredo (approved):

legobeat (changes requested):

zone-live:
> @zone-live I do think that the comments above (which were being written just as this was being merged, apparently) should be resolved until this is ready for prod? Follow-up in #19328

Yes, I'll cc @albertolive about it and provide him context. 👍🏼

PR #18379, opened 2023-03-30
Consume Decrypt Message Manager from @metamask/message-manager

OGPoyraz:
## Explanation
This PR aims to move core `DecryptMessageManager` logic into core, consume `DecryptMessageManager` and create it's own `DecryptMessageController` in extension.

For more info visit: 
https://github.com/MetaMask/MetaMask-planning/issues/415
https://github.com/MetaMask/MetaMask-planning/issues/393

`@metamask/message-manager` PR:
https://github.com/MetaMask/core/pull/1149

## Manual Testing Steps
1. Go test-dapp
2. Scroll down to "Encrypt / Decrypt"
3. Get Encrytiontion Key - Accept Confirmation 
4. Put message to encrypt - Click Encrypt
5. Click Decrypt (This is the step refactored)
6.a. If you cancel it should close Confirmation
6.b. If you accept it should show the text you originally encrypt in the box below 

## Pre-merge author checklist

- [X] I've clearly explained:
  - [X] What problem this PR is solving
  - [X] How this problem was solved
  - [X] How reviewers can test my changes
- [X] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [ ] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
- [ ] PR is linked to the appropriate GitHub issue
- [ ] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.

In this case, a QA Engineer approval will be be required.

PR #18796, opened 2023-04-25
feat: Refactor Transaction Confirmation selector

OGPoyraz:
## Explanation

This PR aims to determine confirmations from `state.metamask.pendingApprovals` instead of domain controller states.

## Manual Testing Steps

There are already e2e tests that test this functionality. (E.g `multiple-transactions.spec.js`, `encrypt-decrypt.spec.js` )

## Pre-merge author checklist

- [X] I've clearly explained:

### DIFF
--- a/app/scripts/background.js
+++ b/app/scripts/background.js
@@ -714,6 +714,7 @@ export function setupController(
   //
 
   updateBadge();
+  controller.txController.addInitialTransactionApprovals(updateBadge);
   controller.txController.on(
     METAMASK_CONTROLLER_EVENTS.UPDATE_BADGE,
     updateBadge,
--- a/app/scripts/background.js
+++ b/app/scripts/background.js
@@ -713,8 +713,9 @@ export function setupController(
   // User Interface setup
   //
 
-  updateBadge();
-  controller.txController.addInitialTransactionApprovals(updateBadge);
+  controller.txController.initApprovals().then(() => {
+    updateBadge();
+  });
   controller.txController.on(
     METAMASK_CONTROLLER_EVENTS.UPDATE_BADGE,
     updateBadge,
--- a/app/scripts/controllers/transactions/index.js
+++ b/app/scripts/controllers/transactions/index.js
@@ -2646,7 +2646,7 @@ export default class TransactionController extends EventEmitter {
     );
   }
 
-  _requestApproval(txMeta) {
+  _requestApproval(txMeta, shouldShowRequest = true) {
     const id = this._getApprovalId(txMeta);
     const { origin } = txMeta;
     const type = ApprovalType.Transaction;
@@ -2661,7 +2661,7 @@ export default class TransactionController extends EventEmitter {
           type,
           requestData,
         },
-        true,
+        shouldShowRequest,
       )
       .catch(() => {
         // Intentionally ignored as promise not currently used
@@ -2694,5 +2694,17 @@ export default class TransactionController extends EventEmitter {
 
   _getApprovalId(txMeta) {
     return String(txMeta.id);
+  }
+
+  _addUnapprovedTxsApprovals() {
+    const unapprovedTxs = this.txStateManager.getUnapprovedTxList();
+    Object.values(unapprovedTxs).forEach((txMeta) => {
+      this._requestApproval(txMeta, false);
+    });
+  }
+
+  addInitialTransactionApprovals(callback) {
+    this._addUnapprovedTxsApprovals();
+    callback();
   }
 }
--- a/app/scripts/controllers/transactions/index.js
+++ b/app/scripts/controllers/transactions/index.js
@@ -2646,7 +2646,10 @@ export default class TransactionController extends EventEmitter {
     );
   }
 
-  _requestApproval(txMeta, showPopup = true) {
+  async _requestApproval(
+    txMeta,
+    { shouldShowRequest } = { shouldShowRequest: true },
+  ) {
     const id = this._getApprovalId(txMeta);
     const { origin } = txMeta;
     const type = ApprovalType.Transaction;
@@ -2661,7 +2664,7 @@ export default class TransactionController extends EventEmitter {
           type,
           requestData,
         },
-        showPopup,
+        shouldShowRequest,
       )
       .catch(() => {
         // Intentionally ignored as promise not currently used
@@ -2696,11 +2699,18 @@ export default class TransactionController extends EventEmitter {
     return String(txMeta.id);
   }
 
-  initApprovals() {
+  /**
+   * Creates approvals for all unapproved transactions in the txStateManager.
+   *
+   * @returns {Promise<void>}
+   */
+  async initApprovals() {
     const unapprovedTxs = this.txStateManager.getUnapprovedTxList();
     return Promise.all(
       Object.values(unapprovedTxs).map((txMeta) =>
-        this._requestApproval(txMeta, false),
+        this._requestApproval(txMeta, {
+          shouldShowRequest: false,
+        }),
       ),
     );
   }
--- a/app/scripts/controllers/transactions/index.js
+++ b/app/scripts/controllers/transactions/index.js
@@ -394,6 +394,22 @@ export default class TransactionController extends EventEmitter {
     });
   }
 
+  /**
+   * Creates approvals for all unapproved transactions in the txStateManager.
+   *
+   * @returns {Promise<void>}
+   */
+  async initApprovals() {
+    const unapprovedTxs = this.txStateManager.getUnapprovedTxList();
+    return Promise.all(
+      Object.values(unapprovedTxs).map((txMeta) =>
+        this._requestApproval(txMeta, {
+          shouldShowRequest: false,
+        }),
+      ),
+    );
+  }
+
   // ====================================================================================================================================================
 
   /**
@@ -2697,21 +2713,5 @@ export default class TransactionController extends EventEmitter {
 
   _getApprovalId(txMeta) {
     return String(txMeta.id);
-  }
-
-  /**
-   * Creates approvals for all unapproved transactions in the txStateManager.
-   *
-   * @returns {Promise<void>}
-   */
-  async initApprovals() {
-    const unapprovedTxs = this.txStateManager.getUnapprovedTxList();
-    return Promise.all(
-      Object.values(unapprovedTxs).map((txMeta) =>
-        this._requestApproval(txMeta, {
-          shouldShowRequest: false,
-        }),
-      ),
-    );
   }
 }
--- a/app/scripts/controllers/transactions/index.js
+++ b/app/scripts/controllers/transactions/index.js
@@ -2646,13 +2646,13 @@ export default class TransactionController extends EventEmitter {
     );
   }
 
-  _requestApproval(txMeta, shouldShowRequest = true) {
+  _requestApproval(txMeta, showPopup = true) {
     const id = this._getApprovalId(txMeta);
     const { origin } = txMeta;
     const type = ApprovalType.Transaction;
     const requestData = { txId: txMeta.id };
 
-    this.messagingSystem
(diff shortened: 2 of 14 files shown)


ui/selectors/approvals.ts
```diff
@@ -1,19 +1,71 @@
 import { ApprovalControllerState } from '@metamask/approval-controller';
 import { ApprovalType } from '@metamask/controller-utils';
+import { transactionMatchesNetwork } from '../../shared/modules/transaction.utils';
+import { TransactionMeta } from '../../shared/constants/transaction';
+import { getCurrentChainId } from './selectors';
 
 type ApprovalsMetaMaskState = {
   metamask: {
     pendingApprovals: ApprovalControllerState['pendingApprovals'];
+    unapprovedTxs: {
```
OGPoyraz, line 10: `unapprovedTxs` doesn't belong `ApprovalsMetaMaskState` but didn't want to create a separate type just for this

ui/helpers/utils/confirm-tx.util.ts
```diff
@@ -86,22 +85,6 @@ export function convertTokenToFiat({
...
- * This is a selector and probably doesn't belong here but its staying for now
- * Note: I did not go so far as to type the entirety of the MetaMask state tree
- * which definitely needs to be done for the full conversion of TypeScript to
- * be successful and as useful as possible.
- * TODO: Type the MetaMask state tree and use that type here.
- *
- * @param state - MetaMask state
- * @returns true if there are unconfirmed transactions in state
- */
-export function hasUnconfirmedTransactions(
```
OGPoyraz, line 99: We consume our selector, this is not needed anymore.

app/scripts/controllers/transactions/index.js
```diff
@@ -2654,4 +2654,16 @@ export default class TransactionController extends EventEmitter {
   _getApprovalId(txMeta) {
     return String(txMeta.id);
   }
+
+  _addUnapprovedTxsApprovals() {
```
OGPoyraz, line 2658: First off, I tried calling this function on constructor, but it didn't work out. Because adding approvals are async operation in `ApprovalController` and  it couldn't be deferred before `updateBadge` call in `background.js`. 

Instead, I've added `addInitialTransactionApprovals` which actually checks and add `unapprovedTxs` to `ApprovalController` approvals. Then it calls callback, in our case it is `updateBadge`.

ui/selectors/approvals.ts
```diff
@@ -1,19 +1,78 @@
...
   return pendingApprovalRequests.length > 0;
 }
+
+export function hasTransactionPendingApprovalsSelector(
+  state: ApprovalsMetaMaskState,
+) {
+  return (
+    hasUnapprovedTransactionsInCurrentNetwork(state) ||
+    // istanbul ignore next: this will be removed after merging https://github.com/MetaMask/metamask-extension/pull/18379
+    hasPendingDecryptApprovals(state) ||
```
OGPoyraz, line 69: This will be changed into `hasPendingApprovalsSelector(state, ApprovalType.EthDecrypt)`only after https://github.com/MetaMask/metamask-extension/pull/18379

Karlitos88:
Ya está quel al final no se quien soy , para perder mi
tiempo y como todos paseando para no ser quien
quereis sea pues solo haceis darme largas lo mando todo tomar por      ...y
gracias por todo así no soy

app/scripts/background.js
```diff
@@ -678,6 +678,7 @@ export function setupController(
   //
 
   updateBadge();
+  controller.txController.addInitialTransactionApprovals(updateBadge);
```
bergarces, line 681: If we end up needing to restore the pending approvals for multiple controllers, we might want to come up with a more generic name so that we can call the same method by just passing a list of controllers here.

But that's only if that ends up being the case. Also, we can change it afterwards if needed, so this is just an observation.

app/scripts/controllers/transactions/index.js
```diff
@@ -2605,7 +2605,7 @@ export default class TransactionController extends EventEmitter {
     );
   }
 
-  _requestApproval(txMeta) {
+  _requestApproval(txMeta, shouldShowRequest = true) {
```
matthewwalsh0, line 2608: Maybe `showPopup` for clarity? And another optional options object to aid readability?

app/scripts/controllers/transactions/index.js
matthewwalsh0: As we're not awaiting the `_requestApproval` method here which in turn isn't returning the promise, won't we still be calling the `callback` immediately which would be the same as calling it after from `background.js`?

app/scripts/controllers/transactions/index.js
```diff
@@ -2654,4 +2654,16 @@ export default class TransactionController extends EventEmitter {
...
   }
+
+  _addUnapprovedTxsApprovals() {
+    const unapprovedTxs = this.txStateManager.getUnapprovedTxList();
+    Object.values(unapprovedTxs).forEach((txMeta) => {
+      this._requestApproval(txMeta, false);
+    });
+  }
+
+  addInitialTransactionApprovals(callback) {
```
matthewwalsh0, line 2665: As mentioned previously, if we're not awaiting anything, then the `callback` shouldn't be needed.

matthewwalsh0, line 2665: Perhaps just `initApprovals` as `Transaction` is implied, and it keeps it generic for any future usage?

app/scripts/metamask-controller.js
```diff
@@ -259,6 +259,7 @@ export default class MetamaskController extends EventEmitter {
         MESSAGE_TYPE.ETH_SIGN,
         MESSAGE_TYPE.PERSONAL_SIGN,
         MESSAGE_TYPE.ETH_SIGN_TYPED_DATA,
+        MESSAGE_TYPE.TRANSACTION,
```
matthewwalsh0, line 262: This will clash with the other pending PR, but we have the `ApprovalType` enum now instead.

ui/selectors/approvals.ts
```diff
@@ -1,19 +1,71 @@
 import { ApprovalControllerState } from '@metamask/approval-controller';
 import { ApprovalType } from '@metamask/controller-utils';
+import { transactionMatchesNetwork } from '../../shared/modules/transaction.utils';
```
matthewwalsh0, line 3: If we're trying to isolate the approval selectors in a new file, I'd recommend not importing any of these transaction helpers to avoid coupling it to the transaction logic.

We could either move the entire new selector out, or add a callback argument such as `matchesCurrentChain`.

ui/selectors/approvals.ts
```diff
@@ -1,19 +1,71 @@
...
+) {
+  const pendingApprovalRequests = getApprovalRequestsByType(
+    state,
+    approvalType,
+  );
+
   return pendingApprovalRequests.length > 0;
 }
+
+export function hasTransactionPendingApprovalsSelector(
```
matthewwalsh0, line 57: I don't think we need the `Selector` suffix as it's implied by the name and arguments, plus not used in some of the existing selector files like `permissions.js`.

ui/selectors/approvals.ts
```diff
@@ -1,19 +1,71 @@
...
+  state: ApprovalsMetaMaskState,
+) {
+  return (
+    hasUnapprovedTransactionsInCurrentNetwork(state) ||
+    hasPendingApprovalsSelector(state, ApprovalType.EthDecrypt) ||
```
matthewwalsh0, line 62: We could avoid some duplication here with `some` like:

```
[ApprovalType.EthDecrypt, ApprovalType.EthSign, ...].some(approvalType => hasPendingApprovalsSelector(state, approvalType)
```
```

File: app/scripts/controllers/transactions/index.test.js [hunk 12]
```diff
@@ -2991,4 +2991,37 @@ describe('Transaction Controller', function () {
       assert.equal(result.type, TransactionType.simpleSend);
     });
   });
+
+  describe('addInitialTransactionApprovals', function () {
+    it('adds unapprovedTxs as approvals', async function () {
+      const txId = '1';
+      txController.addTransaction(
```
matthewwalsh0, line 2998: Is it safer to add two transactions to verify we're iterating and support multiple pending approvals?

app/scripts/background.js
matthewwalsh0: I'd suggested the name `initApprovals` which would be generic enough to support this.

app/scripts/controllers/transactions/index.test.js
OGPoyraz: Done

ui/selectors/approvals.ts
OGPoyraz: Fixed

ui/selectors/approvals.ts
OGPoyraz: Thanks fixed

ui/selectors/approvals.ts
OGPoyraz: Moved out the selector itself.

app/scripts/controllers/transactions/index.js
bergarces: I also think that it'd be best to keep the name used in ApprovalController in this instance. Unless we plan to change that variable there as well.

bergarces:
Do we have a plan for the sinon/assert pipeline error?

OGPoyraz:
> Do we have a plan for the sinon/assert pipeline error?
@bergarces I was going to ask it on dev channel but since this is not a blocker, I will just ignore it and merge.

app/scripts/controllers/transactions/index.js
```diff
@@ -2662,4 +2665,20 @@ export default class TransactionController extends EventEmitter {
   _getApprovalId(txMeta) {
     return String(txMeta.id);
   }
+
+  /**
+   * Creates approvals for all unapproved transactions in the txStateManager.
+   *
+   * @returns {Promise<void>}
+   */
+  async initApprovals() {
```
matthewwalsh0, line 2674: I think the standard is all public methods then private so we can move this method up under the last public method.

ui/selectors/approvals.ts
```diff
@@ -1,19 +1,35 @@
...
-) {
+) => {
   const pendingApprovalRequests = Object.values(
     state.metamask.pendingApprovals,
   ).filter(({ type }) => type === approvalType);
 
+  return pendingApprovalRequests;
+};
+
+export function hasPendingApprovalsSelector(
```
matthewwalsh0, line 25: Should we remove the `Selector` here also?

ui/selectors/transactions.js
```diff
@@ -534,3 +539,32 @@ export const submittedPendingTransactionsSelector = createSelector(
...
+  );
+
+  return filteredUnapprovedTxInCurrentNetwork.length > 0;
+};
+
+export function hasTransactionPendingApprovals(state) {
+  return (
+    hasUnapprovedTransactionsInCurrentNetwork(state) ||
+    [
+      ApprovalType.EthDecrypt,
```
matthewwalsh0, line 563: Not crucial now but this might be best in a local constant such as `TRANSACTION_APPROVAL_TYPES`.

ui/selectors/approvals.ts
OGPoyraz: Fixed

ui/selectors/transactions.js
OGPoyraz: Fixed

app/scripts/controllers/transactions/index.js
OGPoyraz: Fixed

matthewwalsh0 (dismissed):

matthewwalsh0 (approved):

vinistevam (approved):

PR #18818, opened 2023-04-26
Fix for wrong type being assigned to the transaction

### LINKED ISSUE #17865: [Bug]: 10.25 doesn't respect "advanced gas controls" settings, sometimes revert to Legacy txs
### Describe the bug

The "advanced gas control" setting does nothing. Metamask behaves as if it were turned off.

More randomly, Metamask sometimes forces legacy transactions. No baseFee/priorityFee.

Both problems started appearing with 10.25.

### Steps to reproduce

1.

_No response_

### Version

10.25

### Build type

None

### Browser

Chrome

### Operating system

Windows

### Hardware wallet

_No response_

### Additional context

_No response_

jpuri:
Fixes: #17865

The PR fixes issue with EIP-1559 transaction sometimes behave as legacy transaction. The issue is bit hard to replicate, thus fixing it is also hard. @seaona is able to replicate it with internet connection throttling in browser.

I found that default value of `eip1559support` for network is `false`. It actual value is obtained by querying latest blocks in the network. If the internet connect is slow it can happen that it takes time to get latest block and hence check if network support `EIP-1559` - and thus. the default value `false` is used for send flow.

I added a check to see if `eip1559support` is undefined and if so a synchronous call is made to fetch it from network before continuing with the transaction.

@seaona tested this PR and the issue is not replicating in this PR.


NiranjanaBinoy (approved):
LGTM

legobeat (approved):
LGTM

PR #18824, opened 2023-04-26
fix ff diff generation

pedronfigueiredo:
## Explanation

Improved the way the diff is generated on PRs made by external contributors through forks.

Based on the comments on https://github.com/MetaMask/metamask-extension/pull/18508#discussion_r1176812913


.github/workflows/fitness-functions.yml
```diff
@@ -25,5 +27,9 @@ jobs:
           BASE_REF: ${{ github.event.pull_request.base.ref }}
         run: |
           git fetch origin $BASE_REF
```
Gudahtt, line 29: Nit: The `fetch-depth: 0` suggestion was meant as a way to replace this step

```suggestion
```

.github/workflows/fitness-functions.yml
```diff
@@ -24,6 +26,9 @@ jobs:
         env:
           BASE_REF: ${{ github.event.pull_request.base.ref }}
         run: |
-          git fetch origin $BASE_REF
-          git diff origin/$BASE_REF HEAD -- . > diff
+          # The following command generates a diff of changes between the common
+          # ancestor of $BASE_REF and HEAD, and the current commit (HEAD), for
+          # files in the current directory and its subdirectories. The output is
+          # then saved to a file called "diff".
+          git diff "$(git merge-base origin/$BASE_REF HEAD)" HEAD -- . > diff
```
Gudahtt, line 33: I just tried running this through [shellcheck](https://www.shellcheck.net/), and it also suggested two more changes:

```suggestion
          git diff "$(git merge-base "origin/$BASE_REF" HEAD)" HEAD -- . > ./diff
```

The quoting helps ensure this works correctly if `$BASE_REF` includes spaces somehow, and the `./` in front of `diff` ensures that this content is sent to the _file_ `diff` rather than the command `diff`.

Gudahtt (approved):
LGTM!

DDDDDanica (approved):

brad-decker (approved):

PR #18829, opened 2023-04-26
Handle watch asset accept and reject using ApprovalController only

bergarces:
## Explanation

- Updates assets-controllers `TokensController`, which makes it so that only acceptance or rejection of the approval are triggered from the FE. It will no longer call `TokensController`.

Fixes https://github.com/MetaMask/MetaMask-planning/issues/463

## Manual Testing Steps

- Open Test Dapp
- Access EIP 747 section
- Try to add a new ERC20 token

## Pre-merge author checklist

- [X] I've clearly explained:


app/scripts/lib/rpc-method-middleware/handlers/watch-asset.js
```diff
@@ -37,14 +36,10 @@ async function watchAssetHandler(
...
   try {
     const { options: asset, type } = req.params;
-    const handleWatchAssetResult = await handleWatchAssetRequest(asset, type);
-    await handleWatchAssetResult.result;
+    await handleWatchAssetRequest(asset, type);
     res.result = true;
     return end();
   } catch (error) {
-    if (error.message === 'User rejected to watch the asset.') {
-      return end(ethErrors.provider.userRejectedRequest());
```
bergarces, line 46: This is no longer needed, since that's the error we are defining in the FE component for rejection.

ui/pages/confirm-add-suggested-token/confirm-add-suggested-token.js
```diff
@@ -13,7 +14,10 @@ import { getTokens } from '../../ducks/metamask/metamask';
 import ZENDESK_URLS from '../../helpers/constants/zendesk-url';
 import { isEqualCaseInsensitive } from '../../../shared/modules/string-utils';
 import { getSuggestedAssets } from '../../selectors';
```
adonesky1, line 16: Presumably you'll have to repoint this selector since this state will be different as it no longer lives on the TokensController?

bergarces: Good catch, we have a new selector that uses approvals filtered by type. It'll definitely need to be used here.

.yarn/patches/@metamask-assets-controllers-npm-6.0.0-0cb763bd07.patch
bergarces, line 1: Changes from this patch are already included in the new version.

ui/selectors/selectors.js
```diff
@@ -504,7 +504,7 @@ export function getCurrentCurrency(state) {
 export function getTotalUnapprovedCount(state) {
   const { pendingApprovalCount = 0 } = state.metamask;
 
-  return pendingApprovalCount + getSuggestedAssetCount(state);
```
bergarces, line 507: suggested assets already were part of pending approvals from a previous PR.

ui/store/actions.ts
```diff
@@ -2298,44 +2298,6 @@ export function addTokens(
...
-    }
-    dispatch(closeCurrentNotificationWindow());
-  };
-}
-
-export function acceptWatchAsset(
-  suggestedAssetID: string,
-): ThunkAction<void, MetaMaskReduxState, unknown, AnyAction> {
-  return async (dispatch: MetaMaskReduxDispatch) => {
-    dispatch(showLoadingIndication());
```
matthewwalsh0, line 2324: For slower connections, do we lose the loading indicator now?

If so, maybe it's worth looking at a separate ticket to add this into the `resolvePendingApproval` and `rejectPendingApproval` actions.

test/e2e/fixture-builder.js
```diff
@@ -289,7 +289,6 @@ function defaultFixture() {
         allTokens: {},
         detectedTokens: [],
         ignoredTokens: [],
-        suggestedAssets: [],
```
matthewwalsh0, line 292: I think I mentioned on the core PR also, but do we have a policy to remove old unused user data using migrations, or leave it to linger?

matthewwalsh0 (dismissed):

ui/store/actions.ts
bergarces: Up until now, resolving or rejecting an approval would happen very fast.

Now that we have the `expectsResult` option though, we should probably do what you suggest and add the loading indication for resolving actions.

test/e2e/fixture-builder.js
bergarces: I'm going to ask in one of the dev channels.

bergarces: Turns out a migration to get rid of the state in existing installations would be desirable in this case.

app/scripts/migrations/087.ts
bergarces, line 1: Added migration. Based on migration `085`, which has a similar use case.

vinistevam (approved):

mcmire (approved):
Patches to TokensController (at least in terms of taking care of https://github.com/MetaMask/core/pull/1261 and https://github.com/MetaMask/core/pull/1268) look good.

PR #18863, opened 2023-04-27
[Bug]: Enabling the display of token total in SetApprovalForAll 

### LINKED ISSUE #18618: [Bug]: SetApprovalForAll does not display total tokens anymore
### Describe the bug

**Problem**: whenever we trigger setApprovalForAll function for an ERC721 token, we can see how the `Your NFT mayb be at risk` warning does not display anymore the Total number of tokens we are approving.

**Now**:

Screenshot from 2023-04-17 11-59-07

**Before**:

Screenshot from 2023-04-17 12-04-48

### Steps to reproduce

1. Go to test dapp
2. Deploy Collectibles contract
3. Mint some tokens
4. Click Set Approval for All
5. Click Confirm
6. Observe how the warning does not display the Total and check design issues mentioned above

_No response_

### Version

Introduced in 10.26.1

### Build type

None

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

_No response_

NiranjanaBinoy:
## Explanation
Fixes: #18618 

Problem:
 Whenever we trigger setApprovalForAll function for an ERC721 token, we can see how the Your NFT mayb be at risk warning does not display anymore the Total number of tokens we are approving.

Explanation:
`isERC721`, and the `total` was coming up as false and 0, disabling the Total display in the SetApproveForAllWarning.
 isERC721 was coming up as false because instead of comparing `assetStandard` to ERC721, we were comparing it to ERC20. The total was coming up as 0 because of the recent change to the  `fetchTokenBalance`, which changed the format of `tokenBalance` fetched.

### Before

Screenshot 2023-04-27 at 1 12 30 PM

### After

Screenshot 2023-04-27 at 1 11 09 PM

## Manual Testing Steps

1. Go to test dapp
2. Deploy Collectibles contract
3. Mint some tokens
4. Click Set Approval for All
5. Click Confirm
7. Make sure the Total is now displayed in the pop-up warning.


ui/components/app/confirm-page-container/confirm-page-container.component.js
```diff
@@ -140,7 +140,7 @@ const ConfirmPageContainer = (props) => {
       fromAddress,
       global.ethereumProvider,
     );
-    setCollectionBalance(tokenBalance?.balance?.words?.[0] || 0);
+    setCollectionBalance(tokenBalance.toString() || 0);
```
digiwand, line 143: I came across this comment
> `decimals` and `balance` are not converted from `BN`s to `string`s automagically in Firefox (as they are in Chrome) so we need to do this for now.
  >>    decimals: details?.decimals?.toString(10),
       balance: details?.balance?.toString(10),

I wonder if we should do the same thing here 🤔  I haven't looked into this far enough yet 
cc: @adonesky1

NiranjanaBinoy: When I tried base 10, some message was thrown in the console, so I switched to the one without the base. But would like to get some insight into this.

digiwand: hmm. I suppose if we write a test for this we can check it will cover Firefox cases too. Is this something we could add?

NiranjanaBinoy: `BigNumber.toString does not accept any parameters; base-10 is assumed` This is the message in the console when I try `setCollectionBalance(tokenBalance.toString(10) || 0);` it is not exactly an error or warning, but it looks like just a log. I cannot test this in Firefox as the app is not working for Firefox in my local.

NiranjanaBinoy: @digiwand I have updated to `tokenBalance.toString(10)`. As I mentioned above, the message we see on the console for Chrome is not an error or warning but just a log. My thoughts are if this change is not having any adverse effect on Chrome, but it is something needed in Firefox, it would be better to have it in rather than leave it out. What do you think?

ui/components/app/confirm-page-container/confirm-page-container.component.js
```diff
@@ -140,7 +140,7 @@ const ConfirmPageContainer = (props) => {
       fromAddress,
       global.ethereumProvider,
     );
-    setCollectionBalance(tokenBalance?.balance?.words?.[0] || 0);
+    setCollectionBalance(tokenBalance.toString(10) || 0);
```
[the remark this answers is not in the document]
NiranjanaBinoy: Hi @digiwand , `tokenBalance` seems to be of type `any`, but it looks like the correct value is being rendered at the front end without the conversion, so I removed the `.toString(10)` conversion.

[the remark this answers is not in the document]
digiwand: removing `.toString(10)` sounds good to me. 

where are you seeing `tokenBalance` is type `any`? 

It seems that that `collectionBalance` is passed here
```
          <SetApproveForAllWarning
            collectionName={title}
            senderAddress={fromAddress}
            name={fromName}
            isERC721={assetStandard === TokenStandard.ERC20}
            total={collectionBalance}
            onSubmit={onSubmit}
            onCancel={onCancel}
          />
```

and SetApproveForAllWarning is expecting a string 
```
  /**
   * Total number of items that are being approved
   */
  total: PropTypes.string,
```

ui/components/app/confirm-page-container/confirm-page-container.component.js
```diff
@@ -140,7 +140,7 @@ const ConfirmPageContainer = (props) => {
       fromAddress,
       global.ethereumProvider,
     );
-    setCollectionBalance(tokenBalance?.balance?.words?.[0] || 0);
+    setCollectionBalance(tokenBalance || 0);
```
digiwand, line 143: I wonder how `tokenBalance?.balance?.words?.[0]` became `tokenBalance`. Do you know why? Is it possible that we may want to do something like `tokenBalance || tokenBalance?.balance?.words?.[0] || '0'`?

digiwand: a couple of discoveries:
- fetchTokenBalance calls [balanceOf](https://docs.openzeppelin.com/contracts/3.x/api/token/erc1155#ERC1155-balanceOf-address-uint256-)
- [balanceOf](https://docs.openzeppelin.com/contracts/3.x/api/token/erc1155#ERC1155-balanceOf-address-uint256-) returns uint256

digiwand (commented):
I think it would be helpful to add a test

ui/components/app/confirm-page-container/confirm-page-container.component.js
NiranjanaBinoy: There is some recent change to fetchTokenBalance in [this PR](https://github.com/MetaMask/metamask-extension/pull/17891) that could be what changes the structure of returned value. I tried to log the typeOf tokenBalance and the value of tokenBalance itself and I am attaching the results below.

Screenshot 2023-05-11 at 4 45 48 PM

digiwand: Thanks for providing the return value! This is helpful. Looks like it's a BigNumber

ref: https://docs.ethers.org/v5/api/utils/bignumber/

we have `shared/modules/Numeric.ts` which helps us convert values into BigNumbers. Then we have `shared/modules/conversion.utils.ts` which have conversions from BigNumbers to strings. thinking we will want to use a function from the ConversionUtil. I'm unsure why we don't use the [formatUtils method provided in the example](https://docs.ethers.org/v5/api/contract/example/#erc20-meta-methods) (also this [example](https://community.infura.io/t/ethers-js-how-to-retrieve-the-balance-of-an-erc-20-token/6735)). 

I was confused earlier because I was referring to this mock test value in `app/scripts/metamask-controller.test.js`
```
const MOCK_TOKEN_BALANCE = '888';

function MockEthContract() {
  return () => {
    return {
      at: () => {
        return {
          balanceOf: () => MOCK_TOKEN_BALANCE,
        };
      },
    };
  };
}
```

now we know this is a BigNumber object, we can work off of this

digiwand: I confirmed that `.toString` is good to use here
https://docs.ethers.org/v5/api/utils/bignumber/
 
We will want to remove the value `10` we passed to `toString`. Will leave another comment on the updated code. 

Resolving the conversation since it's looking like `tokenBalance?.balance?.words?.[0]` is not needed

ui/components/app/confirm-page-container/confirm-page-container.component.js
```diff
@@ -142,7 +142,7 @@ const ConfirmPageContainer = (props) => {
       fromAddress,
       global.ethereumProvider,
     );
-    setCollectionBalance(tokenBalance?.balance?.words?.[0] || 0);
+    setCollectionBalance(tokenBalance.toString(10) || '0');
```
digiwand, line 145:
```suggestion
    setCollectionBalance(tokenBalance.toString() || '0');
```

The update to use the `toString` method lgtm! One thing is passing a parameter does through a warning. 
https://github.com/ethers-io/ethers.js/blob/master/packages/bignumber/src.ts/bignumber.ts#L201

digiwand, line 145:
```suggestion
    setCollectionBalance(tokenBalance.toString() || '0');
```

The update to use the `toString` method lgtm! One thing is passing a parameter does through a warning. 
https://github.com/ethers-io/ethers.js/blob/master/packages/bignumber/src.ts/bignumber.ts#L201

> BigNumber.toString does not accept any parameters; base-10 is assumed

NiranjanaBinoy: @digiwand  Referring to your previous comment, will it be possible to verify this change without the base won't break in firefox? 
Firefox build is not working in my local, so I cannot test it. :( 

> decimals and balance are not converted from BNs to strings automagically in Firefox (as they are in Chrome) so we need to do this for now.
> 
> decimals: details?.decimals?.toString(10),
> balance: details?.balance?.toString(10),

ui/components/app/confirm-page-container/confirm-page-container.component.js
digiwand: Nice one looking out for the warning message. I think we should follow it and remove the parameter. 

ref: https://github.com/ethers-io/ethers.js/blob/master/packages/bignumber/src.ts/bignumber.ts#L201

---

Now that I've looked at the code some more it looks like the value in [the comment above](https://github.com/MetaMask/metamask-extension/pull/13788/files#r816206665) refers to `BN` object from the 'ethereumjs-util' library. In this PR, this value appears to be a `BigNumber` type from '@ethersproject/bignumber'. 

BN.toString can accept parameters - [ref](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/bn.js/index.d.ts#L75)
BigNumber.toString does not accept any parameters - [ref](https://github.com/ethers-io/ethers.js/blob/master/packages/bignumber/src.ts/bignumber.ts#L201)

Ehsan129:
Stop it.dont send Email agaeen.plz

1299820

در تاریخ دوشنبه ۱۵ مه ۲۰۲۳،‏ ۲۱:۵۴ Ariella Vu ***@***.***>
نوشت:

ui/components/app/confirm-page-container/confirm-page-container.component.js
digiwand:
```suggestion
    setCollectionBalance(tokenBalance.toString() || '0');
```

oops, I moved the conversation back to the previous comment to keep thoughts together. Please see  https://github.com/MetaMask/metamask-extension/pull/18863#discussion_r1194206975 

I think it's ok if it's passing with the GitHub tests. I can be available tomorrow to help with the Firefox build

segun, line 145: Would be very very helpful to add some type of test for this...

NiranjanaBinoy: @segun The test for changes in this PR will be covered in a separate PR #19149  @digiwand  is working on, which will require some clean upon for the confirm-page-container test file as well.

digiwand (approved):
👍

ui/components/app/confirm-page-container/confirm-page-container.component.js
segun: ok

segun (approved):

PR #18913, opened 2023-05-02
[FLASK] `snaps-monorepo@0.33.1-flask.1`

FrederikBolding:
## Explanation

Bumps Snaps Flask packages to `0.33.1-flask.1` and handles any breaking changes.

This in turn bumps `babel/core` to the latest version which has a circular dependency, requiring itself for some reason. A patch was added to remove this, making `babel/core` compatible with LavaMoat.


package.json
```diff
@@ -210,7 +210,18 @@
     "fast-json-patch@^3.1.1": "patch:fast-json-patch@npm%3A3.1.1#./.yarn/patches/fast-json-patch-npm-3.1.1-7e8bb70a45.patch",
     "request@^2.83.0": "patch:request@npm%3A2.88.2#./.yarn/patches/request-npm-2.88.2-f4a57c72c4.patch",
     "request@^2.88.2": "patch:request@npm%3A2.88.2#./.yarn/patches/request-npm-2.88.2-f4a57c72c4.patch",
-    "request@^2.85.0": "patch:request@npm%3A2.88.2#./.yarn/patches/request-npm-2.88.2-f4a57c72c4.patch"
+    "request@^2.85.0": "patch:request@npm%3A2.88.2#./.yarn/patches/request-npm-2.88.2-f4a57c72c4.patch",
+    "@babel/core@^7.12.1": "patch:@babel/core@npm%3A7.21.5#./.yarn/patches/@babel-core-npm-7.21.5-c72c337956.patch",
```
ritave, line 214: what happened here? can you give more explanation in PR description?

ritave (approved):

package.json
FrederikBolding: The latest version of `babel/core` has a circular dependency where it requires itself, this is not allowed in LavaMoat so we patch it out. We had to do the same thing in the `snaps-monorepo`. Updated PR description to reflect this.

Mrtenz (approved):

hmalik88 (approved):

PR #18914, opened 2023-05-02
Bump Circle CI docker image

FrederikBolding:
## Explanation

Stops using deprecated `circleci/node:16-browsers` image and uses `cimg/node:16.20-browsers` instead. 
https://circleci.com/developer/images/image/cimg/node

This updates the Node version we run in CI and fixes the following SES bug: https://github.com/nodejs/node/issues/43496

A few other changes were made to accommodate the new Docker image:
- Stopped using `xset` to check for X-server, this would cause tests to timeout
- Added `sudo apt-get update` to the Chrome installation script because of missing packages required for Chrome installation
- Adjusted Firefox binary location to `opt/firefox/firefox` (we might not have been running the FF version we thought before)

Don't mind the commit log 😄 
image


brad-decker (approved):

.circleci/scripts/chrome-install.sh
```diff
@@ -4,6 +4,8 @@ set -e
 set -u
 set -o pipefail
 
+sudo apt-get update
```
Gudahtt, line 7: Interesting - why was this step necessary?

Gudahtt: Oh I see, missing packages needed by Chrome.

test/e2e/helpers.js
```diff
@@ -107,12 +106,7 @@ async function withFixtures(options, testSuite) {
     }
     const mockedEndpoint = await setupMocking(mockServer, testSpecificMock);
     await mockServer.start(8000);
-    if (
-      process.env.SELENIUM_BROWSER === 'chrome' &&
-      process.env.CI === 'true'
-    ) {
-      await ensureXServerIsRunning();
```
Gudahtt, line 114: We had found that the e2e tests would intermittently fail unless this was running. This took _a long time_ for us to debug. Eventually we settled on this solution after finding that the Chromium project uses a similar strategy for their own internal test suite.

I wonder why it's not required anymore... hopefully this won't bring the intermittent failures back.

FrederikBolding: I'm not sure.. None of the Chrome E2E tests would start when this was being run with it. It seemed like it would never resolve 🤔 

Perhaps `xset` is not available on the new image? It's curious that it didn't throw though.

FrederikBolding: It is only enabled in CI though 🤔

Gudahtt: Ah, interesting. That condition was not present in [the original PR](https://github.com/MetaMask/metamask-extension/pull/11887) where this was introduced.

FrederikBolding: I would say we remove it for now and we can investigate further if it turns out to introduce intermittent test failures?

Gudahtt (approved):
LGTM!

PR #18982, opened 2023-05-03
[MMI] Don't group transactions by nonce if they are custodial Tx

zone-live:
## Explanation

This PR adds a small code fence (where it checks if the `transaction` has a `custodyId`), so that we don't have custodial transactions grouped by nonce in MMI.

## Pre-merge author checklist

- [X] I've clearly explained:

### DIFF
--- a/ui/selectors/transactions.js
+++ b/ui/selectors/transactions.js
@@ -251,7 +251,16 @@ export const nonceSortedTransactionsSelector = createSelector(
         txReceipt,
       } = transaction;
 
-      if (typeof nonce === 'undefined' || type === TransactionType.incoming) {
+      let undefinedNonceOrIncomingTransaction =
+        typeof nonce === 'undefined' || type === TransactionType.incoming;
+
+      ///: BEGIN:ONLY_INCLUDE_IN(build-mmi)
+      // Don't group transactions by nonce if they are custodial TXes
+      undefinedNonceOrIncomingTransaction =
+        undefinedNonceOrIncomingTransaction || transaction.custodyId;
+      ///: END:ONLY_INCLUDE_IN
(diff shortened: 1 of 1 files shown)


ui/selectors/transactions.js
```diff
@@ -251,7 +251,16 @@ export const nonceSortedTransactionsSelector = createSelector(
...
       } = transaction;
 
-      if (typeof nonce === 'undefined' || type === TransactionType.incoming) {
+      let undefinedNonceOrIncomingTransaction =
+        typeof nonce === 'undefined' || type === TransactionType.incoming;
+
+      ///: BEGIN:ONLY_INCLUDE_IN(build-mmi)
+      // Don't group transactions by nonce if they are custodial TXes
+      undefinedNonceOrIncomingTransaction =
+        undefinedNonceOrIncomingTransaction || transaction.custodyId;
```
brad-decker, line 260: It might be best to cast the custodyId to a boolean to help with the inference that we're just checking for its existence, rather than doing anything with its value

brad-decker (changes requested):

ui/selectors/transactions.js
zone-live: Good idea, will do 👍🏼

brad-decker: Also, lets rename 'undefinedNonceOrIncomingTransaction' to something like 'shouldNotBeGrouped' and add a documentation above the variable declaration detailing the 3 specific cases we have for it.

brad-decker (changes requested):

brad-decker (approved):

albertolive (approved):

PR #18983, opened 2023-05-03
fix: support alerts confirmation page without pending transactions

OGPoyraz:
## Explanation

For more info please visit: https://github.com/MetaMask/MetaMask-planning/issues/404
We should only disable alerts if the confirmation type is `add-ethereum-chain` and without pending transactions.

## Screenshots/Screencaps

This PR has no UI chance.

## Pre-merge author checklist

- [X] I've clearly explained:


ui/pages/confirmation/templates/add-ethereum-chain.js
```diff
@@ -181,7 +181,14 @@ function getState(pendingApproval) {
...
 }
 
-function getValues(pendingApproval, t, actions, history) {
+function getValues(
+  pendingApproval,
+  t,
+  actions,
+  history,
+  _setInputState,
+  unnaprovedTxsCount,
```
matthewwalsh0, line 190: Should we put this inside an object such as `state` perhaps so we can group any additional relevant state in future without adding more arguments to the method?

matthewwalsh0, line 190: Should we put this inside an object such as `state` perhaps so we can group any additional relevant state in future without adding more arguments to the method?

ui/pages/confirmation/templates/index.js
```diff
@@ -127,13 +127,15 @@ function getAttenuatedDispatch(dispatch) {
...
  * state of input fields in the templated component.
+ * @param {number} unnaprovedTxsCount - The number of unapproved transactions
  */
 export function getTemplateValues(
   pendingApproval,
   t,
   dispatch,
   history,
   setInputState,
+  unnaprovedTxsCount,
```
matthewwalsh0, line 138: Same here too, so it'd be grouped in both here in the `getTemplateValues` function, but also the `getValues` implementations?

ui/pages/confirmation/templates/add-ethereum-chain.js
```diff
@@ -410,6 +417,7 @@ function getValues(pendingApproval, t, actions, history) {
         ethErrors.provider.userRejectedRequest().serialize(),
       ),
     networkDisplay: !originIsMetaMask,
+    disableAlerts: !(unnaprovedTxsCount > 0),
```
matthewwalsh0, line 420: ❤️

matthewwalsh0, line 420: Do we have to assume that the `switch-ethereum-chain` would also want this value as it has an alert too and we don't want to change the behaviour?

OGPoyraz:
> Would this be a great time to add the first unit tests for both `ConfirmationPage` and the templates? Just for the logic we've updated that is.

Added some basic tests just for this alerts.

ui/pages/confirmation/templates/index.js
OGPoyraz: Fixed.

ui/pages/confirmation/templates/add-ethereum-chain.js
OGPoyraz: I think this comment is not valid anymore, because we disable alerts in their own context now.

ui/pages/confirmation/templates/add-ethereum-chain.js
```diff
@@ -32,6 +32,7 @@ const UNRECOGNIZED_CHAIN = {
 
 const MISMATCHED_CHAIN_RECOMMENDATION = {
   id: 'MISMATCHED_CHAIN_RECOMMENDATION',
+  severity: SEVERITIES.WARNING,
```
OGPoyraz, line 35: I noticed this was missing while running snapshot tests, adding this will remove some required prop warning.

ui/pages/confirmation/confirmation.js
```diff
@@ -91,10 +91,12 @@ const alertStateReducer = produce((state, action) => {
  *
  * @param {object} pendingConfirmation - a pending confirmation waiting for
  * user approval
+ * @param {object} state - The state object consist of required info to determine alerts.
+ * @param state.unapprovedTxsCount
  * @returns {[alertState: object, dismissAlert: Function]} A tuple with
  * the current alert state and function to dismiss an alert by id
  */
-function useAlertState(pendingConfirmation) {
+function useAlertState(pendingConfirmation, { unapprovedTxsCount } = {}) {
```
OGPoyraz, line 99: Destructuring object here intentionally because we don't want to pass entire object to `useEffect` in order to avoid React warnings.

ui/pages/confirmation/confirmation.test.js
```diff
@@ -0,0 +1,127 @@
...
+        expect(
+          getByText('MetaMask does not verify custom networks.'),
+        ).toBeInTheDocument();
+        expect(container.querySelector('.callout')).toBeDefined();
+        expect(container).toMatchSnapshot();
+      });
+    });
+  });
+
+  describe('switch-ethereum-chain', () => {
```
matthewwalsh0, line 75: I appreciate they're technically the same component, but to avoid this file getting too large in future, should we create separate test files for each template?

We could create a test util file with helper functions to avoid unnecessary duplication.

ui/pages/confirmation/confirmation.test.js
```diff
@@ -0,0 +1,127 @@
...
+      };
+      const store = configureMockStore(middleware)(testStore);
+      const { container } = renderWithProvider(<Confirmation />, store);
+      await waitFor(() => {
+        expect(container.querySelector('.callout')).toBeFalsy();
+        expect(container).toMatchSnapshot();
+      });
+    });
+
+    it('should show alert if there are pending txs', async () => {
```
matthewwalsh0, line 96: Do we need the inverse test also to be explicit?

OGPoyraz: It's already been asserted in the first snapshot test, do we want to duplicate this test just to check this?

ui/pages/confirmation/confirmation.test.js
OGPoyraz: Sure that definitely make sense to me.

matthewwalsh0 (approved):

bergarces (approved):
LGTM

PR #18988, opened 2023-05-03
UX Multichain: Added extra margin for native token

### LINKED ISSUE #18987: Token List UI fixes
### Spacing

- We have 0 padding between token list and the tab (only the inner 16px of the cell)
-leave 8px padding between padding and the token list 

Build

Design 

![Screenshot 2023-05-03 at 14.25.07.png](https://images.zenhubusercontent.com/61ef00ec5db2d7c84498dc72/1fd326bd-faa8-4bcb-9595-14c24ece8e5b

NidhiKJha:
This Pr is to add extra 8px gap between the native token and assets tab.

Fixes #18987 

### After

Screenshot 2023-05-03 at 6 18 33 PM

## Manual Testing Steps

- Go to the Assets Tab
- Check the spacing between the native token and tab header is "8px"

## Pre-merge author checklist

- [ ] I've clearly explained:


darkwing (changes requested):

ui/components/multichain/multichain-token-list-item/multichain-token-list-item.scss
```diff
@@ -6,3 +6,7 @@
     }
   }
 }
+
+.native-token {
+  margin-top: 8px; // for the native token of each network, we need to add extra margin on the top
```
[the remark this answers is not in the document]
NidhiKJha: Sure! I have added it based on if it's native-token or not here https://github.com/MetaMask/metamask-extension/pull/18988/commits/fb676b92a66b46830d163fe8582d6604fd95b4e6

vthomas13 (dismissed):

ui/components/app/asset-list/asset-list.js
```diff
@@ -94,6 +94,7 @@ const AssetList = ({ onClickAsset }) => {
           tokenSymbol={primaryCurrencyProperties.suffix}
           secondary={showFiat ? secondaryCurrencyDisplay : undefined}
           tokenImage={balanceIsLoading ? null : primaryTokenImage}
+          className="native-token"
         />
```
darkwing, line 98: Instead of passing this `native-token` and needing to account for that, can there be a `boxProps` prop or something where we add the spacing? We shouldn't be passing arbitrary classnames for JS purposes.

darkwing (changes requested):

ui/components/app/asset-list/asset-list.js
NidhiKJha: Hmm! Unfortunately no. `MultichainTokenListItem` is the same component that we are using to show native token and detected tokens. But this marginTop is required only for the first token cell which in this case is the native token of the network hence we need to add this margin conditionally or add it with CSS. Another approach that I can think of is to add a `margin-bottom` for the first tab using BoxProps since I had previously updated to use Box as a parent component for tab. Updated here https://github.com/MetaMask/metamask-extension/pull/18988/commits/f66dc0af9eaefbfa950789fafb1dd77cbfe976ee .

ui/pages/home/home.component.js
```diff
@@ -700,6 +700,7 @@ export default class Home extends PureComponent {
                 data-testid="home__asset-tab"
                 name={this.context.t('assets')}
                 tabKey="assets"
+                boxProps={{ marginBottom: 2 }}
```
darkwing, line 703: Hmmm, I hate going around on this, but what if we skip the new `boxProps` and simply surround `<AssetList>` with a `<Box marginTop={2}>`?

darkwing (dismissed):

ui/pages/home/home.component.js
NidhiKJha: AssetList is a non-multichain component and we only needs to update the MultichainTokenListItem. I still think going with the className approach and adding css could be a better fix since MultichainTokenListItem accept className as a prop and we can update the css for native token. wdyt?

NidhiKJha: Ah! My bad. Updated!! Thanks @darkwing

darkwing (approved):

vthomas13 (approved):

PR #18347, opened 2023-03-28
e2e test for action metrics

pedronfigueiredo:
## Explanation

End to end tests for this PR https://github.com/MetaMask/metamask-extension/pull/18346

PR #18989, opened 2023-05-03
add scuttle exceptions for JSON and Date

pedronfigueiredo:
## Explanation

As suggested in [this comment](https://github.com/MetaMask/metamask-extension/pull/18347/files#r1182496069), this PR makes changes to the build system atomically for increased visibility for the changelog.

## Screenshots of before the changes

Screenshot 2023-05-03 at 17 09 25
Screenshot 2023-05-03 at 17 02 26

### DIFF
--- a/development/build/index.js
+++ b/development/build/index.js
@@ -125,6 +125,8 @@ async function defineAndRunBuildTasks() {
         'Uint8Array',
         'String',
         'Promise',
+        'JSON',
+        'Date',
         // globals sentry needs to function
         '__SENTRY__',
         'appState',
@@ -132,8 +134,6 @@ async function defineAndRunBuildTasks() {
         'stateHooks',
         'sentryHooks',
         'sentry',
-        'JSON',
-        'Date',
       ],
     });
   }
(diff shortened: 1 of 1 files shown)


brad-decker (dismissed):

DDDDDanica (dismissed):

development/build/index.js
```diff
@@ -132,6 +132,8 @@ async function defineAndRunBuildTasks() {
         'stateHooks',
         'sentryHooks',
         'sentry',
+        'JSON',
```
Gudahtt, line 135: Do we know where these are being used? Are we sure it's related to Sentry?

Gudahtt: I'm guessing that they're used by some dependency not related to Sentry, in which case they'd be more appropriate to list under the "globals chrome driver needs to function (test env)" section

pedronfigueiredo: Not sure where these are being used, so I changed them to the other section as per your suggestion

Gudahtt (approved):
LGTM!

DDDDDanica (approved):

PR #18999, opened 2023-05-03
UX: Multichain: Account Details Fixes

### LINKED ISSUE #18992: Account details UI fixes
### Edit Account name

-Error state state should be shown with the border of the inout in red and the help text left aligned. 

Build

Design

### Password for Private Key 

-Password is not hidden when typing it.
-Ideally we would have an input with a hide/unhide icon, but in the mean time, we can default to hiding the password like we do currently. 
-Label of the input would in medium weight and not bold  

Build

-When the password is incorrect, we show the error copy two times. We should show only the help text in the input. 

-When the password is incorrect and user goes back to the account details we show a "incorrect password" alert that is not related to that screen.

darkwing:
## Explanation

Fixes #18992 

The following Account Details issues were fixed:

1.  Rewrote the Editable label completely to show correct state and not duplicate errors
2. Fixed the font-weight of the password label
3. Removed the duplicate error message on the authentication page
4. Hide the bad password message on the first detail screen, autofocus the password field

## Pre-merge author checklist

- [ ] I've clearly explained:


vthomas13 (dismissed):

ui/components/ui/editable-label/editable-label.js
```diff
@@ -40,48 +49,44 @@ export default class EditableLabel extends Component {
...
-      </div>
+      </Box>
     );
   }
 
   renderReadonly() {
     return (
-      <div className={classnames('editable-label', this.props.className)}>
-        <div className="editable-label__value">{this.state.value}</div>
+      <Box display={DISPLAY.FLEX} alignItems={AlignItems.center} gap={3}>
```
NidhiKJha, line 80: Let's not remove the className since the EditableLabel has been used as a child component in AccountDetailsModal as well and className has been passed as a prop. 
```suggestion
      <Box className={classnames('editable-label', this.props.className)} display={DISPLAY.FLEX} gap={3}>
```

NidhiKJha, line 80:
```suggestion
      <Box className={classnames('editable-label', this.props.className)} display={DISPLAY.FLEX} gap={3}>
```

NidhiKJha (dismissed):
Let's not remove the className since the EditableLabel has been used as a child component in AccountDetailsModal as well and className has been passed as a prop.

ui/components/ui/editable-label/editable-label.js
darkwing: Good shout!

vthomas13 (approved):

NidhiKJha (approved):

PR #19014, opened 2023-05-04
Update CODEOWNERS: remove requirement for supply-chain team to approv…

danjm:
…e lavamoat policy changes

## Explanation

Now that the supply-chain team on github has write access to the metamask-extension repo, all PRs that change lavamoat policy files require approval from one of you four (or leotm). This access was given as a result of this thread, https://consensys.slack.com/archives/GTQAGKY5V/p1683200894794789, but I think the requirement to approve all such PRs was not intended.

That requirement is occuring because of this line in the CODEOWNERS file https://github.com/MetaMask/metamask-extension/blob/develop/.github/CODEOWNERS#L12

Prior to the supply-chain team being granted write access, that line in the CODEOWNERS file was ignored, but now it is coming into effect. We might want to change the CODEOWNERS file going forward, but for now we will temporarily remove this to unblock the PR and release pipelines.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -11,6 +11,7 @@
 **/snaps/**                          @MetaMask/snaps-devs
 **/flask/**                          @MetaMask/snaps-devs
 development/                         @MetaMask/extension-devs @kumavis
+lavamoat/                            @MetaMask/extension-devs @MetaMask/supply-chain @MetaMask/snaps-devs
 
 # The .circleci/ folder instructs Circle CI on the process by which it
 # should test, build and publish releases of our application. Due to the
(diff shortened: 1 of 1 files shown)


brad-decker (dismissed):

DDDDDanica (dismissed):

kumavis:
>  you may want to keep the LM line but change the group. Otherwise you'll have the same problem that we started with where snaps team has to review policy changes

rekmarks (dismissed):

kumavis (dismissed):

kumavis (approved):

brad-decker (approved):

PR #1197, opened 2017-03-08
Allow ENS Names in Send Form

danfinlay:
I know we're going to deprecate the send form hopefully soon, but the plumbing will still be useful.

I've modified our normal `input` component to be an `ens-input` component, which is aware when its contents match the `*.ens` format, and indicates if the name resolves to anything on the current network.

If it does, it gives a little icon that you can click to copy.  Maybe it should be a link to a blockchain explorer, I'm open to thoughts from @VladTod and @eshon.

Here's what I've got so far:
ens-input

To create this, I made my own ENS resolving modules, published publicly on npm:
- [eth-ens-namehash](https://www.npmjs.com/package/eth-ens-namehash)
- [ethjs-ens](https://www.npmjs.com/package/ethjs-ens)

Fixes #1130

PR #19068, opened 2023-05-09
Set network status to "unknown" when ID is invalid

Gudahtt:
## Explanation

We now set the network status to "unknown" rather than "unavailable" when the network ID is invalid. This better reflects what we know when this happens, and it makes the network controller better aligned with the core network controller.

This was accomplished by using a regular error for the network ID assertion rather than using `assert` directly. `assert` would throw an error with a `code` property, which resutled in us treating it like an RPC error.

This isn't tested, but it was found in the course of porting unit tests from core to extension. It will be covered by these tests, which will be added in the next PR.

This change should have no functional impact because we treat these two network statuses as equivalent. The distinction between unknown and unavailable is useful only for debugging.

This relates to [#1197](https://github.com/MetaMask/core/issues/1197)

## Manual Testing Steps

No functional impact

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/app/scripts/controllers/network/network-controller.ts
+++ b/app/scripts/controllers/network/network-controller.ts
@@ -307,10 +307,9 @@ function isErrorWithCode(error: unknown): error is { code: string | number } {
  * @param value - The value to check.
  */
 function assertNetworkId(value: any): asserts value is NetworkId {
-  assert(
-    /^\d+$/u.test(value) && !Number.isNaN(Number(value)),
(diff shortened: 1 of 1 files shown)


Gudahtt:
Tested here: https://github.com/MetaMask/metamask-extension/pull/19070

legobeat (approved):
SGTM!

brad-decker (approved):

PR #19076, opened 2023-05-09
Stop failing pull request creation due to branch validation in .githu/scripts/label-prs.ts

danjm:
The changes in #17603 included code that validated branch names according to conventions introduced by that PR. PR creation fails if branch names are invalid. Included in that is code the prevents PRs from being created if there are no associated issues.

While we may want to enforce branch name validation in the future, we should wait until we have an agreed set of conventions. Also, we probably don't want to require that all PRs have an associated issue, so we might want to reconsider that aspect of validation as well.

For now, this PR comments out the code that would cause PR creation to fail if branch names are "invalid".

### DIFF
--- a/.github/scripts/label-prs.ts
+++ b/.github/scripts/label-prs.ts
@@ -21,12 +21,15 @@ async function main(): Promise<void> {
 
   let issueNumber = await getIssueNumberFromPullRequestBody();
   if (issueNumber === "") {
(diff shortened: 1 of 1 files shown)


danjm:
I was wrong in my PR description above, in the sense that the validation is an opt-in pre-commit hook. This change is not needed.

PR #19102, opened 2023-05-11
Ensure the GasDetailsItem component can handle a tx with a maxPriorityFee of 0

### LINKED ISSUE #19091: [Bug]: Cannot read properties of undefined (reading 'maxPriorityFeePerGas')
### Describe the bug

Screenshot from 2023-05-10 19-44-15

### Steps to reproduce

1. Visit a sites such as uniswap or camelot.exchange or swapfish.fi swap pages
2. connect the wallet
3. select a coin to swap (that requires spending approval)
4. click approve which opens metamask to set the spending cap
5. click approve in metamask
6. the error then appears instead of going to the confirmation/gas fee dialogue

```shell
Error messages is displayed in the attached screenshot to this bug report
```

### Version

10.30.1

### Build type

None

### Browser

Chrome

### Operating system

Linux

### Hardware wallet

_No response_

### Additional context

I downgraded metamask to version 10.29 and the error no longer occurs.

danjm:
Fixes https://github.com/MetaMask/metamask-extension/issues/19091

## Explanation

The `GasDetailsItem` component had this code prior to this PR:

```
            maxPriorityFeePerGas ||
            hexWEIToDecGWEI(transactionData.txParams.maxPriorityFeePerGas)
```

This causes an error when `maxPriorityFeePerGas` is `0` and `transactionData.txParams` is `undefined`. `transactionData.txParams` will only be defined if we are in the send flow, and so `getCurrentDraftTransaction` returns a transaction. So the error would occur if we are on the confirm screen and `maxPriorityFeePerGas` is `0`

This situation can happen in two ways:
1. A dapp creates a transaction and explicitly sets the `maxPriorityFeePerGas` to 0
2. A dapp creates a transaction and doesn't set a `maxPriorityFeePerGas`, and this occurs on a network where [our transaction controller defaults the `maxPriorityFeePerGas` to 0](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/controllers/transactions/index.js#L1041)

Defaulting the `maxPriorityFeePerGas` to zero will only happen on certain networks, arbitrum being the currently known example. (fyi, this bug was first reported by 5 different users on arbitrum)

The `maxPriorityFeePerGas` being zero would cause  `maxPriorityFeePerGas || hexWEIToDecGWEI(transactionData.txParams.maxPriorityFeePerGas)` to resolve to `hexWEIToDecGWEI(transactionData.txParams.maxPriorityFeePerGas)` because in `GasDetailsItem` if `maxPriorityFeePerGas` is 0, it will be a `Number` and not a string. This is ultimately because of the logic and handling of types in `useMaxPriorityFeePerGasInput`

This PR fixes the present bug by properly handling the possible types within the `GasDetailsItem` component. A unit test is also added that would have failed prior to this PR.

Meanwhile, the handling of types in `useMaxPriorityFeePerGasInput`, in particular when dealing with `0` or equivalent values, should be revisited and very likely cleaned up in some way. However, I leave that work to the future: https://github.com/MetaMask/metamask-extension/issues/19103

### Before

https://github.com/MetaMask/metamask-extension/assets/7499938/1c2fa068-e28e-4f0d-87c8-4cbc410fbdcb

### After

https://github.com/MetaMask/metamask-extension/assets/7499938/51c6b961-016c-48d2-92e1-2ce9718b7d76

## Manual Testing Steps

1. Switch to the Arbitrum network
2. Go to https://metamask.github.io/test-dapp/
3. Scroll down to the form at the bottom
4. Enter an ERC20 token contract address in the `to` field, e.g. `0xFd086bC7CD5C481DCC9C85ebE478A1C0b69FCbb9` (which is USDT on the arbitrum chain)
5. Select a type 2 transaction
6. enter a value for "Max Fee" but not for "Max Priority Fee"
7. In the "Data" field, enter the data necessary for an ERC20 token `approve` call. e.g. `0x095ea7b3000000000000000000000000000000000022d473030f116ddee9f6b43ac78ba3ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff`
8. When the confirmation popup loads, enter an approval amount and click "Next".
9. The second step of the token approval flow should render without error.

You should also be able to repeat the same steps as above, but on another network and in step 6, explictly set "Max Priority Fee" to `0`

Both of the above scenarios should fail on `develop` or `master` (with an error at step 9), but succeed on this branch

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/ui/components/app/gas-details-item/gas-details-item.js
+++ b/ui/components/app/gas-details-item/gas-details-item.js
@@ -52,12 +52,12 @@ const GasDetailsItem = ({ userAcknowledgedGasMissing = false }) => {
 
   const maxPriorityFeePerGasToRender = (
     maxPriorityFeePerGas ??
-    hexWEIToDecGWEI(transactionData?.txParams?.maxPriorityFeePerGas ?? '0x0')
+    hexWEIToDecGWEI(transactionData.txParams?.maxPriorityFeePerGas ?? '0x0')
   ).toString();
 
   const maxFeePerGasToRender = (
     maxFeePerGas ??
-    hexWEIToDecGWEI(transactionData?.txParams?.maxFeePerGas ?? '0x0')
+    hexWEIToDecGWEI(transactionData.txParams?.maxFeePerGas ?? '0x0')
   ).toString();
 
   return (
--- a/ui/components/app/gas-details-item/gas-details-item.js
+++ b/ui/components/app/gas-details-item/gas-details-item.js
@@ -51,17 +51,13 @@ const GasDetailsItem = ({ userAcknowledgedGasMissing = false }) => {
   }
 
   const maxPriorityFeePerGasToRender = (
-    maxPriorityFeePerGas === undefined
-      ? hexWEIToDecGWEI(
-          transactionData?.txParams?.maxPriorityFeePerGas ?? '0x0',
-        )
-      : maxPriorityFeePerGas
+    maxPriorityFeePerGas ??
+    hexWEIToDecGWEI(transactionData?.txParams?.maxPriorityFeePerGas ?? '0x0')
   ).toString();
 
   const maxFeePerGasToRender = (
-    maxFeePerGas === undefined
-      ? hexWEIToDecGWEI(transactionData?.txParams?.maxFeePerGas ?? '0x0')
-      : maxFeePerGas
+    maxFeePerGas ??
+    hexWEIToDecGWEI(transactionData?.txParams?.maxFeePerGas ?? '0x0')
   ).toString();
 
   return (
--- a/ui/components/app/gas-details-item/gas-details-item.js
+++ b/ui/components/app/gas-details-item/gas-details-item.js
@@ -50,6 +50,20 @@ const GasDetailsItem = ({ userAcknowledgedGasMissing = false }) => {
     return null;
   }
 
+  const maxPriorityFeePerGasToRender = (
+    maxPriorityFeePerGas === undefined
+      ? hexWEIToDecGWEI(
+          transactionData?.txParams?.maxPriorityFeePerGas ?? '0x0',
+        )
+      : maxPriorityFeePerGas
+  ).toString();
+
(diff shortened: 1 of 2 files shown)


ui/components/app/gas-details-item/gas-details-item.js
```diff
@@ -50,6 +50,16 @@ const GasDetailsItem = ({ userAcknowledgedGasMissing = false }) => {
...
 
+  const maxPriorityFeePerGasToRender = (
+    maxPriorityFeePerGas ??
+    hexWEIToDecGWEI(transactionData?.txParams?.maxPriorityFeePerGas ?? '0x0')
+  ).toString();
+
+  const maxFeePerGasToRender = (
+    maxFeePerGas ??
+    hexWEIToDecGWEI(transactionData?.txParams?.maxFeePerGas ?? '0x0')
+  ).toString();
```
Gudahtt, line 61: Nit: The `?` after `transactionData` seems unnecessary; it's guaranteed to at least be an empty object (see the `useDraftTransactionGasValues` hook)


```suggestion
  const maxPriorityFeePerGasToRender = (
    maxPriorityFeePerGas ??
    hexWEIToDecGWEI(transactionData.txParams?.maxPriorityFeePerGas ?? '0x0')
  ).toString();

  const maxFeePerGasToRender = (
    maxFeePerGas ??
    hexWEIToDecGWEI(transactionData.txParams?.maxFeePerGas ?? '0x0')
  ).toString();
```

Gudahtt (approved):
LGTM!

digiwand (approved):
lgtm! 

looking forward to updating the `useMaxPriorityFeePerGasInput` handling some day so we can handle one type rather than both string and numbers

legobeat (approved):
lgtm!
though a default value of `0` (as opposed to passing a default `0x0` and then parsing, as here) would be nicer, don't let that be a blocker.

pedronfigueiredo (approved):

PR #19105, opened 2023-05-11
Update useDraftTransactionGasValues

digiwand:
## Explanation

- rename to useDraftTransactionWithTxParams
- return transactionData directly instead of wrapped in an object 

no visible ux/ui changes


legobeat (approved):
All for the name change.
I don't understand the motivation behind changing the return type as the old version seems more extensible.
I assume this facilitates some other change and I don't see why not, though.

jpuri (approved):

digiwand:
@legobeat I'm not sure why the older version would be more extendable. I thought the transactionData object would handle extensibility unless I'm missing something (e.g. `transactionData.aNewProperty`).

NiranjanaBinoy (approved):

digiwand:
@legobeat 

makes sense. Thanks for double-checking too! I checked with the original author before merging and this change is good to go

PR #19169, opened 2023-05-16
Swaps UI update

dan437:
## Explanation
We've redesigned our Swaps feature to be easier and faster to use.

There are still a few todos (see below), but we can start getting feedback on code and functionality now to move faster. Most of existing business logic was reused from current Swaps. We keep the old design as a fallback solution, which is controlled via a feature flag. Once we verify that the new redesign is performing well, we will remove the old design from the codebase to make it cleaner.

TODOs:
- Do more QA testing

## Screenshots
image

## Manual Testing Steps
- Click on the "Swap" button on the main page of MetaMask
- Do our regular Swaps testing, but with the new redesign (different tokens, networks, transaction settings, etc.)


ui/pages/swaps/prepare-swap-page/prepare-swap-page.js
```diff
@@ -0,0 +1,1084 @@
+import React, { useContext, useEffect, useState, useCallback } from 'react';
```
dan437, line 1: Most of the logic here is reused from the Build Quote page: https://github.com/MetaMask/metamask-extension/blob/develop/ui/pages/swaps/build-quote/build-quote.js

ui/pages/swaps/prepare-swap-page/review-quote.js
```diff
@@ -0,0 +1,1309 @@
+import React, {
```
dan437, line 1: Most of the logic here is reused from the View Quote page: https://github.com/MetaMask/metamask-extension/blob/develop/ui/pages/swaps/view-quote/view-quote.js

ui/pages/swaps/selected-token/selected-token.js
```diff
@@ -0,0 +1,69 @@
+import React, { useContext } from 'react';
```
dan437, line 1: Most of the code here is reused from this component: https://github.com/MetaMask/metamask-extension/blob/develop/ui/pages/swaps/dropdown-search-list/dropdown-search-list.js

ui/pages/swaps/transaction-settings/transaction-settings.js
```diff
@@ -0,0 +1,332 @@
+import React, { useState, useEffect, useContext } from 'react';
```
dan437, line 1: Most of the code here is reused from the SlippageButtons component: https://github.com/MetaMask/metamask-extension/blob/develop/ui/pages/swaps/slippage-buttons/slippage-buttons.js

darkwing:
Awesome work @dan437 !  I have a few comments:

## 1.  Adding tokens simply because we requested a swap quote
I think this is current behavior but adding a token to the user's token list simply because they requested a quote seems like unnecessary clutter
SCR-20230519-jnmf


## 2. Swap details spacing
These controls and text all look really scrunched together -- could use more standardized spacing, maybe from design system?
SCR-20230519-jmlc

## 3.  Settings spacing
These controls and text all look really scrunched together -- could use more standardized spacing, maybe from design system?
SCR-20230519-jmdk

## 4.  Error off the map
I'm seeing this in other places, especially onboarding, so no idea why this is stretching:
SCR-20230519-jizq

## 5.  Max
I feel like we should disable the "MAX" button if the value is `0`
SCR-20230519-jgst

## 6.  Accessibility
I don't think I could get to these dropdowns via keyboard
SCR-20230519-jetr

ui/pages/swaps/exchange-rate-display/index.scss
```diff
@@ -16,34 +16,8 @@
     font-weight: bold;
   }
 
-  &__switch-arrows {
+  &__quote-rate {
+    color: var(--color-primary-default);
```
darkwing, line 20: Can this be added as a `color` property to `Box`?

ui/pages/swaps/index.scss
```diff
@@ -123,6 +137,14 @@
     flex: 1;
   }
 
+  &__header-back {
+    @include H7;
+
+    color: var(--color-primary-default);
+    cursor: pointer;
+    padding-left: 24px;
```
darkwing, line 145: Can `color` and `padding-left` be added as box props?

app/scripts/controllers/swaps.js
```diff
@@ -104,23 +104,34 @@ const initialState = {
...
+      getEIP1559GasFeeEstimates,
+      onNetworkStateChange,
+    },
+    state,
+  ) {
     this.store = new ObservableStore({
-      swapsState: { ...initialState.swapsState },
+      swapsState: {
+        ...initialState.swapsState,
+        swapsFeatureFlags: state?.swapsState?.swapsFeatureFlags || {},
```
danjm, line 124: Why was the `swapsFeatureFlags` property explicitly added here? This would always initialize the feature flag state to what it was during the last user session. Is that intended?

ui/pages/swaps/prepare-swap-page/review-quote.js
```diff
@@ -0,0 +1,1307 @@
...
+  }
+  let gasTotalInWeiHex = calcGasTotal(maxGasLimit, maxFeePerGas || gasPrice);
+  if (multiLayerL1FeeTotal !== null) {
+    gasTotalInWeiHex = sumHexes(
+      gasTotalInWeiHex || '0x0',
+      multiLayerL1FeeTotal || '0x0',
+    );
+  }
+
+  const { tokensWithBalances } = useTokenTracker(swapsTokens, true);
```
danjm, line 315: Did you check the edge case we mentioned on slack, where the user selects a token to swap that they have not yet added to their token list? The possible failure, given that this PR removes the `destinationTokenAddedForSwap` and associated logic, is that token balances are not calculated correctly on this page for destination tokens that are not in the users list.

app/scripts/controllers/swaps.js
dan437: Glad you asked! The reason for this is the following: if we would turn off the feature flag for this new UI redesign (in case there is a blocking issue), we would anyway [show the new redesign by default](https://github.com/MetaMask/metamask-extension/pull/19169/files#diff-2493772b1f68622e8d2eae0660b50d2765884331c60418c86e5315389fb2fd8eR345) before redirecting to the previous design once we get a response from the `/featureFlags` endpoint, which is not user-friendly. This way we will remember which feature flags were used last time and use those by default, before we get fresh ones from the `/featureFlags` endpoint.

nikoferro:
Hey @dan437, can you debounce fetching the code when the user is typing? Right now it seems it makes the requests to get the quotes just as soon as the user starts typing which feels sluggish and with some lag if the user keeps typing after a couple of characters.

Also in the case the user keeps adding numbers, are you cancelling previous requests or how is this handled? Just wondering about an scenario where the user inputs 1 and then 0 (trying to swap 10), and for some reason the request to get a quote for 1, is slower than the request for quotes for 10. So you get a quote to swap 1 instead of 10

nikoferro:
Hey @dan437 , does the new redesign consider this scenario? Right now swaps on full screen looks a bit weird since the layout goes to the bigger size but swaps remains the same.

Screenshot 2023-06-13 at 15 39 35

dan437:
Hi @nikoferro, thanks a lot for this valuable feedback. We briefly discussed it internally and here are my responses:
- Debouncing: we have it in the prepare-swap-page.js component: https://github.com/MetaMask/metamask-extension/pull/19169/files#diff-8b6b5a47eeec0494313f7bfc546bce1f3e0fbb16bb56d1092884e376ce394095R628 It's the same debouncing that's already in production. I also noticed that the extension is a bit slow when doing `yarn start`, but if I use a build created via `yarn dist`, the experience is very smooth, including the Amount field in Swaps.
- Expanded view: the long nav bar was added recently. The expanded view's width for Swaps is similar to the Send feature, but I will discuss with our designers how we can make it better with the new nav bar.
- ETH -> ETH swap: I wasn't able to reproduce ETH -> ETH, since the Token To field should not show a token that's in the Token From field. If you are able to reproduce it, please, let me know.

ui/components/component-library/text-field/text-field.js
```diff
@@ -38,6 +38,7 @@ export const TextField = ({
   readOnly,
   required,
   size = Size.MD,
+  testId,
```
darkwing, line 41: Should this be defaulted to `''` since it's not required @georgewrmarshall ?

ui/components/ui/typography/typography.js
```diff
@@ -80,6 +80,7 @@ export default function Typography({
   marginLeft,
   boxProps = {},
   className,
+  testId,
```
darkwing, line 83: Same here @georgewrmarshall .

ui/ducks/swaps/swaps.js
```diff
@@ -208,6 +209,9 @@ const slice = createSlice({
     setSwapsSTXSubmitLoading: (state, action) => {
       state.swapsSTXLoading = action.payload || false;
     },
+    setTransactionSettingsOpened: (state, action) => {
+      state.transactionSettingsOpened = action.payload || false;
```
darkwing, line 213: Maybe `Boolean(payload)` instead?

ui/pages/swaps/list-with-search/list-with-search.js
```diff
@@ -0,0 +1,180 @@
...
+        await handleSearchTokenForImport(trimmedNewSearchQuery);
+        return;
+      }
+      setItems(results);
+    }, 350);
+  };
+
+  useEffect(() => {
+    handleSearch(searchQuery);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
darkwing, line 108: Is this eslint line necessary?

ui/pages/swaps/prepare-swap-page/view-quote-price-difference.js
```diff
@@ -0,0 +1,107 @@
...
+  const t = useContext(I18nContext);
+
+  let priceDifferenceTitle = '';
+  let priceDifferenceMessage = '';
+  let priceDifferenceClass = '';
+  if (priceSlippageUnknownFiatValue) {
+    // A calculation error signals we cannot determine dollar value
+    priceDifferenceTitle = t('swapPriceUnavailableTitle');
+    priceDifferenceMessage = t('swapPriceUnavailableDescription');
+    priceDifferenceClass = GasRecommendations.high;
```
darkwing, line 43: Can these first values be assigned at the time of `let`, and then replaced with if `! priceSlippageUnknownFiatValue`.

ui/pages/swaps/searchable-item-list/item-list/item-list.component.js
```diff
@@ -107,7 +108,10 @@ export default function ItemList({
               <div className="searchable-item-list__labels">
                 <div className="searchable-item-list__item-labels">
                   {primaryLabel ? (
-                    <span className="searchable-item-list__primary-label">
+                    <span
+                      className="searchable-item-list__primary-label"
+                      data-testid="searchable-item-list-primary-label"
+                    >
```
darkwing, line 114: Do we want use `Text` instead of `span` here @georgewrmarshall

darkwing:
Seeing this weird state where I can submit a swap that I don't have a balance for, for roughly 1 second before I'm told I don't have enough ETH to pay gas fee:


https://github.com/MetaMask/metamask-extension/assets/46655/8ceffa0e-a169-4414-bb08-e77c1d3d9803

darkwing:
Can we keep the token list within the bounds of the popup viewport?

SCR-20230614-jrgs

ui/ducks/swaps/swaps.js
dan437: Elegant!

ui/pages/swaps/list-with-search/list-with-search.js
dan437: It only silences a warning, but code works without it as well, so I've removed it.

When I pass the `handleSearch` dependency to remove the warning, it triggers this effect many times. That's why I didn't add it into the dependency list. Maybe there is a better way where the `handleSearch` fn doesn't trigger this effect.

ui/pages/swaps/searchable-item-list/item-list/item-list.component.js
dan437: This PR only adds `data-testid` attributes in this existing file. We have a user story to refactor a few remaining components in Swaps to use the new DS components (we use them in most of the redesigned Swaps already), so I would prefer do it in another PR.

dan437:
> Can we keep the token list within the bounds of the popup viewport?
> 
> SCR-20230614-jrgs

Yes, I discussed it with our designers a few weeks ago and there is a [ticket](https://github.com/MetaMask/metamask-extension/issues/19190) for adding a new design system component called `ModalBody`, which should allow scrolling only for content, instead of having scrolling for the whole page as we have it now. Before the new `ModalBody` component is available, I will do that via custom CSS.

ui/components/app/confirm-gas-display/confirm-legacy-gas-display/__snapshots__/confirm-legacy-gas-display.test.js.snap
```diff
@@ -42,6 +44,7 @@ exports[`ConfirmLegacyGasDisplay should match snapshot 1`] = `
       >
         <h6
           class="box box--margin-top-1 box--margin-bottom-1 box--flex-direction-row typography typography--h6 typography--weight-normal typography--style-normal typography--color-text-alternative"
+          data-testid=""
```
danjm, line 47: @dan437 Why are these empty `data-testid=""` here?

dan437: I've added an empty string as a default value based on this: https://github.com/MetaMask/metamask-extension/pull/19169#discussion_r1229766775

However, it seems it would be better to keep it as I had it initially.

davibroc (approved):
I have completed the tests for this and found no issues

darkwing (approved):

danjm (approved):

PR #19172, opened 2023-05-16
Bump dependencies of network-controller

### LINKED ISSUE #19099: Update core packages in support of introducing `@metamask/network-controller`
This ticket is a prerequisite of #17931. In other words, before replacing the NetworkController that exists in the extension with NetworkController from the `@metamask/network-controller` package, we will need to first upgrade other packages that the extension uses that `@metamask/network-controller` also depends on.

The list of packages includes:

- [ ] `@metamask/assets-controllers`. This is currently at ^6.0.0, but needs to be at ^7.0.0 because it depends on `@metamask/network-controller` directly.
- [ ] `@metamask/gas-fee-controller`. This is currently at ^4.0.1, but needs to be at ^5.0.0 because it depends on `@metamask/network-controller` directly. 

There doesn't seem to be any breaking changes of note for either of these packages (only the bump for `@metamask/network-controller`, but that's perfectly fine as it matches what we are doing).

As a side note, it would be great to get a new release of `@metamask/smart-transactions-controller` installed. Currently it relies on `@metamask/network-controller` v2.0.0, but it's been updated on `main`, so if we make a new release we can deduplicate `@metamask/network-controller` and save a bit on the bundle size.

mcmire:
## Explanation

In preparation for introducing `@metamask/network-controller`, we want to ensure that any packages that depend on this package are up to date so that we are using the same version of `@metamask/network-controller` in the dependency tree.

The extension uses two packages that depend on
`@metamask/network-controller`: `@metamask/assets-controllers` and `@metamask/gas-fee-controller`. This commit upgrades them to the latest version. There are no breaking changes in the new release of these packages so there are no accompanying compatibility changes.

Fixes #19099.

## Manual Testing Steps

This is an internal change and should have no user-facing impact. That said, it could affect the list of NFTs/tokens and it could affect gas estimates. So it might be good to double-check that you can add a new token (or that the token autodetection works) and that if you create a transaction you see the gas estimates.

## Pre-merge author checklist

- [x] I've clearly explained:

- [x] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone


mcmire:
I think the Socket errors above may false negatives. `ansi-colors` is not a dependency of either `@metamask/assets-controllers` or `@metamask/gas-fee-controller`.

yarn.lock
```diff
@@ -34646,6 +34594,35 @@ __metadata:
   languageName: node
   linkType: hard
 
+"web3-provider-engine@npm:^16.0.5":
```
Gudahtt, line 34597: We can get rid of this by running `yarn dedupe`

mcmire:
Rebased with `develop` and resolved conflicts.

Gudahtt (approved):
LGTM!

pedronfigueiredo (approved):

pedronfigueiredo (approved):

PR #18903, opened 2023-05-01
Release MultiChain 0.5

darkwing:
SCR-20230525-jgwm

## Explanation

This pull request represents the release of the Multichain 0.5 design work.  This includes:

* A completely new application header, which contains:
   1.  A new network picker, which displays as only an avatar in the popup and as a full dropdown in full screen mode
   2. A new account picker
   3. A new connected icon which displays in popup mode
   4. A new global menu which contains controls that were formally in the account menu as well as account options menu
* A new token list
* A new token details popover

This PR is meant to achieve the following:

* Removes the `process.env.MULTICHAIN` flag and usage, exposing the new UI
* Removes legacy UI components, like the old network picker and account dropdown, that are no longer in use.
* Updates dozens of E2Es, since this core functionality and navigation paths have changed
* Removes legacy localization strings.

This PR is obviously large.  I've done my best to make each commit and its message meaningful so the reviewer can review by commit if they so choose.

## Pre-merge author checklist

- [ ] I've clearly explained:
  - [ ] What problem this PR is solving
  - [ ] How this problem was solved
  - [ ] How reviewers can test my changes
- [ ] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [ ] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
- [ ] PR is linked to the appropriate GitHub issue
- [ ] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.

In this case, a QA Engineer approval will be be required.

PR #19180, opened 2023-05-17
Chromedriver v113

seaona:
## Explanation
Update Chromedriver and Chrome to last stable version for e2e tests.

### Before

```
"chromedriver": "^111.0.0"
CHROME_VERSION='111.0.5563.64-1'
```

### After

```
"chromedriver": "^113.0.0"
CHROME_VERSION='113.0.5672.126-1'
```

## Manual Testing Steps
Verify that e2e tests are run successful on ci or locally

## Pre-merge author checklist

- [X] I've clearly explained:


legobeat (approved):
Looking good after some minor patching up - just chance or is this actually less flakey wrt e2e timeouts?

Thanks for opening @seaona

jpuri (approved):

PeterYinusa (approved):

test/e2e/benchmark.js
```diff
@@ -79,12 +79,17 @@ async function profilePageLoad(pages, numSamples, retries) {
     if (runResults.some((result) => result.navigation.lenth > 1)) {
       throw new Error(`Multiple navigations not supported`);
     } else if (
-      runResults.some((result) => result.navigation[0].type !== 'navigate')
+      runResults.some(
+        (result) =>
+          result.navigation[0] && result.navigation[0].type !== 'navigate',
```
Gudahtt, line 84: Hmm, this shouldn't be empty. This suggests that this benchmark is broken.

development/build/index.js
```diff
@@ -127,6 +127,8 @@ async function defineAndRunBuildTasks() {
         'Promise',
         'JSON',
         'Date',
+        'Proxy',
```
Gudahtt, line 130: This is a notable change that should be called out in the PR description at the very least, since it expands the available global APIs

Gudahtt (approved):
LGTM! 

I left one comment about additional detail needed for the PR description, and the benchmark changes deserve further investigation, but these changes all look good.

test/e2e/benchmark.js
Gudahtt: Good point. I don't recall why those conditions were there.

Agreed that throwing a more explicit error would be helpful. If we don't have at least one navigation entry, it either means that the [1-second pause on this line](https://github.com/MetaMask/metamask-extension/blob/53acfe271d6ab70f93a7efaf302f6e7750db30f0/test/e2e/benchmark.js#L31) isn't long enough for the page to finish loading, or that something is stopping us from using the Navigation API (e.g. LavaMoat scuttling).

darkwing:
Getting this merged will be a huge help with https://github.com/MetaMask/metamask-extension/pull/18903 .  Thank you @seaona !

development/build/index.js
seaona: mmm I am not sure the implications of this change :thinking: let's see if @legobeat can add a bit more context, so we cna merge the PR. Thank you both :pray:

test/e2e/benchmark.js
seaona: that's a v good callout @Gudahtt  and thank you @legobeat for asking in Slack too. 
The results for this PR are mostly `NaN` which seems to be broken now. 

image

development/build/index.js
mcmire: Is it enough to add something like this to the PR description?

```
Update LavaMoat configuration to add `Proxy` to the list of globals
```

(I'm not sure how exactly to characterize this change either, I just know that it is necessary as I ran into the same issue when attempting to update Chromedriver.)

HowardBraham:
Replaced by #19648

PR #19197, opened 2023-05-18
Await approval request in transaction controller

matthewwalsh0:
## Explanation

Refactor the transaction controller to decouple the public methods from the concept of confirmation by instead awaiting the approval request promise and then performing the approve or cancel logic, all in a single method execution.

At a technical level, this includes:

- Updating the `newUnapprovedTransaction` and `addUnapprovedTransaction` methods to await approval and continue with the post-approval logic.
- Changing the `approveTransaction` and `cancelTransaction` controller methods to private.
- Adding an optional `requiresApproval` property to `addUnapprovedTransaction` to skip approval.
- Updating the `updateAndApproveTx`, `cancelTx`, and `cancelTxs` actions to resolve or reject the approval using the `ApprovalController`.
- Updating the swaps reducer to add the transaction and skip approval in a single method call.
- Moving swaps specific transaction logic after creating the transaction but before automatic approval, into the transaction controller.

## Manual Testing Steps

Verify all transaction related flows including:

- Transfers
- Contract Operations
- Swaps

## Pre-merge author checklist

- [x] I've clearly explained:


Taniel13:
> ## Explanation
> 
> Refactor the transaction controller to decouple the public methods from the concept of confirmation by instead awaiting the approval request promise and then performing the approve or cancel logic, all in a single method execution.
> 
> At a technical level, this includes:
> 
> - Updating the `newUnapprovedTransaction` and `addUnapprovedTransaction` methods to await approval and continue with the post-approval logic.
> - Adding an optional `requiresApproval` property to `addUnapprovedTransaction` to skip approval.
> - Updating the `updateAndApproveTx`, `cancelTx`, and `cancelTxs` actions to resolve or reject the approval using the `ApprovalController`.
> - Updating the swaps reducer to add the transaction and skip approval in a single method call.
> - Moving swaps specific transaction logic after creating the transaction but before automatic approval, into the transaction controller.
> 
> ## Manual Testing Steps
> 
> Verify all transaction related flows including:
> 
> - Transfers
> - Contract Operations
> - Swaps
> 
> ## Pre-merge author checklist
> 
> - [x] I've clearly explained:
>   - [x] What problem this PR is solving
>   - [x] How this problem was solved
>   - [x] How reviewers can test my changes
> - [x] Sufficient automated test coverage has been added
> 
> ## Pre-merge reviewer checklist
> 
> - [ ] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
> - [ ] PR is linked to the appropriate GitHub issue
> - [ ] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone
> 
> If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.
> 
> In this case, a QA Engineer approval will be be required.
>

app/scripts/controllers/transactions/index.js
```diff
@@ -1835,6 +1656,252 @@ export default class TransactionController extends EventEmitter {
...
+        `TransactionController - invalid transactionType value: ${transactionType}`,
+      );
+    }
+
+    // If a transaction is found with the same actionId, do not create a new speed-up transaction.
+    if (actionId) {
+      let existingTxMeta =
+        this.txStateManager.getTransactionWithActionId(actionId);
+      if (existingTxMeta) {
+        this.emit('newUnapprovedTx', existingTxMeta);
```
bergarces, line 1691: Are we emitting this event only for the unit tests?

If so, we might want to think on how to refactor them, even if it's in a future PR.

app/scripts/controllers/transactions/index.js
```diff
@@ -1835,6 +1656,252 @@ export default class TransactionController extends EventEmitter {
...
+      const securityProviderResponse = await this.securityProviderRequest(
+        txMeta,
+        txMethodType,
+      );
+
+      txMeta.securityProviderResponse = securityProviderResponse;
+    }
+
+    this.addTransaction(txMeta);
+    this.emit('newUnapprovedTx', txMeta);
```
bergarces, line 1763: Same comment as the previous `newUnapprovedTx` event, but this one is even weirder as `addTransaction` emits a ``${txMeta.id}:unapproved`` with exactly the same data and I cannot see if anyone is listening.

Odd, although we might want to keep them just in case, if they don't seem necessary I'd put a comment for future maintainers.

app/scripts/controllers/transactions/index.js
```diff
@@ -1835,6 +1656,252 @@ export default class TransactionController extends EventEmitter {
...
+
+    txMeta = await this.addTransactionGasDefaults(txMeta);
+
```
bergarces, line 1766: Totally optional and subjective, but I think it'd be clearer for anyone else reviewing if this whole block of code that comes from `swaps.js` UI logic was moved into a private method, guarded here by just the transaction type. Something like:

```
if (transactionType === TransactionType.swap) {
  this._whateverNameGivenToTheSwapSpecificWork(...)
}
```

But maybe this comment is good enough for any reviewer to see where it comes from and why it's here.

app/scripts/controllers/transactions/index.js
```diff
@@ -798,99 +728,29 @@ export default class TransactionController extends EventEmitter {
...
-
-    // Add actionId to txMeta to check if same actionId is seen again
-    // IF request to create transaction with same actionId is submitted again, new transaction will not be added for it.
-    if (actionId) {
-      txMeta.actionId = actionId;
+      actionId,
+      options,
+    );
+    if (isExisting) {
+      return txMeta.status === TransactionStatus.unapproved
```
bergarces, line 743: From what I can see, `_waitForTransactionFinished` resolves when a `${id}:finished` event is fired.

However, that event is only fired when the status is changed to `submitted`, `rejected` or `failed`.

Since we are just awaiting if the initial status is `unapproved`, is there any chance that we could be missing out on awaiting for the tx from any of the other states (`approved`, `signed`, `dropped`, `confirmed`, `pending`). It might be the case that this is impossible, in which case we are good, but just mentioning it so that it's not overlooked.

app/scripts/controllers/transactions/index.js
```diff
@@ -2662,56 +2729,70 @@ export default class TransactionController extends EventEmitter {
...
+      result?.success();
     } catch (error) {
-      log.error('Failed to reject transaction approval request', error);
+      const transaction = this.txStateManager.getTransaction(txId);
+
+      if (transaction && transaction.status === TransactionStatus.unapproved) {
+        if (error.code === errorCodes.provider.userRejectedRequest) {
+          await this._cancelTransaction(txId, actionId);
+        } else {
+          await this._failTransaction(txId, error, actionId);
```
bergarces, line 2768: This one returns void.

app/scripts/controllers/transactions/index.js
```diff
@@ -1361,10 +1205,8 @@ export default class TransactionController extends EventEmitter {
    *
    * @param {number} txId - the tx's Id
    * @param {string} actionId - actionId passed from UI
-   * @param opts - options object
-   * @param opts.hasApprovalRequest - whether the transaction has an approval request
    */
-  async approveTransaction(txId, actionId, { hasApprovalRequest = true } = {}) {
+  async approveTransaction(txId, actionId) {
```
bergarces, line 1209: This is a weird one but... now that this is not called externally via `TransactionController` instance, but only via `PendingTransactionTracker` via passing the method in the constructor... should it be private?

app/scripts/controllers/transactions/index.js
```diff
@@ -356,42 +356,44 @@ export default class TransactionController extends EventEmitter {
...
+    const finishedPromise = isCompleted
+      ? Promise.resolve(initialTxMeta)
+      : this._waitForTransactionFinished(txId);
+
+    if (!isExisting && !isCompleted) {
+      try {
+        await this._requestTransactionApproval(initialTxMeta);
+      } catch (error) {
+        // Errors generated from final status using finished event
+      }
```
cryptotavares, line 371: This is a farfetched case, but if the `_requestTransactionApproval` ends up throwing after setting the transaction status to approved but before setting it to submitted, then we will end up with a pending Promise (`finishedPromise`) that will never resolve and will be blocking the execution flow. Should we or even do we want to handle transactions that were not submitted but were a approved? 🤔 

(I guess that this could already happen before anyway, so any solution should be done on a subsequent PR)

app/scripts/controllers/transactions/index.js
matthewwalsh0: Didn't really like this as I wrote it. I've removed the event and am instead using a spy on the state manager.

app/scripts/controllers/transactions/index.js
matthewwalsh0: I've left all the state manager events but removed `newUnapprovedTx` as it was now unused.

app/scripts/controllers/transactions/index.js
matthewwalsh0: Good idea, done.

app/scripts/controllers/transactions/index.js
matthewwalsh0: Done.

app/scripts/controllers/transactions/index.js
matthewwalsh0: I didn't notice `PendingTransactionTracker` was constructed inside this controller, so changed to private to encapsulate everything.

ui/store/actions.ts
```diff
@@ -1154,18 +1103,22 @@ export function addUnapprovedTransactionAndRouteToConfirmationPage(
  * @param method
  * @param txParams - the transaction parameters
  * @param type - The type of the transaction being added.
+ * @param options
+ * @param options.requireApproval
+ * @param options.extraMeta
```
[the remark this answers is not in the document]
matthewwalsh0: Updated.

app/scripts/controllers/transactions/index.js
matthewwalsh0: Good spot, I've fixed the condition inside the error handling in `_requestTransactionApproval` so it will now correctly cancel or fail the transaction if it's in any incomplete status. Meaning any errors here are just for information, and the transaction will always be `submitted`, `rejected`, or `failed` at this point.

danjm:
One issue I notice while reviewing this PR is we now have two different uses of the term "approval" within the transaction controller. This is not a huge deal, and it is a problem that existed before this PR, but the changes in this PR make it more noticeable.

To clarify, we have:
1. the concept of a "swapApproval", which is a transaction that calls the ERC20 `approve` method to grant the swaps contract the ablity to transfer tokens
2. the `ApprovalController` and the approvals it controllers

And so it can lead to some confusion for the reader of the code. For example we have code like this:
```
finalApproveTxMeta = await addUnapprovedTransaction(
          undefined,
          { ...approveTxParams, amount: '0x0' },
          TransactionType.swapApproval,
          {
            requireApproval: false,
```
Where `swapApproval` and `requireApproval` are right next to each other, but the `Approval`s in those two property names have entirely different meanings.

Now, this is not a huge deal, because when you read the transaction controller itself it is clear and well written. But a possible source of confusion. Including in `metamask-controller.js`, where we have the handling of the `newSwapApproval` event, but also lots of references to the `approvalController` and approvals that come from that controller, but no documentation to distinguish the `newSwapApproval` from the other sort of approvals.

We probably shouldn't address this problem in this PR, as it was a pre-existing problem. So I'd suggest a clean up task to follow this PR that renames the "swapApproval" to something else, perhaps "swapERC20Approve"

app/scripts/controllers/transactions/index.js
```diff
@@ -1646,136 +1395,519 @@ export default class TransactionController extends EventEmitter {
...
+    }
+
+    const swapsMeta = swapOptions?.meta;
+
+    if (!swapsMeta) {
+      return txMeta;
+    }
+
+    if (transactionType === TransactionType.swapApproval) {
+      this.emit('newSwapApproval', txMeta);
```
danjm, line 1725: What's the rationale for emitting an event here instead of just calling the swaps controller method here?

danjm: Oh, maybe because the transactions controller has to be instantiated before the swaps controller?

app/scripts/background.js
```diff
@@ -676,10 +676,8 @@ export function setupController(
   //
   // User Interface setup
   //
```
danjm, line 678: Why do we no longer need to wait for `initApprovals()` to call `updateBadge()`?

ui/store/actions.ts
```diff
@@ -1663,15 +1623,21 @@ export function cancelTxs(
...
           new Promise<void>((resolve, reject) => {
-            const actionId = generateActionId();
-            callBackgroundMethod('cancelTransaction', [id, actionId], (err) => {
-              if (err) {
-                reject(err);
-                return;
-              }
-
-              resolve();
-            });
```
danjm, line 1674: It took me a little while to figure out how the transaction state would correctly be updated now that we are not calling `cancelTransaction` directly. But I think I get it now:
1. `rejectPendingApproval` is called, with the second parameter being the error created by `ethErrors.provider.userRejectedRequest().serialize()`
2. `this.approvalController.reject` is called with an error constructed from its second parameter
3. Steps 1 and 2 will always be happening while we are awaiting `this._requestApproval` in `_requestTransactionApproval` the transaction controller
4. If the user rejects the request, `this._requestApproval` will throw an error
5. The error handler in `_requestTransactionApproval` will check that the error code matches what was set in step 2, and then call `this._cancelTransaction`

Does that sound right?

danjm: Also, I notice that we are passing `ethErrors.provider.userRejectedRequest().serialize()` here, but in `rejectPendingApproval` in `metamask-controller.js` we assume that the error parameter is not serialized (`new EthereumRpcError(error.code, error.message, error.data)`).

Does deserialization happen automatically by some part of our ui<>background communication code?

danjm (changes requested):
This looks good!

I have left a few questions and when they are answered I can approve.

app/scripts/controllers/transactions/index.js
matthewwalsh0: Exactly, a circular dependency. We would need the `SwapsController` to exist first to provide the callback in the constructor, but the `SwapsController` requires the `TransactionController` for callbacks in its constructor.

An event enables us to create the listener after both are already constructed.

app/scripts/background.js
matthewwalsh0: The previous implementation of `initApprovals` wouldn't resolve until the requests were approved, meaning the badge number would be zero at that point anyway.

Plus I noticed we already listened to the `updateBadge` event from the approval controller below anyway, meaning the badge is already reacting to the creation of the approval requests.

ui/store/actions.ts
matthewwalsh0: Exactly right.

The `serialize` method is a bit misleading as it's not serialising to a string, but rather a simple object with only the public properties remaining, including `code`.

https://github.com/MetaMask/rpc-errors/blob/v4.0.3/src/classes.ts#L44

danjm (dismissed):
Nice work!

OGPoyraz (approved):

vinistevam (approved):

bergarces (approved):

PR #19209, opened 2023-05-18
Import a token after a user receives it from a confirmed transaction

blurpesec:
## Explanation
@cloudonshore noticed that transactions that DApps create for users to sign on MetaMask that result in token transfers to users does not result in those tokens being automatically added to the user's app state.

This is especially problematic when users use DApps to swap for new tokens that they don't already have, as this will essentially result in the funds "disappearing" to the user until they add the asset.

This may result in many erroneous support requests about "incorrect" or "missing" assets or balances as well as a poor overall user experience.

This PR addresses this issue by creating a mechanism by which:
1) When pending transactions are confirmed, MetaMask will loop through the transaction receipt's transaction logs and identify all erc20 transfers
2) Filter the transfers by ones where the recipient of the transfer is the sender of the transaction (currently, this is assumed to be the user's account. note: this isn't necessarily true when paymasters are used - but i don't think MetaMask does this for now).
3) For each unique `log.address` (asset contract address), the app will save the token info in the same mechanism as "Import Token" while firing off an analytics event indicating the token was added (@ todo: this analytics event should probably be called something different?)

## Manual Testing Steps
- Find a DApp that allows the user to swap from one asset to another.
- Find a new asset that isn't already added to the your state.
- Create a transaction to exchange from some known asset to a new asset using a DApp with MetaMask as the transaction signer.
- Approve transaction on MetaMask.
- Wait for transaction to confirm.
- Check that the new asset is added to the user's assets list.

## Pre-merge author checklist

- [x] I've clearly explained:


blurpesec:

cloudonshore (dismissed):

shared/lib/transactions-controller-utils.js
```diff
@@ -35,6 +35,52 @@ export function calcTokenAmount(value, decimals) {
...
+  return `0x${addressLogTopic.replace(/^0x000000000000000000000000/u, '')}`;
+}
+
+/**
+ * Extracts token transfers from a txReceipt by filtering through logs.
+ *
+ * @param txReceipt - A transaction receipt.
+ * @returns {{ contractAddress: string, to: string, from: string }} A list of token transfers from a txReceipt.
+ */
+export function getTokenTransfersFromTxReceipt(txReceipt) {
```
bergeron, line 48: I want to make sure this only returns ERC-20 tokens and not ERC-721 NFTs.  There are some existing issues where NFT contracts mistakenly end up in the token state instead of the NFT state.

Since both ERC-20 and ERC-721 transfer events hash to `0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef` I wonder if this also needs as check for `topics.length === 3` to distinguish them.

app/scripts/controllers/transactions/pending-tx-tracker.js
```diff
@@ -223,6 +241,57 @@ export default class PendingTransactionTracker extends EventEmitter {
...
+
+        const receivedTokens = tokens
+          .filter((token) => !isEmpty(token))
+          .reduce((acc, token) => {
+            acc[token.contractAddress.toLowerCase()] = {
+              ...token,
+              image: formatIconUrlWithProxy(txMeta.chainId, token.address),
+            };
+            return acc;
+          }, {});
```
salimtb, line 274: there is no need to use filter then reduce , you can just add if statement inside your reduce callback funcion.
e.g:
`const receivedTokens = tokens.reduce((acc, token) => {
  if (!isEmpty(token)) {
    const { contractAddress, ...restToken } = token;
    acc[contractAddress.toLowerCase()] = {
      ...restToken,
      image: formatIconUrlWithProxy(txMeta.chainId, token.address),
    };
  }
  return acc;
}, {});`

shared/lib/transactions-controller-utils.js
```diff
@@ -35,6 +35,52 @@ export function calcTokenAmount(value, decimals) {
...
+  const txReceiptLogs = txReceipt?.logs;
+  if (!txReceiptLogs || txReceipt?.status === '0x0') {
+    return [];
+  }
+  return txReceiptLogs
+    .map((txReceiptLog) => {
+      if (!txReceiptLog.topics) {
+        return undefined;
+      }
```
salimtb, line 57: you can use a concise version and keep it on one line :
`if (!txReceiptLog.topics) return undefined`

shared/lib/transactions-controller-utils.js
```diff
@@ -35,6 +35,52 @@ export function calcTokenAmount(value, decimals) {
...
+
+      return {
+        to: stripAddressFromLogTopic(transferToAddressTopic),
+        from: stripAddressFromLogTopic(transferFromAddressTopic),
+        contractAddress: txReceiptLog.address,
+      };
+    })
+    .filter((transfer) => Boolean(transfer));
+}
```
salimtb, line 82: same here , you can use only reduce function with if statement instead of map then filter

PR #19230, opened 2023-05-19
Show the right "balance needed" value if we fallback from STX to regular Swaps

dan437:
## Explanation
Only set `ethBalanceNeededStx` if it's a smart transaction, refactoring.

### DIFF
--- a/ui/pages/swaps/view-quote/view-quote.js
+++ b/ui/pages/swaps/view-quote/view-quote.js
@@ -201,6 +201,8 @@ export default function ViewQuote() {
   );
   const swapsNetworkConfig = useSelector(getSwapsNetworkConfig, shallowEqual);
   const unsignedTransaction = usedQuote.trade;
+  const isSmartTransaction =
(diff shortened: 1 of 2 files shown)


adonesky1 (dismissed):
LGTM!

ui/pages/swaps/view-quote/view-quote.js
```diff
@@ -460,7 +462,7 @@ export default function ViewQuote() {
     : null;
 
   let ethBalanceNeededStx;
-  if (smartTransactionsError?.balanceNeededWei) {
+  if (isSmartTransaction && smartTransactionsError?.balanceNeededWei) {
```
mcmire, line 465: I'm not intimately familiar with this logic — is the issue that when `isSmartTransaction` flips from `true` to `false`, `smartTransactionsError` could still have something in it? If so, then would it make sense to reset `smartTransactionsError`?

adonesky1 (approved):

darkwing (approved):

PR #19296, opened 2023-05-25
Ensure custom provider configs have chain ID and RPC URL

mcmire:
## Explanation

Update NetworkController so that when it is creating the network client based on configuration for a custom RPC endpoint, it verifies that the configuration object contains both a chain ID and RPC URL. This check is already present on the core side; bringing it over makes the tests more consistent so we can compare them more easily.

Related to https://github.com/MetaMask/core/issues/1197.

For reference, see core: https://github.com/MetaMask/core/blob/3db352a7688bb1171fe1c39831c50f41877a0b97/packages/network-controller/src/NetworkController.ts#L369

## Manual Testing Steps

There should be no user-facing changes. All tests should pass.

## Pre-merge author checklist

- [x] I've clearly explained:


app/scripts/controllers/network/network-controller.ts
```diff
@@ -596,16 +596,6 @@ export class NetworkController extends EventEmitter {
       return;
     }
 
-    if (!chainId) {
```
mcmire, line 599: This scenario is now impossible. If a provider is set then `#configureProvider` must have been called and we would have already checked that `chainId` is set. (This check is not present on the core side.)

Gudahtt (approved):
LGTM!

brad-decker (approved):

PR #19304, opened 2023-05-26
Add sonarqube scans to the extension

pedronfigueiredo:
## Explanation

Add Sonarqube scanning to the extension.

TODO:
- [x] Add SONAR_TOKEN secret on the extension
- [ ] Change long-lived branch regex

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -1,0 +1,18 @@
+sonar.projectKey=metamask-extension
+sonar.organization=consensys
+
+# This is the name and version displayed in the SonarCloud UI.
+sonar.projectName=MetaMask Extension
+#sonar.projectVersion=1.0
+
+# Root for sonar analysis.
+sonar.sources=app/
+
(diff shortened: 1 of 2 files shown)


brad-decker (approved):

sonar-project.properties
```diff
@@ -0,0 +1,18 @@
...
+#sonar.projectVersion=1.0
+
+# Root for sonar analysis.
+sonar.sources=app/
+
+# Excluded project files from analysis.
+#sonar.exclusions=
+
+# Inclusions for test files.
+sonar.test.inclusions=**.test.**
```
DDDDDanica, line 15: shall we also include e2e tests `xxx.spec.js`

pedronfigueiredo: Good question. I don't think so because Sonar considers unit tests only for coverage.

DDDDDanica (approved):

PR #19306, opened 2023-05-26
[FLASK] Allow Snaps to use `eth_accounts` as a revokable permission

FrederikBolding:
## Explanation

Allows Snaps to access `eth_accounts` as a revokable permission. This means that snaps now can use `eth_accounts` and `eth_requestAccounts` to prompt the user for account selection. This permission can be revoked again in the snaps settings whenever.

This PR is mainly UI changes to make the account connection flow work with these changes and UI additions for permission revoking.

This PR needs the latest, unpublished version of the snaps monorepo to work completely.

### DIFF
--- a/ui/components/app/permission-cell/permission-cell.js
+++ b/ui/components/app/permission-cell/permission-cell.js
@@ -21,6 +21,7 @@ import {
 import { formatDate } from '../../../helpers/utils/util';
 import { useI18nContext } from '../../../hooks/useI18nContext';
 import Tooltip from '../../ui/tooltip';
+import { PermissionCellOptions } from './permission-cell-options';
 
 const PermissionCell = ({
   title,
@@ -29,6 +30,7 @@ const PermissionCell = ({
   avatarIcon,
   dateApproved,
   revoked,
+  showOptions,
 }) => {
   const t = useI18nContext();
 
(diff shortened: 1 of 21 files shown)


ui/components/app/permission-cell/permission-cell.js
hmalik88, line 1: Shouldn't the additions here be fenced?

FrederikBolding: The permission cell component is currently only used in snaps screens, so I don't think that's necessary.

ui/pages/permissions-connect/permissions-connect.component.js
```diff
@@ -268,30 +277,57 @@ export default class PermissionConnect extends Component {
...
-          >
-            <Icon
-              name={IconName.ArrowLeft}
-              marginInlineEnd={1}
-              size={IconSize.Xs}
+      <Box
+        style={{
+          ///: BEGIN:ONLY_INCLUDE_IN(snaps)
+          marginBottom:
+            targetSubjectMetadata.subjectType === SubjectType.Snap && '14px',
```
GuillaumeRx, line 293: this should be moved out of `style` in favour of Box `marginBottom` own prop and use either `2` or `4` for the value

GuillaumeRx: After looking at it it seems like with the legacy style it's too much pain to make this change.

GuillaumeRx (approved):

hmalik88 (approved):

PR #19312, opened 2023-05-26
Fix flakey tests in metamask controller tests

NEllusion:
## Explanation

In `metamask-controller.test.js` we have a test which assets that we only ever reset the state once when the metamask controller is initialized. Context on the change can be found in https://github.com/MetaMask/metamask-extension/pull/18344.

However, we were not correctly resetting a couple of our `sinon.spies` after each test run which meant that tests reliant on the number of times a function would be called were not accurate. This had not been surfaced due to the fact that the tests which relied on these spies were always run first (see test below).

https://github.com/MetaMask/metamask-extension/blob/acad431508319ee82ac346cfc74713edf5cf8f8a/app/scripts/metamask-controller.test.js#L266-L269

In addition, when doing our MV3 tests, we are still creating an MV2 controller in our setup which results in having the same browser polyfill used for both an MV2 and MV3 controller initialization. Making this refactor ensures we no longer need to include a `MetaMaskControllerMV3.prototype.resetStates.resetHistory();` before each of our mv3 tests

## Pre-merge author checklist

- [x] I've clearly explained:


app/scripts/metamask-controller.test.js
```diff
@@ -180,8 +181,10 @@ describe('MetaMaskController', function () {
   before(async function () {
     globalThis.isFirstTimeProfileLoaded = true;
     await ganacheServer.start();
-    sinon.spy(MetaMaskController.prototype, 'resetStates');
-    sinon.spy(MetaMaskControllerMV3.prototype, 'resetStates');
+    resetStateSpies = [
+      sinon.spy(MetaMaskController.prototype, 'resetStates'),
```
Gudahtt, line 185: Typically our strategy for keeping tests isolated is to construct spies/mocks/fakes fresh for each test, then discard them at the end by calling `sinon.restore()`. Would this work here as well?

That would look like moving these two spies to `beforeEach`.

And for the cleanup step, we could add `sinon.restore` to `afterEach`, or we could switch these two mocks to use `sandbox.spy` rather than `sinon.spy` so they're covered by the pre-existing `sandbox.restore` in `beforeEach` already. (ultimately we should be using the default sandbox here, but fixing that is a bigger problem).

app/scripts/metamask-controller.test.js
```diff
@@ -286,1336 +250,1411 @@ describe('MetaMaskController', function () {
...
-          getVersion: () => 'foo',
-        },
-        browser: browserPolyfillMock,
-        infuraProjectId: 'foo',
-        isFirstMetaMaskControllerSetup: false,
+    describe('should reset states on first time profile load', function () {
+      it('is an empty test', function () {
+        assert.ok(true);
+      });
+
```
Gudahtt, line 269: I'm guessing this isn't meant to be here

```suggestion
```

app/scripts/metamask-controller.test.js
Gudahtt: Seems like a good idea to me! It's not so hard to review with "hide whitespace" toggled

Gudahtt (approved):

DDDDDanica (approved):

PR #19325, opened 2023-05-29
Log incoming/outgoing messages between all MM contexts

weizman:
This PR is based on https://github.com/MetaMask/post-message-stream/pull/93 & https://github.com/MetaMask/extension-port-stream/pull/46.

This will log all messages exchanged between MM contexts (background, ui, contentscript, inpage, etc) only in DEBUG mode.

> Order of devtools screens in the following demonstrating image is: UI / CONTENT-SCRIPT / BACKGROUND:

Screenshot 2023-05-29 at 20 31 21


weizman:
moved to https://github.com/MetaMask/metamask-extension/pull/20658

PR #19326, opened 2023-05-29
feat: add new linea mainnet network

VGau:
## Explanation

- Added new Linea mainnet network as a default network.
- Updated `@metamask/eth-json-rpc-infura` package version to 8.1.0 to support linea networks
- ⚠️ Linea mainnet network won't be display on MM before the release date. This network is under feature toggle for now.

See: https://github.com/ConsenSys/zkevm-monorepo/issues/551

### Before
Screenshot 2023-06-05 at 1 16 41 PM

### After
Screenshot 2023-06-05 at 1 51 30 PM
Screenshot 2023-06-05 at 1 51 49 PM

## Pre-merge author checklist

- [x] I've clearly explained:


app/scripts/controllers/network/create-network-client.ts
```diff
@@ -57,7 +57,11 @@ export function createNetworkClient(
   const rpcApiMiddleware =
     networkConfig.type === NetworkClientType.Infura
       ? createInfuraMiddleware({
-          network: networkConfig.network,
+          network:
+            // TODO: this condition is to test the linea mainnet implemenation because the mainnet infrastructure is not yet ready. TO BE REMOVED
+            networkConfig.network === 'linea-mainnet'
```
Gudahtt, line 62: Is this still needed? If so, when is this planned to be removed?

shared/constants/network.ts
```diff
@@ -289,17 +303,22 @@ export const BUILT_IN_NETWORKS = {
...
   },
   [NETWORK_TYPES.MAINNET]: {
     networkId: NETWORK_IDS.MAINNET,
     chainId: CHAIN_IDS.MAINNET,
     blockExplorerUrl: `https://etherscan.io`,
   },
+  [NETWORK_TYPES.LINEA_MAINNET]: {
+    networkId: NETWORK_IDS.LINEA_MAINNET,
+    chainId: CHAIN_IDS.LINEA_MAINNET,
+    blockExplorerUrl: 'https://lineascan.build',
```
Gudahtt, line 320: This site does not seem to be working

ui/pages/settings/networks-tab/networks-list-item/networks-list-item.js
```diff
@@ -95,13 +97,13 @@ const NetworksListItem = ({
           <UrlIcon
             className="networks-tab__content__icon-with-fallback"
             fallbackClassName="networks-tab__content__icon-with-fallback"
-            name={label}
+            name={label || getNetworkLabelKey(labelKey)}
           />
         )
       )}
-      {network.isATestNetwork && (
+      {network.isATestNetwork && network.chainId !== CHAIN_IDS.LINEA_GOERLI && (
```
Gudahtt, line 104: Why was this added condition needed here?

ui/helpers/utils/i18n-helper.tsx
```diff
@@ -165,3 +166,13 @@ async function fetchRelativeTimeFormatData(languageTag: string): Promise<Json> {
   );
   return await response.json();
 }
+
+export function getNetworkLabelKey(network: string): string {
```
Gudahtt, line 170: Why do these two networks need a different "label key" than the others?

app/scripts/controllers/network/create-network-client.ts
VGau: Yes it is needed because we use the testnet RPC url for now. We will update this url once the linea mainnet network is ready.

shared/constants/network.ts
VGau: Yes this url is not working for the moment but will be the one for linea mainnet :)

ui/helpers/utils/i18n-helper.tsx
VGau: It is because `app/_locales` `messages.json` files, use networks name as keys. `linea-goerli` and `linea-mainnet` networks have this `-` character in their name . `app/_locales` `message.json` files cannot have keys containing `-` character.
That's why I use `lineaGoerli` and `lineaMainnet` as keys in `app/_locales` files and that's why I need this `getNetworkLabelKey` function to map real network names with `app/_locales` `message.json` files keys.

ui/pages/settings/networks-tab/networks-list-item/networks-list-item.js
VGau: It is a temporay condition because linea mainnet network use the linea goerli rpc url and chainId

shared/modules/network.utils.ts
```diff
@@ -56,3 +57,15 @@ export function isTokenDetectionEnabledForNetwork(chainId: string | undefined) {
 function isSafeInteger(value: unknown): value is number {
   return Number.isSafeInteger(value);
 }
+
+export async function shouldShowLineaMainnet(rpcUrl: string): Promise<boolean> {
```
danjm, line 61: Nit: if `currentDateHigherThanReleaseDate` is false, then we can return false before we make the `jsonRpcRequest`

ui/components/multichain/network-list-menu/network-list-menu.js
```diff
@@ -62,7 +63,19 @@ export const NetworkListMenu = ({ onClose }) => {
 
   const completedOnboarding = useSelector(getCompletedOnboarding);
 
+  const [lineaMainnetReleased, setLineaMainnetReleased] = useState(false);
```
danjm, line 66: This will result in a network request every time the network menu is rendered. We could instead call `shouldShowLineaMainnet` in the constructor of the preferences controller, and store the result in preferences controller state, so that once mainnet is released, further requests are not needed.

app/scripts/controllers/network/create-network-client.ts
Gudahtt: Hiding the Linea mainnet network prior to release makes sense, but I don't see the purpose in making it act like Linea Goerli. If we do this, then it will require another code change to make the network work properly. If that's so, then why bother adding it all until that point?

VGau: Indeed :) The idea was to put the linea-goerli rpc url for the linea mainnet to test the integration but I think it is better to put the real rpc url. I will change that thanks :)

ui/helpers/utils/i18n-helper.tsx
Gudahtt: Thanks, makes sense.

This does go against a guideline we have about using variables for the `t` function - we prefer to use string literals instead so that we can more effectively find unused localized messages and remove them. But this was a pre-existing pattern in this case and you're just extending it, so this seems fine for this PR

ui/components/multichain/network-list-menu/network-list-menu.js
Gudahtt: Great point. Do we need the `eth_chainId` call for this? It would be ideal to avoid network call side-effects if possible, and just use the timer.

ui/helpers/utils/i18n-helper.tsx
VGau: I agree that is not a very clean solution but it is the only solution I found. If you have another suggestion feel free to share it  😄

Gudahtt: Our solution elsewhere is to create a mapping of keys (e.g. network name) to localized _messages_ (not message keys), e.g. as we've done here: https://github.com/MetaMask/metamask-extension/blob/develop/shared/notifications/index.js

While I would prefer that solution, I'm actually glad that you haven't done it here because it would increase the scope of the PR too much. I'm just mentioning it here for future reference.

shared/modules/network.utils.ts
```diff
@@ -56,3 +57,20 @@ export function isTokenDetectionEnabledForNetwork(chainId: string | undefined) {
...
+  const currentDateHigherThanReleaseDate =
+    new Date().getTime() > Date.UTC(2023, 6, 11);
+
+  if (!currentDateHigherThanReleaseDate) {
+    return false;
+  }
+
+  let chainId: string | null;
+  try {
+    chainId = (await jsonRpcRequest(rpcUrl, 'eth_chainId')) as string;
```
Gudahtt, line 71: What is the purpose of this chain ID check?

VGau: It just to check if the linea mainnet rpc url is working. We can choose another endpoint if you want :)

ui/components/multichain/network-list-menu/network-list-menu.js
danjm: We do need a network call, because they don't want to show the network before infura's Linea mainnet is actually live. So the logic is supposed to be: don't show the network before this date AND don't show it before infura's mainnet is live. The date based condition ensures that the network won't be shown before the planned launch date, the network call check allows the launch date to be moved.

danjm (approved):
This looks good to me

Gudahtt (changes requested):
Looks good apart from the issue with the network call; I can approve after that is resolved

shared/modules/network.utils.ts
danjm: good point

danjm (changes requested):
removing my approval because of the network call issue

shared/modules/network.utils.ts
danjm: I think there might be a product decision to make here though. Effectively the network would be hidden at launch for anyone not on an infura network. How many people do we think that is, and is that number acceptable given the goals of the network launch?

danjm: There is an alternative, that would require product approval, where we should the network after the date and only make a network request once the user attempts to switch to the linea network, and if it fails we could show a custom error message (e.g. "Thanks for your interest in Linea Mainnet, it will be available soon!"), or hide the network from the list or just fallback to our current failure mode when failing to connect to a network (loading spinner and then the option to try again or switch networks)

Gudahtt: That seems like a good idea to me!

Gudahtt (dismissed):
LGTM!

danjm (approved):

Gudahtt (approved):

PR #19337, opened 2023-05-30
adapt chaining approval flow

bergarces:
## Explanation

Adopts the new `ApprovalController` version with the `startFlow` and `endFlow` methods.

None of the confirmations have been adapted to use these methods yet, this is an initial PR just to enable its future use.

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/568

## Pre-merge author checklist

- [X] I've clearly explained:


ui/pages/confirmation/confirmation.js
```diff
@@ -32,11 +32,13 @@ import {
   ///: END:ONLY_INCLUDE_IN
   getUnapprovedTemplatedConfirmations,
   getUnapprovedTxCount,
+  hasPendingApprovalFlows as hasPendingApprovalFlowsSelector,
```
bergarces, line 35: Should we append the word `Selector` at the end of our boolean selector methods that start with `has`? Otherwise we are going to find this problem very often.

Another options would be not having selectors that are not prefaced by the word `get` or `select`, which seems to be the naming convention as far as I can see (https://redux.js.org/style-guide/#name-selector-functions-as-selectthing).

That'd mean having to check the length every time to get this behaviour.

app/scripts/lib/rpc-method-middleware/handlers/add-ethereum-chain.js
bergarces, line 1: These changes are to be completely removed before merging. Here just to show an example whilst the PR is in Draft.

PR #19358, opened 2023-05-31
Update mv3 minimum version in manifest to chrome v88

danjm:
MV3 only works on chrome versions 88 and later. Without the change in this PR, attempts to upload to the chrome store will fail.


Gudahtt (approved):
LGTM!

darkwing (approved):

Issue #19271, opened 2023-05-23
Bump core packages that have been released in 53.0.0


mcmire:
[v53.0.0 of the `core` monorepo](https://github.com/MetaMask/core/pull/1385) will introduce many breaking changes. It would be better to address these breaking changes before we introduce the new version of the NetworkController into the extension.

This is the list of packages that the extension relies on and which will be upgraded with v53.0.0 (listed in alphabetical order but not necessarily the order that these should be addressed):

- [x] `@metamask/address-book-controller`
- [x] `@metamask/announcement-controller`
- [x] `@metamask/approval-controller`
- [x] `@metamask/assets-controllers`
- [x] `@metamask/base-controller`
- [x] `@metamask/controller-utils`
- [x] `@metamask/gas-fee-controller`
- [x] `@metamask/message-manager`
- [x] `@metamask/notification-controller`
- [x] `@metamask/permission-controller`
- [x] `@metamask/rate-limit-controller`
- [x] `@metamask/signature-controller`

These controllers rely on events or data from the extension version of NetworkController. This may impact what can be upgraded individually and what must be upgraded along with NetworkController:

- **TokenListController** (from `@metamask/assets-controllers`)
  - Assumes chain ID is a hex string
  - Subscribes to the ObservableStore on the NetworkController to keep the chain ID up to date
- **PreferencesController** (local to extension)
  - Receives the `provider` from the NetworkController, but doesn't use it
- **TokensController** (from `@metamask/assets-controllers`)
  - Subscribes to the ObservableStore on the NetworkController to keep the chain ID up to date
  - Receives the provider from NetworkController and uses it to make contract calls
- **AssetsContractController** (from `@metamask/assets-controllers`)
  - Subscribes to the ObservableStore on the NetworkController to keep the chain ID up to date
  - Receives the provider from NetworkController and uses it to make contract calls
- **NftController** (from `@metamask/assets-controllers`)
  - Subscribes to the ObservableStore on the NetworkController to keep the chain ID up to date
  - Reuses methods within AssetsContractController to access info about contracts
- **NftDetectionController** (from `@metamask/assets-controllers`)
  - Subscribes to the ObservableStore on the NetworkController to keep the chain ID up to date
- **MetaMetricsController** (local to extension)
  - On demand, reads the ObservableStore on the NetworkController to access parts of the current provider config (`type`, `rpcUrl`, `chainId`)
- **GasFeeController** (from `@metamask/gas-fee-controller`)
[...]

PR #19366, opened 2023-05-31
Update `@metamask/gas-fee-controller` to v6

Gudahtt:
## Explanation

The `@metamask/gas-fee-controller` package has been updated to v6. This version was part of the [core monorepo v53](MetaMask/core#1385) release. The remaining packages released as part of v53 will be updated in later PRs.

This release included a number of breaking changes, but most of them do not affect the extension:

* Bump to Node 16
  * The extension already uses Node.js v16
* The `getChainId` constructor parameter now expects a `Hex` return type rather than a decimal string
  * The extension was already passing in a `getChainId` parameter that returned `Hex`
* The gas fee controller messenger now requires the `NetworkController:stateChange` event instead of the `NetworkController:providerConfigChange` event
  * This does not apply if `onNetworkStateChange` and `getChainId` are provided to the constructor, which is the case here.
* Update `@metamask/network-controller` dependency and peer dependency
  * This dependency is only used for types, and none of the type changes affect how the extension interacts with this controller.
  
The one change that did have an impact is that the constructor parameter `onNetworkStateChange` now expects event handlers to be passed the full network state.

Relates to #19271

## Manual Testing Steps

No functional changes

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -545,10 +545,14 @@ export default class MetamaskController extends EventEmitter {
         this.networkController.getProviderAndBlockTracker().provider,
       // NOTE: This option is inaccurately named; it should be called
       // onNetworkDidChange
-      onNetworkStateChange: networkControllerMessenger.subscribe.bind(
-        networkControllerMessenger,
-        'NetworkController:networkDidChange',
-      ),
+      onNetworkStateChange: (eventHandler) => {
+        networkControllerMessenger.subscribe(
+          'NetworkController:networkDidChange',
+          () => {
+            eventHandler(this.networkController.store.getState());
+          },
+        );
+      },
       getCurrentNetworkEIP1559Compatibility:
         this.networkController.getEIP1559Compatibility.bind(
           this.networkController,
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -549,7 +549,11 @@ export default class MetamaskController extends EventEmitter {
         networkControllerMessenger.subscribe(
           'NetworkController:networkDidChange',
           () => {
-            eventHandler(this.networkController.store.getState());
+            const state = this.networkController.store.getState();
+            if (process.env.IN_TEST) {
+              state.providerConfig.chainId = CHAIN_IDS.MAINNET;
+            }
+            eventHandler(state);
           },
         );
       },
--- a/app/scripts/metamask-controller.js
+++ b/app/scripts/metamask-controller.js
@@ -549,9 +549,15 @@ export default class MetamaskController extends EventEmitter {
         networkControllerMessenger.subscribe(
           'NetworkController:networkDidChange',
           () => {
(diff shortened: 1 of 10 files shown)


DDDDDanica (dismissed):

DDDDDanica:
LGTM !! 💯

app/scripts/metamask-controller.js
```diff
@@ -545,10 +545,14 @@ export default class MetamaskController extends EventEmitter {
         this.networkController.getProviderAndBlockTracker().provider,
       // NOTE: This option is inaccurately named; it should be called
```
mcmire, line 546: Nit: I think we can drop this comment now. The fact that we are listening to `networkDidChange` and the fact that the callback takes the NetworkController state are unrelated.

mcmire (dismissed):
Don't know how I missed this before. One minor but non-blocking thing. LGTM.

app/scripts/metamask-controller.js
```diff
@@ -545,10 +545,18 @@ export default class MetamaskController extends EventEmitter {
...
-        networkControllerMessenger,
-        'NetworkController:networkDidChange',
-      ),
+      onNetworkStateChange: (eventHandler) => {
+        networkControllerMessenger.subscribe(
+          'NetworkController:networkDidChange',
+          () => {
+            const state = this.networkController.store.getState();
+            if (process.env.IN_TEST) {
+              state.providerConfig.chainId = CHAIN_IDS.MAINNET;
```
mcmire, line 554: @Gudahtt I noticed your last commit that added this. In which tests is the chain ID getting set to mainnet?

app/scripts/metamask-controller.js
Gudahtt: I agree those are unrelated, but I thought the comment was suggesting that the name should match the event. Right now it wrongly suggests that the handler should be called on each state change, when really it's called on each network switch.

app/scripts/metamask-controller.js
Gudahtt: None, but the test builds have always used `1` as the chain ID for test updates. See the `getChainId` method for example. I updated this to follow suit since that's how it worked before. That necessitated updating the state to overwrite the chain ID as well, since some of the tests include network switching and this new version of the gas controller uses this state now (whereas before it didn't).

Gudahtt: Maybe it would be simpler use the real chain ID instead, and update the e2e test mocks to expect the chain ID to be 1337. Hmm.

Gudahtt:
Moving this back into draft until CI is passing

app/scripts/metamask-controller.js
Gudahtt: It was simple to remove this; we're now using the real chain ID instead

mcmire (approved):
Changes look good!

legobeat (approved):

PR #19377, opened 2023-06-01
[FLASK] `snaps@0.34.0-flask.1`

FrederikBolding:
## Explanation

Bumps Snaps Flask packages to `0.34.0-flask.1` and handles any breaking changes. 

This also bumps `metamask/permission-controller` to the latest version in both stable and Flask.

The PermissionController bump is a breaking change and therefore some changes renaming the permission specification property `targetKey` to `targetName` was included in this PR. 

Since there is no stable release of the Snaps Platform yet, I have mocked and replaced the stable snaps dependencies with the Flask dependencies in all unit tests where required. This is required because of the aforementioned breaking change, once we release a stable version of the snaps packages, this can be removed again.


FrederikBolding:
@SocketSecurity ignore fast-xml-parser@4.2.2
@SocketSecurity ignore strnum@1.0.5

yarn.lock
```diff
@@ -4433,6 +4433,26 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@metamask/permission-controller@npm:^4.0.0":
```
Gudahtt, line 4436: Nit: Perhaps we could add a resolution to remove the duplicate copy of the permission controller. The only uses at the moment are in dead code.

Gudahtt (approved):
LGTM!

The strategy of using `-flask` packages for unit tests but not in the actual code doesn't seem ideal. But hopefully we can improve that soon.

app/scripts/controllers/permissions/snaps/snap-permissions.test.js
```diff
@@ -8,6 +8,14 @@ import {
   buildSnapRestrictedMethodSpecifications,
 } from './snap-permissions';
 
+// Temporarily replace the snaps packages with the Flask versions.
+jest.mock('@metamask/snaps-controllers', () =>
+  jest.requireActual('@metamask/snaps-controllers-flask'),
+);
+jest.mock('@metamask/rpc-methods', () =>
+  jest.requireActual('@metamask/rpc-methods-flask'),
+);
```
Mrtenz, line 17: Why does this not work otherwise? Does this not break in production?

FrederikBolding: No, it is code-fenced out in production but not in unit tests

Mrtenz (approved):

PR #19383, opened 2023-06-01
Fix EIP-1559 network compatibility check 

vinistevam:
## Explanation

This PR aims to fix a bug that occurs when a user switches the network and performs a transaction, triggering the check for EIP-1559 compatibility. The bug causes an incorrect determination of whether the network is compatible or not with EIP-1559. This PR addresses the issue by not assigning incorrectly the compatibility with EIP-1559 when middleware returns `null` getting the latest block. If the above happens from the perspective of the user it won't change anything and as the compatibility with EIP-1559 was not determined the first time due to being unable to retrieve the latest block, it will try to determine again, [here](https://github.com/MetaMask/core/blob/main/packages/network-controller/src/NetworkController.ts#L918).

Note: the extension is now using the NetworkController from core repo so the changes in the code from this PR were moved to [PR](https://github.com/MetaMask/core/pull/1457) keeping here only the new `@metamask/network-controller` version.

### Before

Once the user switches the network even when the bug happens and the middleware resolves the latest block with `null` the network switch used to be concluded successfully with network status `available` but the `EIPS[1559]` in `networkDetails` would always return `false` even when the network supports the EIP-1559.

So the next time the compatibility is checked is when you perform a transaction using EIP-1559 and the check returns the `EIPS[1559]` in `networkDetails`  incorrectly determined (in this case) when the network switch happened, throwing an error mentioned in the bug.

Screenshot from 2023-05-26 16-54-04

### After
If middleware returns `null` getting the latest block from the perspective of the user, it won't change anything but the compatibility with EIP-1559 will not be set  and when the user performs a transaction as the compatibility with EIP-1559 was not determined the first time due to being unable to retrieve the latest block, it will try to determine it again, [here](https://github.com/MetaMask/core/blob/main/packages/network-controller/src/NetworkController.ts#L918)

[simulate-middleware-resolves-latest-block-null-on-switch-network.webm](https://github.com/MetaMask/core/assets/45455812/6c6dc9b2-f0e2-489f-aa0a-18504edc561f)

**Leaving the first proposal as a reference**
It retries 3 times until the middleware provides the latest block so the check can be performed. If after the retries the block is still null It throws and prevents the user to switch the network. It's the behaviour expected if the middleware/provider responds with an error.
[middleware-resolves-last-block-null](https://github.com/MetaMask/metamask-extension/assets/45455812/5d1627bf-2659-4655-b0da-a84ee73e5481)

## Manual Testing Steps

- Open MetaMask and switch to a network compatible with EIP-1559 (sepolia/goerli)
- The switch should occur as normal
- Go to [the test dapp](https://metamask.github.io/test-dapp/)
- Perform a transaction using the "Send EIP 1559 Transaction" button
- Transaction should be submitted as normal

## Pre-merge author checklist

- [x] I've clearly explained:


app/scripts/controllers/network/network-controller.ts
```diff
@@ -839,14 +839,34 @@ export class NetworkController extends EventEmitter {
...
    *
    * @param provider - A provider, which is guaranteed to be available.
+   * @param maxRetries - The maximum number of retries to attempt fetching the latest block. Defaults to 3.
    * @returns A promise that resolves to true if the network supports EIP-1559
-   * and false otherwise.
+   * and false otherwise. Throws an error if unable to fetch the last block.
    */
   async #determineEIP1559Compatibility(
     provider: SafeEventEmitterProvider,
+    maxRetries = 3,
```
matthewwalsh0, line 848: As this is a private method so only called once, is it clearer to move this to a file level constant?

app/scripts/controllers/network/network-controller.ts
```diff
@@ -839,14 +839,34 @@ export class NetworkController extends EventEmitter {
...
   ): Promise<boolean> {
-    const latestBlock = await this.#getLatestBlock(provider);
-    return latestBlock?.baseFeePerGas !== undefined;
+    let latestBlock;
+    const retryInterval = 500;
+
+    for (let retry = 0; retry <= maxRetries; retry++) {
```
matthewwalsh0, line 853: Is `retries` a little ambiguous as technically it would mean attempts after the initial one so this results in four total attempts? Perhaps we instead call them `attempts` with a `maxAttempts` constant?

app/scripts/controllers/network/network-controller.ts
```diff
@@ -839,14 +839,34 @@ export class NetworkController extends EventEmitter {
...
+      latestBlock = await this.#getLatestBlock(provider);
+
+      if (retry < maxRetries && !latestBlock) {
```
matthewwalsh0, line 856: Would this logic be cleaner if we put it inside a `getLatestBlockWithRetry` then can early return as soon as we have a block, or return null if we ever complete the loop?

app/scripts/controllers/network/network-controller.ts
```diff
@@ -839,14 +839,34 @@ export class NetworkController extends EventEmitter {
...
+        await new Promise((resolve) => setTimeout(resolve, retryInterval));
+      } else {
+        break;
+      }
+    }
+    if (!latestBlock) {
+      throw new Error('Unable to fetch last block');
+    }
+
+    if (latestBlock?.baseFeePerGas !== undefined) {
```
matthewwalsh0, line 866: We shouldn't need the optional chaining here now as we null check above.

app/scripts/controllers/network/network-controller.ts
```diff
@@ -839,14 +839,34 @@ export class NetworkController extends EventEmitter {
...
+      return true;
+    }
+    return false;
```
matthewwalsh0, line 869: Can we simply return `latestBlock.baseFeePerGas !== undefined`?

matthewwalsh0 (dismissed):

cryptotavares (dismissed):

OGPoyraz (dismissed):

PR #19390, opened 2023-06-01
feature(on-ramp): add address and chain id to url params

wachunei:
This reverts commit 82599466259b6dec1946e6e82e07b89c4405a88a.

## Explanation

This PR reverts 82599466259b6dec1946e6e82e07b89c4405a88a, which deleted the current address and chain ID from the buy crypto links to the portfolio.

## Manual Testing Steps

When clicking any "buy" or "buy \<Token\>" link, it must include selected address and chain id.

## Pre-merge author checklist

- [x] I've clearly explained:


wachunei:
@legobeat Although before the portfolio integration it was sent in plain text as part of the query params in every case, it is completely reasonable to have the concerns about the URL being recorded by either the browser and its extensions or at a network level by VPNs or proxies. 

What do you think we deliver it to the portfolio in a hashed format and then the portfolio compares the hashes from the local state with the incoming param?

We have forgotten to mention this will also help avoid a critical mistake for the user when clicking on the buy button in the extension from a secondary account: if the portfolio does not have the context of the selected address and it is connected with a different account then the user could be end up completing the flow and sending the funds to a different account.

PR #19403, opened 2023-06-02
Widen the number of test files that TS sees

mcmire:
## Explanation

Currently, all test files are excluded from TypeScript's purview. That
means that if any test file fails to typecheck, we won't know about it.
In fact if we were to remove all test files from the `exclude` list,
then some files would indeed fail to be typechecked.

At the moment all we care about is the unit test file for
NetworkController. Due to how the `include`/`exclude` rules work in
`tsconfig.json`, we can't "un-exclude" that file after it's excluded, so
to fix this, we simply exclude fewer things.

While we're at it, we also:

- Add `esModuleInterop: true` to address warnings we sometimes see
  around imports
- Add `noEmit: true` so that we can run `tsc` to test typechecking on
  its own without producing any compiled files (so that ESLint doesn't
  accidentally pick them up)
- Clean up `tsconfig.json` slightly

Tangentially related to https://github.com/MetaMask/core/issues/1197.

## Manual Testing Steps

As with other PRs, there should be no user-facing changes. Beyond this:

- `yarn tsc` should not produce any errors.
- `yarn start` (or whatever command you may use to build the extension locally) should not produce any errors.
- CI should pass.

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/lavamoat/browserify/beta/policy.json
+++ b/lavamoat/browserify/beta/policy.json
@@ -774,7 +774,6 @@
         "URL": true,
         "clearInterval": true,
         "clearTimeout": true,
-        "console.error": true,
         "console.info": true,
         "console.log": true,
         "setInterval": true,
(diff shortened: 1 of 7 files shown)


digiwand (approved):
thx for the helpful explanation! Happy to see more TS support

Gudahtt (approved):
LGTM!

Gudahtt:
I rebased this just to get CI passing; there was an audit failure that I thought might need an update (though it may have just been a superfluous false positive)

PR #19411, opened 2023-06-02
Add e2e tests for queuing multiple transactions

vinistevam:
## Explanation

This PR introduces end-to-end (E2E) tests to cover the following scenarios:

- User creates multiple pending transactions, queues them, and subsequently confirms the pending transactions.
- User creates multiple pending transactions, queues them, and subsequently rejects the pending transactions.

See: https://github.com/MetaMask/MetaMask-planning/issues/579

## Pre-merge author checklist

- [x] I've clearly explained:


test/e2e/tests/multiple-transactions.spec.js
```diff
@@ -95,4 +101,125 @@ describe('Confirm transactions', function () {
...
+        // initiates a transaction from the dapp
+        await openDapp(driver);
+        // creates first transaction
+        await driver.clickElement({
+          text: 'Send EIP 1559 Transaction',
+          tag: 'button',
+        });
+        await driver.waitUntilXWindowHandles(3);
+        const windowHandles = await driver.getAllWindowHandles();
+        const extension = windowHandles[0];
```
matthewwalsh0, line 129: This confused me a little, can we be super explicit and call this `extensionTab` or something?

test/e2e/tests/multiple-transactions.spec.js
```diff
@@ -95,4 +101,125 @@ describe('Confirm transactions', function () {
...
+        const confirmation = windowHandles[2];
+        await driver.switchToWindowWithTitle('E2E Test Dapp', windowHandles);
```
matthewwalsh0, line 131: DO we need to use `switchToWindowWithTitle` here? Will `windowHandles[1]` not be the test dApp?

test/e2e/tests/multiple-transactions.spec.js
```diff
@@ -95,4 +101,125 @@ describe('Confirm transactions', function () {
...
+        await driver.switchToWindow(extension);
+        await driver.delay(regularDelayMs);
+        await driver.clickElement('[data-testid="home__activity-tab"]');
+
+        await driver.wait(async () => {
+          const confirmedTxes = await driver.findElements(
+            '.transaction-list__completed-transactions .transaction-list-item',
+          );
+          return confirmedTxes.length === 2;
+        }, 10000);
```
matthewwalsh0, line 160: Is there an existing constant we can use here?

test/e2e/tests/multiple-transactions.spec.js
```diff
@@ -95,4 +101,125 @@ describe('Confirm transactions', function () {
...
+        await driver.waitForElementNotPresent('.loading-overlay__spinner');
+        // confirms first transaction
+        await driver.clickElement({ text: 'Confirm', tag: 'button' });
+
+        await driver.waitUntilXWindowHandles(2);
+        await driver.switchToWindow(extension);
+        await driver.delay(regularDelayMs);
+        await driver.clickElement('[data-testid="home__activity-tab"]');
+
+        await driver.wait(async () => {
```
matthewwalsh0, line 155: Do we need a custom `wait` here? Is it worth adding an optional timeout to `findElements` like `waitForSelector` has?

test/e2e/tests/multiple-transactions.spec.js
```diff
@@ -95,4 +101,125 @@ describe('Confirm transactions', function () {
...
+        await driver.switchToWindow(extension);
+        await driver.delay(regularDelayMs);
+        await driver.clickElement('[data-testid="home__activity-tab"]');
+
+        const noTransactions = await driver.findElements(
+          '.transaction-list__empty-text',
+        );
+        assert.equal(
+          await noTransactions[0].getText(),
+          'You have no transactions',
```
matthewwalsh0, line 220: Is it safe to use text as a selector? Would it be more stable to verify there is no entries in the table?

test/e2e/tests/multiple-transactions.spec.js
```diff
@@ -95,4 +101,125 @@ describe('Confirm transactions', function () {
       },
     );
   });
+
+  it('creates multiple queued transactions, then confirms', async function () {
```
matthewwalsh0, line 105: The existing tests appear to do the same but with a fixture to provide the transactions rather than end to end.

For the sake of time and maintainability, should we delete the previous ones if these encompass them?

test/e2e/tests/multiple-transactions.spec.js
vinistevam: you are right, I pushed the changes now it's using `const dApp = windowHandles[1]`.

test/e2e/tests/multiple-transactions.spec.js
vinistevam: I've run some tests in the pipeline and locally and it can be simplified by removing the custom await.

test/e2e/tests/multiple-transactions.spec.js
vinistevam: It's been used like that in other places, but I got to a better result asserting if `.transaction-list__empty-text` is present and visible and the list of transitions is not present.

test/e2e/tests/multiple-transactions.spec.js
vinistevam: 100%, I've removed the test using fixtures as the new ones are covering end to end.

OGPoyraz (approved):

test/e2e/tests/multiple-transactions.spec.js
```diff
@@ -71,27 +95,50 @@ describe('Confirm transactions', function () {
...
         });
-        await driver.clickElement({ text: 'Reject', tag: 'button' });
-        await driver.waitForSelector({
-          text: 'Reject 3 transactions',
-          tag: 'a',
+        await driver.waitUntilXWindowHandles(3);
+        const windowHandles = await driver.getAllWindowHandles();
+        const extension = windowHandles[0];
+        const confirmation = windowHandles[2];
+        await driver.switchToWindowWithTitle('E2E Test Dapp', windowHandles);
```
matthewwalsh0, line 109: Very minor, but another switch here but I assume we already have the handle in `windowHandles[1]`?

matthewwalsh0 (approved):

PR #19446, opened 2023-06-05
Add e2e tests for queuing multiple signing operations

vinistevam:
## Explanation

This PR introduces end-to-end (E2E) tests to cover the following scenarios:

- User creates multiple pending signing operations (eth signing, personal signing, typed data, v4 and  v3), queues them, and subsequently confirms the pending transactions.

See: https://github.com/MetaMask/MetaMask-planning/issues/582

## Pre-merge author checklist

- [x] I've clearly explained:


bergarces (dismissed):

OGPoyraz (dismissed):

OGPoyraz (dismissed):

test/e2e/tests/eth-sign.spec.js
```diff
@@ -101,4 +101,87 @@ describe('Eth sign', function () {
...
+        await driver.waitForSelector({
+          text: 'Reject 2 requests',
+          tag: 'a',
+        });
+
+        const title = await driver.findElement(
+          '.request-signature__content__title',
+        );
+        const origin = await driver.findElement('.request-signature__origin');
+        assert.equal(await title.getText(), 'Signature request');
```
matthewwalsh0, line 157: Is it better to use non-text selectors to prevent them being too brittle, and also support other locales?

test/e2e/tests/signature-request.spec.js
```diff
@@ -8,17 +8,45 @@ const {
...
+  accounts: [
+    {
+      secretKey:
+        '0x7C9529A67102755B7E6102D6D950AC5D5863C98713805CEC576B945B15B71EAC',
+      balance: convertToHexValue(25000000000000000000),
+    },
+  ],
+};
+
+async function verifyAndAssertSignTypedDataV4(driver) {
```
matthewwalsh0, line 21: Why is it only the typed tests have these helpers to verify and approve but not `ethSign` or `personal`? Are they applicable to both? Could they go in a common location to be used across all three?

test/e2e/tests/signature-request.spec.js
```diff
@@ -88,18 +165,31 @@ describe('Sign Typed Data V4 Signature Request', function () {
   });
 });
 
+async function verifyAndAssertSignTypedDataV3(driver) {
```
matthewwalsh0, line 168: Is it the standard to keep all the helper functions at the top of the file before the tests themselves?

test/e2e/tests/signature-request.spec.js
```diff
@@ -128,34 +218,80 @@ describe('Sign Typed Data V3 Signature Request', function () {
...
+        await driver.clickElement('#signTypedDataV3Verify');
+        const recoveredAddress = await driver.findElement(
+          '#signTypedDataV3VerifyResult',
         );
+        assert.equal(await recoveredAddress.getText(), publicAddress);
+      },
+    );
+  });
+
+  it('can queue multiple Signature Requests and confirm', async function () {
```
matthewwalsh0, line 242: Why is it we refer to these typed tests more generally as `Signature Requests` rather than `typed signs`? Would that be more explicit?

test/e2e/tests/signature-request.spec.js
```diff
@@ -210,20 +345,70 @@ describe('Sign Typed Data Signature Request', function () {
...
         );
-        const origin = await driver.findElement('.request-signature__origin');
-        const message = await driver.findElements(
-          '.request-signature__row-value',
+        assert.equal(await recoveredAddress.getText(), publicAddress);
+      },
+    );
+  });
+
+  it('can queue multiple Signature Request and confirm', async function () {
```
matthewwalsh0, line 366: Is there much difference between the V3 and V4 tests, would it be worth using a single template function with arguments for the validation methods perhaps?

I wonder if we could even abstract something for all of the signature types to remove as much duplication as possible across all three files?

test/e2e/tests/eth-sign.spec.js
```diff
@@ -101,4 +101,87 @@ describe('Eth sign', function () {
...
+        assert.equal(await origin.getText(), DAPP_URL);
+
+        const personalMessageRow = await driver.findElement(
+          '.request-signature__row-value',
+        );
+        const personalMessage = await personalMessageRow.getText();
```
matthewwalsh0, line 163: This terminology is confusing me as `personal` is an alternate signature type. Do we just mean the message body to sign? Is it worth renaming this for clarity?

test/e2e/tests/signature-request.spec.js
vinistevam: I've improved the signature request to use two templates one for the normal signature request and the other to support multiple operations. I also added some helpers in `ethSign` and `personal` but when I tried to expand the template that I've created to signature request it got really confusing and fragile.

test/e2e/tests/signature-request.spec.js
vinistevam: I found one example at the end of the page so I moved the other at the end as well.

test/e2e/tests/signature-request.spec.js
vinistevam: Renamed to clarify each test.

test/e2e/tests/signature-request.spec.js
vinistevam: I got to a point where I managed to abstract between V1, V3 and V4 but when I tried to expand to other signings it got fragile. I could give it another try.

bergarces (approved):

matthewwalsh0 (approved):

PR #19472, opened 2023-06-06
Upgrade assets-controllers to v9

mcmire:
## Explanation

The `@metamask/assets-controllers` package has been updated to v9. This
covers changes that were made as a part of the [v53][1] and [v55][2]
releases.

There were a lot of breaking changes that needed to be addressed as a
result of the upgrade. Please see the changelog notes for those releases
for the full list.

In addition, there were a couple of Yarn patches that were made to bring
in a partial set of functionality from these versions. These have been
removed, since they are no longer needed.

[1]: https://github.com/MetaMask/core/releases/tag/v53.0.0
[2]: https://github.com/MetaMask/core/releases/tag/v55.0.0

Progresses #19271.

## Manual Testing Steps

There should be no user-facing changes. The `assets-controllers` covers a swath of functionality which should work the same way as before, including but not limited to:

- Token detection
  - Balances should appear correctly for tokens in the token list, especially NFTs
- Sending and receiving tokens (ERC-20, ERC-721, ERC-1155)
  - Current information about tokens should appear when sending tokens with proper exchange rates, especially NFTs
- Swaps
  - Current information about tokens should appear with proper exchange rates

## Pre-merge author checklist

- [x] I've clearly explained:


mcmire:
This is a Storybook dependency so it's okay. This could go away in the future if/when we upgrade Storybook itself.

@SocketSecurity ignore error-stack-parser@2.1.4

app/scripts/controllers/detect-tokens.test.js
```diff
@@ -246,23 +246,7 @@ describe('DetectTokensController', function () {
...
-                ...networkState.providerConfig,
-                chainId: convertHexToDecimal(
-                  networkState.providerConfig.chainId,
-                ),
-              },
-            };
-            return cb(modifiedNetworkState);
-          },
-        ),
+      onNetworkStateChange: network.store.subscribe.bind(network.store),
```
mcmire, line 249: The original motivation for this change was to stop converting this chain ID to decimal since we don't need to do this. However, there's no real reason why we need to listen for `networkDidChange` in these tests. We probably don't even need a real NetworkController here, but removing that would be a bigger change.

app/scripts/metamask-controller.js
```diff
@@ -317,27 +317,20 @@ export default class MetamaskController extends EventEmitter {
...
-            ...networkState,
-            providerConfig: {
-              ...networkState.providerConfig,
-              chainId: hexToDecimal(networkState.providerConfig.chainId),
-            },
-          };
-          return cb(modifiedNetworkState);
-        });
-      },
+      onNetworkStateChange: this.networkController.store.subscribe.bind(
```
mcmire, line 331: This and other changes like this were made because the assets controllers expect the NetworkController state to contain a hex chain ID.

jest.config.js
```diff
@@ -48,6 +51,7 @@ module.exports = {
     '<rootDir>/app/scripts/lib/**/*.test.ts',
     '<rootDir>/app/scripts/lib/createRPCMethodTrackingMiddleware.test.js',
     '<rootDir>/app/scripts/migrations/*.test.js',
+    '<rootDir>/app/scripts/migrations/*.test.ts',
```
mcmire, line 54: An ancillary change, but this allows us to write migration tests in TypeScript rather than JavaScript.

test/helpers/setup-helper.js
```diff
@@ -73,14 +73,15 @@ const popoverContent = window.document.createElement('div');
...
 // eslint-disable-next-line no-shadow
-const fetch = require('node-fetch');
+const { default: fetch, Headers, Request, Response } = require('node-fetch');
 
-/* eslint-disable-next-line no-shadow */
-const { Headers, Request, Response } = fetch;
 Object.assign(window, { fetch, Headers, Request, Response });
+// some of our libraries currently assume that `fetch` is globally available,
+// so we need to assign this for tests to run
+global.fetch = fetch;
```
mcmire, line 84: With the `assets-controller` upgrade, the tests for DetectTokensController started failing. A real NetworkController gets used in these tests, and that causes the block tracker to start, which makes a call to the network through the provider. The fetch middleware uses a utility function from `controller-utils` to do this, and that ultimately uses `global.fetch` to make the request. Explicitly assigning `fetch` like this made those tests work again.

I was surprised that we haven't needed to do this before. In upgrading `assets-controllers`, `controller-utils` also got upgraded. However, `isomorphic-fetch` was removed in `controller-utils` back in 3.0.0, and here we've upgraded `controller-utils` from 3.4.0 to 4.0.0. Perhaps some other library that previously imported `isomorphic-fetch` is no longer doing so.

app/scripts/migrations/088.test.ts
```diff
@@ -0,0 +1,638 @@
...
+      data: {
+        NftController: {
+          allNftContracts: {
+            '0x111': {
+              '16': ['contract 1', 'contract 2'],
+              '32': ['contract 3', 'contract 4'],
+            },
+            '0x222': {
+              '64': ['contract 5', 'contract 6'],
+              '128': ['contract 7', 'contract 8'],
```
adonesky1, line 94: I realize the migration doesn't actually touch inside these arrays but wondering if they should be arrays of objects atleast to match the real shape?

app/scripts/migrations/088.test.ts
```diff
@@ -0,0 +1,638 @@
+import { migrate } from './088';
+
+describe('migration #88', () => {
+  it('updates the version metadata', async () => {
+    const oldStorage = {
+      meta: { version: 0 },
```
adonesky1, line 6: should these versions all be 87? Or does it not really matter?

mcmire: It doesn't matter! But I suppose to go with your previous comment about making these tests more realistic I can change this to 87.

adonesky1 (dismissed):
Looks great! Just a few non-blocking comments. I also took it for a spin and didn't see any issues in the flows that would be effected (didn't fully test the migrations though).

mcmire:
@adonesky1 Hmm... some e2e tests are failing for some reason. I'll take a look.

mcmire:
Putting this back in draft for now until I resolve the CI issues.

app/scripts/migrations/088.test.ts
mcmire: Sure! Fixed in [d274c4d](https://github.com/MetaMask/metamask-extension/pull/19472/commits/d274c4d6672dcdf0e12c497aff706d3c665e7d2d).

mcmire:
Paired with @adonesky1 on this and CI now passes! This is now ready for review again.

test/e2e/mock-e2e.js
```diff
@@ -343,7 +343,7 @@ async function setupMocking(server, testSpecificMock) {
     });
 
   await server
-    .forGet('https://token-api.metaswap.codefi.network/token/0x539')
+    .forGet('https://token-api.metaswap.codefi.network/token/1337')
```
adonesky1, line 346: this API accepts either decimal or hexadecimal formatted chainIds

adonesky1 (dismissed):
LGTM!

Gudahtt:
Looks good! Can approve after the fixture builder is updated

jest.config.js
```diff
@@ -11,6 +11,9 @@ module.exports = {
     '<rootDir>/app/scripts/lib/**/*.js',
     '<rootDir>/app/scripts/lib/createRPCMethodTrackingMiddleware.js',
     '<rootDir>/app/scripts/migrations/*.js',
+    '<rootDir>/app/scripts/migrations/*.ts',
+    '!<rootDir>/app/scripts/migrations/*.test.js',
+    '!<rootDir>/app/scripts/migrations/*.test.ts',
```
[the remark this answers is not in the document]
mcmire: Ah, yes, good catch. Updated in [845824b](https://github.com/MetaMask/metamask-extension/pull/19472/commits/845824b19f3bef68e3c8bdd33ae07341f738a917)

test/e2e/fixture-builder.js
```diff
@@ -519,7 +520,7 @@ class FixtureBuilder {
     return this.withNftController({
       allNftContracts: {
         '0x5cfe73b6021e818b776b421b1c4db2474086a7e1': {
-          1337: [
+          [toHex(1337)]: [
```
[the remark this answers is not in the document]
mcmire: Good catch! Updated in [afc73ad](https://github.com/MetaMask/metamask-extension/pull/19472/commits/afc73ad06e8da1b36e78a9738fe38818920389ef)

Gudahtt (approved):
LGTM!

adonesky1 (approved):

PR #19486, opened 2023-06-07
Replace NetworkController w/ core version

### LINKED ISSUE #17931: Replace the extension network controller with the shared network controller
Once all of the API normalization work has been completed, and once we've replaced `web3-provider-engine` on `@metamask/network-controller`, we should delete the extension network controller and replace it with `@metamask/network-controller`.

This issue is the final step of the "[Merge the two network controllers](https://github.com/metamask/core/issues/753)" epic.

mcmire:
## Explanation

This commit fulfills a **long-standing** desire to get the extension using the same network controller as mobile by removing NetworkController from this repo and replacing it with NetworkController from the `@metamask/network-controller` package.

The new version of NetworkController is different the old one in a few ways:

- The new controller inherits from BaseControllerV2, so the `state` property is used to access the state instead of `store.getState()`. All references of the latter have been replaced with the former.
- As the new controller no longer has a `store` property, it cannot be subscribed to; the controller takes a messenger which can be subscribed to instead. There were various places within  MetamaskController where the old way of subscribing has been replaced with the new way. In addition, DetectTokensController has been updated to take a messenger object so that it can listen for NetworkController state changes.
- The state of the new controller is not updatable from the outside. This affected BackupController, which dumps state from NetworkController (among other controllers), but also loads the same state into NetworkController on import. A method `loadBackup` has been added to NetworkController to facilitate this use case, and BackupController is now using this method instead of attempting to call `update` on NetworkController.
- The new controller does not have a `getCurrentChainId` method; instead, the chain ID can be read from the provider config in state. This affected MmiController. (MmiController was also updated to read custom networks from the new network controller instead of the preferences controller).
- ~The default network that the new controller is set to is always Mainnet (previously it could be either localhost or Goerli in test mode, depending on environment variables).~ Technically this is true from the `core` perspective, but I've brought back the old logic so there should be no functional difference on this side of things.

Fixes #17931.

## Manual Testing Steps

The new NetworkController should be functionally equivalent to the old one, so there should be no user-facing changes. That said, it would probably worth testing anything that involves the network (which, I know, is basically everything) to ensure that it works as before.

## Pre-merge author checklist

- [x] I've clearly explained:


mcmire:
Blocked by https://github.com/MetaMask/core/pull/1421

app/scripts/controllers/detect-tokens.js
```diff
@@ -208,8 +218,8 @@ export default class DetectTokensController {
     this.interval = DEFAULT_INTERVAL;
   }
 
-  getChainIdFromNetworkStore(network) {
-    return network?.store.getState().providerConfig.chainId;
+  getChainIdFromNetworkStore() {
```
mcmire, line 221: The argument for this method was removed because the thing it is trying to access is already accessible as a property.

app/scripts/controllers/mmi-controller.js
```diff
@@ -549,22 +549,25 @@ export default class MMIController extends EventEmitter {
...
-      const network = this.preferencesController
-        .getFrequentRpcListDetail()
-        .find((item) => parseInt(item.chainId, 16) === chainId);
-      this.networkController.setRpcTarget(
-        network.rpcUrl,
-        network.chainId,
-        network.ticker,
-        network.nickname,
-      );
+      const foundNetworkConfiguration = Object.values(
```
mcmire, line 558: I'm not sure why this didn't cause issues before, but we must have missed this when migrating the custom networks list to NetworkController.

app/scripts/metamask-controller.actions.test.js
```diff
@@ -96,6 +97,17 @@ describe('MetaMaskController', function () {
         ]),
       );
     metamaskController = new MetaMaskController({
+      initState: {
```
mcmire, line 100: The default network for NetworkController is now Mainnet, but these tests assumed that it is localhost.

app/scripts/metamask-controller.js
```diff
@@ -1311,19 +1310,11 @@ export default class MetamaskController extends EventEmitter {
...
-              providerConfig: {
-                ...networkState.providerConfig,
-              },
-            };
-            return cb(modifiedNetworkState);
-          });
-        },
-        getNetwork: () =>
-          this.networkController.store.getState().networkId ?? 'loading',
+        onNetworkStateChange: networkControllerMessenger.subscribe.bind(
```
mcmire, line 1313: It seemed that previously we were copying the state but making no other changes to the state. I've replaced this with a direct bind to the `stateChange` event.

mcmire:
`mockttp` is a development dependency, so the above warning is acceptable.

@SocketSecurity ignore @httptoolkit/subscriptions-transport-ws@0.9.19

app/scripts/metamask-controller.actions.test.js
Gudahtt: We can make the default network whatever we want, since we set the initial state. Could we make this behave the same as before?

app/scripts/controllers/mmi-controller.js
Gudahtt: Looks like this was introduced after that change, but was in PR for a long time. Probably a merge conflict.

app/scripts/metamask-controller.js
Gudahtt: Looks like this was originally done to work around the need to rename `provider` to `providerConfig` (see https://github.com/MetaMask/metamask-extension/commit/e0ff039078e7ffd54c464aa856de13b41d30ea9a), but I must have missed that it was no longer necessary after that was finally renamed in https://github.com/MetaMask/metamask-extension/commit/d1cea85f3360801e9753d5ef126fd49797cd1ffe#diff-6fbff2cfe97ac01b77296ef2122c7e0a5b3ff6a84b584b4d1a87482f35eea3d6

app/scripts/controllers/mmi-controller.js
mcmire: Ah, that makes sense.

app/scripts/metamask-controller.js
```diff
@@ -1039,29 +1044,24 @@ export default class MetamaskController extends EventEmitter {
...
-        this.networkController.store.subscribe((state) => {
-          if (previousNetworkId !== state.networkId) {
-            listener();
-            previousNetworkId = state.networkId;
-          }
-        });
+        networkControllerMessenger.subscribe(
+          'NetworkController:stateChange',
+          () => listener(),
+          ({ networkId }) => networkId,
```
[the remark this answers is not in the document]
mcmire: It appears that if you pass a selector function to `subscribe` it changes the behavior of `publish` such that the handler will only get called if the value that the selector function returns changes. Is that true? I'm seeing this line in ControllerMessenger: https://github.com/MetaMask/core/blob/633660f579894692a96350e1a94de8d14c24aef0/packages/base-controller/src/ControllerMessenger.ts#L465

test/e2e/metamask-ui.spec.js
```diff
@@ -37,6 +40,37 @@ describe('MetaMask', function () {
       'test-dapp',
       'dist',
     );
+    fixtureServer = new FixtureServer();
```
Gudahtt, line 43: Do we still need these changes, now that the initial controller state has been updated?

test/e2e/metamask-ui.spec.js
```diff
@@ -19,6 +19,7 @@ const dappPort = 8080;
 
 describe('MetaMask', function () {
   let driver;
+  let fixtureServer;
```
Gudahtt, line 22: Two more lines to remove

Gudahtt (approved):
LGTM!

test/e2e/metamask-ui.spec.js
mcmire: Addressed in [61d22a6](https://github.com/MetaMask/metamask-extension/pull/19486/commits/61d22a697528f6ccb84938e8bd78deef8082516a).

brad-decker (approved):
Code LGTM and works great locally!

PR #19487, opened 2023-06-07
fix(e2e): flaky test-e2e-chrome on add-custom-network.spec.js

HowardBraham:
## Explanation

The `test-e2e-chrome` on `add-custom-network.spec.js` has been very flaky recently, failing perhaps more than 50% of the time.

This PR should make it a bit less flaky.

### DIFF
--- a/test/e2e/tests/add-custom-network.spec.js
+++ b/test/e2e/tests/add-custom-network.spec.js
@@ -101,14 +101,14 @@ describe('Custom network', function () {
       },
     );
   });
-  it("don't add bad rpc custom network", async function () {
+
+  it('don’t add bad rpc custom network', async function () {
     await withFixtures(
       {
         dapp: true,
         fixtures: new FixtureBuilder()
           .withPermissionControllerConnectedToTestDapp()
           .build(),
-        ganacheOptions,
         title: this.test.title,
       },
       async ({ driver }) => {
@@ -140,16 +140,23 @@ describe('Custom network', function () {
           'MetaMask Notification',
           windowHandles,
         );
+
+        const errMsg1 = 'verify the network details';
+        await driver.findElement({
+          tag: 'a',
+          text: errMsg1,
+        });
+
         await driver.clickElement({
           tag: 'button',
           text: 'Approve',
         });
 
-        const errMsg =
+        const errMsg2 =
           'Chain ID returned by the custom network does not match the submitted chain ID.';
         await driver.findElement({
           tag: 'span',
-          text: errMsg,
+          text: errMsg2,
         });
 
         const approveBtn = await driver.findElement({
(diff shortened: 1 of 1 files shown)


test/e2e/tests/add-custom-network.spec.js
```diff
@@ -398,21 +405,13 @@ describe('Custom network', function () {
           tag: 'div',
         });
 
-        await driver.clickElement({
-          tag: 'button',
-          text: 'Delete',
-        });
+        // Click first Delete button
+        await driver.clickElement('button.btn-danger');
```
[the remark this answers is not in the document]
DDDDDanica: You are correct ~ i went back to check the implementation and it includes `findClickableElement` before clicking. I guess we could leave it like this for now, and keep monitoring to see if the button is still found and located ~ 🙏

DDDDDanica (approved):

DDDDDanica (approved):

NidhiKJha (approved):

test/e2e/tests/add-custom-network.spec.js
Gudahtt, line 409: Nit: This seems like a step in the wrong direction. The test is now more brittle because it depends upon the specific tag and classname used, which might change. It's also less readable (it's not clear this is a "Delete" button, but it was clear beforehand).

Generally I recommend using just a `text` selector if possible, with a `data-testid` selector as well if necessary. Text selectors are better for readability in that it makes the test more closely match what a user sees when going through this flow, and they're better for maintainability because the text won't change during a refactor. `data-testid` isn't as readable, but it's also nice for maintainability because it flags in the code where a test relies upon an element.

I'm curious what motivated the change - was there a second "Delete" button on the page somewhere? Or was this clicking "Delete" before the Aritrum page rendered (i.e. when the previous network details were shown)?

Gudahtt:
Thanks for looking into this! I noticed this test suite is still flaky though ([it failed again when this PR was merged](https://app.circleci.com/pipelines/github/MetaMask/metamask-extension/48095/workflows/963b87a1-6071-4ab3-bc4e-6aabc6d5f422/jobs/1373994))

test/e2e/tests/add-custom-network.spec.js
HowardBraham: Yes, there are two "Delete" buttons in a row, and I was worried about the driver getting them confused.

image image

HowardBraham:
> I noticed this test suite is still flaky though

I have a theory that this error message may be interfering sometimes.  Any suggestions?

image

test/e2e/tests/add-custom-network.spec.js
Gudahtt: Ahh I see. Hmm. I guess a data-testid could work to differentiate between them

HowardBraham:
> Interesting; so the state is setup with an invalid network

I looked into this further.  It's not set up with an invalid network.  It just sometimes displays that error message for half a second before it decides that actually it's correct.  Should we stop that error message from appearing on a network that's already added?

PR #19507, opened 2023-06-08
Reduce e2e test flakiness

Gudahtt:
## Explanation

Three e2e tests have been updated to fix a possible race condition causing intermittent e2e test failures.

In each of the updated tests, the test checks the current network. The check is performed as a two-step process: locate the current network element, then check the text to ensure it's correct.

This fails when the test driver finds the element before it re-renders. If the test runs too quickly, it compares the text before the switch is shown on screen, and the test fails.

Instead the tests use the element locator to describe what they want. This tells the test driver to keep looking until the conditions are met, ensuring the test doesn't fail unless the network switch takes longer than the default timeout (which should not happen).

This is a good example of why we should avoid using assertions on elements in e2e tests. Express your assertions as locators instead to make the test more resilient in the case where the test runs before the next render.

## Manual Testing Steps

Check that the affected e2e test suites pass

## Pre-merge author checklist

- [x] I've clearly explained:

### DIFF
--- a/test/e2e/tests/chain-interactions.spec.js
+++ b/test/e2e/tests/chain-interactions.spec.js
@@ -55,10 +55,10 @@ describe('Chain Interactions', function () {
         await driver.switchToWindow(extension);
 
         // verify networks
-        const networkDisplay = await driver.findElement(
-          '[data-testid="network-display"] p',
-        );
-        assert.equal(await networkDisplay.getText(), 'Localhost 8545');
+        await driver.findElement({
+          css: '[data-testid="network-display"]',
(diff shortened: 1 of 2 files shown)


mcmire (approved):
There are other jobs that occasionally fail (such as the `user-actions-benchmark` one failing now) but this certainly helps a ton.

Gudahtt:
Yeah this is nowhere near the most flaky; the CircleCI stats indicate that the one I started looking into here has only failed once in the past month. But, it's the one I hit yesterday.

DDDDDanica (approved):

PR #19511, opened 2023-06-08
Integrating ppom-validator with extension

jpuri:
The PR integrates PPOM-Validator in extension.


package.json
```diff
@@ -255,6 +255,7 @@
     "@metamask/permission-controller": "^3.2.0",
     "@metamask/phishing-controller": "^3.0.0",
     "@metamask/post-message-stream": "^6.0.0",
+    "@metamask/ppom-validator": "git@github.com:MetaMask/ppom-validator.git#aa3fcb9e7850e0d44bc00be8ac466cde1b3ee18b",
```
jpuri, line 258: Above will be replaced with a dependency from npm registry.

jpuri:
e2e are failing with this error coming from `@blockaid/ppom-mock`:

Screenshot 2023-06-16 at 3 54 09 PM

app/scripts/metamask-controller.js
```diff
@@ -3777,6 +3806,10 @@ export default class MetamaskController extends EventEmitter {
       engine.emit('notification', message),
     );
 
+    // BEGIN:ONLY_INCLUDE_IN(build-blockaid);
+    engine.push(createPPOMMiddleware(this.ppomController));
+    // END:ONLY_INCLUDE_IN(build-blockaid);
+
```
cryptotavares, line 3812: Is there a specific reason for this to be the first middleware that we are pushing? I am wondering if the ppom middleware should run after the `createDupeReqFilterMiddleware` (ensuring that in MV3 we are not processing repeated requests). Also not sure if the ones to append `origin`, `tabId` and the `createLoggerMiddleware` shouldn't be added before.
What do you think?

app/scripts/lib/indexed-db-backend.ts
```diff
@@ -0,0 +1,185 @@
...
+                );
+              };
+            })
+            .catch((error) => {
+              reject(error);
+            });
+        })
+        .catch((error) => {
+          reject(error);
+        });
```
cryptotavares, line 129: Instead of using Promise chainning we should use async/await wrapped in a Promise. IMO it improves the code readibility. Something like this:
```js
return new Promise<void>(async (resolve, reject) => {
  try {
    await validateChecksum(key, data, checksum);
    const objectuStore = await  this._getObjectStore('readwrite');
    const request = objectStore.put({ ...key, data });

    request.onsuccess = () => {
      resolve();
    };

    request.onerror = (event) => {
      reject(
        new Error(
          `Error writing data: ${(event.target as any)?.error}`,
        ),
      );
    };
  } catch (error) {
    reject(error);
  }
});
```
(the same pattern could be used on the other methods)

What do you think?

app/scripts/metamask-controller.js
jpuri: There is not specific reason for this to be first. It can go after all these middleware's. Let me update the PR.

jpuri: Done

app/scripts/lib/indexed-db-backend.ts
jpuri: @cryptotavares : this seems not valid eslint code. Please check https://eslint.org/docs/latest/rules/no-async-promise-executor

app/scripts/lib/indexed-db-backend.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  private dbVersion: number;
+
+  constructor(storeName: string, dbVersion: number) {
+    this.storeName = storeName;
+    this.dbVersion = dbVersion;
+  }
+
+  private async _getObjectStore(
+    mode: IDBTransactionMode,
```
segun, line 39: I'm a bit confused. Where does the IDBTransactionMode type come from, there are no imports on this file.

segun (dismissed):

segun:
Looks good, just need got fix the failing checks.

jpuri:
@metamaskbot update-policies

jpuri:
@metamaskbot update-policies

Gudahtt:
@metamaskbot update-policies

builds.yml
```diff
@@ -59,6 +59,14 @@ buildTypes:
     isPrerelease: true
     manifestOverrides: ./app/build-types/flask/manifest/
 
+  blockaid:
```
FrederikBolding, line 62: I think what you want is just a feature called `blockaid`, not a buildType

jpuri: We do not want this feature to be part of flask, only adding feature and not including it in buildType is breaking build

development/build/static.js
```diff
@@ -196,6 +196,11 @@ function getCopyTargets(shouldIncludeLockdown, shouldIncludeSnow) {
       dest: `runtime-lavamoat.js`,
       pattern: '',
     },
+    {
+      src: getPathInsideNodeModules('@blockaid/ppom-mock', 'dist/'),
```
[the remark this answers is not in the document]
jpuri: Good point, I made that change

app/scripts/lib/indexed-db-backend.ts
```diff
@@ -0,0 +1,185 @@
+type StorageKey = {
+  name: string;
+  chainId: string;
+};
+
+type StorageBackend = {
```
matthewwalsh0, line 6: Can we use the type defined in PPOM storage rather than duplicating?

I appreciate it couples it to Blockaid but that is where the interface exists and currently only imported in the blockaid build type.

app/scripts/lib/indexed-db-backend.ts
```diff
@@ -0,0 +1,185 @@
...
+  const hashString = Array.from(new Uint8Array(hash))
+    .map((b) => b.toString(16).padStart(2, '0'))
+    .join('');
+
+  if (hashString !== checksum) {
+    throw new Error(`Checksum mismatch for key ${key}`);
+  }
+};
+
+export class IndexedDBBackend implements StorageBackend {
```
matthewwalsh0, line 28: Should we be more explicit here and call this something like IndexexDBPPOMStorage as it exists to support that and is using its interface?

app/scripts/lib/indexed-db-backend.ts
```diff
@@ -0,0 +1,185 @@
...
+  private storeName: string;
+
+  private dbVersion: number;
+
+  constructor(storeName: string, dbVersion: number) {
+    this.storeName = storeName;
+    this.dbVersion = dbVersion;
+  }
+
+  private async _getObjectStore(
```
matthewwalsh0, line 38: If we're creating new Typescript files, should we stick to the `#` private syntax as that seems to be the standard we're pushing in core now also?

matthewwalsh0, line 38: Is it best practice to avoid using the `async` syntax if we're not using any `await` syntax as we're already explicitly returning a promise?

app/scripts/lib/indexed-db-backend.ts
```diff
@@ -0,0 +1,185 @@
...
+
+              request.onerror = (event) => {
+                reject(
+                  new Error(
+                    `Error writing data: ${(event.target as any)?.error}`,
+                  ),
+                );
+              };
+            })
+            .catch((error) => {
```
matthewwalsh0, line 123: Why are we nesting here? Should we not validate then get, then create the handlers? That way we only need a single catch for the entire chain?

Goldmouth81: Withdrawal transfer and deposit

app/scripts/lib/indexed-db-backend.ts
```diff
@@ -0,0 +1,185 @@
...
+    });
+  }
+
+  public async dir(): Promise<StorageKey[]> {
+    return new Promise<StorageKey[]>((resolve, reject) => {
+      this._getObjectStore('readwrite')
+        .then((objectStore) => {
+          const request = objectStore.getAllKeys();
+
+          request.onsuccess = (event) => {
```
matthewwalsh0, line 161: It seems all these methods create promises using the `onsuccess` and `onerror` methods.

Would it simplify this logic and avoid a lot of duplication if we create a private helper method such as `promisifyRequest` which converts any `request` to a `Promise` and then we can convert the majority of these methods to linear `async` code using `awaits`? We could resolve to the `event` or the `event.target.result` if each operation has one.

This method for example could become:

```
const objectStore = await this._getObjectStore('readwrite');
const result = await this._promisifyRequest(objectStore.getAllKeys());
return result.map( ([name, chainId]: string[]) => ({ name, chainId }) );
```

app/scripts/metamask-controller.js
```diff
@@ -618,6 +624,22 @@ export default class MetamaskController extends EventEmitter {
       this.phishingController.setStalelistRefreshInterval(30 * SECOND);
     }
 
+    ///: BEGIN:ONLY_INCLUDE_IN(blockaid)
+    this.ppomController = new PPOMController({
```
matthewwalsh0, line 628: As mentioned in the other PR, would an alternate option be to check the blockaid user preference here before creating this instance?

development/build/static.js
```diff
@@ -108,7 +110,7 @@ module.exports = function createStaticAssetTasks({
   }
 };
 
-function getCopyTargets(shouldIncludeLockdown, shouldIncludeSnow) {
+function getCopyTargets(shouldIncludeLockdown, shouldIncludeSnow, includePPOM) {
```
matthewwalsh0, line 113: Rather than adding a boolean for a single feature, would it be more scalable to pass the `activeFeatures` array here so we can add further conditions to these targets in future without more arguments?

jpuri: Done

app/scripts/metamask-controller.js
jpuri: We need to inject the controller instance in middleware engine, I do not think that is possible once engine initialisation is completed. That is reason I preferred this approach.

app/scripts/lib/indexed-db-backend.ts
jpuri: I updated the PR, please check.

app/scripts/lib/indexed-db-backend.ts
jpuri: This is now resolved.

app/scripts/metamask-controller.js
jpuri: Also, I would avoid writing blockaid specific code to enable/disable blockaid in MetamaskController.

app/scripts/metamask-controller.js
```diff
@@ -210,6 +213,9 @@ import {
 } from './controllers/permissions';
 import createRPCMethodTrackingMiddleware from './lib/createRPCMethodTrackingMiddleware';
 import { securityProviderCheck } from './lib/security-provider-helpers';
+///: BEGIN:ONLY_INCLUDE_IN(blockaid)
+import { IndexexDBPPOMStorage } from './lib/indexed-db-backend';
```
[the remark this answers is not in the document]
jpuri: Fixed

alokeutpal (approved):

legobeat (dismissed):
- [x] https://github.com/MetaMask/metamask-extension/pull/19511#discussion_r1251618461
- [ ] https://github.com/MetaMask/metamask-extension/pull/19511#discussion_r1253402388

package.json
```diff
@@ -251,6 +251,7 @@
     "@metamask/permission-controller": "^4.0.0",
     "@metamask/phishing-controller": "^3.0.0",
     "@metamask/post-message-stream": "^6.0.0",
+    "@metamask/ppom-validator": "^0.0.1",
```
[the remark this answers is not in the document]
alokeutpal: With 100% difficulty, metamax, the same thing happens every time I download the app, reset the wallet. I will not be able to see the token in the Metamax wallet while it is alive. All tokens are either blocked or locked. The whole system is designed to prevent people from working. Thank you, I am experiencing difficulties, so I wrote, do not take it otherwise. Simplify the system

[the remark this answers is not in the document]
alokeutpal: With 100% difficulty, metamax, the same thing happens every time I download the app, reset the wallet. I will not be able to see the token in the Metamax wallet while it is alive. All tokens are either blocked or locked. The whole system is designed to prevent people from working. Thank you, I am experiencing difficulties, so I wrote, do not take it otherwise. Simplify the system

alokeutpal (approved):

app/scripts/metamask-controller.js
matthewwalsh0: I appreciate it adds complexity but as we're talking about a user preference, I'd argue that's a client responsibility.

Plus any required client code would just be to detect the preference change and call suitable methods on the controller such as `disable`.

As for the middleware, a future option could be to create a generic `DisableController` which could be constructed with a callback to detect the controller is disabled, a list of method names, and the original controller instance.

It could then use a proxy to detect calls to the provided methods and either throw or pass to the original controller depending on if the feature was disabled. We'd then do all bindings to this wrapper instead of the internal feature controller.

matthewwalsh0 (approved):

package.json
```diff
@@ -462,6 +463,7 @@
     "eslint-plugin-react": "^7.23.1",
     "eslint-plugin-react-hooks": "^4.2.0",
     "eslint-plugin-storybook": "^0.6.12",
+    "fake-indexeddb": "^4.0.1",
```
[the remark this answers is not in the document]
jpuri: Hey @legobeat : this is only dev dependency, not sure but does that effects ?

package.json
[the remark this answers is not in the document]
jpuri: These are issues we are working on for PPOM. PPOM code is already code fenced thus as I understand these modules will not be part of final build.

builds.yml
jpuri: We have not introduced new build type in the PR, only a new feature

jpuri:
@legobeat : can you plz re-review.

app/scripts/metamask-controller.js
jpuri: `DisableController` is a good idea.

Let me make PreferenceController change in a following PR.

cryptotavares (approved):

app/scripts/lib/indexed-db-backend.test.ts
```diff
@@ -0,0 +1,93 @@
...
+        { name: 'fake_name', chainId: '5' },
+        enc.encode('fake_data'),
+        'XXX',
+      );
+    }).rejects.toThrow('Checksum mismatch');
+  });
+
+  it('should fail to read if checksum does not match', async () => {
+    const indexDBBackend = new IndexedDBPPOMStorage('PPOMDB', 1);
+    await expect(async () => {
```
digiwand, line 49: [nit] could merge this test with the test above. They have the same test phrase

PR #19536, opened 2023-06-08
Add a queue for dapps accessing different networks

BelfordZ:
## Explanation

Currently not working:

- if you call sendTx from dapp 1, then from dapp 2 you call requestPermissions, when you reject the sendTx, it ends up rejecting the requestPermissions too even though u havent seen it yet
- methods that open a confirmation dialog, but where the confirmation does not depend on selected network, should not queue requests from other chains (should just let them through since the confirmation is not chain-specific)

## Pre-merge author checklist

- [ ] I've clearly explained:

### DIFF
--- a/app/scripts/controllers/network/network-controller.ts
+++ b/app/scripts/controllers/network/network-controller.ts
@@ -1152,7 +1152,10 @@ export class NetworkController extends EventEmitter {
   getNetworkConfigurationForChainId(chainId: string) {
     console.log(this.store.getState().networkConfigurations);
     for (const networkConfigId in this.store.getState().networkConfigurations) {
-      if (this.store.getState().networkConfigurations[networkConfigId].chainId === chainId) {
+      if (
+        this.store.getState().networkConfigurations[networkConfigId].chainId ===
+        chainId
+      ) {
         return networkConfigId;
       }
     }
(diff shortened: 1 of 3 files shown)


app/scripts/controllers/network/network-controller.ts
```diff
@@ -1148,4 +1148,14 @@ export class NetworkController extends EventEmitter {
       networkConfigurations,
     });
   }
+
+  getNetworkConfigurationForChainId(chainId: string) {
```
BelfordZ, line 1152: should split this off into seperate PR, since it will need to go into core/network-controller as well.

app/scripts/controllers/selected-network-controller.ts
```diff
@@ -0,0 +1,156 @@
...
+const getDefaultState = () => ({
+  domains: {},
+});
+
+type Domain = string;
+type ChainId = string;
+type RequestQueue = Record<Domain, Promise<unknown>[]>;
+
+export type SelectedNetworkControllerState = {
+  domains: Record<Domain, ChainId>;
```
BelfordZ, line 47: Should make it an array of chainIds so that its future-proofed.

app/scripts/metamask-controller.js
```diff
@@ -3788,6 +3797,98 @@ export default class MetamaskController extends EventEmitter {
...
+          console.log('setting selected network', req.params[0].chainId);
+          this.selectedNetworkController.setChainForDomain(
+            origin,
+            req.params[0].chainId,
+          );
+          return;
+        }
+
+        const chainIdForRequest =
+          this.selectedNetworkController.getChainForDomain(req.origin);
```
adonesky1, line 3846: This shouldn't be required with the multichain network controller

PR #19581, opened 2023-06-13
[MMI] Added code fences in whats new popup

albertolive:
## Explanation

Adding the MMI portfolio notifications without disrupting much of the existing MM logic while at the same time having it prepared to add new MMI-specific notifications in the future. Added code fences to three existing files:
1. metamask-controller: We added our own notifications
2. whats-new-popup: We added new logic just for MMI
3. home-component: Added needed MMI prop

https://consensyssoftware.atlassian.net/browse/MMI-2503

## Pre-merge author checklist

- [X] I've clearly explained:


zone-live (dismissed):

brad-decker (changes requested):

ui/components/app/whats-new-popup/whats-new-popup.js
```diff
@@ -316,22 +382,25 @@ export default function WhatsNewPopup({ onClose }) {
...
+
+          ///: BEGIN:ONLY_INCLUDE_IN(build-mmi)
+          renderNotification = renderFirstNotification;
+          ///: END:ONLY_INCLUDE_IN
+
+          return renderNotification(
+            notification,
+            idRefMap,
+            history,
+            isLast,
```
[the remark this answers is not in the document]
albertolive: Fixed, thanks! 🙌

NidhiKJha (dismissed):

albertolive:
@metamaskbot update-policies

ui/components/app/whats-new-popup/whats-new-popup.js
```diff
@@ -321,22 +388,26 @@ export default function WhatsNewPopup({ onClose }) {
...
-                history,
-                isLast,
-              );
+          // Choose the appropriate rendering function based on the id
+          let renderNotification =
+            notificationRenderers[id] || renderSubsequentNotification;
+
+          ///: BEGIN:ONLY_INCLUDE_IN(build-mmi)
+          renderNotification = renderFirstNotification;
+          ///: END:ONLY_INCLUDE_IN
```
brad-decker, line 397: So... this is an interesting optimization, but I don't think it would pass typescript safety. renderFirstNotification and renderSubsequentNotification have different arguments, so even though the additional parameters for renderFirstNotification are on the end of the function parameters it would fail to compile in typescript due to extra params existing. In order to make this work you should probably change both renderFirst and renderSubsequent take an options bag ({ notification, idRefMap }) instead of positional parameters. When you add jsdoc types to them, you can define a new type using @typedef for the options and pass that type as the type for the first parameter to each function. Even if you don't reference the 4 additional properties on renderFirst in renderSubsequent it would still type check.

brad-decker (changes requested):

albertolive:
@metamaskbot update-policies

zone-live (approved):

ui/components/app/whats-new-popup/whats-new-popup.js
brad-decker: Yup!

brad-decker (approved):

PR #19634, opened 2023-06-16
Fix searchable item list height

dan437:
## Explanation
We want to use the max-height property to make the list flexible in case there is only 1 item (e.g. after filtering).


danjm (approved):

darkwing (approved):
Good catch!

davibroc (approved):

Issue #19491, opened 2023-06-07
[Flaky e2e test]: Send eth


DDDDDanica:
**Occurring in pipelines:**
`send-eth`:
https://app.circleci.com/pipelines/github/MetaMask/metamask-extension/47456/workflows/35fc96c2-5680-4128-889d-bd6e5f770774/jobs/1353477
https://app.circleci.com/pipelines/github/MetaMask/metamask-extension/48456/workflows/2efaf1f5-93a8-49af-a56d-383e50e6bee1/jobs/1386320
https://app.circleci.com/pipelines/github/MetaMask/metamask-extension/48456/workflows/2efaf1f5-93a8-49af-a56d-383e50e6bee1/jobs/1386320

**Investigation & proposed solution:**

Issue #19625, opened 2023-06-15
[Flaky e2e test]: Add account


DDDDDanica:
**Error message:**
```
<testcase classname="Add account" name="It should be possible to remove an account imported with a private key, but should not be possible to remove an account generated from the SRP imported in onboarding" time="6.601"><failure>Expected values to be strictly equal:
+ actual - expected

+ &#x27;Account 2&#x27;
- &#x27;2nd account&#x27;

      + expected - actual

      -Account 2
      +2nd account
```

```
<testcase classname="Add account" name="should not affect public address when using secret recovery phrase to recover account with non-zero balance" time="9.396"><failure>Expected values to be strictly equal:
+ actual - expected

+ &#x27;0x0Cc...afD3&#x27;
- &#x27;0x3ED...CC59&#x27;
     ^

      + expected - actual

      -0x0Cc...afD3
      +0x3ED...CC59
```

**Occurring in pipelines:**
MV2: 
https://app.circleci.com/pipelines/github/MetaMask/metamask-extension/48678/workflows/e355e2b1-b25f-4179-9ccf-d00f7ec65099/jobs/1392975/steps
MV3:
https://app.circleci.com/pipelines/github/MetaMask/metamask-extension/48845/workflows/6ca3ebbd-ac6c-43c4-b5b6-5d0ddc0377e4/jobs/1398661
https://app.circleci.com/pipelines/github/MetaMask/metamask-extension/48845/workflows/6ca3ebbd-ac6c-43c4-b5b6-5d0ddc0377e4/jobs/1399397

**Investigation & proposed solution:**
improvement for selector:
```
        const secondAccountCreated = await driver.findElement(
          '[data-testid="account-menu-icon"]',
        );
        assert.equal(await secondAccountCreated.getText(), '2nd account');
```

PR #19645, opened 2023-06-16
feature(19625): eliminate e2e flakyness for add-account and send-eth

DDDDDanica:
Fixes:
https://github.com/MetaMask/metamask-extension/issues/19491
https://github.com/MetaMask/metamask-extension/issues/19625

## Pre-merge author checklist

- [ ] I've clearly explained:

### DIFF
--- a/test/e2e/tests/send-eth.spec.js
+++ b/test/e2e/tests/send-eth.spec.js
@@ -6,7 +6,6 @@ const {
   openDapp,
   assertAccountBalanceForDOM,
   logInWithBalanceValidation,
-  regularDelayMs,
 } = require('../helpers');
 const FixtureBuilder = require('../fixture-builder');
 
@@ -426,9 +425,7 @@ describe('Send ETH from inside MetaMask to a Multisig Address', function () {
         await inputAmount.fill('1');
 
         // Continue to next screen
-        await driver.delay(regularDelayMs); // wait for Next button is enabled
         await driver.clickElement({ text: 'Next', tag: 'button' });
-        await driver.delay(regularDelayMs); // wait for Next button is enabled
         await driver.clickElement({ text: 'Confirm', tag: 'button' });
 
         // Go back to home screen to check txn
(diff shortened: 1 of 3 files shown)


test/e2e/tests/send-eth.spec.js
```diff
@@ -426,10 +426,9 @@ describe('Send ETH from inside MetaMask to a Multisig Address', function () {
         await inputAmount.fill('1');
 
         // Continue to next screen
-        await driver.findClickableElement({ text: 'Next', tag: 'button' });
+        await driver.delay(regularDelayMs); // wait for Next button is enabled
```
danjm, line 429: Why is this `delay` needed? Was `findClickableElement` timing out?

danjm (dismissed):

test/e2e/tests/send-eth.spec.js
DDDDDanica: In terms of removing old `findClickableElement, is because in `await driver.clickElement({ text: 'Next', tag: 'button' });` we have `findClickableElement` as part of that function. 

And the reason for adding extra delay here, is we have 10s time out for `clickElement`, and sometimes it was not enough for button to be enabled. One thing we can do is to pass a new timeout inside `clickElement` to override original `this.timeout`.

```
  async clickElement(rawLocator) {
    const locator = this.buildLocator(rawLocator);
    const element = await this.findClickableElement(locator); 
    await element.click();
  }
```

Gudahtt: These CI machines aren't _that_ slow; 10  seconds should be ample time for the form validation to run. Something doesn't seem right here.

DDDDDanica: Makes sense, Gonna remove the delays completely and run a few times to see if this is the real problem 🙏

Gudahtt (approved):
LGTM!

pedronfigueiredo (approved):

PR #19506, opened 2023-06-08
Fix `testDev` builds with LavaMoat enabled

Gudahtt:
## Explanation

The build script compiles the LavaMoat runtime during each root task, but it was not recognizing `testDev` as a root task. As a result, all `testDev` builds were broken unless LavaMoat was disabled.

The list of root tasks has been updated to include `testDev`. It should now be complete.

## Manual Testing Steps

* Delete the `dist` directory (if it exists)
* Run `yarn start:test`
* Run any e2e test (e.g.  `yarn test:e2e:single --browser firefox ./test/e2e/tests/account-details.spec.js`) to see that e2e tests run successfully

## Pre-merge author checklist

- [x] I've clearly explained:
  - [x] What problem this PR is solving
  - [x] How this problem was solved
  - [x] How reviewers can test my changes
- [x] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [ ] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
- [ ] PR is linked to the appropriate GitHub issue
- [ ] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.

In this case, a QA Engineer approval will be be required.

PR #19293, opened 2023-05-25
Implement tests for multiple service worker restarts on the mv3 build

pedronfigueiredo:
## Explanation

- The simple send ETH flow should work as usual after multiple service worker restarts
- DApp interactions for signatures should continue to work after multiple service worker restarts 
- DApp interactions for ETH transfers should continue to work after multiple service worker restarts 
- Wallet should lock after the service worker restarts

* Fixes #16689

## Manual Testing Steps

At the moment, e2e tests with the mv3 build are not running. As such, these tests must be run manually with the following commands:

```
yarn
yarn build:mv3
IN_TEST=true yarn test:e2e:single test/e2e/mv3/multiple-restarts.spec.js --browser=chrome --leave-running
```

## Pre-merge author checklist

- [x] I've clearly explained:
  - [x] What problem this PR is solving
  - [x] How this problem was solved
  - [x] How reviewers can test my changes
- [x] Sufficient automated test coverage has been added

## Pre-merge reviewer checklist

- [ ] Manual testing (e.g. pull and build branch, run in browser, test code being changed)
- [ ] PR is linked to the appropriate GitHub issue
- [ ] **IF** this PR fixes a bug in the release milestone, add this PR to the release milestone

If further QA is required (e.g. new feature, complex testing steps, large refactor), add the `Extension QA Board` label.

In this case, a QA Engineer approval will be be required.

PR #19648, opened 2023-06-16
Update Chromedriver to v114

mcmire:
## Explanation

This avoids version mismatch errors when running e2e tests with Chrome locally.

Note that in order to make this change, we have to instruct LavaMoat to ignore two more globals which are used internally by Chromedriver:

- `Proxy`
- `ret_nodes`

The second global definitely seems like something specific to Chromedriver, but the first one — Proxy — is part of EcmaScript. Always ignoring this seems to go against the spirit of LavaMoat. To address this, then, this commit not only upgrades Chromedriver but also adds a guard around these two exceptions — along with other Chromedriver-specific LavaMoat exceptions we've added in the past — so that they only apply for test builds. This ensure that in the production version of MetaMask, access to Proxy is properly sandboxed.

Inspired by https://github.com/MetaMask/metamask-extension/pull/19180.

## Manual Testing Steps

You should be able to run all of the Chrome e2e tests locally. In addition, CI should pass.

## Pre-merge author checklist

- [x] I've clearly explained:


DDDDDanica (dismissed):

development/build/index.js
```diff
@@ -82,34 +82,38 @@ async function defineAndRunBuildTasks() {
...
+      // globals sentry needs to function
+      '__SENTRY__',
+      'appState',
+      'extra',
+      'stateHooks',
+      'sentryHooks',
+      'sentry',
+    ];
+
+    if (entryTask === 'test') {
```
Gudahtt, line 113: There are two entry tasks for test builds: "test" and "testDev".

We also have constants for each of them. I have another PR open now where I've replaced the other entry task references in this file with constants: https://github.com/MetaMask/metamask-extension/pull/19506

Gudahtt:
Creating a separate list of scuttling exceptions for test builds is a great idea. However, the current list of supposedly Chromedriver-related scuttling exceptions is a bit suspect. If there was an entry there relied upon in a production build, it wouldn't be caught by CI. As such we should probably do some manual testing on the production build before reducing that list.

development/build/index.js
mcmire: Ah, okay. I was going off the `shouldScuttle` array above. Your changes make sense. I'll wait until that PR is merged and then rebase this.

mcmire:
@Gudahtt Right, fair enough. I can do that.

mcmire:
@Gudahtt Ah sorry I misread your comment. It sounds like reducing the list can be something we do later. For now I've only added the `if` statement around `Proxy` and `ret_nodes`. Hopefully that addresses your concerns.

Gudahtt (dismissed):
LGTM!

brad-decker (dismissed):

development/build/index.js
```diff
@@ -89,52 +148,7 @@ async function defineAndRunBuildTasks() {
...
-        'Number',
-        'Array',
-        'crypto',
-        'Function',
-        'Uint8Array',
-        'String',
-        'Promise',
-        'JSON',
-        'Date',
-        'Proxy',
```
mcmire, line 129: It appears that `Proxy` may have slipped into https://github.com/MetaMask/metamask-extension/pull/19293 by accident, so I moved it back out.

mcmire:
^ Fixed conflicts and re-pushed.

Gudahtt (dismissed):

mcmire:
Sorry one more approval if you please ^

Gudahtt (approved):

danjm (approved):

PeterYinusa (approved):

PR #19653, opened 2023-06-18
Fix autolock field to accept decimals in Firefox

mcmire:
## Explanation

The autolock field on the Settings screen — the field that allows users to set the duration that MetaMask will wait for until automatically locking — does not always accept decimal numbers. This breaks the e2e test for this feature as it attempts to set this field to "0.1".

More specifically, the React component responsible for this field passes whatever the user inputs through the `Number` function immediately and then uses this to repopulate the input. Therefore, if the user enters "3" followed by a ".", `Number("3.")` will be called. This evaluates to the number 3, and "3" becomes the new value of the field. As a result, the "." can never be typed.

Curiously, this behavior only happens in Firefox; Chrome seems to keep the "." in the input field when it's typed. This happens because `onChange` event doesn't seem to get fired until a number is typed *after* the ".". This may be due to underlying differences in the DOM between Chrome and Firefox.

Regardless, always passing the input through `Number` creates other odd behavior, such as the fact that the input can never be cleared (because `Number("")` evaluates to 0).

This commit solves these problems by saving the "raw" version of the user's input as well as the normalized version. The raw version is always used to populate the input, whereas the normalized version is saved in state.

#### Chrome

It's hard to see, but 1) pressing Backspace when it says "0" does nothing and 2) adding numbers after the decimal does nothing.

https://github.com/MetaMask/metamask-extension/assets/7371/6612b8b0-13e7-41dc-937c-7c6aeff7dc65

#### Firefox

It's hard to see, but 1) pressing Backspace when it says "0" does nothing, 2) adding decimals or anything after the decimal resets the field to "0".

https://github.com/MetaMask/metamask-extension/assets/7371/4dd3a5fe-5557-4570-89f9-057df9d6d2cf

#### Chrome

https://github.com/MetaMask/metamask-extension/assets/7371/c9ac9b7c-5521-465e-87ee-fa4947dd4ef7

#### Firefox

https://github.com/MetaMask/metamask-extension/assets/7371/11292675-939f-4c3d-b9ce-c01635371eda

## Manual Testing Steps

Fundamentally you should ensure that if you go into Settings > Advanced, update the autolock setting to 1, save, then wait 1 minute, MetaMask should automatically lock. You should also see an error (and should not be able to save) if you enter a number past 10080.

With this change, however, you should also be able to enter 0.1, and MetaMask should autolock in 5 seconds.

Also, if you empty the field, it should default to 5 minutes (this has always been the placeholder value for this input, it's just that it never got used).

All of this behavior should be the same in both Chrome and Firefox.

Past this, all other settings on the page should work as intended.

## Pre-merge author checklist

- [x] I've clearly explained:


ui/components/ui/text-field/text-field.component.js
```diff
@@ -241,6 +241,8 @@ const TextField = ({
     inputProps.InputProps.inputProps['data-testid'] = dataTestId;
   }
 
+  // inputProps.InputProps.inputProps.step = '0.1';
+
```
Gudahtt, line 245: Nit: leftover comment

```suggestion
```

ui/pages/settings/advanced-tab/advanced-tab.component.js
```diff
@@ -601,21 +601,41 @@ export default class AdvancedTab extends PureComponent {
...
+  handleLockChange(autoLockTimeLimitBeforeNormalization) {
     const { t } = this.context;
-    const autoLockTimeLimit = Math.max(Number(time), 0);
 
-    this.setState(() => {
-      let lockTimeError = '';
+    if (autoLockTimeLimitBeforeNormalization === '') {
+      this.setState({
+        autoLockTimeLimitBeforeNormalization,
+        autoLockTimeLimit: '5',
```
Gudahtt, line 610: I guess this is the default value. Was this in effect before as a default?

mcmire: It has already sort of been the default; I noticed that the `placeholder` attribute of the input is set to "5". However, we never effectively used the default because it has been impossible for users to clear the field (it always appear as "0"). I suppose to be consistent with the current behavior we could change the placeholder to "0" instead of doing this.

Gudahtt: Leaving the default as zero makes sense, especially if that's already how it works in practice today.

Changing the default to 5 might result in this setting remaining on when users try to disable it. e.g. if I clear the field, I probably don't expect that to result in "5" being entered.

mcmire: Edit: [It looks 0 is the default according to `advanced-tab.container.js`.](https://github.com/MetaMask/metamask-extension/blob/develop/ui/pages/settings/advanced-tab/advanced-tab.container.js#L40) Following this and the current behavior I've changed the placeholder to 0 and the default value to 0 here: [511b2f2](https://github.com/MetaMask/metamask-extension/pull/19653/commits/511b2f25b815c5de1e7eb0a4c4a1628943f4b3cf)

ui/pages/settings/advanced-tab/advanced-tab.component.js
```diff
@@ -601,21 +601,41 @@ export default class AdvancedTab extends PureComponent {
...
+  handleLockChange(autoLockTimeLimitBeforeNormalization) {
     const { t } = this.context;
-    const autoLockTimeLimit = Math.max(Number(time), 0);
 
-    this.setState(() => {
-      let lockTimeError = '';
+    if (autoLockTimeLimitBeforeNormalization === '') {
+      this.setState({
+        autoLockTimeLimitBeforeNormalization,
+        autoLockTimeLimit: '0',
```
Gudahtt, line 610: Nit: The default auto lock time limit is now set in 3 different places (the parent component, here, and the placeholder). Perhaps we could introduce a constant to ensure they stay in sync?

Gudahtt (dismissed):
LGTM!

ui/pages/settings/advanced-tab/advanced-tab.component.js
mcmire: Hmm you're right. A constant would be good. Since you've approved I'll wait to see if any other changes are necessary and either add that here or in a different PR.

mcmire: Okay, references consolidated here: [2069663](https://github.com/MetaMask/metamask-extension/pull/19653/commits/20696634cf5de72b2674f6d0388453441e23051a)

ui/components/ui/text-field/text-field.component.js
mcmire: Good catch — removed in [c457767](https://github.com/MetaMask/metamask-extension/pull/19653/commits/c457767c0c8c2a00dfb1b099d2d42813b3d440e3).

Gudahtt (approved):
LGTM!

brad-decker (approved):

adonesky1 (approved):

